PR #3533, opened 2018-04-12
PDF renderer updates

christianmemije:
### Summary

The first commit changes:
- Adds linear loader. 
- Set's progress to null and 1 appropriately, 
- Limit zoom out and zoom in, mainly to prevent a negative scale value.
- Make short display default view.
- On scale changes, all pages are scaled.
- Account for the page padding.
- Update pdfjs

More
-  Adds a virtual scroller to improve performance and reduce memory usage.
- Moves fullscreen button to the right side.
- Save pdf location in local storage. 

Issues
* There is a miscalculation of the position that happens on scale.
* Assumes all pages are the same height. That was already the case?

Before
before
before2

After
after

after2

localhost_8000_learn_ ipad

### References

----

### Contributor Checklist

### Reviewer Checklist


christianmemije:
Not ready for review.

kolibri/core/assets/src/mixins/contentRenderer.js
```diff
@@ -97,5 +97,9 @@ export default {
       type: Boolean,
       default: false,
     },
+    containerStyle: {
+      type: Object,
```
rtibbles: Hrm, could you add a validator for this? Have tried to add validators for any object prop type in here.

kolibri/plugins/coach/assets/src/views/lessons/LessonContentPreviewPage/ContentArea.vue
```diff
@@ -21,6 +21,11 @@
       :available="content.available"
       :extraFields="content.extra_fields"
       :interactive="false"
+      :containerStyle="{
+        'height': 'calc(100vh - 350px)',
+        'max-height': '100vh',
```
rtibbles: `vh` is not supported on Android browser before 4.4, so I believe we are trying to avoid this?

kolibri/plugins/learn/assets/src/views/content-page.vue
```diff
@@ -192,6 +193,18 @@
       licenceDescriptionIsVisible: false,
     }),
     computed: {
+      containerStyle() {
+        if (
+          this.content.kind === ContentNodeKinds.DOCUMENT ||
```
rtibbles: I'm a little bit iffy about this - why can't this be delegated to the styling of the outermost div of the individual content renderers? Hard coding it by kind here seems to be making assumptions about the content renderer implementations.

Is this particular behaviour needed for the full screen functionality? If so, should we have some special 'full screen mixin' that enables this in a more systematic way?

kolibri/plugins/learn/assets/src/views/content-page.vue
```diff
@@ -9,22 +9,23 @@
...
-      :id="content.id"
-      :kind="content.kind"
-      :files="content.files"
-      :contentId="content.content_id"
-      :channelId="channelId"
-      :available="content.available"
-      :extraFields="content.extra_fields"
-      :initSession="initSession"
-    >
+    <template v-if="!content.assessment">
```
rtibbles: What issue is this solving? Might be good to add a comment in the template so the reason for this is documented.

christianmemije: I removed the slot from the content renderer because It makes sizing the content-renderer harder. For example setting a max-height on the content-renderer with scroll set to overflow will cause the button to also scroll. I think the button should be separate from the content-renderer.

kolibri/plugins/learn/assets/src/views/content-page.vue
christianmemije: The main reason for this was that if we add style to the content-renderer, it applies it to top most div of the content-renderer, where in fact we might want to actually apply style to the actual `<component/>`. An alternative to this would be to use a deep css selector. So I would be abe to do something like 

`content-page.vue`
```css
.content-renderer
  >>>.content-renderer-component
    width: 300px
```

rtibbles:
Hrm, very performant on my machine (able to render a 500 page PDF with no issues).

Checking it out on one of the Instant Schools tablets, and it gets stuck on at the loading screen for a PDF of any size.

kolibri/plugins/learn/assets/src/views/content-page.vue
DXCanas: I agree this is a little iffy. This component shouldn't have to care about the content kind for styles.

We could elevate the `component` to the top of the `template`. Adding child components for the `not renderable` `not available` and `loading` states as options for the `:is` would provide all the same functionality with directly applicable styles.

kolibri/core/assets/src/mixins/contentRenderer.js
DXCanas: Not super sure about this approach. See https://github.com/learningequality/kolibri/pull/3533/files/f3b90c0da2901c46fbd01c8ed7317a32fd4f988b#r186909606

kolibri/plugins/learn/assets/src/views/content-page.vue
DXCanas: Since both options in the `if/else` use a button, why can't we recycle the same button across both? Wouldn't that let us `v-if` only the component?

rtibbles: Yes, this can be simplified, I think.

kolibri/core/assets/src/mixins/contentRenderer.js
christianmemije: @DXCanas I ended up reverting this.

PR #3535, opened 2018-04-12
Coach contents updates

jonboiser:
1. Adds coach-content label in some places missed the first time, including Learner Lessons, which only shows coach content for Learners when in a Lesson or Exam
1. Handles special case where coach-content label is displayed differently for a topic with 1 coach content resource
1. Make adjustments to content cards when title is very long + needs coach content indicator
1. Use ContentNode `in_lesson` filter in one spot
1. Make the coach-content resource counts conditional on whether importing or exporting in import/export flows
1. Show coach-content indicator at the Channel level for Coach reports
1. Put coach-content indicator on channel cards in import/export UI

Todo:


kolibri/content/serializers.py
```diff
@@ -40,12 +40,18 @@ def to_representation(self, instance):
             on_device_resources = descendants.exclude(kind=content_kinds.TOPIC).filter(available=True).count()
             on_device_file_size = local_files.filter(available=True).aggregate(Sum('file_size'))['file_size__sum'] or 0
 
+            num_coach_contents = ContentNodeGranularSerializer(
```
lyw07: I'm not sure if I'm getting it right, but I feel like calling the ContentNodeGranularSerializer to only get `num_coach_contents` results in calling other functions in ContentNodeGranularSerializer which we don't need? I'm thinking that it could run faster if we get a copy of the function `get_num_coach_contents` in the ChannelMetadataSerialize and call it directly from there?

jonboiser: Sure, I can try that.

lyw07: thanks! Other than that, the python code looks great!

jonboiser: @lyw07  Just added a patch that extracted `get_num_coach_contents` to a utility and removed the methods from the three serializers.

lyw07: looks good! @jonboiser thanks for the update

PR #3519, opened 2018-04-10
Set different tree_ids for different databases.

rtibbles:
### Summary
* Every content database only has one content node tree in it.
* As a consequence, every content database node has a tree_id of 1.
* This used to no be an issue, because we only used old version databases, and the import logic in Kolibri accounted for this.
* Unfortunately, some fool failed to include this in the 'naive import' logic, so when we upgraded the exported content databases on Studio, so that they now used the naive import logic, every imported content node tree got its tree_id imported from the content database, rather than being set to the next available one.
* This PR rectifies that heinous oversight, by adding a mapper for the 'naive import' case.

### Reviewer guidance
* Import two recently published channels.
* Inspect the database to confirm that the content nodes have *different* tree_id properties.

----

### Contributor Checklist

- [ ] Contributor has fully tested the PR manually
- [ ] PR has the correct target branch and milestone
- [ ] PR has 'needs review' or 'work-in-progress' label
- [ ] If PR is ready for review, a reviewer has been added. (Don't use 'Assignees')
- [ ] If there are any front-end changes, before/after screenshots are included
- [ ] If this is an important user-facing change, PR or related issue has a 'changelog' label

### Reviewer Checklist

- [ ] Automated test coverage is satisfactory
- [ ] Reviewer has fully tested the PR manually
- [ ] PR has been tested for [accessibility regressions](http://kolibri-dev.readthedocs.io/en/develop/manual_testing.html#accessibility-a11y-testing)
- [ ] External dependencies files were updated (`yarn` and `pip`)
- [ ] Documentation is updated
- [ ] Link to diff of internal dependency change is included
- [ ] CHANGELOG.rst is updated for high-level changes
- [ ] Contributor is in AUTHORS.rst

PR #3484, opened 2018-04-06
Hid back arrow in "<-- Select Content" when downloading channel

HansGam:
### Summary
See #2745 

apr-05-2018 23-15-35

…

### Reviewer guidance
Go to imports and download a large channel like KA (en). You should not be able to exit from within the page

…

### References
Fixed issue #2745 . 

…

----

### Contributor Checklist

- [x] Contributor has fully tested the PR manually
- [x] PR has the correct target branch and milestone
- [x] PR has 'needs review' or 'work-in-progress' label
- [x] If PR is ready for review, a reviewer has been added. (Don't use 'Assignees')
- [ ] If there are any front-end changes, before/after screenshots are included
- [ ] If this is an important user-facing change, PR or related issue has a 'changelog' label

### Reviewer Checklist

- [ ] Automated test coverage is satisfactory
- [ ] Reviewer has fully tested the PR manually
- [ ] PR has been tested for [accessibility regressions](http://kolibri-dev.readthedocs.io/en/develop/manual_testing.html#accessibility-a11y-testing)
- [ ] External dependencies files were updated (`yarn` and `pip`)
- [ ] Documentation is updated
- [ ] Link to diff of internal dependency change is included
- [ ] CHANGELOG.rst is updated for high-level changes
- [ ] Contributor is in AUTHORS.rst

PR #3538, opened 2018-04-13
Cancellable channel import

### LINKED ISSUE #2041: Make Content Metadata Import process happen within a transaction
## Summary

Following the merge of #1947, content metadata is imported into the main database, for now, this involves deleting all existing content metadata for a channel, which, if done during active classroom use would result in a disjointed user experience for active learners.

In order to avoid this, it would be better for the deletion and import to happen within a single transaction, so that it is committed all at once, can be cancelled/rolled back without issue, and just generally doesn't mess things about!

### LINKED ISSUE #2731: more shared-resource availability annotation tests?
### Observed behavior

we weren't able to reproduce, but @jonboiser and I may have observed something like this scenario:

* two channels (A and B) with some overlap (shared) resources are imported
* channel A is deleted
* when trying trying to export from channel B, the content doesn't show up. They were hidden in the UI because the `available` annotation was set to `false`

### Expected behavior

no resources or topics should be unavailable when until there are no more references

### User-facing consequences

potential edge case bugs

### Steps to reproduce

I wasn't able to reproduce

### Context

* Kolibri version: 0.7-beta

----

cc @rtibbles

### LINKED ISSUE #2745: import/export: hide back arrow link in '<- Select content'
### Observed behavior

When in this state, the link at the top doesn't do anything:

image

### Expected behavior

The link should be hidden

### User-facing consequences

Users will try to get out when loading metadata takes too long, and when clicking that link doesn't work they'll do something more drastic like pressing the browser 'back' button or refreshing. Both of these will have negative consequences.

### Steps to reproduce

* delete all your channels
* go to import a large channel like KA (en)
* try to get out

----

refs #2721, #2732

### LINKED ISSUE #3246: Content channel disappears after UPDATE
WHERE: on http://35.229.41.226/  (creds `devowner:admin123`) which runs Kolibri 0.7.2b2.dev+git-1-gdd27d31

### Observed behavior

Trying to UPDATE a channel makes it disappear 

### Expected behavior

Channel remains and contains updated content.

### User-facing consequences

Users cannot update content channels. Will have to delete then re-import.

### Errors and logs

Nothing interesting in logs:

```
INFO:kolibri.content.management.commands.importchannel:Downloading data for channel id 75e2205cb9c449c99a12e558520e5eb1
^M  0%|          | 0/55296 [00:00<?, ?it/s]^[[37mINFO     Channel 75e2205cb9c449c99a12e558520e5eb1 already exists in database, cleaning up ContentNodes^[[0m
INFO:kolibri.content.utils.channel_import:Channel 75e2205cb9c449c99a12e558520e5eb1 already exists in database, cleaning up ContentNodes
INFO:kolibri.content.utils.channel_import:Importing ContentTag data
INFO:kolibri.content.utils.channel_import:Importing ContentNode_has_prerequisite data
INFO:kolibri.content.utils.channel_import:Importing ContentNode_related data
INFO:kolibri.content.utils.channel_import:Importing ContentNode_tags data
INFO:kolibri.content.utils.channel_import:Importing ContentNode data
INFO:kolibri.content.utils.channel_import:Importing Language data
INFO:kolibri.content.utils.channel_import:Importing File data
INFO:kolibri.content.utils.channel_import:Importing LocalFile data
INFO:kolibri.content.utils.channel_import:Importing AssessmentMetaData data
INFO:kolibri.content.utils.channel_import:Importing ChannelMetadata data
INFO:kolibri.content.utils.annotation:Setting availability of File objects based on LocalFile availability
INFO:kolibri.content.utils.annotation:Setting availability of non-topic ContentNode objects based on File availability
INFO:kolibri.content.utils.annotation:Setting availability of ContentNode objects with children for 3 levels
INFO:kolibri.content.utils.annotation:Setting availability of ContentNode objects with children for level 3
INFO:kolibri.content.utils.annotation:Setting availability of ContentNode objects with children for level 2
INFO:kolibri.content.utils.annotation:Setting availability of ContentNode objects with children for level 1
```

### Steps to reproduce

initial state
1 initial state

select channel to import
2 select channel for import

prompted for update
3 prompted to update

after completing update channel is no longer there
4 after completing update channel is no longer there

rtibbles:
### Summary
* Resolves the availability after channel update bug that was fixed by #3519, but in a cleaner way.
* Uses cascade deletion in SQLAlchemy to delete previously existing content database rows related to a channel
* Does channel import updating and deletion within a transaction to allow rollback
* Implements cancellation in the channelimport command
* Changes the task api endpoint to make channelimport a cancellable job
* Updates the import export wizard to include a loading channel metadata interstitial state
* On navigation back from this state, attempt to cancel the channel import task
* On refresh, redirect to the manage content page
* ~Fix bug where we were fetching the list of channels twice on the manage content page~ Cherry picked to 0.9 and merged from there.
* Fix bug where we were reimporting channel metadata even if it was imported, just because we had not downloaded any content yet

### Reviewer guidance
This is ready for review, but I anticipate needing to make updates due to the following two issues:
I think I am cancelling the channel import task, but I am not certain.
When I cancel the task and navigate back to the same channel, I get an error - any help appreciated!

### References
Fixes #2041
Fixes #2745
Fixes #3246
Fixes #2731
Supersedes #3484 

----

### Contributor Checklist

### Reviewer Checklist


jtamiace:
@rtibbles I'm not sure whether it's actually being cancelled, although there are animations...? Here are screencasts of the user flows I tested:

**Scenario 1:**
Import from central server --> select available channel --> click back button during metadata download
Clicking back did not cancel the metadata download
https://drive.google.com/file/d/1Dss4Fk-EG1UCHnuFasGVy45VNE56l2o-/view?usp=sharing

**Scenario 2:**
Import from central server --> select available channel --> let metadata load --> view available content
Content did not load properly from an available channel:
https://drive.google.com/file/d/19riQsnHoMi4GIE4iM1T5cNGcuwhqROTO/view?usp=sharing
Khan Academy (en) did not load (terminal error @ 3:45)
https://drive.google.com/file/d/1sIV65QBG5W9SLEBk98tyNVGiGA4ISvDt/view?usp=sharing

**Scenario 3:**
Import from central server --> select available channel --> select available content --> begin import --> cancel import
Could not get to this point

jonboiser:
@rtibbles @jtamiace The cancel button might not be showing up, because the version of the progress bar that is shown has `cancellable = false`. I don't think it will automatically cancel if you press browser-back either, which might be something we might want.


https://github.com/learningequality/kolibri/blob/401f4e9bc78492f996c3dedbe08d262d44162064/kolibri/plugins/device_management/assets/src/views/select-content-page/index.vue#L8-L17

rtibbles:
I implemented the backward flow of the state machine to initiate the cancellation, and the API endpoint is being called, I am just not sure it is cancelling as I expect on the backend.

rtibbles:
Hrm, using both the back arrow in the interface, and the back button, I am seeing the importchannel task get cancelled. Not replicating the flows you are seeing.

Perhaps we can look at this together next week @jtamiace?

Also, on @jonboiser's point, is it helpful to have a 'cancel' button visible as we do during content import, or *only* have it triggered by the back behaviour? (my intuition is that we would not want to have the cancel button, as that would put us in an intermediate state that would only be resolved by pressing back).

jonboiser:
We could have pressing the cancel button have as a second side-effect, going backwards in the wizard:

1. GIVEN we are in the middle of an `importchannel` task
1. WHEN I click 'cancel' next to the progress bar
1. THEN the `importchannel` tasks is canceled
1. THEN I am redirected back to the list of channels on Studio or USB drive.

I think, when I tested it, I did not press the `<-` button, but just the browser's back button, which I do not think triggers the cancelation, since the `transitionWizard` action is not run in that case.

rtibbles:
The cancel task endpoint is being called both from pressing the back arrow, and using the back button (back button behaviour is ensured by this line: https://github.com/learningequality/kolibri/pull/3538/files#diff-b665a7b21a71fac18c77aca42751eebdR89 - where we initiate the backwards step of the wizard if it is currently on the loading metadata state).

Immediately after browser back:
image

rtibbles:
Will add the cancel button with back functionality.

rtibbles:
Now has cancel button also. I am pretty sure this is now functioning as intended.

kolibri/tasks/api.py
```diff
@@ -301,6 +302,9 @@ def canceltask(self, request):
         if 'task_id' not in request.data:
             raise serializers.ValidationError(
                 "The 'task_id' field is required.")
+        if type(request.data['task_id']) is not str:
```
jonboiser: On my local server, I'm falling into this branch because the type is `unicode` instead of `str`

```
ipdb> request
<rest_framework.request.Request object at 0x10924b690>
ipdb> request.data
{u'task_id': u'7261de84d9bd4685a0955bc2d03331bb'}
ipdb> request.data['task_id']
u'7261de84d9bd4685a0955bc2d03331bb'
ipdb> type(request.data['task_id'])
<type 'unicode'>
ipdb>
```

rtibbles: On python2.7? Maybe we don't have the right imports in this file.

rtibbles: I added this line because of a completely unparseable error I was getting when and object got passed in instead of a task id.

rtibbles: Have changed it to a py2/3 compatible test now, and added a test.

kolibri/content/utils/sqlalchemybridge.py
```diff
@@ -123,22 +125,43 @@ def prepare_bases():
...
     # Calling Base.prepare() means that Base now has SQLALchemy ORM classes corresponding to
     # every database table that we need
-    Base.prepare()
+
+    def _gen_relationship(base, direction, return_fn,
+                          attrname, local_cls, referred_cls, **kw):
+        if direction is interfaces.ONETOMANY:
+            kw['cascade'] = 'all, delete-orphan'
+        # Add cascade behaviour on deletion
+        return generate_relationship(base, direction, return_fn,
```
bonidukic, line 159: @rtibbles Just an observation: you wrapped line 154 even though one-lining it would be shorter than e.g. 133. I guess this might be a good case for https://github.com/learningequality/kolibri/issues/3609.

rtibbles: I think that was probably because I copied it initially from a code sample that was pep8 compliant.

kolibri/content/utils/channel_import.py
```diff
@@ -222,20 +239,72 @@ def table_import(self, model, row_mapper, table_mapper, unflushed_rows):
...
+        # as SQL Alchemy ORM does not support INSERT ... ON DUPLICATE KEY UPDATE style queries,
+        # as not available in SQLite, only MySQL as far as I can tell:
+        # http://hackthology.com/how-to-compile-mysqls-on-duplicate-key-update-in-sql-alchemy.html
+        RowEntry = self.destination.session.query(DestinationRecord).get(data[model._meta.pk.name])
+        if RowEntry:
+            for key, value in data.items():
+                setattr(RowEntry, key, value)
+        else:
+            RowEntry = DestinationRecord(**data)
+        self.destination.session.merge(RowEntry)
```
aronasorman, line 253: If we're running through each record, is this a potential source of slowdown?

kolibri/content/utils/channel_import.py
```diff
@@ -222,20 +239,72 @@ def table_import(self, model, row_mapper, table_mapper, unflushed_rows):
...
+
+    def check_and_delete_existing_channel(self):
```
aronasorman, line 255: Why do we need to delete a channel in import?

rtibbles: We only reimport when the channel version has changed now, and we can't guarantee that nodes that previously existed exist in the new version.

We could try to do a diff to do a more selective update, but this is simpler and produces the same result.

kolibri/content/utils/channel_import.py
rtibbles: Yes, there might be potential for optimisation here, but the mapping code needs to operate on the records in the current implementation. Possible we could encode that in direct sql queries.

PR #3379, opened 2018-03-16
Intern: Build Tool Decoupling CLI

akshaymahajans:
### Summary

  I replaced the webpack_json management command with a command line tool, build_tools/webpack_json.py, that allows you to initialize and build specified kolibri plugins, rather than building every single plugin (as the management command did).

image

…

### Reviewer guidance

 * A reviewer can attempt to run the command " python webpack_json.py" with either of the following options:
1. -i option = enter an input file with a list of kolibri plugins you'd like to build.
2. -b option = enter a python list of plugins, or a space-separated list of plugins to build.

…

### References

 -Related to the management command webpack_json.py

…

----

### Contributor Checklist

- [- ] Contributor has fully tested the PR manually
- [- ] PR has the correct target branch and milestone
- [ ] PR has 'needs review' or 'work-in-progress' label
- [ ] If PR is ready for review, a reviewer has been added. (Don't use 'Assignees')
- [ ] If there are any front-end changes, before/after screenshots are included
- [ ] If this is an important user-facing change, PR or related issue has a 'changelog' label

### Reviewer Checklist

- [ ] Automated test coverage is satisfactory
- [ ] Reviewer has fully tested the PR manually
- [ ] PR has been tested for [accessibility regressions](http://kolibri-dev.readthedocs.io/en/develop/manual_testing.html#accessibility-a11y-testing)
- [ ] External dependencies files were updated (`yarn` and `pip`)
- [ ] Documentation is updated
- [ ] Link to diff of internal dependency change is included
- [ ] CHANGELOG.rst is updated for high-level changes
- [ ] Contributor is in AUTHORS.rst

Issue #950, opened 2017-02-18
Allow plugins to be built standalone


aronasorman:
## Summary

Right now, looks like we need to clone both a Kolibri repository and the specific plugin we need to build, and then run the devserver to have it serve the assets. It's hard on a devops and CI perspective to start building the plugins:

- Which version of Kolibri do I need to clone to get a plugin building? We'll like to pin a plugin to a version of Kolibri in the long term.
- I'm basically just replicating the Kolibri build pipeline, but with one small change. This consumes both devops and computing resources.
- What specific steps should I take to get a plugin building and releasable? Right now, the plugin build process is very dev-focused. Hence I'm having a hard time to figure out what steps I need to do on a continuous integration perspective. `yarn install && yarn run build`, `pip install -e && make assets` all are no dice. Running `make dist` inside the plugin itself doesn't work too.

## Possible solutions

The goal is to have `make dist` produce a working `whl` file that's ready to be uploaded to PyPI.

1. When `make dist` is run, clone Kolibri on Git based on a certain tag, and then run `yarn run build` or some other magic using the cloned Kolibri dir. `make clean` should remove said directory.
1. Similar to the one above, but only fetch the exact files that are needed to build. So maybe only the vue files and the frontend build dependencies?
1. Make the plugin fully standalone (has all the dependencies needed to build its assets). Not sure what this would involve, but is the cleanest solution.

rtibbles:
1 would be easiest to transition to.

2 would require internal knowledge about what is going on inside Kolibri, so would be fragile to changes in Kolibri's structure and API - another alternative would be for Kolibri versions to publish a versioned npm package that plugins could then depend on for build related information. Again, this seems brittle and just a mediated form of 1.

3 This would be ideal, however, all Kolibri frontend modules have a direct dependency on Kolibri, and I can't see how we would break this dependency (unless we published some sort of external npm module as in 2).

indirectlylit:
Another long-term possibility: decouple the build front-end system from the Kolibri application. Make the build system a stand-alone set of utilities which, when run, gets pointed at a particular version of Kolibri and some set of plugins.

Shorter-term, I would be very hesitant to advocate this because so many pieces of the infrastructure are still in flux, so adding yet another repo could introduce a lot of friction.

indirectlylit:
> We'll like to pin a plugin to a version of Kolibri in the long term.

[...]

PR #3542, opened 2018-04-16
JS Build Decoupling from Django stack

rtibbles:
### Summary
Supersedes #3379 

from that PR:

  I replaced the webpack_json management command with a command line tool, build_tools/webpack_json.py, that allows you to initialize and build specified kolibri plugins, rather than building every single plugin (as the management command did).

image

…

### Reviewer guidance

 * A reviewer can attempt to run the command " python webpack_json.py" with either of the following options:
1. -plugin_file option = enter an input file with a list of kolibri plugins you'd like to build.
2. -plugins option = enter a python list of plugins, or a space-separated list of plugins to build.

### References
First step to addressing #950

----

### Contributor Checklist

### Reviewer Checklist


build_tools/webpack_json.py
```diff
@@ -0,0 +1,67 @@
+import argparse
+import json
+import logging
+import os
+import sys
+
+from customize_build import load_plugins_from_file
+
+from kolibri.core.webpack.hooks import WebpackBundleHook
```
lyw07, line 9: According to the buildkite error, I think we might need to call `set_env()` before doing this import

rtibbles: Right, will check if there is a way to make this less of an issue.

rtibbles: Have followed the suggestion from `cli.py` to move set_env behaviour into `kolibri.__init__`, so this is now working, and we should always be able to use kolibri as an imported library without special setup!

build_tools/webpack_json.py
```diff
@@ -0,0 +1,63 @@
...
+        else:
+                logger.info('Successfully initialized {}'.format(module))
+
+def main():
+    logger = logging.getLogger(__name__)
+    logging.StreamHandler(sys.stdout)
+    logger.setLevel(logging.INFO)
+    parser = argparse.ArgumentParser()
+
+    parser.add_argument('--plugin_file', help='the filepath to which you\'d like to run plugins from', type=str, default=None)
```
lyw07, line 34: Sorry I'm not sure if I understand this correctly, and I don't really know about webpack. So my understanding is that we can now enable any plugins we want using `python webpack_json.py --plugin_file/--plugins`. Is this usage only for dev mode? Thank you!

rtibbles: Yes, although this command is currently invoked from inside the webpack build process. For now, this still uses the environment variable as the primary means for configuration of the list of build plugins, but these command arguments are here to allow more flexibility in the future.

If you want to test this out you can run `BUILD_TIME_PLUGINS=test.txt yarn run devserver ` where `test.txt` is the path to a file with a list of plugin module paths you want to build.

lyw07: thanks for the info!

Issue #3549, opened 2018-04-16
make 'empty table' header handling more consistent


indirectlylit:
### Observed behavior

In some places such as the class list, we hide the table headers. In other places, 

| ![image](https://user-images.githubusercontent.com/2367265/38836017-a24c82be-4181-11e8-803e-5de940d2a20a.png)  | ![image](https://user-images.githubusercontent.com/2367265/38836030-ae52d752-4181-11e8-82e1-5903828d6db9.png)  |
|--|--|
| ![image](https://user-images.githubusercontent.com/2367265/38836070-cc7d6d5a-4181-11e8-9515-9091f51a6fac.png) | ![image](https://user-images.githubusercontent.com/2367265/38836085-d8a70f5a-4181-11e8-9590-199d8ec54d5f.png) |
| ![image](https://user-images.githubusercontent.com/2367265/38836114-f0c01974-4181-11e8-8d3f-a47961e1bbc2.png) | ![image](https://user-images.githubusercontent.com/2367265/38836132-002faa82-4182-11e8-921a-6fa948c679c5.png) |
| ![image](https://user-images.githubusercontent.com/2367265/38836227-36e92706-4182-11e8-877a-f8c0a0cb4809.png) | ![image](https://user-images.githubusercontent.com/2367265/38836238-41a6969c-4182-11e8-960e-46d66cf4b782.png) |
| ![image](https://user-images.githubusercontent.com/2367265/38836286-6b81c7d4-4182-11e8-91b2-10f9af8a799c.png) | ![image](https://user-images.githubusercontent.com/2367265/38836331-936bfbe8-4182-11e8-8bb3-c696bb3792b4.png) |

### Expected behavior

consistent handling

see also #3400:

> When there are no users enrolled/assigned, it might be cleaner not to show the "Full name" and "Username" table headers, since they just look like they're floating in free space

Issue #3400, opened 2018-03-22
Text and alignment on class management page


jamalex:
### Observed behavior

![image](https://user-images.githubusercontent.com/618416/37796100-a0b73dca-2dd3-11e8-9972-9ce0ca4ea35a.png)

### Expected behavior

When there are no users enrolled/assigned, it might be cleaner not to show the "Full name" and "Username" table headers, since they just look like they're floating in free space. The "You don't have any" messages may also look better left-aligned?

(Separately, and maybe this would violate the style guide/conventions, but it might aid discovery of the action buttons if they were contextualized to their sections -- so, "Assign coaches" button next to the "Coaches" section, and "Enroll learners" next to the "Learners" section.

### User-facing consequences

Users might be confused when first encountering this page, before users are added.

### Context

0.9.0 beta demo site

PR #3550, opened 2018-04-16
updates to tables

indirectlylit:
### Summary

re #3400 and #3549

* makes the empty states consistent:
  * consistent typography
  * always show header
* removes empty header space

Old - see screenshots in  #3549

new:

| image |
|--|
| image |
| image |
| image |
| image |

### Reviewer guidance

Make sure this all looks okay

----

### Contributor Checklist

### Reviewer Checklist

### DIFF
--- a/kolibri/core/assets/src/views/core-table.vue
+++ b/kolibri/core/assets/src/views/core-table.vue
@@ -37,8 +37,6 @@
   .core-table
     width: 100%
     font-size: 14px
-    // without this, we get inconsistent column widths
-    table-layout: fixed
 
   >>>thead
     border-bottom: solid 1px $core-grey
--- a/kolibri/plugins/coach/assets/src/views/lessons/LessonResourceUserSummaryPage.vue
+++ b/kolibri/plugins/coach/assets/src/views/lessons/LessonResourceUserSummaryPage.vue
@@ -28,7 +28,7 @@
 
     <!-- TODO consolidate with facility_management user-list -->
     <section>
-      <core-table v-if="userData.length">
+      <core-table>
         <thead>
           <tr>
             <th class="visuallyhidden core-table-icon-col">
@@ -141,7 +141,7 @@
         </tbody>
       </core-table>
 
-      <p v-else>
+      <p v-if="!userData.length">
         {{ $tr('userTableEmptyMessage') }}
       </p>
 
--- a/kolibri/plugins/coach/assets/src/views/reports/channel-list-page.vue
+++ b/kolibri/plugins/coach/assets/src/views/reports/channel-list-page.vue
@@ -1,20 +1,16 @@
 <template>
 
   <div>
+
     <breadcrumbs />
     <template v-if="showRecentOnly">
       <h1>{{ $tr('recentTitle') }}</h1>
       <p v-if="standardDataTable.length">{{ $tr('showingRecent', { threshold }) }}</p>
-      <p v-else>{{ $tr('noRecent', { threshold }) }}</p>
     </template>
     <template v-else>
       <h1>{{ $tr('topicsTitle') }}</h1>
-      <p v-if="!standardDataTable.length">{{ $tr('noChannels') }}</p>
     </template>
-    <core-table
-      v-if="standardDataTable.length"
-      :caption="$tr('channelList')"
-    >
+    <core-table :caption="$tr('channelList')">
       <thead slot="thead">
         <tr>
           <th class="core-table-icon-col"></th>
@@ -46,6 +42,12 @@
         </tr>
       </tbody>
     </core-table>
+
+    <template v-if="!standardDataTable.length">
+      <p v-if="showRecentOnly">{{ $tr('noRecent', { threshold }) }}</p>
+      <p v-else>{{ $tr('noChannels') }}</p>
+    </template>
+
   </div>
 
 </template>
--- a/kolibri/plugins/coach/assets/src/views/reports/recent-items-page.vue
+++ b/kolibri/plugins/coach/assets/src/views/reports/recent-items-page.vue
@@ -4,8 +4,8 @@
 
     <breadcrumbs />
     <h1>{{ $tr('title') }}</h1>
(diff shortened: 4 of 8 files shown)


kolibri/core/assets/src/views/core-table.vue
```diff
@@ -37,8 +37,6 @@
   .core-table
     width: 100%
     font-size: 14px
-    // without this, we get inconsistent column widths
-    table-layout: fixed
```
indirectlylit, line 41: @christianmemije would you take a look at the changes to the tables?

I wasn't sure what the intended behavior of `table-layout: fixed` was, but after a brief inspection the tables look better without it?

christianmemije: cc @DXCanas

kolibri/plugins/coach/assets/src/views/class-list-page.vue
```diff
@@ -18,8 +18,7 @@
         <caption class="visuallyhidden">{{ $tr('tableCaption') }}</caption>
         <thead slot="thead">
           <tr>
-            <th class="core-table-icon-col"></th>
```
christianmemije: Do we want the name column to also contain the icon?

kolibri/plugins/coach/assets/src/views/lessons/LessonSummaryPage/ResourceListTable.vue
```diff
@@ -3,19 +3,14 @@
   <core-table>
     <thead slot="thead">
       <tr>
-        <th class="core-table-icon-col">
+        <th class="core-table-main-col" colspan="3">
```
christianmemije: Right here for example, the name header is right above the move arrows. Feels too far from the actual name column.

localhost_8000_coach_ ipad

kolibri/plugins/coach/assets/src/views/reports/channel-list-page.vue
```diff
@@ -46,6 +42,12 @@
         </tr>
       </tbody>
     </core-table>
+
+    <template v-if="!standardDataTable.length">
```
christianmemije, line 46: Thought on adding `v-if="standardDataTable.length"` to the tbody so we don't just have an empty with an empty tbody?

christianmemije: I take that back. Don't think it is semantically correct to omit the tbody.

kolibri/plugins/device_management/assets/src/views/manage-permissions-page/user-grid.vue
```diff
@@ -8,8 +8,7 @@
     <core-table v-else>
```
christianmemije: Need to also update how we handle an empty state here ^

kolibri/plugins/coach/assets/src/views/reports/recent-items-page.vue
```diff
@@ -8,12 +8,12 @@
     <core-table v-if="standardDataTable.length">
```
christianmemije: Need to also update how we handle the empty state here.

kolibri/plugins/coach/assets/src/views/reports/item-list-page.vue
```diff
@@ -20,12 +20,12 @@
     <core-table>
```
christianmemije: Also handle empty state

kolibri/plugins/coach/assets/src/views/lessons/LessonSummaryPage/ResourceListTable.vue
```diff
@@ -3,19 +3,14 @@
   <core-table>
```
christianmemije: Handle empty state.

kolibri/plugins/coach/assets/src/views/lessons/LessonResourceUserSummaryPage.vue
```diff
@@ -31,11 +31,7 @@
       <core-table v-if="userData.length">
```
christianmemije: Handle empty state.

christianmemije:
`group-section` is missing an empty state. And in`class-list-page`, we handle the empty state differently using the `auth-message`. Should we make this more consistent?

kolibri/core/assets/src/views/core-table.vue
indirectlylit: okay the goal was to get the two tables on the class roster page to line up with each other. However I think it has too many other negative consequences so let's maybe live with that for now

kolibri/plugins/coach/assets/src/views/class-list-page.vue
indirectlylit: I think that could work, too. but yeah the goal here was to get all the headers left-aligned with the page

kolibri/plugins/coach/assets/src/views/lessons/LessonSummaryPage/ResourceListTable.vue
indirectlylit: yeah, that's fair. okay I'll revert those changes

kolibri/plugins/coach/assets/src/views/lessons/LessonSummaryPage/ResourceListTable.vue
indirectlylit: this one is handled: https://github.com/learningequality/kolibri/blob/develop/kolibri/plugins/coach/assets/src/views/lessons/LessonSummaryPage/index.vue#L41

kolibri/plugins/coach/assets/src/views/reports/item-list-page.vue
indirectlylit: created new issue for this one due to string freeze: #3554

indirectlylit:
> group-section is missing an empty state

updated

> in class-list-page, we handle the empty state differently using the auth-message. Should we make this more consistent?

This one maybe seems reasonable as-is because we also have special handling for just one class

Issue #3515, opened 2018-04-10
The remove option in Lesson does not work all the time


amit3107:
### Observed behavior

In the lesson feature, remove option is temporarily showing that the resource is removed but when we go back and then enter the lesson, the resources are still there.

### Expected behaviour

In the lesson feature, remove option must remove a resource from the lesson.

### User-facing consequences

Will not be able to remove a resource.

### Steps to reproduce

1. Login as admin and create a lesson in the class with 2-4 resources.
2. Try to remove the resource, go back and check whether the resources were removed or not.

Kolibri version : kolibri 0.9.0
Operating system : ubuntu 14.04
Browser : Chrome

### screenshots

![l1](https://user-images.githubusercontent.com/38091824/38563493-a4080546-3cfa-11e8-8af1-083343c7a363.png)
![l2](https://user-images.githubusercontent.com/38091824/38563494-a47ae200-3cfa-11e8-82e1-6027828bc262.png)
![l3](https://user-images.githubusercontent.com/38091824/38563495-a4c11d4c-3cfa-11e8-90fb-e31c6a4ccd91.png)
![l4](https://user-images.githubusercontent.com/38091824/38563497-a532a0c0-3cfa-11e8-8dba-8a5966f54715.png)

DXCanas:
So, this happens because the API `patch` isn't made until that snackbar goes away. The idea is you can batch remove a ton of different resources and have all of that equate to a single API call (The removals stack into a single operation that you can undo). Or, if you end up hitting undo, it would be only 1 call rather than 2 (or however many items it is you removed).

The problem, I think, is that we're navigating to a page that doesn't have a `workingResources` state property before snackbar disappears. If you navigate to "add resources", this works fine, but only because `workingResources` is carried over.

I tried fixing this by passing a duplicate of the array (pass-by-value rather than pass-by-reference) to the autosave function. But I'm pretty sure that, because the snackbar's `hideCallback` isn't called until it's actually hidden (which may be on another page), it doesn't work as planned. 

Potential quick fixes _could_ be binding the callback (though that might not work, since the element will have been destroyed by this point anyway?) or passing the data needed as parameters (with defaults?) to the callback.

[...]

PR #3551, opened 2018-04-16
Resource Removal Autosave Deoptimization

DXCanas:
### Summary

Quick fix for #3515 

After talks with @indirectlylit, thought it'd be best to just remove all server call optimizations until we have a more bulletproof solution or a new design.

----
#### Notes on the original fix 

`SaveTime = snackbar dismissal time + debounce time`
Removes a resource after `SaveTime` as long as one stays inside the coach plugin. 

If someone moves to a page that _is *not*_ scoped by a lesson before `SaveTime`, the temporary "optimistic" state they were just looking at is reset to its server representation. This could be confusing, and I'm thinking through the best way to avoid the issue.

The solution I'm leaning towards (if getting the `onRouteLeave` guard is out of the question) it to hold a persistent representation of the lesson's "working" state at a higher level than `pageState` within the coach app. But then we're maintaining yet another representation or resources. 

The reason `SaveTime` has to include both times: 

The snackbar is what allows a user the opportunity to "undo" their removal. The amount of time the snackbar is up represents the app's "uncertainty" as to whether or not the user _really_ wants to remove the resource.

The debounced `save` function is included because it holds any other actions happening to the `workingState`: reorders, removals, additions, etc. The most recent save is the one that happens and it carries the state of all the previous ones. 

### Reviewer guidance

…

### References

…

----

### Contributor Checklist

### Reviewer Checklist

### DIFF
--- a/kolibri/plugins/coach/assets/src/state/actions/lessons.js
+++ b/kolibri/plugins/coach/assets/src/state/actions/lessons.js
@@ -251,7 +251,8 @@ export function showLessonResourceSelectionTopicPage(store, classId, lessonId, t
 }
 
 function getResourceCache(store, resourceIds) {
-  const { resourceCache } = store.state.pageState;
+  // duplicate data to remove reliance on pageState throughout the entire method
+  const { resourceCache } = Object.assign({}, store.state.pageState);
   const nonCachedResourceIds = [];
 
   if (resourceCache) {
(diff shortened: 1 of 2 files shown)


rtibbles:
I think this gets at something deeper that concerns me, the conflation of page state with other data in the store. Personally, I would rather that different pages did not use the same top level object entry (assuming we had an efficient mechanism for garbage collection).

indirectlylit:
As an aside, I believe this issue is analogous to the difficulties we've had saving the most recent progress state when someone navigates away from content.

The way many web apps handle this is to set an 'unsaved-changes' flag somewhere, and then if the user navigates away while that's true you pop up an alert that says "You have unsaved changes - wait?" This pattern works when it's rare (due to fast or infrequent saves) but would get frustrating if it occurred frequently.

Issue #2197, opened 2017-09-14
core API - pluggable routing / nav bar items


indirectlylit:
We need a flexible and forward-looking way for application-level plugins (e.g. manage, learn, coach) to register themselves as wanting to add a widget to the nav bar.

This is might not be as simple as just a list of names and URLs:

* Needs to take into account ordering, user permissions, icons, etc
* potentially support more complex widgets than just text + link (e.g. we at one point discussed displaying learner points in the nav bar)
* Might also need to support sub-section links

rtibbles:
I believe this is fixed, except for the sub-section piece, which would come more under the 'plugins for plugins'?

indirectlylit:
yes, thanks

PR #3559, opened 2018-04-18
Sidenav and redirect architecture proposal

### LINKED ISSUE #2025: Make Sidebar pluggable
## Summary

Currently we hard code which URLs we expose in the main navigation sidebar.

Given that we have a plugin driven architecture, this is suboptimal, because only plugins that correspond to our anticipated plugin manifest will ever be exposed through the sidebar.

This also makes it difficult to implement the kind of suggestion made by @indirectlylit https://github.com/learningequality/kolibri/issues/1914#issuecomment-318434198 to implement a more material design style navigation system.

It is proposed that the sidebar be made pluggable for the following purposes:

- Bootstrapping SPA endpoints from the Django backend, to populate the main sidebar navigation items like 'Learn', 'Coach' etc. This would require the backend to supply properties such as: display name (internationalized on the backend), an identifier label. It would also require the design of some sort of ordering system, either based on 'priority' (i.e. how close to the top it should be), or 'preference' (where in the nav bar does it want to sit?). The latter system seems preferable to include the preference for appearing below the divider bar in the side nav. In either system, tie breaking could be delegated to alphabetical sorting on the label.

- SPAs on the frontend registering their sub navigation items to be displayed under their top level nav link on the sidebar.

### LINKED ISSUE #2423: update side-nav to use router links
image

Currently, the side-nav uses keen's `ui-menu`. This means that links cannot be right-clicked to copy the URL or control-clicked to open them in a new tab. If we switch to using vue `router-link` objects, this will be addressed.

### LINKED ISSUE #3819: Add validation for roles, priority, and section properties of side nav components
#3559 introduced side nav components that can plug into the side nav. While all of the above properties are optional on these components, we currently do no validation on them to ensure that when the properties are set, they are set appropriately.

rtibbles:
### Summary
* We currently hard code navigation items in the side nav (#2197)
* This makes it harder to make Kolibri truly pluggable (#2025)
* We also don't use purely links in the side nav, which makes it hard to use 'open in new tab' (#2423)
* To address this, this PR updates the current hooks that had previously existed for generating navigation items.
* It adds WebpackBundleHooks as mixins for NavigationHooks on each of the plugins that need to register a frontend navigation item. This then uses the regular WebpackBundleHook machinery to specify the vue component that defines the navigation item as the entry file for the bundle.
* Within this vue component, the nav item is defined, and then explicitly registered against the navComponents core util.
* The sidenav then populates itself from these components.
* Components are filtered and sorted by their UserKind specified in their `role` property, and can optionally be added to the `account` section, which is also used to populate the appbar dropdown.
* To prevent hardcoding of plugin URLs in the core login action, a `RedirectHook` is created to allow plugins to specify how Kolibri's core URL redirect behaviour should function, depending on installed plugins.
* This is then extended to add 'first_login' flagging for first login specific behaviour.

N.B. first login specific behaviour currently only works if the frontend says it is the first login, but it may be desirable to flag on the user themselves whether this is the case.

### Reviewer guidance

This seems ready to go now, but might need some more documentation, and possibly some additional validation on the priority, role, and section properties of the registered components.

image

Only change is now all drop down menu items have icons too:
image

### References
Fixes #2025, solves #2197, and closes #2423.

This now also fixes #3819 also, also. Also.

----

### Contributor Checklist

### Reviewer Checklist


kolibri/core/assets/src/utils/navItems.js
```diff
@@ -0,0 +1,11 @@
+export let menuItems = [];
+
+export function setMenuItems(items) {
+  menuItems = items;
+}
+
+export let userMenuItems = [];
```
indirectlylit: recommend calling this module something more general, e.g. `customConfig`.

Maybe also consider putting everything in a single big JS config object (à la webpack config) rather than defining separate JS variables for each thing:

```
customConfig = {
  sideNavLinks: {...},
  userMenuItems: {...},
}
```

That puts us in a better position to spec and manage the API for "The Config Object" without having to manage lots of separate JS imports, bootstrap code, etc.

kolibri/core/assets/src/views/app-bar.vue
```diff
@@ -167,16 +158,42 @@
...
+        }
+        if (navItem.role === UserKinds.SUPERUSER) {
+          return this.isSuperuser;
+        }
+        if (navItem.role === UserKinds.ANONYMOUS) {
+          return !this.isUserLoggedIn;
+        }
+        if (navItem.role === UserKinds.LEARNER) {
+          return this.isLearner || this.isCoach || this.isAdmin || this.isSuperuser;
+        }
```
indirectlylit: hmm...

Not sure I love the reliance on `role` for this. I feel like it's baking in some assumptions about our menu API that might not hold up in the long-term

kolibri/core/assets/src/utils/navItems.js
```diff
@@ -0,0 +1,11 @@
+export let menuItems = [];
+
+export function setMenuItems(items) {
```
indirectlylit: maybe call these "sideNavLinks"?

kolibri/core/assets/src/utils/navItems.js
```diff
@@ -0,0 +1,11 @@
...
+  menuItems = items;
+}
+
+export let userMenuItems = [];
+
+export function setUserMenuItems(items) {
+  userMenuItems = items;
+}
```
indirectlylit: setters might not be necessary? can we just assign directly to the exported variable?

kolibri/core/assets/src/views/custom-ui-menu/menu-option.vue
```diff
@@ -1,41 +1,44 @@
...
-        >
-          {{ label }}
+  <li>
+    <a
+      :href="link"
+      class="ui-menu-option"
+      role="menuitem"
+      :class="classes"
+      :tabindex="(isDivider || disabled) ? null : '0'"
+    >
```
indirectlylit: yay it's a link!

kolibri/core/assets/src/views/custom-ui-menu/menu-option.vue
```diff
@@ -48,6 +51,8 @@
   import UiIcon from 'keen-ui/src/UiIcon.vue';
   import UiRippleInk from 'keen-ui/src/UiRippleInk.vue';
 
+  const flippableIcons = ['exit_to_app'];
```
indirectlylit: seems a little strange to define here?

kolibri/core/assets/src/views/side-nav.vue
```diff
@@ -65,21 +64,30 @@
...
+  import { menuItems } from 'kolibri.utils.navItems';
+
+  // Explicit ordered list of roles for nav item sorting
+  const navItemRoleOrder = [
+    UserKinds.LEARN,
+    UserKinds.COACH,
+    UserKinds.ADMIN,
+    UserKinds.CAN_MANAGE_CONTENT,
+    UserKinds.SUPERUSER,
+  ];
```
indirectlylit, line 96: hmm..

kolibri/core/assets/src/views/side-nav.vue
```diff
@@ -216,36 +153,111 @@
...
+          return this.isSuperuser;
+        }
+        if (navItem.role === UserKinds.ANONYMOUS) {
+          return !this.isUserLoggedIn;
+        }
+        if (navItem.role === UserKinds.LEARNER) {
+          return this.isLearner || this.isCoach || this.isAdmin || this.isSuperuser;
+        }
+      },
+      addDividers(itemList) {
```
indirectlylit: the behavior of this function is not obvious

(I didn't take the time to parse it, and I think it will give future devs pause)

kolibri/deployment/default/views.py
```diff
@@ -1,14 +1,46 @@
...
+                url = url or get_url_by_role(user_kinds.SUPERUSER, first_login=first_login)
+            if user_kinds.ADMIN in roles:
+                url = url or get_url_by_role(user_kinds.ADMIN, first_login=first_login)
+            if user_kinds.COACH in roles:
+                url = url or get_url_by_role(user_kinds.COACH, first_login=first_login)
+            url = url or get_url_by_role(user_kinds.LEARNER, first_login=first_login)
+        else:
+            url = get_url_by_role(user_kinds.ANONYMOUS)
+
+        return HttpResponseRedirect(url)
```
indirectlylit: (I still personally believe we should redirect to the last app they used, rather than assuming their assigned permission role is the source of truth about their actual lived role.)

kolibri/core/assets/src/utils/navItems.js
rtibbles: Well, doing it like this allows access to each item directly from the module namespace, whereas exporting an object as the default export from a module (which would allow the same thing) breaks our linting rules.

kolibri/core/assets/src/views/app-bar.vue
rtibbles: I disagree to some extent, as restriction of menu items by currently logged in user role is the main active component of our side navigation right now. I am not sure what the alternative is, or are you suggesting this is insufficiently general? I would rather keep it constrained at first, and then look to change the API if we have additional use cases.

kolibri/core/assets/src/utils/navItems.js
rtibbles: This might be fine, yeah - I will check, I did it like this, as I recall having weird issues with assigning to module exports from outside the module.

kolibri/core/assets/src/views/custom-ui-menu/menu-option.vue
rtibbles: Yeah - this should probably be defined inside the ui-icon component so that we can just configure flipping information by isRtl on that.

kolibri/core/assets/src/views/side-nav.vue
rtibbles: This is how we currently order side nav items, and I also think sorting by 'user role' and then just alphabetically might be the simplest way to get rid of the `priority` field.

kolibri/core/assets/src/views/side-nav.vue
rtibbles: Yeah, it would need additional documentation and cleanup.

kolibri/deployment/default/views.py
rtibbles: I am not sure we've heard strong feedback that this is desirable for users, but we have heard that coach users should be redirected to coach.

Happy to be wrong about this, but this does, for now, replicate the previous behaviour without having to hardcode plugin URLs in the core.

rtibbles:
> it makes the logic much harder to reason about and control

This was my main concern, and I would rather it is simplified.

> it invites hacks like using priority = 1.00001 or priority = 999999 as plugins compete with each other

I think just using the role as the top level sort, and then sorting alphabetically could stop this.

> I really dislike how the config is defined as subclasses of NavigationHook and UserMenuHook and then obliquely dumped into JS with json.dumps([h.menu_item for h in UserMenuHook().registered_hooks]). It makes it very difficult to understand what the schema of the config variables will be.

This could be addressed, as the `menu_item` method from the hook does not need to be there, it could be in the tag instead, so that the schema of the config variables is explicitly there, I can also add validation to it that would error out and give meaningful errors if there was a misconfigured Hook - for example, at least one of the properties is optional, so that should be more evident). As for subclassing, that's how hooks work, and they rely on the inheritance to register unique entries (I don't really want to argue about that implementation, because I didn't write it, and I am not fussed either way).


> Concretely, I'd be interested in exploring a strategy where the configuration of these menus (including ordering, divider locations, etc) is explicitly encoded in a config file with a well-defined schema, like webpack. This config file could be overwritten, extended, or transformed during the custom build process. It could even reference custom implementations of vue nav item components that have their own internal logic and references to vuex state.

This is precisely what I am trying to avoid as this gives us very little over what we are currently doing (except some slight convenience when we make a custom build for a third party). On a purely theoretical level, this still breaks the pluggability of plugins (if you deactivate the 'facility_management' plugin, its sidenav entry would still be there, whereas with this approach, it would disappear).

On a practical level, another big motivation for me of this work was to be able to include 'prototype' plugins in our distributed mainline Kolibri builds, but that are deactivated by default. These plugins can then be activated, and if they necessitate a top level side nav link, then they will be available to the end user.

kolibri/core/assets/src/views/app-bar.vue
indirectlylit: Specifically, I'm worried that baking the concept of 'role' into into the plugin API makes it much harder to evolve that concept later. We've already seen roles evolved a bit (superusers, device permissions, and class scoped coaches).

More generally, trying to schematize these items I think could could impede evolution of our design by locking us into the way we happen to do things now. For example, what if we decided that we wanted to add a number badge (like unread notifications, points) or other more complex interactions (pop-ups, etc)?

That's why I was thinking it would be good if plugins could drop in their own menu item component implementations, almost slot-style. These could either re-use a core `<k-side-nav-item>` implementation, or build their own from scratch. Similarly, visibility logic could depend on any core or plugin state...

indirectlylit:
> This is precisely what I am trying to avoid as this gives us very little over what we are currently doing (except some slight convenience when we make a custom build for a third party). On a purely theoretical level, this still breaks the pluggability of plugins (if you deactivate the 'facility_management' plugin, its sidenav entry would still be there, whereas with this approach, it would disappear).
> 
> On a practical level, another big motivation for me of this work was to be able to include 'prototype' plugins in our distributed mainline Kolibri builds, but that are deactivated by default. These plugins can then be activated, and if they necessitate a top level side nav link, then they will be available to the end user.

I think that concern is a little bit of a strawman. If a plugin is deactivated, of course the corresponding menu items should be hidden, no?

Maybe a stronger way to frame the critique is: my suggestion would probably make it harder to _dynamically_ install and uninstall plugins that weren't originally present at build time.

(I know we've talked in the past about how necessary - or possible - this is. Maybe it's gotten closer with the build decoupling system?)

kolibri/deployment/default/views.py
indirectlylit: yes

kolibri/core/assets/src/views/side-nav.vue
indirectlylit: a way to gauge ... If a developer is interested in adding or removing dividers based on mockups from designers, will it be straightforward to know how to implement the mockups

kolibri/core/assets/src/views/app-bar.vue
rtibbles: Yeah, as I noted in the general thread, this is not impossible to achieve. My main concern is that it makes the API more brittle due to the increased flexibility.

Changing the structure of the core vuex suddenly has implications that may not be immediately obvious (because they are referenced inside these additional side nav definitions) - if we were to do this, only allowing references to the core getters would be preferable, but that may well rob much of the flexibility you seem to want to allow.

kolibri/core/assets/src/views/side-nav.vue
rtibbles: Well, part of this design (by necessity) is that we no longer have direct control over the exact positioning of dividers. In this implementation, they are added at boundaries between sections (so, when section 0 ends, and section 1 begins). There is no explicit control over this, in order to allow for dynamic expansion of the side nav. This may well be undesirable, but unless we follow the build time configuration approach you suggest, we cannot guarantee the contents of the side nav completely.

rtibbles:
> I think that concern is a little bit of a strawman. If a plugin is deactivated, of course the corresponding menu items should be hidden, no?

How? If this is done at build time (as you are suggesting) then it will not be deactivated when the plugin is deactivated. Precisely because it should be hidden if deactivated is why I have implemented a non-build time configuration based system.

> Maybe a stronger way to frame the critique is: my suggestion would probably make it harder to dynamically install and uninstall plugins that weren't originally present at build time.

If the information is there to deactivate a side nav item at run time, almost all systems I can conceive of for achieving this goal allow for both 'build time present' and 'build time absent' plugins to be activated and deactivated.

There is one scenario where I can imagine this not being the case: In order to achieve the ability to deactivate only build time plugins, at build time, metaprogramming is used to generate code to inject a 'activation' signal into the frontend showing that the items associated with this should be activated. What this is doing is basically doing what I am doing here with more magic (having a way of injecting which nav items should be activated through some plugin registry driven means), but has the drawback of disallowing non-build time present plugins from achieving the same thing.

> (I know we've talked in the past about how necessary - or possible - this is. Maybe it's gotten closer with the build decoupling system?)

The present architecture is still designed for the future goal of allowing this to happen, I think there is still an open question about *whether* this is something that people truly want (although, it certainly seems like people do want to change how Kolibri behaves - and it would be nice to prevent people from feeling the need to fork to achieve this) - it remains to be seen whether we can build a sufficiently tractable plugin API that people feel comfortable working with, but I think unless we have very strong reasons not to, I would rather maintain this goal than abandon it (especially as not doing this introduces us as a strong bottleneck on customized use of Kolibri).

In terms of possibility, yes, it is getting closer. Now that it is possible to build the frontend code without the Django stack, we are closer to that. Another important piece will be versioning and settling the coreAPI, and also then making that exportable so that it can be referenced by external plugins. I think you had some good ideas about limiting the scope of this (while still maintaining our somewhat catchall approach to the coreAPI that we currently have!)

From our discussion, I get the sense that perhaps the issues may be less about 'build time configuration' vs 'dynamic configuration', and maybe more about how much of this gets encoded in the frontend code vs the backend code? If the configuration of individual navigation items was done in frontend code would you still be as concerned about the use of build time configuration?

kolibri/core/assets/src/views/app-bar.vue
indirectlylit: > Changing the structure of the core vuex suddenly has implications that may not be immediately obvious (because they are referenced inside these additional side nav definitions) - if we were to do this, only allowing references to the core getters would be preferable, but that may well rob much of the flexibility you seem to want to allow.

No, I totally agree with you that allowing long-term, unfettered access to core state is undesirable.

I'd imagine that in the same way we've talked about having some subset of the core components be 'public' in the component API spec, we can have a subset of 'public vuex getters' that we commit to maintain compatibility on even if the rest of the core system evolves underneath.

kolibri/core/assets/src/views/side-nav.vue
indirectlylit: right

kolibri/core/assets/src/views/custom-ui-menu/menu-option.vue
rtibbles: Will not resolve this here, but rather follow up with an issue to make a kIcon wrapper which can import a list of icons that should be flipped for RTL purposes.

indirectlylit:
is this ready for final test & merge?

rtibbles:
I think it is good for manual testing. I would like to add more automated tests on the backend for some of the hook behaviour.

rtibbles:
Might also be good to add some more validation and documentation on the hook as well.

rtibbles:
This is now refactored and rebased to latest develop.

kolibri/core/assets/src/views/app-bar.vue
```diff
@@ -75,28 +72,9 @@
...
             />
           </core-menu-option>
-          <core-menu-option
-            :label="$tr('signOut')"
-            @select="kolibriLogout"
-          />
+          <logout v-if="isUserLoggedIn" />
         </template>
 
-        <template v-else slot="options">
```
christianmemije, line 84: No longer seeing sign in option. Intended behavior?

kolibri/core/assets/src/mixins/nav-components.js
```diff
@@ -0,0 +1,49 @@
...
+import { UserKinds } from 'kolibri.coreVue.vuex.constants';
+
+export default {
+  methods: {
+    filterByRole(navItem) {
+      if (!navItem.role) {
+        // No role defined, so always show
+        return true;
+      }
+      if (navItem.role === UserKinds.COACH) {
```
christianmemije, line 18: Minor, but could we order these hierarchically?  Maybe use a switch statement?

kolibri/core/assets/src/views/core-base.vue
```diff
@@ -78,18 +75,6 @@
     },
     mixins: [responsiveWindow],
     props: {
-      // This prop breaks the separation between core and plugins.
```
christianmemije, line 81: :tada: :tada: :tada:

kolibri/core/assets/src/views/side-nav.vue
```diff
@@ -233,15 +131,21 @@
...
-      },
       mobile() {
         return this.windowSize.breakpoint < 2;
       },
       footerMsg() {
         return this.$tr('poweredBy', { version: __version });
       },
+      menuOptions() {
+        const topComponents = navComponents
+          .filter(component => component.section !== 'account')
```
christianmemije: Reference the constant?

kolibri/core/assets/src/views/app-bar.vue
rtibbles: No, not intended, thanks for checking, will double check.

rtibbles: Ah, it's in the side nav, but not the app bar. Will fix!

kolibri/core/views.py
```diff
@@ -38,3 +47,47 @@ def set_language(request):
...
+        """
+        Redirects user based on the highest role they have for which a redirect is defined.
+        """
+        first_login = request.session.get("first_login", False)
+        if request.user.is_authenticated():
+            url = None
+            if request.user.is_superuser:
+                url = url or get_url_by_role(user_kinds.SUPERUSER, first_login)
+            roles = set(Role.objects.filter(user_id=request.user.id).values_list('kind', flat=True).distinct())
+            if user_kinds.ADMIN in roles:
```
christianmemije, line 84: Use else if statements instead?

kolibri/core/assets/src/views/side-nav.vue
rtibbles: Oh whoops, missed this one.

kolibri/core/views.py
rtibbles: I thought about that, but it doesn't quite do it, because you need to be able to fall through from an admin user to a coach user to a learner (so if nothing is defined for an admin, they still get redirected to wherever a coach or a learner would end up).

kolibri/core/assets/src/mixins/nav-components.js
indirectlylit: I've come to believe switch statements should be avoided in JS.

here's someone on the internet with a similar opinion: https://medium.com/chrisburgin/rewriting-javascript-replacing-the-switch-statement-cfff707cf045

rtibbles: Missed this comment, but I think this is readable, clear, and gets the job done.

With the validation I am doing, we can also guarantee that one of these conditions will trigger.

Issue #3554, opened 2018-04-18
handle empty state of coach report item list page


indirectlylit:
### Observed behavior

In #3550 we made the 'empty states' of tables more consistent.

We could not update this instance due to string freeze, so we'll do it in 0.10.0

https://github.com/learningequality/kolibri/blob/develop/kolibri/plugins/coach/assets/src/views/reports/item-list-page.vue#L20

### Expected behavior

Same empty-state behavior as other tables

### User-facing consequences

Slight inconsistency

### Context

0.9.1

jonboiser:
#3591

PR #3591, opened 2018-04-24
Add empty state for coach content item report

jonboiser:
FOr #3554, adds a message if, in a coach content report, a topic somehow doesn't have any children.

screen shot 2018-04-24 at 1 12 20 pm

### DIFF
--- a/kolibri/plugins/coach/assets/src/views/reports/item-list-page.vue
+++ b/kolibri/plugins/coach/assets/src/views/reports/item-list-page.vue
@@ -121,43 +121,27 @@
     },
     methods: {
       genRowLink(row) {
+        const rowIsTopic = row.kind === ContentNodeKinds.TOPIC;
+        const params = {
+          classId: this.classId,
+          channelId: this.pageState.channelId,
+          [rowIsTopic ? 'topicId' : 'contentId']: row.id,
+        };
         if (TopicReports.includes(this.pageName)) {
-          if (row.kind === ContentNodeKinds.TOPIC) {
-            return {
-              name: PageNames.TOPIC_ITEM_LIST,
-              params: {
-                classId: this.classId,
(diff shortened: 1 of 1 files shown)


rtibbles:
Code looks fine.

@radinamatic @carinediaz @jtamiace can we get a copy read here? `No Items` for the empty coach report state. See screenshot above for context.

radinamatic:
`Items` is not a user-friendly term, I'd favor something more explicit, for example: `There are no resources or exercises in this topic`.

jtamiace:
Agree with Radina, and I'm not sure whether we still mention `items` anywhere else in the UI? If we do, think we've been moving towards defining everything as `resource(s)`. Would shorten it even more to `No exercises or resources in this topic`

jonboiser:
I think this component is used for both reporting on Topics/Resources and Learners for a specific Resource. We need a message for both cases (no sub-topics/sub-resources, and no learners).

1. http://kolibribeta.learningequality.org/coach/#/96516155886defde0d170572c905ed73/learners/e369a0858c0c80cbff128d54973bba2a/f9d3e0e46ea25789bbed672ff6a399ed
1. http://kolibribeta.learningequality.org/coach/#/96516155886defde0d170572c905ed73/recent/1ceff53605e55bef987d88e0908658c5/350444683e72512b8d7e54af52e0b541

radinamatic:
Is there a specific reason different scenarios require the same end result, other to save dev time? Specific message would decrease user's uncertainty and offer much better feedback than a completely generic terms...

jonboiser:
This particular UI is only for listing the items of topics. I was confused because this listing can be shown for a Learner (like in the screenshot), but also for the entire class (e.g. if we were under the "Recent" or "Channels" tab).

So we just need to decide on something to the effect of "No exercises or resources in this topic"

Issue #3520, opened 2018-04-11
add kolibri command to change the location of the content folder


indirectlylit:
### Observed behavior

On a number of occasions, we've gotten feedback that folks want the content folder in a different location than the primary Kolibri directory. This often occurs even after a user has imported content, which means that we may need to migrate existing content to the new locations.

### Expected behavior

We should provide a new command to migrate the content directory to a new location. One possibility might be something like:

```bash
kolibri content directory  # displays current directory
kolibri content directory /media/external  # migrates the content to a new directory
```

However any command is fine if it serves the general purpose.

It would be reasonable to place some restrictions on this command if helpful, e.g. that the Kolibri server can't be running or be started while it's running.

This may also require some additional checks during Kolibri startup, since now there would be a chance that Kolibri might start without the content directory being available, or the content directory might become unavailable while Kolibri is running. These cases might require errors to get bubbled up to the end-user on the front-end.

### User-facing consequences

Users are not able to set up their systems and leverage higher-capacity hard drives as desired

benjaoming:
Which are the user scenarios for this?

1. backup purposes
1. seeding a deployment

indirectlylit:
My understanding was that 'moving contents to another disk' is the scenario we've heard requests for.

Would it be useful to follow-up and find specifics?

benjaoming:
@indirectlylit thanks -no I think this clarification is sufficient. Seeding a deployment should probably follow a different pattern, but could perhaps use `kolibri manage content` as a component in that procedure.

PR #3594, opened 2018-04-25
Command `contentdirectory` to migrate content in Kolibri

lyw07:
### Summary

This PR creates a new management command `contentdirectory` to migrate content in KOLIBRI_HOME to a new location. It saves the path to the current content directory in `KOLIBRI_HOME/kolibri_settings.json`.

The `contentdirectory` command contains two subcommands.
1. Show - `kolibri manage contentdirectory show`
This show command shows the current content directory.
This command has an option `--kolibri-home` to check whether the current content directory is inside KOLIBRI_HOME or not.

2. Migrate - `kolibri manage contentdirectory migrate <destination>`
This migrate command helps migrate the content from current content directory to **destination/content** directory.
For example, if the original content directory is `~/.kolibri/content`, then the new location is `<destination>/content`.

There are four cases in using the migrate command:

- If the `<destination>` does not exist, then the command will create the `<destination>` folder and then move the current content directory into the new location.
- If the `<destination>` exists but doesn't have any content inside, then the command will do the same thing as above.
- If the `<destination>` exists and has content inside, but the user does not want to keep the content, then the command will delete the `<destination>` folder and then create a new folder to move the current content directory in there.
- If the `<destination>` has content inside and the user wants to keep the content there, then the command will move the content from current content directory to the destination while keeping the old content in the destination folder.

I also add a sanity check for the existence of the content directory before starting Kolibri.

### Reviewer guidance

1. Run `kolibri manage contentdirectory show` to see the current content directory
2. Run `kolibri manage contentdirectory migrate <destination>` to migrate to a new location. Check to see if the content exists in the new location.

> Note: There are two things I'm not sure about my code.
The first one is https://github.com/learningequality/kolibri/compare/develop...lyw07:contentdirectory-command#diff-3d768d4c80976bbf31fa767882807fe4R98, in which I check if there are files inside `<destination>/content/databases` to see if the destination folder has content available. I'm not sure if this check is sufficient.
The second one is https://github.com/learningequality/kolibri/compare/develop...lyw07:contentdirectory-command#diff-d2fc08a4eefe48248e5c777e07b056fdR50, in which I check if the current content directory exists before starting kolibri. I put down my concerns in the comments.

Please let me know if you have any problems reading my code or have any suggestions regarding my concerns above. Thank you!

### References

Issue: https://github.com/learningequality/kolibri/issues/3520

----

### Contributor Checklist

### Reviewer Checklist


lyw07:
cc @indirectlylit the command I create here is a little different from the command you described in the issue. Please feel free to comment about things I need to change. Thanks!

jamalex:
Cool! Just a high-level question first, without having looked at the code yet: After moving the content, we don't have a systematic cross-platform way to persist the path change so that Kolibri will look to the new location, do we? So once they move it, they need to do some other configuration on their own, which we need to make sure they know. Right now, changing the content directory requires creating a custom settings file and then referencing it every time you start Kolibri (or in an environment variable). And with the PEX, it's not actually even technically possible since it can't reference external Python paths to find the settings module. This is a great example of why things like the content directory should be specified in some other sort of config file, rather than in a Django settings file. This relates to some brainstorming @benjaoming and I have done in this doc: https://docs.google.com/document/d/1v-_B7dFlEw9EAOfMSU78IkfhMX9DihzN7jjjxzEUfBk/edit

lyw07:
@jamalex Yes! Regarding the question, I'm currently putting the `CONTENT_DIRECTORY` variable in the `KOLIBRI_HOME/kolibri_settings.json` file. We have already referenced the file every time when we start kolibri (I think this file was set up for the plugins?).

rtibbles:
Yeah, this has been our go to place to put these kinds of things. It just isn't clear that this is a sustainable way forward.

@jamalex @benjaoming @aronasorman have we made any progress on the 'sorting out Kolibri settings' project?

jamalex:
>have we made any progress on the 'sorting out Kolibri settings' project?

Not in practical terms. Discussing right now in #sysops, around DB config stuff too. We should, though.

benjaoming:
We have previously discussed ideas of importing and exporting for seeding purposes.

Would suggest to shorten the command to just `content`. Could be:

```
kolibri manage content show
kolibri manage content export <location> [--channels=<channel-id>[, <channel-id>]]
kolibri manage content import <location> [--channels=<channel-id>[, <channel-id>]]
kolibri manage content delete [--channels=<channel-id>[, <channel-id>]]
kolibri manage content set-directory <location>
```

I like brainstorming CLIs :)

lyw07:
@benjaoming thank you for the info! I'm thinking maybe instead of `show` we could do `show-directory` to make it more clear to the users? 
Also, since currently we have `kolibri manage content importchannel` etc, do you think it would be okay for me to remove the dash so that it looks like `kolibri manage content showdirectory`? 
Thank you!!

benjaoming:
@lyw07 ah, I thought the `show` command would show the channels in the content directory. I've updated the example in the comment above. I don't know if this function is useful.. but I can imagine that it's nice to be able to verify if a location (current or another location) contains expected channels.

lyw07:
@benjaoming Thank you! Will update based on the info

rtibbles:
@lyw07 could you update to leverage the just merged options file work?

lyw07:
@rtibbles yes! thank you!

lyw07:
@jamalex @benjaoming Updated the code with the settings from `options.ini`. Also moved the command code from `kolibri/deviceadmin` to `kolibri/content`.
Currently, the commands are:
`kolibri manage getdirectory`,
`kolibri manage setdirectory`

I didn't shorten the commands to be `kolibri manage content <command>`, because it influences other commands such as `importchannel` and `exportchannel`. I'm not sure if we should start to shorten the content commands from this pull request. Thank you!

benjaoming:
>  I'm not sure if we should start to shorten the content commands from this pull request. Thank you!

IMO we should, because it's better to do it now than later! @jamalex what do you think?

I'll set the changelog label on this, so we can sync up on whether to add release notes for this..

kolibri/content/management/commands/setdirectory.py
```diff
@@ -0,0 +1,141 @@
...
+        try:
+            server.get_status()
+            self.stderr.write(self.style.ERROR(
+                "Cannot migrate content while Kolibri is running, please run:\n"
+                "\n"
+                "    kolibri stop\n"
+            ))
+            raise SystemExit()
+        except server.NotRunning:
+            pass
```
benjaoming: If you look at the `dbbackup/dbrestore` commands, you should find some code to use for when the database isn't SQLite, meaning that we should fail?

rtibbles: Why should it fail when the DB is not SQLite? This code doesn't impact the default database, right?

kolibri/content/management/commands/setdirectory.py
```diff
@@ -0,0 +1,141 @@
...
+        # TODO: This check may not be sufficient
+        dest_database_directory = os.path.join(dest_directory, 'databases')
+        dest_content_exists = os.path.exists(dest_database_directory) and \
+            os.listdir(dest_database_directory)
+
+        # If destination has content inside
+        if dest_exists and dest_content_exists:
+            self.copy_when_dest_content_exists(dest_directory, current_directory)
+        else:
+            self.copy_when_dest_content_dne(dest_exists, dest_directory, current_directory)
```
benjaoming: Shouldn't the directory just be created if it doesn't exist?

kolibri/content/management/commands/setdirectory.py
lyw07: Sorry I'm not sure about this... What I'm trying to implement here is that when the server is running, then users shouldn't be able to change the content directory.

kolibri/content/management/commands/setdirectory.py
```diff
@@ -0,0 +1,141 @@
...
+            raise SystemExit()
+
+    def copy_when_dest_content_dne(self, dest_exists, dest_directory, current_directory):
+        """
+        If the destination defined by the user does not exist,
+        or the destination exists but does not have content inside,
+        or the destination exists and has content inside but the user does not
+        want to keep the content during the migration,
+        then remove the destination if exists and start the copying process.
+        """
```
benjaoming: This function doesn't handle any user interaction and seems to just remove `dest_directory`, so perhaps the comment can be a bit more specific, or the order of calls and conditions can be re-arranged a bit?

kolibri/content/test/test_contentdirectory.py
```diff
@@ -0,0 +1,119 @@
...
+        with self.assertRaises(SystemExit):
+            call_command('setdirectory', 'test')
+            remove_mock.assert_not_called()
+
+    @patch('kolibri.content.management.commands.setdirectory.update_options_file')
+    @patch('kolibri.content.management.commands.setdirectory.rmtree')
+    @patch('kolibri.content.management.commands.setdirectory.copy_tree')
+    @patch('kolibri.content.management.commands.setdirectory.input', return_value='yes')
+    @patch('kolibri.content.management.commands.setdirectory.os.listdir')
+    @patch('kolibri.content.management.commands.setdirectory.os.path.exists', return_value=True)
```
benjaoming: A lot of patching here :)

kolibri/utils/sanity_checks.py
```diff
@@ -42,3 +44,25 @@ def check_port_availability(host, port):
...
         sys.exit(1)
+
+
+"""
+TODO:
+1. Checking the existence of content directory might not be enough.
+    it could be possible that the user deletes certain files/folders in the
+    content directory instead of the whole folder. But I'm not sure if it would
+    be better to do a full check here, since I assume it will take a
+    lot more time for kolibri to start
```
benjaoming: I think this is okay. Instead of providing some weird merge functionality, let the user decide to either migrate to a CLEAN target over be clear that EVERYTHING is overwritten.

...can't see the real benefit of merging stuff, especially if it doesn't work anyways (database-wise things need merging as well).

kolibri/content/management/commands/setdirectory.py
lyw07: ah thank you! I just realized that after I refactored some of my code, this function becomes not really meaningful

kolibri/utils/sanity_checks.py
lyw07: thank you for the suggestion!

kolibri/content/test/test_contentdirectory.py
lyw07: yes... 😓 will try to remove some to make it simple

kolibri/utils/sanity_checks.py
benjaoming: Just to clarify... I meant "weird" as in "hard to explain in a way that the user will quickly understand how it works" :)

kolibri/content/management/commands/setdirectory.py
benjaoming: Ah okay, that makes sense @rtibbles - I was just wondering about content discovery when changing content directories, so if the database files remain untouched, that is totally fine.

indirectlylit:
Created a new issue for cleanup of the commands - #3727

I noticed one discrepancy regarding the treatment of the `content` directory. Observe:

```bash
> kolibri manage getdirectory
Current content directory is /Users/d/kolibri/010x/content.

> kolibri manage setdirectory /Users/d/kolibri/c2
Current content directory is /Users/d/kolibri/c2/content.

> kolibri manage setdirectory /Users/d/kolibri/010x/content
Current content directory is /Users/d/kolibri/010x/content/content.
```

If you try feeding output to the input, it creates nested `content` directories. I'd also recommend removing the trailing period on the output to make it easier to copy.

Finally, looks like there are a couple failing tests.

thanks!

lyw07:
@indirectlylit thank you for creating the issue and testing the pr! I will remove the `content` directory at the end of every result of content migration. I will also fix the tests. thanks!

benjaoming:
> Currently, the commands are:
> 
>     kolibri manage getdirectory,
>     kolibri manage setdirectory

Hi @lyw07 

These are really good changes, just wanna flag my suggesting for name changing the commands.

I think that the above commands should be changed for having some consistency, expecting future commands.

But the other reasoning is that they aren't very self-explanatory ATM: Which directory are we getting and setting? :)

I don't wanna say that `kolibri manage content <command>` is the only way, it could also be `kolibri manage content-<command>`, whereby all commands would show up directly in `kolibri manage help`. However, it may also result in a cluttered experience -- note the current state is already pretty cluttered :)

Another possible benefit of `kolibri manage content <command>` is that it will force all the `content` commands to go into the same application and make DRY easy.

lyw07:
Thank you for the suggestions @benjaoming! I agree that the current commands look a bit messy and not organized, and `kolibri manage content <command>` is way more clean. 
I saw that @indirectlylit created this issue: https://github.com/learningequality/kolibri/issues/3727 to discuss it. 
I think the only question I have is whether we want this change to go into 0.10. 
Thank you!

indirectlylit:
thanks!

agreed that the naming could be better. However first we have some much higher priority 0.10.0 items to push through first

indirectlylit:
> But the other reasoning is that they aren't very self-explanatory ATM: Which directory are we getting and setting?

Maybe as a quick fix for this - just call them `get_content_dir` and `set_content_dir` or similar? Then we can address the subcommand namespacing stuff in #3727

indirectlylit:
@lyw07 ran into this issue after testing the latest changes. I think it deleted all my content?

I was trying to move the content from

```
/Users/d/kolibri/010x/content/content
```

to 

```
/Users/d/kolibri/010x/content
```

logs:

```bash
~/P/l/r/k010x> kolibri manage getdirectory
INFO:kolibri.utils.options:Option RUN_MODE in section [Deployment] being overridden by environment variable KOLIBRI_RUN_MODE
INFO:kolibri.core.apps:Running Kolibri with the following settings: kolibri.deployment.default.settings.base
Current content directory is /Users/d/kolibri/010x/content/content

~/P/l/r/k010x> kolibri manage setdirectory /Users/d/kolibri/010x/content/
INFO:kolibri.utils.options:Option RUN_MODE in section [Deployment] being overridden by environment variable KOLIBRI_RUN_MODE
INFO:kolibri.core.apps:Running Kolibri with the following settings: kolibri.deployment.default.settings.base
INFO     Current content directory is /Users/d/kolibri/010x/content/content
INFO:kolibri.content.management.commands.setdirectory:Current content directory is /Users/d/kolibri/010x/content/content
INFO     Migrating the content into /Users/d/kolibri/010x/content
INFO:kolibri.content.management.commands.setdirectory:Migrating the content into /Users/d/kolibri/010x/content
  File "/Users/d/Projects/learning_equality/repos/k010x/.venv/bin/kolibri", line 11, in <module>
  File "/Users/d/Projects/learning_equality/repos/k010x/kolibri/utils/cli.py", line 635, in main
  File "/Users/d/Projects/learning_equality/repos/k010x/kolibri/utils/cli.py", line 421, in manage
  File "/Users/d/Projects/learning_equality/repos/k010x/.venv/lib/python3.6/site-packages/django/core/management/__init__.py", line 364, in execute_from_command_line
    utility.execute()
  File "/Users/d/Projects/learning_equality/repos/k010x/.venv/lib/python3.6/site-packages/django/core/management/__init__.py", line 356, in execute
  File "/Users/d/Projects/learning_equality/repos/k010x/.venv/lib/python3.6/site-packages/django/core/management/base.py", line 283, in run_from_argv
  File "/Users/d/Projects/learning_equality/repos/k010x/.venv/lib/python3.6/site-packages/django/core/management/base.py", line 330, in execute
  File "/Users/d/Projects/learning_equality/repos/k010x/kolibri/content/management/commands/setdirectory.py", line 42, in handle
    self.migrate(content_directory, destination)
  File "/Users/d/Projects/learning_equality/repos/k010x/kolibri/content/management/commands/setdirectory.py", line 78, in migrate
    shutil.copytree(current_directory, dest_directory)
  File "/Users/d/Projects/learning_equality/repos/k010x/.venv/lib/python3.6/shutil.py", line 309, in copytree
    names = os.listdir(src)
FileNotFoundError: [Errno 2] No such file or directory: '/Users/d/kolibri/010x/content/content'

~/P/l/r/k010x> kolibri manage getdirectory
INFO:kolibri.utils.options:Option RUN_MODE in section [Deployment] being overridden by environment variable KOLIBRI_RUN_MODE
INFO:kolibri.core.apps:Running Kolibri with the following settings: kolibri.deployment.default.settings.base
Current content directory /Users/d/kolibri/010x/content/content does not exist.
```

benjaoming:
On tech debt and getting 0.10 released:

* Who's going to use this command? The original issue talks about support feedback.
* Should it be documented, then?
* Are answers to questions going to be floating around with a CLI we know should be changed?

I would suggest finding the right CLI firstly. If other people start using this CLI, we have to make deprecations, release notes etc..

I take it from @indirectlylit's original issue that we are already pretty intended on some kind of `kolibri manage content <command>`-like layout.

indirectlylit:
Agreed - it's better if we don't have to deprecate/rename commands. On the other hand, these commands will be rarely used.

I targeted #3727 to 0.10.0 and we'll definitely do it if folks on the team have bandwidth or shortly thereafter otherwise

indirectlylit:
result of discussion today for this issue:

* remove "get directory" command
* migrating will be `kolibri manage content movedirectory`

kolibri/content/management/commands/content.py
```diff
@@ -0,0 +1,156 @@
...
+        )
+        movedir_subparser.add_argument(
+            'destination',
+            type=str,
+            help='Destination of the new content directory.'
+        )
+
+    def handle(self, *args, **options):
+        if options['command'] == 'movedirectory':
+            self.move_content_directory(options['destination'])
```
benjaoming, line 37: This is nice and clean :)

kolibri/content/management/commands/content.py
```diff
@@ -0,0 +1,156 @@
...
+        # Check exsitence of current directory before running the commands.
+        content_directory = OPTIONS['Paths']['CONTENT_DIR']
+        destination = os.path.abspath(os.path.expanduser(dst))
+
+        if not os.path.exists(content_directory):
+            self.stderr.write(
+                self.style.ERROR(
+                    '\nCurrent content directory {} does not exist.'
+                ).format(content_directory))
+            raise SystemExit()
```
benjaoming: This should be `raise SystemExit(1)`, as well as with the other errors -- otherwise the command exits as if it went well, which makes it hard to call from other scripts.

kolibri/content/management/commands/content.py
```diff
@@ -0,0 +1,156 @@
...
+        shutil.rmtree(databases_src)
+
+        # copy the storage folder only when the source folder contains it
+        if os.path.exists(storage_src):
+            self.copy_content(storage_src, storage_dst)
+            shutil.rmtree(storage_src)
+
+        self.update_config_content_directory(dst)
+
+    def handle_user_input(self, src, dst):
```
benjaoming: This method is a bit too generic in its naming. But you will also be able to significantly reduce the complexity by inverting the question. So instead of asking "do you want to keep x?", try asking:

> Destination already exists: /x/y/z
>
> Do you want to overwrite it completely? (y/N)

...then refactor to match.

Notice that the capital letter is the default answer that occurs by pressing ENTER, this is a convention of some kind.

kolibri/content/management/commands/content.py
```diff
@@ -0,0 +1,156 @@
...
+        """
+        user_answer = input(
+            self.style.WARNING(
+                'The destination {} has content inside.\n'
+                'Do you want to keep the content? (Yes/No)'.format(dst)))
+
+        # If the user does not want to keep the content in the destination,
+        # remove the databases folder and storage folder in the destination
+        # directory
+        if user_answer == 'no' or user_answer == 'No':
```
benjaoming: Easier to write: `user_answer.strip().lower() == "no"`, which also works with `  no  `, `NO` and the even more unlikely `nO` :)

kolibri/utils/cli.py
```diff
@@ -593,6 +594,10 @@ def main(args=None):  # noqa: max-complexity=13
         port = _get_port(arguments['--port'])
         check_other_kolibri_running(port)
 
+        # Check if the content directory exists when Kolibri runs after the first time.
+        if os.path.exists(version_file()):
+            check_content_directory_exists()
```
benjaoming: Instead, can it be added in `_first_run()` ?

https://github.com/lyw07/kolibri/blob/1b37ca8ab4a7b16b009b257a151933ecd53f2174/kolibri/utils/cli.py#L182

kolibri/utils/sanity_checks.py
```diff
@@ -42,3 +44,14 @@ def check_port_availability(host, port):
...
+
+def check_content_directory_exists():
+    content_directory = OPTIONS['Paths']['CONTENT_DIR']
+
+    # Check if the content directory exists
+    if not os.path.exists(content_directory):
+        logger.error(
+            "The content directory {} does not exist.".format(content_directory)
+        )
+        sys.exit(1)
```
benjaoming, line 57: Should this also check if it's writable?

kolibri/content/management/commands/content.py
lyw07: thanks! I will change them

kolibri/content/management/commands/content.py
lyw07: thank you for the information! I will change the name and the question

kolibri/utils/sanity_checks.py
lyw07: yes I will implement that into the function!

kolibri/content/management/commands/content.py
benjaoming: Actually, this could also be: `user_answer.strip().lower() in ["no", "n"]`

kolibri/utils/cli.py
```diff
@@ -636,6 +634,8 @@ def main(args=None):  # noqa: max-complexity=13
         return
 
     if arguments['start']:
+        # Check if the content directory exists when Kolibri runs after the first time.
+        _first_run()
```
benjaoming: Ah, sorry, I have been giving ill-advice, will do a commit directly to your branch to fix it!

lyw07: thank you for the commit!

radinamatic:
Looking good @lyw07! :+1: 

I tested several times moving the content back and forth from the default location on both the system's HD, small 10 years old USB stick, and the VM network location.

In all the test cases the migration went smoothly, Kolibri was able to access the content even after the system restart, and further/subsequent import was successful! Good catch to add the message for stopping the server :wink: 

win10 start running - oracle vm virtualbox_174

One thing that I noticed is when the `databases` and `storage` folders get moved to a new location with already present other folders, it may be confusing to the users afterwards. Since it was deliberate to move just those two folders and not the `.kolibri\content`, I'll add the recommendation in the documentation to maybe first create a new folder on the destination, `KolibriContent` or similar, and **then** run the command, so it becomes easier to find it later.

indirectlylit:
thanks for the thorough testing!

Before merging .. just want to confirm that's the expected behavior @radinamatic ?

benjaoming:
One behavior remains undefined: What happens if _not_ overwriting a dirty destination folder? Will subsequent processes fail if destination file already exists? Will such files be skipped? Is there a resume function? Will contents that are not in the db be detected and imported?

I think it's okay to leave this undefined for now and merge, but we should probably open an issue to further define this.

https://github.com/lyw07/kolibri/blob/8a0bb46b2b6678141781f4ca6e92834a5834aafc/kolibri/content/management/commands/content.py#L104-L128

rtibbles:
Re: @radinamatic's comment, should we add an extra user prompt asking if we should create a 'content' directory where the user said, or just use the directory as the content directory? Or are we overthinking it?!

radinamatic:
> should we add an extra user prompt asking if we should create a 'content' directory

This could improve the user experience, maybe not essential immediately, but on second pass. The majority of users won't even know what is exactly that they are "moving"... Command is called `movedirectory`, but what is moved around are 2 subdirectories, so there's a slight discrepancy. It might not be a problem if the destination drive is empty, but could cause confusion otherwise, especially in case of issues, if they have to check how many databases are there, etc... Unhappy paths, in short :wink:  

If we want to merge it as is, I'll add the step in the docs, and it will be clearer.

radinamatic:
Added [instructions](https://github.com/learningequality/kolibri-docs/pull/28/commits/2f0a0eaccd5ea0601f5f290565fe70d599625901) in the user docs.

benjaoming:
> Or are we overthinking it?!

@rtibbles try reading `man rsync` :)

@lyw07 great work, glad that it got in, thanks for addressing all the feedback!

Issue #3513, opened 2018-04-10
improve predictability of search


indirectlylit:
### Observed behavior

Currently, searching results in some very unexpected behaviors. It seems to be "fuzzy" but in the wrong ways - e.g. it will pick up on completely unrelated content due to spelling mappings, while not finding synonyms of common words.

### Expected behavior

Search should at least be predictable: if something appears in the results, it should be obvious why it's there. This will encourage people to trust and use it.

As a simple starting point, we should remove the behavior which causes results to appear based on complex character mappings.

This will reduce the number of results for each search query, but make what does get returned more likely to be relevant to the user.

### User-facing consequences

Search is confusing and misleading because it often returns completely unrelated content

-----

cc @christianmemije @ralphiee22

indirectlylit:
* one example: https://github.com/learningequality/kolibri/issues/1713
* some related if slightly different feedback: https://github.com/learningequality/clearinghouse/issues/8
* also - feedback that someone searched for "kung fu" and received results about cooking...

PR #3599, opened 2018-04-27
improve search by looking for whole string and parsed string

ralphiee22:
### Summary

This is an attempt to **_really_** simplify the logic for searching. The new logic is to search the whole string "`math is cool`" and the string broken up "`math`", "`is`", "`cool`" in the title and description of content nodes. 
This PR may merit more discussion in the consequences of this strategy, as well as the different strategies that could be implemented. 
One example is that users must now spell things correctly to get adequate search results.

### Reviewer guidance

Download some channels and try out search.

### References

https://github.com/learningequality/kolibri/issues/3513
----

### Contributor Checklist

### Reviewer Checklist

### DIFF
--- a/kolibri/content/api.py
+++ b/kolibri/content/api.py
@@ -113,14 +113,14 @@ class ContentNodeFilter(IdFilter):
         # remove duplicate content items based on content_id
         content_id_set = set()
         id_list = []
-        for id_pair in matches.values('content_id', 'id')[:30]:
+        for id_pair in matches.values('content_id', 'id'):
             if id_pair['content_id'] in content_id_set:
                 pass
             else:
                 content_id_set.add(id_pair['content_id'])
                 id_list.append(id_pair['id'])
 
-        return matches.filter(id__in=id_list)
+        return matches.filter(id__in=id_list[:30])
 
     def filter_recommendations_for(self, queryset, name, value):
         """


indirectlylit:
Cool! Does this implement any of the result relevance-sorting that we discussed?

From what I can tell glancing at the code, ordering is undefined and searching for `math is cool` might return a result with only `math` in the description above a result with `math is cool` in the title.

If possible, I'd imagine the order to be:
* exact full match in title
* exact full match in description
* other partial matches, if possible in similar order

kolibri/content/api.py
```diff
@@ -98,19 +96,31 @@ class Meta:
...
+        # split string up by spaces and search each individual string in title or description
+        str_arr = value.split()
+        for string in str_arr:
+            query.add(Q(title__icontains=string) | Q(description__icontains=string), Q.OR)
+        matches = queryset.filter(query)
+
+        # remove duplicate content items based on content_id
+        content_id_set = set()
+        id_list = []
+        for id_pair in matches.values('content_id', 'id')[:30]:
```
indirectlylit: looks like this will truncate the number of results _before_ de-duplication rather than after, so you'll end up with fewer than 30 results

indirectlylit:
> ordering the results is a bit cumbersome

Here's an untested idea, just brainstorming for fun.

But maybe we could do something like build up the results incrementally. Starting cheaper and more precise, and build up to more complex and broader?

Two strange ideas:
* Grab a somewhat arbitrarily-sized buffer of possible results, with the expectation that some will be dupes. Don't grab too many to prevent them all from being loaded into memory.
* On each iteration, exclude previously-found results in the query

```python
# return the result of and-ing a list of queries
def intersection(queries):
    query = queries.pop()
    for q in queries:
        query &= q
    return query

# return the result of or-ing a list of queries
def union(queries):
    query = queries.pop()
    for q in queries:
        query |= q
    return query

# return up to 30 results
def filter_search(self, queryset, name, value):

    # queries ordered by relevance priority
    all_queries = [
        # exact match in title
        Q(title__icontains=value),
        # exact match in description
        Q(description__icontains=value),
        # all words in title
        intersection([ Q(title__icontains=w) for w in words ]),
        # all words in description
        intersection([ Q(description__icontains=w) for w in words ]),
        # any words in title
        union([ Q(title__icontains=w) for w in words ]),
        # any words in description
        union([ Q(description__icontains=w) for w in words ]),
    ]

    results = []
    content_ids = set()
    MAX_RESULTS = 30
    BUFFER_SIZE = 100 # grab some extras, but not too many

    # iterate over each query type, and build up search results
    for query in all_queries:
        # in each pass, don't take any items already in the result set
        exclusion_query = union([ Q(content_id__is=cid) for cid in content_ids ])
        matches = queryset.filter(query).exclude(exclusion_query)[:BUFFER_SIZE]

        for match in matches.values():
            if match['content_id'] in content_ids:
                continue
            content_ids.add(match['content_id'])
            results.append(match)
            if len(results) >= MAX_RESULTS:
                return results

    return results
```

indirectlylit:
Alternatively, it looks like this is a place where maybe we could do a more basic search in SQLite, and more advanced searches in PostgreSQL. In particular, it looks like postgres could do:

* [distinct results based on content id](https://docs.djangoproject.com/en/2.0/ref/models/querysets/#django.db.models.query.QuerySet.distinct)
* [ranked search based on a list of terms](https://docs.djangoproject.com/en/2.0/ref/contrib/postgres/search/#searchrank)

jonboiser:
I suggested to Ralphie that to get the ranking, you can pass the first result set through a library like Lunr either on the front-end (via lunr.js) or on the back-end (via lunr.py). So 1) do a simple keyword search to get, say 50, result in no particular order, then 2) pass it through a lunr-type library to get the relevance-ordered version.

ralphiee22:
With @indirectlylit's method, it sounds reasonable, but of course the items are not ranked, but could be with @jonboiser's suggestion of using the lunr library. Also, I am unsure of how efficient it would be since we would be running the queries over all the contentnodes in each run.

indirectlylit:
With the method I suggested, the items would be ranked _across_ queries, just not _within_ queries

Agreed, no idea about performance. But the goal was that not all queries would be run all the time - in theory a match could be found sooner and it would bail. But it's true that queries with no matches woul be very expensive...

Issue #3600, opened 2018-04-29
When using postgres, unable to import content nodes with long field values


jamalex:
### Observed behavior

When running Kolibri with a Postgres DB backend, trying to import content that contains a `ContentNode` with `description` or `license_description` longer than 400 chars throws an error (see below).

### Expected behavior

When we run with a SQLite backend, it [happily ignores](https://chrisbartos.com/articles/sqlite-doesnt-care-about-your-max_length/) the violation of the `max_length` attribute on these fields. Postgres is more strict (as it should be).

Expectation would be one of the following:
- Don't allow published databases to have strings > 400 char in these fields to start with. The ship's probably already sailed on that one...
- Truncate these strings during import, to respect the 400 char limit. Simple, but kind of ugly.
- Increase this limit within the Kolibri schema. This is probably preferred...

### User-facing consequences

Prevents use of Postgres backend for many content sources (e.g. KA). We now have users wanting/trying to use Postgres, and we want to do performance testing on the same.

### Errors and logs

```
INFO:kolibri.content.utils.channel_import:Importing ContentNode data
Traceback (most recent call last):
  File ".virtualenvs/kolibri/bin/kolibri", line 11, in <module>
    load_entry_point('kolibri', 'console_scripts', 'kolibri')()
  File "kolibri/utils/cli.py", line 653, in main
    start(port, daemon=daemon)
  File "kolibri/utils/cli.py", line 265, in start
    update()
  File "kolibri/utils/cli.py", line 249, in update
    update_channel_metadata()
  File "kolibri/content/utils/annotation.py", line 38, in update_channel_metadata
    import_channel_from_local_db(channel_id)
  File "kolibri/content/utils/channel_import.py", line 383, in import_channel_from_local_db
    import_manager.import_channel_data()
  File "kolibri/content/utils/channel_import.py", line 238, in import_channel_data
    raise e
sqlalchemy.exc.DataError: (psycopg2.DataError) value too long for type character varying(400)
```

### Steps to reproduce

1. Use a custom settings file (e.g. `pg_settings.py`) with something like:

```
from __future__ import absolute_import, print_function, unicode_literals

from kolibri.deployment.default.settings.base import *  # noqa

DATABASES = {
    'default': {
        'ENGINE': 'django.db.backends.postgresql',
        'USER': 'postgres',
        'PASSWORD': 'postgres',
        'NAME': 'kolibri',
    },
}
```
2. 
3. Run `kolibri manage migrate` to initialize the database.
[...]

PR #3601, opened 2018-04-29
Add content import regression test for #3600, with long descriptions

jamalex:
### Summary

Test for #3600, to ensure the issue gets caught by travis postgres testing.

### Reviewer guidance

No need to review yet, just checking that tests fail with the regression test for now.

### References

Tests #3600.

----

### Contributor Checklist

### Reviewer Checklist


jamalex:
Aaaaaand tests now pass. :dancing_men:

jamalex:
Going to add some notes here for want of a better place for now (CC @bonidjukic @jredrejo), with some comparison stats.

jamalex:
Comparison when running `kolibri manage generateuserdata`:

SQLite:
    real    35m51.222s
    user    23m3.964s
    sys     7m54.384s


Postgres:
    real    28m28.898s
    user    8m35.400s
    sys     0m13.916s

jamalex:
Querying `/api/channel/?file_sizes=true`:

SQLite: 12.357s
Postgres: 5.907s

jamalex:
Running `kolibri manage importcontent network 1ceff53605e55bef987d88e0908658c5` (with actual downloading disabled, just the prep portion):

SQLite: 3.9s
Postgres: 1.9s

jamalex:
So, in general, for read operations, it seems like Postgres may be around twice as fast. More testing needed to determine write speed diffs (the `generateuserdata` test is quite CPU heavy, rather than just DB, so it's likely not a great representation).

jamalex:
Did a quick (non-concurrent) write test:

```
import uuid
import datetime
from kolibri.logger.models import *
u = FacilityUser.objects.all()[0]

%timeit ContentSummaryLog.objects.create(user=u, channel_id=uuid.uuid4().hex, content_id=uuid.uuid4().hex, start_timestamp=datetime.datetime.now())
```

Times were very similar (around 5-6ms). Could be that it's a simple operation, and hence much of the time is eaten up by non-DB stuff. Or also the fact that we're not pushing concurrency. More testing needed.

kolibri/content/test/test_channel_import.py
```diff
@@ -319,45 +320,39 @@ def create_channel():
 
 DATA_PATH_TEMPLATE = os.path.join(os.path.dirname(__file__), '../fixtures/{name}_content_data.json')
 
-class NaiveImportTestCase(ContentNodeTestBase, TransactionTestCase):
-    """
-    Integration test for naive import - this is run using a TransactionTestCase,
+
+class ContentImportTestBase(TransactionTestCase):
```
rtibbles, line 324: Because this is an instance of `TestCase`, I am pretty sure this is going to get executed in the test suite, but I'm not really sure what it is testing, as it has neither a name, nor a legacy schema associated with it.

This should inherit from object, and then you will have to mix in this, ContentNodeTestBase, and TransactionTestCase into NaiveContentImport.

rtibbles: I think this should make everything else work out how I have intended, with regards to inheritance for prior versions also.

rtibbles: Oh, no, I am wrong. This is fine. Merging.

PR #3607, opened 2018-05-01
First pass at ini-based options file, for persistent user-accessible configuration

jamalex:
### Summary

Prototype of new options.ini file-based configuration.

### Reviewer guidance

In response to all the feedback, this PR became a bit of a rabbit hole (and also triggered a bunch of side-effect linting that clutters it up). It *could* be useful to look at it commit-by-commit, except that it changed a lot from the first few commits to later.

### References

As discussed in https://docs.google.com/document/d/1v-_B7dFlEw9EAOfMSU78IkfhMX9DihzN7jjjxzEUfBk/edit#heading=h.otycu89b2guh

----

### Contributor Checklist

### Reviewer Checklist


kolibri/utils/options.py
```diff
@@ -0,0 +1,103 @@
...
+
+
+def update_env_from_options_file(ini_path):
+
+    conf = read_options_file(ini_path)
+
+    for section in conf.sections:
+        for key, val in conf[section].items():
+            # read each setting into environment variable, prefixed with KOLIBRI_, if not already defined
+            os.environ.setdefault("KOLIBRI_{key}".format(key=key), str(val))
```
jamalex: I coded this based on what we discussed, but I do wonder whether it actually makes sense shoehorning the values into the env vars only to pluck them right back out again. Might make more sense to read the env vars and use them to override the values in a `get_options` method, and then not modifying the env at all?

kolibri/utils/options.py
```diff
@@ -0,0 +1,103 @@
...
+    conf = ConfigObj(ini_path, configspec=configspec, write_empty_values=True)
+
+    validation = conf.validate(Validator(), preserve_errors=True)
+
+    # loop over any errors with config values, and clean up
+    for section_list, key, error in flatten_errors(conf, validation):
+        if error is False:
+            error = 'missing value or section.'
+        if key is not None:
+            # remove the value that didn't validate, to ignore it (rather than bailing)
```
jamalex: The alternative (bailing) may be preferable.

jamalex: (Current behavior: output an error message, and then fall back to the default value, but keep running)

kolibri/utils/options.py
```diff
@@ -0,0 +1,103 @@
...
+    "DATABASE_USER = string(default='')",
+    "DATABASE_HOST = string(default='localhost')",
+
+    "[Paths]",  # these are defined relative to KOLIBRI_HOME (or absolute)
+    "CONTENT_ROOT = string(default=content)",
+
+    "[Deployment]",
+    "LISTEN_PORT = integer(default=8080)",
+
+]
```
jamalex: This is configobj's standard "configspec" method for doing validation. See http://configobj.readthedocs.io/en/v5.0.1/configobj.html#validate

kolibri/utils/options.py
```diff
@@ -0,0 +1,103 @@
...
+    "DATABASE_NAME = string(default='')",
+    "DATABASE_PASSWORD = string(default='')",
+    "DATABASE_USER = string(default='')",
+    "DATABASE_HOST = string(default='localhost')",
+
+    "[Paths]",  # these are defined relative to KOLIBRI_HOME (or absolute)
+    "CONTENT_ROOT = string(default=content)",
+
+    "[Deployment]",
+    "LISTEN_PORT = integer(default=8080)",
```
indirectlylit: maybe `HTTP_PORT`? (wondering if there may eventually be other ports like websockets or https)

kolibri/utils/options.py
```diff
@@ -0,0 +1,103 @@
...
+
+    "[Database]",
+    "DATABASE_ENGINE = option('sqlite', 'postgres', default='sqlite')",
+    "DATABASE_NAME = string(default='')",
+    "DATABASE_PASSWORD = string(default='')",
+    "DATABASE_USER = string(default='')",
+    "DATABASE_HOST = string(default='localhost')",
+
+    "[Paths]",  # these are defined relative to KOLIBRI_HOME (or absolute)
+    "CONTENT_ROOT = string(default=content)",
```
indirectlylit: we could choose a conventional way to name variables that hold paths. E.g. `KOLIBRI_HOME_DIR`, `KOLIBRI_CONTENT_DIR`, etc

kolibri/utils/options.py
indirectlylit: Bailing might make issues easier to debug

kolibri/utils/options.py
indirectlylit: that sounds good to me if the resulting behavior is equivalent

kolibri/utils/options.py
jamalex: This is for direct backwards compatibility to support `KOLIBRI_LISTEN_PORT`. We *could* add a new primary name for this, but then we would need to code in legacy support and figure out precedence rules if both env vars were defined, etc. Worth it? :man_shrugging:

kolibri/utils/options.py
jamalex: Should be equivalent from the end-user's perspective, I think?

kolibri/utils/options.py
jamalex: Yeah, it's a tradeoff, I guess -- if daemonized, could be hard to understand why the server stopped loading (e.g. after an upgrade -- although presumably we would want to always maintain back-compatibility with old ini files). I agree that bailing is probably "safest", though. Will wait for more opinions.

kolibri/utils/options.py
jamalex: Yeah, happy to change this name if desired. If we change `KOLIBRI_HOME` to `KOLIBRI_HOME_DIR`, though, it could be a headache in terms of maintaining support for both the old name and the new name?

(The `_DIR` convention is something we've been using in our Django settings files -- the `_ROOT` suffix is a Django convention, e.g. for `STATIC_ROOT` and `MEDIA_ROOT`, and indicates it's a base directory under which there are subdirectories with their own structure)

indirectlylit:
Regarding variable naming, this is probably the best chance we have to establish a consistent settings API.

If the cost is a few one-time special cases to handle existing environment variables, IMO it's worth it.

jamalex:
> Regarding variable naming, this is probably the best chance we have to establish a consistent settings API.
> 
> If the cost is a few one-time special cases to handle existing environment variables, IMO it's worth it.

I would personally love to see `KOLIBRI_HOME` renamed to `KOLIBRI_DATA_DIR`, as it's more representative of what goes there (what makes a home a home?), and because it matches the KOLIBRI_DATA directory name we use when we export/import to a drive. Just thinking through the implications, however, so we're not assuming it's going to be simple:
- There are dozens of references to `KOLIBRI_HOME` across the codebase, and also in plugins. 
- Some references are to the envvar, others to `conf.KOLIBRI_HOME`, others to `settings.KOLIBRI_HOME`.
- Maintaining backwards compatibility means maintaining the alias in all 3 contexts. So we'd look at both `os.environ["KOLIBRI_HOME"]` and `os.environ["KOLIBRI_DATA_DIR"]`, support both `settings.KOLIBRI_HOME` and `settings.KOLIBRI_DATA_DIR`, etc. Dev confusion can be minimized by comments in the code, but it does add some complexity.
- But yeah, better sooner than later if we're going to do it. Can you remind me *why* we're doing it again? Because now I want to do it, but I'm not sure why. :laughing:

jamalex:
Did a quick scan of all the env vars we reference. Others include:

- Custom build stuff: `BUILD_TIME_PLUGINS`, `RUN_TIME_PLUGINS`, `DEFAULT_SETTINGS_MODULE`, and `EXTRA_REQUIREMENTS` (probably don't need to be considered here)
- Externally defined vars: `READTHEDOCS` and `TOX_ENV` (can't be changed, as we don't set them)
- Specifying Studio URL: `CENTRAL_CONTENT_DOWNLOAD_BASE_URL` (would probably be good to standardize with a `KOLIBRI_` prefix, and include in the options.ini)

kolibri/utils/options.py
jamalex: Going to bail here, yeah. Not bailing in the "extra values" case below, though, so we don't need to worry about deprecating old settings. They'll still get a warning.

kolibri/deployment/default/settings/base.py
```diff
@@ -126,25 +128,38 @@
...
+            'NAME': OPTIONS['DATABASE_NAME'],
+            'PASSWORD': OPTIONS.get('DATABASE_PASSWORD') or '',
+            'USER': OPTIONS.get('DATABASE_USER') or '',
+            'HOST': OPTIONS.get('DATABASE_HOST') or '',
+        }
+    }
 
 # Content directories and URLs for channel metadata and content files
 
+_content_root = os.path.join(KOLIBRI_HOME, OPTIONS.get('CONTENT_ROOT') or 'content')
```
lyw07: Sorry if I'm not getting it right - Is the CONTENT_ROOT variable specifically for the `contentdirectory` command?

kolibri/deployment/default/settings/base.py
```diff
@@ -39,6 +39,8 @@
     os.path.join(KOLIBRI_MODULE_PATH, "locale"),
 ]
 
+OPTIONS = options.get_options_from_env()
```
benjaoming: I think all of this logic should be concealed in `kolibri.utils.conf` -- because this module is expected to be loaded before the Django settings stack and can therefore define settings in one single place.

So for instance:

```
from kolibri.utils import conf

DATABASES = conf.global_settings.DATABASES
```

Another thing: `conf.global_settings` can be made a completely read-only instance after instantiating to guarantee against wrong use.

A friend of mine just tipped me that even for rigorous cases, they would replace `os.environ` with a read-only dict. It's maybe not nice to do against other external environments, but for testing purposes it could be interesting to avoid any changes beyond loading the environment in `kolibri.utils.conf`

kolibri/deployment/default/settings/base.py
benjaoming: In the above line, there are defaults being specified outside of `kolibri.utils.conf` -- perhaps better to just specify in one place.

kolibri/deployment/default/settings/base.py
```diff
@@ -126,25 +128,38 @@
...
-CONTENT_DATABASE_DIR = os.path.join(KOLIBRI_HOME, 'content', 'databases')
+CONTENT_DATABASE_DIR = os.path.join(_content_root, 'databases')
 if not os.path.exists(CONTENT_DATABASE_DIR):
     os.makedirs(CONTENT_DATABASE_DIR)
 
 # Directory and URL for storing de-duped content files for all channels
-CONTENT_STORAGE_DIR = os.path.join(KOLIBRI_HOME, 'content', 'storage')
+CONTENT_STORAGE_DIR = os.path.join(_content_root, 'storage')
 if not os.path.exists(CONTENT_STORAGE_DIR):
     os.makedirs(CONTENT_STORAGE_DIR)
```
benjaoming: These settings remind of our "derived settings" which caused issues in KA Lite. It's very nice to define them in a non-configurable way, but having them sit in the django settings module is perhaps not the right place, since it implies that they can be configured? Maybe they can? But should they?

requirements/base.txt
```diff
@@ -4,6 +4,7 @@ djangorestframework==3.7.7
 django==1.11.10  # pyup: >=1.11,<2
 six==1.11.0
 colorlog==3.1.2
+configobj==5.0.6
```
benjaoming: Initially, in the .ini proposal, it was partly also because we wouldn't have to add dependencies IIRC :)

kolibri/deployment/default/settings/base.py
jamalex: >I think all of this logic should be concealed in kolibri.utils.conf

I avoided using that module itself because it's cluttered with a lot of top-level stuff very specific to parsing `kolibri_settings.json`, so putting this new logic in there would be really messy.

>avoid any changes beyond loading the environment in kolibri.utils.conf

I'm almost done with a new version of this PR that doesn't modify `os.environ` at all (just looks to it for overrides -- everything will be funneled through the strict conf file config as you suggested). Stay tuned.

kolibri/deployment/default/settings/base.py
jamalex: >Is the CONTENT_ROOT variable specifically for the contentdirectory command

Partially, but not solely. It's the option that the user can modify to change where Kolibri looks for content. It will also be used by your management command.

>In the above line, there are defaults being specified outside of kolibri.utils.conf -- perhaps better to just specify in one place.

In an update to this PR that I'm going to push shortly, (almost) all defaults and validation etc will be done in the `options` module.

kolibri/deployment/default/settings/base.py
jamalex: >These settings remind of our "derived settings" which caused issues in KA Lite.

Critical difference here: I prefixed `_content_root` with an underscore so that it's not exported -- i.e. it's not a Django setting, and can't be overridden in Django settings files, so we don't end up with the "I changed this and the derived stuff didn't change" confusion.

>It's very nice to define them in a non-configurable way, but having them sit in the django settings module is perhaps not the right place, since it implies that they can be configured? Maybe they can? But should they?

We will not be recommending to regular end users that they ever use Django settings to configure anything. `CONTENT_STORAGE_DIR` and `CONTENT_DATABASE_DIR` are for a specific "power user" case that is already being utilized by us on our demo servers -- having a single shared `CONTENT_STORAGE_DIR` across demo instances, but separate `CONTENT_DATABASE_DIR` directories, so that we can leverage deduplicated content storage while showing different (but sometimes overlapping) content across demo servers.

kolibri/utils/options.py
jamalex: Update to PR will have a clean way to handle legacy envvars.

requirements/base.txt
jamalex: The problem was that `configparser` isn't able to read/write config files while preserving order, comments, etc, while `configobj` can (as well as having nicer validation logic, etc).

We want to be able to write out options.ini templates that include comments, and modify existing files to update the content directory setting for @lyw07's PR.

jamalex:
Update on current status:

- [Updated the overall design](https://github.com/learningequality/kolibri/pull/3607/commits/843ddc7469edc8067abc665f048427e162307d89) so that it doesn't modify env vars at all, just reads from them to feed into overriding options (prior to validation logic) in a systematic way. We'd still recommend to most users that they use the ini file.
- [Moved](https://github.com/learningequality/kolibri/pull/3607/commits/f53c32e790802770f4d9cb791860cf47472c9a8d) `settings.CENTRAL_CONTENT_DOWNLOAD_BASE_URL` into a config option (before, it was an env var that directly went into a Django setting). This is part of an overall philosophical push (that I'm putting out there for feedback) that Django settings files should primarily be used for Django settings, not as a place for us to funnel all our config options through. Throughout Kolibri, we can directly reference the standardized (and validated) config options via `kolibri.utils.conf.OPTIONS`.

An additional thing I'm thinking could be good:

- Right now we hardcode the `SECRET_KEY` into the base Django settings, and since most users won't be specifying a custom settings module and overriding this value, it may be safest for us to add an ini option for this, and write a random value for the option to the ini file (if it doesn't have one defined in there already).

kolibri/deployment/default/settings/base.py
```diff
@@ -126,32 +125,41 @@
...
-    'default': {
-        'ENGINE': 'django.db.backends.sqlite3',
-        'NAME': os.path.join(KOLIBRI_HOME, 'db.sqlite3'),
-        'OPTIONS': {
-            'timeout': 100,
+if conf.OPTIONS['Database']["DATABASE_ENGINE"] == "sqlite":
+    DATABASES = {
+        'default': {
+            'ENGINE': 'django.db.backends.sqlite3',
+            'NAME': os.path.join(KOLIBRI_HOME, conf.OPTIONS['Database']['DATABASE_NAME'] or 'db.sqlite3'),
```
jamalex: Since `DATABASE_NAME` is used for both sqlite and postgres, couldn't cleanly specify the default filename in the options, so I'm leaving it here for now.

kolibri/content/management/commands/importchannel.py
```diff
@@ -34,7 +36,7 @@ def add_arguments(self, parser):
             help="Download the database for the given channel_id."
         )
 
-        default_studio_url = settings.CENTRAL_CONTENT_DOWNLOAD_BASE_URL
+        default_studio_url = conf.OPTIONS['Urls']['CENTRAL_CONTENT_BASE_URL']
```
indirectlylit, line 50: A few questions and comments

1. In the future if we're allowing people to manage a list of these values from the web UI, would there still be one default value coming from settings? Or would we eventually deprecate this setting? Or... do we move eventually to a setup where the _.ini_ file is actually updated based on modifications to device settings from the web UI?
2. We don't really use the term 'central' [in the UI or documentation](http://kolibri.readthedocs.io/en/latest/manage.html#manage-content). We call it 'Studio', though maybe we don't want to do that here? something like "content import server" might be more aligned with existing terminology
3. Any benefit to making this two settings (hostname and protocol), and losing the potentially ambiguous "base url" framing?

jamalex: >In the future if we're allowing people to manage a list of these values from the web UI, would there still be one default value coming from settings? Or would we eventually deprecate this setting? Or... do we move eventually to a setup where the .ini file is actually updated based on modifications to device settings from the web UI?

My thinking: there's always still the single central "hard-coded" (ini-coded) one. The user-provided ones are stored in the database.

>We don't really use the term 'central' in the UI or documentation. We call it 'Studio', though maybe we don't want to do that here? something like "content import server" might be more aligned with existing terminology

Don't see a perfect answer, yeah. It won't always be Studio, could be e.g. the country-level NITA server in Uganda. Don't really want to encourage the mindset of "many Studios", either. But I don't imagine this will be a heavily utilized option for end users anyway.

>Any benefit to making this two settings (hostname and protocol), and losing the potentially ambiguous "base url" framing?

We did this in KA Lite and I've regretted it ever since -- became a big headache to maintain. Beyond those two components, there's also port, path, etc. We're using "base url" phrasing in Morango as well.

jamalex:
:mag: :mag_right: :mag: :mag_right: :mag: :mag_right: :mag: :mag_right: 
:mag: :mag_right: Added tests!!  :mag: :mag_right: 
:mag: :mag_right: :mag: :mag_right: :mag: :mag_right: :mag: :mag_right:

kolibri/deployment/default/settings/base.py
rtibbles: I guess, could we define these in `conf` instead, as you have with the others? I think this is a nice pattern. I think we can stomach the one time cost of updating our demo servers to use .ini files, no?

rtibbles:
Looking good - I think I would like us to complete the migration of non-Django configuration out of the settings file, and also I assume this needs functions for writing to the config file too! (As well as writing out a config file if it doesn't exist already)

kolibri/deployment/default/settings/base.py
jamalex: A couple thoughts about adding those to the options.ini (if that's what you mean):

- There isn't really any end-user use case for configuring them independently, so it seems more likely to lead to confusion than to help. I was thinking we could have a flag in the options spec dictionary called "show_in_docs" that could be set to `False` for options like this we want to keep hidden (want that for `RUN_MODE` in any case, which we specifically don't want end users messing with).
- More importantly, if we *did* move these to the ini file, we would then face the "derived options" issue, where the default for one option depends on the values of another option. Having the primary content dir specified in the ini, and the derived sub-division into storage/database specified in the settings file, avoids this issue and has a clear flow of precedence.

jamalex:
> I assume this needs functions for writing to the config file too! (As well as writing out a config file if it doesn't exist already)

"Need" is a strong word! :) -- it works fine without it, but it could be good? And yeah, we'll need a writing function before it can be used for @lyw07's PR. Main question in terms of writing out a template is whether we leave the fields empty (and lean on the internally specified defaults, which has the advantage of adapting in case those defaults change across versions), or write the default values into the template file. A middle-ground could be writing a comment above the option line saying what the default is (but then this still won't get updated either when the software updates).

> I would like us to complete the migration of non-Django configuration out of the settings file

There's a lot of grey area there, so it would be good to enumerate which ones you'd like to migrate before initial merge. My first take below:

Definitely good to migrate now:
- SECRET_KEY

Would definitely be possible to migrate, though very low value to end users:
- QUEUE_JOB_STORAGE_PATH

Any others you think would be good?

jamalex:
Another high-level question I'm still not so sure on is how we select and name the "sections". Should it be like it is now, by "kind", with various path settings in one section, or should content-related stuff of various formats be grouped together into one section?

kolibri/deployment/default/settings/base.py
rtibbles: Nope, didn't want to expose them in the ini file, just thinking it is good to not give multiple paths to define variables. As these are are purely Kolibri config, they don't need to be in django settings at all, and can be defined in conf based on the value you are using here.

jamalex: Ah, but how will they then be overridden independently, for the use case described earlier in this comment thread?

>CONTENT_STORAGE_DIR and CONTENT_DATABASE_DIR are for a specific "power user" case that is already being utilized by us on our demo servers -- having a single shared CONTENT_STORAGE_DIR across demo instances, but separate CONTENT_DATABASE_DIR directories, so that we can leverage deduplicated content storage while showing different (but sometimes overlapping) content across demo servers.

jamalex:
Updates:

- Now has a function called `update_options_file` for changing (or adding) option values in the file.
- Held off on any sort of overall "template" creation for now, since I'm unsure about what it should contain, as described above.
- Didn't move SECRET_KEY in here for now, since it really doesn't need to be end-user configurable (just random).

kolibri/deployment/default/settings/base.py
jamalex: I think the main thing I'm unclear on is what "define these in conf instead" means. Could you clarify the proposal?

rtibbles: Oh, well, the use case you describe is no really needed, right? It doesn't matter if different demo servers share a data folder, because they don't draw from that folder (the only time this is not the case is on upgrade, but we can  make this behaviour conditional on upgrade from a very old version).

So, let's not allow them to be configured separately, because there is is no need, and just define these derived values in the conf module?

rtibbles: The other option is just to keep the database folder completely empty.

jamalex: >we can make this behaviour conditional on upgrade from a very old version

Not quite sure what you mean -- do you mean modify the `.data_version` to trick it into thinking it doesn't need to run the upgrade code even after we upgrade? What if we have other upgrade logic that *should* be run after upgrade?

>The other option is just to keep the database folder completely empty.

Like... with a cron job that runs every 30 mins or something...? We want to be able to update the content on these demo servers regularly (via the UI), and they download stuff into that folder.

jamalex: > we can make this behaviour conditional on upgrade from a very old version

Oh, I see what you mean -- change the checking logic so it only happens when the old version is < some cutoff.

jamalex: I think we can actually solve this on demo servers in a totally different way -- not use settings at all, and just symlink the content/storage directories to a shared location -- I think we actually started using that strategy, in some cases, and it should be sufficient.

So yeah, I can move these out of the settings. In terms of where to put it, I still feel weird putting too much at the top level of `kolibri.utils.conf`, because [that code](https://github.com/learningequality/kolibri/blob/develop/kolibri/utils/conf.py) is all very oriented around module-level variables and functions related to the `kolibri_settings.json` stuff. I'm thinking to move that stuff to its own module, and have the conf module be the place all this stuff is bound together and exposed. May try a pass at that now.

jamalex: Actually, I think the best place to put this will be in https://github.com/learningequality/kolibri/blob/develop/kolibri/content/utils/paths.py

(And then I'll hold off on a big refactor of where things in conf live, until we revisit `kolibri_settings.json` more broadly -- including a rename of that file!)

jamalex:
Per discussion thread, removed the CONTENT_STORAGE_DIR and CONTENT_DATABASE_DIR settings altogether, and now all references needing to use these paths access them via the functions under `kolibri.content.utils.paths`, which in turn draw from `conf.OPTIONS["Paths"]["CONTENT_DIR"]`.

Anything else?

kolibri/deployment/default/settings/base.py
benjaoming: Yes indeed - have always used symlinks for such tasks... the reason for the necessity of this setting is AFAIK just Windows file systems, which don't support symlinks.

Maybe we should also check if our disk space measurement resolves when subdirs are symlinked.

benjaoming: @jamalex 

> Critical difference here: I prefixed `_content_root` with an underscore so that it's not exported -- i.e. it's not a Django setting, and can't be overridden in Django settings files, so we don't end up with the "I changed this and the derived stuff didn't change" confusion.

Because `CONTENT_STORAGE_DIR` (not sure if still the name) is a setting that's derived from `_content_root`, the user can still pick a value for `CONTENT_STORAGE_DIR` that's inconsistent with `_content_root`.

So then the problem becomes "the user changed something that was derived from `_content_root`, which we thought was an un-changeable internal value" :)

It becomes a lot better if the variable is renamed `_default_content_root_prefix` :)

rtibbles:
Just need to manually test it, but this looks good to go.

kolibri/utils/tests/test_cli.py
```diff
@@ -168,8 +171,12 @@ def start_mock(port, *args, **kwargs):
...
 
         test_port = 1234
-        # ENV VARS are always a string
-        os.environ['KOLIBRI_LISTEN_PORT'] = str(test_port)
+
+        os.environ['KOLIBRI_HTTP_PORT'] = str(test_port)
+
+        # force a reload of conf.OPTIONS so the environment variable will be read in
+        from kolibri.utils import conf
+        conf.OPTIONS.update(options.read_options_file(settings.KOLIBRI_HOME))
```
rtibbles: Wait, why are you reading KOLIBRI_HOME from settings? Should read it from conf!

(We should also just take it out of the Django settings, but you didn't do that).

rtibbles:
CONTENT_DIR also works.

benjaoming:
Great leap forward!

Should we open an issue to document it over in kolibri-docs?

rtibbles:
Done!

kolibri/deployment/default/settings/base.py
jamalex: >Yes indeed - have always used symlinks for such tasks... the reason for the necessity of this setting is AFAIK just Windows file systems, which don't support symlinks.

Our demo servers (only case where content storage and database dirs need to be different from each other) all run on Linux.

>Maybe we should also check if our disk space measurement resolves when subdirs are symlinked.

That's based on a database query, not a disk scan.

>CONTENT_STORAGE_DIR (not sure if still the name)

No longer exists (see final version of this PR).

>It becomes a lot better if the variable is renamed _default_content_root_prefix :)

Also no longer exists. No content directories in settings module at all anymore.

PR #3608, opened 2018-05-02
Refactor coach/exams actions

jonboiser:
Some refactors to cleanup and remove code

1. Combine all the one-off translators for snackbars messages into one object
1. Use `id` instead of `pk` when requesting ContentNodes (so we don't rename pk -> id anymore)
1. Consolidate the almost-the-same activate/deactivateExam actions
1. Move in-cache exam updater to a mutation
1. Mechanical refactors
1. remove unused `previewExam` function

### DIFF
--- a/kolibri/plugins/coach/assets/src/state/actions/exam.js
+++ b/kolibri/plugins/coach/assets/src/state/actions/exam.js
@@ -107,13 +107,11 @@ export function showExamsPage(store, classId) {
   return ConditionalPromise.all(promises).only(
     samePageCheckGenerator(store),
     ([exams]) => {
-      const pageState = {
+      store.dispatch('SET_PAGE_STATE', {
         exams: _examsState(exams),
         examsModalSet: false,
         busy: false,
-      };
-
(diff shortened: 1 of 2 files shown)


kolibri/plugins/coach/assets/src/state/actions/exam.js
```diff
@@ -143,50 +123,28 @@ export function setExamsModal(store, modalName) {
...
-
-        store.dispatch('SET_EXAMS', exams);
+        store.dispatch('SET_EXAM_STATUS', { examId, isActive });
         setExamsModal(store, false);
-
         createSnackbar(store, {
-          text: createTranslator('examActivateSnackbar', {
-            examIsNowActive: 'Exam is now active',
-          }).$tr('examIsNowActive'),
+          text: snackbarTranslator.$tr(isActive ? 'examIsNowActive' : 'examIsNowInactive'),
```
indirectlylit, line 134: will this work as expected? For some reason I thought we couldn't use expressions inside of `$tr` functions (cc @rtibbles)

PR #3612, opened 2018-05-04
Use public facility endpoint for fullfacilitysync

ralphiee22:
### Summary

Use the public facility endpoint. It was using the internal facility endpoint which serialized the dataset object's attributes. This broke the api call that the management command `fullfacilitysync` calls. We update the endpoint because we do not serialize the facility dataset object and only return the ID.

### Reviewer guidance

Use the mgmnt command! 
cc: @jamalex and I tested this change though and it works.

----

### Contributor Checklist

### Reviewer Checklist

### DIFF
--- a/kolibri/auth/management/commands/fullfacilitysync.py
+++ b/kolibri/auth/management/commands/fullfacilitysync.py
@@ -5,6 +5,7 @@ import requests
 from django.core.exceptions import ValidationError
 from django.core.management import call_command
 from django.core.validators import URLValidator
+from django.urls import reverse
 from django.utils.six.moves import input
 from morango.certificates import Certificate
 from morango.certificates import Filter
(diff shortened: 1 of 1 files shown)


kolibri/auth/management/commands/fullfacilitysync.py
```diff
@@ -30,7 +34,7 @@ def add_arguments(self, parser):
 
     def get_dataset_id(self, base_url, dataset_id):
         # get list of facilities and if more than 1, display all choices to user
-        facility_url = urljoin(base_url, 'api/facility/')
+        facility_url = urljoin(base_url, 'api/public/v1/facility/')
```
rtibbles: Can we not do a `reverse` on the url here?

ralphiee22: this is true!

jamalex: If there's anywhere we might *not* want to use URL reversal, it's on public API endpoints, since we specifically need to ensure they don't change, as the client and server may be running different versions. Doesn't hurt to use it here, but in tests we should just hardcode (I think that's what we do in Morango tests).

PR #3619, opened 2018-05-06
Add experimental management command to spin up Kolibri server via nginx + uwsgi

jamalex:
### Summary

[uWSGI](https://uwsgi-docs.readthedocs.io/en/latest/) is a more efficient alternative to our built-in pure-Python cherrypy wsgi server. It also links directly into Nginx, via sockets (not HTTP proxying), which makes things even faster. This PR adds a management command (`kolibri manage startnginx`) that does the following:

- Creates an Nginx config based on how Kolibri is configured (paths, port, etc)
- Starts Nginx (or restarts it if it was already running)
- Starts uWSGI (using a bundled config file that corresponds to the Nginx config)
- Keeps these running until the user interrupts uWSGI with CTRL-C.
- Cleans up the Nginx config, and stops (or restarts) Nginx appropriately.

### Reviewer guidance

Caveats:

- Won't work on Windows
- Will error out if the port is already taken
- Doesn't currently have a "daemonized" mode

For now, I would consider this experimental (primarily for testing and performance benchmarking purposes -- CC @bonidjukic @jredrejo ), though it seems to be working well for me here. 

### References

…

----

### Contributor Checklist

### Reviewer Checklist


jamalex:
On a simple Django template page, transaction rate approximately doubles with 100 clients.

With CherryPy:
```
jamalex@ace:~$ siege -t 20S -c 100 http://127.0.0.1:8080/user/
** SIEGE 3.0.8
** Preparing 100 concurrent users for battle.
The server is now under siege...
Lifting the server siege..      done.

Transactions:		         811 hits
Availability:		      100.00 %
Elapsed time:		       19.24 secs
Data transferred:	       29.08 MB
Response time:		        2.07 secs
Transaction rate:	       42.15 trans/sec
Throughput:		        1.51 MB/sec
Concurrency:		       87.25
Successful transactions:         811
Longest transaction:	       17.01
Shortest transaction:	        0.11
```

With uWSGI:
```
jamalex@ace:~$ siege -t 20S -c 100 http://127.0.0.1:8080/user/
** SIEGE 3.0.8
** Preparing 100 concurrent users for battle.
The server is now under siege...
Lifting the server siege..      done.

Transactions:		        1675 hits
Availability:		      100.00 %
Elapsed time:		       19.21 secs
Data transferred:	       60.07 MB
Response time:		        1.11 secs
Transaction rate:	       87.19 trans/sec
Throughput:		        3.13 MB/sec
Concurrency:		       96.65
Successful transactions:        1675
Longest transaction:	        4.48
Shortest transaction:	        0.14
```

kolibri/deployment/management/commands/startnginx.py
```diff
@@ -0,0 +1,154 @@
...
+from django.conf import settings
+from django.core.management.base import BaseCommand
+from django.core.management.base import CommandError
+
+try:
+    import subprocess
+except ImportError:
+    subprocess = None
+
+# from kolibri.utils.conf import OPTIONS
```
jamalex: Once https://github.com/learningequality/kolibri/pull/3607 is merged, I'll update this to use the `conf.OPTIONS` value to retrieve the port.

kolibri/deployment/default/wsgi.py
```diff
@@ -6,11 +6,10 @@
...
 """
-
-import os
-
 from django.core.wsgi import get_wsgi_application
 
-os.environ.setdefault("DJANGO_SETTINGS_MODULE", "kolibri.deployment.default.settings.base")
+from kolibri.utils.cli import initialize
+
+initialize()
```
jamalex, line 13: Feels maybe a bit weird importing something from cli to use here, but didn't want to do a huge refactor of that code at this stage. Plus, it does what we want!

benjaoming:
Hi @jamalex 

I have to disagree with the scope of this, if it's intended for releasing: We should not be introducing this type of code for benchmarking and performance tests.

My main concern is that it invite users to use it. But more-over, that this code is going to get introduced, anyways.

It is system-specific code, that needs to be aligned and maintained for whatever system the deployment is for. Configuration files are also subject for further user modification.

The way that the debian packages work is to create and manage these configuration files in `/etc`, sensible to package updates etc.. and it can change system files from other packages through "diversions".

This implementation requires system-specific `sudo` escalations inside our own app, based on assumptions of an out-side environment.

For instance, using `service kolibri <command>` is not compliant with system.d.

For quick benchmarking purposes, it seems fine... but we shouldn't be releasing it IMO.

jamalex:
> My main concern is that it invite users to use it.

Short-term, so we're not exposing an interim solution to end users, I'm wondering whether I should move this management command into a plugin (that's disabled by default).

> this functionality is going to get introduced, anyways, in .deb pkgs -- so the real benchmarking could be done based on this

Do we have a timeline for that? From our past experiences, the Debian work seems like it can take months, and lead to rabbitholes. We do want something easy in the interim.

> The way that the debian packages work is to create and manage these configuration files in /etc

Would love to learn more about how you will be approaching managing the contents of these config files dynamically. There are several values in the nginx config that depend on internal mutable Kolibri options (see also #3607), including:
- The port to listen on
- The static file directory
- The content directory

The idea with this PR (in terms of the direction it pushes, not what it accomplishes in its current form) is having an easy command that relies only on nginx and uwsgi being installed, and allows you to dynamically spin up and shut down a Kolibri server in a way that parallels `kolibri start` -- streamlining all the pieces, taking into account state & variables, conducting initialization and sanity checks, and managing shutdown and cleanup. Right now this is done in a fairly platform dependent way, but it could be extended to support a broader set of platforms.

benjaoming:
@jamalex it doesn't have to live in a plugin per say.. I don't think a management command needs more than to be a Django application.

Could for instance live in an app project, `kolibri-benchmark-tools` ?

benjaoming:
@jamalex re: debian pkg - yes, it takes a lot of time, especially because of supporting various versions of system.d/upstart etc, and even Nginx and Uwsgi and Python versions etc.

So I wouldn't be optimistic about the timeline, but one way to push it out faster, would be to limit the support to the main supported platforms... for instance, Ubuntu 16.04 and Raspbian.

kollivier:
What about adding a 'There Be Dragons Here' or YMMV warning to indicate that this script is used primarily for benchmarking and that you use it for any other purpose at your own risk?

With my primary dev box being OS X, I personally do see value in having a setup option that is distro agnostic, even if it won't get me 100% to where I need to be. (I'm actually tempted to crib this for Studio after doing some performance testing today and wondering if the Django dev server was significantly affecting my profiling times.)

benjaoming:
@kollivier it's also about avoiding having to maintain a code base of stuff that one doesn't know if in production or not.... when it pops up again in 2 years :)

kollivier:
@benjaoming True, but if any user doing using this for production was specifically warned not to,  we can politely point out that the command warned them against it, and hopefully point them towards some docs on how to solve their problem the right way.

Also, one other idea that might address your concerns somewhat - what if only the configure_nginx step ran, and it spit out simple instructions for how to make changes to the config and start it, rather than auto-starting it? (Or perhaps make a non-default argument like --run-benchmarking-server that runs the start steps.) So it would serve more as a helper script for getting your Kolibri nginx install going, and would only be intended to run once, rather than be the actual command to run the server. (Obviously command name would have to change to something like configurenginx in this case.)

jamalex:
> So it would serve more as a helper script for getting your Kolibri nginx install going, and would only be intended to run once, rather than be the actual command to run the server. (Obviously command name would have to change to something like configurenginx in this case.)

That's essentially what we used to do in KA Lite:
https://github.com/learningequality/ka-lite/blob/cc2e7814e7c3306a13b4b32a16cf9da55ab199ab/kalite/main/management/commands/nginxconfig.py

It looks like that was recently removed:
http://ka-lite.readthedocs.io/en/latest/installguide/release_notes.html?highlight=nginxconfig#command-cleanup

@benjaoming, can you tell us a bit about how that's been replaced in KA Lite, and how you approach things like what I described earlier in this thread:

>Would love to learn more about how you will be approaching managing the contents of these config files dynamically. There are several values in the nginx config that depend on internal mutable Kolibri options

kolibri/deployment/management/commands/startnginx.py
indirectlylit: it's merged!

benjaoming:
> @benjaoming, can you tell us a bit about how that's been replaced in KA Lite, and how you approach things like what I described earlier in this thread:

@jamalex it was replaced with a debian package (`ka-lite-raspberry-pi`) that supports Nginx and depends on it. So everything is automatically setup. Even some extra settings for Nginx to optimize it.

https://github.com/learningequality/ka-lite-installers/blob/master/debian/debian/nginx.conf
+
https://github.com/learningequality/ka-lite-installers/blob/master/debian/debian/ka-lite.scripts-common#L439

> Would love to learn more about how you will be approaching managing the contents of these config files dynamically. There are several values in the nginx config that depend on internal mutable Kolibri options

The Debian pkg can ship with a default configuration for the system user `kolibri`. It is intended that some files are placed on the system as "read only", and others are there for the user to adjust. Critical is to not invite users to change files that are managed by dpkg, as it will give warnings during upgrades, where users have to merge changes manually (or select between local version and package maintainers version).

I think we could consider a `settings.ini.dist` kind of pattern, whereby if `settings.ini` doesn't exist then the package will copy from `settings.ini.dist`. I'm not sure what non-default settings would be essential for the system yet.

For the Nginx configuration, we need to define `STATIC_ROOT`, `MEDIA_ROOT` etc. to be served in the right locations -- but only for the default setup. If users change it, they would have to adjust their Nginx configuration as well.

benjaoming:
> This functionality is necessary to move forward on the performance testing project.

I agree, let's move it forward such that it only appears for benchmarking purposes?

Just being cautious about repeating what happened for KA Lite, where the code inside KA Lite was eventually removed (and was a headache to support).

kolibri/deployment/default/settings/base.py
```diff
@@ -63,6 +63,7 @@
     'django_filters',
     'kolibri.auth.apps.KolibriAuthConfig',
     'kolibri.content',
+    'kolibri.deployment',
```
benjaoming, line 64: This *could* potentially also just live in `deviceadmin` and be called for instance `devserver_nginx`

`kolibri manage devserver_nginx`

Just to save a bit on the app name spaces and not use `startnginx`.

kolibri/deployment/management/commands/startnginx.py
```diff
@@ -0,0 +1,154 @@
...
+context = {
+    "kolibri_home": settings.KOLIBRI_HOME,
+    # "http_port": OPTIONS["Deployment"]["HTTP_PORT"],
+    "http_port": os.environ.get("KOLIBRI_LISTEN_PORT", 8080),
+    "static_root": settings.STATIC_ROOT,
+    "content_database_dir": settings.CONTENT_DATABASE_DIR,
+    "content_storage_dir": settings.CONTENT_STORAGE_DIR,
+}
+
+nginx_config = """
```
benjaoming: Contents of the UWSGI config also live and be written to a temporary named file? Or be set as command line args?

benjaoming: I mean, be specified as a Python string, rather than a `.ini` file in the python package.

indirectlylit:
created https://github.com/learningequality/kolibri-installer-debian/issues/33 to track the longer-term production version

jredrejo:
https://github.com/learningequality/kolibri-network-layer will do it for installations based on Debian packages. Not an universal solution, but it makes sense on servers using Debian apt to install nginx & uwsgi.

PR #3667, opened 2018-05-15
Add autofix for improperly imported content trees.

rtibbles:
### Summary
The release of fully schema compatible content databases on Studio caused ContentNode trees to be imported into Kolibri with identical `tree_id`s. This was fixed in #3519, and so should no longer occur.

However, there may be some installations out there that are suffering from this issue, and would appreciate a fix. Rather than try to make them go through a manual procedure, this attempts to auto-remedy the situation for them.

First, it checks to see if there are multiple channels with a `tree_id` of 1. If so, it attempts to reimport the smallest channels if the content databases are present. If not present, it just presents a warning.

### Reviewer guidance
Check the code, check the tests. I can send you a messed up db if it is helpful for testing.

----

### Contributor Checklist

### Reviewer Checklist

### DIFF
--- a/kolibri/content/utils/annotation.py
+++ b/kolibri/content/utils/annotation.py
@@ -32,6 +32,8 @@ def update_channel_metadata():
     If we are potentially moving from a version of Kolibri that did not import its content data,
     scan through the settings.CONTENT_DATABASE_DIR folder for all channel content databases,
     and pull the data from each database if we have not already imported it.
+    Additionally, fix any potential issues that might be in the current content database from bugs
+    in a previous version.
     """
     from .channel_import import import_channel_from_local_db
     channel_ids = get_channel_ids_for_content_database_dir(settings.CONTENT_DATABASE_DIR)
--- a/kolibri/content/utils/annotation.py
+++ b/kolibri/content/utils/annotation.py
@@ -10,6 +10,7 @@ from sqlalchemy import func
 from sqlalchemy import select
 
 from .channels import get_channel_ids_for_content_database_dir
+from .paths import get_content_database_file_path
 from .paths import get_content_file_name
 from .paths import get_content_storage_file_path
 from .sqlalchemybridge import Bridge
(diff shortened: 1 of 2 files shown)


kolibri/content/utils/annotation.py
```diff
@@ -36,7 +38,41 @@ def update_channel_metadata():
...
+        for channel_id in tree_id_one_channel_ids:
+            channel_sizes[channel_id] = ContentNode.objects.filter(channel_id=channel_id).count()
+        # Get sorted list of ids by increasing number of nodes
+        sorted_channel_ids = sorted(channel_sizes, key=channel_sizes.get)
+        # Loop through all but the largest channel, delete and reimport
+        count = 0
+        from .channel_import import import_channel_from_local_db
+        for channel_id in sorted_channel_ids[:-1]:
+            # Double check that we have a content db to import from before deleting any metadata
+            if os.path.exists(get_content_database_file_path(channel_id)):
```
ralphiee22, line 65: Maybe this can be wrapped in a transaction?

rtibbles: Unfortunately, it can't, because the deletion is using the Django ORM, and the import is using SQLAlchemy.

This can be rectified in 0.10, however, where it can be.

kolibri/content/utils/annotation.py
```diff
@@ -36,7 +38,41 @@ def update_channel_metadata():
     for channel_id in channel_ids:
         if not ChannelMetadata.objects.filter(id=channel_id).exists():
             import_channel_from_local_db(channel_id)
-            set_availability(channel_id)
+    fix_multiple_trees_with_id_one()
```
indirectlylit, line 43: Minor Q: the docstring says

> If we are potentially moving from a version of Kolibri that did not import its content data

However, now this function seems a bit broader in scope, so might want to tweak the comment a bit?

rtibbles: True, true. Can update shortly.

Issue #3663, opened 2018-05-15
add relevance-sorted search results


indirectlylit:
### Observed behavior

In #3599 we greatly simplified search behaviors, and tried to make things a bit more predictable for end-users.

However, one of the consequences is that don't do any kind of ranking, and so less relevant results might get displayed instead of more-relevant results. 

### Expected behavior

relevance-ranked results

For some discussion, see https://github.com/learningequality/kolibri/pull/3599#issuecomment-385132331

### User-facing consequences

important results might be obscured

### Context

0.10.0

rtibbles:
One important dimension of relevance: current language.

PR #3670, opened 2018-05-16
new and improved search

ralphiee22:
### Summary

We implement various search filter queries on the complete set of contentnodes. We allow a maximum of 30 results so if the 1st strategy fills up those results, then we do not run any of the subsequent queries. 

### Reviewer guidance

use search and let me know what you think!

### References

https://github.com/learningequality/kolibri/issues/3663

----

### Contributor Checklist

### Reviewer Checklist


kolibri/content/api.py
```diff
@@ -103,26 +104,55 @@ def filter_search(self, queryset, name, value):
...
+            # exact match in title
+            Q(title__icontains=value),
+            # all words in title
+            intersection([Q(title__icontains=w) for w in words]),
+            # exact match in description
+            Q(description__icontains=value),
+            # all words in description
+            intersection([Q(description__icontains=w) for w in words]),
+            # any stemmed words in title
+            union([Q(title__icontains=stemmer.stem(w)) for w in words]),
```
ralphiee22: Note: we stem the words in the last 2 strategies as a last ditch effort to get some relevant results. Without the stemming, something like "sciences" would not get any hits.

kolibri/content/api.py
```diff
@@ -37,7 +38,7 @@
 
 
 logger = logging.getLogger(__name__)
-
+stemmer = Porter2Stemmer()
```
indirectlylit: This seems a bit risky because it assumes the search queries are in English. Results could be unpredictable when applied to other languages

indirectlylit:
we'll wait to merge this until after AB testing

ralphiee22:
http://www.nltk.org/api/nltk.stem.html#nltk.stem.snowball.SnowballStemmer
Library that provides stemming for various languages.

kolibri/content/api.py
```diff
@@ -103,26 +102,55 @@ def filter_search(self, queryset, name, value):
...
-            if id_pair['content_id'] in content_id_set:
-                pass
-            else:
-                content_id_set.add(id_pair['content_id'])
-                id_list.append(id_pair['id'])
-
-        return matches.filter(id__in=id_list[:30])
+        # return the result of and-ing a list of queries
+        def intersection(queries):
+            return reduce(lambda x, y: x & y, queries)
```
indirectlylit, line 108: `reduce` doesn't work in python3

kolibri/content/api.py
```diff
@@ -103,26 +103,55 @@ def filter_search(self, queryset, name, value):
...
+                # filter the dupes
+                if match.content_id in content_ids:
+                    continue
+                # add new, unique results
+                content_ids.add(match.content_id)
+                results.append(match)
+
+                # bail out as soon as we reach the quota
+                if len(results) >= MAX_RESULTS:
+                    return results
```
jamalex, line 151: Adding the fuzzed results back in as the last "query" could help fill in the gaps if there are typos.

jamalex: Or maybe it should be handled specially -- and only be called when there are *no* exact results. And pass a flag back indicating that it's a fuzzy set of results, so the UI can indicate "No exact results found for your search. Some possible matches are shown below."

indirectlylit:
see also https://github.com/learningequality/kolibri/issues/3682

kolibri/content/api.py
indirectlylit: sounds reasonable .. but I think the stemmed metaphone stuff was english-only?

jamalex: Works best with EN, yeah. Other European languages it could still be helpful for as well.

indirectlylit: created a new issue here: https://github.com/learningequality/kolibri/issues/3684

indirectlylit:
build is failing due to https://github.com/learningequality/kolibri/issues/3678

PR #3705, opened 2018-05-24
fix for Kolibri on Ubuntu 18.04

lyw07:
### Summary

This pull request includes:
* remove `distutils` in `kolibri/utils/env.py` and use `platform` module instead to make kolibri work on Ubuntu 18.04.
* refactor code in `build_tools/install_cexts.py`. Previously, the code I wrote installs the specific `cryptography` wheel file directly, which makes it install dependencies according to the platform that runs `make dist`. That is to say, since we build kolibri on ubuntu from buildkite, it could be possible that we run linux 64 bit package on a linux/windows 32 bit machine, which would potentially cause error. 
To fix it, I download the cryptography wheel file and its dependencies for different platforms and install the wheel files one by one.

### Reviewer guidance

1. kolibri should be run successfully on Ubuntu 18.04
2. on Linux 32/64 bit machines and Windows 32 bit machines, kolibri has `cryptography` 2.0.3 bundled inside.
### References

Related discussion: https://learningequality.slack.com/archives/CAA2D6BV0/p1526598696000099

----

### Contributor Checklist

### Reviewer Checklist


build_tools/install_cexts.py
```diff
@@ -6,82 +6,90 @@
 import subprocess
 import sys
 
+import requests
```
rtibbles, line 8: I think @aronasorman was trying to make the build tools require no external libraries - perhaps this could use urllib instead?

build_tools/install_cexts.py
```diff
@@ -6,82 +6,90 @@
...
+def _install_bs4_if_needed():
+    """
+    Check if beautifulsoup4 module is installed.
+    """
     try:
         imp.find_module('bs4')
-        print('bs4 module exists')
     except ImportError:
-        subprocess.call(['pip', 'install', 'bs4'])
+        subprocess.call(['pip', 'install', 'beautifulsoup4'])
```
rtibbles: Oh, we do have dependencies, anyway. @aronasorman should we have a build dependencies requirements file that we just use?

rtibbles (commented):
I don't think I know enough about what's going on in here to do a full review. Have left some comments.

build_tools/install_cexts.py
lyw07: ah I see. then I can change it to urllib3

build_tools/install_cexts.py
aronasorman: I thought we already did have a build deps file -- it's in here: https://github.com/learningequality/kolibri/blob/develop/requirements/build.txt

aronasorman: @lyw07 any reason you couldn't just put beautifulsoup in there?

lyw07: hmm, I don't think so.. I don't remember why I didn't put it in the requirements. I will update it.

build_tools/install_cexts.py
lyw07: @rtibbles I just confirmed with Aron that it's okay to use requests now, so I didn't make the change here.

requirements/build.txt
```diff
@@ -8,3 +8,4 @@ sphinx-autobuild
 pex<1.3
 wheel
 setuptools>=2.2,<20.11          # needed to get pex working # pyup: ignore
```
rtibbles, line 10: Do we need to have requests here too?

lyw07: Hmm I don't think we need to put requests here? Since in the Dockerfile, it installs both `dev.txt` and `build.txt`, and `dev.txt` calls `base.txt` which contains requests?

build_tools/install_cexts.py
benjaoming: `requirements/build.txt` exists for that purpose, but `requests` is already a dep of other deps.

requirements/build.txt
benjaoming: @lyw07 we do -- I think it's supposed to be that you only need to run `pip install -r requirements/build.txt` to build Kolibri. The other stuff is for dev'ing, testing, runtime etc.

lyw07: @benjaoming that makes sense! so I guess I can add `requests` in `requirements/build.txt` even though it's already in `base.txt`? Thank you!

rtibbles:
Anything blocking updates on this @lyw07?

lyw07:
@rtibbles ah I'm just not sure if I should put `requests` in `requirements/build.txt` now. Thanks!

requirements/build.txt
benjaoming: Yes, `requirements/build.txt` does not install anything from `requirements/base.txt`, which might be renamed to `runtime.txt` to be more precise :)

benjaoming: (but not now) :)

lyw07: @benjaoming thank you!

rtibbles:
@lyw07 if something is needed for the build, it should be in the build requirements. Duplicates are fine between the build.txt and other requirements because they are disjoint.

lyw07:
@rtibbles thank you! I've updated the code with the `requests` in `build.txt`

benjaoming (approved):
If tests pass, LGTM!

benjaoming:
@rtibbles does this conclude 0.9.3?

PR #3706, opened 2018-05-24
Refactor docker files

aronasorman:
This is the first step towards Continuous Delivery!

There are now 3 Dockerfiles:

1. `deploy/build.dockerfile` -- the old dockerfile, used to just build Kolibri and produce the installers.

2. `deploy/minimal-env.dockerfile` -- a new dockerfile that simply runs a PEX file from a URL.

3. `Dockerfile` -- a new dockerfile that not only builds the PEX file based on the current source code, but then runs and serves it under port 8080.


Dockerfile
```diff
@@ -1,27 +1,26 @@
...
+RUN yarn install
+RUN make dist
+RUN ls -l /kolibri/dist/
+
+FROM python:2.7.15
 
-VOLUME /kolibridist/  # for mounting the whl files into other docker containers
-# add buildkite pipeline specific installation here:
-CMD cd /kolibri && pip install -r requirements/dev.txt && pip install -r requirements/build.txt && pip install -e . && yarn install && make dist pex && cp /kolibri/dist/* /kolibridist/
+COPY --from=0 /kolibri/dist/kolibri*.whl .
```
aronasorman, line 23: Just copying over the whl file to a fresh docker image makes the resulting deployment image smaller!

aronasorman:
Looks like PEX files are still successfully built!

ivanistheone:
Hi Aron, sorry it took so long to review. I've got some thing to add to the entrypoint script. I want to switch it to bash instead of python and add this function that registers facility:
https://github.com/learningequality/cloud-kolibri-demo/blob/master/config/setupkolibri.template.sh#L8  (otherwise very annoying to have to go through facility wizard every time)

It might be also useful to add a function to importchannel and importcontent from a comma-separated list provided as env var.

Can I take over this branch for today/tomorrow and push to it?

aronasorman:
@ivanistheone sounds good to me!

aronasorman:
@ivanistheone sounds good to me!

rtibbles:
@ivanistheone could you use the `provisiondevice` management command instead?https://github.com/learningequality/kolibri/blob/develop/kolibri/core/device/management/commands/provisiondevice.py

ivanistheone:
Moved work here: https://github.com/learningequality/kolibri/pull/3775

rtibbles:
Closing in favour of #3775

Issue #877, opened 2017-02-08
Use CSP middleware to block online sources in DEBUG mode


benjaoming:
## Summary

With KA Lite, we debugged some issues fetching online sources with JavaScript.

During this process, I had a theory that external content (fetched from our update/content server) was blocked due to inadequate security definitions. This wasn't true, however a CSP middleware came out of it.

I made it so it only triggers in `DEBUG=True` mode. The advantage is that we can ensure that ALL resources fetched from online sources etc. are blocked while we are developing. Or explicitly allowed. That way, we can ensure that we don't accidentally fetch an online source during testing or development that won't actually be available in an offline context.

```
class CSPMiddleware:
    """
    Implements the Content-Security-Policy response header, which
    conforming user-agents can use to restrict the permitted sources
    of various content.
    See http://www.w3.org/TR/CSP/
    """

    def process_response(self, request, response):

        # Check for debug view
        status_code = response.status_code
        if not settings.DEBUG or status_code in (403, 404, 500):
            return response

        header = 'Content-Security-Policy'
        if getattr(settings, 'CSP_REPORT_ONLY', False):
            header += '-Report-Only'

        response[header] = "default-src 'self' 'unsafe-eval' 'unsafe-inline' data: *.learningequality.org; img-src data: *; script-src 'self' *.learningequality.org 'unsafe-eval' 'unsafe-inline'"
        return response
```

## How to reproduce

1. Use an external JS library that has an online-dependent behavior
1. Don't notice it during development or testing
1. Be baffled when stuff breaks offline.

## Real-life consequences (anything community should be aware of, for instance how it affects your deployment)

I think the threat is real, as we don't necessary profile the networking behavior of all client-side browser scripts.

benjaoming:
Btw. we could also block `eval` and `inline` if we want to be even more strict. There could be a security policy in the future or in school environments that increases what JS can and cannot do.

jamalex:
Very cool! A perhaps marginally more efficient way to have it only run in development is to have it only added to `MIDDLEWARE` in the dev settings module.

jamalex:
> we could also block eval and inline if we want to be even more strict

Would be interesting to see whether this is transitive into sandboxed iframes. We probably wouldn't want to prevent them from doing inline JS or whatever they want (within the confines of the `sandbox` specifications), but we similarly *would* want to prevent them from trying to load online resources as described in the initial proposal above.

benjaoming:
[...]

PR #3707, opened 2018-05-25
Updates to response headers for ZIP endpoints for security and functionality

jamalex:
### Summary

Fixes some issues we've been seeing, and proactively protects us against others.

1. Add "Content-Security-Policy" header: More aggressive protections against content being loaded from other servers (to highlight issues that might only show up offline, avoid unwanted bandwidth usage, and protect against data leakage once we add HTML5 app data access methods -- CC @kollivier).
1. Use Django decorator to block XFrameOptionsMiddleware instead of overwriting header, to avoid "Invalid 'X-Frame-Options' header encountered" JS error (which still worked, but caused unnecessary panic).
1. Add "Access-Control-Allow-Origin" header in response to OPTIONS requests: Prevent CORS errors when HTML5 apps make AJAX requests to read files from within zip.

Many (most? all?) of these changes should also be ported over [to Studio](https://github.com/learningequality/studio/blob/4ecb84457e78df35ac14c13eec7c5829c0aa0781/contentcuration/contentcuration/views/zip.py#L30) (CC @ivanistheone @jayoshih @micahscopes).

### Reviewer guidance

I made a zip file that tests all the edge cases, which I've used for doing fairly extensive cross-browser testing on BrowserStack:
[a0de2cf520524f207eed87cc7056ac6b.zip](https://github.com/learningequality/kolibri/files/2037706/a0de2cf520524f207eed87cc7056ac6b.zip)

You can overwrite one of your existing content folder zip files with that file (renaming appropriately) so as to load it in a Kolibri sandboxed iframe easily. Current filename matches the zip for ContentNode: "Channels > PhET Interactive Simulations (en) > Physics > Motion > Gravity And Orbits".

(More manual testing would also be good, on more actual content)

Screenshot of the test zip in action:

image

The main "failure" case is that in old old versions of some browsers, "Content-Security-Policy" is ignored. This just leads to it being possible to load external resources from other domains into the iframe -- doesn't compromise anything critical.

### References

Related to (but different from) https://github.com/learningequality/kolibri/issues/877 -- which is about applying Content-Security-Policy externally to the iframes, in Kolibri itself -- CC @benjaoming.

----

### Contributor Checklist

### Reviewer Checklist

### DIFF
--- a/kolibri/content/views.py
+++ b/kolibri/content/views.py
@@ -19,6 +19,16 @@ mimetypes.init([os.path.join(os.path.dirname(__file__), 'constants', 'mime.types
 class ZipContentView(View):
 
     @xframe_options_exempt
+    def options(self, request, *args, **kwargs):
+        """
+        Handles OPTIONS requests which may be sent as "preflight CORS" requests to check permissions.
+        """
+        response = HttpResponse()
+        response["Access-Control-Allow-Origin"] = "*"
+        response["Access-Control-Allow-Methods"] = "GET, OPTIONS"
+        return response
+
+    @xframe_options_exempt
     def get(self, request, zipped_filename, embedded_filepath):
         """
         Handles GET requests and serves a static file from within the zip file.
--- a/kolibri/content/views.py
+++ b/kolibri/content/views.py
@@ -2,8 +2,10 @@ import mimetypes
 import os
 import zipfile
 
-from django.http import Http404, HttpResponse
-from django.http.response import FileResponse, HttpResponseNotModified
+from django.http import Http404
+from django.http import HttpResponse
+from django.http.response import FileResponse
+from django.http.response import HttpResponseNotModified
 from django.views.generic.base import View
 from le_utils.constants import exercises
 
@@ -78,6 +80,11 @@ class ZipContentView(View):
         # Newer versions of Chrome block iframes from loading
         # solution is to override xframe options with any string (https://stackoverflow.com/a/6767901)
         response['X-Frame-Options'] = "GOFORIT"
+
+        # restrict CSP to only allow resources to be loaded from the Kolibri host, to prevent info leakage
+        # (e.g. via passing user info out as GET parameters to an attacker's server), or inadvertent data usage
+        host = request.build_absolute_uri('/').strip("/")
+        response["Content-Security-Policy"] = "default-src 'self' 'unsafe-inline' 'unsafe-eval' data: " + host
 
         return response
 
--- a/kolibri/content/views.py
+++ b/kolibri/content/views.py
@@ -16,6 +16,15 @@ from .utils.paths import get_content_storage_file_path
 # https://www.thecodingforums.com/threads/mimetypes-guess_type-broken-in-windows-on-py2-7-and-python-3-x.952693/
 mimetypes.init([os.path.join(os.path.dirname(__file__), 'constants', 'mime.types')])
 
+
+def _add_access_control_headers(request, response):
+    response["Access-Control-Allow-Origin"] = "*"
+    response["Access-Control-Allow-Methods"] = "GET, OPTIONS"
+    requested_headers = request.META.get("HTTP_ACCESS_CONTROL_REQUEST_HEADERS", "")
+    if requested_headers:
+        response["Access-Control-Allow-Headers"] = requested_headers
+
+
 class ZipContentView(View):
 
     @xframe_options_exempt
@@ -24,8 +33,7 @@ class ZipContentView(View):
         Handles OPTIONS requests which may be sent as "preflight CORS" requests to check permissions.
         """
         response = HttpResponse()
-        response["Access-Control-Allow-Origin"] = "*"
-        response["Access-Control-Allow-Methods"] = "GET, OPTIONS"
+        _add_access_control_headers(request, response)
         return response
 
     @xframe_options_exempt
@@ -86,8 +94,8 @@ class ZipContentView(View):
         # ensure the browser knows not to try byte-range requests, as we don't support them here
         response["Accept-Ranges"] = "none"
 
-        # allow all origins so that content can be read from within zips within sandboxed iframes
-        response["Access-Control-Allow-Origin"] = "*"
+        # add headers to ensure AJAX requests will be permitted for these files, even from a null origin
+        _add_access_control_headers(request, response)
 
         # restrict CSP to only allow resources to be loaded from the Kolibri host, to prevent info leakage
         # (e.g. via passing user info out as GET parameters to an attacker's server), or inadvertent data usage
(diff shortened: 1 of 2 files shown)


kolibri/content/views.py
```diff
@@ -15,6 +18,17 @@
...
 class ZipContentView(View):
 
+    @xframe_options_exempt
+    def options(self, request, *args, **kwargs):
+        """
+        Handles OPTIONS requests which may be sent as "preflight CORS" requests to check permissions.
+        """
+        response = HttpResponse()
+        response["Access-Control-Allow-Origin"] = "*"
+        response["Access-Control-Allow-Methods"] = "GET, OPTIONS"
```
ivanistheone: We might also need `Access-Control-Allow-Headers`.

This is what I get for the file from earlier today, running the code from this PR:
http://35.196.143.130/learn/#/topics/c/10f41b318dc546dabe0a596aaa537a53

Cross-Origin Request Blocked: The Same Origin Policy disallows reading the remote resource at http://35.196.143.130/zipcontent/213988fd537e4a608e2b86be9632d92d.zip/xml/course_config.xml. (Reason: missing token ‘x-requested-with’ in CORS header ‘Access-Control-Allow-Headers’ from CORS preflight channel).

In FF and Chrome.

benjaoming:
Looks good, @jamalex -- you can also only enforce it with `DEBUG=True` if you want to release it with that extra caution (then of course it's only going to be useful for testing purposes, and not a security improvement).

ivanistheone:
@jamalex Just deployed this PR to a demo server: http://35.196.143.130/ and I have good news and bad news:

The HTML app that Swapnil was working on works now: http://35.196.143.130/learn/#/topics/c/10f41b318dc546dabe0a596aaa537a53
Woo hoo... one head of the CORS dragon has been chopped off!

On the other hand, there seems to be still some issues with the GeoGebra applet:
http://35.196.143.130/learn/#/topics/c/03380b18bcac527c965c60294c675fe6 

Chrome console:
```
Uncaught DOMException: Blocked a frame with origin "null" from accessing a cross-origin frame.
    at w (http://35.196.143.130/zipcontent/3e0c56d1cf2f246e4d14af91406818c3.zip/GeoGebra/HTML5/5.0/web3d/web3d.nocache.js:4:133)
    at v (http://35.196.143.130/zipcontent/3e0c56d1cf2f246e4d14af91406818c3.zip/GeoGebra/HTML5/5.0/web3d/web3d.nocache.js:3:671)
    at m (http://35.196.143.130/zipcontent/3e0c56d1cf2f246e4d14af91406818c3.zip/GeoGebra/HTML5/5.0/web3d/web3d.nocache.js:9:7)
    at http://35.196.143.130/zipcontent/3e0c56d1cf2f246e4d14af91406818c3.zip/GeoGebra/HTML5/5.0/web3d/web3d.nocache.js:10:51
    at HTMLDocument.d (http://35.196.143.130/zipcontent/3e0c56d1cf2f246e4d14af91406818c3.zip/GeoGebra/HTML5/5.0/web3d/web3d.nocache.js:6:71)
```

Firefox console similar stack trace with message: `SecurityError: Permission denied to access property "document" on cross-origin object`


The geogebra applet is from test channel:
https://contentworkshop.learningequality.org/channels/43c08c5c7a7d5365b8a0fe5b0b7b28ad/edit/43c08c5

Download testzip using `wget 'https://contentworkshop.learningequality.org/content/storage/3/e/3e0c56d1cf2f246e4d14af91406818c3.zip' -O testzip.zip` if you want to inspect js code that's causing the issue.

kollivier:
@ivanistheone The dependency that is causing that error shows the following URL in my Chrome: webpack:///./src/statics/turnstile-compiled.js

Is there a file like that in the package? Or is this some webpack magic that is trying to reference the file at that URL?

Also, I've noticed that web3d internally is trying to read/write cookies, which the iframe is forbidding, so that could be causing issues with loading as well.

kollivier:
Related question, if cookies and local storage access are blocked, do HTML5 apps have any (pre-existing) options for persisting app state when used in this mode?

rtibbles:
This error would seem to point to a clue: `sessionStorage not supportedFailed to read the 'sessionStorage' property from 'Window': The document is sandboxed and lacks the 'allow-same-origin' flag.` - would the sessionStorage be shared between the iframe and the containing frame if we allowed this, or is this just banning different scripts within the iframe affecting the iframe's session storage?

If it's the latter, then we should allow it, and this stuff should probably work.

jamalex:
> if cookies and local storage access are blocked, do HTML5 apps have any (pre-existing) options for persisting app state when used in this mode?

Nope, don't think so.

> sessionStorage not supported

Yeah, we'll need to modify the code on-the-fly (did a bunch of this in CK-12) to remove/mock calls to things like this.

> SecurityError: Permission denied to access property "document" on cross-origin object

I faced this problem on CK-12 as well, as PLIX contain an embedded iframe to render the interactive widget thingie. There's no way to communicate directly between embedded iframes and the main html5 app as they don't have origins (would be nice if you could set "pseudo-origins" in the sandbox, but that would have other security implications).

In CK-12, I went the route of replacing it with postMessage:
https://github.com/learningequality/sushi-chef-ck12/blob/master/chef.py#L894

jamalex:
Any changes requested in here that can be addressed by the Kolibri code itself, rather than needing HTML5 app code changes? If not, someone willing to approve?

jamalex:
Also, in case anyone wants to do more cross-browser testing, here's an updated zip file that tests the header stuff:
[a0de2cf520524f207eed87cc7056ac6b.zip](https://github.com/learningequality/kolibri/files/2050389/a0de2cf520524f207eed87cc7056ac6b.zip)

kollivier:
> There's no way to communicate directly between embedded iframes and the main html5 app as they don't have origins (would be nice if you could set "pseudo-origins" in the sandbox, but that would have other security implications).

What about actually giving them different origins via subdomains? e.g. instead of http://localhost:8000/zipcontent/zipid, we do http://zipid.localhost:8000/, which then gives each zip a unique origin. Then we use the same origin CORS restriction, but allow for AJAX communication with the root Kolibri domain for things like the HTML assessment API. Would this approach still allow apps to have access to content in Kolibri that they shouldn't have access to?

ivanistheone:
@kollivier I think changing the domain would be a much more complex strategy (this PR is with fixes we hope to ship soon, i.e. with 0.10)

ivanistheone:
@jamalex Here is a link to the demo server with both CORS HTML5 apps for testing:
http://35.196.143.130/learn/#/topics/43c08c5c7a7d5365b8a0fe5b0b7b28ad

I'm going to approve this PR because all reamining issues seem to be edge cases to be handled by cheffing.

ivanistheone (approved):

indirectlylit:
thanks!

PR #3711, opened 2018-05-25
remove vestigial py_modules code dump

indirectlylit:
### Summary

Currently, our dev docs dump a lot of code into the search results, which makes searching less useful:

image

It appears this is just a dump of code, which is super unhelpful when better tools exist on github or in editors:

image

It also doesn't address anything on the front-end side, so the inconsistency is confusing.

### Reviewer guidance

Make sure this change makes sense

----

### Contributor Checklist

### Reviewer Checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -60,10 +60,6 @@ output/*/index.html
 # Sphinx
 docs/_build
 
-# api-docs
-docs/py_modules/kolibri*rst
-docs/py_modules/modules.rst
-
 # node
 node_modules
 npm-debug.log
--- a/Makefile
+++ b/Makefile
@@ -68,8 +68,6 @@ clean-pyc:
 	find . -name '*~' -exec rm -f {} +
 
 clean-docs:
-	rm -f docs/py_modules/kolibri*rst
-	rm -f docs/py_modules/modules.rst
 	$(MAKE) -C docs clean
 
 lint:
@@ -90,7 +88,6 @@ coverage:
 	coverage report -m
 
 docs: clean-docs
-	sphinx-apidoc -d 10 -H "Python Reference" -o docs/py_modules/ kolibri kolibri/test kolibri/deployment/ kolibri/dist/
 	$(MAKE) -C docs html
 
 release:
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -133,9 +133,6 @@ html_theme = 'default'
 on_rtd = os.environ.get('READTHEDOCS', None) == 'True'
 
 
-if on_rtd:
-    os.system("sphinx-apidoc --doc-project='Python Reference' -f -o . ../kolibri ../kolibri/test ../kolibri/deployment/ ../kolibri/dist/")
-
 if not on_rtd:  # only import and set the theme if we're building docs locally
     import sphinx_rtd_theme
     html_theme = 'sphinx_rtd_theme'


benjaoming:
It's dev docs...

The point is to write docstrings and being able to read them somewhere, with their glorious formatting involved, cross-referencing from the real docs to the API docs etc.

I think you are forgetting that people are also being onboarded in this project, and API docs are a way to access the codebase.

benjaoming:
The issue that you had searching for `i18n` seems to be fairly simple: You get hits for `i18n`, look through them and you'll find something useful.

This is exactly the same in many Python projects hosted on Sphinx, I don't really see much complaining.

We can of course isolate the API docs further, but we loose the ability to quickly cross-reference. E.g.:

http://kolibri-dev.readthedocs.io/en/develop/architecture/plugins.html#core-plugin-example

benjaoming:
> It appears this is just a dump of code, which is super unhelpful when better tools exist on github or in editors:

It's not, it's our docstring, which developers are supposed to be writing.

indirectlylit:
Maybe something is not working correctly then? Because the search results are code dumps, not docstrings. See for example: https://kolibri-dev.readthedocs.io/en/develop/_modules/kolibri/core/views.html?highlight=i18n

> I think you are forgetting that people are also being onboarded in this project, and API docs are a way to access the codebase.

I'm sure that we share the same goal of helping new developers onboard more easily.

I'm not convinced that the way we currently integrate code searching and browsing into the developer docs is conducive to new dev onboarding, because it adds noise to search results and skips all front-end code.

benjaoming:
Solution 1: Do nothing, it's an upstream issue

I don't think looking through 8 search results (in the example from the description) is much noise to be concerned with -- instead, why not see the search-able source code as a nice holistic addition to the kolibri-dev docs? Here you have 1 simple search field that indexes everything relevant.

(We _could_ open an issue that frontend API docs should somehow also be generated)

14 days ago, ReadTheDocs devs started discussing a new search backend, so maybe this is an opportunity to pitch in that source code results should have lower search score. Referencing it directly, in case they can benefit from this issue: https://github.com/rtfd/readthedocs.org/issues/4094

Solution 2: Can something be configured?

* The search backend might be configurable, but I haven't looked into that: http://www.sphinx-doc.org/en/master/web/searchadapters.html

* Source documentation can maybe be deleted, meaning to wipe everything in `docs/_modules` or maybe [sphinx-apidoc](http://www.sphinx-doc.org/en/master/man/sphinx-apidoc.html) offers configuration in this regard.

indirectlylit:
From my perspective, I think an optimal short-term situation might look something like:

* Preserve the ability to embed python module docstrings and code references in the docs. This is clearly useful, and results in some nice results, e.g. [this page](http://kolibri-dev.readthedocs.io/en/develop/architecture/uap/index.html)
* Filter out python code from search results. I suppose we disagree on this point, but I think it's detrimental to have them in there when (a) we can recommend better ways of searching the code, and (b) they provide an incomplete (and therefore misleading) code search

Here's another concrete example - try [searching the docs for "models"](http://kolibri-dev.readthedocs.io/en/develop/search.html?q=models&check_keywords=yes&area=default#). There are a couple useful results in there, but picking them out will be left as an exercise to the reader :)

rtibbles:
_Cui bono?_ I am not particularly convinced that having the code base in the dev docs particularly helps anyone, unless it's via the deliberate link from another part of the documentation - in which case we can do the explicit import as we do in some places to bring them into the codebase.

Having actively onboarded many (N > 10) people into Kolibri, I have yet to see a single person look at the code on the dev docs (it is hard enough to get them to look at the docs at all). I think anything that reduces the irrelevant information is probably going to have a net benefit, and if it seems to be having a deleterious effect, we can always reverse this commit.

indirectlylit:
I've re-tested and confirmed that this PR works as intended:

We do _not_ lose the ability to cross-reference from docs to code, which I believe was @benjaoming's primary concern. Let me know if you're seeing something different.

benjaoming:
I can confirm that this was my main concern!

rtibbles (approved):

benjaoming:
I wiped the build, and there are now 4 results instead of 8. Annoyingly one of the four are a 404 on a now deleted source code page :)

benjaoming:
http://kolibri-dev.readthedocs.io/en/develop/search.html?q=i18n&check_keywords=yes&area=default#

Issue #3678, opened 2018-05-17
versioning logic is preventing legitimate build from running


indirectlylit:
### Observed behavior

Builds produced by #3670 won't run

### Expected behavior

* Builds should run
* version.py should log warnings but not raise assertion errors

### User-facing consequences

cannot deploy a demo server with contents of PR

### Errors and logs

```
Traceback (most recent call last):
  File "/Users/d/Downloads/k.pex/.bootstrap/_pex/pex.py", line 365, in execute
  File "/Users/d/Downloads/k.pex/.bootstrap/_pex/pex.py", line 293, in _wrap_coverage
  File "/Users/d/Downloads/k.pex/.bootstrap/_pex/pex.py", line 325, in _wrap_profiling
  File "/Users/d/Downloads/k.pex/.bootstrap/_pex/pex.py", line 408, in _execute
  File "/Users/d/Downloads/k.pex/.bootstrap/_pex/pex.py", line 466, in execute_entry
  File "/Users/d/Downloads/k.pex/.bootstrap/_pex/pex.py", line 471, in execute_module
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/runpy.py", line 170, in run_module
    mod_name, loader, code, fname = _get_module_details(mod_name)
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/runpy.py", line 109, in _get_module_details
    return _get_module_details(pkg_main_name)
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/runpy.py", line 101, in _get_module_details
    loader = get_loader(mod_name)
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/pkgutil.py", line 464, in get_loader
    return find_loader(fullname)
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/pkgutil.py", line 474, in find_loader
    for importer in iter_importers(fullname):
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/pkgutil.py", line 430, in iter_importers
    __import__(pkg)
[...]

PR #3736, opened 2018-05-31
Fine tune Kolibri CherryPy params, allowing options

### LINKED ISSUE #3720: Optimize CherryPy settings (thread_pool)
### Observed behavior

In performance testing, @bonidjukic and @jredrejo found that `server.thread_pool=30` was the worst possible setting. Incidentally that's the current setting :tada: 

 - [ ] Urgently lower `server.thread_pool` (5? 10?) - Urgent for 0.10.0?
 - [ ] React to further optimization recommendations

### Expected behavior

We use the settings that are optimal for a lower scale device and can then be tweaked by someone with more hardware resources?

### User-facing consequences

Failing requests! Request failure rate at high load is quite possible and predictable.

### Errors and logs

https://docs.google.com/spreadsheets/d/1QNHI554kXaa0CoKrHzVcb3UhfYRsjUa0X87-7Ep9AjE/edit#gid=0

https://docs.google.com/spreadsheets/u/1/d/1M77qhocswGq-U16IuY6wi3heqlUq6i3zAQOYsRCqZ_E/edit#gid=0

### Steps to reproduce

Velox

jredrejo:
### Summary
This PR will change the options used to run kolibri server in production.
The options are configurable using the options.ini file

 fixes #3720

### Reviewer guidance
To check how these server options apply kolibri must be rebuilt.
I did it using `python setup.py install`

### Contributor Checklist

### Reviewer Checklist

### DIFF
--- a/kolibri/utils/options.py
+++ b/kolibri/utils/options.py
@@ -33,10 +33,10 @@ option_spec = {
         },
     },
     "Server": {
-        "THREADS_POOL": {
+        "THREAD_POOL": {
             "type": "integer",
             "default": 10,
-            "envvars": ("CHERRYPY_THREADS_POOL",),
+            "envvars": ("CHERRYPY_THREAD_POOL",),
         },
         "SOCKET_TIMEOUT": {
             "type": "integer",
--- a/kolibri/utils/options.py
+++ b/kolibri/utils/options.py
@@ -32,6 +32,28 @@ option_spec = {
             "envvars": ("KOLIBRI_DATABASE_HOST",),
         },
     },
+    "Server": {
+        "THREADS_POOL": {
+            "type": "integer",
+            "default": 10,
+            "envvars": ("CHERRYPY_THREADS_POOL",),
+        },
+        "SOCKET_TIMEOUT": {
+            "type": "integer",
+            "default": 10,
+            "envvars": ("CHERRYPY_SOCKET_TIMEOUT",),
+        },
+        "QUEUE_SIZE": {
+            "type": "integer",
+            "default": 5,
+            "envvars": ("CHERRYPY_QUEUE_SIZE",),
+        },
+        "QUEUE_TIMEOUT": {
+            "type": "float",
+            "default": 0.1,
+            "envvars": ("CHERRYPY_QUEUE_TIMEOUT",),
+        },
+    },
     "Paths": {
         "CONTENT_DIR": {
             "type": "string",
--- a/kolibri/utils/options.py
+++ b/kolibri/utils/options.py
@@ -33,25 +33,25 @@ option_spec = {
         },
     },
     "Server": {
-        "SERVER_THREAD_POOL": {
+        "CHERRYPY_THREAD_POOL": {
             "type": "integer",
             "default": 10,
-            "envvars": ("KOLIBRI_SERVER_THREAD_POOL",),
+            "envvars": ("KOLIBRI_CHERRYPY_THREAD_POOL",),
         },
-        "SERVER_SOCKET_TIMEOUT": {
+        "CHERRYPY_SOCKET_TIMEOUT": {
             "type": "integer",
             "default": 10,
-            "envvars": ("KOLIBRI_SERVER_SOCKET_TIMEOUT",),
+            "envvars": ("KOLIBRI_CHERRYPY_SOCKET_TIMEOUT",),
         },
-        "SERVER_QUEUE_SIZE": {
+        "CHERRYPY_QUEUE_SIZE": {
             "type": "integer",
             "default": 5,
-            "envvars": ("KOLIBRI_SERVER_QUEUE_SIZE",),
+            "envvars": ("KOLIBRI_CHERRYPY_QUEUE_SIZE",),
         },
-        "SERVER_QUEUE_TIMEOUT": {
+        "CHERRYPY_QUEUE_TIMEOUT": {
             "type": "float",
             "default": 0.1,
-            "envvars": ("KOLIBRI_SERVER_QUEUE_TIMEOUT",),
+            "envvars": ("KOLIBRI_CHERRYPY_QUEUE_TIMEOUT",),
         },
     },
     "Paths": {
--- a/kolibri/utils/options.py
+++ b/kolibri/utils/options.py
@@ -33,25 +33,25 @@ option_spec = {
         },
     },
     "Server": {
-        "THREAD_POOL": {
+        "SERVER_THREAD_POOL": {
             "type": "integer",
             "default": 10,
-            "envvars": ("CHERRYPY_THREAD_POOL",),
+            "envvars": ("KOLIBRI_SERVER_THREAD_POOL",),
         },
-        "SOCKET_TIMEOUT": {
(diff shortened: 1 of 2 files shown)


kolibri/utils/options.py
```diff
@@ -32,6 +32,28 @@
             "envvars": ("KOLIBRI_DATABASE_HOST",),
         },
     },
+    "Server": {
```
benjaoming, line 35: This new section name might sound generic, but it's IMO a good name, as opposed to `HTTP_SERVER`, because we would risk having maybe way too many sections.

kolibri/utils/options.py
```diff
@@ -32,6 +32,28 @@
...
+        "THREADS_POOL": {
```
benjaoming: While the ENVVAR states that these are threads for cherrypy, the setting key itself doesnt?

benjaoming (dismissed):
Should be branched off `release-v.0.10.x`: you can do that locally, cherry-py the commit and overwrite the remote branch. The PR will update accordingly.

benjaoming:
Flagging this build error on the side: Something went wrong in the way things got rebased and `git --describe` no longer works. Merging this would introduce a wrong tag. This confirms that assertions and exceptions are raised with good reasons @indirectlylit @rtibbles :)

I still prefer to stick to the "paranoid" handling of `git --describe`, until all unexpected behavior is accounted for :)

```
AssertionError: Version detected from git describe --tags, but it's inconsistent with kolibri.__version__.__version__ is: (0, 10, 0, u'beta', 1), tag says: (0, 11, 0, 'alpha', 1).
```

Referring to #3678

kolibri/utils/options.py
jredrejo: Yep, I thought of using `Server`, `HTTP_SERVER`or even `CherryPy`, but as you say it could lead us to have too many sections. `Server`can be used for more settings in the future.

kolibri/utils/options.py
jredrejo: I didn't want to do the keys so long, and as far as they are under the `Server` Section, I considered it implicit.

CHANGELOG.rst
```diff
@@ -3,6 +3,15 @@ Release Notes
...
 
+0.9.2
+-----
+
+ - Various bug fixes
+ - Languages: English, Arabic, Bengali, Chinyanja, Farsi, French, Hindi, Marathi, Portuguese, Spanish, Swahili, Tamil,   Telugu, Urdu, Yoruba, and Zulu
+
+See a `more detailed list <https://github.com/learningequality/kolibri/issues?q=milestone%3A0.9.2+label%3Achangelog>`_ of changes on Github.
+
+
```
benjaoming: @indirectlylit how did this get here, I wonder?

indirectlylit:
re #3720

kolibri/utils/options.py
```diff
@@ -32,6 +32,28 @@
...
+        },
+        "QUEUE_SIZE": {
+            "type": "integer",
+            "default": 5,
+            "envvars": ("CHERRYPY_QUEUE_SIZE",),
+        },
+        "QUEUE_TIMEOUT": {
+            "type": "float",
+            "default": 0.1,
+            "envvars": ("CHERRYPY_QUEUE_TIMEOUT",),
```
indirectlylit: This naming seems to violate [our convention](https://docs.google.com/document/d/1v-_B7dFlEw9EAOfMSU78IkfhMX9DihzN7jjjxzEUfBk/edit#heading=h.otycu89b2guh) that all env vars should have the `KOLIBRI_` prefix.

At a higher level, is there a strong need to make these configurable? I think we should be careful about expanding the scope of the settings too quickly.

(cc @jamalex)

jredrejo: I was not aware of that convention. I will change it.

About the other question: we talked to make them configurable so we can test the different options in the partners' installations. I think it's worth to have them. The defaults are the optimized parameters we have tuned in our tests but it could happen in other environments with more/less users, better/worse bandwidth, equipement, etc. that the parameters should be changed.

Feel free to close this PR if you consider having these options is a bad idea. In such case I will do a new PR to set the best defaults we can tune in kolibri.

kolibri/utils/options.py
```diff
@@ -32,6 +32,28 @@
             "envvars": ("KOLIBRI_DATABASE_HOST",),
         },
     },
+    "Server": {
+        "SERVER_THREAD_POOL": {
+            "type": "integer",
+            "default": 10,
+            "envvars": ("KOLIBRI_SERVER_THREAD_POOL",),
+        },
```
benjaoming: Not sure if a previous comment of mine has been lost during force-pushing or never made it. Anyways, I would suggest renaming `SERVER_THREAD_POOL` to `CHERRYPY_THREAD_POOL` and the envvar to `KOLIBRI_CHERRYPY_THREAD_POOL`.

kolibri/utils/options.py
```diff
@@ -32,6 +32,28 @@
             "envvars": ("KOLIBRI_DATABASE_HOST",),
         },
     },
+    "Server": {
+        "SERVER_THREAD_POOL": {
```
benjaoming: Not sure if a previous comment of mine has been lost during force-pushing or never made it. Anyways, I would suggest renaming `SERVER_THREAD_POOL` to `CHERRYPY_THREAD_POOL` and the envvar to `KOLIBRI_CHERRYPY_THREAD_POOL`.

The same thing for the other variables.

The reason is that they don't apply to the whole server, and we might have alternatives to CherryPy, disable it etc.

benjaoming (changes requested):
LGTM - except that names can be more accurate.

benjaoming (approved):
LGTM - all my feedback addressed, thanks @jredrejo :)

kolibri/utils/server.py
```diff
@@ -151,7 +151,10 @@ def run_server(port):
     # Configure the server
     server.socket_host = LISTEN_ADDRESS
     server.socket_port = port
-    server.thread_pool = 30
+    server.thread_pool = conf.OPTIONS["Server"]["CHERRYPY_THREAD_POOL"]
+    server.socket_timeout = conf.OPTIONS["Server"]["CHERRYPY_SOCKET_TIMEOUT"]
+    server.accepted_queue_size = conf.OPTIONS["Server"]["CHERRYPY_QUEUE_SIZE"]
+    server.accepted_queue_timeout = conf.OPTIONS["Server"]["CHERRYPY_QUEUE_TIMEOUT"]
```
indirectlylit, line 157: for posterity, I was curious what the cherrypy defaults look like:

| conf | val |
|--|--|
| `cherrypy.server.thread_pool` | 10 |
| `cherrypy.server.socket_timeout` | 10 |
| `cherrypy.server.accepted_queue_size` | -1 |
| `cherrypy.server.accepted_queue_timeout` | 10 |

jredrejo: @indirectlylit  yes, those are the defaults. You can check them in https://github.com/cherrypy/cherrypy/blob/3eb859974d4a5582fefb978916d1b60c7d3adc27/cherrypy/_cpserver.py#L57-L85

Issue #3510, opened 2018-04-10
MVP: auto-resume for content importing from network


indirectlylit:
### Observed behavior

We've repeatedly observed that importing content over a flaky internet connection is a frustrating process because every time a connection times out or otherwise fails, the process needs to be restarted from scratch.

### Expected behavior

If an import fails partway through during the content download phase, the import system should not immediately stop the job and report the import has having failed.

Instead, import command should remain active and try to resume importing where it left off for some period of time, before finally ending and reporting status as failed.

A possible implementation might look something like:

* If an import fails due to timeout or lost connection during the download phase
  * delete any partially-downloaded files if necessary
  * into an 'attempting resume' mode
  * every 5 minutes try to resume
  * if no new data is downloaded after 1 day, report a failure

This change is simply an MVP to make it a bit more viable to download large channels over flaky connections, and does not involve any changes to the UI.

### User-facing consequences

Without this change, users on flaky connections have very little chance of successfully downloading a large channel.

PR #3741, opened 2018-06-01
auto-resume for content importing from network

lyw07:
### Summary

This PR implements auto-resume for network content import. When there is an connection error or timeout error, it will wait for 30 seconds to retry. 
I edited the file `content_test.json` to have two content files that exist on studio server so that I can access the FileDownload.next() instead of only mocking the FileDownload. Please feel free to let me know if there's a better way to handle the unit test. Thank you!

### Reviewer guidance

To test it, please start a network content import, and then turn off the wifi. After some time, turn the wifi back on to continue the import.

### References

https://github.com/learningequality/kolibri/issues/3510

----

### Contributor Checklist

### Reviewer Checklist

### DIFF
--- a/kolibri/content/management/commands/importcontent.py
+++ b/kolibri/content/management/commands/importcontent.py
@@ -5,7 +5,6 @@ from time import sleep
 from django.core.management.base import CommandError
 from requests.exceptions import ConnectionError
 from requests.exceptions import HTTPError
-from requests.exceptions import Timeout
 
 from ...utils import annotation
 from ...utils import import_export_content
@@ -17,6 +16,7 @@ from kolibri.utils import conf
 # constants to specify the transfer method to be used
 DOWNLOAD_METHOD = "download"
 COPY_METHOD = "copy"
+RETRY_STATUS_CODE = [502, 503, 504, 521, 522, 523, 524]
 
 logging = logger.getLogger(__name__)
 
@@ -152,35 +152,12 @@ class Command(AsyncCommand):
 
                 finished = False
                 while not finished:
-                    try:
-                        with filetransfer, self.start_progress(total=filetransfer.total_size) as file_dl_progress_update:
-                            for chunk in filetransfer:
-                                if self.is_cancelled():
-                                    filetransfer.cancel()
-                                    finished = True
-                                    break
-                                length = len(chunk)
-                                overall_progress_update(length)
-                                file_dl_progress_update(length)
-
+                    finished, increment = self._start_file_transfer(
+                        f, filetransfer, overall_progress_update)
+                    if increment == 2:
                         file_checksums_to_annotate.append(f.id)
-                        finished = True
-
-                    # When there is an Internet connection error or timeout error,
-                    # wait for 30 seconds to retry
-                    except (ConnectionError, Timeout) as e:
-                        logging.error("An error occured during content import: {}".format(e))
-                        for i in range(30):
-                            if self.is_cancelled():
-                                self.cancel()
-                                finished = True
-                                break
-                            sleep(1)
-
-                    # When the file cannot be found on the server or disk, skip it
-                    except (HTTPError, OSError):
-                        number_of_skipped_files += 1
-                        overall_progress_update(f.file_size)
+                    else:
+                        number_of_skipped_files += increment
 
             annotation.set_availability(channel_id, file_checksums_to_annotate)
 
@@ -191,6 +168,54 @@ class Command(AsyncCommand):
 
             if self.is_cancelled():
                 self.cancel()
+
+    def _start_file_transfer(self, f, filetransfer, overall_progress_update):
+        """
+        Start to transfer the file from network/disk to the destination.
+        Return value:
+            * True, 2 - successfully transfer the file.
+            * True, 1 - the file does not exist so it is skipped.
+            * True, 0 - the transfer is cancelled.
+            * Fail, 0 - the transfer fails and needs to retry.
+        """
+        try:
+            with filetransfer, self.start_progress(total=filetransfer.total_size) as file_dl_progress_update:
+                for chunk in filetransfer:
+                    if self.is_cancelled():
+                        filetransfer.cancel()
+                        return True, 0
+                    length = len(chunk)
+                    overall_progress_update(length)
+                    file_dl_progress_update(length)
+            return True, 2
+
+        except Exception as e:
+            logging.error("An error occured during content import: {}".format(e))
+
+            # When there is an Internet connection error or timeout error,
+            # or HTTPError where the error code is one of the RETRY_STATUS_CODE,
+            # return False, 0 to retry the file transfer, or return True, 0 to
+            # indicate the cancellation
+            if (isinstance(e, ConnectionError) or
+                    (isinstance(e, HTTPError) and e.response.status_code in RETRY_STATUS_CODE)):
+                return self._sleep_before_retry(), 0
+
+            # Skip the file if it does not exist on the server or disk
+            elif ((isinstance(e, HTTPError) and e.response.status_code == 404) or
+                    (isinstance(e, OSError) and e.errno == 2)):
+                overall_progress_update(f.file_size)
+                return True, 1
+
+            else:
+                raise e
+
+    def _sleep_before_retry(self):
+        for i in range(30):
+            if self.is_cancelled():
+                self.cancel()
+                return True
+            sleep(1)
+        return False
 
     def handle_async(self, *args, **options):
         if options['command'] == 'network':
--- a/kolibri/content/management/commands/importcontent.py
+++ b/kolibri/content/management/commands/importcontent.py
@@ -5,6 +5,7 @@ from time import sleep
 from django.core.management.base import CommandError
 from requests.exceptions import ConnectionError
 from requests.exceptions import HTTPError
+from requests.exceptions import Timeout
 
 from ...utils import annotation
 from ...utils import import_export_content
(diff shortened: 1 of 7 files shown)


kolibri/content/management/commands/importcontent.py
```diff
@@ -147,34 +149,42 @@ def _transfer(self, method, channel_id, path=None, node_ids=None, exclude_node_i
...
-                except HTTPError:
-                    overall_progress_update(f.file_size)
-
-                except OSError:
-                    number_of_skipped_files += 1
-                    overall_progress_update(f.file_size)
+                finished = False
+                while not finished:
+                    try:
+                        with filetransfer:
```
aronasorman: You can stack with statements, decreasing indentation: https://stackoverflow.com/questions/893333/multiple-variables-in-python-with-statement

aronasorman (dismissed):
Some changes

kolibri/content/management/commands/importcontent.py
lyw07: yes, I will do that! thank you for the suggestion!

kolibri/content/utils/transfer.py
```diff
@@ -141,7 +142,10 @@ def __init__(self, *args, **kwargs):
         super(FileDownload, self).__init__(*args, **kwargs)
 
     def start(self):
-        assert not self.started, "File download has already been started, and cannot be started again"
+        # If a file download was stopped by Internet connection error,
+        # then reopen the temp file to write to it
+        if self.started:
+            self.dest_file_obj = open(self.dest_tmp, "ab")
```
indirectlylit: how should I test that this is working as expected? Does it allow the server to be restarted in the middle of a transfer?

kolibri/content/management/commands/importcontent.py
```diff
@@ -147,34 +149,42 @@ def _transfer(self, method, channel_id, path=None, node_ids=None, exclude_node_i
...
+                    except ConnectionError:
+                        for i in range(30):
+                            if self.is_cancelled():
+                                self.cancel()
+                                finished = True
+                                break
+                            sleep(1)
+
+                    # When the file cannot be found on the server or disk, skip it
+                    except (HTTPError, OSError):
```
indirectlylit: This is skipping files on _all_ HTTP errors, not just `404`s. Some (e.g. `503`) should trigger a retry, similar to a timeout. Other `4xx` and `5xx` errors should probably cause an actual import failure.

indirectlylit (commented):
this seems to be working very well - nice job!

I tested a few different variations:

* turned off wifi on computer
* turned off wifi router
* turned off modem

it recovered from all. Left a couple comments and questions inline

kolibri/content/management/commands/importcontent.py
indirectlylit: cc @aronasorman you'd probably have a better sense for how we want to handle various error codes that might be [returned from cloudflare](https://support.cloudflare.com/hc/en-us/articles/115003011431/#cloudflare_specific_codes)

maybe `502`, `503`, `504`, `521`, `522`, `523`, and `524` should all trigger the retry logic?

indirectlylit:
FYI @lyw07 I pushed a commit to your branch to add another decimal point to the import progress.

| Old | New |
|--|--|
| image | image |

kolibri/content/utils/transfer.py
lyw07: Yes, to test it, I just created a new channel `03ff778aac984323a2a23216871022b5` with a large video file. Please turn off the wifi after the file `KOLIBRI_HOME/content/storage/b/f/bf6ce40da9e61bf2c133c3b1ceebf05d.mp4.transfer` exists in the filesystem (It takes about 10-20 seconds after the import starts). And then feel free to turn the wifi back on

kolibri/content/management/commands/importcontent.py
lyw07: I updated the code to catch the `Timeout` error as well so that it should cover the errors with status codes such as `522`, `524`. I think other errors with the status codes that trigger the retry logic should be covered by `ConnectionError`. So when the `HTTPError` is caught, it should only get the errors that don't trigger the retry logic?

kolibri/content/management/commands/importcontent.py
```diff
@@ -147,34 +150,43 @@ def _transfer(self, method, channel_id, path=None, node_ids=None, exclude_node_i
...
+                    # When there is an Internet connection error or timeout error,
+                    # wait for 30 seconds to retry
+                    except (ConnectionError, Timeout) as e:
+                        logging.error("An error occured during content import: {}".format(e))
+                        for i in range(30):
+                            if self.is_cancelled():
+                                self.cancel()
+                                finished = True
+                                break
+                            sleep(1)
```
indirectlylit: Hmm... if I'm reading this correctly, it doesn't actually retry the file!

From what I can tell, it will go back up to the [top of the loop](https://github.com/learningequality/kolibri/pull/3741/files#diff-303c0883bcd56c13a0e4d895905cc6d6R131) and try to transfer the _next_ `f`, not retry the _current_ `f`.

lyw07: I think my logic here is
ConnectionError/Timeout --> sleep for 30 seconds --> `finished` still equals to `False` --> still in the while loop (line 154) --> retry the current `f`
or
ConnectionError/Timeout --> sleep for several seconds but user cancels the import at some point --> `finished = True` --> exit the while loop --> transfer the next `f`

So as long as the user doesn't cancel the import, it should go to line 154 to retry the current `f` I think?

kolibri/content/management/commands/importcontent.py
indirectlylit: There are three possible outcomes when we encounter either `HTTPError` or `OSError`:

1. _retry_ the same file again
2. _skip_ the file and try the next one
3. _fail_ the entire transfer

| Exception | Condition | Outcome |
|--|--|--|
| `HTTPError` | `e.code == 404` | skip |
| `HTTPError` | `e.code in (502, 503, 504, 521, 522, 523, 524)` | retry |
| `HTTPError` | `else` | fail |
| `OSError` | `e.errno == errno.ENOENT` | skip |
| `OSError` | `else` | fail |

indirectlylit: > I updated the code to catch the Timeout error as well so that it should cover the errors with status codes such as 522, 524. I think other errors with the status codes that trigger the retry logic should be covered by ConnectionError. So when the HTTPError is caught, it should only get the errors that don't trigger the retry logic?

I'm having trouble finding documentation which says that 5xx error codes raise `Timeout` or `ConnectionError`. If so, that's great! Do you have references?

kolibri/content/management/commands/importcontent.py
indirectlylit: Ah, I see - you are correct. My fault!

There might be some opportunities here to break out some helper functions and reduce the depth of loop nesting. For example, either the stuff inside `while not finished:` or `for f in files_to_download:` could potentially be its own helper function. I wouldn't worry about it for this PR - just a side note.

kolibri/content/management/commands/importcontent.py
lyw07: Thank you for the table! It's very clear information.
I didn't find documentation that explains specific error codes for different exceptions, but I found [this doc](https://github.com/requests/requests/blob/v2.18.4/docs/user/quickstart.rst#errors-and-exceptions) that somewhat explains what errors cause `ConnectionError` and `Timeout`. It is my assumption that the errors mentioned in the doc above are related to certain error codes...

indirectlylit: would it be possible to write tests which emulate the five scenarios above and ensure that the expected behavior happens?

lyw07: hmm I'm not sure about emulating the second scenario, but I can try with the other four first!

indirectlylit: yeah I believe you should be able to mock the `filetransfer` object with some dummy code that raises variations of the exceptions listed in the table? I think @jamalex wrote that wrapper so he might have some pointers

lyw07: I think I find the way to test the second scenario now, will update the tests soon!

kolibri/content/management/commands/importcontent.py
jamalex: Another pattern we could try here in future is throwing a custom `TransferCanceled` exception that gets caught at a top level, so we don't need to have as many flags etc.

jamalex: We will also want to explore (in a future PR) the possibility of _resuming_ downloads when they error partway (looking at how much is already in the `.transfer` file, and doing byte-range requests to just read from the URL starting at a particular offset). I think in Kenya I tried download the same single video like a dozen times, starting from scratch each time.

kolibri/content/management/commands/importcontent.py
jamalex: Since the filetransfer object uses `requests.get`, the most robust method might be to mock that to return those status codes, and then see how the error bubbles up to your code.

kolibri/content/utils/transfer.py
jamalex: Hmm, you're appending, but might it be starting the download from the beginning (if the previous connected timed out, for example)? Have you checked that it doesn't end up with a larger than expected (and corrupt) file?

jamalex (commented):
Nice! Some questions.

kolibri/content/utils/transfer.py
lyw07: Yes I checked that the file size is the same as the original from studio, and it's not corrupted.

kolibri/content/management/commands/importcontent.py
lyw07: @jamalex Thank you for the suggestions! Will implement these in a future PR :)

kolibri/content/utils/transfer.py
jamalex: I've been playing with this for a while, and I have a few things I learned in the process.

Due to [this line](https://github.com/learningequality/kolibri/blob/develop/kolibri/content/utils/transfer.py#L141), we often download content in big chunks. On my computer, downloading the 90MB video file above doesn't ping the progress callback or write to disk at all until it's completed (if I comment out that line, I get granular progress reporting again). This made me worried that perhaps you didn't see corruption because it was starting from an empty file when "resuming". However, when I remove that line and have it use smaller chunks, I don't see corruption either. But when you interrupt the connection and let it time out, then reconnect, it seems to start downloading the file from the beginning anyway (the file size and progress bar both start back at 0). So I don't think this chunk of code is doing what you intended anyway.

I'd recommend we remove the "resume partial file" code here, as I don't think it's working, and then open an issue to address that (which I think can probably be done using the "byte-range" method I discussed in another comment). Instead, for now probably best to just cancel/close the file object and start a new one, if it times out.

jamalex (changes requested):
I did some testing, and added another comment. I think some of my earlier comment threads were lost, due to a rebase/force-push, so I can't tell whether they were addressed. Can you still access them via your emails? (And I guess it's best not to do any sort of rebasing until the very end, or before reviews start)?

kolibri/content/utils/transfer.py
lyw07: thank you for testing @jamalex ! I probably rebased before, but I saw all the comments are still in the `conversation` page. Some of them are hidden under `show outdated`. But thank you for the suggestion! I will not rebase so early in the future PRs.
I agree that I didn't see the corruption because the `.transfer` file was empty. 
Since the file object will be closed when it gets an exception and exits the `with` statement, I think I'm going to just change the open mode from `ab` to `wb` to start the new `.transfer` file object.

jamalex (approved):

indirectlylit:
bam

PR #3757, opened 2018-06-04
Webpack4 upgrade

rtibbles:
### Summary
Upgrades frontend build process to use webpack 4.

### Reviewer guidance
Does it build?

In spite of not producing particular performance improvements, am still submitting this to keep us up to date.

----

### Contributor Checklist

### Reviewer Checklist


indirectlylit:
> Does it build?

almost!

```
  Error: Cannot find module 'extract-text-webpack-plugin'
```

rtibbles:
Well, it builds locally - just not for karma. Actually having trouble getting karma to run at all, unfortunately.

rtibbles:
It now builds.

kolibri/core/assets/test/heartbeat.spec.js
```diff
@@ -0,0 +1,221 @@
+/* eslint-env mocha */
```
jonboiser: should probably change this to /* eslint-env jest */

rtibbles: Yeah, can actually get rid of all of these, because I just added the jest env to the base eslint config.

rtibbles: And will set that to only apply to .spec.js files, so that we don't let jest globals weirdly sneak into our client source code.

rtibbles:
Have now added an env override for `spec.js` files so that they are automatically treated as `jest` env. This means we shouldn't need any special eslint comments in any test files.

__mocks__/http.js
```diff
@@ -0,0 +1,38 @@
...
+  request.end();
+  return request;
+}
+
+http.__setCode = __setCode;
+http.__setEntity = __setEntity;
+http.__setHeaders = __setHeaders;
+http.request = request;
+
+module.exports = http;
```
indirectlylit, line 38: I don't see any reference to this file, and it has no comments indicating what it's used by.

* Is there a better location for it?
* can we add a comment to the top explaining why it exists?

indirectlylit (approved):
builds and runs!

indirectlylit:
tests run too - neat new UI!

Issue #3635, opened 2018-05-09
Support variable page heights in pdf renderer 


christianmemije:
### Observed behavior

Currently variable page heights is not supported.

### Expected behavior

Variable page heights to be supported.
https://github.com/Akryum/vue-virtual-scroller#variable-height-mode

### User-facing consequences

Unusable pdf.

### Steps to reproduce

Check out a african story book pdf.

christianmemije:
Addressed in #3760

PR #3760, opened 2018-06-04
Handle pdfs that have different sized pages

christianmemije:
### Summary

Handle pdfs that have different sized pages
Addresses #3635

### Reviewer Guidance
View a pdf that has different sized pages. 
The Global Digital Library - Book Catalog channel and the English topic can come in handy

----

### Contributor Checklist

### Reviewer Checklist

### DIFF
--- a/kolibri/plugins/document_pdf_render/assets/src/views/index.vue
+++ b/kolibri/plugins/document_pdf_render/assets/src/views/index.vue
@@ -6,7 +6,7 @@
     allowfullscreen
   >
     <k-linear-loader
-      v-if="documentLoading"
+      v-if="documentLoading || firstPageHeight === null"
       class="progress-bar"
       :delay="false"
       :type="progress > 0 ? 'determinate' : 'indeterminate'"
@@ -17,7 +17,7 @@
       <recycle-list
         ref="recycleList"
         :items="pdfPages"
-        :itemHeight="pageHeight + 16"
+        :itemHeight="itemHeight"
         :emitUpdate="true"
         :style="{ height: `${elSize.height}px` }"
         class="pdf-container"
@@ -26,13 +26,13 @@
       >
         <template slot-scope="{ item }">
           <pdf-page
-            :pdfPage="item.page"
             :key="item.index"
             :pageNum="item.index + 1"
-            :pageReady="item.resolved"
+            :pdfPage="pdfPages[item.index].page"
+            :pageReady="pdfPages[item.index].resolved"
+            :firstPageHeight="firstPageHeight || 0"
+            :firstPageWidth="firstPageWidth || 0"
             :scale="scale || 1"
-            :defaultHeight="pageHeight || 0"
-            :defaultWidth="pageWidth || 0"
           />
         </template>
       </recycle-list>
@@ -95,8 +95,8 @@
 
   // How often should we respond to changes in scrolling to render new pages?
   const renderDebounceTime = 300;
-
   const scaleIncrement = 0.25;
+  const MARGIN = 16;
 
   export default {
     name: 'pdfRender',
@@ -113,8 +113,6 @@
       scale: null,
       timeout: null,
       totalPages: null,
-      pageHeight: null,
-      pageWidth: null,
       firstPageHeight: null,
       firstPageWidth: null,
       pdfPages: [],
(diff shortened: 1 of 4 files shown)


kolibri/plugins/document_pdf_render/assets/src/views/index.vue
```diff
@@ -17,7 +17,7 @@
       <recycle-list
         ref="recycleList"
         :items="pdfPages"
-        :itemHeight="pageHeight + 16"
+        :itemHeight="itemHeight"
```
indirectlylit, line 20: nice change

kolibri/plugins/document_pdf_render/assets/src/views/index.vue
```diff
@@ -113,8 +113,6 @@
       scale: null,
       timeout: null,
       totalPages: null,
-      pageHeight: null,
-      pageWidth: null,
```
indirectlylit, line 117: nice

kolibri/plugins/document_pdf_render/assets/src/views/pdfPage.vue
```diff
@@ -43,27 +43,41 @@
...
-        return this.height || this.defaultHeight;
+      actualHeight() {
+        if (this.pageReady) {
+          return this.pdfPage.view[3];
+        }
+      },
+      actualWidth() {
+        if (this.pageReady) {
+          return this.pdfPage.view[2];
+        }
```
indirectlylit: It's good to explicitly return `null` so the behavior looks intentional and well-defined.

In cases like this, I personally like to check for the exceptional case (and bail early) rather than the primary case so that the primary logic isn't indented:

```javascript
actualWidth() {
  if (!this.pageReady) {
    return null;
  }
  return this.pdfPage.view[2];
}
```

indirectlylit (commented):
these changes look good. I haven't tested yet

kolibri/plugins/document_pdf_render/assets/src/views/index.vue
```diff
@@ -177,28 +176,17 @@
...
-            this.windowSize.breakpoint > 3
-              ? 1
-              : (this.elSize.width - 2 * pageMargin) / this.firstPageWidth;
-
-          // set default height and width properties
-          const initialViewport = firstPage.getViewport(this.scale);
-          this.pageHeight = initialViewport.height;
-          this.pageWidth = initialViewport.width;
+          this.firstPageWidth = firstPage.view[2];
+          this.scale = this.elSize.height / (this.firstPageHeight + MARGIN);
```
rtibbles, line 182: Ah, so we dynamically set the scale, based on the size of the page and the available viewport, rather than setting the viewport by the scale? Neat.

kolibri/plugins/document_pdf_render/assets/src/views/pdfPage.vue
```diff
@@ -43,27 +43,43 @@
...
+        return this.pdfPage.view[3];
+      },
+      actualWidth() {
+        if (!this.pageReady) {
+          return null;
+        }
+        return this.pdfPage.view[2];
+      },
+      heightToWidthRatio() {
+        return this.actualHeight / this.actualWidth || this.firstPageHeight / this.firstPageWidth;
```
rtibbles, line 73: To check I am understanding it correctly - what we are doing here is differentially scaling pages if they are of a different size, so that they all take up the same amount of vertical space, regardless of their dimensions, right?

rtibbles (approved):
Code looks good. Manually tested and works like a dream.

kolibri/plugins/document_pdf_render/assets/src/views/index.vue
christianmemije: Yup, it isn't ideal, but it is definitely much better than loading the entire pdf. Plus we have those handy zoom buttons.

kolibri/plugins/document_pdf_render/assets/src/views/pdfPage.vue
christianmemije: Exactly.

indirectlylit: maybe add an explanatory comment to that effect?

indirectlylit: whoops we're already merged

PR #3743, opened 2018-06-01
Conditionally encode when reading stopword file

ralphiee22:
### Summary

Specify encoding for reading stopwords file on python2/3.

### Reviewer guidance

…

### References

Hopefully solves this problem https://github.com/learningequality/kolibri-installer-windows/issues/72#issuecomment-393454573

----

### Contributor Checklist

- [ ] Contributor has fully tested the PR manually
- [x] PR has the correct target branch and milestone
- [x] PR has 'needs review' or 'work-in-progress' label
- [x] If PR is ready for review, a reviewer has been added. (Don't use 'Assignees')
- [ ] If there are any front-end changes, before/after screenshots are included
- [ ] If this is an important user-facing change, PR or related issue has a 'changelog' label

### Reviewer Checklist

- [ ] Automated test coverage is satisfactory
- [ ] Reviewer has fully tested the PR manually
- [ ] PR has been tested for [accessibility regressions](http://kolibri-dev.readthedocs.io/en/develop/manual_testing.html#accessibility-a11y-testing)
- [ ] External dependencies files were updated (`yarn` and `pip`)
- [ ] Documentation is updated
- [ ] Link to diff of internal dependency change is included
- [ ] CHANGELOG.rst is updated for high-level changes
- [ ] Contributor is in AUTHORS.rst

Issue #2541, opened 2017-10-27
Changing language settings on devserver results in UnicodeException


alanchenz:
### Observed behavior

When I try to change the language to Arabic, Hindi, or Portuguese (unicode languages), the app crashes and throws UnicodeDecodeError.

### Expected behavior

After changing language, app should still run with all text in the selected language.

### User-facing consequences

Users in areas that use such languages would not be able to use the app to their language preferences.

### Errors and logs

Relevant errors and tracebacks from:

* the command line
* `~/.kolibri/kolibri.log`
* the browser console

```
ERROR    Internal Server Error: /management/facility
Traceback (most recent call last):
  File "c:\users\alanc\kolibri\kolibri\lib\site-packages\django\core\handlers\base.py", line 174, in get_response
    response = self.process_exception_by_middleware(e, request)
  File "c:\users\alanc\kolibri\kolibri\lib\site-packages\django\core\handlers\base.py", line 172, in get_response
    response = response.render()
  File "c:\users\alanc\kolibri\kolibri\lib\site-packages\django\template\response.py", line 160, in render
    self.content = self.rendered_content
  File "c:\users\alanc\kolibri\kolibri\lib\site-packages\django\template\response.py", line 137, in rendered_content
    content = template.render(context, self._request)
  File "c:\users\alanc\kolibri\kolibri\lib\site-packages\django\template\backends\django.py", line 95, in render
    return self.template.render(context)
  File "c:\users\alanc\kolibri\kolibri\lib\site-packages\django\template\base.py", line 206, in render
    return self._render(context)
  File "c:\users\alanc\kolibri\kolibri\lib\site-packages\django\template\base.py", line 197, in _render
    return self.nodelist.render(context)
  File "c:\users\alanc\kolibri\kolibri\lib\site-packages\django\template\base.py", line 992, in render
    bit = node.render_annotated(context)
  File "c:\users\alanc\kolibri\kolibri\lib\site-packages\django\template\base.py", line 959, in render_annotated
    return self.render(context)
  File "c:\users\alanc\kolibri\kolibri\lib\site-packages\django\template\loader_tags.py", line 173, in render
    return compiled_parent._render(context)
  File "c:\users\alanc\kolibri\kolibri\lib\site-packages\django\template\base.py", line 197, in _render
    return self.nodelist.render(context)
  File "c:\users\alanc\kolibri\kolibri\lib\site-packages\django\template\base.py", line 992, in render
    bit = node.render_annotated(context)
[...]

PR #3769, opened 2018-06-07
switch from open to io.open for text file read/write

indirectlylit:
### Summary

In https://github.com/learningequality/kolibri-installer-windows/issues/72 we noticed that characters aren't being rendered properly in Windows using Python 3

This PR attempts to find _all_ instances of `open` that were used on text files, and replace them with `io.open`, using an explicit encoding.

### Reviewer guidance

Please check that this all makes sense

### References

* https://github.com/learningequality/kolibri/commit/715c2907e97e1937f61d1f13c8cbde3273553b31
* https://github.com/learningequality/kolibri/commit/e8d9aead2554dc5792cd9565be864a9e846cfd56
* https://github.com/learningequality/kolibri/pull/3743

----

### Contributor Checklist

### Reviewer Checklist

### DIFF
--- a/build_tools/customize_build.py
+++ b/build_tools/customize_build.py
@@ -5,6 +5,7 @@ the DJANGO_SETTINGS_MODULE environment variable.
 
 For more detail see the documentation in __init__.py
 """
+import io
 import os
 import sys
 import tempfile
@@ -26,11 +27,11 @@ def load_plugins_from_file(file_path):
         if file_path.startswith('http'):
             print("Downloading plugins manifest from {file_path}".format(file_path=file_path))
             _, path = tempfile.mkstemp(suffix=".txt", text=True)
-            with open(path, 'w') as f:
+            with io.open(path, mode='w', encoding='utf-8') as f:
                 r = requests.get(file_path)
                 f.write(r.content)
             file_path = path
-        with open(file_path, 'r') as f:
+        with io.open(file_path, mode='r', encoding='utf-8') as f:
             plugins_cache[file_path] = [plugin.strip() for plugin in f.readlines() if plugin.strip()]
     return plugins_cache[file_path]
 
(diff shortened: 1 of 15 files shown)


build_tools/customize_build.py
```diff
@@ -26,11 +27,11 @@ def load_plugins_from_file(file_path):
         if file_path.startswith('http'):
             print("Downloading plugins manifest from {file_path}".format(file_path=file_path))
             _, path = tempfile.mkstemp(suffix=".txt", text=True)
-            with open(path, 'w') as f:
+            with io.open(path, mode='w', encoding='utf-8') as f:
                 r = requests.get(file_path)
                 f.write(r.content)
             file_path = path
-        with open(file_path, 'r') as f:
+        with io.open(file_path, mode='r', encoding='utf-8') as f:
```
mrpau-richard: @indirectlylit I think you need to use the `KWARGS_IO_READ` when reading file. `mode='w', encoding='utf-8'` will only work at the `Python 3`.
```
# Use encoded text for Python 3 (doesn't work in Python 2!)
KWARGS_IO_READ = {'mode': 'r', 'encoding': 'utf-8'}
KWARGS_IO_WRITE = {'mode': 'w', 'encoding': 'utf-8'}

# Use binary file mode for Python 2 (doesn't work in Python 3!)
if sys.version_info < (3,):
    KWARGS_IO_READ = {'mode': 'rb'}
    KWARGS_IO_WRITE = {'mode': 'wb'}
```

mrpau-richard (changes requested):

build_tools/customize_build.py
indirectlylit: My understanding was that by switching from `open` to `io.open` I could avoid this discrepancy?

* https://docs.python.org/3.3/howto/pyporting.html#text-files
* https://stackoverflow.com/a/10975371

indirectlylit:
hmm looks like it's not building... checking on it

build_tools/customize_build.py
mrpau-richard: Yeah, I tested it. `KWARGS_IO_READ` not needed here.

mrpau-richard (approved):

build_tools/customize_build.py
mrpau-richard: Waiting for the build to test.

mrpau-richard:
@indirectlylit the fix confirmed at the Windows 7. Non-english characters are now displaying properly.
screen shot 2018-06-08 at 11 21 54 am

indirectlylit:
great!

might also help with #2541

Issue #3384, opened 2018-03-19
Building dist is slow because compilemessages includes kolibri/dist


benjaoming:
### Observed behavior

Much of the build time is spent like this:

![screenshot from 2018-03-19 11-50-16](https://user-images.githubusercontent.com/374612/37591494-07ca2082-2b6c-11e8-9c40-41ec5cd26bdb.png)

### Expected behavior

Should skip `kolibri/dist` since everything there is third-party.

### User-facing consequences

n/a

### Errors and logs

n/a

### Steps to reproduce

`make dist`

### Context

n/a

indirectlylit:
would be great to speed things up a bit

benjaoming:
Sorry, I didn't see the new milestone, just went and fixed it immediately :)

rtibbles:
This has been fixed already.

PR #3776, opened 2018-06-08
Improve build time (theoretically)  by compiling messages *before* populating kolibri/dist

benjaoming:
### Summary

There's no immediate way to exclude child folders when running `compilemessages` (that would be an interesting upstream fix).

However, by moving the call to `compilemessages` in before calling `staticdeps`, then `kolibri/dist` will be empty provided that it's a clean environment. Otherwise run `make clean`.

### Reviewer guidance

n/a - if builds work I say go?

### References

#3384

----

### Contributor Checklist

### Reviewer Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -152,7 +152,7 @@ buildconfig:
 	git checkout -- kolibri/utils/build_config # restore __init__.py
 	python build_tools/customize_build.py
 
-dist: writeversion staticdeps staticdeps-cext buildconfig assets translation-django-compilemessages
+dist: writeversion translation-django-compilemessages staticdeps staticdeps-cext buildconfig assets
 	python setup.py sdist --format=gztar --static > /dev/null # silence the sdist output! Too noisy!
 	python setup.py bdist_wheel --static
 	ls -l dist


benjaoming:
@indirectlylit this is ready to review.

indirectlylit (changes requested):
Can you share your numbers? I'm not reproducing a speed up.

Here's what I see:


pre-change:

```
Start +2018-06-12T15:33:35-0700
End   +2018-06-12T15:38:32-0700
                      4:53
```

post-change

```
Start +2018-06-12T15:47:42-0700
End   +2018-06-12T15:53:05-0700
                      5:23
```

benjaoming:
@indirectlylit is your `kolibri/dist` clean when you compile the *before* numbers?

benjaoming:
@indirectlylit Sorry if what I meant by "speed up" was perhaps a bit of an overstatement. You cannot measure runtime using a system's clock. You need to count CPU cycles, memory, i/o etc.

indirectlylit:
>  is your `kolibri/dist` clean when you compile the before numbers?

Yes, I ran `make clean` before each run.

Obviously not a very scientific test - I only ran it one time - but it looked 10% slower which is why I wanted to make sure we're not inadvertently making the situation worse.

benjaoming:
Am including this in #4056 after looking at the same message gathering outputs for too many times while dev'ing :)

benjaoming:
But you are right, this doesn't work very well.

I see that because of calling `kolibri manage compilemessages`, we are also going through migrations etc.. the initial kolibri setup.

```
python -c "import kolibri; print(kolibri.__version__)" > kolibri/VERSION

Current version is now 0.11.0a3.dev0+git.191.g3e2b898c0
# Change working directory to kolibri/ such that compilemessages
# finds only the .po files nested there.
cd kolibri && PYTHONPATH="..:$PYTHONPATH" python -m kolibri manage compilemessages
INFO     Kolibri running for the first time.
INFO     We don't yet use pre-migrated database seeds, so you're going to have to wait a bit while we create a blank database...

```

benjaoming:
I'm going to have a look at this in #4056 

Also, there's this perhaps new problem, these are all-together easier to solve in one go, because I'm invoking the same build commands so many times.

image

PR #3786, opened 2018-06-11
Parameterize core-modal buttons + more

### LINKED ISSUE #3311: fix core modal @enter events, parameterize buttons, and conform to style guide
### Observed behavior

In many cases in the code base - [example](https://github.com/learningequality/kolibri/blob/develop/kolibri/plugins/coach/assets/src/views/exams-page/rename-exam-modal.vue) - we use a pattern like this:

```html
  <core-modal @cancel="closeHandler">
    <form @submit.prevent="submitHandler">
      ...
      <div class="core-modal-buttons">
        <k-button
          type="button"
          :text="$tr('cancel')"
          appearance="flat-button"
          @click="close"
        />
        <k-button
          type="submit"
          :text="$tr('save')"
          :primary="true"
        />
      </div>
    </form>
  </core-modal>
```

instead of the expected pattern like this:

```html
  <core-modal @cancel="closeHandler" @enter="submitHandler">
    ...
    <div class="core-modal-buttons">
      <k-button
        type="button"
        :text="$tr('cancel')"
        appearance="flat-button"
        @click="close"
      />
      <k-button
        type="submit"
        :text="$tr('save')"
        :primary="true"
        :disabled="submitting"
      />
    </div>
  </core-modal>
```

It might be that this is a work-around for double-submission or some other bug?

### Expected behavior

We shouldn't need to add a form with special event handling in every modal. If this pattern is necessary it should be built into the core modal.

While we're at it, we can update `core-modal` to take the button configuration as prop inputs rather than slots. This would allow us to deprecate the `core-modal-buttons` div convention:
One possible interface:

```html
  <core-modal
    @cancel="closeHandler"
    @enter="submitHandler"
    :secondaryBtn="$tr('cancel')"
    :primaryBtn="$tr('save')"
  >
    ...
  </core-modal>
```

### User-facing consequences

potential for bugs in modal submission due to difficult developer experience

### Context

conversation here: https://github.com/learningequality/kolibri/pull/3301#discussion_r172960515

christianmemije:
### Summary

DRYs up usage of core-modal by taking in button props.
Fix #3311 

### Reviewer guidance

Test every single modal please. 

----

### Contributor Checklist

### Reviewer Checklist


indirectlylit:
looking good! always a good sign:

image

let's also update the style guide as part of this

kolibri/core/assets/src/views/core-modal.vue
```diff
@@ -56,40 +101,57 @@
...
       },
-      invisibleTitle: {
+      submitText: {
+        type: String,
+        required: false,
+      },
+      cancelText: {
+        type: String,
+        required: false,
+      },
```
indirectlylit: I'm a bit torn. On one hand, I appreciate that this saves a bunch of boilerplate. On the other hand, keeping it as a slot would be more flexible if (and probably when) we decide to make an exception to the rule and change the button arrangement.

My inclination is to go with the more flexible option and keep it a slot, or at least to preserve that opportunity if need-be...

christianmemije: Added an actions slot.

kolibri/core/assets/src/views/k-modal.vue
```diff
@@ -0,0 +1,312 @@
...
+   */
+  export default {
+    name: 'kModal',
+    components: {
+      kButton,
+    },
+    mixins: [responsiveWindow],
+    $trs: {
+      // error alerts
+      errorAlert: 'Error in',
```
indirectlylit: this string should be parameterized like `Error in {field}`

(I realize this was a pre-existing issue)

kolibri/core/assets/src/views/k-modal.vue
```diff
@@ -0,0 +1,312 @@
+<template>
+
+  <!-- Accessibility properties for the overlay -->
+  <transition name="fade">
+    <div
+      class="modal-overlay"
+      @keyup.esc="emitCancelEvent"
```
christianmemije, line 7: Replaced keydown with keyup. More standard.

kolibri/core/assets/src/views/k-modal.vue
```diff
@@ -0,0 +1,312 @@
...
+    },
+    destroyed() {
+      window.removeEventListener('focus', this.focusElementTest, true);
+      window.removeEventListener('scroll', this.preventScroll, true);
+      // Wait for events to finish propagating before changing the focus.
+      // Otherwise the `lastFocus` item receives events such as 'enter'.
+      window.setTimeout(() => this.lastFocus.focus());
+    },
+    methods: {
+      setContentSectionMaxHeight: debounce(function() {
```
indirectlylit: could you add some comments about the purpose and intended behavior of this function?

kolibri/core/assets/src/views/k-modal.vue
```diff
@@ -0,0 +1,312 @@
...
+        if (this.$refs.title && this.$refs.actions) {
+          this.maxContentHeight =
+            this.windowSize.height -
+            this.$refs.title.clientHeight -
+            this.$refs.actions.clientHeight -
+            32;
+        }
+      }, 100),
+      emitCancelEvent() {
```
christianmemije: Oops. I guess both of these events should not be emitted if submitDisabled/cancelDisabled.

kolibri/core/assets/src/views/k-modal.vue
```diff
@@ -0,0 +1,312 @@
...
+    overflow-y: auto
+
+  .actions
+    text-align: right
+    padding: 24px
+    button
+      margin: 0
+
+  .actions button:last-of-type
+    margin-left: 16px
```
indirectlylit, line 324: recommend:

* remove the [built-in margin](https://github.com/learningequality/kolibri/blob/develop/kolibri/core/assets/src/views/buttons-and-links/buttons.styl#L30) on buttons
* add a new core component, `k-button-group`, which is responsible for adding margins between buttons

kolibri/plugins/coach/assets/src/views/assignments/AssignmentCopyModal.vue
```diff
@@ -1,65 +1,45 @@
...
 
-  <core-modal
-    :title="modalTitle"
-    @cancel="closeModal()"
-  >
+  <div>
     <!-- Classroom Selection Form -->
-    <div v-if="stage===Stages.SELECT_CLASSROOM">
+    <k-modal
+      v-if="stage===Stages.SELECT_CLASSROOM"
```
indirectlylit: `stage === Stages.SELECT_CLASSROOM`

(pre-existing issue)

kolibri/plugins/coach/assets/src/views/exams/exams-page/preview-exam-modal.vue
```diff
@@ -1,22 +1,18 @@
...
-      :primary="false"
-      @click="close"
-    />
     <transition mode="out-in">
       <k-circular-loader
         v-if="loading"
         :delay="false"
       />
-      <div v-else>
+      <div v-else @keyup.enter.stop>
```
indirectlylit, line 15: why is `@keyup.enter.stop` necessary here? (add a comment)

kolibri/plugins/facility_management/assets/src/views/user-page/reset-user-password-modal.vue
```diff
@@ -1,68 +1,50 @@
...
+  <k-modal
     :title="$tr('resetPassword')"
+    :submitText="$tr('save')"
+    :cancelText="$tr('cancel')"
+    :submitDisabled="isBusy"
+    @submit="submitForm"
     @cancel="displayModal(false)"
   >
-    <form @submit.prevent="submitForm">
+    <p>{{ $tr('username') }}: <strong>{{ username }}</strong></p>
```
indirectlylit: the `:` should be moved inside the `username` string

(I realize this was a pre-existing issue)

kolibri/core/assets/src/views/k-modal.vue
indirectlylit: I don't actually believe this is true?

indirectlylit (approved):
* This looks good, but requires some conflict resolution.
* Filed one follow-up issue: #3813 
* If you haven't started yet, don't worry about the 'button group' comments
* would recommend keeping the event on button `down` instead of `up` because it's consistent with `click` events .. but curious if you have other documentation showing that `up` is actually more conventional

PR #3787, opened 2018-06-12
0.10.x into develop

rtibbles:


indirectlylit (approved):
tests are failing for unknown reason, but it's important to get develop synchronized. let's merge now and ask forgiveness later

PR #3800, opened 2018-06-13
Add triggered android apk build pipeline

rtibbles:
### Summary
Re-enables Android builds as a triggered pipeline so as not to hold up builds.

Testing PR

This seems like a useful model as we try to automate building of more and more installation targets? Also, we could restrict the branches to just release branches if needed.

### Contributor Checklist

### Reviewer Checklist

### DIFF
--- a/.buildkite/build_apk.sh
+++ b/.buildkite/build_apk.sh
@@ -6,7 +6,7 @@ set -euo pipefail
 yes y | rm -r kolibri-kivy/ || true 2> /dev/null
 git clone https://github.com/learningequality/kolibri-kivy.git
 cd kolibri-kivy
-git checkout v0.1.0-beta1
+git checkout develop
 cd ..
 
 # copy in the latest whl to make it available to the Docker build script
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -14,10 +14,10 @@ steps:
 
   - wait
 
-  # - label: Build the Android APK file
-  #   command: mkdir -p dist && .buildkite/build_apk.sh
-
-  # - wait
-
   - label: Upload artifacts
     command: .buildkite/setup_and_upload_artifact.sh && docker image prune -f -a
+
+  - block: "Create release artifacts?"
+
+  - label: Build the Android APK file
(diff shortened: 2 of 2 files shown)


.buildkite/build_apk.sh
```diff
@@ -6,7 +6,7 @@ set -euo pipefail
 yes y | rm -r kolibri-kivy/ || true 2> /dev/null
 git clone https://github.com/learningequality/kolibri-kivy.git
 cd kolibri-kivy
-git checkout v0.1.0-beta1
+git checkout develop
```
aronasorman, line 9: I prefer if we pin this to a specific version, based on my experience with the KA Lite windows installer. Pinning it allows us to track which installers-repo commit produced an android installer.

.buildkite/pipeline.yml
```diff
@@ -14,10 +14,10 @@ steps:
...
 
-  # - label: Build the Android APK file
-  #   command: mkdir -p dist && .buildkite/build_apk.sh
-
-  # - wait
-
   - label: Upload artifacts
     command: .buildkite/setup_and_upload_artifact.sh && docker image prune -f -a
+
+  - block: "Create release artifacts?"
```
aronasorman, line 20: Yes pls.

aronasorman (changes requested):

.buildkite/build_apk.sh
rtibbles: My reason for not pinning this right now was that my assumption is that there will be rapid iteration on this for 0.11. Once we have stabilized for release again, very happy to pin it.

aronasorman: Shouldn't we do rapid iteration on the android repo itself then, rather than a workflow comprising of `push test commit on android repo` -> `trigger a build on kolibri`?

rtibbles: Except we have no way of triggering the build from the android repo, which is why having it unpinned in Kolibri is helpful. How do we do iteration on the windows installer repo currently?

aronasorman (approved):
I'm approving this, with the assumption that we'll be changing the apk commit it's pulling in to a tag or commit.

PR #3808, opened 2018-06-15
fix scrolling

### LINKED ISSUE #3793: scroll bar should be on edge of screen, not within margin
### Observed behavior

regression caused by #3782 - scroll bar has margin to its right:

screen shot 2018-06-13 at 11 09 01 am

reported here: https://github.com/learningequality/kolibri/pull/3778#issuecomment-396801750

### Expected behavior

scroll bar should be on page edge

### User-facing consequences

layout looks broken

### Context

0.10.0-beta2

indirectlylit:
### Summary

fixes #3793

----

### Contributor Checklist

### Reviewer Checklist

### DIFF
--- a/kolibri/core/assets/src/views/app-body.vue
+++ b/kolibri/core/assets/src/views/app-body.vue
@@ -10,8 +10,8 @@
       :delay="false"
     />
     <template v-else>
-      <error-box v-if="error" />
       <div class="wrapper">
+        <error-box v-if="error" />
         <slot></slot>
       </div>
     </template>
--- a/kolibri/core/assets/src/views/app-body.vue
+++ b/kolibri/core/assets/src/views/app-body.vue
@@ -11,7 +11,9 @@
     />
     <template v-else>
       <error-box v-if="error" />
-      <slot></slot>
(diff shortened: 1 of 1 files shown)


christianmemije (changes requested):
The error-box needs to also have this max-width, so instead wrap the entire template.

```html
<div class="wrapper" v-else>
  <error-box v-if="error" />
  <slot></slot>
</div>
```

indirectlylit:
good catch. pushed update

christianmemije (approved):
Acceptable.

kolibri/core/assets/src/views/app-body.vue
```diff
@@ -10,8 +10,10 @@
       :delay="false"
     />
     <template v-else>
```
christianmemije, line 12: You could have replaced the template with the div.

indirectlylit: good catch - PR here #3817

indirectlylit: (I see now that's what you wrote above originally)

PR #3795, opened 2018-06-13
updates to developer docs

indirectlylit:
### Summary

Wide-ranging set of changes including:

* fixing errors
* making headers and language consistent
* updating and deleting outdated information
* adding a new section on development workflow
* updating release process

### Reviewer guidance

In particular, please review _development_workflow.html_ which is a new document and tries to outline some higher-level advice about development on Kolibri.

----

### Contributor Checklist

- [x] Contributor has fully tested the PR manually
- [x] PR has the correct target branch and milestone
- [x] PR has 'needs review' or 'work-in-progress' label
- [x] If PR is ready for review, a reviewer has been added. (Don't use 'Assignees')
- [ ] If there are any front-end changes, before/after screenshots are included
- [ ] If this is an important user-facing change, PR or related issue has a 'changelog' label

### Reviewer Checklist

- [ ] Automated test coverage is satisfactory
- [ ] Reviewer has fully tested the PR manually
- [ ] PR has been tested for [accessibility regressions](http://kolibri-dev.readthedocs.io/en/develop/manual_testing.html#accessibility-a11y-testing)
- [ ] External dependencies files were updated (`yarn` and `pip`)
- [ ] Documentation is updated
- [ ] Link to diff of internal dependency change is included
- [ ] CHANGELOG.rst is updated for high-level changes
- [ ] Contributor is in AUTHORS.rst

PR #3814, opened 2018-06-18
Update local versions to conform with PEP 440

benjaoming:
### Summary

Update to match [PEP 440](https://www.python.org/dev/peps/pep-0440/#local-version-identifiers)

> Local version identifiers MUST comply with the following scheme:
> 
>      <public version identifier>[+<local version label>]
> 
> (...)
> 
> To ensure local version identifiers can be readily incorporated as part of filenames and URLs, and to avoid formatting inconsistencies in hexadecimal hash representations, local version labels MUST be limited to the following set of permitted characters:
> 
> * ASCII letters ([a-zA-Z])
> * ASCII digits ([0-9])
> * periods (.)

Working on a Docker build to create Debian builds, I had to be able to predict the file names in `dist/` output, created from `make dist`.

The file names were automatically altered to match PEP 440 conventions, which on one hand is great. On the other, it meant that `kolibri --version` would be `1.2.3+git-1-abcdef123` and the file name would be ``kolibri-1.2.3+git.1.abcdef123.tar.gz`` (dots instead of hyphens).

So the reasons for this PR:

1) Changing our automated version generation for consistency with PEP 440
2) Avoiding some ugly/fuzzy regular expression or file pattern `kolibri-*.tar.gz` match when writing Docker automation that uses output in `dist/`.

### Reviewer guidance

Trying to make gradual, positive improvements. Let me know, if there is something we might as well include here!

You can check the file names of the Buildkite artifacts :)

### References

n/a - came up while working on https://github.com/learningequality/kolibri-installer-debian/issues/3

----

### Contributor Checklist

### Reviewer Checklist

### DIFF
--- a/kolibri/utils/version.py
+++ b/kolibri/utils/version.py
@@ -38,7 +38,7 @@ Confused? Here's a table:
 | Final        | Canonical, only     | N/A                 | N/A                       | 0.1.0, 0.2.2,                       |
 |              | information used    |                     |                           | 0.2.post1                           |
 +--------------+---------------------+---------------------+---------------------------+-------------------------------------+
-| dev release  | (1, 2, 3, 'alpha',  | Fallback            | timestamp of latest       | 0.4.0.dev020170605181124-f1234567   |
+| dev release  | (1, 2, 3, 'alpha',  | Fallback            | timestamp of latest       | 1.2.3.dev0+git.123.f1234567         |
 | (alpha0)     | 0), 0th alpha = a   |                     | commit + hash             |                                     |
 |              | dev release! Never  |                     |                           |                                     |
 |              | used as a canonical |                     |                           |                                     |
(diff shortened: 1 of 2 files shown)


kolibri/utils/version.py
```diff
@@ -38,40 +38,32 @@
...
-|              | your release        |                     |                           |                                     |
-|              | transforms into a   |                     |                           |                                     |
-|              | X.Y.0b0.dev{suffix} |                     |                           |                                     |
-|              | release, which in   |                     |                           |                                     |
-|              | most cases should   |                     |                           |                                     |
-|              | be assigned to the  |                     |                           |                                     |
-|              | preceding release   |                     |                           |                                     |
-|              | type.               |                     |                           |                                     |
+| beta0, rc0,  | DO NOT USE          | Fallback            | timestamp of latest       | 1.2.3b0.dev+git.123.f1234567        |
```
benjaoming, line 65: This was added for clarity. Let's not use these sort of weird and undefined version numbers ever :)

kolibri/utils/version.py
```diff
@@ -235,7 +227,10 @@ def get_version_from_git(get_git_describe_string):
     else:
         major, minor, patch = version_split
 
-    suffix = m.group('suffix')
+    # We need to replace "-" with ".". Namely, this is done automatically in
+    # the naming of source dist .whl and .tar.gz files produced by setup.py.
+    # See: https://www.python.org/dev/peps/pep-0440/#local-version-identifiers
+    suffix = m.group('suffix').replace("-", ".")
```
benjaoming, line 233: Here is the actual fix.

benjaoming: I looked at the code to analyze if `suffix` will always be a regex match group, and for now, I prefer that this fails loudly, if for some insane reason there is a match in `git describe` but it fails to detect data for a suffix. So not checking for `NoneType` is intended.

indirectlylit (approved):
Looks good to me.

Just a heads-up that I have a PR open trying to improve the documentation of our actual [release process](https://github.com/learningequality/kolibri/pull/3795/files#diff-d65b50da06e8fbcc94e7ac875d52ed66R3) and [development stages](https://github.com/learningequality/kolibri/pull/3795/files#diff-4afaee0ab4ee3c3d38cd5656fe9db157R1).

If you have any feedback let me know

Issue #3652, opened 2018-05-14
Simple login screen language dropdown causes layout issues


bonidjukic:
### Observed behavior
On the new simple login screen, the language dropdown behaves a bit strange, to say the least. I've tested the issue on Chrome and Firefox, running on Ubuntu 17.10.

#### Chrome
- version: 65.0.3325.181 (Official Build) (64-bit), 66.0.3359.170 (Official Build) (64-bit)
- description: It makes the entire language footer jump up.
- screencast gif: 

![kolibribeta-peek-chrome](https://user-images.githubusercontent.com/2150991/39999790-4f91adaa-578a-11e8-923c-e823ef117d5c.gif)

#### Firefox
- version: 59.0.2 (64-bit)
- description: 
  - if I click on the "More languages" text, dropdown jumps down and browser gets a vertical scroll
  - if I click on the caret icon, dropdown jumps down, but there is no vertical scroll so it's actually impossible to select anything
- screencast gif: 

![kolibribeta-peek-ff](https://user-images.githubusercontent.com/2150991/39999931-bc055c34-578a-11e8-8200-a1ed3345c46e.gif)

### Expected behavior
I would maybe expect the dropdown to jump up, I believe that would resolved the described issue.

### User-facing consequences
Mostly inconvinience with regards to the content jumping and vertical scroll, but in some cases also the inability to select a language from the dropdown (if the caret icon is clicked on Firefox).

### Errors and logs
n/a

### Steps to reproduce
Visit http://kolibridemo.learningequality.org.

### Context
 * Kolibri version: version which is deployed on http://kolibridemo.learningequality.org
 * Operating system: Ubuntu 17.10
 * Browser: 
    * Chrome: 65.0.3325.181 (Official Build) (64-bit), 66.0.3359.170 (Official Build) (64-bit)
    * Firefox: 59.0.2 (64-bit)

indirectlylit:
@christianmemije  mind taking a look at this?

indirectlylit:
Beyond the obvious layout issues, the typography in this bottom section is all over the place:

![image](https://user-images.githubusercontent.com/2367265/41688113-96e22f18-749f-11e8-8171-fe51400ea338.png)

![image](https://user-images.githubusercontent.com/2367265/41688112-92c28ef0-749f-11e8-9f40-b237069cbeb7.png)

Multiple weights, colors, baselines, and underline styles.

indirectlylit:
[...]

PR #3829, opened 2018-06-20
updated language switcher

indirectlylit:
### Summary

As reported in #3652, the language switcher is kind of a mess.

This PR simplifies the UI drastically and makes it consistent with the other language switcher in the main app:

* use a globe icon to indicate the current language
* use a modal - more effective at handling a large number of languages than the select box

| Before | After |
|--|--|
| image | image |

| select 1 | select 2 | same modal for both |
|--|--|--|
| image | image | image |

The main downside I can think of is that there's no longer a few other visible languages. However, I'd argue that this is a reasonable thing to give up...

### Reviewer guidance

Please review both the design and implementation

### References

#3652

----

### Contributor Checklist

### Reviewer Checklist

### DIFF
--- a/kolibri/core/assets/src/core-app/apiSpec.js
+++ b/kolibri/core/assets/src/core-app/apiSpec.js
@@ -42,7 +42,6 @@ import kTextbox from '../views/k-textbox';
 import kNavbar from '../views/k-navbar';
 import kNavbarLink from '../views/k-navbar/link';
 import logo from '../views/logo';
-import languageSwitcherMixin from '../views/language-switcher/mixin.js';
 import languageSwitcherList from '../views/language-switcher/list.vue';
 import immersiveFullScreen from '../views/immersive-full-screen';
 import elapsedTime from '../views/elapsed-time';
@@ -162,7 +161,6 @@ export default {
     mixins: {
       responsiveWindow,
       responsiveElement,
-      languageSwitcherMixin,
       contentRenderer: contentRendererMixin,
       fullscreen,
     },
--- a/kolibri/core/assets/src/views/language-switcher/list.vue
+++ b/kolibri/core/assets/src/views/language-switcher/list.vue
@@ -83,7 +83,8 @@
           .filter(lang => availableLanguages[lang] !== undefined)
           .filter(lang => lang !== currentLanguage)
           .map(lang => availableLanguages[lang])
-          .slice(0, this.numVisibleLanguages);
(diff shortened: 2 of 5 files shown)


jtamiace:
I think the enhancement to open into a modal is great and so is the globe icon and current language but would like to still advocate for displaying a few languages in different scripts to help especially less digitally literate users. 

The UI components here (footer vs dropdown) are completely different so I would argue that consistency is not a good reason to take them out. There's more screen real-estate afforded in the footer than a dropdown, so why not use it? 

screen shot 2018-06-20 at 5 18 06 pm

I just deleted components with inspect element, but I wouldn't expect someone who's entirely new to computers to know where to change the language here, even with the globe icon.

indirectlylit:
okay, I'll look at a variation with a few languages displayed

khangmach:
- I like the switch to a "Change language" button and also the usage of the modal. 
- The globe icon definitely helps this section stand out
- RE: showing more languages: Accommodating those who aren't computer literate is important...although I think the chances are still  low for a user to chance upon their native language even if we display 2+ languages outside of the modal
- I think we can do without the "Selected language" label text, esp if we include the globe icon

jtamiace:
> the chances are still low for a user to chance upon their native language even if we display 2+ languages outside of the modal

The point is more so to use the display of different scripts to signal that there is likely a mechanism to change language nearby. 

I guess another argument against displaying scripts is that a less digitally literate user still may not understand that signal, and they'll be in a social environment anyway where they can ask someone why Kolibri is in a language they don't understand.

indirectlylit:
Seems like the idea is that by showing foreign-looking characters/words juxtaposed, we can evoke the idea of "multiple languages". That seems to be strategy with icons like these:

image image image image

If we show the names of a few languages, we're kind of using them as abstract icons...

_UNLESS_ we had a way of knowing the "most likely" languages. If we knew that, then it actually becomes really useful to show them up front as quick-links.

indirectlylit:
Update:

Now shows a few language links randomly chosen from our most common languages:

* arabic
* english
* spanish
* french
* hindi
* swahili

The number of languages shown is variable based on the browser width.

image

indirectlylit:
also fixes https://github.com/learningequality/kolibri/issues/3851

khangmach (approved):

jtamiace (approved):

christianmemije (approved):

PR #3849, opened 2018-06-21
Handle single-card carousel.

DXCanas:
### Summary

Removing ability to shift pages if none exists.

There is logic to handle a click that goes past the range of the current array... but I figured this would be simpler. 

Note to @christianmemije: This actually doesn't have to do with the `:key` value? Tried setting it to a constant value, and all cards still loaded initially. I think `:key` is only checked when Vue is evaluating whether or not it needs to replace an element? Not sure where these would be getting deduped, though? 

### Reviewer guidance

…

### References

…

----

### Contributor Checklist

### Reviewer Checklist


indirectlylit (approved):
thanks!

PR #3854, opened 2018-06-22
Optimize contentnode endpoint for faster browsing

bonidukic:
### Summary
Most of the work in this PR is focused on improving the `contentnode` endpoint by splitting it into two endpoints, creating a new, lighter option, `contentnode_slim`, to be used for content browsing.

Additional optimizations have also been added with a goal to increase the content browsing speed, such as:
- ORM prefetching to avoid additional db queries
- caching results to avoid additional db queries
- reducing the number of queries being done to get the number of nodes flagged as `coach_content`
- moved `ContentNode` progress annotation to the frontend (lazy-loaded)

### Graphs

slim_total_requests

slim_requests_per_s

### Reviewer guidance
- A couple of tests have been added to test the `contentnode_slim` endpoint, but I'm open to suggestions on how to increase test coverage if necessary
- Check if _Learn_ and _Recommended_ pages are working properly
- Keep in mind that this PR should not remove any of the existing functionalities, but strictly optimize

----

### Contributor Checklist

### Reviewer Checklist

### DIFF
--- a/kolibri/core/content/api.py
+++ b/kolibri/core/content/api.py
@@ -480,9 +480,7 @@ class ContentNodeSlimViewset(viewsets.ReadOnlyModelViewSet):
     pagination_class = OptionalPageNumberPagination
 
     def prefetch_related(self, queryset):
-        return queryset.prefetch_related(
-            'files',
-        ).select_related('lang')
+        return queryset.prefetch_related('files')
 
     def get_queryset(self, prefetch=True):
         queryset = models.ContentNode.objects.filter(available=True)
--- a/kolibri/core/content/api.py
+++ b/kolibri/core/content/api.py
@@ -473,6 +473,64 @@ class ContentNodeViewset(viewsets.ReadOnlyModelViewSet):
         return Response(serializer.data)
 
 
+class ContentNodeSlimViewset(viewsets.ReadOnlyModelViewSet):
+    serializer_class = serializers.ContentNodeSlimSerializer
+    filter_backends = (DjangoFilterBackend,)
+    filter_class = ContentNodeFilter
+    pagination_class = OptionalPageNumberPagination
+
+    def prefetch_related(self, queryset):
+        return queryset.prefetch_related(
+            'files',
+        ).select_related('lang')
+
+    def get_queryset(self, prefetch=True):
+        queryset = models.ContentNode.objects.filter(available=True)
+        if prefetch:
+            return self.prefetch_related(queryset)
+        return queryset
+
+    def get_object(self, prefetch=True):
+        """
+        Returns the object the view is displaying.
+        You may want to override this if you need to provide non-standard
+        queryset lookups.  Eg if objects are referenced using multiple
+        keyword arguments in the url conf.
+        """
+        queryset = self.filter_queryset(self.get_queryset(prefetch=prefetch))
+
+        # Perform the lookup filtering.
+        lookup_url_kwarg = self.lookup_url_kwarg or self.lookup_field
+
+        assert lookup_url_kwarg in self.kwargs, (
+            'Expected view %s to be called with a URL keyword argument '
+            'named "%s". Fix your URL conf, or set the `.lookup_field` '
+            'attribute on the view correctly.' %
+            (self.__class__.__name__, lookup_url_kwarg)
+        )
+
+        filter_kwargs = {self.lookup_field: self.kwargs[lookup_url_kwarg]}
+        obj = get_object_or_404(queryset, **filter_kwargs)
+
+        # May raise a permission denied
+        self.check_object_permissions(self.request, obj)
+
+        return obj
+
+    @detail_route(methods=['get'])
+    def ancestors(self, request, **kwargs):
+        cache_key = 'contentnode_ancestors_{pk}'.format(pk=kwargs.get('pk'))
+
+        if cache.get(cache_key) is not None:
+            return Response(cache.get(cache_key))
+
+        ancestors = list(self.get_object(prefetch=False).get_ancestors().values('pk', 'title'))
+
+        cache.set(cache_key, ancestors, 60 * 10)
+
+        return Response(ancestors)
+
+
 class ContentNodeGranularViewset(mixins.RetrieveModelMixin, viewsets.GenericViewSet):
     serializer_class = serializers.ContentNodeGranularSerializer
 
--- a/kolibri/core/content/api.py
+++ b/kolibri/core/content/api.py
@@ -71,7 +71,9 @@ class ChannelMetadataViewSet(viewsets.ReadOnlyModelViewSet):
     filter_class = ChannelMetadataFilter
 
     def get_queryset(self):
-        return models.ChannelMetadata.objects.all().order_by('-last_updated')
+        return models.ChannelMetadata.objects.all() \
+            .order_by('-last_updated') \
+            .prefetch_related('root__lang')
 
 
 class IdFilter(FilterSet):
--- a/kolibri/core/content/api.py
+++ b/kolibri/core/content/api.py
@@ -494,7 +494,7 @@ class ContentNodeSlimViewset(viewsets.ReadOnlyModelViewSet):
         """
         Returns the object the view is displaying.
         You may want to override this if you need to provide non-standard
-        queryset lookups.  Eg if objects are referenced using multiple
+        queryset lookups. Eg if objects are referenced using multiple
         keyword arguments in the url conf.
         """
         queryset = self.filter_queryset(self.get_queryset(prefetch=prefetch))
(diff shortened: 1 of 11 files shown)


kolibri/core/content/management/commands/generate_schema.py
```diff
@@ -31,6 +32,10 @@ def add_arguments(self, parser):
         parser.add_argument('version', type=str)
 
     def handle(self, *args, **options):
+        # Restrict to Python 3 due to issues with `json.dump` on Python 2
```
indirectlylit: what sort of issues?

bonidukic:
```
(kolibri)  boni@t470s  ~/projects/learning-equality/kolibri  (feature/optimize-contentnode-endpoint) >  kolibri manage generate_schema 3
/home/boni/.venvs/kolibri/local/lib/python2.7/site-packages/psycopg2/__init__.py:144: UserWarning: The psycopg2 wheel package will be renamed from release 2.8; in order to keep installing from binary please use "pip install psycopg2-binary" instead. For details see: <http://initd.org/psycopg/docs/install.html#binary-install-from-pypi>.
  """)
Installed 25 object(s) from 1 fixture(s)
  File "/home/boni/.venvs/kolibri/bin/kolibri", line 9, in <module>
  File "/home/boni/projects/learning-equality/kolibri/kolibri/utils/cli.py", line 631, in main
  File "/home/boni/projects/learning-equality/kolibri/kolibri/utils/cli.py", line 421, in manage
  File "/home/boni/projects/learning-equality/kolibri/kolibri/dist/django/core/management/__init__.py", line 364, in execute_from_command_line
    utility.execute()
  File "/home/boni/projects/learning-equality/kolibri/kolibri/dist/django/core/management/__init__.py", line 356, in execute
  File "/home/boni/projects/learning-equality/kolibri/kolibri/dist/django/core/management/base.py", line 283, in run_from_argv
  File "/home/boni/projects/learning-equality/kolibri/kolibri/dist/django/core/management/base.py", line 330, in execute
  File "/home/boni/projects/learning-equality/kolibri/kolibri/core/content/management/commands/generate_schema.py", line 65, in handle
    json.dump(data, f)
  File "/usr/lib/python2.7/json/__init__.py", line 190, in dump
    fp.write(chunk)
TypeError: write() argument 1 must be unicode, not str
```

bonidukic: @indirectlylit FYI I discussed this with @rtibbles and we agreed that since this is internal command that it might be easiest to require Python 3 - but if you feel we should support Python 2 as well here I can play with this to figure out something that works?

bonidukic: I found this SO answer helpful earlier: https://stackoverflow.com/a/36008538/6817428

indirectlylit: Main issue is that currently builds _only_ work in py2, so that leaves us with an inconsistent dev experience: https://github.com/learningequality/kolibri/blob/develop/Makefile#L122

@benjaoming do you recall why py2 is necessary for building?

benjaoming: Yes, we only build on Python 2. That's the only way currently to have a Py2+3 build.

So we have to have everything py2+3, there is no nice way to handle exceptions to the rule IMO.

bonidukic: I mean, it's not like I changed any previous behaviour here, this wasn't working on Python 2 anyway, I just added a message to the user to avoid them having to figure out what's going on.

benjaoming: Oh right okay, then let's just say that someone at some point in time will have to find out that this `generate_schema` doesn't work on python 2 and fix that if it's a problem :)

bonidukic: Yep, I'll try to take a look at it within this PR, AFAIK there is a difference between using `json.dumps()` and `json.dump()` with respect to the `ensure_ascii` (among other things) so there might be an easy solution.

kolibri/core/content/api.py
```diff
@@ -473,6 +475,62 @@ def all_content(self, request, **kwargs):
...
+        cache_key = 'contentnode_slim_ancestors_{pk}'.format(pk=kwargs.get('pk'))
+
+        if cache.get(cache_key) is not None:
+            return Response(cache.get(cache_key))
+
+        ancestors = list(self.get_object(prefetch=False).get_ancestors().values('pk', 'title'))
+
+        cache.set(cache_key, ancestors, 60 * 10)
+
+        return Response(ancestors)
```
rtibbles, line 531: There are probably other list and detail endpoints from the ContentNode viewset that can be ported to ContentNodeSlim - but let's get this fixed and merged and then iterate!

rtibbles:
Test errors are one Python linting error, and one testing error that probably only appeared on Postgresql due to ordering issues (as we were only making an assertion about the first item in the response).

kolibri/core/content/management/commands/generate_schema.py
bonidukic: FYI: I removed this commit from this PR and created a separate PR with a compatibility fix: https://github.com/learningequality/kolibri/pull/4020

kolibri/core/content/api.py
```diff
@@ -473,6 +475,62 @@ def all_content(self, request, **kwargs):
...
+            'named "%s". Fix your URL conf, or set the `.lookup_field` '
+            'attribute on the view correctly.' %
+            (self.__class__.__name__, lookup_url_kwarg)
+        )
+
+        filter_kwargs = {self.lookup_field: self.kwargs[lookup_url_kwarg]}
+        obj = get_object_or_404(queryset, **filter_kwargs)
+
+        # May raise a permission denied
+        self.check_object_permissions(self.request, obj)
```
rtibbles, line 516: We can not bother with permission checks on ContentNode, as they have no permissions on them.

kolibri/plugins/learn/assets/src/state/mutations/index.js
```diff
@@ -35,12 +35,22 @@ export default {
...
-      if (topic) {
-        topic.progress = progress.progress_fraction;
+      const contentNode = state.pageState.contents.find(node => node.id === progress.pk);
+      if (contentNode) {
+        contentNode.progress = progress.progress_fraction;
       }
     });
   },
+  SET_RECOMMENDED_NODES_PROGRESS(state, progressArray) {
+    Object.keys(state.pageState).forEach(function(key) {
```
rtibbles: Might want to make this explicit and use `['recommended', 'popular', 'nextSteps']` instead of reading the object keys of the `pageState`.

rtibbles (changes requested):
One small code tweak. Will be manually testing also.

rtibbles:
Manual testing looks good to me.

kolibri/plugins/learn/assets/src/state/mutations/index.js
bonidukic: I was thinking about that but wasn't sure about the general verbosity philosophy so I went with this, but will gladly update as per your suggestion.

kolibri/core/content/api.py
bonidukic: I'm guessing that in that case we want to remove the `check_object_permissions` check on both `ContentNodeViewset` and `ContentNodeSlimViewset`?

kolibri/plugins/learn/assets/src/state/mutations/index.js
rtibbles: My main reason for explicitness is just in case there are other keys on the pageState that aren't one of these three.

kolibri/core/content/api.py
rtibbles: Yeah, we could probably have a 'non-permissioned' Viewset base class.

rtibbles: Might be other things we could exclude too.

kolibri/plugins/learn/assets/src/state/mutations/index.js
bonidukic: Yes, I understand completely, my thinking was that if another key was to be added to `pageState` it would be another set of _Recommended_ nodes which we'd want to annotate anyway - but I agree that listing them explicitly is definitely safer and someone will have to add it to the mutation manually if they see that it's not working rather than causing regression if the key added to `pageState` wasn't a list of nodes.

rtibbles (approved):
Updates made, manual testing works!

kolibri/core/content/api.py
rtibbles: Can manage this in future cleanup.

Issue #3861, opened 2018-06-22
Deduplication of 'Resume' and 'Most popular' happens on client side


indirectlylit:
### Observed behavior

these sections may not have as many items as they ought to:

![image](https://user-images.githubusercontent.com/2367265/41802395-4608b4ee-7635-11e8-928a-d85784f5c8d2.png)

### Expected behavior

de-duping should occur server-side so more items are shown

### User-facing consequences

less useful landing page

### Steps to reproduce

…

### Context

0.10.0

Issue #3847, opened 2018-06-21
If a topic appears in "Next Steps," it takes the user to a content renderer page 


jtamiace:
<img width="1072" alt="screen shot 2018-06-21 at 3 24 05 pm" src="https://user-images.githubusercontent.com/6668144/41748545-3960b922-7567-11e8-94d9-7f7ded283015.png">
<img width="1083" alt="screen shot 2018-06-21 at 3 24 13 pm" src="https://user-images.githubusercontent.com/6668144/41748546-397feb1c-7567-11e8-814c-1296d1011d2a.png">

### Expected behavior

User should be taken to the topic and not experience infinite load state

### User-facing consequences

User will wait forever

### Steps to reproduce

1. Log in as a facility user
2. Complete/master the last item in a topic
3. Go to "Recommended"
4. Click the topic card that appears in "Next Steps"

### Context

0.10
MacOS
Chrome

DXCanas:
@jtamiace Looking over this, one of the reasons it routes to a content page _only_ is because there's a dedicated "recommended" content page with key differentiators being the tab and breadcrumbs highlighted
![image](https://user-images.githubusercontent.com/9877852/41867101-ff5d8ac2-7866-11e8-80d3-c48cc53c3b1b.png)
Should we do anything special for "recommended" topics?

jtamiace:
Hmm... it goes against the original design to keep content that appears in Recommended scoped to Recommended only, but feel like it would make sense to jump the user to that topic in Channels. Would there be huge issues to make that link across tabs?

DXCanas:
>  it goes against the original design to keep content that appears in Recommended scoped to Recommended only

Huh, I didn't know that was the case. I we can probably make it so that they all go their place in the topic tree? I

jtamiace:
> we can probably make it so that they all go their place in the topic tree?

Yeah at this point, that would make sense to me.

PR #3884, opened 2018-06-24
Dedupe content

ralphiee22:
### Summary

Remove duplicates from popular and resume api responses. 
Also only filter against content nodes that currently exist in the database.

### Reviewer guidance

Changes are on http://kolibri-beta.learningequality.org/learn/#/recommended.
sign in as admin to see resume response.
Pay attention to load time?

### References

#3861 

----

### Contributor Checklist

### Reviewer Checklist

### DIFF
--- a/kolibri/content/api.py
+++ b/kolibri/content/api.py
@@ -224,11 +224,21 @@ class ContentNodeFilter(IdFilter):
             .annotate(Count('content_id')) \
             .order_by('-content_id__count')
 
-        most_popular = queryset.filter(content_id__in=list(content_counts_sorted[:10]))
+        most_popular = queryset.filter(content_id__in=list(content_counts_sorted[:20]))
+
+        # remove duplicate content items
+        deduped_list = []
+        content_ids = set()
+        for node in most_popular:
+            if node.content_id not in content_ids:
+                deduped_list.append(node)
+                content_ids.add(node.content_id)
+
+        queryset = most_popular.filter(id__in=[node.id for node in deduped_list])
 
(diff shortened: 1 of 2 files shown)


DXCanas:
There are some things happening with recommended page that are making me wonder if this is the best approach. @jtamiace 

Because of #3847, we're considering removing the "recommended content" page and treating the items that pop up (and their duplicates) in the carousel like we do in search results:
search results page

Not entirely sure what's going on here, but would this remove items duplicate content items in different topic trees?

ralphiee22:
That is a correct assumption and actually quite problematic. For recommended we basically pick a random content node which could be in any channel and return that to the user. 
For resume, its more of an issue I am worried about because we could return content nodes for a channel which they haven't even explored.

kolibri/content/api.py
```diff
@@ -217,22 +217,34 @@ def filter_popular(self, queryset, name, value):
...
+
+        # remove duplicate content items
+        deduped_list = []
+        content_ids = set()
+        for node in most_popular:
+            if node.content_id not in content_ids:
+                deduped_list.append(node)
+                content_ids.add(node.content_id)
+
+        queryset = most_popular.filter(id__in=[node.id for node in deduped_list])
```
indirectlylit, line 237: is it necessary to do a second query? can you not just return the `deduped_list`?

ralphiee22: the result of this query is returned to another filter method (ex. `filter_by_role`). And the logic in there expects another queryset. If we return a list then we can't do methods on it like `.exclude`.

indirectlylit (approved):

PR #3886, opened 2018-06-25
Reduce processing time in Loggers viewsets

jredrejo:
### Summary
Using a profiler I have checked loggers save function (called from the Frontend using `sessionModel.save( ...` ) spends its time in the backend along 4 main sections:

antes

These sections have been studied and modified accordingly in this PR:
1. generic.py : get_object() can be replaced by a simple `model.objects.get` using Django ORM. Adding checks for the appropiate permissions
2. serializers.py : data() is used to 'deserializate' the same object that has been received, thus the object is serialized + deserializated. I have replaced it by just returning the same object that is being received plus the fields that are actually static methods. These fields are included in the Serializers in a new variable called `update_fields`
According to what I have been able to check, this could cause problems only in the case of models that are not in sync, but I can't think of any possible case for the Loggers.
3. Validation: DRF validates each field twice (looking at the code it seems it's due to some legacy code). Most of the validation code has been kept, while removing unneeded steps.
4. Perform_update part has not been modified as it seems optimal using current DRF/Django ORM code. 

Also, point 3 allow us now to remove #2442 : Validation now  happens only on the changed fields, and saving in the database will use only validated data, thus we improve a bit point 4 as well.

So the PR creates a new class to be inherited by the Loggers ViewSets. It also adds a couple of methods to KolibriModelSerializer to replace part of the DRF validation. These changes are making that, according to the profile, logging time is reduced 36% aproximately:

despues

### Reviewer guidance
I'd like to know if I have missed some possible origin of problems with this PR

----

### Contributor Checklist

### Reviewer Checklist


indirectlylit:
just a minor note that you have some linting errors. Recommend setting up [pre-commit](http://kolibri-dev.readthedocs.io/en/develop/start/getting_started.html#pre-commit)

kolibri/core/logger/api.py
```diff
@@ -55,6 +58,35 @@ def filter_learner_group(self, queryset, name, value):
             target_user=F("user"),
         )
 
+class LoggerViewSet(viewsets.ModelViewSet):
+    def update(self, request, *args, **kwargs):
+        partial = kwargs.pop('partial', False)
+        model = self.queryset.model
+        lookup_url_kwarg = self.lookup_url_kwarg or self.lookup_field
+        try:
+            instance = model.objects.get(id=self.kwargs[lookup_url_kwarg])
```
jamalex, line 68: One of the really important things that `get_object` does is check the appropriate permissions for the current user against the object being updated:

https://github.com/encode/django-rest-framework/blob/master/rest_framework/generics.py#L101

I'm guessing that's where a good portion of the savings may be coming from?

If the permission check is slowing things down, we can do some optimizations there. We have a very general-purpose permission system that allows for complex relationships between users and objects to be defined. See:
https://kolibri-dev.readthedocs.io/en/develop/architecture/uap/concepts_and_definitions.html
and
https://kolibri-dev.readthedocs.io/en/develop/architecture/uap/implementation.html

But we can look into "short-circuiting" it in this common case (if it isn't already) where the log is directly associated with the current user.

jredrejo: Hi @jamalex, good catch!. I hadn't seen the permissions checks. I have added it to this PR.
Obviously, this reduces the performance improvement (this check is making one sql query, I think it could be cached instead of looking to the db for the same permission in every request, but that's another story).
Anyway, you can see in the chart below that that's not where most of the time is spent, the `get_object_or_404` function takes longer because it makes a query for each of the fields of the object, instead of fetching the whole object in one query.
has_permissions

indirectlylit (approved):
did some basic testing and things seem to be working!  Merging in so we can test it more broadly

PR #3893, opened 2018-06-26
Add "deprovision" management command to clear out user data from the DB

jamalex:
### Summary

Needed for when an implementing or hardware partner (Nalanda, World Possible, etc) provisions a device or device image with content, but then wants to clear out the temporary facility and user data and allow the end user to provision the device from scratch (but without needing to re-import content).

### Reviewer guidance

To test, make sure to back up your database (if it has anything you care about). Then, run `kolibri manage deprovision` and answer the prompts. Then start the server back up and check that you're taken to the onboarding wizard. After you complete the wizard, the content page should still show whatever content you had in there before.

----

### Contributor Checklist

### Reviewer Checklist

- [x] Automated test coverage is satisfactory

### DIFF
--- a/kolibri/auth/management/commands/deprovision.py
+++ b/kolibri/auth/management/commands/deprovision.py
@@ -1,0 +1,73 @@
+import logging as logger
+import sys
+
+from django.db.models.signals import post_delete
+from django.utils.six.moves import input
+from morango.certificates import Certificate
+from morango.models import Buffer
+from morango.models import DatabaseIDModel
+from morango.models import DeletedModels
+from morango.models import Store
+
+from kolibri.auth.models import FacilityDataset
+from kolibri.auth.models import FacilityUser
+from kolibri.core.device.models import DevicePermissions
+from kolibri.core.device.models import DeviceSettings
+from kolibri.logger.models import AttemptLog
+from kolibri.logger.models import ContentSessionLog
+from kolibri.logger.models import ContentSummaryLog
+from kolibri.tasks.management.commands.base import AsyncCommand
+from kolibri.utils.cli import server
+
+logging = logger.getLogger(__name__)
+
+MODELS_TO_DELETE = [
+    AttemptLog, ContentSessionLog, ContentSummaryLog, FacilityUser, FacilityDataset, Certificate,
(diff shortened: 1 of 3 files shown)


kolibri/auth/management/commands/deprovision.py
```diff
@@ -0,0 +1,73 @@
...
+
+        # safest not to run this command while the server is running
+        status_code, _ = server.get_urls()
+        if status_code == server.STATUS_RUNNING:
+            logging.error("The Kolibri server is currently running. Please stop it and then re-run this command.")
+            sys.exit(1)
+
+        # ensure the user REALLY wants to do this!
+        confirm_or_exit("Are you sure you wish to deprovision your database? This will DELETE ALL USER DATA!")
+        confirm_or_exit("ARE YOU SURE? If you do this, there is no way to recover the user data on this device.")
```
ralphiee22, line 69: maybe add one more so the user can be really sure? (kidding)

ralphiee22 (approved):
Manually tested.

ralphiee22:
is this something we want to include in the documentation? or is it more we just pass this info along to our partners?

indirectlylit (approved):
server running check doesn't work with devserver. I filed a follow-up issue here: #3902

> is this something we want to include in the documentation?

I think we probably do want to include it, right @jamalex ? (cc @radinamatic)

radinamatic:
If I understand correctly, this command is needed only for higher-level implementing/hardware partners who make/clone multiple devices for their deployments. As such, I'm not sure if the correct place for documenting this command is the general end-user documentation.

We should rather start another document for this type of partners and collect all the special-needs commands developed for them.

indirectlylit:
sounds fine to me!

fisherbryan:
Hi @radinamatic 
A technical document would be perfect that would assist us in remote support to collect data from the command prompt. Please consider this greatly! We are using it as the main part of business and to clone  multiple local servers, but we are unable to have a sufficient way of changing facility and customising a centre using it. 

We have just over 50 centre with the same Facility name. We would like to be able to rename a facility with out having to re-download the same content . And also move users + their data over to a new device should the current hardware device fail for any reason. Some technical documentation will chnge our lives :)

PR #3915, opened 2018-06-28
block double clicks

### LINKED ISSUE #3914: double-clicking or -tapping on content cards causes unexpected navigation
### Observed behavior

We've observed learners who believe that they need to double-click on content cards. When the pages are already cached, this causes the learner to navigate _two_ levels deep through the topic tree because the page loads so fast.

### Expected behavior

We should try to prevent clicks to the page that are made within the first 750ms or so, since these are most likely to be inadvertent.

### User-facing consequences

unexpected navigation

### Steps to reproduce

* navigate down the topic tree and then back up so that the inner topics get cached
* double-click on a top-level topic

you'll jump two levels deep in the tree

### Context

0.10.0

indirectlylit:
### Summary

In Kakuma it was observed that some learners thought they needed to double-tap on content cards in order to navigate. This caused them to jump two levels deep in the tree.

This change adds a blocking layer for half a second after navigating to a new page, which has the effect of absorbing the second click in a double-click.

Quoting a comment from the code:

It's something of a hack. A better strategy would be to create a new `setLoading` action which handles both `state.core.loading` and `state.core.blockDoubleClicks` with a single function.

### Reviewer guidance

does this make sense?

### References

fixes #3914 which was observed by @jamalex 

----

### Contributor Checklist

### Reviewer Checklist

### DIFF
--- a/kolibri/core/assets/src/kolibri_app.js
+++ b/kolibri/core/assets/src/kolibri_app.js
@@ -64,7 +64,7 @@ export default class KolibriApp extends KolibriModule {
       mutations: this.mutations,
     });
     return getCurrentSession(store).then(() => {
-      Promise.all([
+      return Promise.all([
         // Invoke each of the state setters before initializing the app.
         ...this.stateSetters.map(setter => setter(this.store)),
       ]).then(() => {
--- a/kolibri/core/assets/src/state/actions.js
+++ b/kolibri/core/assets/src/state/actions.js
@@ -174,6 +174,22 @@ function handleApiError(store, errorObject) {
 }
 
 /**
+ * Used to prevent inadvertent actions if a user double-clicks to navigate
+ *
+ * Something of a hack. A better strategy would be to create a new
+ * `setLoading` action which handles both `state.core.loading` and
+ * `state.core.blockDoubleClicks` with a single function.
+ */
+function blockDoubleClicks(store) {
+  if (!store.state.core.blockDoubleClicks) {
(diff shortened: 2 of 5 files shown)


jonboiser:
If this situation is unique to the Learn Page Resource Cards layout, couldn't we just add a UI blocker to that page, instead of to every page?

jonboiser:
Since all of the routes affected are part of the Learn plugin, it might be possible to add a `afterEach` hook to the Learn plugin router and make that hook the sole caller of `blockDoubleClicks`.

indirectlylit:
so you guys don't like the idea I posted above?

>  A better strategy would be to create a new `setLoading` action which handles both `state.core.loading` and `state.core.blockDoubleClicks` with a single function.

(The only reason I didn't do that was to avoid potentially risky changes in 0.10.0)

indirectlylit:
I think the card layouts were the ones most-likely to be affected, but I don't think there's harm in applying it globally. Unless you can think of an issue there?

jonboiser:
I think the original double-click issue might be hyper specific to the card layout, since it may be the only page where a card will be replaced with another card.

So the proportionate solution, I think, would be to delay rendering or apply the mask to those replacement cards. This could even be done in a super-local way in the Carousel or Card-Grid components.

indirectlylit:
Wouldn't anything that coincidentally happened to appear in the same location as the double-click on the next page be triggered?

indirectlylit:
@christianmemije @jonboiser updated to use `afterEach`

jonboiser (approved):

PR #3942, opened 2018-06-29
Update to vuex 3

jonboiser:
### Summary

Updates vuex from v1 to v3 and updates vuex usage.

Main things:

1. Organize all plugins' vuex assets into a single module `pluginModule.js` that is passed into its `KolibriApp` instance.
1. Organize the `core` state into modules for session, logging, etc.
1. Changes kolibri_app to accept a single object `pluginModule` that is then inserted into the `kolibri.coreVue.store` singleton.
1. Move `samePageCheckGenerator` to a `util` module since it isn't used as a true action in the app.
1. Everywhere in app, switch to new `store` API, including the use of `store.getters` , `store.dispatch`, etc.
1. Everywhere in app, switch to new component vuex binding API, using `mapState`, `mapActions`, `mapGetters`, `mapMutations`.
1. Fix components that initialize their `data()` from vuex (must query the store directly instead of using a vuex-to-computed binding).

Notes:

1. In Coach page, did not inject report-related getters into the store, since they were not type-safe in most cases.
1. In Device Management page, _did_ inject content wizard-related getters, but wrapped them in a way that they were type-safe when outside of the wizard.

#### Formal checks

1. Check that all calls of `store.dispatch` match `store.dispatch('myAction', payload)` and that the type of `payload` matches the `payload` of `myAction`
1. Check that all calls of `store.commit` match `store.commit('MY_MUTATION', payload)` and that the type of `payload` matches the `payload` of `MY_MUTATION`
1. Check that, in all components, that all vuex bindings used in the `<script>` or `<template>` are present. That is, they appear in a `mapGetters/Action/Mutations` wrapper, and that the getters/actions/mutations are actually register in the store.
1. Check that, in all components, all bound actions and mutations are called correctly (similar to check in (1) (2)).

#### Manual checks

If all formal checks are okay, the app should work the same, but test everything, especially pages that depend on a polymorphic `pageState` object.

### References

…

----

### Contributor Checklist

### Reviewer Checklist


package.json
```diff
@@ -56,7 +56,7 @@
     "vue-intl": "2.1.1",
     "vue-markdown": "^2.1.3",
     "vue-router": "^2.1.1",
-    "vuex": "^1.0.0"
+    "vuex": "3.0.1"
```
indirectlylit, line 59: recommend keeping the `^`

kolibri/core/assets/src/heartbeat.js
```diff
@@ -75,12 +74,12 @@ export class HeartBeat {
    * @return {Promise} promise that resolves when the endpoint check is complete.
    */
   checkSession() {
+    const { currentUserId: userId, connected, reconnectTime } = store.getters;
```
indirectlylit: > `currentUserId: userId`

Minor comment: this is helpful to minimize extra changes and changes of regressions in this PR. However in general, I think remapping variable names like this makes makes object destructuring harder to understand.

kolibri/core/assets/src/state/store.js
```diff
@@ -1,231 +1,33 @@
...
+    store.replaceState({
+      core: { ...coreModule.state },
+      ...(pluginModule.state || {}),
+    });
+    forEach(pluginModule.modules, (module, name) => {
+      store.registerModule(name, module);
+    });
+  }
+  return store;
+}
```
indirectlylit, line 39: Would appreciate some comments in here. It's a fairly meta function so it's non-obvious what the intended behavior and usage is.

kolibri/plugins/coach/assets/src/state/mutations/lessonsMutations.js
```diff
@@ -52,8 +50,8 @@ export function SET_LESSON_REPORT(state, report) {
 
 export function ADD_TO_RESOURCE_CACHE(state, node) {
   if (node && node.pk) {
-    const channelObject = getChannelObject(state, node.channel_id);
-
+    // TODO remove dependency on channel title
```
indirectlylit, line 53: mind briefly elaborating?

kolibri/plugins/coach/assets/src/views/lessons/LessonResourceSelectionPage/searchTools/index.vue
```diff
@@ -26,14 +17,13 @@
 
 <script>
 
-  // import searchBox from '../../../../../../../learn/assets/src/views/search-box/';
```
indirectlylit, line 29: thank you!

kolibri/plugins/coach/assets/src/views/reports/summary-section.vue
```diff
@@ -1,166 +0,0 @@
...
-  };
-
-</script>
-
-
-<style lang="stylus" scoped>
-
-  @require '~kolibri.styles.definitions'
-
-</style>
```
indirectlylit, line 166: 👋

indirectlylit (approved):
A few minor comments. Things work for me after a brief run-through.

Let's get this merged!

kolibri/plugins/coach/assets/src/state/mutations/lessonsMutations.js
jonboiser: This code could probably be refactored to not need `channelTitle` (or have it passed in with the payload) and remove this getter-like code from the mutation. Vuex 3 seems to discourage using getters within mutations (since it does not provide them to the functions as it does for getters and actions).

indirectlylit: >  Vuex 3 seems to discourage using getters within mutations

makes sense

PR #3944, opened 2018-06-29
Use file based sessions.

### LINKED ISSUE #3921: Kolibri should explicitly set cookie name
### Observed behavior

When running Kolibri and Studio at the same time in the same browser, Kolibri will immediately sign the user out and require them to sign in again. This process will repeat for as long as both apps are running.

This happens because Kolibri and Studio both currently use the Django default cookie name, 'sessionid'. This causes Kolibri and Studio to set each other's sessionid after various actions.

### Expected behavior
When you sign in, you stay signed in, Studio or not. :)

### User-facing consequences
This is primarily an internal developer issue, though it could happen on devices that deploy other local services in addition to Kolibri that are using Django and keep the default cookie name.

### Steps to reproduce
- Start Studio and sign in
- Start Kolibri and sign in
- Return to Kolibri and try to perform an action, like importing content
Result: It says you are signed out and asks you to sign in again. Once you sign in again, this process will repeat.

### Context
- Latest Kolibri from develop branch as of 6-28-18
- Latest Studio from develop branch as of 6-28-18
- OS X 10.13 (High Sierra)
- Any browser, tested with Safari, Chrome, Chromium, and FIrefox

rtibbles:
### Summary
A potential performance improvement?

### Reviewer guidance
Does it make things faster under load?

Fixes #3921 

----

### Contributor Checklist

### Reviewer Checklist

### DIFF
--- a/kolibri/deployment/default/settings/base.py
+++ b/kolibri/deployment/default/settings/base.py
@@ -339,6 +339,19 @@ JS_REVERSE_EXCLUDE_NAMESPACES = ['admin', ]
 
 ENABLE_DATA_BOOTSTRAPPING = True
 
+# Session configuration
+
+SESSION_ENGINE = "django.contrib.sessions.backends.file"
+
+SESSION_FILE_PATH = os.path.join(conf.KOLIBRI_HOME, "sessions")
+
+if not os.path.exists(SESSION_FILE_PATH):
+    if not os.path.exists(conf.KOLIBRI_HOME):
+        raise RuntimeError("The KOLIBRI_HOME dir does not exist")
+    os.mkdir(SESSION_FILE_PATH)
+
+SESSION_COOKIE_NAME = "kolibri"
+
 SESSION_EXPIRE_AT_BROWSER_CLOSE = True
 
 SESSION_COOKIE_AGE = 600
--- a/kolibri/core/auth/test/test_api.py
+++ b/kolibri/core/auth/test/test_api.py
@@ -4,9 +4,10 @@ from __future__ import unicode_literals
 
 import collections
 import sys
+from importlib import import_module
 
 import factory
-from django.contrib.sessions.models import Session
+from django.conf import settings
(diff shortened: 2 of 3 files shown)


kolibri/deployment/default/settings/base.py
```diff
@@ -339,6 +339,19 @@
...
+SESSION_ENGINE = "django.contrib.sessions.backends.file"
+
+SESSION_FILE_PATH = os.path.join(conf.KOLIBRI_HOME, "sessions")
+
+if not os.path.exists(SESSION_FILE_PATH):
+    if not os.path.exists(conf.KOLIBRI_HOME):
+        raise RuntimeError("The KOLIBRI_HOME dir does not exist")
+    os.mkdir(SESSION_FILE_PATH)
+
+SESSION_COOKIE_NAME = "kolibri"
```
bonidukic, line 353: We should probably name the cookie `sessionid` if we do decide to continue with this method because setting it to `kolibri` breaks tools which send the cookies manually (like Velox e.g.)

rtibbles: Note that I added this because of this issue: https://github.com/learningequality/kolibri/issues/3921

bonidukic: Got it, makes sense.

bonidukic:
Velox test results:
- 30 concurrent users
- exercise channel
- 20 runs for `develop` + 20 runs for `PR#3944`
- sqlite

We're looking at an improvement of ~5% on average, which is not much, but I think it might be worth it, especially since it's a fairly simple change which can be easily reverted if necessary.

number of requests 30 concurrent users

requests 2fs 30 concurrent users

bonidukic (approved):

indirectlylit:
> ~5% on average, which is not much, but I think it might be worth it

agree!

rtibbles:
I will fix the tests then!

PR #3955, opened 2018-07-02
Optimize speed of deletion of old content metadata prior to new import

jamalex:
### Summary

Deleting old content metadata before importing a new version of a channel was super slow. This PR improves the performance in 2 ways:

1. Uses manual SQL queries to delete from specific tables, rather than doing Pythonland cascade deletion.
2. Don't delete data that's going to be overwritten efficiently with REPLACE anyway.

**Times using KA (en) channel:**

- Before: 143.8s to delete, 15.9s to import, total 159.7s
- After: 5.5s to delete, 18.5s to import, total 24s

Speedup: 26x on deletion, 6.7x overall for import updates

### Reviewer guidance

Using a primary database (db.sqlite3) that has an old version of a content database imported into it (e.g. [here's one for KA v10](https://slack-files.com/T0KT5DC58-FBGV1RECC-3b258244f3)), try importing the new version of channel. Watch the logs, and take into account the fact that a large portion of the time may be spent on downloading the new database file.

- The time from when "Older version X of channel..." is shown until "Importing ContentTag data" is shown is the deletion time. This is the part that should be much faster.
- The time from when "Importing ContentTag data" is shown to when "Setting availability of ContentNode objects with children for level 1" is shown is the remainder of the metadata import and annotation (this part shouldn't change much between old versions and this version).

----

### Contributor Checklist

### Reviewer Checklist

### DIFF
--- a/kolibri/content/utils/channel_import.py
+++ b/kolibri/content/utils/channel_import.py
@@ -1,6 +1,7 @@
 import logging as logger
 
 from django.apps import apps
+from django.db.models.fields.related import ForeignKey
 from sqlalchemy.exc import OperationalError
 from sqlalchemy.exc import SQLAlchemyError
 from sqlalchemy.sql import text
@@ -13,7 +14,6 @@ from .sqlalchemybridge import Bridge
 from .sqlalchemybridge import ClassNotFoundError
 from kolibri.content.apps import KolibriContentConfig
 from kolibri.content.legacy_models import License
-from kolibri.content.models import AssessmentMetaData
 from kolibri.content.models import ChannelMetadata
 from kolibri.content.models import CONTENT_SCHEMA_VERSION
 from kolibri.content.models import ContentNode
@@ -36,7 +36,6 @@ merge_models = [
     ContentTag,
     LocalFile,
     Language,
-    ChannelMetadata,
 ]
 
 
@@ -396,38 +395,60 @@ class ChannelImport(object):
 
         return True
 
-    def _can_use_optimized_pre_deletion(self):
-        # check whether we can skip fully deleting these three models, if we'll be using REPLACE on them anyway
-        for model in [ContentNode, File, AssessmentMetaData]:
-            mapping = self.schema_mapping.get(model, {})
-            row_mapper = self.generate_row_mapper(mapping.get('per_row'))
-            table_mapper = self.generate_table_mapper(mapping.get('per_table'))
-            if not self.can_use_sqlite_attach_method(model, row_mapper, table_mapper):
-                return False
-        return True
+    def _can_use_optimized_pre_deletion(self, model):
+        # check whether we can skip fully deleting this model, if we'll be using REPLACE on it anyway
+        mapping = self.schema_mapping.get(model, {})
+        row_mapper = self.generate_row_mapper(mapping.get('per_row'))
+        table_mapper = self.generate_table_mapper(mapping.get('per_table'))
+        return self.can_use_sqlite_attach_method(model, row_mapper, table_mapper)
 
     def delete_old_channel_data(self, old_tree_id):
 
-        # construct a template for deleting records for models that have a foreign key onto ContentNode
-        delete_related_template = "DELETE FROM {table} WHERE contentnode_id IN (SELECT id FROM content_contentnode WHERE tree_id = '{tree_id}')"
+        # construct a template for deleting records for models that foreign key onto ContentNode
+        delete_related_template = """
+            DELETE FROM {table}
+                WHERE {fk_field} IN (
+                    SELECT id FROM {cn_table} WHERE tree_id = '{tree_id}'
+                )
+        """
 
-        # construct a query for deleting old ContentNode records
-        delete_contentnode_query = "DELETE FROM content_contentnode WHERE tree_id = '{}'".format(old_tree_id)
+        # construct a template for deleting the ContentNode records themselves
+        delete_contentnode_template = "DELETE FROM {table} WHERE tree_id = '{tree_id}'"
 
-        if self._can_use_optimized_pre_deletion():
-            # if the external database is attached and there are no mappings for the relevant tables,
+        # we want to delete all content models, but not "merge models" (ones that might also be used by other channels), and ContentNode last
+        models_to_delete = [model for model in self.content_models if model is not ContentNode and model not in merge_models] + [ContentNode]
+
+        for model in models_to_delete:
+
+            # we do a few things differently if it's the ContentNode model, vs a model related to ContentNode
+            if model is ContentNode:
+                template = delete_contentnode_template
+                fields = ["id"]
+            else:
+                template = delete_related_template
+                fields = [f.column for f in model._meta.fields if isinstance(f, ForeignKey) and f.target_field.model is ContentNode]
+
+            # if the external database is attached and there are no incompatible schema mappings for a table,
             # we can skip deleting records that will be REPLACED during import, which helps efficiency
-            delete_related_template += " AND NOT id IN (SELECT id FROM sourcedb.{table})"
-            delete_contentnode_query += " AND NOT id IN (SELECT id FROM sourcedb.content_contentnode)"
+            if self._can_use_optimized_pre_deletion(model):
+                template += " AND NOT id IN (SELECT id FROM sourcedb.{table})"
 
-        # construct queries for deletion of File and AssessmentMetaData model records
-        delete_file_query = delete_related_template.format(table="content_file", tree_id=old_tree_id)
-        delete_assessment_query = delete_related_template.format(table="content_assessmentmetadata", tree_id=old_tree_id)
+            # run a query for each field this model has that foreignkeys onto ContentNode
+            for field in fields:
 
-        # run our deletion queries, starting with the related models (as they depend on the ContentNode records still existing)
-        for query in [delete_file_query, delete_assessment_query, delete_contentnode_query]:
-            self.destination.session.execute(text(query))
-            self.check_cancelled()
+                # construct the actual query by filling in variables
+                query = template.format(
+                    table=model._meta.db_table,
+                    fk_field=field,
+                    tree_id=old_tree_id,
+                    cn_table=ContentNode._meta.db_table
+                )
+
+                # check that the import operation hasn't since been cancelled
+                self.check_cancelled()
+
+                # execute the actual query
+                self.destination.session.execute(text(query))
 
     def check_cancelled(self):
         if callable(self.cancel_check):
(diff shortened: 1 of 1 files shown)


kolibri/content/utils/channel_import.py
```diff
@@ -376,33 +378,57 @@ def check_and_delete_existing_channel(self):
...
+            row_mapper = self.generate_row_mapper(mapping.get('per_row'))
+            table_mapper = self.generate_table_mapper(mapping.get('per_table'))
+            if not self.can_use_sqlite_attach_method(model, row_mapper, table_mapper):
+                return False
+        return True
+
+    def delete_old_channel_data(self, old_tree_id):
+
+        # construct a template for deleting records for models that have a foreign key onto ContentNode
+        delete_related_template = "DELETE FROM {table} WHERE contentnode_id IN (SELECT id FROM content_contentnode WHERE tree_id = '{tree_id}')"
```
rtibbles: Would be nice if we could avoid hardcoding the SQL table names and column names - if we can derive them from the SQLAlchemy table definitions would make this feel a bit DRYer.

kolibri/content/utils/channel_import.py
```diff
@@ -376,33 +378,57 @@ def check_and_delete_existing_channel(self):
...
+            # we can skip deleting records that will be REPLACED during import, which helps efficiency
+            delete_related_template += " AND NOT id IN (SELECT id FROM sourcedb.{table})"
+            delete_contentnode_query += " AND NOT id IN (SELECT id FROM sourcedb.content_contentnode)"
+
+        # construct queries for deletion of File and AssessmentMetaData model records
+        delete_file_query = delete_related_template.format(table="content_file", tree_id=old_tree_id)
+        delete_assessment_query = delete_related_template.format(table="content_assessmentmetadata", tree_id=old_tree_id)
+
+        # run our deletion queries, starting with the related models (as they depend on the ContentNode records still existing)
+        for query in [delete_file_query, delete_assessment_query, delete_contentnode_query]:
```
rtibbles: Will need to add queries to delete the M2M tables for ContentTag to ContentNode, and the ContentNode prerequisite, and related tables.

rtibbles (changes requested):
Makes sense to me, some cleanup would be nice, and this may fail for channels with tags/prerequisites/related nodes.

kolibri/content/utils/channel_import.py
```diff
@@ -376,33 +377,79 @@ def check_and_delete_existing_channel(self):
...
-            self.destination.session.flush()
+
         return True
 
+    def _can_use_optimized_pre_deletion(self, model):
+        # check whether we can skip fully deleting this model, if we'll be using REPLACE on it anyway
+        mapping = self.schema_mapping.get(model, {})
+        row_mapper = self.generate_row_mapper(mapping.get('per_row'))
+        table_mapper = self.generate_table_mapper(mapping.get('per_table'))
+        return self.can_use_sqlite_attach_method(model, row_mapper, table_mapper)
```
jamalex, line 403: Updated this method to be model-specific, so we can selectively use these methods for whichever tables might allow it.

radinamatic:
Waiting for the asset to start testing! 👍

rtibbles (approved):
Much cleaner. I will defer to mr travis for final merge.

radinamatic:
> Altro che volare... ❤️ 

PEX on Ubuntu 16.04, VM with 2GB RAM, single core:

- 18s to delete
- 41s to import

```INFO 2018-07-03 03:08:00,600 ping Attempting a ping.
INFO 2018-07-03 03:08:00,648 ping data: {"instance_id": "c394cef1b732b2e5761db4bc69258993", "platform": "Linux-4.4.0-128-generic-x86_64-with-Ubuntu-16.04-xenial", "database_id": "a50ab60977814c40b16870f5da2b8a4c", "version": "0.10.0.post1", "system_id": "", "mode": "LE-testingHD-U16043", "language": "en", "sysversion": "2.7.12 (default, Dec  4 2017, 14:50:18) \n[GCC 5.4.0 20160609]", "uptime": 0, "node_id": "baa72ad1f5714b23762a"}
INFO 2018-07-03 03:08:00,781 utils Started new MESSAGEPROCESSOR thread ID#lake-pasta-ack-july
INFO 2018-07-03 03:08:00,781 utils Started new SCHEDULER thread ID#washington-lima-utah-nebraska
INFO 2018-07-03 03:08:00,785 utils Started new WORKERMESSAGEHANDLER thread ID#jig-lithium-spaghetti-fillet
INFO 2018-07-03 03:08:01,232 ping Ping succeeded! (response: {}) Sleeping for 1440.0 minutes.
INFO 2018-07-03 03:09:27,796 importchannel Downloading data for channel id 1ceff53605e55bef987d88e0908658c5
INFO 2018-07-03 03:09:35,471 channel_import Older version 10 of channel 1ceff53605e55bef987d88e0908658c5 already exists in database; removing old entries so we can upgrade to version 11
INFO 2018-07-03 03:09:57,587 channel_import Importing ContentTag data
INFO 2018-07-03 03:09:57,592 channel_import Importing ContentNode_has_prerequisite data
INFO 2018-07-03 03:09:57,594 channel_import Importing ContentNode_related data
INFO 2018-07-03 03:09:57,594 channel_import Importing ContentNode_tags data
INFO 2018-07-03 03:09:57,595 channel_import Importing ContentNode data
INFO 2018-07-03 03:10:04,243 channel_import Importing Language data
INFO 2018-07-03 03:10:04,245 channel_import Importing File data
INFO 2018-07-03 03:10:26,887 channel_import Importing LocalFile data
INFO 2018-07-03 03:10:28,522 channel_import Importing AssessmentMetaData data
INFO 2018-07-03 03:10:28,952 channel_import Importing ChannelMetadata data
INFO 2018-07-03 03:10:31,519 annotation Setting availability of File objects based on LocalFile availability
INFO 2018-07-03 03:10:36,025 annotation Setting availability of non-topic ContentNode objects based on File availability
INFO 2018-07-03 03:10:38,093 annotation Setting availability of ContentNode objects with children for 5 levels
INFO 2018-07-03 03:10:38,094 annotation Setting availability of ContentNode objects with children for level 5
INFO 2018-07-03 03:10:38,249 annotation Setting availability of ContentNode objects with children for level 4
INFO 2018-07-03 03:10:38,296 annotation Setting availability of ContentNode objects with children for level 3
INFO 2018-07-03 03:10:38,300 annotation Setting availability of ContentNode objects with children for level 2
INFO 2018-07-03 03:10:38,307 annotation Setting availability of ContentNode objects with children for level 1```

**Comment by indirectlylit** [comment]
thank you so much @rtibbles, @radinamatic, and especially @jamalex

**Comment by radinamatic** [comment]
For comparison, the whole process in 0.10b6 PEX took 15 minutes:

```INFO 2018-07-03 03:26:55,360 importchannel Downloading data for channel id 1ceff53605e55bef987d88e0908658c5
INFO 2018-07-03 03:27:04,300 channel_import Older version 10 of channel 1ceff53605e55bef987d88e0908658c5 already exists in database; removing old entries so we can upgrade to version 11
ERROR 2018-07-03 03:38:10,908 exception The request's session was deleted before the request completed. The user may have logged out in a concurrent request, for example.
INFO 2018-07-03 03:41:09,141 channel_import Importing ContentTag data
INFO 2018-07-03 03:41:20,440 channel_import Importing ContentNode_has_prerequisite data
INFO 2018-07-03 03:41:20,761 channel_import Importing ContentNode_related data
INFO 2018-07-03 03:41:20,785 channel_import Importing ContentNode_tags data
INFO 2018-07-03 03:41:20,807 channel_import Importing ContentNode data
INFO 2018-07-03 03:41:31,090 channel_import Importing Language data
INFO 2018-07-03 03:41:31,116 channel_import Importing File data
INFO 2018-07-03 03:42:14,689 channel_import Importing LocalFile data
INFO 2018-07-03 03:42:18,336 channel_import Importing AssessmentMetaData data
INFO 2018-07-03 03:42:21,676 channel_import Importing ChannelMetadata data
INFO 2018-07-03 03:42:32,727 annotation Setting availability of File objects based on LocalFile availability
INFO 2018-07-03 03:42:39,198 annotation Setting availability of non-topic ContentNode objects based on File availability
INFO 2018-07-03 03:42:43,658 annotation Setting availability of ContentNode objects with children for 5 levels
INFO 2018-07-03 03:42:43,668 annotation Setting availability of ContentNode objects with children for level 5
INFO 2018-07-03 03:42:43,822 annotation Setting availability of ContentNode objects with children for level 4
INFO 2018-07-03 03:42:43,868 annotation Setting availability of ContentNode objects with children for level 3
INFO 2018-07-03 03:42:43,881 annotation Setting availability of ContentNode objects with children for level 2
INFO 2018-07-03 03:42:43,887 annotation Setting availability of ContentNode objects with children for level 1```

**Comment by jamalex** [comment]
Wooooooooo! 14m down to 22s for the deletion portion!

**Comment by benjaoming** [comment]
> Uses manual SQL queries to delete from specific tables, rather than doing Pythonland cascade deletion.

@jamalex This doesn't happen if you call `delete()` on a QuerySet. It happens if you call `delete()` on an object instance.

I am just reciting this out of memory, but also a distinct feeling that I've done it before :) If you don't mind, could you try and see if things are still working if you make this change?

In case the below can work, I suggest opening up a "tech debt" labeled issue.. I'm getting a bit worried that we're using SQL statements, even just very small and very contained ones :)

```python
delete_related_template = """
    DELETE FROM {table}
        WHERE {fk_field} IN (
            SELECT id FROM {cn_table} WHERE tree_id = '{tree_id}'
        )
"""
```

Becomes:

```python
function delete_ids_by_fk_field_in_tree(DeleteFromModel, ChannelModel, fk_field, tree_id):
    # Construct the IN query
    kwargs={fk_field + "__in": ChannelModel.objects.filter(tree_id=tree_id).values('id')}

    # Construct and execute the DELETE query
    DeleteFromModel.filter(**kwargs).delete()
```

rtibbles:
@benjaoming note that we can't do this in Django, this was moved from Django to SQLAlchemy in #3538 in order to allow the deletion to happen within the same transaction as the channel import process that happens within SQLAlchemy (as I have yet to find a way to make Django and SQLAlchemy share a transaction).

The slowness that @jamalex was referring to was mostly in the SQLAlchemy reflection of the relationships defined within Django (which when triggering its cascade behaviour would somehow do something more inefficient than the Django defined cascade behaviour).

The only open question for this is whether translating these statements into SQLAlchemy core syntax as opposed to raw SQL - but ultimately the core syntax is just a Pythonic way of constructing SQL expressions.

benjaoming:
> ultimately the core syntax is just a Pythonic way of constructing SQL expressions

Yes, agreed -- I don't know about advantages of SQLAlchemy, I wasn't aware that these components were living there.

In Django, the advantage would be about controlling model and field references in Python, as opposed to coding them into SQL strings + Being DB agnostic + Also, custom managers may influence querysets in some global way.

Anyhow, alright, seems like we'll stick to the raw SQL :)

benjaoming:
Thanks for explaining @rtibbles :)

PR #3969, opened 2018-07-05
Fix implicit derived dependencies that occur at build-time dep bundling

benjaoming:
CI on `develop` branch for #3968


indirectlylit (approved):

benjaoming:
Alright, since this was merged, I'll cherry-pick it for other branches.

Sorry about not being explicit, I only opened this for verifying that it was passing tests, but intend to merge it from 0.6.x and up..

indirectlylit:
ah, whoops.

>  intend to merge it from 0.6.x and up..

Yes, let's do that and then just resolve the conflict when it gets to develop. No need to cherrypick

benjaoming:
Yup, can work on that tomorrow, but you're welcome to keep this merged as is, no reverting necessary. At least then `develop` is still operational 'til then.

PR #3975, opened 2018-07-05
Network utilities for parsing addresses and URLs, storing NetworkLocation, and loading content P2P

jamalex:
### Summary

Backend utilities to support:
https://trello.com/c/y6RN8gSm/187-manual-port-address-entry

@ralphiee22 is working on some other pieces of this (the /api/public/v1/channel... endpoints), after which this will be complete.

### Reviewer guidance

You can test creating and listing NetworkLocation records at http://127.0.0.1:8080/api/networklocation/

(it won't let you add addresses that aren't serving up the /api/public/info endpoint, i.e. instances of Kolibri running from this PR, or Studio after https://github.com/learningequality/studio/pull/919 is merged in)

In the list view, should show "available" for any address it's currently able to connect to.

@jonboiser For integration with the frontend:

- GET /api/networklocation/ to get the list of current addresses, annotated with their availability.
- POST /api/networklocation/ (with JSON with "base_url" key, and optionally "device_name" key, which will otherwise be filled in based on the hostname of the device it connects to) to add an address. If it succeeds, it'll return the full record with info about the other server, otherwise it'll return a 400 with an error constant code id (https://github.com/learningequality/kolibri/pull/3975/files#diff-e81537358153cccdb34f57e882df426aR7)
- On the startremotechannelimport, startremotecontentimport, and remotechannel endpoints, add a "baseurl" parameter (JSON key or GET param, as appropriate) to specify the server to sync/query against. The value passed for this is the base URL itself (e.g. http://192.51.22.1:8080/), not the ID of the NetworkLocation etc.

----

### Contributor Checklist

### Reviewer Checklist


kolibri/core/content/api.py
```diff
@@ -673,20 +667,23 @@ def list(self, request, *args, **kwargs):
...
         """
-        return self._cache_kolibri_studio_channel_request(identifier=pk)
+        baseurl = request.GET.get("baseurl", None)
+        return self._make_channel_endpoint_request(identifier=pk, baseurl=baseurl)
 
     @list_route(methods=['get'])
     def kolibri_studio_status(self, request, **kwargs):
         try:
-            resp = requests.get(get_channel_lookup_url())
+            # TODO: switch to /api/public/info/ endpoint once that's live on Studio
```
jamalex: TODO after https://github.com/learningequality/studio/pull/919 is deployed

jonboiser:
Tried POSTING with a bad address like "abc" and am getting a `error_constants.NETWORK_LOCATION_NOT_FOUND` instead of a URL formatting error.

jonboiser:
On the DRF API GUI, it looks like instance ID, OS, application, and kolibri version are writeable. Should they be read-only?

jamalex:
> Tried POSTING with a bad address like "abc" and am getting a error_constants.NETWORK_LOCATION_NOT_FOUND instead of a URL formatting error.

I think that might be because "abc" technically isn't a bad hostname (even if it's not a FQDN). For example, it works to put in "localhost" as the address. And on a LAN, basically any alphanumeric string might be a valid local network host.

Here are some examples of "bad addresses":
- 11111.11.1111.11
- qqqq://hello.com
- abc abc (with a space in it)

jamalex:
> On the DRF API GUI, it looks like instance ID, OS, application, and kolibri version are writeable. Should they be read-only?

Good plan, made them read-only now.

kolibri/core/content/api.py
ralphiee22: you can update this now

kolibri/core/discovery/utils/network/client.py
```diff
@@ -0,0 +1,56 @@
...
+    def __init__(self, base_url=None, address=None, **kwargs):
+        """If an explicit base_url is already known, provide that. If a vague address is provided, we can try to infer the base_url"""
+        if not base_url and not address:
+            raise Exception("You must provide either a `base_url` or `address` argument")
+        self.session = requests.Session(**kwargs)
+        if base_url:
+            self.base_url = self._attempt_connections([base_url])
+        else:
+            # normalize the URL and try a number of variations until we find one that's able to connect
+            logger.info("Attempting connections to variations of the URL: {}".format(base_url))
```
ralphiee22, line 23: will we actually log the `base_url` if it not supposed to be defined by the time we get here?

kolibri/core/discovery/utils/network/client.py
```diff
@@ -0,0 +1,56 @@
...
+            self.base_url = self._attempt_connections(get_normalized_url_variations(address))
+
+    def _attempt_connections(self, urls):
+        # try each of the URLs in turn, returning the first one that succeeds
+        for url in urls:
+            try:
+                logger.info("Attempting connection to: {}".format(url))
+                response = self.get("/api/public/info/", base_url=url, timeout=5, allow_redirects=True)
```
ralphiee22, line 31: once my PR is merged should be able to use `get_info_url`

ralphiee22 (approved):

kolibri/core/content/api.py
ralphiee22: after my PR is merged you can use the `get_info_url`

kolibri/core/discovery/migrations/0002_auto_20180825_1744.py
```diff
@@ -0,0 +1,36 @@
+# -*- coding: utf-8 -*-
```
ralphiee22: since we haven't merged this, could probably just make a single migration file for now.

Issue #3141, opened 2018-01-31
localize max-length validation messaging in text box


indirectlylit:
### Observed behavior

The 'max length' user feedback is not currently localized: https://github.com/learningequality/Keen-UI/blob/17911401dff3d65f78bc1edbc32acf675864da03/src/UiTextbox.vue#L84

### Expected behavior

All user-facing numerals should be localized. In this case, we need to define a translation string that includes both the current and max lengths, i.e.:

```
$trs {
  lengthInfo: '{current_val, number, integer} / {max_val, number, integer}',
}
```

### User-facing consequences

Users of non-western languages might struggle with text input across the application.

### Context

see initial work here: #3107

indirectlylit:
@christianmemije in order to fix this, are we going to need to fork the component?

PR #4011, opened 2018-07-10
Localize texts in coach-content-label

### LINKED ISSUE #3970: number of coach resources is not internationalized
### Observed behavior

On topics in Learn, the number shows as a 'latin'-style numeral:

image

### Expected behavior

Number should be properly internationalized

### User-facing consequences

users will not be able to read the text

### Steps to reproduce

Log in as a coach and view a topic in 'Learn' that has coach-specific content

### Context

0.10.0

jonboiser:
### Summary

Localize the title meta and label text for coach-content-label.

screen shot 2018-07-10 at 1 42 27 pm

### Reviewer guidance

1. Install some coach content
1. Look in different UIs where label occurs in a RTL language.

### References

Fixes #3970
----

### Contributor Checklist

### Reviewer Checklist

### DIFF
--- a/kolibri/core/assets/src/views/coach-content-label.vue
+++ b/kolibri/core/assets/src/views/coach-content-label.vue
@@ -5,7 +5,7 @@
       <mat-svg name="local_library" category="maps" />
     </ui-icon>
     <span class="counter" v-if="isTopic">
-      {{ value }}
+      {{ $formatNumber(value) }}
     </span>
   </div>
 
@@ -42,7 +42,7 @@
     },
     $trs: {
       coachResourceLabel: 'Coach resource',
-      topicTitle: 'Contains {count} {count, plural, one {coach resource} other {coach resources}}',
+      topicTitle: 'Contains {count, number, integer} {count, plural, one {coach resource} other {coach resources}}',
     },
   };
 


kolibri/core/assets/src/views/coach-content-label.vue
```diff
@@ -33,9 +33,12 @@
       },
     },
     computed: {
+      valueText() {
+        return this.$formatNumber(this.value);
```
indirectlylit: I _think_ the more correct fix might actually be to use the [number type](https://formatjs.io/guides/message-syntax/#number-type) in ICU syntax like:

```
topicTitle: 'Contains {count, number} {count, plural, one {coach resource} other {coach resources}}',
```

@rtibbles I think you added the `$formatNumber` function? Do you know under what circumstances it's appropriate to use? (And then we can document it in our [i18n docs](http://kolibri-dev.readthedocs.io/en/develop/references/i18n.html).)

rtibbles: `$formatNumber` is the vue-intl method for just formatting a number (in the case where it is used outside of any particular text string). As you note, if used within the context of a string, then the ICU syntax should be used (which ultimately will do, in essence, the `$formatNumber` call under the hood during the string interpolation).

indirectlylit: Hmm... looks like I might be wrong. 

Searching the code for the string `, number}` returns bunch of examples that don't get properly internationalized. For instance the setup wizard:

```javascript
$trs: {
  progressIndicator: 'Step {currentStep, number} of {totalSteps, number}',
},
```

image

This to seems fairly problematic to me ...

rtibbles: Hrm, seems like an issue with vue-intl then, it should be running that conversion on the interpolation. Can you open an issue over there, and I'll investigate?

indirectlylit: https://github.com/learningequality/vue-intl/issues/17

rtibbles: It also seems like we need to vendor ui-textbox in order to internationalize the `0/50` also.

jonboiser: I think the `valueText` computed string is still necessary, since the coach label is basically just `<icon> <number>`, and the number needs to be localized.

The `titleText` computed string can be localized by changing `{count}` to `{count, number}` and fixing the issue with vue-intl.

indirectlylit: > It also seems like we need to vendor ui-textbox in order to internationalize the `0/50` also.

yes, I think that's right (can also submit a patch upstream). Had #3141 open for that already

> I think the `valueText` computed string is still necessary, since the coach label is basically just `<icon> <number>`, and the number needs to be localized.

There are ~a fair number of~ four strings in the code that are only a localized number, with no text to translate - for example:

https://github.com/learningequality/kolibri/blob/0aa7dabaa28059416691f5d9f5cb63696a33910c/kolibri/plugins/coach/assets/src/views/exams/exam-report-page/index.vue#L229

However these strings might be confusing for translators, so I can see an argument for using `$formatNumber` instead. In either case we should try to be consistent.

Does `$formatNumber` support all the modifiers and variations that ICU syntax supports?

rtibbles: Yes: https://github.com/learningequality/vue-intl#formatnumber

indirectlylit (approved):

Issue #4014, opened 2018-07-10
Content import progress bar not reaching 100% 


jeepurs:
### Observed behavior

The progress bar for importing a content channel via USB stopped at a % below 100% even though it was finished.

### Expected behavior

For cases where the import process has to skip files or import incomplete files, force 100% at the end of import. 

### User-facing consequences

Confusion about the finished or unfinished status of content import

### Errors and logs

…

### Steps to reproduce

1. Export a channel to USB but interrupt it so it doesn't finish
2. On another device, import the channel via USB to Kolibri
3. Expect the progress bar to reach 100% eventually. It doesn't

### Context

Kolibri v0.10
Raspbian (R Pi 3+)
Chromium browser

![img_20180710_143338](https://user-images.githubusercontent.com/5150483/42542060-edb62032-8459-11e8-963f-ed48791d36d3.jpg)

jamalex:
>Export a channel to USB but interrupt it so it doesn't finish

Would likely also happen if you export just a subset (not just if you interrupt).

lyw07:
Hi! I tried to import an entire channel from Studio to Kolibri, and then export only a subset of the channel to USB. After that, I deleted the entire channel in Kolibri and tried to import the entire channel from USB. I can see from the logs that some number of files are skipped, but the progress bar reached 100%. I'm not sure which step I didn't follow correctly. Do you mind giving me more instructions to debug? Thank you!

jeepurs:
I would try importing on a completely different server if you want to replicate what i did exactly.

lyw07:
I tried again to import to a different server but still couldn't reproduce the issue... Here are my steps:
1. import the entire channel from Studio to Kolibri
2. export only a subset of the channel to USB
3. create a virtual machine, and set the USB's KOLIBRI_DATA folder to be a shared folder of the virtual machine (this should work as an external drive to the virtual machine)
4. try to import the entire channel from the shared folder (external drive). The progress bar reached 100%, and the log shows that 73 files are skipped.

indirectlylit:
I wonder if this is the same issue as #614, which was closed as not reproducible?

benjaoming:
Fixed in #4059 and released in 0.10.1

PR #4017, opened 2018-07-11
Fix incomplete progress bar

lyw07:
### Summary

This PR is to fix the issue that the size of the source file is smaller than the file size indicated in the database, which means the source file is corrupted. In the github issue #4014, the contentnode that is imported has three local files. The first local file doesn't exist, so it is skipped. The second local file exists, so it is successfully imported. The third local file exists in the disk but has file size `0`, so nothing is added to the progress bar, which caused the error that the progress bar didn't reach 100%. 
The fix is to skip the corrupted files like this, by directly adding the file sizes to the progress bar and not actually importing the files. 

### Reviewer guidance

The data to test is here: https://slack-files.com/T0KT5DC58-FBNQ0K9QB-b27524fe7d
1. put the data folder into an external drive and try to import the entire channel `Lebanon LE Training` from the external drive.
2. check whether the progress bar reached 100%.
3. note that since the channel only contains one content node, and this content node is not successfully imported, the channel will not show up in the imported channel list even after the progress bar reached 100%.

I also updated some error messages. Please feel free to let me know if you think the wording could be improved. Thank you!

### References

#4014 

----

### Contributor Checklist

### Reviewer Checklist

### DIFF
--- a/kolibri/content/management/commands/importcontent.py
+++ b/kolibri/content/management/commands/importcontent.py
@@ -167,7 +167,7 @@ class Command(AsyncCommand):
 
             if number_of_skipped_files > 0:
                 logging.warning(
-                    "{} files are skipped, because they do not exist.".format(
+                    "{} files are skipped, because errors occured during the import.".format(
                         number_of_skipped_files))
 
             if self.is_cancelled():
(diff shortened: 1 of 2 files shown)


lyw07:
tests are failing. I will fix them soon

rtibbles:
Tested, works. Might be a bit misleading for the user in this particular test case (it showing success but nothing importing), but this fixes the issue!

rtibbles (approved):
Works, and is about as good as we can do until we can do more granular reporting of task progress.

lyw07:
@rtibbles thank you! Yes I agree that it's very misleading to users. I guess it would be better if it fails as 0%? But in this case, we might need to update the progress not per local file but per content node?

Issue #3309, opened 2018-03-06
Various errors when exiting a Exercises and PDFs in a Lesson


jonboiser:
Getting different errors which seem to be due to quirks of the individual content renderers.

To reproduce:

1. Create, active, assign a Lesson with a PDF doc and/or Perseus Exercise
1. Go to the Lesson and view the Resource
1. Hit the "<-" button on the top bar
1. In the console, observe the Error.

This does not seem to mess with logging or affect any part of the User Experience, but does point to some code issues.

Don't see this error with HTML5 apps or videos. Couldn't find a channel with just MP3 audio.

## Exiting an exercise

Get this error
```
{request: {…}, url: "/api/attemptlog/?1520367692407=1520367692407", raw: XMLHttpRequest, status: {…}, headers: {…}, …}
entity
:
end_timestamp
:
["This field may not be null."]
__proto__
:
Object
```

## Exiting a PDF

```
[Vue warn]: Error in event handler for "updateProgress": "TypeError: Cannot read property 'channel_id' of undefined"

found in

---> <ContentRender> at kolibri/core/assets/src/views/content-renderer/index.vue
       <LearnContent> at kolibri/plugins/learn/assets/src/views/content-page/index.vue
         <LessonResourceViewer> at kolibri/plugins/learn/assets/src/views/classes/LessonResourceViewer.vue
           <AppBody> at kolibri/core/assets/src/views/app-body/index.vue
             <CoreBasePage> at kolibri/core/assets/src/views/core-base.vue
               <Root>
```

indirectlylit:
thanks. since it doesn't seem to have an impact we can probably live with it for a bit but definitely don't want it to linger

rtibbles:
My main concern here is that these are data logging errors, although it's possible this is happening because of our use of `beforeDestroy` handlers to try to preserve data.

indirectlylit:
yes if this is affecting logging progress on-close we should probably aim for 0.8

(ref #2355)

DXCanas:
> yes if this is affecting logging progress

Progress is logged in a few different places - this affects the progress that gets logged on scroll, but progress gets logged at the end of the "target" amount of time anyway - we can bypass the `onScroll` log and just have update on an interval more often?

DXCanas:
Been finding a bunch of false positive leads on this.

[...]

Issue #2355, opened 2017-10-05
Don't use beforeDestroy to affect tracking and progress logging logic


indirectlylit:
Vue lifecycle hooks - particularly `beforeDestroy` - should not be used to trigger application logic.

The vue component lifecycle hooks are tied to the DOM lifecycle, not the application lifecycle. `beforeDestroy` should only be used for low-level cleanup operations like removing DOM event listeners, clearing timers, and cleaning up callbacks.

Application logic that might currently be triggered by `beforeDestroy` should instead be triggered by whatever vuex action or state change actually _caused_ the component to be destroyed.

-----

refs:

* https://github.com/learningequality/kolibri/pull/2034#issuecomment-325529393, [resulting change](https://github.com/learningequality/kolibri/pull/2045/files#diff-0ca7a9fe05f901d2fb1805decc333920L293)
* #2354
* #4018

-----

Some specific things to look for:

* `this.$emit('stopTracking');`
* `this.saveAttemptLogMasterLog(false);`
* `this.stopTracking();`

rtibbles:
Won't fix.

PR #4018, opened 2018-07-11
Fix logging race condition in assessment-wrapper

jonboiser:
### Summary

Fixes issue where the `success` getter in `assessment-wrapper` errored out because of incompatible `pageState` at the `beforeDestroy` hook.

### Reviewer guidance

1. As a Learner, Interact with an exercise
1. Leave the exercise
1. You should not see any errors in the console

### Contributor Checklist

### Reviewer Checklist

### DIFF
--- a/kolibri/plugins/learn/assets/src/views/assessment-wrapper/index.vue
+++ b/kolibri/plugins/learn/assets/src/views/assessment-wrapper/index.vue
@@ -175,19 +175,24 @@ oriented data synchronization.
         default: () => Promise.resolve(),
       },
     },
-    data: () => ({
-      ready: false,
-      itemId: '',
-      shake: false,
-      firstAttemptAtQuestion: true,
-      complete: false,
-      correct: 0,
-      itemError: false,
-      hintWasTaken: false,
-      // Attempted fix for #1725
-      checkingAnswer: false,
-      checkWasAttempted: false,
-    }),
+    data() {
+      return {
+        ready: false,
+        itemId: '',
+        shake: false,
+        firstAttemptAtQuestion: true,
+        complete: false,
+        correct: 0,
+        itemError: false,
+        hintWasTaken: false,
+        // Attempted fix for #1725
+        checkingAnswer: false,
+        checkWasAttempted: false,
+        // Placing these here so they are available at beforeDestroy
+        masteryModel: this.$store.state.pageState.content.masteryModel,
+        assessmentIds: this.$store.state.pageState.content.assessmentIds,
+      };
+    },
     computed: {
       ...mapGetters(['isUserLoggedIn']),
       ...mapState({
@@ -197,8 +202,6 @@ oriented data synchronization.
           (state.core.logging.mastery.pastattempts || []).filter(attempt => attempt.error !== true),
         userid: state => state.core.session.user_id,
         content: state => state.pageState.content,
-        assessmentIds: state => state.pageState.content.assessmentIds,
-        masteryModel: state => state.pageState.content.masteryModel,
         randomize: state => state.pageState.content.randomize,
       }),
       recentAttempts() {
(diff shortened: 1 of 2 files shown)


kolibri/plugins/learn/assets/src/views/assessment-wrapper/index.vue
```diff
@@ -175,19 +175,24 @@ oriented data synchronization.
...
+        complete: false,
+        correct: 0,
+        itemError: false,
+        hintWasTaken: false,
+        // Attempted fix for #1725
+        checkingAnswer: false,
+        checkWasAttempted: false,
+        // Placing these here so they are available at beforeDestroy
+        masteryModel: this.$store.state.pageState.content.masteryModel,
+        assessmentIds: this.$store.state.pageState.content.assessmentIds,
```
indirectlylit, line 193: clever workaround

indirectlylit (commented):
related issue: #2355

bonidukic:
@jonboiser Actually, I still see the errors. I'm attaching my developers console log. I can also make a video again if necessary. Let me know.

[localhost-1531348623144.log](https://github.com/learningequality/kolibri/files/2186426/localhost-1531348623144.log)

jonboiser:
The response I'm getting back from `/api/attemptlog` is that `end_timestamp` cannot be null when saving an attempt log.

I think this has been an issue for a long time, but may be race-condition related as well.

bonidukic:
@jonboiser You're right, it comes back to this issue: https://github.com/learningequality/kolibri/issues/3309

jonboiser:
That's the error, but the stack trace doesn't make sense to me. The Failed request to `api/attemptlog` looks like it originates from a call of `ContentSummaryLogResource.save()`, which is supposed to result in a PATCH on a different endpoint, 

**content summary endpoint**
https://github.com/jonboiser/kolibri/blob/d3facff498f4987ee8892e68f724031f0d045926/kolibri/plugins/coach/api_urls.py#L18

**where the `.save()` is called**
https://github.com/jonboiser/kolibri/blob/d3facff498f4987ee8892e68f724031f0d045926/kolibri/core/assets/src/state/modules/core/actions.js#L411

jonboiser:
@bonidjukic I added a line to provide a default `end_timestamp` to make the error go away, but might be missing the root problem, since I assume that timestamp should be set somewhere along the way before leaving the Exercise.

rtibbles:
Is it possible that the vuex state has been cleared by the time it gets to this line in the save? (assuming that the routing is the thing triggering the teardown of the component, and that then causes the pageState to be reset.

This reminds me of the issue with debounced saving of examlogs, where the state was getting cleared before it could be executed. I resolved that by explicitly passing arguments into the action, rather than relying on reading potentially volatile state inside the action.

jonboiser:
I've been producing these errors most reliably by doing this:

1. Visiting an exercise
1. Leaving the exercise without 'attempting' it. (leaving  `end_timestamp` at `null`)

I think this is the issue: that an attemptlog is being saved even if there are no attempts. Is this incorrect behavior possibly caused by the shifting vuex state?

jonboiser:
It seems that `assessment-wrappers` `beforeDestroyHook` will call `saveAttemptLog` regardless of whether there has been an attempt. Maybe there should be a guard around that save call?

rtibbles:
Hrm... no, perhaps the issue is more that we should be checking if any attempt happened, and only saving in that case. It doesn't really make sense to save an attempt where no attempts were made!

jonboiser:
@rtibbles I added a guard in the assessment-wrappers `beforeDestroy` hook: It will only trigger the saveAttemptLog/saveMasteryLog actions if there was at least one attempt on the Exercise.

It infers this by looking at the `core.logging.attempt.interaction_history` array's length.

kolibri/plugins/learn/assets/src/views/assessment-wrapper/index.vue
```diff
@@ -285,7 +289,9 @@ oriented data synchronization.
     },
     watch: { exerciseProgress: 'updateExerciseProgressMethod' },
     beforeDestroy() {
-      this.saveAttemptLogMasterLog(false);
+      if (this.currentInteractions > 0) {
+        this.saveAttemptLogMasterLog(false);
```
rtibbles, line 293: Good check!

bonidukic:
I can confirm this fixes the initial issue. 
It should also fix https://github.com/learningequality/kolibri/issues/3309 (at least in part).

rtibbles (approved):

Issue #3727, opened 2018-05-30
define conventions for kolibri commands


indirectlylit:
### Observed behavior

Let's decide on conventions for manage commands.

Two recent examples:

* https://github.com/learningequality/kolibri/pull/3594#issuecomment-388210709
* https://github.com/learningequality/kolibri/pull/3619#pullrequestreview-119193589

Other inconsistencies:

* shell
* plugin
* some older commands

### Expected behavior

commands are consistent

### User-facing consequences

not too much at the moment, but longer-term we want a clean command-line UI

indirectlylit:
In #3594 we decided to use the command 

```bash
kolibri manage content movedirectory
```

At the risk of bike-shedding that decision, I wanted to propose one convention update before committing to it:

> Multi-word commands should be `snake_case`.

In this case, the command would become `kolibri manage content move_directory`

Mashing multiple words together without any clear separator makes them hard to read, which is especially problematic because we won't be internationalizing the commands any time soon.

(In this case, the command scans to me like "moved irectory")

lyw07:
I agree that `move_directory` sounds more clear. I will make changes to the command name.

benjaoming:
> Multi-word commands should be snake_case.

Where is the quote from @indirectlylit ? IIRC, most django management commands are without underscores or dashes, like `makemigrations`.

rtibbles:
Yeah, it is definitely more djangoic to not underscore.

rtibbles:
I believe that the quote is from himself, @benjaoming.

indirectlylit:
Yes, it wasn't a quote - it was a proposal .. Note that I also updated the issue title from "kolibri manage commands" to "kolibri commands".

Agreed that we should not violate conventions if possible. However we should also give Kolibri users the best possible experience, irrespective of Django conventions they likely won't be aware of.

Perhaps we need to treat the `manage` subcommands as an advanced implementation detail and proxy to them. E.g.

```
kolibri content move-directory
```

could proxy to 

```
kolibri manage content movedirectory
```

This would allow all user-facing commands to have a short, clean form (like `start` and `language`).

Note that [GNU recommends dashes between words](https://www.gnu.org/software/libc/manual/html_node/Argument-Syntax.html):

> Option names are typically one to three words long, with hyphens to separate words. Users can abbreviate the option names as long as the abbreviations are unique.

benjaoming:
[...]

PR #4019, opened 2018-07-11
Execute vacuum command to shrink the wal size

### LINKED ISSUE #4006: sqlite3-wal got too big file size.
### Observed behavior
After I import the Khan Academy(English) channel and browse its downloadable contents the `sqlite3-wal` changed the file size to almost 1gb. This happened without downloading any contents from Khan Academy(English).

### Errors and logs
screen shot 2018-07-10 at 2 24 37 pm

### Context

 * Kolibri version: 0.11-0a4
 * Operating system: Windows 10
 * Browser: firefox 61.0

jredrejo:
### Summary
 * To remove fragmentation produced by importing/deleting channels
 * When Kolibri starts, let's execute a vacuum command: Created a Django management command for it.
 * When db is detached after importing, it executes a vacuum command (Not found a way to do it while ensuring the db is not in use, discarded by now)

### Reviewer guidance

The wal file size can be compared before and after applying this PR following the steps commented in #4006 

### References

 * resolve #4006 

### Contributor Checklist

### Reviewer Checklist


kolibri/core/content/utils/channel_import.py
```diff
@@ -469,13 +469,24 @@ def try_attaching_sqlite_database(self):
...
         if self.destination.engine.name == "sqlite":
             try:
                 self.destination.session.execute(text("DETACH 'sourcedb'".format(path=self.source_db_path)))
             except OperationalError:
                 # silently ignore if the database was already detached, as then we're good to go
                 pass
+            try:
+                # Importing produces a big fragmentation in the db,
+                # let's vacuum it.
+                self.destination.session.close()
```
rtibbles: Let's do this during the `end` method below as the final thing we do, rather than here. That way we have completely finished with our SQLAlchemy connection.

indirectlylit:
This looks good to me. Seems like for long-running servers we still will have an issue, but this will help in many cases as a work-around

kolibri/core/sqlite/management/commands/vacuumsqlite.py
```diff
@@ -0,0 +1,36 @@
...
+
+from django.core.management.base import BaseCommand
+from django.db import connections
+from django.db import DEFAULT_DB_ALIAS
+
+logging = logger.getLogger(__name__)
+
+
+class Command(BaseCommand):
+    help = 'Vacuum project sqlite database to optimize it and reduce wal file size'
```
indirectlylit: > Vacuum Kolibri's SQLite database to optimize it and reduce the .wal file size

kolibri/core/sqlite/management/commands/vacuumsqlite.py
```diff
@@ -0,0 +1,36 @@
...
+
+    def add_arguments(self, parser):
+        parser.add_argument(
+            '--database', action='store', dest='database', default=DEFAULT_DB_ALIAS,
+            help='Nominates a database to vacuum. Defaults to the "default" database.',
```
indirectlylit: > Specifies the database to vacuum. Defaults to the "default" database.

kolibri/core/sqlite/management/commands/vacuumsqlite.py
```diff
@@ -0,0 +1,36 @@
...
+                cursor = connection.cursor()
+                cursor.execute('VACUUM;')
+                connection.close()
+            except Exception as e:
+                new_msg = (
+                    "Vacuum of database %s couldn't be executed. Possible reasons:\n"
+                    "  * There is an open transaction in the db.\n"
+                    "  * There are one or more active SQL statements.\n"
+                    "  * The SQL was invalid.\n"
+                    "The full error: %s") % (connection.settings_dict['NAME'], e)
```
indirectlylit: According to [the docs](https://docs.python.org/2/library/stdtypes.html#str.format), using `.format()`

> is the new standard in Python 3, and should be preferred to the `%` formatting

indirectlylit (commented):
minor notes

indirectlylit (commented):
As discussed earlier, let's update this to run at regular intervals (e.g. every 6, 12, or 24 hours) using the same iceqube-based scheduling mechanism as the ping-back uses.

kolibri/utils/server.py
```diff
@@ -100,6 +103,19 @@ def run(self):
...
+
+    @classmethod
+    def start_command(cls):
+        thread = cls()
+        thread.daemon = True
+        thread.start()
+
+    def run(self):
+        # Try to do the vacuum every 3 hours
+        call_command("vacuumsqlite", interval=180)
```
indirectlylit: In order to ensure that this gets thoroughly tested internally, we could set the interval to something much more aggressive like 10 minutes. Then we'd simply plan on increasing the interval before publicly releasing. Thoughts?

benjaoming:
> As discussed earlier, let's update this to run at regular intervals (e.g. every 6, 12, or 24 hours) using the same iceqube-based scheduling mechanism as the ping-back uses.

@indirectlylit I don't feel comfortable about that :)

We already have unsolved "malformed db" issues. Doing parallel raw database-level operations is IMO something we need to be careful about. Iceqube isn't IMO a library that we built with the intend to do system administration tasks.

I think @jredrejo 's idea to put this in Cron jobs is okay, but I prefer a user controlled operation to begin with. Like what we have/had with compacting mailboxes in Thunderbird and Outlook :)

kolibri/core/sqlite/management/commands/vacuumsqlite.py
```diff
@@ -0,0 +1,49 @@
+import logging as logger
```
benjaoming: Ahaha, first line and I have an objection - sorry :) Why alias this and then assign `logging` further down? I would prefer if we could have a strict, consistent logging pattern. So just:

```
import logging
```

kolibri/core/sqlite/management/commands/vacuumsqlite.py
```diff
@@ -0,0 +1,49 @@
...
+
+    def add_arguments(self, parser):
+        parser.add_argument(
+            '--database', action='store', dest='database', default=DEFAULT_DB_ALIAS,
+            help='Specifies the database to vacuum. Defaults to the "default" database.',
+        )
+        parser.add_argument(
+            '--interval', action='store', dest='interval', default=0, type=int,
+            help='Specifies the interval (in minutes) to run the process continuosly. If 0, no repetition will happen',
+        )
```
benjaoming: Can we remove the interval+sleep thing?

While I think that it will be okay when running `kolibri stop`, I'm unsure if this might end up blocking other operations, leave unclosed db connections, PID files etc. Certainly, if integrating with `iceqube`, it would require a well-defined behavior for "eternal" threads and how they are shut down. For some purposes, `iceqube` should wait for a thread to finish nicely, and for others, it shouldn't.

The easiest solution is to 1) not run a loop but just execute `VACUUM` once and 2) avoid integrating system maintenance in `iceqube`, because it's not intended for this.

kolibri/core/sqlite/management/commands/vacuumsqlite.py
```diff
@@ -0,0 +1,49 @@
...
+            connection.close()
+        except Exception as e:
+            new_msg = (
+                "Vacuum of database {db_name} couldn't be executed. Possible reasons:\n"
+                "  * There is an open transaction in the db.\n"
+                "  * There are one or more active SQL statements.\n"
+                "The full error: {error_msg}").format(db_name=connection.settings_dict['NAME'], error_msg=e)
+            logging.error(new_msg)
+        else:
+            logging.info("Sqlite database Vacuum finished.")
```
benjaoming: Maybe consistently write upper-case `VACUUM` to indicate that it's a command for SQLite? :)

kolibri/utils/server.py
```diff
@@ -65,6 +65,9 @@ def start(port=8080):
     # start the pingback thread
     PingbackThread.start_command()
 
+    # Do a db vacuum periodically
+    VacuumThread.start_command()
```
benjaoming, line 69: Unconfigurable thread behavior breaks the intention of having this under `cron` control.

I'm a bit confused at this point, because there are 3 approaches:

1) Iceqube (which IMO isn't perfect)
2) Cron jobs (of course .deb pkg only)
3) A basic Python thread

kolibri/utils/server.py
benjaoming: That's a great idea @indirectlylit !

kolibri/deployment/default/settings/base.py
```diff
@@ -62,6 +62,7 @@
     'kolibri.core.auth.apps.KolibriAuthConfig',
     'kolibri.core.content',
     'kolibri.core.logger',
+    'kolibri.core.sqlite',
```
benjaoming: Let's avoid another application. Can this live in `kolibri.core.deviceadmin` ?

kolibri/core/sqlite/management/commands/vacuumsqlite.py
```diff
@@ -0,0 +1,49 @@
...
+    def handle(self, *args, **options):
+        database = options['database']
+        connection = connections[database]
+        interval = options['interval']
+        if connection.vendor == "sqlite":
+            while True:
+                self.perform_vacuum(connection)
+                if not interval:
+                    break
+                logging.info("Next Vacuum in {interval} minutes.".format(interval=interval))
```
benjaoming: `debug` level this.

benjaoming (changes requested):
Couple of small things - in general, I think we need a decision on which direction we want to take system administrative tasks.

kolibri/utils/server.py
jredrejo: @benjaoming  you said in a previous comment you are against of running vacuum periodically, or maybe I misunderstood you?

kolibri/core/sqlite/management/commands/vacuumsqlite.py
jredrejo: Initially I coded it as you say, but later I changed it to this way for consistency reasons.
Kolibri uses always `logging.info...` not `logger.info`

kolibri/core/sqlite/management/commands/vacuumsqlite.py
jredrejo: I don't understand your objections. Also I did initially only on start and other people requested to have it run periodically to ensure long-running servers are optimized too. If, as Jamie said in slack, our users don't run Kolibri as a server for long periods of time, this does not matter. Anyway, I think this is a discussion that should be held in Slack between the intersted people.

kolibri/core/sqlite/management/commands/vacuumsqlite.py
jredrejo: why? other similar commands in Kolibri use info for these things (look a the ping command)

kolibri/core/sqlite/management/commands/vacuumsqlite.py
jredrejo: Done

kolibri/deployment/default/settings/base.py
jredrejo: I wrote  it in sqlite after asking for Richard's advice. Deviceadmin also can make sense, could you discuss it with @indirectlylit & @rtibbles  to avoid having to move it backwards and forwards again?

kolibri/utils/server.py
jredrejo: We don't have the intention of running it under cron as it's not possible in Windows/pex installations. It's done using point 3 of your list . It seems to be the only possible universal solution.

kolibri/core/sqlite/management/commands/vacuumsqlite.py
benjaoming: Huh? Oh, do we have a rogue pattern somewhere? Could you ref that? Sounds like a cleanup issue then...

kolibri/core/sqlite/management/commands/vacuumsqlite.py
benjaoming: Sorry, I have been confused by the many inputs on Slack. At first, I thought it was a Cron job, then I saw @indirectlylit 's comment about making it an Iceqube job, and now I see that it's a daemonized thread.

kolibri/core/sqlite/management/commands/vacuumsqlite.py
benjaoming: I am just trying to limit our default level user logs. If you `debug` level it, it will be visible with `kolibri <command> --debug`

kolibri/deployment/default/settings/base.py
benjaoming: If we don't manage to divide our various management commands into meaningful application spaces, we'll end up with way too many applications, and we'll have a pattern where we invent an application for simple tasks instead of encouraging reuse etc. between various utility functions. A Django application is best defined by also having models etc -- but since we did have a couple of things that didn't require models, we have bundled them into `deviceadmin`. IMHO of course :)

kolibri/utils/server.py
benjaoming: No, see my previous comment - I just don't get where the job should be controlled from. If we are doing this in its own thread, let's stick to that and make the behavior configurable and possible to disable.

benjaoming: For instance, it shouldn't run on Postgres.

benjaoming: Ah, okay, the thread just exits in other db environments - but for instance, Postgres does actually have a VACUUM command :)

https://www.postgresql.org/docs/9.5/static/sql-vacuum.html

kolibri/core/sqlite/management/commands/vacuumsqlite.py
benjaoming: Writing it lowercase wasn't what I meant, but okay, let's stick to the more important stuff, don't wanna nitpick :)

kolibri/deployment/default/settings/base.py
indirectlylit: Doesn't matter one way or the other to me. (IMO we should not be exposing Django `manage` management commands directly to end users anyway, but that's a completely different conversation for #3727)

Putting it in `deviceadmin` seems reasonable as a way to prevent app bloat if that's a concern

benjaoming: > we should not be exposing Django manage management commands directly to end users anyway

I'm :+1: with this - just that I find it fine to keep such functions wrapped as convenient management commands for advanced users -- then later on, we can build UIs that invoke them.

kolibri/core/sqlite/management/commands/vacuumsqlite.py
jredrejo: @benjaoming  everywhere, just do a grep for it in the code. All kolibri code is that way.

kolibri/core/sqlite/management/commands/vacuumsqlite.py
jredrejo: @benjaoming yes, I know it and I don't dislike your suggestion. My only objection is the same I had for the logging part: I'd prefer to keep the consistency with other similar cases.

kolibri/utils/server.py
jredrejo: Yes, I know it (believe me, do never run that command in a postgresql database in production, setting autovacuum is much better for postgresql), but if you check the code you will see the vacuumsqlite command checks if the db is sqlite, if not, it just does nothing, so it does not run on Postgresql

kolibri/deployment/default/settings/base.py
jredrejo: Perfect, movement done.

kolibri/utils/server.py
jredrejo: @indirectlylit what do you think about the way to alter the interval for testing? should we hardcode it or create a new parameter in options.ini?

kolibri/core/sqlite/management/commands/vacuumsqlite.py
benjaoming: I wouldn't say "all" of it :)

Far from :) `ag "import logging$" --py kolibri` shows a majority of "normal" patterns.

Opened up https://github.com/learningequality/kolibri/issues/4057

kolibri/utils/server.py
indirectlylit: the easiest thing to do is just hard-code it, and open up an issue targeted to 0.11.0 reminding us to change it back before release :)

benjaoming (approved):
Thanks for addressing my input @jredrejo !

This seems fine, but configurability should be added, such that users can control whether and when to call the `VACUUM` command, which btw. will also be relevant for Postgres (in which time, we'll have to research its behavior).

jredrejo:
@benjaoming Executing vacuum in Postgres takes a long, long time. Autovacuum in postgresql does not delay the server normal work so, setting it up in the server config is the best thing to do.

However, in my experience, autovacuum is not enough in postgresql. After some time the db grows too much and becomes slow, but in this case vacuum must be executed periodically (every week in servers with a high load, around 1 million inserts by day, or monthly with fewer inserts). 

In my opinion, we should cross that bridge when we come to it. I don't think we could do it in a standard method. Depending on the server, different intervals and execution times (nightly if possible) should be considered.

Also, we could think of executing it after kolibri stops, but bearing in mind that vacuum might take easily 30 minutes or more, depending on the size of the database.

indirectlylit (approved):
alrighty - looks like we're ready to merge this in! let's keep an eye out for issues :D

benjaoming:
> In my opinion, we should cross that bridge when we come to it.

Totally agree!

PR #4036, opened 2018-07-17
[WIP] Hashi: Html5 app bridge library

kollivier:
### Summary
Adds a bridge library that enables HTML5 apps to load and save state, as well as to communicate assessment and other data to Kolibri for tracking progress and performance.

### Reviewer guidance

Test cases are not in place yet, but will be soon. For now this PR is mostly for code review as the feature work progresses.

### References

…

----

### Contributor Checklist

### Reviewer Checklist


kolibri/core/assets/src/state/actions.js
```diff
@@ -49,14 +49,22 @@ const timeThreshold = 60; // Update logs if 60 seconds have passed since last up
  */
 
 function _contentSummaryLoggingState(data) {
+  // In older versions of Kolibri, extra_fields was being
+  // set to a string rep of an object rather than set to
+  // an actual object. Hence here we need to ensure that
+  // if we do get a string, we convert it to an object.
+  var extraFields = data.extra_fields;
```
rtibbles: Should use `let` for ES6 reassignable variable.

kolibri/core/assets/src/mixins/contentRenderer.js
```diff
@@ -93,6 +93,10 @@ export default {
       default: () => defaultLanguage,
       validator: languageValidator,
     },
+    extraFields: {
```
rtibbles, line 96: :+1:

kolibri/core/assets/src/state/store.js
```diff
@@ -108,6 +108,10 @@ export const coreMutations = {
   SET_LOGGING_COMPLETION_TIME(state, time) {
     state.core.logging.summary.completion_timestamp = time;
   },
+  SET_LOGGING_CONTENT_STATE(state, contentState) {
+    // TODO: Consider whether we want to save these to the session log as well.
```
rtibbles: Probably not, as we would never reload state from this, so this would be entirely for analytics purposes.

packages/hashi/package.json
```diff
@@ -0,0 +1,20 @@
...
+  "name": "hashi",
+  "version": "0.1.0",
+  "description": "A library for integrating HTML5 apps and games with Kolibri.",
+  "main": "hashi.js",
+  "scripts": {
+    "test": "npm-run-all -p --race test:server cypress:open",
+    "cypress:open": "cypress open",
+    "test:server": "http-server ./test -p 8085"
+  },
+  "author": "Kevin Ollivier",
```
rtibbles: We normally label our packages as Learning Equality author.

packages/hashi/src/storage.js
```diff
@@ -0,0 +1,124 @@
...
+        // keys have list ordering, so we use a list to ensure order is not changed.
+        // Per the spec, key ordering is defined by user agent, so we don't have to
+        // model any specific sorting behavior, just be consistent.
+        this.sessionKeys = [];
+        this.initMessageSent = false;
+        this.dataReceived = false;
+    }
+
+    _sendMessage(name, data) {
+        var message = {
```
rtibbles: You're using an ES6 class, so we should use ES6 variable declarations too!

packages/hashi/src/storage.js
```diff
@@ -0,0 +1,124 @@
...
+        };
+        window.parent.postMessage(JSON.stringify(message), '*');
+    }
+
+    initData() {
+      var timeout = 5000;
+      var pinkyPromise = null;
+      if (!this.initMessageSent) {
+        var self = this;
+        pinkyPromise = new Promise(function(resolve, reject) {
```
rtibbles: Whimsical, certainly - but perhaps not terrible descriptive?

rtibbles: If you use arrow functions `(resolve, reject) => {}` it will bind `this` to `this` in the scope in which it is defined, and you can dispense with the `self` declaration.

rtibbles (commented):
A few comments to be going on with.

kollivier:
Okay, @rtibbles the retro Javascript conventions should be gone now! Also am now webpacking the library so that it won't require an ES6-compatible browser to run. Still need to write the Gherkin tests for this, but it should be in good enough shape for ePub state saving hacking.

kolibri/core/assets/src/state/modules/core/actions.js
```diff
@@ -41,14 +41,20 @@ const timeThreshold = 120; // Update logs if 120 seconds have passed since last
  */
 
 function _contentSummaryLoggingState(data) {
+  var extraFields = data.extraFields;
```
christianmemije: Lets use `let` instead of `var`

kolibri/core/assets/src/state/modules/core/actions.js
```diff
@@ -555,6 +561,22 @@ export function updateTimeSpent(store, forceSave = false) {
   }
 }
 
+/**
+ * Update the content state in the summary log
+ * To be called periodically by interval timer
+ * Must be called after initContentSession
+ * @param {boolean} forceSave
+ */
+export function updateContentState(store, contentState, forceSave = false) {
```
christianmemije, line 570: What are the use cases for using or not using force save?

kolibri/core/assets/src/views/ContentRenderer/index.vue
```diff
@@ -204,6 +210,9 @@
         this.$emit('updateProgress', ...args);
         heartbeat.setActive();
       },
+      updateContentState(...args) {
+        this.$emit('updateContentState', ...args);
+      },
```
christianmemije, line 215: Add `heartbeat.setActive()` here too @rtibbles ?

kolibri/plugins/html5_app_renderer/assets/src/views/Html5AppRendererIndex.vue
```diff
@@ -69,6 +71,31 @@
       }
       this.$emit('stopTracking');
     },
+    methods: {
+      iframeMessageReceived(event) {
+        if (!event) {
+          return;
+        }
+        var message = JSON.parse(event.data);
```
christianmemije: `const`

kolibri/plugins/html5_app_renderer/assets/src/views/Html5AppRendererIndex.vue
```diff
@@ -69,6 +71,31 @@
...
+        if (message.action === 'stateUpdated') {
```
christianmemije, line 81: Might want to add these constants somewhere so we are not directly comparing strings.

kolibri/plugins/html5_app_renderer/assets/src/views/Html5AppRendererIndex.vue
```diff
@@ -69,6 +71,31 @@
...
+          this.$emit('updateContentState', message.params);
+        } else if (message.action === 'hashiInitialized') {
+          var iframe = document.getElementsByTagName('iframe')[0].contentWindow;
```
christianmemije: Let's scope this by instead adding a ref to the iframe in the template `ref="iframe"` and then we can do
```js
const iframe = this.$refs.iframe.contentWindow;
```
https://vuejs.org/v2/api/#ref

kolibri/plugins/html5_app_renderer/assets/src/views/Html5AppRendererIndex.vue
```diff
@@ -69,6 +71,31 @@
...
+
+          // On guest access, the user will be signed out, so just return
+          // an empty key value store in that case.
+          var contentState = {};
```
christianmemije: Let

kolibri/plugins/html5_app_renderer/assets/src/views/Html5AppRendererIndex.vue
```diff
@@ -69,6 +71,31 @@
...
+          if (this.extraFields && this.extraFields.contentState) {
+            contentState = this.extraFields.contentState;
+          }
+          var data = {
```
christianmemije: const

kolibri/plugins/html5_app_renderer/assets/src/views/index.vue
```diff
@@ -0,0 +1,112 @@
+<template>
```
christianmemije: Woops, this file should be removed. It was renamed to Html5AppRendererIndex.vue

package.json
```diff
@@ -93,6 +93,7 @@
     "jest-serializer-vue": "^2.0.2",
     "json-loader": "^0.5.4",
     "launch-editor-middleware": "^2.2.1",
+    "lerna": "^2.11.0",
```
christianmemije, line 96: Where is this used?

kollivier: It allows us to create separate npm packages for Hashi, since HTML5 apps supporting Kolibri will need to depend on it.

packages/hashi/src/storage.js
```diff
@@ -0,0 +1,123 @@
...
+    initData() {
+      const timeout = 5000;
+      let initDataPromise = null;
+      if (!this.initMessageSent) {
+        initDataPromise = new Promise((resolve, reject) => {
+            function parentMessageReceived(event) {
+              if (this.timer) {
+                clearTimeout(this.timer);
+              }
+              var message = JSON.parse(event.data);
```
christianmemije: const

kolibri/core/assets/src/state/modules/core/actions.js
kollivier: It's basically used as protection from overwhelming the database with a bunch of rapid changes to the content state, similar to the logic in updateProgress. (I basically cribbed that logic.) forceSave is to be used when you need an immediate update and the caller has been careful to ensure messages aren't coming from multiple points, which is why it defaults to false. 

Note that due to the latency of the HTML5 app <-> Kolibri message passing, Hashi itself has an app-local copy of the localStorage that exists during the app lifetime, so writes are not usually time-critical. The situation may be different for your ePub reader, and since you don't have to deal with things like postMessage, using forceSave may be a better choice there.

kolibri/plugins/html5_app_renderer/assets/src/views/index.vue
kollivier: Ugh, thought I checked for this, sorry!

kolibri/plugins/html5_app_renderer/assets/src/views/Html5AppRendererIndex.vue
kollivier: Ah, nice! I think @rtibbles was trying to explain this to me during a previous review but I got caught up in other changes and forgot to double-back to this one.

kollivier:
Wanted to say thanks @rtibbles and @christianmemije for being patient and taking the time to turn this PR into a Vue and ES6 tutorial! ;) I learned a few new tricks and am on the way to unlearning some of my bad "old man" (classic? original?) Javascript habits. I just pushed the remaining ES6 / Vue updates.

Now I'm going to look at how I can edit the history to split out the Hashi parts of this change.

kolibri/plugins/html5_app_renderer/assets/src/views/Html5AppRendererIndex.vue
kollivier: Is there a place you'd recommend to put these sorts of constants?

rtibbles:
> I learned a few new tricks and am on the way to unlearning some of my bad "old man" (classic? original?) Javascript habits

I think the preferred term is `vintage`!

kolibri/plugins/html5_app_renderer/assets/src/views/Html5AppRendererIndex.vue
christianmemije: I would save in a constants file inside the hashi package. Then we can reference them from here using `import { MESSAGES } from 'hashi/src/constants';`.

kollivier:
This PR was split into two parts, one for the renderer state saving and the other for the actual Hashi library. This PR is thus no longer valid, so closing it out.

PR #4037, opened 2018-07-17
Optimize cherry pick

rtibbles:
### Summary
Cherry picking optimizations from 0.10 and 0.11 to improve performance.

Did a quick manual test, app still seems to work.

Note - I did this with a 0.10 db, and no issues, so it is possible we could ask Nalanda to test with 0.10 also.

----

### Contributor Checklist

### Reviewer Checklist


kolibri/deployment/default/settings/base.py
```diff
@@ -343,6 +342,19 @@
...
+# Session configuration
+
+SESSION_ENGINE = "django.contrib.sessions.backends.file"
+
+SESSION_FILE_PATH = os.path.join(conf.KOLIBRI_HOME, "sessions")
+
+if not os.path.exists(SESSION_FILE_PATH):
+    if not os.path.exists(conf.KOLIBRI_HOME):
+        raise RuntimeError("The KOLIBRI_HOME dir does not exist")
+    os.mkdir(SESSION_FILE_PATH)
```
jamalex, line 354: Hmm, why do we not just use `os.makedirs` here, instead of relying on KOLIBRI_HOME having already been created? Seems risky for fresh installs?

kolibri/utils/server.py
```diff
@@ -150,7 +150,10 @@ def run_server(port):
     # Configure the server
     server.socket_host = LISTEN_ADDRESS
     server.socket_port = port
-    server.thread_pool = 30
+    server.thread_pool = 10
+    server.socket_timeout = 10
+    server.accepted_queue_size = 30
+    server.accepted_queue_timeout = 0.1
```
jamalex, line 156: So do these qualify as cherry py cherry picks, then?

jamalex (approved):

jamalex:
>Note - I did this with a 0.10 db, and no issues, so it is possible we could ask Nalanda to test with 0.10 also.

That would migrate their DB to 0.10, though, right? So it's not just testing, it would permanently update the DB.

rtibbles:
Yes, that was my point - they could safely be migrated to a 0.10 db, and I don't think it would have any ill effects.

kolibri/deployment/default/settings/base.py
rtibbles: Because if you have got this far and KOLIBRI_HOME does not exist then something has already gone wrong.

rtibbles: (this gets checked during set_env much earlier in the load cycle).

kolibri/utils/server.py
jredrejo: yep, that was the first intention, later it was replaced by the options.ini settings, defaulting to these same values.

benjaoming:
Merging 0.9.x to 0.10.x, I'm going to deliberately leave out this part:

```
        # Clear old sessions up
        call_command("clearsessions")
```

benjaoming:
This PR has made it slightly difficult to merge 0.9.x into 0.10.x

PR #4048, opened 2018-07-19
API resource refactor

### LINKED ISSUE #3072: Using 'id' consistently in API responses
`id` and `pk` are used seemingly interchangeably in responses from the REST API , which is kind of confusing.

There seems to be a preference in the JS code for `id`, as when a response uses `pk`, it is often renamed to `id`. `pk` is also hints at the underlying database, while `id` is a little more abstract.

rtibbles:
### Summary
Attempts to simplify the API Resource machinery by getting rid of the use of class inheritance in favour of initialization parameters.
Fixes a few reported issues with the API Resource layer:
* No longer relies solely on the existence of an id for whether a model exists on the server already or not, a new flag is now set on newly created models, so that they will be POSTed rather than PATCHed.
* Instead of resolving the attributes of the model directly, instead resolves a (cheaply computed) deepClone of the attributes to prevent the model from becoming reactive if set onto vuex state.
* Moves getParams for models into the getModel method and uses them to specify caching keys, meaning that models with the same id but different getParams will be cached separately.
* Due to the above change, removed the ability to pass in extra getParams to the `fetch` method.

This PR also adds schema checking in dev mode, to give warnings to developers as they are adding new API Resources regarding the available API Resources on the server, and the methods those resources make available.

This PR also adds a range of helper functions to the Resource class to allow methods to be very simply written for accessing additional endpoints on a DRF viewset. See the TaskResource and ContentNodeSlimResource for details.

This PR further migrates all the content node recommendation filters to be separate endpoints on the content node slim endpoint, utilizing the utility functions above.

Finally, it changes all our returned ids to `id` not `pk`.

### Reviewer guidance
I have manually tested and not found anything broken yet, but it is possible I have missed things.

I would like some feedback on what I have done, and whether it makes sense. Some of the work could probably use some additional automated testing, but most of the refactor is covered by existing tests.

Finally, one thing I have not done is to alter the redundant `getCollection().fetch()` idiom, which could be more succinctly done as `fetchCollection()`. The only reason I have not done this is because it is a major change to the API, but should be relatively trivial to implement. I will happily do so here based on feedback.

### References
Fixes #3072 

…

----

### Contributor Checklist

### Reviewer Checklist


kolibri/core/assets/src/api-resources/channel.js
```diff
@@ -5,10 +5,8 @@ import { Resource } from '../api-resource';
  * ChannelResource.getModel(channel_id).delete()
  *
  * @example Only get the channels that are "available" (i.e. with resources on device)
- * ChannelResource.getCollection().fetch({ available: true })
+ * ChannelResource.getCollection({ available: true }).fetch()
```
indirectlylit, line 8: yes!

kolibri/core/assets/src/api-resources/contentNode.js
```diff
@@ -1,125 +1,17 @@
...
-    }
-    return promise;
-  }
-
-  getCopiesCount(getParams = {}) {
-    const collection = this.createCollection({}, getParams, []);
-    collection.url = (...args) => this.urls[`${this.name}-copies-count`](...args);
-    return collection;
-  }
-
```
indirectlylit, line 81: yes!

kolibri/core/assets/src/api-resource.js
```diff
@@ -552,32 +574,35 @@ export class Resource {
     );
   }
 
+  cacheName(endpointName) {
+    return endpointName ? `endpoint-${endpointName}` : 'default';
+  }
+
+  getCache(type, endpointName) {
```
indirectlylit: A lot of the methods in these classes look internal - could we prefix them with underscores to make clear that they're not part of the public API?

kolibri/core/assets/src/api-resource.js
```diff
@@ -665,169 +669,184 @@ export class Resource {
...
+    return this.getCollection(getParams, detailName, id).fetch();
   }
 
-  unCacheCollection(resourceIds = {}, getParams = {}) {
-    if (!this.hasResourceIds) {
-      if (Object.keys(resourceIds).length && Object.keys(getParams).length) {
-        throw TypeError(
-          `resourceIds and getParams passed to getCollection method of ${this.name} ` +
-            'resource, which does not use resourceIds, only pass getParams for this resource'
+  fetchListCollection(listName, getParams = {}) {
```
indirectlylit: `fetchListModel`, `fetchDetailCollection`, `fetchListCollection` are similarly named.. could use some docs

kolibri/core/assets/src/api-resources/index.js
```diff
@@ -1,96 +1,31 @@
...
+export { default as ExamLogResource } from './examLog';
+export { default as ExamAttemptLogResource } from './examAttemptLog';
+export { default as FacilityDatasetResource } from './facilityDataset';
+export { default as UserProgressResource } from './userProgress';
+export { default as ContentNodeProgressResource } from './contentNodeProgress';
+export { default as DeviceProvisionResource } from './deviceProvision';
+export { default as DevicePermissionsResource } from './devicePermissions';
+export { default as RemoteChannelResource } from './remoteChannel';
+export { default as LessonResource } from './lesson';
+export { default as AttemptLogResource } from './attemptLog';
```
indirectlylit, line 31: weird :)

kolibri/core/content/api.py
```diff
@@ -525,12 +384,150 @@ def ancestors(self, request, **kwargs):
...
+            for node in resume:
+                if node.content_id not in content_ids:
+                    deduped_list.append(node)
+                    content_ids.add(node.content_id)
+
+            queryset = resume.filter(id__in=[node.id for node in deduped_list])
+
+        serializer = self.get_serializer(queryset, many=True)
+        return Response(serializer.data)
+
```
indirectlylit, line 530: nice new routes to replace the filters

indirectlylit (commented):
Looks like some productive and positive changes!

* perhaps we could split `api-resource.js` into three files and call the classes and files `APIResourceModel`, `APIResourceCollection`, and `APIResource` or something more explicit?
* I'm a bit confused about the `set` methods. It seems like it's like `save`, but doesn't actually affect server state? (I know this pre-existed the refactor)

kolibri/core/assets/src/api-resource.js
rtibbles: Definitely

indirectlylit (approved):
approved, pending a few updates as described above

kolibri/plugins/coach/assets/src/state/actions/reports.js
```diff
@@ -401,19 +400,21 @@ export function setReportSorting(store, { sortColumn, sortOrder }) {
...
-        collection_id: classId,
-        last_active_time: threshold,
-      }).fetch();
+      const recentReportsPromise = RecentReportResource.fetchCollection({
+        getParams: {
+          channel_id: channelId,
+          content_node_id: channelData.root,
+          collection_kind: UserScopes.CLASSROOM,
+          collection_id: classId,
+          last_active_time: threshold,
```
jonboiser, line 415: @rtibbles I was getting validation errors sending the long-form date string to the server as a filter param. I made the errors go away it by converting to IsoString. Is that the right fix?

```diff
-          last_active_time: threshold,
+          last_active_time: threshold.toISOString(),
```

rtibbles: Seems right, before this refactor, the threshold was simply being ignored, so no error was being produced.

Issue #4058, opened 2018-07-20
module 'lib' has no attribute 'Cryptography_HAS_SSL_ST'


lyw07:
### Observed behavior

This is an error reported by our user:
Getting a cryptography/OpenSSL related error when running any kolibri command on Odroid c2, with Ubuntu 16.04. 
The error traceback is in the `Errors and logs` section.

### Expected behavior

Users should be able to run any kolibri command successfully.

### User-facing consequences

Users cannot use kolibri

### Errors and logs

```
Traceback (most recent call last):

  File "/usr/bin/kolibri", line 9, in <module>

    load_entry_point('kolibri==0.10.0', 'console_scripts', 'kolibri')()

  File "/usr/lib/python3/dist-packages/pkg_resources/__init__.py", line 542, in load_entry_point

    return get_distribution(dist).load_entry_point(group, name)

  File "/usr/lib/python3/dist-packages/pkg_resources/__init__.py", line 2569, in load_entry_point

    return ep.load()

  File "/usr/lib/python3/dist-packages/pkg_resources/__init__.py", line 2229, in load

    return self.resolve()

  File "/usr/lib/python3/dist-packages/pkg_resources/__init__.py", line 2235, in resolve

    module = __import__(self.module_name, fromlist=['__name__'], level=0)

  File "/usr/lib/python3/dist-packages/kolibri/utils/cli.py", line 24, in <module>

    from . import server  # noqa

  File "/usr/lib/python3/dist-packages/kolibri/utils/server.py", line 8, in <module>

    import requests

  File "/usr/lib/python3/dist-packages/kolibri/dist/requests/__init__.py", line 84, in <module>

    from urllib3.contrib import pyopenssl

  File "/usr/lib/python3/dist-packages/kolibri/dist/urllib3/contrib/pyopenssl.py", line 46, in <module>

    import OpenSSL.SSL

  File "/usr/lib/python3/dist-packages/kolibri/dist/OpenSSL/__init__.py", line 8, in <module>

    from OpenSSL import crypto, SSL

  File "/usr/lib/python3/dist-packages/kolibri/dist/OpenSSL/SSL.py", line 115, in <module>

    if _lib.Cryptography_HAS_SSL_ST:

AttributeError: module 'lib' has no attribute 'Cryptography_HAS_SSL_ST'
```

### Steps to reproduce

The following steps can reproduce this error from a working kolibri instance
1. Have a fresh installation of kolibri 0.10 on any machine
[...]

Issue #2975, opened 2017-12-16
Travis' new build stages & build matrix optimizations


benjaoming:
High-level issues:

 * [ ] Tox fetches dependencies both into its virtualenv and in `kolibri/dist` via `make staticdeps`
 * [ ] Assets and dependencies are applied to build environments that they might be irrelevant to, for instance building docs? Linting? Conditional tests, like build tests
 * [ ] Conditional tests aren't conditional enough. The whole build environment should be exited as quickly as possible if relevant files aren't changed. For instance, don't build docs, when nothing in the docs has been edited. Don't check licenses of dependencies if dependencies haven't changed etc.

I would propose that we stage our Travis builds to shorten build times, this is inspired by CherryPy's setup.

@ralphiee22 has also mentioned the new test tags introduced in Django 1.10 so we can chunk our tests into a slow and fast set.

1. Linting
1. Building assets, unifying this as 1 build for all tests
1. Upload assets to S3
   https://docs.travis-ci.com/user/build-stages/#Sharing-files-between-jobs-via-S3
1. Basic tests in 1 Python version
1. Expanded tests to all tox envs
1. Conditional tests

See: https://blog.travis-ci.com/2017-09-12-build-stages-order-and-conditions

Related: #3321 

![image](https://user-images.githubusercontent.com/374612/34075348-9fc152f2-e2c4-11e7-9fb2-838928dbd29a.png)

rtibbles:
Sounds good - at the same time, is it still important to have our build tests in Travis? Buildkite should be performing this test for us, effectively.

rtibbles:
We should also fix this issue: https://github.com/learningequality/kolibri/issues/551 which should increase confidence across Python versions.

benjaoming:
I'm a big fan of Travis for our test matrix stuff and now these new stages :)

I don't think Buildkite is mature for the testing and linting yet for a couple of reasons. But it seems great for building installers and deploying stuff :)

rtibbles:
I didn't want buildkite to test anything or lint anything - simply that we do a build every time on buildkite, so our tests specifically for the Python build pipeline are duplicative.

benjaoming:
Oooooh "build tests" sorry, kind of ambiguous :) But yes, those should be removed. Or at least used for the actual tests.

rtibbles:
I guess another question here is: if you don't edit any python files in a PR, should we still run the Python tests, and vice versa, if you don't edit any frontend files.

PR #4059, opened 2018-07-20
Fix the error when OpenSSL version is incompatible with the system cryptography version

lyw07:
### Summary

This PR is to fix the error when we don't have bundled cryptography available for the platform, and a system cryptography exists in the machine but is not compatible with kolibri's bundled OpenSSL.
The error is described in #4058 
Basically, the change is that the paths of cryptography and OpenSSL are only added to PYTHONPATH when the path of cryptography exists. That means, when kolibri don't have cryptography that suits the user's platform available, the OpenSSL module bundled inside kolibri will not be imported. Whether users have an old version of cryptography in the system or not will not cause the error anymore.

### Reviewer guidance

The steps to reproduce the error are listed in the issue.
To test the PR, install cryptography 1.2.3 when the platform doesn't have available bundled cryptography in kolibri, and see if the error still occurs.

### References

#4058 

----

### Contributor Checklist

### Reviewer Checklist

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -62,13 +62,15 @@ matrix:
     - python: "2.7"
       env:
         - TOX_ENV=npmbuild
-    - python: "3.4"
+    - python: "2.7"
       env:
-        - TOX_ENV=py3.4
-        - RUN_WITH_CEXT="y"
+        - TOX_ENV=nocext
       script:
-       - pip install -r requirements/build.txt
-       - make staticdeps-cext
+       - tox -e $TOX_ENV
+    - python: "2.7"
+      env:
+        - TOX_ENV=cext
+      script:
        - tox -e $TOX_ENV
 
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,59 +23,54 @@ matrix:
     - python: "2.7"
       env:
         - TOX_ENV=pythonbuild2.7
-        - TRAVIS_NODE_VERSION="6"
     - python: "3.4"
       env:
         - TOX_ENV=pythonbuild3.4
-        - TRAVIS_NODE_VERSION="6"
     - python: "3.5"
       env:
         - TOX_ENV=pythonbuild3.5
-        - TRAVIS_NODE_VERSION="6"
     - python: "3.6"
       env:
         - TOX_ENV=pythonbuild3.6
-        - TRAVIS_NODE_VERSION="6"
     - python: "2.7"
       env:
         - TOX_ENV=pythonlint
-        - TRAVIS_NODE_VERSION="6"
     - python: "2.7"
       env:
         - TOX_ENV=licenses
-        - TRAVIS_NODE_VERSION="6"
     - python: "2.7"
       env:
         - TOX_ENV=py2.7
-        - TRAVIS_NODE_VERSION="6"
     - python: "3.5"
       env:
         - TOX_ENV=docs
-        - TRAVIS_NODE_VERSION="6"
     - python: "3.4"
       env:
         - TOX_ENV=py3.4
-        - TRAVIS_NODE_VERSION="6"
     - python: "3.5"
       env:
         - TOX_ENV=py3.5
-        - TRAVIS_NODE_VERSION="6"
     - python: "3.6"
       env:
         - TOX_ENV=py3.6
-        - TRAVIS_NODE_VERSION="6"
     - python: "2.7"
       env:
         - TOX_ENV=node6.x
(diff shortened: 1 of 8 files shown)


kolibri/utils/env.py
```diff
@@ -5,7 +5,7 @@
 
 def get_cext_path(dist_path):
```
benjaoming: Okay to rename this to `add_cext_to_sys_path` ?

benjaoming: Renamed :)

benjaoming:
@lyw07 I've added tests that install the C extensions in one of the Travis jobs.

lyw07:
@benjaoming thank you for the renaming and adding tests!

benjaoming:
I can't find evidence that the Tox environment is using the C extensions at this point, so let me do some more work on that, and this should be good to go...

lyw07:
thank you! @benjaoming

benjaoming:
Oh wow, just no way to do full tests w/o the C extensions.

image

Update: Travis also timed out the jobs, so it really is impossible.

benjaoming:
Hi @lyw07 

I unfortunately cannot assign you as a reviewer - but that would be quite suitable since now I've added a bunch of changes to the test setup :)

The intention is to have a test environment without C extensions - but as it turns out, we cannot run the full test suite. That's just too slow.

Since other test improvements are coming up, see #2975, I'm okay about at least just having a few quick tests w/o C extensions and leaving it at that.

What do you think?

.travis.yml
```diff
@@ -23,59 +23,56 @@ matrix:
     - python: "2.7"
       env:
         - TOX_ENV=pythonbuild2.7
-        - TRAVIS_NODE_VERSION="6"
```
benjaoming, line 26: Just typed in '6' as the version in the one particular place where NVM installs node :)

kolibri/utils/debian_check.py
```diff
@@ -13,7 +13,7 @@ def check_debian_user():
 
     with open("/etc/kolibri/username", "r") as f:
         kolibri_user = f.read().rstrip()
-    current_user = os.environ["USER"]
```
benjaoming, line 16: This was necessary to run tests on a Debian system that already has Kolibri installed!

pytest.ini
```diff
@@ -1,4 +1,8 @@
 [pytest]
 testpaths = test/ kolibri/
-DJANGO_SETTINGS_MODULE = kolibri.deployment.default.settings.test
 norecursedirs = dist tmp* .svn .*
+# Settings for pytest-django
+django_find_project = false
```
benjaoming, line 5: Stops pytest-django from spending time searching for Django projects.

pytest.ini
```diff
@@ -1,4 +1,8 @@
...
+DJANGO_SETTINGS_MODULE = kolibri.deployment.default.settings.test
+# Settings for pytest-pythonpath
+python_paths = kolibri/dist
```
benjaoming, line 8: This ensures that `kolibri/dist` is ALWAYS part of the path environment when invoking pytest -- For now, it is interesting in case it helps us find clutter between our test environment and our distributed environment. But once we start running tests on our build artifacts, it won't be necessary.

.travis.yml
lyw07: that sounds very nice!

tox.ini
```diff
@@ -45,12 +42,57 @@ commands =
     coverage run -p kolibri stop
     sh -c 'kolibri manage makemigrations --check'
     # Run the actual tests
-    py.test {posargs:--cov=kolibri --color=no}
+    py.test {posargs:--cov=kolibri --cov-report= --cov-append --color=no}
```
benjaoming, line 45: 1) Suppressed the annoying coverage report output in Travis @jonboiser 
2) Appends to an existing report, such that results from above coverage isn't overwritten.

requirements/test.txt
```diff
@@ -1,5 +1,3 @@
--r base.txt
--r cext.txt
```
benjaoming, line 2: Stops our tests from unconditionally installing things that might (or should) be found in `kolibri/dist`

lyw07:
Hi @benjaoming,
Thank you for adding those changes in the PR! It will definitely help us find out the issues more easily. I agree that adding some quick tests without C extensions could be good if doing full tests without C extensions is too slow :)

benjaoming (approved):
Tested on a system with native cryptography 2.1.4 - no regressions observed.

Issue #4023, opened 2018-07-13
add facility setting to disable 'guest' browsing


indirectlylit:
### Observed behavior

Some organizations prefer that their users always be logged in

### Expected behavior

A facility setting can be added, e.g. "Disable guest access".

When this setting is active:

* "Access as guest" is hidden on the log-in page
* When a user visits a page without an active session, they are redirected to the log-in page

### User-facing consequences

Sometimes students are accessing as a guest because it's easier than dealing with a log-in

### Context

k0.10.0

----

cc @jtamiace @rtibbles this seem reasonable?

jtamiace:
Seems reasonable to me. The setup wizard would be a touchpoint for this change, so will think about how this new setting should fit in

jtamiace:
It might make sense for this to be the third step in the setup wizard

<img width="1163" alt="screen shot 2018-07-16 at 2 27 39 pm" src="https://user-images.githubusercontent.com/6668144/42784564-7737c794-8904-11e8-93ff-7c8e0d8ef37f.png">

Here's the rest of the onboarding prototype for review
https://www.figma.com/file/CiaGXfI6OMOjBF9e8hr1c1Ap/Onboarding-0.11-updates

PR #4071, opened 2018-07-24
Allow guest access setting added

ralphiee22:
### Summary

_NOTE: This PR does not add the guest access option to the setup wizard (as mentioned in https://github.com/learningequality/kolibri/issues/4023), but this PR can be used as a basis to work on those pages._

"Continue as guest" button will be hidden if guest access is not allowed. 
**_Before_**
screen shot 2018-07-24 at 1 58 21 pm

**_After_**
screen shot 2018-07-24 at 1 59 20 pm

Under facility management settings, one can toggle access as guest button. 
screen shot 2018-07-24 at 1 56 11 pm

Added middleware to redirect page requests if 1. facility does not allow guest access & 2. user is not authenticated

### Reviewer guidance

Toggle access as guest setting and try navigating unauthenticated.
Also try from fresh install into setup wizard steps.

### References

https://github.com/learningequality/kolibri/issues/4023

----

### Contributor Checklist

### Reviewer Checklist


kolibri/plugins/user/middleware.py
```diff
@@ -0,0 +1,24 @@
...
+USER_SIGNIN_URL = reverse("kolibri:user:user")
+
+
+class RedirectToSignInPageIfNoGuestAccessAndNoActiveSession(MiddlewareMixin):
+    """
+    Redirect to the sign in page if facility does not allow guest access and user
+    has not signed into the facility.
+    """
+
+    def process_request(self, request):
```
christianmemije: We should also allow a few more urls like `set_language` and `admin`, etc url?

ralphiee22: can you confirm if `set_language` is blocked by the redirect?

christianmemije: Confirming.

kolibri/plugins/user/middleware.py
```diff
@@ -0,0 +1,36 @@
...
+ALLOWED_PATH_LIST = [
+    'deviceprovision',
+    'kolibri:setupwizardplugin:setupwizard',
+    'kolibri:set_language',
+    'session-list'
+    'admin:index',
+]
+
+
+class RedirectToSignInPageIfNoGuestAccessAndNoActiveSession(MiddlewareMixin):
```
ralphiee22: Will something like this https://github.com/learningequality/kolibri/blob/develop/kolibri/plugins/setup_wizard/middleware.py#L17 be needed for this middleware as well? cc: @rtibbles @jamalex

kolibri/plugins/user/middleware.py
```diff
@@ -0,0 +1,36 @@
+from django.conf import settings
+from django.core.urlresolvers import reverse
+from django.shortcuts import redirect
+from django.utils.deprecation import MiddlewareMixin
+
+from kolibri.core.auth.models import Facility
+from kolibri.core.auth.models import KolibriAnonymousUser
+
+ALLOWED_PATH_LIST = [
```
rtibbles: This middleware should be in core, not in the user plugin (the setup wizard is an anti-pattern in this respect, and not to be copied).

We should also not have an explicit list of paths that are allowed, as this means that no other plugin can ever exempt itself from the guest access restriction.

Instead, we should use a view decorator to exclude it from this middleware (see, for example, the CSRF exempt decorator in Django).

kolibri/plugins/user/middleware.py
rtibbles: Fortunately, this change is now superfluous thanks to: https://github.com/learningequality/kolibri/blob/develop/kolibri/core/device/middleware.py#L16 - this shortcuts any other middleware in the case that the user agent is the Windows GUI.

kolibri/plugins/user/middleware.py
```diff
@@ -0,0 +1,36 @@
...
+        if settings.DEBUG:  # allow anonymous rest api browsing
+            allowed_paths.append('/api')
+        if any(request.path.startswith(path_prefix) for path_prefix in allowed_paths):
+            return
+
+        if isinstance(request.user, KolibriAnonymousUser):
+            dataset = getattr(Facility.get_default_facility(), 'dataset', None)
+            if dataset and not dataset.allow_guest_access:
+                if not request.path.startswith(reverse("kolibri:user:user")):
+                    return redirect(reverse("kolibri:user:user"))
```
rtibbles: You should always redirect to `/`. The plugin system takes over from there to determine where is best to redirect to. c.f. https://github.com/learningequality/kolibri/blob/develop/kolibri/core/views.py#L72.

kolibri/plugins/user/middleware.py
```diff
@@ -0,0 +1,36 @@
...
+    """
+    Redirect to the sign in page if facility does not allow guest access and user
+    has not signed into the facility.
+    """
+
+    def process_request(self, request):
+        # Don't redirect for URLs that are required for the setup wizard
+        allowed_paths = [reverse(name) for name in ALLOWED_PATH_LIST]
+        if settings.DEBUG:  # allow anonymous rest api browsing
+            allowed_paths.append('/api')
```
rtibbles: Note there are api endpoints that do not end in API. Maybe it would be better just to allow all Views that are made from DRF viewsets? (if people want to browse our API while not logged in, I don't see that as being a problem).

kolibri/plugins/user/middleware.py
```diff
@@ -0,0 +1,36 @@
...
+        if any(request.path.startswith(path_prefix) for path_prefix in allowed_paths):
+            return
+
+        if isinstance(request.user, KolibriAnonymousUser):
+            dataset = getattr(Facility.get_default_facility(), 'dataset', None)
```
rtibbles: Don't want to change the design, I just think we should note this in the user facing design, as setting the facility setting will *only* make a difference if they are doing this for the default facility (so perhaps we should just hide this setting if someone happens to not be modifying the default facility for this device).

rtibbles (changes requested):
Implementation needs a significant change in approach, I think. See comments.

rtibbles:
Also, one other thing - we should be mocking webpack asset building for template view URLs, I will update my smoke test PR with this to get it merged, and then you can use the same logic.

kolibri/plugins/user/middleware.py
ralphiee22: regarding this, could we perhaps change the implementation of `get_default_facility` to read from a cookie in that way an admin can change the setting for any facility they happen to be logged into on a device?

rtibbles: The default facility is the default facility!

I think it is better just to be explicit that the default facility has this kind of 'override' behaviour for the whole device when we let them set the setting. (the other alternative is that this is actually a device setting, but I think that might be even more confusing for users)

kolibri/core/middleware/signin_page.py
```diff
@@ -0,0 +1,30 @@
...
+        if settings.DEBUG:
+            if hasattr(view_func, 'cls'):
+                if issubclass(view_func.cls, APIView):
+                    return None
+
+        if isinstance(request.user, KolibriAnonymousUser):
+            dataset = getattr(Facility.get_default_facility(), 'dataset', None)
+            if dataset and not dataset.allow_guest_access:
+                if not request.path.startswith(reverse("kolibri:user:user")) and request.path != '/' and not request.path.startswith(reverse('admin:index')):
+                    return redirect('/')
```
rtibbles, line 30: :+1:

kolibri/plugins/user/middleware.py
```diff
@@ -0,0 +1,36 @@
+from django.conf import settings
```
rtibbles: Looks like this file should have been deleted?

rtibbles (changes requested):
Looks like there's one file that needs clean up.

rtibbles (approved):

rtibbles:
Manually tested and works.

Issue #3963, opened 2018-07-04
Version  0.10 Server SSL Errors


nobicycle:
Hi. I noticed v0.10 was out and did a clean install, deleting the old kolibri directory.
I selected Khan Academy for import. "Generating channel listing. This could take a few minutes…" took so long I looked in the logs, attached. Seems like there are various issues.

[server.log](https://github.com/learningequality/kolibri/files/2161824/server.log)
[kolibri.log](https://github.com/learningequality/kolibri/files/2161825/kolibri.log)

indirectlylit:
@nobicycle thanks for the report!

Could you please provide:

* install method (pex, pypi, etc)
* OS and version of OS

nobicycle:
Sorry.
Arch( Manjaro) Linux
$ uname -a
Linux tower 4.14.52-1-MANJARO #1 SMP PREEMPT Tue Jun 26 15:24:31 UTC 2018 x86_64 GNU/Linux

Method: pip install in a vitualenv

benjaoming:
It seems that the error occurs in the middle of downloading of channel id 1ceff53605e55bef987d88e0908658c5

@nobicycle thanks a lot for filing this. I suspect this is related to some kind of missing error-retry. So probably a sub task failed, and there is no user feedback in the UI unfortunately. You could potentially retry this, it doesn't seem like something that would persist.

```
INFO:kolibri.content.management.commands.importchannel:Downloading data for channel id 1ceff53605e55bef987d88e0908658c5

  0%|          | 0/162202624 [00:00<?, ?it/s]
  1%|▏         | 2097152/162202624 [00:05<07:10, 371926.49it/s]venv/lib/python3.6/site-packages/kolibri/dist/tqdm/_monitor.py:89: TqdmSynchronisationWarning: Set changed size during iteration (see https://github.com/tqdm/tqdm/issues/481)
  TqdmSynchronisationWarning)

  3%|▎         | 4194304/162202624 [00:12<07:35, 346983.86it/s]
  4%|▍         | 6291456/162202624 [00:17<07:10, 361964.45it/s]
  5%|▌         | 8388608/162202624 [00:22<06:35, 388820.85it/s]
  6%|▋         | 10485760/162202624 [00:26<06:02, 418468.50it/s]
  8%|▊         | 12582912/162202624 [00:30<05:42, 437342.04it/s]
  9%|▉         | 14680064/162202624 [00:34<05:16, 465434.39it/s]
[...]

PR #4078, opened 2018-07-26
add retry logic to importchannel and catch timeout/ssl exceptions

lyw07:
### Summary

This PR is to add retry logic to `importchannel` command and catch additional exceptions in the retry logic - `Timeout` error and `OpenSSL.SSL.Error: [('SSL routines', 'ssl3_get_record', 'decryption failed or bad record mac')]`

I also clean up the code by moving the retry logic from `importcontent` into `utils/import_export_content.py` so we can reuse the code in `importchannel` to reduce redundancy.

Since some platforms won't have `OpenSSL` module after the fix in https://github.com/learningequality/kolibri/pull/4059, I have the code below to handle the case
```
try:
    import OpenSSL
    SSLERROR = OpenSSL.SSL.Error
except ImportError:
    import requests
    SSLERROR = requests.exceptions.SSLError
```

### Reviewer guidance

When the internet is not available, it's possible to reproduce the `ReadTimeout` error, which is a subclass of `Timeout` error.
I'm still not sure how to reproduce the `SSLError` in kolibri. But you can produce the error by the script here: https://github.com/requests/requests/issues/1906#issuecomment-201120959. I used it to see how the exception looks.

### References

https://github.com/learningequality/kolibri/issues/3963

----

### Contributor Checklist

### Reviewer Checklist

### DIFF
--- a/kolibri/content/management/commands/importchannel.py
+++ b/kolibri/content/management/commands/importchannel.py
@@ -1,5 +1,6 @@
 import logging as logger
 import os
+from time import sleep
 
 from django.core.management.base import CommandError
 
@@ -98,6 +99,9 @@ class Command(AsyncCommand):
         finished = False
         while not finished:
             finished = self._start_file_transfer(filetransfer, channel_id, dest)
+            if self.is_cancelled():
+                self.cancel()
+                break
 
     def _start_file_transfer(self, filetransfer, channel_id, dest):
         progress_extra_data = {
@@ -120,16 +124,22 @@ class Command(AsyncCommand):
                 if self.is_cancelled():
                     try:
                         os.remove(dest)
-                    except IOError:
+                    except IOError as e:
+                        logging.error("Tried to remove {}, but exception {} occured.".format(
+                            dest, e))
                         pass
                     self.cancel()
                 return True
 
         except Exception as e:
-            # The return value of retry_import is
-            # * True, 0 - the transfer is cancelled.
-            # * False, 0 - the transfer fails and needs to retry.
-            return retry_import(self, e, 'channel')[0]
+            logging.error("An error occured during channel import: {}".format(e))
+            retry_import(e, skip_404=False)
+
+            logging.info('Waiting for 30 seconds before retrying import: {}\n'.format(
+                filetransfer.source))
+            sleep(30)
+
+            return False
 
     def handle_async(self, *args, **options):
         if options['command'] == 'network':
--- a/kolibri/content/management/commands/importchannel.py
+++ b/kolibri/content/management/commands/importchannel.py
@@ -6,6 +6,7 @@ from django.core.management.base import CommandError
 from ...utils import channel_import
 from ...utils import paths
 from ...utils import transfer
+from ...utils.import_export_content import retry_import
 from kolibri.core.errors import KolibriUpgradeError
 from kolibri.tasks.management.commands.base import AsyncCommand
 from kolibri.utils import conf
@@ -94,14 +95,17 @@ class Command(AsyncCommand):
 
         logging.debug("Destination: {}".format(dest))
 
+        finished = False
+        while not finished:
+            finished = self._start_file_transfer(filetransfer, channel_id, dest)
+
+    def _start_file_transfer(self, filetransfer, channel_id, dest):
         progress_extra_data = {
             "channel_id": channel_id,
         }
 
-        with filetransfer:
-
-            with self.start_progress(total=filetransfer.total_size) as progress_update:
-
+        try:
+            with filetransfer, self.start_progress(total=filetransfer.total_size) as progress_update:
                 for chunk in filetransfer:
 
                     if self.is_cancelled():
@@ -119,6 +123,13 @@ class Command(AsyncCommand):
                     except IOError:
                         pass
                     self.cancel()
+                return True
+
+        except Exception as e:
+            # The return value of retry_import is
+            # * True, 0 - the transfer is cancelled.
+            # * False, 0 - the transfer fails and needs to retry.
+            return retry_import(self, e, 'channel')[0]
 
     def handle_async(self, *args, **options):
         if options['command'] == 'network':
--- a/kolibri/content/utils/import_export_content.py
+++ b/kolibri/content/utils/import_export_content.py
@@ -1,5 +1,3 @@
-from time import sleep
-
 from django.db.models import Sum
 from le_utils.constants import content_kinds
 from requests.exceptions import ConnectionError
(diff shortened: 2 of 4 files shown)


kolibri/content/management/commands/importchannel.py
```diff
@@ -119,6 +123,13 @@ def _transfer(self, method, channel_id, baseurl=None, path=None):
                     except IOError:
                         pass
```
benjaoming: Could you add a comment or log entry here, while we're at it? :)

> `"Tried to remove {}, but exception {} occurred".format(dest, e)`

What I mean is that the file isn't expected to be missing, so if it is.. eh that's weird.

kolibri/content/utils/import_export_content.py
```diff
@@ -61,3 +77,40 @@ def get_num_coach_contents(contentnode, filter_available=True):
             .count()
     else:
         return 1 if contentnode.coach_content else 0
+
+
+def retry_import(command, e, name, f=None, overall_progress_update=None):
+    logging.error("An error occured during channel import: {}".format(e))
```
benjaoming: Logging the exception in a different module than it occurred, makes it harder to troubleshoot.

You can think of the function `retry_import` as doing what it is named to do.. the logging part is beyond that, so it's more natural that the [callee](https://en.wikipedia.org/wiki/Calling_convention).

kolibri/content/utils/import_export_content.py
```diff
@@ -61,3 +77,40 @@ def get_num_coach_contents(contentnode, filter_available=True):
             .count()
     else:
         return 1 if contentnode.coach_content else 0
+
+
+def retry_import(command, e, name, f=None, overall_progress_update=None):
```
benjaoming: This needs a docstring! :)

kolibri/content/utils/import_export_content.py
```diff
@@ -61,3 +77,40 @@ def get_num_coach_contents(contentnode, filter_available=True):
...
+            isinstance(e, ConnectionError) or
+            isinstance(e, Timeout) or
+            (isinstance(e, HTTPError) and e.response.status_code in RETRY_STATUS_CODE) or
+            (isinstance(e, SSLERROR) and 'decryption failed or bad record mac' in str(e))):
+        return sleep_before_retry(command), 0
+
+    # Skip the file if it does not exist on the server or disk.
+    # This only applies to content import not channel import.
+    elif (
+            name == 'content' and
```
benjaoming: Can we remove the whole `name` parameter and instead use deliberate `skip_404=True` kwarg?

kolibri/content/utils/import_export_content.py
```diff
@@ -61,3 +77,40 @@ def get_num_coach_contents(contentnode, filter_available=True):
...
+    else:
+        raise e
+
+def sleep_before_retry(command):
+    logging.info('Waiting for 30 seconds before retrying...')
+    for i in range(30):
+        if command.is_cancelled():
+            command.cancel()
+            return True
+        sleep(1)
```
benjaoming: This busy polling is too busy. I think you can sleep for 30 seconds and then check if the job is canceled after that.

The 30 seconds should be a constant expressed somewhere.

kolibri/content/utils/import_export_content.py
```diff
@@ -61,3 +77,40 @@ def get_num_coach_contents(contentnode, filter_available=True):
...
+            (
+                (isinstance(e, HTTPError) and e.response.status_code == 404) or
+                (isinstance(e, OSError) and e.errno == 2))):
+        overall_progress_update(f.file_size)
+        return True, 1
+
+    else:
+        raise e
+
+def sleep_before_retry(command):
```
benjaoming: Normalize it into the callee - it's only called once AFAIK, so you shouldn't split it into its own function.

kolibri/content/utils/import_export_content.py
```diff
@@ -61,3 +77,40 @@ def get_num_coach_contents(contentnode, filter_available=True):
...
+    logging.info('Waiting for 30 seconds before retrying...')
```
benjaoming: If you move the logging to where the try is called, you can log it from somewhere where you have much more context. So for instance `"Waiting {} seconds before retrying download: {}".format(...)` (just sketching)

benjaoming (changes requested):
Not in all cases, but in general, logging patterns are better as:

```
logger.info("Now calling x()")
y = x()
logger.info("Called x(), got {}".format(y))
```

...because you get the more interesting module path.

Can you post some output from running channel imports while having server errors? What I would like to know especially is if the user is properly informed. Let's say that downloading from `s3.amazon.com` is failing, I would expect warnings like this:

```
Got exception 'SSLWantReadError' while trying to get destination "/x/y.mp4" from "https://s3.amazon.com"
Resuming download at 15.4%
Waiting 30 seconds for Retry 4 of 10...
```

Or something like that - let me know if you have questions :)

mrpau-julius:
Base on my test, This error will appeared when the the computer turn to sleep mode and I try to turn off the wifi within 10 secs and turn it on again same error appeared and wait for 30 secs to continue download the progress.

screen shot 2018-07-27 at 5 07 27 pm

kolibri/content/management/commands/importchannel.py
lyw07: Yes I will add a log here

kolibri/content/utils/import_export_content.py
lyw07: ah yes that makes sense. I will move the log to the module that the exception occurs.

kolibri/content/utils/import_export_content.py
lyw07: Yes thanks! I will add a docstring there

kolibri/content/utils/import_export_content.py
lyw07: yes I can do that!

kolibri/content/utils/import_export_content.py
lyw07: I think we originally chose to check if the job is cancelled every second during this 30-second period because a user might want to cancel the job during this 30-second sleep time. And I'm afraid that the user may think the button is not working if he/she clicks the button to cancel the job at the first second of this 30-second period...

kolibri/content/utils/import_export_content.py
lyw07: yes! thank you for the suggestion!

kolibri/content/utils/import_export_content.py
benjaoming: Just clarifying this part:

> Logging the exception in a different module than it occurred, makes it harder to troubleshoot.

This is because the logs are created with `logger = logging.getLogger(__name__)` so when invoking the logger, it's better to do in a place where it's most relevant to the log message :)

kolibri/content/utils/import_export_content.py
benjaoming: I understand - what I worry is that if someone is downloading 50 GB on an unstable connection, could this cause thousands of polls during retries?

lyw07: ah yes, that's true. It makes sense to me. How about checking if the user cancels the import every 10 or 15 seconds instead of doing every second?

indirectlylit: >  thousands of polls during retries

@benjaoming by 'polls' are you referring to `command.is_cancelled()`? Is this an expensive operation?

benjaoming: @indirectlylit likely not, the storage of Iceqube is in memory afaik, but whenever something is busy polling, you are going to see a reaction from someone like me, who doesn't like it :)

For instance, assuming Iceqube is thread-safe, that implies getting a read lock every time you poll something in the job queue.

benjaoming (approved):
Thanks for addressing my feedback @lyw07 !

benjaoming:
@lyw07 In the hope that it's unrelated to this PR, I restarted a couple of failing jobs in Travis because it had been stuck for some time, and when it started again, it was mentioning corrupt files.

lyw07:
@benjaoming Thanks! If it's the builds for this PR, I think it's because I rebased the commit and pushed it again before the builds finish, since I realized that I forgot to change the test cases

benjaoming (approved):
Great logging btw - looks really useful for users who may want to understand and troubleshoot their connection from logs.

PR #4092, opened 2018-08-01
Bump to 0.10.1 dev series

benjaoming:
Current builds are being released as 0.10.0

Will consolidate release process, so we catch this in the future subsequent to final releases.

### DIFF
--- a/kolibri/__init__.py
+++ b/kolibri/__init__.py
@@ -15,7 +15,7 @@ env.set_env()
 
 #: This may not be the exact version as it's subject to modification with
 #: get_version() - use ``kolibri.__version__`` for the exact version string.
-VERSION = (0, 10, 0, 'final', 0)
+VERSION = (0, 10, 1, 'alpha', 0)
 
 __author__ = 'Learning Equality'
 __email__ = 'info@learningequality.org'


indirectlylit:
@benjaoming note that in the docs I have specified that "Patch releases only have betas, not alphas":

https://kolibri-dev.readthedocs.io/en/develop/references/release_process.html#release-process

This has been a matter of convention & convenience so far to help clarify that patches should never revert back to an 'alpha' state of stability. We can update the docs or the version file.

benjaoming:
@indirectlylit 

If we don't bump to the next dev series after a stable release, all of the subsequent builds are mis-labeled as the previous stable release..

indirectlylit:
Yes understood. In the past though we've bumped to beta, not alpha.

benjaoming:
There's an intermediate process between the final 0.1 first 0.2-beta1, this is expressed as 0.2.dev0 - I think "beta 0" is ill-defined in this sense, so it's better to state "alpha 0", covering pre-beta stages (and pre-alpha in the event we released an alpha as with 0.11).

benjaoming:
Does anything need updates in the release workflow?

indirectlylit:
> There's an intermediate process between the final 0.1 first 0.2-beta1, this is expressed as 0.2.dev0

It's not clear to me where this requirement comes from. E.g. is it something technical in our versioning code, a PEP-440 convention, or general release process convention?

For previous releases we've (sometimes) gone directly from final to beta-1, and it hasn't been an issue. See e.g.

* ce46f8b1e14ef14554c14741d63b0e80f8641a91
* 32699f301a8b8ffb47712eea2a8a8339ebd158ff

The advantage has been that we can simply say in releases branches, code is always either final or beta. The disadvantage is that there's an implicit beta-1, and it hasn't been clear to me whether to actually create a tagged release for that or not... I think that's what you mean by the "intermediate process"?

I agree that the whole thing is a bit ambiguous and we should update the release process [release process](https://kolibri-dev.readthedocs.io/en/develop/references/release_process.html#create-a-release-branch) docs to be explicit

benjaoming:
> It's not clear to me where this requirement comes from. E.g. is it something technical in our versioning code, a PEP-440 convention, or general release process convention?

If the `VERSION` tuple is a _final_ version, then the rest of the logic is switched off. This is necessary, to avoid needlessly (and dangerously!) creating `git` subprocesses to generate the version string.

Therefore, if you for instance bump to `(0, 10, 1, 'final')`, then everything build on `release-v0.10.x` will be `0.10.1` and this is a mis-labeling of all subsequent PRs and automated builds leading up to the first pre-release of a `0.10.2`.

> always either final or beta

This assumption is already broken in 0.11.0 series ;)

indirectlylit:
I think we're talking past each other again :)

PR #4121, opened 2018-08-07
Maintain source translation files in git + update languages for 0.10.x

benjaoming:
### Summary

Very few translations have been missing in 0.10 - guaranteed missing from builds from a clean environments.

Shipped translations originate from 0.10, so there are just very few translations missing.

Cause of issue: We did not have any source translation files in the repo, and they didn't get generated when downloading from CrowdIn. So a clean checkout would lack the source language, and the CrowdIn CLI would not be able to pattern match the contents of the empty `kolibri/local/en`

- [x] update Make target `translation-crowdin-download`
- [x] add new translations

### Reviewer guidance

n/a - anything I don't get here?

In my experience, it's totally conventional practice to maintain source languages in a repo :)

### References

Don't think we opened issues, but @radinamatic flagged this many times :)

----

### Contributor Checklist

### Reviewer Checklist


benjaoming:
image

rtibbles:
Can anyone manually testing this, just check that running `yarn run devserver` with no code changes does not result in a dirty git tree?

radinamatic:
It's not the test that @rtibbles was asking for, but the asset built from this PR installs without problems on Windows 10, displays the current translations from Crowdin that were not included in the latest release, and presents no other issues upon cursory testing.

benjaoming:
I'm trying to rethink if we can instead add a dependency in the Makefile such that `translation-crowdin-download` depends on `translation-extract`. That will rebuild source files before fetching translations from CrowdIn. The problem is if the source files on the current branch are different of course, but it can avoid maintaining a larger file set in the repo.

benjaoming:
I pulled in b935231 because `make assets` was an implicit prerequisite for `make translations-extract`, and I couldn't execute without b935231. I keep bumping in to time-consuming local build issues because of this, so I hope it's fine to get this into the release branch. It's been in `develop` for some months without any reports.

benjaoming:
This should be good to go now, @radinamatic if you see the Spanish translation showing up correctly, please approve the PR :)

radinamatic:
This looks good @benjaoming 👍 

Could you make sure to pull the latest translations, again...? 😊 
I´ve been tinkering and tightening the Spanish translation even more, again 😉

radinamatic (approved):
LGTM! Merge away! 👍

Makefile
```diff
@@ -163,7 +164,10 @@ dist: writeversion staticdeps staticdeps-cext buildconfig assets translation-dja
 pex: writeversion
 	ls dist/*.whl | while read whlfile; do pex $$whlfile --disable-cache -o dist/kolibri-`cat kolibri/VERSION | sed 's/+/_/g'`.pex -m kolibri --python-shebang=/usr/bin/python; done
 
-translation-extract: assets
+translation-extract: clean-build
+	@echo ""
+	@echo "!! This assumes that you are running with latest assets. Run 'make assets' if in doubt !!"
+	@echo ""
```
indirectlylit, line 170: This change has left our documentation out-of-date, and the command itself is no longer named correctly.

ref: #4228

benjaoming: Sorry that you had issues debugging this and that changes left inaccurate information in the documentation. I misunderstood what was going on, thought that `makemessages` was still doing extractions for frontend at this point, but I can see that everything seems to be handled in `frontend_build/src/extract_$trs.js`.

Maybe we should move this out of assets building and into `translations-extract` since it's only essential at this point? I mean by adding `node frontend_build/src/extract_\$trs.js` in the target?

rtibbles: It is useful for it to happen dynamically during asset building as that keeps the frontend messages meaningful, plus the webpack process allows us to automatically identify which strings should be extracted and which plugin they are associated with.

In short, no.

benjaoming: Oh, it is a pre-requisite for messages to display in English (the source language) ?

rtibbles: Yes, otherwise the message id is used!

benjaoming: Ah, okay, that's no good - and we need to use these Message IDs instead of the English source strings?

rtibbles: Yeah, the source strings can get quite complex and having them referenced by id means we don't have to go looking for them in the source of the templates!

benjaoming: Yup, there's no use removing it from `make assets` then.

Would it still be useful to add `node frontend_build/src/extract_\$trs.js` to `make translations-extract` ?

rtibbles: Just running that will not do anything, because it is a plugin for webpack, so the asset pipeline needs to be run when generating messages.

benjaoming: Ah, okay, so putting all this together: There is no way to extract frontend messages separately from the assets building, this has to be done in `make assets`?

rtibbles: Yes, unless we want to do some fairly extensive code refactoring :)

benjaoming: Nah, let's skip that :) I think it's a part of the architecture of message extraction in JS: It cannot be detected by an outside toolchain, but relies on the runtime environment (which kind of becomes part of the toolchain).

PR #4158, opened 2018-08-13
Android Tweaks

DXCanas:
- Resolves https://github.com/learningequality/kolibri-installer-android/issues/5
- Resolves https://github.com/learningequality/kolibri-installer-android/issues/16, which re-broke due to the tqdm package importing multiprocessing?
- Resolves https://github.com/learningequality/kolibri-installer-android/issues/35

### DIFF
--- a/kolibri/core/discovery/utils/filesystem/posix.py
+++ b/kolibri/core/discovery/utils/filesystem/posix.py
@@ -17,7 +17,7 @@ OSX_MOUNT_PARSER = re.compile("^(?P<device>\S+) on (?P<path>.+) \((?P<filesystem
 #  /dev/sdb2 on /media/user/KEEPOD type ext4 (rw,nosuid,nodev,uhelper=udisks2)
 LINUX_MOUNT_PARSER = re.compile("^(?P<device>\S+) on (?P<path>.+) type (?P<filesystem>\S+)", flags=re.MULTILINE)
 
-# Regex parser for the output of 'mount' on Android, which contains raws that looks like:
+# Regex parser for the output of 'mount' on Android, which contains rows that looks like:
 #  /dev/block/bootdevice/by-name/userdata /data ext4 rw,seclabel,nosuid,nodev,noatime,noauto_da_alloc,data=ordered 0 0
 # Note that access to /proc/ is restricted in later versions of Android. Will break the app.
 RAW_MOUNT_PARSER = re.compile("^(?P<device>\S+) (?P<path>\S+) (?P<filesystem>\S+)", flags=re.MULTILINE)
--- a/kolibri/core/discovery/utils/filesystem/posix.py
+++ b/kolibri/core/discovery/utils/filesystem/posix.py
@@ -17,9 +17,10 @@ OSX_MOUNT_PARSER = re.compile("^(?P<device>\S+) on (?P<path>.+) \((?P<filesystem
 #  /dev/sdb2 on /media/user/KEEPOD type ext4 (rw,nosuid,nodev,uhelper=udisks2)
 LINUX_MOUNT_PARSER = re.compile("^(?P<device>\S+) on (?P<path>.+) type (?P<filesystem>\S+)", flags=re.MULTILINE)
 
-# Regex parser for the contents of `/proc/mounts` (mostly needed for Android), which contains rows that looks like:
+# Regex parser for the output of 'mount' on Android, which contains raws that looks like:
 #  /dev/block/bootdevice/by-name/userdata /data ext4 rw,seclabel,nosuid,nodev,noatime,noauto_da_alloc,data=ordered 0 0
-RAW_MOUNT_PARSER = re.compile("^(?P<device>\S+) (?P<path>.+) (?P<filesystem>\S+)", flags=re.MULTILINE)
+# Note that access to /proc/ is restricted in later versions of Android. Will break the app.
+RAW_MOUNT_PARSER = re.compile("^(?P<device>\S+) (?P<path>\S+) (?P<filesystem>\S+)", flags=re.MULTILINE)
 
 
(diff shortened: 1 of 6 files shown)


DXCanas:
@aronasorman what's left to make this ready to merge? I tried importing `multiprocessing` rather than the `myhero` dummy package, but I'm getting a bunch of flake8 errors.

aronasorman:
I remember we needed to get space usage working properly before we can actually import content. But that can be a separate PR.

DXCanas:
@aronasorman ah you're right. 

@jamalex I'm still trying to figure out some JNI issues with Kevin, but Aron and I have a hunch that the import page isn't working because of some space calculation stuff in Morango - was that what you were talking about when I first kicked off?

indirectlylit:
closing PR until it's time to revisit this work

DXCanas:
@jamalex @ralphiee22 The shim is effective in granting access to the import from studio workflow, but I get this error:

quickmemo _2018-11-06-16-56-34

When I try to open up a channel's page. Are file size reports a morango thing?

indirectlylit:
> DO NOT ACKNOWLEDGE. I JUST NEED THE ARTIFACTS

Acknowledged

jamalex:
> Are file size reports a morango thing?

No, they're calculated here:
https://github.com/learningequality/kolibri/blob/develop/kolibri/core/discovery/utils/filesystem/posix.py#L89

If you have a QPython shell, you could try some of those things interactively on an Android device.

(Note: this is about free disk space, not file sizes)

kolibri/core/tasks/management/commands/base.py
```diff
@@ -3,7 +3,8 @@
 
 from django.core.management.base import BaseCommand
 from iceqube.exceptions import UserCancelledError
-from tqdm import tqdm
```
DXCanas, line 6: @aronasorman @lyw07 

Could we make this an optional dependency so that we don't need to monkey patch like this?

What's the appropriate place to file this issue?

lyw07: sorry I'm a bit confused.. why are we doing this here? @DXCanas

DXCanas: if I recall, `tqdm` imports `multiprocessing`, which breaks things in Android @lyw07. This shim is a hacky way of fixing it. 

And I think that `tqdm` is being used in iceqube

rtibbles: tqdm is not being used in iceqube, it is being used inside this module.

I think what would be best is just to do the try except on the import in here, and in the except import your mock tqdm object.

kolibri/core/tqdm_shim.py
```diff
@@ -0,0 +1,17 @@
+from __future__ import print_function
+try:
+    # import multiprocessing
+    import myhero
+except ImportError as e:
+    # import our shim
+    class tqdm(object):
+        def __init__(*args, **kwargs):
+            print("tqdm __init__ shim called!")
```
rtibbles: Can probably skip this, object should have an `__init__` method already.

kolibri/core/tqdm_shim.py
```diff
@@ -0,0 +1,17 @@
...
+
+        def update(*args, **kwargs):
+            print("tqdm update shim called!")
```
rtibbles: Should change these to `logging.info` - see dev docs for how to import logging properly - maybe call with the value that was passed in, so that instead of a progress bar, we see the logged progress in the log?

kolibri/core/discovery/utils/filesystem/posix.py
```diff
@@ -17,9 +17,10 @@
 #  /dev/sdb2 on /media/user/KEEPOD type ext4 (rw,nosuid,nodev,uhelper=udisks2)
 LINUX_MOUNT_PARSER = re.compile("^(?P<device>\S+) on (?P<path>.+) type (?P<filesystem>\S+)", flags=re.MULTILINE)
 
-# Regex parser for the contents of `/proc/mounts` (mostly needed for Android), which contains rows that looks like:
+# Regex parser for the output of 'mount' on Android, which contains raws that looks like:
```
indirectlylit: typo `raws`?

indirectlylit (approved):
I haven't tested, but nothing in here looks problematic to me

kolibri/core/discovery/utils/filesystem/posix.py
DXCanas: On it!

benjaoming:
This is targeted for 0.11.x, but the `tqdm` removal was targeted for 0.12?

indirectlylit:
this can be retargeted to 0.12

DXCanas:
I targeted at 0.11 because I’ve been testing with it most. I’ve built a few
with 0.12 but had several styling issues (this wasn’t that recent).

indirectlylit:
originally the goal was to release with 0.11 but that boat sailed, so we can just target 0.12 now

benjaoming:
Thanks :) I was also wondering, but given the motivation behind the `tqdm` removal, I hope we can get this in for 0.12 :+1:

indirectlylit:
Yeah I approved it a while ago - @DXCanas are you still working on it?

indirectlylit:
changed target to `develop` which introduced conflict in `base.txt`

DXCanas:
Not working on it anymore. I'll handle the conflict then merge?

DXCanas:
@indirectlylit rebased. Figured it would be a cleaner merge that way

benjaoming:
> 18.18% of diff hit (target 51.2%) 

Is there a way @DXCanas that this can be tested, or is it too Android-specific? Some mocks perhaps?

indirectlylit:
good call @benjaoming -

I opened a new issue and targeted it to 0.12.1 so this doesn't become a blocker: #4907

PR #4169, opened 2018-08-20
Fix the issue that progress bar could reach over 100%

lyw07:
### Summary

This PR is to resolve the issue reported by @laurenlichtman that the progress bar could reach over 100% when downloading content.
The problem behind it is that the Internet becomes unreachable in the middle of downloading one file. For example, it could be 20% of this file. Since we implemented the retry logic, the file will be downloaded again starting from 0%. So when the program finishes downloading the content, the progress bar will be shown as 100%+20% = 120%.
To fix the issue, I restore the previous progress bar value to make the progress bar go back to the state before the failed file download so that it reflects the real procedure when the program starts downloading this file from 0%. 

Before the fix:
https://slack-files.com/T0KT5DC58-FCBQ8AHN0-62858c08bf

After the fix:
https://slack-files.com/T0KT5DC58-FCC1PTVKL-3d49150cf8

### Reviewer guidance

I have a testing channel `vijaz-tapuz` that only contains one large video file. 
To test the PR, we can turn off the Internet in the middle of downloading this file, and then turn it on again.

### References

https://learningequality.slack.com/archives/C0KSZ9FEG/p1533704314000091

----

### Contributor Checklist

### Reviewer Checklist

### DIFF
--- a/kolibri/content/management/commands/importcontent.py
+++ b/kolibri/content/management/commands/importcontent.py
@@ -183,6 +183,10 @@ class Command(AsyncCommand):
             * False, 0 - the transfer fails and needs to retry.
         """
         try:
+            # Save the current progress value
+            original_value = self.progresstrackers[0].progress
+            original_progress = self.progresstrackers[0].get_progress()
+
             with filetransfer, self.start_progress(total=filetransfer.total_size) as file_dl_progress_update:
                 # If size of the source file is smaller than the the size
                 # indicated in the database, it's very likely that the source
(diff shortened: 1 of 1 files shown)


kolibri/content/management/commands/importcontent.py
```diff
@@ -183,6 +183,10 @@ def _start_file_transfer(self, f, filetransfer, overall_progress_update):
             * False, 0 - the transfer fails and needs to retry.
         """
         try:
+            # Save the current progress value
+            original_value = self.progresstrackers[0].progress
```
rtibbles: Looks like this list can potentially be empty? At least according to the failing test.

lyw07: ah thank you for pointing it out! Sorry I should have noticed it earlier.

rtibbles:
Tested manually, and I saw the progress reset before my eyes!

rtibbles (approved):
It works!

PR #4183, opened 2018-08-23
Search improvements

rtibbles:
### Summary
* Make search a separate list endpoint.
* Return additional metadata with search results to allow filtering by channel_ids, and kinds
* Return total available search results
* Generalize the unused pagination logic in api-resource.js to allow this arbitrary metadata storage.
* Add 'view more' to search results to allow loading of additional search results, 30 at a time.
* Add 'update search' button to trigger new search based on change search string or altered filters
* Add disabled state for simple link KButton

### Reviewer guidance
Can you still search?
Can you find things you want?
Can you filter and still find things?

image

image

image

----

### Contributor Checklist

### Reviewer Checklist

### DIFF
--- a/kolibri/core/assets/src/views/KSelect.vue
+++ b/kolibri/core/assets/src/views/KSelect.vue
@@ -133,7 +133,6 @@
     methods: {
       handleChange(newSelection) {
         this.selection = newSelection;
-        this.$emit('change', newSelection);
       },
     },
   };
--- a/kolibri/core/assets/src/views/KSelect.vue
+++ b/kolibri/core/assets/src/views/KSelect.vue
@@ -2,7 +2,10 @@
 
   <UiSelect
     class="k-select"
-    :class="{'k-select-inline': inline}"
+    :class="{
+      'k-select-inline': inline,
+      'k-select-disabled': disabled,
+    }"
     :value="selection"
     :options="options"
     :label="label"
@@ -144,6 +147,10 @@
     width: 150px;
     margin-right: 16px;
     vertical-align: bottom;
+  }
+
+  .k-select-disabled /deep/ .ui-select__label-text.is-inline {
+    cursor: default;
   }
 
   /deep/ .ui-select__display-value {
--- a/kolibri/core/assets/src/views/KSelect.vue
+++ b/kolibri/core/assets/src/views/KSelect.vue
@@ -133,6 +133,7 @@
     methods: {
       handleChange(newSelection) {
         this.selection = newSelection;
+        this.$emit('change', newSelection);
       },
     },
   };
--- a/kolibri/core/assets/src/views/buttons-and-links/buttons.scss
+++ b/kolibri/core/assets/src/views/buttons-and-links/buttons.scss
@@ -152,4 +152,8 @@ a.button {
   &:hover:focus {
     outline: $core-outline;
   }
+  &:disabled {
+    color: $primary-flat-disabled-color;
+    cursor: default;
+  }
 }
--- a/kolibri/core/content/api.py
+++ b/kolibri/core/content/api.py
@@ -581,9 +581,9 @@ class ContentNodeSearchViewset(ContentNodeSlimViewset):
         # Use unfiltered queryset to collect channel_ids and kinds metadata.
         unfiltered_queryset = self.get_queryset()
 
-        channel_ids = unfiltered_queryset.filter(all_queries_filter).values_list('channel_id', flat=True).distinct()
+        channel_ids = unfiltered_queryset.filter(all_queries_filter).values_list('channel_id', flat=True).order_by('channel_id').distinct()
 
-        content_kinds = unfiltered_queryset.filter(all_queries_filter).values_list('kind', flat=True).distinct()
+        content_kinds = unfiltered_queryset.filter(all_queries_filter).values_list('kind', flat=True).order_by('kind').distinct()
 
         serializer = self.get_serializer(results, many=True)
         return Response({
--- a/kolibri/core/content/api.py
+++ b/kolibri/core/content/api.py
@@ -576,14 +576,14 @@ class ContentNodeSearchViewset(ContentNodeSlimViewset):
         # If no queries, just use an empty Q.
         all_queries_filter = union(all_queries) or Q()
 
-        total_results = len(set(queryset.filter(all_queries_filter).values_list('content_id', flat=True)))
+        total_results = queryset.filter(all_queries_filter).values_list('content_id', flat=True).distinct().count()
 
         # Use unfiltered queryset to collect channel_ids and kinds metadata.
         unfiltered_queryset = self.get_queryset()
 
-        channel_ids = set(unfiltered_queryset.filter(all_queries_filter).values_list('channel_id', flat=True))
+        channel_ids = unfiltered_queryset.filter(all_queries_filter).values_list('channel_id', flat=True).distinct()
 
-        content_kinds = set(unfiltered_queryset.filter(all_queries_filter).values_list('kind', flat=True))
+        content_kinds = unfiltered_queryset.filter(all_queries_filter).values_list('kind', flat=True).distinct()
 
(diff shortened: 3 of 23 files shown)


rtibbles:
@carinediaz would appreciate input on the new strings introduced here.

carinediaz:
@rtibbles `view more` and `update search` make sense to me. I think they're both terms most people should understand and translate.

kolibri/core/content/api.py
```diff
@@ -553,6 +497,100 @@ def resume(self, request, **kwargs):
...
+                results.append(match)
+
+                # bail out as soon as we reach the quota
+                if len(results) >= MAX_RESULTS:
+                    break
+            # bail out as soon as we reach the quota
+            if len(results) >= MAX_RESULTS:
+                break
+
+        total_results = len(set(queryset.filter(union(all_queries)).values_list('content_id', flat=True)))
```
ralphiee22: does `queryset.filter(union(all_queries)).values_list('content_id', flat=True).distinct().count()` not give the correct results?

rtibbles: Oh, good catch. I am always wary of using `distinct` in SQLite, but it's fine when we've already selected the `content_id`.

rtibbles: Can do similar distinct calls for the two below too.

ralphiee22 (approved):
👍

rtibbles:
Note: holding off on merge until frontend review has also happened.

kolibri/core/assets/src/views/buttons-and-links/buttons.scss
```diff
@@ -152,4 +152,8 @@ a.button {
   &:hover:focus {
     outline: $core-outline;
   }
+  &:disabled {
```
christianmemije, line 155: Let's add this disabled state to button examples in the style guide.

kolibri/plugins/learn/assets/src/views/SearchBox.vue
```diff
@@ -106,9 +227,18 @@
       },
       search() {
         if (this.searchQuery !== '') {
```
christianmemije: Let's add .trim to the searchQuery v-model.

kolibri/plugins/learn/assets/src/views/SearchBox.vue
```diff
@@ -194,4 +323,25 @@
     background-color: $core-action-dark;
   }
 
+  .filter-icon {
+    margin-right: 12px;
+    vertical-align: middle;
+  }
+
+  .filter:nth-of-type(1) {
+    margin-right: 32px;
```
christianmemije: Let's reduce this to 16px. Let's also add bottom margins to the filters to reduce crowding. Maybe 16px? 
localhost_8000_learn_

kolibri/plugins/learn/assets/src/views/SearchPage.vue
```diff
@@ -4,25 +4,36 @@
...
-    <SearchBox />
+    <SearchBox :filters="true" />
 
     <p v-if="!searchTerm">{{ $tr('noSearch') }}</p>
 
     <template v-else>
-      <h1 class="search-results">{{ $tr('showingResultsFor', { searchTerm }) }}</h1>
+      <h1 class="search-results">{{ $tr('showingResultsFor', {
+        searchTerm,
+          results: contents.length,
```
christianmemije: Weird indent?

kolibri/plugins/learn/assets/src/views/SearchPage.vue
```diff
@@ -43,9 +55,10 @@
     $trs: {
       searchPageHeader: 'Search',
       noSearch: 'Search by typing in the box above',
-      showingResultsFor: "Results for '{searchTerm}'",
+      showingResultsFor: "Results for '{searchTerm}' ({results} of {totalResults})",
```
christianmemije: We should not show the `({results} of {totalResults})` part of this string when there are no results. Also would this part of the string be better off being on it's own below the heading? Also, maybe `1-30 of 70` instead of `30 of 70`?

christianmemije (changes requested):
Definitely improved search. :1st_place_medal:  Left some minor comments.

kolibri/plugins/learn/assets/src/views/SearchBox.vue
christianmemije: Also, keep the vertical alignment of the icons to baseline?

kolibri/core/assets/src/views/buttons-and-links/buttons.scss
rtibbles: image
?

kolibri/plugins/learn/assets/src/views/SearchPage.vue
rtibbles: I think this got auto-linted this way, I did not have the additional indent when I manually edited.

rtibbles: Confirmed, the linter says no to non-weird indent.

kolibri/plugins/learn/assets/src/views/SearchPage.vue
rtibbles: We don't show this string when there are no results, there is a separate `noResults` string that gets displayed instead, so think this should be fine as is.

kolibri/plugins/learn/assets/src/views/SearchBox.vue
rtibbles: image

rtibbles:
Updated.

kolibri/plugins/learn/assets/src/views/SearchPage.vue
christianmemije: The heading is displayed as long as there is a search query.

rtibbles: Oh, sorry, misread the template logic. We should probably make this a v-else block.

rtibbles: Have updated.

rtibbles: Have updated.

christianmemije:
Spoke to @khangmach and we decided on a few changes. 
* Instead of having an update search button, let's auto trigger a new search upon the selection of a new filter option.
* Let's change the view more button to a raised secondary to make it more prominent. 
* Let's move the num of total results to below the heading `N results`

rtibbles:
Updated search behaviour:
search_filter

Updated text:
image

Updated view more:
image

Updated empty state:
image
(note that this also includes an update to k-select that sets the pointer to default when disabled)

kolibri/plugins/learn/assets/src/views/SearchPage.vue
```diff
@@ -59,7 +57,7 @@
     $trs: {
       searchPageHeader: 'Search',
       noSearch: 'Search by typing in the box above',
-      showingResultsFor: "Results for '{searchTerm}' ({results} of {totalResults})",
+      showingResultsFor: "{totalResults} results for '{searchTerm}'",
```
christianmemije: Need to use the plural syntax.

rtibbles: So near, yet so fars.

rtibbles: Updated.

kolibri/core/assets/src/views/KSelect.vue
```diff
@@ -133,6 +133,7 @@
     methods: {
       handleChange(newSelection) {
         this.selection = newSelection;
+        this.$emit('change', newSelection);
```
christianmemije: Let's document this event for the style guide.

christianmemije (approved):
Left one comment. Tested and working nicely.

PR #4198, opened 2018-08-27
Bump to 0.9.5.dev

benjaoming:
Maintainer PR

From: @rtibbles 

> Also, Nalanda reported that the 0.9.4-beta5 they are running was displaying to the user as 0.9.4 final.

### DIFF
--- a/kolibri/__init__.py
+++ b/kolibri/__init__.py
@@ -6,7 +6,7 @@ from .utils.version import get_version
 
 #: This may not be the exact version as it's subject to modification with
 #: get_version() - use ``kolibri.__version__`` for the exact version string.
-VERSION = (0, 9, 4, 'final', 0)
+VERSION = (0, 9, 5, 'alpha', 0)
 
 __author__ = 'Learning Equality'
 __email__ = 'info@learningequality.org'


kolibri/__init__.py
```diff
@@ -6,7 +6,7 @@
 
 #: This may not be the exact version as it's subject to modification with
 #: get_version() - use ``kolibri.__version__`` for the exact version string.
-VERSION = (0, 9, 4, 'final', 0)
```
rtibbles, line 9: Oh wait... when did we bump to 0.9.4 final? That was never released...

benjaoming: Woah, strange, I had understood that it was ready to be released and this is really weird... I'm pretty sure that I did this tag and release, and the tag is reported to exist, I just can't find it in the Release page or in the Tag list.

image

benjaoming: https://github.com/learningequality/kolibri/releases/edit/v0.9.4

benjaoming: https://github.com/learningequality/kolibri/releases/v0.9.4

benjaoming: image

benjaoming: Why on earth doesn't it show on the Releases list https://github.com/learningequality/kolibri/releases ??

rtibbles: That release also says there are 683 commits to release-v0.9.x since that release too... which does not seem right.

rtibbles: Can we just delete the tag? There are also no installers on that release.

benjaoming: @rtibbles yes, I'll delete the tag from command line, can't actually find it in the Github interface.

benjaoming: It is no more!

benjaoming: Alright, a correct v0.9.4 is now tagged.

benjaoming: This should mean that we can bump the branch by merging this PR. :crossed_fingers:

rtibbles (approved):

PR #4225, opened 2018-09-07
Async log writing

rtibbles:
### Summary
Writing logs is time consuming.
Why wait for the disk to be free before we complete a request?
Stash requests to update logs on the disk in memory, and then just periodically write them to disk in a batch.
Adds tests for the PATCH behaviour.
Adds tests for the loq queue behaviour.

### Reviewer guidance
Do logs still get written?
Does data get lost?
What tests should I write for this?
What are the potential failure cases?
How can I do this better?

----

### Contributor Checklist

### Reviewer Checklist

### DIFF
--- a/kolibri/logger/api.py
+++ b/kolibri/logger/api.py
@@ -25,6 +25,7 @@ from .serializers import ExamLogSerializer
 from .serializers import MasteryLogSerializer
 from .serializers import TotalContentProgressSerializer
 from .serializers import UserSessionLogSerializer
+from .tasks import add_to_save_queue
 from kolibri.auth.api import KolibriAuthPermissions
 from kolibri.auth.api import KolibriAuthPermissionsFilter
 from kolibri.auth.filters import HierarchyRelationsFilter
@@ -71,24 +72,13 @@ class LoggerViewSet(viewsets.ModelViewSet):
             raise Http404
         serializer = self.get_serializer(instance, data=request.data, partial=partial)
         serializer.is_valid(raise_exception=True)
-        self.perform_update(serializer)
 
-        if getattr(instance, '_prefetched_objects_cache', None):
-            # If 'prefetch_related' has been applied to a queryset, we need to
-            # forcibly invalidate the prefetch cache on the instance.
-            instance._prefetched_objects_cache = {}
-        default_response = dict(request.data)
-        # First look if the computed fields to be updated are listed:
-        updating_fields = getattr(serializer.root, 'update_fields', None)
-        # If not, fetch all the fields that are computed methods:
-        if updating_fields is None:
-            updating_fields = [field for field in serializer.fields if getattr(serializer.fields[field], 'method_name', None)]
-        for field in updating_fields:
-            method_name = getattr(serializer.fields[field], 'method_name', None)
-            if method_name:
-                method = getattr(serializer.root, method_name)
-                default_response[field] = method(instance)
-        return Response(default_response)
+        def save_instance():
+            self.perform_update(serializer)
+
+        add_to_save_queue(save_instance)
+
+        return Response({})
 
 
 class ContentSessionLogFilter(BaseLogFilter):
--- a/kolibri/logger/tasks.py
+++ b/kolibri/logger/tasks.py
@@ -1,42 +1,69 @@
+import logging as logger
 import threading
 import time
 
 from django.db import transaction
 
+logging = logger.getLogger(__name__)
+
+
 class AsyncLogQueue():
 
     def __init__(self):
 
+        # Value in seconds to determine the sleep time between log saving batches
         self.log_saving_interval = 5
 
-        self.active_queue = []
+        # Where new log saving functions are appended
+        self.queue = []
 
-        self.inactive_queue = []
+        # Where the to be executed log saving functions are stored
+        # once a batch save has been invoked
+        self.running = []
 
     def append(self, fn):
-        self.active_queue.append(fn)
+        """
+        Convenience method to append log saving function to the current queue
+        """
+        self.queue.append(fn)
 
-    def toggle_active_queue(self):
-        old_active_queue = self.active_queue
-        new_active_queue = self.inactive_queue
-        self.active_queue = new_active_queue
-        self.inactive_queue = old_active_queue
+    def toggle_queue(self):
+        """
+        Method to swap the queue and running, to allow new log saving functions
+        to be added to the queue while previously added functions are being executed
+        and cleared without fear of race conditions dropping saves.
+        """
+        old_queue = self.queue
+        new_queue = self.running
+        self.queue = new_queue
+        self.running = old_queue
 
-    def clear_queue(self):
-        self.inactive_queue = []
+    def clear_running(self):
+        """
+        Reset the running list to drop references to already executed log saving functions
+        """
+        self.running = []
 
-    def run_queue(self):
-        if self.inactive_queue:
+    def run(self):
+        """
+        Execute any log saving functions in the self.running list
+        """
+        if self.running:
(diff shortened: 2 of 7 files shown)


kolibri/logger/tasks.py
```diff
@@ -0,0 +1,87 @@
...
+        if self.running:
+            # Do this conditionally to avoid opening an unnecessary transaction
+            with transaction.atomic():
+                for fn in self.running:
+                    try:
+                        fn()
+                    except Exception as e:
+                        # Catch all exceptions and log, otherwise the background process will end
+                        # and no more logs will be saved!
+                        logging.debug("Exception raised during background log saving: ", e)
```
aronasorman, line 60: Should this be a `logging.warning`? Seems pretty severe when our logs fail to save.

kolibri/logger/tasks.py
```diff
@@ -0,0 +1,87 @@
...
+    def run(self):
+        """
+        Execute any log saving functions in the self.running list
+        """
+        if self.running:
+            # Do this conditionally to avoid opening an unnecessary transaction
+            with transaction.atomic():
+                for fn in self.running:
+                    try:
+                        fn()
```
aronasorman, line 56: I think it's more efficient to save all the new logs through a bulk create method:

```
Log.objects.create(obj1, obj2, obj3)
```

Rather than having each model be saved individually: ` Log.objects.create(obj1), Log.objects.create(obj2), etc.`

Since it allows the Django ORM to use bulk create SQL queries.

kolibri/logger/tasks.py
```diff
@@ -0,0 +1,87 @@
...
+                    except Exception as e:
+                        # Catch all exceptions and log, otherwise the background process will end
+                        # and no more logs will be saved!
+                        logging.debug("Exception raised during background log saving: ", e)
+
+    def start(self):
+        while True:
```
aronasorman, line 63: Nice! Much clean.

kolibri/logger/tasks.py
rtibbles: I am only using these for updates, not for creates, and morango has special sauce in its save method that would get skipped by a bulk update.

Also, the models being saved can be heterogeneous!

kolibri/logger/tasks.py
rtibbles: Fair!

aronasorman (dismissed):

rtibbles:
Removing from the 0.9.5 milestone, as this appears to be unnecessary for performance in 0.9.x and is a risky change.

@indirectlylit @jredrejo @benjaoming would appreciate any input on whether this is something we should get into 0.11, or whether we it needs more design.

kolibri/logger/api.py
```diff
@@ -71,24 +72,13 @@ def update(self, request, *args, **kwargs):
...
-        updating_fields = getattr(serializer.root, 'update_fields', None)
-        # If not, fetch all the fields that are computed methods:
-        if updating_fields is None:
-            updating_fields = [field for field in serializer.fields if getattr(serializer.fields[field], 'method_name', None)]
-        for field in updating_fields:
-            method_name = getattr(serializer.fields[field], 'method_name', None)
-            if method_name:
-                method = getattr(serializer.root, method_name)
-                default_response[field] = method(instance)
-        return Response(default_response)
```
indirectlylit, line 91: do we just never use this response?

kolibri/logger/tasks.py
indirectlylit: Without a bulk SQL query, it's not clear to me that this batching strategy would get us the performance benefits we're hoping for: we have the same code paths running, just distributed differently over time.

I think this batching is worth pursuing, but we would need different queues for each model type so that they can be properly batched.

indirectlylit (commented):
agree that without batching at the SQL level, this might not be worth the added complexity

we also might want a little more control over the frequencies of writing for different model types.


One option would be to turn `AsyncLogQueue` into something like


```python
class AsyncBatchQueue():
    def __init__(self, update_function, interval):
         ...
    def append(self, data):
         ...
```

Then you can instantiate a new queue for each model type, with a custom update function which takes a list of data objects as its argument.

kolibri/logger/tasks.py
benjaoming: Moving `transaction.atomic()` out as a decorator of `run()` seems nicer to me. It might even be moved to `start` in case that's where the queue is handled.

If the db operation becomes a bulk operation, it may be atomic in itself and not require the transaction flag.

kolibri/logger/api.py
jredrejo: We do, there are some loggers that require a response with the logged data, including some computed fields, after the serialization saves the field.

kolibri/logger/tasks.py
rtibbles: Note that we currently *cannot* do bulk updates/saves because then the morango updates that are required (setting dirty bits etc) would not get done.

Also, wrapping multiple SQL statements in a single transaction is significantly more efficient than opening a new transaction for each statement (which is what would happen without this batch writing). c.f. https://medium.com/@JasonWyatt/squeezing-performance-from-sqlite-insertions-971aff98eef2

kolibri/logger/api.py
rtibbles: I'm not sure that we actually need this in the frontend at any point though, as things get updated optimistically in the frontend (would need some double checking of this assumption though).

kolibri/logger/tasks.py
jredrejo: I think this will benefit the performance because the enqueued task is releasing the server thread of these tasks. The response will be faster and more requests per second can be attended.
As Devon says, the database will have the same impact, but avoiding writing locks in the server thread loop will give to the user a better experience. The database writing queue will be the same, but it should not affect so much to the clients.

kolibri/logger/api.py
jredrejo: If things have not changed since I did the serializers optimization, we needed it for some loggers. As an example, the MasteryLogSerializer response includes totalattempts and pastattempts that are calculated querying the Attemptlog table. If now these values are calculated in the frontend, we can discard that code. If not, it's still needed.

rtibbles: Yeah, the totalattempts and pastattempts are optimistically updated on the frontend, so I think we can not worry about returning them.

kolibri/logger/tasks.py
rtibbles: Yeah, that's definitely also the case!

rtibbles: The main concern might be that `POST` operations (as this only affects updates) might be slowed down by the background log updating, which could cause issues.

jredrejo: Yes, I would like to have something similar to the network traffic policy: add a priority label to the saves, thus data that is more important is always saved, and less important data can be discarded in case of crisis.

kolibri/logger/api.py
jredrejo: There are more fields to check: 
- 'progress', 'score' in ExamLogSerializer
- 'currentmasterylog' in ContentSummaryLogSerializer
- 'progress' in TotalContentProgressSerializer

rtibbles:
Wrapping multiple write operations, whether inserts, updates, or deletes in a single transaction is the first pass optimization for SQLite performance: https://blog.devart.com/increasing-sqlite-performance.html, so I don't really buy the idea that we need to do something more complex than I have to see some significant performance improvements.

I did a quick check, and with 1000 logs on my machine, without the transaction wrapping it took ~9.79s, when wrapped in a transaction it took ~6.12s.

indirectlylit:
Ah, I missed the `with transaction.atomic():` block. Critique retracted!

PR #4226, opened 2018-09-08
Calculate ChannelMetadata public api fields

ralphiee22:
### Summary

Adds fields to `ChannelMetadata` that will be returned by the public channel api. We now calculate these fields during channel import, as well as during a migration.
Added a test for calculating fields during data migration. 
We now have a blacklist of models to exclude during channel import, such as the `Language` M2M field on `ChannelMetadata`.

### Reviewer guidance

make sure you have content locally and 
A. go to http://0.0.0.0:8000/api/remotechannel?baseurl=http://0.0.0.0:8000/ to see public api.
B. migrate old dbs
C. import channel/content from studio

----

### Contributor Checklist

### Reviewer Checklist


kolibri/core/content/models.py
```diff
@@ -348,6 +348,15 @@ class ChannelMetadata(models.Model):
     # Minimum version of Kolibri that this content database is compatible with
     min_schema_version = models.CharField(max_length=50)
     root = models.ForeignKey(ContentNode)
+    # precalculated fields during annotation/migration
+    published_size = models.IntegerField(default=0, null=True, blank=True)
+    total_resource_count = models.IntegerField(default=0, null=True, blank=True)
+    included_languages = models.ManyToManyField(
+        "Language",
+        related_name='channels',
+        verbose_name=_("languages"),
```
rtibbles: To reduce confusion for translators, I have been trying to remove gettext calls for explicitly developer facing code, so I think we can unwrap this.

kolibri/core/content/serializers.py
```diff
@@ -143,23 +140,14 @@ def get_language(self, instance):
...
     def get_icon_encoding(self, instance):
         return instance.thumbnail
 
-    def get_published_size(self, instance):
-        channel_nodes = ContentNode.objects.filter(channel_id=instance.id)
-        return _total_file_size(_files_for_nodes(channel_nodes).filter(available=True))
+    def get_included_languages(self, instance):
+        return list(instance.included_languages.all().values_list('id', flat=True))
 
     def get_last_published(self, instance):
```
rtibbles, line 150: Hrm - this conflates the last updated time with the last publish time, do we want to do this?

kolibri/core/content/serializers.py
```diff
@@ -124,13 +123,11 @@ class Meta:
         )
 
 
-class PublicChannelSerializer(KolibriModelSerializer):
```
rtibbles, line 128: Presumably this is ok because we do not directly serialize the last updated time?

rtibbles: Is it explicitly an issue because of the validation logic?

rtibbles: Oh, I guess my first point doesn't matter because `last_updated` doesn't need this either...

kolibri/core/content/test/test_data_migrations.py
```diff
@@ -0,0 +1,100 @@
...
+from django.test import TestCase
+
+from kolibri.core.content.models import ChannelMetadata as RealChannelMetadata
+from kolibri.core.content.models import ContentNode as RealContentNode
+
+
+# Modified from https://www.caktusgroup.com/blog/2016/02/02/writing-unit-tests-django-migrations/
+
+
+class TestMigrations(TestCase):
```
rtibbles: Is this the same as the code in auth? Could we have a common base class for testing this, rather than repeating?

kolibri/core/content/utils/annotation.py
```diff
@@ -185,6 +187,52 @@ def recurse_availability_up_tree(channel_id):
...
+    # commit the transaction
+    trans.commit()
+
+    elapsed = (datetime.datetime.now() - start)
+    logger.debug("Availability annotation took {} seconds".format(elapsed.seconds))
+
+    bridge.end()
+
+
+def recurse_availability_coach_content(channel_id):
```
rtibbles: This function should be renamed, it's not about the availability of coach content, per se, but the number of coach contents.

kolibri/core/content/utils/annotation.py
```diff
@@ -185,6 +187,52 @@ def recurse_availability_up_tree(channel_id):
...
+
+    ContentNodeClass = bridge.get_class(ContentNode)
+
+    ContentNodeTable = bridge.get_table(ContentNode)
+
+    connection = bridge.get_connection()
+
+    node_depth = bridge.session.query(func.max(ContentNodeClass.level)).scalar()
+
+    logger.info('Setting availability of coach content ContentNode objects with children for {levels} levels'.format(levels=node_depth))
```
rtibbles: `Setting totals of coach content Contentnode objects`

kolibri/core/content/utils/annotation.py
```diff
@@ -216,13 +264,7 @@ def recurse_availability_up_tree(channel_id):
...
             )
 
-        logger.info('Setting availability of ContentNode objects with children for level {level}'.format(level=level))
-        # Only modify topic availability here
-        connection.execute(ContentNodeTable.update().where(
-            and_(
-                ContentNodeTable.c.level == level - 1,
-                ContentNodeTable.c.channel_id == channel_id,
-                ContentNodeTable.c.kind == content_kinds.TOPIC)).values(available=exists(available_nodes)))
+        logger.info('Setting availability of coach content ContentNode objects with children for level {level}'.format(level=level))
```
rtibbles: `Setting totals of coach content Contentnode objects`

kolibri/core/content/utils/annotation.py
```diff
@@ -238,7 +280,7 @@ def recurse_availability_up_tree(channel_id):
     trans.commit()
 
     elapsed = (datetime.datetime.now() - start)
-    logger.debug("Availability annotation took {} seconds".format(elapsed.seconds))
+    logger.debug("Availability of coach content annotation took {} seconds".format(elapsed.seconds))
```
rtibbles: `Topic coach content annotation`

kolibri/core/content/utils/annotation.py
```diff
@@ -251,3 +293,41 @@ def set_availability(channel_id, checksums=None):
...
+
+def calculate_published_size(channel):
+    content_nodes = ContentNode.objects.filter(channel_id=channel.id)
+    channel.published_size = _total_file_size(_files_for_nodes(content_nodes).filter(available=True))
+    channel.save()
+
+
+def calculate_total_resource_count(channel):
+    content_nodes = ContentNode.objects.filter(channel_id=channel.id)
+    channel.total_resource_count = content_nodes.filter(available=True).exclude(kind=content_kinds.TOPIC).count()
```
rtibbles, line 328: Does studio dedupe by `content_id`? If it doesn't, I kind of think it should, and so should Kolibri.

rtibbles (changes requested):
A bit of cleanup would be handy. Some questions and comments.

rtibbles:
Seems to work as intended.
Note that the URL is actually: http://127.0.0.1:8000/api/content/remotechannel/?baseurl=http://127.0.0.1:8000/

One issue is that the public URL returns channels with no available resources (I realize this is not a result of this PR, but this should be fixed).

kolibri/core/content/serializers.py
ralphiee22: this is up for debate. we could add another field which is actually the last updated time. but the public API relies on `last_published` for filling in some fields, so we need to carry that data over from studio. Also, when doing P2P content syncing, it seems more useful to know when a channel has been updated (perhaps with more content?) then knowing the last published. Could also add a comment which explains this discrepancy.

kolibri/core/content/test/test_data_migrations.py
ralphiee22: will do!

kolibri/core/content/test/test_annotation.py
```diff
@@ -99,7 +100,7 @@ def test_one_content_node_available(self):
 
     def test_all_content_nodes_available_coach_content(self):
         ContentNode.objects.exclude(kind=content_kinds.TOPIC).update(available=True, coach_content=True)
-        recurse_availability_up_tree(channel_id="6199dde695db4ee4ab392222d5af1e5c")
+        recurse_availability_coach_content(channel_id="6199dde695db4ee4ab392222d5af1e5c")
```
rtibbles: Looks like you need to update the name of this function in this test?

kolibri/core/public/api.py
```diff
@@ -62,7 +62,7 @@ def _get_channel_list_v1(params, identifier=None):
                             .filter(Q(lang__id__icontains=language_id) | Q(files__lang__id__icontains=language_id)).values_list('tree_id', flat=True)
         channels = channels.filter(Q(root__lang__id__icontains=language_id) | Q(root__tree_id__in=matching_tree_ids))
 
-    return channels.distinct()
+    return channels.filter(root__available=True).distinct()
```
rtibbles, line 65: :+1:

rtibbles (approved):

PR #4232, opened 2018-09-10
Add more server info

### LINKED ISSUE #4063: add mechanism to track and report method of installation in 'server info' page and error reports
### Observed behavior

Currently we can report the kolibri version number and operating system, but not mechanism of installation (pex vs pip vs deb vs exe)

### Expected behavior

We should have the mechanism of installation available to the app for use in e.g. the server info page and error reports

### User-facing consequences

When users submit error reports, it's more complicated for us to ask "how did you install it" than to simply say "copy and paste this text"

### Context

0.10.0

jredrejo:
### Summary
With these changes, Koliri device info page will show
* Server Installer: A discovery of the kind of installation used to run Kolibri.
* Server type: Webserver header description, if any.

…

### Reviewer guidance
After login into kolibri, going to /device/#/info will show the server installer and type info in the screen
detectado_pex
detectado_code

### References

 * Resolves #4063 

### Contributor Checklist

### Reviewer Checklist


kolibri/utils/server.py
```diff
@@ -322,3 +323,39 @@ def get_urls(listen_port=None):
...
+        if 'start' in sys.argv:
+            if launcher.endswith('.pex'):
+                install_type = 'PEX file'
+            elif launcher == '/usr/bin/kolibri':
+                install_type = 'Debian package'
+            elif '\\Scripts\\kolibri' in launcher:
+                paths = sys.path
+                for path in paths:
+                    if 'kolibri.exe' in path:
+                        install_type = 'Windows installer'
```
jamalex: Would probably be good to have these be constants, and render the (internationalized) human-readable text in the frontend UI -- that way I can use this function programmatically, also.

kolibri/utils/server.py
```diff
@@ -322,3 +323,39 @@ def get_urls(listen_port=None):
...
+        - 'unknown': this function has not been able to guess it
+    """
+    install_type = 'unknown'
+    if len(sys.argv) > 1:
+        launcher = sys.argv[0]
+        if 'start' in sys.argv:
+            if launcher.endswith('.pex'):
+                install_type = 'PEX file'
+            elif launcher == '/usr/bin/kolibri':
+                install_type = 'Debian package'
```
benjaoming: This will not work for someone doing a different *nix package :) I'm not saying that I know of an RPM or something.. but well, could we just add something like `launcher == '/usr/bin/kolibri' and os.path.isfile("/etc/apt/sources.list")` ?

Quick suggestion: I'm also interested in knowing the built version of a package.. so for instance, during .deb package build-time, I could dump `dpkg-parsechangelog --show-field Version` in `kolibri/VERSION_PKG` and we make this file a convention for storing the packaged version number?

kolibri/utils/server.py
```diff
@@ -322,3 +323,39 @@ def get_urls(listen_port=None):
...
+                paths = sys.path
+                for path in paths:
+                    if 'kolibri.exe' in path:
+                        install_type = 'Windows installer'
+                        break
+            else:
+                install_type = 'Executable compiled from code'
+        else:
+            if 'runserver' in sys.argv:
+                install_type = 'Development code'
```
jamalex: Where does the "pip install" method get caught? That's another one we can't track good download stats for.

kolibri/utils/server.py
```diff
@@ -322,3 +323,39 @@ def get_urls(listen_port=None):
...
+            elif launcher == '/usr/bin/kolibri':
+                install_type = 'Debian package'
+            elif '\\Scripts\\kolibri' in launcher:
+                paths = sys.path
+                for path in paths:
+                    if 'kolibri.exe' in path:
+                        install_type = 'Windows installer'
+                        break
+            else:
+                install_type = 'Executable compiled from code'
```
jamalex: What does this one mean?

benjaoming: Maybe let this remain "unknown" or "other".

It's also not compiled :)

jamalex (dismissed):

kolibri/utils/server.py
```diff
@@ -322,3 +323,39 @@ def get_urls(listen_port=None):
...
+        - 'Debian package'
+        - 'Windows installer'
+        - 'Compiled from code or from a Whl file'
+        - 'Development code'
+        - 'unknown': this function has not been able to guess it
+    """
+    install_type = 'unknown'
+    if len(sys.argv) > 1:
+        launcher = sys.argv[0]
+        if 'start' in sys.argv:
```
benjaoming: Anyone can do `kolibri manage runserver` on their installation, thereby bypassing the below conditions.. also, a development installation can run `kolibri start`. I think it's better remove `if 'start' in sys.argv`.

kolibri/utils/server.py
```diff
@@ -322,3 +323,39 @@ def get_urls(listen_port=None):
         return STATUS_RUNNING, urls
     except NotRunning as e:
         return e.status_code, []
+
+
+def installation_type():
```
benjaoming: Maybe moving this to `kolibri.utils.version` since it's not supposed to depend on anything external, except maybe a `VERSION_PKG` well-known file location? (see below suggestion)

benjaoming (changes requested):
Have a couple of suggestions here :)

kolibri/utils/server.py
jredrejo: if you do `kolibri start`in  a development installation you are not running it from development, but from the 'Compiled code..' . With 'Development code' I want to mean exactly that: running from django-server while coding. Any other kind of call using start will require a compilated version.

kolibri/utils/server.py
jredrejo: You are running kolibri from the 'binary' generated after running `pip install -e .` not from a package, installer , django server, etc.

kolibri/utils/server.py
jredrejo: in 'Compiled from code or from a Whl file'

kolibri/utils/server.py
jredrejo: you mean the outputs of the function as constants I guess, right? Yes, I like the idea.

kolibri/utils/server.py
```diff
@@ -345,7 +348,14 @@ def installation_type():
             if launcher.endswith('.pex'):
                 install_type = 'PEX file'
             elif launcher == '/usr/bin/kolibri':
-                install_type = 'Debian package'
+                install_type = 'Debian package (Manual)'
+                try:
+                    apt_cache = str(check_output(['apt-cache', 'show', 'kolibri']))
+                except CalledProcessError:  # kolibri package not installed!
+                    apt_cache = ''
+                    install_type = 'Executable compiled from code'
```
benjaoming: It's still not compiled :)

kolibri/utils/server.py
```diff
@@ -345,7 +348,14 @@ def installation_type():
...
+                if '.deb' in apt_cache and 'ilename' in apt_cache:
+                    install_type = 'Debian package (using apt)'
```
benjaoming: It's interesting that this can be detected, but what are the circumstances of having `Filename:` in the apt cache? Will the `Filename` field always be in `apt-cache` even if the source isn't a PPA or installed from a local .deb?

Isn't it a safer bet to leave `apt-cache` out of the equation of something that's supposed to work with very simple means? I think that detecting a `VERSION_PKG` file is a good start, we could add that in the package..

kolibri/utils/server.py
jredrejo: I know, it's Python. Do you have a better idea on how to call it? Just executable? I want to diferentiate between the kolibri executable and running it using yarn.

kolibri/utils/server.py
jredrejo: Well, it's not unknown :)

kolibri/utils/server.py
jredrejo: Yes, I thought of that possibility when doing this modification yesterday and was pending to find a better way.  Detecting VERSION_PKG would be a better start if we had different versions, but that's not the case as far as I know. There are other fields in `apt-cache` that allow detecting this safely. I am implementing them.

kolibri/utils/server.py
jredrejo: I have changed it by 'Binary created after building the sources'

kolibri/utils/server.py
jredrejo: Done

kolibri/utils/server.py
jredrejo: I have redone it,

kolibri/utils/server.py
jredrejo: Renamed to Binary created after building the sources

kolibri/utils/server.py
jredrejo: I have changed the way detection is done for linux.. `start` is now considered only in one case. I think you were right when saying that start could be used after a Deb package was installed. The latest changes should solve that ambiguity.

kolibri/utils/server.py
benjaoming: It's not a "binary" neither. It's just "other" IMO.

kolibri/utils/server.py
```diff
@@ -322,3 +325,46 @@ def get_urls(listen_port=None):
...
+        - installed_pex:  A PEX file
+        - installed_dpkg: A Debian package installed manually
+        - installed_apt: A Debian package installed using apt (usually from a ppa)
+        - installed_windows: Windows Installer
+        - installed_whl: Binary created after building the sources
+        - installed_dev: Development server
+        - installed_unknown: this function has not been able to guess it
+    """
+    install_type = 'installed_unknown'
+    if len(sys.argv) > 1:
```
benjaoming: If it's run from uWSGI, I don't think we have any any executable at all. Could the logic build around this? I think you should consider the input that I gave about a `VERSION_PKG` file generated by the installers, that can alleviate a lot of the conditions and make it easier to mock+test code :)

benjaoming (dismissed):

kolibri/utils/server.py
benjaoming: But `apt-cache` can mis-fire if you have several installations, for instance for developers and testers. This will become confusing.

> Detecting VERSION_PKG would be a better start if we had different versions, but that's not the case as far as I know. 

We do! We may release several builds of the Debian and Windows installers after source releases. Have done this with the Debian installer... it's unfortunately a bit complicated so we do see issues!

kolibri/utils/server.py
benjaoming: @jredrejo I still think that all this logic pertains a kind of versioning of the distributed package. I think it should be moved to `kolibri.utils.version`

kolibri/utils/server.py
jredrejo: Hi, 
1. yes, for uWSGI it will say `unknown` but this PR also adds the info about the web server used, which will say uwsgi. 
2. About `VERSION_PKG`: we can add it when it exists, but it doesn't. In any case having it would replace 3 lines of code by 1 line of code, the 3 lines used to check if the package was installed manually or using apt. The lines to ensure the package is installed should not be removed. I don't think it's very important, but in any case, we can't use it until `VERSION_PKG` exists. I would even say that's not a warranty: nobody avoids you to download the package from the ppa repository and install it manually ;)

jredrejo:
@jamalex , @benjaoming  can you review the code again after the latest changes? In case you approve them I would add tests for the code. I didn't want to do it before agreeing on the implementation.

kolibri/plugins/device_management/assets/src/modules/deviceInfo/handlers.js
```diff
@@ -12,6 +12,10 @@ export function getDeviceInfo() {
     const data = response.entity;
     data.server_time = new Date(data.server_time);
     data.content_storage_free_space = bytesForHumans(data.content_storage_free_space);
+
+    if (response.headers.Server.includes('0.0.0.0')) data.server_type = 'Kolibri internal server';
+    else data.server_type = response.headers.Server;
+
```
indirectlylit, line 18: What are the possible values of this header?

kolibri/utils/server.py
benjaoming: > we can add it when it exists, but it doesn't

I will at some point mention that this change doesn't have any tests. Since all of the tests would require some level of mocking, I don't think that it's a problem to not have the file. If you move all this stuff to `kolibri.utils.version`, you will definitely see what I mean by mocking :smile:

kolibri/plugins/device_management/assets/src/modules/deviceInfo/handlers.js
radinamatic: I was wondering about the same...?

jredrejo: It depends on the used server: uwsgi, Apache, Nginx, each server shows a string with its name and the version. It would serve us to give the better helping instructions knowing the exact server type and version they have installed (if any). If they are running just  `kolibri start ...` it will show 'Kolibri internal server`

kolibri/utils/server.py
jredrejo: I don't think moving it to utils.version is a good idea, as this is utils.server information. Anyway I will do it if more people think it's better. In any case tests don't depend on the place where the functions are. I just want to know if the functionality is approved to do tests for them. I don't want to work on tests for functionalities that might be removed or changed before being merged.

kolibri/plugins/device_management/assets/src/modules/deviceInfo/handlers.js
benjaoming: @jredrejo it's a great idea! I can easily picture a simple user who's actually running IIAB or Rachel in future releases. This user will not be aware of the HTTP server that's in front of Kolibri.

radinamatic: Agreed, sounds like a very useful piece of information! 👍

kolibri/utils/server.py
indirectlylit: yes, let's pin down the expected behavior and afterwards we can add tests

indirectlylit:
@jredrejo let's put this on ice until we decide on the actual UI with the designers

jredrejo:
@indirectlylit ok, ping me back when decissions are taken to study if it's worth working on the tests.

jtamiace:
After feedback from @indirectlylit and @jamalex, here is a proposal for the UI 

device info 0 11

indirectlylit:
@jredrejo I'll submit a quick PR to you tomorrow to get the strings and layout it, and then you can flesh out the details

indirectlylit:
@jredrejo - https://github.com/jredrejo/kolibri/pull/1

indirectlylit (approved):
I'm going to merge this in for string freeze; @jredrejo you can add tests in a new PR.

thanks all!

radinamatic:
Darn, @indirectlylit is too fast for me... :joy: 

Looking good when installed locally, one minor issue only: when user downloads the text file and opens it in Notepad on Windows, it seems to be missing the hard end-of-the-line, which makes the items difficult to distinguish. Opening in Notepad++ or pasting from clipboard does not have this issue.

#### Windows 7

win7 start running - oracle vm virtualbox_194

#### Windows 10

win10 start running - oracle vm virtualbox_195

indirectlylit:
thanks @radinamatic - made a new issue here: https://github.com/learningequality/kolibri/issues/4315

benjaoming:
> @jredrejo you can add tests in a new PR

https://github.com/learningequality/kolibri/issues/4319

PR #4249, opened 2018-09-17
Inline nav items

rtibbles:
### Summary
#3559 created glorious, wonderful, pluggable side nav items.

However, the approach that was eventually taken resulted in many little webpack bundles being included into the template to support this.

Also, in spite of having no styling inside them, these curious little bundles also produced empty css files that were useless, but added an extra request to page load.

### Reviewer guidance
This PR attempts to resolve the above problems thusly:
* Firstly, it adds an `inline` property that allows JS and CSS for `WebpackFrontendHook`s to be inlined into the Django template.
* It applies this `inline` property to the `NavigationHook` class, so that any time this is mixed in with the `WebpackFrontEndHook`, then the nav items are inlined in the template.
* After some frustrating investigation, it became apparent that the empty css blocks were due to vue-loader insisting that there was some css, so this was resolved by removing the empty `<style>` blocks from the side nav components.

Caveats:
* This only affects production, development stays the same, as I didn't want to implement fetching the files from the webpack dev server.
* Applying the `inline` property to the `NavigationHook` class appears a slightly risky approach at first glance, as it depends which order the mixins are added to the child hooks - so this might need some cleanup.
* The inlining could break relative path references for files, so it might not be possible to use an image in a nav item or other inlined code.
* The inlining could be potentially detrimental to performance (I have attempted to mitigate this by caching the reading of static files from disk), as it increases the size of the template which is never cached by Django according to our current caching setup. In addition, these static files are normally served by CherryPy, and hence never normally have to traverse the Django server in production. This could be further mitigated once 0.10 is merged into develop, and we add additional caching logic to allow generated HTML to be better cached.

### References
Some of the code for accessing static files is modified from [django-compressor](https://django-compressor.readthedocs.io/en/latest/), a project we had previously used in KA Lite, but eventually removed.

As far as I can tell, it does not allow the intermixing of JS and CSS when compressing frontend assets, so would be harder to quickly drop in for our purposes, but given that there is no CSS currently for side nav items, it could potentially be used. However, when I tried an initial naive implementation, it gave errors.

----

### Contributor Checklist

### Reviewer Checklist

### DIFF
--- a/kolibri/core/webpack/hooks.py
+++ b/kolibri/core/webpack/hooks.py
@@ -18,7 +18,7 @@ import time
 from functools import partial
 
 from django.conf import settings as django_settings
-from django.contrib.staticfiles import finders
+from django.contrib.staticfiles.finders import find as find_staticfiles
 from django.contrib.staticfiles.storage import staticfiles_storage
 from django.core.cache import cache
 from django.utils.functional import cached_property
@@ -87,8 +87,8 @@ class WebpackBundleHook(hooks.KolibriHook):
     # : Kolibri version for build hashes
     version = kolibri.__version__
 
-    # When being included for synchronous loading, should the source files
-    # for this be inlined?
+    # : When being included for synchronous loading, should the source files
+    # : for this be inlined?
     inline = False
 
     def __init__(self, *args, **kwargs):
@@ -121,17 +121,17 @@ class WebpackBundleHook(hooks.KolibriHook):
         """
         global _JSON_STATS_FILE_CACHE
         try:
-            if not _JSON_STATS_FILE_CACHE.get(self.unique_slug) or django_settings.DEBUG:
+            if not _JSON_STATS_FILE_CACHE.get(self.unique_slug) or getattr(django_settings, 'DEVELOPER_MODE', False):
                 with io.open(self._stats_file, mode='r', encoding='utf-8') as f:
                     stats = json.load(f)
-                if django_settings.DEBUG:
+                if getattr(django_settings, 'DEVELOPER_MODE', False):
                     timeout = 0
                     while stats['status'] == 'compiling':
-                        time.sleep(getattr(settings, 'WEBPACK_POLL_INTERVAL', 0.1))
-                        timeout += getattr(settings, 'WEBPACK_POLL_INTERVAL', 0.1)
+                        time.sleep(0.1)
+                        timeout += 0.1
                         with io.open(self._stats_file, mode='r', encoding='utf-8') as f:
                             stats = json.load(f)
-                        if timeout >= getattr(settings, 'WEBPACK_POLL_INTERVAL', 1.0):
+                        if timeout >= 5:
                             raise WebpackError('Webpack compilation still in progress')
                     if stats['status'] == 'error':
                         raise WebpackError('Webpack compilation has errored')
@@ -155,7 +155,7 @@ class WebpackBundleHook(hooks.KolibriHook):
             if any(list(regex.match(filename) for regex in settings.IGNORE_PATTERNS)):
                 continue
             relpath = '{0}/{1}'.format(self.unique_slug, filename)
-            if django_settings.DEBUG:
+            if getattr(django_settings, 'DEVELOPER_MODE', False):
                 f['url'] = f['publicPath']
             else:
                 f['url'] = staticfiles_storage.url(relpath)
@@ -247,7 +247,7 @@ class WebpackBundleHook(hooks.KolibriHook):
     def frontend_messages(self):
         global _JSON_MESSAGES_FILE_CACHE
         lang_code = get_language()
-        if not _JSON_MESSAGES_FILE_CACHE.get(self.unique_slug, {}).get(lang_code) or django_settings.DEBUG:
+        if not _JSON_MESSAGES_FILE_CACHE.get(self.unique_slug, {}).get(lang_code) or getattr(django_settings, 'DEVELOPER_MODE', False):
             frontend_message_file = self.frontend_message_file(lang_code)
             if frontend_message_file:
                 with io.open(frontend_message_file, mode='r', encoding='utf-8') as f:
@@ -324,14 +324,15 @@ class WebpackBundleHook(hooks.KolibriHook):
         """
         filename = None
         # First try finding the file using the storage class.
-        # This is skipped in DEBUG mode as files might be outdated
-        if not django_settings.DEBUG:
+        # This is skipped in DEVELOPER_MODE mode as files might be outdated
+        # Or may not even be on disk.
+        if not getattr(django_settings, 'DEVELOPER_MODE', False):
             filename = staticfiles_storage.path(basename)
             if not staticfiles_storage.exists(basename):
                 filename = None
         # secondly try to find it with staticfiles
-        if not filename and finders:
-            filename = finders.find(url2pathname(basename))
+        if not filename:
+            filename = find_staticfiles(url2pathname(basename))
         return filename
 
     def get_filecontent(self, url):
--- a/kolibri/core/webpack/hooks.py
+++ b/kolibri/core/webpack/hooks.py
@@ -9,6 +9,7 @@ from __future__ import absolute_import
 from __future__ import print_function
 from __future__ import unicode_literals
 
+import codecs
 import io
 import json
 import logging
@@ -17,13 +18,16 @@ import time
 from functools import partial
 
 from django.conf import settings as django_settings
+from django.contrib.staticfiles import finders
 from django.contrib.staticfiles.storage import staticfiles_storage
 from django.utils.functional import cached_property
 from django.utils.safestring import mark_safe
+from django.utils.six.moves.urllib.request import url2pathname
 from django.utils.translation import get_language
 from django.utils.translation import get_language_info
 from django.utils.translation import to_locale
 from pkg_resources import resource_filename
+from six import text_type
 
 import kolibri
 from . import settings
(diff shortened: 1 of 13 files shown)


kolibri/core/webpack/hooks.py
```diff
@@ -82,6 +87,10 @@ class WebpackBundleHook(hooks.KolibriHook):
     # : Kolibri version for build hashes
     version = kolibri.__version__
 
+    # When being included for synchronous loading, should the source files
+    # for this be inlined?
```
benjaoming: If you write `#: ...`, the documentation will be picked up by Sphinx for API docs of the `inline` property. Otherwise, it's just ignored as a code comment.

kolibri/core/webpack/hooks.py
```diff
@@ -17,13 +18,17 @@
 from functools import partial
 
 from django.conf import settings as django_settings
+from django.contrib.staticfiles import finders
```
benjaoming: Oh this naming is sad... `finders.find` :) Could you do `from django.contrib.staticfiles.finders import find as staticfiles_find` ?

kolibri/core/webpack/hooks.py
```diff
@@ -17,13 +18,17 @@
...
 from django.contrib.staticfiles.storage import staticfiles_storage
+from django.core.cache import cache
```
benjaoming: Maybe not relevant yet, but Django can have different caches defined. This is interesting because caches have different sizes, invalidation rules etc. So we _could_ have one especially for cached assets.

benjaoming: ...which was then invalidated on every initialization in DEBUG mode and at every detected version change.

benjaoming (approved):
I only have suggestions for improving a couple of things, but nothing blocking.

benjaoming (dismissed):
Okay, maybe... would really like a test because coverage of this code is 27%.

rtibbles:
Fair enough, not needed for string freeze, so this can hold a while.

kolibri/core/webpack/hooks.py
rtibbles: Yeah, the only slight issue that I have with using debug mode in this way is that it conflates "building assets dynamically with webpack" with "in debug mode", which are different. Can do a cheeky refactor to see if I can decouple these.

kolibri/core/webpack/hooks.py
rtibbles: OK, but it was django what done it!

benjaoming: I know, nothing in this world is perfect, but Django is almost :)

indirectlylit: >  Django is almost :)

oh my...

kolibri/core/webpack/hooks.py
```diff
@@ -256,12 +265,25 @@ def sorted_chunks(self):
...
+                else:
+                    yield js_tag.format(url=chunk['url'])
             elif chunk['name'].endswith('.css'):
-                yield css_tag.format(url=chunk['url'])
+                if self.inline:
+                    src = self.get_filecontent(chunk['url'])
+                if src is not None:
+                    yield inline_css_tag.format(src=src)
+                else:
+                    yield css_tag.format(url=chunk['url'])
```
indirectlylit, line 296: Mind adding some comments within each of these conditionals?

In particular, it's not clear to me why `self.inline` would be `True`, but `src` might be `None` - that seems like an error?

indirectlylit (commented):
Note, I pushed couple updates per @benjaoming's comments.

I'm trying to test this locally and encountering odd behavior: compilation seems to complete and I see

```
ℹ ｢wdm｣: Child media_player_module:
       54 modules
Child user_module_login_nav_side_nav:
       10 modules
Child document_pdf_render_module:
       161 modules
Child default_frontend:
       1371 modules
Child style_guide_module:
       408 modules
Child learn_module_side_nav:
       9 modules
Child coach_module:
       857 modules
Child facility_management_side_nav:
       10 modules
Child user_module:
       217 modules
Child user_module_user_profile_nav_side_nav:
       10 modules
Child setup_wizard:
       244 modules
Child learn_module:
       635 modules
Child html5_app_renderer_module:
       70 modules
Child facility_management_module:
       461 modules
Child device_management_side_nav:
       10 modules
Child device_management_module:
       548 modules
Child coach_side_nav:
       10 modules
ℹ ｢wdm｣: Compiled successfully.
```

But then I get

```


ERROR    Internal Server Error: /user/
  File "/Users/d/Projects/learning_equality/repos/k011x/.venv/lib/python2.7/site-packages/django/core/handlers/exception.py", line 41, in inner
  File "/Users/d/Projects/learning_equality/repos/k011x/.venv/lib/python2.7/site-packages/django/core/handlers/base.py", line 217, in _get_response
  File "/Users/d/Projects/learning_equality/repos/k011x/.venv/lib/python2.7/site-packages/django/core/handlers/base.py", line 215, in _get_response
    response = response.render()
  File "/Users/d/Projects/learning_equality/repos/k011x/.venv/lib/python2.7/site-packages/django/template/response.py", line 107, in render
    self.content = self.rendered_content
  File "/Users/d/Projects/learning_equality/repos/k011x/.venv/lib/python2.7/site-packages/django/template/response.py", line 84, in rendered_content
    content = template.render(context, self._request)
  File "/Users/d/Projects/learning_equality/repos/k011x/.venv/lib/python2.7/site-packages/django/template/backends/django.py", line 66, in render
    return self.template.render(context)
  File "/Users/d/Projects/learning_equality/repos/k011x/.venv/lib/python2.7/site-packages/django/template/base.py", line 207, in render
    return self._render(context)
  File "/Users/d/Projects/learning_equality/repos/k011x/.venv/lib/python2.7/site-packages/django/template/base.py", line 199, in _render
    return self.nodelist.render(context)
  File "/Users/d/Projects/learning_equality/repos/k011x/.venv/lib/python2.7/site-packages/django/template/base.py", line 990, in render
    bit = node.render_annotated(context)
  File "/Users/d/Projects/learning_equality/repos/k011x/.venv/lib/python2.7/site-packages/django/template/base.py", line 957, in render_annotated
    return self.render(context)
  File "/Users/d/Projects/learning_equality/repos/k011x/.venv/lib/python2.7/site-packages/django/template/loader_tags.py", line 177, in render
    return compiled_parent._render(context)
  File "/Users/d/Projects/learning_equality/repos/k011x/.venv/lib/python2.7/site-packages/django/template/base.py", line 199, in _render
    return self.nodelist.render(context)
  File "/Users/d/Projects/learning_equality/repos/k011x/.venv/lib/python2.7/site-packages/django/template/base.py", line 990, in render
    bit = node.render_annotated(context)
  File "/Users/d/Projects/learning_equality/repos/k011x/.venv/lib/python2.7/site-packages/django/template/base.py", line 957, in render_annotated
    return self.render(context)
  File "/Users/d/Projects/learning_equality/repos/k011x/.venv/lib/python2.7/site-packages/django/template/defaulttags.py", line 322, in render
    return nodelist.render(context)
  File "/Users/d/Projects/learning_equality/repos/k011x/.venv/lib/python2.7/site-packages/django/template/base.py", line 990, in render
    bit = node.render_annotated(context)
  File "/Users/d/Projects/learning_equality/repos/k011x/.venv/lib/python2.7/site-packages/django/template/base.py", line 957, in render_annotated
    return self.render(context)
  File "/Users/d/Projects/learning_equality/repos/k011x/.venv/lib/python2.7/site-packages/django/template/loader_tags.py", line 72, in render
    result = block.nodelist.render(context)
  File "/Users/d/Projects/learning_equality/repos/k011x/.venv/lib/python2.7/site-packages/django/template/base.py", line 990, in render
    bit = node.render_annotated(context)
  File "/Users/d/Projects/learning_equality/repos/k011x/.venv/lib/python2.7/site-packages/django/template/base.py", line 957, in render_annotated
    return self.render(context)
  File "/Users/d/Projects/learning_equality/repos/k011x/.venv/lib/python2.7/site-packages/django/template/library.py", line 203, in render
    output = self.func(*resolved_args, **resolved_kwargs)
  File "kolibri/core/content/templatetags/content_tags.py", line 37, in content_renderer_assets
    return webpack_asset_render(hooks.ContentRendererHook, async=True)
  File "kolibri/core/webpack/utils.py", line 24, in webpack_asset_render
    hook.render_to_page_load_async_html()
  File "kolibri/core/content/hooks.py", line 94, in render_to_page_load_async_html
    urls = [chunk['url'] for chunk in self.bundle]
  File "kolibri/core/webpack/hooks.py", line 153, in bundle
    for f in self._stats_file_content["files"]:
  File "/Users/d/Projects/learning_equality/repos/k011x/.venv/lib/python2.7/site-packages/django/utils/functional.py", line 35, in __get__
    res = instance.__dict__[self.name] = self.func(instance)
  File "kolibri/plugins/hooks.py", line 203, in inner
    return func(instance, *args, **kwargs)
  File "kolibri/core/webpack/hooks.py", line 135, in _stats_file_content
    raise WebpackError('Webpack compilation still in progress')
WebpackError: Webpack compilation still in progress
ERROR    "GET /user/ HTTP/1.1" 500 25763
ERROR:django.server:"GET /user/ HTTP/1.1" 500 25763
```

rtibbles:
Note, this still doesn't have tests, but all other comments have been addressed.

indirectlylit:
@jonboiser would you have bandwidth to test this?

indirectlylit:
this looks like an improvement - the CSS blocks are no longer there. But seems the JS is still not inlined?

image

indirectlylit:
(note, I'm no longer running into the weird webpack issue)

indirectlylit (approved):
merging in for integration testing; filed follow-up issue here: #4366

PR #4252, opened 2018-09-17
Allow unicode usernames

### LINKED ISSUE #4205: Cannot create non-ascii usernames
### Observed behavior

I tried to create a user with username `иван` (ivan in cyrillic) but was not able to.

### Expected behavior

Create a new user  OR  show validation error.

### User-facing consequences

Cannot create usernames in other locales.

### Errors and logs

UI doesn't show a validation error, but JS console has this:

screen shot 2018-08-30 at 5 01 06 pm

### Steps to reproduce

  - Go to localhost:8080/facility/#/users
  - Choose ADD USER
  - Enter a non-ascii username (and valid password etc)
  - Click ADD

Actual: nothing

Expected: user created or validation error "username ... invalid; use ascii characters"

### Context

 * Kolibri version: 0.10.3.dev
 * Operating system: Linux
 * Browser: Firefox

rtibbles:
### Summary
Previously our backend disallowed unicode usernames because our validator used `^\w+$` - on the frontend we used an inverse test to exclude specific characters ```[\s\`~!@#$%^&*()\-+={}\[\]\|\\\/:;"'<>,\.\?]```. This replicates the frontend regex on the backend.

It also updates the account signup page to properly leverage the new error codes being returned in our more standardized validation API.

image

### Reviewer guidance
Can you make a user with a username `иван`?

### References
Fixes #4205

----

### Contributor Checklist

### Reviewer Checklist


kolibri/core/assets/src/constants.js
```diff
@@ -103,6 +103,11 @@ const ERROR_CONSTANTS = {
   NOT_FOUND: 'NOT_FOUND',
   INVALID_CREDENTIALS: 'INVALID_CREDENTIALS',
   MISSING_PASSWORD: 'MISSING_PASSWORD',
+  USER_ALREADY_IN_GROUP_IN_CLASS: 'USER_ALREADY_IN_GROUP_IN_CLASS',
+  MAX_LENGTH: 'MAX_LENGTH',
+  INVALID: 'INVALID',
+  PERMISSION_DENIED: 'PERMISSION_DENIED',
+  NOT_AUTHENTICATED: 'NOT_AUTHENTICATED',
```
indirectlylit, line 110: some new constants are added in this PR, but it's not clear to me why?

It doesn't look like they're thrown by the backend or handled by the frontend

rtibbles: I added INVALID, because it is thrown during the validation process, to explicitly handle that. I added the others in order to mirror the list of core python error messages that have been defined, with the expectation that these should be kept in lockstep.

ivanistheone (approved):
Tested two paths:

1. ADD USER from http://localhost:8000/facility/#/users 
2. CREATE AN ACCOUNT http://localhost:8000/user/#/create_account 

using unicode usernames and password and both worked.

kolibri/core/assets/src/constants.js
rtibbles: Going to merge this PR and submit a clean up PR, as will looking into this more, I found some of the backend constants being referenced in plain text in the frontend, which goes against our usual practices.

PR #4254, opened 2018-09-17
new crowdin tooling and custom font generation

indirectlylit:
### Summary

* removes the need for the java crowdin client
* automates many crowdin tasks that needed to be performed manually before
* only downloads and uploads supported languages
* generates custom Noto fonts based on the app text
* introduces sophisticated font loading process
  * custom font subsets are loaded inline, and available immediately for all browsers
  * for newer browsers, all fonts for all languages are available on-demand
  * for older browsers, fonts for the current language are loaded asynchronously
  * for all browsers, we avoid flashes of invisible text, ensuring rapid readability
* introduces two variants of Noto: a UI font with constrained vertical size for use in UI widgets, and a Content font with unconstrained vertical size, for use in text blocks and paragraphs

To do in a follow-up PR:

* unit tests for new scripts
* load `VueIntl` locale data from a `<link>` like the CSS
* ensure Django has static file caching and cache busting working
* Look at building the CSS files during `make assets` rather than checking them into git
* introduce content font in more places in the app

### Reviewer guidance

Some specific things to look at:

* updated i18n documentation and Makefile commands
* sustainability of process
* handling of locale codes

### References

| Before | After |
|--|--|
| image  | image |
| image | image |

N/A

----

### Contributor Checklist

### Reviewer Checklist


indirectlylit:
This is very close to finished. I'm still struggling a little with font-face unicode ranges to prevent fonts from being download when they're not necessary.

Besides that important update, I still want to:

* load `VueIntl` locale data from a `<link>` like the CSS
* finish the documentation updatse
* ensure Django has static file caching and cache busting working
* Look at building the CSS files during `make assets` rather than checking them into git

indirectlylit:
This is ready for initial review. Still to-do:

* finish documentation and Makefile changes
* incorporate 'content' font instead of 'ui' font in places where we have paragraphs of text

kolibri/core/assets/src/utils/i18n.js
```diff
@@ -26,7 +29,7 @@ export const languageDirections = {
   RTL: 'rtl',
 };
 
-const defaultLocale = 'en';
```
rtibbles: Why this change?

kolibri/core/assets/src/utils/i18n.js
```diff
@@ -154,16 +157,13 @@ export function createTranslator(nameSpace, defaultMessages) {
...
-export function setUpVueIntl() {
+function _setUpVueIntl() {
   /**
    * Use the vue-intl plugin.
    **/
   const VueIntl = require('vue-intl');
-
-  vue.use(VueIntl, { defaultLocale });
-
+  vue.use(VueIntl, { DEFAULT_LOCALE });
```
rtibbles: This is not a valid option to VueIntl, it takes two options on initialization: `defaultLocale` and `defaultFormats`, `DEFAULT_LOCALE` is not a valid option.

kolibri/core/assets/src/utils/i18n.js
```diff
@@ -175,25 +175,82 @@ export function setUpVueIntl() {
...
 
-  if (global.languageCode) {
-    currentLanguage = global.languageCode;
-    vue.setLocale(currentLanguage);
-    setLanguageDensity(currentLanguage);
+  vue.setLocale(currentLanguage);
+  if (global.coreLanguageMessages) {
+    vue.registerMessages(currentLanguage, global.coreLanguageMessages);
+  }
+  importVueIntlLocaleData().forEach(localeData => VueIntl.addLocaleData(localeData));
```
rtibbles, line 185: We don't need to add the locale data for every locale, only the current locale.

rtibbles: (sorry, missed this in the previous review of this change).

kolibri/core/assets/src/utils/i18n.js
```diff
@@ -154,16 +157,13 @@ export function createTranslator(nameSpace, defaultMessages) {
   return new Translator(nameSpace, defaultMessages);
 }
 
-export function setUpVueIntl() {
+function _setUpVueIntl() {
   /**
```
rtibbles, line 161: Would be good to note that this should only be initialized after the function that comes after it, because that is where the `currentLanguage` gets set.

kolibri/core/assets/src/utils/i18n.js
```diff
@@ -175,25 +175,82 @@ export function setUpVueIntl() {
...
+      htmlEl.classList.add(FULL_FONTS);
+      logging.debug(`Loaded full font for '${currentLanguage}'`);
+    })
+    .catch(function() {
+      logging.warn(`Could not load full font for '${currentLanguage}'`);
+    });
 }
 
-export function setUpIntl() {
+export function i18nSetup(skipPolyfill = false) {
```
rtibbles, line 236: I can't find any place where this optional arg gets used?

kolibri/core/public/api.py
```diff
@@ -54,13 +54,24 @@ def _get_channel_list_v1(params, identifier=None):
...
         channels = ChannelMetadata.objects.all()
 
-    if keyword != '':
-        channels = channels.filter(Q(name__icontains=keyword) | Q(description__icontains=keyword))
-
-    if language_id != '':
-        matching_tree_ids = ContentNode.objects.prefetch_related('files') \
-                            .filter(Q(lang__id__icontains=language_id) | Q(files__lang__id__icontains=language_id)).values_list('tree_id', flat=True)
-        channels = channels.filter(Q(root__lang__id__icontains=language_id) | Q(root__tree_id__in=matching_tree_ids))
+    if keyword != "":
```
rtibbles, line 57: Drive by linting, or did something change in here?

kolibri/core/assets/src/utils/vue-intl-locale-data.js
```diff
@@ -7,10 +7,11 @@
 module.exports = function() {
   const data = [];
   data.push(require('vue-intl/locale-data/ar.js'));
-  data.push(require('vue-intl/locale-data/bn.js'));
```
rtibbles, line 10: As I mentioned above, maybe this could be an object, rather than an array, and just register the one we need for the current locale?

rtibbles (commented):
A few questions. Manual testing seems fine - we could probably optimize the timing/position of the default font loading, but it's probably not a huge issue currently!

.flake8
```diff
@@ -0,0 +1,3 @@
+[flake8]
```
benjaoming: This should be added in `setup.cfg`, saves the additional file.

Not sure if we want to ignore these two errors globally? What's the reason?

build_tools/i18n/crowdin.py
```diff
@@ -0,0 +1,396 @@
+# -*- coding: utf-8 -*-
```
benjaoming, line 1: This looks like a tool that should be open sourced ;)

build_tools/i18n/crowdin.py
```diff
@@ -0,0 +1,396 @@
...
+"""
+For usage instructions, see:
+    https://kolibri-dev.readthedocs.io/en/develop/references/i18n.html
```
benjaoming, line 4: If you use docopt, it's slightly easier to get documentation, CLI instructions and parsing in one go.

kolibri/core/assets/src/keen-config/variables.scss
```diff
@@ -1,3 +1,11 @@
-$brand-primary-color: #996189;
-$font-stack: 'Noto Sans';
+@import '~kolibri.styles.definitions';
+
+$brand-primary-color: $core-action-normal;
```
benjaoming, line 3: What's "$core-action-normal"? It doesn't sound like a color code :)

benjaoming (commented):
I'm wondering if the work done by `fonts.py` can just be done once? Can we generate all the subsets and check them into the repo or is it a necessary step for every build?

benjaoming:
Excited to see the work on a Python-based CrowdIn API interaction. I didn't feel much like doing this work earlier, so am glad you didn't shy away :)

Could we move this into a separate project, though? Benefits: 1) Can reuse in other repos 2) Others can help us maintain it. 3) We exchange some goodwill with CrowdIn folks who are really helpful towards Kolibri :)

kolibri/core/assets/src/utils/i18n.js
indirectlylit: drive-by cleanup, setting module constants to capital-case. will revert

kolibri/core/assets/src/utils/i18n.js
indirectlylit: Yeah, this was the change where I loaded everything synchronously because it was faster than loading one thing async.

In a future PR I'd like to the data from a `<link>` like I'm doing with the CSS

kolibri/core/assets/src/utils/i18n.js
indirectlylit: added comment

kolibri/core/assets/src/utils/i18n.js
indirectlylit: good catch - just updated.

This was intended for the Jest setup: https://github.com/learningequality/kolibri/pull/4254/files#diff-f65f9af04c968c709f187a673e5e91c3L15

.flake8
indirectlylit: I've been using the [black code formatter](https://github.com/ambv/black) on this code.

Black is fully PEP8 compliant, whereas flake8 is not for rules E203 and W503: https://github.com/ambv/black#line-breaks--binary-operators

kolibri/core/public/api.py
indirectlylit: I'm not entirely sure what happened, but I was getting linting errors from this and a few other lines.

I think it had to do with this change: https://github.com/learningequality/kolibri/pull/4254/commits/8e385ce98aa790d7777bc96561719c455d804513

which was done in order to use the `black` code formatter: https://github.com/learningequality/kolibri/pull/4254/commits/8e385ce98aa790d7777bc96561719c455d804513#r223027684

kolibri/core/assets/src/keen-config/variables.scss
indirectlylit: originally it was the normal color associated with UI elements that trigger an action; our color variables are admittedly in need of refactor though.

.flake8
indirectlylit: > This should be added in setup.cfg, saves the additional file.

thanks!

indirectlylit:
> I'm wondering if the work done by fonts.py can just be done once? Can we generate all the subsets and check them into the repo or is it a necessary step for every build?

It's currently set up to do the former: generate manually and then check them in.

However I do think it might be smart to eventually move it to part of asset building so that we don't pollute our repo with frequently-changing binary font files.

kolibri/core/assets/src/utils/vue-intl-locale-data.js
indirectlylit: follow-up issue: https://github.com/learningequality/kolibri/issues/4364

kolibri/core/assets/src/utils/i18n.js
indirectlylit: follow-up issue: https://github.com/learningequality/kolibri/issues/4364

build_tools/i18n/crowdin.py
indirectlylit: follow-up issue: https://github.com/learningequality/kolibri/issues/4364

indirectlylit:
> Could we move this into a separate project, though?

I think that would be cool, but certainly not as part of this PR. Currently it's very much tied to our own process, and implements a small subset of their API - it's not at all a generic crowdin client.

build_tools/i18n/crowdin.py
indirectlylit: see note below - https://github.com/learningequality/kolibri/pull/4254#issuecomment-427394100

indirectlylit:
>  we could probably optimize the timing/position of the default font loading, but it's probably not a huge issue currently!

I'm curious what you have in mind? (This is something that I spent a little while working on in this PR)

indirectlylit:
@benjaoming @rtibbles tried to address comments or migrate follow-up to #4364

let me know if there's anything else you think is important to address in here

benjaoming:
> I think that would be cool, but certainly not as part of this PR. Currently it's very much tied to our own process, and implements a small subset of their API - it's not at all a generic crowdin client.

Will suggest this:

Once it's necessary to use this code in one of the other projects (docs mainly), let's move our Python-based CrowdIn client to a separate project and release it on PyPi. We'll only maintain features that we need. Others can scratch their own itches, and we can tell CrowdIn that they are also very welcome to jump in.

benjaoming:
Should `frontend_build` be ignored in test coverage?

rtibbles:
> I'm curious what you have in mind? (This is something that I spent a little while working on in this PR)

At the moment, the font loading only starts after the default frontend JS has been loaded and executed (if you look at the waterfall on the network tab, there is a fairly large gap between the last JS asset being loaded, and the fonts starting to load), so if we could shift some of this font loading out of the default frontend it could be faster. That said, it's not clear that it matters, so not worried about it for now!

benjaoming (approved):
Solid work, and glad to see that these sort of complexities are buried in our build tools. This isn't a given, often stuff like this starts to enter the actual source code.. known from KA Lite :)

indirectlylit:
> At the moment, the font loading only starts after the default frontend JS has been loaded and executed (if you look at the waterfall on the network tab, there is a fairly large gap between the last JS asset being loaded, and the fonts starting to load), so if we could shift some of this font loading out of the default frontend it could be faster. That said, it's not clear that it matters, so not worried about it for now!

Yeah, could be nice to move that earlier.

However note that by that point, the glyphs necessary to render all application text in the current language have already been loaded inline, so only text like topic descriptions might not be styled using Noto at that point.

indirectlylit:
> if we could shift some of this font loading out of the default frontend it could be faster.

@rtibbles, added this item to the follow-up issue.

merging in now to get it in for integration testing; feel free to follow up with additional feedback

PR #2881, opened 2017-12-09
Error checking for base-url.

ralphiee22:
### Summary

Validate url that is passed in through command line. Check if `instance_ids` are equal (by querying a morango information endpoint) which means a device is syncing with itself. Fix issue where it prompts for username again.

### Reviewer guidance

…

### References

…

----

### Contributor Checklist

- [x] PR has the correct target milestone
- [x] PR has 'needs review' or 'work-in-progress' label
- [x] If PR is ready for review, it has been assigned or requests review from someone and labeled as 'needs review'
- [x] PR has been fully tested manually
- [x] External dependencies files were updated (`yarn` and `pip`)
- [x] Link to diff of internal dependency change is included

### Reviewer Checklist

- [ ] Automated test coverage is satisfactory
- [ ] PR has been fully tested manually

PR #4276, opened 2018-09-18
basic kolibri manage command for testing server performance

### LINKED ISSUE #4270: add basic kolibri manage command for testing server performance
### Observed behavior

We often get reports of performance issues in the field, but it can be challenging to debug with remote users in other timezones with limited connectivity.

### Expected behavior

Let's add a diagnostics command that can be run to help us debug performance issues on peoples' servers. We can request that people run the command when they're experiencing poor performance, and then send us the output.

Some statistics we might include:

* Number of logged-in users
* CPU load
* total and used RAM 
* OS process list
* metrics about the time to hit some test endpoints:
  * a basic ping
  * a processor-intensive command
  * a DB-intensive command
* time to download a large static file (can this be generated so it doesn't need to be distributed?)

If there are other things that would be helpful to include, let's do so. Conversely, things above seem too difficult, let's skip them.

### User-facing consequences

easier to help us debug performance issues

jredrejo:
### Summary
This PR creates two new commands that should help to get performance information of the machine where kolibri is being run. The new commands are called `profile` and `benchmark`

##### profile command

The command can be executed, currently, in two ways:

- `kolibri manage profile`
- `kolibri manage profile --num_samples=60`

In both cases the information is written in the file KOLIBRI_HOME/performance/date_time_performance.csv. 
`--num_samples` is the number of times the information is collected and saved, every 10 seconds. 60 is the default value, thus the execution will last 10 minutes if num_samples is not changed.

Each log line contains, in this order, this information:
- Date and time of the command execution
- Number of Kolibri active sessions (including guest sessions)
- Number of Kolibri logged users
- Number of Kolibri users that have done some interaction in the last minute
- Total percentage of use of the CPU
- Total used memory
- Total available memory
- Number of processes executed in the server
- Percentage of cpu used by Kolibri
- Percentage of memory used by Kolibri

Also, while the command is being executed, the Kolibri server will profile every request, writing the information on a second file at KOLIBRI_HOME/performance/date_time_requests_performance.csv
For each request, one line with this information is saved:
- Timestamp
- Request path
- Time spent processing the request
- Memory (in Kbytes) used by the kolibri process when the request came in
- Memory (in Kbytes) used by the kolibri process when the response was sent
- Percentage of use of cpu by the Kolibri process when the request came in
- Percentage of use of cpu by the Kolibri process when the response was sent
- One flag indicating if this request is the slowest since the analysis was started

##### benchmark command

To execute it 
`kolibri manage benchmark`

It will ouput grouped information about the server where Kolibri is being run and the resources kolibri is using. It must be considered as a snapshot of an instant, results can vary depending on the load the server has. 
benchmark

### Reviewer guidance
This is a finished work. It supports Windows and Linux, but not MacOs servers.

Some things that should be take into account:
- Profile of server requests will be enabled only if the option `PROFILE` in the `[Server]` section of options.ini is defined to True. `profile`command will work if this option is not set, but it will provide only server information logs, the requests logs won't be created or updated.
- In case profiling is enabled, while the profile command is running a lock file will be created in KOLIBRI_HOME, so every Kolibri thread/worker will be able to detect it and profile its requests.
- Benchmark command uses a few available urls that don't require authentication. The timing output is not very representative. We could think of creating some endpoints doing intensive cpu/database work to get a better idea on how comfortable Kolibri is in the server ¿ideas?

### References
Resolves #4270 

### Contributor Checklist

### Reviewer Checklist

- [x] PR has been tested for [accessibility regressions](http://kolibri-dev.readthedocs.io/en/develop/manual_testing.html#accessibility-a11y-testing)
- [x] Contributor is in AUTHORS.rst


kolibri/core/analytics/management/commands/profile.py
```diff
@@ -0,0 +1,78 @@
...
+            logger.error('Database unavailable, impossible to retrieve users and sessions info')
+
+        # Machine usage information
+        used_cpu = str(psutil.cpu_percent())
+        used_memory = str(psutil.virtual_memory().used / pow(2, 20))  # In Megabytes
+        free_memory = str(psutil.virtual_memory().available / pow(2, 20))  # In Megabytes
+        total_processes = str(len(psutil.pids()))
+
+        # Kolibri usage information
+        kolibri_pid = None
```
ralphiee22: should be possible to directly get the PID of the running kolibri instance from a file
https://github.com/learningequality/kolibri/blob/develop/kolibri/utils/server.py#L71

jredrejo: Yes, I will change it if I am able to build a multiplatform solution for the whole command not using psutil. If finally I have to use psutil, it might be better to keep it as it is to avoid importing another module (os)

benjaoming: Yup, reliable mechanism should be to use `~/.kolibri/server.pid`. If this file isn't reliable, then we have other issues :)

Notice also about upcoming UWSGI instances, that profiling isn't trivial and it would be nice if this code in general kind of made explicit where it isn't going to work. Or maybe we should put big disclaimer for the command in the docstring?

kolibri/core/analytics/management/commands/profile.py
```diff
@@ -0,0 +1,78 @@
...
+logger = logging.getLogger('profiler')
+
+
+class Command(BaseCommand):
+    help = "Logs performance/profiling info in the server running Kolibri"
+
+    def add_arguments(self, parser):
+        parser.add_argument(
+            '--interval', action='store', dest='interval', default=0, type=int,
+            help='Specifies the interval (in seconds) to run the process continuosly. If 0, no repetition will happen',
```
benjaoming: It would be nice with some idea of what the desired interval is? I wouldn't know whether to put logs every second or every minute. Is the use case something like:

1) User reports slow server
2) We ask for a performance log, like `kolibri manage profile --interval=X`
3) User uploads performance log

benjaoming:
Great idea to use logging and put the log together with the other logs in a separate file!

jredrejo:
This is the approach I am working on. I have not finished it yet , but I want  to share it so it can be stopped if it's not considered a good approach:
1. Requests have to be measured by the kolibri server, the can't be measured by a command.
2. `kolibri manage profile` will make a request to the kolibri server, to a route _/api/perfrequests/PID_, where pid is the PID of the `profile`command being executed
3. the _/api/perfrequests/PID_ request will activate a middleware that will:
    1. check the process with the provided pid is running
    2. measure: 
        - cpu load after every request execution
        - time needed for every request execution
        - memory difference between the start 
    3. log in a conf.KOLIBRI_HOME/performance_requests.log the measured data
    4. stops as soon as the process with the PID is not active. i.e. when the `kolibri manage profile` is stopped 

I am considering the possibility of logging only the requests that consume more time than the 10 slowest, but that would exclude the info of small/ping requests.

benjaoming:
@jredrejo - I like that structure!

The TCP port of the currently running instance is available in `server.pid` btw.. see the implementation of `kolibri status`.

kolibri/core/analytics/management/commands/profile.py
indirectlylit: Sounds like the intent is that most users will run with the default of 0?

I agree that it's not clear to the user what an appropriate interval is.

I think having no options is better. If there are certain metrics that we believe fluctuate rapidly, we should take multiple samples and report the averages.

kolibri/core/analytics/management/commands/profile.py
```diff
@@ -0,0 +1,82 @@
...
+                kolibri_proc = psutil.Process(kolibri_pid)
+                kolibri_mem = str(kolibri_proc.memory_info().vms / pow(2, 20))
+                kolibri_cpu = str(kolibri_proc.cpu_percent())
+            except psutil.NoSuchProcess:
+                pass  # Kolibri server is not running
+
+        collected_information = (active_sessions, active_users, active_users_minute, used_cpu,
+                                 used_memory, free_memory, total_processes, kolibri_cpu, kolibri_mem)
+        logs = ','.join(collected_information)
+        return logs
```
indirectlylit: minor comment, but there's no need to have the `logs` variable. Can just do:

```python
return ", ".join(collected_information)
```

----

More importantly though...

I don't think that this output format is appropriate for our use-case, which is generally to paste into the forums or a github issue:

```
INFO:profiler:0,0,0,33.6,12393.09765625,5876.06640625,351,None,None
```

My recommendation would be to format it as a tabulated list of data, e.g.:


```
INFO:profiler

Sessions
* Active sessions (10 min):      10
* Active sessions (1 min):       7
* Active users in (10 min):      6
* Active users in (1 min):       3

CPU
* Total processes:               351
* Used CPU:                      33.6%
* Kolibri CPU usage:             22.3%

Memory
* Used memory:                   9.3 GB
* Total memory:                  16.0 GB
* Kolibri memory usage:          56.8 MB
```

A couple notes:

* the format above would display nicely as both Markdown-formatted text or monospace text, so it will paste very easily into the forums and github issues
* formatted numbers with consistent precision and units

A final thought... perhaps it would make sense to also include the information you recently added to the device info page, e.g.

```
Device info
* Version:                       (version)
* OS:                            (os)
* Installer:                     (installer)
* Server:                        (server_type)
* Database:                      (database_path)
* Device name:                   (device_name)
* Free disk space:               (content_storage_free_space)
* Server time:                   (server_time)
* Server timezone:               (server_timezone)
```

That way a single `kolibri manage profile` gives us all the relevant info in one place?

kolibri/core/analytics/management/commands/profile.py
```diff
@@ -0,0 +1,82 @@
...
+        except ValueError:
+            pass  # corrupted Kolibri PID file
+
+        if kolibri_pid:
+            try:
+                kolibri_proc = psutil.Process(kolibri_pid)
+                kolibri_mem = str(kolibri_proc.memory_info().vms / pow(2, 20))
+                kolibri_cpu = str(kolibri_proc.cpu_percent())
+            except psutil.NoSuchProcess:
+                pass  # Kolibri server is not running
```
indirectlylit: For the first pass of this command, I think we should require that the server be running and exit with an error if its not.

The command currently outputs misleading results and no warnings when Kolibri isn't running.

kolibri/core/analytics/management/commands/profile.py
```diff
@@ -0,0 +1,82 @@
+import logging
+import time
+from datetime import timedelta
+
+import psutil
```
indirectlylit: need to update `requirements.txt`

indirectlylit:
> Output as a log, with csv format, is it fine?

See comment above: I believe the format should be amenable to human reading, and the most common method of sharing will be to copy/paste into forums or github issues.

> Do we need the complete list of the processes names that are running in the server (it will have several hundred of processes in a normal Linux machine)

No

> Options to add to the command, if any

As noted above, the `interval` option is a bit confusing and should be removed.

If we think that some of the reported metrics are highly variable and should be smoothed over a small window with multiple samples, we should simply do that that behind the scenes and report the final averages as the results.

> Information to be added/removed from the log

Some information on content currently available would be useful - e.g. total number of content nodes and total number of channels.

> Requests have to be measured by the kolibri server, the can't be measured by a command.

Could you elaborate? I don't quite understand this

> `kolibri manage profile` will make a request to the kolibri server, to a route _/api/perfrequests/PID_, where pid is the PID of the `profile` command being executed ... check the process with the provided pid is running ... stop when `kolibri manage profile` is stopped

I don't quite understand what the goal is here. It sounds like we're trying to somehow synchronize the two processes and get information about the effects on CPU and memory of triggering the requests...?

For a first pass, I feel like this could be much simpler. Something like:

1. We create three new profiling endpoints:
    * _/api/profile/read_ does just DB-reads, perhaps loading the metadata of all current channels
    * _/api/profile/write_ does some DB-writes, perhaps creating a dummy user, logging some activity, and deleting it
    * _/api/profile/no-op_ returns immediately
2. `kolibri manage profile` makes a few requests to each of the endpoints, times how long they take, and reports the average times

This would at least give us some baseline information about how quickly the database and app server are able to respond.

kolibri/core/analytics/management/commands/profile.py
jredrejo: Yes, the intent was more users would run it only once.
Maybe we should change it to be running for 30 or 60 seconds by default. Specially thinking of the request metrics, they need requests to be done for some time.

kolibri/core/analytics/management/commands/profile.py
jredrejo: Yeah, I agree. I will change it.

kolibri/core/analytics/management/commands/profile.py
jredrejo: If we use the log format, adding that tabulated information can not have sense. We should decide if we either go for a log format so we can see how the performance evolves in time, or we just generate a tabulated report (or even adding this second part as an option).
From a developer point of view, who can receive the file provided by the users to find out where problems are,  I consider more util having the log format because the timeline provides more feedback. Nevertheless some other people can prefer the report.

kolibri/core/analytics/management/commands/profile.py
jredrejo: Yes, but not yet. My plans include trying to do it in pure python if possible. psutil requires C compiled libraries and we are running away from them in the installer. So far, I haven't found any alternative, but once we have a clear decission on the metrics/info to retrieve I will take a last look to the windows world to try to avoid psutil. If there are not alternatives, we'll have to add it to requirements.txt and modify the installers.

kolibri/core/analytics/management/commands/profile.py
benjaoming: What's the lower bound? I mean, at some point the profiling would turn into a DOS attack? :)

30 seconds sounds okay, that way an impatient user would maybe run the profiler for 5-10 minutes and then there would still be enough data?

How much time would we want the profiling to be running for? Say, it's a classroom and there are 30 client devices, so it'll have some fluctuation...

jredrejo: No, it will never be a DOS attack, it can affects performance because it writes in a file in the disk, however the impact should be minimal. It's not something making requests to the server, but measuring how long, how much cpu and how much memory the actual requests take.
Of course I would not have it active by default, but having it running for one hour or more in a classroom where we want to detect problems should not suppose an added problem, it should not be noticeable for the user.

jredrejo:
@indirectlylit I have updated the code to complete the architecture I had designed. I know it's not what you had in mind after reading your comments. I think the taken approach is not incompatible with your ideas, let's discuss them to decide if we implement both or just one of them.

One prior note, all the data the issue #4270 required is provided by the current code. I have added more info after reading the comments, excepting
>Some information on content currently available would be useful - e.g. total number of content nodes and total number of channels.
which I will add too.

So, going to the arquitecture:
The taken approach is trying to log all the possible information while Kolibri is in real use. To do it, two logs in csv format are generated. 

1. First log file, `performance.log` One is generated by the command itself, with information about the server performance.
2. The the other log file, `requests_performance.log` is generated by the server when the command is in execution.  It includes information about the execution of every request the server processes. 

About your caveats:
1 Having a human readable file is totally doable, containing averages, max and min over the list of data that the current logs are saving. I think it's good to have the csv format, or csv and human format, but if csv has to be discarded, it should not be a problem. 
I think users will  probably prefer the human readable file. Nevertheless to debug a problem, I think the csv is more useful as it contains more detailed information.
2. Your proposal 
>We create three new profiling endpoints: 

has, from my point of view, one advantage: no need to add the profiling middleware that now has been added, the command can take all the measurements. However I think it has an imporant disadvantage: those endpoints are not the real server data. If we need to debug a malfunctioning server I would prefer to have data from the requests the users are actually doing. If all what we want is some numbers/stats of the performance of kolibri in the server, then I like your idea. It's just a kind of benchmarking tool, not to collect debugging info.

The hardest part is done, let's talk about these two last points to decide how to end the application.

kolibri/core/analytics/api.py
```diff
@@ -0,0 +1,38 @@
...
+    client_ip, _ = get_client_ip(request)
+    local_request = client_ip == '127.0.0.1' or client_ip == '::1'
+    if not local_request:
+        return HttpResponseNotFound()
+    try:
+        command_pid = int(pid)
+    except ValueError:
+        # provided pid parameter is not an integer
+        return HttpResponseNotFound()
+    MetricsMiddleware.disabled = False
```
benjaoming: Question: What if this only happens inside the currently running instance of Kolibri, but there are several workers? Is this relevant, or should we document it somewhere? Can we detect multi-worker environments and fail the profiling?

kolibri/core/analytics/management/commands/profile.py
benjaoming: That's fair - I think what you are measuring should have a chance of finding a platform-independent & pure-python way of working.. preferably without further dependencies of course :)

kolibri/core/analytics/api.py
jredrejo: Yes, it's relevant. I have been thinking about it and discussing it with Richard and Devon as well: I thought of using the cache to notify other workers, but I can't use the cache because we have safe threads. The only option would be to have some persistent write accesible to all the workers. By now I want to avoid the situation where every request has to read from the database or the disk. For the first proof of concept, multi-worker environments will profile only the thread capturing that request.

I will evaluate the impact of just checking if a file lock exists in every request. If it's not noticeable (probably it's not, I supposse Django processing time is longer enought to make such check inappreciable ), that would be the easiest way to go: the profiler commands creates the lock file, and the middleware self-activates if it detects the file, and deactivates when the command process stops. This approach will work in a multi-worker environment.

benjaoming: I like that approach, and thanks for the update! I agree that checking for the existence of a file isn't expensive. We could perhaps switch profiling on/off through `options.ini`.. anyways, for the proof of concept I think it's all good and the multi-worker environments aren't yet relevant AFAIK (but soon-ish).

kolibri/core/analytics/pskolibri/__init__.py
```diff
@@ -0,0 +1,270 @@
...
+from kolibri.core.analytics.pskolibri.common import memoize_when_activated
+from kolibri.core.analytics.pskolibri.common import NoSuchProcess
+from kolibri.core.analytics.pskolibri.common import PY3
+
+POSIX = os.name == "posix"
+WINDOWS = os.name == "nt"
+LINUX = sys.platform.startswith("linux")
+MACOS = sys.platform.startswith("darwin")
+_TOTAL_PHYMEM = None
+_timer = getattr(time, 'monotonic', time.time)
```
benjaoming, line 60: There are a couple of these spread around -- maybe they could be maintained in `kolibri.utils.compat`? Slight tweaks for the names would be prefixing with `IS_`, like `IS_POSIX` etc...

jredrejo: I haven't been able to find other constants or requests to find out the OS. I have added them all in the common.py file in pskolibri, so they are available for the whole application there.
kolibri.utils.compat seems to be focused in different versions of python compatibility, not in the OS.

kolibri/core/analytics/management/commands/benchmark.py
```diff
@@ -74,8 +74,8 @@ class Command(BaseCommand):
     help = "Outputs performance info and statistics of usage for the running Kolibri instance in this server"
 
     def handle(self, *args, **options):
-        if not LINUX:
-            print("This OS is not supported yet")
+        if not LINUX and not WINDOWS:
+            print("This OS is not yet supported")
             sys.exit(0)
```
benjaoming: I think the exit code should be non-zero?

kolibri/core/analytics/management/commands/benchmark.py
```diff
@@ -0,0 +1,137 @@
...
+    * Free disk space:               (content_storage_free_space)
+    * Server time:                   (server_time)
+    * Server timezone:               (server_timezone)
+    """
+    help = "Outputs performance info and statistics of usage for the running Kolibri instance in this server"
+
+    def handle(self, *args, **options):
+        if not LINUX and not WINDOWS:
+            print("This OS is not yet supported")
+            sys.exit(0)
```
benjaoming: I think the exit code should be non-zero?

jredrejo: Honestly, I also thought it, but then I thought that not being able to use it is not an error... I don't have a strong opinion, both values could be valid. I will change it happily if it's the common opinion.

benjaoming: I'll gladly have that opinion to make this easier to decide :)

Let's return non-zero... zero should indicate that a command was successfully completed... not being able to run the command's purpose (benchmarks) is not quite successful, although the program terminates without exceptions. We had some unanticipated events in the past because of this exact kind of assumption.. just can't recall what it was.

benjaoming: Here's one.. but am pretty sure there's been another recent case... https://github.com/learningequality/kolibri/pull/2881

kolibri/core/analytics/management/commands/benchmark.py
```diff
@@ -0,0 +1,137 @@
...
+
+        try:
+            kolibri_cpu, kolibri_mem = get_kolibri_use()
+        except KolibriNotRunning:
+            sys.exit("Profile command executed while Kolibri server was not running")
```
benjaoming: This should also be non-zero for the same reasons.. Does the string cause this? There is at least one more similar `sys.exit` call, so maybe just search for that string, I'll hold off with the duplicate code comments :)

kolibri/core/analytics/measurements.py
```diff
@@ -0,0 +1,173 @@
...
+            kolibri_mem = str(kolibri_proc.memory_info().vms / pow(2, 10))
+            kolibri_cpu = str(kolibri_proc.cpu_percent())
+        except psutil.NoSuchProcess:
+            # Kolibri server is not running
+            raise KolibriNotRunning()
+
+    return (kolibri_cpu, kolibri_mem)
+
+
+class KolibriNotRunning(Exception):
```
benjaoming: There's also `kolibri.utils.server.NotRunning` :)

kolibri/core/analytics/pskolibri/common.py
```diff
@@ -0,0 +1,152 @@
...
+    pass
+
+
+class AccessDenied(Exception):
+    """Exception raised when permission to perform an action is denied."""
+
+    pass
+
+
+def memoize(fun):
```
benjaoming, line 54: This should exist elsewhere? Have you looked? Maybe in Django?

kolibri/core/analytics/pskolibri/common.py
```diff
@@ -0,0 +1,152 @@
...
+
+    cache = {}
+    wrapper.cache_activated = False
+    wrapper.cache_activate = cache_activate
+    wrapper.cache_deactivate = cache_deactivate
+    return wrapper
+
+
+def open_binary(fname, **kwargs):
+    return open(fname, "rb", **kwargs)
```
benjaoming: Is this necessary?

kolibri/core/analytics/pskolibri/common.py
```diff
@@ -0,0 +1,152 @@
...
+
+def open_text(fname, **kwargs):
+    """On Python 3 opens a file in text mode by using fs encoding and
+    a proper en/decoding errors handler.
+    On Python 2 this is just an alias for open(name, 'rt').
+    """
+    if PY3:
+        kwargs.setdefault('encoding', ENCODING)
+        kwargs.setdefault('errors', ENCODING_ERRS)
+    return open(fname, "rt", **kwargs)
```
benjaoming: Would `io.open` be relevant here? Try looking around the codebase for examples.

kolibri/core/analytics/pskolibri/common.py
```diff
@@ -0,0 +1,152 @@
...
+import functools
+import os
+import sys
+from collections import namedtuple
+
+PY3 = sys.version_info[0] == 3
+POSIX = os.name == "posix"
+WINDOWS = os.name == "nt"
+LINUX = sys.platform.startswith("linux")
+MACOS = sys.platform.startswith("darwin")
```
benjaoming, line 14: I still think all this stuff should be in `kolibri.utils.compat` -- not sure if you've seen my previous comment?

kolibri/core/analytics/pskolibri/_pswindows.py
```diff
@@ -0,0 +1,316 @@
+"""Windows platform implementation."""
```
benjaoming, line 1: This seems quite complicated :) How has it been tested? Is the code from elsewhere or did you write it from scratch?

kolibri/core/analytics/pskolibri/_psposix.py
```diff
@@ -0,0 +1,34 @@
+"""Routines common to all posix systems."""
+from __future__ import absolute_import
+from __future__ import division
+
+import errno
+import os
+
+
+def pid_exists(pid):
```
benjaoming: This implementation already exists in `kolibri.utils.system`

kolibri/core/analytics/pskolibri/_psosx.py
```diff
@@ -0,0 +1,3 @@
+"""macOS platform implementation."""
+from __future__ import absolute_import
+from __future__ import division
```
benjaoming, line 3: Seems empty :)

kolibri/core/analytics/pskolibri/_pslinux.py
```diff
@@ -0,0 +1,309 @@
+"""Linux platform implementation."""
```
benjaoming, line 1: Did you write all this? :)

kolibri/core/analytics/measurements.py
```diff
@@ -0,0 +1,173 @@
...
+    return (homepage_time, recommended_time, channels_time)
+
+
+def get_process_pid():
+    """
+    Returns the pid of the command in execution
+    :returns: Integer number with the pid of this command
+    """
+    this_command = psutil.Process()
+    return this_command.pid
```
benjaoming: You should be able to avoid the call to `psutil` here by using the `.pid` file as a cross-platform alternative? Is there an issue with that? In that case, it's pretty bad since we rely on the `.pid` file for other purposes...

benjaoming (changes requested):
There's a lot of code coming in here, sorry I can't review it all right now.. the pure-Python ps implementation is what I'm mostly unable to comment on (because of its complexity) and curious about... could you can add some more general comments and discussion about it? Did you implement all of it or find it in some tested source? Anything that is known to fail or not work? Etc.

benjaoming:
> ### Reviewer guidance
>
> This is a finished work.

Nice try :smile: :smile: :stuck_out_tongue:

benjaoming:
From @jredrejo: 

> initially psutil was the Python library that I used because it's multiplatform and it provides all the functions needed to build the commands, but it required compilation of C code. That's why I have been working in a pure python replacement for the functions we need.  Pushed code is based in psutil, I called it  pskolibri, but not C is required (well, I used a lot of C in the windows part but using ctypes, so no compilation is needed). I have tested it in Linux and in a clean Windows virtual machine, with only Kolibri 0.10 installed, and works perfectly without adding any dependencies.

Is this then a modified version of the original `psutil` library where the C parts are stripped out?

benjaoming:
Have tested this locally, the terminal output is beautifully formatted :heart_eyes:

kolibri/core/analytics/measurements.py
jredrejo: This is the pid of the command being executed, not from Kolibri. For kolibry I rely on the implementation we already have.

kolibri/core/analytics/measurements.py
jredrejo: Thanks, I didn't know it. Fixing it.

kolibri/core/analytics/pskolibri/_pslinux.py
jredrejo: For the Linux part, most if taken from psutils. I removed the parts we didn't need and simplified what could be simplified or reused from Kolibri. Most of this file ideas are not my merit , even if it took me work to do it.

kolibri/core/analytics/pskolibri/_psosx.py
jredrejo: It is. OSX version is not started, not even sure if it is worth doing it.

kolibri/core/analytics/pskolibri/_psposix.py
jredrejo: This is a general function, not for Kolibri, but for any process. It's needed when Django commands are executed out of the Kolibri server.

kolibri/core/analytics/pskolibri/_pswindows.py
jredrejo: If you look at it, the code is not complicated. It just uses calls to the standard Windows API. All of them are well documented in https://docs.microsoft.com/en-us/windows/desktop/api/index.
The main problem was to pass the C format to Python types, but even for it, the code is quite standard, nothing fancy nor imaginative.
I tested all the functions in windows xp and windows 7. 
Only Process.cmdline was out of my skills using the windows API. So after trying it for some time I gave up and I relied in the `wmic` command using subprocess. wmic is available in all the windows versions since 2002.

kolibri/core/analytics/pskolibri/common.py
jredrejo: I saw it and I answered you previously.
compat was used for Python versions compatibility, not for OS compatibilty.
Also this is a complete new module, used in the analytics part. I don't see any benefit of doing it in compat if it's being used in analytics.

kolibri/core/analytics/pskolibri/common.py
jredrejo: Yes, I looked and didn't find it. 
This errors are for OS errors, I don't think Kolibri has faced them before.

kolibri/core/analytics/pskolibri/common.py
jredrejo: Well, as needed as any other function that's used often. It's just a way of organizing the code.

kolibri/core/analytics/pskolibri/common.py
jredrejo: yeah, it could make sense for Python 3 only. For compatibility with Python 2 I prefer to keep using the old file open method.

kolibri/core/analytics/management/commands/benchmark.py
jredrejo: Perfect, changed

benjaoming:
Hi @jredrejo - I think that to be on the safe side and to ensure that we can easily track where the code origins from, let's write a docstring in the top of those `pskolibri` modules.

E.g.

```
"""
This is a stripped down version of psutil. See: https://github.com/giampaolo/psutil/

psutil is distributed under BSD license reproduced below.

Copyright (c) 2009, Jay Loden, Dave Daeschler, Giampaolo Rodola'
All rights reserved.


Redistribution and use in source and binary forms, with or without modification,
are permitted provided that the following conditions are met:

 * Redistributions of source code must retain the above copyright notice, this
   list of conditions and the following disclaimer.
 * Redistributions in binary form must reproduce the above copyright notice,
   this list of conditions and the following disclaimer in the documentation
   and/or other materials provided with the distribution.
 * Neither the name of the psutil authors nor the names of its contributors
   may be used to endorse or promote products derived from this software without
   specific prior written permission.

THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR
ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
(INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON
ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
(INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
"""
```

Notice the `Redistributions of source code must retain the above copyright notice` :)

kolibri/core/analytics/management/commands/profile.py
```diff
@@ -0,0 +1,83 @@
...
+            print("This OS is not yet supported")
+            sys.exit(1)
+
+        interval = 10  # the measures are taken every 10 seconds
+        _, port = get_kolibri_process_info()
+        if port:
+            this_pid = get_process_pid()
+            try:
+                urlrequest.urlopen('http://localhost:{port}/api/analytics/activate/{pid}/'.
+                                   format(port=port, pid=this_pid))
```
benjaoming: We can set a timeout parameter here to be sure that it's what we want? If there is 10 seconds intervals, the timeout could be `interval/2` ?

Also, below, IIRC, it will raise some timeout exception if the socket works but times out...

kolibri/core/analytics/measurements.py
benjaoming: Ah cool thanks for responding.

You can still get the `pid` with `os.getpid()` ? Then trim the `pskolibri` a bit more, in case that's helpful :)

kolibri/core/analytics/pskolibri/__init__.py
benjaoming: My comment doesn't appear next to the code that I originally wrote it for, but I think you responded to what I meant :)

The thing that I would think is that if we don't keep generic compatibility layers in a centralized place, we end up with a lot of modules re-inventing the same structures. So I thought that detecting the OS seems like a pretty generic thing that we could perhaps benefit from moving outside.

benjaoming: But I understand that it's perhaps part of the original `psutil` and if we do other things than just deleting code, it gets a bit hard to maintain...

benjaoming: One day, a new Linux kernel og Windows 12 will come and ask us some questions about what's going on here :)

So I guess I think it's fine to keep these constants together with their original code! :+1:

kolibri/core/analytics/pskolibri/common.py
benjaoming: Something weird happened with that comment thread, sorry about that. I see it now.

No, never mind moving this.

kolibri/core/analytics/pskolibri/common.py
benjaoming: What if opening a file in binary write mode? :)

kolibri/core/analytics/pskolibri/common.py
benjaoming: No, that's true -- now I looked, too.. was just pretty sure that this stuff used to be shipped. But Django also removed the bundled `lru_cache` not long ago.

benjaoming (changes requested):
Few more things, especially the license stuff.

kolibri/core/analytics/pskolibri/common.py
indirectlylit: actually `io.open` [is available in Python 2.6](https://docs.python.org/2/library/io.html), and is the best way I've found to write cross-version file opening.

See previous work here: #3769

Recommend using `io.open` instead of `open` everywhere to avoid issues and simplify logic.

indirectlylit: > everywhere

or rather - everywhere we are opening _text_ files

kolibri/core/analytics/pskolibri/_psosx.py
jredrejo: After chatting with @indirectlylit , we are not going to do OSX by now

kolibri/core/analytics/measurements.py
jredrejo: This change won't save any code in pskolibri. If you look at the Process class in __init__.py, it uses os.getpid() to get the pid when instantiating a new Process if the pid is not provided. It's needed to use some of the methods this class has.

In any case, I like  your suggestion, the code I wrote there is a clear case of overengineering. I am changing it.

kolibri/core/analytics/management/commands/profile.py
jredrejo: Yes, we can, and I did it in the beginning, but I removed it because @indirectlylit told me to avoid these parameters to make user life easier :)
If you both agree I have no problem in restoring it from the git history.

kolibri/core/analytics/pskolibri/common.py
jredrejo: The pskolibri code only use reads, never writes.

kolibri/core/analytics/pskolibri/common.py
jredrejo: Ok, changed.

jredrejo:
@benjaoming  I haven't been able to find the link in the code to the license changes you propose. Maybe you didn't see the license header in pskolibri/__init.py__ It gives credit to the psutil library (which uses a BSD license, so we are totally safe on that side). As this is the module main file I don't think another header is needed in all the files (also taking into account not all the files contain psutil code)

kolibri/core/analytics/pskolibri/__init__.py
```diff
@@ -0,0 +1,292 @@
...
+import os
+import sys
+import time
+
+from kolibri.core.analytics.pskolibri.common import LINUX
+from kolibri.core.analytics.pskolibri.common import memoize_when_activated
+from kolibri.core.analytics.pskolibri.common import NoSuchProcess
+from kolibri.core.analytics.pskolibri.common import PY3
+from kolibri.core.analytics.pskolibri.common import WINDOWS
+# This module is heavily based on the psutil module
```
benjaoming: Could this be at the top instead, as a docstring? That would make it more visible.

The intention is of course both that we give attribution and mention licenses.. but _also_ that we, ourselves, can make future decisions about this codebase based on its vendored origin.. and not somehow that we wrote this ourselves :)

benjaoming:
@jredrejo are you using pre-commit hooks? I'm getting this error when committing to this branch, which is caught by the Flake8 hook:

```
flake8 kolibri/core/analytics/ 
kolibri/core/analytics/pskolibri/__init__.py:231:54: F821 undefined name 'long'
```

See: https://kolibri-dev.readthedocs.io/en/latest/start/getting_started.html?highlight=pre-commit#pre-commit

benjaoming:
The reason that I added the whole license is that the license itself specifies to do so:

```
 * Redistributions of source code must retain the above copyright notice, this
   list of conditions and the following disclaimer.
```

jredrejo:
> The reason that I added the whole license is that the license itself specifies to do so:
> 
> ```
>  * Redistributions of source code must retain the above copyright notice, this
>    list of conditions and the following disclaimer.
> ```

Yep, it's fine, it's a BSD so it gives you permission to do literally what you want, but under no circumstances I would like to silent the credits to the psutil authors. I just didn't see the need to include it in every file. Having it in the __init__.py is perfect.

jredrejo:
> @jredrejo are you using pre-commit hooks? I'm getting this error when committing to this branch, which is caught by the Flake8 hook:
> 
> ```
> flake8 kolibri/core/analytics/ 
> kolibri/core/analytics/pskolibri/__init__.py:231:54: F821 undefined name 'long'
> ```
> See: https://kolibri-dev.readthedocs.io/en/latest/start/getting_started.html?highlight=pre-commit#pre-commit

@benjaoming I do use them, and can't reproduce the problem you had. Actually , looking at the code I guess you have the problem because you're using Python3 while I am using Python2. `long` exists in Python 2, but not in Python 3, that's exactly what that line of code checks ;)

kolibri/core/analytics/pskolibri/__init__.py
jredrejo: @benjaoming 
 I am not sure of how you want the text to be wording. Please, feel free to change it yourself. I am totally in favour of mentioning others work.

benjaoming:
> can't reproduce the problem you had

Right @jredrejo - I wasn't thinking very carefully about that, using Python 2 is also the choice of all our linting! Thanks for reminding!

benjaoming (approved):
Looking very much forward to having this included, tested and seeing results from users! Thanks @jredrejo ! :+1:

radinamatic:
@jredrejo Let's get together and work on some good user documentation for this command! :+1:

indirectlylit:
@jredrejo this new code currently throws an unhandled error on Mac when simply starting the devserver. We probably need to add a platform conditional around the import and initialization.

```
Unhandled exception in thread started by <function check_errors.<locals>.wrapper at 0x10995f378>
  File "/Users/d/Projects/learning_equality/repos/k011x/.venv/lib/python3.6/site-packages/django/utils/autoreload.py", line 228, in wrapper
    fn(*args, **kwargs)
  File "/Users/d/Projects/learning_equality/repos/k011x/.venv/lib/python3.6/site-packages/django/core/management/commands/runserver.py", line 146, in inner_run
    handler = self.get_handler(*args, **options)
  File "/Users/d/Projects/learning_equality/repos/k011x/.venv/lib/python3.6/site-packages/django/contrib/staticfiles/management/commands/runserver.py", line 28, in get_handler
    handler = super(Command, self).get_handler(*args, **options)
  File "/Users/d/Projects/learning_equality/repos/k011x/.venv/lib/python3.6/site-packages/django/core/management/commands/runserver.py", line 67, in get_handler
    return get_internal_wsgi_application()
  File "/Users/d/Projects/learning_equality/repos/k011x/.venv/lib/python3.6/site-packages/django/core/servers/basehttp.py", line 47, in get_internal_wsgi_application
    return import_string(app_path)
  File "/Users/d/Projects/learning_equality/repos/k011x/.venv/lib/python3.6/site-packages/django/utils/module_loading.py", line 20, in import_string
    module = import_module(module_path)
  File "/Users/d/Projects/learning_equality/repos/k011x/.venv/lib/python3.6/importlib/__init__.py", line 126, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 994, in _gcd_import
  File "<frozen importlib._bootstrap>", line 971, in _find_and_load
  File "<frozen importlib._bootstrap>", line 955, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 665, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 678, in exec_module
  File "<frozen importlib._bootstrap>", line 219, in _call_with_frames_removed
  File "/Users/d/Projects/learning_equality/repos/k011x/kolibri/deployment/default/wsgi.py", line 15, in <module>
    application = get_wsgi_application()
  File "/Users/d/Projects/learning_equality/repos/k011x/.venv/lib/python3.6/site-packages/django/core/wsgi.py", line 14, in get_wsgi_application
    return WSGIHandler()
  File "/Users/d/Projects/learning_equality/repos/k011x/.venv/lib/python3.6/site-packages/django/core/handlers/wsgi.py", line 151, in __init__
    self.load_middleware()
  File "/Users/d/Projects/learning_equality/repos/k011x/.venv/lib/python3.6/site-packages/django/core/handlers/base.py", line 80, in load_middleware
    middleware = import_string(middleware_path)
  File "/Users/d/Projects/learning_equality/repos/k011x/.venv/lib/python3.6/site-packages/django/utils/module_loading.py", line 20, in import_string
    module = import_module(module_path)
  File "/Users/d/Projects/learning_equality/repos/k011x/.venv/lib/python3.6/importlib/__init__.py", line 126, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 994, in _gcd_import
  File "<frozen importlib._bootstrap>", line 971, in _find_and_load
  File "<frozen importlib._bootstrap>", line 955, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 665, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 678, in exec_module
  File "<frozen importlib._bootstrap>", line 219, in _call_with_frames_removed
  File "/Users/d/Projects/learning_equality/repos/k011x/kolibri/core/analytics/middleware.py", line 11, in <module>
    import kolibri.core.analytics.pskolibri as psutil
  File "/Users/d/Projects/learning_equality/repos/k011x/kolibri/core/analytics/pskolibri/__init__.py", line 73, in <module>
    raise NotImplementedError('platform %s is not supported' % sys.platform)
NotImplementedError: platform darwin is not supported
```

benjaoming:
image

@jredrejo is this a creationist easter egg? :D

indirectlylit:
running in linux, prior to starting Kolibri:

```
vagrant@vagrant-ubuntu-trusty-64 ~> ./kolibri.pex manage benchmark
/home/vagrant/.pex/install/kolibri-0.11.0a5.dev0+git.421.g9ea67fa-py2.py3-none-any.whl.efc443a782f9b8a23b9528431dced3c371ef8029/kolibri-0.11.0a5.dev0+git.421.g9ea67fa-py2.py3-none-any.whl/kolibri/dist/cext/cp27/Linux/cp27mu/x86_64/cryptography/hazmat/primitives/constant_time.py:26: CryptographyDeprecationWarning: Support for your Python version is deprecated. The next version of cryptography will remove support. Please upgrade to a 2.7.x release that supports hmac.compare_digest as soon as possible.
  utils.DeprecatedIn23,
  File ".bootstrap/_pex/pex.py", line 365, in execute
  File ".bootstrap/_pex/pex.py", line 293, in _wrap_coverage
  File ".bootstrap/_pex/pex.py", line 325, in _wrap_profiling
  File ".bootstrap/_pex/pex.py", line 408, in _execute
  File ".bootstrap/_pex/pex.py", line 466, in execute_entry
  File ".bootstrap/_pex/pex.py", line 471, in execute_module
  File "/usr/lib/python2.7/runpy.py", line 180, in run_module
    fname, loader, pkg_name)
  File "/usr/lib/python2.7/runpy.py", line 72, in _run_code
    exec code in run_globals
  File "/home/vagrant/.pex/install/kolibri-0.11.0a5.dev0+git.421.g9ea67fa-py2.py3-none-any.whl.efc443a782f9b8a23b9528431dced3c371ef8029/kolibri-0.11.0a5.dev0+git.421.g9ea67fa-py2.py3-none-any.whl/kolibri/__main__.py", line 10, in <module>
    main(args=sys.argv[1:])
  File "/home/vagrant/.pex/install/kolibri-0.11.0a5.dev0+git.421.g9ea67fa-py2.py3-none-any.whl.efc443a782f9b8a23b9528431dced3c371ef8029/kolibri-0.11.0a5.dev0+git.421.g9ea67fa-py2.py3-none-any.whl/kolibri/utils/cli.py", line 638, in main
  File "/home/vagrant/.pex/install/kolibri-0.11.0a5.dev0+git.421.g9ea67fa-py2.py3-none-any.whl.efc443a782f9b8a23b9528431dced3c371ef8029/kolibri-0.11.0a5.dev0+git.421.g9ea67fa-py2.py3-none-any.whl/kolibri/utils/cli.py", line 428, in manage
  File "/home/vagrant/.pex/install/kolibri-0.11.0a5.dev0+git.421.g9ea67fa-py2.py3-none-any.whl.efc443a782f9b8a23b9528431dced3c371ef8029/kolibri-0.11.0a5.dev0+git.421.g9ea67fa-py2.py3-none-any.whl/kolibri/dist/django/core/management/__init__.py", line 364, in execute_from_command_line
    utility.execute()
  File "/home/vagrant/.pex/install/kolibri-0.11.0a5.dev0+git.421.g9ea67fa-py2.py3-none-any.whl.efc443a782f9b8a23b9528431dced3c371ef8029/kolibri-0.11.0a5.dev0+git.421.g9ea67fa-py2.py3-none-any.whl/kolibri/dist/django/core/management/__init__.py", line 356, in execute
  File "/home/vagrant/.pex/install/kolibri-0.11.0a5.dev0+git.421.g9ea67fa-py2.py3-none-any.whl.efc443a782f9b8a23b9528431dced3c371ef8029/kolibri-0.11.0a5.dev0+git.421.g9ea67fa-py2.py3-none-any.whl/kolibri/dist/django/core/management/base.py", line 283, in run_from_argv
  File "/home/vagrant/.pex/install/kolibri-0.11.0a5.dev0+git.421.g9ea67fa-py2.py3-none-any.whl.efc443a782f9b8a23b9528431dced3c371ef8029/kolibri-0.11.0a5.dev0+git.421.g9ea67fa-py2.py3-none-any.whl/kolibri/dist/django/core/management/base.py", line 330, in execute
  File "/home/vagrant/.pex/install/kolibri-0.11.0a5.dev0+git.421.g9ea67fa-py2.py3-none-any.whl.efc443a782f9b8a23b9528431dced3c371ef8029/kolibri-0.11.0a5.dev0+git.421.g9ea67fa-py2.py3-none-any.whl/kolibri/core/analytics/management/commands/benchmark.py", line 85, in handle
    get_requests_info()
  File "/home/vagrant/.pex/install/kolibri-0.11.0a5.dev0+git.421.g9ea67fa-py2.py3-none-any.whl.efc443a782f9b8a23b9528431dced3c371ef8029/kolibri-0.11.0a5.dev0+git.421.g9ea67fa-py2.py3-none-any.whl/kolibri/core/analytics/measurements.py", line 95, in get_requests_info
    return (homepage_time, recommended_time, channels_time)
UnboundLocalError: local variable 'homepage_time' referenced before assignment
```

indirectlylit:
testing in an ubuntu VM, it seems to report VSX rather than RSS memory usage:

```
Memory
* Used memory:                    325 Mb
* Total memory:                   489 Mb
* Kolibri memory usage:           846076 Kb
```
`ps aux`: 


```
USER       PID %CPU %MEM    VSZ   RSS TTY      STAT START   TIME COMMAND
vagrant   6070  1.5 25.9 846076 130056 ?       Sl   00:50   0:04 /usr/bin/python ./kolibri.pex start
```

Also recommend formatting in Mb for consistency

indirectlylit:
last note - the `profile` commands seem to work well, but recommend a few minor changes to the logging:

* change extension from `.log` to `.csv` so it's handled better by the OS (e.g. open in spreadsheet program)
* in the first line of the files, add column headers to make the file self-documenting
* for each run of the `profile` command, start new log files with the file names prefixed by a human-readable date/time

-----

that's all! this is looking awesome nice job

jredrejo:
> @jredrejo this new code currently throws an unhandled error on Mac when simply starting the devserver. We probably need to add a platform conditional around the import and initialization.
> 
> ```
> Unhandled exception in thread started by <function check_errors.<locals>.wrapper at 0x10995f378>
> Traceback (most recent call last):
>   File "/Users/d/Projects/learning_equality/repos/k011x/.venv/lib/python3.6/site-p
....
>     import kolibri.core.analytics.pskolibri as psutil
>   File "/Users/d/Projects/learning_equality/repos/k011x/kolibri/core/analytics/pskolibri/__init__.py", line 73, in <module>
>     raise NotImplementedError('platform %s is not supported' % sys.platform)
> NotImplementedError: platform darwin is not supported
> ```

I clearly failed on the code part where the OS was not supported. All the errors you saw in MacOS should have been fixed.

jredrejo:
> testing in an ubuntu VM, it seems to report VSX rather than RSS memory usage:
> 
> ```
> Memory
> * Used memory:                    325 Mb
> * Total memory:                   489 Mb
> * Kolibri memory usage:           846076 Kb
> ```
> `ps aux`:
> 
> ```
> USER       PID %CPU %MEM    VSZ   RSS TTY      STAT START   TIME COMMAND
> vagrant   6070  1.5 25.9 846076 130056 ?       Sl   00:50   0:04 /usr/bin/python ./kolibri.pex start
> ```
> Also recommend formatting in Mb for consistency

I took those decissions on purpose:

- vms memory takes into account the swapped memory as well. rss is just RAM.
- Units in kilobytes seemed better for me for the process information, as usually is a low number compared to the OS number

I have changed both things to follow your advice.

jredrejo:
> last note - the `profile` commands seem to work well, but recommend a few minor changes to the logging:
> 
> * change extension from `.log` to `.csv` so it's handled better by the OS (e.g. open in spreadsheet program)
No problem, done (not pushed yet)

> * in the first line of the files, add column headers to make the file self-documenting

This change has forced to stop using logger and change to use the csv library. (not pushed yet)

> * for each run of the `profile` command, start new log files with the file names prefixed by a human-readable date/time
> 

This brings me the only doubt: This will end creating many files in the .kolibri folder, 2 per each time profiler is run. To avoid missing all the files in that folder, do you think we should consider using another directory to store them? maybe a new folder inside .kolibri?

radinamatic:
> consider using another directory to store them? maybe a new folder inside .kolibri?

I would agree with this, especially if we are going to direct people to that location, it would be helpful to have them separated from the rest. :+1:

indirectlylit:
thanks for all these changes!

> I took those decisions on purpose

I was mainly just confused because it said something like 800000 Kb which didn't make sense to me - it was much higher than I expected

kolibri/core/analytics/middleware.py
```diff
@@ -0,0 +1,135 @@
...
+
+    def process_response(self, request, response):
+        """
+        Calculate and output the page generation details
+        Log output consist on:
+        Datetime, request path, request duration, memory before, memory after requests is finished,
+        cpu load before, cpu load after the request is finished, max
+        Being `max` True or False to indicate if this is the slowest request since logging began.
+        """
+        self.check_start_conditions()
```
rtibbles, line 138: I am not sure how much of a performance hit this is going to cause, but it seems like we are touching the disk for every request (regardless of whether the profiling is turned off or not), in order to be able to check if the profiling command is currently being run.

As running the profiling command will require being able to invoke the Kolibri command from the command line anyway, could we instead require that the profiling command is run before the server is started, and then do a startup check on the middleware in the `__init__` method: https://docs.djangoproject.com/en/1.11/topics/http/middleware/#marking-middleware-as-unused ?

rtibbles: Oh, sorry - I overstated the issue, as you are checking the conf file value. We could perhaps just throw `MiddlewareNotUsed` during `__init__` if the conf value is not set?

kolibri/core/api_urls.py
```diff
@@ -10,5 +10,5 @@
     url(r'^device/', include('kolibri.core.device.api_urls')),
     url(r'^lessons/', include('kolibri.core.lessons.api_urls')),
     url(r'^discovery/', include('kolibri.core.discovery.api_urls')),
-    url(r'^public/', include('kolibri.core.public.api_urls')),
+    url(r'^public/', include('kolibri.core.public.api_urls'))
```
rtibbles: Why remove the trailing comma? I don't think we've enforced this, but I think there are good reasons to retain them! https://medium.com/@nikgraf/why-you-should-enforce-dangling-commas-for-multiline-statements-d034c98e36f8

benjaoming: Great blog post. In Python, the issue is even worse: a missing comma causes string concatenation between two lines, which will cause issues at runtime and may not be caught in testing.

kolibri/core/analytics/middleware.py
jredrejo: I am not sure if I u nderstand what you mean: do you want to throw an error when kolibri starts if the conf value is  not set? The idea was to leave it not set by default. Only if you are going to profile you need to activate it.

jredrejo: Forget my previous comment. I got it, and implemented it. It was a very good idea :100:

kolibri/core/analytics/middleware.py
```diff
@@ -67,11 +71,15 @@ class MetricsMiddleware(MiddlewareMixin):
...
     """
-    slowest_request = 'unknown'
     slowest_request_time = 0
     disabled = True
     command_pid = 0
 
+    def __init__(self, get_response=None):
+        super(MetricsMiddleware, self).__init__(get_response=get_response)
+        if not conf.OPTIONS["Server"]["PROFILE"]:
+            raise MiddlewareNotUsed('Request profiling is not enabled')
```
rtibbles, line 81: :+1:

rtibbles (approved):
Looks good to me!

indirectlylit:
it's merge time!

great job @jredrejo

PR #4277, opened 2018-09-18
Make any cross-plugin url references safe to plugin removal.

rtibbles:
### Summary
Continuing cleanup of cross-plugin references.

Previous to this PR we had some bare references to URLs, this was unsafe for plugin purposes, and also due to the new `URL_PATH_PREFIX` option which means these links would have broken.

This PR resolves these two issues by referencing URLs from the kolibri `urls` object, and then only attempting to invoke the URL function if it exists.

### Reviewer guidance
Do these links still function?

Question:
1) The pattern in here could be made more general by making the URLs object return a function that returns `undefined` when the URL is not defined, which would avoid having to reimplement undefined checking on the URL function itself - would this be useful? My sense is it might be overkill.

----

### Contributor Checklist

### Reviewer Checklist

### DIFF
--- a/kolibri/core/assets/src/core-app/__mocks__/urls.js
+++ b/kolibri/core/assets/src/core-app/__mocks__/urls.js
@@ -1,10 +1,18 @@
-const urls = new Proxy(
-  {},
-  {
-    get() {
-      return () => '';
-    },
-  }
-);
+let returnedUrl = 'test';
+
+const urlsObject = {
+  __setUrl(url) {
+    returnedUrl = url;
+  },
+};
+
+const urls = new Proxy(urlsObject, {
+  get(obj, prop) {
+    if (obj[prop]) {
+      return obj[prop];
+    }
+    return () => returnedUrl;
+  },
+});
 
 export default urls;
--- a/kolibri/plugins/coach/assets/src/views/ClassListPage.vue
+++ b/kolibri/plugins/coach/assets/src/views/ClassListPage.vue
@@ -9,9 +9,9 @@
     >
       <KExternalLink
         slot="details"
-        v-if="isAdmin"
+        v-if="isAdmin && createClassUrl"
         :text="$tr('noClassesDetailsForAdmin')"
-        href="/facility"
+        :href="createClassUrl"
       />
     </AuthMessage>
 
@@ -70,6 +70,7 @@
   import orderBy from 'lodash/orderBy';
   import KRouterLink from 'kolibri.coreVue.components.KRouterLink';
   import KExternalLink from 'kolibri.coreVue.components.KExternalLink';
+  import urls from 'kolibri.urls';
   import { PageNames } from '../constants';
   import { filterAndSortUsers } from '../../../../facility_management/assets/src/userSearchUtils';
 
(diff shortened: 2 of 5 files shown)


indirectlylit:
looks like the unit test may have found a legitimate breakage!

`a` not found...

```
    > 56 |       expect(els.AuthMessage().find('a').attributes().href).toEqual('/facility');
```

indirectlylit:
This is getting into interesting territory... the plugins are still coupled by knowledge of the `urls` object keys, but with this change they at least kind of handle the case when the plugin isn't loaded it looks like?

I'm wondering if maybe this is a good opportunity to start thinking about the plugin API a little more, in particular how we cross-reference "permalinks" across pluggable apps. For example, what if we had a global function like

```javascript
plugin = getPlugin('plugin_id')
```

Then you could start doing things like

```html
<KExternalLink v-if="plugin.app_loaded" :href="plugin.base_url">
```

This could potentially allow plugins to register both the top-level Django URL as well as internal Vue router URLs:

```html
<KExternalLink v-if="plugin.app_loaded" :href="plugin.getUrl('url_id')">
```

rtibbles:
I think I need to update the test, one for the case where the link should be there and one where it shouldn't. The new behaviour is that the link is not shown if it is falsy, which all urls are for the default mocking! Maybe I should update the mock to return something by default.

rtibbles:
Yeah, I did think about doing something more generalized, particularly when I was writing out those long url paths, but I decided not to over implement for now.

indirectlylit:
> decided not to over implement for now

yeah, that's usually my attitude also :)

To me the most important functionality is allowing one plugin to be able to know whether another plugin is available.

-----

A slight tangent that would probably be better for a conversation...

My perception is colored because I've never quite understood how auto-generating a Django URL reversal object is doing much in the way of decoupling for us, because the keys in the object still need to be hard-coded.

So not only are we still coupled, but to figure out what Python code is triggered by a particular JS call, a newcomer might follow this path:

* Kolibri Resource layer API
* _Kolibri Resource layer internals_
* _Django JS reverse-generated javascript snippets_
* _Webpack module integration_
* _Kolibri 'Hooks' for hoisting into templates_
* _Django `reverse()`-generated IDs_
* DRF-generated URLs
* A Django view

I've often wished we could cut out the italicized items and hard-code URLs directly into the `Resource` object intializers:

```javascript
export default new Resource({
  resource_url: '/api/some/great/resource',
});
```

In other words, rather than

```javascript
console.log(`${urls['kolibri:devicemanagementplugin:device_management']()}#/content`);
```

we could just hardcode URLs in a core `pluginURLs.js` file like

```javascript
export const DeviceBaseURL = '/device/';
export const DeviceURLs = {
  CONTENT_PAGE: `${DeviceBaseURL}#/content`,
  PERMISSIONS_PAGE: `${DeviceBaseURL}#/permissions`,
  INFO_PAGE: `${DeviceBaseURL}#/info`,
};
```

and then

```javascript
import { DeviceURLs } from 'kolibri.pluginURLs';
console.log(DeviceURLs.CONTENT_PAGE);
```

I realize that there are at least a few issues with this:

1. Maintaining a hard-coded `pluginURLs.js` file would be time-consuming and error-prone. Possible alternative: auto-generated, like we're starting to do with language files?
2. Dynamic resolution and installation of plugins isn't supported. However it's not supported by the Django `reverse()` system either
3. It doesn't handle the `URL_PATH_PREFIX` issue out of the box, but that could be addressed by bootstrapping just that one variable into the system from Django rather than the whole URLs object

indirectlylit (approved):
looks good, with my comments noted above

PR #4286, opened 2018-09-20
Update languages

indirectlylit:
### Summary

This PR will be a pre-requisite to the upcoming i18n refactor work

* updates with the latest languages
* simplifies the directory structure a bit - move all files into LC_MESSAGES
* don't try to lazy-load language files in order to avoid a costly async wait on the critical path. Take the lesser evil and increase file size

### Reviewer guidance

check it out, see if it makes sense

### References

These changes (in some form) will be a prerequisite for #4254

----

### Contributor Checklist

### Reviewer Checklist


kolibri/core/assets/src/utils/vue-intl-locale-data.js
```diff
@@ -4,123 +4,25 @@
...
+  data.push(require('vue-intl/locale-data/my.js'));
+  data.push(require('vue-intl/locale-data/nyn.js'));
+  data.push(require('vue-intl/locale-data/pt.js'));
+  data.push(require('vue-intl/locale-data/sw.js'));
+  data.push(require('vue-intl/locale-data/ta.js'));
+  data.push(require('vue-intl/locale-data/te.js'));
+  data.push(require('vue-intl/locale-data/ur.js'));
+  data.push(require('vue-intl/locale-data/yo.js'));
+  data.push(require('vue-intl/locale-data/zu.js'));
+  return data;
```
rtibbles, line 27: This will work for now, but I worry about the sustainability of this approach in the long term.

kolibri/core/webpack/hooks.py
```diff
@@ -230,7 +230,7 @@ def frontend_message_file(self, lang_code):
             file_path = os.path.join(
                 path,
                 to_locale(lang_code),
-                "LC_FRONTEND_MESSAGES",
+                "LC_MESSAGES",
```
rtibbles, line 233: Is this just to get everything in the same folder on crowdin?

rtibbles (approved):
I would like to think how to handle the data better as we scale, but this'll do for now.

kolibri/core/webpack/hooks.py
indirectlylit: no, on crowdin we skip this folder entirely.

I would have preferred to mirror the crowdin directory structure in our repo, but it seems that the assumed existence of an `LC_MESSAGES` directory is built into the Django translation machinery.

As far as I can tell, adding a separate `LC_FRONTEND_MESSAGES` was never actually necessary and may have just been an off-the-cuff developer decision

kolibri/core/assets/src/utils/vue-intl-locale-data.js
indirectlylit: Yes agreed. In an upcoming PR I'll have a similar issue with font subset files for each language.

Eventually I'd like to get your help with figuring out how to inline individual files based on the current language, but for now loading everything is still faster than adding an async request.

(Note this will only become critical when we add Chinese, Japanese, or Korean because these font files are enormous)

indirectlylit:
> I would like to think how to handle the data better as we scale, but this'll do for now.

absolutely agree

PR #4287, opened 2018-09-20
allow proper translation of assignment modal

### LINKED ISSUE #3882: Insufficient options for correct i18n in AssignmentChangeStatusModal
### Observed behavior
When observed in UI language with gendered adjectives as Spanish, option labels in the status change modal for Lessons and Exams will always display incorrectly in one of those two. 

At first I thought it was my error while translating, but after digging deeper, it seems to be more than that. If I'm interpreting the code correctly, the `AssignmentChangeStatusModal.vue` component is (re)used in both cases, even though in languages other than English, Lessons and Exams might be adjectives of different genders. In Spanish *Lesson* is **Activa** but *Exam* is **Activo**.

selection_005

I can appreciate the "reuse" philosophy, but in this case the result is too DRY for the correct i18n. Since we are past the string freeze, I've changed the translation on Crowdin to verbs (**Activar/Desactivar**), and my suggestion is to do the same for the next release, namely, [make the following change](https://github.com/learningequality/kolibri/blob/develop/kolibri/plugins/coach/assets/src/views/assignments/AssignmentChangeStatusModal.vue#L79):

```
      activeOption: 'Activate',
      inactiveOption: 'Deactivate', 
```

Not much we can do in the `assignments/StatusIcon.vue` at this point, but it is probably going to need to get a bit wet in the future.

### User-facing consequences
Incorrect UI text that reduces understanding.

indirectlylit:
### Summary

fixes #3882 by adding additional parameters

### Reviewer guidance

make sure it works

Using a parameterized component here might be overkill?

----

### Contributor Checklist

### Reviewer Checklist

### DIFF
--- a/kolibri/plugins/coach/assets/src/views/exams/ExamReportPage/ManageExamModals.vue
+++ b/kolibri/plugins/coach/assets/src/views/exams/ExamReportPage/ManageExamModals.vue
@@ -14,6 +14,8 @@
       v-else-if="examsModalSet === AssignmentActions.CHANGE_STATUS"
       :modalTitle="$tr('changeExamStatusTitle')"
       :modalDescription="$tr('changeExamStatusDescription')"
+      :modalActiveOption="$tr('changeExamStatusActive')"
+      :modalInactiveOption="$tr('changeExamStatusInactive')"
       :active="exam.active"
       @changeStatus="handleChangeStatus"
       @cancel="setExamsModal(null)"
@@ -149,6 +151,8 @@
     $trs: {
       changeExamStatusTitle: 'Change exam status',
       changeExamStatusDescription: 'Learners can only see active exams',
+      changeExamStatusActive: 'Active',
+      changeExamStatusInactive: 'Inactive',
       copyExamTitle: 'Copy exam',
       copyExplanation: 'Copy this exam to',
       assignmentQuestion: 'Assign exam to',
--- a/kolibri/plugins/coach/assets/src/views/assignments/AssignmentChangeStatusModal.vue
+++ b/kolibri/plugins/coach/assets/src/views/assignments/AssignmentChangeStatusModal.vue
@@ -9,12 +9,12 @@
   >
     <p>{{ modalDescription }}</p>
     <KRadioButton
-      :label="$tr('activeOption')"
+      :label="modalActiveOption"
       :value="true"
       v-model="activeIsSelected"
     />
     <KRadioButton
-      :label="$tr('inactiveOption')"
+      :label="modalInactiveOption"
       :value="false"
       v-model="activeIsSelected"
     />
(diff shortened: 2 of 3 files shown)


kolibri/plugins/coach/assets/src/views/exams/ExamReportPage/ManageExamModals.vue
```diff
@@ -149,6 +151,8 @@
     $trs: {
       changeExamStatusTitle: 'Change exam status',
       changeExamStatusDescription: 'Learners can only see active exams',
+      changeExamStatusActive: 'Active',
```
rtibbles, line 154: I really wish there was a way for us to provide more context to translators from our json files - this makes it possible to fix, but still not clear that it should be appropriately gendered!

rtibbles: I think potentially, the best way for us to provide additional context would be... to not upload json files. This would require some parsing back and forth, but with your deprecation of the Crowdin client in #4254, that may not be such a big deal, as we can define the conversion process in there!

rtibbles (approved):
It works - while it's only 2 instances, I think there is sufficient overlap for it to be OK - especially, as you mention, we are trying to align the two experiences.

kolibri/plugins/coach/assets/src/views/exams/ExamReportPage/ManageExamModals.vue
radinamatic: @indirectlylit @rtibbles You were too fast for me here, and I'm afraid that I don't understand what exactly was "fixed" here... Is it possible to get a screenshot or similar?

"Aligning experiences" is a worthy goal, but in the line 154 I still see the adjectival form `Active`... :confused: 

My recommendation was to switch to verb, `Activate/Deactivate` which would avoid the whole mess, is there a reason why that could not have been be implemented?

indirectlylit: > I don't understand what exactly was "fixed" here... 

Previously there were only two strings (`activeOption` and `inactiveOption`) that had to be used with both exams and lessons.

Now there are four strings, so they can be customized to the modified noun:

* `changeLessonStatusActive`: Activas
* `changeLessonStatusInctive`: Inactivas
* `changeExamStatusActive`: Activos
* `changeExamStatusInctive`: Inactivos

> Is it possible to get a screenshot or similar?

the screenshots would look exactly the same because the Lesson and Exam forms are the same in English

> My recommendation was to switch to verb

I can do that too, but usually we reserve verbs for buttons so I was hesitant to...

Can still do that if you think it's better - I would loop @jtamiace in on that

radinamatic: > Now there are four strings

Now I get it, thanks... No need for screenshots! :+1: 
This resolves the immediate concern, the only thing that translators would need is a bit more context, so we should put back on the table the discussion about solving the context-on-Crowdin issue as soon as this release is done.

I also understand the hesitation, and the will to stick to `button=verb` convention, and now I'm hesitant too (we do demand action, but still... :thinking: )

PR #4291, opened 2018-09-20
Add validation warning to username input on login.

### LINKED ISSUE #4068: provide validation of usernames in login screen
### Observed behavior

Currently, we permit anything to be entered in the username field. We've observed learners trying to enter in their full names (i.e., first and last name with a "space") instead of their user name.

### Expected behavior

Login field should validate usernames based on our allowed characters (e.g. spaces)

### User-facing consequences

Users will have assistance logging in and understanding usernames

rtibbles:
### Summary
Previously we let people submit any username for login (aside from trimming the input).

This now validates the username based on our username validator and shows a validation error if they have input something invalid (spaces are not caught unless a character is entered subsequent to the space, due to the trim on the `v-model`).

It also disables the submit button until the form is valid.

### Reviewer guidance
Can you log in with a valid username?
Can you try to log in with an invalid username? (hopefully not!)

### References
Fixes #4068

----

### Contributor Checklist

### Reviewer Checklist

### DIFF
--- a/kolibri/plugins/user/assets/src/views/SignInPage/index.vue
+++ b/kolibri/plugins/user/assets/src/views/SignInPage/index.vue
@@ -126,6 +126,7 @@
   import KExternalLink from 'kolibri.coreVue.components.KExternalLink';
   import KTextbox from 'kolibri.coreVue.components.KTextbox';
   import CoreLogo from 'kolibri.coreVue.components.CoreLogo';
+  import { validateUsername } from 'kolibri.utils.validators';
   import UiAutocompleteSuggestion from 'keen-ui/src/UiAutocompleteSuggestion';
   import UiAlert from 'keen-ui/src/UiAlert';
   import responsiveWindow from 'kolibri.coreVue.mixins.responsiveWindow';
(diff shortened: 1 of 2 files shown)


kolibri/plugins/user/assets/src/views/SignInPage/index.vue
```diff
@@ -201,16 +203,19 @@
...
       },
-      usernameIsInvalidText() {
+      usernameErrorText() {
         if (this.usernameBlurred || this.formSubmitted) {
           if (this.username === '') {
             return this.$tr('required');
           }
         }
+        if (this.username && !validateUsername(this.username)) {
+          return this.$tr('usernameNotAlphaNumUnderscore');
```
indirectlylit: I wonder if we should do one of:

* rename `validateUsername` to `validateUsernameIsAlphanumeric`
* return `this.$tr('usernameNotAlphaNumUnderscore')` from `validateUsername`

Currently this logic assumes that `validateUsername` doesn't, for example, check max length

indirectlylit (dismissed):
This update changes the logic slightly such that error messages might be shown when `this.usernameBlurred` is `false` and the form has not yet been submitted, which goes against the form validation conventions we've established.

indirectlylit:
> It also disables the submit button until the form is valid.

This also goes against our established conventions

christianmemije:
In `usernameIsInvalid` we are now kind of repeating the logic that happens in `usernameErrorText`. I also agree with waiting until blur and not disabling the submit button if invalid to keep consistency.

christianmemije (approved):
LGTM :100:

PR #4327, opened 2018-09-26
Don't aggressively cache recommendation endpoints.

rtibbles:
### Summary
Reduces the cache time on content recommendations from 20 years to 10 minutes.

### Reviewer guidance
Check the headers on the responses, they should be set to cache for 10 minutes from now.
----

### Contributor Checklist

### Reviewer Checklist

### DIFF
--- a/kolibri/content/api.py
+++ b/kolibri/content/api.py
@@ -57,7 +57,7 @@ def cache_forever(some_func):
             request = args[0]
             request = kwargs.get('request', request)
         except IndexError:
-            request = None
+            request = kwargs.get('request', None)
         if isinstance(request, HttpRequest):
             if any(map(lambda x: x in request.GET, ['popular', 'next_steps', 'resume'])):
                 timeout = 600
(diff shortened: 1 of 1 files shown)


ralphiee22 (approved):
Manually tested 👍

kolibri/content/api.py
```diff
@@ -47,7 +48,20 @@ def cache_forever(some_func):
...
     def wrapper_func(*args, **kwargs):
         response = some_func(*args, **kwargs)
-        patch_response_headers(response, cache_timeout=cache_timeout)
+        # This caching has the unfortunate effect of also caching the dynamically
+        # generated filters for recommendation, this quick hack checks if
+        # the request is any of those filters, and then applies less long running
+        # caching on it.
+        timeout = cache_timeout
+        try:
+            request = args[0]
```
ralphiee22, line 57: if we get an `IndexError`, can we guarantee the request object does not exist in the kwargs?

rtibbles: Hrm, true - probably not. I stopped here because in manual testing the args was sufficient. Will update.

indirectlylit: Is the idea that this won't work if somehow request is passed as a keyword?

```python
wrapper_func(request=request)
```

Is this even possible?

Issue #4301, opened 2018-09-21
Should choose content format that has available renderer


jamalex:
### Observed behavior

When there are multiple presets available for a piece of content, but there's no renderer capable of handling the first one, it errors out:

![image](https://user-images.githubusercontent.com/618416/45853477-85f3e680-bcfa-11e8-8d47-b78ca017d55b.png)

### Expected behavior

In the case above, we *do* have a PDF renderer, so it could have fallen back to that file and used that renderer.

Additionally: we should ensure we use the new le_utils "order" field to select the highest priority content preset from those available (and renderable).

### Steps to reproduce

1. Create a channel with a content node and an epub
1. Download the channel to kolibri (with epub renderer disabled)
1. Load that content node under /learn/

### Context

0.11 (develop)

jamalex:
Longer-term: we should add in the drop-down for switching between available non-supplementary presets.

indirectlylit:
May want to consider targeting this at 0.10?

rtibbles:
I think so, @indirectlylit - this is a backwards compatibility concern otherwise.

PR #4345, opened 2018-09-28
Only try to display renderable files in a content node.

DXCanas:
### Summary

A fix for #4301, done in close collaboration with @rtibbles


indirectlylit (approved):
looks good to me. merging in so we can try to get it in 0.11 also

rtibbles:
Just as a sanity check, this all looks hunky-dory to me!

Issue #4334, opened 2018-09-26
errors during startup (database locked, DatabaseIDModel does not exist)


indirectlylit:
### Observed behavior

Attempting to run `kolibri start` on the most recent `develop` branch I get the errors below. This does not happen with the devserver.

At this point, I cannot load Kolibri.

### Expected behavior

No errors, Kolibri loads

### User-facing consequences

sadness, regret

### Errors and logs

```
INFO     Running 'kolibri start' in foreground...
INFO:kolibri.utils.cli:Running 'kolibri start' in foreground...
INFO     Kolibri running on:

INFO:kolibri.utils.cli:Kolibri running on:

	http://127.0.0.1:8080/
	http://192.168.1.2:8080/

INFO     Attempting a ping.
INFO:kolibri.core.analytics.management.commands.ping:Attempting a ping.
Exception in thread Thread-1:
Traceback (most recent call last):
  File "/Users/d/Projects/learning_equality/repos/k011x/.venv/lib/python3.6/site-packages/morango/models.py", line 71, in get_or_create_current_database_id
    return cls.objects.get(current=True)
  File "/Users/d/Projects/learning_equality/repos/k011x/.venv/lib/python3.6/site-packages/django/db/models/manager.py", line 85, in manager_method
    return getattr(self.get_queryset(), name)(*args, **kwargs)
  File "/Users/d/Projects/learning_equality/repos/k011x/.venv/lib/python3.6/site-packages/django/db/models/query.py", line 380, in get
    self.model._meta.object_name
morango.models.DoesNotExist: DatabaseIDModel matching query does not exist.

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/Users/d/Projects/learning_equality/repos/k011x/.venv/lib/python3.6/site-packages/django/db/backends/utils.py", line 64, in execute
    return self.cursor.execute(sql, params)
  File "/Users/d/Projects/learning_equality/repos/k011x/.venv/lib/python3.6/site-packages/django/db/backends/sqlite3/base.py", line 328, in execute
    return Database.Cursor.execute(self, query, params)
sqlite3.OperationalError: database is locked

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/usr/local/Cellar/python/3.6.5/Frameworks/Python.framework/Versions/3.6/lib/python3.6/threading.py", line 916, in _bootstrap_inner
    self.run()
[...]

PR #4357, opened 2018-10-02
use lock objects on the threads run methods

ralphiee22:
### Summary

It seems that the vacuum and ping command are trying to access the database at the same time during startup. This causes some important models to not be generated by the ping command, as well as database is locked errors. Although, kolibri starts per usual even given this error. This PR adds a blocking lock to the threads so one process can not execute until the other has finished.

### Reviewer guidance

`make assets`
`kolibri start`
is there any database is locked errors?

### References

https://github.com/learningequality/kolibri/issues/4334

----

### Contributor Checklist

### Reviewer Checklist

### DIFF
--- a/kolibri/utils/server.py
+++ b/kolibri/utils/server.py
@@ -47,7 +47,7 @@ DAEMON_LOG = os.path.join(conf.KOLIBRI_HOME, "server.log")
 LISTEN_ADDRESS = "0.0.0.0"
 
 # use locks so vacuum doesn't conflict with ping
-lock = threading.Lock()
+vacuum_db_lock = threading.Lock()
 
 
 class NotRunning(Exception):
@@ -106,8 +106,7 @@ class PingbackThread(threading.Thread):
         thread.start()
 
     def run(self):
-        with lock:
-            call_command("ping")
+        call_command("ping")
 
 
 class VacuumThread(threading.Thread):
@@ -119,9 +118,8 @@ class VacuumThread(threading.Thread):
         thread.start()
 
     def run(self):
-        with lock:
-            # Try to do the vacuum every 3 hours
-            call_command("vacuumsqlite", interval=10)
+        # Try to do the vacuum every 3 hours
+        call_command("vacuumsqlite", interval=10)
 
 
 def stop(pid=None, force=False):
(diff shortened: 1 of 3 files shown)


kolibri/utils/server.py
```diff
@@ -46,6 +46,9 @@
 # Currently non-configurable until we know how to properly handle this
 LISTEN_ADDRESS = "0.0.0.0"
 
+# use locks so vacuum doesn't conflict with ping
+lock = threading.Lock()
```
benjaoming: Maybe rename this lock to reflect its intention. What is it locking?

kolibri/utils/server.py
```diff
@@ -115,8 +119,9 @@ def start_command(cls):
         thread.start()
 
     def run(self):
-        # Try to do the vacuum every 3 hours
-        call_command("vacuumsqlite", interval=10)
+        with lock:
+            # Try to do the vacuum every 3 hours
```
benjaoming: Wondering if this comment makes sense when it says `interval=10` ?

kolibri/utils/server.py
```diff
@@ -115,8 +119,9 @@ def start_command(cls):
...
+            call_command("vacuumsqlite", interval=10)
```
benjaoming: Isn't it the `vacuumsqlite` command that does the blocking wait? In that case, it should have the lock as an argument in the call or import it from `kolibri.utils.server`?

benjaoming: ...otherwise we risk a deadlock because it never yields the lock for the other waiting thread?

benjaoming (changes requested):
I like the solution with locks, those are straight-forward.

Regarding testing: Just like the error showed up, perhaps merging this as is would reveal if it was the sole cause.

kolibri/utils/server.py
ralphiee22: planning to change that according to the tagged issue

benjaoming (approved):
Regarding testing: Just like the error showed up, perhaps merging this as is would reveal if it was the sole cause.

kolibri/utils/server.py
```diff
@@ -116,7 +119,7 @@ def start_command(cls):
 
     def run(self):
         # Try to do the vacuum every 3 hours
-        call_command("vacuumsqlite", interval=10)
+        call_command("vacuumsqlite", interval=True)
```
indirectlylit: can we rename this from `interval` to `scheduled`, `repeat`, or something similar?

indirectlylit (approved):
thank you!

PR #4381, opened 2018-10-09
Testing Github Checks

benjaoming:
### Summary

Having deleted the deprecated Travis integration, I had to close the old PR since it continued to have it enabled.

 * [ ] Sort out if all currently open PRs maybe have to be re-opened
 * [ ] Sort out what should happen to the Read The Docs implementation which seems to be status based as well

### Reviewer guidance

None for now.


rtibbles:
I think we're back on the right track now!

PR #4387, opened 2018-10-10
i18n tooling bug fix and messages update

rtibbles:
### Summary
Don't import from django in build tools.
Update message files.

### Contributor Checklist

### Reviewer Checklist


build_tools/i18n/utils.py
```diff
@@ -52,6 +51,19 @@
...
 
 
+def to_locale(language):
+    """
+    Turns a language name (en-us) into a locale name (en_US).
+    """
+    p = language.find('-')
+    if p >= 0:
+        if len(language[p + 1:]) > 2:
+            return language[:p].lower() + '_' + language[p + 1].upper() + language[p + 2:].lower()
```
indirectlylit, line 61: what is an example where this code would get run?

could we simplify this function to something like 

```python
codes = language.split("-");
if len(codes) == 2:
    return "{}_{}".format(codes[0], codes[1].upper())
return codes[0]
```

or does that not handle all cases?

rtibbles: This is directly copied from the Django source, the only thing I removed was the `lower` optional kwarg, which renders the second half into lower case.

The second part of the lang code can be title case, if the second piece is a script code, e.g. zh_Hans or zh_Hant, hence the conditional behaviour if the second part is longer than 2 characters.

PR #4402, opened 2018-10-12
Bump develop to 0.12.0.dev and merge 0.11.x into develop

benjaoming:
Routine merge. Contains documentation stuff.


rtibbles (approved):

rtibbles:
Looks good, will merge when we get buildkite going again, just to make sure we don't break the build!

benjaoming:
@rtibbles as you can see from the latter commits, I had to fix some versioning stuff... this should probably be backported to 0.11 afterwards.. but I'm still unsure why the pattern matching is failing in some cases, and not failing in others.

rtibbles:
Do we have test cases for the different weird failures?

benjaoming:
@rtibbles no :) the thing is that they happened when calling Git with the wrong command line argument.. a reg ex pattern.. it's out of scope IMO :)

benjaoming:
Just flagging: this isn't ready yet, because the version is still not properly detected from `git describe`

Previous Buildkite error was about mal-detecting a .whl file, where there was a '.' (dot) in the filename instead of a '.' (dash).

benjaoming:
Oh wait... it is fine: We have:

```
$ git describe --tags
v0.11.0-beta1-6-g8ce8e617f
```

That does only know about the previous 0.11 tag. We are now 0.12 alpha 0 mode.. so that mode does not know or care about what `git describe` says because of the current logic. It *should* use the commit hash though... but this requires a new type of parsing, which we can hopefully just skip for now and be happy about the date thing. I'll add a TODO.

If the change below fixes this, I think we can push further changes...

PR #4417, opened 2018-10-14
Content Node fetching cleanup

### LINKED ISSUE #4415: Content Node resources are still fetched even when there are no ids to fetch
### Observed behavior
In some places we fetch a set of content nodes by using an `ids` filter. On some occasions, these fetches occur even if this filter is empty - this may lead to odd/unintended behaviour.

### Expected behavior
Only do the fetch if ids is not empty.

rtibbles:
### Summary
* Adds conditional logic to prevent fetches of empty `ids` arrays.
* Uses ContentNodeSlim when only summary information about ContentNodes is being displayed.
* Removes unnecessary call to ContentNode endpoint when coach report API returns `num_coach_contents` data already.

### Reviewer guidance
Do exams and lessons display properly when they have no resources in them?
Does the channels coach report still display properly, (especially if a channel has coach contents)?
Does the Lesson Summary coach report display properly (especially if a lesson has coach contents)?

### References
Fixes #4415

----

### Contributor Checklist

### Reviewer Checklist

### DIFF
--- a/kolibri/core/assets/src/exams/utils.js
+++ b/kolibri/core/assets/src/exams/utils.js
@@ -54,12 +54,17 @@ function getExamReport(store, examId, userId, questionNumber = 0, interactionInd
         const questionSources = exam.question_sources;
 
         const questionList = createQuestionList(questionSources);
+        let contentPromise;
 
-        const contentPromise = ContentNodeResource.fetchCollection({
-          getParams: {
-            ids: questionSources.map(item => item.exercise_id),
-          },
-        });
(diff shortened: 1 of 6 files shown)


jonboiser (approved):
Looks good, but I think the first solution I had in mind was to either implement this logic at the Viewset level, or the in the JS Resource class's fetchCollection method. Then we could avoid writing the conditionals, since it seems it is almost always a mistake to call a contentnode* endpoint with no filters.

PR #4427, opened 2018-10-15
Font updates

### LINKED ISSUE #4392: fonts are broken in grids
### Observed behavior

some fonts are not Noto Sans, e.g. "2 resources in this lesson" below

image

### Expected behavior

consistent fonts

### User-facing consequences

UI looks wonky

### Context

0.11

### LINKED ISSUE #4418: Custom icons in Video Player are missing
### Observed behavior
@christianmemije reports:
Video player is missing custom icons, never got the points snackbar or completion icon

screen shot 2018-10-23 at 3 56 58 pm

### Expected behavior
All icons should show up.

### User-facing consequences
Alarm, confusion, despair.

### Steps to reproduce
Play a video?

### Context
 * 0.11.0-beta1

indirectlylit:
### Summary

* simplify i18n update process to a single command, `make i18n-download branch=[...]`
* update documentation
* major bug fix where subsets weren't being properly generated (missing critical ligatures in some languages)
* removes 'content' font variant - not particularly helpful and lots of extra bandwidth
* fixes #4392 (font rendering in grids)
* fixes #4418 (buttons missing in video player)

### Reviewer guidance

this PR is mostly for code review. Will submit a subsequent PR with regenerated files

### References

continuation of #4254

----

### Contributor Checklist

### Reviewer Checklist


Makefile
```diff
@@ -36,12 +36,11 @@ help:
 	@echo "--------------------"
 	@echo ""
 	@echo "i18n-extract: extract all strings from application (both front- and back-end)"
-	@echo "i18n-crowdin-upload branch=<crowdin-branch>: upload strings to Crowdin"
-	@echo "i18n-crowdin-download branch=<crowdin-branch>: download strings from Crowdin and compile"
-	@echo "i18n-crowdin-stats branch=<crowdin-branch>: output information about translation status"
+	@echo "i18n-upload branch=<crowdin-branch>: upload sources to Crowdin"
+	@echo "i18n-download branch=<crowdin-branch>: download strings from Crowdin and regenerate files"
+	@echo "i18n-stats branch=<crowdin-branch>: output information about translation status"
```
benjaoming, line 43: The intention was to highlight the interaction with CrowdIn and have a command just for that.

I'm not sure if I think it's a great idea to merge Makefile commands into monolithic commands if they are already split for either intentional or natural reasons... it removes ability to quickly do a targeted run.

Also, now we can't do any font work without CrowdIn access and vice versa?

Makefile
```diff
@@ -36,12 +36,11 @@ help:
...
 	@echo "i18n-django-compilemessages: compiles .po files to .mo files for Django"
 	@echo "i18n-install-font name=<noto-font>: Downloads and installs a new or updated font"
-	@echo "i18n-generate-fonts: Generates custom fonts and CSS based on current languages and app strings"
```
benjaoming, line 44: I liked having this separately :)

benjaoming: Could it just depend on the `i18n-crowdin-download` recipe instead?

Makefile
benjaoming: Suggestion: Could we keep the `i18n-download` command as an alias for `i18-crowdin-download + i18n-generate-font`?

benjaoming (dismissed):
Not sure that the "merging" of the Makefile recipes into one monolithic recipe will serve us.

Developer time is often wasted when waiting for long scripts to run that the developer doesn't understand and therefore does not achieve to cherry-pick essentials from. At least, I'm speaking for myself here :)

rtibbles:
I am inclined to agree with Ben that it can be helpful to have intermediate steps, even if there is one 'command to rule them all' that we ordinarily run.

Makefile
benjaoming: Alternatively, as mentioned above, we could make a `i18n-download` recipe that just depends on `i18-crowdin-download` + `i18-generate-fonts`..

indirectlylit:
I'm willing to add the granular commands back in, but the purpose is not clear it me because it's possible to run the individual steps by inspecting the makefile and running the individual commands listed there.

Having the Makefile include granular commands that are just proxies to other commands seems unnecessarily indirect and not particularly helpful?

At a high level by merging commands I was trying to cut out the number of places that need to be maintained, which right now are:

* developer docs rst files
* documentation/help text for command-line interfaces in individual scripts
* argument parsers in individual scripts
* documentation/help text for command-line interfaces in Makefile
* argument parsers in Makefile

indirectlylit:
In addition to the issues noted above, I find Makefiles to be a very poor place to construct complex command-line UIs because:

1. there's no sense of a "private function" in Makefiles - every command becomes part of the public interface, so code re-use is difficult
2. you can cleanly call other commands _before_ a given command, but to call one _after_ you actually have to launch a new `make` process
3. unlike Python's [`argparse`](https://docs.python.org/3/library/argparse.html) or Node's [`commander`](https://github.com/tj/commander.js/), it's challenging to cleanly incorporate help text, validation, etc

I think Makefiles are very good at doing things you'd use a Bash script for, but we shouldn't try to make them much more than that...

benjaoming:
>  it's possible to run the individual steps by inspecting the makefile and running the individual commands listed there

The Make recipies package easier/simpler invocation, together with documentation.. in this case for memorizing the `branch` argument. You can try to run just `make`, you will see the help texts. So it's nice to have them available, even if just very few lines.

There's also a hidden disruptive cost everytime that development procedures are changed, but that's another thing.

> In addition to the issues noted above, I find Makefiles to be a very poor place to construct complex command-line UIs because:

That would be a long discussion :smile: 

> I  think Makefiles are very good at doing things you'd use a Bash script  for, but we shouldn't try to make them much more than that...

That's also the effort that I try to put in it. A _KISS_ way of having build commands that are easy to understand for everyone in the community and easy to document.

indirectlylit:
Do you have any thoughts on this?

> At a high level by merging commands I was trying to cut out the number of places that need to be maintained, which right now are:
> 
> * developer docs rst files
> * documentation/help text for command-line interfaces in individual scripts
> * argument parsers in individual scripts
> * documentation/help text for command-line interfaces in Makefile
> * argument parsers in Makefile

benjaoming:
I don't know if adding Docker to the equation will help maintaining all this.. probably the contrary! But it will simplify invocation (hopefully) of the most normal commands and documentation. However, the i18n stuff will remain on the side.

Makefile
```diff
@@ -1,5 +1,5 @@
-# Specifies the targets that should ALWAYS have their recipies run
-.PHONY: help clean clean-pyc clean-build clean-assets writeversion lint test test-all coverage docs release i18n-crowdin-upload i18n-crowdin-download
+# List most target names as 'PHONY' to prevent Make from thinking it will be creating a file of the same name
+.PHONY: help clean clean-assets clean-build clean-pyc clean-docs lint test test-all assets coverage docs release test-namespaced-packages staticdeps staticdeps-cext writeversion buildconfig pex i18n-extract-frontend i18n-extract-backend i18n-extract i18n-django-compilemessages i18n-upload i18n-download i18n-regenerate-fonts i18n-stats i18n-install-font docker-clean docker-whl docker-deb docker-deb-test docker-windows docker-demoserver docker-devserver
```
indirectlylit, line 2: @benjaoming are we using any of our make targets to actually create files of the same name, except maybe `dist`?

I updated this to include essentially everything based on my understanding [from the docs](http://www.gnu.org/software/make/manual/make.html#Phony-Targets) and [this SO post](https://stackoverflow.com/questions/2145590/what-is-the-purpose-of-phony-in-a-makefile)

> `.PHONY: help clean clean-assets clean-build clean-pyc clean-docs lint test test-all assets coverage docs release test-namespaced-packages staticdeps staticdeps-cext writeversion buildconfig pex i18n-extract-frontend i18n-extract-backend i18n-extract i18n-django-compilemessages i18n-upload i18n-download i18n-regenerate-fonts i18n-stats i18n-install-font docker-clean docker-whl docker-deb docker-deb-test docker-windows docker-demoserver docker-devserver`

benjaoming: @indirectlylit yes, that's the case, so failing to mark a command as ".PHONY" would not likely have any immediate consequences, except in the case of `make dist` because of the `dist` dir.

We _could_ use this pattern of adding `.PHONY` right next to each command... https://gist.github.com/ajmazurie/d8edac58f6774ed4b8cc7dc9c0c81a90

indirectlylit: yes that looks easier to maintain

indirectlylit:
FYI I have another update coming soon to handle arbitrary languages that might be encountered in our content

@rtibbles @benjaoming would you like to review/merge this first, or should I just add to it?

Makefile
rtibbles: This discussion `make`s me think that we, in general, overload the Makefile.

build_tools/i18n/utils.py
```diff
@@ -51,16 +52,24 @@
...
 
+# Keys used in simple internal 'font info' dictionaries
+KEY_NAME = "name"
+KEY_HAS_UI = "has_ui_variant"
+
 
 def to_locale(language):
     """
     Turns a language name (en-us) into a locale name (en_US).
+    Logic is derived from Django so be careful about changing it.
```
rtibbles, line 63: I enjoy the juxtaposition of this comment and the extensive diff that follows it :)

rtibbles (approved):
Code makes sense to me.

build_tools/i18n/utils.py
indirectlylit: ha that was the `black` autoformatter

indirectlylit: logic is the same!

Makefile
benjaoming: A long Makefile isn't complex, it's just long :)

benjaoming (approved):
Comments addressed, thanks @indirectlylit ! :+1:

indirectlylit:
thanks - waiting to merge pending #4461

indirectlylit:
merging in - fonts may break briefly, bug #4461 coming soon

PR #4458, opened 2018-10-18
Fix channel file size summaries in import/export workflows

### LINKED ISSUE #3925: Disable 'select all' at topics  with zero importable contents (importing from USB)
### Observed behavior

(1) During import (more) from external drive workflow, topics that have zero available leaf nodes still show a clickable 'select all' checkbox. If a user were to click it, the 'resources selected' estimate gets increased, even though in practice nothing will get imported.

screen shot 2018-06-28 at 2 06 22 pm

(2) This also happens if everything on the device has already been imported (and cannot be imported again)

screen shot 2018-06-28 at 2 14 36 pm

(3) Another thing that can happen is a user could toggle a topic whose child topics are also empty. Navigating to the child topic will show an empty page but with a checked 'select-all' checkboxes. 

### Expected behavior

1. There is a reasonable empty state and no select-all checkboxes for these completely-unavailable topics
2. In scenario (3), the system could automatically remove the node from the to-import list to correct the import estimate.

### LINKED ISSUE #4452: file/resource size discrepancy in Export
### Observed behavior

from https://github.com/learningequality/kolibri/issues/4409#issuecomment-429510003 @radinamatic  says:

> Not sure if it's connected, but when trying to export to local drive, total size & number of resources in the channel is reported as 0, together with the number of resources already imported on the device, while their size is apparently reported correctly.
> 
> ubuntu 16 04 3 start running - oracle vm virtualbox_024

@jonboiser says:

> The file/resource size discrepancy in Export is an issue that seems to have stemmed from some of the speed up work in #3885.

### Expected behavior

> Need to make sure that the full size statistics are available at parts of the workflow that require it. Previously, we relied on all the file/resource stats being available because they came as a group (now they can be selected as needed)

### User-facing consequences

incorrect information about file sizes

### Steps to reproduce

export, and compare resource count with size

### Context

0.11

jonboiser:
### Summary

1. This fixes the incorrect displays of file/resources size for channels. How? By explicitly requesting these statistics and placing them in the `transferredChannel` part of `manageContent.wizard.state` (and having views rely on this data), rather than in `manageContent.channelList`, which does not have the full set of counts. Fixes #4452 
1. Disables the 'select all' checkbox in import/export at topics. mostly Fixes #3925 

### Reviewer guidance

1. See Gherkin stories for channel sizes. Test to see that the numbers are correct for all kinds of workflows. If testing against Khan Academy English version 13, there are 12473 resources, 39 GB file size.
1. To test 'select all', import a subset of a channel with some empty topics. Try to export the channel to a USB drive. When going to an empty topic, the select-all button should be disabled.

### References

…

----

### Contributor Checklist

### Reviewer Checklist

### DIFF
--- a/integration_testing/features/admin/admin-manage-channel-sizes.feature
+++ b/integration_testing/features/admin/admin-manage-channel-sizes.feature
@@ -1,0 +1,26 @@
+Feature: Admin Device Management Channel Sizes
+  Admin should see the correct file sizes and resource counts regarding a channel
+
+  Background:
+    Given I am signed into Kolibri as an admin user
+      And I am on the *Device > Channels* page
+
+  Scenario: Correct file sizes when transferring content
+    When I am doing an <transfer_workflow>
+      And I go to the *Select Content Page* for <channel_name>
+    Then In the *Size: Total size* cell, it says <total_file_size>
+      And In the *Size: On your device* cell, it says <on_device_file_size>
+      And In the *Resources: Total size* cell, it says <total_resources>
+      And In the *Resources: On your device* cell, it says <on_device_resources>
+
+Examples:
+| channel_name        | on_device_file_size | total_file_size | on_device_resources | total_resources |
+| Khan Academy        | 1 GB                | 10 GB           | 10                  | 100             |
+
+Examples:
+| transfer_workflow               |
+| Import from Kolibri Studio      |
+| Import more from Kolibri Studio |
+| Export to a USB drive           |
+| Import from a USB drive         |
+| Import from a Peer              |
--- a/kolibri/plugins/device_management/assets/src/modules/wizard/actions/selectContentActions.js
+++ b/kolibri/plugins/device_management/assets/src/modules/wizard/actions/selectContentActions.js
@@ -1,28 +1,31 @@
 import client from 'kolibri.client';
 import urls from 'kolibri.urls';
 import { downloadChannelMetadata } from '../utils';
+import { getChannelWithContentSizes } from '../apiChannelMetadata';
 
 /**
  * Transitions the import/export wizard to the 'load-channel-metadata' interstitial state
  *
  */
-export function loadChannelMetaData(store) {
+export function loadChannelMetadata(store) {
   let dbPromise;
   const { transferredChannel } = store.state.manageContent.wizard;
   const channelOnDevice = store.getters['manageContent/channelIsInstalled'](transferredChannel.id);
 
-  // Downloading the Content Metadata DB
-  if (!channelOnDevice) {
-    // Update metadata when no content db has been downloaded
-    dbPromise = downloadChannelMetadata(store);
-  } else if (!channelOnDevice.available && channelOnDevice.version < transferredChannel.version) {
-    // If channel _is_ on the device, but not "available" (i.e. no resources installed yet)
-    // _and_ has been updated, then download the metadata
+  // If channel _is_ on the device, but not "available" (i.e. no resources installed yet)
+  // _and_ has been updated, then download the metadata
+  const newChannelDbAvailable =
+    channelOnDevice &&
+    !channelOnDevice.available &&
+    channelOnDevice.version < transferredChannel.version;
+
+  // Update metadata when no content db has been downloaded or if it is stale
+  if (!channelOnDevice || newChannelDbAvailable) {
     dbPromise = downloadChannelMetadata(store);
   } else {
     // If already on device, then skip the DB download, and use on-device
     // Channel metadata, since it has root id.
-    dbPromise = Promise.resolve(channelOnDevice);
+    dbPromise = getChannelWithContentSizes(transferredChannel.id);
   }
 
   // Hydrating the store with the Channel Metadata
@@ -32,8 +35,6 @@ export function loadChannelMetaData(store) {
       // Replacing them here with canonical type from ChannelResource.
       store.commit('manageContent/wizard/SET_TRANSFERRED_CHANNEL', {
         ...channel,
-        total_resources: transferredChannel.total_resources,
-        total_file_size: transferredChannel.total_file_size,
         version: transferredChannel.version,
         public: transferredChannel.public,
       });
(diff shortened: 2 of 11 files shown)


radinamatic:
Will try to test this asap, just a couple of doubts:

> See Gherkin stories for channel sizes

There are no Gherkin scenarios (yet) covering specifically channel sizes, mostly because those change regularly. I'm guessing the idea is to test all the _import_ scenarios making note of the _current_ channel sizes...? 

Currently, full KA EN channel has 34646 resources as reported on Studio and on my just-updated-to-version-13 dev repo. Where does the number of 12473 resources come from?

jonboiser:
> Currently, full KA EN channel has 34646 resources as reported on Studio and on my just-updated-to-version-13 dev repo. Where does the number of 12473 resources come from?

I think you might have found another bug :(

The `devicechannel` API (which is is derived from the on-device channel db) returns this

```
author: ""
available: true
description: "Khan Academy content for English."
id: "1ceff53605e55bef987d88e0908658c5"
...
total_file_size: 39646022107
total_resources: 12473
version: 13
```

the `remotechannel` API (e.g. from studio) returns this:

```
description: "Khan Academy content for English."
id: "1ceff53605e55bef987d88e0908658c5"
included_languages: {gu: "ગુજરાતી", zh-TW: "正體字/繁體字", fr-CA: "Français (Canada)", sco: "Scots", gn: "Avañeẽ", gl: "Galego",…}
lang_code: "en"
lang_name: "English"
last_updated: "2018-09-25T14:02:25.135777Z"
name: "Khan Academy (English)"
public: true
...
total_file_size: 39891286668
total_resources: 34646
version: 13
```

rtibbles:
@jonboiser @radinamatic this is because @ralphiee22 has deduped the resources by `content_id` in order to not artificially inflate resource counts. He has had a matching PR merged on Studio, but I don't think it's made it into production yet.

radinamatic:
> I think you might have found another bug :(

Yep, that's me, talent to step into mud holes... 😈 

Ok, until somebody steps in to fill that hole, what should I take as the correct number to test this PR?

jonboiser:
@radinamatic I'm not sure. I think basically the "total" file/resources counts should be consistent between workflows. (e.g. no more having 0/0)

On this PR, at least, the counts should be the smaller counts I put in the original post.

jonboiser:
Also, I added a basic gherkin file to remind testers there are **5** different import/export flows.

radinamatic:
Thank you for the pickle! 💯

radinamatic:
Tested on Windows 10, mostly hits and some misses, almost certainly not related to this PR, but with roots in other mud holes.

**Import from Studio** & **Import from USB** - all's good!



**Export to USB**



Values in all 4 cells apparently well reported, but still one thing: (not related to this fix as I know some resources can be annotated in more than one node, but still potentially confusing to the user): _If I imported only 23 resources, how come now I can choose from 48+10 to export? If total is 12k resources, why there are almost 25k in subtopics?_

But when user arrives to the subtopic they know they can actually export, there is no info about the number of resources. Was this intentional?



**PtP Import more from Kolibri Beta**



Again, stuff related to this PR seems to be working OK, but I lost a ton of time trying to figure out why it was not importing all the content correctly. Hint - not only the PtP device needs to be running the Kolibri 0.11 in order to be saved as a network address, but it also needs to have the same channel version as the one on the device where user wants to import to (cc @jamalex). Here I first imported portion of KA version 13 from Studio, and then tried fruitlessly to import from Kolibri Beta several times, no errors reported, until the bulb finally went on and I confirmed that Kolibri Beta had KA version 12. Anyways, I'm venting, and will put all this in another issue...

Case 2 (`Select all` and empty topics)

Not sure how to test this 😕 
This version of my testing channel has an empty topic on Studio, but it does not get imported at all, and therefore cannot be exported either.

radinamatic:
>  testing channel has an empty topic on Studio, but it does not get imported at all

Or it is imported, just hidden in Kolibri...?

jonboiser:
@radinamatic 	By "empty topic", I mean a topic that has no "available"/downloaded nodes on the server (but is not actually empty in the channel on Studio). For example, If I only download "Math" on KA, then the other subjects (PHysics, English, etc.) will be empty and have no content to export to a USB drive.

When you get to such an empty-on-my-device topic, there should be no checkable nodes, and the "select all" checkbox on the top of the list should also be disabled.

jonboiser:
With respect to the large numbers in the "Export to USB" screenshot (e.g. "48 of 21833 resources on your device"), 21,833 is larger than the total (de-duplicated) resource count for the channel (12473), so my theory is that the counts (`total_resources`) there also need to be de-duplicated in The ContentNodeGranular endpoint.

radinamatic:
@jonboiser Thanks for clarifying that, I was being too literal... :woman_facepalming: 

I think 2. should be fixed too, but I'll recheck (VM is on my other laptop), and confirm!

radinamatic:
@jonboiser Does not work as stated in the 2., but I don't mind this, to say the truth. Basically, nothing that is actually imported on the server, isn't even displayed for export. Would that be a more straightforward solution than selectively disabling the `Select all`?

radinamatic:
Ah, I now see that you have committed something new... Does it need a separate testing?

jonboiser:
@radinamatic 	I got confused about the exact testing scenario for (2). You can get "empty topics" when **importing from a USB drive**, but not the other kinds of workflows. At the moment, we list the channel as if it were complete on USB drive (so all of the topics are listed), but if you were to navigate deeper into the tree, you may eventually discover that the drive has none of the content.

screen shot 2018-10-19 at 2 18 55 pm

radinamatic:
So with the both previous and the latest built asset, the only disabled `Select all` checkbox is the very last one when you navigate the topic. 

In this particular *Import from USB* scenario, only the first 2 *Early math* subtopics are actually on the drive (highlighted green). Wouldn't it be more helpful if *all* the checkboxes of "empty" topics were disabled?

jonboiser:
Yeah, it would be ideal if importing from a USB drive were similar to exporting to one (where you only see topics that are available). At the moment, we cannot look at a given topic and know immediately that there are no "leaf nodes" (i.e. resources) at the end.

In fact, if you were to select one of the non-highlighted topics (which are actually empty), the UI will add the total sizes to the "Content selected" estimate.

Disabling "select all" at empty topics avoids a situation that is probably pretty rare (why would somebody select all in an empty topic?), but prevents adding non-available nodes and artificially increasing the Content selected estimate.

radinamatic:
Understood, possible-workaround-for-the-time-being!

Well, if it works as intended, merge away! :slightly_smiling_face:

radinamatic (approved):
Tested, and working as stated! :+1:

PR #4461, opened 2018-10-19
Moar fonts!!

indirectlylit:
### Summary

Ready for review/merge.

Updates:

* Changes the way that source fonts are installed and referenced. Now download _all_ fonts from the Noto Github repo, and pins us to a particular version of the repo to make things more reproducible and to change source font versions more predictably. Upgrading fonts will mean pointing to a new git reference in the github repo.
* Uses Git LFS to store all font files. Let's try it for a bit and see how it works - we can always explore other strategies if this introduces problems.
* Added some final tweaks and fixes for font loading quirks that exist even in modern browsers
* Adds a new 'common' subset font in addition to the existing language-specific subset font. This slightly increases the amount of data and requests made by the client, but should reduce the size of our kolibri distribution by a few megabytes
* Updates with the latest language data

### Reviewer guidance

Recommend going through commit by commit rather than looking at the overall files changed

### References

re #4427

----

### Contributor Checklist

### Reviewer Checklist


benjaoming:
Hi @indirectlylit - A brain stormy suggestion: One way of working around this could be to package the fonts separately, in their own repo and as their own package. But are they a 1:1 asset of Kolibri with a 1:1 versioning life cycle? Because in that case, it will be a disadvantage. On the other hand, if we only build the fonts at the first release of a new series, then we can have for instance kolibri 0.11.* _depends on_ kolibri-fonts 0.11.

If we can get around having LFS, that would be great - it seems to come with its own platform specific set of [installation instructions](https://help.github.com/articles/installing-git-large-file-storage/)... Would you know if the installation of LFS would be required to check out and build the repo?

indirectlylit:
>  One way of working around this could be to package the fonts separately, in their own repo and as their own package

Yeah, I could see this potentially working. However I would definitely not want to make it a PyPi or NPM package - we currently do that with the perseus renderer and it's really annoying to maintain.

Putting them in a separate repo and just downloading them could work well ... in fact that's what I currently do with the Google Noto repo: https://github.com/learningequality/kolibri/pull/4461/files#diff-da3b8b5004c2cbbf1173eb5a436beaaeR244

> Would you know if the installation of LFS would be required to check out and build the repo?

required to do a real build so font files get included, but optional e.g. for backend development

indirectlylit:
@rtibbles this is ready for review. Recommend looking at individual commits rather than overall files changed

indirectlylit:
cc @christianmemije would you mind testing this also?

christianmemije:
Should update this line https://github.com/learningequality/kolibri/blob/develop/kolibri/plugins/style_guide/assets/src/views/StyleGuideIndex.vue#L80

indirectlylit:
thanks - updated

kolibri/core/assets/src/styles/fonts/noto-manifest.json
```diff
@@ -1,34 +1,105 @@
-[
-  {
-    "has_ui_variant": false,
-    "name": "NotoSans"
+{
+  "fonts": {
+    "NotoSans": {
+      "bold_url": "https://raw.githubusercontent.com/googlei18n/noto-fonts/b20c9c07139b22eb38013b1f80f9a5222c6058c6/phaseIII_only/hinted/ttf/NotoSans/NotoSans-Bold.ttf",
```
rtibbles: Seems odd that we are storing this info in our assets folder when it is, presumably, only used by the build_tools/i18n functionality?

kolibri/deployment/default/settings/base.py
```diff
@@ -173,7 +173,11 @@
 # https://en.wikipedia.org/wiki/List_of_ISO_639-1_codes
 # http://helpsharepointvision.nevron.com/Culture_Table.html
 
-LANGUAGES = i18n.get_supported_languages(KOLIBRI_MODULE_PATH)
+# Kolibri format
+KOLIBRI_SUPPORTED_LANGUAGES = i18n.get_supported_language_info(KOLIBRI_MODULE_PATH)
```
rtibbles: I think this is better defined in the `i18n` module, as far as possible, we avoid defining Kolibri specific variables in the Django settings file, because the assumption is that they could, in principle, be overwritten by a third party user.

kolibri/utils/i18n.py
```diff
@@ -33,7 +33,7 @@ def get_installed_app_locale_path(appname):
         return module_locale_path
 
 
-def get_supported_languages(kolibri_module_path):
+def get_supported_language_info(kolibri_module_path):
```
rtibbles: We can probably refactor this function to not need the argument, and just use the relative path (there's no reason that the file should move relative to this).

```suggestion
def get_supported_language_info():
    """
    Returns a list of dicts with all the language information
    """
    file_path = os.path.join('..', "locale", "supported_languages.json")
    with io.open(file_path, encoding="utf-8") as f:
        return json.load(f)

KOLIBRI_SUPPORTED_LANGUAGES = get_supported_language_info()
```

kolibri/core/assets/src/views/language-switcher/LanguageSwitcherList.vue
```diff
@@ -12,14 +12,15 @@
       />
     </UiIconButton>
 
-    <span class="selected">
-      {{ selectedLanguage }}
+    <span class="selected" :title="selectedLanguage.english_name">
```
rtibbles, line 15: I am not a fan of this, why english?

rtibbles (changes requested):
One thing I'd definitely like to see changed (`KOLIBRI_SUPPORTED_LANGUAGES` location), a couple of questions, and a bit concerned about the English only tooltips.

kolibri/core/assets/src/styles/fonts/noto-manifest.json
indirectlylit: it's not in `/kolibri/core/static`, it's in `/kolibri/core/assets/src/styles/fonts`.

I put it there because I was thinking of it as a "source" file, though in reality it's more similar to stuff in `node_modules` or `site_packages`...

Where do you recommend saving it to?

rtibbles: My assumption, because I could only see it being used in build_tools/i18n is that it would be stored in that folder?

kolibri/utils/i18n.py
indirectlylit: Yeah I agree.

I've been blindly following how it was originally implemented - it looked weird but I assumed there was a reason for it so I didn't want to refactor... If you think it's safe to update I'll do that

kolibri/core/assets/src/views/language-switcher/LanguageSwitcherList.vue
indirectlylit: It stemmed from [this recommendation](https://github.com/learningequality/clearinghouse/issues/274) from @radinamatic:

>  consider fitting in both the English and the native forms, like in Windows and macOS above. This could improve the visibility of the available languages, and could also make QA a bit easier - in the future it may become necessary to test a specific locale for something, and the native forms in non-Latin alphabets are really difficult to distinguish.

I agree that really every language name should be translated into every other language, and the tooltips should be localized the same as all other application text. However, I think there's also value and not much harm in adding English tooltips now and translating them in the next release.

indirectlylit: (cc @radinamatic)

kolibri/utils/i18n.py
rtibbles: The only reason it was like this was because we were not using the information in the json file in any way except to generate the Django `LANGUAGES` setting previously. Seems safe to me!

kolibri/core/assets/src/views/language-switcher/LanguageSwitcherList.vue
rtibbles: :okay:

radinamatic: I second that the added value here outweighs the potential harm of having (even) English-only labels/tooltips. 

If the concern is that this would add unnecessary burden the translators workload, then *as a translator* I'd say that having 20 one-word strings more (atm, let's say 20 more in a year, optimistically) with language names that will almost certainly be offered as a MT suggestions that require only 2 clicks to approve, is far, far less of a headache than many of the already present Kolibri UI strings, and many more to come in future... :slightly_smiling_face:

rtibbles: Yeah, not worried about the translation, just the untranslated part!

indirectlylit: created a follow-up issue for this: #4483

rtibbles:
A completely unrelated comment. For science.
@bubbles of 3 in 5 minutes

rtibbles:
In the meantime, this is a completely unrelated test:
@bubbles of 3 in 5 minutes

indirectlylit:
@rtibbles  comments addressed - please re-review

rtibbles (approved):

PR #4464, opened 2018-10-23
Move linting out of the build chain

### LINKED ISSUE #4377: Separate linting of frontend assets from loaders in the build pipeline
### Observed behavior
Currently we predominantly use loaders to apply linting to all files. These are done as pre-loaders, meaning that all files must be linted before being passed to other loaders for code transpilation. Some quick measurements with https://www.npmjs.com/package/speed-measure-webpack-plugin show that these loaders take up a large amount of the compilation time on a cold start build, leading to poorer dev performance.

### Expected behavior
Our dev processes should auto-lint code, but it is not necessary for this to happen prior to any build processes, especially as during development we do not fail on linting.

There are two alternatives:
1) To put all linting into plugins so that they happen asynchronously to the main build (and do not slow down transpilation).
2) To put all linting into separate processes that run separate watchers based on glob paths.

The advantages of 1) are that we already do this for the prettier linting. The advantages of 2) is that it makes our linting process completely separate, it also means that we have a clearly defined set of linting that we can do either as a pre-commit hook or a CI check, and that we can now lint all frontend files automatically, regardless of whether they are in the build pipeline or not, such as test spec files.

### User-facing consequences
Don't keep devs a-waiting?

rtibbles:
### Summary
Lint loaders are a convenient mechanism as they avail themselves of the convenient rules mechanism of webpack to determine which files they should effect.
However, because loaders are applied sequentially to code, every time the dev server is started, all the code in the entire codebase is linted. This slows down a cold dev build startup time a lot.
This PR resolves that by consolidating all frontend linting logic into the `pretter_frontend` function/command line tool.
Further, it adds a `monitor` mode to allow watching of files, only running linting on these files when changes occur. The watcher it uses is `chokidar`, the same watcher used under the hood by webpack, brunch, etc.
As a bonus, I decided to also parallelize our webpack builds while I was at it, which means that all of your CPU cores now belong to webpack.

### Reviewer guidance
When you start the devserver, does your computer nearly become unresponsive, and hangouts start stuttering, and the fan start going like it's starting to achieve lift off? Do you no longer have time to make a pour over coffee in the time it takes for the dev server to cold start?

Places I would like suggestions:
* Should I limit the number of cores that it uses (I can do this by forking n - 1 processes and then running multiple devservers inside each child process)?
* I think there is probably a bit too much terminal spam in there at the moment, from a couple of sources:
    1. The translation extraction plugin is using the deprecated `plugin` interface in webpack, this warning now gets spammed N times (for N webpack bundles).
    2. The Python part of the frontend still initializes some parts of the Kolibri stack - I would like this not to be the case at all, so that this part could be completely silent if possible.
* Parallelization of a full build for local debug testing purposes?

Should these hold up merging, or should I do these as follow up issues?

Finally, this removed what I *thought* was the barrier to hot module reloading in webpack, and I tried using it. While the hot update JS was loaded into the frontend, I did not see any HMR behaviour.

### References
Fixes #4377

----

### Contributor Checklist

### Reviewer Checklist


frontend_build/src/webpackdevserver.js
```diff
@@ -1,37 +1,60 @@
...
+    process.exit(code);
+  });
+} else {
+  const index = Number(process.env.index);
+  const bundle = bundles[index];
+  const port = devServerConfig.port + index;
+  const address = devServerConfig.address;
+  const basePath = devServerConfig.basePath;
+  const publicPath = 'http://' + address + ':' + port + '/' + (basePath ? basePath + '/' : '');
+  bundle.output.publicPath = publicPath;
```
rtibbles, line 30: For this curious, this (and the extra option below) is how I attempted to activate HMR.
```suggestion
  bundle.output.publicPath = publicPath;
  bundle.plugins.push(new webpack.HotModuleReplacementPlugin());
```

frontend_build/src/webpackdevserver.js
```diff
@@ -1,37 +1,60 @@
...
+  var server = new WebpackDevServer(compiler, {
+    // webpack-dev-server options
+
+    // contentBase: "http://localhost:3000/",
+    // Can also be an array, or: contentBase: "http://localhost/",
+
+    // Set this as true if you want to access dev server from arbitrary url.
+    // This is handy if you are using a html5 router.
+    historyApiFallback: false,
+
```
rtibbles, line 41:
```suggestion
hot: true,
```

frontend_build/src/prettier-frontend.js
```diff
@@ -26,93 +49,162 @@ function prettierFrontend({ file, write, encoding = 'utf-8', prettierOptions })
...
+            })
+            .then(output => {
+              if (output.results) {
+                messages.push(
+                  stylelintFormatter(
+                    output.results.map(message => {
+                      message.line += lineOffset;
+                      message.column += columnOffset;
+                      return message;
+                    })
```
indirectlylit, line 93: getting very deeply nested here .. maybe some intermediary variables to flatten things a bit?

frontend_build/src/prettier-frontend.js
```diff
@@ -26,93 +49,162 @@ function prettierFrontend({ file, write, encoding = 'utf-8', prettierOptions })
...
+        if (esLintOutput.messages.length) {
+          messages.push(esLintFormatter([esLintOutput]));
+        }
+        return linted;
+      }
+      function lintScss(code, { lineOffset = 0, columnOffset = 0 } = {}) {
+        options.parser = 'scss';
+        let linted = prettier.format(code, options);
+        if (linted.trim() !== code.trim()) {
+          notSoPretty = true;
```
indirectlylit, line 75: aww

indirectlylit (approved):
This seems to be working fairly well. My inclination would be to just merge it in and iterate on as we learn and get feedback rather than letting it linger in a feature branch.

One thing I noticed at one point which was rather weird (maybe related to multiple threads?) was that it said "compilation succeeded" but then I still got an "in progress" error. If in fact that's what happened, it would be best to only report "compilation succeeded" once _all_ threads have finished?

.pre-commit-config.yaml
```diff
@@ -17,19 +17,7 @@
...
-          entry: node node_modules/eslint/bin/eslint.js
-          language: system
-          files: \.(js|vue)$
-      -   id: frontend-stylelint
-          name: Lint SCSS
-          description: This hook checks that SCSS conforms to our Stylelint standards.
-          entry: node node_modules/stylelint/bin/stylelint.js
-          language: system
-          files: \.(scss|vue)$
+          files: \.(js|vue|scss)$
```
christianmemije, line 23: We should also lint css files?

christianmemije:
OS froze completely. Had to restart. Mission accomplished?

rtibbles:
Ouch - yeah, we might want to not use all cores.

rtibbles:
I am doing a fairly substantial refactor and rewrite of this too as a follow up, so if we can merge this as is, a lot of the clean up will follow.

Also happy to make minor tweaks if needed.

indirectlylit:
if it literally caused his cpu to freeze and crash, might not want merge quite yet? :)

rtibbles:
Well, it sounds like an OS issue, but I'll close this in favour of the much larger PR that is about to land.

PR #4494, opened 2018-10-31
Cache dataset_id lookups for logs and remove related lookups

ralphiee22:
### Summary

The main motivation behind this PR is to optimize deserialization (serialized store models being loaded into the django app layer). When we deserialize, for each model we call `save`, which calls `ensure_dataset`, which does lots of db related lookups. This adds up and makes this operation slow.

Performance Improvement for same number of records:
**BEFORE:**
```
('func: ', 'deserialization')
('queries:', 2955)
('READS: ', 2363)
('WRITES: ', 582)
took: 9.25s
```

**AFTER:**
```
('func: ', 'deserialization')
('queries:', 1655)
('READS: ', 1063)
('WRITES: ', 582)
took: 5.29s
```

Proposed solution in this PR:
Change related lookup calls in `infer_dataset` calls from `self.user.dataset` -> `self.user.dataset_id` (reduces read queries by one).
Also caching dataset lookup calls for logs (saves one read query per log). 

Possible point of contention: One thing is that sqlite does not enforce foreign key constraints so it may be possible to save an object with a non existent `dataset_id`.
Possible solution: We can enable foreign key constraints (if the sqlite version supports it) which would avoid the above https://www.sqlite.org/foreignkeys.html#fk_enable

### Reviewer guidance

Do tests pass?
Does using the app as normal work or noticing any strange behavior?

----

### Contributor Checklist

### Reviewer Checklist

### DIFF
--- a/kolibri/core/auth/models.py
+++ b/kolibri/core/auth/models.py
@@ -29,6 +29,7 @@ from django.contrib.auth.models import AbstractBaseUser
 from django.contrib.auth.models import AnonymousUser
 from django.contrib.auth.models import UserManager
 from django.core import validators
+from django.core.cache import cache
 from django.core.exceptions import ObjectDoesNotExist
 from django.core.exceptions import ValidationError
 from django.db import models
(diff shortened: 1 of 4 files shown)


kolibri/core/auth/models.py
```diff
@@ -159,6 +160,23 @@ class AbstractFacilityDataModel(FacilityDataSyncableModel):
...
+    def cache_related_dataset_lookup(self, related_obj_name):
+        """
+        Attempt to get the dataset_id either from the cache or the actual related obj instance.
+
+        :param related_obj_name: string representing the name of the related object on this model
+        :return: the dataset_id associated with the related obj
+        """
+        field = self._meta.get_field(related_obj_name)
+        key = '{id}_{db_table}_dataset'.format(id=getattr(self, field.attname), db_table=field.related_model._meta.db_table)
+        cached_value = cache.get(key)
```
rtibbles: Mostly stylistic, but I quite like doing this as a single if statement:

```suggestion
        dataset_id = cache.get(key)
        if dataset_id is None:
            dataset_id = getattr(self, related_obj_name).dataset_id
            cache.set(key, dataset_id, 60 * 10)
        return dataset_id
```

kolibri/core/auth/models.py
```diff
@@ -159,6 +160,23 @@ class AbstractFacilityDataModel(FacilityDataSyncableModel):
     class Meta:
         abstract = True
 
+    def cache_related_dataset_lookup(self, related_obj_name):
```
rtibbles: Maybe `cached_related_dataset_lookup`?

rtibbles (commented):
Makes sense, and the reduced reads speak for themselves, a couple of stylistic comments mostly.

indirectlylit:
seems like a significant improvement. What needs to be done to get this merged in?

ralphiee22:
@indirectlylit Nothing, can be merged as is.

indirectlylit (approved):

PR #4495, opened 2018-10-31
Use ImmersiveToolbar in  UserPermissionsPage

jonboiser:
### Summary

Removes usage of ImmersiveFullScreen in UserPermissionsPage and replaces it with the ImmersiveToolBar.

screen shot 2018-10-31 at 10 52 03 am

### Reviewer guidance

1. Workflow should be visually similar to the older version
1. Check accessing the UserPermissionsPage when logged in under different permissions (guest, learner, admin, manage-content only, etc.). The toolbar title should be different.
1. Check exiting the page under different permissions (exiting should go to different places).

### References

…

----

### Contributor Checklist

### Reviewer Checklist

### DIFF
--- a/kolibri/plugins/device_management/assets/src/modules/userPermissions/handlers.js
+++ b/kolibri/plugins/device_management/assets/src/modules/userPermissions/handlers.js
@@ -1,6 +1,12 @@
 import { DevicePermissionsResource, FacilityUserResource } from 'kolibri.resources';
-import ConditionalPromise from 'kolibri.lib.conditionalPromise';
 import samePageCheckGenerator from 'kolibri.utils.samePageCheckGenerator';
+import { createTranslator } from 'kolibri.utils.i18n';
+
+const translator = createTranslator('UserPermissionToolbarTitles', {
+  loading: 'Loading user permissions…',
+  goBackTitle: 'Go Back',
+  invalidUserTitle: 'Invalid user ID',
+});
 
 /**
  * Serially fetches Permissions, then FacilityUser. If returned Promise rejects,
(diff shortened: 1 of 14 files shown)


kolibri/plugins/device_management/assets/src/views/DeviceIndex.vue
```diff
@@ -18,7 +19,12 @@
     </transition>
 
     <div>
-      <component :is="currentPage" />
+      <component
+        :is="currentPage"
+        :appBarTitle.sync="childAppBarTitle"
+        :appBarLink.sync="childAppBarLink"
+        :immersivePageIcon.sync="immersivePageIcon"
```
indirectlylit: This feels like something that might be more conventionally done with vuex?

Curious on the thinking here to using `.sync`

jonboiser: I'd rather not expand Vuex state if I can help it.

Long-term, if we want to manage this toolbar using vuex, we could maybe extract all of the toolbar/corebase-related stuff into a module. Currently, it's tied to the `manageContent` and wizard workflows.

Using the `.sync` modifier is just a convenience since it avoid writing a handler like `@updateChildAppBarTitle="appBarTitle = $event"`. This shorthand could probably used in more places, even if the communication isn't always two-way and it helps get rid of differently-named handlers.

indirectlylit: Yeah, my intuition is that the toolbar/corebase state would be better handled in a dedicated vuex module.

Within the code base, currently our use of 2-way communication is limited to using `v-model` in custom user input components, where there is a clear "value" associated with the component.

Using `.sync` like this seems like it saves some vuex boilerplate but breaks the spirit of one-directional data flow. I believe we should avoid letting it become a precedent.

If you do want to merge it in as-is, recommend:

1. add a comment indicating that it is a temporary fix, and should not be copy-pasted to other parts of the app
2. file a follow-up issue around migrating toolbar/corebase state to vuex

thanks!

indirectlylit: > This shorthand could probably used in more places, even if the communication isn't always two-way and it helps get rid of differently-named handlers.

Are you referring to `v-model` specifically? (filed #4499 - feel free to elaborate on it)

kolibri/plugins/device_management/assets/src/views/UserPermissionsPage/index.vue
```diff
@@ -191,14 +189,27 @@
...
         this.permissions.can_manage_content || this.permissions.is_superuser;
     },
+    beforeDestroy() {
+      this.updateParentAppBar();
+    },
     methods: {
       ...mapActions('userPermissions', ['addOrUpdateUserPermissions']),
       ...mapActions(['createSnackbar']),
+      updateParentAppBar(options = {}) {
+        this.$emit('update:appBarTitle', options.backPageText || '');
```
DXCanas: Mmm. Backboney.

jonboiser:
Changed things. Now there is a `coreBase` vuex module that manages all of the props for CoreBase. The Wizard and Permissions pages and their handlers interact with this module to control what text appears on the Toolbar.

indirectlylit:
> Changed things

what do you think about the new patterns?

indirectlylit:
@jonboiser if you merge in the latest `develop` it should build

kolibri/plugins/device_management/assets/src/modules/userPermissions/handlers.js
```diff
@@ -1,6 +1,12 @@
 import { DevicePermissionsResource, FacilityUserResource } from 'kolibri.resources';
-import ConditionalPromise from 'kolibri.lib.conditionalPromise';
 import samePageCheckGenerator from 'kolibri.utils.samePageCheckGenerator';
+import { createTranslator } from 'kolibri.utils.i18n';
+
+const translator = createTranslator('UserPermissionToolbarTitles', {
```
DXCanas, line 5: hrm. Are there alternatives to using the translator here?

kolibri/plugins/device_management/assets/src/modules/userPermissions/handlers.js
```diff
@@ -41,21 +47,39 @@ function fetchUserPermissions(userId) {
...
+  // Don't request any data if not an Admin
+  if (!store.getters.isSuperuser) {
+    setUserPermissionsState({ user: null, permissions: {} });
+    setAppBarTitle(translator.$tr('goBackTitle'));
+    stopLoading();
+    return Promise.resolve();
+  }
+
+  // CoreBase parameters for loading state
+  setAppBarTitle(translator.$tr('loading'));
```
DXCanas, line 63: Hmm I haven't seen this pattern anywhere. We can't display the loading page?

DXCanas (commented):
Also getting 404s on the device permission page 
image

kolibri/plugins/device_management/assets/src/modules/userPermissions/handlers.js
jonboiser: It would still look weird. The loading page only fills the area below the App Bar, so it's helpful to have "Loading" text on the immersive toolbar. Otherwise it would visually change to something like this "<- Device", which I think looks odd.

indirectlylit: When I test this it's super fast, so the loading message flickers in and out very briefly:

flicker

Are there situations where the loading actually takes long enough to warrant a feedback indicator? (Unless the user actually needs it, I usually avoid trying to show loading feedback)

indirectlylit:
looks like the 404 errors pre-dated this PR. opening a new issue: #4558

indirectlylit (approved):

kolibri/plugins/device_management/assets/src/modules/userPermissions/handlers.js
indirectlylit: created a follow-up issue for this: #4559

PR #4524, opened 2018-11-07
Optimize and externalise build tooling

### LINKED ISSUE #950: Allow plugins to be built standalone

### LINKED ISSUE #3653: Upgrade Node 6.x => 8.x (LTS)
### Observed behavior

The more recent Node 8 is an LTS and comes in the official Ubuntu 18.04 repos, just `apt install nodejs`.

> Note that the LTS lifespan for 8.x will end on December 31st, 2019.

### Expected behavior

Let's stay on par with an LTS

### User-facing consequences

None!

Original 8.0.0 changelog:

https://github.com/nodejs/node/blob/master/doc/changelogs/CHANGELOG_V8.md#2017-05-30-version-800-current-jasnell

### LINKED ISSUE #4377: Separate linting of frontend assets from loaders in the build pipeline

### LINKED ISSUE #4443: Build and distribute everything in kolibri.plugins
### Observed behavior

0.10 had a WIP of the new epub plugin distributed. It wasn't enabled, and thus not built.

However, upon downgrading from 0.11, the issue turns up: Because the plugin exists, yet the stats file isn't present because it wasn't enabled+built when we built the 0.10 distributed package.

This is unaffected by @jredrejo 's fix for avoiding to enable plugins that don't exist.. because in this case, they do, but they're broken :)

CC: @rtibbles 

### Expected behavior

1. That any plugin present in `kolibri.plugins` is built, regardless of whether it's part of the default plugins set for that version.
2. We might add a check for the stats file and skip loading the plugin if it isn't there.. while initializing and enabling plugins.

We propose to add a glob pattern in `default_plugins.txt` like `kolibri.plugins.*` to avoid maintaining this list moving forwards.

### User-facing consequences

We can have problems if we change the default set of plugins, by adding something that isn't enabled or removing a plugin from the enabled set.

### Errors and logs

See the error in #4438 

### Steps to reproduce

Install 0.11 and run it, then downgrade to 0.10 and run it.

### Context

0.11 pre-release (possibly there aren't any further issues)

rtibbles:
### Summary
* Makes optimizations to the Kolibri frontend build pipeline by putting linting in a separate process
* Consolidates all frontend linting into a single frontend command
* Creates a `kolibri-tools` npm package using yarn workspaces that exposes commands for linting, building, and testing.
* Makes the `kolibri-tools` package exportable for use by external plugins to build their frontend assets standalone, without requiring a copy of the kolibri source repo (although a pip installed version of Kolibri will still be necessary, for the WebpackFrontendAssetHook import).
* Creates a `kolibri` npm package that creates an (somewhat complete, but with extant issues for some use cases) exported copy of the Kolibri Core API currently only intended for use in tests of external plugins (as any imports of the core API will need to be properly resolved for the purposes of executing tests).
* Simplifies the frontend build logic, consolidating all base webpack configuration generation in a function in the `webpack.config.base.js` that takes a single object with data about the frontend plugin as input, and returns the appropriate webpack config as output.
* Adds an experimental `multi-processor` mode for all build commands. This can be flagged with the `--multi` flag, or by setting `KOLIBRI_BUILD_MULTI=true` as an env var. This will aggressively use all resources available on the machine to attempt to complete the build as fast as possible.
* Upgrades the required node version to 10.
* Deletes the capability to do core API injection, as it was not amenable to the way that this PR refactored the build pipeline. 
* Deletes ununsed `events` and `once` hashes defined on frontend plugins. Has not removed all the associated internal code from the mediator yet, but this should probably also be cleaned up.
* Adds the ability to specify build plugins using globs in Python paths.

### Reviewer guidance
Install Node.js 10.
Make sure that all the build and linting related commands in the package.json still work.
Test out the multi processor option.
Check the logging during builds and linting to see if it meets developer needs.

### References
Fixes #950 
Fixes #3653 
Fixes #4377
Fixes #4443 

### Contributor Checklist

### Reviewer Checklist


aronasorman:
image

ralphiee22:
Leaving this here as a todo.
Change port number between watching plugin frontend assets and kolibri frontend assets.

rtibbles:
@ralphiee22 what I'll do is add an option for 'starting port' for the webpack devservers so that plugins can set a different port by default if they want to, seems like the simplest solution. Work for you?

package.json
```diff
@@ -31,115 +39,18 @@
...
-    "webpack": "^4.6.0",
-    "webpack-bundle-size-analyzer": "^2.0.2",
-    "webpack-bundle-tracker": "https://github.com/learningequality/webpack-bundle-tracker",
-    "webpack-cli": "^2.0.15",
-    "webpack-dev-middleware": "^3.1.2",
-    "webpack-dev-server": "3.1.3",
-    "webpack-hot-middleware": "2.12.1",
-    "webpack-merge": "^4.1.0",
-    "webpack-rtl-plugin": "^1.6.0",
+    "stylelint-scss": "^3.3.2",
```
indirectlylit: why is `stylelint-scss` here and not in `packages/kolibri-tools`?

Also I see `responselike` is used in `__mocks__/http.js` - shouldn't that be in `packages/kolibri-tools` or maybe `kolibri/core`?

packages/publish.js
```diff
@@ -0,0 +1,75 @@
+const path = require('path');
+const execSync = require('child_process').execSync;
+const versionTools = require('./kolibri-tools/lib/version');
+
+/*
+ * Step 1 - Set the version of the kolibri package by the current Kolibri version
+ */
```
indirectlylit, line 7: super minor style point, but would prefer these comments be single-line so this file is easier to scan

also, a block comment at the top describing what this file does would be helpful. What are we publishing and to where?

packages/kolibri/package.json
```diff
@@ -0,0 +1,35 @@
...
+    "rest": "^1.3.2",
+    "screenfull": "^3.3.2",
+    "seededshuffle": "^0.1.1",
+    "shave": "^2.1.3",
+    "vue": "^2.5.17",
+    "vue-router": "^2.1.1",
+    "vuex": "^3.0.1"
+  },
+  "devDependencies": {
+    "kolibri-tools": "0.12.0-dev.1"
```
indirectlylit, line 30: this is fine because it's an internal dependency, but if we ever make it public I'd prefer to call it `kolibri-dev-tools`

packages/eslint-plugin-kolibri/.gitignore
```diff
@@ -0,0 +1,2 @@
+# Do not ignore this lib dir
```
indirectlylit, line 1: a 'because' clause would be helpful

indirectlylit:
Working on review now. One quick question - how (if at all) will this affect the Perseus dev workflow?

kolibri/core/package.json
```diff
@@ -0,0 +1,39 @@
+{
+  "name": "kolibri-core",
```
indirectlylit, line 2: In the notes you say

> Creates a `kolibri` npm package that creates an (somewhat complete, but with extant issues for some use cases) exported copy of the Kolibri Core API currently only intended for use in tests of external plugins (as any imports of the core API will need to be properly resolved for the purposes of executing tests).

I'm a little unclear what this is and how it works.

As an aside, wouldn't the package be `kolibri-core` rather than `kolibri`?

rtibbles:
You will now be able to build Perseus without doing a full build of Kolibri as a whole. So you can build perseus standalone.

packages/kolibri-tools/lib/cli.js
```diff
@@ -0,0 +1,297 @@
...
+        if (currentlyCompiling === 0) {
+          buildLogging.info('All builds complete!');
+          if (completionCallback) {
+            completionCallback(bundleData);
+          }
+        }
+      }
+      const children = [];
+      for (let index = 0; index < numberOfBundles; index++) {
+        if (multi) {
```
indirectlylit, line 167: This works well on my machine! Building takes noticeably shorter, and makes my computer noticeably hotter :)

If you wanted to prevent all threads from kicking off simultaneously, you could maybe keep a count `numBuildsRunning` and then do something like

```javascript
function build() { /* create child process */ }

function buildEventually() {
  if (numBuildsRunning > numCores - 1) {
    setTimeout(75, buildEventually);
  } else {
    build();
  }
}

buildEventually();
```

kolibri/core/assets/src/kolibri_module.js
```diff
@@ -25,9 +25,6 @@ export default class KolibriModule {
     // __kolibriModuleName is replaced during webpack compilation with the name derived from
     // the Python module name and the name of the class that defines the frontend kolibriModule.
     this.name = __kolibriModuleName;
-    // Insert the event hashes defined in the Python code.
-    this.events = __events;
-    this.once = __once;
```
indirectlylit, line 30: yay!

kolibri/core/assets/src/views/KGrid/extra-units.css
```diff
@@ -78,7 +77,7 @@
 }
 
 .pure-u-7-7 {
-  width: 100.0%;
+  width: 100%;
```
indirectlylit: I don't know for sure whether the extra precision was originally necessary - it's based on pure.css grids and they do some fairly tricky browser compatibility stuff. I'd prefer to ignore the file from linting and keep the decimal precision as-is to be on the safe side if possible.

indirectlylit (approved):
Looks great! I like it a lot.

Minor comments and questions, but overall nothing blocking

indirectlylit:
note on logging: when the threads are parallelized, some of the messages make less sense because they are interleaved:

```
Kolibri: Completed device_management_side_nav bundle in 4.138s!
Kolibri: Completed user_module_user_profile_nav_side_nav bundle in 4.209s!
ℹ ｢wdm｣:    12 modules
ℹ ｢wdm｣: Compiled successfully.
ℹ ｢wdm｣:    12 modules
ℹ ｢wdm｣: Compiled successfully.
Kolibri: Completed coach_side_nav bundle in 4.108s!
Kolibri: Completed user_module_login_nav_side_nav bundle in 4.084s!
ℹ ｢wdm｣:    12 modules
ℹ ｢wdm｣: Compiled successfully.
Kolibri: Completed learn_module_side_nav bundle in 3.961s!
ℹ ｢wdm｣:    12 modules
ℹ ｢wdm｣: Compiled successfully.
ℹ ｢wdm｣:    11 modules
ℹ ｢wdm｣: Compiled successfully.
Kolibri: Completed facility_management_side_nav bundle in 4.227s!
ℹ ｢wdm｣:    12 modules
ℹ ｢wdm｣: Compiled successfully.
```

Might as well just hide the ones with no context completely

packages/kolibri-tools/yarn-error.log
```diff
@@ -0,0 +1,14515 @@
+Arguments: 
+  /home/richard/.nvm/versions/node/v10.12.0/bin/node /home/richard/.yarn/bin/yarn.js add webpack-file-loader
+
+PATH: 
+  /home/richard/.nvm/versions/node/v10.12.0/bin:/home/richard/.virtualenvs/kolibripy2/bin:/home/richard/.yarn/bin:/home/richard/.config/yarn/global/node_modules/.bin:/usr/local/bin:/home/richard/.cargo/bin:/home/richard/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/games:/usr/local/games:/snap/bin:/usr/lib/jvm/java-8-oracle/bin:/usr/lib/jvm/java-8-oracle/db/bin:/usr/lib/jvm/java-8-oracle/jre/bin
```
indirectlylit: whoops - looks like this error log was committed by accident. My current PR has some yarn issues, too. I'll fix those those up once this is updated

indirectlylit: nevermind - addressed it here: https://github.com/rtibbles/kolibri/pull/16/commits/62932beb37bc2cf89ccfc8dd228817bc585d2307

packages/kolibri-tools/lib/cli.js
indirectlylit: (as we discussed, this wouldn't work: I didn't realize the processes stay running rather than restart on each rebuild)

package.json
```diff
@@ -1,25 +1,33 @@
...
+    "devserver": "npm-run-all --parallel django-devserver watch lint:watch",
+    "bundle-stats": "kolibri-tools build stats --file ./build_tools/build_plugins.txt",
+    "clean": "kolibri-tools build clean --file ./build_tools/build_plugins.txt",
+    "preinstall": "node ./packages/kolibri-tools/lib/npm_deprecation_warning.js",
+    "lint": "kolibri-tools lint",
+    "lint:fix": "yarn run lint -w",
+    "lint:watch": "yarn lint:fix -m",
+    "generate-locale-data": "node ./build_tools/i18n/intl_code_gen.js",
+    "build-kolibri": "yarn workspace kolibri run build",
+    "build-kolibri-tools": "yarn workspace kolibri-tools run build",
```
indirectlylit, line 30: I'm unclear on what `build-kolibri` and `build-kolibri-tools` are for, and how they differ from `build` above.

`build-kolibri` fails with 

```
(k012x)d@mac ~/P/l/r/k012x (external_build)> yarn run build-kolibri
yarn run v1.12.3
$ yarn workspace kolibri run build
warning Missing version in workspace at "/Users/d/Projects/learning_equality/repos/k012x/kolibri/plugins/demo_server", ignoring.
$ node ./build/index.js
internal/modules/cjs/loader.js:582
    throw err;
    ^

Error: Cannot find module '/Users/d/Projects/learning_equality/repos/k012x/packages/kolibri/build/index.js'
    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:580:15)
    at Function.Module._load (internal/modules/cjs/loader.js:506:25)
    at Function.Module.runMain (internal/modules/cjs/loader.js:741:12)
    at startup (internal/bootstrap/node.js:285:19)
    at bootstrapNodeJSCore (internal/bootstrap/node.js:739:3)
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
error Command failed.
Exit code: 1
Command: /Users/d/.nvm/versions/node/v10.14.1/bin/node
Arguments: /usr/local/Cellar/yarn/1.12.3/libexec/lib/cli.js run build
Directory: /Users/d/Projects/learning_equality/repos/k012x/packages/kolibri
Output:

info Visit https://yarnpkg.com/en/docs/cli/workspace for documentation about this command.
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

package.json
```diff
@@ -1,25 +1,33 @@
...
-    "test-jest": "jest --config jest_config/jest.conf.js",
-    "test-jest-cov": "jest --config jest_config/jest.conf.js --coverage",
+    "test-jest": "kolibri-tools test --extraConfig ./jest.conf.js",
+    "test-jest-cov": "yarn run test-jest --coverage",
     "coverage": "yarn run test-jest-cov",
-    "build": "webpack --config ./frontend_build/src/webpack.config.prod.js",
-    "test-build": "yarn run lint-js && webpack --config ./frontend_build/src/webpack.config.prod.js --bail",
-    "makemessages": "webpack --config ./frontend_build/src/webpack.config.trs.js",
-    "watch": "node ./frontend_build/src/webpackdevserver.js",
+    "build": "kolibri-tools build prod --file ./build_tools/build_plugins.txt",
```
indirectlylit, line 15: The build command currently clears the console after each module is built. It would be better if it didn't, because helpful information such as bundle sizes gets hidden.

(Note - I'm using `multi` mode)

packages/kolibri-tools/lib/cli.js
```diff
@@ -0,0 +1,297 @@
...
+    list,
+    []
+  )
+  .option(
+    '--pluginPaths <pluginPaths...>',
+    'An explicit comma separated list of explicit file paths to plugins',
+    list,
+    []
+  )
+  .option('-m, --multi', 'Run using multiple cores to improve build speed', false)
```
indirectlylit: Might want to make the default `true`?

indirectlylit (commented):
a few more comments based on recent HMR work

.eslintrc.js
```diff
@@ -1,124 +1 @@
...
-    'import/no-duplicates': ERROR,
-    'import/newline-after-import': ERROR,
-    'import/order': ERROR,
-
-    // Custom vue rules
-    'kolibri/vue-filename-and-component-name-match': ERROR,
-    'kolibri/vue-component-registration-casing': ERROR,
-  },
-};
+module.exports = require('kolibri-tools/.eslintrc');
```
indirectlylit, line 1: As far as I can tell, vue files are no longer being linted after this change. This is true both of `<script>` and `<template>` sections.

To test this, try adding `props: ['simpleprop']` to a component before and after this PR. Before, it will complain

```
eslint:vue/require-prop-types Props should at least define their types.
```

but after it will not.

Currently I believe this is a config issue and not related to `lint.js` because the behavior happens whether I run the `devserver` or try to set up linting in sublimetext.

Investigating further...

indirectlylit (dismissed):
noticed a regression

.eslintrc.js
indirectlylit: Confirmed that moving `.eslintrc.js` back to the root directory and updating the `import/resolver` allows `eslint` to work correctly in Sublime.

However, even with this change, `yarn run lint` behaves strangely:

* `<template>` errors are thrown without context, e.g. `Kolibri linter: Invalid array length` (which I believe comes from `HTMLHint`, not `eslint`)
* Vue style rules like `require-prop-types` are not reported
* Syntax errors in the `<script>` section are reported, but again lack context (filename)

packages/kolibri-tools/lib/cli.js
```diff
@@ -0,0 +1,297 @@
...
+    if (!bundleData.length) {
+      cliLogging.log('No valid bundle data was returned from the plugins specified');
+      process.exit(1);
+    }
+    const buildModule = {
+      [modes.PROD]: 'production.js',
+      [modes.DEV]: 'webpackdevserver.js',
+      [modes.I18N]: 'i18n.js',
+      [modes.STATS]: 'bundleStats.js',
+    }[mode];
```
indirectlylit, line 143: This currently returns `undefined` when running `make clean`, which causes the line below to error out

.eslintrc.js
rtibbles: I hate magic import resolvers. Will play around with this and try to add some testing.

Will look at the context issues again, I was trying to use the built in reporters for each linter to maintain context as much as possible, but it's possible that invoking them with the file contents and not the file itself may be causing issues. Thanks for checking this out!

kolibri/core/package.json
rtibbles: No, the `kolibri` package is constructed with reference to this `kolibri-core` package, but is a separate entity. The main purpose of calling the exported package `kolibri` is for consistency of naming between our internal references to the kolibri core API, and people using it externally.

kolibri/core/assets/src/views/KGrid/extra-units.css
rtibbles: Righto, can also add linting exclude comments.

package.json
rtibbles: These are the build commands for the two exportable/publishable npm modules.

rtibbles: I think I must have committed this code part way through a refactor of how I was doing the building...

rtibbles: I have building working again, but the `kolibri` package is still very much WIP.

package.json
rtibbles: This happens without multi also. Will have to look and see how to stop webpack from doing this.

packages/kolibri-tools/lib/cli.js
rtibbles: Have brought clean in line with the other modes.

.eslintrc.js
rtibbles: I have tested this out, and this all seems to be working properly now, and eslint is working correctly in Sublime for me also. Let me know if this is still causing issues as the PR is now.

indirectlylit: working!

indirectlylit (approved):
great job

PR #4526, opened 2018-11-08
Load plugins from build_plugins.txt when env var BUILD_TIME_PLUGINS is empty

lyw07:
### Summary

This PR is to fix the issue that the webpack build files for the plugin `demo_server` are missing in the latest 0.11 release. 
This is because in [env.list](https://github.com/learningequality/kolibri/blob/v0.11.0/docker/env.list#L1), the environment variable `BUILD_TIME_PLUGINS` is set to be empty string, so the code
```
os.environ.setdefault(
    "BUILD_TIME_PLUGINS", os.path.realpath(os.path.join(os.path.dirname(__file__), "build_plugins.txt"))
)
```
does not change the value to `build_tools/build_plugins.txt`.
This leads to the value of variable `build_list` to be `[]`

### Reviewer guidance

to test the change, we can either extract the whl file and check if the build files exist in `kolibri/plugins/demo_server` or enable the plugin first and then run kolibri to see if kolibri server starts successfully.

### References

…

----

### Contributor Checklist

### Reviewer Checklist

### DIFF
--- a/build_tools/webpack_json.py
+++ b/build_tools/webpack_json.py
@@ -12,9 +12,6 @@ from kolibri.plugins.registry import initialize
 logging.basicConfig(format='%(levelname)s:%(message)s', level=logging.DEBUG)
 logging.StreamHandler(sys.stdout)
 
-os.environ.setdefault(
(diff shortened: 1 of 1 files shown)


rtibbles (approved):
The right fix, and it gives the right result - the build stats and static files are now present for the demo server plugin!

PR #4529, opened 2018-11-08
0.11 into develop

rtibbles:
### Summary
Cascade merge. Only conflict was the version in `kolibri/__init__.py`.

----


christianmemije (approved):
Did a quick sanity test, and looks good to me!

Issue #4422, opened 2018-10-14
Session and Summary Log CSV files take a long time to generate


rtibbles:
### Observed behavior
When there is a reasonable amount of data, the CSV summaries both take a long time to generate.
This can lead a user to think either nothing is happening, or if there is a strict timeout enforced by, e.g. nginx, then it can prevent access to them.

### Expected behavior
Generated the logs asynchronously and let the user know when they are ready to be downloaded.

### User-facing consequences
Inability to download important data.

### Steps to reproduce
On a server with a at least the fake data generated on it, attempt to download either of the CSV files from the Facility Data tab.

### Context
* v0.11.0-beta1

khangmach:
![csv generation](https://user-images.githubusercontent.com/25045166/47186876-d7d26100-d2e6-11e8-8e4e-c1cc944db6f2.png)

@rtibbles  Here's something we could do:

**Left:** 
once the user hits the download button, show a snackbar notification that lets the user know the preparation to generate the file is taking place. The snackbar persists on the screen throughout the application and is cancelable. Cancelling the action stops file generation. 

**Right:** 
once the preparation is done dismiss the previous snackbar and show a new one saying the file is now downloadable. This one is also dismissable. ALTERNATIVELY, and also which may be preferable and quicker, is to automatically start the download after the file is fully prepared (aka after snackbar one dismisses)

**Some notes:**
- The transition of switching snackbars works pretty much exactly like the one we show when Kolibri is not connected to a network and when the network is re-established.
- Inspiration is heavily drawn from Google Drive's download feature (try downloading a folder). Didn't want to invent something new when there's already something that works well out there!

rtibbles:
Yeah, drawing on existing designs seems like a good idea here. What happens if you navigate away from Google Drive and back again while this process is happening, do you have to restart the process?

rtibbles:
(more generally, how transient do we want this to be?)

khangmach:
<img width="507" alt="screen shot 2018-10-23 at 11 13 48 am" src="https://user-images.githubusercontent.com/25045166/47381410-cc02e800-d6b4-11e8-875b-0f03197db827.png">
Currently when you navigate away from gdrive through another URL or closing the tab you get this message and if you navigate away, you lose the process. When you browser back to the original gdrive page, the process doesn't restart. 

I think it makes sense for our log file process to continue if you navigate to a different kolibri page, but maybe not if you exit that specific kolibri browser tab.

[...]

PR #4531, opened 2018-11-13
Async log export

jredrejo:
### Summary
In #4422 a problem when downloading Session and Summary logs in big databases.
This PR adds several changes:
1. After doing a profiling analysis of the exporting code, it was clear there was a big performance problem in the DRF serializer. This commit solves  it https://github.com/jredrejo/kolibri/commit/1c131bb4e897e005c07fd1b2998cd11112d9d315 
Tested in an I7, 16 Gb RAM laptop shows needed time passes from more than 70 seconds to less than 11 seconds. Done with a database having 13432 to be exported. These are the profiling charts:
Develop:
develop_contentsessionlogs
Branch with commit fixing the performance problem:
branch_contensessionlogs

2. Creates a Django command to execute the creation of the csv file asynchronously,
`kolibri manage exportlogs` (by default dumps `session`on a temp file)
`kolibri manage exportlogs --output-file=new_file.csv --log-type=summary`  are all the possible arguments.

3. Creates a new api entry to add a task that will executes the previous command
/api/tasks/tasks/startexportlogcsv/
It's a POST request that accepts `logtype` and `filepath`as parameters and returns the info of the job task where filepath is also included.

### Reviewer guidance
1. To download the generate csv file I thought of using the content/DownloadContentView class, but a) it only allows GET and b) it might be more logical creating an specific endpoint for these csv logs. Opinions?
2. To test the code, you can download the logs in a big database using the UI. Then, it can be tested either using the command line with  `kolibri manage exportlogs` or using the web UI

Frontend
Before clicking on the button to generate a log (having one previously generated):
before

After summary logs have been generated and while session logs are being generated:
after

### References
Solves #4422 

### Contributor Checklist

### Reviewer Checklist


ralphiee22:
Have you thought about caching `get_facility_name` as well under `LogCSVSerializerBase`? Since this is doing another 2 reads to get the user, and the facility name for each log object (even though they may be the same across many logs). 
On an unrelated note, I find it weird that in the function we only put the facility name if the object has a user, when each log has to be associated with a facility. We should actually be doing `Facility.objects.get(dataset_id=obj.dataset_id).name` cc: @jamalex

ralphiee22:
Ahh I just seen the `get_queryset` method where you prefetch the user and facility models. Ignore my previous comment.

jonboiser:
1. Downloading from the data page still works as usual (but with very few logs on my instance, so handling of very large logs wasn't tested)
1. Downloading with the management command (on macOS) or POSTing to `startexportlogcsv` places the CSV in a seemingly randomly-generated path like `/var/folders/21/nlh_hhnj1p3g92bgnbl7p42c0000gn/T/tmpN_9dGC.csv`. Whereas when you download on the browser it goes to `~/Downloads` with a consistent filename like `content_summary_logs.csv`. It's also not clear which log file I downloaded (content session or summary?). Would it be possible or desirable to make the management command download work the same as on the UI (i.e. be more opinionated on the default output file location)?

jredrejo:
@jonboiser great you tested it on macOS, about your doubts:
1. Frontend is not changed yet, I am working on it. The download should be much faster than previously on large logs anyway . I am making an async frontend as suggested in #4422 

2. I have always been reluctant to create predictable filenames in a web application and the command is going to be used from the web UI. However, it's done to be usable also from the cli, so it make sense to use the same filenames. About the location, there might be problems: my first thought is leaving it in the same directory the command is executed, but it could have permissions problems. WHen using temp filenames that problem didn't exist. Unless a better idea is suggested I will leave the file in the folder and show an error if the user has not write permissions on it or if the file already exists.

jonboiser:
>  I will leave the file in the folder and show an error if the user has not write permissions on it or if the file already exists.

That sounds like a behavior most users will expect (even though I saw the log that the file was written to `/var` I thought it would have been copied automatically to my `pwd`). If the file already exists, though, Chrome and other browsers will append something like `(2)` to the end to make it unique.

kolibri/core/tasks/api.py
```diff
@@ -334,8 +335,49 @@ def localdrive(self, request):
...
+        job_type = "EXPORTSUMMARYLOGCSV" if log_type == "summary" else "EXPORTSESSIONLOGCSV"
+
+        job_metadata = {
+            "type": job_type,
+            "started_by": request.user.pk,
+        }
+
+        job_id = get_client().schedule(
+            call_command,
+            "export_log_csv",
```
indirectlylit: We tend to follow the (rather unfortunate) django convention of not using underscores to separate words in command names.

Also, recommend not mentioning "csv" in the command name and API endpoints: In the future, we might want to allow alternate export representations, like JSON.

So this command could just be `exportlogs`

kolibri/core/logger/csv.py
```diff
@@ -13,6 +17,60 @@
...
+    if hasattr(request, 'session'):
+        filepath = request.session.get('csv_file_{}'.format(log_type), None)
+    else:
+        filepath = None
+    # if the file does not exist on disk, return a 404
+    if filepath is None or not os.path.exists(filepath):
+        raise Http404('There is no csv export file for {} available'.format(log_type))
+
+    # generate a file response
+    response = FileResponse(open(filepath, 'rb'))
```
indirectlylit, line 87: For Py2/Py3 compatibility I think this needs to be:

```python
io.open(filepath, mode='r', encoding='utf-8')
```

kolibri/core/logger/management/commands/export_log_csv.py
```diff
@@ -0,0 +1,101 @@
...
+        with self.start_progress(total=self.total_rows) as progress_update:
+            for item in serializer.data:
+                progress_update(1)
+                self.data.append(item)
+
+        return True
+
+    def _create_file(self, buffer):
+        try:
+            with open(self.filepath, 'w') as f:
```
indirectlylit: This throws an error in Py3 with unicode:

```
WARNING:iceqube.worker.backends.base:Job 4d281c4cd08049f695b2baa8936178b6 raised an exception: Traceback (most recent call last):
  File "/Users/d/Projects/learning_equality/repos/k012x/.venv/lib/python3.6/site-packages/iceqube/worker/backends/inmem.py", line 75, in handle_finished_future
    result = future.result()
  File "/usr/local/Cellar/python/3.6.5/Frameworks/Python.framework/Versions/3.6/lib/python3.6/concurrent/futures/_base.py", line 425, in result
    return self.__get_result()
  File "/usr/local/Cellar/python/3.6.5/Frameworks/Python.framework/Versions/3.6/lib/python3.6/concurrent/futures/_base.py", line 384, in __get_result
    raise self._exception
  File "/usr/local/Cellar/python/3.6.5/Frameworks/Python.framework/Versions/3.6/lib/python3.6/concurrent/futures/thread.py", line 56, in run
    result = self.fn(*self.args, **self.kwargs)
  File "/Users/d/Projects/learning_equality/repos/k012x/.venv/lib/python3.6/site-packages/iceqube/worker/backends/inmem.py", line 149, in wrap
    raise e
  File "/Users/d/Projects/learning_equality/repos/k012x/.venv/lib/python3.6/site-packages/iceqube/worker/backends/inmem.py", line 145, in wrap
    return f(*args, **kwargs)
  File "/Users/d/Projects/learning_equality/repos/k012x/.venv/lib/python3.6/site-packages/iceqube/common/classes.py", line 120, in y
    return func(*args, **kwargs)
  File "/Users/d/Projects/learning_equality/repos/k012x/.venv/lib/python3.6/site-packages/django/core/management/__init__.py", line 131, in call_command
    return command.execute(*args, **defaults)
  File "/Users/d/Projects/learning_equality/repos/k012x/.venv/lib/python3.6/site-packages/django/core/management/base.py", line 330, in execute
  File "/Users/d/Projects/learning_equality/repos/k012x/kolibri/core/tasks/management/commands/base.py", line 99, in handle
    return self.handle_async(*args, **options)
  File "/Users/d/Projects/learning_equality/repos/k012x/kolibri/core/logger/management/commands/export_log_csv.py", line 101, in handle_async
    self._create_file(buffer)
  File "/Users/d/Projects/learning_equality/repos/k012x/kolibri/core/logger/management/commands/export_log_csv.py", line 70, in _create_file
    f.write(buffer)
TypeError: write() argument must be str, not bytes
```

Looks DRF returns binary strings so it needs to be:

```python
with open(self.filepath, 'wb') as f:
```

indirectlylit (dismissed):
This is looking great, and I think we're 90% of the way there. However there are a few issues:

----

When I click 'download', python starts eating all my CPU, and the download never starts:

downloading

Note that this is on Python 3, and occurred after making the unicode-related changes I mentioned inline.

There are a few discrepancies with the [UI details as specified](https://docs.google.com/document/d/1pgjYFbONTNgm2taUu8gSDPMMJ61MtWlvInJtYR0cdkg/edit#heading=h.1lusfn3rajl5):

1. When a log file has already been generated, it should say "Generated `X` minutes ago" and the link should be "Regenerate log file", not "Generate log file"
2. The loader should just have a 'spinner' icon. No linear loading visualization and no percentage
3. The UI doesn't poll until the user triggers an export. This means that if two users are looking the page or two tabs are open, they won't stay in sync
4. We are relying on client state to store the name of the file to be downloaded (e.g. `filepath: "/var/folders/n0/_qbhl2qn1ng8hjsqgdmwdq1h0000gp/T/tmpznvsz6lv.csv"`). Since this is stored in client state, so if you refresh the page or log out and back in again, you lose the reference to the file. It's also a security concern since we're leaking absolute paths to the client, and maybe even allowing arbitrary files to be downloaded?

I think a better strategy would be to make generate a new folder inside the kolibri home directory. There you can save your files with known names (e.g. `~/.kolibri/log_export/summary.csv`). You can simultaneously generate a file that has a timestamp for when that was exported  (e.g. `~/.kolibri/log_export/summary_time.txt`), and this will allow you to display the "Generated `X` minutes ago" text

indirectlylit:
> To download the generate csv file I thought of using the content/DownloadContentView class, but a) it only allows GET and b) it might be more logical creating an specific endpoint for these csv logs. Opinions?

I agree (b) is the better option. `DownloadContentView` seems specific to content, and would make it more difficult to do some of the things described above

kolibri/core/logger/management/commands/export_log_csv.py
jredrejo: Thanks, fixed

kolibri/core/logger/csv.py
jredrejo: Thanks, fixed

kolibri/core/tasks/api.py
jredrejo: Done

jredrejo:
@indirectlylit 

> Note that this is on Python 3, and occurred after making the unicode-related changes I mentioned inline.

Yes, that was the issue, it has been fixed after changing the way open is used.

jredrejo:
@indirectlylit 

> The loader should just have a 'spinner' icon. No linear loading visualization and no percentage

can I know the rationale of this decission?. From an user point of view I would like to have a percentage or  a progress bar to have an idea of how long the process will take. If that's not needed I don't see the reason of having made an async task for generating the files.
I don't know how rock graved are the UI specs, but in this case I must confess I think there are a few points that could be improved. I thought the mockup was provisional: the percentage, some of the font sizes, or even the position of the Download button before the generate log text are things I think could be re-evaluated.

jredrejo:
@indirectlylit 

> We are relying on client state to store the name of the file to be downloaded (e.g. filepath: "/var/folders/n0/_qbhl2qn1ng8hjsqgdmwdq1h0000gp/T/tmpznvsz6lv.csv"). Since this is stored in client state, so if you refresh the page or log out and back in again, you lose the reference to the file. It's also a security concern since we're leaking absolute paths to the client, and maybe even allowing arbitrary files to be downloaded?

This is the part where I have more doubts. I am aware of the flaws of the implementation, mainly the fact of losing the information when reloading the page. However, in the original proposal I see two problems:

1. I think it is more insecure writing files in the disk with a known name, that's why temp filenames are more secure. With the current implementation, in the worst scenario a useless file is left, but no way for an attacker to guess in advnace the name those files will have. I agree ~/.kolibri/log_export is a better place to store them, instead of the dangerous tmp folder, but I don't like having a predictable filename there.
2. WIth the current implementation, two different users from two different sessions can generate , if they have permissions, their log files at the same time. With the proposed solution, if one user generates a dump file, it will be overwritten later by a second user, not knowing why he can't download a file that has already been generated. I think this can be a confussing scenario for the end users.

Said that, the changes to do the proposal implementation are easy and should be ready in one day. Having git we can rollback if needed.

indirectlylit:
regarding mockups:

The mockups were not intended to specify font sizes, layout, etc - that's all stuff that we can easily tweak. However they were intended to specify functional behaviors, in terms of what the content displayed in various states. It would be best to iterate on functional design before implementing, or to raise the question during implementation. I have thoughts on showing percentages, etc, but I think it would be best to take this to a voice conversation and loop in Khang.

----

regarding files and temp files:

We may be talking about different attack vectors?

Tempfiles are useful for isolating different processes running under the same user from each other. However if an attacker has access to the `KOLIBRI_HOME` directory, it means they already have user-level access to the operating system and therefore they can already access all the data in the unencrypted sqlite DB.

I'm much more concerned with leaking internal information from the OS to web clients. In the current implementation, I'm fairly sure that as an external attacker, I could craft a malicious request that downloads arbitrary files from the OS because of

```python
def download_csv_file(request, log_type):
    if hasattr(request, 'session'):
        filepath = request.session.get('csv_file_{}'.format(log_type), None)
```

I can send `csv_file_summary` = `/Users/jredrejo/.ssh/id_rsa` 😈

indirectlylit:
> With the proposed solution, if one user generates a dump file, it will be overwritten later by a second user, not knowing why he can't download a file that has already been generated

I think it should be fairly clear that the file can't be downloaded because a new one is being generated. This is the same behavior we use in import/export of content, where all users have a shared view of server state. In general, it will be highly unlikely for multiple admins to be simultaneously trying to re-export logs. These changes are more to make sure that refreshing the page and opening multiple tabs for a single user works as expected.

jredrejo:
@indirectlylit I have done all the changes you mentioned to fulfill the specs.

indirectlylit (dismissed):
Hey @jredrejo this is looking very good. I made a PR to clean up the layout and text a bit: https://github.com/jredrejo/kolibri/pull/2/ - let me know what you think

I see two final issues with the PR:

1. It looks like the tasks are never "cleaned up". Old "completed" tasks continue to be accumulated and sent:

image

The desired behavior here is a bit different from content import/export, where we have the user manually acknowledge and dismiss the most recent import.

2: Log import tasks are showing up in the content import/export view:

image

To reproduce this:

* open CSV export in one tab and the content management page in another
* generate a log file

We'll need to add some logic to the import/export task viewer to filter by task type.

jredrejo:
@indirectlylit 

> 1: It looks like the tasks are never "cleaned up". Old "completed" tasks continue to be accumulated and sent:

Fixed

> 2: Log import tasks are showing up in the content import/export view:

Fixed (had to add a filter in the device_management plugin because it was considering all the existing tasks were related to it.

indirectlylit:
Looking great!

I'm still seeing a little bit of unexpected interaction between the content and data pages though. My guess is that it's caused by issue #4604 that you reported?

see below:

export

jredrejo:
@indirectlylit about 

> I'm still seeing a little bit of unexpected interaction between the content and data pages though. My guess is that it's caused by issue #4604 that you reported?

Yes, that's the problem: as canceling a single task has been nulled in #4604 the only option I had was deleting all completed tasks, what causes that glitch
I see three possible options:
1. Lazy and quick option: this is not very important so we can leave it as it is.
2. Conservative and quick option: I can add a `cancelSingleTask` function to do what `cancelTask` should do, so this problem is solved. 
3. Perfect but maybe long term solution: Fixing #4604 either making deviceManagement call `clearTasks` or fixing the issue that hack was solving so it can use `cancelTask` working as it should.

indirectlylit (approved):
I've updated your existing issue at #4604 to reflect the user-facing consequences above and given it a P2 priority in 0.12.

Time to get this merged!

PR #4520, opened 2018-11-06
Update front end docs (WIP)

christianmemije:
### Summary

Update front end docs (WIP)

### References

…

----

### Contributor Checklist

- [ ] Contributor has fully tested the PR manually
- [ ] PR has the correct target branch and milestone
- [ ] PR has 'needs review' or 'work-in-progress' label
- [ ] If PR is ready for review, a reviewer has been added. (Don't use 'Assignees')
- [ ] If there are any front-end changes, before/after screenshots are included
- [ ] If this is an important user-facing change, PR or related issue has a 'changelog' label

### Reviewer Checklist

- [ ] Automated test coverage is satisfactory
- [ ] Reviewer has fully tested the PR manually
- [ ] PR has been tested for [accessibility regressions](http://kolibri-dev.readthedocs.io/en/develop/manual_testing.html#accessibility-a11y-testing)
- [ ] External dependencies files were updated (`yarn` and `pip`)
- [ ] Documentation is updated
- [ ] Link to diff of internal dependency change is included
- [ ] CHANGELOG.rst is updated for high-level changes
- [ ] Contributor is in AUTHORS.rst

PR #4456, opened 2018-10-18
adds docs about our github label conventions

indirectlylit:
### Summary

Adds a summary of our github label conventions. I'll also add descriptions to the labels in github

### Reviewer guidance

Make sure both the docs and the descriptions make sense

### References

fixes #3835

----

### Contributor Checklist

- [x] Contributor has fully tested the PR manually
- [x] PR has the correct target branch and milestone
- [x] PR has 'needs review' or 'work-in-progress' label
- [x] If PR is ready for review, a reviewer has been added. (Don't use 'Assignees')
- [ ] If there are any front-end changes, before/after screenshots are included
- [ ] If this is an important user-facing change, PR or related issue has a 'changelog' label

### Reviewer Checklist

- [ ] Automated test coverage is satisfactory
- [ ] Reviewer has fully tested the PR manually
- [ ] PR has been tested for [accessibility regressions](http://kolibri-dev.readthedocs.io/en/develop/manual_testing.html#accessibility-a11y-testing)
- [ ] External dependencies files were updated (`yarn` and `pip`)
- [ ] Documentation is updated
- [ ] Link to diff of internal dependency change is included
- [ ] CHANGELOG.rst is updated for high-level changes
- [ ] Contributor is in AUTHORS.rst

PR #4536, opened 2018-11-20
flatten developer docs

indirectlylit:
### Summary

Removes a layer of nesting on our dev docs, so that more pages are visible at a glance:

| before | after |
|--|--|
| image | image |

Specifically, flattened the 'start here', 'architecture', and 

### Reviewer guidance

make sure this makes sense and still works

### References

follow-up from #4520

----

### Contributor Checklist

### Reviewer Checklist


benjaoming:
So, maybe the flat structure is a bit over the top in terms of the number of sections?

Proposing:

The actual problem is that the **Architecture** section contained **Frontend architecture**, and that section is full of things that didn't specifically pertain to "Architecture". So maybe it would be better to have a top-level section simply called **Frontend** and keep the other sections put?

That way, frontend developers save a click and will see their documentation elevated to the same level as other developer topics (which is totally warranted).

While we're at it:

I don't understand the "Builds" section and would like to nest that in Architecture, since it's basically discussing the Build architecture.

benjaoming:
It was a nice point in the [What Nobody Tells You About Documentation](https://www.youtube.com/watch?v=t4vKPhjcMZg) talk that we should try to separate the background documentation from the basic practical instructions/howtos.

indirectlylit:
> So, maybe the flat structure is a bit over the top in terms of the number of sections?

I disagree - for me personally it's a big improvement to be able to visually scan them rather than manually click open little toggles to find a section. Unless you think it's wrong or actually harder to use, I'd prefer to make this change or maybe we should get a second opinion

> While we're at it

I'd really prefer to just get this merged in and do follow up work in another PR. e.g. #4456 is another one that I haven't had time to come back to and so it's de facto blocked

benjaoming:
Agreed that the current nesting of sections in Architecture doesn't work. There is also a confusing overlap between Architecture and Reference. Some sections in Architecture appear more as References. AND there's an overlap or mis-placements of getting started guides for i18n and QA inside References. In short: Nothing really seems to have a convincing, natural place in the docs at present.

> do follow up work in another PR

Opened a new issue: https://github.com/learningequality/kolibri/issues/4543

I suggest to put "Builds" in "Architecture", so that's a contradiction to flattening the structure :) I generally don't think it's a good idea to flatten the structure when the problem is that the existing structure isn't being followed and doesn't work well.

The structure is supposed to serve the purpose of finding information through navigation. Eliminating the structure will give us a short-term gain and after that a long-term problem of not having a structure. It doesn't scale very well, so to speak :)

I think we could benefit from selecting better first level topics and then re-organizing under those? Well, if not in this PR, then definitely soon. But for the flattening, do feel free. If it makes life better for someone, that's :+1:

indirectlylit:
>  There is also a confusing overlap between Architecture and Reference. Some sections in Architecture appear more as References. AND there's an overlap or mis-placements of getting started guides for i18n and QA inside References. In short: Nothing really seems to have a convincing, natural place in the docs at present.

Yes, exactly :)

Just a moment ago for example, I went to look for documentation on Webpack. That could reasonably live under any of Architecture, Reference, or Build and I actually clicked all three of them before finding it

> The structure is supposed to serve the purpose of finding information through navigation. Eliminating the structure will give us a short-term gain and after that a long-term problem of not having a structure. It doesn't scale very well, so to speak 

Yes, very true.

However adding structure before we need it seems like adding abstractions in code too early: we have no way of knowing what the right "classes" are yet, and well-meaning but misguided speculation is how we ended up with the Architecture/References mess. Granted, flat doesn't scale – but we can cross that bridge when we need to.

To me as a user it's always felt frustrating to be forced to find information through navigation when there are only a dozen pages.

> Opened a new issue: #4543

thank you!

> But for the flattening, do feel free. If it makes life better for someone, that's 👍

thank you!

benjaoming (approved):
:+1: 

It's great if this makes life easier. Let's try to find time to re-organize the sections and split them and add placeholders etc. for a better future layout.

indirectlylit:
> I don't understand the "Builds" section and would like to nest that in Architecture, since it's basically discussing the Build architecture.

updated to address this note

benjaoming:
Moving Builds to Architecture is nice, thanks!

PR #4540, opened 2018-11-24
Hashi implementation for localStorage, sessionStorage, and cookies

rtibbles:
### Summary
Implements a 'no dependency' version of Hashi building on original work by @kollivier in #4501 (and, as such, supersedes it).

Sets up our HTML5 zip file endpoint on Kolibri to serve up a special template that loads Hashi and then initializes it.

The inner Hashi mocks the localStorage, sessionStorage, and document.cookie interfaces inside the iframe, allowing HTML5 apps contained therein to access them as if they were not sandboxed, but still safely.

Once this has been setup, it sends a ready event to any external Hashi that may be listening that it is ready. Once it receives a return ready event, it loads up the actual HTML for the page and writes that into the document.

This inner Hashi then communicates with a hashi object external to the iframe that is setup by the HTML5AppRenderer, that then communicates changes in persistent state to be saved into the extraFields object on the ContentSummaryLog.

### Reviewer guidance
Read the tests.
For a full manual integration test, run the `make_test_zip.py` file in the scripts folder to generate a zip file that will utilize the three shimmed features. If it doesn't error, and on reload you see increasing counts of loading and last loaded dates, it is working!

I also deleted the 15 second 'completion' for HTML5 apps, because it is silly, but not sure what to replace it with.

### References
Supersedes #4501

----

### Contributor Checklist

### Reviewer Checklist


packages/hashi/src/cookie.js
```diff
@@ -0,0 +1,222 @@
...
+    });
+    return this.__removeExpiredCookies(data);
+  }
+
+  __now() {
+    return new Date(Date.now() + this.__nowDiff);
+  }
+
+  __setNowDiff(nowDiff) {
+    this.__nowDiff = nowDiff;
```
kollivier, line 121: Could you add a comment somewhere in this class about the purpose of __nowDiff? Is this an time sync issue with the bridge and using async events?

packages/hashi/src/mainClient.js
```diff
@@ -0,0 +1,69 @@
+import Mediator from './mediator';
+import LocalStorage from './localStorage';
+import Cookie from './cookie';
+import { events, nameSpace } from './hashiBase';
+
+export default class Hashi {
```
kollivier: I'm already finding inner / outer Hashi is hard to keep straight during the code review. How about names like SandboxHost (for the iFrame side) and SandboxBridge/Client (for the app side)?

packages/hashi/src/mainClient.js
```diff
@@ -0,0 +1,69 @@
...
+  constructor({ iframe, now } = {}) {
+    this.events = events;
+    this.iframe = iframe;
+    this.mediator = new Mediator(this.iframe.contentWindow);
+    this.storage = {
+      localStorage: new LocalStorage(this.mediator),
+      cookie: new Cookie(this.mediator),
```
kollivier, line 21: I saw earlier that we initialize sessionStorage, but it seems there's no key for this here. Is that intentional?

packages/hashi/test/iframeClient.spec.js
```diff
@@ -0,0 +1,86 @@
...
+    mockXHRClass.prototype.addEventListener = jest.fn();
+    mockXHRClass.prototype.setRequestHeader = jest.fn();
+    window.XMLHttpRequest = mockXHRClass;
+    xhrOpen = window.XMLHttpRequest.prototype.open;
+    hashi = new Hashi();
+  });
+  describe('constructor method', () => {
+    it('should bind a listener to a readycheck event to send a ready event', () => {
+      hashi.mediator.sendMessage = jest.fn();
+      return new Promise(resolve => {
```
kollivier, line 15: Just wondering, do we need to actually return the promise in this test rather than, say, assign it to a variable? I know the assertions is checked in the then block and it's all async, just not used to seeing a test return like this.

packages/hashi/src/mainClient.js
kollivier: Disregard, upon further review, I realized that we're doing this in the inner Hashi only since we don't need to persist sessionStorage.

packages/hashi/test/mediator.spec.js
```diff
@@ -0,0 +1,266 @@
...
+      mediator.__messageHandlers[nameSpace] = testData;
+      mediator.removeMessageHandler({ event, nameSpace });
+      expect(mediator.__messageHandlers).toEqual({
+        [nameSpace]: {
+          [event]: [],
+        },
+      });
+    });
+  });
+});
```
kollivier, line 266: So test! Much coverage! More than reviewing your code, this was a Jest tutorial for me! :)

packages/hashi/src/mediator.js
```diff
@@ -0,0 +1,88 @@
+/*
+ * This class manages all message listening and sending from the postMessage
+ * layer. All interfaces that need to message via the postMessage layer should
+ * do so via an instance of this class.
+ */
+
+class Mediator {
```
kollivier, line 7: Mediators++! :)

packages/hashi/src/mediator.js
```diff
@@ -0,0 +1,88 @@
...
+  }
+
+  handleMessage(ev) {
+    const { nameSpace, event, data } = ev.data;
+    // nameSpace and event should be defined, otherwise, it's not our message!
+    if (typeof nameSpace === 'undefined' || typeof event === 'undefined') {
+      return;
+    }
+    if (this.__messageHandlers[nameSpace] && this.__messageHandlers[nameSpace][event]) {
+      this.__messageHandlers[nameSpace][event].forEach(callback => {
```
kollivier, line 22: Message stacks, nice!

kollivier (changes requested):
Looks awesome, thanks! Crazy to see how evolved this is over what I had. :) 

I think the only thing I feel is a blocker is having both sides of the bridge being named Hashi. While it's not a technical problem, I don't want to punt this and have it fall off the radar.

packages/hashi/test/iframeClient.spec.js
rtibbles: This is necessary for async testing in Jest (and mocha) - returning the Promise lets the test suite know that the test is unfinished as long as the Promise has not resolved, so essentially, when you return a promise from the test, it adds a `finally` invocation to the end of the promise chain that you return, and only assesses whether the test has failed after that has been reached.

packages/hashi/src/cookie.js
rtibbles: Can add a comment, for sure.

This is because generally we have found that client side clocks can be really unreliable. Server clocks can also be unreliable, but this allows us to set all times relative to the Kolibri server, and not at the whims of the client (so this is allowing us to import the Kolibri server time functionality without having to directly import this from Kolibri).

packages/hashi/src/mainClient.js
rtibbles: Sure, I just got caught up with the chopsticks :)

packages/hashi/test/mediator.spec.js
rtibbles: There was a reason I suggested it was good reading at the content dev meeting!

One nice thing I liked was that I also used Jest to do a mixture of unit and bit more 'integration-y' testing (although, obviously not full end to end). Within the mediator, and each of the shim specs, I mostly mocked and stubbed out the other functionality, but in the sandbox client, and external client tests, I ran a bit closer to real code to test out some of the mediator messaging functionality a bit better.

packages/hashi/src/mediator.js
rtibbles: Fun fact, during this process I discovered that Vue devtools uses `window.postMessage` for its messaging!

packages/hashi/src/mediator.js
rtibbles: Listen once, act many!

rtibbles:
Changed class names, but didn't update filenames. Also noticed something as I was adding comments that would have been a fairly major bug, so added a test for it, and fixed it.

packages/hashi/README.md
```diff
@@ -0,0 +1,36 @@
...
+
+`lerna bootstrap`
+
+Step 2: Switch to the Hashi package root dir
+
+`cd packages/hashi`
+
+Step 3: Build Hashi and test HTML5 zip
+
+`yarn run build`
```
kollivier, line 24: Just realized that we're no longer building the localstorage_test.zip file via `yarn run build`, but instructions below expect that to happen. We should either update the docs or the build script.

kollivier (changes requested):
To fully support Funza, the following changes need made:

- Sandbox needs to specify "allow-same-origin" permission to load multimedia assets loaded via JS
- BaseStorage.getItem should return null instead of undefined if a key does not exist
- Storage item keys need to be accessible as properties on the storage object, to enable `if ("key" in localStorage)` conditional tests. (override hasOwnProperty? I didn't know the right ES6 way to do this, and my naive attempts had no effect, so calling in the calvary!)

packages/hashi/README.md
rtibbles: Oh yeah, I didn't update the readme at all! Good extra thing to do.

rtibbles:
Agree with suggestions except enabling `allow-same-origin`, because if we do that, we basically have no sandboxing, and this entire exercise is pointless. Once you add `allow-same-origin` the iframe has access to all storage APIs, and the parent window object. I am going to look into the funza zip more to see what is going on with their asset loading in particular.

rtibbles:
Have now updated the README. Further, I have the Storage objects as compatible as possible with `delete` and `in` operators, as well as direct property assignment. These should be covered by additional tests to guarantee the behaviour.

For Funza in particular, the attempts I made at inserting the appropriate CORS flags disrupted rendering, not because of the CORS flags, but because I overrode the built in `Image` constructor. Instead, I have pursued a strategy of patching Pixi.js upstream to properly check for `window.origin === null`. See: https://github.com/pixijs/pixi.js/pull/5268

jamalex:
Just starting to review this (will be focusing on the Python piece, since I'm more familiar with it and @kollivier has the other parts covered well).

One high-level note: The PR description is fantastic, and helps to explain how this whole beautiful Rube Goldberg rigmarole works. Could that be included into the dev docs?

kolibri/core/content/templates/content/hashi.html
```diff
@@ -0,0 +1,14 @@
+{% load static %}
+<!DOCTYPE html>
+<html>
+    <head>
+        <meta charset="utf-8">
+        <meta http-equiv="X-UA-Compatible" content="IE=Edge,chrome=1">
+        <meta name="viewport" content="width=device-width, initial-scale=1">
```
jamalex, line 7: Do these get overridden when the Hashi content is loaded in? I'm wondering whether certain apps may need their own specific meta values in order to render nicely?

kolibri/core/content/test/test_zipcontent.py
```diff
@@ -18,14 +19,19 @@ class ZipContentTestCase(TestCase):
...
     test_name_1 = "testfile1.txt"
     test_str_1 = "This is a test!"
     test_name_2 = "testfile2.txt"
     test_str_2 = "And another test..."
 
     def setUp(self):
 
-        self.client = Client()
+        # Fetch with this header by default to run through non-hashi related behaviour.
+        self.client = Client(HTTP_X_REQUESTED_WITH="XMLHttpRequest")
```
jamalex, line 37: Might be clearer to have both `self.xhr_client` and `self.nonxhr_client` defined at the top here? Would also reduce needing to define new clients in each of the non-XHR tests.

kolibri/core/content/test/test_zipcontent.py
```diff
@@ -98,3 +106,60 @@ def test_access_control_allow_headers(self):
...
+    def test_non_xhr_request_for_html_return_hashi(self):
+        client = Client()
+        response = client.get(self.zip_file_base_url)
+        template = loader.get_template('content/hashi.html')
+        hashi_snippet = template.render({}, None)
+        self.assertEqual(response.content.decode(), hashi_snippet)
+
+    def test_xhr_request_for_html_return_html(self):
+        response = self.client.get(self.zip_file_base_url)
+        self.assertEqual(b"".join(response.streaming_content).decode(), self.index_str)
```
jamalex, line 142: Particular reason to use `streaming_content` instead of just `content`?

kolibri/core/content/views.py
```diff
@@ -27,6 +30,13 @@ def _add_access_control_headers(request, response):
         response["Access-Control-Allow-Headers"] = requested_headers
 
 
+def _add_content_security_policy_header(request, response):
+    # restrict CSP to only allow resources to be loaded from the Kolibri host, to prevent info leakage
+    # (e.g. via passing user info out as GET parameters to an attacker's server), or inadvertent data usage
+    host = request.build_absolute_uri(OPTIONS['Deployment']['URL_PATH_PREFIX']).strip("/")
```
jamalex: Awesome, I hadn't realized CSP supported restricting to subpaths! Could that actually be used to restrict requests to only be under the current zip file? That would prevent the Studio errors we've seen about accidental inclusion of too many ".." in font and image references, thereby hitting other server resources.

That could potentially interfere with the "ZIP dependency" (factoring-out resources like a library into a second shared zip) approach. But I'm not 100% on the naive version of that approach anyway; we may want to approach in a more intentional way.

jamalex (approved):
Masterpiece! Now I understand why so much typing was happening next to me here!

I see huge potential for a really interesting (and useful) blogpost about how we approach safe sandboxing in general, and Hashi.

Only a couple small suggestions from my side. Only looked at the Python code in detail. Skimmed the JavaScript. SKIPPED THE YARN LOCK FILE ALTOGETHER!

kolibri/core/content/views.py
jamalex: (we could at least restrict it to the /zipcontent/ endpoint, perhaps?)

radinamatic:
> Rube Goldberg rigmarole

Zap, add another thing I'll have to google and read through :100:

kollivier (approved):

kolibri/core/content/templates/content/hashi.html
rtibbles: Yes, the entire `documentElement.innerHTML` is replaced when the HTML is loaded, see here: https://github.com/learningequality/kolibri/pull/4540/files#diff-96bb9088342cae39df0fd69f29b92a64R32

kolibri/core/content/test/test_zipcontent.py
rtibbles: Because Django's `FileResponse` returns a streaming response!

kolibri/core/content/views.py
rtibbles: Actually, reading in more detail, it looks like we can't restrict by a subpath.
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/default-src#Sources

Can only specify scheme, host, and port. Will update this so that it doesn't give a false sense of security in here.

PR #4544, opened 2018-11-30
Create a script to automate the integration testing worksheet.

### LINKED ISSUE #4432: Create a script to automate the creation of the new integration testing tab at the spreadsheets.
### Observed behavior
To create a new [integration testing](https://docs.google.com/spreadsheets/d/1tol65GwoBdqzcsQyH0NHsWARGGKRJrbT6q1UeGH7kHE/edit?pli=1#gid=505990701) it needs to manually copy the base template and edit each scenarios to point it to the Gherkin `.feature` file link.

### Expected behavior
Create a script that can automate the creation of the new integration testing tab in the spreadsheets. It will also update all the scenarios links with the provided version.

mrpau-richard:
### Summary
I created a script to automate the new integration testing worksheet.

To make the script work on the BuildKite, the `GOOGLE_APPLICATION_CREDENTIALS` must have an access or share to [Integration testing spreadsheet](https://docs.google.com/spreadsheets/u/1/d/1tol65GwoBdqzcsQyH0NHsWARGGKRJrbT6q1UeGH7kHE/edit#gid=822923449).

### Reviewer guidance
The current behaviour on this script is to copy the base template of the Integration testing spreadsheet and update each `.Gherkin .feature files` links.

### References
This fixes #4432

### Contributor Checklist

### Reviewer Checklist


.buildkite/create_integration_testing_worksheet.py
```diff
@@ -0,0 +1,48 @@
...
+WKS_CELLS_RANGE = 100
+WKS_COLUMN = 'B'
+
+SCOPE = ['https://spreadsheets.google.com/feeds',
+         'https://www.googleapis.com/auth/drive']
+
+
+def copy_worksheet():
+	google_account = gspread.authorize(ServiceAccountCredentials.from_json_keyfile_name(WKS_CREDENTIALS, SCOPE))
+	wks = google_account.open(WKS_NAME)
```
indirectlylit: This is risky because according to the [docs](https://gspread.readthedocs.io/en/latest/api.html#gspread.Client.open)

> If there’s more than one spreadsheet with same title the first one will be opened.

[`open_by_key`](https://gspread.readthedocs.io/en/latest/api.html#gspread.Client.open_by_key) would be more reliable

.buildkite/create_integration_testing_worksheet.py
```diff
@@ -0,0 +1,48 @@
...
+	wks_id = wks.get_worksheet(WKS_INDEX_COPY).id
+	return wks.duplicate_sheet(wks_id, insert_sheet_index=1, new_sheet_id=None, new_sheet_name=WKS_TAG)
```
indirectlylit: Having a sheet per test run all in a single document looks a bit unwieldy:

image

Have you guys (cc @radinamatic) considered alternate structures?

For example, one possibility might be a new document per release branch, and then having a sheet per test run? In other words, you could have a structure like:

* _[folder]_ Kolibri testing spreadsheets
  * _[spreadsheet]_ 0.10.x
    * _[sheet]_ 0.10.0-beta1
    * _[sheet]_ 0.10.0-beta2
    * _[sheet]_ 0.10.0
    * _[sheet]_ 0.10.1-beta1
    * _[sheet]_ ...
  * _[spreadsheet]_ 0.11.x
    * _[sheet]_ 0.11.0-dev-notifications
    * _[sheet]_ 0.11.0-beta1
    * ...

Also, just a heads-up that the spreadsheet currently exists in a KA Lite folder in addition to the Kolibri folder:

ka

.buildkite/create_integration_testing_worksheet.py
```diff
@@ -0,0 +1,48 @@
...
+
+def edit_new_worksheet():
+	worksheet = copy_worksheet()
+	for count in range(WKS_CELLS_RANGE):
+		current_wks = WKS_COLUMN + str(count+1)
```
indirectlylit: isn't this `current_cell`?

.buildkite/create_integration_testing_worksheet.py
```diff
@@ -0,0 +1,48 @@
...
+
+
+WKS_CREDENTIALS = os.getenv("GOOGLE_APPLICATION_CREDENTIALS")
+WKS_TAG = os.getenv("BUILDKITE_TAG")
+
+WKS_TEMPLATE_TAG = 'release-v0.11.x'
+WKS_NAME = "Integration testing with Gherkin scenarios"
+WKS_INDEX_COPY = 0
+WKS_CELLS_RANGE = 100
+WKS_COLUMN = 'B'
```
indirectlylit: `INDEX_COPY`, `CELLS_RANGE`, and `COLUMN` could all use more descriptive variable names. For example, `SHEET_TEMPLATE_INDEX`, `MAX_ROW_COUNT`, and `FEATURE_LINKS_COLUMN`.

.buildkite/create_integration_testing_worksheet.py
```diff
@@ -0,0 +1,48 @@
...
+import os
+
+
+WKS_CREDENTIALS = os.getenv("GOOGLE_APPLICATION_CREDENTIALS")
+WKS_TAG = os.getenv("BUILDKITE_TAG")
+
+WKS_TEMPLATE_TAG = 'release-v0.11.x'
+WKS_NAME = "Integration testing with Gherkin scenarios"
+WKS_INDEX_COPY = 0
+WKS_CELLS_RANGE = 100
```
indirectlylit: Looks like we're assuming there will never be more than 100?

.buildkite/create_integration_testing_worksheet.py
```diff
@@ -0,0 +1,48 @@
...
+		final_value = val.replace(WKS_TEMPLATE_TAG, WKS_TAG)
+		if str(final_value):
+			worksheet.update_acell(current_wks, final_value)
+
+def main():
+    edit_new_worksheet()
+
+
+if __name__ == "__main__":
+    main()
```
indirectlylit, line 353: You're mixing tabs and spaces for indentation.

Spaces are standard - recommend configuring your editor.

.buildkite/create_integration_testing_worksheet.py
```diff
@@ -0,0 +1,48 @@
...
+	return wks.duplicate_sheet(wks_id, insert_sheet_index=1, new_sheet_id=None, new_sheet_name=WKS_TAG)
+
+def edit_new_worksheet():
+	worksheet = copy_worksheet()
+	for count in range(WKS_CELLS_RANGE):
+		current_wks = WKS_COLUMN + str(count+1)
+		val = worksheet.acell(current_wks, value_render_option='FORMULA').value
+		final_value = val.replace(WKS_TEMPLATE_TAG, WKS_TAG)
+		if str(final_value):
+			worksheet.update_acell(current_wks, final_value)
```
indirectlylit: This assumes our stories never change!

We have to actually read the contents of our feature files from the git repo and re-write all the rows and sections based on them.

indirectlylit (commented):
this is exciting!

It won't work as-is, but it's a good first step

.buildkite/create_integration_testing_worksheet.py
benjaoming: It could be a copy-paste thing. But should be caught by the pre-commit hooks.

@mrpau-richard could you try this in your virtualenv:

```
pip install -r requirements/dev.txt
pre-commit install
git commit --amend
```

benjaoming: The flake8 command should give you this error: https://lintlyci.github.io/Flake8Rules/rules/W191.html

.buildkite/create_integration_testing_worksheet.py
mrpau-richard: Noted, Thanks.

mrpau-richard:
@cpauya suggested to create a `.json config file` at the `develop/integration_testing/features` as the base template of the Integration testing spreadsheet. The script will use it, to get the `.feature` name and path. And will solve the issue of the name arrangement and path of each `.feature' file in the worksheet.

This is the sample format of the `.json` file.
```
{

    {
        'role': 'Super admin', 
        'scenarios': [
            {
                'feature': 'Super admin setup wizard'
                'path': '/features/super-admin/super-admin-setup-wizard.feature'
            },
            {
                'feature': 'Super admin sign-in'
                'path': '/features/super-admin/super-admin-setup-wizard.feature'
            }

        ],

    },
    {
        'role': 'Facility admin'
        'scenarios': [
            {
                'feature': 'Admin sign-in'
                'path': '/features/admin/admin-sign-in.feature'
            }
        ]
    }
    {
        'role': 'Coach'
        'scenarios': [
            {
                'feature': 'Coach sign-in'
                'path': '/features/coach/coach-sign-in.feature'
            }
        ]
    }
    {
        'role': 'Learner'
        'scenarios': [
            {
                'feature': 'Learner sign-in'
                'path': '/features/learner/learner-sign-in.feature'
            }
        ]
    }
    {
        'role': 'Guest'
        'scenarios': [
            {
                'feature': 'Guest change language'
                'path': '/features/guest/change-language.feature'
            }
        ]
    }

}
```

cc @benjaoming @indirectlylit

benjaoming:
> And will solve the issue of the name arrangement and path of each `.feature' file in the worksheet.

As mentioned in the meeting, I really think you should solve this by means of the file name. This will make it much easier to view the .feature files and locate them in Github and with filesystem navigators.

Entering each file in a .json will be additional work and open up for new errors. JSON is nice, but if the main purpose is human readability, it's not the best choice.

I'm wondering if everything can be achieved through:

1) Enumerating files and directories
2) Using symlinks for reused scenarios. Symlinks can have different names from their target, so its possible to reuse a scenario in another context.

Symlinks work in Git.. maybe have issues on Windows, not sure. But it may not be important.

indirectlylit:
Agreed. Currently the JSON file shown above is redundant with the existing file structure, which means extra maintenance cost and high likelihood of user error.

For an example of grabbing all the information about the git tree, see my recent work [parsing a font git repo](https://github.com/learningequality/kolibri/blob/develop/build_tools/i18n/noto_source.py#L183). Documentation of the [git tree API is here](https://developer.github.com/v3/git/trees/).

The main complexity I can think of is that currently there is a distinction between file names and the "human-readable" names inside the .feature files. I think you'll want to avoid having to read the contents of every file.

My recommendation therefore would be:
1. Use the file name as the source of truth, and use this string to populate the spreadsheet. Maybe switch to using underscores as word separators, and replace the underscores with spaces when you populate the spreadsheet.
2. Inside feature files, delete all name information so that we don't need to keep the contents of the file in sync with the file names. The first line of each file would become just `Feature:`

(cc @radinamatic)

benjaoming:
@indirectlylit 

Should this refactor perhaps take place in a separate PR?

indirectlylit:
Depends how long this work takes to complete:

* For the next week or so, no one is editing feature files so it's fine to do all in one PR
* If this work takes longer than another week, we'll start introducing conflicts as people start making modifications

cpauya:
@benjaoming You are right, JSON is poor for human-readability, we can use [YAML](https://stackoverflow.com/a/1729545/845481).  One advantage of this format is that we can document relationships (or dependencies) between scenarios with this document too.

Here's an example YAML for reference:

```yaml
purpose: >
  This is a list of platforms and Gherkin stories to be used for
  integration testing.  A script will be used to auto-generate
  a Google Doc worksheet based on the items here.
platforms: 
  - os: Windows 10
    pythons: 
      - python: Python 2.7.6 64-bit
        browsers: [Chrome, FF 61.0.1 64-bit, IE]

      - python: Python 3.4 64-bit
        browsers: [Chrome, FF 61.0.1 64-bit, IE]
  - os: Ubuntu 16.04
    pythons:
      - python: Python 3.5.2
        browsers: [Chrome 68.0.3440.84, FF 61.0.1 64bit]

      - python: Python 2.7
        browsers: [Chrome 68.0.3440.84, FF 61.0.1 64bit]
  - os: Raspbian
    pythons:
      - python: Python 3.2
        browsers: [Chromium, Firefox]
gherkin stories:
  - roles:
    - role: Super Admin
      path: features/super-admin
      scenarios:
        - feature: Super admin setup wizard
          filename: super-admin-setup-wizard.feature

        - feature: Super admin non-existent .feature yet
          filename: 

        - feature: Super admin sign-in
          filename: super-admin-setup-wizard.feature
    - role: Facility admin
      path: features/admin/
      scenarios:
        - feature: Admin sign-in
          filename: admin-sign-in.feature
```

Advantages of using a master-list file:

1. Single point of reference for integration testing scenarios in the source-code repo instead of looking at the Google Doc Integration Testing spreadsheet. Documentation for OS platforms, Python interpreters, web browsers and versions used for testing.

1. Adding scenarios in the auto-generated worksheet which has no equivalent `.feature` file just yet.  Refer to the [`IT Template`](https://docs.google.com/spreadsheets/d/1tol65GwoBdqzcsQyH0NHsWARGGKRJrbT6q1UeGH7kHE/) worksheet to see some rows without Github links but were used in testing.

1. Sorting of scenarios in the worksheet.

1. Script can create a worksheet (or other spreadsheets like MS Excel) based on the master-list contents with no need for a Google Doc template.

Disadvantages of using a master-list file:

1. Maintenance work for the master-list file.  (But we are already doing this in the Google Docs Integration Testing document.)

1. Prone to user-errors when adding/editing/deleting `.feature` files.

I think this opens another train of thought than originally planned for this script.  

Thoughts @mrpau-richard and @radinamatic?

benjaoming:
> One advantage of this format is that we can document relationships (or dependencies) between scenarios with this document too.

This is what I think we can obtain with symlinks. Here's an example of what I mean - nevermind the actual scenarios, but notice the links and the enumeration of files and folders.

```
.
├── 001-deviceadmin
│   ├── 001-admin-login.feature -> ../base/admin-login.feature
│   ├── 002-admin-can-view-import-dashboard.feature
│   ├── 003-admin-can-logout.feature -> ../base/logout.feature
│   ├── 004-learner-can-login.feature -> ../base/learner-login.feature
│   └── 005-learner-cannot-view-import-dashboard.feature
├── 002-useradmin
│   ├── 001-admin-can-login.feature -> ../base/admin-login.feature
│   └── 002-admin-can-create-a-coach.feature
└── base
    ├── admin-login.feature
    ├── learner-login.feature
    └── logout.feature
```

mrpau-richard:
Now I understand how the symlinks work on the Github. And we can use it as our bases on the Integration testing scenarios. But it will create a symlink scenario files at the repo.

indirectlylit:
> Symlinks work in Git.. maybe have issues on Windows, not sure. But it may not be important.

We do need to allow windows users to edit feature files

mrpau-richard:
I think it's better to do the YAML for the reference of all Gherkin scenarios, one of the advantages of it is the YAML can be edited in any platform. Let's just create a test of all the feature path if it exist when there is new a PR.

benjaoming:
What I meant was that symlinks via git on Windows *might* have issues. But it doesn't look like they have anymore?

https://github.community/t5/How-to-use-Git-and-GitHub/Windows-symlinks-do-they-play-well-with-Git-projects/td-p/4964

mrpau-richard:
This is now using the `ORDER.txt` to arrange the feature files at the Integration testing spreadsheet.

These are the remaining task:
- [x] Create some comments on how the script works
- [x] Fix the [structures](https://github.com/learningequality/kolibri/pull/4544#discussion_r237933193) of the integration testing spreadsheet (I need suggestions in this)
- [x] Read about the Google [write APIs](https://developers.google.com/analytics/devguides/config/mgmt/v3/limits-quotas) limits
- [ ] Setup the Google api credentials at the BuildKite


cc @benjaoming @indirectlylit @radinamatic @cpauya

.buildkite/create_integration_testing_worksheet.py
```diff
@@ -6,61 +6,106 @@
...
+    return str_name.strip()
+
+
+def get_role_name(str_arg):
+    return str_arg.replace("-", " ").capitalize() + " scenarios"
+
+
+def get_feature_dir_path():
+    patent_path = os.path.dirname(os.getcwd())
+    return patent_path + "/integration_testing/features/"
```
cpauya: I don't understand what is `patent_path`?  Maybe a misspelling for `parent_path`? 😸

.buildkite/create_integration_testing_worksheet.py
```diff
@@ -6,61 +6,106 @@
...
+def check_file_exist(str_arg):
+    return os.path.exists(str_arg)
+
+
+def check_name(str_arg1, str_arg2):
+    if str_arg2 == "":
+        return False
+    elif check_file_exist(str_arg1) and str_arg2[0] != "#":
+        return True
+    return False
```
cpauya: A nitpick, but maybe this could be re-written as:
```python
def check_name(...
    if str_arg2 != "":
        if check_file_exists(...
            return True
    return False
```

.buildkite/create_integration_testing_worksheet.py
```diff
@@ -6,61 +6,106 @@
...
+
+
```
cpauya, line 97: This `fetch_feature_files()` function is also where we call the Google API to update or insert spreadsheet rows, right?  

So instead of making an API call for every feature file that we read (every loop iteration), you can put those features in a list during the file read loop first.  When done with the loop, you then make the API call using the batch update function as @benjaoming mentioned: https://developers.google.com/sheets/api/reference/rest/v4/spreadsheets/batchUpdate - will it still be quota-limited as you said?

.buildkite/create_integration_testing_worksheet.py
```diff
@@ -6,61 +6,106 @@
 # Environment Variable/s:
     * GOOGLE_APPLICATION_CREDENTIALS = Your service account key.
 """
-import itertools
 import os
 
 import gspread
 from oauth2client.service_account import ServiceAccountCredentials
```
cpauya, line 16: Nitpick, for "beautiful" codes to read: 😸 
1. Order your `import` statements alphabetically.
2. Put a new line after the last `import` and before the first `from ... import ...`.

cpauya (changes requested):
Some nitpicks but the `batchUpdate()` Google API call is important to try if it solves the API quota limit issue.

mrpau-richard:
Hi guys, I did implemented the batch update functionality to the script to fix the Google API write limits. The only missing is to setup and run this at the BuildKite.

CC @benjaoming @indirectlylit @cpauya

benjaoming:
Thanks for the update @mrpau-richard  - you need to merge in or rebase to the latest upstream `develop` to fix the build errors.

.buildkite/create_integration_testing_worksheet.py
```diff
@@ -6,61 +6,106 @@
...
+    worksheet = spreadsheet.get_worksheet(SHEET_INDEX)
+    order_name = "/ORDER.txt"
+    role_order_path = feature_dir + order_name
+    counter = 3
+    with open(role_order_path, 'r') as read_role:
+        role_lines = read_role.readlines()
+        for role in role_lines:
+            role_file_name = role.strip()
+            role_path = feature_dir + role_file_name + "/"
+            role_order_path = role_path + order_name
```
cpauya: Please avoid re-using the same variable/s (example: `role_order_path`) inside inner code blocks to prevent confusion.

.buildkite/create_integration_testing_worksheet.py
```diff
@@ -6,61 +6,106 @@
...
+def check_name(str_arg1, str_arg2):
+    if str_arg2 == "":
+        return False
+    elif check_file_exist(str_arg1) and str_arg2[0] != "#":
+        return True
+    return False
+
+
+def fetch_feature_files():
+
```
cpauya, line 104: Please put a comment on what this function does.

cpauya (dismissed):
Your updates are better than the previous.  Please note my requests for comment/s to help the reviewer/s.

mrpau-richard:
The script is now working at the BuildKite. Thanks to @aronasorman he helped me to upload my Google credentials at the server.

To check it here is the sample build https://buildkite.com/learningequality/kolibri/builds/9233#f4362155-10bc-4acd-ac03-cf4b7e49712d

I think this PR is ready to merge.

cc @radinamatic @benjaoming @cpauya @indirectlylit

rtibbles:
If this is intended for 0.12.0 please update the milestone and retarget to 0.12.x branch.

indirectlylit:
are we ready for re-review?

mrpau-richard:
I updated the target branch to 0.12.x, please review this again. Thanks.

.buildkite/build_worksheet.sh
```diff
@@ -0,0 +1,47 @@
...
+fi
+
+PYTHON_CMD="$PYTHON_PATH .buildkite/create_integration_testing_worksheet.py"
+$PYTHON_CMD
+if [ $? -ne 0 ]; then
+    echo ".. Abort!  Can't execute '$PYTHON_CMD'."
+    exit 1
+fi
+
+buildkite-agent artifact upload '.buildkite/spreadsheet-link.txt'
```
indirectlylit, line 47: Caveat: I'm not super familiar with buildkite so I don't know best practices.

That said, this file seems a little strange to me: I would not have expected a shell script to be creating a virtual environment, installing pip dependencies one at a time, and then spawning other pipeline processes.

Instead, I would have expected something like:

* the build would be isolated in a docker environment, so a python virtualenv would not be needed
* the dependencies for this pipeline step would be defined in a pip `requirements.txt` file or a Pipenv `Pipfile`
* the individual steps (install dependencies, run `create_integration_testing_worksheet.py`, upload artifacts) are specified as multiple commands in the 'Create integration testing' block of the `pipeline.yml` file

.buildkite/create_integration_testing_worksheet.py
```diff
@@ -0,0 +1,353 @@
...
+    Check the role and feature file name.
+    """
+    if str_arg == "" or str_arg[0] == "#":
+        return False
+    return True
+
+
+def create_artifact(str_arg):
+    buikite_path = get_kolibri_path() + "/.buildkite/"
+    txt_path = buikite_path + 'spreadsheet-link.txt'
```
indirectlylit: typo  - `buikite_path`

.buildkite/create_integration_testing_worksheet.py
```diff
@@ -0,0 +1,353 @@
...
+        end_index_counter = end_index + 3
+        role_counter += 1
+
+    cell_range = SHEET_TPL_COLUMN + str(SHEET_TPL_START_VALUE) \
+        + ":" + SHEET_TPL_COLUMN + str(SHEET_TPL_START_VALUE + len(feature_contents))
+    cell_list = worksheet.range(cell_range)
+    cell_counter = 0
+    for cell in cell_list:
+        try:
+            feature_vall = feature_contents[cell_counter]
```
indirectlylit: `feature_vall` - typo?

indirectlylit (approved):
overall this looks good to me. I left one concern I have about how the process is integrated into the pipeline.

I haven't tested this, but feel free to merge, especially if it helps with integration testing

.buildkite/build_worksheet.sh
mrpau-richard: @indirectlylit I created another [issue](https://github.com/learningequality/kolibri/issues/4953) for this one. I need time to read more about docker.

mrpau-richard:
@indirectlylit thanks for reviewing this. I have a new commit about the typo errors and created a new [issue](https://github.com/learningequality/kolibri/issues/4953) about the build process of the spreadsheet. For now I think we can merge this PR if you have no other comments.

benjaoming:
Great job @mrpau-richard - is there a particular integration testing spreadsheet generated in this PR?

indirectlylit:
@benjaoming it's hard-coded to point at this one 
* https://github.com/learningequality/kolibri/pull/4544/files#diff-6577c7f97f02f5843710d3511b33f8f3R20
* https://docs.google.com/spreadsheets/d/1kVhg0evo9EV2aDo10KdIIjwqsoT4rISR7dJzf6s_-RM/

PR #4565, opened 2018-12-07
Actionable Notifications for Coaches - Backend

jredrejo:
### Summary

 * Insert notifications events into the db 
 * Exposes api 

### Reviewer guidance
 * Saving of `ResourceIndividualCompletion`,  `LessonResourceIndividualCompletion`,  `QuizIndividualCompletion` and `LessonExerciseIndividualNeedsHelpEvent` are made by now
 * API  /coach/api/notifications/?collection_id=<classroom_id> ( or <learngroup_id>) (filters to provide only current day notifications)

### References
Notifications are created in an async thread, with code based on the work from #4225 

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated


kolibri/core/notifications/models.py
```diff
@@ -0,0 +1,93 @@
...
+    Multiple = "MultipleUnsuccessfulAttempts"
+
+
+@python_2_unicode_compatible
+class KolibriNotification(models.Model):
+    id = UUIDField(primary_key=True)
+    notification_type = models.CharField(max_length=200, choices=[(t, t.value) for t in NotificationType], blank=True)
+    user_id = UUIDField()
+    lesson_id = UUIDField(null=True)
+    classroom_id = UUIDField()
```
indirectlylit, line 105: Some of the metadata here is specific to the class of notifications that we are currently building: "learner progress notifications" or something like that.

I think it's probably too early to worry about making this new Django app a generalized notification system that can encompass all notification types (e.g. the information we currently poll from iceqube), although I do think it's appropriate to have that goal in the back of our minds.

In the meantime maybe we should rename `KolibriNotification` to `LearnerProgressNotification` so that there's no assumption of generality?

kolibri/core/notifications/signals.py
```diff
@@ -0,0 +1,77 @@
...
+    filtered_lessons = Lesson.objects.filter(
+        lesson_assignments__collection__in=learner_groups,
+        is_active=True,
+        resources__regex=r"" + content_id + ""
+    ).distinct()
+    lesson_resources = {lesson.collection_id: (lesson.id, lesson.resources) for lesson in filtered_lessons}
+    return lesson_resources
+
+
+@receiver(post_save, sender=ContentSummaryLog)
```
indirectlylit: There's an interesting perspective that [signals decrease code readability](https://lincolnloop.com/blog/django-anti-patterns-signals/), and that it's equally effective to call this function directly from `ContentSummaryLog.save()`.

I tend to agree because this function is already tightly coupled to the structure of `ContentSummaryLog`. I believe both strategies allow us to migrate to a background queue, but maybe I'm missing something and signals make that easier.

thoughts?

indirectlylit (commented):
Overall this approach looks good to me, and is essentially what I believe we've talked about.

(Longer-term we can think about what a generalized notification system might look like, how consumers subscribe to certain event types, and how producers create those events, and how authorization is handled. My sense is we're going to eventually want to distinguish two kinds of notifications: those that can live only in-memory and disappear if the server is restarted, versus those that persist indefinitely... But I don't think we need to pin any of that down now.)

kolibri/core/notifications/models.py
jredrejo: ok, no problem at all

kolibri/core/notifications/signals.py
jredrejo: One of my first approaches was using `ContentSummaryLog.save()`. But with that method the actual save happens after this function is executed. This has two undesired side-effects:
1. saving of `ContentSummaryLog` is delayed by the function
2. any aggregated query made by the function does not have the instance of ContentSummaryLog in the calculations

Point 2 made me discard using `ContentSummaryLog.save()`

jredrejo: Definitely,I will move this to be executed in an iceqube task launched by `ContentSummaryLog.save()`

kolibri/core/notifications/models.py
```diff
@@ -0,0 +1,100 @@
...
+
+        # No opinion for all other scenarios
+        return None
+
+
+class NotificationType(Enum):
+    Resource = "ResourceIndividualCompletion"
+    Quiz = "QuizIndividualCompletion"
+    Help = "LessonResourceIndividualNeedsHelpEvent"
+    Lesson = "LessonResourceIndividualCompletion"
```
jonboiser: This should just be "LessonIndividualCompletion" (a learner completed a lesson).

kolibri/core/notifications/signals.py
```diff
@@ -0,0 +1,177 @@
...
+                notifications.append(lesson_notification)
+
+    save_notifications(notifications)
+
+
+@receiver(post_save, sender=ExamLog)
+def parse_exam_log(sender, instance, **kwargs):
+    if not instance.closed:
+        return
+    user_classrooms = instance.user.memberships.all()
```
jonboiser: This check may not be necessary. A learner can only be assigned a Lesson or Exam if they belong to a classroom. So if a Learner is completing a quiz, then that means they have been assigned an exam through a classroom (and have the membership).

This _might_ be guarding against an unusual case where a student somehow loses that membership, but still does the exam somehow, but I'm not entirely sure that's possible in practice.

kolibri/core/notifications/models.py
jredrejo: Done

kolibri/core/notifications/signals.py
jredrejo: Done

kolibri/core/notifications/signals.py
jredrejo: Done in an async thread. Chosen that option based on the work from Richard on #4225 because creating and executing a Django command everytime a logger saves is a clear overhead. The queue is set to be procesed every 5 seconds.

rtibbles:
@jredrejo it looks like you have failing tests for two reasons:
1) Your migration is not Python 2/3 compatible, so Py2.7 tests think that there is a migration missing.
2) The postgres tests are failing because you have only done a database configuration for sqlite, and not for postgres as well - look below the SQLite config in the settings base.py!

jredrejo:
@rtibbles all tests pass now. Migration problems where due to the use of the Enum class, that I have replaced by a custom made class.
Postgres problems where due to the `postgres_test.py` setup that overrides `base.py` config. In a normal environment base.py worked fine with postgres because the router was not added, but in tests the router was added causing the problem.  I have fixed it removing the `NotificationsRouter` in the model by checking settings in runtime.

kolibri/core/notifications/models.py
```diff
@@ -0,0 +1,115 @@
...
+
+
+@python_2_unicode_compatible
+class LearnerProgressNotification(models.Model):
+    id = UUIDField(primary_key=True)
+    notification_type = models.CharField(max_length=200, choices=NotificationType.choices(), blank=True)
+    user_id = UUIDField()
+    classroom_id = UUIDField()
+    contentnode_id = UUIDField(null=True)
+    contentnode_id = UUIDField(null=True)
```
indirectlylit: Is it a typo that this line is duplicated?

It might be helpful to store the content ID in addition to the node ID.

kolibri/core/notifications/models.py
```diff
@@ -0,0 +1,115 @@
...
+
+
+class NotificationType(myEnum):
+    Resource = 'Resource'
+    Quiz = 'Quiz'
+    Help = 'Help'
+    Lesson = 'Lesson'
+    values = {'Resource': 'ResourceIndividualCompletion',
+              'Quiz': 'QuizIndividualCompletion',
+              'Help': 'LessonResourceIndividualNeedsHelpEvent',
```
indirectlylit: is there a 'needs help' event for exercises that are not in a lesson?

kolibri/core/notifications/models.py
```diff
@@ -0,0 +1,115 @@
...
+    id = UUIDField(primary_key=True)
+    notification_type = models.CharField(max_length=200, choices=NotificationType.choices(), blank=True)
+    user_id = UUIDField()
+    classroom_id = UUIDField()
+    contentnode_id = UUIDField(null=True)
+    contentnode_id = UUIDField(null=True)
+    lesson_id = UUIDField(null=True)
+    quiz_id = UUIDField(null=True)
+    reason = models.CharField(max_length=200, choices=HelpReason.choices(), blank=True)
+    timestamp = DateTimeTzField(default=local_now)
```
indirectlylit, line 110: Is there any risk that our notification querying logic will be messed up if the server clock gets reset? Should we maybe add a monotonically increasing per-class integer field?

indirectlylit (commented):
Hey @jredrejo - just doing a final review and trying to make sure our notification model is fully expressive of the info we need.

A few notes and questions:

1. Could you update this PR to also include 'started' events? (refs #4660)
2. Currently the enumerated values all have the word 'Individual' in them, e.g. `ResourceIndividualCompletion`. However this seems redundant because _all_ notifications are about an individual
3. In the course of working on the user-facing strings, I ended up [enumerating all possible event types](https://github.com/learningequality/kolibri/blob/develop/kolibri/plugins/coach/assets/src/views/new/shared/status/statusStrings.js) and I think it might imply a slightly different way to parameterize the notification types: instead of a single `notification_type` field, what do you think of having two fields: `object_type` (`quiz`, `lesson`, `resource`, `exercise`) and `event_type` (`started`, `completed`, `needs_help`)? This parameterization would map a little closer to the front-end implementation

kolibri/core/notifications/models.py
jredrejo: It was a typo. It's fixed now. According to the specs, the frontend only needs the node id, it does not use the content id (if needed, adding channel id could make sense as well)

kolibri/core/notifications/models.py
jredrejo: No, all the events are triggered only in a Lesson context. Exercises made out of a lesson don't trigger any event.

kolibri/core/notifications/models.py
jredrejo: Yes, it might happen, https://github.com/jredrejo/kolibri/blob/notifications_backend_phase_1/kolibri/plugins/coach/api.py#L165 would be affected. 
However I don't get how you want to do it. In case an increasing integer field is used, how do you know how many rows you want to provide? Are you thinking of adding another field to save the latest provided row and send the difference? If that's the case, I can't see how to manage different coaches accesing to the same classroom.

jredrejo:
Hi @indirectlylit 

> * Could you update this PR to also include 'started' events? (refs #4660)
Yes, working on it. I have left a comment in #4660 with some doubts.


> * Currently the enumerated values all have the word 'Individual' in them, e.g. `ResourceIndividualCompletion`. However this seems redundant because _all_ notifications are about an individual

I recall having this conversation with @jonboiser before, and we agreed to change the api events removing the `Individual` part of the event names. Howver for the values he gave reasons to keep the Individual part as it seems there might be Group or Class events in the future. In case that's not gonna happen, of course I can remove them, but it will break current frontend, so it's good all of us agree on the change.

> * In the course of working on the user-facing strings, IndividualI ended up [enumerating all possible event types](https://github.com/learningequality/kolibri/blob/develop/kolibri/plugins/coach/assets/src/views/new/shared/status/statusStrings.js) and I think it might imply a slightly different way to parameterize the notification types: instead of a single `notification_type` field, what do you think of having two fields: `object_type` (`quiz`, `lesson`, `resource`, `exercise`) and `event_type` (`started`, `completed`, `needs_help`)? This parameterization would map a little closer to the front-end implementation

It depends, if we are going to continue adding those `event_types` to all the `object_types` so more combinations will appear, it makes totally sense. If we are going to keep only this four or five events, it might be a bit of overengineering. Also, this has to be coordinated with the frontend too as it would break current development there @jonboiser

jredrejo:
@indirectlylit :


> * Currently the enumerated values all have the word 'Individual' in them, e.g. `ResourceIndividualCompletion`. However this seems redundant because _all_ notifications are about an individual
> * In the course of working on the user-facing strings, I ended up [enumerating all possible event types](https://github.com/learningequality/kolibri/blob/develop/kolibri/plugins/coach/assets/src/views/new/shared/status/statusStrings.js) and I think it might imply a slightly different way to parameterize the notification types: instead of a single `notification_type` field, what do you think of having two fields: `object_type` (`quiz`, `lesson`, `resource`, `exercise`) and `event_type` (`started`, `completed`, `needs_help`)? This parameterization would map a little closer to the front-end implementation

These two points are now done. DB model and api has changed now. This is an example of the object provided by the API:
```
    {
        "timestamp": "2019-01-15T20:28:48.636805+01:00",
        "user_id": "286aa5a9bfed5ad1430a19d01e65402b",
        "classroom_id": "9da65157a8603788fd3db890d2035a9f",
        "lesson_id": "209711dfff083fb689ff2ac7dd296bd9",
        "contentnode_id": "0d62c6ce0b1949f8b5d27e377983227a",
        "object": "Resource",
        "event": "Completed",
        "type": "ResourceCompleted"
    },
```

indirectlylit:
merging to a new shared feature branch, `coach-data` (cc @jonboiser)

indirectlylit:
work to continue here: https://github.com/learningequality/kolibri/pull/4733

jredrejo:
@indirectlylit  
- added pagination
- added  `after` parameter to recover from id:
http://localhost:8000/coach/api/notifications/?collection_id=9da65157a8603788fd3db890d2035a9f&after=3

returns all the notifications with id >3 

http://localhost:8000/coach/api/notifications/?collection_id=9da65157a8603788fd3db890d2035a9f
returns all notifications for today.



@jonboiser serializer now returns also  1) user’s full name; 2) lesson title (if there is one); 3) resource title as you requested

indirectlylit:
thanks @jredrejo! 

(refs: https://github.com/learningequality/kolibri/pull/4733/commits)

Issue #1255, opened 2017-04-19
standardize on strings as type returned for all IDs


indirectlylit:
Some IDs are returned from the API as integers (like users and facilities), and others are strings (like channels and content nodes).

We should probably standardize on strings everywhere. This may be happening anyway as part of the Morango upgrade? We'll likely need to do some client-side work to remove any assumptions about integer-based ID types.

cc @ralphiee22 @jamalex @rtibbles

rtibbles:
More confusingly, I have come across places where the id is an integer but is serialised as a string, and in other places the id is an integer and serialised as such.

DXCanas:
@rtibbles could it be that the int strings are JSON?

rtibbles:
If I recall correctly, the difference seems to be whether the id is the primary key, or a reference to a primary key of a related model, which means directly comparing the two is difficult.

Either way, I agree with having one way of sending them to the front end.

rtibbles:
Note that this is now somewhat mitigated, due to the introduction of Morango, where all syncable model ids are now strings.

PR #4586, opened 2018-12-11
Cache key conversion

ralphiee22:
### Summary

When passing an integer as the id to the cache key mechanism, it does not coerce the id to a string creating mismatching cache keys. 
Also `const cacheKey`  used to be locally scoped to within the if statement, so we make its scope local to the function. This solves an issue of creating another `cacheKey` variable but scoped as `cacheKey$1`, which does not resolve correctly.

### Reviewer guidance

Do api calls still work? Does caching still work?

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/core/assets/src/api-resource.js
+++ b/kolibri/core/assets/src/api-resource.js
@@ -618,7 +618,9 @@ export class Resource {
         {},
         ...Object.keys(allParams)
           .sort()
-          .map(paramKey => ({ [paramKey]: allParams[paramKey] }))
+          .map(paramKey => ({
+            [paramKey]: paramKey === this.idKey ? String(allParams[paramKey]) : allParams[paramKey],
+          }))
       )
     );
   }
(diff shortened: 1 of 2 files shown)


indirectlylit:
I thought IDs were always UUID strings? Where do we use numbers?

ralphiee22:
Testing the plugin, we passed a user model which has an integer ID.

indirectlylit:
I think that's the real bug then - see: #1255

ralphiee22:
But for people who build plugins, if they create new models then we don't expect them to standardize on having to pass in strings for IDs.

indirectlylit:
I'd sort of prefer to put in validation and throw an exception rather than being more permissive on the allowed inputs.

It's a fairly minor point, but in API design I think strict type and schema checking is more maintainable and easier to document than flexibility. It's always easier to add flexibility than to take it away.

rtibbles:
I would agree, except that the reason this issue arose is because we are already being permissive in our inputs, and coercing the input, which led to this issue in the first place (otherwise, the cache key would not be in an inconsistent state). So this is a fix for current intended behaviour that is acting inconsistently.

rtibbles (approved):

Issue #4543, opened 2018-11-29
Developer Docs structural update for new top-level sections


benjaoming:
### Observed behavior

The top-level structure does not reflect the lower levels, and there are intersections and overlaps in the nested themes.

Navigating to a subsection is therefore counter-intuitive.

I think the fundamental issue has been that our technical docs are often written+structured with a bit too much immediate ease/comfort of the author :) And flaws in the structure propagated further confusion of the ideas of top-level elements in our original structure were. I'm not even sure if we discussed them properly or questioned them subsequently. @indirectlylit has done great work to counter this, but the overall structure hasn't given us enough to work with in an iterative setting.

Hence the suggestion to reorganize a bit.

It's inspired by the [What Nobody Tells You About Documentation](https://www.youtube.com/watch?v=t4vKPhjcMZg) talk.

### Expected behavior

Better navigation that either eliminates thematic overlaps or introduces adequate cross-references.

A draft is shared here with the full rework, please update it and use it to reflect on the scope of changes (which I think are realistic) : 

https://docs.google.com/document/d/15A1YfgTSgrCnb2FTzTTrRfHTrK6G4AHJesGiexVYPfQ/edit

So in this structure, I put _Developer *how-to*_ in the first section, and _References_ in the last. _Architecture_ is renamed to _Background_.

This hopefully creates a better yes/no situation for choosing where to put a specific item:

 * Does it instruct a developer on _how to_ specifically do something? Put it in _Developer how-tos_.
 * Is it a reference? I.e. does it show and explain details from a concrete implementation. Put it in References.
 * Is it an explanation of choices or introduction to more abstract relations in the stack? Put it in _Background_.
 * Is it a mix of the above? Split it up. Do not mix references and background. Do not mix how-tos and backgrounds. Do not mix how-tos and references.
 * Does it instruct about how to use or install Kolibri? Don't put it here!

indirectlylit:
I like the idea of being more intentional about choosing "what type of documentation am I writing?"

Pasting his overview slide for reference:

![image](https://user-images.githubusercontent.com/2367265/49261809-88f50c80-f3f8-11e8-9742-2abf6eec7f6d.png)

Also [here's some research](https://www.nngroup.com/articles/flat-vs-deep-hierarchy/) about the use of hierarchies in websites. The summary:

1. Content is more **discoverable** when it's not buried under multiple intervening layers. All other things being equal, deep hierarchies are more difficult to use.
[...]

PR #2798, opened 2017-12-01
Make core-snackbar button text white

christianmemije:
### Summary

Make core-snackbar button text white

localhost_8000_learn_ 14

### Reviewer guidance

Go offline to make disconnected snackbar appear. 

### References

Fixes #2759 

----

### Contributor Checklist

- [x] PR has the correct target milestone
- [x] PR has 'needs review' or 'work-in-progress' label
- [x] If PR is ready for review, it has been assigned or requests review from someone and labeled as 'needs review'
- [x] PR has been fully tested manually
- [ ] Documentation is updated
- [ ] External dependencies files were updated (`yarn` and `pip`)
- [ ] Link to diff of internal dependency change is included
- [x] Screenshots of any front-end changes are in the PR description
- [x] PR does not introduce [accessibility regressions](http://kolibri.readthedocs.io/en/develop/dev/manual_testing.html#accessibility-a11y-testing)
- [ ] CHANGELOG.rst is updated for high-level changes
- [ ] You've added yourself to AUTHORS.rst if you're not there

### Reviewer Checklist

- [ ] Automated test coverage is satisfactory
- [ ] PR has been fully tested manually

PR #4587, opened 2018-12-11
Migrate all core theming to vuex driven dynamic styles

rtibbles:
### Summary
#4524 did a lot of good things, but by separating each build into a separate, independent process, it broke the ability to do dynamic coreAPI injection at build time. Along with this, it also broke the ability to do custom styling of Kolibri at build time using a build time specified external plugin.

This PR rectifies some of this in a slightly different way by migrating all styling that hitherto has been done by variables specified in `core-theme.scss` to be driven by a core object in Vuex state.

As a bonus, it also adds a colour customization tool in dev mode to let people dynamically change styles.
…

### Reviewer guidance
Check that styling is still applied correctly in the many components that have been refactored - I still have some more work to do manually verifying all this also, although I was mostly checking as I went, and did a post-pass of high visibility areas.

Also read the additions to the docs and see if they make sense!

### References
No current issues this addresses, but will enable custom theming for accessibility and cosmetic purposes in future.

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated


docs/frontend_architecture/core.rst
```diff
@@ -47,7 +47,39 @@ These modules would now be available for import anywhere with the following stat
...
+------------------
+
+In addition to some of the global SCSS variables, we also leverage Vuex state to drive overall theming of the application, in order to allow for more flexible theming (either for accessibility or cosmetic purposes). As such, much of our colour styles are defined in Javascript variables kept in Vuex state, which are then programmatically accessed using Vue.js style bindings in order to apply stylings inline to elements.
+
+This works for most use cases, however, inline styles cannot apply pseudo-selectors (e.g. ':hover', ':focus', '::before') - in order to accomplish this, we leverage the `Aphrodite<https://github.com/Khan/aphrodite>`__ library to generate computed classes that apply these stylings through a programmatically generated style sheet. This should be generally performant, but we are limiting their use for now to cases where we have pseudo selectors and styles that may conflict with pseudo-selector applied styles (to have clear selector precedence). In order to apply a style using Aphrodite, define a style object as a computed property, similarly to how you might for a Vue.js style binding. Pseudo-selectors can be encoded within this object thusly:
```
indirectlylit: Also for transitions.

One thing that's [been discussed lately](https://github.com/learningequality/kolibri/issues/4543) is distinguishing 'how-to' guides, references, and understanding-oriented discussions. If the goal here is mostly how-to, we could probably shorten it a bit.

For example:

> There are two cases where dynamic styles cannot be applied directly to DOM elements:
> 
> * styles using [pseudo-classes](https://developer.mozilla.org/en-US/docs/Web/CSS/Pseudo-classes)
> * styles applied during [Vue transitions](https://vuejs.org/v2/guide/transitions.html)
> 
> For these cases, it's necessary to define a "computed class" using the `$computedClass` function. This returns an auto-generated class name which can be used like a standard CSS class name. Under the hood, this uses [Aphrodite](https://github.com/Khan/aphrodite) to create unique classes for each set of inputs given, so be careful not to abuse this feature!

docs/frontend_architecture/core.rst
```diff
@@ -47,7 +47,39 @@ These modules would now be available for import anywhere with the following stat
...
+
+.. code-block:: javascript
+
+  pseudoStyle() {
+    return {
+      ':hover': {
+        backgroundColor: white,
```
indirectlylit: let's actually reference a style here so it's clear what the point is

kolibri/core/assets/src/core-app/apiSpec.js
```diff
@@ -181,14 +181,14 @@ export default {
...
   styles: {
-    theme,
     definitions,
     keenVars,
     buttonAndLinkStyles,
   },
   urls,
   utils: {
     contentNode,
+    colour,
```
indirectlylit, line 191: 🇬🇧

docs/frontend_architecture/core.rst
rtibbles: Am I being very thick? Is `background-color` not a style?

indirectlylit: sorry I wasn't clear - `white` is not a vuex-driven style :)

rtibbles: Ah, gotcha. yes.

kolibri/core/assets/src/core-app/constructor.js
```diff
@@ -64,6 +66,11 @@ export default class CoreApp {
       mediator.setReady();
     };
 
+    if (process.env.NODE !== 'production') {
+      const colourPicker = require('../utils/colourPicker').default;
+      colourPicker.start();
+    }
```
indirectlylit, line 62: This is fun, but I don't think it should be turned on by default in the devserver.

Also, `colourPicker.js` is unnecessary. You can just drop `ColourPicker.vue` in `CoreBase.vue`.

 Maybe add a prop on `CoreBase` like

```javascript
showColorPicker: {
  type: Boolean,
  default: false,
}
```

and then add to the template

```html
...

    <GlobalSnackbar />
    <ColourPicker v-if="showColorPicker" />
  </div>

</template>
```

kolibri/core/assets/src/views/CoreSnackbar/index.vue
```diff
@@ -108,23 +108,12 @@
...
     margin: 16px;
-
-    /deep/ .ui-snackbar__action-button {
-      font-weight: bold;
-      color: $core-bg-light;
-    }
-
-    /deep/ .ui-button--type-secondary:hover:not(.is-disabled) {
-      background-color: rgba(255, 255, 255, 0.05);
-    }
```
indirectlylit, line 127: Looks like we lost some styles (regression of #2798):

image

indirectlylit (approved):
couple small comments. Overall I think it's best to get this in the mainline and open new issues as we find them

kolibri/core/assets/src/core-app/constructor.js
rtibbles: Yeah, I did it like this so that it wouldn't even get bundled into the production build.

indirectlylit:
getting some travis failures too

PR #4588, opened 2018-12-12
Fixes sign-up validation

### LINKED ISSUE #4575: <Improvement> Not showing user is exist notification on UI.
### Observed behavior
We are able to create user using CREATE AN ACCOUNT option which is on login page of kolibri. But when someone uses existing username to create account, it will not show any kind of existing user notification on UI.
Not able to distinguish whether account is exist or not.
### Expected behavior
It must show existing username notification on UI if user is exist.

### Steps to reproduce
1. Login with Admin and go to the facility.
2. Click on settings.
3. Select Allow learners to create accounts.
4. Logout and click on CREATE AN ACCOUNT button and use existing username to create account.

### Context

 * Kolibri version : Kolibri 0.11.0 
 * Operating system : ubuntu 14.04
 * Browser : Chrome

indirectlylit:
### Summary

* fixes regression - display validation when username already exists (#4575)
* fixes issue where our 'unknown error' window was hidden
* implements this screen as a standard `core-base` immersive view rather than a custom special case, and delete a lot of unneeded styles and code

### Reviewer guidance

On the error handling side, I'm not at all clear what the intended API is. This change seems to fix the problem, but now we have three different styles of raising errors, with different schemas for `detail`:

A simple string:

```python
            raise serializers.ValidationError(
                detail='An account with that username already exists.',
                code=error_constants.USERNAME_ALREADY_EXISTS
            )
```

A dict of lists of strings:

```python
            raise serializers.ValidationError(detail={'username': ['An account with that username already exists.']},
                                              code=error_constants.USERNAME_ALREADY_EXISTS)
```

A dict of strings:

```python
            raise serializers.ValidationError(detail={'classroom': 'This user is already in a group in this class'},
                                              code=error_constants.USER_ALREADY_IN_GROUP_IN_CLASS)
```

This makes no sense to me and I can't find any documentation about how it's intended to be used.

### References

fixes #4575

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/core/assets/src/views/CoreBase.vue
+++ b/kolibri/core/assets/src/views/CoreBase.vue
@@ -1,9 +1,7 @@
 <template>
 
   <div>
-    <!-- temporary hack, resolves flicker when using other templates -->
-    <template v-if="navBarNeeded">
-
+    <template v-if="!fullScreen">
       <ImmersiveToolbar
         v-if="immersivePage"
         :appBarTitle="toolbarTitle || appBarTitle"
@@ -13,7 +11,6 @@
         :height="headerHeight"
         @nav-icon-click="$emit('navIconClick')"
       />
-
       <template v-else>
         <AppBar
           ref="appBar"
@@ -41,12 +38,12 @@
           @toggleSideNav="navShown=!navShown"
         />
       </template>
-
     </template>
 
     <AppBody
       :topGap="appBodyTopGap"
       :bottomGap="bottomMargin"
+      :fullScreen="fullScreen"
     >
       <AuthMessage
         v-if="notAuthorized"
@@ -100,10 +97,10 @@
         required: false,
         default: '',
       },
-      // Prop that determines whether to show nav components
-      navBarNeeded: {
+      // Prop that determines whether to show nav components and provide margins
+      fullScreen: {
         type: Boolean,
-        default: true,
+        default: false,
       },
       // reserve space at the bottom for floating widgets
       bottomMargin: {
@@ -134,11 +131,13 @@
         required: false,
         default: false,
       },
+      // generally a 'back' or 'close' icon
       immersivePageIcon: {
         type: String,
         required: false,
         default: 'close',
       },
+      // link to where the 'back' button should go
       immersivePageRoute: {
         type: Object,
         required: false,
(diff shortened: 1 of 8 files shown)


kolibri/core/auth/serializers.py
```diff
@@ -55,8 +55,10 @@ class FacilityUserSignupSerializer(FacilityUserSerializer):
 
     def validate_username(self, value):
         if FacilityUser.objects.filter(username__iexact=value).exists():
-            raise serializers.ValidationError(detail={'username': ['An account with that username already exists.']},
-                                              code=error_constants.USERNAME_ALREADY_EXISTS)
+            raise serializers.ValidationError(
+                detail='An account with that username already exists.',
+                code=error_constants.USERNAME_ALREADY_EXISTS
+            )
```
indirectlylit, line 61: My guess is that this needs to be handled differently because some DRF magic is changing the result since it's inside `validate_username`?

kolibri/core/assets/src/views/CoreBase.vue
```diff
@@ -100,10 +97,10 @@
         required: false,
         default: '',
       },
-      // Prop that determines whether to show nav components
-      navBarNeeded: {
+      // Prop that determines whether to show nav components and provide margins
+      fullScreen: {
```
DXCanas, line 101: Seems this introduces a new flavor `core-base`:
- `fullScreen`
- `immersivePage`
- "regular"

You think it's worth revisiting our approach to setting that "mode"? 

I remember some ideas floating around about transitioning to a vuex module approach? Seems to be the direction that https://github.com/learningequality/kolibri/pull/4495 was headed in.

indirectlylit: This was mostly just renaming `navBarNeeded` to `fullScreen` so that it could live up to its full potential. It allowed me to delete all this special-case stuff: https://github.com/learningequality/kolibri/pull/4588/files#diff-ab033c27dbeccc48ef01312c1fee6894L3

indirectlylit: > You think it's worth revisiting our approach to setting that "mode"?

I don't think so, no. Current implementation seems to work fine after this change

indirectlylit:
@khangmach this PR updates the sign-up page to use our standard 'immersive modal' rather than the custom layout used previously. Does this work?

| before | after |
|--|--|
| image | image |

khangmach:
@indirectlylit yup seems fine to me

jonboiser (approved):

indirectlylit:
@ralphiee22 would you mind weighing in on the question I put in the PR description? I'd like to clarify the API for generating errors.

Issue #2279, opened 2017-09-25
ios momentum scrolling not working


indirectlylit:
report: no momentum scrolling on mobile. Chrome app, iPhone 6, iOS 10.3.3

HansGam:
.

jonboiser:
We reopened this because the `-webkit-overflow-scrolling :touch` rule interacts with the "scrolling pane" CSS or code to make the App Bar unusable on iOS, so it was removed in #5100.

indirectlylit:
from https://css-tricks.com/snippets/css/momentum-scrolling-on-ios-overflow-elements/

```css
.class {
  overflow-y: scroll; /* has to be scroll, not auto */
  -webkit-overflow-scrolling: touch;
}
```

indirectlylit:
this also relates to #2722 and #5326

My sense is that we'll want to try and address all three at once by refactoring how scrolling and the app bar are handled more generally

jonboiser:
Trying to add this back in is causing a lot of problems, especially with how the layout is dynamically changing. It seems to have something to do with iOS's desire to have scrollable items to be immediately visible to enable this momentum scrolling.

Here are some problems I've already seen:

1. The side navigation drawer not appearing
1. 'main' div not appearing in Learn > Classes, Search, Channels, etc. while waiting for data from the backend.
1. Unable to dynamically set top positions for items in CoreBase (e.g. if you dynamically add a top positioning to a div, everything above it will be blank, even if you intended something to be there).

It doesn't seem to be a problem on Coach pages, though, so maybe we should figure out what is wrong with our layouts before enabling momentum scroll, otherwise we would need to test every page of the UI to make sure we didn't bring in more problems.

indirectlylit:
I've come across evidence that having `body` be the main scrolling pane rather than an internal `div` might be a way to refactor our layout. If you have a simple scrolling `body`, iOS will apply momentum scrolling behaviors by default without any vendor-prefix CSS hacks.

jonboiser:
I've got that to work, but had to move the scroll listener from the nested div to `window`. There also seems to be a little bit of a performance drop on big pages like the User Page with thousands of users, but works well with simpler pages.

Issue #2158, opened 2017-09-12
can't scroll PDFs on mobile safari (iOS)


indirectlylit:
Go to page, try to scroll PDF. Doesn't work in embedded or full-screen mode.

Reproducible in Browserstack:

![image](https://user-images.githubusercontent.com/2367265/30351642-40c4d8c0-97d1-11e7-9ed5-c4e97792da37.png)

kollivier:
Unable to reproduce in iOS 9+ on device or simulator in the general case, and iOS v7 was notably buggy after the huge UI rework, so I am going to close this issue.

Issue #1046, opened 2017-03-09
Videos do no load on iPhone


radinamatic:
## Summary

So far I didn't manage to see any video on "real" iPhone, but they do work on Android phones, even though it's messy trying to start/stop them because of the #1045. 

## System information

 - Version: 0.2.0alpha5.dev35
 - Operating system:  iPhone 5S
 - Browser: all (Safari, Chrome, Firefox)

## Screenshots

![video](https://cloud.githubusercontent.com/assets/1457929/23767259/d403eade-04bc-11e7-905f-b08d26957bda.JPG)

![video2](https://cloud.githubusercontent.com/assets/1457929/23767288/ea0f90e4-04bc-11e7-80cd-f71be03948b1.JPG)

christianmemije:
https://github.com/videojs/video.js/issues/2945

christianmemije:
This is the error that is thrown:
`VIDEOJS: ERROR: (CODE:4 MEDIA_ERR_SRC_NOT_SUPPORTED) The media could not be loaded, either because the server or network failed or because the format is not supported. `

hhsnopek:
@christianmemije I just chased this issue down a rabbit hole and finally found a solution for me. When I was compressing my video I didn't set the proper encoding for [H.264](https://trac.ffmpeg.org/wiki/Encode/H.264). If you use `ffmpeg`, add the flags `-profile:v baseline -level 3.0` to support all iOS devices. I hope this helps! Good luck

rtibbles:
Oh yes, I remember us having this issue with KA Lite on iOS devices also.

oezh:
> @christianmemije I just chased this issue down a rabbit hole and finally found a solution for me. When I was compressing my video I didn't set the proper encoding for [H.264](https://trac.ffmpeg.org/wiki/Encode/H.264). If you use `ffmpeg`, add the flags `-profile:v baseline -level 3.0` to support all iOS devices. I hope this helps! Good luck

@hhsnopek Where do you add those flags?

indirectlylit:
@oezh see https://ffmpeg.org/

PR #4589, opened 2018-12-12
Add full URLs in exercises to ensure they always load

### LINKED ISSUE #1049: Images in exercises do no load on iPhone
## Summary

Starting to get really annoyed with iPhone... Oh wait, I never liked it in the first place! :unamused: 

## System information

 - Version: 0.2.0alpha5.dev35
 - Operating system:  iPhone 5S
 - Browser: all (Safari, Chrome, Firefox)

## How to reproduce

Try any of these (I could post more, but these will do for start):

https://instantschools.learningequality.org/learn/#/topics/11d9de56da744f98877cc9fe710bb78d/content/44d60f677fa04938afdba9807e46b8ea

https://instantschools.learningequality.org/learn/#/topics/11d9de56da744f98877cc9fe710bb78d/content/678f22a8bfec4f5bb4969d0817a23499

https://instantschools.learningequality.org/learn/#/topics/11d9de56da744f98877cc9fe710bb78d/content/dddefe5a72c84a37b74f22e9e2c54784

## Screenshots

exercise2

exercise4

rtibbles:
### Summary
Previously we were just replacing a relative path to origin for image URLs inside perseus json blobs. On some browsers these were not then properly resolved, resulting in images being missing.

This fixes that by always creating a full URL for replacing in the json blob. It attempts to construct this with a URL derived from the `HTTP_REFERER` header, which appears to be widely supported, as this will give the host that the frontend request this from, not the host that Kolibri is running on (in case Kolibri is behind a proxy).

It also uses this to more accurately set our Content Security Policy header, so that it still works even behind a poorly configured proxy.

### Reviewer guidance
Check that Perseus images still work.
Check that they now work on older iOS.
See my tests, do they cover everything?

### References
Fixes #1049

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/core/content/test/test_zipcontent.py
+++ b/kolibri/core/content/test/test_zipcontent.py
@@ -87,8 +87,8 @@ class ZipContentTestCase(TestCase):
         self.assertEqual(response.get("Content-Security-Policy"), "default-src 'self' 'unsafe-inline' 'unsafe-eval' data: http://testserver")
 
     def test_content_security_policy_header_http_referer(self):
-        response = self.client.get(self.zip_file_base_url + self.test_name_1, HTTP_REFERER="http://testserver")
-        self.assertEqual(response.get("Content-Security-Policy"), "default-src 'self' 'unsafe-inline' 'unsafe-eval' data: http://testserver")
+        response = self.client.get(self.zip_file_base_url + self.test_name_1, HTTP_REFERER="http://testserver:1234/iam/a/real/path/#thatsomeonemightuse")
+        self.assertEqual(response.get("Content-Security-Policy"), "default-src 'self' 'unsafe-inline' 'unsafe-eval' data: http://testserver:1234")
 
     def test_access_control_allow_origin_header(self):
         response = self.client.get(self.zip_file_base_url + self.test_name_1)
@@ -112,11 +112,11 @@ class ZipContentTestCase(TestCase):
         response = self.client.get(self.zip_file_base_url + self.test_name_3, HTTP_REFERER=server_name)
         self.assertEqual(
             response.content.decode('utf-8'),
-            self.test_str_3.replace("$" + exercises.IMG_PLACEHOLDER, (server_name + self.zip_file_base_url).strip("/")))
+            self.test_str_3.replace("$" + exercises.IMG_PLACEHOLDER, (server_name.replace('http:', '') + self.zip_file_base_url)).strip("/"))
 
     def test_json_image_replacement_no_http_referer_header(self):
         server_name = "http://testserver"
         response = self.client.get(self.zip_file_base_url + self.test_name_3)
         self.assertEqual(
             response.content.decode('utf-8'),
-            self.test_str_3.replace("$" + exercises.IMG_PLACEHOLDER, (server_name + self.zip_file_base_url).strip("/")))
+            self.test_str_3.replace("$" + exercises.IMG_PLACEHOLDER, (server_name.replace('http:', '') + self.zip_file_base_url)).strip("/"))
--- a/kolibri/core/content/test/test_zipcontent.py
+++ b/kolibri/core/content/test/test_zipcontent.py
@@ -110,9 +110,13 @@ class ZipContentTestCase(TestCase):
     def test_json_image_replacement_http_referer_header(self):
         server_name = "http://testserver"
         response = self.client.get(self.zip_file_base_url + self.test_name_3, HTTP_REFERER=server_name)
-        self.assertEqual(response.content, self.test_str_3.replace("$" + exercises.IMG_PLACEHOLDER, (server_name + self.zip_file_base_url).strip("/")))
+        self.assertEqual(
+            response.content.decode('utf-8'),
+            self.test_str_3.replace("$" + exercises.IMG_PLACEHOLDER, (server_name + self.zip_file_base_url).strip("/")))
 
     def test_json_image_replacement_no_http_referer_header(self):
         server_name = "http://testserver"
         response = self.client.get(self.zip_file_base_url + self.test_name_3)
-        self.assertEqual(response.content, self.test_str_3.replace("$" + exercises.IMG_PLACEHOLDER, (server_name + self.zip_file_base_url).strip("/")))
+        self.assertEqual(
+            response.content.decode('utf-8'),
+            self.test_str_3.replace("$" + exercises.IMG_PLACEHOLDER, (server_name + self.zip_file_base_url).strip("/")))
--- a/kolibri/core/content/test/test_zipcontent.py
+++ b/kolibri/core/content/test/test_zipcontent.py
@@ -5,6 +5,7 @@ import zipfile
 
 from django.test import Client
 from django.test import TestCase
+from le_utils.constants import exercises
 
 from ..models import LocalFile
 from ..utils.paths import get_content_storage_file_path
@@ -22,6 +23,8 @@ class ZipContentTestCase(TestCase):
     test_str_1 = "This is a test!"
     test_name_2 = "testfile2.txt"
     test_str_2 = "And another test..."
+    test_name_3 = "testfile3.json"
+    test_str_3 = "A test of image placeholder replacement ${placeholder}".format(placeholder=exercises.IMG_PLACEHOLDER)
 
     def setUp(self):
 
@@ -41,6 +44,7 @@ class ZipContentTestCase(TestCase):
         with zipfile.ZipFile(self.zip_path, "w") as zf:
             zf.writestr(self.test_name_1, self.test_str_1)
             zf.writestr(self.test_name_2, self.test_str_2)
+            zf.writestr(self.test_name_3, self.test_str_3)
 
         self.zip_file_obj = LocalFile(id=self.hash, extension=self.extension, available=True)
         self.zip_file_base_url = self.zip_file_obj.get_storage_url()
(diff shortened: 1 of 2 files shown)


kolibri/core/content/views.py
```diff
@@ -27,6 +30,35 @@ def _add_access_control_headers(request, response):
...
+        # Reconstruct the parsed URL using only the scheme(0) and host + port(1)
+        zipcontent = urlunparse((parsed_referrer_url[0], parsed_referrer_url[1], zipcontent, '', '', ''))
+    return zipcontent.encode()
+
+
+def get_host(request, parsed_referrer_url):
+    if parsed_referrer_url:
+        host = urlunparse((parsed_referrer_url[0], parsed_referrer_url[1], '', '', '', ''))
+    else:
+        host = request.build_absolute_uri(OPTIONS['Deployment']['URL_PATH_PREFIX'])
```
kollivier, line 60: If the page is refreshed or reloaded, will the parsed_referrer_url still be set? If not, do we want to have some sort of guard for the `http://kolibri` case and revert to the old behavior of not adding a host to the URL when it happens? That way, at least for some other browsers that handle the `http:/zipcontent/` URLs will continue to work in this case.

kolibri/core/content/views.py
```diff
@@ -79,11 +113,11 @@ def get(self, request, zipped_filename, embedded_filepath):
                 response = FileResponse(zf.open(info), content_type=content_type)
                 file_size = info.file_size
             else:
+                image_prefix_url = generate_image_prefix_url(zipped_filename, parsed_referrer_url)
```
indirectlylit: Minor note:

There's some ambiguity about the type of `parsed_referrer_url` because it can be implicitly `None`  from `get_referrer_url` or a tuple-like thing from `urlparse`.

Might be cleaner to avoid passing this value around between functions.

For example could do:

```python
image_prefix_url = generate_image_prefix_url(zipped_filename, request.META.get('HTTP_REFERER'))
```

And then you can move the `urlparse` logic inside a redefined

```python
def generate_image_prefix_url(zipped_filename, http_referer=None):
```

That avoids passing around the `None`/`urlparse` thing.

indirectlylit: ah didn't notice you use it below, too

kolibri/core/content/views.py
rtibbles: Yes, all requests to fetch JSON are done from within an XHR request, rather than a direct page load, so the header should always be set.

What you are suggesting is what I have done - if the referrer header is not set (which it will not be on e.g. a direct page load) then it falls back to the old behaviour.

kolibri/core/content/views.py
rtibbles: Yeah, the only reason I moved it out into separate functions was for code complexity linting reasons.

rtibbles:
Looks like my code is not Python3 compatible!

kolibri/core/content/views.py
```diff
@@ -99,7 +133,7 @@ def get(self, request, zipped_filename, embedded_filepath):
 
         # restrict CSP to only allow resources to be loaded from the Kolibri host, to prevent info leakage
         # (e.g. via passing user info out as GET parameters to an attacker's server), or inadvertent data usage
-        host = request.build_absolute_uri(OPTIONS['Deployment']['URL_PATH_PREFIX']).strip("/")
+        host = get_host(request, parsed_referrer_url)
```
rtibbles: Realized that actually this may not help in the only case that matters (loading an HTML page inside an iframe), as the referer header may not be set.

rtibbles: Oh, I am incorrect - marvellous. Looks like the `Referer` header *is* set on the request for an iframe's src.

kolibri/core/content/views.py
rtibbles: Decided to just directly invoke it from inside the functions using it, as it should not be an expensive operation.

kolibri/core/content/views.py
kollivier: Calling `build_absolute_uri` was being called before when setting the `host` variable for the `Content-Security-Policy` header, so that part is the same, but it was not being used when doing the `content_with_path` replacement, so I wanted to double-check that `build_absolute_uri` wouldn't ever return a bad host name in this case. (As if it did, `http:/zipcontent/...` will at least work in some browsers, whereas `http://kolibri/zipcontent/...` won't work anywhere.) Very possible you learned a way to ensure `get_absolute_uri` and `get_host` don't return `http://kolibri/` after our discussion on Slack, but just wanted to double-check it.

rtibbles: This function is only generating the host for the CSP header. See the `generate_image_url_prefix` function for the behaviour for image placeholder substitution URL generation - there the old behaviour is preserved, we just then add the previously defined path to the schema + host of the referer header if available.

kollivier: Gah, sorry, I'm blind. :( Thanks for clarifying!

kollivier (approved):

kolibri/core/content/test/test_zipcontent.py
```diff
@@ -82,6 +86,10 @@ def test_content_security_policy_header(self):
         response = self.client.get(self.zip_file_base_url + self.test_name_1)
         self.assertEqual(response.get("Content-Security-Policy"), "default-src 'self' 'unsafe-inline' 'unsafe-eval' data: http://testserver")
 
+    def test_content_security_policy_header_http_referer(self):
+        response = self.client.get(self.zip_file_base_url + self.test_name_1, HTTP_REFERER="http://testserver")
```
jamalex: Perhaps put a longer URL, with a path, to make sure that gets stripped down properly.

kolibri/core/content/views.py
```diff
@@ -27,6 +30,37 @@ def _add_access_control_headers(request, response):
...
+
+def get_host(request):
+    parsed_referrer_url = get_referrer_url(request)
+    if parsed_referrer_url:
+        host = urlunparse((parsed_referrer_url[0], parsed_referrer_url[1], '', '', '', ''))
+    else:
+        host = request.build_absolute_uri(OPTIONS['Deployment']['URL_PATH_PREFIX'])
+    return host.strip("/")
+
+
```
jamalex, line 63: Would it be nicer to extract these functions to a utils file? Or do you think they'll only ever be used here?

kolibri/core/content/views.py
```diff
@@ -27,6 +30,37 @@ def _add_access_control_headers(request, response):
         response["Access-Control-Allow-Headers"] = requested_headers
 
 
+def get_referrer_url(request):
+    if request.META.get('HTTP_REFERER'):
+        # If available user HTTP_REFERER to infer the host as that will give us more
```
jamalex: user -> use

jamalex (approved):
Nice! A couple small questions/tweaks, but otherwise looks good to go!

benjaoming:
@rtibbles @indirectlylit 

In #1049 it was mentioned:

> we currently block access to Kolibri via iOS safari

Is this still active and should it then be unblocked?

benjaoming:
@rtibbles Re: PR description

> It also uses this to more accurately set our Content Security Policy header, so that it still works even behind a poorly configured proxy.

That is great, and it might become relevant pretty soon that we have an Nginx proxy in front of Kolibri, meaning that both `localhost:80` and `localhost:8080` might be active at the same time. We'll want of course to figure out all this such that eventually Kolibri only exists behind one port or that this is somehow configurable. Btw. not so related to the CSP... but do not assume the host name is ever fixed. Kolibri can be accessed from both `http://localhost`, `http://1.2.3.4` and `http://my.own.dns`. We can probably use Django's `ALLOWED_HOSTS` setting in this relation, since it's solving the same type of issues.

rtibbles:
Yeah, I have no intention of assuming the host name is fixed! Is there a better way than what I have done for systematically determining the host name via which the request was made?

rtibbles:
> Is this still active and should it then be unblocked?

I think there are a few other issues that have iOS Safari tags on them, we might want to double check those first.

benjaoming:
> Is there a better way than what I have done for systematically determining the host name via which the request was made?

I think that the `HTTP_REFERER` isn't necessary set in all cases. If someone just straight-up types an address. Or if applications like RACHEL are deep-linking to resources. Or in cases where the browser has blocked it.

Could `request.META.get("HTTP_HOST")` together with `request.META.get("HTTPS")` be fine?

We could also add a utility function in case that someone at some point desires more functionality? Like `get_absolute_content_path(request, path)`, which could in some unknown scenario be pointing to a different location for content.

benjaoming:
> I think there are a few other issues that have iOS Safari tags on them, we might want to double check those first.

Hmm, but can the iOS reporter then actually test the builds in this PR?

kollivier:
I was unable to reproduce #1046 and #2158 when I was testing on iOS, and both of those tickets are more than a year old so it's possible that changes to Kolibri fixed those issues. I haven't tested #2279 but I definitely don't feel like that's a blocker, and aside from this issue those are the only three that are listed for iOS / Safari.

I would vote for removing the block, and re-considering if we start seeing a lot of new error reports from iOS / Safari users.

rtibbles:
Thanks for testing them, @kollivier! When I've fixed up this PR and merged, can remove Safari from the blacklist in a follow up PR.

indirectlylit:
sounds good to me!

indirectlylit:
@benjaoming 

> Hmm, but can the iOS reporter then actually test the builds in this PR?

yes we've determined that under the hood, chrome and safari are using the same webkit renderer

benjaoming:
@rtibbles [something has gone wrong](http://community.learningequality.org/t/missing-images-graphs-questions/911/37), and absolute URLs are appended to their relative paths:

> One last additional bit of info. Looking at the failed links found in the debugger interface, I see the resources that load correctly have a correctly formed URL. Whereas, those that don’t load have this duplicated frontend of the URL as shown here.

> `http://10.0.2.100:8080/learn/http://10.0.2.100:8080/zipcontent/eca85362c5a5c8370b08cae21edb69be.perseus/images/0041a87d6104cbbdb130f3b63baf4b57523e5acd.svg`

Would you be able to add a regression test for this case, too?

rtibbles:
Yeah, looks like this issue here is that Perseus does not expect the scheme to be included in the URL, will fix this up.

rtibbles:
> Could request.META.get("HTTP_HOST") together with request.META.get("HTTPS") be fine?

@benjaoming the issue with using `HTTP_HOST` is that it is not sent on the request, but added by Django, so this may differ from the external route that the frontend made the request to (so this would bypass any nginx or other proxy).

rtibbles:
Have replicated the base issue again on Browserstack, using iPhone 6S, and have now confirmed a fix (needed to remove the `http` scheme from the generated URL).

kolibri/core/content/views.py
rtibbles: I think we can leave them here for now. If they are needed elsewhere, we can move them.

PR #4599, opened 2018-12-13
Pin nodejs version in Dockerfile

lyw07:
### Summary

This PR is to fix the issue that nodejs version cannot be pinned when installing from PPA, because the pinned version got removed from PPA when a new version is released.

This PR introduces a new environment variable in Dockerfile `NODE_VERSION` to pin the version of nodejs, downloads the deb installer of specific nodejs from https://deb.nodesource.com/node_10.x/pool/main/n/nodejs and installs it.

### Reviewer guidance

Please check if the build runs successfully.

### References

Example error on buildkite: https://buildkite.com/learningequality/kolibri/builds/8795#0f5c102b-e95f-40c9-9f90-1c4d26fd1490

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/docker/base.dockerfile
+++ b/docker/base.dockerfile
@@ -1,29 +1,30 @@
 FROM ubuntu:bionic
 
-# install latest python and nodejs
+ENV NODE_VERSION=10.14.1
+
+# install latest python
 RUN apt-get update && \
     DEBIAN_FRONTEND=noninteractive apt-get install -y \
       curl \
-      software-properties-common
-
-# Install nodejs and add 'hold' such that it doesn't get upgraded
-RUN curl -sL https://deb.nodesource.com/setup_10.x | bash -
+      software-properties-common \
+      gettext \
+      git \
+      git-lfs \
+      psmisc \
+      python2.7 \
+      python-pip \
+      python-sphinx
 
 # add yarn ppa
 RUN curl -sS https://dl.yarnpkg.com/debian/pubkey.gpg | apt-key add -
 RUN echo "deb https://dl.yarnpkg.com/debian/ stable main" | tee /etc/apt/sources.list.d/yarn.list
 
+# install nodejs and yarn
 RUN apt-get update && \
-    DEBIAN_FRONTEND=noninteractive apt-get install -y \
-      gettext \
-      git \
-      git-lfs \
-      nodejs=10.14.1-1nodesource1 \
-      psmisc \
-      python2.7 \
-      python-pip \
-      python-sphinx \
-      yarn
+    curl -sSO https://deb.nodesource.com/node_10.x/pool/main/n/nodejs/nodejs_$NODE_VERSION-1nodesource1_amd64.deb && \
+    dpkg -i ./nodejs_$NODE_VERSION-1nodesource1_amd64.deb && \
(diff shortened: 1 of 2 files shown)


benjaoming:
Hi @lyw07 

The solution fits well, I think that what is described about keeping the files themselves in the archive but removing them from the `Releases` repository manifest is indeed what's happening.. and this would be the way to fix it.

> This PR is to fix the issue that nodejs version cannot be pinned when installing from PPA, because the pinned version got removed from PPA when a new version is released.

Nit-pick: This isn't technically speaking a PPA. While the differences are very small, a PPA is a repository that is hosted via Launchpad and has the specific feature of `apt-add-repository` being able to fetch and add the PGP signing keys for that particular repository. Maybe there are other differences, too.. for instance, with a Launchpad PPA, you can only add packages for currently active Ubuntu dists. The node-js repository is actually just a repository, so users have to install the repository keys manually before using it :)

Oh, some more background: Because of a bit of complexities using `apt-add-repository`, I have actually defaulted to using the generic way of manually adding keys. Effectively treating our PPA as a normal repository. This was more of a documentation issue, because I couldn't figure out how to write "but if you are on system B, C, D, please use this method".

docker/base.dockerfile
```diff
@@ -1,29 +1,30 @@
...
-      curl \
-      software-properties-common
-
-# Install nodejs and add 'hold' such that it doesn't get upgraded
-RUN curl -sL https://deb.nodesource.com/setup_10.x | bash -
-
-# add yarn ppa
-RUN curl -sS https://dl.yarnpkg.com/debian/pubkey.gpg | apt-key add -
-RUN echo "deb https://dl.yarnpkg.com/debian/ stable main" | tee /etc/apt/sources.list.d/yarn.list
+ENV NODE_VERSION=10.14.1
```
benjaoming, line 3: I think this should be in the `env.list` such that we have everything collected neatly in one place :)

docker/base.dockerfile
```diff
@@ -1,29 +1,30 @@
...
 
+# install latest python
```
benjaoming: Not sure if that comment is very accurate. It wasn't before, neither and that's my fault :)

benjaoming (approved):
A couple of comments, but feel free to merge. It's a nice and clean solution.

docker/base.dockerfile
lyw07: @benjaoming yes I was putting it into env.list first, but it seems like during build time, it doesn't read from env.list file. Also I notice that `docker build` doesn't have the `--env-file` option (https://docs.docker.com/engine/reference/commandline/build/#options). So I guess `env.list` is only available during `docker run` not `docker build`? In this case, we need to have the node version in image build time to download the deb installer, so I feel like putting ENV in the Dockerfile seems to be the only way to make it accessible during build time.

lyw07:
@benjaoming thank you so much for the full explanation! I agree with you that the word "PPA" is not accurate here. It's great to learn something more about "PPA" vs repository!

Also I will push another commit to fix the comment issue you mentioned above. Thanks!

PR #4615, opened 2018-12-17
Add NotificationCard component with $trs

jonboiser:
### Summary
These are placeholder components for the 1) individual notification card and 2) the "activity feed" that should have all of their localizable strings.

### Reviewer guidance

…

### References

…

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated


kolibri/plugins/coach/assets/src/views/NotificationCard.vue
```diff
@@ -0,0 +1,32 @@
...
+<script>
+
+  export default {
+    name: 'NotificationCard',
+    components: {},
+    props: {},
+    computed: {},
+    methods: {},
+    $trs: {
+      dotSeparatedHeader: '{string1} ● {string2}',
```
indirectlylit: This should not be a translation string: the dot is something we don't want translators to touch.

indirectlylit (commented):
looks good - one small change requested

kolibri/plugins/coach/assets/src/views/NotificationCard.vue
indirectlylit: Small note for future reference: we need to explicitly list non-ascii characters in our font-generation code: https://github.com/learningequality/kolibri/blob/develop/build_tools/i18n/fonts.py#L417

The dot is already listed, so we're good here, but just a heads-up.

build_tools/i18n/fonts.py
```diff
@@ -415,6 +415,7 @@ def _get_common_strings():
         "“",
         "”",
         "•",
+        "●",
```
indirectlylit, line 418: that's a big dot :)

indirectlylit (approved):

PR #4582, opened 2018-12-11
update network address strings

indirectlylit:
### Summary

Cleans up the network address strings a little.

In the first page:

* renames 'New address' to 'Add new address' to make it a verb
* move it below the info message so the ordering flows a bit better
* reword info message to be consistent with other parts of the app and less judgmental

In the second page, added explanatory text for the fields

| before | after |
|--|--|
| image  | image |
| image | image |

### Reviewer guidance

Do these changes look good?

### References

There's no issue filed, but it came up during the initial hack session that these fields were a bit confusing

----

### Contributor Checklist

PR process:

- [x] PR has the correct target branch and milestone
- [x] PR has 'needs review' or 'work-in-progress' label
- [x] If PR is ready for review, a reviewer has been added. (Don't use 'Assignees')
- [ ] If this is an important user-facing change, PR or related issue has a 'changelog' label
- [ ] If this includes an internal dependency change, a link to the diff is provided

Testing:

- [x] Contributor has fully tested the PR manually
- [x] If there are any front-end changes, before/after screenshots are included
- [ ] Gherkin stories have been updated
- [ ] Unit tests have been updated

### Reviewer Checklist

- Automated test coverage is satisfactory
- PR is fully functional
- PR has been tested for [accessibility regressions](http://kolibri-dev.readthedocs.io/en/develop/manual_testing.html#accessibility-a11y-testing)
- External dependency files were updated if necessary (`yarn` and `pip`)
- Documentation is updated
- Contributor is in AUTHORS.rst

PR #4634, opened 2018-12-19
add explanation for import options

indirectlylit:
### Summary

Adds descriptive text to the three import options

| Before | After |
|--|--|
| image | image |

Reproduced below:

* **Kolibri Studio (online)** - Download content directly from Learning Equality if you have access to the public internet
* **Local network or internet** - Transfer content between two running instances of Kolibri or a mirror of Kolibri Studio
* **Attached drive or memory card** - Copy content directly to the computer running Kolibri if you have a drive that has been pre-loaded with content

### Reviewer guidance

Feedback on the strings would be appreciated

### References

https://github.com/learningequality/clearinghouse/issues/321

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/plugins/device_management/assets/src/views/ManageContentPage/SelectNetworkAddressModal/AddAddressForm.vue
+++ b/kolibri/plugins/device_management/assets/src/views/ManageContentPage/SelectNetworkAddressModal/AddAddressForm.vue
@@ -137,9 +137,10 @@
       },
     },
     $trs: {
-      addressDesc: "The network address can be an IP like '10.0.0.10' or a URL like 'example.com':",
+      addressDesc:
+        "The network address can be an IP and port like '192.168.0.100:8080' or a URL like 'example.com':",
       addressLabel: 'Full network address',
-      addressPlaceholder: 'e.g. http://123.456.7.89:8080',
+      addressPlaceholder: 'e.g. 192.168.0.100:8080',
       cancelButtonLabel: 'Cancel',
       errorCouldNotConnect: 'Could not connect to this network address',
       errorInvalidAddress: 'Please enter a valid IP address, URL, or hostname',
--- a/kolibri/plugins/device_management/assets/src/views/ManageContentPage/SelectTransferSourceModal/SelectImportSourceModal.vue
+++ b/kolibri/plugins/device_management/assets/src/views/ManageContentPage/SelectTransferSourceModal/SelectImportSourceModal.vue
@@ -96,11 +96,11 @@
       selectLocalRemoteSourceTitle: 'Select a source',
       loadingMessage: 'Loading connections…',
       studioDescription:
-        'Download content directly from Learning Equality if you have access to the public internet',
+        "Import content channels from Learning Equality's library if you are connected to the internet",
       networkDescription:
-        'Transfer content between two running instances of Kolibri or a mirror of Kolibri Studio',
+        'Import content channels from Kolibri running on another device, either in the same local network or on the internet',
       localDescription:
-        'Copy content directly to the computer running Kolibri if you have a drive that has been pre-loaded with content',
+        'Import content channels from a drive. Channels must first be exported onto the drive from another Kolibri device with existing content',
     },
     methods: {
       ...mapActions('manageContent/wizard', ['goForwardFromSelectImportSourceModal']),
--- a/kolibri/plugins/device_management/assets/src/views/ManageContentPage/SelectTransferSourceModal/SelectImportSourceModal.vue
+++ b/kolibri/plugins/device_management/assets/src/views/ManageContentPage/SelectTransferSourceModal/SelectImportSourceModal.vue
@@ -25,18 +25,21 @@
         :value="ContentSources.KOLIBRI_STUDIO"
         :disabled="kolibriStudioIsOffline || formIsDisabled"
         :autofocus="!kolibriStudioIsOffline"
+        :description="$tr('studioDescription')"
       />
       <KRadioButton
         v-model="source"
         :label="$tr('localNetworkOrInternet')"
         :value="ContentSources.PEER_KOLIBRI_SERVER"
         :disabled="formIsDisabled"
+        :description="$tr('networkDescription')"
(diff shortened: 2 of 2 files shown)


indirectlylit:
note: recently made a similar update in #4582

radinamatic:
> Download content directly from Learning Equality if you have access to the public internet

First concern here is the word `public`... As opposed to ...`private`, or to differentiate from `local` in the second option? Might be confusing. And the term `directly` to me suggests that other paths are somehow `indirect` which may introduce some confusion. Maybe drop both words altogether?

> Transfer content between two running instances of Kolibri or a mirror of Kolibri Studio

Even though technically true, it may be better to stick to the word `import` instead of `transfer`. It would also be better to avoid the word `instance`, might be to technical. Maybe:
`Import content from Kolibri running on another device, or from a mirror of Kolibri Studio`

The _another_ device does not need to be in the same local network, right? It could simply be reachable by IP?

> Copy content directly to the computer running Kolibri if you have a drive that has been pre-loaded with content

Again, let's stick to the verb `import` for consistency, and the avoid the term `directly`. I'd suggest simply: 
`Import from a drive that has been pre-loaded with content`

lauradanforth:
Agree with most of @radinamatic 's comments above. My thoughts and suggested changes, which incorporates her suggestions:

**Kolibri Studio (online) - Download content directly from Learning Equality if you have access to the public internet**
Proposed change: "Kolibri Studio (online) - Download content from Learning Equality's library if you are connected to the Internet."

**Local network or internet - Transfer content between two running instances of Kolibri or a mirror of Kolibri Studio**
Thoughts: What does mirror of Kolibri Studio mean? I would take this out as it probably adds further confusion.
Proposed change: "Import content from Kolibri running on another device connected to the same network."

**Attached drive or memory card - Copy content directly to the computer running Kolibri if you have a drive that has been pre-loaded with content**
Thoughts: The biggest area of confusion around this is that it can only be content channels (not say direct adding of PDFs) and around how to get the channels onto the drive.
Proposed change: "Import content channels from a drive. A channel must first be exported onto the drive from an installation of Kolibri with existing content." (this may be too wordy but I think it needs to get at the steps of import, then export, then import into another installation)

laurenlichtman:
I agree with Laura's edit re: Kolibri Studio (I like the use of the word "library" because we don't usually mention LE in Kolibri itself)

I also agree with Laura's comment re: mirror and was going to say the same. I could also envision potential i18n issues with that as well.

I also agree with Laura's last comment re: attached drive and eliminating the use of the word "computer" as it may not always be a computer. 

Great suggestions @lauradanforth!

radinamatic:
**1º** Agreed with @lauradanforth 's suggestion:

> Kolibri Studio (online) - Download content from Learning Equality's library if you are connected to the Internet

Although I would prefer us to maintain consistency and keep using `import` instead of `download`.

**2º** `Import content from Kolibri running on another device connected to the same network` is not completely accurate, because this option allows for both - import from Kolibri in the same network, _*and*_ from outside the local LAN, as long as the IP and/or domain is known: one could technically put our [kolibribeta.learningequality.org](kolibribeta.learningequality.org), or NITA's [e-learning.education.go.ug](e-learning.education.go.ug), for example. Maybe: 

`Import content from Kolibri running on another device, either in the same local network or on Internet`

And this still does not cover the option of having the _real_ Studio mirror somewhere, right @indirectlylit?

**3º** `Import content channels from a drive. A channel must first be exported onto the drive from an installation of Kolibri with existing content`: I understand Laura's thinking, but my concern here would be that we only, and for the first time, mention _channels_ here. In my head this could induce thinking "Is this the only way to import (full) channel...?" So either we drop the word _channel_ here, or add it to the first two options too:

`Import content channels from a drive. Channels must first be exported onto the drive from another Kolibri device with existing content` (and add _channel_ to the first two options)

`Import content from a drive. Content must first be exported onto the drive from another Kolibri device`

indirectlylit:
I'll let the team continue to iterate - I'm okay with what ever you guys decide. Quickly responding to two items:

* **Regarding the verb:** I had intentionally not used 'import' in my original proposal because it seemed redundant. I chose the verbs 'download', 'transfer', and 'copy' to try and convey extra nuances of what kind of import it is. (I'm totally fine with not using this strategy)
* **Regarding Studio mirrors:** I do think this is something that a subset of our users will encounter, but I'm not sure that it's phrased right, because people will not be using Studio mirrors for curation - just content distribution. So technically "content distribution server" or just "content mirror" would be more accurate than "Studio mirror"

lauradanforth:
Following @radinamatic 's comments, I agree with being consistent on use of the word channel. I think it's critical to have it in the external drive description, so I'd advocate for putting in all descriptions then.

So, taking Radina's other suggestions into consideration:
`Kolibri Studio (online) - Import content from Learning Equality's library if you are connected to the Internet`

`Local network or internet - Import content from Kolibri running on another device, either in the same local network or on Internet`

`Attached drive or memory card - Import content channels from a drive. Channels must first be exported onto the drive from another Kolibri device with existing content`

I'm still confused by what the mirror of Studio would mean here and still advocate for leaving it out

radinamatic:
Ok, after including _channels_ in the first two options, we have:

- Kolibri Studio (online)
Import content channels from Learning Equality's library if you are connected to the internet

- Local network or internet
Import content channels from Kolibri running on another device, either in the same local network or on the internet

- Attached drive or memory card
Import content channels from a drive. Channels must first be exported onto the drive from another Kolibri device with existing content

I de-capitalized `Internet` to be consistent.

We may skip the mention of Studio mirrors for now, and reevaluate introducing it further on, when we have actual user case(s).

indirectlylit:
awesome thanks @lauradanforth @radinamatic!

indirectlylit:
changes made - @DXCanas ready for final review & merge

DXCanas (approved):
WFM. Found a bug while testing, resolved here:
https://github.com/learningequality/kolibri/pull/4649

PR #4657, opened 2018-12-26
When a lesson is set to active, clears the masterylog

### LINKED ISSUE #4574: allow exercises to be re-mastered within the context of a lesson
### Observed behavior

* learner completes an exercise on their own
* coach assigns the exercise in a lesson
* learner is shown as already having completed the exercise in the lesson's progress report

### Expected behavior

learner should re-master the exercise in the context of the lesson in order to show progress

### User-facing consequences

progress through lessons is currently slightly misleading

### Context

0.11

jredrejo:
### Summary

  When Lesson changes from inactive to active:
   * All the summarylogs for the affected resources are reset
   * All attemptlogs for the affected summarylogs are deleted

### Reviewer guidance
1. Using a learner, do some exercise after browsing some of the channels. Logout the learner.
2. Create a lesson for a group where the previous learner is included, and add the previous exercise inside this lesson
3. Set the lesson to active
4. Log in using the same learner as in the point 1, go to the created lesson: all the exercises must not be marked as done.

### References
resolves #4574 

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated


indirectlylit:
Thanks @jredrejo!

Would you mind working with @radinamatic to add a Gherkin story for this?

kolibri/core/lessons/serializers.py
```diff
@@ -128,7 +132,29 @@ def update(self, instance, validated_data):
...
+                )
+            content_ids = [resource['content_id'] for resource in instance.resources]
+            summarylogs_ids = set(
+                ContentSummaryLog.objects.filter(content_id__in=content_ids)
+                .filter(user_id__in=user_ids)
+                .exclude(completion_timestamp__isnull=True)
+                .values_list('id', flat=True)
+            )
+            AttemptLog.objects.filter(masterylog__summarylog__in=summarylogs_ids).delete()
+            ContentSummaryLog.objects.filter(id__in=summarylogs_ids).update(completion_timestamp=None, progress=0.0)
```
rtibbles, line 157: This will work, just trying to think through how this interacts with the 'MasteryLog' model also. Will need to play around with this locally.

kolibri/core/lessons/serializers.py
```diff
@@ -128,7 +132,29 @@ def update(self, instance, validated_data):
...
+                    )
+                )
+            content_ids = [resource['content_id'] for resource in instance.resources]
+            summarylogs_ids = set(
+                ContentSummaryLog.objects.filter(content_id__in=content_ids)
+                .filter(user_id__in=user_ids)
+                .exclude(completion_timestamp__isnull=True)
+                .values_list('id', flat=True)
+            )
+            AttemptLog.objects.filter(masterylog__summarylog__in=summarylogs_ids).delete()
```
rtibbles, line 156: Hrm, we do not want to delete the existing attemptlogs - missed this before.

This is why bumping to a new masterylog is important, because then we focus on the attemptlogs for this particular mastery log, without wiping out previous user engagement.

kolibri/core/lessons/serializers.py
```diff
@@ -128,7 +132,29 @@ def update(self, instance, validated_data):
                 lesson_id=instance.id,
                 collection_id__in=(set(current_group_ids) - set(new_group_ids))
             ).delete()
-
+        if validated_data.get("is_active", False):
```
rtibbles, line 135: The other problem here is that toggling from 'active' to 'inactive' and then back to 'active' will clear any learner's engagement with the content also. Hrm.

rtibbles:
@jredrejo I chatted this through with @indirectlylit and I think it might not be possible to do this neatly in the shortcut way I had originally proposed, sorry.

Plan now is to postpone implementation of this until we can do better contextual information to allow it, and just not implement this for 0.12.

Will close, and then we can iterate on the implementation plan from there. Thanks for taking a stab at this!

PR #4658, opened 2018-12-26
Update attemptlog after 4 interactions in the same exercise

### LINKED ISSUE #4639: trigger 'needs help' notifications even if hints are not used
### Observed behavior

In the current implementation of 'needs help' notifications, we ignore consecutive unsuccessful attempts to solve questions within an exercise.

Instead, we look at consecutive unsuccessful `AttemptLog` objects, which don't directly map to a user's mental model of an "attempt" (mainly for DB write performance reasons). Only a single unsuccessful `AttemptLog` object is generated even when a learner unsuccessfully attempts to solve a question in an exercise multiple times in a row.

In practice, this means that we're relying on the presence and usage of hints in order to trigger 'needs help' notifications because users will never proceed to the next question and therefore never trigger consecutive `AttemptLog`s. If the learner cannot proceed past a question (the question doesn't have a hint or the learned doesn't use it), the coach will never be notified.

### Expected behavior

Currently our exercises tell people to "try again" on a questions in exercises, no matter how many times they get it wrong. This is the user mental model of an 'attempt', and I would propose that it's this model that should drive 'needs help' notifications to coaches.

### User-facing consequences

Coaches might not get notified that a learner needs help, even when they do

-----

cc @rtibbles @jredrejo @jonboiser

jredrejo:
### Summary
This is a small change to force the update of the attemptlog logger when an user is stalled in an exercise.
It would be nice to trigger it only if no hints are left, but that's not possible unless 
kolibri-exercise-perseus-plugin is modified to provide that info when triggering its `hintTaken` event.

### Reviewer guidance
Just force 4 failed answers in the same exercise, the attemptlog is patched and the notification should be created.

### References
resolves #4639 

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated


indirectlylit:
clever work-around!

rtibbles (approved):
This works, will want to monitor if this causes any issues for performance, but this is the right way to handle it!

PR #4663, opened 2018-12-29
models - doc updates

indirectlylit:
### Summary

 * Updates to documentation
   * clarify text and fix a few mistakes
   * fix nav bar, which previously had two links to the 'models' sections because it was embedded in the index
 * ~Renames `AttemptLog` to `ExerciseAttemptLog`~

### Reviewer guidance

Cleaning up the docs, partly as a way to make sure I understand them.

~Also renamed the `AttemptLog` model to be less vague.~ Is this even possible in a post-Morango world?

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/core/content/models.py
+++ b/kolibri/core/content/models.py
@@ -1,34 +1,35 @@
 """
-This is one of the Kolibri core components, the abstract layer of all contents.
-To access it, please use the public APIs in api.py
+These models are used in the databases of content that get imported from Studio.
 
-The ONLY public object is ContentNode
 
-Developer note: If you modify the schema here, it has implications for the content
-import pipeline, including Kolibri Studio where the imported content databases
-are created.
+*DEVELOPER WARNING regarding updates to these models*
 
-Currently, Kolibri Studio has a modified copy of this models.py for generating backwards compatible
-content databases. Changes here should be propagated to that copy in order to allow for
-generation of databases with the changed schema.
-TODO: (rtibbles) achieve this by abstract base models that are instantiated in both applications.
+If you modify the schema here, it has implications for the content import pipeline
+because other systems also read and write these databases.
 
-As such, if models or fields are added or removed, or their type is changed, the CONTENT_SCHEMA_VERSION
-value must be incremented, with an additional constant added for the new
-version. e.g. a new constant VERSION_3 = '3', should be added, and CONTENT_SCHEMA_VERSION set to
+Specifically, Kolibri Studio has a modified version of this models.py for generating
+backwards-compatible content databases. Changes made here should be propagated to Studio
+in order to allow for generation of databases with the updated schema.
+
+In order to track updates to models or fields, the CONTENT_SCHEMA_VERSION value must be
+incremented, with an additional constant added for the new version.
+E.g. a new constant VERSION_3 = '3', might be added, and CONTENT_SCHEMA_VERSION set to
 VERSION_3.
 
 In addition, the new constant should be added to the mappings dict in
 ./utils/channel_import.py with an appropriate ChannelImport class associated.
-The mappings dict is a dict of content schema versions, with an associated ChannelImport class
-that will allow proper import of that content database into the current content schema for Kolibri.
+This map associates content schema versions with associated ChannelImport classes,
+which allows proper importing of that content database into the current content schema
+for Kolibri.
 
 If the new schema requires inference of the field when it is missing from old databases
 (i.e. it does not have a default value, or cannot be null or blank), then all the
-ChannelImport classes for previous versions must be updated to infer this data from old databases.
+ChannelImport classes for previous versions must be updated to infer this data from old
+databases.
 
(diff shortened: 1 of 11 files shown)


radinamatic:
> Cleaning up the docs, partly as a way to make sure I understand them.

Oops, who let you in on the secret of why _I_ like writing docs? :blush:

rtibbles (changes requested):
Please revert the model name change, as this has database level implications for something that is just a matter of terminology.

indirectlylit:
reverted model name change and pushed a few more updates

>  this has database level implications

also vague ;-)

I'd like to better understand this. The current name made it harder for me to understand the model structure because it wasn't clear how it's distinct from `BaseAttemptLog`.

rtibbles:
It renames the database table when you rename the model. Hence it creating a database migration.

indirectlylit:
yeah I ran `makemigrations` and it created one as expected. I don't understand the issue?

indirectlylit:
Are you saying that migrations should be avoided unless absolutely necessary?

In general I think keeping names as accurate as possible is helpful for onboarding new devs. However if migrations have side effects that negatively affect users, that's a trade-off to consider. My understanding was that a table rename migration would be cheap?

rtibbles:
Normally it would not be a concern, but it is a syncable model, so we should avoid unnecessary migrations, yes.

indirectlylit:
Yeah I didn't change the `morango_model_name` field so I figure it was OK but with the disadvantage of slight inconsistency.

Anyways this particular issue is not important - just trying to do some drive-by cleanup. More important are the docs updates - want to make sure my changes correctly reflect reality.

Regarding our long-term ability to refactor models: I expect we'll eventually want more flexibility in order to avoid accumulating a lot of historical baggage like what happened with [MS Office 97-2003 formats](https://www.joelonsoftware.com/2008/02/19/why-are-the-microsoft-office-file-formats-so-complicated-and-some-workarounds/). Or alternatively if models really are 4eva, we'll want to favor names that may seem overly specific, and perhaps establish some naming conventions.

jamalex:
> Yeah I didn't change the morango_model_name field so I figure it was OK but with the disadvantage of slight inconsistency.

This should be the case, yeah -- won't affect syncing.

Main concern I would have is whether the migration might be slow on a large
table (and this is generally one of our largest tables), because I recall
something about certain things in sqlite involving creating a new table and
moving across, but that doesn't seem to be the case for a table rename,
from my quick search.

[DR - edited to remove email cruft]

kolibri/core/content/models.py
```diff
@@ -1,34 +1,35 @@
...
 
-Developer note: If you modify the schema here, it has implications for the content
-import pipeline, including Kolibri Studio where the imported content databases
-are created.
+*DEVELOPER WARNING regarding updates to these models*
 
-Currently, Kolibri Studio has a modified copy of this models.py for generating backwards compatible
-content databases. Changes here should be propagated to that copy in order to allow for
-generation of databases with the changed schema.
-TODO: (rtibbles) achieve this by abstract base models that are instantiated in both applications.
```
rtibbles, line 14: Did you create an issue in lieu of this TODO?

kolibri/core/logger/models.py
```diff
@@ -237,8 +243,8 @@ class Meta:
 
 class AttemptLog(BaseAttemptLog):
     """
-    This model provides a summary of a user's engagement within a particular interaction with an
-    item/question in an assessment
+    This model provides a summary of a user's interactions with a question in an exercise.
```
rtibbles: This is inaccurate, as soon we will have assessments that are not exercises.

rtibbles:
A couple more comments.

kolibri/core/logger/models.py
indirectlylit: That's what I was trying to clarify earlier when I renamed the classes. My understanding was that `AttemptLog` was specifically for exercises, and `BaseAttemptLog` is the base class?

kolibri/core/content/models.py
indirectlylit: https://github.com/learningequality/kolibri/issues/4702

kolibri/core/logger/models.py
rtibbles: `AttemptLog` is for any assessment attempt taken within the context of a `ContentNode`, `BaseAttemptLog` is the base class for `AttemptLog` and `ExamAttemptLog`. `ExamAttemptLog` is specifically for assessment attempts taken within the context of Quizzes (nee exams).

indirectlylit: got it. updated to

> This model provides a summary of a user's interactions with a question in a content node.
> (Think of it like a ContentNodeAttemptLog to distinguish it from ExamAttemptLog and BaseAttemptLog)

rtibbles (approved):

PR #4688, opened 2019-01-05
Fixes exercise detail view issues

### LINKED ISSUE #4610: Not able to move in the channel of coach tab.
### Observed behavior
Not able to move or click on anything in the channel of coach tab.
To understand this issue in more detail please watch attached video.

### Expected behavior
It allow user to click anywhere in the kolibri.

### Steps to reproduce
1. Solve same exercise using two/three learners.
2. Login with coach user and go to the coach tab.
3. Click on Channel tab.
4. Move to the channel where learner solve exercises.
5. click on learner name and click on every question.
6. Try  to click on back button.

### Context
 * Kolibri version : kolibri 0.11.0
 * Operating system: Ubuntu 14.04 
 * Browser : Chrome

### Video
[VID_20181214_192532 (1).mp4.zip](https://github.com/learningequality/kolibri/files/2680347/VID_20181214_192532.1.mp4.zip)

### LINKED ISSUE #4622: Not able to move in recent of coach tab.
### Observed behavior
Not able to move or click on anything in the recent of coach tab.

### Expected behavior
It must allow user to click anywhere in the kolibri.

### Steps to reproduce
1. Solve the exercise.
2. Login with coach.
3. go to the recent tab.
4. follow the path from which channel you solved exercise.
5. Click on exercise.
6. Check attempts.
7. click on other question and see you will not able to move anywhere in the recent tab.

### Context
 * Kolibri version : Kolibri 0.11.0
 * Operating system : ubuntu 14.04
 * Browser : chrome

### LINKED ISSUE #4624: Not able to move in learner of coach tab.
### Observed behavior
Not able to move or click on anything in the learner of coach tab.

### Expected behavior
It must allow user to click anywhere in the kolibri.

### Steps to reproduce
1. Solve the exercise.
2. Login with coach.
3. go to the learner tab.
4. follow the path from which channel you solved exercise.
5. Click on exercise.
6. Check attempts.
7. Click on other question and see you will not able to move anywhere in the recent tab.

### Context
 * Kolibri version : kolibri 0.11.0
 * Operating system : Ubuntu 14.04
 * Browser : Chrome

rtibbles:
### Summary
Fixes issues with exercise detail views.
Set channelId on state.

### Reviewer guidance
Does the exercise detail view function properly in all circumstances (learner, channel, recent, and lessons).

### References
Fixes #4610, closes #4622, and resolves #4624

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/plugins/coach/assets/src/modules/exerciseDetail/handlers.js
+++ b/kolibri/plugins/coach/assets/src/modules/exerciseDetail/handlers.js
@@ -73,6 +73,7 @@ export function showExerciseDetailView(
           interactionIndex,
           summaryLog: summaryLog[0],
           user,
+          channelId,
         };
       });
     },


kolibri/plugins/coach/assets/src/modules/coreCoach/actions.js
```diff
@@ -43,7 +43,7 @@ export function resetModuleState(store, { toRoute, fromRoute }) {
     return store.dispatch('examCreation/resetExamCreationState');
   }
   const moduleName = pageNameToModuleMap[fromRoute.name];
-  if (moduleName) {
+  if (moduleName && fromRoute.name !== toRoute.name) {
```
indirectlylit: > Don't reset state when navigating between instances of the same page.

can you elaborate on why? I would expect this to be necessary

rtibbles: Because we avoid setting core loading to true when we navigate between similar instances of the same page, this causes breakage as the state is cleared, and the component tries to read from nested state that is now undefined.

rtibbles: The alternative is to fully specify the nested state in the initial state (but will still cause odd flashes as the state is cleared and then waits to load).

rtibbles: (generally, it is also unnecessary on navigation from the same page to the same page, as we fully specify the state in the handler, and so any previous state is overwritten when the new state is committed).

kolibri/plugins/coach/assets/src/views/reports/LearnerExerciseDetailPage/index.vue
```diff
@@ -79,7 +95,7 @@
             },
           };
         }
-        return undefined;
+        return {};
```
indirectlylit: this will result in a different behavior because `undefined` is falsy and an empty object is truthy. was this intentional?

indirectlylit (approved):
confirmed that this addresses the problem shown in the video

I'm a bit uncomfortable with globally updating the `resetModuleState` action because it's used in other places, and I'm worried it will have unexpected effects if assumptions don't hold everywhere

kolibri/plugins/coach/assets/src/views/reports/LearnerExerciseDetailPage/index.vue
rtibbles: Yes, it prevents syntax errors from trying to read properties of undefined.

rtibbles:
Reverted to the minimal fix required, there is still a slight flash in the interface when switching between questions, but it's not broken, just visually perturbing.

Issue #4308, opened 2018-09-21
Error occurred during content import 


arky:
Perhaps harmless log messages during importing content from Kolibri studio. It would be good to know what is the root cause of this content corruption logs? 

Is it because of network latency (as am testing this from Phnom Penh, Cambodia)? Or you are doing sanity checks before the file is completely downloaded?  

Testing Kolibri 0.10.2 pip package.

PS: Did anyone notice the typo 'occured' ? 

```
 27%|##6       | 1318251/4891502 [00:04<0ERROR    An error occured during content import: File https://studio.learningequality.org/content/storage/f/9/f913f4f7e30d7ee5a6f084b15fafee96.png is corrupted.5.14it/s]
ERROR:kolibri.content.management.commands.importcontent:An error occured during content import: File https://studio.learningequality.org/content/storage/f/9/f913f4f7e30d7ee5a6f084b15fafee96.png is corrupted.
 59%|#####8    | 2874761/4891502 [00:08<ERROR    An error occured during content import: File https://studio.learningequality.org/content/storage/3/0/30b531e1e0774bb69cb73c31aee684cb.png is corrupted.22.36it/s]s]
ERROR:kolibri.content.management.commands.importcontent:An error occured during content import: File https://studio.learningequality.org/content/storage/3/0/30b531e1e0774bb69cb73c31aee684cb.png is corrupted.
/usr/local/lib/python2.7/dist-packages/kolibri/dist/tqdm/_monitor.py:89: TqdmSynchronisationWarning: Set changed size during iteration (see https://github.com/tqdm/tqdm/issues/481)
  TqdmSynchronisationWarning)
 67%|######7   | 3292225/4891502 [00:10<ERROR    An error occured during content import: File https://studio.learningequality.org/content/storage/a/8/a82353a706cc5f02f0bfeeb1f6647533.png is corrupted.73it/s]/s]
ERROR:kolibri.content.management.commands.importcontent:An error occured during content import: File https://studio.learningequality.org/content/storage/a/8/a82353a706cc5f02f0bfeeb1f6647533.png is corrupted.
 92%|#########2| 4518122/4891502 [00:13<ERROR    An error occured during content import: File https://studio.learningequality.org/content/storage/6/e/6ec1913bac09950fddcf27d6faa122e3.png is corrupted.3092.40it/s]
[...]

Issue #2648, opened 2017-11-21
check python dependencies for known security issues and upgrade any necessary


indirectlylit:
in preparation for 0.7, let's audit for security issues. 

Sub-set of #1597

indirectlylit:
cc @aronasorman @benjaoming results of safety check:

```
~/P/l/r/k2 (develop) $ 
safety check -r requirements/base.txt 
Warning: unpinned requirement 'cherrypy' found, unable to check.
Warning: unpinned requirement 'morango' found, unable to check.
safety report
checked 25 packages, using default DB
---
-> django, installed 1.9.7, affected >=1.9,<1.9.10, id 25718
-> django, installed 1.9.7, affected >=1.9,<1.9.11, id 33075
-> django, installed 1.9.7, affected >=1.9,<1.9.11, id 33076
-> django, installed 1.9.7, affected >=1.9,<1.9.11, id 25734
-> django, installed 1.9.7, affected >=1.9,<1.9.13, id 33302
-> django, installed 1.9.7, affected >=1.9,<1.9.8, id 25721
-> tqdm, installed 4.8.3, affected <4.11.2, id 34741
```

aronasorman:
```
Warning: unpinned requirement 'cherrypy' found, unable to check.
Warning: unpinned requirement 'morango' found, unable to check.
```
These two requirements use either specific version or commit zip files from Github, so they're technically pinned, just not using PyPI.

aronasorman:
They look like this:
```
https://github.com/cherrypy/cherrypy/archive/v6.2.0.zip#egg=cherrypy
https://github.com/learningequality/morango/archive/ef260d263c92a4d3689e9ae742e25b57b52656e2.zip#egg=morango
```

indirectlylit:
looks like cherrypy 6.2.0 is okay

benjaoming:
@aronasorman @indirectlylit CherryPy is VERY MUCH updated in https://github.com/learningequality/kolibri/pull/1719

Maybe we should make this a separate target then? I updated it and didn't find any issues... would be nice to stop relying on this super-old + pinned version...

PR #4718, opened 2019-01-11
remove tqdm dependency

indirectlylit:
### Summary

TQDM puts out new releases [very often](https://github.com/tqdm/tqdm/releases), and has in the past caused us issues with thread safety, spurious errors, android issues, security issues, and excess logging.

### Reviewer guidance

I don't think there is any strong reason to include this dependency, and I believe it's simpler to remove it entirely.

### References

Some past TQDM-related stuff:

* https://github.com/learningequality/kolibri/pull/4158
* https://github.com/learningequality/kolibri/issues/2648
* https://github.com/learningequality/kolibri/issues/3963
* https://github.com/learningequality/kolibri/issues/4308
* https://github.com/learningequality/kolibri/pull/2654

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/core/tasks/management/commands/base.py
+++ b/kolibri/core/tasks/management/commands/base.py
@@ -3,7 +3,10 @@ from collections import namedtuple
 
 from django.core.management.base import BaseCommand
 from iceqube.exceptions import UserCancelledError
-from tqdm import tqdm
+import logging
+
+logging.basicConfig(level=logging.INFO)
+
 
 Progress = namedtuple(
     'Progress',
@@ -30,18 +33,12 @@ class ProgressTracker():
         self.level = level
         self.update_callback = update_callback
 
-        # initialize the tqdm progress bar
-        self.progressbar = tqdm(total=total)
-
     def update_progress(self, increment=1, message="", extra_data=None):
-
-        self.progressbar.update(increment)
-
         self.progress += increment
-
         self.message = message
-
         self.extra_data = extra_data
+
+        logging.info('Progress: {:.0%} | {} | {}'.format(1.0 * self.progress / self.total, self.message, self.extra_data))
 
         if callable(self.update_callback):
             p = self.get_progress()
@@ -60,8 +57,7 @@ class ProgressTracker():
         return self.update_progress
 
     def __exit__(self, *exc_details):
-        if self.progressbar is not None:
-            self.progressbar.close()
+        pass
 
 
 class AsyncCommand(BaseCommand):
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -8,7 +8,6 @@ configobj==5.0.6
 docopt==0.6.2
 django-mptt==0.9.1
 djangorestframework-csv==2.1.0
-tqdm==4.24.0                     # progress bars
 requests==2.20.1
 cherrypy==13.0.1  # Temporarily pinning this until CherryPy stops depending on namespaced package, see #2971 # pyup: <13.1.0
 tempora<1.13  # derived no-upper-bounds dependency of cherrypy causing issues  # pyup: <1.13


rtibbles:
Yeah, also I found the way that it tries to dynamically update itself tends to cause weird things in the terminal.

kolibri/core/tasks/management/commands/base.py
```diff
@@ -3,7 +3,10 @@
 
 from django.core.management.base import BaseCommand
 from iceqube.exceptions import UserCancelledError
-from tqdm import tqdm
+import logging
```
rtibbles, line 6: Our standard way to import and configure logging is this:
```
import logging
```
Then reference `logger` for log methods.
Because the logger is not namespaced, it is not getting sent to any of our registered loggers in the settings, and I don't see any progress info in the terminal.

kolibri/core/tasks/management/commands/base.py
```diff
@@ -30,19 +33,13 @@ def __init__(self, total=100, level=0, update_callback=None):
...
-
-        self.progressbar.update(increment)
-
         self.progress += increment
-
         self.message = message
-
         self.extra_data = extra_data
 
+        logging.info('Progress: {:.0%} | {} | {}'.format(1.0 * self.progress / self.total, self.message, self.extra_data))
```
rtibbles, line 41: Would be good to condition the third element of the progress tracker on not being `None`.

rtibbles (changes requested):
Couple of small tweaks. Tested the suggested logger change and fixed behaviour.

kolibri/core/tasks/management/commands/base.py
rtibbles: Might also be good to set this as `debug` to avoid flooding people's log files.

rtibbles:
Note that some of our tests also explicitly check the progressbar (hence the test failure above).

kolibri/core/tasks/management/commands/base.py
benjaoming: Recently, I wondered about writing a pre-commit hook for this common error :)

kolibri/core/tasks/management/commands/base.py
benjaoming: We can also write this straight to sys.stderr and not do any linebreak? This overwrites the current line, such that we don't flood the command line.

What is the need for having it in the user log? If we log that a process has started, there will be a timestamp, and if it breaks underway, it's easy to assume what has happened through the traceback... in between progress updates don't add value to the log, it just makes the log less useful? Or is it just me?

indirectlylit: I'm happy to remove it entirely

kolibri/core/tasks/management/commands/base.py
indirectlylit: I copied this from elsewhere in the codebase :)

kolibri/core/tasks/management/commands/base.py
rtibbles: Yeah, just not outputting it seems fine as well.

rtibbles: (but maybe sys.stdout is a nice compromise)

lyw07: I think we also need to remove this line: https://github.com/learningequality/kolibri/blob/develop/kolibri/core/content/management/commands/importcontent.py#L225 in the import retry logic, since we are not having progressbar attribute anymore.

benjaoming: Here's a snippet that I wrote for KA Lite. So because it keeps overwriting the same line, you get the effect of a progress bar:

```python
while not done:
    done = do_something()
    fraction = get_progress()
    sys.stdout.write("\rProgress: [{}{}] ({}%)".format(
        "=" * (int(fraction * 100) / 2),
        " " * (50 - int(fraction * 100) / 2),
        int(fraction * 100),
    ))
```

indirectlylit:
cc @DXCanas want to take over?

kolibri/core/tasks/management/commands/base.py
DXCanas: Wait so stdout or stderr?

rtibbles: Can probably just use @benjaoming's snippet verbatim, so stdout looks fine.

indirectlylit: is this desired behavior?

does it behave strangely in a multi-threaded or multi-process environment with multiple things writing to the same terminal?

DXCanas: Carriage return kind of messes up log statements during dev, but I'll leave it? Alternatively, I can just add use a newline, though we'd be missing out on the in-place replacement.

DXCanas:
Superseded by #4158

kolibri/core/tasks/management/commands/base.py
benjaoming: 1. It depends on who you are targeting with the output: A user should see clean output with the progress bar uninterrupted, whilst a dev will understand that it can get a bit messy.
1. To mitigate the clutter in the developers/debuggers log output: Avoid updating the progress bar too often, make a callback that only fires once every other second or even less.

Issue #4571, opened 2018-12-08
upgrade notifications using pingback


indirectlylit:
### Observed behavior

Currently users have no way of knowing when new versions are available

### Expected behavior

[See spec](https://docs.google.com/document/d/1h5hyHttxTrZpQq7j0sPCBLNlcV7PcaJ5CrtlJvVZBdM/edit)

### User-facing consequences

Currently many users who actually have internet access might not know new versions are available

### Context

0.11

indirectlylit:
@ralphiee22 @mrpau-dev let's make sure that these notifications are part of our integration testing

@ralphiee22 would you be able to write gherkin stories?

indirectlylit:
this was built

PR #4731, opened 2019-01-15
Pingback Notification Feature

ralphiee22:
### Summary

This adds the APIs needed to query notification objects, as well as dismissed notification objects. When querying notification objects we exclude those already dismissed by the user. 
Also adds notification modal for superusers and admins. Valid notifications will show up one after another, as the user dismisses them.

### Reviewer guidance

1. From the command line, `kolibri manage ping --once --server http://nutritionfacts-staging.nanoapp.io`.
2. Log in as admin or superuser
3. See message!
Try hitting `ok` and navigating to other pages. Ensure it does not show up when navigating between pages, or until logging back in.
Do the same above for dismissing the notification, but ensure it never shows up again (unless logging in as other user).

### References

https://github.com/learningequality/kolibri/issues/4571
MVP upgrade notifications doc
https://docs.google.com/document/d/1h5hyHttxTrZpQq7j0sPCBLNlcV7PcaJ5CrtlJvVZBdM/edit?pli=1#

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated


kolibri/core/analytics/migrations/0001_initial.py
```diff
@@ -0,0 +1,39 @@
...
+from django.db import migrations
+from django.db import models
+
+
+class Migration(migrations.Migration):
+
+    initial = True
+
+    dependencies = [
+        migrations.swappable_dependency(settings.AUTH_USER_MODEL),
```
ralphiee22: I think this may need to be changed to `FacilityUser`.

kolibri/core/analytics/serializers.py
```diff
@@ -0,0 +1,18 @@
...
+    class Meta:
+        model = PingbackNotification
+        fields = ('id', 'version_range', 'timestamp', 'link_url', 'i18n')
+
+
+class PingbackNotificationDismissedSerializer(serializers.ModelSerializer):
+
+    class Meta:
+        model = PingbackNotificationDismissed
+        fields = ('user', 'notification')
```
indirectlylit, line 19: minor style note: I find separating the `serializers.py` and `api.py` into two files to be unhelpful indirection. They're tightly coupled so I'd recommend just moving these classes to `api.py`

kolibri/core/analytics/api_urls.py
```diff
@@ -0,0 +1,11 @@
+from rest_framework import routers
+
+from .api import PingbackNotificationDismissedViewSet
+from .api import PingbackNotificationViewSet
+
+router = routers.SimpleRouter()
+
+router.register(r'pingbacknotification', PingbackNotificationViewSet, base_name='pingbacknotification')
+router.register(r'pingbacknotificationdismissed', PingbackNotificationDismissedViewSet, base_name='pingbacknotificationdismissed')
```
indirectlylit, line 9: From a client API perspective, seems a little awkward to expose both DB tables. This seems like a server-side implementation detail.

Might be cleaner to have just a single endpoint rather than two:

* For a particular user, the notifications can be annotated with their dismissal status
* By default the notification collection can be filtered to only show non-dismissed items
* The user can dismiss a notification by setting the dismissal flag on a particular notification, which is mapped to the `PingbackNotificationDismissed` table in the serializer

indirectlylit: (cc @rtibbles this mirrors some comments you were making to me about how the logging API can be cleaned up by reducing the 1:1 model/API endpoint mapping we've traditionally used.)

kolibri/core/assets/src/views/CoreBase/index.vue
```diff
@@ -259,11 +274,44 @@
...
+        const notification = this.notifications[0];
+        // check if translated message is available for current language
+        if (notification.i18n[currentLanguage] !== undefined) {
+          languageCode = currentLanguage;
+        }
+        return {
+          id: notification.id,
+          title: notification.i18n[languageCode].title,
+          msg: notification.i18n[languageCode].msg,
+          linkText: notification.i18n[languageCode].linkText,
```
jamalex: now called `link_text` for consistency

kolibri/core/analytics/utils.py
```diff
@@ -169,7 +177,19 @@ def extract_channel_statistics(channel):
     }
 
 
-def cmp_semver(kolibri_version, version_range):
-    if version_range == '*':
-        return True
-    return semver.match(kolibri_version, version_range)
+@transaction.atomic
+def create_and_update_notifications(data, source):
+    if data.get('messages'):
```
jamalex: This means that if we don't receive any messages, we won't do anything. But in that case we'd want to mark all messages as inactive. So, probably best to just do `messages = data.get('messages', [])` and continue from there.

jamalex:
The following changes (and a sample message) have been posted to the nutritionfacts staging server:
https://github.com/learningequality/nutritionfacts/commit/b33e19400ae639cbcf2b2e9b312d37493eb1e566#diff-5b7513e7bc7d64d348fd8d3f2222b573

Can be tested using:
`kolibri manage ping --once --server http://nutritionfacts-staging.nanoapp.io`

indirectlylit:
merging in to make this easier to test

indirectlylit (approved):

PR #4733, opened 2019-01-15
WIP - coach data feature branch

indirectlylit:
### Summary

…

### Reviewer guidance

…

### References

…

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated


jredrejo:
@indirectlylit  
- added pagination
- added  `after` parameter to recover from id:
http://localhost:8000/coach/api/notifications/?collection_id=9da65157a8603788fd3db890d2035a9f&after=3

returns all the notifications with id >3 

http://localhost:8000/coach/api/notifications/?collection_id=9da65157a8603788fd3db890d2035a9f
returns all notifications for today.



@jonboiser serializer now returns also  1) user’s full name; 2) lesson title (if there is one); 3) resource title as you requested

kolibri/core/logger/test/test_api.py
```diff
@@ -133,101 +133,101 @@ def tearDown(self):
...
+#         self.assertEqual(len(response.data), expected_count)
+
+#     def test_learner_group_log_filtering(self):
+#         self.client.login(username=self.admin.username, password=DUMMY_PASSWORD, facility=self.facility)
+#         response = self.client.get(reverse('kolibri:core:contentsummarylog-list'), data={"learner_group": self.learner_group.id})
+#         expected_count = ContentSummaryLog.objects.filter(user__pk=self.user2.id).count()
+#         self.assertEqual(len(response.data), expected_count)
+
+#     def tearDown(self):
+#         self.client.logout()
```
indirectlylit: @jredrejo we usually recommend deleting code rather than commenting it out, unless you expect this to be very temporary

jredrejo: Yep, it's a temp change to try to reproduce this problem in Travis. I will restore it tomorrow

indirectlylit: coolio

kolibri/core/notifications/utils.py
```diff
@@ -0,0 +1,36 @@
...
+            return cache[key]
+        except KeyError:
+            ret = cache[key] = fun(*args, **kwargs)
+            return ret
+
+    def cache_clear():
+        """Clear cache."""
+        cache.clear()
+
+    cache = {}
```
rtibbles, line 34: Could perhaps modify this to use Django cache?

rtibbles (approved):
Nothing jumped out at me as egregious - let's get this in and iterate further!

kolibri/core/notifications/utils.py
jredrejo: @rtibbles if we don't plan to move from using LocMemCache to other kind as redis or memcache, yes, it would be equivalent. Please, confirm me if you prefer me to change it to use Django cache.

rtibbles: One nice thing about using the Django cache, even without using a different backend, is that we can let the cache expire regularly, and not risk having stale values stored in the memoize?

PR #4737, opened 2019-01-16
Hashi and content renderer cleanup

### LINKED ISSUE #4724: Hashi no hash-y
### Observed behavior
Currently we build hashi without a build hash. This means that in the future if we upgrade our Hashi functionality, this may not get cache busted.

### Expected behavior
New versions of Hashi should cache bust old versions!

### User-facing consequences
Weird, unpredictable Hashi behaviour, depending on previous use.

### LINKED ISSUE #4725: Not possible to 'complete' an HTML5 App
### Observed behavior
In introducing Hashi, the 15s insta-points was removed from HTML5 apps. Now impossible to register any progress.

### Expected behavior
What should progress look like for an arbitrary HTML5 app?

### User-facing consequences
Confusion?

rtibbles:
### Summary
Hashi did not have a build hash for each build. This left it vulnerable to never being upgraded.
This fixes this by adding a build has to the build, and then using that to calculate an ETag on the zipcontent endpoint, so that the ETag gets updated for html files fetched from the endpoint, based on the hashi tag. This will allow cache busting when the hashi version is updated.

In addition, this reinstate completion for HTML5 apps, with a 5 minute timer.

Further, it adds a minor bug fix to session progress logging, whereby we were not maxing out the session progress at `1`.

### Reviewer guidance
Does the HTML5 renderer still work as expected?
Given that I am doing a vary-on the `XHR` header, perhaps all of the etag business is unnecessary, because the resource will be reloaded as long as the last time something was loaded was with XHR if the current request was not with XHR. Thoughts?

### References
Fixes #4724, and closes #4725

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/core/assets/src/state/modules/core/actions.js
+++ b/kolibri/core/assets/src/state/modules/core/actions.js
@@ -562,7 +562,7 @@ export function updateProgress(store, { progressPercent, forceSave = false }) {
   /* Calculate progress based on progressPercent */
   // TODO rtibbles: Delegate this to the renderers?
   progressPercent = progressPercent || 0;
-  const sessionProgress = sessionLog.progress + progressPercent;
+  const sessionProgress = Math.min(1, sessionLog.progress + progressPercent);
   const summaryProgress = summaryLog.id
     ? Math.min(1, summaryLog.progress_before_current_session + sessionProgress)
     : 0;
--- a/kolibri/core/content/templates/content/hashi.html
+++ b/kolibri/core/content/templates/content/hashi.html
@@ -6,7 +6,7 @@
         <meta http-equiv="X-UA-Compatible" content="IE=Edge,chrome=1">
         <meta name="viewport" content="width=device-width, initial-scale=1">
         <meta name="google" content="notranslate">
-        <script src="{% static "content/hashiframe.js" %}"></script>
+        <script src="{% static hashi_path %}"></script>
     </head>
     <body>
         {% include "kolibri/loading_snippet.html" %}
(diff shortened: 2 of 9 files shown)


indirectlylit:
FYI tests are failing in this PR

kolibri/core/assets/src/state/modules/core/actions.js
```diff
@@ -519,7 +519,7 @@ export function updateProgress(store, { progressPercent, forceSave = false }) {
   /* Calculate progress based on progressPercent */
   // TODO rtibbles: Delegate this to the renderers?
   progressPercent = progressPercent || 0;
-  const sessionProgress = sessionLog.progress + progressPercent;
+  const sessionProgress = Math.min(1, sessionLog.progress + progressPercent);
```
rtibbles, line 565: Hrm, this should be `Math.max`, I think...

rtibbles: Oh no - this is right, nvm...

kollivier (approved):
Change itself looks good to go, but I think we really need to rethink the point allocation metric. Five minutes may be too long for some of the simpler simulations, where you can run through them within a couple minutes, and it can also be way too low for some other experiences.  We should seriously consider either allowing customization, or, my preference, just awarding points based on time spent with a max of 500 total to be earned.

ralphiee22:
I am getting completion of html 5 apps in PHET in around 1m30s. Is this expected?

PR #4741, opened 2019-01-17
work on routing of plan pages

indirectlylit:
Work on migrating older coach management/plan pages to the new URL routing structure and giving them independent core base components outside of the `CoachIndex`

cc @jonboiser


indirectlylit:
self-merging

PR #4749, opened 2019-01-18
work on quiz editing

indirectlylit:
cc @jonboiser


indirectlylit:
@jonboiser note that this PR changes the vuex state as we discussed so that maps and arrays are swapped (maps are now state and arrays are getters)

PR #4774, opened 2019-01-23
add Sentry clients

indirectlylit:
### Summary

Adds Sentry clients to both the front- and backend, off by default, and configurable using options.ini or environment variables.

### Reviewer guidance

Is this structure acceptable? @rtibbles suggested maybe using an external settings file - is that actually necessary?

### References

* Instructions on using Sentry: https://github.com/learningequality/kolibri/pull/4774/files#diff-1eed66c9b6a7aa504cc4be09853b801fR103
* Follow-up issue on migrating this functionality to an external plugin: #4775

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/core/assets/src/kolibri_app.js
+++ b/kolibri/core/assets/src/kolibri_app.js
@@ -3,23 +3,7 @@ import router from 'kolibri.coreVue.router';
 import Vue from 'kolibri.lib.vue';
 import store from 'kolibri.coreVue.vuex.store';
 import heartbeat from 'kolibri.heartbeat';
-import * as Sentry from '@sentry/browser';
-import logger from 'kolibri.lib.logging';
 import KolibriModule from 'kolibri_module';
-
-export const logging = logger.getLogger(__filename);
-
-if (global.sentryEnabled) {
-  logging.warn('Sentry error logging is enabled - this disables all local error reporting!');
-  logging.warn(
-    '(see https://github.com/vuejs/vue/issues/8433 and https://docs.sentry.io/platforms/javascript/vue/)'
-  );
-  Sentry.init({
-    dsn: `https://${global.sentryConfig.publicKey}@sentry.io/${global.sentryConfig.projectId}`,
-    release: __version,
-    integrations: [new Sentry.Integrations.Vue({ Vue })],
-  });
-}
 
 /*
  * A class for single page apps that control routing and vuex state.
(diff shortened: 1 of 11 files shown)


requirements/base.txt
```diff
@@ -30,3 +30,4 @@ python-dateutil==2.7.5
 ifcfg==0.17
 sqlalchemy==1.2.10
 user-agents==1.1.0
+sentry-sdk==0.6.9
```
rtibbles, line 33: I am a little ambivalent about bundling this into our released builds.

kolibri/core/assets/src/kolibri_app.js
```diff
@@ -3,8 +3,24 @@ import router from 'kolibri.coreVue.router';
 import Vue from 'kolibri.lib.vue';
 import store from 'kolibri.coreVue.vuex.store';
 import heartbeat from 'kolibri.heartbeat';
+import * as Sentry from '@sentry/browser';
+import logger from 'kolibri.lib.logging';
 import KolibriModule from 'kolibri_module';
 
+export const logging = logger.getLogger(__filename);
+
+if (global.sentryEnabled) {
```
rtibbles: Can avoid bundling this into the default frontend bundle by making a separate bundle, and conditionally loaded tag. Will still be able to access the global kolibri Vue through `kolibri.lib.vue`.

requirements/base.txt
indirectlylit: would it have to become a plugin?

indirectlylit: In general I'm a big fan of decoupling this and making it optional, but would need help implementing it as such

indirectlylit:
made two updates:

* sentry loads async on client-side
* simplified options just two variables

requirements/base.txt
jamalex: Could have it not be in the requirements file at all, and make the import conditional (try/except)? Then, install it specifically in the demo environment? However, that would need to be coupled with either a) using something other than the PEX for spinning up a demo server (as we want to do anyway), or b) enabling external Python paths for PEX (as we also want to do anyway)

indirectlylit:
follow-up issue: #4775

requirements/base.txt
indirectlylit: > However, that would need to be coupled with either a) using something other than the PEX for spinning up a demo server (as we want to do anyway), or b) enabling external Python paths for PEX (as we also want to do anyway)

yup - precisely what I wrote in #4775 :)

jamalex (approved):
Looks good to me. But agree with removing the dependency from requirements.txt, to avoid installer bloat.

PR #4785, opened 2019-01-24
Write a dynamic KOLIBRI_HOME/nginx.conf

benjaoming:
### Summary

This is meant to support the implementation in https://github.com/learningequality/kolibri-server/pull/7

 * It avoids calls to `kolibri` CLI itself, but it *is* depending on Kolibri having been spun up (happens during installation of the `kolibri` deb package if system service is activated)
 * It avoids defining Nginx configuration in a monolithic way and instead assumes control is outside (i.e. the `kolibri-server` package

Additional tasks:

 * [x] Build a test release .deb for `kolibri` based on this test branch (thanks Buildkite)
 * [ ] ....then build a test release `kolibri-server` .deb and try with this one

### Reviewer guidance

Which other things do we need from internal kolibri?

I also have a feeling that we will need a touch reload file for UWSGI. See: https://uwsgi-docs.readthedocs.io/en/latest/articles/TheArtOfGracefulReloading.html

### References

https://github.com/learningequality/kolibri-server/issues/1

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated


kolibri/utils/conf.py
```diff
@@ -160,3 +160,50 @@ def enable_default_plugins():
...
+            "    alias  {{ static_dir }};\n"
+            "}\n"
+            "location /content {\n"
+            "    alias  {{ content_dir }};\n"
+            "}"
+        ).format(
+            static_dir=static_root,
+            content_dir=get_content_dir_path(),
+        )
+        nginx_conf_file.write(configuration)
```
jredrejo, line 209: why not writing the PORT here too?

jredrejo: Also, to avoid a chicken-egg  problem, I think we should write in this file the path to KOLIBRI_HOME and place it in /etc/kolibri, not in KOLIBRI_HOME.
Currently, the kolibri package is assuming the KOLIBRI_HOME is always in ~/.kolibri. I am not sure this is always true.

benjaoming: > Also, to avoid a chicken-egg problem, I think we should write in this file the path to KOLIBRI_HOME and place it in /etc/kolibri, not in KOLIBRI_HOME.

This file is written by the `kolibri` process owned by the user with `KOLIBRI_HOME`. This user does not have permissions to write files in `/etc/`.

benjaoming: > why not writing the PORT here too?

I haven't considered this yet. The `listen` directive is quite sensitive because it doesn't allow to be specified several times AFAIK. If there is a way around it, could be to specify a parameter or variable for the rest of the configuration to pick up.

jredrejo: > This file is written by the kolibri process owned by the user with KOLIBRI_HOME. This user does not have permissions to write files in /etc/.

Yes, but there are always solutions to that /var/run or some other place with 777 permissions. In any case if we are sure the path is going to be always  ~/.kolibri, this is not needed. But if that's not the case we need a solution.

> I haven't considered this yet. The listen directive is quite sensitive because it doesn't allow to be specified several times AFAIK.

I don't get it, why is it going to be specified several times?  It should be only one, with the port Kolibri is setup to use.

benjaoming: > Yes, but there are always solutions to that /var/run or some other place with 777 permissions. In any case if we are sure the path is going to be always ~/.kolibri, this is not needed. But if that's not the case we need a solution.

The environment that launches `kolibri` knows and defines `KOLIBRI_HOME`, for instance by running `source /etc/default/kolibri`. This is already solved in the `kolibri-server` package: https://github.com/learningequality/kolibri-server/pull/4

benjaoming: > I don't get it, why is it going to be specified several times? It should be only one, with the port Kolibri is setup to use.

When I say that I haven't investigated something, I think it's best not to discuss details before actually investigating it. This was my way of saying why I didn't include it. You can probably come up with reasons why a configuration directive gets specified several times in a configuration stack.

benjaoming:
@jredrejo can you use the Buildkite asset .deb from this PR to test a version of `kolibri-server` that should integrate with this?

indirectlylit:
added 'work-in-progress' label. when it's ready for review, please update the checklist and labels

benjaoming:
I would propose to review this change as is: It creates a pattern for adding Nginx-configuration with contents generated from Kolibri, whilst not yet touching on some rather sensitive issues that I think we should discuss aside in separate issues:

1) How to reload Nginx
2) Whether the `listen` directive for Nginx can be set this way, and how we deprecate/rename/ignore the current `HTTP_PORT` or `KOLIBRI_LISTEN_PORT` settings in this regard.

jredrejo:
@benjaoming  @jamalex  I have included most of the code from this PR into the kolibri-server package, moving the nginx configuration to that package instead of being reponsability of the kolibri application.
https://github.com/learningequality/kolibri-server/blob/develop/kolibri_server_setup.py

I think we can close this PR as the functionality is done there.

benjaoming:
@jredrejo I like your implementation over there and thanks for copying in my code comments, too :)

One small concern that I would have about moving it is that I envision that we re-generate this file when Kolibri knows about configuration changes or needs to reload Nginx or Uwsgi. Can this be done by externalizing this code still?

jredrejo:
Nope, it will require restarting the daemon. It seems reasonable to be. Most of the services require restarting them after configuration changes are done.
It's not something that's done frequently.

PR #4791, opened 2019-01-25
Add CHERRYPY_START option to allow running Kolibri background services

rtibbles:
### Summary
For best compatibility with a uwsgi setup, add an option to disable running the cherrypy server on Kolibri start command.

### Reviewer guidance
Does everything still work as expected?

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated


jredrejo (changes requested):
@rtibbles I like the proposal and the code that implements it, but I would prefer not having it mixed with  renaming kolibri/utils/time.py → kolibri/utils/time_utils.py
I am not saying that change is not a good idea, it's just it 'taints' this PR and makes it harder to understand why 17 files have been changed for this purpose.

benjaoming:
Agree with @jredrejo, this is why I initially gave up on looking at it :)

benjaoming:
I have one question that I really really really need to ask. I have asked it before. It's an honest question.

We do this because we want an environment of multiple, disposable separate Kolibri worker processes to run without Iceqube. So running without Iceqube is a thing, but it almost solves itself because the entry point is then the wsgi application and not the CLI.

Here is the question:

Does Iceqube even support this? I know that it has a persistent queue. But does it expect jobs and results to be added and removed by separate processes?

(these are the problems that are solved by redis/rabbitmq + celery/channels).

jredrejo:
> Does Iceqube even support this? I know that it has a persistent queue. But does it expect jobs and results to be added and removed by separate processes?

yes, it does. 
I even created a python script that started Iceqube instead of Kolibri. It worked, but I discarded it because most of the commands we run in the queue require Django. Including all the needed libraries and configs was as heavy as just using Kolibri and required maintaining two codebases, so it didn't make sense.

Otoh, planning more features for the queue or deciding if Iceqube is or is not good enough for Kolibri is out of the scope of this PR. I think that is a different story that can be commented in the Icequebe repository or in a different issue.

benjaoming:
@jredrejo 
>> Does Iceqube even support this? I know that it has a persistent queue. But does it expect jobs and results to be added and removed by separate processes?
>
> yes, it does.

~I can't quite wrap my head around how it would work with the experiment that you have done?~

Sorry, okay, I understand what's going on and start recalling what was done in Iceqube by reading this part of the code: https://github.com/learningequality/iceqube/blob/master/src/iceqube/storage/backends/inmem.py ("inmem" or "in memory" is usually the opposite of "persistent" or "shared between processes")

rtibbles:
> I like the proposal and the code that implements it, but I would prefer not having it mixed with renaming kolibri/utils/time.py → kolibri/utils/time_utils.py
> I am not saying that change is not a good idea, it's just it 'taints' this PR and makes it harder to understand why 17 files have been changed for this purpose.

This was required in order to be able to import the Python `time` module into `kolibri/utils/server.py`, otherwise it was importing the local `time` module instead, and I couldn't execute `time.sleep`!

jredrejo (approved):

Issue #4790, opened 2019-01-25
Exercises missing when importing from any other peer apart from studio


intelliant01:
### Observed behavior

Trying to import a channel via peer sync i.e. ```Local network or internet``` source, all other content but exercises are being imported. Importing from studio or a thumbdrive then all of the exercises copy fine.

Tried with a few other channels as well which have exercises (For e.g. ```CK-12 > Math > Elementary Math > Grade 1 > Addition and Subtraction facts to 10```). So it is not just a KA issue.

Rather it seems that the exercises created using Perseus framework are importing but not the rest.

### Expected behavior

All exercises part of the topics imported should have been imported but they weren't. 

### User-facing consequences

Confusing results for all.

### Errors and logs

[snippet_kolibri.log](https://github.com/learningequality/kolibri/files/2795923/snippet_kolibri.log)

### Steps to reproduce

Select to import all contents from any of these channel topics via peer sync i.e. ```Local network or internet``` wherein the source has been seeded initially from the studio -
- ```Khan Academy (English) > Math >1st grade > Place value > Numbers 0 to 120```
- ```CK-12 > Math > Elementary Math > Grade 1 > Addition and Subtraction facts to 10```

After import match with source if all the exercises have been imported.

### Context

Kolibri 0.11.1 - Source on Linux, Firefox 64.0.2
Kolibri 0.11.1 - Destination on Windows 10, Firefox 64.0.2

rtibbles:
Thanks for the report, this seems very problematic!

rtibbles:
One thing I note in the log file is this line:
`ERROR 2019-01-25 16:52:06,943 importcontent An error occurred during content import: File http://192.168.4.34:8080/content/storage/7/3/735d4c10bac106a8a4196f8618207f6b.perseus is corrupted.`
Where an exercise file appears to be corrupted, wondering if this might be causing the issue (or perhaps we are wrongly flagging exercise files as corrupted).

intelliant01:
@rtibbles I had also noticed this. It is possible that this message is in error as I found most of the Perseus based exercises imported correctly but there is no way I can be conclusive on this.

rtibbles:
It's possible that the files are being transferred, but because of the error we might not be then making them as imported.

PR #4793, opened 2019-01-25
Check the file size on disk after the file transfer

lyw07:
### Summary

This PR is to fix the issue that the exercises are mistakenly considered as corrupted when transferring over local network. The issue behind that is we compare the size of the file over the network with the size in the database. Instead, we should compare the size of the file on the disk to the number in the database. Cherrypy compresses the files, so the size of the file over the network is smaller than the size in the database.

### Reviewer guidance

1. Check if we can successfully import exercises in `CK-12 > Math > Elementary Math > Grade 1 > Addition and Subtraction facts to 10` from local network.
2. Check that if a file on the external drive is corrupted, whether the file would be skipped. I tested this by removing a file and then creating a random file with the same name.

### References

https://github.com/learningequality/kolibri/issues/4790
https://community.learningequality.org/t/import-channel-content-from-network-location/985

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/core/content/management/commands/importcontent.py
+++ b/kolibri/core/content/management/commands/importcontent.py
@@ -197,15 +197,6 @@ class Command(AsyncCommand):
             original_progress = self.progresstrackers[0].get_progress()
 
             with filetransfer, self.start_progress(total=filetransfer.total_size) as file_dl_progress_update:
-                # If size of the source file is smaller than the the size
-                # indicated in the database, it's very likely that the source
-                # file is corrupted. Skip this file.
-                if filetransfer.total_size < f.file_size:
-                    e = "File {} is corrupted.".format(filetransfer.source)
-                    logger.error("An error occurred during content import: {}".format(e))
-                    overall_progress_update(f.file_size)
-                    return True, 1
-
                 for chunk in filetransfer:
                     if self.is_cancelled():
                         filetransfer.cancel()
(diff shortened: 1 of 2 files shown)


kolibri/core/content/management/commands/importcontent.py
```diff
@@ -197,22 +197,25 @@ def _start_file_transfer(self, f, filetransfer, overall_progress_update):
...
                     file_dl_progress_update(length)
+
+                # If size of the destination file is smaller than the size
+                # indicated in the database, it's very likely that the source
+                # file is corrupted. Skip importing this file.
+                dest_file_size_on_disk = os.path.getsize(filetransfer.dest)
+                if dest_file_size_on_disk < f.file_size:
+                    e = "File {} is corrupted.".format(filetransfer.source)
+                    logger.error("An error occurred during content import: {}".format(e))
+                    os.remove(filetransfer.dest)
```
rtibbles, line 215: Could we add an additional test to guarantee the cleanup behaviour also?

lyw07: Yes I will do that!

lyw07: I didn't add an additional test but test the behavior in an existing test for corrupted file. Please feel free to let me know if you think I need additional test only for the cleanup behavior. Thank you!

rtibbles: Sounds good!

kolibri/core/content/management/commands/importcontent.py
```diff
@@ -197,22 +197,25 @@ def _start_file_transfer(self, f, filetransfer, overall_progress_update):
...
+                    overall_progress_update(f.file_size - dest_file_size_on_disk)
```
rtibbles, line 216: I feel like we need to make this correction to both progress processes, if the wire size is smaller than the size on disk?

PR #4800, opened 2019-01-26
Move all URL setting on kolibriGlobal object to initialization.

rtibbles:
### Summary
We now use a dynamic webpack public path, but this only got set after urls had been loaded from the backend.
This PR puts all URL reversal into the template prior to default frontend initialization, and moves all URL setting into the entry point of the default frontend bundle.
This means that no asynchronous webpack chunk loading can happen prior to the webpack public url being properly set.

### Reviewer guidance
Does our Sentry integration load its chunk properly in production mode?
Do static images still load properly?

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/core/assets/src/core-app/index.js
+++ b/kolibri/core/assets/src/core-app/index.js
@@ -1,3 +1,12 @@
+// Shim Array includes for ES7 spec compliance
+require('array-includes').shim();
+// polyfill for older browsers
+// TODO: rtibbles whittle down these polyfills to only what is needed for the application
+require('core-js');
+
+// Do this before any async imports to ensure that public paths
+// are set correctly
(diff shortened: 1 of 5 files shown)


indirectlylit (commented):
I can't tell if this is user error, but I'm having problems with this. As far as I can tell, `global.sentryDSN` is not defined here:

https://github.com/learningequality/kolibri/pull/4800/files#diff-e46818b93613a96345b99fe7bbe04beaR35

even though it's been set properly on the backend

FYI to test, you can easily create a dummy Sentry project and set the DSN

indirectlylit (approved):
merging in because nothing seems 'more broken' - will continue to test Sentry

PR #4802, opened 2019-01-27
accessible drag and drop, plus improved keyboard focus rings

### LINKED ISSUE #4764: make list sorting keyboard accessible
### Observed behavior

follow-up from #4716

### Expected behavior

…

### User-facing consequences

…

### Errors and logs

…

### Steps to reproduce

…

### Context

…

indirectlylit:
### Summary

* makes drag and drop keyboard-accessible
* improves consistency and visibility of focus rings, particularly on buttons and links

| focus styling before | focus styling after |
|--|--|
| outline-old |  outline |

Rearranging with keyboard:

rearrange

### Reviewer guidance

tab around, make sure things work

### References

fixes #4764

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated


radinamatic:
1. While navigating with keyboard through the main tabs on the Coach page, in Firefox both the list item and tab title get separate focus:

tab-ff

That does not happen in Chrome, nor does it happen with Lessons, Quizzes, Groups and Learner sub-tabs. The only difference I see in FF Accessibility properties is the presence of 2 sections inside list items in former and only one in latter:

kolibri - mozilla firefox_002

2. The above leads to issue that the [`aria-label` for the tab svg icon](https://github.com/learningequality/kolibri/pull/4802/files#diff-110059d230b5b8580be3842e82c79648R10) here is not necessary and repetitive, as it gets read from the tab title itself - see how the element's accessible name resolves into `Class Home CLASS HOME`.

3. Visual focus ring is a bit too thin and light for my taste, is it possible to make it wider and/or maybe darker?

4. Keyboard navigation works well, resource moves down or up with both <kbd>Enter</kbd> and <kbd>spacebar</kbd> keys, accessible name is appropriate, and the only qualm I would have is with the focus ring that is confusing for the resources that can be moved both up _and_ down. It does not seem sufficiently clear to me which element is actually in focus, as the ring kind of engulfs or overlaps all of them:

contar 4 - kolibri - mozilla firefox_001

The `grip` icon is actually unnecessary when navigating by keyboard - is it possible to detect that the user used the <kbd>Tab</kbd> key to focus, and render the `grip` icon invisible until either up or down arrows are in focus?

khangmach (approved):

radinamatic (dismissed):
I'd like for us to solve the points 1. and 2. [from my comment](https://github.com/learningequality/kolibri/pull/4802#issuecomment-457980468) before merging, but I would not consider 3. and 4. blockers, and those can be slated to 0.12.1.

indirectlylit:
I was never able to replicate exactly what is shown in the screenshot, but I made the following changes:

* explicitly removed text from tab order
* removed aria label from SVG
* removed code which may have been hiding the outline on the tabs
* increased the width of the outlines to 3px and darkened it a bit
* on small icon buttons and tabs, made the outline smaller
* made the grip even dimmer when using keyboard

radinamatic:
Re-tested, 1, 2, 3 fixed, merge away! 

kolibri - mozilla firefox_010

rtibbles:
Note that some frontend tests are currently failing.

PR #4808, opened 2019-01-28
Handle adding large topics to exams

### LINKED ISSUE #4740: exam creation breaks when clicking 'select all' on a large topic
### Observed behavior

Just filing this is a note for later to investigate further. I'm currently experiencing this in our `coach-data` working branch. Haven't tested in `develop` yet.

Clicking 'select all' on KA Math when creating an exam does two things:

* browser freezes and CPU goes to 100% for a longer period than I've been willing to wait
* server reports a 'URI string too long' type error

I briefly looked at the request (but forget which endpoint) and it indeed had an array with thousands of items.

### Expected behavior

buttery smoothness

### User-facing consequences

:/

### Errors and logs

`414 Request-URI Too Large`

### Steps to reproduce

select-all in KA math

### Context

`coach-data` branch

rtibbles:
### Summary
Currently, adding large topics to quizzes causes two issues:
1) It makes a request with too many exercise ids in it to the `node_assessments` endpoint, resulting in a 411 error.
2) It adds new exercises to the data store using a `unionBy` operation, and then various methods for verifying if a topic should be checked or indeterminately checked using searches over arrays for membership.
3) Lastly, there were some issues in the backend code that caused these issues to arise for large channels, even when they were only partially imported.

This PR deals with these issues thusly:
1) Only make a request to the `node_assessments` endpoint if the number of selected exercises is less than the maximum number of questions in a quiz. This prevents the 411.
2) The first issue is troublesome, but the second was the major contributor to the browser freezing. This is resolved by using an object rather than an array to store selected exercises - this allows uniquely updating to be done much more efficiently, and for any methods not to have to do array searches over potentially very long arrays. All references to `selectedExercises` should now be updated to reflect that it is now an `id: entity` object store.
3) Ensures `node_assessments` and `descendant_assessments` endpoints both filter by availability.

### Reviewer guidance
Ensure that the entire quiz creation work flow works as expected, including adding and removing large topics.

Did I catch every reference to `selectedExercises`?

### References
Fixes #4740

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/plugins/coach/assets/src/constants/examConstants.js
+++ b/kolibri/plugins/coach/assets/src/constants/examConstants.js
@@ -1,3 +1,5 @@
 export const Modals = {
   PREVIEW_EXAM: 'PREVIEW_EXAM',
 };
+
+export const MAX_QUESTIONS = 50;
--- a/kolibri/plugins/coach/assets/src/modules/examCreation/actions.js
+++ b/kolibri/plugins/coach/assets/src/modules/examCreation/actions.js
@@ -187,7 +187,7 @@ export function filterAndAnnotateContentList(childNodes) {
 }
 
 export function updateSelectedQuestions(store) {
-  if (!store.state.selectedExercises.length) {
+  if (!Object.keys(store.state.selectedExercises).length) {
     store.commit('SET_SELECTED_QUESTIONS', []);
     return Promise.resolve();
   }
@@ -196,7 +196,7 @@ export function updateSelectedQuestions(store) {
     // If there are more exercises than questions, no need to fetch them all so
     // choose N at random where N is the the number of questions.
     const exerciseIds = shuffled(
-      uniq(store.state.selectedExercises.map(exercise => exercise.id)),
+      Object.keys(store.state.selectedExercises),
       store.state.seed
     ).slice(0, store.state.numberOfQuestions);
 
--- a/kolibri/plugins/coach/assets/src/modules/examCreation/actions.js
+++ b/kolibri/plugins/coach/assets/src/modules/examCreation/actions.js
@@ -10,6 +10,7 @@ import { getContentNodeThumbnail } from 'kolibri.utils.contentNode';
 import router from 'kolibri.coreVue.router';
 import { ContentNodeKinds } from 'kolibri.coreVue.vuex.constants';
 import { PageNames } from '../../constants';
+import { MAX_QUESTIONS } from '../../constants/examConstants';
 import { createExam } from '../examShared/exams';
 import selectQuestions from './selectQuestions';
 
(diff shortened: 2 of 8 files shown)


kolibri/plugins/coach/assets/src/constants/examConstants.js
```diff
@@ -1,3 +1,5 @@
 export const Modals = {
   PREVIEW_EXAM: 'PREVIEW_EXAM',
 };
+
+export const MAX_QUESTIONS = 50;
```
indirectlylit, line 5: this can be re-used in a couple more places:

* https://github.com/learningequality/kolibri/blob/develop/kolibri/plugins/coach/assets/src/views/plan/CreateExamPage/index.vue#L242
* https://github.com/learningequality/kolibri/blob/develop/kolibri/plugins/coach/assets/src/views/plan/CreateExamPage/CreateExamPreview.vue#L264

kolibri/plugins/coach/assets/src/modules/examCreation/actions.js
```diff
@@ -31,19 +32,20 @@ export function removeFromSelectedExercises(store, exercises) {
...
-
 export function updateAvailableQuestions(store) {
   const { selectedExercises } = store.state;
+  // Only bother checking this if there is any doubt that we have sufficient
+  // questions available. If we have selected more exercises than we allow questions
+  // then we are sure to have this.
   if (selectedExercises.length > 0) {
-    return ContentNodeResource.fetchNodeAssessments(selectedExercises.map(ex => ex.id)).then(
-      resp => {
+    if (MAX_QUESTIONS > Object.keys(selectedExercises).length) {
```
indirectlylit, line 41: makes sense

indirectlylit (approved):
tried basic quiz creation, and works for me

one small change requested and then good to merge

indirectlylit:
>  using an object rather than an array to store selected exercises

Yes, I'm finding in coach reports that objects of keys are a more natural state representation than arrays.

The result coming back from the server is an array, but the first thing I do is create a map: https://github.com/learningequality/kolibri/blob/develop/kolibri/plugins/coach/assets/src/modules/classSummary/index.js#L159

Then I provide vuex getters to return the array form as necessary: https://github.com/learningequality/kolibri/blob/develop/kolibri/plugins/coach/assets/src/modules/classSummary/index.js#L95

kolibri/plugins/coach/assets/src/constants/examConstants.js
rtibbles: I have already updated it in the former, changing the latter.

Issue #4329, opened 2018-09-26
loading-spinner.gif is large and poorly exported


indirectlylit:
### Observed behavior

Our loading spinner is 100KB. It's also poorly exported:

* the outer edges are aliased against the transparent background
* the inner edges are blurry, as though scaled from pixels rather then exported from vectors, which works against the compression
* each frame of the gif re-draws all previous frames, wasting bytes

![loading-spinner](https://user-images.githubusercontent.com/2367265/46098655-c4195c00-c179-11e8-9e52-6946c9c58ae2.gif)

### Expected behavior

The loader should either be a well-exported and optimized gif, or it should be done as an inline SVG.

[The original is here](https://drive.google.com/file/d/0B9ZzDms8cSNgRnlSbzJjOWxGMzQ/view)

### User-facing consequences

ironically slow-loading loading indicator

### Context

0.11

MisRob:
Hi!

I've been watching you for some time, I appreciate what you do :)! I checked developer docs and now I would like to pick a couple of smaller issues to get acquainted with the project and start contributing eventually when I have some free time - can I assign myself to this one and to some other free issues after this one is done ? Thanks.

PR #4833, opened 2019-01-29
Optimize loader

MisRob:
#### Description
Replace poorly exported gif loader with inline SVG (optimized with [svgo](https://github.com/svg/svgo)) + css animation.

#### Manual checks
 - locally - Safari, Chrome, Firefox (latest)
- Browserstack - IE9 (NOK), IE10, IE11, Edge

IE9 doesn't support CSS animations => loader is not working properly there (looks simply like static image). In [dev docs](https://kolibri-dev.readthedocs.io/en/develop/manual_testing.html#cross-browser-and-os-testing) I noticed the following browser support criteria:
> IE 9+ on Windows XP and up

On the other hand, I needed to temporarily remove `supported_browser` context processor to be able to test the behaviour in IE9 and IE10. So I am not sure - should I prepare fallback when the user will see "Unsupported browser" message in IE9 anyway? Or should the doc be updated? It's my first attempt to contribute to Kolibri so any guidance on browsers support will be appreciated.

#### Before

loading-spinner

#### After

kolibri-loader

#### How to test the loader

Probably not the smartest way but I was able to preview the loader in the app after commenting out [FE bundle import](https://github.com/learningequality/kolibri/blob/develop/kolibri/core/templates/kolibri/base.html#L37)
(EDIT: I realized network throttling could be more useful for review :)

#### What to pay attention to

The animation is now smoother since I could use the full power of CSS animations. Though, I am not sure if that's OK - should I imitate the original gif more accurately or do you have any other preferences (timing adjustments..)?

### References

#4329 

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated


MisRob:
Hello, it's my first contribution - I went through contributor checklist but it seems I am not authorized to add a label or a reviewer so I'm leaving it like it is now. I suppose only core team members can do that or have I missed something? Thanks :)

indirectlylit:
this is great! will review asap

khangmach (approved):

jamalex:
Awesome, thanks for the contribution!!

MisRob:
My honour. I appreciate what you've managed with Kolibri and KALite. I've been planning to check codes for a while :)

Now I can see coverage is a bit lower. I am not 100% sure since I don't know the codebase, but issues don't seem to be related. Is it all right?

(I was considering adding a test but to be honest I wasn't sure how to test the loader in base template - seems there's a lot of async stuff happening in the background. If I would want to test the template only, it seems it would require mocking Webpack hooks somehow and I am not sure if such test would be useful in this case. Or maybe some e2e test could help (though that would probably require some throttling)? I checked integration tests briefly so I am not sure if something like this would be suitable. So if you have any recommendations :)

rtibbles:
> Now I can see coverage is a bit lower. I am not 100% sure since I don't know the codebase, but issues don't seem to be related. Is it all right?

This is fine - this is mostly related to the fact that the SVG is many more lines than the IMG tag that you replaced. We don't currently have a good mechanism for testing your change, so I wouldn't expect you to come up with a test!

rtibbles (approved):
Code looks good to me!

MisRob:
OK thanks for the explanation

PR #4837, opened 2019-01-29
Add needs help flagging to lesson content status.

### LINKED ISSUE #4817: 'help needed' is not currently being reported by the class_summary_api for content
### Observed behavior

Haven't implemented yet:

https://github.com/learningequality/kolibri/blob/develop/kolibri/plugins/coach/class_summary_api.py#L27-L34

### Expected behavior

This query runs fairly often, so we need to be careful about performance. Currently 'needs help' would require a join with the attempt logs, which would be run for every single content item in every lesson in the class.

I'm thinking it would be preferable to add a `help_needed` field to the summary log to make this cheaper. The basic idea would be to update the `help_needed` flag at the same time as notifications are generated.

### User-facing consequences

currently 'needs help' is not reported in our reports

### Context

0.12

----

cc @jredrejo @rtibbles I could use a little help with this if either of you have time

rtibbles:
### Summary
Adds needs help flagging to our summary coach report API endpoint.

### Reviewer guidance
Check that when needs help is flagged, it shows in the return.
Check that when a learner then completes, it shows as completed instead.

### References
Fixes #4817

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/plugins/coach/class_summary_api.py
+++ b/kolibri/plugins/coach/class_summary_api.py
@@ -1,16 +1,17 @@
+from django.db.models import Max
 from django.db.models import Sum
-from rest_framework import viewsets
+from django.shortcuts import get_object_or_404
 from rest_framework import serializers
+from rest_framework import viewsets
 from rest_framework.response import Response
 
 from kolibri.core.auth import models as auth_models
-from kolibri.core.logger import models as logger_models
 from kolibri.core.content.models import ContentNode
-from kolibri.core.lessons.models import Lesson
 from kolibri.core.exams.models import Exam
-
-
-from django.db.models import Max
+from kolibri.core.lessons.models import Lesson
+from kolibri.core.logger import models as logger_models
+from kolibri.core.notifications.models import LearnerProgressNotification
+from kolibri.core.notifications.models import NotificationEventType
 
 
 NOT_STARTED = "not_started"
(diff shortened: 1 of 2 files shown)


kolibri/plugins/coach/class_summary_api.py
```diff
@@ -145,7 +224,7 @@ def data(Serializer, queryset):
 
 class ClassSummaryViewSet(viewsets.ViewSet):
     def retrieve(self, request, pk):
-        classroom = auth_models.Classroom.objects.filter(id=pk).get()
+        classroom = get_object_or_404(auth_models.Classroom, id=pk)
```
indirectlylit, line 208: ah, cool

kolibri/plugins/coach/class_summary_api.py
```diff
@@ -19,23 +20,82 @@
...
+    # existence of this key in the set in order to see whether this user has been flagged as needing
+    # help.
+    lookup_key = "{user_id}-{node_id}"
+    needs_help = {
+        lookup_key.format(user_id=n[0], node_id=n[1]): n[2] for n in LearnerProgressNotification.objects.filter(
+            classroom_id=classroom.id,
+            notification_event=NotificationEventType.Help,
+            lesson_id__in=[lesson["id"] for lesson in lesson_data],
+        ).values_list("user_id", "contentnode_id", "timestamp")
+    }
```
indirectlylit, line 51: These dictionary comprehensions and tuple lookups obfuscate the intended logic. Some intermediary variables would help self-document the code:

```python
learner_tuples = LearnerProgressNotification.objects.filter(
    classroom_id=classroom.id,
    notification_event=NotificationEventType.Help,
    lesson_id__in=[lesson["id"] for lesson in lesson_data],
).values_list("user_id", "contentnode_id", "timestamp")

needs_help = {}
for (user_id, contentnode_id, timestamp) in learner_tuples:
    needs_help["{}-{}".format(user_id, contentnode_id)] = timestamp
```

---

Side note - I like the strategy of generating a new key. In vuex I used nested objects, but this seems cleaner

indirectlylit (approved):
sweet

PR #4844, opened 2019-01-31
adding tests for data helpers

indirectlylit:
(replaces #4843)

### Summary

This adds tests for most of the data helpers which help transform base state into common intermediary patterns.

@jonboiser this also touches but doesn't really change those `STATUSES` constants you'd mentioned to me.

### Reviewer guidance

I skipped a few functions for now, but they follow the same structure as some others. Can finish them later

### References

…

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated


kolibri/plugins/coach/assets/src/modules/classSummary/__test__/dataHelpers.spec.js
```diff
@@ -0,0 +1,109 @@
+import Vuex from 'vuex';
+import summaryModule from '../index';
+
```
rtibbles, line 3: Might be worth using the status constants here, rather than strings to make the tests less brittle?

rtibbles (approved):
No red flags for me, one suggestion, but can be cleaned up later.

kolibri/plugins/coach/assets/src/modules/classSummary/__test__/dataHelpers.spec.js
indirectlylit: interesting

the expected values are tightly coupled to the sample data - could you elaborate on how constants would decrease brittleness?

PR #4903, opened 2019-02-02
Fixes #4898: Return a status of STARTED as soon as user has opened exam

### LINKED ISSUE #4898: When starting Exams, Notifications and Attempt logs can be out of sync
### Observed behavior

1. As a learner, start an exam, but do not do any of the questions
1. After some time, a Notification will be recorded for the exam with a status of "Started"
1. However, querying the class summary API, the 'exam learner status' for that exam will have a status of "NotStarted".

This can lead to an inconsistency, where the Notification will update the in-memory classSummary data and cause the dashboard and reports to show that 1 learner has "started" an exam, but if you were to refresh (and get the on-server class summary data without the updating notification), it will revert to showing 0 learners starting the exam.

### Expected behavior

Since, the UI intends to use Notifications to patch class summary in real time, the two notions of "Exam Started" should match to avoid situations like the on described above.

### User-facing consequences

inconsistent / fluctuating values in reports

### Steps to reproduce

…

### Context

0.12.0 a 7

jamalex:
### Summary

Per https://github.com/learningequality/kolibri/issues/4898, we previously only marked an exam as started after the first answer had been submitted (`ExamAttemptLog` exists). With this PR, we mark an exam as started as soon as it's opened (based on whether an `ExamLog` exists).

After:
image

### Reviewer guidance

In one browser, watch exam report. In another, log in as learner and open exam. In first browser, note that user now shows as having started.

### References

Fixes #4898.

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/plugins/coach/class_summary_api.py
+++ b/kolibri/plugins/coach/class_summary_api.py
@@ -108,9 +108,8 @@ class ExamStatusSerializer(serializers.ModelSerializer):
     def get_status(self, exam_log):
         if exam_log.closed:
             return COMPLETED
-        elif exam_log.attemptlogs.values_list("item").count() > 0:
+        else:
             return STARTED
-        return NOT_STARTED
 
     def get_num_correct(self, exam_log):
         return (


indirectlylit (approved):
deleted code is the best code

kolibri/plugins/coach/class_summary_api.py
```diff
@@ -108,9 +108,8 @@ class ExamStatusSerializer(serializers.ModelSerializer):
     def get_status(self, exam_log):
         if exam_log.closed:
             return COMPLETED
-        elif exam_log.attemptlogs.values_list("item").count() > 0:
+        else:
```
indirectlylit, line 111: in this case I would typically remove the `else` clause completely

PR #4906, opened 2019-02-02
Ensure the snackbar does not reappear after the resource page is left

### LINKED ISSUE #4848: spurious 'removed 0 resources' notification
### Observed behavior

observed by @jtamiace during 0.12.0 alpha 7 testing:

> image

### Expected behavior

Notifications should not be displayed until the user has made a change

### User-facing consequences

possible confusion

### Errors and logs

…

### Steps to reproduce

…

### Context

0.12.0 alpha 5

jredrejo:
### Summary
Ensures the snackbar does not appear after the page is left in Lessons and Quiz plans.

### Reviewer guidance
Instruccions to reproduce the issue are at #4848 
Using this PR code the snackbar will disappear after leaving the resource that has been added.

### References
Closes #4848 

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/plugins/coach/assets/src/views/plan/PlanLessonSelectionContentPreview.vue
+++ b/kolibri/plugins/coach/assets/src/views/plan/PlanLessonSelectionContentPreview.vue
@@ -50,8 +50,11 @@
         return false;
       },
     },
+    beforeDestroy() {
+      this.clearSnackbar();
+    },
     methods: {
-      ...mapActions(['createSnackbar']),
+      ...mapActions(['createSnackbar', 'clearSnackbar']),
       ...mapActions('lessonSummary', ['addToResourceCache']),
       handleAddResource(content) {
         this.$store.commit('lessonSummary/ADD_TO_WORKING_RESOURCES', content.id);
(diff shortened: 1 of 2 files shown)


kolibri/plugins/coach/assets/src/views/plan/PlanLessonSelectionContentPreview.vue
```diff
@@ -50,8 +50,11 @@
         return false;
       },
     },
+    beforeDestroy() {
+      this.clearSnackbar();
+    },
```
indirectlylit, line 55: I think this should be okay here because `clearSnackbar` is a core action.

However just a heads-up that (I believe) triggering actions and other logic from `beforeDestroy` hooks is an anti-pattern, because by the time the component is getting destroyed, app vuex state is generally unpredictable. In general I recommend that `beforeDestroy` only gets used for cleanup like removing event listeners and tearing down third-party libraries (like video.js).

Sometimes it's hard to find a reasonable alternative to `beforeDestroy`, though.

jredrejo: Yes, I also dislike the moment this action is done, but what other way can we detect the person is leaving this component?
Anyway, this is a workaround, the main problem here is the SnackBar is being re-mounted whenever any of these components are destroyed. It's the mount event in the snackbar what's restarting the timeout. I haven't been able to find what is triggering the `mounted` event once again, lost in the internal vue & vuex multiple internal hooks and triggers. @indirectlylit do you have any idea of why the SnackBar is mounted when moving through components? If that event is cached we could avoid using the `beforeDestroy` event.

rtibbles (approved):
Merging until we can find a better pattern!

PR #4914, opened 2019-02-04
Don't exit if database is locked while vacuuming

jredrejo:
### Summary
When running kolibri  under uwsgi, we are keeping Kolibri process running too , to be able to execute the background tasks.
This can drive to a race condition if Kolibri is executing vacuum: the database is locked, uwsgi dies and the users can't access to the server.

This PR waits for one minute if the database connection fails and try it again one more time.

### Reviewer guidance
The only way to review it is to have the chance of doing a big vacuum executin when kolibri starts and run kolibri under uwsgi at the same time.

### References
No issues existing, just feedback from kolibri-server tests

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/deployment/default/wsgi.py
+++ b/kolibri/deployment/default/wsgi.py
@@ -6,6 +6,7 @@ It exposes the WSGI callable as a module-level variable named ``application``.
 For more information on this file, see
 https://docs.djangoproject.com/en/1.8/howto/deployment/wsgi/
 """
+import logging
 import os
 import time
 
@@ -13,9 +14,20 @@ from django.core.wsgi import get_wsgi_application
 from django.db.utils import OperationalError
 
 os.environ.setdefault("DJANGO_SETTINGS_MODULE", "kolibri.deployment.default.settings.base")
+logger = logging.getLogger(__name__)
 
-try:
-    application = get_wsgi_application()
-except OperationalError:  # It happens when sqlite vacuum is being executed. The db is locked
-    time.sleep(60)  # vacuum should not take longer than 1 minute
-    application = get_wsgi_application()  # try again one last time
+application = None
+tries_remaining = 6
+interval = 10
+while not application and tries_remaining:
+    try:
+        logger.info("Starting Kolibri")
+        application = get_wsgi_application()
+    except OperationalError:
+        # an OperationalError happens when sqlite vacuum is being executed. the db is locked
+        logger.info("DB busy. Trying again in {}s".format(interval))
+        tries_remaining -= 1
+        time.sleep(interval)
+        application = get_wsgi_application()  # try again one last time
+if not application:
+    logger.error("Could not start Kolibri")
--- a/kolibri/deployment/default/wsgi.py
+++ b/kolibri/deployment/default/wsgi.py
@@ -7,9 +7,15 @@ For more information on this file, see
 https://docs.djangoproject.com/en/1.8/howto/deployment/wsgi/
 """
 import os
+import time
 
 from django.core.wsgi import get_wsgi_application
+from django.db.utils import OperationalError
 
 os.environ.setdefault("DJANGO_SETTINGS_MODULE", "kolibri.deployment.default.settings.base")
 
-application = get_wsgi_application()
+try:
+    application = get_wsgi_application()
+except OperationalError:  # It happens when sqlite vacuum is being executed. The db is locked
+    time.sleep(60)  # vacuum should not take longer than 1 minute
+    application = get_wsgi_application()  # try again one last time
--- a/kolibri/utils/cli.py
+++ b/kolibri/utils/cli.py
@@ -273,7 +273,6 @@ def start(port=None, daemon=True):
     :param: port: Port number (default: 8080)
     :param: daemon: Fork to background process (default: True)
     """
-
     run_cherrypy = conf.OPTIONS["Server"]["CHERRYPY_START"]
 
     # This is temporarily put in place because of
@@ -663,7 +662,11 @@ def main(args=None):  # noqa: max-complexity=13
         return
 
     if arguments['start']:
-
+        try:
+            with open(server.STARTUP_LOCK, 'w') as f:
+                f.write("%d\n%d" % (os.getpid(), port))
+        except (IOError, OSError):
+            logger.warn('Impossible to create file lock to communicate starting process')
         # Check if the content directory exists when Kolibri runs after the first time.
         check_content_directory_exists_and_writable()
 
--- a/kolibri/utils/cli.py
+++ b/kolibri/utils/cli.py
@@ -663,8 +663,7 @@ def main(args=None):  # noqa: max-complexity=13
 
     if arguments['start']:
         try:
-            with open(server.STARTUP_LOCK, 'w') as f:
-                f.write("%d\n%d" % (os.getpid(), port))
+            server._write_pid_file(server.STARTUP_LOCK, port)
         except (IOError, OSError):
             logger.warn('Impossible to create file lock to communicate starting process')
         # Check if the content directory exists when Kolibri runs after the first time.
--- a/kolibri/utils/server.py
+++ b/kolibri/utils/server.py
@@ -310,9 +310,9 @@ def get_status():  # noqa: max-complexity=16
         except (TypeError, ValueError):
             raise NotRunning(STATUS_STOPPED)
 
-        if not os.path.isfile(PID_FILE):
-            # There is no PID file (created by server daemon)
-            raise NotRunning(STATUS_STOPPED)  # Stopped
+    if not os.path.isfile(PID_FILE):
+        # There is no PID file (created by server daemon)
+        raise NotRunning(STATUS_STOPPED)  # Stopped
 
     # PID file exists and startup has finished, check if it is running
     try:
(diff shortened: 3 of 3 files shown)


benjaoming:
Hi @jredrejo - great find - should we maybe do a systemd integration for `kolibri-server` that blocks UWSGI from starting before after the kolibri system service is up? Could maybe be done by having a dummy systemd service that depends on kolibri and blocks uwsgi?

```
+----------------+          +----------------------------------+
|kolibri start   +--------->+Calls "kolibri start", which does |
|(systemd)       |          |not daemonize before returning    |
+----------------+          |from initializing jobs like       |
                            +vacuuming                         |
                            +--------------+-------------------+
                                           |
 +-----------------+                       |
 |uWSGI can now    |                       v
 |start the apps   |         +-------------+-------------------+
 |in /etc/uwsgi/   +<--------+systemd state: kolibri target    |
 |apps-enabled/    |         |ready.                           |
 |                 |         +---------------------------------+
 +-----------------+
```

benjaoming:
Oh yeah, this would also want us to avoid daemonizing before doing the initial VACUUM? Is that safe / better / should it be configurable through a flag like `--init-blocking`?

jredrejo:
@benjaoming 
> Hi @jredrejo - great find - should we maybe do a systemd integration for `kolibri-server` that blocks UWSGI from starting before after the kolibri system service is up? 

I liked that idea and have made a change in kolibri, so it does not start the server until all the backend tasks are finished. It affects to  CherryPy or the backend loop.
@indirectlylit @rtibbles @jamalex please, take a look, this touches some critical parts of the starting process.

kolibri/utils/cli.py
```diff
@@ -663,7 +662,11 @@ def main(args=None):  # noqa: max-complexity=13
         return
 
     if arguments['start']:
-
+        try:
+            with open(server.STARTUP_LOCK, 'w') as f:
+                f.write("%d\n%d" % (os.getpid(), port))
```
benjaoming: You can probably use `_write_pid_file()` for this?

kolibri/utils/server.py
```diff
@@ -285,22 +296,24 @@ def get_status():  # noqa: max-complexity=16
...
-                    raise NotRunning(STATUS_STARTING_UP)
-                # It's dead so assuming the startup went badly
-                else:
-                    raise NotRunning(STATUS_FAILED_TO_START)
-            # Couldn't parse to int or empty PID file
-            except (TypeError, ValueError):
-                raise NotRunning(STATUS_STOPPED)
-        raise NotRunning(STATUS_STOPPED)  # Stopped
+    # Is there a startup lock?
+    if os.path.isfile(STARTUP_LOCK):
```
benjaoming, line 300: This looks good, and could improve the scenarios where the server is considered to be stopped... only theoretically, I can't see a situation where the PID_FILE wouldn't be written.

One thing I was asked repeatedly about was if we could have this situation where Kolibri fails to start because it has crashed and the startup lock is left behind. So we really really have to use it conservatively.

A worst case that happened in KA Lite was that some project had written system startup scripts that just deleted the pid lock and started KA Lite no matter what :)

benjaoming (approved):
If you could reuse the `_write_pid_file()`, that'd be great. Otherwise, LGTM! :+1:

benjaoming (dismissed):
Oh wait actually one thing: Could you write a .travis.yml entry that ensures that Kolibri does NOT start while vacuuming? Mockup:

```
# Here is the logic, tested:
`exit 0` && ! `exit 0` || echo "Kolibri should never start while vacuuming" ; exit 1

# Here is an implementation?
`kolibri start` && ! `kolibri start` || echo "Kolibri should never start while vacuuming" ; exit 1
```

kolibri/utils/cli.py
jredrejo: Done, as a curiosity: this function is not being used in the code. All the places where a pid file is being written are ignoring it.

kolibri/utils/server.py
jredrejo: From your comment I am not sure if you are just explaining it or requesting something more to be done.

jredrejo:
> Oh wait actually one thing: Could you write a .travis.yml entry that ensures that Kolibri does NOT start while vacuuming? Mockup:

I don't understand this point. The main purpose of this PR is to avoid that, Kolibri will not start while vacuuming (nor until pingback and clearsessions are finished too). I.e. with this PR `kolibri start` will always wait for all this booting parts to finish before starting cherrypy (or the background loop if cherrypy is not enabled)

kolibri/utils/server.py
benjaoming: I am not explaining anything else to be done, just that since you have recently looked at the code, there is a bit of background to it. I did not put a giant "startup lock" around everything because there was a concern that we would get in to a situation where Kolibri's own startup lock mechanism was blocking it from starting up.

This is just background, otherwise I would have mentioned it in the actual review.

benjaoming:
This behavior has no test, I suggested a way to test it, nothing more.

kolibri/deployment/default/wsgi.py
```diff
@@ -7,9 +7,15 @@
...
+from django.db.utils import OperationalError
 
 os.environ.setdefault("DJANGO_SETTINGS_MODULE", "kolibri.deployment.default.settings.base")
 
-application = get_wsgi_application()
+try:
+    application = get_wsgi_application()
+except OperationalError:  # It happens when sqlite vacuum is being executed. The db is locked
+    time.sleep(60)  # vacuum should not take longer than 1 minute
+    application = get_wsgi_application()  # try again one last time
```
indirectlylit: Waiting a whole minute might lead someone to believe things are broken if they are waiting around for it? 
 Maybe this should be a loop with some logging? e.g.

```python
application = None
tries_remaining = 6
interval = 10
while not application and tries_remaining:
    try:
        logging.info("Starting Kolibri")
        application = get_wsgi_application()
    except OperationalError:
        # an OperationalError happens when sqlite vacuum is being executed. the db is locked
        logging.info("DB busy. Trying again in {}s".format(interval))
        tries_remaining -= 1
        time.sleep(interval)
        application = get_wsgi_application()  # try again one last time
if not application:
    logging.error("Could not start Kolibri")
```

kolibri/utils/server.py
indirectlylit: Below we check whether the PID exists and raise an exception if it doesn't. Perhaps a safer action would be to log a warning, delete the startup lock, and continue with startup?

indirectlylit (commented):
overall looks fine to me. I agree that it's better to err on the side of trying to start (and failing) vs potentially blocking ourselves from starting due to a faulty lock file

kolibri/deployment/default/wsgi.py
jredrejo: yep, I like that idea.

jredrejo:
@indirectlylit 
> Below we check whether the PID exists and raise an exception if it doesn't. Perhaps a safer action would be to log a warning, delete the startup lock, and continue with startup?

I agree, in fact I would say that `STARTUP_LOCK` file could be an empty, not need to check anything inside. We have a `PID_FILE` for that purpose. I thought to write in `STARTUP_LOCK` the same that was written in other locks, but, honestly, now I can't think of any reason to write anything there.

jredrejo:
@indirectlylit
> 
> > Below we check whether the PID exists and raise an exception if it doesn't. Perhaps a safer action would be to log a warning, delete the startup lock, and continue with startup?
> 
> I agree, in fact I would say that `STARTUP_LOCK` file could be an empty, not need to check anything inside. We have a `PID_FILE` for that purpose. I thought to write in `STARTUP_LOCK` the same that was written in other locks, but, honestly, now I can't think of any reason to write anything there.

I have double checked the code and added more comments to explain the situation. I think the code is correct. There were some concept problems in the comments we both did before:

 1. This code is to check the status of the server, it has not any influence on the startup of the system, the exceptions raised here don't stop/resume the server behaviour. They are use just to report the status as an output of `kolibri status`
  2. The current flow of the startup is:
     1. `STARTUP_LOCK` is created
     2. clear sessions, ping and vacuum commands are executed
     3. `PID_FILE` is created
     4. When vacuum finishes, `STARTUP_LOCK` is deleted
     5. CherryPy or the backend loop starts

i.e. an existing or wrong `STARTUP_LOCK` does not alter the server to start. Only a  never ending vacuum command could do it. I can't think of any reason for vacuum to last forever. Even with a corrupted database it finishes soon raising an error.

indirectlylit:
changed base to 0.12.x

indirectlylit:
some tests failing and a merge conflict. Let us know when ready for re-review

jredrejo:
@indirectlylit tests are passing now. However, the PR was branched from develop, after being changed to merge in release-v0.12.x has brought several commits from mrpau-richard for the buildkite configuration.
Can you give me any advice on how to proceed with this?

rtibbles:
Could you try rebasing onto release 0.12 and force pushing? Should exclude the spare commits.

indirectlylit:
If rebasing doesn't work, I would:

1. make a fresh local `temp` branch off of `release-v0.12.x`
2. [cherry pick the ranges of commits](https://www.tollmanz.com/git-cherry-pick-range/) that you're interested in into `temp`
3. locally delete `save_lockins`
4. rename `temp` to `save_lockins`
5. force push to github

jredrejo:
@indirectlylit  thanks!
Steps done and everything went well. Ready to merge.

indirectlylit (approved):

PR #4984, opened 2019-02-09
Change server persistence logic based on page visibility

### LINKED ISSUE #3903: clean up API of HeartBeat class
### Observed behavior

In #3900 we added a new `stop` method alongside the `start` method. However there's already `setActive`,  `setConnected`, `setInactive`, and `wait` methods, all of which sound somewhat related but different.

### Expected behavior

Recommend clarifying what's being started/stopped/set etc. E.g. `setUserActiveFlag`, `stopHeartbeatPoll`?

Also, recommend clarifying which of these are protected, internal methods. E.g. `_signOutDueToInactivity`?

### User-facing consequences

developer confusion

### Context

0.10.0

rtibbles:
### Summary
Adds a page visibility tracker to reduce the aggressiveness of heartbeat behaviour and to pause log saving, while the page is not visible.

Also cleans up some unused code (including the sessions middleware which is a no op in Django - the presence of it in settings is only used to trigger some backwards compatibility behaviour that we do not need).

Finally, it moves all CSRF setting to be the responsibility of the session endpoint, to remove any user specific data from the base template to pave the way for better caching.

### Reviewer guidance
From an incognito browser, can you change the language on the home page? (issue that the `csrf_token` was added to the `base.html` to fix). Can you make posts etc to endpoints as a coach/admin?

Does everything still run as expected when you background a tab and then return to it after 10 minutes?

Now also fixes #3903

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated


kolibri/core/assets/src/core-app/baseClient.js
```diff
@@ -4,11 +4,43 @@
...
+ *
+ * @returns {Client}
+ */
+const csrf = interceptor({
+  request: function handleRequest(request) {
+    var headers, name, token;
+
+    headers = request.headers || (request.headers = {});
+    name = request.csrfTokenName || 'X-CSRFToken';
+    token = request.csrfToken || cookiejs.get('csrftoken');
```
indirectlylit, line 39: It's not clear where these names are coming from - I assume we have control over them on the server-side?

If we know the names of the header and the cookie ahead of time, I would expect the `||` logic to be unnecessary, and the values simply hard-coded here, perhaps with a comment indicating what middleware sets them on the server.

kolibri/core/assets/src/heartbeat.js
```diff
@@ -70,12 +68,20 @@ export class HeartBeat {
   setInactive() {
     this.active = false;
   }
-  wait() {
+  get delay() {
```
indirectlylit: recommend prefixing internal functions and state with `_` so it's clear that these are not part of the class instance's public API

kolibri/core/assets/src/heartbeat.js
```diff
@@ -23,12 +23,10 @@ const timeoutReconnectTime = 60;
 
 const minReconnectTime = 5;
 
+const activeDelay = 240000;
```
indirectlylit: few minor thoughts. Maybe...

* constant variable names to `SCREAMING_SNAKE_CASE`
* indicate here how this value is chosen? E.g. "the server closes the session after 10 minutes so we should poll at least every 5 minutes"
* store in seconds for consistency with reconnect time

kolibri/core/assets/src/state/modules/session.js
```diff
@@ -72,6 +74,9 @@ export default {
   },
   mutations: {
     CORE_SET_SESSION(state, value) {
+      const serverTime = value.server_time;
+      setServerTime(serverTime);
```
indirectlylit: I don't think we should call functions with non-vuex side effects inside mutations. Maybe we need a `setSession` action?

indirectlylit (commented):
looks good. haven't had a chance to test yet, but here are some small notes from a visual code review

kolibri/core/assets/src/core-app/baseClient.js
rtibbles: This is vendored from our rest client, I did briefly simplify this code, but it bought very little, and it seemed safer just to bring in the code as is.

indirectlylit: cool. worth maybe just adding a comment indicating where on the backend these are handled?

kolibri/core/assets/src/heartbeat.js
rtibbles: Have addressed this in latest commits.

kolibri/core/assets/src/heartbeat.js
rtibbles: Have addressed this in latest commits.

kolibri/core/assets/src/state/modules/session.js
rtibbles: Have addressed this in latest commits.

kolibri/core/assets/src/core-app/baseClient.js
rtibbles: They aren't anywhere in our backend code, it's all in Django magic. Can link to the Django CSRF docs if helpful.

indirectlylit:
one test failure

rtibbles:
0 test failures!

kolibri/core/assets/src/heartbeat.js
```diff
@@ -15,68 +16,98 @@ import {
...
+const MIN_RECONNECT_TIME = 5;
+
+// The usual time to wait between session check pings, this is set
+// at 4 minutes, as this will maintain activity in the user activity
+// log in the Kolibri server, if the user has been active in this time,
+// as this is windowed at activity being logged in the last 5 minutes.
+// In the case of the page being backgrounded, we double this, which is
+// sufficient to maintain the current Kolibri session, but will start a
+// new user session log when activity is detected again.
+const ACTIVE_DELAY = 240;
```
indirectlylit, line 44: thanks for all the comments!

indirectlylit (approved):
tried both scenarios, works for me

kolibri/core/assets/src/heartbeat.js
```diff
@@ -15,68 +16,98 @@ import {
...
-    this.timerId = setTimeout(this.beat, reconnectTime * 1000 || this.delay);
-    return this.timerId;
+    if (!store.getters.connected && reconnectTime) {
+      // If we are currently engaged in exponential backoff in trying to reconnect to the server
+      // use the current reconnect time preferentially instead of the standard delay.
+      // The reconnect time is stored in seconds, so multiply by 1000 to give the milliseconds.
+      return reconnectTime;
+    }
+    // If page is not visible, don't poll as frequently, as user activity is unlikely.
+    return store.state.pageVisible ? ACTIVE_DELAY : ACTIVE_DELAY * 2;
```
jredrejo, line 104: is there any objetive reason to the x2 factor when the page is not visible, or it's just a guess?

rtibbles: It's less frequent, but not so infrequent that it would cause a session timeout with our default setting.

PR #4994, opened 2019-02-10
Allow better caching of SPA launch pages

rtibbles:
### Summary
Sets up language URL prefixing to allow SPA launch pages to be cached with a minimal number of Vary headers set.

Sets up a 300 second cache on all SPA launch pages.

### Reviewer guidance
Does everything work as expected? Do redirects work well?

### Reference
Dependent on #4984.

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated


kolibri/core/decorators.py
```diff
@@ -275,3 +276,19 @@ def wrapped_view(*args, **kwargs):
...
+    Set appropriate Vary on headers on a view that specify there is
+    no user specific data being rendered in the view.
+    """
+    def inner_func(*args, **kwargs):
+        request = args[0]
+        del request.session
+        response = view_func(*args, **kwargs)
+        headers = ['accept-encoding', 'accept', 'user-agent']
+        patch_response_headers(response, cache_timeout=300)
+        response['Vary'] = ', '.join(headers)
```
indirectlylit: could be simplified, unless the intent was to merge these headers with some existing headers?

```python
response['Vary'] = 'accept-encoding, accept, user-agent'
```

kolibri/core/decorators.py
```diff
@@ -275,3 +276,19 @@ def wrapped_view(*args, **kwargs):
         return view_func(*args, **kwargs)
     wrapped_view.signin_redirect_exempt = True
     return wraps(view_func)(wrapped_view)
+
+
+def cache_no_user_data(view_func):
+    """
+    Set appropriate Vary on headers on a view that specify there is
+    no user specific data being rendered in the view.
+    """
```
indirectlylit, line 336: how does this interact with bootstrapped data and authorization to access certain pages and data?

indirectlylit (commented):
Looks good from a cursory review.

Do you have a sense for what kind of performance benefits this achieves?

I'm a bit concerned that this touches so many files so close to 0.12.0, and particularly stuff that's harder to test like i18n and caching behaviors.

kolibri/core/decorators.py
rtibbles: This shouldn't be used on pages that bootstrap user data into the page. I have also removed most instances of bootstrapped user data in the page, because it would interfere with this.

rtibbles:
We can hold off on this until 0.12.1 - I mostly want to have a few things in our pocket to allow performance improvements if we discover that we need them for 0.12.0.

indirectlylit:
setting milestone to 0.12.1

note there's a conflict in kolibri_app.js

rtibbles:
Not a hugely representative methodology, but I ran siege with default settings (25 concurrent users), and these are the results on current release-v0.12.x:
```
Transactions:		        1000 hits
Availability:		      100.00 %
Elapsed time:		       65.42 secs
Data transferred:	        6.90 MB
Response time:		        1.60 secs
Transaction rate:	       15.29 trans/sec
Throughput:		        0.11 MB/sec
Concurrency:		       24.53
Successful transactions:        1000
Longest transaction:	        5.91
Shortest transaction:	        0.18
```

These are the results when run on this PR:
```
Transactions:		        1000 hits
Availability:		      100.00 %
Elapsed time:		       10.00 secs
Data transferred:	        6.76 MB
Response time:		        0.25 secs
Transaction rate:	      100.00 trans/sec
Throughput:		        0.68 MB/sec
Concurrency:		       24.72
Successful transactions:        1000
Longest transaction:	        3.69
Shortest transaction:	        0.06
```

Testing conditions definitely favour this PR and what it is doing, but it is still a roughly 6 times speed up on average response time.

rtibbles:
The other thing to note, obviously, is that this is moving some bootstrapping out of user pages, so this will result in more requests to the server, so we would probably want to do a live load test to properly assess the impact of this.

kolibri/core/decorators.py
```diff
@@ -275,3 +276,19 @@ def wrapped_view(*args, **kwargs):
...
+    def inner_func(*args, **kwargs):
+        request = args[0]
+        del request.session
```
benjaoming, line 339: Do we want a decorator that just deletes the entire session data? Where will this be used? Normally, it would seem like a logout procedure. This doesn't match that well with the `__docstring__` of the decorator.

kolibri/core/device/middleware.py
```diff
@@ -1,17 +1,70 @@
...
+from django.urls import is_valid_path
 from django.utils import translation
+from rest_framework.views import APIView
 
 from .translation import get_language_from_request
+from kolibri.utils.conf import OPTIONS
 
 
-class KolibriLocaleMiddleware(LocaleMiddleware):
+class KolibriLocaleMiddleware:
```
benjaoming: Why was the inheritance removed? It would be okay to override all the current methods, but keeping the inheritance somehow also means keeping the contract of what a `LocaleMiddleware` is supposed to do.

kolibri/core/context_processors/custom_context_processor.py
```diff
@@ -33,12 +33,10 @@ def pass_browser_entry(agent, entry):
...
 def supported_browser(request):
-    if 'supported_browser' not in request.session:
-
-        user_agent = user_agents.parse(request.META.get('HTTP_USER_AGENT', ''))
-        request.session['supported_browser'] = all(
-            pass_browser_entry(user_agent, entry) for entry in browser_requirements)
-    return {'supported_browser': request.session['supported_browser']}
+    user_agent = user_agents.parse(request.META.get('HTTP_USER_AGENT', ''))
+    return {
+        'supported_browser': all(pass_browser_entry(user_agent, entry) for entry in browser_requirements)
```
jredrejo: I see the returned value is being removed from the session. 
wouldn't it be good to use the session to cache the value and return it if it already exixts? it's being called/recalculated from the base.html template whenever the SPA changes.

kolibri/core/decorators.py
jredrejo: I understand the utility of this decorator except for the point Ben comments, why is it doing `del request.session`?
All the other lines in the decorator make sense for me, but I don't understand the rationale of deleting the session when switching between learn/learn.html and user/user.html

kolibri/core/device/middleware.py
jredrejo: @benjaoming I think the reasons are explained in the related PR #4984 :

> Also cleans up some unused code (including the sessions middleware which is a no op in Django - the presence of it in settings is only used to trigger some backwards compatibility behaviour that we do not need).

benjaoming: I think it's better to keep it and override changed/unneeded methods. Simply because of maintaining the link to the reference implementation.

kolibri/deployment/default/settings/dev.py
```diff
@@ -11,3 +11,10 @@
 ROOT_URLCONF = 'kolibri.deployment.default.dev_urls'
 
 DEVELOPER_MODE = True
+
+# Create a dummy cache for each cache
```
jredrejo, line 15: thus, nulls the caches, right?

kolibri/core/decorators.py
rtibbles: The reason for deleting the session is that otherwise Django adds a Vary header for cookies onto the request, meaning no caching takes place.

I could achieve the same effect by having this decorator add a "don't add cookie vary header" flag to the view, and then modify other session modifying middleware to obey that directive, but this seemed simplest.

kolibri/core/device/middleware.py
rtibbles: The other reason for not inheriting is that the default Django locale middleware class is an old style middleware, so I would be overriding its methods simply to turn them into a no op (and I am not entirely sure how using new style middleware works when inheriting from the middleware deprecation compatibility class).

kolibri/core/context_processors/custom_context_processor.py
rtibbles: I can either cache the pages, or I can use the session object. A middle ground might be to memoize the calculation based on user agent?

kolibri/deployment/default/settings/dev.py
rtibbles: Yes, I am open to arguments that we should enforce caching in development to expose caching induced bugs to devs, but this will prevent having to do superfluous saves on python files to restart the devserver just to clear the cache.

kolibri/core/context_processors/custom_context_processor.py
jredrejo: Yeah, I hadn't thought of using the user agent, but your idea of memoizing based on it would be the more efficient solution.

kolibri/deployment/default/settings/dev.py
jredrejo: nah, I just wanted to confirm I had understood your purpose correctly.

indirectlylit:
switching base to `develop`

rtibbles:
This PR is now dependent on #5246

rtibbles:
Note that a live load test of this showed improved performance of this over current release 0.12 - although there was a cache warm-up phase where things were a little slower.

rtibbles:
This PR now also removes the GZip middleware which seems to have been a considerable performance bottleneck on our bulkier API endpoints. As an example, here is a siege run against the channel endpoint, with and without GZip:
```
Channel Endpoint:
With GZip:
Transactions:		         500 hits
Availability:		      100.00 %
Elapsed time:		       10.04 secs
Data transferred:	       80.57 MB
Response time:		        0.49 secs
Transaction rate:	       49.80 trans/sec
Throughput:		        8.03 MB/sec
Concurrency:		       24.63
Successful transactions:         500
Longest transaction:	        2.27
Shortest transaction:	        0.08

Without GZip:
Transactions:		         500 hits
Availability:		      100.00 %
Elapsed time:		        5.79 secs
Data transferred:	      111.32 MB
Response time:		        0.24 secs
Transaction rate:	       86.36 trans/sec
Throughput:		       19.23 MB/sec
Concurrency:		       20.75
Successful transactions:         500
Longest transaction:	        3.02
Shortest transaction:	        0.01
```

kolibri/core/device/translation.py
```diff
@@ -3,58 +3,91 @@
...
     If the user requests a sublanguage where we have a main language, we send
     out the main language.
-    If check_path is True, the URL path prefix will be checked for a language
-    code, otherwise this is skipped for backwards compatibility.
     """
 
+    try:
+        # If this is not a view that needs to be translated, return None, and be done with it!
+        if not getattr(resolve(request.path_info).func, "translated", False):
+            return None, False
```
indirectlylit, line 80: technically, I think this should be `None, True`

kolibri/core/device/middleware.py
```diff
@@ -1,34 +1,62 @@
...
-
-class IgnoreGUIMiddleware(object):
     def __init__(self, get_response):
         self.get_response = get_response
-        # One-time configuration and initialization.
 
     def __call__(self, request):
-        # Code to be executed for each request before
-        # the view (and later middleware) are called.
+        language, language_from_path = get_language_from_request(request)
```
indirectlylit: maybe rename to `get_language_and_reason(request)` or something to convey multiple return values?

kolibri/core/assets/src/utils/userAgent.js
```diff
@@ -0,0 +1,116 @@
...
+      }
+    }
+  }
+  if (browser[VERSION]) {
+    browser[VERSION] = browser[VERSION].split('.');
+  }
+  return browser;
+}
+
+export function passesRequirements(browser, requirements) {
```
indirectlylit, line 159: recommend adding a brief comment about the expected content of `browser` and `requirements`

kolibri/core/decorators.py
```diff
@@ -320,3 +321,26 @@ def wrapped_view(*args, **kwargs):
...
+    data into it - this will remove the headers that will make this vary
+    on a per user basis.
+    """
+
+    def inner_func(*args, **kwargs):
+        request = args[0]
+        del request.session
+        response = view_func(*args, **kwargs)
+        patch_response_headers(response, cache_timeout=300)
+        response["Vary"] = "accept-encoding, accept"
```
indirectlylit: interesting. this is the minimal case?

kolibri/core/device/middleware.py
```diff
@@ -1,34 +1,62 @@
...
-class KolibriLocaleMiddleware(LocaleMiddleware):
-    def process_request(self, request):
-        language = get_language_from_request(request)
-        translation.activate(language)
-        request.LANGUAGE_CODE = translation.get_language()
+class KolibriLocaleMiddleware(object):
+    """
+    Copied and then modified into a new style middleware from:
+    https://github.com/django/django/blob/stable/1.11.x/django/middleware/locale.py
+    Also has several other changes to suit our purposes.
```
indirectlylit, line 14: maybe include a description of what the intended logic is, especially with regard to redirects? It's a bit hard to evaluate otherwise

kolibri/core/device/middleware.py
```diff
@@ -1,34 +1,62 @@
...
+                    script_prefix = OPTIONS["Deployment"]["URL_PATH_PREFIX"]
+                    # Insert language after the script prefix and before the
+                    # rest of the URL
+                    language_url = request.get_full_path(
+                        force_append_slash=path_needs_slash
+                    ).replace(script_prefix, "%s%s/" % (script_prefix, language), 1)
+                    return HttpResponseRedirect(language_url)
+
+            if "Content-Language" not in response:
+                response["Content-Language"] = language
```
indirectlylit, line 78: Since this block has some complex nested if/else statements, recommend adding a few more comments to conditional branches indicating the circumstances under which you'd expect the code to reach the branch.

Another thing that can be helpful is to stick `return` statements anywhere the processing is supposed to be "done". This can help reduce ambiguity of intent compared to letting execution fall out of the conditionals to the final `return`.

kolibri/utils/server.py
```diff
@@ -188,6 +194,33 @@ def stop(pid=None, force=False):
...
+            )
+        # otherwise return either the min or max amount
+        return MAX_CACHE if total_memory >= MAX_MEM else MIN_CACHE
+    elif sys.platform.startswith(
+        "darwin"
+    ):  # Considering MacOS has at least 4 Gb of RAM
+        return MAX_CACHE
+    return MIN_CACHE
+
+
```
indirectlylit, line 223: this looks familiar :)

indirectlylit (commented):
This is looking good - I think having languages segmented under their own URLs is a good move, and it makes sense to do it in support of better caching.

There are a lot of possible states in here, given conditional logic for language selection, browser compatibility, and page caching behaviors. Most of my notes relate to future maintainability in terms of clarifying intended behavior.

The test coverage is a great defense against edge cases bugs, but I haven't reviewed them carefully yet.

In terms of manual testing it will be hard to cover every scenario. I think it might be helpful to add some gherkin stories which exercise various paths through the conditionals, at least for the intended "happy paths".

I started testing briefly, but the language switching didn't seem to work on `devserver-hot`:

switch

rtibbles:
Thanks - I'll double check the language switching behaviour from the frontend some more, and see if I can replicate.

kolibri/core/device/translation.py
rtibbles: `The language has been calculated from the path.`

I am using a Russellian theory of descriptions here to return `False`, because of the null referent of `language` (`language` is `None`), the predicate evaluates to `False`, just as `The language has not been calculated from the path` would evaluate to `False` by the same reasoning.

kolibri/core/device/middleware.py
rtibbles: At the very least I should update the doc string of the function - coming up with a concise and meaningful name might be trickier.

kolibri/core/assets/src/utils/userAgent.js
rtibbles: Yeah, JSDocs needed.

kolibri/core/decorators.py
rtibbles: I believe so, not varying on these could cause problems if we sent the wrong mimetype back to a request, or tried to compress for a client that couldn't handle compression.

`accept-charset` is another possibility, but I have never seen our backend add `accept-charset` to the Vary header. My hope is this would be subsumed by language?

kolibri/utils/server.py
rtibbles: Only done it twice so far, so just changed a few names to protect the innocent.

kolibri/core/device/translation.py
indirectlylit: my cat's breath smells like cat food

kolibri/core/device/middleware.py
rtibbles: > Another thing that can be helpful is to stick return statements anywhere the processing is supposed to be "done". This can help reduce ambiguity of intent compared to letting execution fall out of the conditionals to the final `return`.

This is already the case, the return of the redirect is the shortcut return.

rtibbles:
Have now addressed all comments - the language changing behaviour was indeed broken. Have fixed the issue, and also reimplemented redirect to make sure that when you change language it returns you to exactly the same page (including fragment and query param) that you were on previously.

I think this is ready to merge, assuming there are no tests that I have missed.

indirectlylit (approved):
did a bit of testing, and everything seems in order. Let's get it in!

I think there's an open interesting question of "permalinks" and whether they should contain the target language or not...

Issue #4965, opened 2019-02-08
Not enough space still appear even my hard disk is not full


mrpau-julius:
### Observed behavior

I try to import a channel in kolibri studio (online). When I select a channel, I see an error message appeared "Not enough space on your device"  so that I cannot import. 
![not enough space](https://user-images.githubusercontent.com/38421180/52469779-5ec58500-2bc7-11e9-8eda-71bad18d819a.png)

### Expected behavior

If the hard disk is not full the error message will not appeared.

### Context

* Kolibri version 0.12.0 beta 3
 * Operating system Windows 10
 * Browser Firefox 65.0 (64-bit)

indirectlylit:
@mrpau-julius I think you've observed similar behavior in the past. Would you mind linking to related previous issues if you know of them?

jonboiser:
The '-1' in "drive space available" indicates to me that the 'freespace' API is returning an error, see 'getAvailableSpaceOnDrive'.

rtibbles:
Fixed in #4997

Issue #4966, opened 2019-02-08
Device info page error on Windows


mrpau-richard:
### Observed behavior
Device info page has an error on WIndows 7.

### Errors and logs
```
ERROR 2019-02-08 01:50:21,316 exception Internal Server Error: /api/device/deviceinfo/
Traceback (most recent call last):
  File "C:\Python34\lib\site-packages\kolibri\dist\django\core\handlers\exception.py", line 41, in inner
    response = get_response(request)
  File "C:\Python34\lib\site-packages\kolibri\dist\django\core\handlers\base.py", line 187, in _get_response
    response = self.process_exception_by_middleware(e, request)
  File "C:\Python34\lib\site-packages\kolibri\dist\django\core\handlers\base.py", line 185, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "C:\Python34\lib\site-packages\kolibri\dist\django\views\decorators\csrf.py", line 58, in wrapped_view
    return view_func(*args, **kwargs)
  File "C:\Python34\lib\site-packages\kolibri\dist\django\views\generic\base.py", line 68, in view
    return self.dispatch(request, *args, **kwargs)
  File "C:\Python34\lib\site-packages\kolibri\dist\rest_framework\views.py", line 495, in dispatch
    response = self.handle_exception(exc)
  File "C:\Python34\lib\site-packages\kolibri\dist\rest_framework\views.py", line 455, in handle_exception
    self.raise_uncaught_exception(exc)
  File "C:\Python34\lib\site-packages\kolibri\dist\rest_framework\views.py", line 492, in dispatch
    response = handler(request, *args, **kwargs)
  File "C:\Python34\lib\site-packages\kolibri\core\device\api.py", line 93, in get
    info['content_storage_free_space'] = get_free_space()
  File "C:\Python34\lib\site-packages\kolibri\utils\system.py", line 175, in get_free_space
    st = os.statvfs(os.path.realpath(path))
AttributeError: 'module' object has no attribute 'statvfs'

```

### Context
Tell us about your environment, including:
 * Kolibri version: 0.12-beta3
 * Operating system: Windows 7
 * Browser: Firefox 65.0

radinamatic:
Confirmed on both `0.12.0.b3` (Chrome) and `0.12.0.b4` (FF v66):

[...]

PR #4997, opened 2019-02-10
Windows free space regression fix and python code quality fixes

rtibbles:
### Summary
Fixes low risk code quality issues reported by LGTM: https://lgtm.com/projects/g/learningequality/kolibri/alerts/?mode=list&lang=python

Also fixes a regression in free space checking, whereby windows free space checking behaviour is overridden by the default (*nix focused) behaviour.

### Reviewer guidance
Do tests pass? Does free space calculation work on Windows?

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated


radinamatic:
Some errors in the console, but free space checks OK with built asset on Win7 VM, both Chrome & FF:





Where did `Select all` label go in Chrome? It reappears if I uncheck the `overflow-x: hidden`... 🤔 

overflow

cc @indirectlylit @jonboiser

radinamatic (dismissed):
Can't offer opinion on the code innards, but the result LGTM! 😊

rtibbles:
Note that the API errors appear to be 404s, which I believe are expected.

rtibbles:
I had not realized, but the Windows free space error has also previously been documented here: #4966, this PR should fix it, as confirmed by @radinamatic above.

rtibbles:
@radinamatic @mrpau-richard can you confirm if this PR also fixes #4965 ? I think it should.

radinamatic:
Actually, my above comments were meant as a fix-confirmation for #4965, but #4966 seems fixed too! 😉

kolibri/core/analytics/management/commands/benchmark.py
```diff
@@ -78,7 +78,7 @@ def handle(self, *args, **options):
             sys.exit(1)
 
         try:
-            kolibri_cpu, kolibri_mem = get_kolibri_use()
+            get_kolibri_use()
```
jredrejo, line 81: good catch! this was only used to trigger the error

kolibri/core/analytics/pskolibri/__init__.py
```diff
@@ -161,7 +155,6 @@ def cpu_percent():
     utilization as a percentage.
     """
     global _last_cpu_times
-    global _last_per_cpu_times
```
jredrejo, line 164: yeah, this is not needed anymore as we are not calculating cpu usage per core, but per system.

kolibri/core/analytics/utils.py
```diff
@@ -73,9 +73,6 @@ def normalize_version_to_semver(version):
 
     initial, dev = re.match(r"(.*?)(\.dev.*)?$", version).groups()
```
jredrejo: if `initial` is not used anymore, I would prefer to make it evident using the format
`_ , dev = re.match(r"(.*?)(\.dev.*)?$", version).groups()`

kolibri/plugins/coach/api.py
```diff
@@ -156,7 +156,7 @@ def list(self, request, *args, **kwargs):
         Then it fetches and saves the needed information to know how many coaches
         are requesting notifications in the last five minutes
         """
-        response = super(viewsets.ReadOnlyModelViewSet, self).list(request, *args, **kwargs)
+        response = super(ClassroomNotificationsViewset, self).list(request, *args, **kwargs)
```
jredrejo: have you checked this works properly?
I remember having to use the parent class because this produced an infinite loop

jredrejo (dismissed):

kolibri/core/analytics/utils.py
rtibbles: Fair!

kolibri/plugins/coach/api.py
rtibbles: I can revert this - I wanted to only do low risk code quality fixes here, so if this is at all risky, we can leave it as is.

rtibbles: This was risky, and did exactly what you said. I have reverted the change, and added a comment to prevent similar foolhardiness in the future!

kolibri/core/analytics/utils.py
rtibbles: Updated!

ralphiee22 (approved):
Nice!

PR #5003, opened 2019-02-11
More semantically handle malformed zipcontent paths by returning 404s.

### LINKED ISSUE #5002: AssertionError: 'xxxx' is not a valid content storage filename
Sentry Issue: [KOLIBRI-BACKEND-6](https://sentry.io/learningequality/kolibri-backend/issues/877445818/?referrer=github_integration)

```
AssertionError: 'fonts' is not a valid content storage filename
(7 additional frame(s) were not displayed)
...
  File "django/views/decorators/clickjacking.py", line 58, in wrapped_view
    resp = view_func(*args, **kwargs)
  File "django/utils/decorators.py", line 67, in _wrapper
    return bound_func(*args, **kwargs)
  File "django/views/decorators/http.py", line 88, in inner
    res_etag = etag_func(request, *args, **kwargs) if etag_func else None
  File "kolibri/core/content/views.py", line 90, in calculate_zip_content_etag
    zipped_path = get_content_storage_file_path(zipped_filename)
  File "kolibri/core/content/utils/paths.py", line 73, in get_content_storage_file_path
    assert VALID_STORAGE_FILENAME.match(filename), "'{}' is not a valid content storage filename".format(filename)
```

rtibbles:
### Summary
Previously both the etag and get on the zipcontent endpoint would raise a 500 error because of an assertion error in the calculation of the local storage path.

This PR catches these assertion errors and raises a 404.

### Reviewer guidance
Does the zipcontent endpoint work as expected? Do the tests pass?

### References
Fixes #5002

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/core/content/views.py
+++ b/kolibri/core/content/views.py
@@ -87,7 +87,10 @@ def _add_content_security_policy_header(request, response):
 
 
 def calculate_zip_content_etag(request, zipped_filename, embedded_filepath):
-    zipped_path = get_content_storage_file_path(zipped_filename)
+    try:
+        zipped_path = get_content_storage_file_path(zipped_filename)
+    except AssertionError:
+        return None
 
     # if no path, or a directory, is being referenced, look for an index.html file
     if not embedded_filepath or embedded_filepath.endswith("/"):
@@ -99,6 +102,14 @@ def calculate_zip_content_etag(request, zipped_filename, embedded_filepath):
         return hashlib.md5(get_hashi_filename().encode('utf-8')).hexdigest()
 
     return hashlib.md5((zipped_filename + embedded_filepath).encode('utf-8')).hexdigest()
+
+
+def get_path_or_404(zipped_filename):
+    try:
+        # calculate the local file path to the zip file
+        return get_content_storage_file_path(zipped_filename)
+    except AssertionError:
+        raise Http404('"%(filename)s" is not a valid zip file' % {'filename': zipped_filename})
 
 
 class ZipContentView(View):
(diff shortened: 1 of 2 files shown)


kolibri/core/content/views.py
```diff
@@ -87,7 +87,10 @@ def _add_content_security_policy_header(request, response):
 
 
 def calculate_zip_content_etag(request, zipped_filename, embedded_filepath):
-    zipped_path = get_content_storage_file_path(zipped_filename)
+    try:
+        zipped_path = get_content_storage_file_path(zipped_filename)
+    except AssertionError:
+        return None
```
indirectlylit, line 93: My sense is that we're mis-using assertions in `get_content_storage_file_path`:

```python
def get_content_storage_file_path(filename, datafolder=None):
    assert VALID_STORAGE_FILENAME.match(filename), "'{}' is not a valid content storage filename".format(filename)
    return os.path.join(
        get_content_storage_dir_path(datafolder),
        filename[0],
        filename[1],
        filename,
    )
```

I would expect us to instead define and handle a custom `InvalidContentStorageFilename` exception.

[Reference](https://wiki.python.org/moin/UsingAssertionsEffectively):

> Assertions should *not* be used to test for failure cases that can occur because of bad user input or operating system/environment failures, such as a file not being found. Instead, you should raise an exception, or print an error message, or whatever is appropriate. One important reason why assertions should only be used for self-tests of the program is that assertions can be disabled at compile time.

rtibbles: You are probably right, but I didn't want to mess with the API, in case it disrupted other parts of the app. Can dig deeper here if helpful.

indirectlylit: created a follow-up issue here: https://github.com/learningequality/kolibri/issues/5020

indirectlylit (approved):

Issue #4714, opened 2019-01-11
Button name on tablet


swapnil106111:
### Observed behavior
While using learn tab on tablet following things are happened on tablet:
1. CHECK button  shown as CHE...
2. NEXT button shown as NE.

Able to see expected names on Lenovo tab but not on swipe tab.

Lenovo tab model number : TAB3 7 ESSENTIAL
Swipe tab model number : W7

### Expected behavior
It must show whole name on every device.

### Steps to reproduce
1. Go to the learn tab and use on swipe tablet.

### Context
 * Kolibri version : Kolibri 0.11.0
 * Operating system : Ubuntu 14.04
 * Browser : Chrome

### Screenshot

Swipe tablet:
![img_20190107_143723](https://user-images.githubusercontent.com/12776071/51016053-d38bac00-1594-11e9-840d-ed216325b9e3.jpg)

Lenovo tablet:
![img_20190111_113633](https://user-images.githubusercontent.com/12776071/51016179-56146b80-1595-11e9-8b87-754e98838c05.jpg)

rtibbles:
Fixed in #5004

PR #5004, opened 2019-02-11
Remove text ellipsis on overflow in buttons.

rtibbles:
### Summary
Simplest fix for issue reported here: #4714

Without a much more complex JS driven ellipsis algorithm, not sure how else to address it.

### Reviewer guidance
Is this going to cause trouble elsewhere in the app?

### References
#4714

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/core/assets/src/views/buttons-and-links/buttons.scss
+++ b/kolibri/core/assets/src/views/buttons-and-links/buttons.scss
@@ -17,7 +17,6 @@ $raised-shadow: 0 1px 5px rgba(0, 0, 0, 0.2), 0 2px 2px rgba(0, 0, 0, 0.14),
   line-height: 36px;
   text-align: center;
   text-decoration: none;
-  text-overflow: ellipsis;
   text-transform: uppercase;
   white-space: nowrap;
   cursor: pointer;


jonboiser (approved):

indirectlylit:
Something seems a little fishy about this. Probably OK.

The overflow ellipsis should only be shown when the text ... overflows. For some reason the button doesn't seem to be expanding as much as it should:

image

rtibbles:
Yeah, the specificity of the behaviour to a particular tablet type did cause me some concern. At the same time, I'm not sure that the css text ellipsis is particularly helpful on most buttons, considering our tendency to have very short text in buttons. In the screenshot above, there is space for the button text, but it's showing three dots instead.

I think it would be better for KButton to handle very long text a bit more intelligently, but for short text, let things clip, even if it looks ugly, for the sake of better user comprehension.

indirectlylit:
I tried manually adding extra text to the button and it expands as expected, pushing the checkboxes to the right. Agreed that showing the ellipsis will be rare

My guess is that they're overflowing here due to some subtle interaction between the  button label and its surrounding border

PR #5006, opened 2019-02-11
Allow setting of scroll position based on browser history state.

### LINKED ISSUE #3659: scroll position/state is not saved when navigating
### Observed behavior

scroll state is not saved when browsing back/forward using vue router

As [noted](https://github.com/learningequality/kolibri/pull/3636#issuecomment-388961868) by @jonboiser:

> `savedPosition` is always `{x:0, y:0}` when using back/forward button. This might have something to do with our unique usage of vue-router. Maybe if we used `router-view`, `savedPosition` would be correct

### Expected behavior

Scroll state should be preserved.

### User-facing consequences

extra scrolling and lost context

### Steps to reproduce

* going to Learn
* scrolling down on 'recommended'
* clicking something
* clicking back

I would expect it to take me to the scrolled-down location, but it goes to the top of recommended

### Context

kolibri 0.9.0

rtibbles:
### Summary
Monitors CoreBase scrolling, and uses that information to update window.pageYOffset to update the history state.
Sets a vuex state element to indicate the desired starting scroll position for the core base - this might be better implemented as an event, but this was easier.
CoreBase adds a watcher to this vuex state in the created property to set the scrollTop to this value when the router has set it.

### Reviewer guidance
Does it provide the intended scrolling behaviour?
Could this be implemented in a better way? In what way might this be brittle?

scrollmedo

### References
Fixes #3659

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated


ralphiee22 (dismissed):
Works great for coach pages, but not really for other pages. But hey, some working is better than none working?!

jonboiser:
Is the scroll position supposed to be unique for each page? The current behavior seems to apply the same position to the previous page even if the position was set at an even earlier page. For example

1. On Page A I go down 200 pixels, then click a link to Page B.
1. On Page B I do not scroll, but click into Page C
1. At Page C, I go back to Page B (where I did not scroll). On Page B, I am now down 200 pixels.
1. I go back to Page A and have a Y position of 0 pixels (when it was 200 at the first step)

rtibbles:
Hrm, good point - need to clear the vuex data when leaving a page, so that it is only set from the history state.

rtibbles:
Or rather, not the vuex state, but the window offset value - if you navigate to pages without ever scrolling, that doesn't get reset.

kolibri/core/assets/src/views/CoreBase/index.vue
```diff
@@ -333,16 +334,54 @@
...
+        if (this.unwatchScrollHeight) {
+          this.unwatchScrollHeight();
+        }
+        if (this.loading) {
+          // Don't set scroll position until the main content
+          // of coreBase is shown in the DOM.
+          // Create a watcher to monitor changes in loading
+          // to try to set the scrollHeight after the contents
+          // have loaded.
+          this.unwatchScrollHeight = this.$watch('loading', () => {
```
indirectlylit, line 361: I wonder if `unwatchScrollHeight` should be defined on `data` and initialized to `undefined` for explicitness? Seems a little weird to just slap it onto the view model.

indirectlylit (approved):
added 1a27eb6b65a15b93db8479f138679acccde85758 to your PR

kolibri/core/assets/src/views/CoreBase/index.vue
rtibbles: Why? It doesn't need to be reactive, putting it on the view itself is exactly the right place for it, the only thing that might be better is prefixing it with an underscore, but I really don't see any benefit to having it be a watched bit of internal state.

indirectlylit: Yeah agreed – I don't like the fact that it's reactive either. The thing that bothered me is to add it to the `vm` namespace without declaring it anywhere.

There's a [lively discussion here](https://github.com/vuejs/vue/issues/1988) so apparently this is something that keeps people up at night :)

Looks like the original strategy you used is likely fine, and another common pattern is to put it inside `$options`

rtibbles: I do find it a little odd that the github issue is so long. Either you put it in a namespace, or you put it right on the object. Vue doesn't, and shouldn't care!

PR #5116, opened 2019-02-20
Don't check ports if CherryPy is  not going to start

jredrejo:
### Summary
In version 0.12 we have added the option CHERRYPY_START to disable CherryPY server when starting kolibri.
If this option is not enabled we should not check available ports, thus not raising an error if the port selected to run kolibri is already open (when using kolibri-server it will be open by nginx)

### Reviewer guidance
Simple fix looking at the code 

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/utils/cli.py
+++ b/kolibri/utils/cli.py
@@ -617,7 +617,8 @@ def main(args=None):  # noqa: max-complexity=13
 
     if arguments['start']:
         port = _get_port(arguments['--port'])
-        check_other_kolibri_running(port)
+        if OPTIONS["Server"]["CHERRYPY_START"]:
+            check_other_kolibri_running(port)
 
     try:
         initialize(debug=debug)


kolibri/utils/cli.py
```diff
@@ -617,7 +617,8 @@ def main(args=None):  # noqa: max-complexity=13
 
     if arguments['start']:
         port = _get_port(arguments['--port'])
-        check_other_kolibri_running(port)
+        if OPTIONS["Server"]["CHERRYPY_START"]:
+            check_other_kolibri_running(port)
```
rtibbles, line 621: Makes sense - I think I had left it assuming that this would happen before something else occupied the port, but this makes more sense!

rtibbles (approved):

Issue #4973, opened 2019-02-08
Impossible to close the privacy modal with keyboard


radinamatic:
### Observed behavior

When navigating by keyboard user is prevented to close the *Usage and privacy* modal: even when the <kbd>Close</kbd> button is in focus, and the user presses the <kbd>Enter</kbd> key, the modal keeps reopening, until the user closes by mouse click (if they can).

#### Chrome
![modal-close-keyboard-chrome](https://user-images.githubusercontent.com/1457929/52496754-24aab200-2bcc-11e9-8226-7eab54b1168f.gif)

#### Firefox
![modal-close-keyboard-ff](https://user-images.githubusercontent.com/1457929/52496760-2a07fc80-2bcc-11e9-8cd6-7c96313a1672.gif)

*Note the additional focusable element in Firefox (the whole field)* - **should get the `focusable=false` attribute.**

### Expected behavior

Modal should be fully under control by keyboard only.

### User-facing consequences

Being blocked and unable to close the modal.

### Errors and logs
--

### Steps to reproduce

Navigate the *Usage and privacy* modal by keyboard and try to close

### Context

 * Kolibri version: 0.12.0.b3
 * Operating system: Windows 7
 * Browser: Chrome & Firefox

indirectlylit:
(removing performance tag - I think that's typically reserved for 'slowness')

radinamatic:
> performance tag typically reserved for 'slowness'

Noted! 👍

jonboiser:
#4981

PR #5121, opened 2019-02-20
Manually blur text-link-style KButtons when they are clicked or enter-keyed

### LINKED ISSUE #5103: On (at least) one modal, dismissing with enter keypress causes it to come back
### Observed behavior

This was seen with the privacy modal, and is in other places now:

1. Facility > Class Details > Rename Class
1. Privacy link on sign-in page
1. Privacy link on sign-up page
1. Profile page, change password modal
1. "Change" status button for Lessons

It seems to be for the same reasons as #4973 (privacy modal), the "Edit" button maintains focus while the modal is on, so it gets "clicked" again when you dismiss the modal with an Enter keypress.

The common thing between these two cases is that the privacy link and rename use the text-link versions of the Button.

### Expected behavior

Text-type links do not have this side-effect when "clicked" by an enter-keypress

### Context

Kolibri 0.12 latest

jonboiser:
### Summary

Fixes #5103 and the root cause of #4973 by manually blurring the component when it is clicked or enter-keyed. This prevents the link/button from being re-clicked when the user dismisses/submits a modal by hitting Enter (but has the other, hopefully less noticeable side effect of removing focus from that link, so the user will need to refocus it if a keyboard-only user).

### Reviewer guidance

In #5103, there is a list of all the pages where there was a problem with dismissing a modal by keyboard. Go to all of them and test that

1. If you click the link and dismiss the modal with a keyboard, the modal does not re-appear
1. If you "click" the link via keyboard, it''s the same.

And, until you are satisfied, check that other k-button text links that do something other than open a modal still work properly (if they exist).
### References

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/core/assets/src/views/SideNav.vue
+++ b/kolibri/core/assets/src/views/SideNav.vue
@@ -201,9 +201,6 @@
       },
       handleClickPrivacyLink() {
         this.privacyModalVisible = true;
-        // HACK to fix https://github.com/learningequality/kolibri/issues/4973.
-        // For some reason, the button maintains focus even when the modal is opened.
-        this.$refs.privacyLink.$el.blur();
       },
       compareMenuComponents(navComponentA, navComponentB) {
         // Compare menu items to allow sorting by the following priority:
--- a/kolibri/core/assets/src/views/buttons-and-links/KButton.vue
+++ b/kolibri/core/assets/src/views/buttons-and-links/KButton.vue
@@ -9,7 +9,7 @@
     :disabled="disabled"
     tabindex="0"
     @click="handleClick"
-    @keyup.enter.stop.prevent="handlePress"
+    @keyup.enter.stop.prevent="handelPressEnter"
   >
     <slot v-if="$slots.default"></slot>
     <template v-else>{{ text }}</template>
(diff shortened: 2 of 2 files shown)


kolibri/core/assets/src/views/SideNav.vue
```diff
@@ -201,9 +201,6 @@
       },
       handleClickPrivacyLink() {
         this.privacyModalVisible = true;
-        // HACK to fix https://github.com/learningequality/kolibri/issues/4973.
-        // For some reason, the button maintains focus even when the modal is opened.
```
indirectlylit, line 205: is this condition failing?

https://github.com/learningequality/kolibri/blob/develop/kolibri/core/assets/src/views/KModal.vue#L201

jonboiser: I'm not sure. But even if you put the focus on the modal but clicking or tabbing on to it, then close it, you still get the same behavior only when the modal-opening button is an anchor tag (so focus is actually in both places). Might be specific just to `<a>`.

jonboiser: Yeah, I just checked it and it does enter into the focusModal branch when you click them (and this is without this PR).

jonboiser: Yeah, I checked this with the Class Rename button, and it does not enter into the focusModal branch because, I guess the active element is still that link.

For the side-nav privacy link, it _does_ enter the focusModal branch because in that case the link was blurred in the hack we deleted here.

I tried removing the second condition so that focusModal gets called unconditionally, and we still get the same behavior.

indirectlylit (approved):

PR #5130, opened 2019-02-21
Fix detail views

### LINKED ISSUE #5052: inconsistency between difficult question summary and detail reports
### Observed behavior

| summary | detail |
|--|--|
|  "Question 1" | "Repeated addition (1)" |
| "1 of 1 needs help" | three users needing help |
| [Link to page]( http://kolibribeta.learningequality.org/coach/#/4707894249cbe44b7fd142bdb87fe90e/reports/lessons/77c3daea1dcadbdb2f1b8e0a9d278a81/exercises/52afd859b8df5bb1a6a834ddaeaf3f23/questions) | [Link to page](http://kolibribeta.learningequality.org/coach/#/4707894249cbe44b7fd142bdb87fe90e/reports/lessons/77c3daea1dcadbdb2f1b8e0a9d278a81/exercises/52afd859b8df5bb1a6a834ddaeaf3f23/questions/58e624d3713f553fb26c156885e344aa/learners/db6a72b70250db5ed2d69955eece98bd/interactions/0) | 
| image | image |

### Expected behavior

Table summary and detail page should be consistent

### User-facing consequences

possible confusion

### Steps to reproduce

see links to beta site above

### Context

0.12.0 beta 4

----

cc @rtibbles

rtibbles:
### Summary
Detail views for difficult questions and for exercises were not working correctly.
In addition, the list views for difficult questions were not being appropriately filtered based on the context.
This PR fixes this and other related behaviour in several ways.

Commit 1) Only calls a force fetch on the root redirect to the exercise detail view to prevent requesting the same attemptlogs twice.
Commit 2) Only calls a force fetch on the root redirect for the difficult question detail view to prevent requesting the same attemptlogs twice.
Commit 3) Fixes the filtering for the API endpoints that provide statistics for the difficult questions list pages.
Commit 4) Resets the attemptLogMap and the learnerMap state objects on every page navigation, to prevent navigation history from causing contamination of the displayed attempts or learners.
Commit 5) Fixes two issues were the incorrect params were being passed, or not being passed, between handlers, actions, and fetches from the backend.
Commit 6) Adds in additional handling for quiz difficult question lists, to show questions that *no-one* answered, as these are a _difficult_ question, but the backend returns no stats indicating such.
Commit 7) Filters out correct answers after we have set 'no attempt' for quiz detail question views, to prevent correct answers being accidentally infilled as 'no attempt'.
Commit 8) Update the way that we generate question titles, by relying on the `counterInExercise` property set in the exercise data. Update the v0 to v1 mapping function to provide this data also.

### Reviewer guidance
Run through the gherkin stories for difficult questions and for the exercise detail views.

### References
Fixes #5052 

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/core/assets/src/exams/utils.js
+++ b/kolibri/core/assets/src/exams/utils.js
@@ -61,6 +61,10 @@ function convertExamQuestionSourcesV0V1(questionSources, seed, questionIds) {
     exercise_id: question.exercise_id,
     question_id: shuffledExerciseQuestions[question.exercise_id][question.questionNumber],
     title: question.title,
+    counterInExercise:
+      questionIds[question.exercise_id].findIndex(
+        id => id === shuffledExerciseQuestions[question.exercise_id][question.questionNumber]
+      ) + 1,
   }));
 }
 
--- a/kolibri/plugins/coach/api.py
+++ b/kolibri/plugins/coach/api.py
@@ -17,6 +17,7 @@ from kolibri.core.auth.models import Collection
 from kolibri.core.auth.models import FacilityUser
 from kolibri.core.auth.models import LearnerGroup
 from kolibri.core.decorators import query_params_required
+from kolibri.core.exams.models import Exam
 from kolibri.core.lessons.models import Lesson
 from kolibri.core.logger.models import AttemptLog
 from kolibri.core.logger.models import ExamAttemptLog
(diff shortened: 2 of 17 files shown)


kolibri/plugins/coach/assets/src/modules/exerciseDetail/actions.js
```diff
@@ -1,14 +1,15 @@
...
+export function setAttemptLogs(store) {
+  const { learnerId, exercise, attemptId } = store.state;
   return AttemptLogResource.fetchCollection({
     getParams: {
       user: learnerId,
       content: exercise.content_id,
     },
-    force: true,
+    // Only force when accessing the root url
+    force: attemptId === null,
```
indirectlylit, line 12: would this depend in part on where they're coming _from_? e.g. would a direct link to a non-root URL from outside kolibri not work?

kolibri/plugins/coach/assets/src/modules/exerciseDetail/handlers.js
```diff
@@ -47,18 +47,13 @@ function showExerciseDetailView({ learnerId, exerciseId, attemptId = null, inter
...
-          // to the first attempt.
-          if (attemptId === null) {
-            return attemptLogs[0].id;
-          }
-        });
+      return store.dispatch('exerciseDetail/setAttemptLogs').then(attemptLogs => {
+        // No attemptId was passed in, so we should trigger a url redirect
+        // to the first attempt.
+        if (attemptId === null) {
+          return attemptLogs[0].id;
```
indirectlylit, line 54: we may want to explicitly check that there is at least one item here and bail if not

kolibri/plugins/coach/assets/src/modules/questionDetail/actions.js
```diff
@@ -19,7 +20,7 @@ export function setLearners(store, { questionId, exerciseId, quizId, classId, gr
   return resource
     .fetchCollection({
       getParams,
-      force: true,
+      force: !learnerId,
```
indirectlylit, line 23: same Q as before - would this break direct links to non-root pages?

kolibri/plugins/coach/api.py
```diff
@@ -188,57 +189,117 @@ def list(self, request, *args, **kwargs):
...
+                        target_user=F("user"),
+                    )
+                queryset = queryset.distinct()
+        if group_id is not None:
+            collection_id = group_id or classroom_id
+            queryset = HierarchyRelationsFilter(queryset).filter_by_hierarchy(
+                ancestor_collection=collection_id,
+                target_user=F("user"),
+            )
+
```
indirectlylit, line 258: Lots going on here that I don't understand! but it looks tested so I trust it

kolibri/plugins/coach/assets/src/modules/questionList/actions.js
```diff
@@ -35,11 +36,20 @@ export function setItemStats(store, { classId, exerciseId, quizId, lessonId, gro
...
-        question_id: stat.item,
-      }));
+      // If no one attempted one of the questions, it could get missed out of the list
+      // of difficult questions, so use the exam data to fill in the blanks here.
+      stats = item.question_sources.map(source => {
+        const stat = stats.find(
+          stat => stat.item === source.question_id && stat.content_id === source.exercise_id
+        ) || {
+          correct: 0,
+          total: (stats[0] || {}).total || 0,
```
indirectlylit, line 48: do we need to check `stats.length`?

kolibri/core/assets/src/exams/utils.js
```diff
@@ -61,6 +61,10 @@ function convertExamQuestionSourcesV0V1(questionSources, seed, questionIds) {
     exercise_id: question.exercise_id,
     question_id: shuffledExerciseQuestions[question.exercise_id][question.questionNumber],
     title: question.title,
+    counterInExercise:
+      questionIds[question.exercise_id].findIndex(
+        id => id === shuffledExerciseQuestions[question.exercise_id][question.questionNumber]
+      ) + 1,
```
indirectlylit, line 67: thanks

indirectlylit (approved):
haven't tested the scenarios, but reviewed the code. just a couple questions but looks good

kolibri/plugins/coach/assets/src/modules/exerciseDetail/actions.js
rtibbles: It would still work, it just wouldn't force the fetch. If someone refreshed the page, or pasted a link, it would still do the fetch.

kolibri/plugins/coach/assets/src/modules/exerciseDetail/handlers.js
rtibbles: Can check - any idea of what the graceful fallback should be?

kolibri/plugins/coach/assets/src/modules/questionDetail/actions.js
rtibbles: Nope!

kolibri/plugins/coach/api.py
rtibbles: It's the only reason that I do.

kolibri/plugins/coach/assets/src/modules/questionList/actions.js
rtibbles: This is essentially doing that - if `stats[0]` is undefined, then either the Array is malformed, or it is empty.

kolibri/plugins/coach/assets/src/modules/exerciseDetail/handlers.js
rtibbles: Also, with @ralphiee22's PR to set started only when exercises have attempts, this should only happen if someone manually constructs a bad URL, in which case I think it'll just show the error page. This may be sufficient?

radinamatic:
Error while viewing the difficult questions tab:

difq

And nothing available to c/p for the error details:



This is with a fresh DB and no previous interactions whatsoever.

rtibbles:
> Error while viewing the difficult questions tab

Was this for a lesson or a quiz?

radinamatic:
Lesson

radinamatic:
Second learner gets the answer wrong (look for the weird `NaN` in the question number):

difq2

kolibri/plugins/coach/assets/src/modules/exerciseDetail/handlers.js
indirectlylit: yeah sounds good

kolibri/plugins/coach/assets/src/modules/questionList/actions.js
indirectlylit: ah I see

radinamatic:
Reviewing difficult questions in quizzes does not trigger any errors, and everything seems correct! 👍 

difq3

indirectlylit:
@radinamatic I think it'd make sense to merge this in as-is for beta9 and create follow-up issues. thoughts?

radinamatic:
Agreed @indirectlylit , let's retest in beta9 👍

indirectlylit:
follow-up issue here: https://github.com/learningequality/kolibri/issues/5133

@radinamatic mind clarifying?

Issue #5158, opened 2019-02-26
Learner last activity report display invalid date


mrpau-richard:
### Observed behavior
At the coach reports page I have an `invalid date` record on a learner. This might related to https://github.com/learningequality/kolibri/issues/5113#issuecomment-465900520

### Expected behavior
Display the exact date/time. 

### User-facing consequences

…

### Errors and logs
![screen shot 2019-02-26 at 6 54 51 pm](https://user-images.githubusercontent.com/8663934/53408346-35935a00-39f9-11e9-95aa-329f75e118d6.png)
![screen shot 2019-02-26 at 7 05 02 pm](https://user-images.githubusercontent.com/8663934/53408459-768b6e80-39f9-11e9-914f-609a96d52dbe.png)

### Context

Tell us about your environment, including:
 * Kolibri version: 0.12.0-beta9
 * Operating system: macOS 10.13.3
 * Browser: Safari 11.0.3

indirectlylit:
maybe related https://github.com/learningequality/kolibri/issues/5079

rtibbles:
This should be fixed by #5918.

Issue #5113, opened 2019-02-20
Class activity  shows nothing.


swapnil106111:
### Observed behavior
We have observed that class activity shows nothing in class home everytime.

### Expected behavior
It must show class activity.

### Steps to reproduce
1. Click on Class Home in coach report.

### Context:
 * Kolibri version : Kolibri 0.12.0b4
 * Operating system : Ubuntu 14.04
 * Browser : Chrome

### Screenshot
![1](https://user-images.githubusercontent.com/12776071/53088718-940b9480-3530-11e9-9515-fb6921312efb.png)

indirectlylit:
please test in the latest beta

mrpau-julius:
Also in Kolibri latest beta version 7 & 8, class activity are empty and not automatically notify. 
* Browser : Firefox 65.0(64-bit)
![screenshot from 2019-02-21 15 25 11](https://user-images.githubusercontent.com/38421180/53151292-cdeea080-35ed-11e9-9615-1298a55908fa.png)

mrpau-richard:
Same issue on Safari 11.0.3 using Kolibri beta7
![screen shot 2019-02-21 at 4 02 19 pm](https://user-images.githubusercontent.com/8663934/53153049-2c1d8280-35f2-11e9-8b8f-bdc95d6991ee.png)

indirectlylit:
thank you!

I mis-tagged beta 8, but the beta 7 result is concerning

FYI beta 8b tagged here should be correct: https://github.com/learningequality/kolibri/releases/tag/v0.12.0-beta8b

indirectlylit:
I'm still not reproducing. you're seeing this in beta 9 still on all browsers?

rtibbles:
Please let us know if this is still extant for you - concerning if so.

indirectlylit:
Reproduced some weird behaviors, specifically with exercises. Unsure if it's what this issue is about specifically, but perhaps related:

https://drive.google.com/file/d/1g7Wd7NDf92DnyJh64g_Ui3NddW77c_xP/view?usp=sharing

**Scenario above:**

* `[0:00]` learner on the left, coach with network tab on the right
* `[0:17]` press the 'check' button which should write an activity log and mark it as 'started'
* `[0:17]-[0:37]` looking for any indication of notifications, none found
* `[0:47]` refresh, wait. notification still doesn't show up so apparently it's not in the table yet
* `[1:30]` add another attempt log. still doesn't show up
[...]

Issue #5079, opened 2019-02-16
'last activity' can be in the future


indirectlylit:
### Observed behavior

Noticed that the 'last activity' column very briefly said "in 3 seconds", indicating a negative value:

![image](https://user-images.githubusercontent.com/2367265/52892752-7615ec80-314a-11e9-9cb6-bfeac5533f0b.png)

### Expected behavior

We should determine how it happened that the value went negative and prevent it.

One thing to be particularly careful of is that we're always comparing server time to server time, and never server time to client time because they could be very different. This bug might be due to that.

Let's also add a gherkin story that explicitly tests various parts of the app when the server and client times are different. (cc @radinamatic @mrpau-dev)

### User-facing consequences

If server and client times are very different, values in this column may be days or even years off

### Steps to reproduce

unknown, but possibly try changing your local computer time and then browsing the beta server (assuming the theory above is correct)

### Context

0.12.0 beta 6

indirectlylit:
let's just check for worst-case scenario

jonboiser:
I have just seen, in a single session of the same user one resource being something reasonable like "10 min ago", and then another resource (same session, same user) have something in the future like "in 15 minutes", and the videos were viewed within minutes of each other.

But then I logged out/in and the next video had an accurate-looking last activity time.

jredrejo:
This problem seems to have the same origin than #5113

rtibbles:
This may be partially resolved by #5198, but I am doing some follow up work.

rtibbles:
This is caused by a bug in the server time setting code. To replicate this follow these steps:
1) Login.
2) Walk away for ten minutes (or more) leaving the browser tab open - perhaps have a nice soothing cup of tea?
3) Come back.
4) Do anything with any content.
5) Observe the last activity to be ~`<however long you went away for>` in the future.

rtibbles:
Fixed in #5210.

PR #5198, opened 2019-02-26
pass timestamps to notifications

indirectlylit:
### Summary

This attempts to address #5158 and #5079 by explicitly passing timestamps into the notifications rather than generating them at the time the notification object is created

### Reviewer guidance

notification objects are actually created a while after the event occurs, leading to discrepancies between attempt logs and notification history

### References

While working on this, I noticed some very peculiar behaviors in the way that exercise 'started' notifications are (or are not) being sent.

see https://github.com/learningequality/kolibri/issues/5113#issuecomment-467643676

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/core/logger/serializers.py
+++ b/kolibri/core/logger/serializers.py
@@ -111,6 +111,12 @@ class AttemptLogSerializer(KolibriModelSerializer):
                   'end_timestamp', 'completion_timestamp', 'item', 'time_spent', 'user',
                   'complete', 'correct', 'hinted', 'answer', 'simple_answer', 'interaction_history', 'error')
 
+    def create(self, validated_data):
+        instance = super(AttemptLogSerializer, self).create(validated_data)
+        # to check if a notification must be created:
+        wrap_to_save_queue(parse_attemptslog, instance)
+        return instance
+
     def update(self, instance, validated_data):
         instance = super(AttemptLogSerializer, self).update(instance, validated_data)
         # to check if a notification must be created:
--- a/kolibri/core/logger/serializers.py
+++ b/kolibri/core/logger/serializers.py
@@ -20,6 +20,7 @@ from kolibri.core.notifications.api import parse_examlog
 from kolibri.core.notifications.api import parse_summarylog
 from kolibri.core.notifications.tasks import wrap_to_save_queue
 from kolibri.core.serializers import KolibriModelSerializer
+from kolibri.utils.time_utils import local_now
 
 
 class ContentSessionLogSerializer(KolibriModelSerializer):
@@ -59,13 +60,13 @@ class ExamLogSerializer(KolibriModelSerializer):
             instance.completion_timestamp = now()
         instance = super(ExamLogSerializer, self).update(instance, validated_data)
         # to check if a notification must be created:
-        wrap_to_save_queue(parse_examlog, instance)
+        wrap_to_save_queue(parse_examlog, instance, local_now())
         return instance
 
     def create(self, validated_data):
         instance = super(ExamLogSerializer, self).create(validated_data)
         # to check if a notification must be created:
-        wrap_to_save_queue(create_examlog, instance)
+        wrap_to_save_queue(create_examlog, instance, local_now())
         return instance
 
 
--- a/kolibri/core/notifications/api.py
+++ b/kolibri/core/notifications/api.py
@@ -337,6 +337,6 @@ def parse_attemptslog(attemptlog):
                                                lesson.group_or_classroom,
                                                lesson_id=lesson.id,
                                                contentnode_id=contentnode_id,
-                                               reason=HelpReason.Multiple)
+                                               timestamp=attemptlog.start_timestamp)
             notifications.append(notification)
         save_notifications(notifications)
--- a/kolibri/core/notifications/api.py
+++ b/kolibri/core/notifications/api.py
@@ -194,11 +194,6 @@ def parse_summarylog(summarylog):
     Lesson Completed notification.
     """
 
-    # updating an exercise summary log might mean they made their first attempt in the exercise/lesson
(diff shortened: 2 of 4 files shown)


indirectlylit:
cc @radinamatic @mrpau-dev this will require testing the notification scenarios

rtibbles:
Two test failures:
```
_________________ NotificationsAPITestCase.test_create_examlog _________________
self = <kolibri.core.notifications.test.test_api.NotificationsAPITestCase testMethod=test_create_examlog>
save_notifications = <MagicMock name='save_notifications' id='139907087624784'>
    @patch('kolibri.core.notifications.api.save_notifications')
    def test_create_examlog(self, save_notifications):
        examlog = ExamLog.objects.create(exam=self.exam, user=self.user1)
>       create_examlog(examlog)
E       TypeError: create_examlog() takes exactly 2 arguments (1 given)
kolibri/core/notifications/test/test_api.py:199: TypeError
----------------------------- Captured stdout call -----------------------------
Installed 2 object(s) from 1 fixture(s)
_________________ NotificationsAPITestCase.test_parse_examlog __________________
self = <kolibri.core.notifications.test.test_api.NotificationsAPITestCase testMethod=test_parse_examlog>
save_notifications = <MagicMock name='save_notifications' id='139907089515152'>
    @patch('kolibri.core.notifications.api.save_notifications')
    def test_parse_examlog(self, save_notifications):
        examlog = ExamLog.objects.create(exam=self.exam, user=self.user1)
>       parse_examlog(examlog)
E       TypeError: parse_examlog() takes exactly 2 arguments (1 given)
kolibri/core/notifications/test/test_api.py:187: TypeError
----------------------------- Captured stdout call -----------------------------
Installed 2 object(s) from 1 fixture(s)
```

indirectlylit:
there's definitely more going on here

here's a notification which is sending a timestamp _in the future_:

image

indirectlylit:
Some minor notes:

* the 'exercise started' notifications are being created after two minutes when the summary log is updated, regardless of whether there was an attempt. This should have a conditional and only trigger 'started' in the case of an attempt
* conversely, the 'exercise started' notifications are still _not_ being created when an attempt is made. I cannot reproduce the fix from 54d621a4f9d318fe370b7d008ef8ca2880a4903a
* as shown in the screenshot above, sometimes timestamps are in the future. Haven't investigated
* it appears the polling for notifications will cease and not re-started if there's a network error. refresh is necessary to resume polling:
    image

rtibbles:
> * the 'exercise started' notifications are being created after two minutes when the summary log is updated, regardless of whether there was an attempt. This should have a conditional and only trigger 'started' in the case of an attempt

Have fixed this above, so that summarylog updates no longer trigger started notifications for exercises.

> * conversely, the 'exercise started' notifications are still _not_ being created when an attempt is made. I cannot reproduce the fix from [54d621a](https://github.com/learningequality/kolibri/commit/54d621a4f9d318fe370b7d008ef8ca2880a4903a)

Have refined the logic and added additional tests here to ensure this is happening.

> * as shown in the screenshot above, sometimes timestamps are in the future. Haven't investigated

Have not replicated this yet, will continue to investigate.

> * it appears the polling for notifications will cease and not re-started if there's a network error. refresh is necessary to resume polling

Have added catches for polling errors and let polling continue when they do happen.

kolibri/core/logger/serializers.py
```diff
@@ -59,13 +60,13 @@ def update(self, instance, validated_data):
...
         # to check if a notification must be created:
-        wrap_to_save_queue(parse_examlog, instance)
+        wrap_to_save_queue(parse_examlog, instance, local_now())
         return instance
 
     def create(self, validated_data):
         instance = super(ExamLogSerializer, self).create(validated_data)
         # to check if a notification must be created:
-        wrap_to_save_queue(create_examlog, instance)
+        wrap_to_save_queue(create_examlog, instance, local_now())
```
jredrejo, line 69: So, the problem came from the fact the queue saving the notification can have delays, thus timestamp=local_now is not the logger timestamp, but the 'saving notification' timestamp. If that's right this PR makes totally sense.

kolibri/core/notifications/api.py
```diff
@@ -286,32 +285,50 @@ def parse_attemptslog(attemptlog):
...
+
     # More than 3 errors in this mastery log:
     needs_help = len(failed_interactions) > 3
 
-    if needs_help:
-        notifications = []
-        for lesson, contentnode_id in lessons:
+    notifications = []
+    for lesson, contentnode_id in lessons:
+        if needs_help:
```
jredrejo, line 303: why have the needs_help moved to check every lesson? I can't think  how needs_help can affect only to one lesson if the needs come by the attemptlog.

jredrejo (changes requested):

kolibri/core/notifications/api.py
rtibbles: Because I am also now doing the 'started' notification checking for exercises only during the attempt log parsing too. So, the behaviour is the same for the needs help behaviour (the tests still pass!), it's just that we use the same loop to also check the started notifications too.

kolibri/core/logger/serializers.py
rtibbles: Yeah, I think that was definitely causing some issues with timestamping.

jredrejo (approved):

rtibbles:
Have updated this to now do the initial check on attemptlog creation as well as update, which I think is why @indirectlylit was not seeing the intended behaviour. I tested this again, and it seemed to work as intended.

rtibbles:
I have not seen the timestamps in the future issue, but will attempt to replicate that outside the scope of this PR.

PR #5200, opened 2019-02-27
Set on save, not before.

rtibbles:
### Summary
Fixes a subtle bug whereby we would set data on a model before saving. If the save was rejected, this would not roll back.

Fixes this by only setting on the model after the save is successful.

### Reviewer guidance
I have written a test for this behaviour, but one place in the UI is in the user profile:
1. Try to change your username to a username that already exists on the server.
2. Press save.
3. Be told that this is not a valid username.
4. Change Full name to another name.
5. Change username back to original username.
6. Save again.
7. Prior to this fix it would give you the validation error, because the original, invalid username would still be being sent to the client. After this fix, it works as expected.

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/core/assets/src/api-resource.js
+++ b/kolibri/core/assets/src/api-resource.js
@@ -121,10 +121,11 @@ export class Model {
                 payload[key] = attrs[key];
               }
             });
-            this.set(payload);
           } else {
-            this.set(attrs);
-            payload = this.attributes;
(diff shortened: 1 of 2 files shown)


ralphiee22 (approved):
Fixes what it claims.

Issue #2371, opened 2017-10-07
Allow internationalization of known content licenses


rtibbles:
## Summary

Currently all content licenses are hard coded into the database. It would be good to be able to translate known content licenses into the interface language.

radinamatic:
Bumping up this, as it is still extant on 0.11:

#### Learn

![ubuntu 16 04 3 0 11b1 in-context running - oracle vm virtualbox_045](https://user-images.githubusercontent.com/1457929/46926539-7afa5080-d031-11e8-92fb-f32cd3452a45.png)


#### Coach > Preview Lesson resource

![workspace 1_044](https://user-images.githubusercontent.com/1457929/46926549-86e61280-d031-11e8-98f3-67e5621e9f2b.png)

(Crowdin window behind, I was looking for the string making sure I did not miss the translation, before finding this old issue)

indirectlylit:
unfortunately we'll probably have to wait for 0.12 on this since it requires new strings

radinamatic:
Agreed, but there was no `0.12` milestone label yet.

indirectlylit:
true :)

indirectlylit:
todo: look into short-term possibilities for arabic and spanish

indirectlylit:
removing from 0.12 in lieu of https://github.com/learningequality/kolibri/issues/5148

indirectlylit:
refs:

* https://github.com/learningequality/kolibri/issues/5569
* https://github.com/learningequality/kolibri/pull/5205

PR #5205, opened 2019-02-27
Adds some but not all of the creative commons translations

### LINKED ISSUE #5148: hard-code and conditionally display arabic translations of CC licenses
### Observed behavior

we want to do https://github.com/learningequality/kolibri/issues/2371 but doing this through the standard crowdin flow is too time-consuming and risky this close to a release

however there is a strong need to have at least the arabic licenses translated:

> In order to make sure users use the content appropriately (the idea of licensing is not very clear to many [Middle East and North Africa] contexts) and to make sure we are doing our best to monitor that, I would push for having it translated

### Expected behavior

see https://creativecommons.org/licenses/?lang=ar for target licenses. Contact Hiba for assistance if necessary

Recommend we hard-code these into a JSON file and then conditionally display them when Arabic is the selected language

### User-facing consequences

Arabic speaker who are not familiar with western content licensing might not understand the usage terms of the content

### Context

0.12.0

indirectlylit:
### Summary

hard-codes CC license translations into the code-base, based on https://creativecommons.org/licenses/

Adds:

* arabic
* english (official from CC)
* french
* spanish
* brazilian portuguese

| coach | learner |
|--|--|
|image | image |

### Reviewer guidance

This updates the learner content view and the coach lesson management view. Those are the only places I could find license display.

Some notes:

* current implementation embeds full descriptions license descriptions and names within the content, which seems possibly incorrect
* current implementation doesn't distinguish license names vs license IDs, but CC does (e.g. CC BY-NC vs Attribution-NonCommercial)
* this PR does nothing to address non-CC licenses or placeholders like "special permission to use in kolibri"

### References

issues:

* fixes #5148
* refs #2371 but does not fix it

code:

* https://github.com/learningequality/le-utils/blob/master/le_utils/constants/licenses.py#L10
* https://github.com/learningequality/le-utils/blob/master/le_utils/resources/licenselookup.json#L1

internal docs:

* https://drive.google.com/file/d/1FgwHiDY4w2_70tan966lizdbL4yksDPr/view
* https://drive.google.com/file/d/1Fx_6SrglhbHOKajtWU1tzYj2r_WrtJr1/view

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/core/assets/src/utils/licenseTranslations.js
+++ b/kolibri/core/assets/src/utils/licenseTranslations.js
@@ -1,0 +1,166 @@
+// except where specified, these are from
+// https://creativecommons.org/licenses/?lang=ar
+
+export const licenseTranslations = {
+  ar: {
+    'CC BY': {
+      name: 'نَسب المُصنَّف',
+      description:
+        'هذه الرخصة تتيح للآخرين حرية إعادة التوزيع، التعديل، التغيير، والاشتقاق من عملك، سواء أكان ذلك لأغراض تجارية أو غير تجارية، طالما ينسبون العمل الأصلي لك. هذه هي الرخصة الأكثر تسامحاً في مجموعة الرخص المعروضة. وهي الأنسب لضمان أكبر انتشار واستخدام ممكن للمواد المرخَّصة.',
+    },
+    'CC BY-SA': {
+      name: 'نَسب المُصنَّف - الترخيص بالمثل',
+      description:
+        'هذه الرخصة تتيح للآخرين حرية إعادة التوزيع، التعديل، التغيير، والاشتقاق من عملك، سواء أكان ذلك لأغراض تجارية أو غير تجارية، طالما ينسبون العمل الأصلي لك ويرخصون أعمالهم المشتقة تحت نفس الشروط. وهذه الرخصة غالباً تضاهي رخص "الحقوق المتروكة" للبرمجيات الحرة والمفتوحة المصدر. كل الأعمال المشتقة من عملك ستحمل نفس الرخصة، بالتالي جميع الأعمال المشتقة تسمح بالاستخدام التجاري. هذه الرخصة هي المستخدمة من قبل ويكيبيديا، وننصح بها للمواد التي قد تستفيد من الاشتقاق من ويكيبيديا والمشاريع المماثلة في الترخيص.',
+    },
+    'CC BY-ND': {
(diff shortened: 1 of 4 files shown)


kolibri/plugins/coach/assets/src/views/plan/LessonContentPreviewPage/index.vue
```diff
@@ -170,6 +171,21 @@
       content() {
         return this.currentContentNode;
       },
+      translatedLicense() {
+        if (
+          licenseTranslations[global.languageCode] &&
```
rtibbles: `import { currentLanguage } from 'kolibri.utils.i18n';`
Rather than relying on an unparsed global variable.

Could even abstract this slightly more and create a wrapper function in the `i18n` module that does this.

kolibri/core/assets/src/utils/licenseTranslations.js
```diff
@@ -0,0 +1,166 @@
+// except where specified, these are from
+// https://creativecommons.org/licenses/?lang=ar
```
rtibbles, line 2: `lang=<lang-code>` presumably?

rtibbles (changes requested):
At a minimum, I'd want to use the i18n currentLanguage, rather than the global reference.

kolibri/plugins/coach/assets/src/views/plan/LessonContentPreviewPage/index.vue
indirectlylit: agreed - I was wondering why we were using the global at all actually

rtibbles:
> current implementation embeds full descriptions license descriptions and names within the content, which seems possibly incorrect

This is intentional, because of the need to be able to represent custom licenses, which we will always have to present in the source language stored in the DB.

Having translations for the common licenses is definitely a big win though!

rtibbles (approved):

indirectlylit:
> This is intentional

minor issues are

* they're not quite consistent with each other
* the name/ID distinction is not made in content

rtibbles:
Oh, sorry, my comment was more about the full text descriptions being embedded in the database. I think we should definitely solidify the ids as ids, not names.

indirectlylit:
yeah I know - I was responding to two things.

anyways .. minor stuff

PR #5210, opened 2019-02-28
Fix client side 'server' time setting to prevent spooky interactions in the future

### LINKED ISSUE #5079: 'last activity' can be in the future

rtibbles:
### Summary
We now set the 'server time' not at page load, but through the session object.
This meant that it was being updated more than once during a session.
Previously, we were using the performance timing API to calculate the start of the page load request, in order to give a time for diffing.
This meant that as time moved on in the client side without a page load, the diff became larger and larger, proportional to the length of time since page load.
This made time stamps in the future as a result.

### Reviewer guidance
Follow steps to reproduce here: https://github.com/learningequality/kolibri/issues/5079#issuecomment-468082659

See that this no longer occurs.

### References
Fixes #5079 

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/core/assets/src/heartbeat.js
+++ b/kolibri/core/assets/src/heartbeat.js
@@ -153,6 +153,8 @@ export class HeartBeat {
         })
       );
     }
+    // Log the time at the start of the request for time diff setting.
+    const pollStart = new Date();
     return client({
       params: {
         // Only send active when both connected and activity has been registered.
@@ -162,6 +164,8 @@ export class HeartBeat {
       path: this._sessionUrl('current'),
     })
       .then(response => {
+        // Log the time that the poll of the session endpoint ended.
+        const pollEnd = new Date();
         const session = response.entity;
         // If our session is already defined, check the user id in the response
         if (store.state.core.session.id && session.user_id !== currentUserId) {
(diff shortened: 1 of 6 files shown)


kolibri/core/assets/src/heartbeat.js
```diff
@@ -153,6 +153,8 @@ export class HeartBeat {
         })
       );
     }
+    // Log the time at the start of the request for time diff setting.
+    const pollStart = new Date();
```
indirectlylit: suuuper minor but I've always found `new Date()` to be an obfuscated way of getting the current date and time. Could possibly standardize on `Date.now()`

indirectlylit (commented):
hah time dilation

jonboiser (approved):
Manually tested Kolibri with and without this patch

without patch: last activity was quoted at 8 min in the future
with patch: last activity was quoted at ~10 sec in the future (might have been from the notification). after a refresh, last activity was in the past.

kolibri/core/assets/src/heartbeat.js
rtibbles: Except that `Date.now()` returns the time in milliseconds since epoch, whereas `new Date()` returns a date object. Given how I end up using them in the calculation below, `Date.now()` would work, but I wouldn't necessarily want to standardize on it.

indirectlylit:
> last activity was quoted at ~10 sec in the future (might have been from the notification)

I don't quite understand - why are we still seeing future times?

PR #5201, opened 2019-02-27
Fix for notification version-checking and database corruption

jamalex:
### Summary

Some fixes to issues in the notification message/version-checking logic uncovered during changes to nutritionfacts being made to support https://github.com/learningequality/kolibri/pull/4828#issuecomment-462215605

Plus: a tiny fix that should eliminate the database corruption issues we've been seeing for years.

### Reviewer guidance

The version-parsing error can be triggered by calling this on the old code (before this PR):
```KOLIBRI_RUN_MODE=msg-test kolibri --debug manage ping --once --server http://nutritionfacts-staging.nanoapp.io```
and then logging into the UI as an admin. It creates a message with empty string as version_range, which blew up the former logic we had in Kolibri.

I replicated the corruption issue by running `kolibri start` in one terminal, and this in another:
`while true; do KOLIBRI_RUN_MODE=msg-test9 kolibri --debug manage ping --once --server http://nutritionfacts-staging.nanoapp.io; done`
Then clicking around in the UI and logging/in out. After a while it leads to a 500 in the web UI.

----

### Contributor Checklist

PR process:

- [x] PR has the correct target branch and milestone
- [x] PR has 'needs review' or 'work-in-progress' label
- [x] If PR is ready for review, a reviewer has been added. (Don't use 'Assignees')
- [ ] If this is an important user-facing change, PR or related issue has a 'changelog' label
- [ ] If this includes an internal dependency change, a link to the diff is provided

Testing:

- [x] Contributor has fully tested the PR manually
- [ ] If there are any front-end changes, before/after screenshots are included
- [ ] Critical user journeys are covered by Gherkin stories
- [ ] Critical and brittle code paths are covered by unit tests

### Reviewer Checklist

- Automated test coverage is satisfactory
- PR is fully functional
- PR has been tested for [accessibility regressions](http://kolibri-dev.readthedocs.io/en/develop/manual_testing.html#accessibility-a11y-testing)
- External dependency files were updated if necessary (`yarn` and `pip`)
- Documentation is updated
- Contributor is in AUTHORS.rst

Issue #3825, opened 2018-06-20
Database disk image is malformed 


mrpau-richard:
### Observed behavior
Upgrading kolibri from 0.7.2 to 0.10.0b3 will have this `Database disk image is malformed` error when running the `kolibri start` command.

### Errors and logs
```
442 static files copied to '/home/user/.kolibri/static', 3 unmodified.
Operations to perform:
  Apply all migrations: admin, auth, content, contenttypes, device, exams, kolibriauth, lessons, logger, morango, sessions
Running migrations:
  Applying auth.0008_alter_user_username_max_length... OK
  Applying content.0007_auto_20180212_1155... OK
  Applying content.0008_contentnode_coach_content...Traceback (most recent call last):
  File "/usr/lib/python3/dist-packages/kolibri/dist/django/db/backends/utils.py", line 64, in execute
    return self.cursor.execute(sql, params)
  File "/usr/lib/python3/dist-packages/kolibri/dist/django/db/backends/sqlite3/base.py", line 328, in execute
    return Database.Cursor.execute(self, query, params)
sqlite3.DatabaseError: database disk image is malformed

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/usr/bin/kolibri", line 9, in <module>
    load_entry_point('kolibri==0.10.0b3', 'console_scripts', 'kolibri')()
  File "/usr/lib/python3/dist-packages/kolibri/utils/cli.py", line 646, in main
    start(port, daemon=daemon)
  File "/usr/lib/python3/dist-packages/kolibri/utils/cli.py", line 261, in start
    update()
  File "/usr/lib/python3/dist-packages/kolibri/utils/cli.py", line 239, in update
    _migrate_databases()
  File "/usr/lib/python3/dist-packages/kolibri/utils/cli.py", line 176, in _migrate_databases
    call_command("migrate", interactive=False, database=database)
  File "/usr/lib/python3/dist-packages/kolibri/dist/django/core/management/__init__.py", line 131, in call_command
    return command.execute(*args, **defaults)
  File "/usr/lib/python3/dist-packages/kolibri/dist/django/core/management/base.py", line 330, in execute
    output = self.handle(*args, **options)
  File "/usr/lib/python3/dist-packages/kolibri/dist/django/core/management/commands/migrate.py", line 204, in handle
    fake_initial=fake_initial,
  File "/usr/lib/python3/dist-packages/kolibri/dist/django/db/migrations/executor.py", line 115, in migrate
[...]

PR #5218, opened 2019-02-28
remove instructions that made corruption worse not better

indirectlylit:
### Summary

In the case of DB corruption, the current instructions advise users to:

```bash
sqlite3 {path} .dump | sqlite3 fixed.db
cp fixed.db {path}
```

User who run these commands will add another layer of corruption on top of the first.

From the docs in [1.4. Mispairing database files and hot journals](https://www.sqlite.org/howtocorrupt.html), the following actions are all likely to lead to corruption:

* Swapping journal files between two different databases.
* Overwritting a journal file with a different journal file.
* Moving a journal file from one database to another.
* Copying a database file without also copying its journal.
* Overwriting a database file with another without also deleting any hot journal associated with the original database. 

Furthermore by having the user overwrite the old DB without backing it up, it made it impossible for us to get a copy of the original corrupted DB.

### Reviewer guidance

Do the new instructions make sense?

### References

* Hopeful fix for DB corruption in #5201
* Refs #3825

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/utils/cli.py
+++ b/kolibri/utils/cli.py
@@ -636,7 +636,6 @@ def main(args=None):  # noqa: max-complexity=13
                 "end in .sqlite3, .sqlite3-shm, .sqlite3-wal, or .log and then"
                 "contact Learning Equality. Thank you!"
             )
-            sys.exit(1)
         raise
 
     # Alias
--- a/kolibri/utils/cli.py
+++ b/kolibri/utils/cli.py
@@ -629,26 +629,12 @@ def main(args=None):  # noqa: max-complexity=13
     try:
         initialize(debug=debug)
     except (DatabaseError, SQLite3DatabaseError) as e:
-        from django.conf import settings
         if "malformed" in str(e):
-            recover_cmd = (
-                "    sqlite3 {path} .dump | sqlite3 fixed.db \n"
-                "    cp fixed.db {path}"
-                ""
-            ).format(
-                path=settings.DATABASES['default']['NAME'],
-            )
             logger.error(
-                "\n"
-                "Your database is corrupted. This is a "
-                "known issue that is usually fixed by running this "
-                "command: "
-                "\n\n"
-                + recover_cmd
-                + "\n\n"
-                "Notice that you need the 'sqlite3' command available "
-                "on your system prior to running this."
-                "\n\n"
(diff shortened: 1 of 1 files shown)


indirectlylit:
also updated to re-raise exception rather than sys.exit - is this preferable to get the stack trace?

rtibbles (approved):

benjaoming:
Would this mean that the current `dbrestore` command suffers from the same issue? I.e. would it be safe to delete/backup journals upon restoring?

Also, are journals needed when running `dbbackup` or is it better/safer to ignore them, as long as the `dbrestore` command doesn't mix mis-matching databases and journals?

jredrejo:
when doing `dbbackup` journals are not included in the backup, but you can't delete current db journals because there would be data corruption. Actually, if the db is closed when doing the backup there should not be any journal. This is something that should be fixed if there are any.

On the other hand, I am not sure about `dbrestore`. If it relies on internal sqlite restore mechanisms it should take care of the journals, if it's a manual restore, journals should be removed before doing the restore process.

benjaoming:
> This is something that should be fixed if there are any.

Yes :) And that's the question, I don't know the internals of SQLite. We do use SQLite execute statements for the restore itself, but we wipe the db file firstly, not any journal files. See from the line `# Wipe current database file`...

```
    dst_file = settings.DATABASES["default"]["NAME"]

    # Close connections
    db.connections.close_all()

    # Wipe current database file
    if not db.connections["default"].is_in_memory_db():
        with open(dst_file, "w") as f:
            f.truncate()
    else:
        logger.info("In memory database, not truncating: {}".format(dst_file))

    # Setting encoding=utf-8: io.open() is Python 2 compatible
    # See: https://github.com/learningequality/kolibri/issues/2875
    with open(from_file, **KWARGS_IO_READ) as f:
        db.connections["default"].connect()
        db.connections["default"].connection.executescript(f.read())

    # Finally, it's okay to import models and open database connections.
    # We need this to avoid generating records with identical 'Instance ID'
    # and conflicting counters, in case the database we're overwriting had
    # already been synced with other devices.:
    from morango.models import DatabaseIDModel

    DatabaseIDModel.objects.create()

```

PR #5159, opened 2019-02-26
Ensure connections are closed on threads to avoid locks in the database

jredrejo:
### Summary
When the database is accessed from a thread, Django creates a new connection that is not closed when the thread ends, or if the thread is kept alive to perform some periodic task.
This causes two problems:
1. Vacuum does not work as the database is always locked while the threads are alive
2. Sqlite corruption when using threads is described in some of the cases at https://www.sqlite.org/howtocorrupt.html and https://www.sqlite.org/wal.html . Without being able to reproduce these problems we can't ensure this PR will solve them, but keeping the database clean, defragmented, with the wal reduced to 0 most of the time will decrease the probability of having these problems.

There were several cases affected in the code: ping & vacuum commands were leaving 2 connections open permanently. 
From version 0.12, notifications thread were leaving four more connections open.

This PR ensure all the connections are closed inside the threads and clean all the connections before executing vacuum.

### Reviewer guidance
Before applying the code from this PR, executing 
`lsof db.sqlite3` 
the list of open connections will be visible

After applying this PR, the list of connections must be zero whenever no requests are made or vacuum has finished.
Also, sqlite wal must be small or zero, because vacuum works correctly.

### References

Closes #5149 

### Contributor Checklist

PR process:

- [x] PR has the correct target branch and milestone
- [x] PR has 'needs review' or 'work-in-progress' label
- [x] If PR is ready for review, a reviewer has been added. (Don't use 'Assignees')
- [ ] If this is an important user-facing change, PR or related issue has a 'changelog' label
- [ ] If this includes an internal dependency change, a link to the diff is provided

Testing:

- [x] Contributor has fully tested the PR manually
- [ ] If there are any front-end changes, before/after screenshots are included
- [ ] Critical user journeys are covered by Gherkin stories
- [ ] Critical and brittle code paths are covered by unit tests

### Reviewer Checklist

- Automated test coverage is satisfactory
- PR is fully functional
- PR has been tested for [accessibility regressions](http://kolibri-dev.readthedocs.io/en/develop/manual_testing.html#accessibility-a11y-testing)
- External dependency files were updated if necessary (`yarn` and `pip`)
- Documentation is updated
- Contributor is in AUTHORS.rst

PR #5229, opened 2019-03-01
Backport corruption fixes

jredrejo:
Working with 0.12 release we've fixed the corruption problems in #5159  and #5201 
While 0.12 is finally released as stable, and considering some clients are having problems with 0.11 and can keep it using for some time, I think we should backport these fixes and build an updated 0.11.2 compilation.

This PR backports the relevant code from the mentioned PRs

### Summary

Explained in #5159 and #5201 descriptions
…

### Reviewer guidance
Same as decribed in #5201 

### References
Infamous #3825 

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/core/analytics/management/commands/ping.py
+++ b/kolibri/core/analytics/management/commands/ping.py
@@ -6,6 +6,7 @@ from datetime import datetime
 
 import requests
 from django.core.management.base import BaseCommand
+from django.db import connection
 from django.utils.six.moves.urllib.parse import urljoin
 from django.utils.timezone import get_current_timezone
 from django.utils.timezone import localtime
@@ -61,6 +62,8 @@ class Command(BaseCommand):
                     logger.info("Ping succeeded! (response: {})".format(data))
                     if "id" in data:
                         self.perform_statistics(server, data["id"])
+                    connection.close()
+
                 if once:
                     break
                 logger.info("Sleeping for {} minutes.".format(interval))
--- a/kolibri/core/content/utils/annotation.py
+++ b/kolibri/core/content/utils/annotation.py
@@ -2,6 +2,7 @@ import datetime
 import logging
 import os
 
+from django.db import connection
 from le_utils.constants import content_kinds
 from sqlalchemy import and_
 from sqlalchemy import exists
@@ -48,6 +49,7 @@ def update_channel_metadata():
             except (InvalidSchemaVersionError, FutureSchemaError):
                 logger.warning("Tried to import channel {channel_id}, but database file was incompatible".format(channel_id=channel_id))
     fix_multiple_trees_with_id_one()
+    connection.close()
 
 
 def fix_multiple_trees_with_id_one():
--- a/kolibri/core/deviceadmin/management/commands/vacuumsqlite.py
+++ b/kolibri/core/deviceadmin/management/commands/vacuumsqlite.py
@@ -3,6 +3,7 @@ import logging
 import time
 
 from django.core.management.base import BaseCommand
+from django.db import close_old_connections
 from django.db import connections
 from django.db import DEFAULT_DB_ALIAS
 
@@ -31,7 +32,7 @@ class Command(BaseCommand):
         if connection.vendor == "sqlite":
             while True:
                 with vacuum_db_lock:
-                    self.perform_vacuum(connection)
+                    self.perform_vacuum(database)
                 if not scheduled:
                     break
                 current_dt = datetime.datetime.now()
@@ -48,17 +49,21 @@ class Command(BaseCommand):
                 logger.info("Next Vacuum at 3AM local server time (in {} minutes).".format(interval))
                 time.sleep(interval * 60)
 
-    def perform_vacuum(self, connection):
+    def perform_vacuum(self, database):
         try:
+            close_old_connections()
+            connections.close_all()
+            connection = connections[database]
             cursor = connection.cursor()
             cursor.execute('vacuum;')
             connection.close()
         except Exception as e:
+            logger.error(e)
             new_msg = (
                 "Vacuum of database {db_name} couldn't be executed. Possible reasons:\n"
                 "  * There is an open transaction in the db.\n"
                 "  * There are one or more active SQL statements.\n"
-                "The full error: {error_msg}").format(db_name=connection.settings_dict['NAME'], error_msg=e)
+                "The full error: {error_msg}").format(db_name=connections[database].settings_dict['NAME'], error_msg=e)
             logger.error(new_msg)
         else:
             logger.info("Sqlite database Vacuum finished.")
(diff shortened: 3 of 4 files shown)


rtibbles:
The failure of the build for this PR seems to be an issue with our version checking for Debian builds?
```
Successfully tagged learningequality/kolibri-deb:latest
--
  | export KOLIBRI_VERSION=$(cat kolibri/VERSION) && \
  | docker run --env-file ./docker/env.list -v $PWD/dist:/kolibridist "learningequality/kolibri-deb"
  | kolibri-0.11.1.tar.gz
  | uupdate: New Release will be 0.11.1-0ubuntu1.
  | uupdate: new version 0.11.1-0ubuntu1 <= current version 0.12.0~b7-0ubuntu1; aborting!
  | make: *** [docker-deb] Error 1
  | 🚨 Error: The command exited with status 2
```

@lyw07 this seems like an issue if we ever want to release a patch release for an older version?

lyw07:
@rtibbles ~~it might be because we are building the installer based on the docker image cache, but I will look into the issue more to confirm.~~
The failure is because we get the latest source code from kolibri-proposed PPA and build the new version based on that. The latest version on kolibri-proposed is 0.12.0~b7 which is greater than 0.11.1. I will check how to fix this.

jredrejo:
@lyw07 I see we are doing that inside a docker container. I think we could change it to take the sources from github, using the appropiate tag instead. We have had a couple of problems with the current building method as it depends of uploading code changes to the ppa, instead of using git, what seems more logical.

what do you think?

lyw07:
@jredrejo I agree with your idea. Our current approach in the docker file does not allow us to release old versions in many situations, because many source files related to the old versions on Launchpad have been all gone due to superseding packages with newer versions. So taking the sources from github should be the way to go.
However, I'm wondering how we can get the code from https://github.com/learningequality/kolibri-installer-debian. Since we don't tag versions in the kolibri-installer-debian repository, it's going to be hard for us to find out the specific debian files, especially `changelog`, for the old release. Do you have any suggestions? Thank you!

rtibbles:
Do we ever make non-backwards compatible changes to the debian-installer?

lyw07:
from the list of packages in PPA, it seems that we haven't done that before.
https://launchpad.net/~learningequality/+archive/ubuntu/kolibri/+packages?field.name_filter=&field.series_filter=&field.status_filter=&batch=75&direction=backwards&memo=75

jredrejo:
> from the history of PPA, it seems that we haven't done that before.

Yes, but I don't see any reason to self-restrict us. It should be possible (using apt won't work because it will get always the latest vesion, but dpkg will work)

jredrejo:
> However, I'm wondering how we can get the code from https://github.com/learningequality/kolibri-installer-debian. Since we don't tag versions in the kolibri-installer-debian repository, it's going to be hard for us to find out the specific debian files, especially `changelog`, for the old release. Do you have any suggestions? Thank you!

We could tag the last commit for 0.11.1, and do the it from now on for the new releases.

I have done it for 0.11.1 : https://github.com/learningequality/kolibri-installer-debian/releases

lyw07:
yes I think that makes sense. So to make sure I understand things correctly, things we need to do are:
1. Start to tag versions in https://github.com/learningequality/kolibri-installer-debian so we can get related debian files using tags.
2. In `build_debian.dockerfile`, we will get the source code from github with tags instead of using `apt-get source kolibri`. 

Do these sound right? Thank you!

jredrejo:
> yes I think that makes sense. So to make sure I understand things correctly, things we need to do are:
> 
> 1. Start to tag versions in https://github.com/learningequality/kolibri-installer-debian so we can get related debian files using tags.
> 2. In `build_debian.dockerfile`, we will get the source code from github with tags instead of using `apt-get source kolibri`.
> 
> Do these sound right? Thank you!

yes, my only doubt is how we should tag. For this test I have tagged `v0.11.1-0ubuntu3` from the debian/changelog as `v0.11.1` . For this one it is clear as that's the latest 0.11.1 package version. But in the future, we might have v0.12.2-0ubuntu1, then  v0.12.2-0ubuntu2 .. and we don't know in advance which will be the latest v0.12.2. Including the `~ubuntu `part would make build_debian.dockerfile can't find the right release. We should find a tag mechanism to get always the latest packaging tag for the desired kolilbri version.

indirectlylit:
If at all possible, I'd love to put the debian installer on a separate versioning scheme than kolibri. This is what we do with the other distributions:

* [Windows](https://github.com/learningequality/kolibri-installer-windows/releases) is currently on 1.2.1
* [Android](https://github.com/learningequality/kolibri-installer-android/releases) is currently on 0.1.0-beta4
* [Kolibri Server](https://github.com/learningequality/kolibri-server/releases) is on 0.1.0-beta1

I don't think the current strategy of tagging the Debian code with Kolibri version numbers will work long-term. [See discussion here](https://learningequality.slack.com/archives/CAWKSU7KR/p1550700816017000).

Proposed:

* Kolibri has one set of versions (e.g. `0.11.2`)
* Debian source code (in git) has an independent set of versions (e.g. `0.1.3`).
  * We can maintain release branches to help iterate on older versions (e.g. simultaneous `0.1.3` and `0.2.0`)
* Debian packages (in the PPA) use the conventional versioning (e.g. `0.11.2-0ubuntu[N]`, where [N] is just an increasing integer)
* Every published PPA version has a [.changes file](https://www.debian.org/doc/debian-policy/ch-controlfields.html#debian-changes-files-changes) or some similar mechanism which specifies the Kolibri and Debian versions it was built from. Examples:
  * `0.11.2-0ubuntu3` might specify `kolibri v0.11.2; debian v0.1.3`
  * `0.12.3-0ubuntu1` might specify `kolibri v0.12.3; debian v0.2.0`

benjaoming:
I believe we already talked about tagging the Debian installer repo :) Just need to do it properly :)

benjaoming:
Am on this! Will open a PR for `release-v0.11.x` that builds with some ideal branch from `kolibri-installer-debian`

indirectlylit:
I don't think we still need to do this, as most people will just upgrade to 0.12.x.

Feel free to re-open if that's not the case.

benjaoming:
@indirectlylit sorry, I didn't respond to this:

> Every published PPA version has a .changes file or some similar mechanism which specifies the Kolibri and Debian versions it was built from. Examples: 

It's a very interesting file for sure :) The `.changes` file is generated automatically by `dpkg-buildpackage` and AFAIK not something that maintains any kind of human readable versioning. From the description:

> The .changes files are used by the Debian archive maintenance software to process updates to packages. 

What happens, is that when we use `dput`, the Debian uploading tool, we point it to this `.changes` file. It then gets a manifest of all the stuff that it needs to process. This for instance means, that we can release `1.2.3-0ubuntu2` on top of `1.2.3-0ubuntu1` and it WILL NOT upload the source package, but only the Debian rules. The whole thing is rebuild by Launchpad, but we save the trouble of re-uploading Kolibri itself.

It's very clever :) Files already exist and look like this:

image

PR #5224, opened 2019-03-01
Move async notifications thread invocation into server start.

rtibbles:
### Summary
In production mode, the async notifications thread seems to shut down, this may well be because it gets started during Django initialization, which happens prior to daemonization.

This PR fixes this by moving the async notification thread starting to `server.py` `start`. This has the disadvantage of not logging notifications in dev mode, but a workaround can be found.

### Reviewer guidance
Do notifications now get logged in production mode, always?

### References
Fixes #5217 

----

### Contributor Checklist

PR process:

- [x] PR has the correct target branch and milestone
- [x] PR has 'needs review' or 'work-in-progress' label
- [x] If PR is ready for review, a reviewer has been added. (Don't use 'Assignees')
- [ ] If this is an important user-facing change, PR or related issue has a 'changelog' label
- [ ] If this includes an internal dependency change, a link to the diff is provided

Testing:

- [x] Contributor has fully tested the PR manually
- [ ] If there are any front-end changes, before/after screenshots are included
- [ ] Critical user journeys are covered by Gherkin stories
- [ ] Critical and brittle code paths are covered by unit tests

### Reviewer Checklist

- Automated test coverage is satisfactory
- PR is fully functional
- PR has been tested for [accessibility regressions](http://kolibri-dev.readthedocs.io/en/develop/manual_testing.html#accessibility-a11y-testing)
- External dependency files were updated if necessary (`yarn` and `pip`)
- Documentation is updated
- Contributor is in AUTHORS.rst

Issue #5217, opened 2019-02-28
Class activity is not being recorded - beta11


radinamatic:
### Observed behavior

![ubuntu 16 04 3 start running - oracle vm virtualbox_445](https://user-images.githubusercontent.com/1457929/53594141-7bd3ef00-3b9a-11e9-924d-162b65e8a5fa.png)

### Expected behavior
…

### User-facing consequences
…

### Errors and logs
[kolibri.log](https://github.com/learningequality/kolibri/files/2916667/kolibri.log)
[server.log](https://github.com/learningequality/kolibri/files/2916668/server.log)

[Notification DB](https://learningequality.slack.com/files/U0LJG17EK/FGK4N0DRN/notifications.sqlite3)
There should have been several lesson and quiz interactions by user `abbyl` in the last 30 minutes, and more recently quiz interaction by `peterb`

### Steps to reproduce

…

### Context
Version:           0.12.0b11
OS:                Linux-4.4.0-142-generic-x86_64-with-Ubuntu-16.04-xenial
Installer:         apt
Server:            Kolibri internal server

cc @rtibbles @jonboiser @indirectlylit

indirectlylit:
Notes so far:

* the `LearnerProgressNotification` table appears to be empty
* on the surface, looks very similar to #5113
* nothing of interest in the logs

indirectlylit:
still can't reproduce locally. made a PR here with some extra logging: https://github.com/learningequality/kolibri/pull/5221

mrpau-julius:
same issue on kolibri-server.
![image](https://user-images.githubusercontent.com/38421180/53616953-0ad71a80-3c20-11e9-8f6e-55c70c347dde.png)

jredrejo:
Fixed on #5224

PR #5231, opened 2019-03-01
Use a disk queue for the async notifications thread

rtibbles:
### Summary
Anticipates possible issues with memory sharing for the async notiifications queue, and implements a disk based queue (with an appropriate dehydration and hydration layer) to continuously store notifications that need updates.

### Reviewer guidance
Does it work?

At the moment this implementation clears the entire queue before moving on, I imagine it might be desirable to have a `queue_chunk_length` to determine how many are processed in one transaction.

### References
Depends on #5224

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/core/notifications/tasks.py
+++ b/kolibri/core/notifications/tasks.py
@@ -1,11 +1,37 @@
 import logging as logger
+import os
 import threading
 import time
 
+from diskcache import Deque
 from django.db import connection
 from django.db import transaction
 
+from kolibri.core.logger import models
+from kolibri.core.notifications import api
+from kolibri.utils.conf import KOLIBRI_HOME
+
 logging = logger.getLogger(__name__)
+
+
+queue_dir = os.path.join(KOLIBRI_HOME, 'notifications_queue')
+
+
+def dehydrate_spec(fn, instance, *args):
+    return {
+        "function_name": fn.__name__,
+        "model_type": type(instance).__name__,
+        "model_id": instance.id,
+        "args": args,
+    }
+
+
+def hydrate_and_execute_spec(notification_spec):
+    function = getattr(api, notification_spec["function_name"])
+    Model = getattr(models, notification_spec["model_type"])
+    instance = Model.objects.get(id=notification_spec["model_id"])
+    args = notification_spec["args"]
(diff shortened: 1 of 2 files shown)


jredrejo (commented):
I can't think of any problems with this code. 
However I can't think either of problems having the code in memory currently. Having them in different threads only could affect to the order the notifications are inserted, and that's not very important as they have their own timestamp. Thinking in future additions to the notifications it's clearly  healthier to move to a thread-safe implementation as we can't anticipate if that will cause problems.

I think this is a tradeoff between the performance of DiskCache and the memory usage of the previous implementation. If DiskCache is as fast as they claim, we can go with it.  

But we must be aware DiskCache is a queue implementation over sqlite. All the problems we have had with sqlite apply here. First thing I think we should do is activating wal on it to ensure writes are fast. We also should test the behaviour of the journal to decide if we need to do vacuum periodically. 

Also, from an arquitectural point of view I am not too happy: 
We were using a memory queue to be fast before inserting the notifications in a sqlite database. Now we are using a sqlite database to insert the queue that later is read and inserted into another sqlite database. Doesn't it sound odd?

indirectlylit:
intuitively, I have the same feelings as @jredrejo. If #5224 fixes the symptoms in #5217, this change may end up doing good or harm and it's hard to predict which.

could someone remind me why we didn't use iceqube for notifications?

rtibbles:
Yeah, this is meant as a proof of concept in case it is needed, if not, I am happy to stay with an in-memory implementation.

> could someone remind me why we didn't use iceqube for notifications?

I think ultimately we should, but it would have all the disadvantages of this PR in terms of the disk based queue. I think the main issue is that iceqube doesn't currently manage multiple workers? So doing any other async task would halt notification logging.

jredrejo:
> could someone remind me why we didn't use iceqube for notifications?

For the sake of simplicity, that's why the complexity that we are adding now makes me think that this was a MVP and now we need to move to a better thought architecture.
Moving to iceqube, or replacing iceqube with a general solution makes more sense to me that handling two different kind of queues in the same application.

Issue #2258, opened 2017-09-25
add device setting for changing default language


indirectlylit:
The default language for a device is now configurable during setup. However there is currently no UI for modifying this after the fact.

Needs design - cc @khangmach @jtamiace

khangmach:
@indirectlylit there is a "change language" option within the user dropdown

indirectlylit:
The user dropdown updates the language for the current user.

This issue is for the device default, which is set up during initial onboarding. This affects the default language for newly-created users.

indirectlylit:
It also affects the language of the sign-in screen. Will probably need to be added to device settings somewhere:

![image](https://user-images.githubusercontent.com/2367265/30794590-f2f03796-a17d-11e7-9d04-98df7b1fdf2c.png)

jtamiace:
![device setting](https://user-images.githubusercontent.com/6668144/31142351-e8acad02-a82e-11e7-9833-8b1cf828c253.png)

Putting this here so that it doesn't get lost. Still prefer the option on the left, with tab name of "Settings" to minimize the number of tabs in the future "Data" & "Facilities"

Note from @indirectlylit in Slack:

> (minor subtlety is the other language switchers change the global language immediately, whereas this one wouldn’t because the admin’s language preference still takes priority. doesn’t really affect the designs above, just an interaction detail)

jtamiace:
![slice](https://user-images.githubusercontent.com/6668144/31144062-560641a6-a834-11e7-8b19-74754e46d61c.png)

I think we like this one better. In the words of Devon:

> "simple drop-down is simple"

jonboiser:
This will require some new additions to the code:

1. A Viewset to be able to modify the default language in the DeviceSettings model
1. An entirely new section for Device > Settings

PR #5246, opened 2019-03-06
Make device settings language have more impact.

rtibbles:
### Summary
Moves device settings language up in the priority list so that it takes priority over browser accept headers if set.

### Reviewer guidance
If you have no set session language, and no language cookie set, does it default to the device setting language, and not your browser language setting?

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated


kolibri/core/device/translation.py
```diff
@@ -13,7 +14,8 @@
 
 from .models import DeviceSettings
 
-DEVICE_LANGUAGE = None
```
jonboiser, line 16: the `global DEVICE_LANGUAGE` statement on line 29 can be removed too.

jonboiser:
Tested this out by updating the database and it seems to work. I had to add a line to clear the cache on every refresh to avoid waiting 10 minutes. 

1. Changing DB doesn't change if user has set a language cookie
1. changing db does change language if in incognito
1. When language is null, it defaults to english for me

When implementing #2258, we will probably have to clear or re-set the cached language setting when we change the language to make sure that the changes take place immediately after a user updates them.

jonboiser (approved):

rtibbles:
The tricky thing with the cache clearing is that our caches are thread safe, and so when we are running in production mode, the cache only gets cleared in the thread where the change is made. The only way to be sure of making the change is never to cache the device setting value.

jonboiser:
Caching the device setting seems like a pretty minor optimization, I think, so maybe 
 we don't need to cache. The get-language function only gets called once when a user loads the app.

rtibbles:
The get language from request gets called on every request (including API requests), until a user explicitly sets a language using the language switcher, when it gets set on the cookie and the session, and hence the read from the device setting gets by passed.

I think we might still be able to get away with not caching it (in the follow up PR #4994 - which uses language URLs, and so bypasses the need to check the device setting).

PR #5251, opened 2019-03-07
New Feature - Change the default language in DeviceSettings model

### LINKED ISSUE #2258: add device setting for changing default language

jonboiser:
### Summary

1. Adds a Read and Update-only endpoint `/api/device/devicelanguage` for the `DeviceSettings.language_id` field
1. Adds a page under Device > Settings for an admin user to 

GIF starts when I type "Start" in the user name

language changing

### Reviewer guidance

1. Test out the UI for as many languages as you want
1. Check to see that anonymous or non-admin users cannot access the API endpoint 

### References

Fixes #2258

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated


CHANGELOG.rst
```diff
@@ -4,6 +4,13 @@ Release Notes
 List of the most important changes for each release.
 
 
+0.12.1
```
indirectlylit: maybe leave this `0.12.y` until we know the final answer

kolibri/plugins/device_management/assets/src/views/DeviceSettingsPage/index.vue
```diff
@@ -0,0 +1,149 @@
...
+          })
+          .catch(() => {
+            this.saveStatus = 'ERROR';
+          });
+      },
+      getDeviceLanguageSetting,
+      saveDeviceLanguageSetting,
+    },
+    $trs: {
+      browserDefaultLanguage: 'Browser default',
```
rtibbles, line 129: Super!

kolibri/core/device/api.py
```diff
@@ -99,3 +102,29 @@ def get(self, request, format=None):
...
+            'language_id': settings.language_id
+        })
+
+    def patch(self, request):
+        settings = DeviceSettings.objects.get()
+
+        try:
+            request_lang = request.data.get('language_id', '')
+        except (Exception):
+            request_lang = ''
```
indirectlylit: Under what circumstances does this happen?

Recommend only catching the specific exception type you're expecting and letting others raise an exception so that errors aren't lost.

indirectlylit (commented):
code looks good, haven't tested

one small note on exception handling

kolibri/core/device/api.py
jonboiser: If the code falls into this branch, the response will be a 400 a few lines below here.

This can happen if the payload is deserialized into a list or number or something that can't be converted to a dictionary (throwing an AttributeError). JSON parse errors for non-JSON strings are handled by the APIView apparently, and will also throw a 400.

kolibri/plugins/device_management/assets/src/views/DeviceSettingsPage/index.vue
rtibbles: Oh - we should add this option to the setup wizard language picker also.

jonboiser: Good point!

jonboiser: The setup wizard uses the same language picker on the sign-on page. Should we add "Browser default" to all the language selectors, including the modal type?

jonboiser: Making this work via the language cookie for individual users is probably beyond the scope of the feature, and might be a little more complicated to implement.

To add this in the setup wizard, we will probably need to add a special flag to provide "Browser default" as an option to the LanguageSwitcherList.

kolibri/core/device/api.py
indirectlylit: In that case I might do something like

```python
try:
    request_lang = request.data.get('language_id', '')
except AttributeError:
    return HttpResponseBadRequest('Value must be a JSON object: {}'.format(request.data))
```

Then if any other exception occurs, it will throw a 500 which would be more appropriate for an unexpected error.

kolibri/core/device/api.py
```diff
@@ -99,3 +102,29 @@ def get(self, request, format=None):
...
+
+        if request_lang is None or check_for_language(request_lang):
+            settings.language_id = request_lang
+            settings.save()
+            return self.get(request)
+        else:
+            return HttpResponseBadRequest()
```
indirectlylit: This can be made more informative like:

```python
return HttpResponseBadRequest('Unknown language: {}'.format(request_lang))
```

kolibri/core/device/api.py
indirectlylit: also below you're checking for `None`, so maybe this should actually be `request.data.get('language_id')`? (instead of returning an empty string as the default)

kolibri/plugins/device_management/assets/src/views/DeviceSettingsPage/index.vue
rtibbles: We could let the language switcher expose this option, but I think it might be more confusing to users, so let's not give the option to clear the cookie until people ask for it.

> To add this in the setup wizard, we will probably need to add a special flag to provide "Browser default" as an option to the LanguageSwitcherList.

I think this is the probably what we need to do.

kolibri/core/device/api.py
jonboiser: I kind of wanted to keep the error handling simple since there's only one consumer of this API, and 500 errors crash the app. Hopefully, since the try block is simple, AttributeError will catch _all_ of the errors that might occur.

It's also not correct to not leave a specific default for getting 'language_id' since this will happen

```
foo_bar = { "foo": "bar" }
ok = { "language_id": None } # None converted from null

foo_bar.get('language_id') == None # Is acceptable when it shouldn't be
```

kolibri/plugins/device_management/assets/src/views/DeviceSettingsPage/index.vue
jonboiser: Actually, this might be messy too, since the language list expands to the modal (so we'd need to update the modal too). Also, not sure what would happen if you actually selected "Browser default" in the setup wizard, since changing a language on the language list triggers a full refresh.

How about we just put a button below the language list on Step 1 of the Setup Wizard to the effect of "Use the browser default"?

jonboiser: Or can we just consider this PR done and punt the Setup Wizard stuff to a new feature?

rtibbles (approved):
LGTM

rtibbles:
Added a follow up issue for the setup wizard here: https://github.com/learningequality/kolibri/issues/5290

indirectlylit:
adding 'changelog' and 'needs docs update' labels

radinamatic:
@jonboiser Would it be possible to have a Gherkin scenario to test this feature? 
(see https://github.com/learningequality/kolibri-docs/issues/100 for reference)

jonboiser:
@radinamatic 	I think this is how it basically works.

```
Scenario: Changing the default language in facility settings should result in that language being shown when the site is visited for the first time

GIVEN There is an admin user "Admin" and a learner user "Learner"
    AND "Admin" is at the Facility Settings Page
    AND The current default language is "English"
    AND "Learner" has never used Kolibri on their device before
WHEN "Admin" sets the default language to "French"
    AND "Admin" clicks *Save*
THEN "Admin" sees a notification that the settings have been successfully changed.
WHEN "Learner" accesses Kolibri for the first time
THEN The Sign In page (and the rest of the application) is localized in French 
```

radinamatic:
@jonboiser Thank you for quick reply. Steps you describe are more or less what I imagined them to be, but are not working as intended (tested on 0.12.4 on Ubuntu 16.04), I'll file an issue.

indirectlylit:
thanks guys

Issue #5242, opened 2019-03-04
KPageContainer or panel shadows not used consistently (and nowhere in Learn page)


jonboiser:
### Observed behavior

The Learn page is flatter than the other coach/admin pages because the main content is not surrounded by the KPageContainer.

### Expected behavior

Learn Pages are surrounded by the KPageContainer

### Context
0.12rc1

jonboiser:
Other places where the center area is completely flat.

- [x] Exam Reports (ReportsQuizLearnerPage)
- [x] Exercise Reports (ReportsLessonExerciseLearnerPage)
- [x] PlanLessonSelectionContentPreview
- [x] User Profile Page (UserIndex)
- [ ] every LearnIndex sub-page

Issue #5119, opened 2019-02-20
Apply header-icon pairings consistently throughout pages


jonboiser:
### Observed behavior
On the username in Exam Reports
<img width="688" alt="Screen Shot 2019-05-13 at 2 54 52 PM" src="https://user-images.githubusercontent.com/10248067/57657091-21d6a900-758f-11e9-9d01-085bff793723.png">

On the class name in Coach Homepage

<img width="555" alt="Screen Shot 2019-05-13 at 2 53 52 PM" src="https://user-images.githubusercontent.com/10248067/57657036-f5bb2800-758e-11e9-8023-fd287854369e.png">

On the headers for channels when importing/exporting

<img width="943" alt="Screen Shot 2019-05-13 at 2 52 31 PM" src="https://user-images.githubusercontent.com/10248067/57656968-ca383d80-758e-11e9-8fd9-5b928d895876.png">

~On the headers for lesson resources (and probably for resources in learn browsing)~

<img width="711" alt="Screen Shot 2019-03-21 at 2 01 21 PM" src="https://user-images.githubusercontent.com/10248067/54784765-dbde1400-4be1-11e9-8b35-a56446fa4a5d.png">

~For example, in Coach > Plan > Lesson, the "Wednesday" lesson header could use a Lesson Icon to its left to match its corresponding page on Coach > Reportgs~

![screen shot 2019-02-20 at 12 01 13 pm](https://user-images.githubusercontent.com/10248067/53120773-72e27e00-3507-11e9-9055-ef86d3235f55.png)

~Similarly, in Facility > Classes > Edit Class, the "Class One" header could use an icon too (to match a similar page in Coach Reports)~

![screen shot 2019-02-20 at 12 03 05 pm](https://user-images.githubusercontent.com/10248067/53120807-85f54e00-3507-11e9-90f6-adba1285bbc1.png)

~And again in  the user permissions page, we could add an icon next to the user name~

![screen shot 2019-02-20 at 12 08 37 pm](https://user-images.githubusercontent.com/10248067/53121126-4ed36c80-3508-11e9-92f2-4430128d6d6b.png)

* Classroom Name header on Class Homepage

### Expected behavior

Consistent symbology throughout the app, I suppose.

### Context

0.12 latest

rtibbles:
For the lesson example, this appears to be related to: #5143

rtibbles:
@jonboiser did a lot of work on this in #5233 - will leave it open until we have confirmed that all instances are covered.

[...]

PR #5261, opened 2019-03-12
Layout changes for the ContentPage in Learn pages

### LINKED ISSUE #3664: Use immersive page in content-page within Learn
### Observed behavior

Not using immersive page in content-page within Learn

### Expected behavior

Use immersive page in content-page within Learn

### User-facing consequences

Not the best use of space.

### LINKED ISSUE #4706: Fixes to resource pages
### Observed behavior
Some outstanding issues with the current exercises page that we need to address:
- Severe lack of screen estate affording the user to interact with the exercise. The presence of top-nav/breadcrumbs/and previously the learn tabs all push down elements and components that correspond to the exercise question (including answer input areas and answer choice selections). 
- Reliance on including navigational elements ever-present on every page. This is valuable usability for the act and intention of navigation to/from various places (e.g. channels tab, import) but shouldn't be present in its current form within pages that stress primary purpose to be heads down focus and engagement (content pages). 
- Legibility of the page suffers. Text everywhere, lack of a clear direction

### Expected behavior and projected improvements
- More clean space to work with
- Standardize on the top-nav style (we already have this applied to lesson resource pages)
- Clear differentiation between "Question area" and "answer areas" (big)

### Design changes
- Top-nav swap
- Eliminate redundancy of title-ception (option 2 ideal)
- Maintain breadcrumb affordance via showing "location", e.g the name of parent topic. Or "From" 
- Eliminate current forms of breadcrumbs
- Add "In-progress" "Completed" labels near the appropriate icon

### Specs and mockup
topnav comparisons

jonboiser:
### Summary

1. Implements layout changes for ContentPage (where you view videos, exercises, etc.), per #4706 

Screen Shot 2019-03-12 at 12 11 31 PM

### Reviewer guidance

1. Verify that parent topic links only appear when browsing content
1. Verify that parent topic links do not appear when viewing a lesson resource
1. Check on the link for the back arrow "<-" as make sure it behaves as described in #4706 for all of the different situations

### References

Fixes #4706 
Fixes #3664 

…

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/plugins/learn/assets/src/views/ContentPage.vue
+++ b/kolibri/plugins/learn/assets/src/views/ContentPage.vue
@@ -249,8 +249,9 @@
         return this.$tr('recommended');
       },
       parentTopic() {
-        if (this.content.breadcrumbs.length > 0) {
-          return this.content.breadcrumbs[this.content.breadcrumbs.length - 1];
+        const { breadcrumbs = [] } = this.content;
+        if (breadcrumbs.length > 0) {
+          return breadcrumbs[breadcrumbs.length - 1];
         }
       },
       progress() {
--- a/kolibri/plugins/learn/assets/src/views/ContentPage.vue
+++ b/kolibri/plugins/learn/assets/src/views/ContentPage.vue
@@ -14,22 +14,13 @@
     />
 
     <div v-if="parentTopic" class="topic-link">
-      <mat-svg
-        ref="folderIcon"
-        class="folder-svg"
-        category="file"
-        name="folder"
-      />
-      <KTooltip
-        reference="folderIcon"
-        :refs="$refs"
-      >
-        {{ $tr('topicLocationTooltip') }}
-      </KTooltip>
-      <KRouterLink
-        :text="parentTopic.title"
-        :to="$router.getRoute('TOPICS_TOPIC', { id: parentTopic.id })"
-      />
+      <KLabeledIcon :tooltipText="$tr('topicLocationTooltip')">
+        <KIcon slot="icon" topic />
+        <KRouterLink
+          :text="parentTopic.title"
+          :to="$router.getRoute('TOPICS_TOPIC', { id: parentTopic.id })"
+        />
+      </KLabeledIcon>
     </div>
 
     <ContentRenderer
@@ -155,7 +146,8 @@
   import CoachContentLabel from 'kolibri.coreVue.components.CoachContentLabel';
   import DownloadButton from 'kolibri.coreVue.components.DownloadButton';
   import KRouterLink from 'kolibri.coreVue.components.KRouterLink';
-  import KTooltip from 'kolibri.coreVue.components.KTooltip';
+  import KLabeledIcon from 'kolibri.coreVue.components.KLabeledIcon';
+  import KIcon from 'kolibri.coreVue.components.KIcon';
   import { isAndroidWebView } from 'kolibri.utils.browser';
   import UiIconButton from 'kolibri.coreVue.components.UiIconButton';
   import markdownIt from 'markdown-it';
@@ -189,8 +181,9 @@
       AssessmentWrapper,
       MasteredSnackbars,
       UiIconButton,
+      KIcon,
+      KLabeledIcon,
       KRouterLink,
-      KTooltip,
     },
     metaInfo() {
       // Do not overwrite metaInfo of LessonResourceViewer
@@ -356,12 +349,6 @@
 
 
 <style lang="scss" scoped>
-
-  .folder-svg {
-    margin-right: 8px;
-    margin-bottom: -2px;
-    vertical-align: bottom;
-  }
 
   .coach-content-label {
     margin: 8px 0;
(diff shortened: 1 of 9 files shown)


kolibri/plugins/learn/assets/src/views/ContentPage.vue
```diff
@@ -327,6 +357,12 @@
 
 <style lang="scss" scoped>
 
+  .folder-svg {
+    margin-right: 8px;
+    margin-bottom: -2px;
+    vertical-align: bottom;
+  }
```
indirectlylit: let's use the `KLabeledIcon` component here to avoid adding special-case layout styling.

If desired, we could can add a new prop to handle tooltips which would be valuable in other places also

kolibri/core/assets/src/views/icons/KLabeledIcon.vue
```diff
@@ -1,22 +1,33 @@
 <template>
 
   <span class="labeled-icon-wrapper">
-    <div class="icon">
+    <div ref="icon" class="icon">
```
indirectlylit: cc @khangmach -

Currently the tooltip and 'up' link feel a bit awkward:

* they're below the header rather than above
* the icon and link look visually distinct, and only the link is clickable but only the icon has the tooltip
* the phrase "Resource is located in this topic" will be challenging for young learners to understand

> image

For reference, below is how we've started doing 'up' links in coach tools, which looks more clear and doesn't need a tooltip. I feel like we should just use this pattern rather than introducing a new one:

> image


The other thing we should probably do is be consistent about when we use type icons. For example here's the parent topic view:

> image

Should we be bringing back type icons in headers? (We could right-align the 'progress' icon if this would be desirable)

indirectlylit (commented):
Hooray! Moving to the full-screen is awesome.

My only concern is with 'up' links, as described below

kolibri/core/assets/src/views/icons/KLabeledIcon.vue
indirectlylit: i.e. something like:

| topic view | content view |
|--|--|
| image | image |

* parent back links always above header, always purple clickable
* in-progress icons top-right in both cards and pages
* type icons top-left in both cards and pages

indirectlylit: final question: should we be wrapping the content in a `KPageContainer`? They still look a little naked with the grey background:

> image

jonboiser: There's a separate issue for the KPageContainer here #5242. There are multiple pages where it's not used as well, so it would be nice if it were all fixed in one patch.

jonboiser: Same with the header+icon pairings in #5119. Should try to fix this holistically.

indirectlylit: ooooh, @jonboiser just brought up a great point.

I _think_ we'd decided to start going with the convention that in full-screen windows, we'd do:

* × [current thing]
* ← [parent thing]

In other words, back buttons are labeled by what you're going back to, and close buttons are labeling the thing you're about to close. Is that right @khangmach?

In that case, everything becomes much simpler:

> image

khangmach: @indirectlylit Yes we can just use the implementation above and just have the parent name in the topnav. Also, is the gold star floating right an updated implementation? Haven't seen that yet

jonboiser: The only changes I'm going to make are to

1. Make the label to the "<-" the parent topic name (if in browse mode)
1. Make the label to the "<-" the lesson name (if in lesson mode)
1. Remove the topic back link.

Moving icons around is a broader design system issue that should maybe be addressed in #5119

indirectlylit: > Also, is the gold star floating right an updated implementation? 

that was a recommendation to mirror the layout of content cards:

image

and also to be a bit more similar to coach reports, which put the progress info in a separate column rather than inline with the title

kolibri/core/assets/src/views/icons/KLabeledIcon.vue
```diff
@@ -1,22 +1,33 @@
...
     <div class="label">
       <slot></slot>
     </div>
+    <KTooltip
+      v-if="tooltipText"
+      reference="icon"
+      :refs="$refs"
+    >
+      {{ tooltipText }}
+    </KTooltip>
```
indirectlylit: new tooltip prop is now unused

let's remove this until we have a concrete need for it. Then we can decide whether it's appropriate to have the tooltip on just the icon versus on the entire labeled icon

indirectlylit (approved):
this looks great!

one more small change requested and then feel free to merge

PR #1726, opened 2017-06-24
Prettier for linting

rtibbles:
## Summary

Uses prettier for linting instead of eslint.

Migrates app to use ES6 imports wherever possible.

Aligns Kolibri API Resources imports to be in the same form as other core imports:
e.g. `import { ContentNodeResource } from 'kolibri.resources';`

Upgrades node requirement to Node v6.

Uses headless Chrome instead of PhantomJS (PhantomJS started hating on my JS for reasons I couldn't understand).

Cleans up installed packages.

Updates tests.

## TODO

- [X] Have tests been written for the new code?
- [X] Has documentation been written/updated?
- [X] New dependencies (if any) added to requirements file
- [x] Add an entry to CHANGELOG.rst
- [x] Add prettier linting to git pre-commit hooks?

## Reviewer guidance

The last commit is the prettier rewrite, so looking at commits before that may be more instructive. Best review is to make sure the app still looks and runs the same. I've done a fair amount of testing, but I could have missed something.

PR #5286, opened 2019-03-19
apply 'Black' formatter

indirectlylit:
### Summary

* apply formatting
* add pre-commit hook
* updates to docs

### Reviewer guidance

* should we do this?
* is this the right time to do this?

### References

https://github.com/ambv/black

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/.buildkite/create_integration_testing_worksheet.py
+++ b/.buildkite/create_integration_testing_worksheet.py
@@ -22,7 +22,7 @@ SPREADSHEET_TITLE = "Integration testing with Gherkin scenarios"
 
 # Use to get the Kolibri version, for the integration testing spreadsheet
 SHEET_TAG = os.getenv("BUILDKITE_TAG")
-SHEET_TPL_COLUMN = 'B'
+SHEET_TPL_COLUMN = "B"
 SHEET_TPL_START_VALUE = 5
 SHEET_INDEX = 0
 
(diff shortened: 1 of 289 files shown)


jredrejo:
I like Black and use it when developing my code, but I don't think applying it massively to the whole code base is a good idea as we are losing a good part of the git history with it, together with the utility of `git blame`.
I would prefer to have it applied for new changes instead.

Also, as far as I know it only works with Python 3.x, so having it as a mandatory or pre-commit hook will force everybody to use python 3.x for development. That might not be a bad thing, but it's good to remark it.

indirectlylit:
thanks @jredrejo - very good feedback

> I would prefer to have it applied for new changes instead.

Yeah, I had a similar thought and this is definitely a possibility. It may be that applying it globally doesn't have a huge benefit.

The main reason I like doing it all at once is to avoid this scenario:

* I open up a file that hasn't been edited in a while
* I make one small change
* Black makes a dozen other changes and screws up my little diff

We did a similar 'big bang' reformat with Prettier in the front-end (#1726) and it hasn't caused more issues in practice than any other refactor might. The previous commit is always a click away at least in github: 

image

> Also, as far as I know it only works with Python 3.x, so having it as a mandatory or pre-commit hook will force everybody to use python 3.x for development. That might not be a bad thing, but it's good to remark it

Kind of, but not quite. Py2 and Py3 need to both be available on the path, but you can still develop in Py2.

Turns out that pre-commit runs its hooks in separate virtual environments, so the dev environment can be Py2 and still use Black: https://github.com/learningequality/kolibri/pull/5286/files#diff-82c3b36709926de3a1b27bea0d8258ccR2

Made a note of this here: https://github.com/learningequality/kolibri/pull/5286/files#diff-9bc8bf6e8d9db020cefce1fbc0426795R77

rtibbles:
I am fine with the big bang reformat, for much the reasons you outline @indirectlylit.

I think now would be a good time to bring our pre-commit hooks and our pylinting test build into line as well. At the moment, our pre-commit hooks are more rigorous than our pylint test step, could you also update our pylinting travis steps to apply all of the pre-commit hooks as well, so that we enforce as well as facilitate the linting?

aronasorman:
> Turns out that pre-commit runs its hooks in separate virtual environments, so the dev environment can be Py2 and still use Black: https://github.com/learningequality/kolibri/pull/5286/files#diff-82c3b36709926de3a1b27bea0d8258ccR2

TIL! That makes it significantly more palatable on Studio too.

indirectlylit:
> could you also update our pylinting travis steps

I'll take a look at that. I also wanted to update our devserver to format in real-time, similar to prettier, when developing in py3

rtibbles:
> I also wanted to update our devserver to format in real-time, similar to prettier, when developing in py3

Could we use the virtualenv trick from the precommit hook to run the formatter in real time regardless of which virtualenv you were currently in?

indirectlylit:
maybe if we can find where `pre-commit` stores its venv and then hack into those site-packages, but it feels a bit wrong to mess with those internals...

indirectlylit:
suppose we could try to mimic what pre-commit does. sounds like a lot of effort :)

indirectlylit:
Updated PR to resolve conflicts and apply formatting.

Haven't yet added Travis or devserver live update support, but perhaps should do these in a second PR?

rtibbles:
Can also do each in a separate PR, fine by me.

rtibbles (approved):

Issue #5280, opened 2019-03-16
Remove obsolete code related to app bar icon


MisRob:
When working on #5279 I noticed that following place seems to be dead https://github.com/learningequality/kolibri/blob/develop/kolibri/core/assets/src/views/AppBar.vue#L24-L27

It was introduced by this [commit](https://github.com/learningequality/kolibri/commit/2566eac930198119ef34332796d42d762378e668) and seems to have dependency on certain css variables that I wasn't able to find anywhere in current version. After some initial examination, it seems that there is no way how to reach this place from any parent component.

PR #5279, opened 2019-03-16
Adjust toolbar icon and title spacing

MisRob:
### Summary

Fix left and right margins of AppBar hamburger as requested in #4931.

I noticed that applying new style within AppBar component only would introduce inconsistency (loud especially when icon is in hovered state) because exactly the same icon + title pattern occurs at more places that use KeenUI toolbar as a base component. So I tried to apply new style to all of them, but I am not sure if that's desired so let's see what you think, I can just move the style back to AppBar only. @khangmach could you please check it? I will ask Aron for access to mockups meanwhile :)

I didn't want to pollute global styles for such purpose so I created a base component in core lib that could serve as a wrapper for all common toolbar modifications.

#### App bar

Before
AppBar-before

After
AppBar-after
AppBar-after-left
AppBar-after-right

#### Immersive toolbar
Before
ImmersiveToolbar-before

After
ImmersiveToolbar-after

#### Progress toolbar

Before
ProgressToolbar-before

After
ProgressToolbar-after

### Reviewer guidance

- check app toolbar
- visit setup wizard to check progress toolbar
- visit a lesson item to check immersive toolbar

### References

Solves #4931.

----

### Contributor Checklist

PR process:

- [x] PR has the correct target branch and milestone
- [x] PR has 'needs review' or 'work-in-progress' label
- [x] If PR is ready for review, a reviewer has been added. (Don't use 'Assignees')
- [ ] If this is an important user-facing change, PR or related issue has a 'changelog' label
- [ ] If this includes an internal dependency change, a link to the diff is provided

Testing:

- [x] Contributor has fully tested the PR manually
- [x] If there are any front-end changes, before/after screenshots are included
- [ ] Critical user journeys are covered by Gherkin stories
- [ ] Critical and brittle code paths are covered by unit tests

### Reviewer Checklist

- Automated test coverage is satisfactory
- PR is fully functional
- PR has been tested for [accessibility regressions](http://kolibri-dev.readthedocs.io/en/develop/manual_testing.html#accessibility-a11y-testing)
- External dependency files were updated if necessary (`yarn` and `pip`)
- Documentation is updated
- Contributor is in AUTHORS.rst

PR #5287, opened 2019-03-19
Remove obsolete code related to app bar title icon

MisRob:
### Summary

When working on #5279 I noticed that the following place seems to be dead code related to [one of previous implementations of the app bar](https://github.com/learningequality/kolibri/blob/2566eac930198119ef34332796d42d762378e668/kolibri/core/assets/src/views/app-bar/index.vue). It was probably meant for customizing a title icon but in a current version, it cannot be even reached from any parent component and hamburger is rendered there all the time.

### References

 #5280

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/core/assets/src/views/AppBar.vue
+++ b/kolibri/core/assets/src/views/AppBar.vue
@@ -21,11 +21,6 @@
         />
       </UiIconButton>
 
-      <div>
-        <div class="app-bar-title-icon"></div>
-        {{ title }}
-      </div>
-
       <div slot="actions">
         <slot name="app-bar-actions"></slot>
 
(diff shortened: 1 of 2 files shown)


jonboiser:
Is this deleted div not visible because it doesn't have a slot  within the UiToolbar?

MisRob:
This div is visible actually even when there's no explicitly defined slot - it's propagated down to KeenUI toolbar through the default slot. It's just difficult to recognize it because the title is the same as the title passed in through `title` prop to KeenUI toolbar. So, if you override `{{ title }}` directly in the component, it would be visible if I am not mistaken.

It's totally possible that I missed something, just still can't see any way how to use this piece of code for some customization of AppBar (title is the same at both places, and icon is simply inside the AppBar component). In case someone would need to customize a title icon, I think that a different implementation would be needed with a current version of the component (defining slots within AppBar or something similar and using hamburger only as a default icon).

jonboiser (approved):

PR #5296, opened 2019-03-21
Calculate CherryPy default thread pool depending on the server memory

jredrejo:
### Summary
Sets the default value for CherryPy thread pool depending on the server memory.
Based on tests done on several kolibri installations, it's clear the default value of 10 threads is not enough if there are many connections. However, more threads means more memory in use.
This PR uses a simple rule to set the threads for servers having at least 2 Gb of RAM.
Less than 10 Gb will leave the default  CherryPy value of 10.

Notice this code must not have any effect on MacOS as psutils is not supported in this OS.

### Reviewer guidance

`kolibri start` should work properly in Windows, Linux & MacOs
- In case options.ini has a value set for the threads, it must be kept
- The value for the thread pool must depend on the memory of the machine running kolibri

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/utils/options.py
+++ b/kolibri/utils/options.py
@@ -5,6 +5,29 @@ from configobj import ConfigObj
 from configobj import flatten_errors
 from configobj import get_extra_values
 from validate import Validator
+try:
+    import kolibri.core.analytics.pskolibri as psutil
+except NotImplementedError:
+    # This module can't work on this OS
+    psutil = None
+
+
+def calculate_thread_pool():
+    """
+    Returns the default value for CherryPY thread_pool
+    It is calculated based on the best values obtained in
+    several partners installations.
+    The value must be between 10 (default CherryPy value) and 200.
+    Servers with more memory can deal with more threads.
+    Calculations are done for servers with more than 2 Gb of RAM
+    """
+    pool = 10
+    if psutil:
+        total_memory = psutil.virtual_memory().total / pow(2, 30)  # in Gb
+        if total_memory > 2:
+            pool = int(45 * total_memory - 80)  # 10 for 2 Gb,100 for 4 Gb
+            pool = 200 if pool > 200 else pool
+    return pool
 
 
 option_spec = {
@@ -44,7 +67,7 @@ option_spec = {
         },
         "CHERRYPY_THREAD_POOL": {
             "type": "integer",
-            "default": 10,
+            "default": calculate_thread_pool(),
             "envvars": ("KOLIBRI_CHERRYPY_THREAD_POOL",),
         },
         "CHERRYPY_SOCKET_TIMEOUT": {
--- a/kolibri/utils/options.py
+++ b/kolibri/utils/options.py
@@ -1,5 +1,6 @@
 import logging.config
 import os
+import sys
 
 from configobj import ConfigObj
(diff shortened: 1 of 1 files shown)


kolibri/utils/options.py
```diff
@@ -5,6 +5,29 @@
...
+    Returns the default value for CherryPY thread_pool
+    It is calculated based on the best values obtained in
+    several partners installations.
+    The value must be between 10 (default CherryPy value) and 200.
+    Servers with more memory can deal with more threads.
+    Calculations are done for servers with more than 2 Gb of RAM
+    """
+    pool = 10
+    if psutil:
+        total_memory = psutil.virtual_memory().total / pow(2, 30)  # in Gb
```
rtibbles, line 27: Will virtual memory include swap space?

kolibri/utils/options.py
```diff
@@ -5,6 +5,29 @@
...
+        if total_memory > 2:
+            pool = int(45 * total_memory - 80)  # 10 for 2 Gb,100 for 4 Gb
```
rtibbles, line 29: Is there a reason not to have a higher floor, especially given the issues we have seen of sockets not closing?

kolibri/utils/options.py
jredrejo: No, it's only physical memory. We can add swap as well, but If the server expands to use swap memory I think it would be too slow.

rtibbles: OK, the `virtual_memory` part was making me think it might include swap space - I too thought that would be too slow!

kolibri/utils/options.py
jredrejo: If you refer to the default value of 10, I just wanted to keep the default value CherryPy provides ( https://github.com/cherrypy/cherrypy/blob/master/cherrypy/_cpserver.py#L78 ) I considered the authors have reasons to keep it so low.

jredrejo: Looking at the comments after the hack session it seems having a starting 60 threads would be a better option, right?

rtibbles: I think so - 60 was definitely better than 10 on the Pi, I don't know if that is the definitive number, but it was also very good performance wise for 25ish users on the Windows laptop.

kolibri/utils/options.py
```diff
@@ -1,10 +1,36 @@
...
+            pool = 200 if pool > 200 else pool
```
rtibbles, line 30: I did some local tests on my machine (only using siege on `127.0.0.1:8080`), and increasing the thread pool beyond 100 seemed to have a negative impact on performance. Is it possible that after a certain amount of threads, the context switching issues that we had previously seen (and made us reduce the default thread pool value) might start to dominate?

I wonder if we should start at 50, and then max out at 100 to be a little more conservative in this PR?

indirectlylit (approved):
tested on Mac and this works fine. Someone else should test on linux but the change looks great!

kolibri/utils/options.py
jredrejo: In the tests we did in Honduras, with 8 Gb of RAM, 150 was working better than 100.

PR #5297, opened 2019-03-21
Update scanforcontent command and annotation to handle missing files properly

rtibbles:
### Summary
* Updates local file annotation to mark localfiles as unavailable when they do not exist on disk.
* Updates the scanforcontent command to import all locally available content databases and 

### Reviewer guidance
Does the `scanforcontent` command now mark content nodes as unavailable if their primary files are removed?

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/core/content/management/commands/scanforcontent.py
+++ b/kolibri/core/content/management/commands/scanforcontent.py
@@ -1,9 +1,16 @@
+import logging
+
 from django.core.management.base import BaseCommand
 
-from ...utils.annotation import set_local_file_availability_from_disk
-from ...utils.annotation import update_content_metadata
+from ...utils.annotation import annotate_content
+from ...utils.channel_import import FutureSchemaError
+from ...utils.channel_import import import_channel_from_local_db
+from ...utils.channel_import import InvalidSchemaVersionError
 from ...utils.channels import get_channel_ids_for_content_database_dir
+from kolibri.core.content.models import ChannelMetadata
 from kolibri.core.content.utils.paths import get_content_database_dir_path
+
+logger = logging.getLogger(__name__)
 
 
 class Command(BaseCommand):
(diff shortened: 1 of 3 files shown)


indirectlylit (approved):
works great! Requires a restart of the server - I assume this is expected?

cc @radinamatic we may want to document this eventually

indirectlylit:
retargeting to `develop`. Let's wait to merge until `develop` has been synced with `0.12.x`

radinamatic:
Yes!!! Awesome to have this, will be added to user docs in the following update (still have to integrate a couple of pending notes from the most recent one).

rtibbles:
In spite of the merge of 0.12 into develop, this still has more commits than necessary, will rebase to develop.

radinamatic:
I don't recall ever actually documenting this, but so much has changed since then. @rtibbles Is there anything form here that is worth adding to 0.13 docs update?

rtibbles:
I think it can be added, but with the caveat that it will override any granular import decisions previously made, so it should only be used for initial setup of a server, not during usage.

indirectlylit:
we can wait until after the 0.13.0 crunch to document this

radinamatic:
Ok, 0.13.0 crunch is over, is it time to document this? @indirectlylit 

* What is the use case for this?
* `kolibri manage scanforcontent`?

rtibbles:
Basically, scanforcontent is for when you want content availability to be determined entirely by what is on disk. It will wipe out any user determined 'availability'.

Its main use case is in copying large amounts of files directly into a content folder, run the command, and then they will be marked as available.

indirectlylit:
> I think it can be added, but with the caveat that it will override any granular import decisions previously made, so it should only be used for initial setup of a server, not during usage
...
> It will wipe out any user determined 'availability'

Perhaps we should document this command in the `kolibri --help` rather than the user docs, since it's so risky? Even on initial setup, it might not properly reflect what users want for some large channels like KA  (due to the Indian curriculum)

PR #5306, opened 2019-03-22
Limit threads to 50 to 100 range.

rtibbles:
### Summary
Follow up from #5296 to limit threads to a range of 50 to 100.

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/utils/options.py
+++ b/kolibri/utils/options.py
@@ -25,7 +25,7 @@ def calculate_thread_pool():
     MAX_POOL = 150
     if psutil:
         MIN_MEM = 2
-        MAX_MEM = 4
+        MAX_MEM = 6
         total_memory = psutil.virtual_memory().total / pow(2, 30)  # in Gb
         # if it's in the range, scale thread count linearly with available memory
         if MIN_MEM < total_memory < MAX_MEM:
--- a/kolibri/utils/options.py
+++ b/kolibri/utils/options.py
@@ -22,14 +22,19 @@ def calculate_thread_pool():
     Servers with more memory can deal with more threads.
     Calculations are done for servers with more than 2 Gb of RAM
     """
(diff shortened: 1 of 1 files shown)


kolibri/utils/options.py
```diff
@@ -22,14 +22,19 @@ def calculate_thread_pool():
...
+    pool = MIN_POOL
     if psutil:
+        min_mem = 2
+        max_mem = 4
         total_memory = psutil.virtual_memory().total / pow(2, 30)  # in Gb
-        if total_memory > 2:
-            pool = int(45 * total_memory - 80)  # 10 for 2 Gb,100 for 4 Gb
-            pool = 200 if pool > 200 else pool
+        if min_mem < total_memory < max_mem:
+            pool = int(MAX_POOL * float(total_memory) / max_mem)
```
indirectlylit: This I think works because `MIN_POOL / MAX_POOL` currently equals `min_mem / max_mem` but I believe it may not work as expected otherwise. perhaps something like:

```python
MIN_POOL + (MAX_POOL - MIN_POOL) * float(total_memory) / max_mem
```

rtibbles: Yes, can update.

jredrejo (commented):
In the tests we did in Honduras, with 8 Gb of RAM, 150 was working better than 100.
I would not limit it to 100, it's too low from my pov. I had not tested 200, but 150 worked better than 100 if the server can handle it.

indirectlylit:
pushed commit cbf60605015838e205095f875097c840b58fcbb6

- increase max thread count to 150
- use multiple return statements instead of updating a local variable
- clean up comments

indirectlylit (approved):

Issue #5196, opened 2019-02-26
Use appropriate permissions for ClassSummary endpoint


jonboiser:
### Observed behavior

In #5209, we fixed an issue where Promise rejections from unauthenticated requests to the notifications API was being swallowed by UI code (and not showing an AuthMessage component to users). We also added the most basic Permissions to the ClassSummary endpoint so that it would also return a 400 error for unauthenticated users.

However, the Permissions for Class Summary are too basic, and should be refined to be identical to permissions to AttemptLog, Groups, Exams, Lessons APIs, etc. (stuff ClassSummary aggregates).

There is a class `KolibriReportPermissions` that is used on some of the above-referenced APIs, that might be extended (or used as-is, possibly) on ClassSummary.

### Expected Behavior

ClassSummary permissions are exactly the same as endpoints that provide access to coach-only resources (exams, lessons, groups, logs, etc.)

### (Old) Observed behavior

1. When attempting to view a Coach page when an anonymous user, there is no "AuthMessage" component that explains why the page is not accessible. It's just a blank page. (No app bar either)
1. On Facility and Device pages, you do see that message, but on Device it's actually wrapped in KPageContainer (in Facility it's not)

![screen shot 2019-02-26 at 2 24 24 pm](https://user-images.githubusercontent.com/10248067/53450947-6f049f00-39d2-11e9-91e1-d99d1dcdd1e5.png)

### Expected behavior

1. When unauthenticated, accessing the coach pages shows an AuthMessage

### Context
0.12beta9

indirectlylit:
might relate to work done in https://github.com/learningequality/kolibri/issues/5014 ?

jonboiser:
I just tried this out, and the AuthMessage is showing up on coach pages, but you can also see the class name on the app bar.

It seems that the classSummary endpoint is open to unauthenticated users (hence the class name). It is the notifications endpoint that fails with the 403 error, which ends up showing the AuthMessage.

The Classroom API returns a 200, but it is an empty array (since anonymous users are not coaches or admins to any class).

rtibbles:
Some fixes in #5209 - still need proper permissions on the class summary endpoint.

rtibbles:
Fixed in #5308

Issue #5305, opened 2019-03-22
Better error handling for various edge cases (like 404s from deleting non-existent records)


jonboiser:
### Observed behavior

Here are some places where we could replace the catch-all 'AppError' (appears with 'handleApiError' dispatches) with in-UI error messages (or no error at all).

1. ~Deleting a user from the facility when that user has already been deleted. (No error should be shown)~
1. Deleting a class that has already been deleted
1. ~Visiting a Coach link for a class that has been deleted (it's handled like a 403 error)~
1. ~Repeat for the various other things that can be deleted in Kolibri, as these will all lead to 404 errors of some kind.~
1. ~Assigning an existing user a username that has already been given (by another user in another session, for example)~
1. ~Creating a new user in a similar situation as above. There is incorrect logic that is intended to handle this in UserCreateModal (but the code incorrect, so this situation ends up showing the AppError page)~
1. Attempting to access a classroom page when you are a coach not assigned to the classroom.
1. ~Attempting to view a Classroom List or Assignments page as a Learner when not logged in (#5360 )~

These errors fall into several broad categories:

1. Attempting access a non-existent object (e.g. because it's been deleted or a UUID has been mis-typed). Issue #5666 
1. Attempting to modify a deleted object
1. Attempting to delete a deleted object
1. Attempting to create an object, and the new object violates some rule (e.g. unique name)
1. Attempting to access an object for which you don't have the proper permissions.

In general, look where `handleApiError` is used and ask whether this can be replaced with no error, or an error placed within the current UI.

### Expected behavior

In the edge cases described above, errors are handled as prescribed.

### User-facing consequences

…

### Errors and logs

…

### Steps to reproduce

…

### Context

…

indirectlylit:
do these situations need design and/or user-facing strings?

jonboiser:
Some of these don't require any new UI, like deleting something that's already deleted. (e.g. if you deleted an already-deleted faciltiyuser, dont show an error, remove the facilityuser from the table, the app user doesn't need to see anything).

We don't handle going to links for non-existent things (e.g. visiting a class homepage for deleted class), so that will require new design and strings.

I've asked @MisRob 	if she can look for more overlooked edge cases to complete the list in the OP, and then maybe we can handle the different categories in more depth, design-wise.

MisRob:
[...]

PR #5308, opened 2019-03-23
Class summary permissions

MisRob:
### Summary

- a classroom summary can be accessed only by admins, facility coaches or class coaches granted with the classroom permissions
- FE authorized messages condition fix

A couple of first commits include some related refactoring of `KolibriReportPermissions`. At first, I tried to use it as is, but I found out that I would have to add a third view condition in it to be able to use it and it wasn't very readable after this change. Also, every view seemed to have a quite different auth conditions. So I divided it into two permissions classes - `LessonReportPermissions` and `ClassroomNotificationsPermissions`. Initially, I was planning re-using `LessonReportPermissions` for a new class summary auth checks but it showed up that there are still quite important differences and using another permissions class seemed like a bit better option to me.

#### Manual verification steps performed

I visited classroom summary as a facility admin (can access), a facility coach (can access), a class coach (can access), an another class coach (cannot access), a learner (cannot access) and an anonymous user (cannot access).

#### Screenshots of UI changes introduced with this PR

| Learner - before  | Learner - after |
| ------------- | ------------- |
| Learner-before  | Learner-after
  |

| Another class coach - before  | Another class coach - after |
| ------------- | ------------- |
| AnotherClassCoach-before | AnotherClassCoach-after  |

### Reviewer guidance

- Does new class summary permission make sense? Does it work?
- Do lesson report and classroom notifications permissions still work as expected?

It's my first task related to auth and permissions in Kolibri - I tried to be careful and study all those roles as much as possible, but I don't feel confident about all possible relationships and also it's been a while since I worked on BE so please check everything thoroughly. Thanks in advance :) 
…

### References

Solves #5196 
Follow-up issue #5309
…

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/plugins/coach/api.py
+++ b/kolibri/plugins/coach/api.py
@@ -76,10 +76,25 @@ class LessonReportViewset(viewsets.ReadOnlyModelViewSet):
     queryset = Lesson.objects.all()
 
 
+class ClassroomNotificationsPermissions(permissions.BasePermission):
+    """
+    Allow only users with admin/coach permissions on a collection.
+    """
+
+    def has_permission(self, request, view):
+        collection_id = view.kwargs.get('collection_id')
+
+        allowed_roles = [role_kinds.ADMIN, role_kinds.COACH]
+        try:
+            return request.user.has_role_for(allowed_roles, Collection.objects.get(pk=collection_id))
+        except (Collection.DoesNotExist, ValueError):
+            return False
+
+
 @query_params_required(collection_id=str)
 class ClassroomNotificationsViewset(viewsets.ReadOnlyModelViewSet):
 
-    permission_classes = (KolibriReportPermissions,)
+    permission_classes = (ClassroomNotificationsPermissions,)
     serializer_class = LearnerNotificationSerializer
     pagination_class = OptionalPageNumberPagination
     pagination_class.page_size = 10
--- a/kolibri/plugins/coach/api.py
+++ b/kolibri/plugins/coach/api.py
@@ -38,38 +38,6 @@ class OptionalPageNumberPagination(pagination.PageNumberPagination):
     page_size_query_param = "page_size"
 
 
-class KolibriReportPermissions(permissions.BasePermission):
-
-    # check if requesting user has permission for collection or user
-    def has_permission(self, request, view):
-        if isinstance(view, LessonReportViewset):
-            report_pk = view.kwargs.get('pk', None)
-            if report_pk is None:
-                # If requesting list view, check if requester has coach/admin permissions on whole facility
-                collection_kind = 'facility'
-                collection_or_user_pk = request.user.facility_id
-            else:
-                # If requesting detail view, only check if requester has permissions on the Classroom
-                collection_kind = 'classroom'
-                collection_or_user_pk = Lesson.objects.get(pk=report_pk).collection.id
-
-        else:
-            if isinstance(view, ClassroomNotificationsViewset):
-                collection_kind = 'classroom'
-            else:
-                collection_kind = view.kwargs.get('collection_kind', 'user')
-            collection_or_user_pk = view.kwargs.get('collection_id', view.kwargs.get('pk'))
-
-        allowed_roles = [role_kinds.ADMIN, role_kinds.COACH]
-        try:
-            if 'user' == collection_kind:
-                return request.user.has_role_for(allowed_roles, FacilityUser.objects.get(pk=collection_or_user_pk))
-            else:
-                return request.user.has_role_for(allowed_roles, Collection.objects.get(pk=collection_or_user_pk))
-        except (FacilityUser.DoesNotExist, Collection.DoesNotExist, ValueError):
-            return False
-
-
 class LessonReportPermissions(permissions.BasePermission):
     """
     List - check if requester has coach/admin permissions on whole facility.
--- a/kolibri/plugins/coach/api.py
+++ b/kolibri/plugins/coach/api.py
@@ -70,8 +70,29 @@ class KolibriReportPermissions(permissions.BasePermission):
             return False
 
 
+class LessonReportPermissions(permissions.BasePermission):
+    """
+    List - check if requester has coach/admin permissions on whole facility.
+    Detail - check if requester has permissions on the Classroom.
+    """
+
+    def has_permission(self, request, view):
+        report_pk = view.kwargs.get('pk', None)
+        if report_pk is None:
+            collection_id = request.user.facility_id
+        else:
+            collection_id = Lesson.objects.get(pk=report_pk).collection.id
+
+        allowed_roles = [role_kinds.ADMIN, role_kinds.COACH]
+
+        try:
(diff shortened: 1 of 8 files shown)


jonboiser:
Test cases for me to go through:

- [x] An admin can access all class reports (notifications, exercise details, tabular reports, etc.)
- [x] A facility coach can access all class reports
- [x] A class coach can access class reports if they are assigned to the class
- [x] A class coach cannot access class reports if they are not assigned to the class
- [x] A (non-superuser) learner cannot access any class reports
- [x] A superuser learner can access any class reports (same as admin)
- [x] An anonymous user cannot access any class reports

kolibri/plugins/coach/api.py
```diff
@@ -38,48 +38,53 @@ class OptionalPageNumberPagination(pagination.PageNumberPagination):
...
+    """
+
+    def has_permission(self, request, view):
+        collection_id = view.kwargs.get('collection_id')
+
+        allowed_roles = [role_kinds.ADMIN, role_kinds.COACH]
+
+        try:
+            return request.user.has_role_for(allowed_roles, Collection.objects.get(pk=collection_id))
+        except (Collection.DoesNotExist, ValueError):
```
jonboiser, line 80: If the Collection/Classroom does not exist, does this return a 403 or a 404?

jonboiser: Okay, it returns a 403. We probably don't have a special UI for this, but it might be nice to separate both situations (later on). Otherwise if you go to a class that was deleted etc., it will just show a "are you logged in" message (when you actually are logged in).

kolibri/plugins/coach/test/helpers.py
```diff
@@ -0,0 +1,55 @@
+from kolibri.core.auth.models import FacilityUser
+
+
+def create_learner(username, password, facility, classroom=None, learner_group=None):
```
jonboiser, line 4: Maybe in the future we can just assume every test user is using "DUMMY_PASSWORD" so it can be set by default in these kinds of helpers.

jonboiser: Same with any other constants, like the facility

jonboiser:
@MisRob can you check if we can/should deprecate any Viewsets that may have been replaced by ClassSummary? I'm not sure if we are still using some of the resource reports like lessonreport, examreport, etc.

kolibri/plugins/coach/api.py
MisRob: Oh yes, that's right. I wasn't thinking about that much and kept the original `KolibriReportPermissions` logic. But definitely agree with you, it would be nice to return 404 instead and then display another message for such situation on FE. I can open an issue.

kolibri/plugins/coach/test/helpers.py
MisRob: 👍

kolibri/plugins/coach/api.py
jonboiser: There's an issue here for error-handling in general #5305

MisRob:
@jonboiser Sure sure, would you prefer to have it done here or in another PR? In case it would be here, the PR would have to wait a while until I'm back coding, but if a couple of days is not a problem, then it's fine, too.

jonboiser:
Yeah, if we want to remove any code, we should do it in a separate PR.

MisRob:
Okay agreed then, it's the first item on my to-do list.

kolibri/plugins/coach/test/helpers.py
MisRob: I've just written it down and am going to include it in tests simplifications proposal.

kolibri/plugins/coach/api.py
```diff
@@ -38,48 +38,53 @@ class OptionalPageNumberPagination(pagination.PageNumberPagination):
...
-            if report_pk is None:
-                # If requesting list view, check if requester has coach/admin permissions on whole facility
-                collection_kind = 'facility'
-                collection_or_user_pk = request.user.facility_id
-            else:
-                # If requesting detail view, only check if requester has permissions on the Classroom
-                collection_kind = 'classroom'
-                collection_or_user_pk = Lesson.objects.get(pk=report_pk).collection.id
-
+        report_pk = view.kwargs.get('pk', None)
```
jonboiser: Is this the typical way to check if it's a list- or detail- view?

jonboiser (approved):
Looks good to me! Breaking up the permissions does seem to make the PermissionsClasses a lot simpler.

indirectlylit:
this all looks good to me, but I also haven't had much experience with the permission classes. @rtibbles would you mind taking a quick look when you're feeling better?

MisRob:
Thanks everyone for checking it.

I was thinking about how to generalize report permissions if needed meanwhile while keeping things simple and I think it could be done by using principles similar to Django mixins/inheritance. Basically, rest's `BasePermissions` could be extended to `ReportPermissions` and this class could be used as a base class for all viewsets using a similar report permissions logic, but every viewset's permissions class would need to specify collection ID itself (and possibly more information if needed). To explain it on an example, permissions for all viewsets I was working with here might look like:

```
class LessonReportPermissions(common.ReportPermissions):
    def get_collection_id(self, request, view):
        report_pk = view.kwargs.get('pk', None)
        if report_pk is None:
            return request.user.facility_id

        return Lesson.objects.get(pk=report_pk).collection.id

class ClassroomNotificationsPermissions(common.ReportPermissions):
    def get_collection_id(self, request, view):
        return view.kwargs.get('collection_id')

class ClassSummaryPermissions(common.ReportPermissions):
    def get_collection_id(self, request, view):
        return view.kwargs.get('pk')
```

and `ReportPermissions` base class would handle the rest of the permissions logic. Right now I think it's not needed probably because there are only about 4 common lines, but I guess it might be useful when implementing more granular error handling or another logic common to all report viewsets.

If you would like to see this interface, I can open a follow-up.

kolibri/plugins/coach/api.py
rtibbles: Yeah, I think in general, returning `True` from the permissions, and then returning a 404 from the Viewset is a good pattern to follow.

kolibri/plugins/coach/class_summary_api.py
```diff
@@ -207,11 +209,23 @@ def data(Serializer, queryset):
...
+    """
+
+    def has_permission(self, request, view):
+        classroom_id = view.kwargs.get('pk')
+        allowed_roles = [role_kinds.ADMIN, role_kinds.COACH]
+
+        try:
+            return request.user.has_role_for(allowed_roles, Collection.objects.get(pk=classroom_id))
+        except (Collection.DoesNotExist, ValueError):
+            return False
```
rtibbles, line 224: As above, can pass `True` here, as the first line of the Viewset will give a 404.

rtibbles:
This looks good to me, and I agree with the comments re: handling permissions for missing items.

The refactor is akin to an appendectomy, there are historical reasons why we had the `KolibriReportPermissions` class, but they no longer apply!

PR #5313, opened 2019-03-26
Run integration testing script inside the docker environment

### LINKED ISSUE #4953: Isolate the Kolibri integration testing spreadsheet build inside the docker environment
### Observed behavior
I have this [PR](https://github.com/learningequality/kolibri/pull/4544) that's using a virtualenv to run the scripts and generate the Integration testing spreadsheet at the BuildKite.

@indirectlylit [comment](https://github.com/learningequality/kolibri/pull/4544#discussion_r254383845) to that PR to isolate the build inside the docker environment.

### Expected behavior

>* the build would be isolated in a docker environment, so a python virtualenv would not be needed
>* the dependencies for this pipeline step would be defined in a pip `requirements.txt` file or a Pipenv `Pipfile`
>* the individual steps (install dependencies, run `create_integration_testing_worksheet.py`, upload artifacts) are specified as multiple commands in the 'Create integration testing' block of the `pipeline.yml` file

mrpau-richard:
### Summary
I use docker environment to run the integration testing spreadsheet script at the BuildKite.

This is a work in progress PR.

### Reviewer guidance
The integration testing spreadsheet script should run inside the docker environment.

### References
This fixes #4953 

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated


indirectlylit:
cc @DXCanas would you mind reviewing this work? I'm not super familiar with Docker and the build process, and would like to get your input

DXCanas:
@indirectlylit sure! Is it ready for review?

.buildkite/build_worksheet.sh
```diff
@@ -2,46 +2,15 @@
...
-virtualenv -p python3 env
-if [ $? -ne 0 ]; then
-    echo ".. Abort!  Can't create virtualenv."
-    exit 1
-fi
+# Go to parent path (where kolibri sources are assumed to live)
+cd "$DIR"
+cd ..
+PARENT_PATH=`pwd`
+cd "$PARENT_PATH"
```
DXCanas: Seems like a lot of unnecessary flip flopping around directories here. Is there any reason for this?

docker/build_testing_sheet.dockerfile
```diff
@@ -0,0 +1,19 @@
...
+COPY . /kolibri
+
+COPY ./kolibri/.buildkite $GOOGLE_SPREADSHEET_CREDENTIALS
+
+CMD cd /kolibri && \
+    pip install -r requirements/testing_sheet.txt
+
+VOLUME /kolibridist/
+
+CMD export GOOGLE_SPREADSHEET_CREDENTIALS=./kolibri/.buildkite/*.json && \
```
DXCanas: Docker's `run` command lets you do this, and includes checks for variable definition
https://docs.docker.com/engine/reference/commandline/run/#set-environment-variables--e---env---env-file

docker/build_testing_sheet.dockerfile
```diff
@@ -0,0 +1,19 @@
+FROM ubuntu:bionic
+
+RUN apt-get update -y
+
+COPY . /kolibri
+
+COPY ./kolibri/.buildkite $GOOGLE_SPREADSHEET_CREDENTIALS
```
DXCanas: Could you clarify what this is meant to do?

docker/build_testing_sheet.dockerfile
```diff
@@ -0,0 +1,19 @@
...
+
+CMD cd /kolibri && \
+    pip install -r requirements/testing_sheet.txt
+
+VOLUME /kolibridist/
```
DXCanas: Any particular reason you opted for a VOLUME rather than a runtime bind mount here? @aronasorman is probably in favor of this approach though.

docker/build_testing_sheet.dockerfile
```diff
@@ -0,0 +1,19 @@
+FROM ubuntu:bionic
+
+RUN apt-get update -y
+
+COPY . /kolibri
```
DXCanas: Is the entire Kolibri repo necessary?  It's likely I overlooked something, but the script seems to pull a lot of its data from the github directly, anyway. Copying in the entire repo this early in the stage makes for a lot of duplicated layers. They end up getting cleaned later in the pipeline, but reducing the amount of copies could have significant speed implications.

requirements/testing_sheet.txt
```diff
@@ -0,0 +1,6 @@
+
```
DXCanas, line 1: Is there a rationale behind putting this in the Kolibri repo's `requirements` folder? Seems to me this is pretty independent of the Kolibri project. Might be better to just bake these into the `Dockerfile`

@indirectlylit thoughts on this?

DXCanas (changes requested):
Build fails for me:

`COPY failed: stat /var/lib/docker/tmp/docker-builder275323553/kolibri/.buildkite: no such file or directory`

This seems to be operating under the assumption that all the work happens at build time, ignoring the run-time side of things. Docker has some baked-in functionality in their `run` command that could clean this up a bit.

I see this potentially taking up a good amount of storage, but it may be inconsequential due to the cleanup command we run later in the pipeline. Besides that, though, the `COPY` command for the entire kolibri directory takes a good while. Correcting that could mean some significant speed improvements.

indirectlylit:
@DXCanas thanks for the review! I believe @mrpau-richard is still on leave, so we shouldn't expect an immediate response

requirements/testing_sheet.txt
indirectlylit: The requirements folder has a lot of 'meta' stuff in it so I don't see any particular issue.

Also the integration gherkin files are in fact part of the Kolibri project - they're checked into the root directory. To me this feels similar to the docs building, testing, and font management requirements.

I think the main question is whether these are _only_ used from within Docker, or if it's also possible to run the migration script outside of docker?

requirements/testing_sheet.txt
```diff
@@ -0,0 +1,6 @@
...
+# These are the requirements to create an Integration testing Spreadsheet
+
+oauth2client
+PyOpenSSL
```
indirectlylit: these should be pinned to specific versions

.buildkite/build_worksheet.sh
mrpau-richard: This is to ensure that I can run or execute a command from the MakeFile when I'm at the `kolibri/.buildkite` path.

docker/build_testing_sheet.dockerfile
mrpau-richard: I'll try this thanks.

docker/build_testing_sheet.dockerfile
mrpau-richard: The integration testing script need the Google api credentials `json` file which is the path of it is in the `GOOGLE_SPREADSHEET_CREDENTIALS ` environment variable to be able to access the script I'll copy it inside the docker environment.

docker/build_testing_sheet.dockerfile
mrpau-richard: Thanks for that information @DXCanas I'll copy on what the script needed in the `./kolibri ` directory.

requirements/testing_sheet.txt
mrpau-richard: It's also possible to run the script outside the docker environment. Running a docker file in the local machine take more time than execute the script directly.

requirements/testing_sheet.txt
mrpau-richard: I'll do it thanks.

mrpau-richard:
@DXCanas & @indirectlylit thanks for reviewing this even if this is a work in progress PR. and apologies for the late response as I'm not active for the last few days.

docker/build_testing_sheet.dockerfile
DXCanas: My understanding, though, is that you're going to copy the contents of the `.buldkite` directory into the _container_, at a directory specified in the environment variable `$GOOGLE_SPREADSHEET_CREDENTIALS`

mrpau-richard: I see what you mean. I'll fix it thanks.

requirements/testing_sheet.txt
DXCanas: I'd like to add that it's only slowness is largely due to the current implementation. Caching could make it just as fast as a native run. I do doubt, though, that it could get any faster than a native run, since the script relies on the GitHub-hosted repo

docker/build_testing_sheet.dockerfile
benjaoming: @DXCanas I was reading about the bind mounts and didn't see why they should be favored or vice versa? Is it important? Is there a recommended blog post about it?

requirements/testing_sheet.txt
benjaoming: Requirements often have to co-exist so it's nice to keep them together, for instance in cases where they depend on each other (like the dev requirements depend on runtime requirements). Also in cases where you'd want to be aware of conflicts. And to simplify version management, so you don't have to look far around the codebase.

.buildkite/pipeline.yml
```diff
@@ -27,23 +27,21 @@ steps:
   - label: Upload artifacts
     command: .buildkite/setup_and_upload_artifact.sh
 
+  - block: "Create integration testing worksheet?"
+  - label: Create integration testing
+    command: .buildkite/build_worksheet.sh
```
benjaoming: The `make target` should be the only entry-point(s) for your job, not the shell script. We should try to achieve a call stack that is more like `pipeline.yml > make > Docker > Python stuff`. And if we can factor the shell script by implementing into our `pipeline.yml` and Makefile, that'd be great, too :)

.buildkite/build_worksheet.sh
```diff
@@ -2,46 +2,15 @@
...
 
-PIP_CMD="$PIP_PATH install gspread==3.1.0"
-echo "Running $PIP_CMD..."
-$PIP_CMD
-if [ $? -ne 0 ]; then
-    echo ".. Abort!  Can't install '$PIP_CMD'."
-    exit 1
-fi
+make docker-build-sheet
```
benjaoming: Looking at this shell  script and how short it is now, could we remove it entirely? In `pipeline.yml`, we call `make docker-build-sheet` followed by `buildkite-agent artifact upload './dist/spreadsheet-link.txt'` (I think maybe there's even a shorter command for that)

.buildkite/create_integration_testing_worksheet.py
```diff
@@ -101,6 +101,7 @@ def create_artifact(str_arg):
     file = open(txt_path, 'w')
     file.write(str_arg)
     file.close()
+    os.chmod(txt_path, 0o777)
```
benjaoming: Nit-picking... but `777` makes a file executable for everyone.. if you use `666`, it will be writeable to everyone, but not executable.

requirements/testing_sheet.txt
benjaoming: Regarding performance, one of the things I'd worked with was to split up our Docker files and Make targets such that building images and running them was a separate thing and that dockers steps and cache entries were re-used.

docker/build_testing_sheet.dockerfile
```diff
@@ -0,0 +1,23 @@
...
+
+VOLUME /kolibridist/
+
+CMD export BUILDKITE_PULL_REQUEST_BASE_BRANCH=$BUILDKITE_PULL_REQUEST_BASE_BRANCH && \
+    export BUILDKITE_BRANCH=$BUILDKITE_BRANCH && \
+    export BUILDKITE_TAG=$BUILDKITE_TAG && \
+    pip3 install -r ./kolibri/requirements/testing_sheet.txt && \
+    python3 ./kolibri/.buildkite/create_integration_testing_worksheet.py && \
+    cd /kolibri/.buildkite && \
+    mv *.txt /kolibridist/
```
benjaoming: Try using `cp` instead of `mv`. This is the only permission-related issue I can think of regarding `remove /var/lib/buildkite-agent/builds/vader-1/learningequality/kolibri/dist/spreadsheet-link.txt: permission denied`

docker/env.list
```diff
@@ -10,6 +10,9 @@ KOLIBRI_WINDOWS_INSTALLER_VERSION=v1.2.1
 # Make sure we don't record these docker runs in pingbacks
 KOLIBRI_RUN_MODE="docker"
 
+# Set the default value of the Google api credentials for the integration testing script.
+GOOGLE_SPREADSHEET_CREDENTIALS=./kolibri/.buildkite/integration-testing-google-api-credential.json
```
benjaoming, line 14: We should maintain entries in `.gitignore` for such secrets.

.buildkite/create_integration_testing_worksheet.py
benjaoming: I'm also wondering about the error from `make docker-clean` that says `Failed to remove "/var/lib/buildkite-agent/builds/vader-1/learningequality/kolibri" (remove /var/lib/buildkite-agent/builds/vader-1/learningequality/kolibri/dist/spreadsheet-link.txt: permission denied)`.

Since normally we didn't have issues removing files generated inside the Docker containers, I think that you should remove this part?

benjaoming (changes requested):
Nice work, let's see if we can simplify it a bit and get rid of the permission error :)

docker/build_testing_sheet.dockerfile
DXCanas: A couple reasons - this volume is being declared at build time, meaning it'll be wiped clean with every build and filled with the contents of that directory. It's the earliest implementation of Docker volumes and it's mostly superseded by the current run-time volume creation.

Generally, we're baking a lot of frequently-changed code during build time and at an early layer. Every rebuild would throw out the subsequent layers.

Runtime bind mounts would be more appropriate here (and in other cases I've described to you) because the cache wouldn't get invalidated by an earlier `COPY`. A later `COPY`  would preserve more cache, but I don't see a benefit in baking source files into the image if they don't need to be there (like if we were using the image for a K8s cluster or something). 

We're using docker for the sole purpose of environment isolation here: why fill the image with files that aren't necessary for environment creation?

mrpau-richard: > Runtime bind mounts would be more appropriate here (and in other cases I've described to you) because the cache wouldn't get invalidated by an earlier COPY. A later COPY would preserve more cache, but I don't see a benefit in baking source files into the image if they don't need to be there (like if we were using the image for a K8s cluster or something).

Let's use the bind mount then.

mrpau-richard:
@DXCanas @benjaoming the permission issue was fixed and I addressed some of your comments I think this PR is ready to merge.

benjaoming (approved):

benjaoming:
Great work @mrpau-richard, thanks for addressing all the feedback - confirming that the spreadsheet link and spreadsheet itself was generated :100:

DXCanas (approved):
Made a quick, single line fix so that we leverage the docker layer cache for pip dependencies. Noted non-blocking faults in #5505, to be worked on at a later time.

indirectlylit:
thanks - merging in.

It looks like there is plenty of cleanup that can be done, but since things are functional it's not an immediate priority.

indirectlylit:
Crap - I totally forgot to fix this PR. It was targeting the wrong branch and brought in some changes that shouldn't be in there

PR #5322, opened 2019-03-28
Don't cause errors when assigning or enrolling users

### LINKED ISSUE #5321: User added twice on the list to enroll in class
### Observed behavior
When admin first selects manually a learner with the checkbox, but then searches for a text using the `Search` field, and already selected learner happens to be among the results and gets selected for the second time, will provoke a `Server Error (500)`.

select-user-twice

### Expected behavior
Ignore the first selection?

### User-facing consequences
UI error

### Errors and logs

[kolibri.log](https://github.com/learningequality/kolibri/files/3015798/kolibri.log)

### Steps to reproduce
1. Go to Facility > Classes > Class N > Enroll Learners
2. Select a learner with checkbox
3. Use the `Search` field to filter with a string that gives as a result the previously selected learner and at least one other.
4. Check the `Select all` checkbox 
5. Click `Confirm` 

### Context
0.12.2.beta3
Ubuntu/Firefox

…

cc @rtibbles

rtibbles:
### Summary
* Uniqifies learner and coach lists before sending them to the backend to prevent constraint violations.
* Force fetches the classroom model to update on coach assignment.

### Reviewer guidance
* Follow steps to replicate from #5321 
* Enroll coaches in a class, return to the class page, see coaches enrolled.

### References
Fixes #5321 

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/plugins/facility_management/assets/src/modules/classAssignMembers/actions.js
+++ b/kolibri/plugins/facility_management/assets/src/modules/classAssignMembers/actions.js
@@ -1,12 +1,13 @@
 import { MembershipResource, RoleResource } from 'kolibri.resources';
 import { UserKinds } from 'kolibri.coreVue.vuex.constants';
+import uniq from 'lodash/uniq';
 
 export function enrollLearnersInClass(store, { classId, users }) {
   return MembershipResource.saveCollection({
     getParams: {
       collection: classId,
     },
-    data: users.map(userId => ({
+    data: uniq(users).map(userId => ({
       collection: classId,
       user: userId,
     })),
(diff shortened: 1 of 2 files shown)


jonboiser (approved):

jonboiser:
The front-end change works, but I think the viewset should probably handle this gracefully too.

radinamatic:
Works well when I pull your branch locally, will test with the asset now, just to be sure!

radinamatic (approved):
Tested on Ubuntu 16.04, looking good! 👍

rtibbles:
> The front-end change works, but I think the viewset should probably handle this gracefully too.

A 500 is definitely not the correct response, I feel like constraint violations should in general return a 400.

Alternatively, I assume you are proposing a non RESTful endpoint that just assigns the appropriate memberships to the passed in user/collection pairs, which also works, but this will do for now!

PR #5325, opened 2019-03-28
update 'es' strings

indirectlylit:
### Summary

activate!

### Reviewer guidance

re-ran an extraction from crowdin on 0.12.2 branch and this is the only change

### References

…

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated


jredrejo (approved):

radinamatic:
@indirectlylit Do you know why the strings with this same context appear twice on Crowdin?
I have an impression that many of the identical strings, with the identical context, often appear twice... Yes, one can reuse the translation of one for the other, but just wondering why 🤔

radinamatic (approved):
Strings look updated correctly! 👍

indirectlylit:
>  Do you know why the strings with this same context appear twice on Crowdin? I have an impression that many of the identical strings, with the identical context, often appear twice... Yes, one can reuse the translation of one for the other, but just wondering why

Can you send me screenshots showing me in what situation you're talking about? I've made settings changes but I'm not always sure how it affects the translator experience

What you're experiencing probably relates to me changing this setting:

> image

I did this as an attempt to have more control over crowdin's translation memory

radinamatic:
> Can you send me screenshots showing me in what situation you're talking about? I've made settings changes but I'm not always sure how it affects the translator experience.

Recorded an animation with steps to find the strings in question.

Translator's experience is not necessarily bad, a part from having to reuse the identical translations over and over. What I'm more wondering is how much is our word count inflated...? 🤔 

seemingly-duplicates

indirectlylit:
thanks, I'll investigate this.

the word count should not be inflated

radinamatic:
> the word count should not be inflated

Good to know! 😌

indirectlylit:
@radinamatic FYI I looked into this today, and I believe the observed behavior is correct.

I just did a test upload of the current state of `develop`. Here's what Spanish looks like:

image

(It currently reports 7317 words total which is actually lower than 0.12.2, and I need to look into why exactly that is.)

To take one example, I searched for the phrase "A class with that name already exists". There are four results:

image

The four results come from two IDs (`ClassCreateModal.duplicateName`, `ClassRenameModal.duplicateName`) and two branches (`release-v0.12.2` and `test-upload`).

In this case, both strings got automatically copied over from their prior versions because both the IDs and the English phrases stayed the same. If either had changed, they strings would show up as untranslated.

Filtering to just the `test-upload` branch shows only the two instances as we would expect:

image

Now the final question is: do we really want these two instances represented twice in each branch? They appear in two different contexts (creating a new class and renaming an existing class). Maybe there are situations where the message might differ based on context, but it seems unlikely. However in this case there are only two instances, so we don't gain much by making them share a single source.

However there are other instances where we'd gain a lot, such as commonly re-used strings like "Save" and "Cancel". I'm starting to think we should make globals for our most commonly re-used strings:

image

radinamatic:
Thank you for investigating this, @indirectlylit!

> do we really want these two instances represented twice in each branch? 
...
> I'm starting to think we should make globals for our most commonly re-used strings

I'm not opposed to adopting the strategy of making globals, but let's also plan for the overhead that will be necessary to break out those globals when we realize that the context warrants different translations, even though the original string is identical.

Issue #4088, opened 2018-07-31
Upgrade eslint-plugin-vue and Configure it to use consistent ordering for custom Vue options


jonboiser:
1. ~Update eslint-plugin-vue to the latest version. Some of the new rules will lead to large changes, so we can apply them gradually.~ Fixed in #5331
1. Our components have two options `metaInfo` and `$trs` that are not considered by the [order-in-components](https://github.com/vuejs/eslint-plugin-vue/blob/master/docs/rules/order-in-components.md) rule. So they appear in different places in different components. Vue-router hooks do not seem to have a prescribed order either. To make them consistent, we can specify a desired place for these three (`$trs`, router hooks, `metaInfo`).

rtibbles:
Second part resolved in #5359

PR #5331, opened 2019-03-30
Upgrade eslint plugin vue

MisRob:
### Summary

- upgrade eslint-plugin-vue to the latest version
- manually fix issues regarding a new rule vue/return-in-computed-property
- run auto fix

### Reviewer guidance

- run lint
- are new return statements OK?

### References

Solves: first part of #4088 
Follow-ups: #5337, #5338

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/plugins/device_management/assets/src/views/SelectContentPage/index.vue
+++ b/kolibri/plugins/device_management/assets/src/views/SelectContentPage/index.vue
@@ -185,6 +185,8 @@
         if (Object.values(ContentWizardErrors).includes(this.status)) {
           return this.status;
         }
+
+        return undefined;
       },
       channelOnDevice() {
         return this.channelIsInstalled(this.transferredChannel.id) || {};
--- a/kolibri/plugins/learn/assets/src/views/ContentPage.vue
+++ b/kolibri/plugins/learn/assets/src/views/ContentPage.vue
@@ -230,6 +230,8 @@
         if (breadcrumbs.length > 0) {
           return breadcrumbs[breadcrumbs.length - 1];
         }
+
+        return undefined;
       },
       progress() {
         if (this.isUserLoggedIn) {
--- a/kolibri/plugins/learn/assets/src/views/ContentPage.vue
+++ b/kolibri/plugins/learn/assets/src/views/ContentPage.vue
@@ -79,10 +79,14 @@
           </UiIconButton>
           <div v-if="licenceDescriptionIsVisible" dir="auto" class="license-details">
             <template v-if="translatedLicense">
-              <p class="license-details-name">{{ licenseName }}</p>
+              <p class="license-details-name">
+                {{ licenseName }}
+              </p>
               <p>{{ licenseDescription }}</p>
             </template>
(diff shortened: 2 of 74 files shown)


indirectlylit:
cool. were these changes made automatically, or did you need to manually update them to get rid of the linting errors?

MisRob:
It was both - you can find manual fixes for the new rule in the second commit and auto fixes in the third, last, commit.

kolibri/plugins/device_management/assets/src/views/SelectContentPage/index.vue
```diff
@@ -185,6 +185,8 @@
         if (Object.values(ContentWizardErrors).includes(this.status)) {
           return this.status;
         }
+
+        return undefined;
```
jonboiser, line 189: Have this return the same type as `this.status`. Is it a string?

kolibri/plugins/learn/assets/src/views/ContentPage.vue
```diff
@@ -230,6 +234,8 @@
         if (breadcrumbs.length > 0) {
           return breadcrumbs[breadcrumbs.length - 1];
         }
+
+        return undefined;
```
jonboiser, line 238: Have this return `null`, so the return type is `Maybe (breadcrumb)`, JS-style.

jonboiser:
This looks good. IT seems the only mandatory new rules from this upgrade are the "Newline after tags" rule and "required return statement in functions" rule.

For the "required return statement" rule, I think the conservative thing that was done was to return `undefined` to make the  function semantically equivalent, but we should probably take a look at some of these spaces and see if we can have a little more discipline on return types. For example, always return a number, or an optional type (e.g. number or null).

rtibbles:
> For the "required return statement" rule, I think the conservative thing that was done was to return undefined to make the function semantically equivalent, but we should probably take a look at some of these spaces and see if we can have a little more discipline on return types. For example, always return a number, or an optional type (e.g. number or null).

I agree with this - if we want to be disciplined, we could require a single type to return (in the case of a `Number` we should use `NaN` instead of `null` to enforce this). The edge case that comes to mind is if there are any instances where we might want to distinguish between `''` and a truly `undefined` value.

indirectlylit:
I think `null` is an appropriate placeholder for combining with any return type when you're trying to say "things are working fine, but here it doesn't exist", similar to `None` in Python.

Usually if I see `undefined` or `NaN` I assume something in the code has malfunctioned, and `''` is of course a valid string.

MisRob:
Yes, for now, I stayed with `undefined` in those properties where I wasn't (at least 99% ) sure that I won't break anything by using more specific type because, well, it's JavaScript and I've seen too much I guess :))

I agree that it would be nice to follow some standards when it comes to return types. Using `null` as expected none value for any type makes sense to me, too. I also see `undefined` or `NaN` rather as something unexpected.

I'm going to open an issue and reference this discussion there.

jonboiser (approved):

PR #5248, opened 2019-03-06
Use Vue observable for state that does not need to be vuex

rtibbles:
### Summary
Ugrades VueJS and related dependencies.
Changes modality state to use a Vue observable object.
Changes dynamic styling state to use a Vue observable object.

### Reviewer guidance
Does modality setting still work? (keyboard tabbing makes buttons outlined)
Do styles still work properly on dynamically styled components?

### References
Implements some suggestions made in https://github.com/learningequality/kolibri/issues/5057#issuecomment-468495847

----

### Contributor Checklist

PR process:

- [x] PR has the correct target branch and milestone
- [x] PR has 'needs review' or 'work-in-progress' label
- [x] If PR is ready for review, a reviewer has been added. (Don't use 'Assignees')
- [ ] If this is an important user-facing change, PR or related issue has a 'changelog' label
- [ ] If this includes an internal dependency change, a link to the diff is provided

Testing:

- [x] Contributor has fully tested the PR manually
- [ ] If there are any front-end changes, before/after screenshots are included
- [ ] Critical user journeys are covered by Gherkin stories
- [ ] Critical and brittle code paths are covered by unit tests

### Reviewer Checklist

- Automated test coverage is satisfactory
- PR is fully functional
- PR has been tested for [accessibility regressions](http://kolibri-dev.readthedocs.io/en/develop/manual_testing.html#accessibility-a11y-testing)
- External dependency files were updated if necessary (`yarn` and `pip`)
- Documentation is updated
- Contributor is in AUTHORS.rst

PR #5332, opened 2019-03-31
Proposal: Simple vuex validation setup with demo code

rtibbles:
### Summary
Allows for vuex state to have an optional schema that specifies the structure and validation of the state, to allow for warnings in dev mode, similarly to Vue prop component validation. Implemented as an optional additional key, for ease, but could be embellished to generate the default state also.

### Reviewer guidance
Does this seem like a useful enhancement?

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/core/assets/src/kolibri_app.js
+++ b/kolibri/core/assets/src/kolibri_app.js
@@ -1,9 +1,13 @@
 import forEach from 'lodash/forEach';
+import isPlainObject from 'lodash/isPlainObject';
 import router from 'kolibri.coreVue.router';
+import logger from 'kolibri.lib.logging';
 import Vue from 'kolibri.lib.vue';
 import store from 'kolibri.coreVue.vuex.store';
 import heartbeat from 'kolibri.heartbeat';
 import KolibriModule from 'kolibri_module';
+
+export const logging = logger.getLogger(__filename);
 
 /*
  * A class for single page apps that control routing and vuex state.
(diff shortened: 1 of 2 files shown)


jonboiser:
I think the API for this looks good, but probably has the same limitations of the `props` validators for components, esp. when validating complicated objects (i.e. if you want to check nested objects, you need to write a complicated validator function).

Maybe if we used an established validator library, we could define the module schema using whatever DSL they prefer (which will probably be richer than just saying `{type: Array}`, etc.)

rtibbles:
I do give some allowance for a nested object structure:
```
              // Otherwise assume it is just a nested object structure
              registerSchema(subSchema, moduleName, [...subPaths, propertyName]);
            }
```
This allows to specify validators at arbitrary depths of Objects, rather than defining a top level Object validator. I could extent this to Arrays quite simply to avoid the complex validator I used as an example.

rtibbles:
Also - any suggestions for a validator library, would definitely be simpler than rolling more from scratch!

indirectlylit:
I think our vuex state will be most maintainable if we generally be constrain it to be:

* literals
* arrays of literals
* if necessary, flat objects or ~arrays of flat objects~ maps from IDs to flat objects.

I'd be interested to see examples of situations where we'd need something more complex than that, especially with vuex modules at our disposal

jonboiser:
For maintainability, maybe less Vuex State, continuing the direction of #5248, esp. replacing the modules for very simple sub-pages with an observable or managing data fetching within the component.

jonboiser:
I'm still trying to think of situations where this validation will save me, though, or catch something that another part of the code (like the Vue runtime) would miss (or that could be avoided by certain coding conventions).

In general, the big `module/SET_STATE` mutations only have one call site in a handler function and are generally predictable. But when that isn't the case, we do do risky things like `SET_STATE(Object.assign(oldState, newMysteriousStuff))`.

At dev time, I feel like the smaller `SET_SOMETHING` mutations are generally safe too. If it a validator were to catch anything, it would maybe be at an integration level, where we see most bugs, because of some unexpected form of data that breaks our assumptions. Even then, I'm not sure I've seen a Sentry error related to crash in Vuex.

indirectlylit:
[relevant article](https://blog.usejournal.com/type-vue-without-typescript-b2b49210f0b) which argues that the most important thing to strive for is intellisense in the editor.

Aside: Regardless of vuex state, observables, or props, I think it's usually cleaner to use flatter objects. However I'll amend "arrays of flat objects" to "maps from IDs to flat objects", which turned out to be much more performant in coach tools...

MisRob:
> Also - any suggestions for a validator library, would definitely be simpler than rolling more from scratch!

https://github.com/jquense/yup looks quite simple and seems to have all the features needed. But I've never used it in a project so far, just heard that it's something like simpler Joi and a bit more suitable for browsers.

But anyway, I also think that there's no need to deal with (many levels) nested structures. So in terms of this issue, this draft seems good to me!

> I'm still trying to think of situations where this validation will save me, though, or catch something that another part of the code (like the Vue runtime) would miss (or that could be avoided by certain coding conventions).

I agree that Vue runtime, edge cases tests or coding conventions would catch most of the problems at the end of the day. On the other hand, sometimes it takes time to track the problem back to store. I would say it's similar to props validators - it's supposed to work as optional prevention. So I still consider it useful. But I guess it really is questionable and also depends on personal preferences.

---

Either way, most appreciated part of the schema is self-documentation for me! You have the idea of how state looks like in detail immediately, especially for arrays or objects. Example: `errors` - from default state it's clear that it's an array but it could be an array of error messages as strings or an array of error objects with a message field. If I want to use it and cannot remember what that thing is, I would need to find this information somewhere in the code. With schema it's obvious. Same applies to object state properties and their fields (what fields are in there, which are required, which are optional...)

rtibbles:
> You have the idea of how state looks like in detail immediately, especially for arrays or objects

I think this might be the main reason this could be useful, it gives a consistent way to document what the state *should* look like.

kolibri/core/assets/src/kolibri_app.js
```diff
@@ -68,6 +72,72 @@ export default class KolibriApp extends KolibriModule {
...
+                }
+              }
+              logging.debug(getter, callback);
+              store.watch(getter, callback);
+            } else {
+              // Otherwise assume it is just a nested object structure
+              registerSchema(subSchema, moduleName, [...subPaths, propertyName]);
+            }
+          }
+        });
```
indirectlylit, line 128: probably deserves its own module

kolibri/core/assets/src/kolibri_app.js
```diff
@@ -68,6 +72,72 @@ export default class KolibriApp extends KolibriModule {
...
+                  if (!subSchema.validator(newValue)) {
+                    fail = true;
+                  }
+                }
+                if (fail) {
+                  logging.error(
+                    `Validation failed for property: ${[...subPaths, propertyName]} in module ${
+                      moduleName ? moduleName : 'root'
+                    }`
+                  );
```
indirectlylit, line 118: similar feedback I make occasionally: recommend `return`ing early rather than "falling through" the conditionals and checking result at the end. For example:

```javascript
function callback(newValue) {
  const prop = [...subPaths, propertyName];
  const mod = moduleName ? moduleName : 'root';
  if (subSchema.type) {
    if (subSchema.type === Object && !isPlainObject(newValue)) {
      logging.error(
        `Validation failed for property: ${property} in module ${mod}: ${newValue} should be Object`
      );
      return;
    } else if (!(newValue instanceof subSchema.type)) {
      logging.error(
        `Validation failed for property: ${property} in module ${mod}: ${newValue} should be ${subSchema.type}`
      );
      return;
    }
  }
  if (subSchema.validator && !subSchema.validator(newValue)) {
    logging.error(
      `Validation failed for property: ${property} in module ${mod}: ${newValue} failed validation.`
    );
  }
}
```

two benefits:

1. makes explicit when processing is 'complete'
2. individually tailored messages for each case

indirectlylit (approved):
+1

rtibbles:
Merging for now, can iterate and clean up in the future.

PR #5333, opened 2019-04-01
automatic Black formatting

### LINKED ISSUE #3609: Add python auto-formatting tooling
In thread below, there is a desire to use something like

* autopep
* black
* yapf

To auto-format the python codebase.

Old issue:

<details>
Don't know if it's a coincidence, but I'm noticing the flake8 config in `setup.cfg` 
```
[flake8]
ignore = E226,E302,E41
max-line-length = 160
max-complexity = 10
exclude = kolibri/*/migrations/*,kolibri/dist/*
```

is really similar to the docs here:

http://pycodestyle.pycqa.org/en/latest/intro.html#error-codes

And `E41` does not seem to be a pycodestyle code either.

`E226` (requires whitespace around math operators) and `E302` (two newlines between top-level symbols in a module) seem like things that are done the majority of the time and all of the time for me personally. So maybe we should just remove the `ignore` settings
</details>

indirectlylit:
### Summary

Adds automatic Black formatting and checking to the devserver, pre-commit, and travis. Should work on Mac and Linux, regardless of Python version.

Also upgrades `pre-commit` to a version which runs checks in parallel by default.

fixes #3609

### Reviewer guidance

* Is the performance good when using the devserver?
* everything look correct and sustainable?

### References

leverages: https://github.com/learningequality/black-fmt

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated


package.json
```diff
@@ -15,19 +15,19 @@
...
     "clean": "kolibri-tools build clean --file ./build_tools/build_plugins.txt",
     "preinstall": "node ./packages/kolibri-tools/lib/npm_deprecation_warning.js",
-    "lint": "kolibri-tools lint",
-    "lint-all": "yarn run lint '{kolibri*/**/assets,packages}/**/*.{js,vue,scss,less,css}' -i '**/node_modules/**','**/static/**','**/packages/kolibri-core-for-export/**'",
-    "lint-all:fix": "yarn run lint-all -w",
-    "lint:fix": "yarn run lint -w",
-    "lint:watch": "yarn lint-all -w -m",
-    "generate-locale-data": "node ./build_tools/i18n/intl_code_gen.js",
+    "fmt-frontend": "yarn run fmt-frontend:check --write",
+    "fmt-frontend:check": "kolibri-tools lint '{kolibri*/**/assets,packages}/**/*.{js,vue,scss,less,css}' --ignore '**/node_modules/**','**/static/**','**/packages/kolibri-core-for-export/**'",
```
rtibbles: Need to update the tox.ini frontend linting job for this (this is why tests are failing currently).

rtibbles:
Not a huge amount to comment on here, but for the formatter command you are using:
```
// define the command
const command = command_array.join(" ");
function cmd() {
  console.log(command);
  child_process.exec(command, (error, stdout, stderr) => {
    // print the output
    console.log(stderr);
    if (!program.watch && error) {
      // exit with an error code for CI
      process.exit(1);
    }
  });
}

// do the work
console.log("Black formatter");
if (program.watch) {
  // start a watch process
  chokidar.watch(program.args).on("change", cmd);
} else {
  cmd();
}
```
This might be problematic - you are running black on `everything` anytime one watched file changes - you'd be better off reading the watched file name as an argument, and then passing that single file to the invocation of black, no?

package.json
```diff
@@ -15,19 +15,19 @@
...
     "bundle-stats": "kolibri-tools build stats --file ./build_tools/build_plugins.txt",
     "clean": "kolibri-tools build clean --file ./build_tools/build_plugins.txt",
     "preinstall": "node ./packages/kolibri-tools/lib/npm_deprecation_warning.js",
-    "lint": "kolibri-tools lint",
-    "lint-all": "yarn run lint '{kolibri*/**/assets,packages}/**/*.{js,vue,scss,less,css}' -i '**/node_modules/**','**/static/**','**/packages/kolibri-core-for-export/**'",
-    "lint-all:fix": "yarn run lint-all -w",
-    "lint:fix": "yarn run lint -w",
-    "lint:watch": "yarn lint-all -w -m",
-    "generate-locale-data": "node ./build_tools/i18n/intl_code_gen.js",
+    "fmt-frontend": "yarn run fmt-frontend:check --write",
```
rtibbles: Note that this command doesn't just format, it also runs linting and errors if linting has failed, so I am not sure this name change is appropriate for the frontend.

package.json
```diff
@@ -15,19 +15,19 @@
...
+    "fmt-frontend:check": "kolibri-tools lint '{kolibri*/**/assets,packages}/**/*.{js,vue,scss,less,css}' --ignore '**/node_modules/**','**/static/**','**/packages/kolibri-core-for-export/**'",
+    "fmt-frontend:watch": "yarn run fmt-frontend:check --write --monitor",
+    "fmt-backend": "yarn run black-fmt --exclude '/(\\.git|\\.tox|\\.venv|build|static|dist|migrations|node_modules)/' .",
```
rtibbles: Will be more efficient to pass an explicit include? Doesn't seem necessary to watch anything other than `*.py` files.

rtibbles:
In local testing, this is using 145% of my local CPU when running in watch mode, with no files changing.

By comparison, the frontend file watcher uses 0% of CPU.

indirectlylit:
yup, I believe it's going into an infinite self-watch loop. Was wondering if you'd run into anything like that in your work?

package.json
indirectlylit: I briefly tried explicit includes, but was concerned that over time we'd forget to keep adding them.

An aside, black's exclude logic caused me hours of grief because the docs claim you can pass multiple `--exclude` inputs but it doesn't work as expected... arg.

rtibbles: If you just use a general glob pattern, this should not be much of an issue. I also think that not using it may be contributing to your infinite watch loop woes.

package.json
indirectlylit: yes... on the other hand it doesn't just lint, it also formats :)

I went with `fmt` because on the backend there are other places that _also_ lint (e.g. flake8), and I wanted to distinguish from that, and because 'lint' seems a bit jargon-y. (It's not used on either the Prettier or Black project pages)

rtibbles: Right, on the backend, definitely, Black is only formatting, but that's not the case for the frontend.

Given that eslint and stylelint also autofix things I don't see that `linting` is a bad general term for the format + warn behaviour.

indirectlylit:
thanks for the feedback! This is very much a WIP - just posting it for initial eyes and thoughts

package.json
indirectlylit: `¯\_(ツ)_/¯` yeah my strategy was to keep the verbs the same for the sake of [simpler docs](https://github.com/learningequality/kolibri/pull/5333/files#diff-9bc8bf6e8d9db020cefce1fbc0426795R359) but if `fmt` feels misleading to you I'm happy to switch it back

lately I suppose I've been conceptualizing "linting" as a subset of "formatting", e.g. 

```bash
format --check
```

indirectlylit: Note that below, I changed the semantics of the command a bit for consistency:

`fmt-frontend": "yarn run fmt-frontend:check --write",
    "fmt-frontend:check    "fmt-frontend:watch":

indirectlylit: Note that below, I changed the semantics of the command a bit for consistency:

* `fmt-thing` formats and lints
* `fmt-thing:check` just lints
* `fmt-thing:watch` watches, formats, and lints

so I might revert that change too

package.json
indirectlylit: unfortunately `black` only takes regex, but I can certainly do this for `chokidar`

package.json
indirectlylit: reverted changes here: https://github.com/learningequality/kolibri/pull/5333/commits/0fb396a16ebd89d24bd5ef8cdcb6ea8caac774a7

indirectlylit:
ready for review

.pre-commit-config.yaml
```diff
@@ -19,7 +19,15 @@ repos:
     -   id: frontend-lint
         name: Linting of JS, Vue, SCSS and CSS files
         description: This hook handles all frontend linting for Kolibri
-        entry: yarn run lint -w
+        entry: yarn workspace kolibri-tools run lint --write
```
rtibbles: This is not a valid command. The only way to invoke this directly is `node ./packages/kolibri-tools/lib/cli.js lint --write`.

I think we may have discovered why I had a base lint command in the package.json - because that could be written as `kolibri-tools lint`.

rtibbles (dismissed):
This PR breaks our frontend linting precommit hook.

In manual testing, I ran the fmt-backend:watch command, CPU spiked to 260%, then eventually subsided to zero. Unfortunately, it happened because the black daemon exited with code 255. This wasn't immediately obvious because the node wrapper itself didn't also exit.

.pre-commit-config.yaml
indirectlylit: This is the same pattern we're currently using for the `build-kolibri-tools` command: https://github.com/learningequality/kolibri/blob/develop/package.json#L31

rtibbles: Yes, because `build` is a valid command inside the `kolibri-tools` package.json - note that this is for building the distributable version of `kolibri-tools` not for invoking the `kolibri-tools` `build` subcommand.

indirectlylit: ah, I see. Maybe we could call the bin alias `kolibri-tools-cli` to disambiguate?

> The only way to invoke this directly is `node ./packages/kolibri-tools/lib/cli.js lint --write`

I believe `yarn run kolibri-tools lint --write` should work too since `kolibri-tools` is getting registered as a runnable binary?

rtibbles: No, I don't believe so.

`yarn/npm run ...` is only used to invoke registered scripts in a package.json, whereas if `kolibri-tools` is registered as a runnable libary it is invoked by `kolibri-tools` (although command line invocation mileage varies - but it works consistently when invoked inside a package.json command).

rtibbles (approved):
Code looks right, performance seems reasonable. Still got a spike up to 145% CPU usage on first start up (after the webpack builds had finished), so I am worried that chokidar might still be watching too many files, but it did subside about 10-15 seconds after the initial webpack build process has finished.

jredrejo:
> This might be problematic - you are running black on `everything` anytime one watched file changes - you'd be better off reading the watched file name as an argument, and then passing that single file to the invocation of black, no?

This is having really bad consequences when having virtual environment directories (in my case I have one for python 2.7 and others for different versions of python 3.x). Cpu goes to the top and commit times out while black-fmt processes stay in the background forever.... I think running it on everything is not correct.

indirectlylit:
this shouldn't have been happening any more - will investigate

Issue #4992, opened 2019-02-09
Create linting rules to check for unused props, data, computed props, and vuex mappers


rtibbles:
### Observed behavior
Currently we have no way of auditing our components to see whether any of our component data is used inside of the component itself. This means that our components slowly accrue cruft over time.

### Expected behavior
Our linting should give an error when any of a prop, data, computed props, vuex mapped getters, and state are unused.

In addition, it could also warn for methods, and mapped actions that are unused anywhere in the component.

MisRob:
I'll try to prepare a first simple version of one or two rules and after we see how it works, I think that we could also possibly ask _eslint-plugin-vue_ folks if they would be interested in using code and implementing those rules right within _eslint-plugin-vue_ ? All of them are really general rules that might be useful in many projects and for us, getting some wider support from Vue community would be also nice I think.

And it seems they already have some of them in their todo list :)

https://github.com/vuejs/eslint-plugin-vue/issues/848
https://github.com/vuejs/eslint-plugin-vue/issues/631

rtibbles:
Yeah, I made this issue because I was surprised that they didn't already exist, anything we can do to have them maintained externally is fine by me!

rtibbles:
Although, I note that neither of these are vuex specific (due to the way that the vuex mappers work, this would probably need to be implemented separately to the no unused vars/props etc).

MisRob:
No problem, I'm actually planning to split it anyway for the first version and use utility functions for some common functionality. I would rather start small. It can be merged into one rule if needed later - this will be easier when compared to breaking one rule into more of them. So we should be able to provide some of them to _eslint-plugin-vue_ and keep the rest of them as our custom rules and eventually merge them if that would be more appropriate for us.

rtibbles:
I think the general tendency in eslint rules (which is continued in eslint-plugin-vue) is to make rules very small, so that they can be turned on and off as desired, so I think it makes perfect sense to separate out vuex specific rules from regular vue ones.

MisRob:
I've started working on it.

PR #5347, opened 2019-04-06
Add eslint rule to find unused properties, data and computed properties

MisRob:
### Summary

1. Implement eslint rule that reports unused properties, data, and computed properties
2. Turn it on and deal with it
3. The last commit contains fix for the following issue: One of the reported errors showed that there was a development mock label in a group exercise lesson report link leading back to a parent lesson page (Coach -> Reports -> Groups –> select a group lesson <lesson_title> -> select an exercise belonging to this lesson and watch backlink label (should be 'Back to "<lesson_title>"' but was 'Back to “Lesson 1"')

**Before**
before

**After**
after

### Reviewer guidance
- Break it :)! Any ideas about possibly untreated situations or edge cases?
- I tried to be careful and check that there is no hidden usage for all properties removed in 
055bfc3 , though I would appreciate if someone had a look
- In 582dcab I removed constants from properties. This could also be solved be leaving them in properties but referencing them using `this`. I guess there's no need to save them to an instance and it's also a bit more visible that they are some common definitions and constants, but I don't have any preference so if someone prefers the second solution, no problem.
- In 9ca2dae I commented out all unused properties related to to commented out template TODO code. I could remove everything but I suppose that it is under development? If it's not the case, I'll remove everything including those template parts.

### References

Partially solves  #4992.

----

As discussed in #4992, as soon as it's ready, I would like to open the rule PR directly in eslint-plugin-vue repo where there's already an issue for this rule opened and proposal accepted.

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/plugins/coach/assets/src/views/reports/ReportsGroupHeader.vue
+++ b/kolibri/plugins/coach/assets/src/views/reports/ReportsGroupHeader.vue
@@ -79,6 +79,7 @@
       group() {
         return this.groupMap[this.$route.params.groupId];
       },
+      /** TODO COACH
       recipients() {
         return this.group.member_ids;
       },
@@ -92,6 +93,7 @@
         }
         return this._.meanBy(statuses, 'score');
       },
+      */
     },
     $trs: {
       back: 'All groups',
--- a/kolibri/plugins/coach/assets/src/views/reports/ReportsGroupReportLessonPage.vue
+++ b/kolibri/plugins/coach/assets/src/views/reports/ReportsGroupReportLessonPage.vue
@@ -93,6 +93,7 @@
     components: {},
     mixins: [commonCoach],
     computed: {
+      /** TODO COACH
       actionOptions() {
         return [
           { label: this.coachStrings.$tr('editDetailsAction'), value: 'ReportsLessonEditorPage' },
@@ -102,6 +103,7 @@
           },
         ];
       },
+      */
       lesson() {
         return this.lessonMap[this.$route.params.lessonId];
       },
--- a/packages/eslint-plugin-kolibri/lib/rules/vue-no-unused-properties.js
+++ b/packages/eslint-plugin-kolibri/lib/rules/vue-no-unused-properties.js
@@ -7,6 +7,17 @@
 const remove = require('lodash/remove');
 const utils = require('eslint-plugin-vue/lib/utils');
 
+const GROUP_PROPERTY = 'props';
+const GROUP_DATA = 'data';
+const GROUP_COMPUTED_PROPERTY = 'computed';
+const GROUP_WATCHER = 'watch';
+
+const PROPERTY_LABEL = {
+  [GROUP_PROPERTY]: 'property',
+  [GROUP_DATA]: 'data',
+  [GROUP_COMPUTED_PROPERTY]: 'computed property',
+};
+
 const getReferencesNames = references => {
   if (!references || !references.length) {
     return [];
@@ -27,16 +38,9 @@ const reportUnusedProperties = (context, properties) => {
   }
 
   properties.forEach(property => {
-    let kind = 'property';
-    if (property.groupName === 'data') {
-      kind = 'data';
-    } else if (property.groupName === 'computed') {
-      kind = 'computed property';
-    }
-
     context.report({
       node: property.node,
-      message: `Unused ${kind} found: "${property.name}"`,
+      message: `Unused ${PROPERTY_LABEL[property.groupName]} found: "${property.name}"`,
     });
   });
 };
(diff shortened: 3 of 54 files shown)


kolibri/plugins/coach/assets/src/views/reports/ReportsGroupHeader.vue
```diff
@@ -79,6 +79,7 @@
       group() {
         return this.groupMap[this.$route.params.groupId];
       },
+      /** TODO COACH
```
rtibbles, line 82: This looks unfinished?

kolibri/plugins/coach/assets/src/views/reports/ReportsGroupReportLessonPage.vue
```diff
@@ -93,6 +93,7 @@
     components: {},
     mixins: [commonCoach],
     computed: {
+      /** TODO COACH
```
rtibbles, line 96: Ditto.

packages/eslint-plugin-kolibri/lib/rules/vue-no-unused-properties.js
```diff
@@ -0,0 +1,138 @@
...
+  });
+};
+
+const reportUnusedProperties = (context, properties) => {
+  if (!properties || !properties.length) {
+    return;
+  }
+
+  properties.forEach(property => {
+    let kind = 'property';
```
rtibbles: Could maybe make each of these kinds a constant at the top of the file, and then refer to them?
```
const PROPERTY = 'property';
const DATA = 'data';
const COMPUTED_PROPERTY = 'computed property';
```
(this is following our convention of using SCREAMING_SNAKE_CASE constants in the frontend, instead of repeating references to strings).

rtibbles (commented):
This is excellent. My read through of the code changes and the linting code make me think this is definitely close to mergeable, but as you suggest, I will also try to break it myself!

indirectlylit:
awesome!

> I tried to be careful and check that there is no hidden usage for all properties

One thing to check for is that many of the users of these components will still be passing in the props. For example, we deleted `id` from `TaskProgress` but that prop is [still being passed in](https://github.com/learningequality/kolibri/blob/develop/kolibri/plugins/device_management/assets/src/views/SelectContentPage/index.vue#L12).

> I removed constants

We had a [related conversation](https://github.com/learningequality/kolibri/pull/5006#discussion_r256683153) a little while go. The general pattern is that some `thing` in `<script>` needs to be accessed in `<template>`, and then we've tended to do one of:

* put `this.thing = thing;` or `this.$options.thing = thing;` in `mounted()` or somewhere similar
* put `thing` in `data()`
* put `thing() { return thing; }` in computed props

There are lots of these scattered around the code base. I don't think the inconsistency has caused us any particular issues, but might be nice to standardize on a strategy.

> I commented out all unused properties related to to commented out template TODO code

I think some of these relate to some 0.12.y coach tools work @jonboiser is working on. @jonboiser would you mind checking if any of these are actually dead code that should be removed?

packages/eslint-plugin-kolibri/lib/rules/vue-no-unused-properties.js
MisRob: Yes, this will be better, thanks. Done!

MisRob:
Thanks for info Devon!

> One thing to check for is that many of the users of these components will still be passing in the props. 

Right, I didn't realize I should also clean this, thanks. It's done in the last four commits.


> We had a related conversation a little while go. The general pattern is that some thing in `<script>` needs to be accessed in `<template>`, and then we've tended to do one of...


OK. It was used only in the script part here but I guess it will be better to keep it consistent and prepared for usage in templates so I moved those constants I removed before back to an instance and referenced them using `this` (2934966).

indirectlylit:
> I guess it will be better to keep it consistent and prepared for usage in templates

all right, that's cool. The way you had it before also seemed fine though

MisRob:
No problem, I don't have any strong opinion on this particular issue and if this is a general tendency ;)

indirectlylit (approved):
I'm gonna go ahead and merge this to prevent merge conflicts.

@jonboiser when you get back, could you follow up on the 'TODO' items and flag if there are any that won't be addressed by your work on coach tools?

jtwaleson:
This is great!! Are there any plans to add this to eslint-plugin-vue? Could an update be pushed to NPM so we can use these rules via eslint-plugin-kolibri?

MisRob:
Hello, we opened a PR in `eslint-plugin-vue` repository one month ago, you can find it here https://github.com/vuejs/eslint-plugin-vue/pull/871 with some more info regarding updates and other, similar, rules.

Issue #5360, opened 2019-04-13
Auth message is not displayed on classes page for unauthenticated users


MisRob:
### Observed behavior

When logged out user tries to access a learner's classes page, an unexpected error is displayed

<img width="709" alt="Screenshot 2019-04-13 at 19 56 57" src="https://user-images.githubusercontent.com/13509191/56083512-4facbe80-5e26-11e9-8a31-9f1c843e7c9b.png">

### Expected behavior

"Did you forget to sign in? You must be signed in as a learner to view this page " shows up.

### User-facing consequences

Just slight confusion (when user clicks "Back to home" as recommended, he will be redirected to Channels page which can be accessed by unauthorized users)

### Steps to reproduce

1. Make sure you're logged out
2. Visit `/learn/#/classes`

### Context

0.12.3a0.dev0+git.162.gc3af46e5

MisRob:
First investigation showed:

When logged out user visits `learn/#/classes`, following happens
- API returns `403`
- When `CoreBase` is mounted within `LearnIndex`, `authorized` prop is not passed in (I suppose that’s because we allow logged out users to access Channels and Recommended sections maybe)
- `authorized` is set to `true` by default in `CoreBase` which causes `notAuthorized` computed prop to return `false`
- =>`<AuthMessage>` is not displayed, and because there’s an `error`, `<AppError />` fallback is triggered

There’s a [code prepared for handling logged out users](https://github.com/learningequality/kolibri/blob/develop/kolibri/plugins/learn/assets/src/views/classes/AllClassesPage.vue#L22) but at the first sight it seems that it cannot be reached.

I suppose that this might be solved by passing this information into `CoreBase` using `authorized` param which would be calculated in `LearnIndex` according to whether the user is logged in or no and which page is accessed.

MisRob:
Or, based on the fact that there are pages that need to control this behavior, I was also thinking if it would make sense to pull out `AuthMessage` and `AppError` from `CoreBase` and handle it rather locally which would make these things a bit more flexible.

jonboiser:
I'm adding this case to issue #5305. In summary, the reason we're seeing this type of error is because the `showAllClassesPage` function handles errors using the catch-all `handleApiError` Vuex action. One solution would be to replace that `handleApiError` in the catch branch with some other reasonable behavior (like redirecting to a different part of the "Learn" page)

[...]

Issue #5309, opened 2019-03-23
Consistent style for auth messages


MisRob:
### Observed behavior

Differently styled auth messages:

<img width="1440" alt="Screenshot 2019-03-23 at 15 33 39" src="https://user-images.githubusercontent.com/13509191/54867913-72e3d100-4d86-11e9-9464-e2aba75ce532.png">

<img width="1439" alt="Screenshot 2019-03-23 at 15 33 26" src="https://user-images.githubusercontent.com/13509191/54867912-70817700-4d86-11e9-9144-3739d6649f9c.png">

(seems to have something in common with `KPageContainer` presence)

### Expected behavior

Coach auth message should be wrapped in a card too.
…

### User-facing consequences
Visual inconsistency

indirectlylit:
agreed!

MisRob:
🆗 thanks for checking it. I'll have a look at it then.

indirectlylit:
Wrapping with `KPageContainer` will be the simplest fix

rtibbles:
Fixed in #5361

PR #5361, opened 2019-04-13
Consistent style for auth messages

MisRob:
### Summary

Wrap all auth messages in a container component to achieve visual consistency - some messages were wrapped and some weren't before - e.g. Device pages:

device

I noticed that Device pages were the only exception actually (and it could be just a side effect of routing pages inside of `KPageContainer`) - so another solution would be to remove the wrapper there but I think I still like the version with a wrapper a bit more, it makes the text to stand out a bit better.

#### Changes screenshots

| | Before  | After |
| ------------- | ------------- | ------------- |
| Facility | facility_before | facility_after |
| Coach | coach_before | coach_after |

I also found another auth message to be wrapped - the one to be displayed when an unauthenticated user tries to visit Learn classes. But it seems to never be displayed right now - reported as #5360.

### Reviewer guidance

- visit `/coach` and `/facility` as logged out user

### References

Solves #5309

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/core/assets/src/views/CoreBase/index.vue
+++ b/kolibri/core/assets/src/views/CoreBase/index.vue
@@ -77,7 +77,10 @@
           :details="authorizationErrorDetails"
         />
       </KPageContainer>
-      <AppError v-else-if="error" />
+      <KPageContainer v-else-if="error">
+        <AppError />
+      </KPageContainer>
+
       <slot v-else></slot>
     </div>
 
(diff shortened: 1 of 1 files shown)


MisRob:
Also, _if_ we're going to wrap all auth messages, then it might be a good idea to also wrap an unexpected error, right?

Screenshot 2019-04-13 at 20 55 03

indirectlylit:
> Also, if we're going to wrap all auth messages, then it might be a good idea to also wrap an unexpected error, right?

yes, please!

MisRob:
OK, I've just added it to this PR:

Screenshot 2019-04-14 at 12 18 15

rtibbles (approved):

PR #5374, opened 2019-04-16
Add unused methods eslint rule

MisRob:
### Summary

- Implement eslint rule that reports unused Vue component methods
- Turn it on + related cleanup

When dealing with public methods, I noticed that there was one occurrence of jsdoc's `@public` tag from the past. I really liked that idea so I used it for documenting all public methods. Then I noticed that in case there's `@public` tag available, `// eslint-disabled-next-line` is somewhat duplicate information. So I experimented with `@public` comment detection in the rule which is now able to ignore methods marked with it. This also gave me enough confidence to turn the rule in error mode instead of warnings only. Does this sound useful?

### Reviewer guidance

As usual :)

- Works fine?
- Are there more scenarios to be treated?
- Is cleanup ok?

### References

Solves third part of #4992.

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/plugins/user/assets/src/views/SignInPage/index.vue
+++ b/kolibri/plugins/user/assets/src/views/SignInPage/index.vue
@@ -286,7 +286,9 @@
       },
     },
     watch: {
-      username: 'setSuggestionTerm',
+      username() {
+        this.setSuggestionTerm();
+      },
     },
     mounted() {
       /*
--- a/packages/eslint-plugin-kolibri/lib/rules/vue-no-unused-methods.js
+++ b/packages/eslint-plugin-kolibri/lib/rules/vue-no-unused-methods.js
@@ -8,25 +8,9 @@ const remove = require('lodash/remove');
 const eslintPluginVueUtils = require('eslint-plugin-vue/lib/utils');
 
 const utils = require('../utils');
+const constants = require('../constants');
 
-const GROUP_METHODS = 'methods';
-
-const PROPERTY_LABEL = {
-  [GROUP_METHODS]: 'method',
-};
-
-const reportUnusedProperties = (context, properties) => {
-  if (!properties || !properties.length) {
-    return;
-  }
-
-  properties.forEach(property => {
-    context.report({
-      node: property.node,
-      message: `Unused ${PROPERTY_LABEL[property.groupName]} found: "${property.name}"`,
-    });
-  });
-};
+const { GROUP_METHODS } = constants;
 
 const create = context => {
   let hasTemplate;
@@ -58,7 +42,7 @@ const create = context => {
       });
 
       if (!hasTemplate && unusedProperties.length) {
-        reportUnusedProperties(context, unusedProperties);
+        utils.reportUnusedProperties(context, unusedProperties);
       }
     })
   );
@@ -76,7 +60,7 @@ const create = context => {
     },
     utils.executeOnRootTemplateEnd(() => {
       if (unusedProperties.length) {
-        reportUnusedProperties(context, unusedProperties);
+        utils.reportUnusedProperties(context, unusedProperties);
       }
     })
   );
--- a/packages/eslint-plugin-kolibri/lib/rules/vue-no-unused-methods.js
+++ b/packages/eslint-plugin-kolibri/lib/rules/vue-no-unused-methods.js
@@ -1,0 +1,99 @@
+/**
+ * @fileoverview Disallow unused methods.
+ */
+
+'use strict';
+
+const remove = require('lodash/remove');
+const eslintPluginVueUtils = require('eslint-plugin-vue/lib/utils');
+
+const utils = require('../utils');
+
+const GROUP_METHODS = 'methods';
+
+const PROPERTY_LABEL = {
+  [GROUP_METHODS]: 'method',
+};
+
+const reportUnusedProperties = (context, properties) => {
+  if (!properties || !properties.length) {
+    return;
+  }
+
+  properties.forEach(property => {
+    context.report({
+      node: property.node,
+      message: `Unused ${PROPERTY_LABEL[property.groupName]} found: "${property.name}"`,
+    });
+  });
+};
+
+const create = context => {
+  let hasTemplate;
+  let unusedProperties = [];
+  let thisExpressionsVariablesNames = [];
+
+  const initialize = {
+    Program(node) {
+      if (!utils.checkVueEslintParser(context)) {
+        return;
+      }
+
+      hasTemplate = Boolean(node.templateBody);
+    },
+  };
+
+  const scriptVisitor = Object.assign(
+    {},
+    utils.executeOnThisExpressionProperty(property => {
+      thisExpressionsVariablesNames.push(property.name);
+    }),
+    eslintPluginVueUtils.executeOnVue(context, obj => {
+      unusedProperties = Array.from(
+        eslintPluginVueUtils.iterateProperties(obj, new Set([GROUP_METHODS]))
+      );
+
+      remove(unusedProperties, property => {
+        return thisExpressionsVariablesNames.includes(property.name);
+      });
+
+      if (!hasTemplate && unusedProperties.length) {
+        reportUnusedProperties(context, unusedProperties);
+      }
+    })
+  );
+
+  const templateVisitor = Object.assign(
+    {},
+    {
+      'VExpressionContainer[expression!=null][references]'(node) {
+        const referencesNames = utils.getReferencesNames(node.references);
+
+        remove(unusedProperties, property => {
+          return referencesNames.includes(property.name);
+        });
+      },
+    },
+    utils.executeOnRootTemplateEnd(() => {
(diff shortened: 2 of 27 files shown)


rtibbles:
> When dealing with public methods, I noticed that there was one occurrence of jsdoc's @public tag from the past. I really liked that idea so I used it for documenting all public methods. Then I noticed that in case there's @public tag available, // eslint-disabled-next-line is somewhat duplicate information. So I experimented with @public comment detection in the rule which is now able to ignore methods marked with it. This also gave me enough confidence to turn the rule in error mode instead of warnings only. Does this sound useful?

Yes.

I think having our linting be responsive to documentation comments rather than linting exclusion comments is much preferred.

packages/eslint-plugin-kolibri/lib/utils.js
```diff
@@ -87,4 +99,30 @@ module.exports = {
...
+    }
+
+    properties.forEach(property => {
+      if (disabledLines && disabledLines.includes(property.node.loc.start.line)) {
+        return;
+      }
+
+      let message = `Unused ${PROPERTY_LABEL[property.groupName]} found: "${property.name}"`;
+      if (property.groupName === GROUP_METHODS) {
+        message = `${message}. If the method is supposed to be public, you might forget to add @public tag.`;
```
rtibbles:
```suggestion
        message = `${message}. If the method is supposed to be public, you might have forgotten to add a @public tag.`;
```

packages/eslint-plugin-kolibri/lib/rules/vue-no-unused-methods.js
```diff
@@ -0,0 +1,89 @@
...
+
+'use strict';
+
+const remove = require('lodash/remove');
+const eslintPluginVueUtils = require('eslint-plugin-vue/lib/utils');
+
+const utils = require('../utils');
+const constants = require('../constants');
+
+const { GROUP_METHODS } = constants;
```
rtibbles: Not necessary to change, but if `GROUP_METHODS` is the only thing you are using from the `constants` variable, then you can write these two lines as : `const { GROUP_METHODS } = require('../constants');` to do the destructuring and assignment in one line.

kolibri/plugins/user/assets/src/views/SignInPage/index.vue
```diff
@@ -286,7 +286,9 @@
       },
     },
     watch: {
-      username: 'setSuggestionTerm',
+      username() {
```
rtibbles, line 289: Hrm, rather than forcing this syntax, oughtn't we just add an additional check as to whether the method is used as a watch callback by string reference?

rtibbles (changes requested):
This looks excellent overall. The one thing I would like added is catching the reference to a method as a string in the watch property, as that is valid vue syntax, this seems like an appropriate thing to allow, and is kind of the inverse case of the vuex state/getter/action linting.

kolibri/plugins/user/assets/src/views/SignInPage/index.vue
indirectlylit: I don't particularly mind forcing the more verbose function syntax. Referencing functions by string value seems unnecessarily magical and [isn't covered in the Vue intro guide](https://vuejs.org/v2/guide/computed.html).

Opaque without knowledge of Vue:

```javascript
    watch: {
      username: 'setSuggestionTerm',
    }
```

Very explicit and readable:

```javascript
    watch: {
      username() {
        this.setSuggestionTerm();
      },
    }
```

Concise and explicit, but maybe less readable:

```javascript
    watch: {
      // does this actually work?
      username: () => this.setSuggestionTerm(),
    }
```

packages/eslint-plugin-kolibri/lib/rules/vue-no-unused-methods.js
MisRob: Right, thanks, will update it tomorrow.

kolibri/plugins/user/assets/src/views/SignInPage/index.vue
MisRob: Yep, as described in the commit message, I didn't find any reference (not even here https://vuejs.org/v2/api/#vm-watch, have I missed something?) and that's why I decided to update the code instead of adding support to the rule.

packages/eslint-plugin-kolibri/lib/utils.js
MisRob: Thanks!

kolibri/plugins/user/assets/src/views/SignInPage/index.vue
rtibbles: It's under the component Options section: https://vuejs.org/v2/api/#watch (not the watch method itself), second item in the example.

MisRob: Ah, sorry, when I posted the link I noticed I'm in the wrong place https://vuejs.org/v2/api/#watch.

MisRob: And I obviously checked the wrong one before. No problem to add the support for this then.

rtibbles: And from that same reference @indirectlylit:
```
Note that you should not use an arrow function to define a watcher
(e.g. searchQuery: newValue => this.updateAutocomplete(newValue)).
The reason is arrow functions bind the parent context, so this will not
be the Vue instance as you expect and this.updateAutocomplete
will be undefined.
```

rtibbles: Also, if the method is reading the newValue (and any of the other values passed into the watcher function), then the object method form becomes even more verbose:
```
    watch: {
      username(newVal, oldVal) {
        this.setSuggestionTerm(newVal, oldVal);
      },
    }
```

MisRob: Yeah it is indeed and I like it because you immediately know what's happening there, but I am quite into functions so it's just a personal preference. No problem to support more syntaxes.

indirectlylit: A few other considerations:

* `watch` is a code smell: new Vue devs will often use it when `computed` would have been more appropriate
* In Kolibri we use `this.$watch(...)` 5x and `watch: { ... }` 32x.
* We use the string syntactic sugar in just 4 components, and one of them actually [looks like a bug](https://github.com/learningequality/kolibri/blob/develop/kolibri/plugins/learn/assets/src/views/TotalPoints.vue#L40) that would have been caught if it were written using the full function syntax.

I vote to ban :)

MisRob: > We use the string syntactic sugar in just 4 components, and one of them actually looks like a bug that would have been caught if it were written using the full function syntax.

After checking that reference in our codebase, this actually sounds like a pretty good reason to me.

MisRob: If we ban it, the question is - should this particular rule be responsible for this? It might be confusing for folks who know how to use the string syntax but don't know about this decision to not support it in our codebase. Because in this situation, they would be basically using valid Vue syntax but the report would say "Unused method found...". I was thinking about adding more help text but it's long enough and anyway, this sounds like something to be forced with another rule?

rtibbles: I think banning should be another rule, yes, so making this rule work for the banned case seems reasonable, then adding a very lightweight 'vue-watch-no-string' rule seems sensible.

rtibbles: I am not certain that the bug would have been caught without this as the bug arose because of the mapped action being removed from the component definition and that not being noticed in review. 

I think the main reason it is extant is that it will only ever get triggered right as someone is logged out, which will almost always trigger a redirect.

That said, banning is fine.

rtibbles (approved):

kolibri/plugins/user/assets/src/views/SignInPage/index.vue
rtibbles: Could you file a follow up issue for this banning?

MisRob: #5380

indirectlylit: thanks guys.

The reason I assumed it would have been caught was that I think this would have resulted in an 'undefined function `fetchPoints`' error?

```javascript
watch: {
  currentUserId() { this.fetchPoints(); },
},
```

rtibbles: Yes - but Vue would have thrown a similar error when the named watcher function didn't exist - so the reason the bug didn't get caught was because the watcher was only being invoked during a page redirect.

PR #5389, opened 2019-04-18
Adds support for a redis-cache option.

rtibbles:
### Summary
Adds an option for using django-redis-cache.

### Reviewer guidance
Does it work? If you are running redis with default settings, adding the follow lines to your options.ini should activate it:
```
[Cache]
CACHE_BACKEND = redis
CACHE_LOCATION = localhost:6379
```
`django-redis-cache` dropped support for Python3.4 in 2.0.0 - so this addition throws a runtime error if it detects it being used in Python3.4.

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/deployment/default/cache.py
+++ b/kolibri/deployment/default/cache.py
@@ -1,0 +1,47 @@
+import sys
+
+from kolibri.utils.conf import OPTIONS
+
+cache_options = OPTIONS["Cache"]
+
+# Default to LocMemCache, as it has the simplest configuration
+default_cache = {
+    "BACKEND": "django.core.cache.backends.locmem.LocMemCache",
+    # Default time out of each cache key
+    "TIMEOUT": cache_options["CACHE_TIMEOUT"],
+}
+
+built_files_prefix = "built_files"
+
+built_files_cache = {
+    "BACKEND": "django.core.cache.backends.locmem.LocMemCache",
+    # Default time out of each cache key
+    "TIMEOUT": cache_options["CACHE_TIMEOUT"],
+}
+
+
+if cache_options["CACHE_BACKEND"] == "redis":
+    if sys.version_info.major == 3 and sys.version_info.minor < 5:
+        raise RuntimeError(
+            "Attempted to use redis cache backend with Python 3.4, please use Python 2.7 or 3.5+"
+        )
+    base_cache = {
+        "BACKEND": "redis_cache.RedisCache",
+        "LOCATION": cache_options["CACHE_LOCATION"],
+        # Default time out of each cache key
+        "TIMEOUT": cache_options["CACHE_TIMEOUT"],
+        "OPTIONS": {"PASSWORD": cache_options["CACHE_PASSWORD"]},
+    }
+    default_cache = dict(base_cache)
+    default_cache["OPTIONS"]["DB"] = 0
+    built_files_cache = dict(base_cache)
+    built_files_cache["OPTIONS"]["DB"] = 1
+
+built_files_cache["KEY_PREFIX"] = built_files_prefix
+
+CACHES = {
+    # Default cache
+    "default": default_cache,
+    # Cache for builtfiles - frontend assets that only change on upgrade.
+    "built_files": built_files_cache,
+}
--- a/kolibri/deployment/default/cache.py
+++ b/kolibri/deployment/default/cache.py
@@ -34,9 +34,9 @@ if cache_options["CACHE_BACKEND"] == "redis":
         "OPTIONS": {"PASSWORD": cache_options["CACHE_PASSWORD"]},
     }
     default_cache = copy.deepcopy(base_cache)
-    default_cache["OPTIONS"]["DB"] = 0
+    default_cache["OPTIONS"]["DB"] = cache_options["CACHE_REDIS_MIN_DB"]
     built_files_cache = copy.deepcopy(base_cache)
-    built_files_cache["OPTIONS"]["DB"] = 1
+    built_files_cache["OPTIONS"]["DB"] = cache_options["CACHE_REDIS_MIN_DB"] + 1
 
 built_files_cache["KEY_PREFIX"] = built_files_prefix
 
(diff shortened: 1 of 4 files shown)


kolibri/deployment/default/cache.py
```diff
@@ -0,0 +1,47 @@
...
+    "BACKEND": "django.core.cache.backends.locmem.LocMemCache",
+    # Default time out of each cache key
+    "TIMEOUT": cache_options["CACHE_TIMEOUT"],
+}
+
+
+if cache_options["CACHE_BACKEND"] == "redis":
+    if sys.version_info.major == 3 and sys.version_info.minor < 5:
+        raise RuntimeError(
+            "Attempted to use redis cache backend with Python 3.4, please use Python 2.7 or 3.5+"
```
aronasorman, line 27: What's the reason for this check?

kolibri/deployment/default/cache.py
```diff
@@ -0,0 +1,47 @@
...
+        )
+    base_cache = {
+        "BACKEND": "redis_cache.RedisCache",
+        "LOCATION": cache_options["CACHE_LOCATION"],
+        # Default time out of each cache key
+        "TIMEOUT": cache_options["CACHE_TIMEOUT"],
+        "OPTIONS": {"PASSWORD": cache_options["CACHE_PASSWORD"]},
+    }
+    default_cache = dict(base_cache)
```
aronasorman: `base_cache` already seems like a dict! What's the reason for a casting it again?

kolibri/utils/options.py
```diff
@@ -45,6 +45,29 @@ def calculate_thread_pool():
...
+            "options": ("memory", "redis"),
+            "default": "memory",
+            "envvars": ("KOLIBRI_CACHE_BACKEND",),
+        },
+        "CACHE_TIMEOUT": {
+            "type": "integer",
+            "default": 300,
+            "envvars": ("KOLIBRI_CACHE_TIMEOUT",),
+        },
+        "CACHE_PASSWORD": {
```
aronasorman, line 60: Should we add an error if these aren't defined, and we're using redis?

aronasorman (commented):
Some Q's!

kolibri/deployment/default/cache.py
rtibbles: To make a copy, so that I can edit it - although, I am now realizing that because it has a subdict, I need to do a deep copy instead. Good catch!

kolibri/utils/options.py
rtibbles: Maybe not for password, as it's not required, but either we do that for location, or as @jamalex suggested, we should have a default location for `localhost:6379` which is the redis default.

aronasorman:
Sounds good! I agree with the localhost default.

Aron
On Thu, Apr 18, 2019 at 17:18, Richard Tibbles <notifications@github.com> wrote:
@rtibbles commented on this pull request.




In kolibri/utils/options.py [https://github.com/learningequality/kolibri/pull/5389#discussion_r276873101] :

> @@ -45,6 +45,29 @@ def calculate_thread_pool():
 
 
 option_spec = {
+    "Cache": {
+        "CACHE_BACKEND": {
+            "type": "option",
+            "options": ("memory", "redis"),
+            "default": "memory",
+            "envvars": ("KOLIBRI_CACHE_BACKEND",),
+        },
+        "CACHE_TIMEOUT": {
+            "type": "integer",
+            "default": 300,
+            "envvars": ("KOLIBRI_CACHE_TIMEOUT",),
+        },
+        "CACHE_PASSWORD": {


Maybe not for password, as it's not required, but either we do that for location, or as @jamalex [https://github.com/jamalex] suggested, we should have a default location for localhost:6379 which is the redis default.

kolibri/deployment/default/cache.py
```diff
@@ -0,0 +1,48 @@
...
+        "BACKEND": "redis_cache.RedisCache",
+        "LOCATION": cache_options["CACHE_LOCATION"],
+        # Default time out of each cache key
+        "TIMEOUT": cache_options["CACHE_TIMEOUT"],
+        "OPTIONS": {"PASSWORD": cache_options["CACHE_PASSWORD"]},
+    }
+    default_cache = copy.deepcopy(base_cache)
+    default_cache["OPTIONS"]["DB"] = 0
+    built_files_cache = copy.deepcopy(base_cache)
+    built_files_cache["OPTIONS"]["DB"] = 1
```
jredrejo: why are we hardcoding the redis databases to be 0 & 1 and not having them as optional?
Thinking of a server where redis is already in use this might taint it. As a sysadmin I would like to be able to use different dbs for different applications.?

requirements/base.txt
```diff
@@ -30,3 +30,5 @@ ifcfg==0.17
 sqlalchemy==1.2.10
 semver==2.8.1
 sentry-sdk==0.7.9
+django-redis-cache==2.0.0
+redis==3.2.1
```
jredrejo, line 34: are there any plans on how to package redis with kolibri in the pex installer?
This decision will affect all the installers deeply.

jredrejo (changes requested):

kolibri/deployment/default/cache.py
rtibbles: Good point - my impression from the `redis-cache` docs was that each Django cache needed a separate database number, but would be good to allow a custom starting number to be specified?

requirements/base.txt
rtibbles: The intention of this PR is simply to make it possible if Redis is available externally (for cloud hosted versions of Kolibri, or for kolibri-server). No plans for bundling yet!

kolibri/deployment/default/cache.py
jredrejo: I think so. I have worked previoulsy in servers where a single redis server was used by different applications and having the option to select the db is important.

rtibbles: OK, can add. I just only wanted to add one option, so that people don't have to configure each database number for each cache separately.

jredrejo (approved):

PR #5391, opened 2019-04-18
Add renderer for Slideshow content

nucleogenesis:
### Summary

This adds all functionality necessary for importing a Slideshow content kind from a Studio instance and displaying the Slideshow via the new slideshow_renderer plugin.
…

### Reviewer guidance

1. Once you have this branch running locally, go to Device and click [Import].
2. Add `http://nucleogenesis-slideshow-content-kind.studio.cd.learningequality.org` as a "Local network or internet" source.
3. Import from that source.
4. You should see "Learning Equality Slideshow Demo" as an available content node. Import that.
5. Find that content in Learn and test.

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated


rtibbles:
Note that since you made the database migration for this, a new one has been added to develop - best to delete your migration and rerun `makemigrations`.

Also, the rebase seems to have done something most peculiar?

indirectlylit:
(updated title b/c WIP is in the label)

nucleogenesis:
@rtibbles @indirectlylit 

Would you be willing to give this a review so we can get it merged?

A working demo of this with a Slideshow content is available on a demo server here:

(login: devowner / admin123)
http://35.229.70.110
http://35.229.70.110/learn/#/topics/c/da16403d0e22582c8ab589dc4e4fce9c  -> The slideshow content directly

rtibbles:
Yeah, now that we've branched for 0.12.y, we can merge this to develop sooner!

kolibri/core/content/migrations/0017_auto_20190415_1855.py
```diff
@@ -1,19 +0,0 @@
...
-
-class Migration(migrations.Migration):
-
-    dependencies = [("content", "0016_auto_20190124_1639")]
-
-    operations = [
-        migrations.AlterField(
-            model_name="channelmetadata",
-            name="published_size",
-            field=models.BigIntegerField(blank=True, default=0, null=True),
```
indirectlylit: We should generally just create new migrations rather than removing and re-creating existing ones. This ensures that every DB out there has a path 'upward'.

Also, we're not consistent about it, but it's good to name the migration something meaningful like `0018_do_something_special.py`

kolibri/plugins/slideshow_render/assets/src/views/SlideshowRendererComponent.vue
```diff
@@ -0,0 +1,214 @@
...
+</script>
+
+
+<style lang="scss" scoped>
+
+  .btn {
+    position: absolute;
+    top: 8px;
+    right: 8px;
+    z-index: 100;
```
indirectlylit: z-indexes should generally follow Material's [default elevations](https://material.io/design/environment/elevation.html#default-elevations)

kolibri/plugins/slideshow_render/assets/src/views/SlideshowRendererComponent.vue
```diff
@@ -0,0 +1,214 @@
...
+        type: Array,
+        default: () => [],
+      },
+    },
+    data: () => ({
+      isInFullscreen: false,
+      manifest: {},
+      slides: null,
+      currentSlide: null,
+      hooperSettings: {
```
indirectlylit: I don't actually see this used anywhere

kolibri/plugins/slideshow_render/assets/src/views/SlideshowRendererComponent.vue
```diff
@@ -0,0 +1,214 @@
...
+        centerMode: true,
+      },
+    }),
+    computed: {
+      slideshow_images: function() {
```
indirectlylit: we conventionally use `camelCase` on the JS side and `snake_case` on the python side

kolibri/plugins/slideshow_render/assets/src/views/SlideshowRendererComponent.vue
```diff
@@ -0,0 +1,214 @@
...
+        const files = this.files;
+        return files.filter(file => file.preset != 'Slideshow Manifest');
```
indirectlylit: We should use programmatic constant strings here like `slideshow_manifest` rather than things that look human-readable.

Otherwise it becomes tempting to display it to the end-user, like in #2371

kolibri/plugins/slideshow_render/assets/src/views/SlideshowRendererComponent.vue
```diff
@@ -0,0 +1,214 @@
...
+      <HooperPagination slot="hooper-addons" />
+    </Hooper>
+  </CoreFullscreen>
+
+</template>
+
+
+<script>
+
+  import _ from 'lodash';
```
indirectlylit: should be written `import orderBy from 'lodash/orderBy';` to avoid pulling in the entire lodash library

kolibri/plugins/slideshow_render/assets/src/views/SlideshowRendererComponent.vue
```diff
@@ -0,0 +1,214 @@
...
+    },
+    mounted() {
+      /*
+        Using the manifest file, get the JSON from the manifest, then
+        use the manifest JSON to get all slide images and metadata.
+      */
+      fetch(this.defaultFile.storage_url, { method: 'GET' })
+        .then(response => {
+          return response.status === 200 ? response.json() : {};
+        })
```
indirectlylit: we will probably need some loading and error states?

see e.g. `handleApiError` and `CORE_SET_PAGE_LOADING`

kolibri/plugins/slideshow_render/assets/src/views/SlideshowRendererComponent.vue
```diff
@@ -0,0 +1,214 @@
...
+          this.manifest = manifest_data.slideshow_data;
+
+          this.slides = _.orderBy(
+            this.manifest.map(image => {
+              return {
+                storage_url: this.slideshow_images.find(
+                  sFile =>
+                    sFile.storage_url
+                      .split('/')
+                      [sFile.storage_url.split('/').length - 1].split('.')[0] == image.checksum
```
indirectlylit: somewhat complex string logic .. maybe define this as a helper method with a unit test

kolibri/plugins/slideshow_render/assets/src/views/SlideshowRendererComponent.vue
```diff
@@ -0,0 +1,214 @@
...
+            ['asc']
+          );
+        });
+    },
+    methods: {
+      handleSlide(payload) {
+        this.currentSlide = this.slides[payload.currentSlide];
+      },
+      slideTextId(id) {
+        return 'descriptive-text-' + id;
```
indirectlylit, line 171: If for some reason we ever wanted to show two slide shows at the same time, we'd need to make this more unique. For example, 

```javascript
let component_id_counter = 0;

export default {
  beforeCreate() {
    this.component_id = component_id_counter.toString();
    component_id_counter += 1;
  },
  methods: {
    slideTextId(id) {
      return `${component_id}-descriptive-text-${id}`;
    }
  }
};
```

kolibri/plugins/slideshow_render/assets/src/views/SlideshowRendererComponent.vue
```diff
@@ -0,0 +1,214 @@
...
+
+<style lang="scss" scoped>
+
+  .btn {
+    position: absolute;
+    top: 8px;
+    right: 8px;
+    z-index: 100;
+    fill: white;
+  }
```
indirectlylit, line 222: suuuper minor but most other places in the code we put a line break between class definitions

kolibri/plugins/slideshow_render/assets/src/views/SlideshowRendererComponent.vue
```diff
@@ -0,0 +1,214 @@
...
+  .slideshow-renderer {
+    position: relative;
+    height: 70vh;
```
indirectlylit: we've tended to avoid the `vh` and `vw` units due to [early android browsers](https://caniuse.com/#feat=viewport-units)

kolibri/plugins/slideshow_render/assets/src/views/SlideshowRendererComponent.vue
```diff
@@ -0,0 +1,214 @@
...
+</template>
+
+
+<script>
+
+  import _ from 'lodash';
+
+  import themeMixin from 'kolibri.coreVue.mixins.themeMixin';
+  import contentRendererMixin from 'kolibri.coreVue.mixins.contentRendererMixin';
+  import responsiveElement from 'kolibri.coreVue.mixins.responsiveElement';
```
indirectlylit, line 59: when possible it's nice to avoid reliance on `responsiveElement` but sometimes it's a necessary evil

kolibri/plugins/slideshow_render/assets/src/views/SlideshowRendererComponent.vue
```diff
@@ -0,0 +1,214 @@
...
+    },
+    mixins: [contentRendererMixin, themeMixin, responsiveElement, responsiveWindow],
+    props: {
+      defaultFile: {
+        type: Object,
+        default: () => {},
+      },
+      files: {
+        type: Array,
+        default: () => [],
```
indirectlylit, line 103: whenever props are complex types (objects or arrays) we should also provide a `validator` function

kolibri/plugins/slideshow_render/assets/src/views/SlideshowRendererComponent.vue
```diff
@@ -0,0 +1,214 @@
...
+          :style="{
+            height: `calc(100% - ${caption_height}px)`
+          }"
+        >
+          <img
+            :src="slide.storage_url"
+            :aria-labelledby="slideTextId(slide.id)"
+          >
+        </div>
+        <div :id="slideTextId(slide.id)" class="hidden-descriptive-text">
```
indirectlylit: we have a global `visuallyhidden` class which should achieve this

kolibri/plugins/slideshow_render/assets/src/views/SlideshowRendererComponent.vue
```diff
@@ -0,0 +1,214 @@
...
+
+  import UiIconButton from 'kolibri.coreVue.components.UiIconButton';
+  import CoreFullscreen from 'kolibri.coreVue.components.CoreFullscreen';
+  import {
+    Hooper,
+    Slide,
+    Navigation as HooperNavigation,
+    Pagination as HooperPagination,
+  } from 'hooper';
+  import '../../styles/custom-hooper.css';
```
indirectlylit: let's move `custom-hooper.css` to the same directory as this component, since `SlideshowRendererComponent.vue` is the only place `Hooper` is used.

Also note that importing the css file like this will drop the styles into the global scope.

I don't know for sure, but we might be able scope it using something like...

```html
<style scoped>
  /deep/ .slideshow-renderer {
    @import './custom-hooper.css';
  }
</style>
```

or some other variation of sass-y goodness

kolibri/plugins/slideshow_render/assets/styles/custom-hooper.css
```diff
@@ -0,0 +1,145 @@
...
+.hooper-pagination {
+  position: absolute;
+  right: 50%;
+  bottom: 0;
+  display: flex;
+  padding: 5px 10px;
+  transform: translateX(50%);
+}
+.hooper-indicators {
+  display: flex;
```
indirectlylit: we should check whether things still work without flexbox since older android browsers don't support

kolibri/plugins/slideshow_render/assets/styles/custom-hooper.css
```diff
@@ -0,0 +1,145 @@
...
+  padding: 0;
+  margin: 0 auto;
+  list-style: none;
+}
+.hooper-indicator.is-active,
+.hooper-indicator:hover {
+  /* background-color: #4285f4; */
```
indirectlylit: dead code

indirectlylit (commented):
haven't tested yet, but here are some small notes - nothing major

kolibri/plugins/slideshow_render/assets/styles/custom-hooper.css
nucleogenesis: @indirectlylit I imagine that it could possibly be made to work without flexbox. Non-active slides are hidden to the side and the position of the active element is placed into view. I wish I'd realized that this wasn't going to work previously I recall now having read that we can't use flex in the style guide.

I'll see what I can do to replicate the effect without flexbox but may opt to try out other OSS plugins that give similar features if I can find one that doesn't make use of flexbox.

indirectlylit: see a similar thread that just occurred with @bjester :)

https://github.com/learningequality/kolibri/pull/5475#discussion_r283904383

if it's too hard to change for now, let's leave it and just make sure we're aware of it and give it extra attention during integration testing

kolibri/plugins/slideshow_render/assets/src/views/SlideshowRendererComponent.vue
```diff
@@ -0,0 +1,214 @@
...
+      100% - 50px
+    ); // Very much not going to be this way once the caption positioning is figured out
+
+    margin: 0 auto;
+    background-repeat: no-repeat;
+    background-position: center center;
+    background-size: contain;
+
+    img {
+      object-fit: contain;
```
bjester: If you don't already, I think you'll need a polyfill for this. IE 11 and old Android browsers do not support `object-fit`

kolibri/plugins/slideshow_render/assets/src/views/SlideshowRendererComponent.vue
nucleogenesis: @indirectlylit 

I refactored it to just return `response.json()` and added a `catch` block which dispatches `handleApiError` with the error object.

As for the `CORE_SET_PAGE_LOADING` - I tried:

```
mounted() {
      /*
        Using the manifest file, get the JSON from the manifest, then
        use the manifest JSON to get all slide images and metadata.
      */
      this.$store.commit("CORE_SET_PAGE_LOADING", true);
      fetch(this.defaultFile.storage_url, { method: 'GET' })
        .then(response => {
          return response.json();
        })
        .then(manifest_data => {
          this.manifest = manifest_data.slideshow_data;

          this.slides = orderBy(
            this.manifest.map(image => {
              return {
                storage_url: this.slideshowImages.find(
                  sFile => checksumFromFile(sFile) == image.checksum
                ).storage_url,
                caption: image.caption,
                sort_order: image.sort_order,
                id: image.checksum,
                descriptive_text: image.descriptive_text,
              };
            }),
            ['sort_order'],
            ['asc']
          );
          this.$store.commit("CORE_SET_PAGE_LOADING", false);
        })
        .catch(error => {
          this.$store.commit("CORE_SET_PAGE_LOADING", false);
          this.$store.dispatch('handleApiError', error);
        });
```

For some reason, this infinitely loops calling the `mounted()` method over and over. I looked a bit into that mutation and I'm not seeing how or why that would re-mount the component. 

Any thoughts on why this might be the case?

Also - FWIW - the functionality of the slideshow is to show a spinning loader until the images are all loaded. I've tested with bandwidth throttling to slow down the networking and it showed the spinner until it was fully loaded and ready.

kolibri/plugins/slideshow_render/assets/src/views/SlideshowRendererComponent.vue
nucleogenesis: Noted! 👍 Thanks Devon

kolibri/plugins/slideshow_render/assets/src/views/SlideshowRendererComponent.vue
nucleogenesis: Thanks for letting me know - I've added validators to both props. Do you think they cover what I'll need to validate well enough?

kolibri/plugins/slideshow_render/assets/src/views/SlideshowRendererComponent.vue
nucleogenesis: I refactored to import using CSS and moved the file. However I had issues when trying to import the CSS using `/deep/`  - since I have `<style scoped>` shouldn't that ensure the Hooper CSS is only used on this component?

kolibri/plugins/slideshow_render/assets/src/views/SlideshowRendererComponent.vue
rtibbles: Sorry, should have caught this sooner - you should use the kolibri `client` module: `import client from 'kolibri.client';` to do any fetches (rather than using the `fetch` API which is not as widely supported: https://caniuse.com/#feat=fetch (e.g. IE11, but also many others).

kolibri/plugins/slideshow_render/assets/styles/custom-hooper.css
nucleogenesis: I think I managed a solid solution. A good point from @bjester is 

>  I think in these scenarios where there is a question of whether this will work correctly, you can take into account the use and what happens when it breaks.

In my case it broke everything - Hooper was very reliant on flex to manage the spacing and flow of the slides.

kolibri/plugins/slideshow_render/assets/src/views/SlideshowRendererComponent.vue
nucleogenesis: Thanks @bjester - I hadn't considered how new object-fit is - I found and implemented a simple and working polyfill. In testing I was also was reminded that IE11 doesn't support fetch() so had to revise that to use XMLHTTPRequest.

kolibri/core/content/migrations/0017_auto_20190415_1855.py
nucleogenesis: @indirectlylit that makes sense. I cannot recall exactly why I deleted this migration. Should I leave it deleted in this instance since the repo includes a migration file that replaces this one?

kolibri/plugins/slideshow_render/assets/src/views/SlideshowRendererComponent.vue
bjester: Nice! Yeah, object-fit is really nice so glad the polyfill worked!

kolibri/plugins/slideshow_render/assets/src/views/SlideshowRendererComponent.vue
nucleogenesis: @rtibbles thanks for letting me know! Done and pushed.

kolibri/core/content/migrations/0017_auto_20190415_1855.py
rtibbles: The other thing to bear in mind is that other migrations might be added which need to be previous to this in the migration tree (e.g. https://github.com/learningequality/kolibri/pull/5497/files#diff-21bfbbd534842850877c325a17b3d32a), so the correct migration (particularly on the develop branch) can end up being a bit of a moving target.

The best thing to do in this case would be to reinstate the previous `0017` migration, then create a new migration just for what you are doing.

If the migration I note above ends up getting merged before your PR, then we will need to recreate the migration to take that into account. Otherwise it can get sorted out by the person who merges it into develop.

nucleogenesis:
@indirectlylit @rtibbles @bjester 

I've gone through all of the feedback above and pushed several commits addressing the issues, particularly those that could affect compatibility.

- Refactored Hooper CSS to work without flexbox
- Replaced my `fetch()` with `kolibri.client`
- Use a polyfill for the CSS `object-fit` to accommodate IE11 and older Android.
- Misc refactoring, code cleanup, formatting fixes.

I've tested in Chrome & Firefox on Linux, IE11 on Windows 7 (Browser Stack) and 10 (Proper Install) and Chrome on Android v4 (Browser Stack).

I have also updated the "Reviewer Guidelines" in the description for the PR which will help you quickly get Slideshow content into your local Kolibri instance and ready for test. Please let me know if you run into any issues!

kolibri/plugins/slideshow_render/assets/src/views/SlideshowRendererComponent.vue
indirectlylit: yes if it works without `/deep/` that's great. Sometimes I've encountered issues where that wasn't the case

kolibri/plugins/slideshow_render/assets/src/views/SlideshowRendererComponent.vue
indirectlylit: I guess I'm actually still left a bit confused as to what form of data these props expect. From the validators I see a schema like:

```
defaultFile: { storage_url: "*json" }
```

and 

```
files: [
  { storage_url: "*" },
  //...
]
```

but then down below I also see a reference to `file.preset`.

I guess what I would expect to see in either comments, validator function, or both is a complete description of the shape of the data for the things that the component is calling a "file" (or maybe more accurately a "file descriptor object"?).

indirectlylit:
Hey @nucleogenesis  -

I pushed a quick commit to resolve some minor conflicts. Let's just follow up on the [migrations thread](https://github.com/learningequality/kolibri/pull/5391#discussion_r283567537) above, and then we can merge this in and iterate in a future PR

integration_testing/features/learner/learner-engage-slideshow.feature
```diff
@@ -0,0 +1,59 @@
+Feature: Learner engages with content of the slideshow kind
```
nucleogenesis, line 1: @radinamatic Would you be able to give this gherkin story a look and let me know if it covers everything you think that should be covered in a story for the slideshow?

integration_testing/features/learner/learner-engage-slideshow.feature
```diff
@@ -0,0 +1,59 @@
...
+      Then the slideshow presents the slide that is associated with the dot's positional order
+
+    Scenario: Engaging with full screen mode
+      When I click the enter full screen button
+      Then I view the slideshow renderer in full screen
+        And I maintain my position in the slideshow
+        And I see left and right arrow buttons on the left and right sides of the screen
+        And I see the exit full screen button in the top right corner of the screen
+        And I can navigate the content by keyboard, swiping or clicking the left and right arrow buttons
+      When I am viewing in full screen mode and click the exit full screen button or hit ESC
```
radinamatic: It doesn't seem necessary to repeat you are in full screen mode here, shorten to `When I click the exit full screen button or hit ESC`.

integration_testing/features/learner/learner-engage-slideshow.feature
```diff
@@ -0,0 +1,59 @@
...
+      Given that <resource> resource is a slideshow
+        When I click the <resource> resource
+        Then I see the *Channels > '<channel>' > '<topic>' > '<resource>'* breadcrumb
+          And I see the <resource> content
+
+    Scenario: Engage with the slideshow content
+      When I view a slide that has caption text
+      Then I see the caption text beneath the slide image
+      When I am viewing the first slide
+      Then I the left-most pagination dot is filled in with color and the rest of them are white
```
radinamatic: Delete the first `I`.

integration_testing/features/learner/learner-engage-slideshow.feature
```diff
@@ -0,0 +1,59 @@
...
+        And the left arrow button is disabled
+        And tapping the left key on my keyboard does nothing
+        And swiping the image to the right does nothing
+      When I am viewing the last slide
+      Then I the right-most pagination dot is filled in with color and the rest of them are white
```
radinamatic: Delete the first `I`.

radinamatic (changes requested):
@nucleogenesis I pulled your branch locally but wasn't able to add `http://nucleogenesis-slideshow-content-kind.studio.cd.learningequality.org` as a "Local network or internet" source and do a proper test. 

Gherkin scenario looks complete otherwise, I left just a couple of comments. :+1:

radinamatic:
Ok, a few suggestions for improvements related to keyboard navigation, feel free to work on them in the following PR, if preferred:

1. Make the visual focus indicator around the full screen button bigger - at the moment it is overlapping the button itself, and it's not easily noticeable.
2. When you navigate the dots with the <kbd>Tab</kbd> key, but then you switch to using arrow keys, these seem to take "precedence" (that is, slides become active as keys are pressed), but the focus and the ring remain where the <kbd>Tab</kbd> key stopped previously, and that got me confused for a moment. I don't know if it would be possible, or even advisable, to make the focus and the ring jump to  the active slide and dot too?

slideshow

radinamatic (approved):

nucleogenesis:
@indirectlylit @radinamatic 

I have opened a PR for the user docs here: https://github.com/learningequality/kolibri-docs/pull/113

nucleogenesis:
> Ok, a few suggestions for improvements related to keyboard navigation, feel free to work on them in the following PR, if preferred:
> 
> 1. Make the visual focus indicator around the full screen button bigger - at the moment it is overlapping the button itself, and it's not easily noticeable.
> 2. When you navigate the dots with the Tab key, but then you switch to using arrow keys, these seem to take "precedence" (that is, slides become active as keys are pressed), but the focus and the ring remain where the Tab key stopped previously, and that got me confused for a moment. I don't know if it would be possible, or even advisable, to make the focus and the ring jump to  the active slide and dot too?
> 
> slideshow

RE: #1 - I checked this on a different content type (PDF) and this seems to be an issue that could be addressed separately from this PR - perhaps we should create a new issue?

RE #2 - That is interesting. I wonder what the ideal solution would be for people who are most inclined to need to use keyboard navigation. To me - FWIW - when I think of the keyboard navigation "focus" functionality, I think of my "focus" indicator as sort of like "where my mouse would be if I were using one". So I wouldn't necessarily expect that focus indicator to also then map to the state of the content that I'm viewing. Rather - I'd consider it to be where hitting the `Enter` key would simulate clicking. In any case - if adjusting how the focus is managed would better serve the most common users of keyboard navigation, then I think that I can make it work accordingly.

Issue #5147, opened 2019-02-25
UNIQUE constraint failed: logger_masterylog.id


indirectlylit:
### Observed behavior

getting a few of these errors on Sentry: 

https://sentry.io/organizations/learningequality/issues/901657313/?project=1378215&query=is%3Aunresolved&statsPeriod=14d&utc=false

### Expected behavior

unique constraint satisfied

### User-facing consequences

unknown

### Errors and logs

```
IntegrityError: UNIQUE constraint failed: logger_masterylog.id
  File "django/core/handlers/exception.py", line 41, in inner
    response = get_response(request)
  File "django/core/handlers/base.py", line 187, in _get_response
    response = self.process_exception_by_middleware(e, request)
  File "django/core/handlers/base.py", line 185, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "django/views/decorators/csrf.py", line 58, in wrapped_view
    return view_func(*args, **kwargs)
  File "rest_framework/viewsets.py", line 116, in view
    return self.dispatch(request, *args, **kwargs)
  File "rest_framework/views.py", line 495, in dispatch
    response = self.handle_exception(exc)
  File "rest_framework/views.py", line 455, in handle_exception
    self.raise_uncaught_exception(exc)
  File "rest_framework/views.py", line 492, in dispatch
    response = handler(request, *args, **kwargs)
  File "rest_framework/mixins.py", line 21, in create
    self.perform_create(serializer)
  File "rest_framework/mixins.py", line 26, in perform_create
    serializer.save()
  File "rest_framework/serializers.py", line 214, in save
    self.instance = self.create(validated_data)
  File "rest_framework/serializers.py", line 943, in create
    instance = ModelClass._default_manager.create(**validated_data)
  File "django/db/models/manager.py", line 85, in manager_method
    return getattr(self.get_queryset(), name)(*args, **kwargs)
  File "django/db/models/query.py", line 394, in create
    obj.save(force_insert=True, using=self.db)
  File "kolibri/core/auth/models.py", line 200, in save
    super(AbstractFacilityDataModel, self).save(*args, **kwargs)
  File "morango/models.py", line 438, in save
    super(SyncableModel, self).save(*args, **kwargs)
  File "morango/utils/uuids.py", line 118, in save
    super(UUIDModelMixin, self).save(*args, **kwargs)
  File "django/db/models/base.py", line 808, in save
    force_update=force_update, update_fields=update_fields)
[...]

Issue #4119, opened 2018-08-07
Error when mastering a Perseus exercise for a second time


jonboiser:
### Observed behavior

Get this error when mastering an exercise for the second time.

e.g. at http://localhost:8000/learn/#/topics/c/69e5e6abf479581483d441b83d7d76f4

I mastered this exercise once, then refreshing and mastering it again attempts a POST to `/api/masterylog` with an incomplete payload

```
{"completion_timestamp":"2018-08-07T17:58:43.148Z","end_timestamp":"2018-08-07T17:58:43.111Z","pastattempts":[{"correct":1},{"correct":1,"hinted":false,"error":false},{"correct":1,"hinted":false,"error":false},{"correct":1,"hinted":false,"error":false},{"correct":1,"hinted":false,"error":false},{"correct":1,"hinted":false,"error":false},{"correct":1,"hinted":false,"error":false}],"totalattempts":7}
```

Resulting in a validation error

```
{"mastery_level":["This field is required."],"start_timestamp":["This field is required."],"summarylog":["This field is required."],"user":["This field is required."]}
```

(It does not have problems if I just continue doing exercises and extend my streak, which results in PATCHING the previous masterylog, I assume)

### Expected behavior

There are no problems when re-mastering an exercise in a new session.

### User-facing consequences

None, because the initial mastery log is in the server. But an uncaught runtime error could have unexpected side effects in the app.

### Errors and logs

…

### Steps to reproduce

1. Master a perseus exercise
1. Continue your mastery streak by getting one more assessment right.
1. Observe no errors
1. Refresh your browser on that exercise's page
1. "Re-master" it by getting one more assessment correct
1. Observe error

### Context

Develop

rtibbles:
Could be related to #5147

indirectlylit:
is this issue still extant, or would it have been addressed by the PR referenced above?

rtibbles:
I think we should close this unless it recurs.

jonboiser:
Just saw this again on the kolibribeta server (version 0.13.0). I don't think #5393 would have fixed it, since it doesn't address what I described in the OP: that we're attempting to PATCH the log with and incomplete payload. The unique ID constraint thing seems like it was different.

pcenov:
@radinamatic I'm not able to replicate this issue in 0.15.1 - no errors observed after attempting to master a Perseus exercise for a second time.

rtibbles:
Closing this as fixed.

PR #5393, opened 2019-04-19
Return calculated masterylog id and avoid constraint failed error

### LINKED ISSUE #5377: Auto sign-out from tablets during normal class
**Observed behavior**
Kids are signout out of their tabs automatically while watching videos or doing quizzes

**Expected behavior**
The user should not be signed out automatically.

**User-facing consequences**
While the kids are in the middle of any exercise or video, they are being automatically signed out. They need to again re-login to continue the session

**Steps to reproduce**
This happens in some of the tabs during a normal classroom session.

**Context**
Kolibri version : Kolibri 0.12.2
Operating system : ubuntu 14.04
Browser : Chrome

ad5ad6e1-e3cf-497c-a107-e26479a172a2

jredrejo:
### Summary
Sometimes the frontend sends two different requests for the same logger to be saved. It seems to happen on high concurrency , probably because two different backend threads receive the same request after the user refreshes the page or insists sending it. Also #4119 gives more reasons for this to happen.

Taking the frontend reasons to do it apart, this PR deals with the backend to avoid the server sending a wrong response, disconnecting the user and avoiding him to continue the normal exercises flow.

On one side, it adds logging for the 400 errors. This will help understanding why this kind of error happens.
On the other side, if the object has already been created by the backend returns it, with its calculated id.
Using create_or_get strategy has not been possible due to the way the id is calculated in the model.

### Reviewer guidance
1. Using  _postman_ or any other REST util, doing a request like http://localhost:8000/api/logger/masterylog/?1555673658140=1555673658140 should returns a 400 error. `kolibri.log` should contain the information of the reasons for the error, indicating the fields that are missed or malformed
2. Using _Firefox_, in the network tab, filter the requests using the '`masterylog`' word. Browsing to the learn tab, when doing non-previously run exercises one request will appear. Then, using the "Edit and resend" option Firefox  has, resend the same request. Before applying this PR a 500 error will happen and the "Sorry! Something went wrong! " message will appear. After applying this PR the user won't notice anything and no error will happen.

### References
Closes: #5147 
Relates: #4119 
Probably will close #5377 because their logs show this same error.

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/core/logger/api.py
+++ b/kolibri/core/logger/api.py
@@ -11,6 +11,7 @@ from django_filters.rest_framework import FilterSet
 from rest_framework import filters
 from rest_framework import status
 from rest_framework import viewsets
+from rest_framework.exceptions import ValidationError
 from rest_framework.response import Response
 
 from .models import AttemptLog
@@ -107,7 +108,7 @@ class LoggerViewSet(viewsets.ModelViewSet):
             obj.id = obj.calculate_uuid()
             final_obj = self.get_serializer(obj)
             return Response(final_obj.data)
-        except Exception as e:
+        except ValidationError as e:
             logger.error("Failed to validate data: {}".format(e))
             return Response(request.data, status.HTTP_400_BAD_REQUEST)
 
--- a/kolibri/core/logger/api.py
+++ b/kolibri/core/logger/api.py
@@ -1,11 +1,15 @@
+import logging
+
 from django.core.exceptions import ObjectDoesNotExist
 from django.db.models.query import F
+from django.db.utils import IntegrityError
 from django.http import Http404
 from django_filters import ModelChoiceFilter
 from django_filters.rest_framework import CharFilter
 from django_filters.rest_framework import DjangoFilterBackend
 from django_filters.rest_framework import FilterSet
 from rest_framework import filters
+from rest_framework import status
 from rest_framework import viewsets
 from rest_framework.response import Response
 
@@ -35,6 +39,8 @@ from kolibri.core.auth.models import FacilityUser
 from kolibri.core.auth.models import LearnerGroup
 from kolibri.core.content.api import OptionalPageNumberPagination
 from kolibri.core.exams.models import Exam
+
+logger = logging.getLogger(__name__)
 
 
 class BaseLogFilter(FilterSet):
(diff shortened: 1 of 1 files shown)


kolibri/core/logger/api.py
```diff
@@ -90,6 +96,21 @@ def update(self, request, *args, **kwargs):
...
+        except IntegrityError:
+            # The object has been created previously: let's calculate its id and return it
+            serializer = self.get_serializer(data=request.data)
+            serializer.is_valid(raise_exception=True)
+            obj = serializer.Meta.model(**serializer.validated_data)
+            obj.id = obj.calculate_uuid()
+            final_obj = self.get_serializer(obj)
+            return Response(final_obj.data)
+        except Exception as e:
+            logger.error("Failed to validate data: {}".format(e))
```
rtibbles, line 112: Would it work to just reraise the exception here, in case it is something more severe than just a validation error?

jredrejo: yep, good idea. I only could reproduce the validation error (`serializer.is_valid(raise_exception=True) ` is what does it) but that does not exclude other possibilities. I will add it.

jredrejo: Done

rtibbles (approved):
I think this is the right approach, as noted the id is calculated using the Model itself, so this is the easiest way to do it. I can't imagine any problematic data loss occurring by just calculating and returning the id, as this is happening before any significant user interaction has occurred.

kolibri/core/logger/api.py
```diff
@@ -90,6 +97,21 @@ def update(self, request, *args, **kwargs):
                 default_response[field] = method(instance)
         return Response(default_response)
 
+    def create(self, request, *args, **kwargs):
+        try:
+            return super(LoggerViewSet, self).create(request, *args, **kwargs)
+        except IntegrityError:
```
indirectlylit, line 103: interesting - like a "create or get" command?

rtibbles: Pretty much - except because the id is calculated from the Model, we just return the model id from an instance of the model.

PR #5398, opened 2019-04-20
No unused Vuex actions and mutations lint rule

### LINKED ISSUE #4992: Create linting rules to check for unused props, data, computed props, and vuex mappers

MisRob:
### Summary

Last #4992 lint rule that disallows unused Vuex mutations and actions + related cleanup.

### Reviewer guidance

- Works fine?
- Are there more scenarios to be treated?
- Is cleanup ok?

### References

Closes #4992.

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/core/assets/src/views/AppBar.vue
+++ b/kolibri/core/assets/src/views/AppBar.vue
@@ -94,7 +94,7 @@
 
 <script>
 
-  import { mapGetters, mapState, mapActions } from 'vuex';
+  import { mapGetters, mapState } from 'vuex';
   import themeMixin from 'kolibri.coreVue.mixins.themeMixin';
   import UiToolbar from 'kolibri.coreVue.components.UiToolbar';
   import UiIconButton from 'kolibri.coreVue.components.UiIconButton';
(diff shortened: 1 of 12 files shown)


rtibbles:
I think the only suggestion I would make is to add a wider range of test cases for the invalid tests.

MisRob:
> I think the only suggestion I would make is to add a wider range of test cases for the invalid tests

I wasn' sure if many "invalid" unit tests are helpful since there's probably an infinite amount of "unused" combinations and if there's a passing test for one scenario, it doesn't necessarily mean that something really similar won't fail. So I'm not sure what exactly to test here with units (unlike for "valid"
tests where it's much more clear).

At least I added a couple of specs for various `map` syntaxes a while ago - looks useful. Any other ideas?

rtibbles:
I think this is sufficient - just covering the major bases you are trying to catch is helpful, mostly because the 'valid' tests would pass without the linting rule being present at all. The valid tests help to make sure that the rule is not throwing false positives, but the invalid tests help to make sure it is not throwing false negatives.

rtibbles (approved):

MisRob:
Yeah, makes sense!

PR #5400, opened 2019-04-21
Custom HTMLHint rules cleanup

### LINKED ISSUE #5337: HTMLHint custom rules cleanup
There's a couple of custom HTMLHint errors. Some of them (possibly all?) might be removed and included in the autoformatting process only. Review custom HTMLHint rules in detail, remove duplicate behavior if any and move rest of the rules to the autoformatting process if possible.

MisRob:
### Summary

- remove duplicate self-closing check from custom HTMLHint rules
- automatically indent top-level template block (this has already worked for script and style blocks) => I could also remove "top-level content surrounded by one empty line" HTMLHint rule for all blocks

After reviewing how custom linting works in detail, I think that eventually we could get rid of all custom HTMLHint checks and move everything to the format process but I didn't want to deal with it in the scope of HTMLHint duplicates cleanup because it seems that it would require some extra refactoring of _lint.js_ and I am not sure if that would be worthy. I also noticed a couple of minor problems that seemed to be caused by the need to coordinate all those tools and custom checks - I fixed one of them in this PR and skipped others for now because I think that a bit more suitable approach might be to define what conventions we want to actually follow and then maybe prepare the lint process again from scratch while trying to simplify it as much as possible and move more responsibility to external tools in general to save some maintenance time spent on _lint.js_ in the future.

### Reviewer guidance

- Are all adjustments OK?

### References

Closes  #5337.

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/packages/kolibri-tools/lib/lint.js
+++ b/packages/kolibri-tools/lib/lint.js
@@ -220,13 +220,24 @@ function lint({ file, write, encoding = 'utf-8', silent = false } = {}) {
           let block;
           // First lint the whole vue component with eslint
           formatted = eslint(source);
+
+          let vueComponent = compiler.parseComponent(formatted);
+
+          // Format template block
+          if (vueComponent.template && vueComponent.template.content) {
+            formatted = insertContent(
+              formatted,
+              vueComponent.template,
+              vueComponent.template.content
+            );
+            vueComponent = compiler.parseComponent(formatted);
+          }
+
           // Now run htmlhint on the whole vue component
           let htmlMessages = HTMLHint.verify(formatted, htmlHintConfig);
           if (htmlMessages.length) {
             messages.push(...HTMLHint.format(htmlMessages, { colors: true }));
           }
-
-          let vueComponent = compiler.parseComponent(formatted);
 
           // Format script block
           if (vueComponent.script) {
--- a/packages/kolibri-tools/lib/lint.js
+++ b/packages/kolibri-tools/lib/lint.js
@@ -191,9 +191,8 @@ function lint({ file, write, encoding = 'utf-8', silent = false } = {}) {
               if (linted.trim() !== (stylinted || code).trim()) {
                 notSoPretty = true;
               }
-              if (linted.trim() !== code.trim()) {
-                styleCodeUpdates.push(() => callback(linted));
-              }
+
+              styleCodeUpdates.push(() => callback(linted));
             })
             .catch(err => {
               messages.push(err.toString());
--- a/kolibri/core/assets/src/styles/definitions.scss
+++ b/kolibri/core/assets/src/styles/definitions.scss
@@ -64,4 +64,3 @@ $login-overlay: #201a21;
   backface-visibility: hidden;
   perspective: 10000px;
 }
-
--- a/packages/kolibri-tools/lib/htmlhint_custom.js
+++ b/packages/kolibri-tools/lib/htmlhint_custom.js
@@ -43,15 +43,6 @@ HTMLHint.addRule({
               event.raw
             );
           }
-          if (match && match[2].length !== 2) {
-            reporter.error(
-              'Top-level content should be indented two spaces.',
-              event.line,
-              event.col,
-              self,
-              event.raw
-            );
-          }
         }
       }
     });
@@ -75,15 +66,6 @@ HTMLHint.addRule({
             if (match[1].length !== 2) {
               reporter.error(
                 'Top-level content should be surrounded by one empty line.',
-                event.line,
-                event.col,
-                self,
-                event.raw
-              );
-            }
-            if (match[2].length !== 2) {
-              reporter.error(
-                'Top-level content should be indented two spaces.',
                 event.line,
                 event.col,
                 self,
--- a/packages/kolibri-tools/lib/htmlhint_custom.js
+++ b/packages/kolibri-tools/lib/htmlhint_custom.js
@@ -10,32 +10,6 @@ function clean(val) {
 }
 
 /*
-  Based on the existing `tag-self-close` rule
-*/
-HTMLHint.addRule({
-  id: '--no-self-close-common-html5-tags',
-  description: 'Self-closing HTML5 tags are not valid.',
-  init: function(parser, reporter) {
-    var self = this;
-    var commonTags = parser.makeMap(`a,audio,b,body,button,canvas,caption,center,dir,div,dl,em,font,footer,
-      form,h1,h2,h3,h4,h5,h6,head,header,html,iframe,label,li,main,map,menu,nav,object,option,output,p,progress,
-      q,script,section,select,span,strong,style,sub,table,tbody,td,textarea,th,thead,time,title,tr,u,ul,var,video`);
-    parser.addListener('tagstart', function(event) {
-      var tagName = event.tagName.toLowerCase();
-      if (event.close && commonTags[tagName]) {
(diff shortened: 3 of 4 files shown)


MisRob:
I haven't noticed that there are also specs for these rules - will update it later.

MisRob:
Done.

packages/kolibri-tools/lib/lint.js
```diff
@@ -191,9 +191,8 @@ function lint({ file, write, encoding = 'utf-8', silent = false } = {}) {
               if (linted.trim() !== (stylinted || code).trim()) {
                 notSoPretty = true;
               }
-              if (linted.trim() !== code.trim()) {
-                styleCodeUpdates.push(() => callback(linted));
```
rtibbles, line 195: What was the issue you were seeing here?

packages/kolibri-tools/lib/lint.js
```diff
@@ -221,14 +220,25 @@ function lint({ file, write, encoding = 'utf-8', silent = false } = {}) {
           let block;
           // First lint the whole vue component with eslint
           formatted = eslint(source);
+
+          let vueComponent = compiler.parseComponent(formatted);
+
+          // Format template block
+          if (vueComponent.template && vueComponent.template.content) {
+            formatted = insertContent(
```
rtibbles, line 228: Did you do this because you were seeing instances of the block indentation and line spacing being stripped by eslint?

rtibbles (commented):
HTMLHint changes look fine, a few questions about the lint modifications.

rtibbles:
Ultimately, I think we should get rid of HTMLHint (sooner rather than later would be great). If we could implement all of our custom template formatting rules in eslint, I think that would be great. We can implement the script indentation using: https://vuejs.github.io/eslint-plugin-vue/rules/script-indent.html, with a base indent of 2.

We could then copy that rule for a 'template' and 'style' indent rule as well, to add the base indent to both of those blocks.

Finally, we would need to have a way of enforcing the blank lines within each block, but I imagine that shouldn't be difficult?

As you say, reducing the number of linting tools would allow us to greatly reduce the complexity of `lint.js` - although currently the main complexity comes from adding stylint integration into the style blocks of Vue components, because stylint only offers an async interface, whereas all the other linters are synchronous. At the very least, we may be able to make the code more readable by using async and await.

packages/kolibri-tools/lib/lint.js
MisRob: `insertContent` (the function responsible for inserting a block while adjusting it's indentation and content spacing) is passed here as `callback` function when styles are being processed. This was working properly only when there were some stylelint updates of the block. When there were none, this condition caused that the correct indentation wasn't applied to style block, for example:

```
<style lang="scss" scoped>
.red {
  color: red;
}
</style>
```

would not be formatted to
```
<style lang="scss" scoped>

  .red {
    color: red;
  }

</style>
```

In such cases, HTMLHint saved it and reported `"'Top-level content should be surrounded by one empty line.'"` and `"Top-level content should be indented two spaces."` so you could still fix it manually. But I removed those two checks from HTMLHint in this PR so I also reviewed existing auto-formatting functionality and adjusted it to work for those scenarios, too.

packages/kolibri-tools/lib/lint.js
MisRob: This logic was added because I noticed that auto-indentation/spacing is already working for `style` and `script` blocks => HTMLHints's `"'Top-level content should be surrounded by one empty line.'"` and `"Top-level content should be indented two spaces."` was duplicate behaviour.

=> I wanted to remove these two duplicate checks from HTMLHint but I noticed that I would still need to leave some of them there because of `template` block which spacing wasn't auto-formatted. I didn't want to make both formatting process and HTMLHint responsible for basically the same thing, so instead of removing those two HTMLHint checks only partially, I added auto-indent/spacing to template block which gave me the freedom to get rid of `"'Top-level content should be surrounded by one empty line.'"` and `"Top-level content should be indented two spaces."` completely.

MisRob:
Yep, exactly, I also believe that most of the logic could be moved to eslint as you say. About `style` blocks, I think we might be able to add indentation from within eslint-plugin-vue, too. Although the parser is currently not being able to retrieve the style block itself, we can still read the whole file there (but would need to check if that's the right way..). Also, we were talking about unused classes detection with Devon so we could also bear in mind that it would be nice if the solution used for accessing `style` block would be working for such use cases as well. For all other `style` block checks, I was thinking if we could just call `stylelint` on the whole `.vue` file instead of extracting `style` blocks? I think that `stylelint` should be already capable of reading style directly from SFCs. Or is there any other reason for doing so in our own process?

MisRob:
I'll wait a while if there's more input on this topic or any updates needed here and then fill in a ticket.

packages/kolibri-tools/lib/lint.js
MisRob: There's probably no need to insert 2 spaces indent again to template block (`vue/html-indent` should take care of it), but there's still need to add an empty line to the beginning/end of the template block and for that purpose, I just re-used the logic we have already and yeah, right now it causes duplicate behavior of 2 spaces indent in template because they're inserted again in the section you referenced - not exactly harmful but also kind of useless and misleading. I suppose that adding `indent` Bool parameter to `insertContent` and setting it to `false` at this particular place could solve this for now. Is this what you mean or is there anything else?

rtibbles: Yeah - maybe the best thing to do is to add this flag for insertContent, and only set it to true for style blocks. You can then add in the `vue/html-indent` and `vue/script-indent` rules to handle the other two cases.

This paves the way for some follow up work that could hopefully get rid of the need for the function all together, by adding a `vue/style-indent` rule, and a `vue/line-space-before-and-after-blocks` rule (or something snappier, maybe!).

rtibbles:
> For all other style block checks, I was thinking if we could just call stylelint on the whole .vue file instead of extracting style blocks? I think that stylelint should be already capable of reading style directly from SFCs. Or is there any other reason for doing so in our own process?

I think this is correct, it should pull out any embedded styles from the vue component file.

Could you file follow up issues from the comments here, and we can merge this as is?

rtibbles:
Overall, I think the desired things are:
1. Eliminate HTMLHint and replace all rules with eslint ones.
2. If possible, run prettier on the entire vue component, and using the eslint-plugin-prettier.
3. Run stylelint on the entire vue component after eslint/prettier has been run on it, to reduce callback/promise hell.

So `lint.js` would be simplified to:
1. If JS, run eslint with prettier plugin.
2. If CSS/SCSS/SASS/LESS, run stylelint (open question of whether we want to run prettier or not on our style files too).
3. If Vue component, run stylelint, then run eslint with prettier.

Some of this is dependent on whether we are able to run prettier with eslint in the way we want, and also whether we want to run prettier on stylint.

rtibbles (approved):
Questions answered, everything else can be filed as follow up issues.

MisRob:
👍 Sounds like a plan. Thanks for a nice summary - I'm going to move it together with other notes to a new ticket in the evening.

rtibbles:
The main tricky thing that I didn't highlight above is the prettier integration. We have it setup to only run prettier on separate blocks of the component, because our overall component styling would clash with prettier. Ideally we could just run prettier on the whole component, but that may never work.

PR #5388, opened 2019-04-18
Update coach content filter in Lesson Management Search Page

jonboiser:
### Summary
1. Implements the change in comments in #5033 
1. Changes how incremental progress is calculated in MediaPlayerIndex (might address #5354, but not confirmed).
1. Move the total points icon (Fixes #5358)

Screen Shot 2019-04-18 at 2 39 42 PM

Screen Shot 2019-04-19 at 1 05 19 PM

### Reviewer guidance

Test the filter with results that have some/no coach contents

### References
Fixes #5033

----

### Contributor Checklist

PR process:

- [x] PR has the correct target branch and milestone
- [x] PR has 'needs review' or 'work-in-progress' label
- [ ] If PR is ready for review, a reviewer has been added. (Don't use 'Assignees')
- [ ] If this is an important user-facing change, PR or related issue has a 'changelog' label
- [ ] If this includes an internal dependency change, a link to the diff is provided

Testing:

- [ ] Contributor has fully tested the PR manually
- [ ] If there are any front-end changes, before/after screenshots are included
- [ ] Critical user journeys are covered by Gherkin stories
- [ ] Critical and brittle code paths are covered by unit tests

### Reviewer Checklist

- Automated test coverage is satisfactory
- PR is fully functional
- PR has been tested for [accessibility regressions](http://kolibri-dev.readthedocs.io/en/develop/manual_testing.html#accessibility-a11y-testing)
- External dependency files were updated if necessary (`yarn` and `pip`)
- Documentation is updated
- Contributor is in AUTHORS.md

PR #5401, opened 2019-04-21
Small clean up from #5388

rtibbles:
### Summary
Small cleanup from #5388.


rtibbles:
Self merging to fix the build.

PR #5406, opened 2019-04-22
Docker speedup

DXCanas:
### Summary

Speeds up build times a modest amount by caching pip packages in a layer and yarn packages in volume (due to our use of [yarn workspaces](https://yarnpkg.com/lang/en/docs/workspaces/). 

Volume can be easily wiped out if the cache becomes an issue. See comments.

### Reviewer guidance

- Test the build! Make sure buildkite isn't complaining and the pex works. 
- This will be harder, but compare build times! See build for this vs others

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/docker/build_whl.dockerfile
+++ b/docker/build_whl.dockerfile
@@ -42,8 +42,6 @@ RUN yarn config set cache-folder /yarn_cache
 # Copy all files in this directory
 COPY . .
 
-VOLUME /kolibridist/  # for mounting the whl files into other docker containers
-
 CMD yarn install --pure-lockfile
     make dist && \
     make pex && \
--- a/docker/build_whl.dockerfile
+++ b/docker/build_whl.dockerfile
@@ -25,16 +25,22 @@ RUN apt-get update && \
     rm nodejs_$NODE_VERSION-1nodesource1_amd64.deb && \
     apt-get install yarn
 
-RUN git lfs install
+RUN git lfs install &&\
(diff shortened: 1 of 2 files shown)


docker/build_whl.dockerfile
```diff
@@ -42,8 +42,6 @@ RUN yarn config set cache-folder /yarn_cache
 # Copy all files in this directory
 COPY . .
 
-VOLUME /kolibridist/  # for mounting the whl files into other docker containers
```
ivanistheone, line 32: Yes, this makes sense to remove.

This was part of an earlier design where the volume for the container that has the build assets was being mounted from another container doing other packaging steps. we seem to have moved away from that design and instead use the `-v $$PWD/dist:/kolibridist` volume mount with assets showing up in /dist
https://github.com/learningequality/kolibri/blob/develop/Makefile#L221

DXCanas: Yeah, i noticed the removal was inconsequential - curious about what the use case was for mounting to another container?

ivanistheone: From what I remember windows packageing container (or another one) needed the .whl from the build-whl container outputs (this was before the `-v $$PWD/dist:/kolibridist`)

rtibbles (approved):

Issue #4616, opened 2018-12-17
Fix demo site banner behavior


khangmach:
### Observed behavior

Some issues:
- Upon page load, banner takes up a lot of space and options "Create an account" and "continue as guest" are the only immediately visible options. 
- Upon closing the banner information, Kolibri logo is never visible.
- Banner position is absolute, which cuts the logo. 
- Verbose text -- needs a cut

<img width="413" alt="screen shot 2018-12-17 at 1 10 36 pm" src="https://user-images.githubusercontent.com/25045166/50115739-2a53cf00-01fd-11e9-870b-3e6d6dfa8897.png">
<img width="364" alt="screen shot 2018-12-17 at 1 10 41 pm" src="https://user-images.githubusercontent.com/25045166/50115751-2f188300-01fd-11e9-8df2-011b8510b29e.png">

### Expected behavior
- Banner to be a div that is relative and is above the rest of the page. This will make the the logo show
![scrollable area](https://user-images.githubusercontent.com/25045166/50116641-a0593580-01ff-11e9-87bb-c879b660833b.gif)

- On page load, an overlay to appear to prevent the user to click on "Create an account"/"continue as guest". Overlay disappears when user presses "CLOSE"
![overlay](https://user-images.githubusercontent.com/25045166/50116738-dc8c9600-01ff-11e9-8bac-3b9442719f38.png)

- Simpler verbiage overall
![demo site language](https://user-images.githubusercontent.com/25045166/50116800-05ad2680-0200-11e9-848e-5b62e1bb28ef.png)

### User-facing consequences
Confusing usability, buggy look, improper logo visibility

rtibbles:
Note: implementation of this would be dependent on having a core banner component.

khangmach:
Can we cut the verbiage and use css to fix the absolute positioning of the banner in the interim while I whip up the banner proposal?

indirectlylit:
see also #4562

laurenlichtman:
@khangmach let's definitely cut the text. I like the edit you proposed with one small change. Can you change the second to last sentence to read: "This demo has features of the latest version of Kolibri, and only contains a sample of the content that is available for offline use."  cc: @lauradanforth as FYI in case she has further edits. 

I think it'd be great if we could proceed as you suggested Khang! Thanks!

nucleogenesis:
@indirectlylit @rtibbles 

I am updating the copy on the banner to match the right side of this image:

[...]

PR #5412, opened 2019-04-23
Update Demo Banner Styling & Copy

nucleogenesis:
### Summary

This pushes the login/signup page view down so that the Kolibri logo is visible when the "Welcome" demo banner is visible.

This also creates a semi-transparent mask over the form page - so a user may not engage with the page until they've closed the banner.

This PR will also include updates to the copy on the banner itself.

### Reviewer guidance

When running in demo mode (enable the demo_server plugin `kolibri plugin kolibri.plugins.demo_server enable`) then go to the sign in page. You will see the banner having pushed the form down and the Kolibri logo is visible. You should not be able to touch the form until the banner is closed.

### References

This PR is in response to this issue (where I just posted an accessibility-related question):
https://github.com/learningequality/kolibri/issues/4616

Also - it would be worth viewing on different screen sizes to be sure - in my testing, it looked decent enough. On mobile, the banner takes up almost all of the screen - but on tablet/laptop sizes, it looked solid in my testing.
----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated


nucleogenesis:
@indirectlylit @rtibbles 

*Copying this from the related issue*

I am updating the copy on the banner to match the right side of this image:

https://user-images.githubusercontent.com/25045166/50116800-05ad2680-0200-11e9-848e-5b62e1bb28ef.png

The last paragraph is giving me a little trouble because it is trying to inline <KExternalLink> components inside of a string defined in the `$trs`.

Here are the relevant `$trs`:

```
demoServerP3: 'To learn more about using Kolibri, take a look at our [documentation guides] and [download the latest release].',
demoServerA1: 'documentation guides',
demoServerA2: 'download the latest release',
```

In the current version - the `demoServerP3` text is written to be followed by a list of the <KExternalLink> components - but the revised version is looking to inline those links.

I tried passing an object like:

```
<p>{{ $tr('demoServerP3', {demo: <KExternalLink href='#' text='$tr("demoServerA1")' />) }}</p>
 /* ... later on ... */
demoServerP3: 'To learn more about using Kolibri, take a look at our {demo} and ...',
```

But Vue told me "This isn't React get that mess out of here". 

Do you have any thoughts on how I can maintain the desired accessibility while keeping the `$trs` all properly "self contained" as meaningful and complete strings of text unto themselves?

indirectlylit:
Looking good! few notes below...

------

> The last paragraph is giving me a little trouble because it is trying to inline components inside of a string defined in the $trs.

Right, unfortunately we cannot currently combine inline markup (formatting or links) with translated text. We'll have to use a bulleted list so that each link is a self-contained translated phrase. (cc @khangmach)


------

| note | image |
|--|--|
| 12px -> 16px margins [from Material Design defaults](https://material.io/design/layout/responsive-layout-grid.html#columns-gutters-margins) | image |
| strange layout behavior observed on Chrome |   layout |

This is a pre-existing issue, but I think we should fix it here - basically _all_ `/user` pages show the banner. It would be better if only the login page showed it:

> image

------

Finally, there's something a bit strange about how we're adding banner-related styles and logic to `UserIndex`, and using `kolibriGlobal` to pass messages. My sense is we should do one of two things:

1. If we want to keep using the plugin strategy, we need to make sure that the `user` plugin knows nothing about the banner. This seems like a lost cause...
2. I think an easier strategy would be to get rid of the demo plugin, and bake the banner into the user plugin. It can be disabled by default and enabled using an [undocumented `options.ini` setting](https://github.com/learningequality/kolibri/blob/develop/kolibri/utils/options.py). (this seems much easier .. curious what you think @rtibbles)

nucleogenesis:
@indirectlylit @rtibbles 

How/where can I get the values that are loaded using the options.ini setting in the frontend? Should it be included somehow in an existing serializer that gets loaded into an 'options' field? That sounds like a much better approach to me and would be a relatively simple refactor.

Also - latest commit fixes the text margins issue and the mask scrolling with the window issue and includes an updated copy with the exception of the last bit that calls the user to read the docs and download/install.

rtibbles:
We don't serialize any values from options into the frontend, so we don't have any prior art. I am a little resistant to move away from the plugin method, mostly because I think otherwise we risk overloading the `options.ini` with non-user facing configuration, but also because I would like us to keep pushing towards Kolibri being properly pluggable (which currently it is in only a relatively limited sense).

We currently have our sentry configuration setup using options, and I think this should also be implemented as a plugin, because it clutters the main code otherwise.

My assumption for how this was to be implemented was to have a core banner component https://material.io/design/components/banners.html that the demo server plugin could then conditionally populate.

To make the conditionality simpler, we could just keep showing the banner, as long as the user is not authenticated, which would serve a similar purpose to checking the URL, so as soon as they login, it disappears.

nucleogenesis:
My initial instinct was to just make sure the banner was first in the DOM by changing its modules.js kind of like:

```
ready() {
    const modalDiv = global.document.createElement('div');
    global.document.body.insertBefore(modalDiv, global.document.body.firstChild);
    this.rootvue = new Vue(Object.assign({ el: modalDiv, store }, DemoServerIndex));
  }
```

That way the banner could be `position:relative` and push the next DOM element (the main-wrapper) down along with it. I ran into some issues where some part of that was `position:absolute` - so the content wasn't being pushed down as expected.

Perhaps I abandoned that path too early and can revisit it because it could both decouple the banner's logic from the User module while keeping the pluggable apps working as they do now.

rtibbles:
I would suggest adding it to the `CoreBase/index.vue` so that it can interact properly with everything that is going on in there if it is shown.

I _think_ we should be able to then set a component into a slot using some vuex state, but this sounds a little messy.

rtibbles:
This would mean that CoreBase would look something like this:

See below!

Then our ready method would look something like:
```
ready() {
    if (!store.getters.isUserLoggedIn) {
        store.commit('SET_CORE_BANNER_SLOT', DemoServerIndex);
    }
  }
```

rtibbles:
Actually, the slot might be superfluous:
```
<template>
   <CoreBanner v-if="bannerContent">
       <component :is="bannerContent"></component>
   </CoreBanner>
...

**Comment by nucleogenesis** [comment]
I'll give that a try later this afternoon!

What of @indirectlylit 's concern about showing the banner only on the sign in page? I believe there is a Vuex state field saying what the current page is (ie, SIGN_IN, SIGN_UP, etc). 

In any case - I'll just try to get the display logic working and we can adjust whether we display only on Sign In or on all User routes.

Thanks!

**Comment by rtibbles** [comment]
> What of @indirectlylit 's concern about showing the banner only on the sign in page? I believe there is a Vuex state field saying what the current page is (ie, SIGN_IN, SIGN_UP, etc).

That's what I was attempting to emulate with the `isUserLoggedIn` getter - I think it makes a similar amount of sense, because the banner shows you the login credentials, and reminds you that you can login. When you are logged in, then it can hide it.

**Comment by nucleogenesis** [comment]
@rtibbles @indirectlylit 

Here is what has changed in the last commit:

- The `ready()` function in the demo_server module.js commits a Vuex mutation storing the DemoServerIndex component in state.
- The `CoreBase/index` component now renders the component from the Vuex state using a getter.
- The events are now local to the component context and do not use `kolibriGlobal` - but I'm still using events to manage the state.

So - there still is a coupling of the banner and another module's component's logic because `CoreBase/index` now has local state determining whether or not to apply certain CSS based on the state.

I could not manage to get everything working properly using the DOM flow when I moved the point of entry for DemoServerIndex into the CoreBase/index. I got close but the transition was offsetting part of the content with two vertical scrollbars and I couldn't get it to work as expected.

In any case - I look forward to hearing your thoughts.

*Richard* - Your example includes a CoreBanner component. I just went through and built it to get it working but I'm wondering what you'd expect a CoreBanner component to do other than serve as an empty wrapper.

Thanks guys!

**Comment by rtibbles** [comment]
> Your example includes a CoreBanner component. I just went through and built it to get it working but I'm wondering what you'd expect a CoreBanner component to do other than serve as an empty wrapper.

I would expect the `CoreBanner` to handle the open/close/dismiss state, and the component inserted into it to handle the content. That way the state changing logic can be entirely encapsulated into the CoreBanner/CoreBase components, without leakage from the DemoServer component.

File: kolibri/core/assets/src/views/CoreBanner.vue [hunk 1]
```diff
@@ -0,0 +1,119 @@
...
+      };
+    },
+    mounted() {
+      this.$root.$on('toggleBannerState', () => {
+        this.bannerClosed = !this.bannerClosed;
+        this.$root.$emit('bannerStateChanged', { bannerClosed: this.bannerClosed });
+      });
+    },
+    $trs: {
+      demoServerHeader: 'Welcome to the Kolibri demo site!',
```
rtibbles: Looks like there may be some left over stuff here from a copy?

rtibbles:
Latest code makes sense to me on a read through - we may want to generalize the variable names if we use this as a banner component in other scenarios. Will do some manual testing and review before merge!

kolibri/plugins/demo_server/assets/src/DemoServerBannerHeader.vue
```diff
@@ -0,0 +1,36 @@
+<template>
+
+  <h1>
+    Welcome to the Kolibri demo site!
```
indirectlylit: unwrapped string

kolibri/plugins/demo_server/assets/src/DemoServerBannerHeader.vue
```diff
@@ -0,0 +1,36 @@
...
+    <KButton
+      class="open-button"
+      :text="'More info'"
+      appearance="flat-button"
+      :primary="true"
+      @click="$root.$emit('toggleBannerState')"
+    />
```
indirectlylit: unwrapped string 'More info'

I'd recommend moving the 'close' and 'more info' buttons into the `CoreBanner` component so that the `demo_server` plugin doesn't need take part in event handling. I think this is what @rtibbles was [recommending here](https://github.com/learningequality/kolibri/pull/5412#issuecomment-488339611)

kolibri/core/assets/src/views/CoreBase/index.vue
```diff
@@ -65,6 +65,12 @@
       :class="fullScreen ? 'scrolling-pane' : 'content'"
       :style="contentStyles"
     >
+      <CoreBanner
+        v-if="demoBannerComponents"
+        :defaultComponent="demoBannerComponents.DemoServerBannerContent"
+        :persistentComponent="demoBannerComponents.DemoServerBannerHeader"
+      />
```
indirectlylit: Rather than registering two independent components passing them in as props, I'd recommend using a slot-and-string-based API for [consistency with the `KModal`](https://github.com/learningequality/kolibri/blob/develop/kolibri/core/assets/src/views/KModal.vue), and pass the title is as a string prop.

If you were to do this, you would end up with an API that looks like:

```html
<CoreBanner v-if="bannerIsVisible" :title="bannerTitle">
  <component :is="BannerComponent" />
</CoreBanner>
```

----

A somewhat unrelated note: saving the component definition to Vuex state will cause it to become a reactive object. Not sure if this is actually bad or not, but intuitively it feels a bit iffy to me...

Another strategy would be to follow the example of how nav bar items are added using a [global `register` function](https://github.com/learningequality/kolibri/blob/develop/kolibri/plugins/coach/assets/src/views/CoachSideNavEntry.vue#L41). The component and title would then become available inside `CoreBase` off of a core JS module rather than off of vuex state.

Combined with parameterizing the title as mentioned above, this might look something like:

```javascript
import DemoServerBannerContent from './DemoServerBannerContent';
import KolibriModule from 'kolibri_module';
import coreBannerContent from 'kolibri.utils.coreBannerContent';
import { createTranslator } from 'kolibri.utils.i18n';

const translator = createTranslator('DemoServer', {
  title: 'Welcome to the Kolibri demo site!',
});

class DemoServerModule extends KolibriModule {
  ready() {
    coreBannerContent.register(DemoServerBannerContent, translator.$tr('title'))
    store.commit('SET_CORE_BANNER_VISIBLE', !store.getters.isUserLoggedIn);
  }
}
```

where only the `SET_CORE_BANNER_VISIBLE` boolean is stored in Vuex and used to hide or show the banner

indirectlylit (dismissed):
this is looking great!

Added some ideas for making it more consistent with existing "prior art", but not absolutely critical. The important thing is getting it tested and merged next week.

The main issues that I'm seeing are:

----

## Scrolling behavior

When you scroll down, you get permanently stuck and can't interact with the site:

scroll

~My understanding from #4616 was that the new design would remove the modal [scrim](https://ux.stackexchange.com/questions/114449/what-is-the-transluscent-background-of-a-modal-dialog-called) entirely.~

[EDIT] - re-reading, it says:

> On page load, an overlay to appear to prevent the user to click on "Create an account"/"continue as guest". Overlay disappears when user presses "CLOSE"

So it sounds like the intent is that scrolling should be disabled when the banner is expanded. However this introduces issues on mobile:

| | |
|--|--|
| unhandled mobile behavior | image |

My recommendation would be to simply remove the scrim and allow the user to scroll down to the login page, whether or not the banner is expanded. @khangmach thoughts? I can't think of any reason to prevent interaction with the login form while the banner is visible.

----

## Closed state

In the closed state, the banner is still overlapping rather than pushing down the content:

image

indirectlylit:
WIP -> label

nucleogenesis:
@indirectlylit The latest commit:

- Moves state management logic and buttons into the CoreBanner component.
- Consolidates DemoServerBannerHeader into DemoServerBannerContent
- Fixes a few unwrapped strings - adds a couple of new wrapped strings for the buttons (openButton and closeButton).
- Uses scoped slots to pass props from CoreBanner to whatever Component it is given to render in CoreBase.
- Removes the semi-transparent overlay, effectively fixing the scrolling issue
- Makes the closed banner relative positioning, so it no longer overlays the page itself

I agree with you, Devon, that there is no real reason to force the user to close the banner in order to interact with the page. One thing that is worth considering is that we are giving usernames and passwords in the banner content and, even though they are simple and short, it could be useful to have them visible while trying to login for a user.

If this were actually to be used as a component in other places, we could implement that semi-transparent mask with a Boolean prop but I don't see it as necessary now.

*RE: Using a register module, removing from Vuex* - I wanted to get everything else worked out and improved before moving this way. If there are no issues with the above I'll rework the module to implement a register function like how you demonstrated above.

khangmach:
Let's ditch the scrim behavior for now and go with what you two suggested: having banner info (opened or closed state) push the rest of the page down

nucleogenesis:
@indirectlylit shall I move forward with removing the Vuex usage?

indirectlylit:
> shall I move forward with removing the Vuex usage?

Yes, I think maintaining consistency with how the side nav is implemented would help set a standard precedent for future similar features

kolibri/plugins/demo_server/assets/src/DemoServerBannerContent.vue
```diff
@@ -0,0 +1,85 @@
...
+    float: right;
+    margin-bottom: 24px;
+  }
+  .open-button {
+    float: right;
+    margin-top: 0;
+  }
+  .banner-header {
+    float: left;
+  }
```
indirectlylit: When there are horizontal layouts, we should specify grid behaviors for the different screen sizes (cc @khangmach).

(Floats tend to cause issues when adjacent items collide with each other, while the grids keep everything nice and consistent.)

For example, the spec might say...

> (Columns not drawn to scale)
> 
> **large size - max-width: 1000px centered**
> 
> ```
> -----------------------------------------------------------------------
> | title: 10/12 cols left-aligned              |     button: 2/12 cols |
> |                                             |         right-aligned |
> -----------------------------------------------------------------------
> ```
> 
> **medium size**
> ```
> ----------------------------------------------------------
> | title: 6/8 cols left-aligned      |   button: 2/8 cols |
> |                                   |      right-aligned |
> ----------------------------------------------------------
> ```
> 
> **small size**
> ```
> ---------------------------------------
> | title: 4/4 cols left-aligned        |
> ---------------------------------------
> | button: 4/4 cols left-aligned       |
> ---------------------------------------
> ```

And this could be implemented with the responsive grid using (roughly):

```html
<KGrid>
  <KGridItem sizes="4, 6, 10">
    {{ title }}
  </KGridItem>
  <KGridItem sizes="4, 2, 2">
    <KButton />
  </KGridItem>
</KGrid>
```

(Unfortunately that makes it a bit harder to drop in the Content component as you have implemented here.)

kolibri/plugins/demo_server/assets/src/DemoServerBannerContent.vue
```diff
@@ -0,0 +1,85 @@
...
+      </li>
+      <li>
+        <KExternalLink
+          href="https://learningequality.org/download/"
+          :text="$tr('demoServerA2')"
+          target="_blank"
+        />
+      </li>
+    </ul>
+  </div>
```
indirectlylit, line 32: Won't necessarily help here, but just a heads-up though that vue has a clever nested `<template>` feature and you can do things like:

```html
<template>
  <div>
    <h1>always shown</h1>
    <template v-if="condition">
      <p>conditionally</p>
      <p>shown</p>
    </template>
  </div>
<template>
```

kolibri/plugins/demo_server/assets/src/DemoServerBannerContent.vue
nucleogenesis: Thanks Khang! I think I can work this out using the KGrid

nucleogenesis:
@khangmach thanks for pointing out the KGrid! That worked out very well.

@indirectlylit Could you give the coreBannerContent module a look when you can? Everything works but this is a newer approach to me so I'm hoping I've not missed anything.

kolibri/core/assets/src/views/CoreBanner.vue
```diff
@@ -0,0 +1,118 @@
...
+    components: { KButton, KGrid, KGridItem },
+    mixins: [themeMixin],
+    data() {
+      return {
+        bannerClosed: false,
+      };
+    },
+    computed: {
+      contentGridSizes() {
+        return this.bannerClosed ? '4, 6, 10' : '4, 8, 12';
```
indirectlylit: `KGridItem` also accepts arrays and numbers for sizes. whole thing could be written inline like 

```html
<KGridItem :sizes="bannerClosed ? [4, 6, 10] : [4, 8, 12]">`
```

kolibri/core/assets/src/views/CoreBanner.vue
```diff
@@ -0,0 +1,118 @@
+<template>
+
+  <div>
```
indirectlylit: unnecessary wrapper div

kolibri/core/assets/src/views/CoreBanner.vue
```diff
@@ -0,0 +1,118 @@
...
+    <div
+      class="banner"
+      :style="{ background: $coreBgLight}"
+      :tabindex="0"
+      role="dialog"
```
indirectlylit: `tabindex` and `role` held over from previous work with scrim I assume

kolibri/core/assets/src/views/CoreBanner.vue
```diff
@@ -0,0 +1,118 @@
...
+        <KGrid>
+          <!-- Grid Content -->
+          <KGridItem :sizes="contentGridSizes">
+            <slot :bannerClosed="bannerClosed"></slot>
+          </KGridItem>
+
+          <!-- Grid Buttons -->
+          <KGridItem v-if="!bannerClosed" sizes="4, 8, 12">
+            <KButton
+              v-if="!bannerClosed"
```
indirectlylit: redundant with `v-if="!bannerClosed"` on parent `KGridItem`

also, note `KGridItem` also accepts sizes in percentages like

```html
<KGridItem size="100" percentage>
```

kolibri/plugins/demo_server/assets/src/DemoServerBannerContent.vue
```diff
@@ -0,0 +1,85 @@
...
+</script>
+
+
+<style lang="scss" scoped>
+
+  .close-button {
+    float: right;
+    margin-bottom: 24px;
+  }
+  .open-button {
```
indirectlylit: these classes can be deleted now

kolibri/plugins/demo_server/assets/src/DemoServerBannerContent.vue
```diff
@@ -0,0 +1,85 @@
+<template>
+
+  <h1 v-if="bannerClosed" class="banner-header">
+    {{ $tr('demoServerHeader') }}
+  </h1>
+  <div v-else>
+    <h1>{{ $tr('demoServerHeader') }}</h1>
```
indirectlylit, line 7: with the new implementation, this can be simplified to just putting the `v-if` on the `<h1>`!

kolibri/core/assets/src/views/CoreBanner.vue
```diff
@@ -0,0 +1,118 @@
...
+    padding-left: 12px;
+    margin: 0 auto;
+
+    h1 {
+      font-weight: bold;
+    }
+  }
+
+  .close-button {
+    float: right;
```
indirectlylit: can replace this with `alignment="right"` on parent `KGridItem`

kolibri/core/assets/src/views/CoreBanner.vue
```diff
@@ -0,0 +1,118 @@
...
+    margin: 0 auto;
+    overflow-y: auto;
+    box-shadow: 0 2px 50px rgba(0, 0, 0, 1);
+  }
+
+  .banner-inner {
+    max-width: 1000px;
+    padding-top: 0;
+    padding-right: 12px;
+    padding-left: 12px;
```
indirectlylit: in Material conventions, paddings are generally multiples of 8px

kolibri/core/assets/src/views/CoreBanner.vue
```diff
@@ -0,0 +1,118 @@
...
+
+  .banner {
+    position: relative;
+    top: 0%;
+    left: 0%;
+    z-index: 7500;
+    width: 100%;
+    margin: 0 auto;
+    overflow-y: auto;
+    box-shadow: 0 2px 50px rgba(0, 0, 0, 1);
```
indirectlylit: we have a set of Material-based built-in drop-shadows. In this case, we should match the login pane lower on the page:

https://github.com/learningequality/kolibri/blob/develop/kolibri/plugins/user/assets/src/views/SignInPage/index.vue#L477


I think I think `top`, `left`, `overflow-y`, and `z-index` are no longer necessary?

indirectlylit (approved):
This looks awesome, well done!

I left a few more notes mostly as FYIs, and pushed some quick cleanup. We're good to go once that builds. Thank you!

kolibri/plugins/demo_server/assets/src/DemoServerBannerContent.vue
indirectlylit: (actually not quite - nevermind)

nucleogenesis:
Thanks for the detailed notes and for the refinement! @indirectlylit

Issue #5370, opened 2019-04-15
Enumerating channel database files includes those that are broken symlinks


jamalex:
### Observed behavior

On a RACHEL, we create a symlink within the `.kolibri/content/databases/` directory to the content databases in the RACHEL modules directory. Uninstalling a module is meant to remove the symlinks, but if the module files are somehow deleted without the cleanup scripts being run, the symlinks may be left behind but be broken.

When we then enumerate the channel database files on upgrade, it includes those that are broken symlinks, but then when we try to actually load them, it fails to load the file.

```
Traceback (most recent call last):
  File "/usr/bin/kolibri", line 9, in <module>
    load_entry_point('kolibri==0.12.2', 'console_scripts', 'kolibri')()
  File "/usr/lib/python3/dist-packages/kolibri/utils/cli.py", line 673, in main
    start(port, daemon=daemon)
  File "/usr/lib/python3/dist-packages/kolibri/utils/cli.py", line 281, in start
    update()
  File "/usr/lib/python3/dist-packages/kolibri/utils/cli.py", line 260, in update
    update_channel_metadata()
  File "/usr/lib/python3/dist-packages/kolibri/core/content/utils/annotation.py", line 44, in update_channel_metadata
    channel_ids = get_channel_ids_for_content_database_dir(get_content_database_dir_path())
  File "/usr/lib/python3/dist-packages/kolibri/core/content/utils/channels.py", line 38, in get_channel_ids_for_content_database_dir
    if os.path.getsize(filename) == 0:
  File "/usr/lib/python3.4/genericpath.py", line 50, in getsize
    return os.stat(filename).st_size
FileNotFoundError: [Errno 2] No such file or directory: '/root/.kolibri/content/databases/197934f144305350b5820c7c4dd8e194.sqlite3'

root@WAPD-235N-Server:~# ls /root/.kolibri/content/databases/
197934f144305350b5820c7c4dd8e194.sqlite3  4da5cdd8cb7d431f9666c50ebcf9290f.sqlite3  a9b25ac9814742c883ce1b0579448337.sqlite3
1ceff53605e55bef987d88e0908658c5.sqlite3  a1e879bdbe7647d3aa73e8e3e753e1dd.sqlite3  e409b964366a59219c148f2aaa741f43.sqlite3
```

### Expected behavior

[...]

PR #5419, opened 2019-04-24
skip/remove broken symlink or corrupted db files

lyw07:
### Summary

This PR is to fix the issue that database files are corrupted or have broken symlinks when enumerating channels.
There are three cases:
1. The database file in the kolibri home directory is corrupted and the channel is not imported to kolibri --> start kolibri --> kolibri enumerates channels in the home directory data folder --> kolibri finds the corrupted database file and raises a Database error --> error gets caught by the program so kolibri skips reading the channel

2. The database file in the USB data folder is corrupted and the channel is not imported to kolibri --> start kolibri and try to import channels from USB --> kolibri enumerates channels in the USB data folder --> kolibri finds the corrupted database file and raises a Database error --> error gets caught by the program so kolibri skips reading the channel

3. The database file has a symbolic link that is broken --> start kolibri --> kolibri enumerates channels in the home directory data folder --> kolibri finds that the database file with broken symlink actually does not exist --> raise FileNotFoundError --> error gets caught by the program and the database file is removed

### Reviewer guidance

Case 1:
1. create a file `~/.kolibri/content/databases/000409f81dbe5d1ba67101cb9fed4530.sqlite3` and write string `test file` inside to make the file corrupted. 
2. start kolibri
3. Check that there is a warning message in the log: `WARNING  Tried to import channel 000409f81dbe5d1ba67101cb9fed4530, but database file was corrupted.`

Case 2:
1. create a file `000409f81dbe5d1ba67101cb9fed4530.sqlite3` inside the USB data folder and write string `test file` inside to make the file corrupted.
2. start kolibri
3. import content from USB
4. Check that there is a warning message in the log: `WARNING  Tried to import channel from database file <USB>/KOLIBRI_DATA/content/databases/000409f81dbe5d1ba67101cb9fed4530.sqlite3, but the file was corrupted.`

Case 3:
1. make sure that the files `/path/000409f81dbe5d1ba67101cb9fed4530.sqlite3` and `<kolibri_home>/content/databases/000409f81dbe5d1ba67101cb9fed4530.sqlite3` do not exist
2. run 
```
import os
os.symlink("/path/000409f81dbe5d1ba67101cb9fed4530.sqlite3", "<kolibri_home>/content/databases/000409f81dbe5d1ba67101cb9fed4530.sqlite3")
```
3. check that `<kolibri_home>/content/databases/000409f81dbe5d1ba67101cb9fed4530.sqlite3` now exists and has a symlink
4. start kolibri
5. check that there is a warning message in the log: `WARNING  Removing nonexistent or empty databases in content database directory '/Users/lingyiwang/.kolibri/content/databases' with IDs: set(['000409f81dbe5d1ba67101cb9fed4530']).
Please import the channels again.`
6. check that the database file `<kolibri_home>/content/databases/000409f81dbe5d1ba67101cb9fed4530.sqlite3` does not exist now

### References

https://github.com/learningequality/kolibri/issues/5370

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/core/content/utils/channels.py
+++ b/kolibri/core/content/utils/channels.py
@@ -6,7 +6,6 @@ from sqlalchemy.exc import DatabaseError
 
 from .paths import get_content_database_dir_path
 from .sqlalchemybridge import Bridge
-from kolibri.core.content.models import ChannelMetadata
 from kolibri.core.discovery.utils.filesystem import enumerate_mounted_disk_partitions
 from kolibri.utils.uuids import is_valid_uuid
 
@@ -42,11 +41,7 @@ def get_channel_ids_for_content_database_dir(content_database_dir):
         if not os.path.exists(filename) or os.path.getsize(filename) == 0:
             db_files_to_remove.add(db_name)
             os.remove(filename)
-            # Delete the channel metadata from the database if exists, so that
-            # users can download the database file again from the channel import page.
-            channel = ChannelMetadata.objects.filter(id=db_name)
-            if channel:
-                channel.delete()
+
     if db_files_to_remove:
         err_msg = (
             "Removing nonexistent or empty databases in content database directory "
--- a/kolibri/core/content/utils/channels.py
+++ b/kolibri/core/content/utils/channels.py
@@ -2,8 +2,11 @@ import fnmatch
 import logging
 import os
 
+from sqlalchemy.exc import DatabaseError
+
 from .paths import get_content_database_dir_path
 from .sqlalchemybridge import Bridge
+from kolibri.core.content.models import ChannelMetadata
 from kolibri.core.discovery.utils.filesystem import enumerate_mounted_disk_partitions
 from kolibri.utils.uuids import is_valid_uuid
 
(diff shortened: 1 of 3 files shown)


kolibri/core/content/utils/channels.py
```diff
@@ -30,18 +33,29 @@ def get_channel_ids_for_content_database_dir(content_database_dir):
...
+        if not os.path.exists(filename) or os.path.getsize(filename) == 0:
+            db_files_to_remove.add(db_name)
             os.remove(filename)
-    if empty_db_files:
-        logger.warning("Removing empty databases in content database directory '{directory}' with IDs: {names}"
-                       .format(directory=content_database_dir, names=empty_db_files))
-    valid_dbs = list(set(valid_db_names) - set(empty_db_files))
+            # Delete the channel metadata from the database if exists, so that
+            # users can download the database file again from the channel import page.
+            channel = ChannelMetadata.objects.filter(id=db_name)
```
rtibbles: Hrm, this seems potentially somewhat destructive. This could completely remove a channel from someone's system just because a DB file got corrupted.

rtibbles (changes requested):
Code generally looks good - I think we should not remove channels from the default database though, as it could cause considerable user confusion and distress if content disappeared.

kolibri/core/content/utils/channels.py
lyw07: yes that makes sense. I'm doing this here because if the user tries to export the channel with a corrupted DB file to USB, then a FileNotFound error will be raised and the user will not be able to export the channel unless they manually download the DB file from studio and put that into Kolibri home directory. Deleting the channel from the default database will not remove the existing content of the channel. If the user go to the channel import page, they will get to download the channel DB file and still see the existing content. However, it will definitely be confusing to them that the channel disappears with no indication. Do you have any suggestions on handling this case? Thank you!

rtibbles: From the user's perspective, deleting the channel from the default database does _delete_ the channel content (because they will not be able to see it in the interface any more).

So, I think it would be better to catch the FileNotFound error in the export step, so that only one user flow is disrupted (export) and not two (interacting with content and export).

lyw07: yes that sounds better! I assume that we can just throw the FileNotFound exception in the export step without doing any extra implementations?

jamalex:
Another case to consider, if it's possible to include here: content files may also be symlinks. Do they currently get marked (on `LocalFile`) as available or unavailable, if the corresponding symlink is broken?

rtibbles:
It uses `os.exist` (although we might want to strengthen that to verify the checksum), how would a broken sym-link register?

lyw07:
@jamalex ~~The files are marked as available currently. If that's the case, should we just skip reading this file so that users will be prevented from exporting the file to USB?~~ The files are actually marked as unavailable.
@rtibbles could you please explain a bit more about "strengthen that to verify the checksum"? I'm not sure if I'm following correctly. 
Thank you!

rtibbles:
We currently just check that the file exists, using `os.exists`. We could make the check more robust, and get the md5 checksum of the file at that file location, and verify that it matches the checksum stored on the `LocalFile` model. If it doesn't match (i.e. it's corrupted, 0 length, or someone has swapped in a different file), we could mark it as unavailable.

lyw07:
thank you! I will update the PR based on that

lyw07:
We handle the content files with broken symlinks here: https://github.com/learningequality/kolibri/blob/release-v0.12.x/kolibri/core/content/utils/annotation.py#L171

Since we have a solution to the issue in 0.12.x branch now, I'm not going to add checksum implementation in this PR. I will do that in a separate PR target develop branch.

ralphiee22 (commented):
Manually tested 👍

rtibbles (approved):

PR #5426, opened 2019-04-25
Make scrolling behaviour more consistent and not rely on setting window properties

rtibbles:
### Summary
* Removes most reliance on vuejs router history scroll behavior
* Utilises the history state key that vuejs uses to independently store a record of scroll positions
* Encapsulates all scroll behaviour inside core base component
* Removes Vuex state for scroll history

### Reviewer guidance
Is scroll state consistent preserved across browser history navigation?

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/core/assets/src/router.js
+++ b/kolibri/core/assets/src/router.js
@@ -4,6 +4,8 @@ import store from 'kolibri.coreVue.vuex.store';
 
 const logging = logger.getLogger(__filename);
 
+const scrollPositions = {};
+
 /** Wrapper around Vue Router.
  *  Implements URL mapping to Vuex actions in addition to Vue components.
  *  Otherwise intended as a mostly transparent replacement to vue-router.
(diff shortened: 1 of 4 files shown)


indirectlylit:
was there a specific issue this was addressing, so I can compare before/after behavior?

rtibbles:
Behaviour reported by @jonboiser that scroll state was being inconsistently set (#3659). This seems to show some improvement, but doesn't resolve all the issues he has seen.

Further, by circumventing the vue router scroll, this makes the scroll state saving work on more browsers (because some would error when explicitly setting the pageYOffset which is read only in the spec).

rtibbles:
Have added more scroll capture behaviour, so that we capture the current scroll height before we leave a route.

jonboiser (approved):

PR #5430, opened 2019-04-25
Fix #4782 epub rendering issues of tables

### LINKED ISSUE #4782: Epub render - does not split tables and font resizing issue
### Observed behavior

1. EPUB files containing tables do not split-render the tables.
2. Font resizing does not work.

Problem can be seen in this capture -

arryfqrrpl

### Expected behavior

- Table should be split over multiple pages.
- Font resizing should work 

Expected behaviour shown in the capture herein from Microsoft edge -

smvgozehpn

### User-facing consequences

- Content not accessible to users when part of tables are hidden from view.
- Inconvenience for readers who may font a bigger or smaller font size.

### Errors and logs

NA

### Steps to reproduce

Upload the attached epub into a studio channel and take it offline in Kolibri app.

### Context

 * Kolibri version - 0.11.0

bjester:
### Summary

- Updated epub render to split tables when it's detected that the table isn't breaking into separate column, as other content does in CSS column layout
  - This mainly affects Firefox, which is a [known limitation](https://caniuse.com/#feat=multicolumn)
- Added styling to give tables a more responsive width, somewhat fixing their rendering in non-fullscreen view

#### Table wrap issue, before
Firefox Table Overflow

#### Table wrap issue, after fix
Firefox Table Overflow - Fixed

#### Non-fullscreen, before
epub-initial-load

#### Non-fullscreen, after
table-responsiveness

### Reviewer guidance

NOTE: Test using `yarn run devserver`, as epub renderer doesn't work with `devserver-hot` command

- [ ] Load epubs content from token `jafal-zovuk` into Kolibri instance
- [ ] Open "reflowable" epubs in Firefox
- [ ] Verify tables do not overflow out of epub viewport
- [ ] Verify no regressions in other browsers, like Chrome 

Notes:
- The section slider/progress bar is misbehaving in Firefox, but hasn't regressed due to these changes
- Chrome still allows tables to collide in non-fullscreen, although not as badly as before

### References

Issue #4782 

----

### Contributor Checklist

PR process:

Testing:

- [ ] ~Critical user journeys are covered by Gherkin stories~

### Reviewer Checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/plugins/document_epub_render/assets/src/views/EpubRendererIndex.vue
+++ b/kolibri/plugins/document_epub_render/assets/src/views/EpubRendererIndex.vue
@@ -277,6 +277,7 @@
           h3: { ...colorStyle },
           h4: { ...colorStyle },
           h5: { ...colorStyle },
+          table: { width: '100%!important' },
           'p:first-of-type::first-letter': { ...colorStyle },
           video: { 'max-width': '100%' },
         };
--- a/kolibri/plugins/document_epub_render/assets/src/views/EpubRendererIndex.vue
+++ b/kolibri/plugins/document_epub_render/assets/src/views/EpubRendererIndex.vue
@@ -474,11 +474,7 @@
             }
 
             // So we'll split the table into a bunch of smaller tables
-            const newTables = this.splitTable(height, table, tag =>
-              contents.document.createElement(tag)
-            );
-
-            newTables.forEach(newTable => {
+            this.splitTable(height, table).forEach(newTable => {
               table.parentElement.insertBefore(newTable, table);
 
               if (newTable.clientHeight < height) {
@@ -493,6 +489,7 @@
                 div.style.maxHeight = `${height - EPUB_VERTICAL_PADDING - diff}px`;
                 div.style.overflowY = 'auto';
 
+                // Move cell contents into div
                 while (cell.firstChild) {
                   div.append(cell.removeChild(cell.firstChild));
                 }
--- a/kolibri/plugins/document_epub_render/assets/src/views/EpubRendererIndex.vue
+++ b/kolibri/plugins/document_epub_render/assets/src/views/EpubRendererIndex.vue
@@ -180,7 +180,7 @@
   import SettingsButton from './SettingsButton';
   import SearchButton from './SearchButton';
 
-  import FixTableWrap from './FixTableWrap';
+  import TableWrapFactory from './TableWrap';
   import { THEMES } from './EpubConstants';
 
   const FONT_SIZE_MIN = 8;
@@ -458,7 +458,7 @@
         // Known issues: https://caniuse.com/#feat=multicolumn
         // Hook into content render, to check tables are breaking properly in column layout
         this.rendition.hooks.content.register(contents => {
-          FixTableWrap.doFix(contents.document);
+          TableWrapFactory.buildFixer(contents.document).fix();
         });
       });
       this.book.on(EVENTS.BOOK.OPEN_FAILED, () => {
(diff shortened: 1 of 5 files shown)


kolibri/plugins/document_epub_render/assets/src/views/EpubRendererIndex.vue
```diff
@@ -467,6 +520,57 @@
...
+            return newTable;
+          }
+
+          tableSet.forEach(row => {
+            tbody.appendChild(row);
+          });
+
+          return newTable;
+        });
+      },
```
indirectlylit: not critical, but might be nice to quarantine these two big code blocks in an external `firefoxTableHack.js` file or similar?

kolibri/plugins/document_epub_render/assets/src/views/EpubRendererIndex.vue
```diff
@@ -276,6 +277,7 @@
           h3: { ...colorStyle },
           h4: { ...colorStyle },
           h5: { ...colorStyle },
+          table: { width: '100%!important' },
```
indirectlylit, line 281: recommend adding a comment indicating why this was necessary for posterity

kolibri/plugins/document_epub_render/assets/src/views/EpubRendererIndex.vue
bjester: That's reasonable. They are large chunks of meaty code. I'll do that.

bjester:
@indirectlylit Updated!

jonboiser:
How does a table cell know to enable the scroll bar? At some font size settings (small to medium) the taller cells are scrollable, but for the larger font sizes, a cell will actually go outside of the viewport (including the bottom border) and will not be scrollable anymore (using e.g. the mouse scroller on the tall table will cause the area outside of the renderer to scroll). This is in non-full screen mode.

kolibri/plugins/document_epub_render/assets/src/views/TableWrap.js
```diff
@@ -0,0 +1,231 @@
...
+    });
+  }
+
+  /**
+   * Makes table "scrollable" by making the cells scrollable. Since the tables are
+   * split by rows, then this table has one row, so we'll scroll the cells.
+   *
+   * @param {Table} table
+   */
+  makeScrollable(table) {
```
jonboiser, line 92: This might need to be called every time the font-size is changed. I see that it's only called when the epub loads initially.

bjester: For me it appears to be called every time for font sizing increasing / decreasing.

bjester:
@jonboiser Unfortunately we can't rely on the browser (Firefox in this case) trigger scrolling based on size constraints and overflow rules. The main approach was to split the table into a set of smaller tables containing at least one row of the original. In the event the content in a single row's cells has a greater height of the viewable area, it will modify the cell contents to allow scrolling to happen.

I can see what you're saying, by increasing the font size, it can come to a point where the contents can no longer display in a reasonable manner. Although, for me, the cell scrolling appears to be doing its best handling the increase in content size.

kolibri/plugins/document_epub_render/assets/src/views/TableWrap.js
bjester: Hmm, maybe it isn't. Let me see.

bjester:
@jonboiser It was not re-rendering the epub contents as I thought when the font size changed. Unfortunately, supporting that outside of re-render means I should probably have it re-attach tables when decreasing the font size. I'm going to close this and come back to it.

PR #5438, opened 2019-04-28
Add directive dynamic argument support to no-unused-.. rules

MisRob:
### Summary

Add support for a[ new directive dynamic arguments syntax](https://vuejs.org/v2/guide/syntax.html#Dynamic-Arguments) to `vue-no-unused-properties` and `vue-no-unused-vuex-properties`.

@jonboiser This should solve the issue you reported.

### Reviewer guidance

Does it work well?

### References

None.

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/packages/eslint-plugin-kolibri/lib/rules/vue-no-unused-properties.js
+++ b/packages/eslint-plugin-kolibri/lib/rules/vue-no-unused-properties.js
@@ -71,6 +71,13 @@ const create = context => {
         });
       },
     },
+    utils.executeOnDirectiveDynamicArg(node => {
+      const argName = utils.getDirectiveDynamicArgName(node);
+
+      remove(unusedProperties, property => {
+        return argName === property.name;
+      });
+    }),
     utils.executeOnRootTemplateEnd(() => {
       if (unusedProperties.length) {
         utils.reportUnusedProperties(context, unusedProperties);
--- a/packages/eslint-plugin-kolibri/lib/rules/vue-no-unused-vuex-properties.js
+++ b/packages/eslint-plugin-kolibri/lib/rules/vue-no-unused-vuex-properties.js
@@ -140,6 +140,13 @@ const create = context => {
         });
       },
     },
+    utils.executeOnDirectiveDynamicArg(node => {
+      const argName = utils.getDirectiveDynamicArgName(node);
+
+      remove(unusedVuexProperties, property => {
+        return argName === property.name;
+      });
+    }),
     utils.executeOnRootTemplateEnd(() => {
       if (unusedVuexProperties.length) {
         utils.reportUnusedVuexProperties(context, unusedVuexProperties);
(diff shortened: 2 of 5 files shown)


rtibbles (dismissed):

rtibbles (approved):
I think this works to allow the use of these properties as the dynamic attribute name.

rtibbles:
I imagine the big issue with dynamic attribute names is that we can't know what prop, data, or vuex getter/state they are referring to until runtime?

So, our linting rule will throw an error on any of the properties that are defined to be referred to by this dynamic attribute, because they are never explicitly defined.

The test case we would want to be able to pass would be:
```
    {
      filename: 'test.vue',
      code: `
        <template>
          <a :[attributeName]="url"> ... </a>
        </template>
        <script>
          export default {
            props: ['attributeName'],
            data() {
               'href': 'http://kolibri.me',
               'src': 'http://kolibri.me/src.js',
            },
          };
        </script>
      `,
    },
```
But we can't make a linting rule that would handle this - so we will have to put eslint ignore comments around the dynamically selected data.

As we can't solve this except with ignore comments, I assume that was not the intent of this PR, and am merging!

MisRob:
🤔 I am not sure - I'm not clear on why this test should pass - I haven't used this syntax myself yet, but if I get the documentation right (do I? https://vuejs.org/v2/guide/syntax.html#Dynamic-Arguments), I think that in the scenario above, if `attributeName` is `"href"` for example, it will just cause that the template will be translated to

```
  <template>
    <a :href="url"> ... </a>
  </template>
```

and both `href` and `src` data fields should be still reported as unused data, is that right?

To actually use `href` from `data`, I would need to write

```
  <template>
    <a :href="href"> ... </a>
  </template>
```

or something like that.

Or have I misunderstood the issue that the example above demonstrates?

rtibbles:
My point was that the test above should fail, because we cannot know in advance which of the `data` attributes might get passed in as the value to the `attributeName` prop.

We can't make a linting rule that would allow this test to pass, so the only way that we could make this pass would be by doing this:
```
    {
      filename: 'test.vue',
      code: `
        <template>
          <a :[attributeName]="url"> ... </a>
        </template>
        <script>
          export default {
            props: ['attributeName'],
            data() {
               // eslint-disable vue-no-unused-properties
               'href': 'http://kolibri.me',
               'src': 'http://kolibri.me/src.js',
               // eslint-enable
            },
          };
        </script>
      `,
    },
```
I think this is the appropriate way to handle this, as dynamic attributes should be used sparingly, and with full awareness of which properties might actually end up being referenced by them. I just wanted to make sure we were all on the same page with respect to this!

MisRob:
> we cannot know in advance which of the data attributes might get passed in as the value to the attributeName prop

I still cannot wrap my head around why do we need to know this and why those `data` fields should be somehow related to `attributeName` property, it seems to be two completely different things to me. Obviously, it's Monday 😄 I'll have a look at the example one more time later.

Issue #5320, opened 2019-03-28
Apply consistent patterns for all modals


jonboiser:
### Observed behavior
This read-only [Google Doc](https://docs.google.com/document/d/1C2mMnCktUYgv2tb7pokRCLr-v2hM3Be9YBeNsOdZcck/edit?usp=sharing) describes how modals should work.

Not all of the modals work as prescribed. The Expected Behavior section tries to enumerates which modals don't conform and how.

Some of the prescriptions overlap with #5305 and #5281 

ChannelTokenModal seems to be the only modal with a form that follows all of the prescribed patterns.

### Expected behavior

**AssignmentDetailsModal (Lessons)**

* The submit and cancel buttons should disable when a request is pending.
* The "New lesson created" snackbar should still be visible after the user is redirected to the summary page. (It is shown briefly on the Lesson List page, and then disappears)

**~AssignmentChangeStatusModal~**

* The submit and cancel buttons should disable when a request is pending.
* The form should disable when a request is pending
* A notification should appear when the request fails.

**AssignmentCopyModal**
* The submit and cancel buttons should disable when a request is pending.
* A notification should appear when the request fails.
* When copying a lesson to a class results in a duplicated title, this should be handled gracefully by renaming it to make it unique. (Currently you get an AppError page)

**AssignmentDeleteModal**
* The submit and cancel buttons should disable when a request is pending.
* A notification should appear when the request fails (for  non 404 reasons).
* Deleting an assignment that has been deleted already should fail gracefully.

**CreateGroupModal**
* The form and cancel button should be disabled when a request is pending
* An error should appear in the form when a group name has already been used (but is only discovered after submitting the request)

**RenameGroupModal**
* The modal should not be immediately dismissed when submitted, but should remain on-screen with disabled form and buttons.
* An error should appear in the form when a group name has already been used (but is only discovered after submitting the request)

**DeleteGroupModal**
* The buttons should be disabled when the request is pending.

**CreateClassModal**

* The cancel button should disable  when the request is pending
* An error should appear in the form when a class name has already been used (but is only discovered after submitting the request)
* There should be a notification if the request suceeds

**AssignmentDetailsModal (Exams)**

* The 'status' section of the form and the buttons should disable when a  request is pending

**ClassDeleteModal**

* The buttons should disable when a request is pending
* A notification should appear if the request succeeds or fails

**UserRemoveConfirmationModal, RemoveFromGroupModal**

* The buttons should disable when a request is pending
[...]

PR #5439, opened 2019-04-29
Have KModal pass through their 'submit'  and 'cancel' events to simplify other modals

jonboiser:
### Summary

1. Changes KModal to have its parent emit 'submit' or 'cancel' events if these listeners have not been set on KModal. The point of this is to remove some boilerplate like `@submit="$emit('submit')` in the components that wrap the base KModal, and to reduce the amount variations on these event names (e.g. `@close`, `@success`, etc.) that appear in the code (in support of future work towards #5320). This makes the events of something like AssignmentCopyModal the same as the basic KModal.
1. For some modals, remove some coupling to Vuex (moving towards a full decoupling of modals from vuex so they are more like simple forms or input elements).
1. Wrap the text in KButton in a span, so it doesn't add extra whitespace when KButton is set to be basic-link
1. Fix a warning coming from CoreMenu

### Reviewer guidance
1. The simplest change is to delete lines like `@submit="$emit('submit')` from wrapped KModals, which should just work. The more complicated handlers like `@submit="handleSubmit` where handleSubmit may also emit or have complicated side-effects may take a little more checking.

### References

…

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated


kolibri/core/assets/src/views/KModal.vue
```diff
@@ -228,16 +228,26 @@
...
+      // Emitted when the submit button or the enter key is pressed
       emitSubmitEvent() {
         if (!this.submitDisabled) {
-          // Emitted when the submit button or the enter key is pressed
-          this.$emit('submit');
+          // If no @submit is set on KModal, have the parent emit the event
+          if (this.$listeners.submit) {
+            this.$emit('submit');
+          } else {
+            this.$parent.$emit('submit');
```
indirectlylit: I believe it should be the responsibility of the wrapper component to define this behavior because having a child modify the behavior of its parent goes against the typical component hierarchy model.

I haven't tried this before, but I think [you can use `v-on="$listeners"`](https://vuejs.org/v2/api/#vm-listeners) like this:

```html
<!-- AssignmentDeleteModal.vue -->
<template>
  <KModal
    :title="modalTitle"
    :submitText="coachStrings.$tr('deleteAction')"
    :cancelText="coachStrings.$tr('cancelAction')"
    v-on="$listeners"
  >
    <p>{{ modalDescription }}</p>
    <p v-if="modalConfirmation">
      {{ modalConfirmation }}
    </p>
  </KModal>
</template>
```

Admittedly it only cuts down one line of boilerplate rather than two, but I think the trade-off for explicitness and parent component control is worth it.

A similar strategy can be used for creating custom wrappers around native HTML elements like `<input>`. 

ref: https://github.com/vuejs/vue/issues/5983

jonboiser: Using `v-on=$listeners` seems to work, but is kind of odd since instead of the Wrapper's child (KModal) defining the event passthrough behavior it's now the parent (of the wrapper) defining the passthrough behavior, namely by getting the event names right.

```
<ModalUser>
    <!-- "submit" will go through, but not "cancel", since the typo -->
    <!-- In either case, ModalUser component will not work -->
    <WrappedModal @submit="handleSubmit" @carncel="handleCancel" />
        <KModal v-on="$listeners />
```

Another thing I liked about the `$parent.$emit` hack is that you don't get duplicated `$emit` events in the devtools

jonboiser: Okay, I can try using `v-on=$listeners` in this refactor, which gives us, in my opinion, three benefits

1. Introduces an interesting new syntax into the codebase that might useful for other things.
1. Still gets rid of some boilerplate
1. Still encourages the consistent use of the events "submit" and "cancel" for all Modal-type components.

indirectlylit (approved):
looks good to me, thank you!

kolibri/core/assets/src/views/UpdateNotification.vue
```diff
@@ -84,7 +85,7 @@
           this.saveDismissedNotification(this.id);
         }
         this.removeNotification(this.id);
-        this.$emit('closeModal');
+        this.$emit('submit');
```
indirectlylit, line 87: is it necessary to emit anything here? I think the `v-on` will do this for you?

kolibri/plugins/coach/assets/src/views/plan/GroupsPage/CreateGroupModal.vue
```diff
@@ -82,21 +82,18 @@
...
     methods: {
-      ...mapActions('groups', ['displayModal', 'createGroup']),
+      ...mapActions('groups', ['createGroup']),
       callCreateGroup() {
         this.formSubmitted = true;
         if (this.formIsValid) {
           this.submitting = true;
           this.createGroup({ groupName: this.name, classId: this.classId }).then(() => {
-            this.$emit('success');
+            this.$emit('submit');
```
indirectlylit, line 91: won't this emit submit twice? Once for `v-on="$listeners"` and once here?

indirectlylit (commented):
actually - question:

It looks like in a number of places we might be emitting events twice? E.g. once immediately due to `v-on...`, and again inside a custom event handler?

kolibri/plugins/coach/assets/src/views/plan/GroupsPage/CreateGroupModal.vue
jonboiser: CreateGroupModal only emits 'submit' once. I think if you explicitly set a `@submit=customListener`, it sort of overwrites the `{submit: parentListener}` hidden in `v-on="$listeners`

jonboiser: Interestingly, though, if you remove this `$emit`, the event from CreateGroupModal does not appear in the devtools, but it seems that it still goes through the GroupsPage, which is not really what I want since i want to wait till the promise is resolved before emitting 'submit'.

I'm just going to remove the `v-on="listeners"` for the handful of modals that might emit the same event asynchronously.

In the future, I would like to to remove async behavior API calls from modals, and delegate them to the parents of modals. So modals just synchronously emit events and data, basically.

indirectlylit: yeah that seems reasonable.

another alternative is for modals that handle their own internal async code, to explicitly emit an event that's _not_ `submit` or `cancel`. For example:

```javascript
.then(() => this.$emit('finished'))
```

jonboiser:
I've changed this so that every KModal simply re-emits "cancel" and/or "submit". Also added a warning on KModal in case someone shows a button, but doesnt actually register a listener for the button click.

indirectlylit (approved):

Issue #5152, opened 2019-02-25
Management command for registering facility to data portal 


ralphiee22:

indirectlylit:
@ralphiee22 mind flagging this issue in the PR which addressed it?

Issue #5151, opened 2019-02-25
Update Kolibri management command to do Data Portal syncing by default (with inverse CSR)


ralphiee22:

indirectlylit:
@ralphiee22 mind flagging this issue in the PR which addressed it?

PR #5442, opened 2019-04-30
Add mgmnt command for claiming a facility on data portal

ralphiee22:
### Summary

Adds management command for claiming a facility on the data portal server. Handles appropriate error checking.

### Reviewer guidance

Are the error messages sufficient?

Test it against the data portal PR 

### References

https://github.com/learningequality/kolibri-data-portal/pull/6
Addresses https://github.com/learningequality/kolibri/issues/5152 && https://github.com/learningequality/kolibri/issues/5151.

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated


kolibri/core/auth/management/commands/claim.py
```diff
@@ -0,0 +1,56 @@
+from django.core.management.base import BaseCommand
```
rtibbles: Might want to call the management command `register`?

kolibri/utils/options.py
```diff
@@ -136,6 +136,13 @@ def calculate_thread_pool():
                 "KOLIBRI_CENTRAL_CONTENT_BASE_URL",
                 "CENTRAL_CONTENT_DOWNLOAD_BASE_URL",
             ),
+        },
+        "DATA_PORTAL_SYNCING_BASE_URL": {
```
rtibbles, line 140: What is the use case for making this customizable? We want to allow syncing against arbitrary Kolibri devices, but there is only one data portal.

rtibbles: If it is just for testing purposes, I'd rather we just change it adhoc.

kolibri/core/utils/portal.py
```diff
@@ -0,0 +1,52 @@
...
+            'certificate_chain': json.dumps(CertificateSerializer(client_cert.get_ancestors(include_self=True), many=True).data),
+            'token': token,
+        }
+
+        # sign the client/server nonce combo to attach to the request
+        message = "{client_nonce}:{server_nonce}".format(**data)
+        data["signature"] = client_cert.sign(message)
+
+        # attempt to claim the facility
+        response = requests.post(urljoin(PORTAL_URL, 'portal/api/claim/'), data=data)
```
rtibbles: This URL seems to be wrong.

kolibri/utils/options.py
ralphiee22: I am imagining the case where we set up custom data portal servers for specific partners. 
But also, it is helpful for testing locally.

kolibri/core/auth/management/commands/claim.py
ralphiee22: Can change all instances of claim to register.

kolibri/core/auth/management/commands/register.py
```diff
@@ -0,0 +1,92 @@
...
+                "This device does not own a certificate for Facility: {}".format(
+                    facility.name
+                )
+            )
+        # an invalid nonce/claim response
+        except exceptions.HTTPError as e:
+            # handle 404 responses from claim endpoint
+            if e.response.status_code == 404:
+                # if the facility does not exist on data portal, try syncing and retry registering
+                if "Facility" in e.response.text:
```
rtibbles: Something about this feels brittle - but at the same time, could the endpoint ever return a 404 for any other reason?

indirectlylit: maybe JSON-formatted errors with error codes?

jamalex: > maybe JSON-formatted errors with error codes?

That was the suggestion I made at https://github.com/learningequality/kolibri-data-portal/pull/6#discussion_r281357052 (because yes, there are already two different ways it can 404 -- missing Facility, or missing LearningCommunity).

kolibri/core/auth/management/commands/register.py
```diff
@@ -0,0 +1,92 @@
...
+from morango.certificates import Certificate
+from requests import exceptions
+
+from kolibri.core.auth.utils import confirm_or_exit
+from kolibri.core.auth.utils import get_facility
+from kolibri.core.utils.portal import claim
+
+
+class Command(BaseCommand):
+    help = "Allows registering of facilities which have been synced to a portal server."
```
jamalex: which -> that

jamalex: In terms of naming, we use "register" here, but "claim" on KDP. For consistency and clarity, perhaps unifying them? I agree that `claim` sounds a bit funny on the client side, so perhaps sticking with `register` and updating the KDP server naming accordingly? And maybe being more explicit with the name too, e.g. `registerfacility`?

kolibri/core/auth/utils.py
```diff
@@ -10,3 +13,41 @@ def confirm_or_exit(message):
...
+            raise CommandError("Facility with ID {} does not exist".format(facility_id))
+    # if no id passed in, assume only one facility on device
+    else:
+        try:
+            facility = Facility.objects.get()
+        except Facility.DoesNotExist:
+            raise CommandError(
+                (
+                    "There are no facilities on this device. "
+                    "Please setup your kolibri instance."
```
jamalex: Perhaps being clearer in how to do so -- "Please initialize your Kolibri installation by starting the server, loading it in the browser, and following the instructions" or something.

kolibri/core/auth/utils.py
```diff
@@ -10,3 +13,41 @@ def confirm_or_exit(message):
...
+                )
+            )
+        except Facility.MultipleObjectsReturned:
+            if noninteractive:
+                raise CommandError(
+                    (
+                        "There are multiple facilities on this device. "
+                        "Please pass in a facility ID."
```
jamalex: Provide instructions on how to do so (mention the `--facility` argument).

kolibri/core/auth/utils.py
```diff
@@ -10,3 +13,41 @@ def confirm_or_exit(message):
     if answer != "yes":
         print("Canceled! Exiting without touching the database.")
         sys.exit(1)
+
+
+def get_facility(facility_id=None, noninteractive=False):
```
jamalex: Given it's specifically a helper function for management commands (`noninteractive`, `CommandError`, etc), perhaps it would belong better in something like `kolibri/core/auth/management/commands/utils.py`? Or if it's only used by the one management command, for now it could just live inside the command file directly?

kolibri/core/auth/management/commands/register.py
```diff
@@ -0,0 +1,92 @@
...
+                            "Facility: {} has been synced. Would you like to retry registering?".format(
+                                facility.name
+                            )
+                        )
+                        claim(token, facility)
+                        self.stdout.write(
+                            "Facility: {} has been successfully registered.".format(
+                                facility.name
+                            )
+                        )
```
jamalex: This is a dupe of lines 32-35; perhaps put them in a helper method on the `Command`?

kolibri/core/auth/management/commands/sync.py
```diff
@@ -0,0 +1,67 @@
...
+            sync_client = network_connection.create_sync_session(
+                client_cert, server_cert
+            )
+            progress_update(1)
+
+            sync_client.initiate_push(Filter(scope_params["dataset_id"]))
+            progress_update(1)
+
+            sync_client.close_sync_session()
+            progress_update(1)
```
jamalex, line 96: Let's chat about how we can unify this management command in some form with `fullfacilitysync` or whatever the older one is called.

kolibri/core/auth/utils.py
```diff
@@ -10,3 +13,41 @@ def confirm_or_exit(message):
...
+                        "Please pass in a facility ID."
+                    )
+                )
+            else:
+                # in interactive mode, allow user to select facility
+                facilities = Facility.objects.all()
+                message = "Please choose a facility to sync:\n"
+                for idx, facility in enumerate(facilities):
+                    message += "{}. {}\n".format(idx + 1, facility.name)
+                idx = input(message)
```
jamalex: Assert or validation of some sort that it's in range, or will the natural exception raised be meaningful enough, do you think?

kolibri/core/utils/portal.py
```diff
@@ -0,0 +1,63 @@
...
+
+def claim(token, facility, client_ip=None):
+
+    # get client ip if not passed in
+    if not client_ip:
+        hostname = socket.gethostname()
+        client_ip = socket.gethostbyname(hostname)
+
+    # request the server for a one-time-use nonce
+    PORTAL_URL = conf.OPTIONS["Urls"]["DATA_PORTAL_SYNCING_BASE_URL"]
```
jamalex, line 16: Perhaps better to have this be the default, but allow it to be overridden as an argument (also in the management command).

kolibri/core/utils/portal.py
```diff
@@ -0,0 +1,63 @@
...
+from kolibri.core.auth.constants.morango_scope_definitions import FULL_FACILITY
+from kolibri.utils import conf
+
+
+def claim(token, facility, client_ip=None):
+
+    # get client ip if not passed in
+    if not client_ip:
+        hostname = socket.gethostname()
+        client_ip = socket.gethostbyname(hostname)
```
jamalex: I feel like I've hit situations where these can throw exceptions, but I can't remember. Also, do we really need the IP here? I feel like it's mostly used with Nonces on the server end, to make sure the client using the nonce is the one that created it. Could we allow `None` or put in a dummy value for this case?

kolibri/core/utils/portal.py
```diff
@@ -0,0 +1,63 @@
...
+        "token": token,
+    }
+
+    # sign the client/server nonce combo to attach to the request
+    message = "{client_nonce}:{server_nonce}".format(**data)
+    data["signature"] = client_cert.sign(message)
+
+    # attempt to claim the facility
+    response = requests.post(
+        urljoin(PORTAL_URL, "portal/api/public/v1/claim/"), data=data
```
jamalex: Should we put that URL path in a constant, as has been the general pattern?

jamalex (changes requested):
Nice! A few suggestions, and one place to talk through

kolibri/core/auth/management/commands/sync.py
ralphiee22: CC: @rtibbles 
We talked about keeping the peer to peer syncing command separate from the data portal syncing command.

kolibri/core/auth/utils.py
ralphiee22: I thought it awkward to live there because it can then be invoked with `manage utils`, even though it would fail.
But if its undocumented, I suppose it could live there?

kolibri/core/utils/portal.py
ralphiee22: perhaps if we use it in more than one place, but I think its fine since this will probably the one place we will be using it.

kolibri/core/utils/portal.py
ralphiee22: so in the case there are more data portal servers?

kolibri/core/auth/utils.py
jamalex: Discussed in person: put in management/utils.py

kolibri/core/utils/portal.py
jamalex: Discussed: In the case of other portal servers, use envvar

kolibri/core/utils/portal.py
jamalex: Discussed: remove client nonce entirely (it's just one-way proving for this command)

jamalex (dismissed):
Couple final things, then good to go!

rtibbles (approved):

PR #5445, opened 2019-05-01
Node.js dependency and tools setup and validation

bjester:
### Summary

- Add `nodeenv` as dev python dependency
- Update documentation to instruct developers to use `nodeenv` for Node.js and Yarn setup
- Updated `engines` in root `package.json` to have `yarn` instead of `npm`
- Replace existing engines validator with one that works with yarn
- Add `@types/jest` for helpful IDE support in JS unit tests

Diff `requirements/dev.txt`:
```
+nodeenv==1.3.3
```

### Reviewer guidance

- Run through clean set up of Node.js and verify documentation instructions and tools work appropriately
- Modify `engines` in `package.json` and verify they are strictly validated 

### References

- Refer to Dr. Tibbles

----

### Contributor Checklist

PR process:

- [ ] ~If this is an important user-facing change, PR or related issue has a 'changelog' label~

Testing:

- [ ] ~If there are any front-end changes, before/after screenshots are included~
- [ ] ~Critical user journeys are covered by Gherkin stories~

### Reviewer Checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/package.json
+++ b/package.json
@@ -48,6 +48,7 @@
   "private": true,
   "dependencies": {},
   "devDependencies": {
+    "@types/jest": "^24.0.12",
     "black-fmt": "https://github.com/learningequality/black-fmt#v0.1.3",
     "kolibri-tools": "0.12.0-beta.3.2",
     "yarn-run-all": "^3.1.1"
@@ -55,8 +56,9 @@
   "optionalDependencies": {
     "fsevents": "*"
   },
+  "engineStrict": true,
   "engines": {
     "node": "10.x",
-    "npm": ">= 3"
+    "yarn": ">= 1.12.3"
   }
 }
--- a/packages/kolibri-tools/lib/cli.js
+++ b/packages/kolibri-tools/lib/cli.js
@@ -1,15 +1,14 @@
 #!/usr/bin/env node
 
+require('colors');
 const os = require('os');
 const path = require('path');
 const program = require('commander');
-const version = require('../package.json').version;
+const checkVersion = require('check-node-version');
+const version = require('../package.json');
 const logger = require('./logging');
-const readWebpackJson = require('./read_webpack_json');
 
-// ensure the correct version of node is being used
-// (specified in package.json)
-require('engine-strict').check();
+const readWebpackJson = require('./read_webpack_json');
 
 const cliLogging = logger.getLogger('Kolibri CLI');
 
@@ -331,4 +330,28 @@ program
     require('jest-cli/build/cli').run();
   });
 
-program.parse(process.argv);
+// Check engines, then process args
+const engines = require(path.resolve(__dirname, '../../../package.json')).engines;
+checkVersion(engines, (err, results) => {
+  if (err) {
+    console.error(err);
+    process.exit(1);
+  }
+
+  if (results.isSatisfied) {
+    program.parse(process.argv);
+    return;
+  }
+
+  for (const packageName of Object.keys(results.versions)) {
+    if (!results.versions[packageName].isSatisfied) {
+      let required = engines[packageName];
+      console.error(`\nERROR:`.red, `Incorrect version of ${packageName}.`);
+      console.error(`ERROR:`.red, `${packageName} ${required} is required.`);
+    }
+  }
+
+  console.log(``);
+
+  process.exit(1);
+});
--- a/packages/kolibri-tools/lib/cli.js
+++ b/packages/kolibri-tools/lib/cli.js
@@ -1,6 +1,5 @@
 #!/usr/bin/env node
 
-require('colors');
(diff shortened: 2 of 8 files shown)


packages/kolibri-tools/package.json
```diff
@@ -22,6 +22,7 @@
     "babel-preset-env": "^1.7.0",
     "buble": "0.19.3",
     "buble-loader": "0.5.0",
+    "check-node-version": "^3.3.0",
```
rtibbles, line 25: Can probably remove `engine-strict` below, as it seems this is taking its place?

package.json
```diff
@@ -48,15 +48,17 @@
   "private": true,
   "dependencies": {},
   "devDependencies": {
+    "@types/jest": "^24.0.12",
```
rtibbles, line 51: Should this be in here, or in the kolibri-tools package.json?

rtibbles (changes requested):
1 (possibly 2) things that I think can be cleaned up. Have not yet manually tested.

packages/kolibri-tools/lib/cli.js
```diff
@@ -331,4 +330,28 @@ program
     require('jest-cli/build/cli').run();
   });
 
-program.parse(process.argv);
+// Check engines, then process args
+const engines = require(path.resolve(__dirname, '../../../package.json')).engines;
+checkVersion(engines, (err, results) => {
+  if (err) {
+    console.error(err);
```
rtibbles: Oh, and use the logger module imported above, that will do standard colour coding for errors for you, and will also mean you don't get `no-console` linting breakages.

`cliLogging.error` will apply the right colouring without needing the `colors` require at the top.

packages/kolibri-tools/lib/cli.js
```diff
@@ -1,15 +1,14 @@
 #!/usr/bin/env node
 
+require('colors');
```
rtibbles: Can drop this and just use the `cliLogging.error` function in place of `console.error` with explicit colouring applied.

packages/kolibri-tools/package.json
bjester: Ah yes, I totally I removed it, but I think I reverted the change when I was trying different replacements.

package.json
bjester: Yes, it needs to be in the project `package.json`. It's only metadata exposing what `jest` provides. Having `jest` encapsulated within kolibri-tools is the reason I've added this.

rtibbles: OK - is it an IDE plugin that is using this metadata? (as we are not using typescript anywhere in the project)

bjester: I think most code-hinting uses TypeScript, even for VSCode. https://github.com/felixfbecker/vscode-php-intellisense

rtibbles: Sublime has not complained hitherto at the lack, so I hadn't noticed!

packages/kolibri-tools/lib/cli.js
bjester: If it's okay with you, I would like to align the logger more closely to how the `console` works. Some proposed changes:
- Allow logger methods to accept multiple messages as different arguments, as does `console`
- Add `.format()` method which will help facilitate color formatting and prefixing
- Use corresponding `console` methods for each, so that messages are output to stdout and stderr respectively; `.error()` and `.warn()` should go to stderr, and the rest should go to stdout.

rtibbles: Yep - fine by me! You may be the first person apart from me to ever look at the file, so it has not received much TLC.

packages/kolibri-tools/lib/logging.js
```diff
@@ -4,38 +4,42 @@
...
-    if (color && prefixed) {
-      console.log(color(this.prefix + msg));
-    } else if (color) {
-      console.log(color(msg));
-    } else {
-      console.log(msg);
-    }
-    /* eslint-enable no-console */
+  write(level, messages, color, prefix = this.prefix) {
+    messages = prefix ? messages.map(msg => this.prefix + msg) : messages;
```
rtibbles: I don't think this is quite intended behaviour - as passing multiple args to a console function prints it all on the same line, I think it only needs one prefix? So perhaps it would be better as:
`messages = prefix ? [prefix].concat(messages) : messages;`

bjester: Yep, whoops.

rtibbles: (which probably also means we can get rid of the space at the end of the prefixes!)

bjester: That's better.
logger

rtibbles (approved):
WFM

PR #5448, opened 2019-05-02
updates to learner quiz view

### LINKED ISSUE #5440: Submit button on top right corner of a quiz
**Observed behavior**
The students sometimes directly press the Submit button on the top right corner of any quiz without even start or just after attempting the first question. This inturn directly completes the quiz.

**Expected behavior**
The Submit button appears in the last page of the quiz so that students don't press it by mistake. On the last page of the quiz we can have the submit button in the place of next button.

**User-facing consequences**
The exams/quizzes get submitted even before they are attempted or completed. This in turn affects the teacher who had prepared and planned for the quiz.

**Steps to reproduce**
Create a quize and check using any tablet.

**Context**
Kolibri version : Kolibri 0.12.2
Operating system : ubuntu 14.04
Browser : Chrome
Screenshot_20190430-153902

indirectlylit:
### Summary

Changes based in part on discussion with the team at design office hours and Ana's design work.

fixes #5440

* adds floating bottom bar
* adds 'Question X of Y' header above exercise
* 'next' and 'previous' button updates
    * remove the word 'question'
    * make them both primary
* 'submit' button updates
    * moves it next to the 'X of Y answered' text
    * makes it a flat secondary button
* layout updates
    * increases screen real estate on larger screens
    * removes question list on smaller screens
    * on mobile screens, 'submit' button moves to inner scrolling pane

**Old view:**

image

**New desktop view:**

desktop

**New tablet view:**

tablet

**New mobile view:**

mobile

### Reviewer guidance

there are also some changes in https://github.com/learningequality/kolibri-exercise-perseus-plugin/pull/188 which this PR leverages in terms of layout cleanup

### References

https://github.com/learningequality/clearinghouse/issues/291

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated


radinamatic:
Looks good... visually. My only concern would be the experience of users who navigate by keyboard, as having the `Submit quiz` button _in between_ the `Next` and `Previous` does not seem optimal.

Even in terms of tasks to be performed, submitting comes _after_ navigating between and answering questions, shouldn't the `Submit quiz` button be located "after" the `Next` and `Previous` ones, not only in the DOM that determines the <kbd>Tab</kbd> order, but visually too?

indirectlylit:
thanks, good feedback!

miyer03:
Few thoughts on this 

1. If the submit button could appear on the last page of the quiz, then there will not be any case when a student, presses the Submit button by mistake in between a quiz.

2. In case we can't go by the above suggestion, would suggest to have some symbols for previous, next and sumbit along with the text <-- for previous  --> for next . Also, agree with the previous suggestion that Submit should appear after next.

khangmach:
slice

based on the feedback I'm seeing here:
- Incorporate arrows into the buttons to show directionality
- Group together like-actions (in this case the navigational actions of previous/next)
- In smaller breakpoints, only show the arrow buttons to save space

miyer03:
@khangmach the above option look good.

indirectlylit:
I ran into two issues with the proposal above:

1. Does not work for smaller breakpoints: 'x of y answered' text and 'submit quiz' button are large enough to dominate the footer
2. Inconsistent with other full-screen footers in the app, which all have primary buttons in the bottom-right

I've tried to work around these while still ensuring semantically-correct a11y ordering for both DOM order (screen readers) and keyboard tab order:

1. Next
2. Previous
3. Submit

Let me know if you think this satisfies all the constraints, including the original primary goal of emphasizing the next/prev functionality over submission.

Updated changes below, with semantic order of buttons highlighted:

### Desktop view

> image

### Desktop view (RTL)

> image

### Tablet view

> image

### Mobile view

> image

khangmach (approved):
Works for me. Nice call hiding the x of y text on smallest breakpoint

radinamatic:
Maybe we don't need to address it right now, but I still find awkward that in the tablet view `SUBMIT` is _before_ `Next` and `Previous`, even visually... :confused:

radinamatic:
What happens when the user tabs between answers, and then into `Next` or `Previous`, is the `SUMBIT` skipped every time?

indirectlylit:
> What happens when the user tabs between answers, and then into Next or Previous, is the SUMBIT skipped every time?

Submit is initially skipped and then gets focus after 'Previous'

radinamatic:
So if I understand correctly, user tabs through the answers, tabs to `Next`, but if they want to go back and correct, they will tab-back to `Previous`, and eventually run into `SUBMIT`, right?

jtamiace (approved):
Looks good to me as well. 

Any thoughts from you guys on switching the order of "SUBMIT QUIZ" and "X of Y questions answered"? I don't know why it feels a bit strange for the action to be floating in space like that

indirectlylit:
> Any thoughts from you guys on switching the order of "SUBMIT QUIZ" and "X of Y questions answered"? I don't know why it feels a bit strange for the action to be floating in space like that

I actually had it like that originally, and it looked even weirder because the button has inner padding which makes the whole thing become visually indented, especially in this view:

image

indirectlylit:
merging in. I'm sure there are improvements that can be made, so let's tackle that in subsequent updates

nucleogenesis:
I have gone through and tested on several tablet/mobile views in BrowserStack and it looks and feels great. 

One thought I had is that as a mobile (small screen) user that I'd have the ability to select a specific question to view like you can in the larger views.

Perhaps the "Question X of Y" portion could be in its own section of the page above the portion that displays the exercise. Then that could display "Question X of Y ↓" and when clicked, display a dropdown list of all questions by number to allow the user to select a specific question.

indirectlylit:
> Perhaps the "Question X of Y" portion could be in its own section of the page above the portion that displays the exercise. Then that could display "Question X of Y ↓" and when clicked, display a dropdown list of all questions by number to allow the user to select a specific question.

that's a great idea. I believe some version of that is being considered for a future iteration - @jtamiace @khangmach this is like what Ana is working on, no?

jtamiace:
Yep she has several different design ideas exploring the show/hide question list approach

Issue #5403, opened 2019-04-22
Add group-level reports and preview button to exercise/resource report pages


jonboiser:
### Observed behavior

Currently, ReportsLesson(Resource|Exercise)LearnerListPage.vue does not have

1. an option to toggle an alternative view that groups learners and presents group-level summary statistics. 
1. a "preview" button that presents the resource or exercise in a way similar to the  exam or lesson creation workflow

### Expected behavior

A way to toggle group reports and preview items is available, along the lines of these designs:

**For ReportsLessonExerciseLearnerListPage**

![Exercise report - groups view (alt)](https://user-images.githubusercontent.com/10248067/56514183-800cf080-64e9-11e9-811f-ced7abadda39.png)

**For ReportsLessonResourceLearnerListPage**
![audio report - learners view](https://user-images.githubusercontent.com/10248067/56514315-cd895d80-64e9-11e9-8523-609f3019787e.png)

Other points:

1. This functionality will account for learners who are in more than one group. For example, if a learner "Pamela" is in two groups "A" and "B", then Pamela's row will appear in both grouped tables, and her progress will be incorporated into the summary statistics.
1. For the Resource version of the report, data for "view count" (to calculate the "Average number of views" statistic) may not be available from the backend
1. Do not add change the table header seen in the ReportsLessonExerciseLearnerListPage mockup to "Answer streak". Keep its current header. 
1. Only add the "Preview" button to the two reports pages discussed above. For example, do not add them to the ReportsGroupReportLessonResourceLearnerListPage.
1. To implement the "Preview" button, it might help to use the strategy used in #5417, where you link to the `RESOURCE_CONTENT_PREVIEW` route and provide extra data in the URL to get the correct exiting behavior (e.g. clicking the "X" button on on the preview page). See https://github.com/learningequality/kolibri/pull/5417/files#diff-59c8f3c9df19d72f38d1c9889771846fL108 for details.
1. The widget with circular icons and counts (e.g. yellow star, blue clock, red exclamation, blank circle) is a new element. The counts mean
    * yellow star - The number of learners in the group who have completed the exercise or resource
    * blue clock - The number (...) who have started the exercise/resource but not finished it
    * red exclamation - The number (...) who "need help" on the exercise (not applicable to non-exercise resources)
    * blank circle - the number (...) who do not fall into the other categories, i.e. the rest.

### User-facing consequences

…

### Errors and logs

…

### Steps to reproduce

…

### Context

…

[...]

PR #5453, opened 2019-05-05
Add group-level reports for an exercise/resource learners

MisRob:
### Summary

Add "View by groups" feature to exercise learners list reports

#### Screenshots

Screenshot 2019-05-05 at 22 04 32
Screenshot 2019-05-05 at 22 04 40

#### Design files

- #5403 screens
- https://www.figma.com/file/D5wXhmWNGyt60SpGBgFucwZU/Coach-reports-0.12?node-id=1385%3A4822I'

#### Notes

- We'll need to discuss if/how to visually treat edge cases like empty groups (only non-empty ones are displayed now) and learners without groups (they're ignored now) - I'm going to ask design folks tomorrow on Slack and eventually update the PR/open a follow-up based on the result.
- At first I tried to add a couple of basic `ReportsLessonExerciseLearnerListPage.vue` specs related to the new feature, but got stuck on router mocks for quite a long time - I tried various ways but I always broke something else. Not sure why yet, maybe I overlooked something. After a quite long time, I gave it up for now, but I kept my test scenarios and will add them as soon as I have a look at how to simply test these views (so far it looked that `commonCoach` mixin's dependencies could cause some of those troubles so I would like to have a closer look).
- It was a bit difficult to track all dependencies to store etc. with `commonCoach` mixin when navigating in coach codes. I noticed that it's being used also for smaller components, but tried to avoid that in a new table component. I was thinking about rather splitting the mixin into helpers and smaller pieces in the future (even when it would cost us some duplicates in components/views)? Would it make sense?
- It's my first feature so would like to ask - when will be the right time to add gherkin stories? After the review?

### Reviewer guidance

1. Create a couple of groups and assign learners
2. Create a lesson containing an exercise
3. Login as some learners of groups you created and try to start/finish the exercise or stop in the middle of it
4. Visit Coach -> Reports -> Lessons -> your lesson with an exercise -> Report -> your exercise (`coach/#/<classId>/reports/lessons/<lessonId>/exercises/<exerciseId>/learners`)

- Toggle "View by groups"
- Check groups and their stats
- Check navigation between an exercise learners list and their detail views

### References

Solves first part of #5403

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/plugins/coach/assets/src/views/reports/ReportsExerciseLearners.vue
+++ b/kolibri/plugins/coach/assets/src/views/reports/ReportsExerciseLearners.vue
@@ -18,7 +18,7 @@
           <KRouterLink
             v-if="showLink(entry)"
             :text="entry.name"
-            :to="entry.exerciseLink"
+            :to="entry.exerciseLearnerLink"
           />
           <template v-else>
             {{ entry.name }}
--- a/kolibri/plugins/coach/assets/src/views/reports/ReportsExerciseLearners.vue
+++ b/kolibri/plugins/coach/assets/src/views/reports/ReportsExerciseLearners.vue
@@ -33,7 +33,7 @@
           />
         </td>
         <td v-if="showGroupsColumn">
-          <TruncatedItemList :items="entry.groups" />
+          <TruncatedItemList :items="getGroupNames(entry)" />
         </td>
         <td>
           <ElapsedTime
@@ -101,6 +101,13 @@
         }
 
         return null;
+      },
+      getGroupNames(entry) {
+        if (!entry || !entry.groups) {
+          return [];
+        }
+
+        return entry.groups.map(group => group.name);
       },
     },
   };
--- a/kolibri/plugins/coach/assets/src/views/reports/ReportsExerciseLearners.vue
+++ b/kolibri/plugins/coach/assets/src/views/reports/ReportsExerciseLearners.vue
@@ -13,7 +13,7 @@
       </tr>
     </thead>
     <transition-group slot="tbody" tag="tbody" name="list">
-      <tr v-for="entry in entries" ref="entry" :key="entry.id">
+      <tr v-for="entry in entries" :key="entry.id" data-test="entry">
         <td>
           <KRouterLink
             v-if="showLink(entry)"
--- a/kolibri/plugins/coach/assets/src/views/reports/ReportsExerciseLearners.vue
+++ b/kolibri/plugins/coach/assets/src/views/reports/ReportsExerciseLearners.vue
@@ -1,0 +1,108 @@
+<template>
+
+  <CoreTable :emptyMessage="coachStrings.$tr('activityListEmptyState')">
+    <thead slot="thead">
+      <tr>
+        <th>{{ coachStrings.$tr('nameLabel') }}</th>
+        <th>{{ coachStrings.$tr('progressLabel') }}</th>
+        <th>{{ coachStrings.$tr('timeSpentLabel') }}</th>
+        <th v-if="showGroupsColumn">
+          {{ coachStrings.$tr('groupsLabel') }}
+        </th>
+        <th>{{ coachStrings.$tr('lastActivityLabel') }}</th>
+      </tr>
+    </thead>
+    <transition-group slot="tbody" tag="tbody" name="list">
+      <tr v-for="entry in entries" ref="entry" :key="entry.id">
+        <td>
+          <KRouterLink
+            v-if="showLink(entry)"
+            :text="entry.name"
+            :to="entry.exerciseLink"
+          />
+          <template v-else>
+            {{ entry.name }}
+          </template>
+        </td>
+        <td>
+          <StatusSimple :status="entry.statusObj.status" />
+        </td>
+        <td>
+          <TimeDuration
+            :seconds="timeDuration(entry)"
+          />
+        </td>
+        <td v-if="showGroupsColumn">
+          <TruncatedItemList :items="entry.groups" />
+        </td>
+        <td>
+          <ElapsedTime
+            :date="elapsedTime(entry)"
+          />
+        </td>
+      </tr>
+    </transition-group>
+  </CoreTable>
+
+</template>
+
+
+<script>
+
+  import CoreTable from 'kolibri.coreVue.components.CoreTable';
+  import ElapsedTime from 'kolibri.coreVue.components.ElapsedTime';
+  import KRouterLink from 'kolibri.coreVue.components.KRouterLink';
+
+  import { coachStringsMixin } from '../common/commonCoachStrings';
+  import StatusSimple from '../common/status/StatusSimple';
+  import TimeDuration from '../common/TimeDuration';
+  import TruncatedItemList from '../common/TruncatedItemList';
+  import { STATUSES } from '../../modules/classSummary/constants';
+
+  export default {
+    name: 'ReportsExerciseLearners',
+    components: {
+      CoreTable,
+      ElapsedTime,
+      KRouterLink,
+      StatusSimple,
+      TimeDuration,
+      TruncatedItemList,
+    },
+    mixins: [coachStringsMixin],
+    props: {
+      entries: {
+        type: Array,
+      },
+      showGroupsColumn: {
+        type: Boolean,
+        default: true,
+      },
+    },
+    data() {
+      return {
+        STATUSES,
+      };
+    },
+    methods: {
+      showLink(entry) {
+        return entry.statusObj.status !== this.STATUSES.notStarted;
+      },
+      timeDuration(entry) {
+        if (entry.statusObj.status !== this.STATUSES.notStarted) {
+          return entry.statusObj.time_spent;
+        }
+
+        return null;
+      },
+      elapsedTime(entry) {
+        if (entry.statusObj.status !== this.STATUSES.notStarted) {
+          return entry.statusObj.last_activity;
+        }
+
+        return null;
+      },
+    },
+  };
+
+</script>
--- a/kolibri/plugins/coach/assets/src/views/reports/ReportsLessonExerciseLearnerListPage.vue
+++ b/kolibri/plugins/coach/assets/src/views/reports/ReportsLessonExerciseLearnerListPage.vue
@@ -37,7 +37,7 @@
           </p>
 
           <ReportsExerciseLearners
-            :entries="getTableEntries([group.id])"
+            :entries="getGroupEntries(group.id)"
             :showGroupsColumn="false"
           />
         </div>
@@ -48,7 +48,7 @@
           <StatusSummary :tally="summaryTally" />
         </p>
 
-        <ReportsExerciseLearners :entries="getTableEntries()" />
+        <ReportsExerciseLearners :entries="allEntries" />
       </div>
     </KPageContainer>
   </CoreBase>
@@ -81,8 +81,7 @@
         return this.lessonMap[this.$route.params.lessonId];
       },
       summaryTally() {
-        const recipients = this.getLearnersForGroups(this.lesson.groups);
-        return this.getContentStatusTally(this.$route.params.exerciseId, recipients);
+        return this.getContentStatusTally(this.$route.params.exerciseId, this.recipients);
       },
       lessonGroups() {
         if (!this.lesson.groups.length) {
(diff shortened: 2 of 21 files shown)


kolibri/plugins/coach/assets/src/views/reports/ReportsExerciseLearners.vue
```diff
@@ -0,0 +1,108 @@
...
+        <th>{{ coachStrings.$tr('progressLabel') }}</th>
+        <th>{{ coachStrings.$tr('timeSpentLabel') }}</th>
+        <th v-if="showGroupsColumn">
+          {{ coachStrings.$tr('groupsLabel') }}
+        </th>
+        <th>{{ coachStrings.$tr('lastActivityLabel') }}</th>
+      </tr>
+    </thead>
+    <transition-group slot="tbody" tag="tbody" name="list">
+      <tr v-for="entry in entries" ref="entry" :key="entry.id">
```
jonboiser: Is the `ref` used somewhere outside of this component?

MisRob: Yes, in specs - I usually prefer to use `ref`s there.

kolibri/plugins/coach/assets/src/views/reports/ReportsLessonExerciseLearnerListPage.vue
```diff
@@ -73,26 +60,78 @@
...
         return this.lessonMap[this.$route.params.lessonId];
       },
-      recipients() {
-        return this.getLearnersForGroups(this.lesson.groups);
+      summaryTally() {
+        const recipients = this.getLearnersForGroups(this.lesson.groups);
+        return this.getContentStatusTally(this.$route.params.exerciseId, recipients);
+      },
+      nonEmptyGroups() {
+        if (!this.groups || !this.groups.length) {
```
jonboiser: We should only show tables for groups that have been assigned the Lesson. `this.groups` will return all of the groups in the classroom.

kolibri/plugins/coach/assets/test/views/reports/ReportsExerciseLearners.spec.js
```diff
@@ -0,0 +1,163 @@
...
+      entries,
+    });
+
+    const rows = wrapper.findAll({ ref: 'entry' });
+    expect(rows.length).toBe(2);
+
+    const firstRow = rows.at(0);
+    const secondRow = rows.at(1);
+
+    expect(getCol(firstRow, 0).html()).toContain('learner1');
```
jonboiser: A snapshot test might be an efficient way to implment this particular test, since it's checking the presentation of this element.

kolibri/plugins/coach/assets/src/views/reports/ReportsLessonExerciseLearnerListPage.vue
MisRob: Oh didn't notice, thanks, good catch. Will check it.

kolibri/plugins/coach/assets/src/views/reports/ReportsExerciseLearners.vue
jonboiser: Which 'tr' does `$refs.entry` refer to? The last one?

In your test is it possible to write the test using something more high-level, like the data from vue-test-utils Wrapper methods, instead of having to access more low-level DOM info from `refs`?

kolibri/plugins/coach/assets/test/views/reports/ReportsExerciseLearners.spec.js
MisRob: I tried to avoid snapshot in this particular case because of "Last Activity" - not sure if mocking time would be useful only because of one test and also didn't want to test another component's responsibility but I'll check if some stub could solve it.

kolibri/plugins/coach/assets/test/views/reports/ReportsExerciseLearners.spec.js
```diff
@@ -0,0 +1,163 @@
...
+
+    expect(wrapper.isVueInstance()).toBe(true);
+  });
+
+  it('renders all entries', () => {
+    const wrapper = initWrapper({
+      entries,
+    });
+
+    const rows = wrapper.findAll({ ref: 'entry' });
```
jonboiser: Could you accomplish this using a CSS selector something else?

kolibri/plugins/coach/assets/test/views/reports/ReportsExerciseLearners.spec.js
```diff
@@ -0,0 +1,163 @@
...
+            },
+          },
+        ],
+      });
+
+      row = wrapper.find({ ref: 'entry' });
+    });
+
+    it("renders learner's name as a link to an exercise", () => {
+      expect(getCol(row, 0).html()).toContain('router-link');
```
jonboiser: You can actually stub KRouterLink so it can be treated as a very simple vue component, and the Wrapper API works on it.

Here's an example https://github.com/learningequality/kolibri/blob/develop/kolibri/plugins/coach/assets/src/views/common/notifications/__test__/NotificationCard.spec.js#L15

kolibri/plugins/coach/assets/src/views/reports/ReportsExerciseLearners.vue
MisRob: Basically, it's used only to find those elements using vue-test-utils `find` - in case there is markup/css selector change, you don't need to update your `find`s, when compared to trying to find an element by a class or tag, that's all. Both ways have some advantages/disadvantages, no problem to change it if you don't use it in the codebase.

kolibri/plugins/coach/assets/test/views/reports/ReportsExerciseLearners.spec.js
jonboiser: If you want to be advanced, you can mock the PC clock to get a consistent "Last Activity" value.

You could also, like you say, stub 'LastActivity' so that it just doesn't appear in the snapshot, or in modified version (maybe just a span with the input date as text, for exapmle).

kolibri/plugins/coach/assets/test/views/reports/ReportsExerciseLearners.spec.js
MisRob: Answered above.

kolibri/plugins/coach/assets/test/views/reports/ReportsExerciseLearners.spec.js
MisRob: Not sure if I can understand this comment - why should I stub it here? I just want to check that the link is there.

kolibri/plugins/coach/assets/test/views/reports/ReportsExerciseLearners.spec.js
MisRob: > If you want to be advanced, you can mock the PC clock to get a consistent "Last Activity" value

That sounds like something to be tested rather in `ElapsedTime` component.

kolibri/plugins/coach/assets/test/views/reports/ReportsExerciseLearners.spec.js
jonboiser: It's just an unusual test, because the actual rendered html will never have a `<router-link>` tag.

A test like
```
expect(getCol(row, 0).find({ name: 'KRouterLink' }).props().to).toEqual(...)
```

Mirrors the implementation a little more closely.

MisRob: Ah, ok, get what you meant now. Yep makes sense.

kolibri/plugins/coach/assets/src/views/reports/ReportsExerciseLearners.vue
jonboiser: We generally reserve the use `ref` if we need to access it in JS application code. Even if it's used in a test, it sort of looks like dead code, and I have the tendency to delete `refs` if I see they aren't used anymore in `methods`, etc.

People who want to avoid using "brittle" CSS selectors have been using this test-specific attribute `data-test`. In the spec, you are using `ref` as an attribute selector, and not really using the `ref` in the same way we use it components (e.g. low-level DOM access).

MisRob: Yes, it's one of the disadvantages indeed.

I like the idea of using the attribute, it combines the best of both worlds, thanks for the tip!

indirectlylit:
> It was a bit difficult to track all dependencies to store etc. with commonCoach mixin when navigating in coach codes. I noticed that it's being used also for smaller components, but tried to avoid that in a new table component. I was thinking about rather splitting the mixin into helpers and smaller pieces in the future (even when it would cost us some duplicates in components/views)? Would it make sense?

That mixin was supposed to be a temporary hack and I never got around to cleaning it up. Feel free to do any cleanup you think would be appropriate.

> It's my first feature so would like to ask - when will be the right time to add gherkin stories? After the review?

You can add them any time. In theory, adding then before review means that you don't need to write reviewer guidance :)

That said, if you want to do them as a separate PR after this is merged, that's fine too - especially if you're trying to get it in before string freeze. Usually in that case I'll open a new issue specifically for the stories.

kolibri/plugins/coach/assets/test/views/reports/ReportsExerciseLearners.spec.js
MisRob: Eventually, I added server time mock. Didn't realize that with stub, I couldn't test if an information is displayed or no.

kolibri/plugins/coach/assets/src/views/reports/ReportsLessonExerciseLearnerListPage.vue
MisRob: Done. I also noticed that all groups were displayed in a table in Groups column so I filtered them by lesson groups, too. Otherwise it would be misleading when switching from groups to the list of all learners.

MisRob:
@jonboiser asked me to move everything to one PR, so I'm adding it here.

Updates:
- comments resolved
- found out what was wrong with my router mocks so I could also add a couple of basic specs for exercise learners report page
- ungrouped learners are displayed in a separate section at the bottom of the page (when a whole class is a recipient)
- empty groups (that are recipients of a lesson) are displayed
- implemented same grouping feature for resources

screencapture-localhost-8000-en-coach-2019-05-07-16_30_47
screencapture-localhost-8000-en-coach-2019-05-07-16_31_35

MisRob:
As discussed with JB, I've just also added Preview buttons to resource/exercise learners list reports to this PR:

Screenshot 2019-05-07 at 21 25 53
Screenshot 2019-05-07 at 21 26 03

This part is not exactly ideal because there are more mechanisms and in some situations, it gets a bit complicated. Also, I didn't work on a correct immersive toolbar title when navigating from reports (we agreed it should display an exercise/resource name instead of "Manage resources"). I was also thinking that maybe changing exercise/resource preview URL might be suitable after adding this feature, since it's not related to "Coach - Plan" only anymore. Anyway, we agreed that some basic functionality is OK so we can close it till Thursday.

I'll start follow-up tickets after the review and will have a look if it could be simplified a bit and also add Gherkin stories later.

When testing this PR, please try to navigate from exercise/resource list to exercise/resource preview, exercise learner detail and back for both "View by groups" checkbox states.

jonboiser:
For the Preview buttons, can you place them to the right of the "<- Back to Lesson Name" link. Unfortunately, the mock up was out of date with this detail.

There is a component called "BackLinkWithOptions" with that will layout the button and back link like this:

Screen Shot 2019-05-07 at 4 13 27 PM

jonboiser:
Next to the group header (h2.group-title), could you add the group icon? You can use the `<h2><KLabeledIcon><KIcon group slot="icon" /></h2>` setup.

jonboiser:
Don't need to fix this, but just noticed that the "Time spent" column is not being live-updated when a user finishes watching a video (not sure about exercises). Screenshot shows row for "b" after the status is live-updated to "Completed", but the timespent column is unchanged.

Screen Shot 2019-05-07 at 4 36 42 PM

kolibri/plugins/coach/assets/src/views/reports/ReportsLessonResourceLearnerListPage.vue
```diff
@@ -82,12 +114,23 @@
...
     name: 'ReportsLessonResourceLearnerListPage',
-    components: {},
+    components: {
+      ReportsResourceLearners,
+      ReportsResourcesStats,
+    },
     mixins: [commonCoach],
+    data() {
+      return {
+        viewByGroups: Boolean(this.$route.query && this.$route.query.groups),
```
jonboiser: I think `this.$route.query` is always going to be truthy, since it's an empty object if no query parameters are provided.

MisRob:
@jonboiser Just pushed all updates related to the feature, new screenshots:

screencapture-localhost-8000-en-coach-2019-05-08-19_49_02
screencapture-localhost-8000-en-coach-2019-05-08-19_49_36

jonboiser (approved):
Looks great!

MisRob:
Yep, I especially like the idea with group icons, it makes orientation much easier, especially when "Ungrouped learners" are present :)

MisRob:
Regarding "Time spent" column issue, I'll have a look at it later together with this PR's follow-ups.

jonboiser:
Okay, I wouldn't worry too much about the Time Spent issue. The tables are "live updated" with data coming from the Notifications API, and I'm not sure if Notifications contain the info to update the Time Spent column right now.

indirectlylit:
this looks awesome, great job

would one of you mind filing an issue re: the time-spent updates?

MisRob:
Thank you. Yes, I'm planning to have a look at it and open an issue.

MisRob:
#5496

Issue #5418, opened 2019-04-24
Allow an exit path from search


khangmach:
### Observed behavior
Currently there isn't a way for a user to easily exit the search view in the LEARN tab. More details here: https://github.com/learningequality/clearinghouse/issues/32. The user has to click on the hamburger menu and then click on the 'Learn' menu option again.

### Expected behavior
Have an easy way to get back to Learn home page. Swap the current appbar with a static back appbar that leads back to the root page of Learn.

Label text: Learn

![end-state](https://user-images.githubusercontent.com/25045166/56693548-8fcb3700-6699-11e9-8054-99f173941fab.png)

### User-facing consequences
Complicates the user experience. Unnecessary navigation back to home

### Steps to reproduce
- Navigate to Learn tab
- Input a search query
- Press search button

### Context
Tell us about your environment, including:
 * Kolibri version: Kolibri 0.12.2
 * Operating system: MacOS
 * Browser: Chrome

jonboiser:
Implemented in #5459, but looks a little different than the description in the original post (different color, label, and back link behavior).

indirectlylit:
This is fine, but the inconsistency with lesson and quiz creation is a little odd

| search in channels | search in quizzes | search in lessons |
|--|--|--|
| ![image](https://user-images.githubusercontent.com/2367265/57497041-cf457600-728a-11e9-96b8-684545581ade.png) | ![image](https://user-images.githubusercontent.com/2367265/57496810-bc7e7180-7289-11e9-95a6-7092103ea130.png) | ![image](https://user-images.githubusercontent.com/2367265/57496934-50503d80-728a-11e9-9a82-389b85289646.png) |

PR #5459, opened 2019-05-06
set immersive AppBar for Search in Learn section

nucleogenesis:
### Summary

This change fixes this issue https://github.com/learningequality/kolibri/issues/5418 by setting the Learn Search page's AppBar to use the Immersive version - clicking the arrow will always return the user to the Learn home page.

Before:
exit-path-before

After:
exit-path-after

…

### Reviewer guidance
- Checkout develop & enter a search input in the Learn section
- See the hamburger menu
- Checkout this PR's branch, repeat the search
- See the back icon
- Click the icon to return to the Learn home page
…

### References

Kolibri Issue: https://github.com/learningequality/kolibri/issues/5418
Clearinghouse Issue: https://github.com/learningequality/clearinghouse/issues/32

…

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/plugins/learn/assets/src/views/LearnIndex.vue
+++ b/kolibri/plugins/learn/assets/src/views/LearnIndex.vue
@@ -139,12 +139,12 @@
         }
         if (this.pageName === PageNames.SEARCH) {
           return {
-            appBarTitle: this.$tr('learnTitle'),
+            appBarTitle: this.$tr('searchTitle'),
             immersivePage: true,
             // Default to the Learn root page if there is no lastRoute to return to.
             immersivePageRoute: this.lastRoute || this.$router.getRoute(PageNames.TOPICS_ROOT),
-            immersivePagePrimary: true,
-            immersivePageIcon: 'arrow_back',
+            immersivePagePrimary: false,
+            immersivePageIcon: 'close',
           };
         }
 
@@ -243,6 +243,7 @@
     $trs: {
       learnTitle: 'Learn',
       examReportTitle: '{examTitle} report',
+      searchTitle: 'Search',
     },
   };
 
--- a/kolibri/plugins/learn/assets/src/views/LearnIndex.vue
+++ b/kolibri/plugins/learn/assets/src/views/LearnIndex.vue
@@ -231,11 +231,13 @@
           return;
         }
 
-        // Destructure the oldRoute into an object with the only four properties.
+        // Destructure the oldRoute into an object with 3 specific properties.
         // Setting this.lastRoute = oldRoute causes issues for some reason.
-        this.lastRoute = (({ name, query, path, params }) => ({ name, query, path, params }))(
-          oldRoute
-        );
+        this.lastRoute = {
+          name: oldRoute.name,
+          query: oldRoute.query,
+          params: oldRoute.params,
+        };
       },
     },
     $trs: {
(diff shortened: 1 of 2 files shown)


kolibri/plugins/learn/assets/src/views/LearnIndex.vue
```diff
@@ -132,6 +132,15 @@
             };
           }
         }
+        if (this.pageName === PageNames.SEARCH) {
+          return {
+            appBarTitle: this.$tr('learnTitle'),
+            immersivePage: true,
+            immersivePageRoute: this.$router.getRoute(PageNames.TOPICS_ROOT),
```
jonboiser: We should return the user to whichever page they were on when they initiated the search, since the search bar appears every where in the Learn page. For example, if they are deep into Khan Academy topics, initiate a search, then exit the search page, we want them to go back to the same topic, not to TOPICS_ROOT (the list of channels).

In the Coach > Lessons > Manage Resources workflow, we use query parameters in that flow's search URL to determine where to return the user when they press the "Exit search" button. We also use similar tricks in Coach reports to send users back to the right page when they view other immersive pages (like the exercise and exam report pages).

Exiting Search in Learn might be tricky because it kind of combines all of those types of pages:

1. When browsing content (pageName is TOPICS_TOPIC or TOPICS_ROOT), you need to specify which page type and topicId you were last in
1. When browsing another page like Class Assignments, you need to specify which page you were last in (including the class identifier).
1. Other pages like "Recommended" is similar to the above (but don't require an identifier).
1. The search page, but with a different search term.

nucleogenesis: @jonboiser I was thinking the same thing. I think that would make more sense and with your notes I imagine I'll have a fairly easy time getting that functionality.

@khangmach what are your thoughts?

khangmach: Returning the user to where they were previously is a good strat! 👍

kolibri/plugins/learn/assets/src/views/LearnIndex.vue
```diff
@@ -208,6 +223,21 @@
...
+      $route: function(newRoute, oldRoute) {
+        // Return if the current route is a search path. Otherwise this method creates
+        // an infinite loop.
+        if (newRoute.name === 'SEARCH' && oldRoute.name === 'SEARCH') {
+          return;
+        }
+
+        // Destructure the oldRoute into an object with the only four properties.
+        // Setting this.lastRoute = oldRoute causes issues for some reason.
+        this.lastRoute = (({ name, query, path, params }) => ({ name, query, path, params }))(
```
jonboiser: Might be simpler to just create a new object from oldRoute. Like `lastRoute = { name: oldRoute.name }`, etc. And I don't think router-link objects need the `path` property. name, query, and params always are enough.

jonboiser (approved):

bjester:
I'm wondering if any of these changes is causing a navigation loop I'm experiencing? If not, I'll add an issue. Here are my reproduce steps:
1. From Learn, search for content
2. Open the content from search
3. Click the back arrow in the header
4. Click the X to close search
5. Return to content opened from search
6. Clicking back arrow and X from now on loop back and forth

jonboiser:
Yeah, at step (4), the last page will be the content viewed at step (3), so clicking X will take you back to that content item.

I think to fix this we might need to add extra logic to the `watch.$route` handler. When, Going from SEARCH to TOPICS_CONTENT, we may not necessarily update the `lastRoute` @nucleogenesis

PR #5475, opened 2019-05-10
Miscellaneous EPUB renderer fixes

bjester:
#### Fixes
- Converts epub.js flow mode to `scrolled` when tables are detected to prevent issues rendering them in paginated columns, resolving https://github.com/learningequality/kolibri/issues/4782
- Next Page navigation button occasionally disappeared in Firefox due to white-space wrapping
- Theme and font size settings were not being saved and restored properly when returning to the renderer
- Section slider did not properly represent current "location" (position in document) when returning with a saved location
- Search sidebar did not report correct number of results nor show results
- Theme selector buttons sometimes did not show corresponding theme background color

#### Other Updates
- Updated `epubjs` version to `0.3.84`
- Refactored out a large number of JS calculated dimensions
- Refactored access to epub.js created elements to operate through epub.js classes as much as possible
- Changed loading state to overlay renderer area instead of using `v-show` since `display: none` element trees naturally don't have dimensions

#### Known Issues
- Increasing font size may cause content in scrolled mode to overflow and be hidden horizontally without a horizontal scrollbar, until either the page is resized (including going fullscreen) or the page is refreshed. New issue to be added.
- In IE 11, going fullscreen may trigger a display error within epub.js causing the "Could not load book" banner to appear, but the document should still be visible.

#### Original Issue
epub-table-overflow

#### Scrolled Flow when tables are present
screenshot-localhost-8000-2019 05 10-14-25-31

#### Search Issue
screenshot-localhost-8000-2019 05 10-14-24-16

#### Search Issue Resolved
screenshot-localhost-8000-2019 05 10-14-24-59

#### Testing documents
- EPUB's shown in the screenshots can be imported via token `jafal-zovuk`
- Other EPUB's for testing can be imported via token `gojad-patot`

#### Notes
- Take note of Known Issues above
- Changes were tested in many different browsers on BrowserStack, so similar testing for verification is suggested

### References

Related issue: https://github.com/learningequality/kolibri/issues/4782
Initial approach: https://github.com/learningequality/kolibri/pull/5430

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/integration_testing/features/learner/learner-engage-epub-content.feature
+++ b/integration_testing/features/learner/learner-engage-epub-content.feature
@@ -1,0 +1,125 @@
+Feature: Learner engages EPUB content
+  Learner engages with EPUB content from Channels tab and EPUB reader capabilities
+
+  Background:
+    Given I am signed in to Kolibri as a Learner user
+      And there are one or more channels imported on the device with EPUB content
+      And I am on the *Topics* page for a channel with EPUB content
+
+  Scenario Outline: Opens and engages with EPUB paginated content
+    When I am on the *Topics* page for <channel>
+    Then I see the *Channels > '<channel>'* breadcrumb
+      And I see all the topics for the channel <channel>
+    When I click the topic <topic>
+    Then I see the *Channels > '<channel>' > '<topic>'* breadcrumb
+      And I see all the the subtopics and content items of the topic <topic>
+      And I see <content_item> content item having a red dog-ear in the top left corner indicating it's a document
+    When I click the <content_item> content item
+    Then I see the *Channels > '<channel>' > '<topic>' > '<content_item>'* breadcrumb
+      And I see the <content_item> content
+    But I shouldn't be able to scroll in the <content_item> content
+    When I click the button to *Go to next page*
+    Then I see the <content_item> content change to the next page
+    When I click the button to *Go to previous page*
+    Then I see the <content_item> content change to the previous page
+      And I see the <content_item> content that I saw before clicking the button to *Go to next page*
+    Examples:
+      | channel              | topic            | content_item                      |
+      | EPub Testing Channel | Richard's EPubs  | The Adventures of Sherlock Holmes |
+
+  Scenario Outline: Opens and engages with EPUB scrolled content
+    When I am on the *Topics* page for <channel>
+    Then I see the *Channels > '<channel>'* breadcrumb
+      And I see all the topics for the channel <channel>
+    When I click the topic <topic>
+    Then I see the *Channels > '<channel>' > '<topic>'* breadcrumb
+      And I see all the the subtopics and content items of the topic <topic>
+      And I see <content_item> content item having a red dog-ear in the top left corner indicating it's a document
+    When I click the <content_item> content item
+    Then I see the *Channels > '<channel>' > '<topic>' > '<content_item>'* breadcrumb
+      And I see the <content_item> content
+      And I am able to scroll the <content_item> content
+    But I do not see buttons to *Go to next page* or *Go to previous page*
+    Examples:
+      | channel              | topic          | content_item      |
+      | EPub Testing Channel | Blaine's EPubs | Epub with Tables! |
+
+  Scenario Outline: Engages with EPUB table of contents
+    When I am on the page for the <content_item> content item
+    Then I see the *Channels > '<channel>' > '<topic>' > '<content_item>'* breadcrumb
+      And I see the <content_item> content
+    When I click the button to *Toggle table of contents side bar*
+    Then I see the book's table of contents overlaid on the left side of the content
+    When I click the button to *Toggle table of contents side bar* again
+    Then I see the book's table of contents disappear
+    When I click the button to *Toggle table of contents side bar*
+    Then I see the book's table of contents overlaid on the left side of the content again
+    When I click on the <section> section of the table of contents
+    Then I see the table of contents side bar disappear
+      And I see the <content_item> content change to the section <section>
+    Examples:
+      | channel              | topic            | content_item                      |
+      | EPub Testing Channel | Richard's EPubs  | The Adventures of Sherlock Holmes |
+
+  Scenario Outline: Engages with EPUB settings
+    When I am on the page for the <content_item> content item
+    Then I see the *Channels > '<channel>' > '<topic>' > '<content_item>'* breadcrumb
+      And I see the <content_item> content
+    When I click the button to *Toggle settings side bar*
+    Then I see settings for *Text Size* and *Theme* overlaid on the right side of the content
+    When I click the button to *Increase text size*
+    Then I see the text size of <content_item> become larger
+    When I click the button to *Decrease text size*
+    Then I see the text size of <content_item> become smaller
+      And I see the text size as it was originally
+    When I click the button to *Set beige theme*
+    Then I see the background of <content_item> become beige
+      And I see the text color of <content_item> become brown
+    When I click the button to *Set grey theme*
+    Then I see the background of <content_item> become grey
+      And I see the text color of <content_item> become white
+    When I click the button to *Set black theme*
+    Then I see the background of <content_item> become black
+      And I see the text color of <content_item> become white
+    When I click the button to *Set high contrast black theme*
+    Then I see the background of <content_item> become black
+      And I see the text color of <content_item> become yellow
+    When I click the button to *Set high contrast white theme*
+    Then I see the background of <content_item> become white
+      And I see the text color of <content_item> become blue
+    When I click the button to *Set white theme*
+    Then I see the background of <content_item> become white
+      And I see the text color of <content_item> become black
+    When I click the button to *Toggle settings side bar* again
+    Then I see the settings side bar disappear
+    Examples:
+      | channel              | topic            | content_item                      |
+      | EPub Testing Channel | Richard's EPubs  | The Adventures of Sherlock Holmes |
+
+  Scenario Outline: Engages with EPUB search
+    When I am on the page for the <content_item> content item and section <section> of the content
+    Then I see the *Channels > '<channel>' > '<topic>' > '<content_item>'* breadcrumb
+      And I see the <content_item> content and <section> in the content
+    When I click the button to *Toggle search side bar*
+    Then I see the search side bar overlaid on the right side of the content
+    When I type <search> into the input to *Enter search query* and click the button to *Submit search query*
+    Then I see search results below the input
+      And I see <search> highlighted in the EPUB content
+    When I click the button to *Toggle search side bar* again
+    Then I see the search side bar disappear
+      And I see the <search> no longer highlighted in the EPUB content
+    Examples:
+      | channel              | topic            | content_item                      | section   | search    |
+      | EPub Testing Channel | Richard's EPubs  | The Adventures of Sherlock Holmes | Chapter 1 | Chapter I |
+
+  Scenario Outline: Engages with EPUB fullscreen mode
+    When I am on the page for the <content_item> content item
+    Then I see the *Channels > '<channel>' > '<topic>' > '<content_item>'* breadcrumb
+      And I see the <content_item> content
+    When I click the button to *Toggle fullscreen*
+    Then I see <content_item> content go fullscreen
+    When I click the button to *Toggle fullscreen* again
+    Then I see the page as it was in the beginning
+    Examples:
+      | channel              | topic            | content_item                      |
+      | EPub Testing Channel | Richard's EPubs  | The Adventures of Sherlock Holmes |
--- a/integration_testing/features/learner/learner-engage-epub-content.feature
+++ b/integration_testing/features/learner/learner-engage-epub-content.feature
@@ -1,125 +1,99 @@
-Feature: Learner engages EPUB content
-  Learner engages with EPUB content from Channels tab and EPUB reader capabilities
+Feature: Learner engages ePUB content
+  Learner needs to be able to engage with ePUB content and use all the ePUB reader features
 
   Background:
     Given I am signed in to Kolibri as a Learner user
-      And there are one or more channels imported on the device with EPUB content
-      And I am on the *Topics* page for a channel with EPUB content
+      And there are one or more channels imported on the device with ePUB content
+      And I am on the *Channels* page for a channel with ePUB content
 
-  Scenario Outline: Opens and engages with EPUB paginated content
-    When I am on the *Topics* page for <channel>
-    Then I see the *Channels > '<channel>'* breadcrumb
-      And I see all the topics for the channel <channel>
-    When I click the topic <topic>
-    Then I see the *Channels > '<channel>' > '<topic>'* breadcrumb
-      And I see all the the subtopics and content items of the topic <topic>
-      And I see <content_item> content item having a red dog-ear in the top left corner indicating it's a document
-    When I click the <content_item> content item
-    Then I see the *Channels > '<channel>' > '<topic>' > '<content_item>'* breadcrumb
-      And I see the <content_item> content
-    But I shouldn't be able to scroll in the <content_item> content
-    When I click the button to *Go to next page*
-    Then I see the <content_item> content change to the next page
-    When I click the button to *Go to previous page*
-    Then I see the <content_item> content change to the previous page
-      And I see the <content_item> content that I saw before clicking the button to *Go to next page*
-    Examples:
-      | channel              | topic            | content_item                      |
-      | EPub Testing Channel | Richard's EPubs  | The Adventures of Sherlock Holmes |
+    Scenario: Browse and find ePUB content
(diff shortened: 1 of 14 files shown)


rtibbles:
Note, this should be targeted to `release-v0.12.y` for the 0.12.y milestone.

bjester:
@rtibbles Thanks

jonboiser:
On the Settings widget, all the themes are gray squares on my instance:

Screen Shot 2019-05-13 at 5 32 01 PM

jonboiser:
~On my computer, on both Chrome and Firefox, the books with tables are going into the single-column "scroll mode", but I can't actually scroll (trying with my MacBook touchpad, two finger drag gesture) and the scroll bars don't appear on the right like in your screenshots.~

~On Firefox, dragging the progress input does seem to move my position, it's not very consistent. Sometimes it takes 1 increment (pressing -> on my keyboard) to move me, some times two.~

jonboiser:
~For other kinds of books like "Accelerando", in Chrome, it shows the loading screen indefinitely, but will start to load properly if you resize the window. (Noticed this by opening the dev tools when I was wondering why the book wouldn't start loading).~

bjester:
@jonboiser Regarding theme colors, I'm not seeing that. That's in both Chrome and Firefox for you?

Regarding not being able to scroll... That's weird. What versions exactly?

screenshot-localhost-8000-2019 05 13-17-38-06

jonboiser:
I forgot to update epubjs locally before trying this. Let me see if it fixes these issues.

jonboiser:
Thankfully, upgrading to the pinned version fixed most of these, but the theme buttons are still all grey for me.

jonboiser:
Aphrodite isn't creating distinct classes for all the theme buttons, it's re-using the same computed class, with a grey background and black text.

bjester:
@jonboiser Yeah, the pinned version of epubjs we were using had several issues that I spent time trying to work around until I tried upgrading and a bunch of things started working correctly. 

Regarding the theme buttons, that makes sense as to why they would all look the same. I did add some strings for `aria-label`s for two of the colors that didn't have them, so perhaps that messed something up. I'm not seeing what you're seeing, but I'll dig in some more.

rtibbles:
Glad that the upgrade went without a hitch! I think I'd tried to upgrade previously, and it had caused some issues.

Feel free to ping me on the Aphrodite issues.

integration_testing/features/learner/learner-engage-epub-content.feature
```diff
@@ -0,0 +1,125 @@
...
+    When I am on the page for the <content_item> content item
+    Then I see the *Channels > '<channel>' > '<topic>' > '<content_item>'* breadcrumb
+      And I see the <content_item> content
+    When I click the button to *Toggle fullscreen*
+    Then I see <content_item> content go fullscreen
+    When I click the button to *Toggle fullscreen* again
+    Then I see the page as it was in the beginning
+    Examples:
+      | channel              | topic            | content_item                      |
+      | EPub Testing Channel | Richard's EPubs  | The Adventures of Sherlock Holmes |
```
indirectlylit: @radinamatic would you mind reviewing the new gherkin stories here?

kolibri/plugins/document_epub_render/assets/src/views/EpubRendererIndex.vue
```diff
@@ -418,35 +382,75 @@
...
+            }, false);
+
+            // Tables have issues rendering, so switch from paginated flow to scrolled
+            if (hasTables) {
+              this.scrolled = true;
+              this.rendition.flow('scrolled');
+              this.rendition.clear();
+              // Re-renders and targets this.savedLocation (if set)
+              return this.rendition.display(this.savedLocation || undefined);
+            }
```
indirectlylit, line 417: Would you mind posting a quick comparison of epubs rendered as scrolled vs paginated? I'm curious how big a difference it "feels like".

If there's a chance that having two different interaction modes is confusing, perhaps it would be preferable for consistency to display _all_ books as scrolled? (cc @jtamiace @khangmach)

kolibri/plugins/document_epub_render/assets/src/views/LoadingError.vue
```diff
@@ -1,11 +1,14 @@
...
-    type="error"
-    :dismissible="false"
-  >
-    {{ $tr('couldNotLoadThisBook') }}
-  </UiAlert>
+  <div class="epub-renderer-error" :class="{loaded: loaded}">
+    <UiAlert
+      type="error"
+
+      :dismissible="false"
```
indirectlylit, line 6: whitespace

kolibri/plugins/document_epub_render/assets/src/views/LoadingError.vue
```diff
@@ -1,11 +1,14 @@
 <template>
 
-  <UiAlert
-    type="error"
-    :dismissible="false"
-  >
-    {{ $tr('couldNotLoadThisBook') }}
-  </UiAlert>
+  <div class="epub-renderer-error" :class="{loaded: loaded}">
```
indirectlylit: can be just `:class="{loaded}"`

kolibri/plugins/document_epub_render/assets/src/views/LoadingError.vue
```diff
@@ -27,4 +36,22 @@
...
-<style lang="scss" scoped></style>
+<style lang="scss" scoped>
+
+  .epub-renderer-error {
+    position: absolute;
+    top: 0;
+    right: 0;
+    bottom: 0;
+    left: 0;
+    z-index: 7;
```
indirectlylit: generally we try to match z-indexes to one of [Material's default elevations](https://material.io/design/environment/elevation.html#default-elevations)

kolibri/plugins/document_epub_render/assets/src/views/LoadingScreen.vue
```diff
@@ -32,9 +43,14 @@
...
   .loading-screen {
     position: absolute;
-    top: 50%;
-    left: 50%;
-    transform: translate(-50%, -50%);
+    top: 0;
+    right: 0;
+    bottom: 0;
+    left: 0;
+    z-index: 7;
```
indirectlylit: z-index Material elevation note - same as above

kolibri/plugins/document_epub_render/assets/src/views/LoadingScreen.vue
```diff
@@ -32,9 +43,14 @@
...
+    display: flex;
+    align-items: center;
+    justify-content: center;
```
indirectlylit: We've tried to avoid using flexbox due to [less-than-complete support](https://caniuse.com/#feat=flexbox) on older Android browsers

kolibri/plugins/document_epub_render/assets/src/views/SettingsSideBar.vue
```diff
@@ -174,11 +180,15 @@
...
       increase: 'Increase',
+      decreaseTextSize: 'Decrease text size',
+      increaseTextSize: 'Increase text size',
       theme: 'Theme',
       setWhiteTheme: 'Set white theme',
       setBeigeTheme: 'Set beige theme',
       setGreyTheme: 'Set grey theme',
       setBlackTheme: 'Set black theme',
+      setHighContrastBlackTheme: 'Set high contrast black theme',
+      setHighContrastWhiteTheme: 'Set high contrast white theme',
```
indirectlylit: We're past string freeze for the `release-v0.12.y` branch, so we can't add or modify strings in this PR. We can make a small follow-up PR targeting `develop`

kolibri/plugins/document_epub_render/assets/src/views/EpubRendererIndex.vue
bjester: The paginated experience feels more like a book because the content is usually rendered with two columns, with the exception of smaller viewports where it will transition to a single column. I feel that the paginated mode is superior for reading books and pure text documents since the readability is better, specifically because of the column layout.

The scrolled mode is definitely superior for documents with tables or inflexible flow typed content that doesn't morph to different layouts. We may find there are other situations where the scrolled mode is superior as well, but when it comes to books, I think it's a poorer experience. To render a book in scrolled mode, it would be preferable to limit the content width which wastes screen real-estate and essentially makes fullscreen mode mostly unnecessary.

I don't think it's confusing to have different modes. Notion essentially does this with tables inside pages, but contextualizes the scrolled vs. paginated/column mode by each block. Mixing the two in the same document I find actually less ideal, so this approach is the best of all options IMO.

kolibri/plugins/document_epub_render/assets/src/views/LoadingError.vue
bjester: Ha where did that come from

kolibri/plugins/document_epub_render/assets/src/views/LoadingError.vue
bjester: Good call.

kolibri/plugins/document_epub_render/assets/src/views/LoadingError.vue
bjester: Okay, I'll have to change some existing z-indexes that don't match that as well.

rtibbles: Yeah, I had noticed some non-compliant ones previously when I was doing some work on the top bar and buttons also, but didn't dig through everything to fix.

kolibri/plugins/document_epub_render/assets/src/views/LoadingScreen.vue
bjester: Understood. This should be just fine though, unless you feel strongly this should be changed.

kolibri/plugins/document_epub_render/assets/src/views/EpubRendererIndex.vue
indirectlylit: sounds good!

kolibri/plugins/document_epub_render/assets/src/views/LoadingScreen.vue
indirectlylit: > This should be just fine though

could you elaborate?

I don't actually know a lot about what parts of flexbox worked on older android browsers, so I've generally just advocated avoiding it. However if it works in some specific cases, we can adjust our conventions accordingly but we'll need to make sure everyone on the team understands when it's permitted.

bjester: The styles in Kolibri run through `autoprefixer` correct? That's what I've seen thus far. With prefixed styles here, it should work just fine.

Some gotchas are with `order`, `justify-content: space-around`, `flex-grow`, and worst of all in my experience, `flex-direction: column`. 

https://benfrain.com/flexbox-gotchas-for-android-2-1-4-3-stock-browser/
https://blog.frankmtaylor.com/2013/10/11/true-story-the-android-web-browser-is-the-honey-badger-of-flexbox/

bjester: Also, if the Android Browser is really supported, I think we should update the list here: https://kolibri.readthedocs.io/en/latest/access.html#compatibility

kolibri/plugins/document_epub_render/assets/src/views/SettingsSideBar.vue
bjester: Oh okay I remember that being announced but I thought this was going into a different release. I'm a bit confused haha.

kolibri/plugins/document_epub_render/assets/src/views/LoadingScreen.vue
indirectlylit: Yup, the styles get auto-prefixed.

I think it was those two articles that made me decide to simply ban `flexbox` from our codebase!

Basically the thinking was that simpler, more conservative conventions are easier to enforce. I'm open to reversing that (especially if it turns out it applies to browsers we don't actually need to support), but my guess is that maybe it's helped us avoid browser-specific layout bugs?

I created a new issue to resolve the browser support question here: https://github.com/learningequality/kolibri-docs/issues/95

kolibri/plugins/document_epub_render/assets/src/views/SettingsSideBar.vue
indirectlylit: * 0.12.3 / x -> upcoming release, no string changes
* 0.12.4 / y -> upcoming "bigger" release with full integration test pass and string changes, currently being translated
* 0.13.0 / develop -> next big feature release

kolibri/plugins/document_epub_render/assets/src/views/LoadingScreen.vue
bjester: Personally, I don't think that's best to ban flexbox. If you continue with the approach of banning CSS syntaxes, you'll end up banning everything. For instance, iOS Safari has some pretty annoying issues with fixed positioning that haven't been fixed as far as I know-- and that's a current browser. So would you ban all fixed positioning based off that? I would say no. And if all developers do that now or did that years ago, web development and design would have not progressed because no one would use new abilities.

In this instance, I truly believe no issues will be present, based off my experience. I think in these scenarios where there is a question of whether this will work correctly, you can take into account the use and what happens when it breaks. In this case, it's a loading animation and if the flexbox doesn't work, it simply displays in the top left corner. Now if we're talking a page template/layout, that's something entirely different-- the fallback is probably a broken experience.

kolibri/plugins/document_epub_render/assets/src/views/SettingsSideBar.vue
bjester: So "currently being translated" means no string additions?

indirectlylit: correct - usually we call that "string freeze" but I think that needs to be renamed :)

kolibri/plugins/document_epub_render/assets/src/views/LoadingScreen.vue
indirectlylit: Yeah I totally trust that this specific case is fine.

My main motivation is about the precedent we set for more junior devs or drive-by contributors: This sort of ban has been a cheaper alternative to rigorously testing every PR in every browser.

We've tended to use caniuse.com as the "source of truth" to guide us. At present, we don't have many instances of intentional 'graceful degradation' in the app - though I'm sure there are places of unintentional degradation, graceful or not!

For what it's worth, on Studio flexbox is allowed along with css variables, and css grid, and other more modern features. My general attitude with Kolibri has been much more conservative because of the unique target audience (folks for whom upgrading browsers is difficult).

Maybe we should take the temperature from the team at some point and see if this kind of blanket ban is more annoying or more helpful?

bjester: Well, if Can I Use is the source of truth and lists it as "partial support" with prefixes, I think you'll need to provide more criteria on what goes and what doesn't. At the moment, it seems a bit arbitrary to me.

There may be a bigger flexbox problem than you're aware... the Pure CSS used for `<KGrid>` uses flexbox!

indirectlylit: right - full support with prefixes was what I meant.

I believe purecss falls back to inline blocks when flexbox isn't available. Though looking at their github page they say Android 4.4+ which is a big surprise!

bjester:
@jonboiser I made a change that should fix the all gray theme buttons you experienced. If you have a moment, it would be great to see if that issue is resolved on your end.

jonboiser:
Yup, the "Theme" buttons are now styled correctly.

kolibri/plugins/document_epub_render/assets/src/views/LoadingScreen.vue
indirectlylit: I'll 'resolve' this thread since it's sort of tangential to this PR. We can either keep it or not, but some follow-up in https://github.com/learningequality/kolibri-docs/issues/95 is definitely warranted

radinamatic:
Tested this and vertical scroll for tabled ePUBs seem like a good alternative, great work!

The only issue that I found is that ToC, settings and search buttons do not close the respective sidebars with a mouse click, as they just shrink for a second, but remain open (both Chrome and FF on Ubuntu 18.04). On another hand, everything works fine when keyboard navigation is used (focus with TAB > first Enter opens > second Enter closes).

It seemed easier if I opened a PR on top of yours for the Gherkin scenarios, hope you don't mind!

bjester:
@radinamatic Thanks Radina! I must have broke something since I thought the toggles were working. I'm resolving that issue now.

indirectlylit (approved):
sweet

PR #5490, opened 2019-05-15
Importability Annotation

### LINKED ISSUE #3269: Import from disk - content shown to be available for import when it is not
### Instructions
**Instructions:**
Import content from a usb flash disk on Linux
### Observed behavior
The channels display seems OK:
http:/<lan>/management/device/#/content/wizard/availablechannels
It seems to show the correct mounted partition  where the KOLIBRI_DATA directory exists.
/mnt
It shows the correct number of channels exported.
7 channels available
import from mnt

However if a channel is selected, eg. Khan Academy, it does not show the content actually available on the flash disk. It seems to be displaying what one would see from the Central Server.
khan academy - shows nothing

-->

…

### Expected behavior

During import from flash disk, the content actually available on the flash disk should be shown. It seems to be displaying what one would see from the Central Server.

### User-facing consequences
Cannot transfer any content

…

### Errors and logs
The flash disk partition was mounted as follows
sudo mount /dev/sdb3 /mnt

```
$ lsblk 
NAME   MAJ:MIN RM   SIZE RO TYPE MOUNTPOINT
sda      8:0    0 232.9G  0 disk 
├─sda1   8:1    0   512M  0 part /boot
└─sda2   8:2    0 232.4G  0 part /
sdb      8:16   1  28.9G  0 disk 
├─sdb1   8:17   1   160M  0 part 
├─sdb2   8:18   1     5G  0 part 
└─sdb3   8:19   1  23.7G  0 part /mnt

-->]$ ls /mnt -l
total 5732
...
drwxr-xr-x 3 me me    4096 Feb 28 00:14 KOLIBRI_DATA
...
```

### Context
Tell us about your environment, including:
 * Kolibri version kolibri-v0.7.1.pex
 * Operating system
Linux kb.barana.net 4.14.21-1-MANJARO #1 SMP PREEMPT Fri Feb 23 00:01:40 UTC 2018 x86_64 GNU/Linux
 * Browser
Vivaldi 
1.14.1077.41 (Stable channel) (64-bit)

This was a fresh install. No content from the Central Server had been fetched.

### LINKED ISSUE #4038: Import - Does not present contents of Export.
### Observed behavior
When performing import, the select tree should present the contents of the exported file. Instead, it presents the contents of the whole channel.

Storage contains 109 MB

storage

The Global Digital Library, for which the English subtree  was completely exported, shows the whole tree without anything selected.  To be useful, the English subtree should be selected; that is, the actual contents of the export file.
 
f32c31a9-bca0-4d9b-b5b8-2062d7ce4b17

### Expected behavior
To be useful, the English subtree should be selected; that is, the actual contents of the export file.

### User-facing consequences
Confusion.
Inability to control what gets imported.

### Steps to reproduce
Perform an import from a local file

### Context
 * Kolibri version .10
 * Linux Manjaro
 * Vivaldi browser

### LINKED ISSUE #4194: scan for availability during import from local drive
### Observed behavior
When I import channel content from attached drive and I select the contents that's not at my local drive, it will start to download but after the download finished the channel local contents still the same.

### Expected behavior
Downloaded channel contents should show at the channel local contents. 

### Errors and logs
screen shot 2018-08-27 at 3 59 24 pm
This is after I selected the local drive channel import.
screen shot 2018-08-27 at 3 52 48 pm
This is after I select import 53mb channel content.
screen shot 2018-08-27 at 3 53 01 pm
The channel content still the same 43mb.

### Steps to reproduce
*  Download less than 1gb channel contents from the studio.
* Export all the channel contents to the local drive.
* Then select Import from local drive and select the channel that you recently exported.
* Select contents that is not on your device and import it. 

### Context

Tell us about your environment, including:
 * Kolibri version: 0.10.2
 * Operating system: Windows 7
 * Browser: Firefox 61.0.1

rtibbles:
### Summary
This PR primarily aims to address the current issue of content import not showing which content is actually available when importing from an external drive or from peer to peer content importing.

In order to achieve this, it makes some fairly large changes to how content import is handled on the backend (and to a lesser extent on the frontend).

Backend changes summary:
* Adds a *lot* of annotation fields to both ContentNode and ChannelMetadata to summarize different aspects of metadata that is useful for content import.
* Runs a per-import annotation that writes to the database in order to filter and return appropriate values to the frontend.
* No longer calculates resource counts and file size counts during API calls, and precomputes them instead.
* No longer gives accurate numbers for channels with duplicates from these precalculated values, in order to achieve efficiency gains.
* Summarizes the level of duplication of resources and file sizes on the channel.
* Summarizes deduplicated resources and file size on the channel.
* Collapses all tree recursion annotation parsing into a single function to avoid looping multiple times and reducing DB writes.
* Precalculates the num_coach_content field.
* *CHANGES THE MEANING OF IMPORTABILITY TO SIGNIFY THAT SOMETHING IS AVAILABLE FOR IMPORT, AND ALSO NOT CURRENTLY IMPORTED*

Frontend changes summary:
* Change messaging to handle times when we cannot give accurate counts due to resource duplication.
* Shows estimated resource counts and filesizes based on estimated counts and the overall channel ratio - *this behaviour may be misleading, and it might be better to just show the unmodified counts*.
* Always runs a task at channel load to annotate importability (on the backend this is cached so that returning to the same page will not rerun it if no changes have been made).
* Changes frontend behaviour to reflect that importability is now mutually exclusive with availability.
* Makes channel import based on files that can be imported, not resources, in case of new thumbnails etc.

### Reviewer guidance
Do all import/export workflows function as expected?
Does the count and file size estimation add more confusion or help with it? (the alternative is to still say it is estimated, but not attempt to reduce the counts).

### References

Fixes #4038
Fixes #4194
Fixes #3269

Some earlier design work here: https://docs.google.com/document/d/1KBLNejVu7dlhZdAodk59cEHQQYZtO7BNe5B6Xz7gi5s/edit#

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated


indirectlylit:
cc @radinamatic in case this needs updated user docs.

also adding a follow-up PR for gherkin stories: https://github.com/learningequality/kolibri/issues/5493

kolibri/core/content/management/commands/annotateimportable.py
```diff
@@ -0,0 +1,58 @@
...
+            "channel_id",
+            type=str,
+            help="Annotate importability on the database for the given channel_id.",
+        )
+
+        parser.add_argument(
+            "--baseurl",
+            type=str,
+            default="",
+            help="The host we will check importability of the content from.",
```
indirectlylit, line 32: is it a base URL or a host name?

kolibri/core/content/management/commands/annotateimportable.py
```diff
@@ -0,0 +1,58 @@
...
+                    )
+                progress_update(1, progress_extra_data)
+            elif options["directory"]:
+                annotate_importability_from_disk(
+                    options["channel_id"], options["directory"]
+                )
+                progress_update(1, progress_extra_data)
+            else:
+                annotate_importability_from_studio(options["channel_id"])
+                progress_update(1, progress_extra_data)
```
indirectlylit, line 58: maybe add a check to make sure that `baseurl` and `directory` are mutually exclusive

kolibri/core/content/models.py
```diff
@@ -215,6 +232,29 @@ class ContentNode(MPTTModel):
...
+    # source.
+    importable_file_size = models.BigIntegerField(default=0, null=True, blank=True)
+    # Total number of importable coach only resources for this node
+    importable_coach_contents = models.IntegerField(default=0, null=True, blank=True)
+    # Total count of on device resources for this node.
+    on_device_resources = models.IntegerField(default=0, null=True, blank=True)
+    # Total file size on disk for this node
+    on_device_file_size = models.BigIntegerField(default=0, null=True, blank=True)
+    # Total number of coach only resources for this node
+    num_coach_contents = models.IntegerField(default=0, null=True, blank=True)
```
indirectlylit, line 257: If I'm understanding correctly, this change assumes that only one import activity is occurring at a time. A simple counter-example:

* Open up the device import in two tabs
* In one tab, import from Studio and in the other tab, import from disk

Storing the metadata directly on the content model could make it more difficult to handle queued/concurrent imports and multi-device-admin scenarios.

Unsure how this would affect performance, but conceptually I could imagine a new `ContentNodeImportAvailability` table and create a concept of an "import session", defined by a user having the granular import UI open...

E.g.:

* `id`
* `content_node` (FK)
* `import_session` (FK)
* `importable` (bool)
* ... other session-specific attributes

kolibri/core/content/utils/annotation.py
```diff
@@ -159,6 +172,13 @@ def set_leaf_node_availability_from_local_file_availability(channel_id):
         .where(ContentNodeTable.c.id == FileTable.c.contentnode_id)
     )
 
+    contentnode_file_size_statement = (
+        select([func.coalesce(func.sum(LocalFileTable.c.file_size), 0)])
+        .select_from(FileTable.join(LocalFileTable))
+        .where(and_(FileTable.c.available == True))  # noqa
+        .where(ContentNodeTable.c.id == FileTable.c.contentnode_id)
+    )
```
indirectlylit, line 180: some comments might be helpful for posterity

kolibri/core/content/utils/importability_annotation.py
```diff
@@ -0,0 +1,437 @@
...
+that is used during the content import process - the starting point for annotation is the
+concept of 'importability' - the implementation here is differently defined from an
+historic use of 'importable' heretofore used, previously 'importable' signified only
+that the particular file in question was available to be imported (it was either assumed
+to be the case in the case of any remote import - or if the file was present on disk when
+importing from USB or an external hard drive).
+
+Henceforth, 'importable' is better understood in a more contextual light, relative to the
+current import operation that is being enacted by the user. Importable means precisely:
+A file is importable if and only if it exists to be imported on the remote server or external
```
indirectlylit, line 12: > A file is importable

Are we using the same terminology for files and for content nodes?

I wonder if there's a valuable distinction to make between "file availability" and "content node importability", since an individual file can not actually be "imported" per se?

kolibri/core/decorators.py
```diff
@@ -275,21 +275,43 @@ def query_params_required(**kwargs):
...
-
-        class Wrapper(cls):
-            def initial(self, request, *args, **kwargs):
+    def _params(view):
+        try:
+            assert issubclass(view, APIView)
+
+            class Wrapper(view):
+                def initial(self, request, *args, **kwargs):
+
```
indirectlylit, line 284: might be helpful to add some quick notes on how this all works from a mechanistic perspective. I rarely implement decorators so seeing functions returning wrapper classes inside try/excepts is all a bit confusing...

indirectlylit (commented):
this is exciting.

haven't tested yet - started with a high level scan of the code

kolibri/core/content/models.py
rtibbles: Yeah, that's my main misgiving about the current implementation - the reason that I did not go even further in this direction is because the content import process is already problematic with simultaneous users.

However, I think your suggestion is more scalable, because we can keep these multiple distinct annotations around for longer and don't have to rewrite them as much.

The other thing I like about this approach is that it doesn't bloat the ContentNodes with a bunch of unnecessary cruft for their core function. Will have a stab at this refactor within the scope of this PR, so that we can avoid having to add and then remove these fields in future.

kolibri/core/content/utils/importability_annotation.py
rtibbles: From a technical point of view, a file is the only thing that can be imported, and then a ContentNode can be imported by virtue of its files being importable - I think there probably is a slight distinction between a file being importable and a ContentNode being importable, but I am not sure I can define it.

indirectlylit:
updating milestone to 0.13.0 so this doesn't become a blocker for 0.12.4

kolibri/core/content/management/commands/annotateimportable.py
rtibbles: It's a base URL, I copied this from the importchannel command, so any errors have propagated.

kolibri/core/decorators.py
rtibbles: Yeah, I probably could have avoided doing this at all, but it seemed appropriate to use on the endpoint I created.

kolibri/core/content/utils/importability_annotation.py
indirectlylit: I meant that from a user's point of view, a content node is the thing being imported

rtibbles: Right, as far as the software is concerned only files can be imported, as far as users are concerned only contentnodes can be, so I'm stuck in the middle with both.

rtibbles:
Needs retargeting to develop - but will do when I do a rebase, post merge of #5497

rtibbles:
I think everything I can salvage from this PR has been salvaged.

indirectlylit:
it lived a good life

rtibbles:
I am not so sure about that.

radinamatic:
@indirectlylit Is it safe to remove the `needs user docs` label?

indirectlylit:
yes, thank you!

Issue #4652, opened 2018-12-22
Log all commands being run + arguments


benjaoming:
### Observed behavior

Getting logs for debugging purposes is a bit difficult, given that the logs do not capture when the user invokes other commands, `kolibri mange migrate ...` or `kolibri plugin ...` etc.

### Expected behavior

Add some `logging.info` call.

### User-facing consequences

Time-consuming explanations for support purposes.

### Errors and logs

n/a

### Steps to reproduce

Run a command like `kolibri manage showmigrations` and see that nothing is logged.

### Context

Community support, Github etc.

lyw07:
should be updated in #5495

Issue #2512, opened 2017-10-24
Update logging + implement log rotation


benjaoming:
### Observed behavior

When kolibri starts, the following message is received:

```
INFO     Going to daemon mode, logging to /home/benjamin/.kolibri/server.log
```

However `server.log` is not the main log anymore.

Currently, we have:

 * `kolibri.log` for general INFO level from Python
 * `debug.log` for general DEBUG level (if switched on?)
 * `server.log` for crash outputs when deamonized

Related issue: Icecube logging is noisy #2212

Changes to this have to be consolidated with docs.

### Expected behavior

I would suggest the following naming for an easier user experience and for UIs that want to integrate a log view (highly recommended)

 * `log/daemon.log` for sporadic crash output produced by the daemon and uncaught by other loggers
 * `log/kolibri.log` for the latest log output
 * `log/kolibri-YYYY-MM-DD.log` for log rotation (dates provided kindly by built-in log rotation, see KA Lite)
 * `log/debug.log` for the latest debug output, not rotated to avoid accumulating huge logs

Icecube issue: Possibly solved by only logging Icecube progress etc. as debug output and accepting that it's noisy.

### User-facing consequences

Makes it easier to dig out logs and finalizes how we want to do logging so we don't have to change it further down the road.

Readiness for UI implementation.

Means we can also amend the issue template regarding where to find the logs.

### Updates

 * ~Changed file extensions from `.log` to `.txt` -- will make it easier for users on Windows to open the files~. Moved: Follow-up from https://github.com/learningequality/kolibri/issues/2512
 * ~Suggestion to put rotated log files in `log/archive/` to reduce user confusion~ Not easy to implement, needs modification of built-in Python logging class.

lyw07:
~~One question I have is about the `log/daemon.log`. We currently don't make the process in macosx go daemon, so the name `daemon.log` may not cover all the platforms we have.~~

Never mind. I think I misunderstood the question.

rtibbles:
I wonder if it would be helpful to have a single log file, but then add rotation?

lyw07:
@rtibbles do you mean getting rid of the server.log file?

rtibbles:
Yeah, given that server.log doesn't contain much, I'm not sure I see a reason for keeping it separate, as long as the log messages are clearly differentiated?

lyw07:
[...]

PR #5494, opened 2019-05-15
Migrate CLI to Click, add Version Upgrade Logic

rtibbles:
### Summary
* Migrates the Kolibri CLI to click
* Adds the ability to do version constrained upgrade logic
* Cleans up accumulated CLI cruft to restrict startup tasks.

### Reviewer guidance
Does the CLI work as expected?
Does the version upgrade logic make sense, is it sufficiently well documented?

### References
Fixes #5675

----

### Contributor Checklist

PR process:

- [x] PR has the correct target branch and milestone
- [x] PR has 'needs review' or 'work-in-progress' label
- [x] If PR is ready for review, a reviewer has been added. (Don't use 'Assignees')
- [ ] If this is an important user-facing change, PR or related issue has a 'changelog' label
- [ ] If this includes an internal dependency change, a link to the diff is provided

Testing:

- [x] Contributor has fully tested the PR manually
- [ ] If there are any front-end changes, before/after screenshots are included
- [ ] Critical user journeys are covered by Gherkin stories
- [ ] Critical and brittle code paths are covered by unit tests

### Reviewer Checklist

- Automated test coverage is satisfactory
- PR is fully functional
- PR has been tested for [accessibility regressions](http://kolibri-dev.readthedocs.io/en/develop/manual_testing.html#accessibility-a11y-testing)
- External dependency files were updated if necessary (`yarn` and `pip`)
- Documentation is updated
- Contributor is in AUTHORS.md

PR #5495, opened 2019-05-16
rename all the log files and add log rotation to kolibri.txt

lyw07:
### Summary

This PR is to
1. rename the log files:
   -  server.log --> logs/daemon.txt
   - debug.log --> logs/debug.txt
   - kolibri.log --> logs/kolibri.txt  

2. put the logging messages from server.log and kolibri.log into kolibri.txt
3. add daily rotation to kolibri.txt and the rotation files are named as logs/archive/kolibri-YYYY-MM-DD.txt
4. add a check before the server starts to move the previous log files to the correct location
5. add more logs in cli.py so that the log files can capture what commands users have invoked for developers to better debug the issues

-----
The custom handler KolibriTimedRotatingFileHandler is to change the default rotation filename from logs/kolibri.txt.YYYY-MM-DD to logs/archive/kolibri-YYYY-MM-DD.txt

The custom formatter KolibriLogFileFormatter is to change the logging information of cherrypy.
For example, **without** the formatter, the logging of cherrypy would look like this:
```
INFO:_cplogging:[21/May/2019:15:59:02] ENGINE Bus STARTING
```
**with** the formatter, the logging looks like this:
```
INFO:cherrypy.error: ENGINE Bus STARTING
```
----
I have tested the following situations:
1. log rotation on windows
2. log rotation on macosx
3. log rotation on ubuntu
4. when there are existing log files with the old name conventions, they will be moved to the new location

### Reviewer guidance

please check if the code makes sense, and the new logging works. Thank you!

### References

https://github.com/learningequality/kolibri/issues/2512
https://github.com/learningequality/kolibri/issues/4652

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/deployment/default/settings/base.py
+++ b/kolibri/deployment/default/settings/base.py
@@ -268,6 +268,10 @@ LOGGING = {
         },
         "simple": {"format": "%(levelname)s %(message)s"},
         "simple_date": {"format": "%(levelname)s %(asctime)s %(module)s %(message)s"},
+        "simple_date_file": {
+            "()": "kolibri.core.logger.utils.formatter.KolibriLogFileFormatter",
+            "format": "%(levelname)s %(asctime)s %(module)s %(message)s",
+        },
         "color": {
             "()": "colorlog.ColoredFormatter",
             "format": "%(log_color)s%(levelname)-8s %(message)s",
@@ -305,15 +309,17 @@ LOGGING = {
             "level": "DEBUG",
             "filters": ["require_debug_true"],
             "class": "logging.FileHandler",
-            "filename": os.path.join(conf.KOLIBRI_HOME, "debug.log"),
+            "filename": os.path.join(conf.LOG_ROOT, "debug.txt"),
             "formatter": "simple_date",
         },
         "file": {
             "level": "INFO",
             "filters": [],
-            "class": "logging.FileHandler",
-            "filename": os.path.join(conf.KOLIBRI_HOME, "kolibri.log"),
-            "formatter": "simple_date",
+            "class": "kolibri.core.logger.utils.handler.KolibriTimedRotatingFileHandler",
+            "filename": os.path.join(conf.LOG_ROOT, "kolibri.txt"),
+            "formatter": "simple_date_file",
+            "when": "midnight",
+            "backupCount": 30,
         },
     },
     "loggers": {
@@ -338,6 +344,8 @@ LOGGING = {
             "level": "INFO",
             "propagate": False,
         },
+        "cherrypy.access": {"handlers": ["file"], "level": "INFO", "propagate": False},
+        "cherrypy.error": {"handlers": ["file"], "level": "INFO", "propagate": False},
     },
 }
 
--- a/kolibri/deployment/default/settings/base.py
+++ b/kolibri/deployment/default/settings/base.py
@@ -344,8 +344,16 @@ LOGGING = {
             "level": "INFO",
             "propagate": False,
         },
-        "cherrypy.access": {"handlers": ["file"], "level": "INFO", "propagate": False},
-        "cherrypy.error": {"handlers": ["file"], "level": "INFO", "propagate": False},
+        "cherrypy.access": {
+            "handlers": ["file", "console"],
+            "level": "INFO",
+            "propagate": False,
+        },
+        "cherrypy.error": {
+            "handlers": ["file", "console"],
+            "level": "INFO",
+            "propagate": False,
+        },
     },
 }
 
--- a/kolibri/utils/cli.py
+++ b/kolibri/utils/cli.py
@@ -316,7 +316,10 @@ def start(port=None, daemon=True):
     # Daemonize at this point, no more user output is needed
     if daemon:
 
-        logger.info("Going to daemon mode, logging to {0}".format(server.KOLIBRI_LOG))
+        from django.conf import settings
+
+        kolibri_log = settings.LOGGING["handlers"]["file"]["filename"]
+        logger.info("Going to daemon mode, logging to {0}".format(kolibri_log))
 
         kwargs = {}
         # Truncate the file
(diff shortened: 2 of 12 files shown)


lyw07:
PR to update the user doc is referenced above ^

kolibri/utils/sanity_checks.py
```diff
@@ -15,6 +15,9 @@
 
 
 def check_other_kolibri_running(port):
+    """
+    Make sure there are no other Kolibri instances running before starting the kolibri.
```
radinamatic: Second Kolibri was not capitalized, but maybe skip the repetition?
```Make sure there are no other Kolibri instances running before starting the server.```

Comment (radinamatic) [reply kolibri/utils/sanity_checks.py]: Òh, this is code comment, and not a message for the user...? :woman_facepalming:

File: kolibri/deployment/default/settings/base.py [hunk 2]
```diff
@@ -338,6 +344,8 @@
             "level": "INFO",
             "propagate": False,
         },
+        "cherrypy.access": {"handlers": ["file"], "level": "INFO", "propagate": False},
```
rtibbles: Is there a reason not to log this to console in parallel to morango and iceqube?

kolibri/utils/conf.py
```diff
@@ -47,6 +47,11 @@
         )
     os.mkdir(KOLIBRI_HOME)
 
+# Create a folder named logs inside KOLIBRI_HOME to store all the log files.
+LOG_ROOT = os.path.join(KOLIBRI_HOME, "logs")
+if not os.path.exists(LOG_ROOT):
+    os.mkdir(LOG_ROOT)
```
rtibbles, line 53: Just to double check, this should happen before we do anything with options that might trigger its logger, right?

kolibri/utils/sanity_checks.py
```diff
@@ -60,3 +69,28 @@ def check_content_directory_exists_and_writable():
             "The content directory {} is not writable.".format(content_directory)
         )
         sys.exit(1)
+
+
+def check_log_file_location():
```
rtibbles, line 74: This is good - can you add a follow up issue to migrate this to a version specific upgrade once my PR (#5494) to allow that is merged (for 0.13.0)?

kolibri/utils/server.py
```diff
@@ -52,7 +52,9 @@
 
 # This is a special file with daemon activity. It logs ALL stderr output, some
 # might not have made it to the log file!
-DAEMON_LOG = os.path.join(conf.KOLIBRI_HOME, "server.log")
+DAEMON_LOG = os.path.join(conf.LOG_ROOT, "daemon.txt")
+
+KOLIBRI_LOG = settings.LOGGING["handlers"]["file"]["filename"]
```
rtibbles: This looks wrong - we are accessing Django settings in the module root here. The server module is imported in the module scope of the cli module, which is executed before Django is initialized. I am a little surprised this has not caused breakage.

Given that this value is only referenced in the `cli.py` file, I feel like we could do without this extra indirection, and just reference the value directly in `cli.py`.

kolibri/utils/cli.py
```diff
@@ -422,11 +424,12 @@ def services(daemon=True):
     # Daemonize at this point, no more user output is needed
     if daemon:
 
+        logger.info("Going to daemon mode, logging to {0}".format(server.KOLIBRI_LOG))
```
rtibbles: Best to just directly import the settings here and reference the logging filename.

kolibri/utils/cli.py
```diff
@@ -315,11 +316,12 @@ def start(port=None, daemon=True):
...
```
rtibbles: Same as below.

rtibbles (changes requested):
One blocker, a nice to have, and some questions.

kolibri/utils/sanity_checks.py
lyw07: oh but thank you for pointing it out! I will change it in another commit

kolibri/deployment/default/settings/base.py
lyw07: no, I was unsure about that so I only put them to "file". I will log the cherrypy messages to console.

kolibri/utils/conf.py
lyw07: yes, because we set up some basic logging in the options I believe.

kolibri/utils/sanity_checks.py
lyw07: yes created it here: https://github.com/learningequality/kolibri/issues/5532

kolibri/utils/server.py
lyw07: I wonder if it's because we have configured `DJANGO_SETTINGS_MODULE` to `kolibri.deployment.default.settings.base` in `conf.py`, we are just getting the variable LOGGING from the settings.base module?
But yeah I will reference this in cli.py directly

rtibbles: When you do `from django.conf import settings` - the `settings` object is not actually the direct settings, but a Django `LazySettings` object. As soon as you try to access a key from the settings object, it will check that it is initialized, if it is not, it will throw an error. That's why I was confused!

What will probably be good in the future is actually to move our logging configuration out of the settings `base.py` and into a separate file so that we can reference it independently of the settings (and also avoid the repetition in `options.py` too). This will have the slight issue of causing problems if someone completely overrides our base settings, but that's their problem, not ours!

lyw07: ah it makes sense to me now. thank you so much for the explanation!

rtibbles (approved):
Code changes look good, log files rotate, and @radinamatic confirms that the installed asset works as expected.

radinamatic:
DEB asset installed on Ubuntu 16.04, seems to be running & importing OK. 

I haven't gone into any deep testing, just on the first glance `daemon.txt` has some additional chars at the beginning and end of each line:

Ubuntu16 04 (start)  Running  - Oracle VM VirtualBox_037

rtibbles:
The chars at the beginning of the line was existing behaviour afaicr.

radinamatic:
Gotcha, not major anyway!

PR #5497, opened 2019-05-16
Content and database related fixes

rtibbles:
### Summary
* Fixes a critical bug that prevents unmigrated databases from migrating (because it tries to access the database at startup, before the relevant table exists).
* Allows the generation of a 'current' content DB schema, based on the current tables from the Kolibri content app - which may have fields not present in a published Content Database (because they are local annotation fields).
* Tests and fixes a subtle bug in our content import logic, whereby a file belonging to an excluded content node would be excluded from import even if a contentnode that depended on it was explicitly included.
* Adds `num_coach_content` as an explicit annotated field on ContentNodes so that it no longer has to be dynamically calculated - also removes the `include_fields` logic that was triggering this behaviour, and references in the frontend that were using it. (the dynamic calculation on ContentNodeGranular endpoint is still used as it has to recalculate for import vs export).
* Caches mounted drive info for use when querying an individual drive's information by id - previously we would scan all drives to get the list of drives, then scan all drives to retrieve the information for one drive. Now it is cached from the previous scan and used if available, otherwise a new scan is initiated.
* Gets rid of the `by_role` Content Node filter, and replaces it with `user_kind` so that the returned data does not depend on user session data, and will not lead to weird caching behaviour.

### Reviewer guidance
Does everything still work?

### References
Mostly cherry picked from #5490 

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated


indirectlylit (approved):
did some cursory testing and everything seems to be fine. However I did not test with external drives, so this should be one major initial focus of initial integration testing in beta1 (cc @mrpau-dev)

PR #5506, opened 2019-05-17
Revert "Run integration testing script inside the docker environment"

indirectlylit:
Reverts learningequality/kolibri#5313


indirectlylit:
will force push instead

PR #5510, opened 2019-05-18
Add version number to wsgi start

rtibbles:
### Summary
We currently log that Kolibri is starting, but not what version number it is. This can make debugging difficult when looking at logs.

This change adds the version number to the logs to help with debugging in the future.

### Reviewer guidance
Does the server still start?

### References
Cherry picked from #5494

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/deployment/default/wsgi.py
+++ b/kolibri/deployment/default/wsgi.py
@@ -13,6 +13,8 @@ import time
 from django.core.wsgi import get_wsgi_application
 from django.db.utils import OperationalError
 
+import kolibri
+
 os.environ.setdefault(
     "DJANGO_SETTINGS_MODULE", "kolibri.deployment.default.settings.base"
 )
@@ -23,7 +25,7 @@ tries_remaining = 6
 interval = 10
 while not application and tries_remaining:
     try:
-        logger.info("Starting Kolibri")
+        logger.info("Starting Kolibri {version}".format(version=kolibri.__version__))
         application = get_wsgi_application()
     except OperationalError:
         # an OperationalError happens when sqlite vacuum is being executed. the db is locked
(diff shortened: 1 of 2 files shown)


kolibri/deployment/default/wsgi.py
```diff
@@ -23,7 +25,7 @@
 interval = 10
 while not application and tries_remaining:
     try:
-        logger.info("Starting Kolibri")
+        logger.info("Starting Kolibri {version}".format(version=kolibri.__version__))
```
jredrejo: logger.info does not work inside uwsgi, replacing it by print, will output the message in uwsgi.log

kolibri/deployment/default/wsgi.py
```diff
@@ -13,6 +13,8 @@
 from django.core.wsgi import get_wsgi_application
 from django.db.utils import OperationalError
 
+import kolibri
```
jredrejo: Following our current policy this import should be at the beginning of the file.

jredrejo (dismissed):

kolibri/deployment/default/wsgi.py
rtibbles: Hrm, the issue I was trying to solve is that this message appears in `kolibri.log` but with no useful version information.
Perhaps it would be better just to move this message into `server.py`'s `start` function instead?

kolibri/deployment/default/wsgi.py
rtibbles: Good point - possible I didn't have pre-commit installed when I made this commit.

kolibri/deployment/default/wsgi.py
jredrejo: yes, then we could use logger and it will appear in kolibri.log.
In any case uwsgi.log contains the traces of any python error when using uwsgi. Maybe it's good to have it there too, in case this file is provided (By the way, all the `logger.info` lines in this wsgi.py archive are useless, so you can use this PR to either remove them or replace them by `print`)

indirectlylit (approved):
thank you!

indirectlylit (approved):

PR #5512, opened 2019-05-19
Annotate num coach contents at startup

### LINKED ISSUE #5511: num_coach_contents is not annotated on existing channels
### Observed behavior
Since the merging of #5497 num_coach_contents is no longer a dynamically calculated field, and instead is calculated on channel import and annotation.

This means that the result will be 0 for any content that was imported prior to 0.12.4.

### Expected behavior
num_coach_contents should accurately represent the number of coach contents available.

### User-facing consequences
Minor confusion.

rtibbles:
### Summary
Annotates num_coach_contents at startup to ensure it is properly annotated after being added as an explicit field.

### Reviewer guidance
Does server startup take drastically longer? Do previously imported channels show the correct number of coach contents?

### References
Fixes #5511 

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/core/content/utils/annotation.py
+++ b/kolibri/core/content/utils/annotation.py
@@ -161,11 +161,10 @@ def update_num_coach_contents():
         ContentNodeTable.update()
         .where(
             # That are not topics
-            ContentNodeTable.c.kind != content_kinds.TOPIC,
+            ContentNodeTable.c.kind
+            != content_kinds.TOPIC
         )
-        .values(
-            num_coach_contents=cast(ContentNodeTable.c.coach_content, Integer())
-        )
+        .values(num_coach_contents=cast(ContentNodeTable.c.coach_content, Integer()))
     )
 
     # Expression to capture all available child nodes of a contentnode
--- a/kolibri/core/content/utils/annotation.py
+++ b/kolibri/core/content/utils/annotation.py
@@ -137,8 +137,8 @@ def fix_multiple_trees_with_id_one():
 
 def update_num_coach_contents():
     """
-    Function to set num_coach_content on topic trees that were imported before
-    the annotation would have been run.
+    Function to set num_coach_content on all topic trees to account for
+    those that were imported before annotations were performed
     """
     bridge = Bridge(app_name=CONTENT_APP_NAME)
 
--- a/kolibri/core/content/utils/annotation.py
+++ b/kolibri/core/content/utils/annotation.py
@@ -70,6 +70,7 @@ def update_channel_metadata():
                     )
                 )
     fix_multiple_trees_with_id_one()
+    update_num_coach_contents()
     connection.close()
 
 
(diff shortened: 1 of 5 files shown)


kolibri/core/content/utils/annotation.py
```diff
@@ -134,6 +135,93 @@ def fix_multiple_trees_with_id_one():
...
+
+    child = ContentNodeTable.alias()
+
+    logger.info("Updating num_coach_content on existing channels")
+
+    # start a transaction
+
+    trans = connection.begin()
+
+    for channel_id in ChannelMetadata.objects.all().values_list("id", flat=True):
```
indirectlylit, line 187: I'm a little confused - does this operate on all channels, or just the ones "that were imported before the annotation would have been run"?

kolibri/core/content/utils/annotation.py
```diff
@@ -134,6 +135,93 @@ def fix_multiple_trees_with_id_one():
...
+                    # In this channel
+                    ContentNodeTable.c.channel_id == channel_id,
+                    # That are not topics
+                    ContentNodeTable.c.kind != content_kinds.TOPIC,
+                )
+            )
+            .values(
+                num_coach_contents=cast(ContentNodeTable.c.coach_content, Integer())
+            )
+        )
```
indirectlylit: If this is being run for all channels anyway, couldn't this update block be moved outside of the `for` loop and made into a single DB update?

```python
# Update all leaf ContentNodes to have num_coach_content to 1 or 0
connection.execute(
    ContentNodeTable.update()
    .where(ContentNodeTable.c.kind != content_kinds.TOPIC)
    .values(num_coach_contents=cast(ContentNodeTable.c.coach_content, Integer()))
)
```

kolibri/core/content/utils/annotation.py
```diff
@@ -134,6 +135,93 @@ def fix_multiple_trees_with_id_one():
...
+    """
+    bridge = Bridge(app_name=CONTENT_APP_NAME)
+
+    ContentNodeClass = bridge.get_class(ContentNode)
+
+    ContentNodeTable = bridge.get_table(ContentNode)
+
+    connection = bridge.get_connection()
+
+    child = ContentNodeTable.alias()
```
indirectlylit, line 151: what kind of object is this, and what's it a child of? I see 

```python
    def get_table(self, DjangoModel):
        return self.get_class(DjangoModel).__table__
```

but I can't find documentation for `Model.__table__` or `Model.__table__.alias`

kolibri/core/content/utils/annotation.py
```diff
@@ -134,6 +135,93 @@ def fix_multiple_trees_with_id_one():
...
+        # Go from the deepest level to the shallowest
+        for level in range(node_depth, 0, -1):
+
+            # Expression to capture all available child nodes of a contentnode
+            available_nodes = select([child.c.available]).where(
+                and_(
+                    child.c.available == True,  # noqa
+                    ContentNodeTable.c.id == child.c.parent_id,
+                )
+            )
```
indirectlylit: can this `available_nodes` calculation be moved outside the `for` loop? Seems that it wouldn't change with each iteration up the tree levels

kolibri/core/content/utils/annotation.py
rtibbles: It is a SQLAlchemy table object. https://docs.sqlalchemy.org/en/13/orm/extensions/declarative/table_config.html#using-a-hybrid-approach-with-table

kolibri/core/content/utils/annotation.py
rtibbles: This operates on all channels, because we have no way of knowing (currently) when a channel was originally imported.

In 0.13 we will be able to use the version upgrade logic to restrict this to only happen on upgrades from before 0.12.4.

kolibri/core/content/utils/annotation.py
rtibbles: Yes - I just copied it directly from the normal logic and didn't optimize.

kolibri/core/content/utils/annotation.py
rtibbles: Yes - by itself its just a statement that is used elsewhere, so it can be put outside the loop.

rtibbles:
Updated based on comments.

kolibri/core/content/utils/annotation.py
indirectlylit: okay yeah, that's what I thought. I was just confused by the comment at the top:

> Function to set num_coach_content on topic trees that were imported before the annotation would have been run.

Maybe clarify to

> Function to set num_coach_content on all topic trees to account for those that were imported before annotations were performed

kolibri/core/content/utils/annotation.py
indirectlylit: cool thanks. could be helpful to someone in the future to drop that link as a comment in `get_table`

indirectlylit (commented):
Looks like a couple test failures, though some might be spurious.

@bjester would you mind giving this a quick test and merge?

kolibri/core/content/utils/annotation.py
rtibbles: I never dared to dream that someone else might want to touch this code, I will dream!

rtibbles:
I think it needs a rebase, some issues occurring from changes upstream.

indirectlylit:
cc @radinamatic for testing

rtibbles:
Now rebased, the failing test has been fixed.

bjester:
Testing these changes.

bjester:
I tested this with a mere 376 coach contents and it was instant on my laptop. I could maybe test this on a slower device in the office tomorrow, but I'm guessing this will be fine performance-wise.

indirectlylit (approved):
sounds like we're good to go!

Issue #5057, opened 2019-02-14
string cleanup and refactoring


indirectlylit:
### Observed behavior

0.12.0 has quite a bit of string-related tech debt:

* `crossComponentTranslator` was used/abused to allow components to use translated strings from other components
* A subset of strings is now in a 'common coach strings' file
* A subset of strings now uses more precise naming conventions to distinguish different parts of speech - for example `previewLabel` (a noun) vs `previewAction` (a verb)
* Many strings defined are no longer used

### Expected behavior

* [x] We should add four more shared string files: one for core, and one for each of the plugins (learn, facility, device)
* [x] We should remove all uses of `crossComponentTranslator` by either moving the string to the new component or moving the string to a shared file
* [x] We should move common strings to the new shared string files and give them explicit, conventional names that reflect their part of speech.
* [x] We should create a new linting rule that warns about unused strings

Note on common strings:

First, we should remove the use of computed props and mixins. Instead, we should use `$formatMessage(namespace, stringID)`. [See discussion](https://learningequality.slack.com/archives/CB37UM23A/p1557769482282900)

Second, we should be judicious about what strings are put in 'common' files because over-doing this can have adverse affects. In particular, sometimes strings that are the same in English actually need to be translated differently in other languages. For example:

* "preview" the verb vs "preview" the noun
* adjectives that need to agree with the thing they're modifying
* words that are context-specific, and just happen to be the same in English

Therefore, we should not put a string in common _unless_ it is:

1. used at least 3 times
2. used in exactly the same way (label, button, etc)
3. used to mean exactly the same thing. All strings added to common files should follow the naming conventions in `commonCoachStrings`

### User-facing consequences

none, but affects the costs and difficulty for devs and transators

### Context

0.12.0

jonboiser:
I would add too that the amount of extraneous data in the DevTools has gone up a lot, making them harder to read for me

1. Extra `computed` items within Components coming from mixins
1. About 20-30 getters for kolibriCoreTheme, one for almost each color and some styles
1. The SET_MODALITY mutation appears in the Vuex feed on every blur (e.g. just clicking a button).

I would like to see some effort in reducing the amount of extra data in devtools.

indirectlylit:
yeah, it's true. Do `$tr` strings show up in devtools though?

rtibbles:
We could use the new Vue observable method to manage some of the watched state (like the modality) without having to put it in the store. Could do a similar refactor to the dynamic styles as well.

[...]

PR #5526, opened 2019-05-21
[WIP] Refactor $trs

nucleogenesis:
### Summary

This WIP PR is working toward refactoring the Kolibri use of $trs to directly address https://github.com/learningequality/kolibri/issues/5057.

**Results**

[Here is a Google Sheet with the latest profiled data.](https://docs.google.com/spreadsheets/d/1_qs5a8g0Rz-dWPiJ5vL9mQoDfhZIwN64eNsohn14KX8/edit#gid=511677294)

Rows/strings highlighted blue are those which I believe could be pulled into a Common file.

_Notes on meanings of each column:_

- `STRING` - The literal text string.
- `DEFINITION` - Is this _row_ displaying data about the namespace+keys where the `STRING` is defined?
- `NAMESPACE` - The namespace in which the row is used or defined.
- `KEY` - The key (aka, param passed to `$tr('key')`) referring to this row's string.
- `COMMON?` - This is only relevant for those with `DEFINITION = NO` - this indicates if the string was used via a Common*Strings mixin. (Currently only the commonCoachStrings mixin's `common$tr()` method added in this PR.
- `PATHNAME` - The exact file where this Namespace+Key was used to refer to the String in this row.

**Issues & Todo**
- [x] Make this work with a relative file path - it's currently dumping CSVs to a hard-coded local path to Jacob's machine.
- [x] The plugin runs when you run `yarn makemessages` and is piggy-backing on the `extract$trs.js` so it'll need its own command and configuration in Webpack
- [ ] A "Use" is basically skipped when the param in $tr() is a variable rather than a string. I could revise it to try to find possible values for the variable given to the $tr() call but it may not be entirely necessary.

**Plan**
- [x] Refactor how common strings are applied, starting with the commonCoachStrings.
- [x] Refactor all crossComponentTranslator calls so that strings they are trying to use from other components are instead defined in the component where crossComponentTranslator is used.
- [x] Then profile the usage of all literal string translations - what namespaces are they used in and with what keys.
- [ ] Using the profiled data, address issues concerning which strings should move to common files, which are never used and can be removed, and which might need to be renamed to better indicate their proper usage.
- [ ] Where it is found to be necessary, create common string files for Learn, Core, Facility, and Device similar to the existing commonCoachStrings.js implementation.
- [ ] Implement a linting warning when a string goes unused altogether.

### Reviewer guidance

TBD

### References

Addressing this issue: https://github.com/learningequality/kolibri/issues/5057

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated


radinamatic:
@indirectlylit At some point, after @nucleogenesis has advanced enough with this, it might be beneficial to make a test branch on Crowdin and upload the strings from here, to see how this translates in the wild :wink:

indirectlylit:
>  it might be beneficial to make a test branch on Crowdin and upload the strings from here

absolutely

indirectlylit:
One quick note on the current audit output:

We'll need to be able to compare strings across plugins and core, not just within them. For example, the words "Cancel" and "Save" are used in multiple plugins, and we'll want to consolidate those into a core string.

nucleogenesis:
> One quick note on the current audit output:
> 
> We'll need to be able to compare strings across plugins and core, not just within them. For example, the words "Cancel" and "Save" are used in multiple plugins, and we'll want to consolidate those into a core string.

@indirectlylit I added a new sheet to that Google Sheets document called "Full Profile" which is organized similarly in that all definitions are listed, then all uses for any given "String of text".

However - now the DEFINITION? column shows nothing for uses, but shows the module in which that string is defined for definitions.

kolibri/plugins/coach/assets/src/views/common/Recipients.vue
```diff
@@ -17,36 +17,19 @@
...
-    mixins: [coachStringsMixin],
     props: {
       groupNames: {
         type: Array,
         required: true,
       },
-      hasAssignments: {
-        type: Boolean,
-        required: true,
-      },
```
indirectlylit: looks like `hasAssignments` and `components` were mistakenly deleted?

kolibri/plugins/learn/assets/src/views/LearnIndex.vue
```diff
@@ -53,7 +52,9 @@
   import LearnTopNav from './LearnTopNav';
   import { ASSESSMENT_FOOTER, QUIZ_FOOTER } from './footers.js';
 
-  const RecommendedSubpageStrings = crossComponentTranslator(RecommendedSubpage);
+  // Bottom toolbar is 111px high on mobile, 113px normally.
+  // We reserve the smaller number so there is no gap on either screen size.
+  const BOTTOM_SPACED_RESERVED = 111;
```
indirectlylit: was this added by accident?

kolibri/plugins/coach/assets/src/views/plan/CreateExamPage/CreateExamPreview.vue
```diff
@@ -146,8 +145,6 @@
   import QuestionListPreview from './QuestionListPreview';
   import CreateExamPage from './index';
 
-  const createExamPageStrings = crossComponentTranslator(CreateExamPage);
```
indirectlylit, line 149: need to also remove `CreateExamPage` import

kolibri/plugins/coach/assets/src/views/plan/CoachExamsPage/index.vue
```diff
@@ -146,6 +146,15 @@
         return { name: PageNames.EXAM_CREATION_ROOT };
       },
     },
+    methods: {
+      genRecipientsString(groups) {
```
indirectlylit: unused method

indirectlylit (commented):
recommend running `yarn run lint-frontend:format`

kolibri/plugins/coach/assets/src/views/common/commonCoachStrings.js
```diff
@@ -131,6 +131,7 @@ const coachStrings = createTranslator('CommonCoachStrings', {
   // formatted values
   integer: '{value, number, integer}',
   combinedLabel: '{firstItem} / {secondItem}',
+  nthExerciseName: '{ name } ({number, number, integer})',
```
indirectlylit: for consistency this can be written `'{name} ({number, number, integer})'`

kolibri/plugins/coach/assets/src/views/common/QuestionDetailLearnerList.vue
```diff
@@ -2,7 +2,7 @@
 
   <div :style="{ backgroundColor: $themeTokens.surface }">
     <h3 class="header">
-      {{ coachStrings.$tr('learnersLabel') }}
+      {{ common$tr('learnersLabel') }}
```
indirectlylit: I don't actually remember the thinking behind renaming `coachStrings` to ` common$tr`. Do you recall?

Perhaps for the sake of this PR we should leave the naming as-is just to make it a smaller diff? The when we add new common files for core and other plugins, we can decide on final naming conventions.

indirectlylit (approved):
pending some linting fixes, this looks good to merge. We can do further iteration in another PR.

Great job getting rid of all the cross component translator usage!

kolibri/plugins/coach/assets/src/views/common/QuestionDetailLearnerList.vue
nucleogenesis: The reason is that we've rewritten the mixin altogether.

So - now the only thing being mixed in is a method `common$tr()` - which takes a key and returns the proper value from the coachStrings object in the commonCoachStrings file.

All of the common files would then have that mixin - might need to rename the mixed-in method when adding a commonCoreStrings file/mixin - or just use `commonCore$tr()` as the core mixin's method name.

packages/kolibri-tools/lib/profile$trs.js
```diff
@@ -0,0 +1,526 @@
...
+               *
+               * It's a string - which is what espree wants anyway - so we can strip the `with()`
+               * altogether, leaving us with a stringified array containing everything we need to
+               * create a thorough AST of the compiled Vue template.               *
+               */
+              const render = template.render.replace(/^.{18}|.{1}$/g, '');
+              ast = espree.parse(render, {
+                ecmaVersion: '2018',
+                sourceType: 'module',
+                ecmaFeatures: { jsx: true, templateStrings: true },
```
nucleogenesis: Worth testing this without the JSX: true - I kind of got squirrely with the options passed to espree while trying to figure out how to get it to parse the render JS without errors.

kolibri/plugins/learn/assets/src/views/LearnIndex.vue
nucleogenesis: Not sure exactly how that worked out as something that I added - perhaps I kept it by mistake while rebasing or something.

PR #5570, opened 2019-05-28
Use substitutions for repeated versions in release_process.rst, update version to something recent

jonboiser:
### Summary

1. Replaces some repeated references to version numbers, git tags, git branches, etc. using RST substitutions in `release_process.rst`
1. Flags in findable comment the (5) current parts of `release_process.rst` where some manual editing is needed to keep consistent with the substitutions.
1. Updates the variables so the documented scenario is going from version `0.13` to `0.14` (the upcoming change)`.
1. Fix some typos, and add a diff code example in the section about making a new release.

### Reviewer guidance

Check that the docs are still correct, but with the new version numbers.

### References

…

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/docs/i18n.rst
+++ b/docs/i18n.rst
@@ -98,7 +98,7 @@ Dynamic values are passed into translation strings as named arguments in an obje
 .. code-block:: javascript
 
   export default {
-    name: 'anothetComponent',
+    name: 'anotherComponent',
     mounted() {
       // outputs 'Henry read 2 stories'
       console.log(this.$tr('msg', {name: 'Henry', count: 2}));
--- a/docs/release_process.rst
+++ b/docs/release_process.rst
@@ -71,7 +71,6 @@ Test to ensure that changes have propagated as expected.
 Create a release branch
 -----------------------
 
-.. UNSUBSTITUTED VERSION NUMBERS IN THIS SECTION
 
 When we're nearing the end of major new feature development, cut a new release branch. If we're getting ready to release version |next_version|, we'd do the following steps:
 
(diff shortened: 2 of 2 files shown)


indirectlylit:
This is a cool idea.

Did you by any chance try parameterizing it with fewer variables? e.g. if you could just set

```rst
.. |this_version| replace:: ``13``
.. |next_version| replace:: ``14``
```

then presumably everything else could be generated like...

```rst
Patch releases (e.g. '0.|next_version|.1') follow the same process
outlined below, except that development occurs exclusively on an
existing release branch (e.g. release-v0.|this_version|.x). 
```

My concern with the large number of variables and manual substitutions is that it will make it more likely for people to introduce errors into the documentation either when editing the docs for clarity or bumping the example versions. This might outweigh the benefit of having the example version numbers closer to the current Kolibri version numbers.

jonboiser:
Unfortunately, one cannot do nested markup in RST, as far as I can tell. So, you can't substitute inside something in the double-backticks. If we didn't have all of the git branches/tags/etc. inside double-backticks, we could simplify things by just providing the two version numbers as a substitution.

jonboiser:
Yeah, I don't think there's a elegant way of doing this in RST. I think, though for managing actual releases, we should try to figure a way to automate getting all the tags, etc. in the Kanban cards and instructions correct and consistent.

And maybe once we reach a large enough number or this page gets a major edit, we can up the number, since, as a reader, it does make the examples easier to scan if they are closer to the actual values I might be working with.

Issue #5567, opened 2019-05-28
Learn page - coach-content icons and annotations are shown to learners


indirectlylit:
### Observed behavior

reported by @jonboiser - learners see coach annotations even though they have no access to that content:

| card | user |
|--|--|
|  ![image](https://user-images.githubusercontent.com/2367265/58511486-be807580-814f-11e9-9cbf-761e220568ad.png) |  ![image](https://user-images.githubusercontent.com/2367265/58511510-cdffbe80-814f-11e9-9bc0-eccbe7644007.png) |

### Expected behavior

Since the coach content is not accessible to learners, these annotations should be hidden completely

### User-facing consequences

meaningless/confusing extra information shown to learners

### Steps to reproduce

* import a channel with a combination of coach and learner content, such as the epub testing channel (`9096a64765c64af18519a9d9509e64a9`)
* view as learner

### Context

0.12.3

rtibbles:
Ah, this is probably because of the simplifications I made, where coach content info is always sent to the frontend, so the indicator just needs to check the user role to see if it should display.

jonboiser:
This is fixed in #5571. The coach-content icon (a blue person reading a book) is only shown on cards if the `session` indicates that the viewer is authenticated and not a learner (meaning they are at least a coach or higher).

PR #5571, opened 2019-05-28
Hide CoachContent annotations from Learners

nucleogenesis:
### Summary
Fixes issue in https://github.com/learningequality/kolibri/issues/5567 where coach content annotation counts were shown to Learners.

### Reviewer guidance
- Have a channel that includes coach content.
- In release-v0.12.y - log in as a Learner and go to that channel in Learn. See the blue icon with a count.
- Checkout this branch - restart server - see that the icon is hidden from the Learner.
- Log out - then go to the same content as a guest user to see that it is also hidden from non-logged in Guest users.
- As a coach, admin, etc you will still see the annotations.

### References
https://github.com/learningequality/kolibri/issues/5567

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/plugins/learn/assets/src/views/ContentCard/index.vue
+++ b/kolibri/plugins/learn/assets/src/views/ContentCard/index.vue
@@ -27,7 +27,7 @@
       </p>
       <div class="footer">
         <CoachContentLabel
-          v-if="hasUserPermissions && !isLearner"
+          v-if="userHasPermissions && !isLearner"
           class="coach-content-label"
           :value="numCoachContents"
           :isTopic="isTopic"
@@ -122,7 +122,7 @@
       },
     },
     computed: {
-      ...mapGetters(['isLearner', 'hasUserPermissions']),
+      ...mapGetters(['isLearner', 'userHasPermissions']),
       isTopic() {
         return this.kind === ContentNodeKinds.TOPIC || this.kind === ContentNodeKinds.CHANNEL;
       },
--- a/kolibri/plugins/learn/assets/src/views/ContentCard/index.vue
+++ b/kolibri/plugins/learn/assets/src/views/ContentCard/index.vue
@@ -27,7 +27,7 @@
       </p>
       <div class="footer">
         <CoachContentLabel
-          v-if="userHasPermissions && !isLearner"
+          v-if="isUserLoggedIn && !isLearner"
           class="coach-content-label"
           :value="numCoachContents"
           :isTopic="isTopic"
@@ -122,7 +122,7 @@
       },
     },
     computed: {
-      ...mapGetters(['isLearner', 'userHasPermissions']),
+      ...mapGetters(['isLearner', 'isUserLoggedIn']),
       isTopic() {
         return this.kind === ContentNodeKinds.TOPIC || this.kind === ContentNodeKinds.CHANNEL;
       },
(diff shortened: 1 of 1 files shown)


kolibri/plugins/learn/assets/src/views/ContentCard/index.vue
```diff
@@ -27,6 +27,7 @@
       </p>
       <div class="footer">
         <CoachContentLabel
+          v-if="!isLearner"
```
jonboiser: Does this catch unauthenticated users as well? (Is an unauthenticated/guest user equivalent to a learner?)

nucleogenesis: Thanks @jonboiser I just pushed a commit addressing that.

kolibri/plugins/learn/assets/src/views/ContentCard/index.vue
```diff
@@ -27,6 +27,7 @@
       </p>
       <div class="footer">
         <CoachContentLabel
+          v-if="hasUserPermissions && !isLearner"
```
jonboiser: I can't find this getter in the code. I see a getter in `modules/session.js` called `getUserPermissions` or `userHasPermissions` is that what you meant?

jonboiser: @nucleogenesis Could you confirm that this is the right getter? I can't find it when searching the code.

nucleogenesis: @jonboiser Great catch - for some reason using a non-existent getter still achieved the correct effect.

In any case - I just pushed it up with the proper getter.

jonboiser: Yeah, that looks like a valid reference now, but I'm still not sure if that's the most straightforward way to check for an unauthenticated user... authenticated learners also have `userHasPermissions = false`.

The permissions object is kind of an indirect way to check if a person is authenticated. Is there something more direct we can check inside `core.state.session`?

jonboiser: ~Maybe we can implement a new straightforward getter `isGuest` and check `state.core.session.kind === 'anonymous'`~

rtibbles: Maybe the simplest way would be to just check that the user is logged in: https://github.com/learningequality/kolibri/blob/release-v0.12.y/kolibri/core/assets/src/state/modules/session.js#L34 and that they are not a learner?

jonboiser: Even better! Show the coach-content annotation iff the user is logged in and the user is not a learner (i.e. a coach or higher in the current scheme)

jonboiser: @nucleogenesis Like @rtibbles suggested, use the `isUserLoggedIn` getter instead of `userHasPermissions`

jonboiser (changes requested):
Use the `isUserLoggedIn` getter to check for (un)authenticated users.

jonboiser (approved):
Looks good to me!

PR #5595, opened 2019-05-30
Hide in-progress quiz scores in coach reports

### LINKED ISSUE #5585: in-progress quiz scores are shown to coaches
### Observed behavior

Scores are shown before the quiz has been submitted, giving an inaccurate portrayal of a learner's performance:

image

### Expected behavior

Scores should not be shown until the learner is finished with it. The 'average score' reported at the top should only include completed quizzes.

### User-facing consequences

Confusing view of quiz results.

### Context

0.12.3

nucleogenesis:
### Summary
This change addresses issue https://github.com/learningequality/kolibri/issues/5585

Two items are described there, only one was found to be an issue:
1. "Started" - Any non-completed quiz scores were being shown in Coach Quiz Reports, so a student who hasn't "Submitted" their quiz might show a 50% when they've only gone through 5/10 questions and have gotten them all right.

2. Ensure that the "Average Score" at the top wasn't taking non-completed scores into its calculations. I did not find that this was happening so no change in this PR affects that calculation.

**NOTE** - This was initially targeted for 0.12.4 but because of additional complications after a Slack discussion, this PR will update the way the `Progress` column's data is made. Anybody who has "Started" their quiz, should have an indicator saying "X of Y Answered" rather than just "Started".

This will require a change to a serializer in class_summary_api.py and associated tests to make the necessary data available to the frontend.

### Reviewer guidance

1. Create a Class with Learners and an active Quiz assigned to that Class.
2. In one browser, log in as a coach and view the report for that quiz, listing all Learners and their respective statuses on that quiz.
3. In a separate browser, log in as one of the Learners (in Incognito / Private Browsing mode).
4. Open the quiz: _See that student's record show "X of Y Started" in the "Progress" column and that there is nothing in the score column.
5. Answer questions (some correctly) - _Note that nothing changes for that user and no score is shown_ Also note that if you view the report simultaneously, the X of Y will update as the learner answers questions.
6. Submit the Quiz as the Learner. _After a second, it will update automatically and show "Completed" and a value % in in the 'Score' column.

Other Learner scenarios to test:
- Submit a quiz with no questions answered
- Submit an incomplete quiz with _no correct_ answers given.
- Submit an incomplete quiz with correct answers given.
- Submit a fully completed quiz with correct answers given.

Also - note that the Average score will only change once you've submitted a quiz as a learner.

### References

https://github.com/learningequality/kolibri/issues/5585

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated


nucleogenesis:
@indirectlylit 

We decided previously in a group call that instead of having the text "Started" - we ought to have a more informative string "X of Y answered" - which indicates to the Coach the progress that a student has made. If I recall correctly - this is because of the label for that column being called "Progress" - "Started" does not indicate a quantitative measure of "Progress" but instead shows more of a "Status".

I've gone through some of the steps that it would take to implement this to the point of having the information visible on page load.

However, before going any further I'd like to get some feedback.

1. I'm adding props to the components that display the user data - is there a better way to get the data needed into the LearnerProgressLabel?
2. Currently, if you're looking at the Quiz Report, and a student starts the quiz, then the "Started" label will show until you reload the page - perhaps this has to do with the query getting the # of answered questions. If you reload, then it will say "X of Y answered" as expected.
3. The `started` status is true even if the user hasn't answered any questions - so it could currently say "0 of Y answered" - could this possibly be confusing? If so - is there a better alternative?
4. Aside from the above, the next changes would be to add a notification that updates quiz progress and to ensure it gets merged/mapped to Vuex appropriately.

Example of what the "X of Y answered" looks like:
quiz-progress-new

kolibri/plugins/coach/assets/src/views/reports/ReportsQuizLearnerListPage.vue
```diff
@@ -110,6 +114,16 @@
     beforeMount() {
       this.filter = this.filterOptions[0];
     },
+    methods: {
+      score(statusObj) {
+        return statusObj.status === this.STATUSES.completed
+          ? statusObj.score || 0.0 // Ensures score will show 0% when null
+          : null;
+      },
```
indirectlylit: Recommend moving this logic inside the `table()` computed prop, and updating `tableRow` to have the value returned here.

(methods are always re-calculated, while the outputs of computed props get cached)

kolibri/plugins/coach/assets/src/views/reports/ReportsQuizLearnerListPage.vue
```diff
@@ -110,6 +114,16 @@
...
+      numAnswered(statusObj) {
+        return statusObj.status === this.STATUSES.started ? statusObj.num_answered : null;
+      },
```
indirectlylit: `numAnswered` can also be made an attribute directly on `tableRow` inside of the `table()` computed prop

kolibri/plugins/coach/assets/src/views/reports/ReportsQuizLearnerListPage.vue
```diff
@@ -42,9 +42,13 @@
               </KLabeledIcon>
             </td>
             <td>
-              <StatusSimple :status="tableRow.statusObj.status" />
+              <StatusSimple 
+                :status="tableRow.statusObj.status" 
+                :answeredQuestionsCount="numAnswered(tableRow.statusObj)" 
+                :totalQuestionsCount="exam.question_count" 
+              />
```
indirectlylit, line 58: The `StatusSimple` is actually just for "learner" statuses. All strings describe something of the form "how many learners are ____". (e.g. 3 of 5 learners have started, 1 learner needs help). In this new variation we're talking about questions, not learners.

There actually already are a couple strings that could be helpful:

https://github.com/learningequality/kolibri/blob/47218d290259abc8f9710943f06d599c00674ce8/kolibri/plugins/coach/assets/src/views/common/status/statusStrings.js#L150-L161

So then we can do something like 

```html
<StatusSimple
  v-if="tableRow.numAnswered !== null && exam.question_count !== null"
  :status="tableRow.statusObj.status"
/>
<KLabeledIcon
  v-else
  icon="inProgress"
  :text="translations.question.completed('ratioShort', {count: tableRow.numAnswered, total: exam.question_count})"
/>
```

where `translations` is available from `statusStringsMixin`.

kolibri/plugins/coach/class_summary_api.py
```diff
@@ -149,10 +152,27 @@ def get_num_correct(self, exam_log):
...
+                .aggregate(
+                    complete__sum=Count(
+                        Case(
+                            When(complete=True, then=1),
+                            default=0
+                        )
+                    )
+                )
+                .get("complete__sum")
+        )
```
indirectlylit, line 167: note that this will get run for every quiz and every student in the class. Might want to do a quick performance check to make sure it's not too heavy

indirectlylit (commented):
This looks great so far.

> I'm adding props to the components that display the user data - is there a better way to get the data needed into the LearnerProgressLabel?

Yes - see my notes inline. We should use a `KLabeledIcon` and the status strings directly, rather than re-using `LearnerProgressLabel`, because they're semantically not quite compatible.

> Currently, if you're looking at the Quiz Report, and a student starts the quiz, then the "Started" label will show until you reload the page - perhaps this has to do with the query getting the # of answered questions. If you reload, then it will say "X of Y answered" as expected.

This seems related to your last point below about notification updates of quiz progress

> The started status is true even if the user hasn't answered any questions - so it could currently say "0 of Y answered" - could this possibly be confusing? If so - is there a better alternative?

I think this is expected because it distinguishes learners who have opened the quiz from those that haven't.

> Aside from the above, the next changes would be to add a notification that updates quiz progress and to ensure it gets merged/mapped to Vuex appropriately.

@jonboiser and @jredrejo originally worked on the front- and back-ends of this, so would be best suited to give you guidance

kolibri/plugins/coach/assets/src/views/reports/ReportsQuizLearnerListPage.vue
nucleogenesis: @indirectlylit I tried applying these changes - but it looks like `statusStringsMixin` also brings along with it a bunch of props (count, verbosity, total) - which are required - which are not passed to `ReportsQuizLearnerListPage` - so I'm getting errors for those props not being provided to the component in which I'm using that mixin.

Also - that `ratioShort` string would be `X of Y completed` - which I don't think provides the meaningful context that `X of Y answered` does in implying that the X and Y are referring to a number of questions. 

In any case - should I define that string directly on the component for simplicity's sake?

I have implemented it as follows now - and the one issue I'm having is that the icon is not the light-blue color but is instead black unless I pass the hex code directly to the KIcon.

```html
<StatusSimple
  v-if="tableRow.statusObj.status !== STATUSES.started"
  :status="tableRow.statusObj.status"
/>
<KLabeledIcon v-else>
  <KIcon slot="icon" inProgress />
  {{ $tr('questionsCompletedRatioLabel', {count: tableRow.statusObj.num_answered, total: exam.question_count}) }}
</KLabeledIcon>
```

kolibri/plugins/coach/class_summary_api.py
nucleogenesis: Do you have any suggestions for how best to test the performance on this?

I reviewed the time it took for the `classSummary` api call to return in the browser - ~250ms with the code above in place.

When I change that function to return 0 - bypassing the DB work altogether - the time on that API call was +/- 20ms - not noticeably faster or slower.

I'm also working on a quiz with ~ 25 learners FWIW.

kolibri/plugins/coach/assets/src/views/reports/ReportsQuizLearnerListPage.vue
nucleogenesis: I worked out a way to just get rid of it - the `num_answered` property is defined on the `statusObj` for each `tableRow` so no need for a method now.

kolibri/plugins/coach/assets/src/views/reports/ReportsQuizLearnerListPage.vue
nucleogenesis: That method is called at line 53 - passed to the `<Score>` component.

I'm pretty sure that we could just ditch the method by calling the `<Score>` component like so:

`<Score v-if="tableRow.statusObj.status === STATUSES.completed" :value="tableRow.statusObj.score || 0.0" />`

Which shows nothing unless the quiz is completed, but ensures that a float is passed to `<Score>` even if we get null for `tableRow.statusObj.score`.

Thoughts?

kolibri/plugins/coach/class_summary_api.py
indirectlylit: okay - thanks for checking! Sounds like we're already fairly slow, but this doesn't change things much.

that process of testing sounds good to me, with the one caveat that the data might not be realistic. if you'd like to test with real data, aron or lingyi can give your our beta server DB.

kolibri/plugins/coach/assets/src/views/reports/ReportsQuizLearnerListPage.vue
indirectlylit: > In any case - should I define that string directly on the component for simplicity's sake?

I think you're right: better to define a new string specifically for this case

> the icon is not the light-blue color but is instead black unless I pass the hex code directly to the KIcon.

Rather than a hex code, you'll want to use the color token defined here: https://github.com/learningequality/kolibri/blob/develop/kolibri/core/assets/src/styles/theme.js#L40

To use this, add the `themeMixin` and then reference `$themeTokens.progress`

kolibri/plugins/coach/class_summary_api.py
nucleogenesis: Thanks, @indirectlylit , I got the DB and tested some larger classrooms using the demo DB and didn't notice any significant change to the performance.

kolibri/plugins/coach/assets/src/views/reports/ReportsQuizLearnerListPage.vue
indirectlylit: sounds fine to me! mainly I just like avoiding method calls, but even that is minor

rtibbles:
I can review the Python code here - is anything else needed aside from merge conflict resolution?

kolibri/core/notifications/api.py
```diff
@@ -285,6 +300,16 @@ def num_correct(examlog):
     )
 
 
+def num_answered(exam_log):
+    return (
+        exam_log.attemptlogs.values_list("item")
+        .order_by("completion_timestamp")
+        .distinct()
+        .aggregate(complete__sum=Count(Case(When(complete=True, then=1), default=0)))
```
rtibbles, line 308: Do we need the `Count(Case(When` here instead of just `Sum(`? Does it not sum over booleans properly? (I can imagine that possibly this might break on postgres, but I think it would work on sqlite at least).

kolibri/core/notifications/api.py
```diff
@@ -285,6 +300,16 @@ def num_correct(examlog):
     )
 
 
+def num_answered(exam_log):
+    return (
+        exam_log.attemptlogs.values_list("item")
```
rtibbles: Hrm, it's an edge case, but there's no guarantee that item ids are unique within quizzes. This is because they are drawn from multiple exercises, where within that exercise each item ids is guaranteed to be unique.

Setting line 305 to:
`exam_log.attemptlogs.values_list("item", "content_id")`
is probably the way to avoid this (admittedly unlikely) edge case.

kolibri/plugins/coach/class_summary_api.py
rtibbles: As this function appears to be identical to the above, note that my suggestions there apply here too.

Another thing that occurs is that we should probably be enforcing this uniqueness at the DB level, rather than having to do the rather tenuous distinct workaround for SQLite that we need to do here (although that won't help with historical data, but would just be good to do for future data hygiene).

@indirectlylit the DRF list serializer issue (serializing every model in a queryset as an independently loaded entity) is also an issue here, and probably the biggest bottleneck to performance.

We could get a small speedup by doing a prefetch on the exam log query here:
https://github.com/learningequality/kolibri/blob/develop/kolibri/plugins/coach/class_summary_api.py#L283
```
        query_exam_logs = logger_models.ExamLog.objects.filter(
            exam__in=query_exams
        ).prefetch_related("attemptlogs").annotate(last_activity=Max("attemptlogs__end_timestamp"))
```

kolibri/core/notifications/api.py
rtibbles: I see that this is also from the `num_correct` function above, so we should probably correct that too.

rtibbles (changes requested):
Would like to fix this edge case before it propagates further in the code - everything else is just suggestion and clarification.

kolibri/core/notifications/api.py
nucleogenesis: I initially tried just using `Sum(` but it didn't work, unfortunately. I just went through and tested again to be sure and for whatever reason, I'm not getting the proper updated data as expected when a learner submits an answer to a question.

kolibri/core/notifications/api.py
nucleogenesis: Thanks for that - I've updated both functions accordingly

kolibri/plugins/coach/class_summary_api.py
```diff
@@ -152,9 +155,27 @@ def get_num_correct(self, exam_log):
             .get("correct__sum")
         )
 
+    def get_num_answered(self, exam_log):
+        return (
+            exam_log.attemptlogs.values_list("item")
```
rtibbles: Still need to make the same update here as was made above:
`examlog.attemptlogs.values_list("item", "content_id")`

nucleogenesis: @rtibbles Should the change also apply to the `get_num_correct` method just above that `get_num_answered` method as well?

rtibbles: Ah, yes - otherwise we might undercount.

rtibbles (commented):
Code looks good to me now - do we need to update any gherkin stories here for testing? @radinamatic would you have some time to do a quick spin?

rtibbles (approved):

radinamatic:
@rtibbles @nucleogenesis I can test this tomorrow evening CET (midday PST), hope it's not urgent!

rtibbles:
Nope - I did a cursory test here and looks good. Feel free to merge on approval @radinamatic.

radinamatic:
@nucleogenesis @indirectlylit @rtibbles 

Changes seem implemented as per steps 1-6 in the **Reviewer guidance**. One comment and one issue found:

I would agree that `X of Y answered` gives better context for the **Progress** column. Quiz can be *started*, but questions are *answered* (or not). Is the change in string still pending?

When coach is looking at the detailed quiz report, it always lags one question: learner has answered, but that last attempt is not displayed to the coach until the page reload.

reload-to-see-answer

This issue persists even after the learner submits the quiz.

reload-to-see-the-last-answer-on-completed-quiz

rtibbles:
@nucleogenesis can you file a follow up issue for the string update?

@radinamatic can you file a follow up issue for the lag issue you see? (I think it is pre-existing to the work in this PR, so I don't want to hold up the merge).

PR #5603, opened 2019-06-03
clean up setup file

indirectlylit:
### Summary

* removed what appeared to be unreachable or unused code
* clean up batch file generation variable naming and comments

### Reviewer guidance

* Did I remove anything that is still being used?
* Should we open any issues to replace the 'todo' comments removed from the code?

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -8,22 +8,51 @@ import logging
 import os
 import sys
 
-try:  # for pip >= 10
-    from pip._internal.req import parse_requirements
-except ImportError:  # for pip <= 9.0.3
-    from pip.req import parse_requirements
 from setuptools import setup
 from setuptools.command.install_scripts import install_scripts
 
 import kolibri
-from kolibri import dist as kolibri_dist
+
+# Windows-specific .bat script template
+WINDOWS_BATCH_TEMPLATE = r"""@echo off
+set mypath=%~dp0
+set pyscript="%mypath%{file_name}"
+set /p line1=<%pyscript%
+if "%line1:~0,2%" == "#!" (goto :goodstart)
+echo First line of %pyscript% does not start with "#!"
+exit /b 1
+:goodstart
+set py_exe=%line1:~2%
+call %py_exe% %pyscript% %*
+"""
+
+# Generate Windows-specific .bat files
+class gen_windows_batch_files(install_scripts):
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -9,66 +9,15 @@
 import os
 import sys
 
-try:  # for pip >= 10
-    from pip._internal.req import parse_requirements
-except ImportError:  # for pip <= 9.0.3
-    from pip.req import parse_requirements
```
benjaoming, line 14: Have you researched the `pip` version on the various systems that we support?

benjaoming (approved):
Agreed, great changes and about time :) We seem to bundle dependencies as a default and the opposite hasn't been required. So the disabled code should go -- and if we need it again, it's over in KA Lite :+1:

setup.py
benjaoming: I think this will break `pip install kolibri` on a default Debian/Raspbian system.

https://packages.debian.org/stable/python-pip

indirectlylit: where is `parse_requirements` used? the only place I saw was inside an `if False` block

rtibbles: Yeah, it was added on the assumption that we would eventually do non-static pip builds for Kolibri, I guess?

benjaoming: Oh right, didn't see it was made redundant further down, please continue! :+1:

rtibbles:
Tried restarting the build to double check, but it's still failing. Given that you've edited setup.py, I assume the error is directly related to the PR, but I can't see where the issue is from a brief inspection of the code.

benjaoming:
It's because we use `setup.py sdist --static`, and that option was our own, which we removed from `argv` before passing it on.. need to correct one or more build instructions, search for `--static`.

jonboiser:
Buildkite is breaking with the `make dist` command. Does that need to be updated?

https://buildkite.com/learningequality/kolibri/builds/11265#92e32dd5-2a5d-4bbb-9a4f-5869265df3bb/549-2279

indirectlylit:
sorry I haven't had a chance to come back to this. will take a look this week

PR #5608, opened 2019-06-03
Release v0.12.x into release v0.12.y

rtibbles:
### Summary
Cascade merge.


jonboiser (approved):
Not sure about the django update, but as long as the new test passes, I think we're good with this merge.

Issue #5597, opened 2019-05-31
File exist error after the upgrade from 0.10 to 0.12.4-beta6


mrpau-richard:
### Observed behavior
I can't start the Kolibri server in Windows 7 it has an file exist error after I upgrade from `0.10` to `0.12.4-beta6`

### Expected behavior
I should be able to start the Kolibri server after the upgrade.

### User-facing consequences
The user might not able to start the Kolibri server after the upgrade.

### Errors and logs
![Screen Shot 2019-05-31 at 10 52 33 AM](https://user-images.githubusercontent.com/8663934/58679302-80749300-8395-11e9-904d-1dfe01e1db76.png)

### Work-around 
Manually delete the kolibri.log file path and restart the kolibri server.

### Steps to reproduce
* Install the Kolibri 0.10 on WIndows 7
* After the installation, upgrate it using the 0.12.4-beta6 Windows installer
* Manually start the Kolibri server using the command prompt.
* The file exist error show up at the command prompt.

### Context

Tell us about your environment, including:
 * Kolibri version:  0.12.4-beta6
 * Operating system: Windows 7
 * Browser: N/A

mrpau-julius:
Same error on windows 10, cannot start the Kolibri server after the upgrade.
![sameErrorwin10](https://user-images.githubusercontent.com/38421180/58682037-97b97d80-83a1-11e9-9662-2a350a7b8aef.png)

radinamatic:
Confirmed that this is the issue only on Windows. Upgrading to 0.12.4b6, either on Ubuntu where 0.10 was running as PEX, or when it was installed from DEB, creates no issues regarding the path of the log files.

lyw07:
thank you for testing! I will have a fix soon

indirectlylit:
nicely done!

rtibbles:
This should be fixed in #5609 but will wait for confirmation before closing.

lyw07:
confirmed by @radinamatic that this has been fixed.

PR #5609, opened 2019-06-03
Fix windows log file error when upgrading version

lyw07:
### Summary

The PR fixes the issue that kolibri cannot start during version upgrade due to an OSError from `os.rename` on Windows. The issue is because `os.rename` [cannot handle the case when the destination file exists on Windows](https://docs.python.org/2/library/os.html#os.rename). 
The solution is to instead use `shutil.move` so that when the destination file exists, the source file will be copied over to the destination and deleted afterwards.

### Reviewer guidance

please see if the error has been fixed on windows. (I tested this on my Windows VM)

### References

https://github.com/learningequality/kolibri/issues/5597

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/utils/sanity_checks.py
+++ b/kolibri/utils/sanity_checks.py
@@ -1,5 +1,6 @@
 import logging
 import os
+import shutil
 import sys
 
 import portend
@@ -93,4 +94,4 @@ def check_log_file_location():
         old_log_path = os.path.join(home, log)
         if os.path.exists(old_log_path):
             new_log_path = os.path.join(home, "logs", log_location_update[log])
-            os.rename(old_log_path, new_log_path)
+            shutil.move(old_log_path, new_log_path)
--- a/kolibri/utils/tests/test_sanity_check.py
+++ b/kolibri/utils/tests/test_sanity_check.py
@@ -43,12 +43,12 @@ class SanityCheckTestCase(TestCase):
             logging_mock.assert_called()
 
     @patch("kolibri.utils.cli.initialize")
-    @patch("kolibri.utils.sanity_checks.os.rename")
+    @patch("kolibri.utils.sanity_checks.shutil.move")
     @patch(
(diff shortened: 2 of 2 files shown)


kolibri/utils/sanity_checks.py
```diff
@@ -93,4 +94,4 @@ def check_log_file_location():
         old_log_path = os.path.join(home, log)
         if os.path.exists(old_log_path):
             new_log_path = os.path.join(home, "logs", log_location_update[log])
-            os.rename(old_log_path, new_log_path)
+            shutil.move(old_log_path, new_log_path)
```
rtibbles, line 97: Ah yes, that will make the difference!

Wonder if we should also just wrap this in a try except clause, so that we don't break startup if the file is corrupted, is actually a broken symlink etc.?

lyw07: I think `shutil.move` doesn't find out source files with broken symlinks. I created a corrupted file and used `shutil.move` to rename it to something else, but no errors showed up.

I have the `os.path.exists` check here: https://github.com/learningequality/kolibri/pull/5609/files/49ee1a89bbb51e9f2500b98883a1f197fb5cb9b9#diff-d2fc08a4eefe48248e5c777e07b056fdR95
so if the source file is corrupted, it will be ignored.

rtibbles (approved):
Code looks good - I don't have a Windows VM to test this on, but this can be double checked in the next beta.

radinamatic:
Yes, @mrpau-dev will test this on `0.12.4.b7`! :+1:

radinamatic:
Tested and confirmed working when upgraded from 0.11.1 to 0.12.4b6 on Windows 10!

lyw07:
🎉thank you for testing! @radinamatic

PR #5625, opened 2019-06-05
openid_connect plugin for BSF

jredrejo:
### Summary
Create a plugin called `openid_connect`, unabled by default, that will allow kolibri users to authenticate against an [OpenID Connect](https://openid.net/connect/) provider.
The setup is ready to work with the [olip-api](https://gitlab.com/bibliosansfrontieres/olip/olip-api) developed by the BSF project

### Reviewer guidance
1 . Ensure you have the server provider working either locally ([Docker setup](#docker-setup)) in an development environment or using a remote installation (as the one at http://70.167.220.187:443/ )
2. set the variable environment API_URL with the url of the server provider
3. enable this plugin in kolibri
4. After login  into the platform with an username existing in the platform (and not existing in the local kolibri database), login into kolibri should be possible just accepting the confirmation request.

### References

- [OpenID Connect specs](https://openid.net/specs/openid-connect-basic-1_0.html#CodeFlow)
- [OIDC glossary ](https://infosec.mozilla.org/guidelines/iam/openid_connect.html)
- [OLIP repositories](https://gitlab.com/bibliosansfrontieres/olip)
- [OLIP project architecture](https://docs.google.com/document/d/1bWLrUtpUtAuHTd_7phSZnH4b08Dn1bHAaB_K2Ge8_hI/edit?usp=sharing)

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated

# Docker Setup
Instructions to set BSF platform environment up running using Docker:
Previous note, there's a chicken-egg problem: The ip of the olip-api container is going to be needed to setup the olip-api container, in the `API_BASE_URL` env variable. In the example below it's 172.17.0.2. In other systems it can be different. The only way to know it is executing `docker inspect olip-api` after creating the container. If it's not the one written below, stop & remove the container and create it again with the assigned ip.
1. `docker run -d --name olip-api   --mount type=bind,source="$(pwd)"/,target=/data  -e HOST_DATA_DIR=/data -e API_BASE_URL=http://172.17.0.2:5002    -e SQLALCHEMY_DATABASE_URI=sqlite:////data/app.db  -e BOX_REPOSITORY_IPN=http://drop.bsf-intranet.org/olip/conf/        bibliosansfrontieres/olip-api-i386`
2. `docker run -d --name olip-dashboard -e API_URL=http://172.17.0.2:5002 bibliosansfrontieres/olip-dashboard-i386` . `API_URL` must use the ip of the `olip-api` container

3. Run kolibri , setting before an environment variable called `API_URL` with the value `http://olip-api:5002/oauth` being `olip-api` the value of the olip-api container ip


jredrejo:
@indirectlylit @rtibbles I think the code is ready to review now.
I would like to know your opinion + ideas on how we could make it work removing the added line at `kolibri/utils/conf.py` . This is the only line that avoid considering the code a clean plugin, as it affects to the kolibri core code. 
If we could set this as an option in `option.ini` or through an environment variable, this PR could be merged with the current Kolibri develop code without causing any damage.

kolibri/plugins/openid_connect/openid_connect_settings.py
```diff
@@ -0,0 +1,34 @@
...
+from kolibri.deployment.default.settings.base import AUTHENTICATION_BACKENDS
+from kolibri.deployment.default.settings.base import INSTALLED_APPS
+
+INSTALLED_APPS += ["mozilla_django_oidc"]
+
+# OIDC SETUP
+AUTHENTICATION_BACKENDS += [
+    "kolibri.plugins.openid_connect.auth.OIDCKolibriAuthenticationBackend"
+]
+OIDC_URL = os.environ.get("API_URL", "http://127.0.0.1:5002/oauth")
```
rtibbles, line 15: Would it be good to set this via the options.ini? (not currently possible for plugins to add to it, but wondering if it might be worthwhile).

jredrejo: yes, it would be much better, but I didn't want to add more code in the Kolibri 'core'. Is it possible to add entries to the options.ini file from a plugin?

rtibbles: Not yet, but I am prototyping something now to allow this. Just wanted to make sure that it would be of use!

requirements/base.txt
```diff
@@ -33,3 +33,4 @@ sentry-sdk==0.7.9
 django-redis-cache==2.0.0
 redis==3.2.1
 html5lib==1.0.1
+mozilla-django-oidc==1.2.2
```
jamalex, line 36: Note that installing this [creates a hard dependency](https://github.com/mozilla/mozilla-django-oidc/blob/master/setup.py#L35) on `cryptography`, which we only have as a soft dependency/recommendation for Kolibri itself (our crypto operations have pure-Python fallbacks when `cryptography` is not available). As long as it gracefully degrades (e.g. the OpenID endpoints don't work, but everything else is fine), then it could be OK. But we likely don't want it to install `cryptography` naively (we handle it specially when building the WHL file, to bundle CEXT builds for different architectures).

CC @lyw07

lyw07: Yes I discussed that with José yesterday. My suggestion was to only install the pure python packages`mozilla-django-oidc` and `josepy` without dependencies and get the remaining dependencies from `kolibri/dist/cext` folder as we have the `cryptography` and its dependencies installed there. One downside is that we don't have cryptography for all platforms installed in kolibri (not Mac OSX). 

We think that we should change the distribution of the c packages in kolibri, which I would like to add as a discussion item in next week's dev meeting

kolibri/plugins/openid_connect/openid_connect_settings.py
rtibbles: When we return to this plugin, this is now possible!

kolibri/plugins/user/templates/user/user.html
```diff
@@ -1,8 +1,10 @@
 {% extends "kolibri/base.html" %}
 {% load user_tags %}
 {% load kolibri_tags %}
+{% load webpack_tags %}
 {% load cache %}
 {% block content %}
   {% user_async_assets %}
   {% user_assets %}
+  {% webpack_asset 'user_module' %}
```
rtibbles, line 9: I am making this change in a parallel PR so there'll be a merge conflict here, happy to resolve any of the plugin related merge conflicts that arise!

PR #5631, opened 2019-06-06
Interactive video transcript

bjester:
### Summary

This adds a new transcript feature to the video player in Kolibri. It utilizes existing text tracks that generate the subtitles (captions) to create a transcript that appears alongside of the video. 

| Before | After |
| --- | --- |
| screenshot-kolibridemo learningequality org-2019 11 06-08-56-23 | screenshot-localhost-8000-2019 11 06-08-40-39 |
|    | screenshot-localhost-8000-2019 11 06-08-40-06 |

In technical terms, this uses the component architecture of Video.js to add and modify behavior within the video's container managed by Video.js. Mixins are used to connect Vue components with the Video.js components to allow for using Vue within the Video.js container. The mixin connection allows for utilizing some of handling that video.js provides as well as the flexibility of Vue components and our infrastructure built upon that.

### Reviewer guidance

Testing areas:
1) Initial state should have subtitles enabled by default if there is matching language with the current Kolibri language.
2) Switching between videos that have different language options and ensuring options are disabled and enabled appropriately.
3) Various video preferences should persist, including subtitle and transcript preferences.
4) Responsiveness to different screen sizes.
5) Performance: There is quite a bit of event handling to produce and render the transcript.

Known issues:
1) Control buttons in the video's menu bar do not have active highlighting like video.js adds by default.
2) Video.js menus are not exclusive when open, and these new custom menus do not change that behavior, meaning more than one menu may be open at once.
3) In Safari, cues do not load in transcript until after first cue becomes active or transcript is toggled.

### References

https://www.notion.so/learningequality/Interactive-Video-Transcript-9ac195ce394942838a85ed22c0ebeaf0

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated


kolibri/core/assets/src/utils/i18n.js
```diff
@@ -95,9 +95,7 @@ const languageDensityMapping = {
   zh: languageDensities.dense,
 };
 
-function languageIdToCode(id) {
-  return id.split('-')[0].toLowerCase();
-}
+export const languageIdToCode = id => id.split('-')[0].toLowerCase();
```
indirectlylit: Super minor question, but curious why you prefer the  `const` and arrow function?

I personally find the `function` style to be more explicit and readable:

```js
export function languageIdToCode(id) {
  return id.split('-')[0].toLowerCase();
}
```

kolibri/plugins/media_player/assets/src/module.js
```diff
@@ -1,10 +1,20 @@
...
     return MediaPlayerComponent;
   }
+
+  get store() {
+    return store;
+  }
+
+  ready() {
+    this.store.registerModule('mediaPlayer', storeModule);
+  }
```
indirectlylit, line 17: The only disadvantage of doing this in vuex versus component `data` that I can think of is it limits us to one video component visible at a time? I can't think of any near-term situation where we'd want multiple videos displayed simultaneously, so should be fine.

kolibri/plugins/media_player/assets/src/views/MediaPlayerIndex.vue
```diff
@@ -259,25 +304,28 @@
...
         this.useSavedSettings();
         this.loading = false;
         this.$refs.player.tabIndex = -1;
 
         this.updateContentStateInterval = setInterval(this.updateContentState, 30000);
       },
       resizePlayer() {
-        const wrapperWidth = this.$refs.wrapper.clientWidth;
+        if (this.isFullscreen) {
+          this.$refs.wrapper.style.height = `100vh`;
```
indirectlylit: We've generally tried to avoid the `vh` unit due to poor support on early android browsers (pre 4.4). We aren't actively testing these anymore, but if there's another way of achieving this without `vh` that would be preferable.

kolibri/plugins/media_player/assets/src/views/MediaPlayerIndex.vue
```diff
@@ -473,18 +506,62 @@
...
-  /** COLOR PALLETTE **/
-  $video-player-color: #212121;
-  // tint if $video-player-color = black-ish, shade if $video-player-color = white-ish
-  $video-player-color-2: tint($video-player-color, 7%);
-  $video-player-color-3: tint($video-player-color, 15%);
-  $video-player-font-color: white;
+  @import './videojs-style/variables';
+
+  .video-js.vjs-fill {
+    z-index: 1;
```
indirectlylit: just flagging that z-indexes in Kolibri should always follow the Material elevations outlined here:

https://material.io/design/environment/elevation.html#default-elevations

(I don't know if this is correct or not, just mentioning)

kolibri/plugins/media_player/assets/src/views/MediaPlayerTranscript/TranscriptCue.vue
```diff
@@ -0,0 +1,185 @@
...
+      cue: Object,
+      mediaDuration: Number,
+      langCode: String,
+      active: {
+        type: Boolean,
+        default: false,
+      },
+    },
+
+    data: () => ({}),
```
indirectlylit: unused `data`

kolibri/plugins/media_player/assets/src/views/MediaPlayerTranscript/TranscriptCue.vue
```diff
@@ -0,0 +1,185 @@
...
+  import themeMixin from 'kolibri.coreVue.mixins.themeMixin';
+  import { getLangDir } from 'kolibri.utils.i18n';
+
+  const SPEAKER_REGEX = /<v ([^>]+)>/i;
+
+  export default {
+    name: 'TranscriptCue',
+
+    mixins: [themeMixin],
+
```
indirectlylit: Suuuper minor, but conventionally in the kolibri repo we haven't been putting whitespace between attributes of the vue component object.

(This should really be handled by `prettier` rather than code review)

kolibri/plugins/media_player/assets/src/views/MediaPlayerIndex.vue
```diff
@@ -453,17 +473,30 @@
...
   .fill-space {
     position: relative;
     width: 100%;
     height: 100%;
+    border: 1px solid transparent;
+  }
+
+  .loading-space {
+    box-sizing: border-box;
+    padding-top: calc(100% * 9 / 16);
```
indirectlylit: any way to avoid using `calc`? There [might be issues](https://caniuse.com/#feat=calc) with android 4.4 and below

kolibri/plugins/media_player/assets/src/views/customButtons.js
```diff
@@ -23,8 +24,4 @@ export class ForwardButton extends videojsButton {
   }
 }
 
-export class MimicFullscreenToggle extends videojsFullscreenToggle {
-  handleClick() {
-    this.player().trigger('mimicFullscreenToggled');
-  }
-}
+ForwardButton.prototype.controlText_ = 'Forward';
```
indirectlylit, line 27: Do 'Replay' and 'Forward' need to be internationalized?

kolibri/plugins/media_player/assets/src/views/MediaPlayerTranscript/TranscriptCue.vue
```diff
@@ -0,0 +1,185 @@
+<template>
+
+  <a
+    role="button"
+    tabindex="0"
+    :class="[
+      'transcript-cue',
+      { active },
+      $computedClass(style)
+    ]"
```
indirectlylit, line 10: one thing I learned recently is that you can include both `class` and `:class` in the same element, e.g.

```
    class="transcript-cue"
    :class="[
      { active },
      $computedClass(style)
    ]"
```

I don't think that's an improvement in this case, but just FYI in case that's new to you too

indirectlylit (commented):
This is looking great!

In a few places there are slightly modern CSS features that are okay for the majority of our users, but if it's possible to avoid them without a lot of extra work, that would be preferable.

From a usability perspective the main thing I'd like to do is try and avoid the nested menus (cc @khangmach @jtamiace).

Proposal:

* have two menus, CC and Language, rather than just one for CC
* CC has two switches, Subtitles and Transcript
* Language has a list of radio buttons for all languages
* If a language is selected while neither Subtitles or Transcript are enabled, Subtitles are automatically turned on

kolibri/core/assets/src/utils/i18n.js
bjester: I believe I just matched the style above this in the file, but I now realize that everything below is the opposite. Your example is fine with me, and has the benefit of better stack traces.

kolibri/plugins/media_player/assets/src/module.js
bjester: Introducing Vuex helped greatly since there is a lot of state to handle. Initially I did not go with using it and wasn't a happy developer.

kolibri/plugins/media_player/assets/src/views/MediaPlayerIndex.vue
bjester: This might need this to be `1` because of Video.js styles, but I'd have to double check.

kolibri/plugins/media_player/assets/src/views/MediaPlayerTranscript/TranscriptCue.vue
bjester: Gotcha. I would like to say that, whitespace is not an enemy! I've found most of the Kolibri code super dense. I once used the same style that LE does for JS, but after trying Allman style, which produces more visual whitespace, found it so much more readable. Just my preference and thoughts!

kolibri/plugins/media_player/assets/src/views/customButtons.js
bjester: They're internationalized within Video.js.

kolibri/plugins/media_player/assets/src/views/MediaPlayerTranscript/TranscriptCue.vue
indirectlylit: That's reasonable – I don't feel strongly on the matter myself. I would just recommend that we systematically change it everywhere rather than have multiple styles in the same codebase.

packages/kolibri-components/src/KResponsiveWindowMixin.js
```diff
@@ -109,11 +109,17 @@ export default {
         If that issue ever gets addressed, we should make them computed props.
       */
       windowBreakpoint: undefined,
+      windowIsPortrait: false,
+      windowIsLandscape: false,
```
indirectlylit, line 113: FYI you'll likely encounter a conflict with a change I made: https://github.com/learningequality/kolibri/pull/6000/files#diff-60e4b3c76101fb0363a4af6b85fb8719R118

Should be an easy resolution, let me know if you need help

bjester: Alright, thanks for the heads up

kolibri/plugins/media_player/assets/src/modules/captions/index.js
```diff
@@ -188,14 +177,31 @@ export default {
...
+
+      store.commit('SET_CUES', cues);
     },
     setActiveCuesFromTrack(store, track) {
-      store.commit('SET_ACTIVE_CUE_LIST', track.activeCues);
+      // In case we get triggered to set active cues but haven't added cues yet, do that now
+      // This helps an issue in Safari where we don't get an updated cue list
+      if (track.cues && track.cues.length !== store.state.cues.length) {
+        store.dispatch('setCuesFromTrack', track);
+      }
```
indirectlylit, line 219: cool

bjester: It doesn't entirely fix the issue in Safari, but it helps. I'll definitely add a followup issue regarding it.

bjester:
Added issue https://github.com/learningequality/kolibri/issues/6034

radinamatic:
Great to be able to activate/deactivate the captions by keyboard! :+1: :clap: 

One issue I run into with this is when the user is navigating up and down the transcript **while video is playing**, video player bar with all the rest of controls disappears, not only visually, but from the TAB order too. This results in user being unable to navigate back and pause/stop or control the playback in any other way. Happens with both positions of the transcript (sidebar and below).

([screencast here, too big for GH attachment](https://drive.google.com/open?id=1YaGAc0lM4TIQsWXQmcYo89I1hWos9cWE))

Not sure what here is the best solution: are we able to keep the control bar visible, and **in the TAB order** while the keyboard focus is inside the transcript?

Second problem might be more hairy: I see a real issue when the video is long, and the list of captions becomes interminable. Could we think about the option to allow users to go to beginning or the end of the transcript with some keyboard shortcut? :thinking: 

I was looking at various patterns at [WAI-ARIA Authoring Practices](https://www.w3.org/TR/wai-aria-practices/), and several (for example [Listbox](https://www.w3.org/TR/wai-aria-practices/#Listbox)) include using <kbd>Home</kbd> and <kbd>End</kbd> keys for precisely this, and add that _Supporting this key is strongly recommended for lists with more than five options_.

Now, I'm not suggesting you refactor the transcript completely to reflect this whole pattern, just to see if we could add something similar to be able to navigate more easily to beginning and the end of the long transcripts

bjester:
@radinamatic 
> One issue I run into with this is when the user is navigating up and down the transcript **while video is playing**, video player bar with all the rest of controls disappears, not only visually, but from the TAB order too. This results in user being unable to navigate back and pause/stop or control the playback in any other way. Happens with both positions of the transcript (sidebar and below).

That's a good catch. I'm not a fan of the control bar disappearing, but that was an existing part of the renderer, so without the transcript, the issue exists with it disappearing from the TAB order.

> Not sure what here is the best solution: are we able to keep the control bar visible, and **in the TAB order** while the keyboard focus is inside the transcript?

That could work, but I think you found a larger issue in that it disappears with or without the transcript. So perhaps a solution shouldn't be dependent on the transcript? Perhaps in keyboard modality, we keep the control bar visible regardless?


> I was looking at various patterns at [WAI-ARIA Authoring Practices](https://www.w3.org/TR/wai-aria-practices/), and several (for example [Listbox](https://www.w3.org/TR/wai-aria-practices/#Listbox)) include using Home and End keys for precisely this, and add that _Supporting this key is strongly recommended for lists with more than five options_.

I think adding support for <kbd>Home</kbd> and <kbd>End</kbd> should be reasonably straightforward. Perhaps longer term we could think about changing the roles of the items in the transcript such that it's traversalable by arrow keys and perhaps <kbd>Tab</kbd> jumps between `Beginning of transcript` and `End of transcript`

radinamatic:
> Perhaps in keyboard modality, we keep the control bar visible regardless

I think this a reasonable solution, considering how much keyboard-only users depend on easily locating controls they need, and control bar disappearing by default makes that more difficult! 

If this approach does not involve too much tweaking, go for it :+1: 

> changing the roles of the items in the transcript such that it's traversalable by arrow keys and perhaps <kbd>Tab</kbd> jumps between `Beginning of transcript` and `End of transcript`

Yes, I like this suggestion a lot! :heart_eyes: 
Let's put it on the 0.14 feature list, and proceed to add the support for  <kbd>Home</kbd> and <kbd>End</kbd> in the scope of this PR.

Thank you!

packages/kolibri-components/src/KThemePlugin.js
```diff
@@ -41,6 +41,9 @@ export default function KThemePlugin(Vue) {
           outline: 'none',
         };
       },
+      $inputModality() {
+        return globalThemeState.inputModality;
+      },
```
indirectlylit, line 46: great

bjester: I had tried to add it as a computed prop solely to the media player component, but it wouldn't update. Although, this works even though it's nearly the same.

jonboiser (approved):

PR #5637, opened 2019-06-07
Add plugin blacklist for when plugins are disabled.

### LINKED ISSUE #5633: add 'disabled' list of plugins alongside 'enabled' in kolibri_settings.json
### Observed behavior

Currently we have no way of distinguishing disabled plugins from newly created plugins.

### Expected behavior

* disabled plugins should remain disabled across upgrades
* newly created default plugins should automatically become enabled on upgrade

### User-facing consequences

…

### Errors and logs

…

### Steps to reproduce

…

### Context

…

rtibbles:
### Summary
Adds a blacklist for plugins, so that plugins that are deliberately disabled can stay disabled, even on upgrade.

### Reviewer guidance
1) To test, disable a core plugin:
`kolibri plugin kolibri.plugins.learn disable`
2) Start the server.
3) Verify that learn is now not available on the kolibri server.
4) Stop the server.
4) Then enter the shell (`kolibri shell`) and enter the following:
```
from kolibri.utils.conf import config
config.enable_default_plugins()
```
5) Start the server.
6) Verify that learn is still not available on the kolibri server.

### References
Fixes #5633

Will do additional cleanup of some of this logic when rebasing #5494.

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/plugins/utils.py
+++ b/kolibri/plugins/utils.py
@@ -1,4 +1,29 @@
+import importlib
+import logging
+
+from django.core.exceptions import AppRegistryNotReady
 from django.core.urlresolvers import reverse
+
+from kolibri.plugins.base import KolibriPluginBase
+from kolibri.utils.conf import config
+
+logger = logging.getLogger(__name__)
+
+
+class PluginDoesNotExist(Exception):
+    """
+    This exception is local to the CLI environment in case actions are performed
+    on a plugin that cannot be loaded.
+    """
+
+
+class PluginBaseLoadsApp(Exception):
+    """
+    An exception raised in case a kolibri_plugin.py results in loading of the
+    Django app stack.
+    """
+
+    pass
 
 
 def plugin_url(plugin_class, url_name):
@@ -7,3 +32,81 @@ def plugin_url(plugin_class, url_name):
             namespace=plugin_class().url_namespace(), url_name=url_name
         )
     )
+
+
+def _is_plugin(obj):
+    return (
+        isinstance(obj, type)
+        and obj is not KolibriPluginBase
+        and issubclass(obj, KolibriPluginBase)
+    )
+
+
+def get_kolibri_plugin(plugin_name):
+    """
+    Try to load kolibri_plugin from given plugin module identifier
+
+    :returns: A list of classes inheriting from KolibriPluginBase
+    """
+
+    plugin_classes = []
+
+    try:
+        plugin_module = importlib.import_module(plugin_name + ".kolibri_plugin")
+        for obj in plugin_module.__dict__.values():
+            if _is_plugin(obj):
+                plugin_classes.append(obj)
+    except ImportError as e:
+        # Python 2: message, Python 3: msg
+        exc_message = getattr(e, "message", getattr(e, "msg", None))
+        if exc_message.startswith("No module named"):
+            msg = (
+                "Plugin '{}' does not seem to exist. Is it on the PYTHONPATH?"
+            ).format(plugin_name)
+            raise PluginDoesNotExist(msg)
+        else:
+            raise
+    except AppRegistryNotReady:
+        msg = (
+            "Plugin '{}' loads the Django app registry, which it isn't "
+            "allowed to do while enabling or disabling itself."
+        ).format(plugin_name)
+        raise PluginBaseLoadsApp(msg)
+
+    if not plugin_classes:
+        # There's no clear use case for a plugin without a KolibriPluginBase
+        # inheritor, for now just throw a warning
+        logger.warning(
+            "Plugin '{}' has no KolibriPluginBase defined".format(plugin_name)
+        )
+
+    return plugin_classes
+
+
+def enable_plugin(plugin_name):
+    plugin_classes = get_kolibri_plugin(plugin_name)
+    for klass in plugin_classes:
+        klass.enable()
+
+
+def disable_plugin(plugin_name):
+    try:
+        plugin_classes = get_kolibri_plugin(plugin_name)
+        for klass in plugin_classes:
+            klass.disable()
+    except PluginDoesNotExist as e:
+        logger.error(str(e))
+        logger.warning(
+            "Removing '{}' from configuration in a naive way.".format(plugin_name)
+        )
+        if plugin_name in config["INSTALLED_APPS"]:
+            config["INSTALLED_APPS"].remove(plugin_name)
+            logger.info("Removed '{}' from INSTALLED_APPS".format(plugin_name))
+        else:
+            logger.warning(
+                (
+                    "Could not find any matches for {} in INSTALLED_APPS".format(
+                        plugin_name
+                    )
+                )
+            )
--- a/kolibri/plugins/utils.py
+++ b/kolibri/plugins/utils.py
@@ -36,6 +36,11 @@ def plugin_url(plugin_class, url_name):
 
 def _is_plugin(obj):
     return (
+        # Check that the object is an instance of type
+        # i.e. that it is a class definition, not an
+        # instantiated class.
+        # Failing to do this will result in the call to
+        # issubclass below blowing up.
         isinstance(obj, type)
         and obj is not KolibriPluginBase
         and issubclass(obj, KolibriPluginBase)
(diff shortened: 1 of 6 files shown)


indirectlylit:
pushing to 0.12.5 because this is mainly for internal dev experince

rtibbles:
Updating to 0.12.6 as needed for UNW work.

kolibri/plugins/utils.py
```diff
@@ -1,9 +1,112 @@
...
+    """
+
+
+class PluginBaseLoadsApp(Exception):
+    """
+    An exception raised in case a kolibri_plugin.py results in loading of the
+    Django app stack.
+    """
+
+    pass
```
indirectlylit, line 26: Non-blocking: `pass` not necessary because there's a docstring

kolibri/plugins/utils.py
```diff
@@ -1,9 +1,112 @@
...
     return reverse(
         "kolibri:{namespace}:{url_name}".format(
             namespace=plugin_class().url_namespace(), url_name=url_name
         )
     )
+
+
+def _is_plugin(obj):
+    return (
+        isinstance(obj, type)
```
indirectlylit, line 44: This could use a comment clarifying the intent. It's not clear from me why we would check whether `obj` is of the same type as the builtin `type()` function

kolibri/plugins/utils.py
```diff
@@ -1,9 +1,112 @@
...
+    plugin_classes = []
+
+    try:
+        plugin_module = importlib.import_module(plugin_name + ".kolibri_plugin")
+        for obj in plugin_module.__dict__.values():
+            if _is_plugin(obj):
+                plugin_classes.append(obj)
+    except ImportError as e:
+        # Python 2: message, Python 3: msg
+        exc_message = getattr(e, "message", getattr(e, "msg", None))
```
indirectlylit, line 68: I think this be just `str(e)`?

kolibri/plugins/utils.py
```diff
@@ -1,9 +1,112 @@
...
+    Try to load kolibri_plugin from given plugin module identifier
+
+    :returns: A list of classes inheriting from KolibriPluginBase
+    """
+
+    plugin_classes = []
+
+    try:
+        plugin_module = importlib.import_module(plugin_name + ".kolibri_plugin")
+        for obj in plugin_module.__dict__.values():
```
indirectlylit, line 63: recommend moving the `for` loop outside of the `try`/`except` blocks to keep it clear where the exceptions are expected to be thrown from

kolibri/plugins/utils.py
```diff
@@ -1,9 +1,112 @@
...
+            "Plugin '{}' has no KolibriPluginBase defined".format(plugin_name)
+        )
+
+    return plugin_classes
+
+
+def enable_plugin(plugin_name):
+    plugin_classes = get_kolibri_plugin(plugin_name)
+    for klass in plugin_classes:
+        klass.enable()
```
indirectlylit, line 96: Non-blocking: maybe 

```python
    for c in plugin_classes:
        c.enable()
```

kolibri/utils/conf.py
```diff
@@ -80,92 +78,126 @@
...
-                ).format(mod=module_path)
-            )
-            changed = True
-
-    if changed:
-        save()
+# These values are encoded on the config dict as sets
+# so they need to be treated specially for serialization
+# and deserialization to/from JSON
+SET_KEYS = ["INSTALLED_APPS", "DISABLED_APPS"]
```
indirectlylit, line 87: maybe

```python
KEY_INSTALLED_APPS = "INSTALLED_APPS"
KEY_DISABLED_APPS = "DISABLED_APPS"
SET_KEYS = [KEY_INSTALLED_APPS, KEY_DISABLED_APPS]
```

then all the hard-coded strings below can be replaced with constants

kolibri/utils/conf.py
```diff
@@ -80,92 +78,126 @@
...
+        )
+        if os.path.isfile(conf_file):
+            try:
+                # Open up the config file and load settings
+                # use default OS encoding
+                with open(conf_file, "r") as kolibri_conf_file:
+                    self.update(json.load(kolibri_conf_file))
+                return
+            except json.JSONDecodeError:
+                pass
```
indirectlylit: can we log a warning here?

kolibri/utils/conf.py
```diff
@@ -80,92 +78,126 @@
...
+                "INSTALLED_APPS": DEFAULT_PLUGINS,
+                #: Everything in this list is removed from the list above
+                "DISABLED_APPS": [],
+            }
+        )
+        if os.path.isfile(conf_file):
+            try:
+                # Open up the config file and load settings
+                # use default OS encoding
+                with open(conf_file, "r") as kolibri_conf_file:
```
indirectlylit, line 107: Non-blocking: I had left this with default OS encoding because people were manually editing their JSON files in local text editors. If we wanted to move towards what we did in #3769 we could provide a UTF-8 fallback e.g.

```python
io.open(conf_file, mode='r', encoding='utf-8')
```

and then start writing UTF-8

kolibri/utils/conf.py
```diff
@@ -80,92 +78,126 @@
...
-
-    if changed:
-        save()
+# These values are encoded on the config dict as sets
+# so they need to be treated specially for serialization
+# and deserialization to/from JSON
+SET_KEYS = ["INSTALLED_APPS", "DISABLED_APPS"]
+
+
+class ConfigDict(dict):
```
indirectlylit, line 90: Non-blocking: normally my instinct is to [avoid subclassing](https://en.wikipedia.org/wiki/Composition_over_inheritance), particularly of built-ins types. In this case, I would personally have kept `config` a simple `dict`, and implemented the methods in this class (`update`, `save`, etc) as simple module-level functions which operate on `config`.

kolibri/utils/conf.py
```diff
@@ -80,92 +78,126 @@
...
+        """
+        Updates current configuration with ``new_values``. Does not save to file.
+        """
+        values_copy = new_values.copy()
+        for key in SET_KEYS:
+            if key in values_copy:
+                values_copy[key] = set(values_copy[key])
+        super(ConfigDict, self).update(values_copy)
+
+    def save(self):
```
indirectlylit, line 131: maybe call this `write_settings_file` or something more explicit

kolibri/utils/conf.py
```diff
@@ -80,92 +78,126 @@
...
+        # use default OS encoding
+        config_copy = self.copy()
+        for key in SET_KEYS:
+            if key in config_copy:
+                config_copy[key] = list(config_copy[key])
+        with open(conf_file, "w") as kolibri_conf_file:
+            json.dump(config_copy, kolibri_conf_file, indent=2, sort_keys=True)
+
+    def autoremove_unavailable_plugins(self):
```
indirectlylit, line 140: it's not clear to me why we would actually want to do this. Seems like it would be better to leave the plugin enabled, but simply warn that it couldn't be loaded?

kolibri/utils/conf.py
```diff
@@ -80,92 +78,126 @@
...
+            self["DISABLED_APPS"].remove(module_path)
+        except KeyError:
+            pass
+
+    def disable_plugin(self, module_path):
+        self["DISABLED_APPS"].add(module_path)
+        try:
+            self["INSTALLED_APPS"].remove(module_path)
+        except KeyError:
+            pass
```
indirectlylit, line 198: don't `enable_plugin` and `disable_plugin` need to include a call to `save`?

indirectlylit (approved):
This works for me. There are a few small changes from the last review I'd recommend before merging:

1. clarify `isinstance` behavior: https://github.com/learningequality/kolibri/pull/5637/files#r296875951
2. move `for` loop outside of `try` block: https://github.com/learningequality/kolibri/pull/5637/files#r296880878
3. log a warning or exit completely on JSON decode error rather than failing silently: https://github.com/learningequality/kolibri/pull/5637/files#r296886177

The other items I think could be addressed but aren't necessary

kolibri/plugins/utils.py
rtibbles: Adds a decent amount of cruft to do that - so adding a comment instead.

PR #5641, opened 2019-06-07
Implement permalinking scheme for content.

rtibbles:
### Summary
Implements a permalinking scheme for content.
Sets up a hook for the learn plugin to redirect based on permalinks.

### Reviewer guidance
These links will be permanent, so we don't want to change them. How does this look?

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/core/content/hooks.py
+++ b/kolibri/core/content/hooks.py
@@ -18,6 +18,7 @@ from django.utils.safestring import mark_safe
 from le_utils.constants import content_kinds
 
 from kolibri.core.webpack.hooks import WebpackBundleHook
+from kolibri.plugins.hooks import KolibriHook
 from kolibri.utils import conf
 
 logger = logging.getLogger(__name__)
@@ -108,3 +109,17 @@ class ContentRendererHook(WebpackBundleHook):
             )
         ]
         return mark_safe("\n".join(tags))
+
+
+class ContentNodeDisplayHook(KolibriHook):
+    """
+    A hook that registers a capability of a plugin to provide a user interface
+    for a content node. When subclassed, this hook should expose a method that
+    accepts a ContentNode instance as an argument, and returns a URL where the
+    interface to interacting with that node for the user is exposed.
+    If this plugin cannot produce an interface for this particular content node
+    then it may return None.
+    """
+
+    def node_url(self, content_node):
+        raise NotImplementedError("This must be overridden by a subclass")
--- a/kolibri/core/content/urls.py
+++ b/kolibri/core/content/urls.py
@@ -1,5 +1,6 @@
 from django.conf.urls import url
 
+from .views import ContentPermalinkRedirect
 from .views import DownloadContentView
 from .views import ZipContentView
 
@@ -16,4 +17,5 @@ urlpatterns = [
         {},
         "downloadcontent",
     ),
+    url(r"^viewcontent$", ContentPermalinkRedirect.as_view(), name="contentpermalink"),
 ]
--- a/kolibri/core/content/views.py
+++ b/kolibri/core/content/views.py
@@ -15,6 +15,7 @@ from django.conf import settings
 from django.contrib.staticfiles import finders
 from django.http import Http404
 from django.http import HttpResponse
+from django.http import HttpResponseRedirect
 from django.http.response import FileResponse
 from django.http.response import HttpResponseNotModified
 from django.template import loader
@@ -30,9 +31,11 @@ from six.moves.urllib.parse import urlunparse
 from .api import cache_forever
 from .decorators import add_security_headers
 from .decorators import get_referrer_url
+from .models import ContentNode
 from .utils.paths import get_content_storage_file_path
 from kolibri import __version__ as kolibri_version
 from kolibri.core.content.errors import InvalidStorageFilenameError
+from kolibri.core.content.hooks import ContentNodeDisplayHook
 
 # Do this to prevent import of broken Windows filetype registry that makes guesstype not work.
 # https://www.thecodingforums.com/threads/mimetypes-guess_type-broken-in-windows-on-py2-7-and-python-3-x.952693/
@@ -319,3 +322,33 @@ class DownloadContentView(View):
         response["Content-Length"] = os.path.getsize(path)
 
         return response
+
+
+class ContentPermalinkRedirect(View):
+    def get(self, request, *args, **kwargs):
+
+        # extract the GET parameters
+        channel_id = request.GET.get("channel_id")
+        node_id = request.GET.get("node_id")
+        content_id = request.GET.get("content_id")
+
+        try:  # first, try to get the node by the unique node_id
+            node = ContentNode.objects.get(id=node_id)
+        except ContentNode.DoesNotExist:  # if it isn't found, fall back to looking for the content_id in the channel
+            node = ContentNode.objects.filter(
+                channel_id=channel_id, content_id=content_id
+            ).first()
+            if (
+                not node
+            ):  # if it's still not found, see if we can find anything with the content_id across any channel
+                node = ContentNode.objects.filter(content_id=content_id).first()
+
+        # build up the target topic/content page URL
+        if node:
+            url = None
+            for hook in ContentNodeDisplayHook().registered_hooks:
+                url = hook.node_url(node)
+            if url:
+                return HttpResponseRedirect(url)
+
+        raise Http404
--- a/kolibri/core/content/views.py
+++ b/kolibri/core/content/views.py
@@ -324,6 +324,44 @@ class DownloadContentView(View):
         return response
 
 
+def get_by_node_id(node_id):
+    """
+    Function to return a content node based on a node id
+    """
+    if node_id:
+        try:
+            return ContentNode.objects.get(id=node_id)
+        except (ContentNode.DoesNotExist, ValueError):
+            # not found, or the id is invalid
+            pass
+
+
+def get_by_channel_id_and_content_id(channel_id, content_id):
+    """
+    Function to return a content node based on a channel_id and content_id
+    """
+    if channel_id and content_id:
+        try:
+            return ContentNode.objects.filter(
+                channel_id=channel_id, content_id=content_id
+            ).first()
+        except ValueError:
+            # Raised if a malformed UUID is passed
+            pass
+
+
+def get_by_content_id(content_id):
+    """
+    Function to return a content node based on a content_id
+    """
+    if content_id:
+        try:
+            return ContentNode.objects.filter(content_id=content_id).first()
+        except ValueError:
+            # Raised if a malformed UUID is passed
+            pass
+
+
 class ContentPermalinkRedirect(View):
     def get(self, request, *args, **kwargs):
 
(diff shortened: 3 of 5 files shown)


kolibri/core/content/views.py
```diff
@@ -319,3 +322,30 @@ def get(self, request, filename, new_filename):
...
+            if not node:  # if it's still not found, see if we can find anything with the content_id across any channel
+                node = ContentNode.objects.filter(content_id=content_id).first()
+
+        # build up the target topic/content page URL
+        if node:
+            url = None
+            for hook in ContentNodeDisplayHook().registered_hooks:
+                url = hook.node_url(node)
+            if url:
+                return HttpResponseRedirect(url)
```
ivanistheone, line 388: HttpResponseRedirect = 302 = "Moved Temporarily"
using HttpResponsePermanentRedirect would result in 301 

I have never seen a HTTP system where the difference mattered so probably either is good, but wanted to flag the two options for your considerations.

kolibri/core/content/views.py
```diff
@@ -319,3 +322,30 @@ def get(self, request, filename, new_filename):
...
+
+    def get(self, request, *args, **kwargs):
+
+        # extract the GET parameters
+        channel_id = request.GET.get("channel_id")
+        node_id = request.GET.get("node_id")
+        content_id = request.GET.get("content_id")
+
+        try:  # first, try to get the node by the unique node_id
+            node = ContentNode.objects.get(id=node_id)
```
ivanistheone: Using `id__startswith` instead of `id` here would allow for partial-hash URLs (e.g. using only first six characters of the node_id as the key).

Same goes for lookups on line `339` and `341`.

kolibri/core/content/urls.py
```diff
@@ -16,4 +17,9 @@
         {},
         "downloadcontent",
     ),
+    url(
+        r'^viewcontent$',
```
ivanistheone: This seems rather longish `http://localhost:8080/viewcontent/928329382938293829382938` and defeats the purpose of having short links.

Could we use something shorter like `/vc/` ?  or `/rc/` (redirect content) ?

I'm not sure if Short URL was part of the use case; if not you can ignore this comment and the [other comment below](https://github.com/learningequality/kolibri/pull/5641/files#r291783115).

rtibbles: The main intention was for a machine parseable permalink - we can definitely extend this later for more human friendly shortlinks.

kolibri/core/content/views.py
indirectlylit: Since the goal is to allow us to change the target location, `HttpResponseRedirect` seems more correct

kolibri/core/content/views.py
```diff
@@ -319,3 +322,33 @@ def get(self, request, filename, new_filename):
...
+            if (
+                not node
+            ):  # if it's still not found, see if we can find anything with the content_id across any channel
+                node = ContentNode.objects.filter(content_id=content_id).first()
+
+        # build up the target topic/content page URL
+        if node:
+            url = None
+            for hook in ContentNodeDisplayHook().registered_hooks:
+                url = hook.node_url(node)
```
indirectlylit, line 386: is the assumption that only one plugin should ever be registered to handle content permalinks? If so, you could use 

https://github.com/learningequality/kolibri/blob/d4a4387dd195421a335516b2abc48407d07e2920/kolibri/plugins/hooks.py#L206-L209

kolibri/core/content/urls.py
```diff
@@ -16,4 +17,5 @@
         {},
         "downloadcontent",
     ),
+    url(r"^viewcontent$", ContentPermalinkRedirect.as_view(), name="contentpermalink"),
```
indirectlylit, line 20: What's the utility of putting this in `core` rather than keeping it namespaced under `learn`?

I don't have a good sense of the use-cases we expect to support by allowing any plugin to implement a `ContentNodeDisplayHook`-defined content view redirect. I'm also interested the precedent we're establishing here for other similar redirects in the future.

A quick list of semi-rhetorical questions...

* Is "view content" meant to imply a Learner display versus e.g. viewing analytics or metadata?
* If it's always Learn, does it imply a specific learning display modality? (self-contained/isolated vs contextualized in a channel, lesson, etc)?
* Will we ever want to parameterize 'view content' further to provide more specific instructions?
* Do we expect to re-use this pattern for other content or DB model types, e.g. channel, facility, user, etc?
* Are there situations where multiple plugins might legitimately register handlers for the same or similar URL patterns?

indirectlylit (commented):
As noted in the comments, I'm trying to think through the longer-term precedent this sets. It seems to me that our approach might fall under one of two high-level directions, outlined below. I'm happy to hop on a call to discuss further if that would expedite the process...

## generalized redirect system

We set up core hooks like in this PR, but generalize a bit more like a `ModelRedirectHook` that handles URLs like:

```
/ref/[model_name]/[model_id]?context=learner_view&shownav=0
```

or even more generally a `RedirectHook` that can handle

```
/redirect/[arbitrary_url_pattern]
```

Plugins can register themselves as being able to handle certain core model types or arbitrary URL patterns...

## specialized redirects

Each plugin remains responsible for maintaining their own redirects, perhaps with the help of some shared core library for consistency and DRY implementation. So in this case, the Learn plugin keeps its own manifest namespaced redirects without needing to involve the hook system, e.g.

```
/learn/redirects/viewcontent/[contentid]
```

kolibri/core/content/views.py
rtibbles: Not necessarily, no, I left it agnostic so far if in the future the nature of the redirect affords things like: conditionally redirecting depending on user agent (to offer feature phone friendly content viewing).

We could make it one only currently - but the main reason for the separation between the core redirect and the learn hook was simply to emphasize the redirection is a core functionality of Kolibri, but actually being able to render a specific content node requires a plugin that can handle this.

kolibri/core/content/urls.py
rtibbles: > Is "view content" meant to imply a Learner display versus e.g. viewing analytics or metadata?

Yes

> If it's always Learn, does it imply a specific learning display modality? (self-contained/isolated vs contextualized in a channel, lesson, etc)?

No - it is up to the individual plugin to determine how the redirect should be resolved. In the case of learn, if we wanted to get smart, we could try to see if that content item existed in any lessons currently assigned to the logged in user and redirect based on that. I have not done that, but the hook is unopinionated about whether this should happen or not.

> Will we ever want to parameterize 'view content' further to provide more specific instructions?

Yes - future versions could take additional GET params, but allow the fallback behaviour encoded here if that Kolibri version does not support these, or the higher specificity is not locally available.

> Do we expect to re-use this pattern for other content or DB model types, e.g. channel, facility, user, etc?

No - I don't think so. Content is different because it plays a very different role across the Kolibri ecosystem and appears consistently in many different Kolibri installations. It is also fundamentally public, whereas most of the other data in Kolibri is private to more or lesser a degree.

Channels might be one other place where we might offer a universal redirect, but I don't have a use case in mind for why yet.

> Are there situations where multiple plugins might legitimately register handlers for the same or similar URL patterns?

Assuming I am understanding this correctly - this is possible, but as all registered plugin urls are namespaced under the plugin name, the plugins would have to be named identically for this to occur (but we should also complain noisily - we don't currently - if this happens).

indirectlylit: > Assuming I am understanding this correctly 

This was essentially the same question as https://github.com/learningequality/kolibri/pull/5641#discussion_r292145083

basically, if multiple plugins have registered the same pattern, how do we break the tie?

rtibbles: Ah, specifically for the redirects? There is no mechanism atm, no. So putting it as only one would be the quick fix for that.

indirectlylit:
bumping this to 0.12.5 so as not to block 0.12.4

kolibri/core/content/views.py
jamalex: Agreed -- one place it can matter is caching, I think.

kolibri/core/content/urls.py
jamalex: > Channels might be one other place where we might offer a universal redirect, but I don't have a use case in mind for why yet.

And that could probably already be handled using this right? Just using the root node ID (which is generally the same as the channel ID)?

kolibri/core/content/views.py
jamalex: That could be good! In practice, might be good to always use at least 8, but could make the URLs less massive.

jamalex: In the absence of a node_id param, would need to make sure this is `""` and note `None`, though, so it doesn't blow up

kolibri/core/content/views.py
jamalex: I quite like the way it's implemented here where in principle one plugin could be set to handle certain types of content, say, but `return None` on the others so they would fall through. A bit similar to Django DB routers.

But there's no way to control precedence, right? We don't have a concept (at least that's user-controllable) of plugin "ordering", right?

jamalex (dismissed):
Looks good to me! Is it worth adding a permalink... link, below content items, so they can be copy-pasted?

One challenge we'll face is that IPs, hosts, and ports change too, or can be source-network-dependent. So "permanent" needs to be taken with a grain of salt.

kolibri/core/content/views.py
rtibbles: Wouldn't starts with "" make it return everything though?

kolibri/core/content/views.py
rtibbles: No - we bounced around wondering how to manage this with side nav items, and ended up on something that replicated the behaviour with minimal configuration required, but it was ultimately based on user permissions required to see the nav item.

kolibri/core/content/views.py
jamalex: Mmm, yeah, so it would blow up anyway... Should probably add an `if` before this section?

kolibri/core/content/views.py
```diff
@@ -319,3 +322,33 @@ def get(self, request, filename, new_filename):
...
+
+        try:  # first, try to get the node by the unique node_id
+            node = ContentNode.objects.get(id=node_id)
+        except ContentNode.DoesNotExist:  # if it isn't found, fall back to looking for the content_id in the channel
+            node = ContentNode.objects.filter(
+                channel_id=channel_id, content_id=content_id
+            ).first()
+            if (
+                not node
+            ):  # if it's still not found, see if we can find anything with the content_id across any channel
```
indirectlylit: can move the comment above the `if` to fix formatting

kolibri/core/content/hooks.py
```diff
@@ -108,3 +109,17 @@ def render_to_page_load_async_html(self):
             )
         ]
         return mark_safe("\n".join(tags))
+
+
+class ContentNodeDisplayHook(KolibriHook):
+    """
+    A hook that registers a capability of a plugin to provide a user interface
+    for a content node. When subclassed, this hook should expose a method that
+    accepts a ContentNode instance as an argument, and returns a URL where the
```
indirectlylit, line 118: Originally, I had read this comment (and the hook name '_content node_ display hook') and interpreted it to mean that the hooks expose a URL like

```/viewcontent/[node_id]```

I didn’t see that it was in fact more like

```/viewcontent?node_id=[node_id]&channel_id=[channel_id]&content_id=[content_id]```

This was the source of some of my confusion in my previous review comments, and makes me feel better about this PR, so I'll approve it.

That said, I do think there is an argument to have the plugins responsible for turning params directly into a local URL, rather than having the `ContentNodeDisplayHook` perform resolution through an intermediary channel node object. Channel nodes are not always the appropriate means of uniquely resolving the final plugin URL (e.g. in a lesson context, quiz, local curation, etc).

One strategy would be to rename `ContentNodeDisplayHook` -> `ContentDisplayHook` and to have some fixed, well-defined vocabulary of GET params parsed in the core hook, e.g. `channel_node_id`, `content_id`, `lesson_id`, etc. Then, each plugin would have direct access to these named params, and can do its own DB lookups rather than relying on being passed a `ContentNode` instance. Then, plugins could implement their own handling for cases like `{content_id, lesson_id}` versus `{channel_node_id}`:

```python
class ContentDisplayHook(KolibriHook):
    CHANNEL_NODE_ID = 'channel_node_id'
    CONTENT_ID = 'content_id'
    LESSON_ID = 'lesson_id'

    def content_url(self, content_params):
        raise NotImplementedError("This must be overridden by a subclass")


class LearnContentHook(ContentDisplayHook):
    def content_url(self, content_params):
        if this.CHANNEL_NODE_ID not in content_params:
            return None  # for example
        obj = ContentNode.objects.get(id=content_params[this.CHANNEL_NODE_ID])
        # ....
```

This might also allow for similar resolution behavior that @jamalex [described below](https://github.com/learningequality/kolibri/pull/5641/files#r292227054).

indirectlylit (approved):

kolibri/core/content/hooks.py
rtibbles: Yeah - we can definitely iterate on the hook implementation, as long as we are fine with the URL scheme, because that's the thing we shouldn't change!

Issue #5563, opened 2019-05-28
In User Page, "Search for user" filter is slow when facility has a large number of users


indirectlylit:
### Observed behavior

With ~800 users in a facility, the 'Search for a user..." box becomes unusably slow:

> ![image](https://user-images.githubusercontent.com/2367265/58498918-e661e000-8133-11e9-99bf-a8ffd0c3301f.png)

### Expected behavior

The filtering is done client-side, so it should be very snappy.

It can be throttled if necessary, but there's something deeper happening here than just updates happening too often.

### User-facing consequences

frozen browser, wasted battery

### Steps to reproduce

Try filtering users on either the beta or demo server

### Context

0.12.3

nucleogenesis:
@indirectlylit 

The issue seems to be with how many of the components we're rendering and not so much with the sorting/filtering logic.

**Testing Method**

[See this gist.](https://gist.github.com/nucleogenesis/682cd841ddc7c9c7fe484830788ce731) - You can try it by saving it to `file.js` and running `node file.js` wherever you want to locally.

I basically extracted the exact logical flow that is taken to sort the data (with a couple of variables renamed). I also took a dataset of 1030 users for this test.

When I run this with 1030 users in the data, it sorts & filters my requests in  0~75ms. This was what I first tried to see if this is the bottleneck and I'm comfortable saying that the filter/logic code itself is okay.

**Performance Profile**

It seems that the issue is somewhere in the rendering process, particularly with the KDropdownMenu and its children. See the following image profile - performed with ~1000 users.

![search-performance](https://user-images.githubusercontent.com/6356129/58578489-c72f8500-81fd-11e9-9546-f855a8310198.png)

I then removed the entire <KDropdownMenu> section from the table, which notably improved performance but wasn't really all that great. The following is a profile without the Dropdowns.

![search-performance-no-dropdown](https://user-images.githubusercontent.com/6356129/58581823-470d1d80-8205-11e9-824f-2542a7c7bc46.png)

Still - there is an average of 1.5s to load each of the <KLabeledIcons> which are just the left-most columns in the table with the icon + full name - so it's far from snappy. Subjectively speaking - the page is very much faster to load and the filter to respond - but it's not so much better that it would be acceptable in my opinion

**Where to Look Next**

1. Paginate?

If the page didn't have to render so many of each of the components, it would be much faster. Is there a general opposition to pagination for a11y or other design purposes?

[...]

PR #5647, opened 2019-06-10
Add pagination to Facility Users table

nucleogenesis:
### Summary
Added pagination to the Facility > Users page to improve performance with a large number of users. See https://github.com/learningequality/kolibri/issues/5563 for some background.

**Before**
pagination-before

**After**
pagination-after

### Reviewer guidance

1. Add a bunch of users to your installation. I did it in the kolibri shell - import `kolibri.core.auth.models` and add a bunch of FacilityUsers.
2. On the target branch, log in and go to Facility > Users and see the table take a long time to load. 
3. Try filtering - see the time it takes to re-render.

Now, checkout this branch... see the pagination. See how the page loads much more quickly. There may still be a smidge of delay between typing and seeing the filtered results.

### References

Ref: https://github.com/learningequality/kolibri/issues/5563

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/plugins/facility_management/assets/src/views/ClassEnrollForm.vue
+++ b/kolibri/plugins/facility_management/assets/src/views/ClassEnrollForm.vue
@@ -21,7 +21,7 @@
     />
 
     <nav>
-      <span>
+      <span dir="auto" class="pagination-label">
         {{ $tr('pagination', { visibleStartRange, visibleEndRange, numFilteredUsers }) }}
       </span>
       <UiIconButton
@@ -29,6 +29,8 @@
         :ariaLabel="$tr('previousResults')"
         :disabled="pageNum === 1"
         size="small"
+        :dir="isRtl ? 'rtl' : 'ltr'"
+        class="pagination-button"
         @click="goToPage(pageNum - 1)"
       >
         <mat-svg
@@ -47,6 +49,8 @@
         :ariaLabel="$tr('nextResults')"
         :disabled="pageNum === 0 || pageNum === numPages"
         size="small"
+        :dir="isRtl ? 'rtl' : 'ltr'"
+        class="pagination-button"
         @click="goToPage(pageNum + 1)"
       >
         <mat-svg
@@ -194,7 +198,13 @@
   .actions-header,
   .footer,
   nav {
-    text-align: right;
+    text-align: end;
+  }
+  .pagination-button[dir='ltr'] {
+    margin-left: 8px;
+  }
+  .pagination-button[dir='rtl'] {
+    margin-right: 8px;
   }
 
 </style>
--- a/kolibri/plugins/facility_management/assets/src/views/ClassEnrollForm.vue
+++ b/kolibri/plugins/facility_management/assets/src/views/ClassEnrollForm.vue
@@ -29,40 +29,20 @@
         :ariaLabel="$tr('previousResults')"
         :disabled="pageNum === 1"
         size="small"
-        :dir="isRtl ? 'rtl' : 'ltr'"
         class="pagination-button"
         @click="goToPage(pageNum - 1)"
       >
-        <mat-svg
-          v-if="isRtl"
-          name="chevron_right"
-          category="navigation"
-        />
-        <mat-svg
-          v-else
-          name="chevron_left"
-          category="navigation"
-        />
+        <KIcon back style="position: relative; top: -1px;" />
       </UiIconButton>
       <UiIconButton
         type="primary"
         :ariaLabel="$tr('nextResults')"
         :disabled="pageNum === 0 || pageNum === numPages"
         size="small"
-        :dir="isRtl ? 'rtl' : 'ltr'"
         class="pagination-button"
         @click="goToPage(pageNum + 1)"
       >
-        <mat-svg
-          v-if="isRtl"
-          name="chevron_left"
-          category="navigation"
-        />
-        <mat-svg
-          v-else
-          name="chevron_right"
-          category="navigation"
-        />
+        <KIcon forward style="position: relative; top: -1px;" />
       </UiIconButton>
     </nav>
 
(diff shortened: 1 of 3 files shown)


jonboiser:
The pagination controls and counters can be incorrect when keyword filtering. Might be some of the logic in the new computed properties.

ACtually, the filtering looks wrong too, i'm sure there are a lot of users with `123` in their user name.

The `sortedFilteredUsers` properties looks correct, but the data is not being put in the table.

Screen Shot 2019-06-13 at 12 05 26 PM

nucleogenesis:
> The pagination controls and counters can be incorrect when keyword filtering. Might be some of the logic in the new computed properties.
> 
> ACtually, the filtering looks wrong too, i'm sure there are a lot of users with `123` in their user name.
> 
> The `sortedFilteredUsers` properties looks correct, but the data is not being put in the table.
> 
> Screen Shot 2019-06-13 at 12 05 26 PM

@jonboiser when you did that screenshot - did you start typing the filter after navigating to deeper pages? I got the same effect as you and put a watcher on the `searchFilter` that resets the pageNum to the first page whenever the `searchFilter` changes, which seems to avoid the issue.

jonboiser:
@nucleogenesis 	That seems to fix it. For me. Just fix the ESLint error and we're good to go.

We're going to wait until after 0.12.5 to release this change, so not going to merge it immediately.

jonboiser:
Since there's some time, could you make two changes:

1. Add an entry in the CHANGELOG for 0.12.6 
1. Add some margin next to the buttons and the pagination label

Screen Shot 2019-06-19 at 2 37 12 PM

1. Check to see if a `dir =auto` attribute is applied to the label. It looks odd in RTL mode:

Screen Shot 2019-06-19 at 2 38 35 PM

kolibri/plugins/facility_management/assets/src/views/UserPage/index.vue
```diff
@@ -48,11 +43,32 @@
       </template>
     </UserTable>
 
+    <nav>
+      <span>{{ $tr('pagination', { visibleStartRange, visibleEndRange, numFilteredUsers }) }}</span>
```
jonboiser: add a `dir="auto"` attribute to fix the way numbers appear in RTL mode. Also add a little bit of margin against the two UIBUttons.

jonboiser: A little bit of space _between_ the buttons might be nice too.

jonboiser: You can copy whatever the margin values are from ClassEnrollForm.vue. And if you add spacing between the buttons, we should copy those over to ClassEnrollForm form for consistency.

nucleogenesis: Thanks @jonboiser I added some spacing to both files and added a dir attr to the buttons as well to more easily target the different styles with appropriate button margins. By putting the margins on the buttons, the label text is also spaced apart better from the buttons now.

kolibri/plugins/facility_management/assets/src/views/ClassEnrollForm.vue
```diff
@@ -194,7 +198,13 @@
   .actions-header,
   .footer,
   nav {
-    text-align: right;
+    text-align: end;
+  }
+  .pagination-button[dir='ltr'] {
```
jonboiser: you might need to make separate rules for the "previous" and "next" buttons. The way this is, in RTL mode the "next" button (i.e. the one on the left in RTL) gets left margin and the "previous" button gets no right margin, so it is squished against the label.

kolibri/plugins/facility_management/assets/src/views/ClassEnrollForm.vue
```diff
@@ -21,14 +21,16 @@
...
-      <span>
+      <span dir="auto" class="pagination-label">
         {{ $tr('pagination', { visibleStartRange, visibleEndRange, numFilteredUsers }) }}
       </span>
       <UiIconButton
         type="primary"
         :ariaLabel="$tr('previousResults')"
         :disabled="pageNum === 1"
         size="small"
+        :dir="isRtl ? 'rtl' : 'ltr'"
```
indirectlylit: what does adding a `dir` attribute to the icon button do?

kolibri/plugins/facility_management/assets/src/views/UserPage/index.vue
```diff
@@ -48,11 +43,38 @@
...
+        type="primary"
+        :ariaLabel="$tr('nextResults')"
+        :disabled="pageNum === 0 || pageNum === numPages"
+        size="small"
+        class="pagination-button"
+        :dir="isRtl ? 'rtl' : 'ltr'"
+        @click="goToPage(pageNum + 1)"
+      >
+        <mat-svg v-if="isRtl" name="chevron_left" category="navigation" />
+        <mat-svg v-else name="chevron_right" category="navigation" />
```
indirectlylit: instead of using `v-if`/`v-else`, recommend using the `is-rtl` global css class to flip the icon conditionally.

Alternatively, you could update the `KIcon` component to handle this icon

indirectlylit (dismissed):
I've added some developer documentation to shed some light on the nuances of RTL:

https://github.com/learningequality/kolibri/pull/5710

Recommend taking a look at that and some other precedent from around the app to tighten up the RTL handling. Let me know if you'd like help.

kolibri/plugins/facility_management/assets/src/views/ClassEnrollForm.vue
nucleogenesis: I added it because I was using attr selectors in the CSS a la `.class[dir='rtl']` - however, I was using hot-reloading so I was definitely seeing something different than what would be expected and - now that I'm not hot-reloading, I see the dir gets added automatically anyway.

nucleogenesis: This allows me to target the styles using an attr selector a la `.pagination-button[dir='rtl'] {}`

kolibri/plugins/facility_management/assets/src/views/ClassEnrollForm.vue
nucleogenesis: @jonboiser I was using hot reloading, which I suspect caused some of the issues where things worked out right on my end but not on yours.

Now that I did the dev using the normal dev server, I revised everything and it should be good to go.

jonboiser (approved):
looks good!

Issue #5646, opened 2019-06-10
sentry issue on ubuntu 14.04


indirectlylit:
### Observed behavior

Observed:

![image](https://user-images.githubusercontent.com/2367265/59227959-97409400-8ba4-11e9-96f7-49c9c674624c.png)

See also:

* https://learningequality.slack.com/archives/CB37UM23A/p1560165318063600
* https://github.com/learningequality/kolibri/issues/4775
* https://github.com/learningequality/kolibri/pull/5346
* https://github.com/learningequality/kolibri-installer-debian/pull/67

### Expected behavior

Installs the same as in 0.12.3.

To-do:

* [x] @indirectlylit and @jredrejo to remove tornado.py in the debian package: https://github.com/learningequality/kolibri-installer-debian/blob/master/debian/rules#L35
* [ ] ~@indirectlylit to try downgrading sentry~ choosing not to downgrade sentry unless absolutely necessary to avoid additional testing

### User-facing consequences

users cannot upgrade

### Steps to reproduce

Install debian package

### Context

ubuntu 14.04

indirectlylit:
cc @mrpau-dev @radinamatic

indirectlylit:
14.04 is not currently part of our test suite, but we have users using it, so perhaps it should be?

radinamatic:
I understand that this is on a case basis, but do we have a plan/strategy until when we are willing to support the OSes that are EOL?

benjaoming:
> 14.04 is not currently part of our test suite, but we have users using it, so perhaps it should be?

Nah, I don't think so.. there should be advantages of dropping support of something? :) But detecting such failure of compiling into Python bytecode is interesting. Added tests for that above.

benjaoming:
We also continue to slightly support 14.04 by supporting Python 3.4.

indirectlylit:
In practice, we have a large number users currently on 14.04 so our dropped support was preemptive. We'll need to keep testing it until we are sure that they are mostly migrated away, or we'll end up simply paying the price later.

benjaoming:
> In practice, we have a large number users currently on 14.04 so our dropped support was preemptive. 

Supporting people using an EOL operating system.. let's talk about this Monday next week :)

PR #5652, opened 2019-06-11
Add pycompile in Travis matrix for Python 3 versions

benjaoming:
### Summary

Try to add better upfront guarantees about ability to compile.

This was due to `async` keywords included in some bundled dependencies that broke on Python 3.4.

See:

https://docs.python.org/3/library/compileall.html

### Reviewer guidance

n/a

### References

https://github.com/learningequality/kolibri/issues/5646

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -110,11 +110,10 @@ before_script:
 # command to run tests
 script:
   # Get static deps in order to try and compile them all for Python 3
-  - make staticdeps
+  - SKIP_PY_CHECK=1 make staticdeps
   # Ensures that we can compile all source code files in this Python 3+ version
   # This behavior is disabled in the 'include' matrix above for Python 2.7
-  - python -m compileall -x py2only kolibri -q
-  - python --version
+  - make staticdeps-compileall
   - tox -e $TOX_ENV
 
 after_success:
--- a/.travis.yml
+++ b/.travis.yml
@@ -114,6 +114,8 @@ script:
   # Ensures that we can compile all source code files in this Python 3+ version
   # This behavior is disabled in the 'include' matrix above for Python 2.7
   - make staticdeps-compileall
+  # Remove staticdeps again, so they are not part of tox
+  - make clean-staticdeps
   - tox -e $TOX_ENV
 
 after_success:
--- a/.travis.yml
+++ b/.travis.yml
@@ -109,6 +109,8 @@ before_script:
 
 # command to run tests
 script:
+  # Get static deps in order to try and compile them all for Python 3
+  - make staticdeps
   # Ensures that we can compile all source code files in this Python 3+ version
   # This behavior is disabled in the 'include' matrix above for Python 2.7
   - python -m compileall -x py2only kolibri -q
--- a/.travis.yml
+++ b/.travis.yml
@@ -44,6 +44,8 @@ matrix:
     - python: "3.5"
       env:
         - TOX_ENV=docs
+      script:
+       - tox -e $TOX_ENV
     - python: "3.4"
       env:
         - TOX_ENV=py3.4
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,6 +23,8 @@ matrix:
     - python: "2.7"
       env:
         - TOX_ENV=pythonlint2
+      script:
+       - tox -e $TOX_ENV
     - python: "3.6"
       env:
         - TOX_ENV=pythonlint3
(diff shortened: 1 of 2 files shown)


rtibbles:
Wait, did we fix this already? I thought it was still extant, if so, why is the test passing?

.travis.yml
```diff
@@ -99,6 +109,9 @@ before_script:
 
 # command to run tests
 script:
+  # Ensures that we can compile all source code files in this Python 3+ version
+  # This behavior is disabled in the 'include' matrix above for Python 2.7
+  - python3 -m compileall -x py2only kolibri -q
```
rtibbles: By this point the travis venv has already been activated, so this should just be `python` not `python3` to trigger the matrix set python version. I think this may be why it is passing when we have not fixed this issue yet.

benjaoming: Good point -- after opening the PR I didn't have time to wait around for the Travis results.

Will try changing it to `python` and add a temporary `python --version`

benjaoming:
> Wait, did we fix this already? I thought it was still extant, if so, why is the test passing?

Yes, it's supposed to be failing :/ Then I want to merge `release-v0.12.y` after a fix is committed so we can see that this is a safe-guard.

rtibbles:
The immediate fix has only been applied on the debian installer repo - I'll be doing a fix on develop to remove the sentry integration to a plugin, so may be best to retarget to there!

benjaoming:
Check!

benjaoming:
Still not sure why this compiles. It fails perfectly on my local box. Will try to get access to Python 3.5.

Travis:

image

Local:

```
➜  kolibri git:(travis-pycompile) ✗ make staticdeps-compileall
bash -c 'python --version'
Python 2.7.15+
# Seems like the compileall module does not return a non-zero exit code when failing
bash -c 'if ( python -m compileall -x py2only kolibri -q | grep SyntaxError ) ; then echo "Failed to compile kolibri/dist/" ; exit 1 ; else exit 0 ; fi'
SyntaxError: invalid syntax
SyntaxError: invalid syntax
SyntaxError: invalid syntax
Failed to compile kolibri/dist/
Makefile:151: recipe for target 'staticdeps-compileall' failed
make: *** [staticdeps-compileall] Error 1
```

benjaoming:
@rtibbles @indirectlylit  this is now correctly failing ONLY for the on Python 3.4 build:

image

I think it was fixed by removing "tornado" in the Debian installer by @jredrejo - but wouldn't we need a more source-oriented fix such that we don't distributed sentry-sdk in our source dist? Can we remove it from `base.txt` and add it to a separate location?

rtibbles:
My suggestion, and what I think is the simplest solution, is to turn the sentry integration into an external plugin. With the enhancements to the plugin API, I think we could make the switch with no change of configuration on the our demo servers except installing and activating the plugin.

rtibbles:
Unless we are still using the pex file...

benjaoming:
So provided that #5773 is merged, we can merge `develop` into this PR and the tests should start working again :)

benjaoming:
Sorry I had initially fine builds, only the correct failure on Python 3.4, but now I see this on 3.5+...

image

...will investigate

benjaoming:
Here we go:

```
*** Error compiling 'kolibri/dist/morango/certificates_x509_old.py'...
  File "kolibri/dist/morango/certificates_x509_old.py", line 123
    print extract_morango_scopes(cert)
                               ^
SyntaxError: invalid syntax
```

Haven't we fixed this before?

benjaoming:
Okay, this was fixed in a more recent Morango release that isn't bumped in our requirements, ref'ing: #5775

benjaoming:
Updated with latest `develop` containing morango version bump.........

benjaoming:
Just Python 3.4 is failing now, and that should be because of the sentry plugin, which is removed in #5773

benjaoming:
Sorry, I force-pushed to get a cleaner commit history, so the trace of former build errors is gone.

What changed was that new `develop` is in, meaning that `sentry_sdk` is removed.

benjaoming:
The adjusted tests w/ pycompile are now passing after `sentry_sdk` got removed.

Meaning that we now have a safe-guard as to whether things are Python 3 friendly. Note that the behavior for now has been disabled for Python 2.7, since the immediate necessity was for pre-compilation of Python 3 packages on Debian.

rtibbles (approved):

PR #5660, opened 2019-06-12
Zeroconf-based Kolibri P2P discovery system

jamalex:
### Summary

We support local network peer connections for content import, and soon, for data syncing. But right now, it's necessary to manually enter the IP and port, which is at best inconvenient, and at worst an insurmountable barrier for many users.

Enter [zeroconf](https://blog.hyperiongray.com/multicast-dns-service-discovery/)! It's specifically designed for this type of situation -- allowing nearby devices on the same network to discover one another and the services they offer.

This PR is a preliminary, but usable, backend implementation of a zeroconf-based "Kolibri service broadcasting and discovery" system. When Kolibri starts up, it will register itself on the local network, and then monitor and track any other instances of Kolibri on the same local network. The frontend code can then call an API endpoint to enumerate the nearby Kolibri instances, including information about what IP/port they're listening on, and what facilities/channels they have loaded on them.

One of the lovely side-effects of using zeroconf (aka mDNS) for this is that it also ends up creating local 
area domain names that can be resolved and used in the browser at least on Linux and OSX:
image

Note: includes some changes I made to https://github.com/learningequality/python-zeroconf

### Reviewer guidance

To test, you'll want to spin up a couple instances of this. To start with, these could be on the same computer, just using different ports and KOLIBRI_HOME directories, e.g.:
```
# run this in one terminal tab
kolibri start

# run this in another tab
KOLIBRI_HOME=~/.kolibri-temp KOLIBRI_LISTEN_PORT=9090 kolibri start
```

Then, login as an admin on each and go to:
http://127.0.0.1:8080/api/discovery/networksearch/
and
http://127.0.0.1:9090/api/discovery/networksearch/

and you should see something like this:

```
[
    {
        "ip": "192.168.1.25",
        "self": true,
        "data": {
            "channels": [
                {
                    "id": "c7eda62c6489554a941058fa883e7c2c",
                    "name": "Better World Ed"
                },
                {
                    "id": "74f36493bb475b62935fa8705ed59fed",
                    "name": "Thoughtful Learning"
                },
                {
                    "id": "f9da12749d995fa197f8b4c0192e7b2c",
                    "name": "PraDigi"
                }
            ],
            "facilities": [
                {
                    "dataset_id": "d265a3809dd02de42570ef5ba5b1bb86",
                    "id": "318aa7335a2ad82434dee8534caf85d9",
                    "name": "a"
                }
            ]
        },
        "port": 8080,
        "host": "b053.kolibri.local",
        "local": true,
        "id": "b053"
    }
]
```

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated

TODO:
- [ ] Add an option in options.ini to disable zeroconf
- [ ] Consider switching service type to `_http._tcp.local` so it's standardized.
- [ ] Rebroadcast service when channels or facilities change

### DIFF
--- a/kolibri/core/discovery/utils/network/search.py
+++ b/kolibri/core/discovery/utils/network/search.py
@@ -16,7 +16,7 @@ from kolibri.core.content.models import ChannelMetadata
 
 logger = logging.getLogger(__name__)
 
-SERVICE_TYPE = "_kolibri._http._tcp.local."
+SERVICE_TYPE = "Kolibri._sub._http._tcp.local."
 LOCAL_DOMAIN = "kolibri.local"
 
 ZEROCONF_STATE = {"zeroconf": None, "listener": None, "service": None}
--- a/kolibri/core/discovery/utils/network/search.py
+++ b/kolibri/core/discovery/utils/network/search.py
@@ -113,6 +113,7 @@ class KolibriZeroconfListener(object):
             "port": info.port,
             "host": info.server.strip("."),
             "data": {key: json.loads(val) for (key, val) in info.properties.items()},
+            "base_url": "http://{ip}:{port}/".format(ip=ip, port=info.port),
         }
         logger.info(
             "Kolibri instance '%s' joined zeroconf network; service info: %s\n"
--- a/kolibri/core/discovery/utils/network/search.py
+++ b/kolibri/core/discovery/utils/network/search.py
@@ -1,0 +1,174 @@
+import atexit
+import json
+import logging
+import socket
+import time
+from contextlib import closing
+
+from zeroconf import get_all_addresses
+from zeroconf import NonUniqueNameException
+from zeroconf import ServiceInfo
+from zeroconf import USE_IP_OF_OUTGOING_INTERFACE
+from zeroconf import Zeroconf
+
+from kolibri.core.auth.models import Facility
+from kolibri.core.content.models import ChannelMetadata
+
+logger = logging.getLogger(__name__)
+
+SERVICE_TYPE = "_kolibri._http._tcp.local."
+LOCAL_DOMAIN = "kolibri.local"
+
+ZEROCONF_STATE = {"zeroconf": None, "listener": None, "service": None}
+
+
+def _id_from_name(name):
+    assert name.endswith(SERVICE_TYPE), (
+        "Invalid service name; must end with '%s'" % SERVICE_TYPE
+    )
+    return name.replace(SERVICE_TYPE, "").strip(".")
+
+
+def _is_port_open(host, port, timeout=1):
+    with closing(socket.socket(socket.AF_INET, socket.SOCK_STREAM)) as sock:
+        sock.settimeout(timeout)
+        return sock.connect_ex((host, port)) == 0
+
+
+class KolibriZeroconfService(object):
+
+    info = None
+
+    def __init__(self, id, port=8080, data={}):
+        self.id = id
+        self.port = port
+        self.data = {key: json.dumps(val) for (key, val) in data.items()}
+        atexit.register(self.cleanup)
+
+    def register(self):
+
+        if not ZEROCONF_STATE["zeroconf"]:
+            initialize_zeroconf_listener()
+
+        assert self.info is None, "Service is already registered!"
+
+        i = 1
+        id = self.id
+
+        while not self.info:
+
+            # attempt to create an mDNS service and register it on the network
+            try:
+                info = ServiceInfo(
+                    SERVICE_TYPE,
+                    name=".".join([id, SERVICE_TYPE]),
+                    server=".".join([id, LOCAL_DOMAIN, ""]),
+                    address=USE_IP_OF_OUTGOING_INTERFACE,
+                    port=self.port,
+                    properties=self.data,
+                )
+
+                ZEROCONF_STATE["zeroconf"].register_service(info, ttl=60)
+
+                self.info = info
+
+            except NonUniqueNameException:
+                # if there's a name conflict, append incrementing integer until no conflict
+                i += 1
+                id = "%s-%d" % (self.id, i)
+
+            if i > 100:
+                raise NonUniqueNameException()
+
+        self.id = id
+
+        return self
+
+    def unregister(self):
+
+        assert self.info is not None, "Service is not registered!"
+
+        ZEROCONF_STATE["zeroconf"].unregister_service(self.info)
+
+        self.info = None
+
+    def cleanup(self, *args, **kwargs):
+
+        if self.info and ZEROCONF_STATE["zeroconf"]:
+            self.unregister()
+
+
+class KolibriZeroconfListener(object):
+
+    instances = {}
+
+    def add_service(self, zeroconf, type, name):
+        info = zeroconf.get_service_info(type, name)
+        id = _id_from_name(name)
+        ip = socket.inet_ntoa(info.address)
+        self.instances[id] = {
+            "id": id,
+            "ip": ip,
+            "local": ip in get_all_addresses(),
+            "port": info.port,
+            "host": info.server.strip("."),
+            "data": {key: json.loads(val) for (key, val) in info.properties.items()},
+        }
+        logger.info(
+            "Kolibri instance '%s' joined zeroconf network; service info: %s\n"
+            % (id, self.instances[id])
+        )
+
+    def remove_service(self, zeroconf, type, name):
+        id = _id_from_name(name)
+        logger.info("\nKolibri instance '%s' has left the zeroconf network.\n" % (id,))
+        if id in self.instances:
+            del self.instances[id]
+
+
+def get_available_instances(timeout=2, include_local=True):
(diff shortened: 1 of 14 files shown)


kolibri/core/discovery/utils/network/search.py
```diff
@@ -0,0 +1,174 @@
...
+        if not _is_port_open(instance["ip"], instance["port"], timeout=timeout):
+            continue
+        instance["self"] = (
+            ZEROCONF_STATE["service"] and ZEROCONF_STATE["service"].id == instance["id"]
+        )
+        instances.append(instance)
+    return instances
+
+
+def register_zeroconf_service(port, id):
```
jredrejo, line 149: Why does this function need the port as a parameter?
Is not possible to get it from the options or reading it from the `PID_FILE` as it is done in other parts of the code?
This `port` option pushes changes in many parts of this PR code that might not be needed.

kolibri/utils/cli.py
```diff
@@ -639,21 +639,31 @@ def _get_port(port):
     return int(port) if port else OPTIONS["Deployment"]["HTTP_PORT"]
 
 
+def _cleanup_before_quitting(signum, frame):
```
jredrejo, line 647: what happens if there's a network outage or a sudden power off of the computer runing Kolibri? Is the announce of the service removed automatically, if not, how long is it kept?

jamalex: There's a 60 second TTL, though in practice it seems to take at least several minutes to "forget".

jamalex: And yes, it no longer announces no matter how it is killed. This is just to have it actively announce its departure so records are updated immediately.

kolibri/core/discovery/utils/network/search.py
jamalex: Ah, PID_FILE could be a good place, yes. I originally wanted to pull it from `conf.OPTIONS`, except it can be overridden with `--port=XYZ`, which doesn't get stored anywhere (except in the PID file, I guess), just passed in to the start command function as an argument, which was the pattern I followed.

kolibri/utils/server.py
```diff
@@ -100,7 +107,7 @@ def start(port=8080, run_cherrypy=True):
     :param: port: Port number (default: 8080)
     """
 
-    run_services()
+    run_services(port=port)
```
benjaoming, line 110: Possibly new Python argument patterns confuse me a bit, so not sure if I'm right here: But the function defines this as positional argument, and is then called with a named argument that will be fed as a positional one?

kolibri/core/discovery/utils/network/search.py
jredrejo: Yes, I understand your concerns but, as I understand the code, the value in the PID file is affected both by conf.OPTIONS and env variables, thus it should be the real value for the port. Taking it from there would avoid many modifications you had to do just to drag the port parameter here.

requirements/base.txt
```diff
@@ -33,3 +33,4 @@ sentry-sdk==0.7.9
 django-redis-cache==2.0.0
 redis==3.2.1
 html5lib==1.0.1
+git+https://github.com/learningequality/python-zeroconf.git@8d0e70029a925862e965519c9d42b4016d694d5c#egg=zeroconf
```
benjaoming: We talked about this yesterday, and do I remember correctly that we wanted to change this?

It requires Git to be locally available at build time, and I think these types of dependencies have been seen as either insecure or just an anti pattern. Could releasing this as a PyPi project `zeroconf-2and3` work?

I think we're seeing more and more small ponds of reasons why we'd want to drop Python 2 support, so the choice of a "2 and 3" for a forked project should not be seen as fateful, anyways.

benjaoming (commented):
This looks so promising!! Am already thinking about what it can contribute to in terms of documentation simplification about static IPs -- that if MDNS is supported widely (IDK) then we could implement the naming of a device in the setup process `<unique-device-name>.kolibri.local`.

benjaoming:
~@jamalex a quick idea: When using custom names, could the device try to look up its own name and report back if there is a conflict?~

Scratch that, you already knew :100:

kolibri/core/discovery/utils/network/search.py
```diff
@@ -0,0 +1,175 @@
...
+
+    def register(self):
+
+        if not ZEROCONF_STATE["zeroconf"]:
+            initialize_zeroconf_listener()
+
+        assert self.info is None, "Service is already registered!"
+
+        i = 1
+        id = self.id
```
benjaoming, line 56: This assigns to a builtin, or is it just me?

kolibri/core/discovery/utils/network/search.py
jamalex: Yep, agreed! Can't see any issues with that approach. Will update.

The PID is also written into there for kolibri-server, yes? Based on what uwsgi/nginx are exposing?

jamalex:
>This looks so promising!! Am already thinking about what it can contribute to in terms of documentation simplification about static IPs -- that if MDNS is supported widely (IDK) then we could implement the naming of a device in the setup process <unique-device-name>.kolibri.local

It's sooooooo close to a possibility, EXCEPT that Android doesn't implement any version of mDNS/zeroconf/etc.... and that's one of our most common client devices. :crying_cat_face: 

What I'm thinking is that we could make lightweight "launcher" apps for Android/Windows/Linux that do nothing other than quickly scan the local network and find the IP for a Kolibri instance (possibly caching/preferring a specific one after first use), and then pop it open in the web browser. Would be a huge timesaver and headache-avoider in itself.

kolibri/utils/server.py
jamalex: Perhaps, yeah -- sometimes I like to be explicit with named arguments, as it makes reading the function call clearer. In this case it doesn't add much value since the variable has the same name as the argument.

requirements/base.txt
jamalex: I was preparing it for release as `zeroconf-py2compat` yesterday, but didn't finish yet.

kolibri/core/discovery/utils/network/search.py
jamalex: Seems so, yep! Perhaps I'll rename the local var to `unique_id`

requirements/base.txt
```diff
@@ -33,4 +33,4 @@ sentry-sdk==0.7.9
 django-redis-cache==2.0.0
 redis==3.2.1
 html5lib==1.0.1
-git+https://github.com/learningequality/python-zeroconf.git@8d0e70029a925862e965519c9d42b4016d694d5c#egg=zeroconf
+zeroconf-py2compat==0.19.3
```
benjaoming, line 36: :tada:

kolibri/core/discovery/utils/network/search.py
```diff
@@ -0,0 +1,175 @@
...
+            "base_url": "http://{ip}:{port}/".format(ip=ip, port=info.port),
+        }
+        logger.info(
+            "Kolibri instance '%s' joined zeroconf network; service info: %s\n"
+            % (id, self.instances[id])
+        )
+
+    def remove_service(self, zeroconf, type, name):
+        id = _id_from_name(name)
+        logger.info("\nKolibri instance '%s' has left the zeroconf network.\n" % (id,))
```
benjaoming, line 125: The newline at the beginning seems wrong?

kolibri/core/discovery/utils/network/search.py
jredrejo: kolibri-server  takes the value from conf.OPTIONS because there's no warranty the part of code in Kolibri that saves the PID has been executed before kolibri-server starts. The env variable option does not have sense when running it as a daemon.
After kolibri finishes the starting process, the PID file will have the same value that is in conf.OPTIONS so I think we can consider we are safe on this side too.

kolibri/core/discovery/utils/network/search.py
benjaoming: Oh and the assigned to a builtin once again. @indirectlylit I think that removing Flake 8 was a mistake, since it checks many things that Black doesn't check. Assignments, code complexity, unused variables.

image

benjaoming:
@jamalex could you try running `pre-commit install` on your checkout? Then reset the PR branch unto the current `develop`, commit everything again and see how it lints?

indirectlylit:
@rtibbles how do you think this work will be affected by https://github.com/learningequality/kolibri/pull/6007

rtibbles:
Looks like any background tasks are managed inside zeroconf - so, apart from superficial merge conflicts, I don't think it would be affected.

indirectlylit:
> Looks like any background tasks are managed inside zeroconf - so, apart from superficial merge conflicts, I don't think it would be affected.

okay cool. wasn't sure if you were going to want to leverage iceqube machinery for this

indirectlylit:
migrating to #6011 and https://github.com/learningequality/kolibri/tree/zeroconf so that @micahscopes can work on it

PR #5681, opened 2019-06-19
Allow plugins to add to Django settings and the core options spec

rtibbles:
### Summary
* Decouples plugin initialization from Django app initialization - this can now happen completely independently of Django.
* Adds the capability to autodiscover a `settings.py` file in the plugin folder (parallel to the autodiscovery of `urls.py` and `api_urls.py`) and add those settings to the default settings module specified for Django.
* Adds the capability to autodiscover an `options.py` file in the plugin folder, and if it contains an `options_spec` object, add that to the core options_spec used by Kolibri. This allows plugins to have options specified in the `options.ini` file.
* The options parsing above also allows overwriting of default values of core Kolibri options - any other attempt to change the options spec will result in errors, but overwriting the default value is OK.
* All validation is still delegated to the core options parser, so a poorly written plugin could break Kolibri until the plugin is disabled.

### Reviewer guidance
* Create a `settings.py` file in a plugin and test that values that are not specified in the core Kolibri settings can be added to the django settings in `kolibri shell`.
* Create an `options.py` file in a plugin with an options_spec object (see the core `options.py` file for the spec it should follow) to add additional options and overwrite a core option default. Verify this from the shell.

### References
This PR depends on #5637

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated


rtibbles:
Updating to 0.12.6 as needed for UNW work.

kolibri/plugins/registry.py
```diff
@@ -19,8 +19,8 @@
     dumped and exported easily.
 
 From a developer's perspective, plugins are Django applications listed
-in ``INSTALLED_APPS`` and are initialized once when the server starts, mean at
-the load time of the django project, i.e. Kolibri.
+in ``INSTALLED_APPS`` and are initialized once when the server starts, before
+the load time of Django.
```
benjaoming: Not sure that I understand this part? How can the plugin be instantiated before Django loads?

kolibri/plugins/registry.py
```diff
@@ -30,8 +30,8 @@
 
 .. note::
 
-    Each app in ``INSTALLED_APPS`` is searched for the special
-    ``kolibri_plugin`` module.
+    Each app in ``ACTIVE_PLUGINS`` in the kolibri conf is searched for the
+    special ``kolibri_plugin`` module.
```
benjaoming, line 35: Will `INSTALLED_APPS` still be the manifesto of active Django applications? Does a plugin still have to be a Django application? Is it expected to be one?

It also seems that other modules are now automatically discovered and loaded. It's important to clarify this, because it complicates the load order by setting up certain assumptions that may be difficult for plugin authors to understand and anticipate.

kolibri/plugins/base.py
```diff
@@ -131,6 +142,31 @@ def api_url_module(self):
...
+        Return a settings module, containing Django settings that this
+        module wants to apply.
+
+        For settings that take a tuple or list, these will be appended to the value from
+        the base settings module set through conventional Django means.
+
+        By default settings.py will be automatically discovered, if you wish to
+        override this behaviour, override this method.
+        """
+        return self._return_module("settings")
```
benjaoming: This is where I get most concerned.

kolibri/plugins/base.py
```diff
@@ -131,6 +142,31 @@ def api_url_module(self):
...
+        Return an options module, containing a config spec as the 'option_spec' value.
+
+        These options should not override the core config spec, but may specify only a new
+        default value for a core config spec option.
+
+        By default options.py will be automatically discovered, if you wish to
+        override this behaviour, override this method.
+        """
+        return self._return_module("options")
+
```
benjaoming, line 224: What if instead we had a separate options file for persistent plugin configuration? `plugin_options.ini` or similar?

benjaoming (commented):
Exciting changes! I'd prefer that this part of architecture went through a proposal process first, but I can also see your proposal quite clearly as code :+1: 

## Architecture concerns

Django apps will not normally change Django settings, since this is a shared environment. They will of course read out settings from a Django project, but they will leave it up to the project to set these values. It's the application's responsibility to define and document its own settings, but never to automatically set them.

It's a principle of choices between clearly defined architecture and delegation of responsibility vs. an unpredictable slope towards plugins not behaving as _plugins_. Sorry about the drama and the bias, but now you know that I'm a bit fundamentalist about this :)

Regarding the automatic detection of `settings.py` and writing of project settings, this was deliberately left out. 1) Because other reusable Django apps don't do it and 2) Because it lead to [a settings chaos in KA Lite](https://github.com/learningequality/ka-lite/blob/master/kalite/settings/base.py#L467). KA Lite applications (albeit undefined plugin API) contained project settings rather than the project, and it was unclear which order their settings should be imported in. It gave developers the wrong set of ideas when for instance they though of a MIDDLEWARE or TEMPLATE_CONTEXT_PROCESSOR to add.

That's also why we have `kolibri.deployment`: To define a set of default project settings for a default set of applications (plugins).

Hooks are defined with a specific purpose in mind, can we solve what we are trying to achieve inside this pattern?

As Kolibri developers, we can also fear our own plugins in the sense that they might lock down development because important plugins start to expect, manage or hog certain aspects of the settings stack. Say, for instance, if a plugin adds or removes entries in `INSTALLED_APPS` or `MIDDLEWARE`. This can lead to all sorts of issues which are hard to solve, as plugins do not specify their Kolibri compatibility and we won't be able to depreciate elements of the proposed `plugin.settings` pattern, whereas for instance a hook can be deprecated or deactivated.

### New questions of precedence and responsibility

I try to imagine some plugins. One plugin A, wants to add middleware that logs a teacher in if it has a configured MAC address. Lets imagine that it sets a welcome message "you were logged in automatically", sensitive to current user language. Then, plugin B is interested in setting up a language based on the IP, for instance in a context with telcos or school networks operating in different languages.

I find the example interesting, because plugin A simply appends to the list of middleware and thinks it's safe. But then plugin B comes along and ruins the assumption :)

plugin_a.options

```python
MIDDLEWARE_CLASSES = ["some.middleware.LoginUserIfTeachersMAC"]
```

plugin_b.options

```python
MIDDLEWARE_CLASSES = ["some.middleware.SetLocaleBasedOnIP"]
```

Which comes first? And should plugins be able to shuffle or remove other entries? Should they be able to read existing values and make conditional changes?

## Plugin author concerns

Am also trying to understand this from a plugin author's perspective:

 * What is the difference between "options" and "settings"? :) I think it's fair to expect that a plugin author would be confused immediately :)
 * Do we maintain the simplest possible way of defining our plugin API that will make it as easy as possible to be a plugin author and maintain our own documentation and example plugin?
 * Can a plugin's `settings.py` import `django.conf.settings` - I assume not, but then the plugin probably wants to have another mechanism for writing default settings based on the contents of `django.conf.settings` ?
 * If plugins are writing their settings schema to Kolibri's `options.ini`, then they have to be managed according to a plugin's independent release cycle.. should we impose very strict limits to how this is expected to work? I mean, for instance plugins should not use the Kolibri options file for storage of user data, nor should they create heaps of schema changes, prompting automatic fixups of `options.ini` and such.

rtibbles:
My basic rationale here is:

Kolibri is a wrapper around a set of Django apps. We customize the behaviour of these Django apps using our options ini to modify settings in a more user friendly way/unified way that allows them to control non-Django behaviour as well through a single interface.

Kolibri plugins are wrappers around one or more Django apps. By plugging into Kolibri they add these Django apps to the set of Django apps that Kolibri wraps. In order to provide default behaviours for these Django apps, sometimes Django settings need to be defined (one example is the work in this PR: https://github.com/learningequality/kolibri/pull/5625 - where the open id connector Django app being used requires settings to be defined to work properly). So, in order for plugins to work properly, sometimes Django settings need to be augmented.

In the current implementation of this PR, plugins are not allowed to override any settings that are defined in the specified `DJANGO_SETTINGS_MODULE`. I think this is a potential weakness, because it means that plugins could unexpectedly break because someone uses a different settings module that then has a clash with the plugin settings. On reflection, this PR could probably be improved by not dynamically modifying whichever `DJANGO_SETTINGS_MODULE` has been defined, and only applying these settings to our `kolibri.deployment.default.settings.base` module.

The only instance in which the plugin can modify existing settings in the `DJANGO_SETTINGS_MODULE` is to add to `INSTALLED_APPS`, `TEMPLATE_LOADERS` and `LOCALE_PATHS` - I think we could make the API more restrictive, and also separate anything that augments existing properties (rather than adding new ones) into separate explicit properties of the plugin if we wanted to.

For concerns re: `MIDDLEWARE` or removing things from `INSTALLED_APPS` etc - this is not possible with the implementation I have here, as it is only additive in nature for the three defined pieces above, and I excluded `MIDDLEWARE` precisely due to the order concerns.

I'll add additional inline comments where I think I could make changes to the current implementation and to highlight where your concerns are possibly pre-empted!

kolibri/plugins/base.py
```diff
@@ -44,6 +46,15 @@ class KolibriPluginBase(object):
     # : Suggested property, not yet in use
     collect_static_on_enable = False
 
+    def __init__(self):
+        if settings.configured:
+            # Check to see if a plugin is being initialized after Django
+            warnings.warn(
```
rtibbles, line 75: This warning helps to identify to the plugin author if their plugin is not being initialized at the right time in order for auto-discovered modules (such as settings and options) to have an effect on the operation of Kolibri.

kolibri/plugins/base.py
rtibbles: This is a convenience method - this can be directly defined as a particular module if so desired.

Could also add a try catch around the import in the `_return_module` method to avoid issues, but I thought it was better to raise errors.

kolibri/plugins/base.py
rtibbles: We could do - but it seemed reasonable to allow plugins to add to the core options.ini (but again, as with settings, not allow them to overwrite - the only exception being to let them change default values).

kolibri/plugins/registry.py
rtibbles: I should update this comment further, it should read:
> From a developer's perspective, plugins are wrappers around Django applications, listed in ``ACTIVE_PLUGINS`` on the kolibri config object. They are initialized before Django's app registry is initialized and then their relevant Django apps are added to the ``INSTALLED_APPS`` of kolibri.

kolibri/plugins/registry.py
rtibbles: `INSTALLED_APPS` is still the manifest of active Django applications - it just isn't the manifest of Kolibri plugins (although it will include all Kolibri plugins).

I have added warnings and other catches to warn and error when these assumptions are being violated. Can add some more test cases to extend these.

jredrejo (approved):
I have taken a look at these changes and I really appreciate them. If we want plugins to be able to add significative  behaviour methods to kolibri without touching kolibri core I think this is the way to go. 
@benjaoming I understand your concerns about the possible conflicts between plugins, but that can happen now too. Look at https://github.com/jredrejo/kolibri/blob/olip-auth/kolibri/plugins/openid_connect/kolibri_plugin.py#L14-L16 to see how I had to workaround these problems. You can see that using the actual plugin architecture we can trigger the same problems you mention. At least, with this PR changes we can have a deterministic idea on the order the plugins are going to be loaded.

@rtibbles nothing to comment on the code, looks very good. However, I would really like to see some documentation to help in the process of creating a plugin. Current documentation is poor (so poor it's useless to build a plugin). Providing a scaffolding plugin with comments on the actions each line of code does could be enough if writting a whole guide can take too long.

jredrejo:
@rtibbles I have found a problem that we have with the current arquitecture and these changes don't solve.
This Django code https://github.com/django/django/blob/master/django/apps/registry.py#L61 is executed after this calls stack:
```
  /datos/le/mio/kolibri/venv/bin/kolibri(11)<module>()
-> load_entry_point('kolibri', 'console_scripts', 'kolibri')()
  /datos/le/mio/kolibri/kolibri/utils/cli.py(667)main()
-> initialize(debug=debug, skip_update=arguments["--skipupdate"])
  /datos/le/mio/kolibri/kolibri/utils/cli.py(193)initialize()
-> django.setup()
  /datos/le/mio/kolibri/venv/local/lib/python2.7/site-packages/django/__init__.py(27)setup()
-> apps.populate(settings.INSTALLED_APPS)
> /datos/le/mio/kolibri/venv/local/lib/python2.7/site-packages/django/apps/registry.py(67)populate(
```
When `django.setup()` is executed, the plugins settings have not been loaded yet, so the `INSTALLED_APPS` the plugin inserts are not included into the `django.apps.registry.Apps`.
This causes this error if any of these `INSTALLED_APPS` has database models:
`RuntimeError: Model class oidc_provider.models.ResponseType doesn't declare an explicit app_label and isn't in an application in INSTALLED_APPS.`

kolibri/plugins/utils/settings.py
```diff
@@ -0,0 +1,103 @@
...
+
+def _validate_module_setting(
+    setting, existing_settings, setting_value, plugin_settings, module_path
+):
+    # Raise an error if overwriting an existing setting that cannot be simply appended
+    # to.
+    if setting in existing_settings and setting not in _tuple_settings:
+        raise ValueError(
+            "Plugin settings should not be used to override Kolibri default settings, use the Django settings module env var"
+        )
```
benjaoming, line 34: :heart:

kolibri/plugins/utils/settings.py
```diff
@@ -0,0 +1,103 @@
...
+    if setting in _tuple_settings and not isinstance(setting_value, (tuple, list)):
+        raise ValueError("{setting} must be a tuple or a list".format(setting=setting))
+    # Warn if this setting has already been modified by another plugin
+    if setting in plugin_settings:
+        warnings.warn(
+            "Plugin {plugin_module} is modifying the {setting} setting, but this was already modified by: {plugins}".format(
+                plugin_module=module_path,
+                setting=setting,
+                plugins=", ".join(plugin_settings[setting]),
+            )
```
benjaoming, line 45: :heart:

benjaoming:
@jredrejo 

Thanks for providing a solid example of why we need this and what problems need to be fixed!

@rtibbles 

Thanks for adding more validation to what is allowed to be set in a shared environment by the plugins.

Suggesting the following:

Move `settings.py` to `autodiscover/django_settings.py` and `options.py` to `autodiscover/kolibri_options.py` -- This would be to further help the judgement of the plugin author and to avoid the confusion between "options" and "settings", where I also think it would be VERY tempting to add other non-related items to these auto-loaded modules simple for the sake of a clean namespace... otherwise plugins could end up with "options", "settings" and "config" :)

indirectlylit:
One of my least favorite things about Django is their strategy of giving files with certain names special behavior, and placing them alongside other normal files.

When looking at Django apps, it's hard to tell which file names were chosen arbitrarily, and which were chosen because they have a special meaning in the Django framework (`urls.py`, `api_urls.py`, `models.py`, `serializers.py`, `views.py`, etc). It would be nice to avoid this ambiguity if possible for plugins.

A possible alternative to auto-discovery would be explicit parameterization of `KolibriPluginBase`.

For example, a `kolibri_plugin.py` file might have:

```python
from . import api_urls
from . import urls
from . import settings
from . import options

class MyPlugin(KolibriPluginBase):
    django_api_urls = api_urls
    django_urls = urls
    django_settings = settings
    kolibri_options = options
```

Then the file naming is decoupled, and the API is explicit

indirectlylit:
(sorry, accidentally hit the wrong button - continuing edit above)

benjaoming:
@indirectlylit I thought about exactly that, but because of the tricky-ness of load-order, I wasn't really sure if it could be achieved. I think one of the fundamental issues is to have these 3 modules separated because they're wanted at 3 different stages in loading Kolibri?

indirectlylit:
I see.

It could also take easily take the module names as strings, and then let the plugin framework import them at the correct time.

```python
class MyPlugin(KolibriPluginBase):
    django_api_urls = "api_urls"
    django_urls = "urls"
    django_settings = "settings"
    kolibri_options = "options"
```

indirectlylit:
That strategy does feel a *little* less valuable, because there's still "weird dynamic importing" happening. However it does help with the API a little bit...

rtibbles:
Note that the 'weird dynamic importing' is deliberate and necessary, because it prevents accidental initialization of Django during plugin initialization.

benjaoming:
I can live with importing string-defined modules. The "well-defined" part seems valuable. I.e. that we can encapsulate these structures in simple templates and documentation. And at the same time ensure that developers aren't "easily tempted" to break the structures because they're complicated.

rtibbles:
Yeah - I like @indirectlylit's suggestion too. We will need to change a few plugins when we merge this though :)

benjaoming:
There are still no plugins "in the wild", right? So everything released is bundled with Kolibri?

rtibbles:
Can you guys stop closing my PR?!

Currently there are mostly only 'internal' external plugins. There is one that is being maintained by a partner, but I don't think these changes should affect this.

benjaoming:
> Can you guys stop closing my PR?!

Haha, weird and apologies, we're really not trying to imply anything. I use `CTRL+enter` to submit comments.

rtibbles:
Have got rid of any magic imports for plugins, and moved the settings enhancement into our base settings configuration.

The only outstanding things here are:
* Whether we want to do something different for `INSTALLED_APPS` and `TEMPLATE_LOADERS` than the current scheme (and remove the `LOCALE_PATHS` modification because the plugin API already defines that).
* Whether we want any mechanism for enhancing `MIDDLEWARE` - I think I agree with @benjaoming that we generally don't want to do this, but I know that @jredrejo might need this for setting up an open ID provider.

rtibbles:
The only other open question I had here was related to which settings we allow to be overridden.

At the moment, it uses the Kolibri base settings as the source of truth for this - it doesn't allow overwriting any settings that are defined in there through plugins.

Another alternative would be to use the Django global settings as the source of truth - this would mean any Kolibri specific settings would be overwritable, while all Django global settings would not be (we could also use the union of the two).

I am not hugely opinionated on either approach - the only slight disadvantage of the current approach I can see is that a plugin could change a Django global setting with radical consequences (when it might be better done using a custom settings file).

kolibri/plugins/base.py
```diff
@@ -37,6 +37,29 @@ class KolibriPluginBase(object):
...
+    # URLs for views that contain
+    # translated content, and hence will be prefixed
+    # with a language prefixs
+    translated_view_urls = None
+
+    # Name of a local module that contains additional settings to augment
+    # Django settings.
+    # For settings that take a tuple or list, these will be appended to the value from
+    # the base settings module set through conventional Django means.
+    django_settings = None
```
benjaoming, line 56: You can use the `#: Comment` syntax to automatically have the comment become a documentation string for the property :) It's not a part of default Sphinx, but we have added it to make inline documentation easier.

benjaoming: Does it still work?

rtibbles: Oh neat! And yes, it does.

indirectlylit:
merging this into a new UN Women feature branch so we can use it for ongoing work

benjaoming:
> merging this into a new UN Women feature branch so we can use it for ongoing work

Thinking if we can have the PR work open in a separate branch, still? Is the `un-women` branch something that is intended for merging into a release branch, or just an accumulation of a few currently unmerged working branches that will eventually be merged?

benjaoming:
Said UN Woment branch: https://github.com/learningequality/kolibri/pull/5722

PR #5696, opened 2019-06-21
Update VERSION variable to 0.12.6

jonboiser:


lyw07 (approved):
will merge after build finishes

PR #5718, opened 2019-06-27
Strings Refactor

nucleogenesis:
### Summary

This PR is the follow-up to https://github.com/learningequality/kolibri/pull/5526 and continues to address this issue https://github.com/learningequality/kolibri/issues/5057.

**In Summary**
- Extract strings with 3+ contextually identical uses into a "common" strings file.
- When used 3+ times across multiple modules (eg, Learn, Core, Coach) then it is extracted to a "Core Common" file.
- Each "Common" file exports a mixin, which provides a module-named method (eg, `coreCommon$tr()` or `coachCommon$tr()`)
- The "Core Common" file is part of the coreVue api spec and can be imported globally.
- Module-specific "Common" files are available directly in their own modules and are only imported within those modules.
- There are some exceptions wherein certain strings were not extracted. These cases are largely those in which the "key" (ie, `requiredLabel`) that would be passed to a `*$tr()` method is determined dynamically, depending on external values. A good example is in Coach's `common.js` where the route has a meta value that serves as a key in map of values which determines the "key" used in the local `$tr()` call to determine the AppBar's label. There are several more like this - and we've basically ignored them.

**Status of Work in Progress**
- There are some more strings to work through that were given by the profiler.
- I believe it would be good to review newly profiled data as the newly generated data could make the next pass substantially simpler.

### Reviewer guidance

- We will want to verify that no strings were extracted mistakenly that ought to be translated separately due to contextual meanings.
- Ensure that all strings are being rendered as expected - we shouldn't see anything like "CoachCommonStrings.coachesLabel" where we would expect to see "Coaches".
- Review the approach, naming conventions and other aspects of the change to be sure that things are semantically consistent and will make the continued use of this system as easy as possible

### References

[The Google Sheet "Second Pass" indicating the strings, definitions, keys, and my last pass of changes based on color-coding](https://docs.google.com/spreadsheets/d/1_qs5a8g0Rz-dWPiJ5vL9mQoDfhZIwN64eNsohn14KX8/edit#gid=1414584889)

https://github.com/learningequality/kolibri/issues/5057

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated


kolibri/core/assets/src/mixins/commonCoreStrings.js
```diff
@@ -0,0 +1,60 @@
...
+  passwordLabel: 'Password',
+  progressLabel: 'Progress',
+  questionNumberLabel: 'Question { questionNumber, number }',
+  resourcesLabel: 'Resources',
+  searchLabel: 'Search',
+  usageAndPrivacyLabel: 'Usage and privacy',
+  usernameLabel: 'Username',
+  usersLabel: 'Users',
+  userTypeLabel: 'User type',
+  viewMoreLabel: 'View more',
```
indirectlylit: this one's an action

indirectlylit (commented):
this is exciting to see!

kolibri/plugins/learn/assets/src/views/MasteredSnackbars/index.vue
```diff
@@ -32,7 +32,7 @@
           type="warning"
           class="alert"
         >
-          {{ $tr('signIn') }}
+          {{ coreCommon$tr('signInLabel') }}
```
indirectlylit: This change looks problematic, and we should be careful if it's indicative of other similar changes.

Here, the string `signIn` defined in this component was actually "Sign in or create an account to save points you earn", not "Sign in".

It's also still defined below. If this was a case where we were consolidating, it should have been deleted.

Finally, "Sign in" at least as used on the main page is an action. (Looks like the title of the page is "User Sign In" which would be a label... and also I think is grammatically suspect...)

indirectlylit (dismissed):
I found a potential issue - might want to check if other similar ones occurred.

kolibri/plugins/learn/assets/src/views/MasteredSnackbars/index.vue
nucleogenesis: I actually grep-replaced this by mistake. I changed another component's `$tr('signIn')` to the coreCommon on one or two other components and didn't realize that this was in the results.

Pushed a fix

indirectlylit: cool! should we delete `signInLabel: 'Sign in',`?

nucleogenesis: @indirectlylit `signInLabel` in Core is used elsewhere

kolibri/core/assets/src/mixins/commonCoreStrings.js
```diff
@@ -0,0 +1,72 @@
...
+
+  // toggles
+
+  // formatted values
+
+  // Errors
+});
+
+export default {
+  methods: {
```
jonboiser: Could you add a console.error or something like that to catch when a 'key' is not available. I think this can be generalized to catch invalid 'args' too, which would be nice.

kolibri/plugins/coach/assets/src/views/common/QuizActive.vue
```diff
@@ -37,8 +37,8 @@
     computed: {
       label() {
         return this.active
-          ? this.coachCommon$tr('quizActiveLabel')
-          : this.coachCommon$tr('quizInactiveLabel');
```
jonboiser, line 41: This might run into the noun-adjective agreement issue (e.g. "active [quiz]" and "active [lesson"] would be translated differently in Spanish.), so I would say to keep 'quizActiveLabel' and 'lessonActiveLabel' separate in coach strings.

kolibri/plugins/coach/assets/src/views/plan/assignments/AssignmentDetailsModal.vue
```diff
@@ -217,14 +218,14 @@
...
           return {
-            activeStatus: this.coachCommon$tr('quizActiveLabel'),
-            inactiveStatus: this.coachCommon$tr('quizInactiveLabel'),
+            activeStatus: this.coachString('activeLabel'),
+            inactiveStatus: this.coachString('inactiveLabel'),
             statusExplanation: this.$tr('activeQuizzesExplanation'),
           };
         }
         return {
-          activeStatus: this.coachCommon$tr('lessonActiveLabel'),
```
jonboiser, line 226: Revert this to preserve adjective-noun agreement I mentioned above. Same with quizzes.

kolibri/core/assets/src/mixins/commonCoreStrings.js
```diff
@@ -0,0 +1,72 @@
+import { createTranslator } from 'kolibri.utils.i18n';
+
+export const coreStrings = createTranslator('CommonCoreStrings', {
+  // actions
+  cancelAction: 'Cancel',
```
jonboiser, line 5: I don't know if this has been an issue, but there is always the possibility that the verb will need to agree with the elided direct object. e.g. "Delete [quiz]" may be translated differently from "Delete [user]" (bracket means it's left out/elided/implied.

kolibri/plugins/learn/assets/src/views/MasteredSnackbars/index.vue
```diff
@@ -103,7 +104,7 @@
       Snackbar,
       UiAlert,
     },
-    mixins: [themeMixin],
+    mixins: [themeMixin, commonCoreStrings],
```
jonboiser: I don't see `this.coreString` used in this component.

kolibri/core/assets/src/mixins/commonCoreStrings.js
indirectlylit: That's a great point. We haven't heard that feedback from from translators, but we honestly don't hear much from them. We have had the analogous case for adjectives, e.g. "Active [lesson]" vs "Active [quiz]".

I think the best strategy would be to name the actions with the elided concept, where appropriate. So for example, we would have independent strings for `deleteUserAction` and `deleteLessonAction`. This will have the effect of lowering the amount of re-use, but will ensure that when strings are reused they are more likely to be correct.

I don't think we'll be able to do this for strings where the elided word is not obvious. For example, `cancelAction`: cancel operation, cancel modal, cancel dialog, cancel activity...? There's no clearly correct object that end users will expect agreement with.

kolibri/plugins/coach/assets/src/views/common/QuizActive.vue
nucleogenesis: @jonboiser Do you know offhand if/where there are other contexts in which the word "Active" would refer to an elided "quiz"? 

If this is the only component, then I will define the string on that component directly with "activeQuizLabel" and "inactiveQuizLabel" rather than using the common "(in)activeLabel" strings.

kolibri/core/assets/src/mixins/commonCoreStrings.js
nucleogenesis: I wonder how many places this is an issue in the codebase currently. 

If there are fewer than 3 cases of needing "Active [quiz]" on fewer than 3 separate components, then we would just define them directly on the component with a properly named action, like @indirectlylit mentioned ie, `quizActiveLabel`.

kolibri/plugins/coach/assets/src/views/common/QuizActive.vue
indirectlylit: > I will define the string on that component directly 

that seems like a safe bet for cases like this one in general

jonboiser (approved):

indirectlylit:
bang!

PR #5730, opened 2019-07-01
Update EPUB renderer styles, fix update notification link

### LINKED ISSUE #5564: non-standard styling in epub renderer
### Observed behavior

**1 - icons are black with purple circles:**

Should be completely black or completely purple

image

**2 - Non-standard elevations and drop shadow usage.** [original reference](https://github.com/learningequality/kolibri/pull/5579#issuecomment-497120316)

Our drop shadows should be kept consistent with elevation. We have a set of [standard drop shadows](https://github.com/learningequality/kolibri/blob/develop/kolibri/core/assets/src/styles/materialDesignElevations.scss) that we're not using here.

It's also not clear to me why the header and footer elements have drop shadows or elevations at all.  This is not consistent with Material [elevation conventions](https://material.io/design/environment/elevation.html) or our other renderers. (cc @khangmach @jtamiace)

| current | flattened |
|--|--|
| image | image  |

**3 - No rounded corners**

In general we avoid hard corners. Standard practice is to use `border-radius: $radius;` which is `4px`.

> image

### Expected behavior

Consistent UI with the rest of the app

### User-facing consequences

small UI inconsistencies

### Steps to reproduce

view an epub

### Context

0.12.3

### LINKED ISSUE #5728: upgrade notification link should open in a new tab
### Observed behavior

Currently, upgrade notification links open in the same tab

### Expected behavior

they should open in a new tab

### User-facing consequences

non-typical behavior when opening a link pointing outside of our app

### Steps to reproduce

Follow the upgrade notification gherkin story, and click the link at the end:
https://github.com/learningequality/kolibri/blob/develop/integration_testing/features/super-admin/super-admin-upgrade-notification.feature

see also https://github.com/learningequality/nutritionfacts/issues/8

### Context

0.12.5

jonboiser:
### Summary
1. Applies light button style when in "YELLOW" (yellow text on black background) theme.
1. Applies a 1dp box shadow to the container around the EPUB, removes shadows from the top and bottom control bars to flatten them.
1. Removes box shadow from left and right side bars so shadows do not spread into the top and bottom bars. I don't follow the suggestion in #5564 since adding rounded corners means that dark backgrounds will peek through. This just makes all the internal panels of the render flattened.
1. Remove the animations that accompany toggling side bars. Now they just (dis)appear instantly.
1. Some refactoring.
1. Make update notification link open in a new window by setting target properly. Fixes #5728

Before
Screen Shot 2019-07-01 at 1 55 17 PM

After

Screen Shot 2019-07-01 at 1 54 51 PM

### Reviewer guidance

1. View an EPUB. Test all of the controls (settings, navigation, etc.)

### References

Fixes #5564 
…

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/plugins/document_epub_render/assets/src/views/EpubRendererIndex.vue
+++ b/kolibri/plugins/document_epub_render/assets/src/views/EpubRendererIndex.vue
@@ -177,7 +177,7 @@
   import SettingsButton from './SettingsButton';
   import SearchButton from './SearchButton';
 
-  import { THEMES } from './EpubConstants';
+  import { THEMES, darkThemeNames } from './EpubConstants';
 
   const FONT_SIZE_MIN = 8;
   const FONT_SIZE_MAX = 32;
@@ -309,7 +309,7 @@
         };
       },
       navigationButtonColor() {
-        return [THEMES.BLACK, THEMES.GREY].some(theme => isEqual(this.theme, theme))
+        return darkThemeNames.some(themeName => isEqual(this.theme.name, themeName))
           ? 'white'
           : 'black';
       },
--- a/kolibri/plugins/document_epub_render/assets/src/views/EpubRendererIndex.vue
+++ b/kolibri/plugins/document_epub_render/assets/src/views/EpubRendererIndex.vue
@@ -21,7 +21,7 @@
 
       <TopBar
         ref="topBar"
-        class="top-bar"
+        class="top-bar-component"
         :isInFullscreen="isInFullscreen"
         @tableOfContentsButtonClicked="handleTocToggle"
         @settingsButtonClicked="handleSettingToggle"
@@ -711,6 +711,7 @@
 
 <style lang="scss" scoped>
 
+  @import '~kolibri.styles.definitions';
   @import './EpubStyles';
 
   $top-bar-height: 36px;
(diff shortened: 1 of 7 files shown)


indirectlylit:
Would you mind including one more change in this PR, and hiding the "previous" button on the first page, and the "next" button on the last page?

I was recently experimenting with RTL books and found it super confusing that both buttons were always visible.

kolibri/plugins/document_epub_render/assets/src/views/EpubRendererIndex.vue
```diff
@@ -734,10 +714,14 @@
   $navigation-button-normal: 52px;
 
   .epub-renderer {
+    @extend %dropshadow-1dp;
```
indirectlylit: rather than a 1dp drop shadow, let's make this a 1px solid line:

image

jonboiser: which color token?

jonboiser: used the grey-300 color

indirectlylit:
looks good. However #5564 and #5728 target 0.12.6, but the PR targets `develop`

indirectlylit (approved):

PR #5731, opened 2019-07-02
work on integrating OIDC provider workflow as a built-in plugin

indirectlylit:
### Summary

Builds off of @jredrejo's work in https://github.com/learningequality/kolibri-un-women-online-plugin/pull/7/

### Reviewer guidance

* Attempts to use the standard sign in page for OIDC sign-in
* Can be activated by running `kolibri plugin kolibri.plugins.oidc_provider_plugin enable`
* Mounts `openid-configuration` at the domain root

### References

…

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/core/assets/src/state/modules/core/actions.js
+++ b/kolibri/core/assets/src/state/modules/core/actions.js
@@ -243,8 +243,14 @@ export function kolibriLogin(store, sessionPayload) {
   Lockr.set(UPDATE_MODAL_DISMISSED, false);
   return SessionResource.saveModel({ data: sessionPayload })
     .then(() => {
-      // Redirect on login
-      redirectBrowser();
+      // OIDC redirect
+      if (sessionPayload.next) {
+        redirectBrowser(sessionPayload.next);
+      }
+      // Normal redirect on login
+      else {
+        redirectBrowser();
+      }
     })
     .catch(error => {
       store.commit('CORE_SET_SIGN_IN_BUSY', false);
--- a/kolibri/deployment/default/settings/dev.py
+++ b/kolibri/deployment/default/settings/dev.py
@@ -4,7 +4,7 @@ from __future__ import unicode_literals
 
 from .base import *  # noqa isort:skip @UnusedWildImport
 
-INSTALLED_APPS += ["rest_framework_swagger"]  # noqa
+INSTALLED_APPS = list(INSTALLED_APPS) + ["rest_framework_swagger"]  # noqa F405
 
 INTERNAL_IPS = ["127.0.0.1"]
 
--- a/kolibri/plugins/base.py
+++ b/kolibri/plugins/base.py
@@ -49,6 +49,11 @@ class KolibriPluginBase(object):
     # with a language prefixs
     translated_view_urls = None
 
+    # Name of a local module that contains url_patterns that define
+    # URLs for views that should be attached to the domain root.
+    # Use with caution! The lack of namespacing is dangerous.
+    root_view_urls = None
+
     # Name of a local module that contains additional settings to augment
     # Django settings.
     # For settings that take a tuple or list, these will be appended to the value from
@@ -177,6 +182,31 @@ class KolibriPluginBase(object):
                 logging.warn(
                     "{plugin} defined {urls} untranslated view urls but the module was not found".format(
                         plugin=self._module_path(), urls=self.untranslated_view_urls
+                    )
+                )
+            return module
+
+    def root_url_module(self):
+        """
+        Return a url module, containing ``urlpatterns = [...]``, a conventional
+        Django application url module.
+
+        Do this separately for endpoints that need to be attached at the root.
+
+        URLs are by default accessed through Django's reverse lookups like
+        this::
+
+            reverse('kolibri:url_name')
+
+        By default this will be discovered based on the root_view_urls
+        property.
+        """
+        if self.root_view_urls:
+            module = self._return_module(self.root_view_urls)
+            if module is None:
+                logging.warn(
+                    "{plugin} defined {urls} root view urls but the module was not found".format(
+                        plugin=self._module_path(), urls=self.root_view_urls
                     )
                 )
             return module
--- a/kolibri/plugins/oidc_provider_plugin/api_urls.py
+++ b/kolibri/plugins/oidc_provider_plugin/api_urls.py
@@ -1,0 +1,17 @@
+from django.conf.urls import include
+from django.conf.urls import url
+from rest_framework import routers
+
+from .views import ProviderInfoView
+
+router = routers.SimpleRouter()
+
+urlpatterns = [
+    url(
+        r"^\.well-known/openid-configuration/?$",
+        ProviderInfoView.as_view(),
+        name="provider-info",
+    ),
+    url(r"^", include("oidc_provider.urls", namespace="oidc_provider")),
+]
+urlpatterns += router.urls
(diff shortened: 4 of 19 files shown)


kolibri/core/assets/src/state/modules/core/actions.js
```diff
@@ -243,8 +243,14 @@ export function kolibriLogin(store, sessionPayload) {
   Lockr.set(UPDATE_MODAL_DISMISSED, false);
   return SessionResource.saveModel({ data: sessionPayload })
     .then(() => {
-      // Redirect on login
-      redirectBrowser();
+      // OIDC redirect
+      if (sessionPayload.next) {
+        redirectBrowser(sessionPayload.next);
```
indirectlylit, line 248: This is where redirect can occur. Unsure if this is actually what's supposed to happen next in the flow?

kolibri/deployment/default/settings/dev.py
```diff
@@ -4,7 +4,7 @@
 
 from .base import *  # noqa isort:skip @UnusedWildImport
 
-INSTALLED_APPS += ["rest_framework_swagger"]  # noqa
+INSTALLED_APPS = list(INSTALLED_APPS) + ["rest_framework_swagger"]  # noqa F405
```
indirectlylit, line 7: `INSTALLED_APPS` was a tuple

kolibri/plugins/oidc_provider_plugin/settings.py
```diff
@@ -0,0 +1,2 @@
+INSTALLED_APPS = ["oidc_provider"]
+OIDC_LOGIN_URL = "/user/#/signin"
```
indirectlylit: Note, switched from `LOGIN_URL` to `OIDC_LOGIN_URL`

Will this work as intended though? I'm unsure of how things will be affected by our language prefixes, and also potentially by the client-side routing

kolibri/plugins/user/assets/src/views/SignInPage/index.vue
```diff
@@ -405,11 +410,15 @@
...
         if (this.formIsValid) {
-          this.kolibriLogin({
+          const sessionPayload = {
             username: this.username,
             password: this.password,
             facility: this.facilityId,
-          }).catch();
+          };
+          if (global.oidcProviderEnabled) {
+            sessionPayload['next'] = this.$route.query.next;
```
indirectlylit: grab the `next` query parameter and add it to the payload

indirectlylit (commented):
@jredrejo @rtibbles left some comments inline

kolibri/plugins/oidc_provider_plugin/settings.py
jredrejo: nope, it has to be `LOGIN_URL` . The provider library requires this name. We can do some monkey patching if you want to add a prefix.

kolibri/deployment/default/settings/dev.py
rtibbles: Ah, right, the plugin settings modification coerces it to a tuple.

kolibri/plugins/base.py
```diff
@@ -181,6 +186,31 @@ def api_url_module(self):
...
+        Django application url module.
+
+        Do this separately for endpoints that need to be attached at the root.
+
+        URLs are by default accessed through Django's reverse lookups like
+        this::
+
+            reverse('kolibri:url_name')
+
+        By default this will be discovered based on the root_view_urls
```
rtibbles, line 201: I would like a very big warning here, just to scare people.

kolibri/plugins/oidc_provider_plugin/api_urls.py
```diff
@@ -0,0 +1,8 @@
+from django.conf.urls import include
+from django.conf.urls import url
+from rest_framework import routers
+
+router = routers.SimpleRouter()
```
rtibbles, line 5: Looks like this is doing nothing, as we are not registering any endpoints against the router.

kolibri/plugins/oidc_provider_plugin/root_urls.py
```diff
@@ -0,0 +1,15 @@
+from django.conf.urls import url
+from rest_framework import routers
+
+from .views import ProviderInfoView
+
+router = routers.SimpleRouter()
```
rtibbles, line 6: Ditto.

kolibri/plugins/user/assets/src/views/SignInPage/index.vue
```diff
@@ -166,6 +168,14 @@
   import LanguageSwitcherFooter from '../LanguageSwitcherFooter';
   import FacilityModal from './FacilityModal';
 
+  // https://davidwalsh.name/query-string-javascript
+  function getUrlParameter(name) {
+    name = name.replace(/[[]/, '[').replace(/[\]]/, '\\]');
+    var regex = new RegExp('[\\?&]' + name + '=([^&#]*)');
```
rtibbles, line 174: Thought we had this somewhere already, but I think it was in Hashi and then I later removed it. Can also polyfill the URL constructor too for this.

kolibri/plugins/user/assets/src/views/SignInPage/index.vue
```diff
@@ -292,6 +302,17 @@
         }
         return { backgroundColor: this.$themeColors.brand.primary.v_900 };
       },
+      oidcProviderFlow() {
+        return global.oidcProviderEnabled && this.nextParam;
+      },
+      nextParam() {
```
rtibbles, line 308: Django ordinarily has support for this, but I had removed it in our implementation - I actually kind of prefer handling it in the frontend, but we could reinstate the backend functionality if that was preferred.

kolibri/plugins/utils/urls.py
```diff
@@ -10,13 +10,16 @@ def get_urls():
...
+        root_url_module = plugin_instance.root_url_module()
         instance_patterns = []
         # Normalize slug
         slug = plugin_instance.url_slug().lstrip("^").rstrip("/") + "/"
         if url_module:
             instance_patterns += i18n_patterns(url_module.urlpatterns, prefix=slug)
         if api_url_module:
             instance_patterns.append(url(slug + "api/", include(api_url_module)))
+        if root_url_module:
+            instance_patterns.append(url("", include(root_url_module)))
```
rtibbles, line 22: What I had suggested was to not mount these within the plugin namespace (so they are doubly 'root' - both in terms of their point of mounting, and the Django namespace in which they are mounted).

Would want to extract this specific part into a separate 'root_urls' function that gets separately called and mounted in the `deployment.default.urls` configuration. This should allow us to remove a lot of the monkey patching above.

kolibri/utils/conf.py
```diff
@@ -77,6 +77,7 @@
         "kolibri.plugins.style_guide",
         "kolibri.plugins.document_epub_render",
         "kolibri.plugins.default_theme",
+        "kolibri.plugins.oidc_provider_plugin",
```
rtibbles, line 80: Do we want this to be the default during dev?

indirectlylit: no, this was supposed to be  removed

kolibri/plugins/base.py
indirectlylit: added warning

kolibri/plugins/user/assets/src/views/SignInPage/index.vue
indirectlylit: looks like [`URLSearchParams`](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams#Browser_compatibility) is the new hotness

rtibbles: Yeah, when you parse a url with the URL constructor, the it generates a URLSearchParams object for the query params.

Issue #5470, opened 2019-05-09
remove the UserTable component


indirectlylit:
### Observed behavior

Our `UserTable` component is serving too many purposes, united only by the fact that they happen to be tables whose rows correspond to users. It's growing into a difficult-to-maintain beast and doesn't do very much to enforce consistency over just using a bare `CoreTable`

### Expected behavior

We should remove the shared `UserTable` component and replace it with non-parameterized `CoreTable`s

### User-facing consequences

None - just code maintainability

N/A

### Context

0.12.3

jonboiser:
Closing, since I believe that `UserTable` covers all of our current user account-displaying needs, and actually duplicating the code for the various workflows its used in would actually increase the maintenance burden. If a new presentational component is needed for users, though, I would recommend just adding a new one instead of adding more props/features to `UserTable`

PR #5751, opened 2019-07-08
Add pagination to manage permissions page, make unified pagination component

### LINKED ISSUE #5741: Add Pagination to Device > Manage Permissions Page
### Observed behavior

The Device > Managed Permissions Page lists every Facility User in one big list. So similar to the Facility > Users page, applying the filter can be slow when there are thousands of users.

### Expected behavior

Following #5647, add a similar pagination UI, so applying the filter isn't slowed down by re-rendering the large list.

### Context

Kolibri 0.13

### LINKED ISSUE #5783: user string not localized - 'x of y' (pages) on Facility > Users
### Observed behavior
`of` in 'x of y' (pages) on **Facility > Users** does not seem properly localized.

Selection_154

Selection_155

### Expected behavior
Fully localized phrase.

### User-facing consequences
Difficulty understanding.

### Errors and logs
```
[Vue Intl] Missing message: "UserPage.pagination" for locale: "es-419", using default message as fallback.
[Vue Intl] Missing message: "UserPage.previousResults" for locale: "es-419", using default message as fallback. vue-intl.dev.js:3158
[Vue Intl] Missing message: "UserPage.nextResults" for locale: "es-419", using default message as fallback.
```

### Steps to reproduce
Switch to language different than English on **Facility > Users**.

### Context
 * Kolibri version: `release-v0.12.x` dev branch
 * Operating system: all
 * Browsers: all

cc @nucleogenesis 
@indirectlylit to correct the milestone (if necessary)

jonboiser:
### Summary

1. Adds PaginatedListContainer, a component that helps create paginated and filtered displays of stuff. Replace existing FacilityUser tables with it
1. Adds pagination to Manage Permissions Page with PaginatedListContainer
1. With unified list component, the message `N - M of Total`should be localized everywhere this paginated list is used Fixes #5783 
1. Increases the default page size from 10 rows to 30

### Reviewer guidance

…

### References

Fixes #5741 

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated


kolibri/core/assets/src/views/PaginatedListContainer.vue
```diff
@@ -0,0 +1,161 @@
+<template>
+
+  <div>
+    <KGrid>
+      <KGridItem sizes="3, 3, 3">
+        <slot name="otherFilter"></slot>
+      </KGridItem>
+      <KGridItem sizes="9, 9, 9" align="right" class="text-filter">
```
indirectlylit: Remember that...

| breakpoint | number of columns |
|--|--|
| small | 4 |
| medium | 8 |
| large | 12 |

so I don't think the sizes as given make sense?

kolibri/core/assets/src/views/PaginatedListContainer.vue
```diff
@@ -0,0 +1,161 @@
...
+        <KFilterTextbox
+          v-model.trim="filterInput"
```
indirectlylit: in the past, using 'trim' has caused us issues. Is the use here intentional?

kolibri/core/assets/src/views/PaginatedListContainer.vue
```diff
@@ -0,0 +1,161 @@
...
+      otherFilters: {
+        type: Object,
+        required: false,
+      },
+    },
+    data() {
+      return {
+        filterInput: '',
+        pageNum: 1,
+        perPage: 10,
```
indirectlylit: My sense is that 10 is way too low .. maybe we should bump this up to 30 or so?

kolibri/plugins/facility_management/assets/src/views/ClassEnrollForm.vue
```diff
@@ -1,50 +1,23 @@
...
-        <KIcon icon="forward" style="position: relative; top: -1px;" />
-      </UiIconButton>
-    </nav>
+    <PaginatedListContainer
+      :items="usersNotInClass"
+      :filterFunction="filterUsers"
+      :filterPlaceholder="$tr('searchForUser')"
+    >
+      <template v-slot:default="{items, filterInput}">
+        <UserTable
```
indirectlylit, line 11: possibly a good time to look at #5470 ?

indirectlylit (commented):
cool!

some quick notes...

kolibri/core/assets/src/views/PaginatedListContainer.vue
jonboiser: this was the existing implementation. i'll just remove it

kolibri/core/assets/src/views/PaginatedListContainer.vue
jonboiser: i don't think it's going to look nice on small or medium, but i'll just split it at 2/2, 3/5, 3/9

kolibri/core/assets/src/views/PaginatedListContainer.vue
jonboiser: sounds good to me

kolibri/core/assets/src/views/PaginatedListContainer.vue
indirectlylit: sometimes for small windows I intentionally split across multiple lines, e.g. 4/4, 8/8, 3/9

indirectlylit: that's also when I'll change alignments, so the right column might have `alignments="left, left, right"`

jonboiser: I set the props to your suggestions

rtibbles:
Anything else blocking here @indirectlylit? I think the `UserTable` issue should probably be resolved separately.

kolibri/plugins/facility_management/assets/src/views/UserPage/index.vue
```diff
@@ -166,59 +140,31 @@
...
       ...mapActions('userManagement', ['displayModal']),
+      emptyMessageForItems(items, filterText) {
+        if (this.facilityUsers.length === 0) {
+          return this.$tr('noUsersExist');
+        } else if (items.length === 0) {
+          return this.$tr('allUsersFilteredOut', { filterText });
+        }
+        return '';
+      },
+      filterUsers(users, filterText, { roleFilter }) {
```
rtibbles: Why does this need to be passed into the function, rather than referenced as `this.roleFilter` within the function?

I feel like this would simplify the API, and would mean that you only need to reset the page number when the `filteredItems` changes in the `PaginatedListComponent`.

rtibbles (commented):
Behaves as I would expect - one question on implementation.

kolibri/plugins/facility_management/assets/src/views/UserPage/index.vue
jonboiser: I'm not sure anymore. My guess is that I needed pass `otherFilters` into the PaginatedListContainer to make its `filteredItems` computed property reactive to changes in `otherFilters` as well as the default textbox input.

Maybe we could avoid having to pass `otherFilters` as a reactive prop by making `filterUsers` itself the reactive prop (make it a computed property that returns a function instead of a static method like it is here).

So if the role value changes, then `filterUsers` would change, and then `filteredItems` would have to update. I think I just didn't think to do that since I haven't really used functions as observable things.

jonboiser: that approach actually does work, but uses a little advanced JS (having to `.bind(this)` the reactive function).

rtibbles: Hrm, I guess my hope was that because `filterUsers` is a method of the component, its `this` would remain bound to the component itself, and then using it inside the computed prop of the `PaginatedListComponent` would access the reactive properties on the `UserPage` component, and then that would trigger recomputation on change.

rtibbles (approved):

Issue #4775, opened 2019-01-23
move Sentry clients to an external plugin


indirectlylit:
### Observed behavior

In #4774 we're adding Sentry clients so we can do error tracking. However, most users don't need or want these since we're using it for internal error reporting on our own servers.

We could make this into an external plugin; however this isn't currently compatible with our `pex` format, which is how we deploy our demo servers.

### Expected behavior

These should be contained in an external plugin that we'll install as-needed. This can occur when either:

1. pex formats support Python packages that are installed to the system
2. we migrate our demo server environments away from pex files

### User-facing consequences

A little extra bloat in our distributions

### Errors and logs

lots of them, hopefully

### Steps to reproduce

See [instructions on using Sentry](https://kolibri-dev.readthedocs.io/en/develop/manual_testing.html#collecting-client-and-server-errors-using-sentry)

### Context

0.12

aronasorman:
When you do this, can you also add support for [sentry environments](https://docs.sentry.io/enriching-error-data/environments/?platform=python)?

indirectlylit:
yeah, good idea.

I was thinking about hooking into the `nutritionfacts` pingback data and sending all of that, including e.g. run mode, channel lists, etc.

indirectlylit:
this has caused additional issues on ubuntu 14.04 because sentry seems to include some python 3.7+ files:

> ![image](https://user-images.githubusercontent.com/2367265/54968211-1a076a80-4f37-11e9-90ad-b75ba8cbcd7c.png)

@jredrejo is manually removing the offending (and unused) files in https://github.com/learningequality/kolibri-installer-debian/pull/67 .. however solving this issue would be a cleaner long-term solution

PR #5773, opened 2019-07-16
Remove sentry-sdk dependency

benjaoming:
# Summary

Removes the Sentry integration.

### Reviewer guidance

We are not using the sentry-sdk plugin for end-users, and it causes issues because of bytecode precompilation on incompatible systems.

### References

#5652
#5646

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/docs/manual_testing.rst
+++ b/docs/manual_testing.rst
@@ -113,7 +113,7 @@ The Sentry SDK Python package is not distributed with Kolibri. In order to use S
 
     pip install sentry-sdk  # might need to run with sudo
 
-If you're running Kolibri using a pex file, you'll need to make sure that the pex inherits the Python path. To do this, ensure that an environment variable ``PEX_INHERIT_PATH`` is set to ``1``.
+If you're running Kolibri using a pex file, you'll need to make sure that the pex inherits a Python path with `sentry_sdk` available. To do this without inheriting the full system path, run the pex from an active virtual environment with `PEX_INHERIT_PATH=1 python kolibri.pex`.
 
 To set up error reporting, you'll need a `Sentry DSN <https://docs.sentry.io/error-reporting/quickstart>`__. These are available from your project settings at ``https://sentry.io/settings/[org_name]/[project_name]/keys/``
 
--- a/docs/manual_testing.rst
+++ b/docs/manual_testing.rst
@@ -107,6 +107,14 @@ Collecting client and server errors using Sentry
 
 `Sentry <https://docs.sentry.io/>`__ clients are available for both backend and frontend error reporting. This can be particularly useful to have running on beta and demo servers in order to catch errors "in the wild".
 
+The Sentry SDK Python package is not distributed with Kolibri. In order to use Sentry, you must install it to your system Python packages and make the SDK available to Kolibri. To install, run
+
+.. code-block:: bash
+
+    pip install sentry-sdk  # might need to run with sudo
+
+If you're running Kolibri using a pex file, you'll need to make sure that the pex inherits the Python path. To do this, ensure that an environment variable ``PEX_INHERIT_PATH`` is set to ``1``.
+
 To set up error reporting, you'll need a `Sentry DSN <https://docs.sentry.io/error-reporting/quickstart>`__. These are available from your project settings at ``https://sentry.io/settings/[org_name]/[project_name]/keys/``
 
 You can set these either in options.ini or as environment variables.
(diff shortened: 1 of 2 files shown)


indirectlylit (dismissed):
see https://github.com/learningequality/kolibri/issues/4775 for the complexities here

benjaoming:
> see #4775 for the complexities here

Oh okay, was asking in Slack about this to figure out if we could `pip install sentry-sdk` on demo servers.

@rtibbles I guess that #5777 will have the same issue, given that a bundled plugin would still require sentry as an external dependency.

@lyw07 do you know a hack that can make our Pex on the demo servers access an external environment? But without changing the Pex itself (I guess we DON'T want it to see the external env). If not, I suggest that we block both #5773 and #5777 until we figure out how to run demo servers without pex files. Shouldn't be too hard.

lyw07:
@benjaoming do you mean finding a way to let the pex be able to import a library outside the pex environment?
If that's the case, we can set the environment variable [PEX_INHERIT_PATH](https://pex.readthedocs.io/en/stable/api/index.html#pex.variables.Variables.PEX_INHERIT_PATH) to the path of the library in the external environment and run pex file with it.

benjaoming:
Perfect, thanks so much @lyw07 - that works

So this means that we can continue using Pex on the demo servers, provided that we launch them in a virtual environment with `sentry_sdk` and set `PEX_INHERIT_PATH=true`

```
➜  workon testpex 
➜  pip install sentry-sdk
(...)
➜  PEX_INHERIT_PATH=true python kolibri-buildkite-build-5773-11590-kolibri-0.13.0.dev0_git.20190716150511.pex manage shell
INFO     Invoking command 'kolibri manage shell'.
Python 3.6.8 (default, Jan 14 2019, 11:02:34) 
[GCC 8.0.1 20180414 (experimental) [trunk revision 259383]] on linux
Type "help", "copyright", "credits" or "license" for more information.
(InteractiveConsole)
>>> import sentry_sdk
>>> print(sentry_sdk.__file__)
/home/user/.virtualenvs/testpex/lib/python3.6/site-packages/sentry_sdk/__init__.py
>>> 
```

I'd say that should work fine, and that this PR is safe to merge. We can already prepare the setup on the demo servers, even though they are still running 0.12.

benjaoming:
As argued above, this is safe to merge @rtibbles @indirectlylit - it is also a prerequisite to see if #5652 works.

indirectlylit:
Sounds good. I'll merge this as soon as someone has actually tested that it works as expected.

@rtibbles would that mean that #5777 is not actually necessary?

benjaoming:
I think it downplays the urgency, but I think #5777 is great, too.

rtibbles:
Not necessary, but useful.

indirectlylit:
sounds good, let's do both

benjaoming:
I verified that it runs and clicked a bit around: Not going to do a full integration testing, but it seems good to go.

indirectlylit:
I'm assuming @aronasorman or @lyw07 need to update the deployment process for our servers before merging this?

My guess is that if we merge as-is, future deployments which have sentry enabled will be broken?

indirectlylit:
here are the current docs:

https://kolibri-dev.readthedocs.io/en/develop/manual_testing.html#collecting-client-and-server-errors-using-sentry

aronasorman:
@indirectlylit cloud kolibri at least, already installs sentry manually. @lyw07 knows best on what we do for the kolibri demo servers. Thanks for the heads up!

benjaoming:
The method is outlined further up..

indirectlylit:
@benjaoming I tried to begin updating the docs here: https://github.com/benjaoming/kolibri/pull/7

Please incorporate the changes and ensure that they are correct.

benjaoming:
:heavy_check_mark: thanks @indirectlylit - for some reason I can't push to your repo, so if you can accept the suggestion, I'll merge to this PR.

benjaoming:
@indirectlylit changes to docs are in! Thanks!

indirectlylit (approved):
cc @lyw07 @jonboiser please note the updated process for future beta server and translation server updates

lyw07:
thank you for the heads-up!

benjaoming:
You can already make these changes on 0.12 deployments, they should be harmless in a virtualenvironment with only `sentry_sdk` available...

PR #5770, opened 2019-07-15
Fix router null reference in coach pages

jonboiser:
### Summary

Fixes a breakage in the Coach plugin's top-level app.js, that might be caused by a race condition in which the `kolibri.lib.router` is null before the `beforeEach` hook is called on it.

See e31fc608c66a24ecd837c4da2122802084372312

### Reviewer guidance

1. Verify that coach page is not working on develop without this patch.
1. Verify that coach page works 100% of the time with this patch.

### Contributor Checklist

PR process:

- [x] PR has the correct target branch and milestone
- [x] PR has 'needs review' or 'work-in-progress' label
- [x] If PR is ready for review, a reviewer has been added. (Don't use 'Assignees')
- [ ] If this is an important user-facing change, PR or related issue has a 'changelog' label
- [ ] If this includes an internal dependency change, a link to the diff is provided

Testing:

- [ ] Contributor has fully tested the PR manually
- [ ] If there are any front-end changes, before/after screenshots are included
- [ ] Critical user journeys are covered by Gherkin stories
- [ ] Critical and brittle code paths are covered by unit tests

### Reviewer Checklist

- Automated test coverage is satisfactory
- PR is fully functional
- PR has been tested for [accessibility regressions](http://kolibri-dev.readthedocs.io/en/develop/manual_testing.html#accessibility-a11y-testing)
- External dependency files were updated if necessary (`yarn` and `pip`)
- Documentation is updated
- Contributor is in AUTHORS.md

PR #5761, opened 2019-07-11
Styleguide fixes and clean-up

indirectlylit:
### Summary

Continuation of #5319. Includes updates based on feedback from @jonboiser 

This PR contains layout and styling updates to the styleguide, but _no changes to the content_.

Specifically:

* Removes the live code editor. My sense is that it caused more issues than it was worth:
  * examples used a second vue instance, and caused the examples to differ from our actual code-base in terms of import statements and `kebab-case` vs `CamelCase`
  * interacted poorly with new dynamic styling
  * related issues: #2416,  #1685, #1686
* updated examples so that they now are real, accurate reflections of components used as used in Kolibri
* switched to a Prism-based code example viewer (https://prismjs.com/) which seemed lighter-weight and more reliable than codemirror, which is [currently being rewritten](https://codemirror.net/6/).
* made code examples a bit more responsive (still not fully though)
* updated side-nav to use material-spec styling, but _not_ necessarily look like Kolibri
* updated component API tables to use our core table styling
* simplified API headers
* general styling updates, including globally bumping line height by a bit
* improved layout
* new color scheme
* mechanism for auto-generating and handling page anchor links
* switches URL to `/design`

### Reviewer guidance

…

### References

…

----

### Contributor Checklist

PR process:

- [x] PR has the correct target branch and milestone
- [x] PR has 'needs review' or 'work-in-progress' label
- [x] If PR is ready for review, a reviewer has been added. (Don't use 'Assignees')
- [ ] If this is an important user-facing change, PR or related issue has a 'changelog' label
- [ ] If this includes an internal dependency change, a link to the diff is provided

Testing:

- [x] Contributor has fully tested the PR manually
- [ ] If there are any front-end changes, before/after screenshots are included
- [ ] Critical user journeys are covered by Gherkin stories
- [ ] Critical and brittle code paths are covered by unit tests

### Reviewer Checklist

- Automated test coverage is satisfactory
- PR is fully functional
- PR has been tested for [accessibility regressions](http://kolibri-dev.readthedocs.io/en/develop/manual_testing.html#accessibility-a11y-testing)
- External dependency files were updated if necessary (`yarn` and `pip`)
- Documentation is updated
- Contributor is in AUTHORS.md

PR #5774, opened 2019-07-16
updates to router and vue initialization

indirectlylit:
### Summary

follow-up from https://github.com/learningequality/kolibri/pull/5770 and e31fc60

As far as I can tell, the async promise behaviors were not necessary, and were causing issues with initialization.

### Reviewer guidance

please look and test

### References

originally stemmed from work in https://github.com/learningequality/kolibri/pull/5761

There, I wanted to be able to initialize the router with alternate options, and the "kolibri app" framework was making that difficult

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated


rtibbles:
The main motivation for waiting for the initial heartbeat poll is that until that has happened the session information is not loaded into vuex state, which can cause inconsistencies or errors in the UI.

If this is not the case, this should be fine.

rtibbles:
Oh, and until the heartbeat has finished the initial retrieve, any API writes to the server will result in a 403 - which could happen if the session endpoint is slow to respond and the app logic kicks off without waiting for it to complete.

So, on second thoughts, we shouldn't do this.

rtibbles:
Third thoughts - after conversation, it seems like Vue Router only starts serving routes after it has been passed into the root Vue component at vue app initialization. So, it seems like this will be non-harmful, because nothing will happen until the heartbeat promise has been fulfilled and all necessary state is in place!

luna215:
Is there a reason this pull hasn't been merged? This will definitely fix a bug that's in currently in the most recent develop branch.

rtibbles (approved):

PR #5777, opened 2019-07-16
Externalize sentry plugin

rtibbles:
### Summary
* Removes sentry integration
* Adds code to allow external plugin of sentry - currently WIP at: https://github.com/learningequality/kolibri-sentry-plugin
* Optimistically updates docs to point at the above repo (installation instructions to follow in the readme there).

### Reviewer guidance
Main thing to note is the new 'context' API for WebpackBundleHooks to allow frontend assets to load Django context into `window[<__moduleName>]` where `__moduleName` is the `bundle_id` property defined on the instance and used to generate the name of the module in all built JS assets.

### References
https://github.com/learningequality/kolibri-sentry-plugin

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/core/webpack/hooks.py
+++ b/kolibri/core/webpack/hooks.py
@@ -77,6 +77,11 @@ class WebpackBundleHook(hooks.KolibriHook):
     # : for this be inlined?
     inline = False
 
+    # : A mapping of key to JSON serializable value.
+    # : This context will be bootstrapped into the window object
+    # : with a key of the unique_slug as a Javascript object
+    context = {}
+
     def __init__(self, *args, **kwargs):
         super(WebpackBundleHook, self).__init__(*args, **kwargs)
 
@@ -272,6 +277,17 @@ class WebpackBundleHook(hooks.KolibriHook):
         else:
             return []
 
+    def context_tag(self):
+        if self.context:
+            return [
+                '<script>window.{bundle} = JSON.parse("{context}");</script>'.format(
+                    bundle=self.unique_slug,
+                    context=json.dumps(self.context),
+                )
+            ]
+        else:
+            return []
+
     def get_basename(self, url):
         """
         Takes full path to a static file (eg. "/static/css/style.css") and
@@ -344,7 +360,7 @@ class WebpackBundleHook(hooks.KolibriHook):
         :param bundle_data: The data returned from
         :return: HTML of script tags for insertion into a page.
         """
-        tags = self.frontend_message_tag() + list(self.js_and_css_tags())
+        tags = self.context_tag() + self.frontend_message_tag() + list(self.js_and_css_tags())
 
         return mark_safe("\n".join(tags))
 
@@ -364,7 +380,7 @@ class WebpackBundleHook(hooks.KolibriHook):
         :returns: HTML of a script tag to insert into a page.
         """
         urls = [chunk["url"] for chunk in self.sorted_chunks()]
-        tags = self.frontend_message_tag() + [
+        tags = self.context_tag() + self.frontend_message_tag() + [
             '<script>{kolibri_name}.registerKolibriModuleAsync("{bundle}", ["{urls}"]);</script>'.format(
                 kolibri_name=conf.KOLIBRI_CORE_JS_NAME,
                 bundle=self.bundle_id,
--- a/kolibri/core/webpack/hooks.py
+++ b/kolibri/core/webpack/hooks.py
@@ -32,7 +32,8 @@ from pkg_resources import resource_filename
 from six import text_type
 
 from kolibri.plugins import hooks
-from kolibri.utils import conf
+from kolibri.utils.js_names import KOLIBRI_CORE_JS_NAME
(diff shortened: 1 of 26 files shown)


indirectlylit:
needs black formatting

rtibbles:
Rebased - have also updated the external plugin with the latest build updates, and it now builds completely independently of Kolibri. Will need to do some sanity checking to make sure it preserves the functionality, but I don't see why it wouldn't!

kolibri/core/webpack/hooks.py
```diff
@@ -272,6 +277,16 @@ def frontend_message_tag(self):
         else:
             return []
 
+    def context_tag(self):
+        if self.context:
+            return [
+                '<script>window.{bundle} = JSON.parse("{context}");</script>'.format(
```
indirectlylit: It might be preferable to keep all the context objects namespaced together. e.g. somewhere at the top 

```js
window.kolibriPluginContextObjects = {};
```

and then for each `context_tag`:

```js
window.kolibriPluginContextObjects['{bundle}'] = JSON.parse('{context}');
```

to prevent plugins from clobbering the global namespace, or potentially using names that aren't valid JS identifiers.

ref: https://github.com/learningequality/kolibri-sentry-plugin/blob/master/kolibri_sentry_plugin/assets/src/module.js#L10

indirectlylit (commented):
looks good!

made one suggestion about namespacing plugin context objects

wasn't able to fully test - should we wait to merge until `kolibri_sentry_plugin` is published?

kolibri/core/webpack/hooks.py
rtibbles: Yeah, that sounds like a good idea. I was also thinking about adding a `getContext` helper function that would mean the plugin itself wouldn't have to worry about the implementation detail of where the context was stored, sound worthwhile?

rtibbles:
Probably good to test this PR in conjunction with an unpublished version of the sentry plugin - any recommendations on the best way to test that it is properly reporting to sentry?

kolibri/core/webpack/hooks.py
indirectlylit: >  I was also thinking about adding a getContext helper function

yes, that's a good idea

indirectlylit: i wonder if 'context' is the right word... is this more 'options'?

indirectlylit:
> any recommendations on the best way to test that it is properly reporting to sentry?

for javascript, i've had luck triggering intentional "is not an attribute of undefined" errors. E.g. add to the code

```js
console.log(window.trigger.error)
```

for python, I think just tossing a `raise` in there worked for me in the past

kolibri/core/webpack/hooks.py
indirectlylit: so, presumably we could use this new functionality here, right?

https://github.com/learningequality/kolibri/pull/5847#discussion_r313025261

rtibbles: Yes - and we can also use it to consolidate all of our `default_frontend` global assignment that currently happens in a bunch of different template tags.

rtibbles:
This is now updated. Have tested logging both backend and frontend errors to sentry and both work. The external plugin is also now published to PyPi: https://pypi.org/project/kolibri-sentry-plugin/

kolibri/utils/kolibri_js_names.json
```diff
@@ -0,0 +1,4 @@
+{
+    "KOLIBRI_CORE_JS_NAME": "kolibriGlobal",
+    "KOLIBRI_JS_PLUGIN_DATA_NAME": "__kolibriPluginData"
+}
```
indirectlylit, line 4: Since there are so many variable names for these strings, it ironically makes it harder to figure out what the actual global variable is actually called: `KOLIBRI_JS_PLUGIN_DATA_NAME`, `__kolibriPluginData`, `kolibriPluginDataName`, `__kolibriPluginDataName`... but all the dev wants to do is inspect it in their console.

I think in this case, it would be preferable to lower the amount of indirection and allow the strings to be defined more than once in the codebase. If the strings are unique, meaningful, and unlikely to change, there's not a huge win from having a single source of truth. So for example:

* `window.theKolibriPluginData`
* `window.theKolibriCoreApp`

These are unique, easily-searched strings and use the [singleton naming convention](https://vuejs.org/v2/style-guide/#Single-instance-component-names-strongly-recommended). Also since they're named similarly, they'll be clustered and easy to find in the dev tools.

indirectlylit (approved):
looks good. One non-blocking comment about global names

kolibri/utils/kolibri_js_names.json
rtibbles: Yeah - as I was doing the extension here I was having a similar thought. Especially as it just moves the 'shared nameness' to the constant variable name, rather than the value itself.

I can clean this up in a follow up.

Issue #5769, opened 2019-07-15
pass buildkite env vars prefixed with KOLIBRI_BUILD_ into docker


lyw07:
### Observed behavior

Currently for UNW custom installers, we have an env var `BACKGROUND_IMAGE` that is defined in buildkite and gets passed into docker through `env.list`. 
As we will probably add more related env vars later for the custom installers, it will be better if we have a script that autogenrate the `env.list` to include all the env vars prefixed with `KOLIBRI_BUILD_` in the current environment.

### Expected behavior

…

### User-facing consequences

…

### Errors and logs

…

### Steps to reproduce

…

### Context

…

rtibbles:
Fixed in #5778

PR #5778, opened 2019-07-16
dynamically add env vars for custom installer build into docker env

lyw07:
### Summary

This PR is to add a script to dynamically add env vars prefixed with KOLIBRI_BUILD_ into docker environment for building the UN Women custom installers.
The script `build_tools/customize_docker_envlist.py` will parse through all the environment variables in the current environment, find all the variables prefixed with `KOLIBRI_BUILD_`, remove the prefix and pass the result name and corresponding value to `env.list` so that docker can read from this file.
The env.list will be restored after running docker container so that the dynamic env vars will not be committed to the repo.

### Reviewer guidance

testing buildkite pipeline:
1. `export KOLIBRI_BUILD_XXX=a`
2. run `make docker-whl` 
3. run `docker exec -it <kolibri_container>` to see if the environment variable `XXX` exists

### References

#5769 

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/build_tools/__init__.py
+++ b/build_tools/__init__.py
@@ -3,7 +3,10 @@ This module is intended to allow easier build time configuration of Kolibri.
 This is currently implemented by allowing additional build configuration options to be configured
 through environment variables.
 In order for these environment variables to propagate through to our docker build environment
-we explicitly list them in the env.list file in the root of the repository.
+we dynamically add them to the env.list file in the root of the repository.
+All the environment variables should be prefixed with KOLIBRI_BUILD_ in the current environment,
+and will be passed into the docker environment without the prefix by running the
+customize_docker_envlist.py script.
 The following environment variables are supported:
 
 BUILD_TIME_PLUGINS, a txt file containing an explicit list of all Kolibri plugins that should
@@ -32,4 +35,6 @@ This environment variable is used in the customize_requirements.py script.
 For BUILD_TIME_PLUGINS, RUN_TIME_PLUGINS, and EXTRA_REQUIREMENTS if the environment variable
 specifies a URL, the scripts will attempt to download a text file from that URL and then use
 that file.
+
+BACKGROUND_IMAGE, an image to display on the Kolibri login page.
 """
--- a/build_tools/__init__.py
+++ b/build_tools/__init__.py
@@ -35,6 +35,4 @@ This environment variable is used in the customize_requirements.py script.
 For BUILD_TIME_PLUGINS, RUN_TIME_PLUGINS, and EXTRA_REQUIREMENTS if the environment variable
 specifies a URL, the scripts will attempt to download a text file from that URL and then use
 that file.
-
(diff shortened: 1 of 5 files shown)


docker/build_whl.dockerfile
```diff
@@ -45,8 +45,8 @@ COPY . .
 
 CMD echo '--- Installing JS dependencies' && \
     yarn install --pure-lockfile && \
-    echo `--- Making whl` && \
```
lyw07, line 48: This is just to fix the error `/bin/sh: 1: ---: not found` when running the dockerfile

build_tools/__init__.py
```diff
@@ -32,4 +35,6 @@
 For BUILD_TIME_PLUGINS, RUN_TIME_PLUGINS, and EXTRA_REQUIREMENTS if the environment variable
 specifies a URL, the scripts will attempt to download a text file from that URL and then use
 that file.
+
+BACKGROUND_IMAGE, an image to display on the Kolibri login page.
```
rtibbles: Should remove this, as it is specific to the UN women plugin.

lyw07: thanks! I will remove it!

docker/env.list
```diff
@@ -29,5 +25,3 @@ KOLIBRI_RUN_MODE="docker"
 # DOCKERMNT_PEX_PATH=
 # Specify path to a .pex file in the /docker/mnt volume, e.g.
 # DOCKERMNT_PEX_PATH=/docker/mnt/kolibri-0.11.0a1.dev_git-37-gb15b923.pex
```
rtibbles, line 27: There doesn't seem to be a new line at the end of this file, so I wonder if we need to make sure that a new line is added before adding any env vars in the python script above.

build_tools/customize_docker_envlist.py
```diff
@@ -0,0 +1,34 @@
...
+    envs = os.environ
+
+    with open(ENVLIST_FILE, "a") as f:
+        for env in envs:
+            # Find all the environment variables prefixed with KOLIBRI_BUILD_
+            # and add them to env.list to pass into docker environment after
+            # removing the prefix.
+            if env.startswith(BUILD_ENV_PREFIX):
+                key = env.replace(BUILD_ENV_PREFIX, "")
+                f.write("{key}={value}\n".format(key=key, value=envs[env]))
```
rtibbles: Per my comment below, it's probably safer to prefix the write with the newline, to make sure we don't accidentally add to an existing line. So:
`f.write("\n{key}={value}".format(key=key, value=envs[env]))`

rtibbles (changes requested):
2 small changes, but the overall implementation looks right!

build_tools/customize_docker_envlist.py
lyw07: makes sense to me. I will make the change soon!

rtibbles (approved):

Issue #3451, opened 2018-03-29
Lint for unused translations 


christianmemije:
### Observed behavior

There are some strings in the codebase that are never referenced. 

### Expected behavior

Warn about unused translations

### User-facing consequences

Wastes time for translators.

I was thinking we can search if the string is referenced inside the `registerFoundMessages` function. The typical case of `$tr('translation'...` is easy to search for but sometimes translations are not referenced like that. Sometimes there is a map of translations that is passed into a function. Or in places like the Perseus exercise renderer, the strings are never actually referenced in our codebase. Thoughts @rtibbles? 

Also handling strings that are intentionally not referenced. Also the best place in the build pipeline to lint for this.

indirectlylit:
the flip side: https://github.com/learningequality/kolibri/issues/1958

indirectlylit:
@nucleogenesis just noticed that this issue exists

Issue #5410, opened 2019-04-23
Simplify linting process


MisRob:
1. Eliminate HTMLHint and replace all rules with eslint ones
2. Eliminate custom indent/spacing functions - some parts already are or can be replaced by `vue/html-indent`, `vue/script-indent`, for some of them we'll need new eslint rules (mostly spacing stuff)
2. If possible, run prettier on the entire vue component, and using the eslint-plugin-prettier.
3. Run stylelint on the entire vue component after eslint/prettier has been run on it, to reduce callback/promise hell.

So `lint.js` would be simplified to:
1. If JS, run eslint with prettier plugin.
2. If CSS/SCSS/SASS/LESS, run stylelint (open question of whether we want to run prettier or not on our style files too).
3. If Vue component, run stylelint, then run eslint with prettier.

References: #5400

### User-facing consequences

None

indirectlylit:
similar issue from last year here: #4050

Issue #4050, opened 2018-07-19
deprecate htmlhint in favor of eslint for template linting


indirectlylit:
### Observed behavior

`htmlhint` has a few issues:

* poor error reporting
* hacky support for vue files
* poor performance
* not maintained

However, it does enforce a number of important rules.

### Expected behavior

If possible, we should migrate all linting to `eslint`.

### User-facing consequences

None

### Context

0.10.0

indirectlylit:
closing because #5410 has more information

PR #5790, opened 2019-07-18
Warn about unused translation strings during frontend-lint

nucleogenesis:
### Summary
This will add a process in the `frontend-lint` command that will warn the user of any unused strings in a given Vue file.

refs: #3451

### Reviewer guidance

Run `yarn lint-frontend` and see the warnings. Check as many of them as you can by opening the file indicated and checking for the following:

- Is the key of the defined translation string never passed to a `$tr()` or `coachCommon$tr()` function?
- Is the key never defined in an array or object so that it could then be passed to a `$tr()` function in an iterator or by referring to the object's key that refers to the defined translation string key?

Example:

```JS
// Presume these defined strings:

$trs: {
    topic: "Topic",
    channel: "Channel",
}

// An object definition would look like
const strMap = {
    TopicKey: 'topic',
    ChannelKey: 'channel',
}

// An array definition would look like:
const listOfStrings = ['topic', 'channel']

// Note that the values defined are the keys of the $trs object.
// So elsewhere in the code, you may see:
this.$tr(strMap['TopicKey']);
// or
listOfStrings.forEach(str => this.coachCommon$tr(str));

// Note - that any strings used this way should NOT appear in the linter's warnings about unused strings.
```

### References

This is a continued iteration on translation strings refactoring in Kolibri per [this issue](https://github.com/learningequality/kolibri/issues/5057).

[Notion project page](https://www.notion.so/learningequality/New-linting-rule-that-warns-about-unused-strings-e6421da8fe90451ab1fdcf18852c6ca2)

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated


indirectlylit:
> Should this also remove those unused strings if the :format postfix is added to the command? If not - then should we mention this in the warning message? Perhaps we can add a warning message like "Linting will not automatically remove unused translation strings.".

I definitely wouldn't want to auto-delete things, because that would cause pain during development. I believe rather than warnings, we can just make them errors. Linting is decoupled from the build process now (thanks @rtibbles!) so making them errors should not cause the build to break. In cases where it's intentional (which happens occasionally) we'll add eslint `ignore` comments.

> Is the current basic approach of implementing the linting code viable - should I seek to insert this process elsewhere in the linting system?

I'd recommend syncing with @MisRob and @jonboiser who have both done some work with custom linting

MisRob (commented):
> Should this also remove those unused strings if the :format postfix is added to the command? If not - then should we mention this in the warning message? Perhaps we can add a warning message like "Linting will not automatically remove unused translation strings.".

I agree with @indirectlylit on this.

> Is the current basic approach of implementing the linting code viable - should I seek to insert this process elsewhere in the linting system?

When I worked on linting, we actually talked about reducing custom logic in _packages/kolibri-tools/lib/lint.js_ file in favor of eslint rules as much as possible. Main reasons were that the file is difficult to read, some logic is conflicting and a couple of others. [This is a PR including some of those issues and a discussion](https://github.com/learningequality/kolibri/pull/5400) I had with @rtibbles on this topic (it also includes some problems I discovered). Also, I'm inserting references to some related tickets https://github.com/learningequality/kolibri/issues/5410, https://github.com/learningequality/kolibri/issues/4050. Bearing all this in mind, I would say that it might be suitable to implement this as `vue-no-unused-translations` [eslint rule](https://github.com/learningequality/kolibri/tree/develop/packages/eslint-plugin-kolibri) (If possible! I would not expect problems since access to template and script should be sufficient for this task. Though I haven't checked all details yet.) On the other hand, from what I've seen so far, I don't think that this logic would complicate lint.js more in a significant way so I don't consider it to be crucial, rather nice to have. But, on the other hand again, another advantage of "eslint rule" approach could be also no need to introduce `ast-traverse` and using the same traversing logic for all our lint checks instead, and also having access to or enhancing our [utilities](https://github.com/learningequality/kolibri/blob/develop/packages/eslint-plugin-kolibri/lib/utils.js) being available for use in all our current and future eslint rules. And last but not least, eslint provides `RuleTester` - could be handy!

So that's just some background from me. I'll try to play with it later (need to setup the whole Kolibri  dev env again because I've been using a different computer) :)!

rtibbles:
I agree with @MisRob that this is probably best implemented as a custom eslint rule, and I think some of the code she has implemented could be copied fairly directly to turn this into a rule.

nucleogenesis:
I didn’t commit the .eslintrc file including this rule because I figured Travis would fail. So be sure to add the rule there to activate it to try it out. 

I also will add some tests for the above.

packages/eslint-plugin-kolibri/lib/rules/vue-no-unused-translations.js
```diff
@@ -0,0 +1,73 @@
...
+'use strict';
+
+const eslintPluginVueUtils = require('eslint-plugin-vue/lib/utils');
+
+const utils = require('../utils');
+const constants = require('../constants');
+
+const { GROUP_$TRS } = constants;
+
+const $TR_FUNCTIONS = ['$tr', 'coachString', 'coreString', 'learnString'];
```
MisRob: Code looks good!

I only have one note for the new version - I would suggest moving translations functions configuration rather to the rule settings. We might need to override it from other projects. So far I only know about Studio and I am actually not 100% sure if these rules are turned on there, but it would make sense to me anyway since this is project specific unlike other logic. I guess that it might be a bit more intuitive for Kolibri folks to just update eslintrc, too.

rtibbles: I concur - our linting tools are currently used on Studio, and while this could be turned off, I think it would be a useful rule to leverage there.

rtibbles: In the kolibri `.eslintrc.js` we could then just directly import the namespaces from the `ProfileStrings.js` file!

rtibbles:
The tests failed linting, btw! Lint :allthethings:!

nucleogenesis:
### Common Strings 

Initially, I figured that I'd be working to check common strings as well. However, that doesn't really work for a linter because it would require a persistent state of the entire codebase - the linting rule that I ~stole~ borrowed from @MisRob 's rules tracks the state for each file individually.

We would need to basically get a list of all defined strings that persists across the linter's traversal of each file - then track their uses for each common namespace. 

I'll do a bit of digging to see if this is possible within a linting rule.

### Added logic for tracking 'dynamic' string uses

Example:

```Javascript
<script>
        export default {
          computed: {
            tooltipText() {
              const kindToLabeLMap = {
                [ContentNodeKinds.TOPIC]: 'topic',
                [ContentNodeKinds.CHANNEL]: 'channel',
                [ContentNodeKinds.EXERCISE]: 'exercise',
                [ContentNodeKinds.VIDEO]: 'video',
                [ContentNodeKinds.AUDIO]: 'audio',
                [ContentNodeKinds.DOCUMENT]: 'document',
                [ContentNodeKinds.HTML5]: 'html5',
                [ContentNodeKinds.EXAM]: 'exam',
                [USER]: 'user',
              };
              const label = kindToLabeLMap[this.kind];
              return label ? this.$tr(label) : '';
            },
          },
          $trs: {
            topic: 'Topic',
            channel: 'Channel',
            exercise: 'Exercise',
            video: 'Video',
            audio: 'Audio',
            document: 'Document',
            html5: 'App',
            exam: 'Quiz',
            lesson: 'Lesson',
            user: 'User',
            slideshow: 'Slideshow',
          },
        }
      </script>
```

Previously, the linter would say that all of the defined strings above were unused because we were tracking whether a string was used or not based on whether or not it ever showed as an argument to the `$tr()` function.

However, in some cases in Kolibri, defined strings' keys are in an array or object and some token representing them is passed to `$tr()`.

Now, if any of the keys that are defined ever show up as a value in an array or as an assigned value in an object, then it is considered to be 'used' and won't show a warning.

rtibbles:
Just to double check, this introspects the object to confirm that the string labels are listed therein?

nucleogenesis:
> Just to double check, this introspects the object to confirm that the string labels are listed therein?

Basically - it goes to any object (`ObjectExpression` in the AST) defined between the `<script>` tags - gets all of the values that are assigned (anything right-hand during the key-value assignment) - and if that matches any of the defined strings' keys (eg, coachesLabel, submitAction) then it considers that a "use" and will not warn in the linting process.

My reasoning is that - it isn't likely that you'll be creating a string in a component called `whateverThingLabel` which you then later assign to an object or put into an array as a value in a context that isn't specifically related to using that `whateverThingLabel` to refer ultimately back to the defined translation string.

It could certainly happen, especially if strings are labeled like they are in my above code example, but as we continue being more strict in naming our strings to better represent their actual meaning, such as by appending their functionality to them with words like "Label" or "Action", the less likely we would be to run into this kind of situation.

The other side of that coin is that we end up warning "Unused translation found for [topic/channel/etc]" in the code in my above comment and any code that works similarly.

nucleogenesis:
I'm not seeing any way to run the linting rule across every file while passing along a bit of stateful information. So - I'd suggest that the LOE of including Common files in this linting process may be higher than the value it would provide and the solution would likely be kind of hacky (reading/writing a temp JSON file to pass state between files being linted).

Although if anybody has a thought on how to go about this please let me know!

EDIT: One thing to consider is that the `yarn makemessages` now creates a "Profile" CSV in the `locale/en/LC_MESSAGES/profiles` directory which breaks down every definition and it's uses. So we may opt to have a once-per-major-release audit of those generated files to manually find any Common strings that were never used. Wouldn't take long to do, especially after it's been done once.

indirectlylit:
For a first pass, skipping common files sounds like a fine call to make.

At least initially, we won't be putting things in common files unless they're used in at least a couple places anyway. Over time it's possible the common files will accumulate "dead strings" but those should be fairly easy to weed out occasionally.

MisRob:
> I'm not seeing any way to run the linting rule across every file while passing along a bit of stateful information. So - I'd suggest that the LOE of including Common files in this linting process may be higher than the value it would provide and the solution would likely be kind of hacky (reading/writing a temp JSON file to pass state between files being linted).

> Although if anybody has a thought on how to go about this please let me know!

I've never tried to pass information between more files and am not aware of how to do that. I've just went through the whole eslint developer documentation and haven't noticed anything helpful so far. If we'd need to do this, I think that it would be better to stay with the first approach indeed, as you suggest. Though I suppose that common strings checks would introduce also the need to track relationship between files in terms of imports and that might be even more complex.. Other rules do not deal with this as well (for example checks for unused methods are performed only in scope of one file and I didn't deal with methods imported using mixins etc.). So I think that it's fine too, especially as long as common strings are not the main source of translations.

rtibbles (approved):
This seems ready to merge to me.

Issue #5678, opened 2019-06-19
Super admin label makes user CoreTable get internal vertical scroll


rtibbles:
### Observed behavior
Unnecessary vertical scrollbar

![image](https://user-images.githubusercontent.com/1680573/59729561-c6ce4c80-91f3-11e9-96e5-70420d25fc19.png)

### Expected behavior
Don't show a scroll bar unless you have to.

### User-facing consequences
Ugly

### Context
 * Kolibri version: 0.12.4
 * Operating system: Ubuntu
 * Browser: Chrome

nucleogenesis:
@rtibbles I'm not able to replicate this on Debian w/ Chromium or Chrome. I've looked at every table where users are listed that I can think of but perhaps I'm not looking in the right place?

rtibbles:
Hrm - let me see if I can still replicate on current release.

rtibbles:
I also can no longer replicate. Let's close this and reopen if it recurs!

bjester:
I can replicate this whenever a user doesn't have name. The fix is fairly straightforward, so I'll throw it in.

bjester:
Reopening for fix: https://github.com/learningequality/kolibri/pull/5919

jonboiser:
Fixed for this particular table in #5919 by making sure rows with an empty full name cell get the same height as other cells.

I think that we should probably not allow facility users to be made without a full name in the importusers command, as this isn't allowed when making a user inside of Kolibri.

rtibbles:
Seems like a reasonable restriction - can you open an issue?

PR #5806, opened 2019-07-24
Add demographic info fields to all user creation workflows

jonboiser:
### Summary

1. Adds demographic info fields (birth year, gender, identifier number) to alluser creation and editing workflows: Sign Up Page, User Create/Edit Page, Profile Page, Setup Wizard Superuser Step
1. Various refactors, including consolidating username, full name, and password textboxes from different workflows into re-usable components
1. updates `importsusers` management command to handle CSV files that provide birth year, id, and gender

### Reviewer guidance

Check all of the workflows:

1. Setup wizard
1. Sign Up Page
1. User Profile Page
1. Create New User
1. Edit Existing User

If demographic data is (not) provided it should be reflected in in the User Profile Page as well as the Facility Users page.

Look out for unusual displays of these data like the special string `DEFER` and empty string `"''"`.

Some other things affected in refactoring

1. Changing own password
1. Changing other user's password
1. Changing Roles in User Edit Page

Check that the "Update your profile" notification is (not) shown in the right situations: Check the Gherkin file in this PR to see different scenarios.

Try out the CSV import and try scenarios that differ from the unit tests
### References

…

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated


CHANGELOG.md
```diff
@@ -2,6 +2,12 @@
 
 List of the most important changes for each release.
 
+## 0.13.0 (unreleased)
+
+### Changed or fixed
+
+- Fixed: Unclean shutdowns on very large databases, due to prolonged database cleanup
+
```
indirectlylit: did the merge work correctly? This shouldn't be in the diff

jonboiser: Yeah, I merged develop into this branch after #5888 was merged into develop.

kolibri/core/auth/migrations/0014_auto_20190815_1421.py
```diff
@@ -0,0 +1,44 @@
...
+            model_name="facilityuser",
+            name="gender",
+            field=models.CharField(
+                blank=True,
+                choices=[
+                    ("MALE", "Male"),
+                    ("FEMALE", "Female"),
+                    ("NOT_SPECIFIED", "Not specified"),
+                    ("DEFERRED", "Defers for later"),
+                ],
```
bjester, line 34: Some check constraints for gender and age could be worthwhile.

packages/kolibri-components/src/KPageContainer.vue
```diff
@@ -47,6 +47,7 @@
 
     padding: 8px 24px 24px;
     margin-top: 24px;
+    overflow-x: scroll;
```
bjester: This is causing a horizontal scrollbar on all page containers.

jonboiser: This might be another case of #5678, with `scroll` working differently on different platforms. In devtools, is there a version of this rule that works for your browser? I could also just move this overflow rule to the wide tables.

kolibri/core/auth/migrations/0014_auto_20190815_1421.py
jonboiser: In auth.models.FacilityUser, I think I get the checks for free by setting the `choices` kwarg. I also wrote a validator for the `birth_year` char field

packages/kolibri-components/src/KPageContainer.vue
bjester: Well, `scroll` makes the scrollbar always show. On Apple platforms, I think there is styling that essentially makes it invisible, but on FF/Chrome on other platforms it always shows the bulky scrollbar. The value `auto` would be good, but there may be cross-browser inconsistencies. Did you add this primarily for the User table?

jonboiser: The UserTable actually already has its own overflow rule. There are other fixed-width very wide, non-table layouts in the app like the Exercise Report that aren't that can overflow the container, so putting this on the KPageContainer is a catch-all.

I'll just change the value to `auto`.

bjester: Okay sounds good.

kolibri/plugins/facility_management/assets/src/views/UserCreatePage.vue
```diff
@@ -0,0 +1,295 @@
...
+  }
+
+  .narrow-container {
+    width: 500px;
+    margin: auto;
+  }
+
+  .form {
+    max-width: 400px;
+    min-height: 950px;
```
bjester: I noticed that the container was quite a bit larger than the actual form. What's the reason for this?

kolibri/plugins/facility_management/assets/src/views/UserCreatePage.vue
```diff
@@ -0,0 +1,295 @@
...
+    margin-right: auto;
+    margin-bottom: 100px;
+    margin-left: auto;
+  }
+
+</style>
```
bjester, line 295: You may want to review the styling here, as it doesn't seem too responsive. On viewport widths less than 550px, it starts to suffer.
screenshot-localhost-8080-2019 08 23-14-28-10

kolibri/plugins/facility_management/assets/src/views/UserCreatePage.vue
jonboiser: Some of the Keen-UI form components have a fixed width of 400px, we make the group the same width for consistency. I think later we'll try to extend the KPageContainer to fit to contents when we need it to.

bjester: But the `min-height`?

jonboiser: That's to make sure everything fits when _everything_ is in the form. The form is the tallest when the User Kind dropdown + Radio button fieldset is there (try editing a user account that's not your own). Without the extra height, the birth year dropdown will overflow the bottom.

kolibri/plugins/facility_management/assets/src/views/UserCreatePage.vue
jonboiser: Cool. Changing the `.narrow-container` rule from `width` to `max-width` should fix this.

kolibri/plugins/facility_management/assets/src/views/UserCreatePage.vue
bjester: Seems like just a `padding-bottom` could be set to always give the bottom buffer needed for the dropdown?

jonboiser: yeah, i can do that

jonboiser: it does make the container jump a bit when you toggle the kind from coach to non-coach.

bjester:
Another minor design observation, some extra space has appeared underneath the input placeholder for the password verification field on the sign up page and the setup wizard account form.
screenshot-localhost-8080-2019 08 23-15-39-42

jonboiser:
I don't see that on my computer. Can you see in dev tools what might be causing that extra space? It almost looks like the placeholder is in mid-transition.

bjester:
Heh, so it appears the extra space is related to something with the font. The `-` character produces a larger line height than the other characters, creating a larger height for the placeholder, and hence the extra space. I don't see the issue [here](http://kolibribeta.learningequality.org/en/user/#/create_account ) but I don't know when that was updated.

jonboiser:
That's weird. Are you able to see if the dash is from a different font-family than the rest of the word?

indirectlylit:
> so it appears the extra space is related to something with the font

@bjester would you mind filing an issue for this? Sounds similar to some strange behaviors we saw with Korean characters

kolibri/core/assets/src/mixins/commonCoreStrings.js
```diff
@@ -44,23 +49,24 @@ export const coreStrings = createTranslator('CommonCoreStrings', {
...
-  // toggles
-
-  // formatted values
-
-  // Errors
+  // Demographic-specific strings
+  genderOptionMale: 'Male',
+  genderOptionFemale: 'Female',
+  genderOptionOther: 'Other',
+  notSpecifiedOption: 'Not specified',
```
indirectlylit: let's call this `genderOptionNotSpecified` because in some languages 'not specified' will need to match the gender of the word 'gender' (woo meta!)

also cc @nucleogenesis are there conventions we're trying to establish for (a) the naming of core strings and (b) the decision of whether or not a string should be in core? I want to make sure we're all on the same page with that

bjester:
> @bjester would you mind filing an issue for this? Sounds similar to some strange behaviors we saw with Korean characters

Done! https://github.com/learningequality/kolibri/issues/5915

bjester:
Status update: All I have left to check are the CSV import and some of the notification scenarios, but otherwise it's looking good so far.

bjester:
@jonboiser I'm not able to close the Usage and Privacy modal when launched from the *update your profile* notification. It closes when I open it from the sign-in page though. There are no errors, the button just doesn't seem to work.

jonboiser:
Cool. Just fixed the PrivacyModal issue and made sure the other PrivacyModals had both their `submit` and `cancel` handlers defined.

bjester (approved):
LGTM! :100:

indirectlylit:
please merge when ready

PR #5834, opened 2019-07-30
Fix Hashi behaviour for Proyecto Descartes HTML5 content

rtibbles:
### Summary
* Previously, Hashi would do a server fetch to get the actual page content after Hashi had been initialized. This led to the window `load` event being called before any on page Javascript for the HTML5 app was around to hear it. To compensate for that, we mocked the window `load` event and fired it again for their benefit.

* Now Hashi loads in a more synchronous manner, so in page Javascript will load before the window `load` event is fired.

* This had the slightly odd side effect of causing Proyecto Descartes HTML5 apps to go into an infinite loop, for reasons that I could not quite discern, but seemed to have to do with writing to the DOM and setting resize listeners within the load event handler.

* To fix this, this PR checks for the `document.readyState` property. The window `load` event is only fired if the property is already set to `'complete'` indicating that the window `load` event has already been fired.

### Reviewer guidance
Import a channel containing Proyecto Descartes content (e.g. `447cedc184ea4345aaedadf2fd6da906`) and import the content node `Habilidades Básicas - ONU Mujeres -> Proyecto Descartes -> Matemáticas -> 10-13 años -> Áreas de polígonos`.

Verify that before this fix viewing the content node will freeze the page and spike CPU to 100%, and that after this fix it runs normally.

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/packages/hashi/src/replaceScript.js
+++ b/packages/hashi/src/replaceScript.js
@@ -74,8 +74,12 @@ export function seq(arr, index) {
       const loadEvent = document.createEvent('Event');
       loadEvent.initEvent('load', true, true);
       document.dispatchEvent(DOMContentLoadedEvent);
-      document.dispatchEvent(loadEvent);
-      window.dispatchEvent(loadEvent);
+      if (document.readyState === 'complete') {
+        // Only call this if the document is already fully loaded
+        // and we need to emulate the window load event.
+        document.dispatchEvent(loadEvent);
+        window.dispatchEvent(loadEvent);
+      }
       const elements = document.querySelectorAll(':not(script)');
       Array.prototype.forEach.call(elements, element => {
         element.dispatchEvent(loadEvent);


kollivier:
From your description, it sounded to me like we can now rely on the browser onload event firing at the right time, is this not the case? Is there another reason we need to fire off our own version of the event?

rtibbles:
In the Proyecto Descartes case, we definitely can. I originally just removed the lines, rather than conditionally executing them, both worked. I added the conditionality to reduce the risk of a regression to other content, in case the browser onload event timing was inconsistent.

kollivier:
If the browser is inconsistent, then that means we're still going to have two onload events firing at unknown times, so I'm concerned we may end up just moving the problem from one set of apps to another. I think if we can't think of a clear scenario where not firing this should fail, we should try without it, as firing the event twice isn't ever a good thing to do if we can at all avoid it.

rtibbles:
It's not an inconsistency in the browser - rather in the initialization of Hashi.

Hashi does not execute the in page scripts until it has received a ready signal via postMessage from outside the sandboxed iframe.

If this ready signal has already been received before it gets to the point of executing the scripts in its initialization logic, it will continue synchronously and execute the scripts. However, if for some reason (perhaps slow data load from the server in Kolibri) the ready signal has not been emitted, then the script execution is run conditionally on receiving that ready signal. This will cause Hashi to release the main thread - as Hashi is the last tag in the page, this will cause the window to fire the `load` event.

When the ready signal fires Hashi will now execute all the scripts in the page, but unless we fire the load event ourselves, then any scripts executed that are waiting for it will never receive it.

We could make this slightly more bullet proof, and change the conditionality of the execute script callback when the ready signal has not been received yet, and instead wait for both the ready signal *and* the `load` event to have been fired. This will guarantee that the executed scripts will only ever see one `load` event.

kollivier:
I think I probably still just don't have a full understanding of how this all works, particularly the precise order of operations for page load. As this is time sensitive, for now, let's leave this PR as-is and I'll test it out with as much HTML5 app content as I can. If everything looks good, I'll give it the green light!

rtibbles:
Yeah - I thought this was the most conservative change, which even though it might still leave other content broken, at least won't break other content that is currently working.

kollivier (approved):
LGTM!

PR #5839, opened 2019-07-31
Add skip navigation link to Core Base (App Bar)

jonboiser:
### Summary

1. Adds a SkipNavigationLink component that is embedded in the AppBar (the component that has the hamburger, tabs, and other nav items that may want to be skipped). It is not embedded in the "ImmersiveBar" since that bar does not have nav items that would be skipped.
1. In CoreBase, wraps the main slot in a div with `role` and `id` equal to "main" that are used by SkipNavigationLink to target and focus when the link is clicked.
1. Fixes some warnings about localized strings

Screen Shot 2019-07-31 at 2 32 58 PM

### Reviewer guidance

1. Take a look at the comments in the new code to see if the decisions made make sense
1. Navigate the app using keyboard only and make sure the link appears when it should and does not when it shouldn't and that it "works" properly.

### References

…

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/core/assets/src/views/CoreBase/index.vue
+++ b/kolibri/core/assets/src/views/CoreBase/index.vue
@@ -87,7 +87,7 @@
         <AppError />
       </KPageContainer>
 
-      <div v-else id="main" role="main" tabindex="-1">
+      <div v-else id="main" role="main" tabindex="-1" class="main">
         <slot></slot>
       </div>
     </div>
@@ -448,8 +448,13 @@
 
   @import '~kolibri.styles.definitions';
 
-  #main {
+  .main {
     height: 100%;
+    // Don't put focus outline on the main div even if still in
+    // keyboard modality
+    &:focus {
+      outline-style: none !important;
+    }
   }
 
   .scrolling-pane {
--- a/kolibri/core/assets/src/views/CoreBase/index.vue
+++ b/kolibri/core/assets/src/views/CoreBase/index.vue
@@ -87,7 +87,9 @@
         <AppError />
       </KPageContainer>
 
-      <slot v-else></slot>
+      <div v-else id="main" role="main" tabindex="-1">
+        <slot></slot>
+      </div>
     </div>
 
     <GlobalSnackbar />
@@ -445,6 +447,10 @@
 <style lang="scss" scoped>
 
   @import '~kolibri.styles.definitions';
+
+  #main {
+    height: 100%;
+  }
 
   .scrolling-pane {
     position: absolute;
--- a/kolibri/core/assets/src/views/CoreBase/index.vue
+++ b/kolibri/core/assets/src/views/CoreBase/index.vue
@@ -450,11 +450,11 @@
 
   .main {
     height: 100%;
-    // Don't put focus outline on the main div even if still in
-    // keyboard modality
-    &:focus {
-      outline-style: none !important;
-    }
+  }
+
+  // When focused by SkipNavigationLink, don't outline non-buttons/links
+  /deep/ [tabindex='-1'] {
+    outline-style: none !important;
   }
 
   .scrolling-pane {
--- a/kolibri/plugins/learn/assets/src/views/classes/AllClassesPage.vue
+++ b/kolibri/plugins/learn/assets/src/views/classes/AllClassesPage.vue
@@ -2,7 +2,7 @@
 
   <div>
     <div v-if="isUserLoggedIn ">
-      <h2>{{ coreString('classesLabel') }}</h2>
+      <h1>{{ coreString('classesLabel') }}</h1>
       <p v-if="!classrooms.length">
         {{ $tr('noClasses') }}
       </p>
--- a/integration_testing/features/common/ORDER.txt
+++ b/integration_testing/features/common/ORDER.txt
@@ -1,3 +1,4 @@
 view-all-content-types.feature
 appbar-scroll.feature
 appbar-user-type.feature
+skip-navigation-link.feature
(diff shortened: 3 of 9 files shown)


rtibbles (commented):
Behaviour matches what is detailed in the gherkin specs for this project. Could we add those directly in this PR?

I don't see any issues in the code itself. @radinamatic any concerns here?

radinamatic:
Having some issues running the VBox atm, as soon as I sort it out will give it a whirl!

indirectlylit:
the text styling is not exactly to spec, but I actually like this a bit better

ref: https://www.notion.so/learningequality/UX-UI-design-0d178fd46e1040b093f47a51cf28a49d

kolibri/core/assets/src/views/CoreBase/index.vue
```diff
@@ -87,7 +87,9 @@
         <AppError />
       </KPageContainer>
 
-      <slot v-else></slot>
+      <div v-else id="main" role="main" tabindex="-1" class="main">
+        <slot></slot>
+      </div>
```
indirectlylit, line 92: This seems fine, but curious whether we could just do

```html
<slot v-else id="main" role="main" tabindex="-1" class="main"></slot>
```

kolibri/core/assets/src/views/CoreBase/index.vue
```diff
@@ -446,6 +448,15 @@
 
   @import '~kolibri.styles.definitions';
 
+  .main {
+    height: 100%;
+    // Don't put focus outline on the main div even if still in
+    // keyboard modality
+    &:focus {
+      outline-style: none !important;
+    }
```
indirectlylit: would you mind showing gifs of what the behavior is with and without this?

indirectlylit (commented):
implementation looks reasonable to me. Haven't tested yet.

radinamatic:
Ok, tested now, and it behaves according to the Gherkin scenario in the specs. Couple of issues and a (slight) proposed change.

1. User menu button lost the visual focus indicator somewhere along the way: when they choose NOT to use the `Skip to content` link, and continue tabbing, in Firefox there is a slight button color change when the user menu button receives focus, but no border/outline (**issue 1a**), and on the next <kbd>Tab</kbd> press the inner text label receives focus (**issue 1b** - there's no need for the label to be focusable).

user-menu-focus-FF

In Chrome the text label does not seem focusable, but the user menu button has no visual focus indicator whatsoever (**issue 1c**).

user-menu-focus-Chrome


2. Maybe we should point the `Skip to content` link to jump to the **first heading (H1)** of the `div.main` instead of the container itself...? The user-facing result would practically be the same , but it might solve the issue @jonboiser encountered when testing in VoiceOver, namely the screen reader output was _everything_ inside the container:

Screen Shot 2019-08-01 at 10 33 03 AM

kolibri/core/assets/src/views/CoreBase/index.vue
jonboiser: Here's what it looks like without this override. It's because the kolibriOutline rule with keyboard modality applies to everything (has a `*` selector)

skip link

jonboiser:
I haven't noticed the differences in tabbing the user drop down before. In Chrome, It always consistently outlines the entire button (and doesn't tab into and outline the username label). I guess that is a bug, if we want the square outline box to show everywhere. (Don't want to fix it in this PR though).

I think I _can_ change the the focus behavior to find the first 'h1' in the main div, if it exists.

kolibri/core/assets/src/views/CoreBase/index.vue
jonboiser: Nope, doesn't work. I think the only API slots expose is `v-slot`.

jonboiser:
@radinamatic 	I'm noticing that the top headers on some of the Learn pages are actually 'h2's, so my strategy won't work. Should they be 'h1's for optimal markup?

jonboiser:
I've updated this with @radinamatic 	's suggestion to focus on the first header.

This has the nice benefit that if a header lies _below_ more navigation like a backwards link or breadcrumbs, the Skip Navigation Link will take you past those as well and save you some extra tabbing to get to the main content. :+1:

It doesn't change the amount of output you get from MacOS voice over. It doesn't pause after the header, and basically goes "header, Classes, [the rest of the content]"

kolibri/plugins/learn/assets/src/views/classes/AllClassesPage.vue
```diff
@@ -2,7 +2,7 @@
 
   <div>
     <div v-if="isUserLoggedIn ">
-      <h2>{{ coreString('classesLabel') }}</h2>
+      <h1>{{ coreString('classesLabel') }}</h1>
```
radinamatic, line 5: If this gets to become H1, then the best practice would be to change content card titles to H2 🙂

radinamatic (approved):
Looking good, let's fix the inconsistency in the headings on the Learn page, and this can be merged.

I filed the #5845 to fix user menu focus outline separately.

jonboiser:
@radinamatic 	on the "Recommended" page the h2s are the section headers like "Popular", "Resume",etc. On that page, should the Content Cards' titles be wrapped under an h3?

indirectlylit:
headers in the cards are difficult to get right because they appear at different header nesting levels. Let's open a new issue for that

indirectlylit:
> I haven't noticed the differences in tabbing the user drop down before. In Chrome, It always consistently outlines the entire button (and doesn't tab into and outline the username label). I guess that is a bug, if we want the square outline box to show everywhere. (Don't want to fix it in this PR though).

Let's open a new issue for this, too. I'd like to move to a place where _all_ elements use the standard square outline box for keyboard tab navigation.

radinamatic:
> headers in the cards are difficult to get right because they appear at different header nesting levels. Let's open a new issue for that

Right, I forgot about that... It actually counts as a _warning/recommendation_, and not an _error_ in a11y checkers, so agreed that we can deal with it in another PR! 👍

PR #5847, opened 2019-08-05
work on common component library

indirectlylit:
### Summary

* Work on the style guide to better support component documentation
* Moved all components that might be thought of as "shareable" with external projects to a new `kolibri-components` yarn work
* Globally register `K*` components and theme
* Removed the `K` prefix from anything not in the shared library. Their time will come eventually!
* decoupled kolibri-specific theme config from general color config

### Reviewer guidance

This is a massive PR, but I think something like this will be necessary if we want to move forward on making re-usable components across our products.

Thoughts are welcome.

### References

…

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated


kolibri/core/assets/src/views/AppBar.vue
```diff
@@ -170,6 +169,7 @@
     },
     created() {
       window.addEventListener('click', this.handleWindowClick);
+      this.$kolibriTheme = global.kolibriTheme;
```
rtibbles, line 172: Why are we directly referencing the global here?

indirectlylit: The top-level `global.kolibriTheme` contains kolibri-specific config, e.g. logo behaviors and styling. The generic stuff is used here: https://github.com/learningequality/kolibri/pull/5847/files#diff-e46818b93613a96345b99fe7bbe04beaR15

rtibbles: I would rather that we don't rely on the global, and have everything go through the theme mixin, so that we have a single point for relying on the globals (similarly to how we only use the language globals in the `i18n` functionality and then everything else relies on imports from that module).

indirectlylit: okay - I can make an independent mixin for the config options that are kolibri-specific

indirectlylit: to be addressed using new `pluginData` functionality

rtibbles: Noting that to do this, it would still have to be defined in a core module somewhere, in case it was to be referenced outside of the `default_frontend` bundle.

rtibbles (approved):
Can merge as soon as Mr Travis says so.

PR #5853, opened 2019-08-06
Content refactor

rtibbles:
### Summary
* Refactors content models to clearly disambiguate imported content metadata from locally generated metadata. Makes no actual change to the schema, just how it is represented in code.
* Fixes a bug in the Postgresql backed queue implementation whereby initializing the DB connection would fail when the port was left undefined (and therefore meant to be the postgres default value).
* Adds the capability to set a target for content import (aside from the default DB), which would allow for schema conversion of imported channel DBs to allow for non-destructive previewing of updated channel DBs.
* Fixes generation of import schema through the `generate_schema` management command, so that we can once again make updates to the import content schema.

### Reviewer guidance
Most importantly is the tests all passing.
Secondly is if the doc strings in the new model modules are sufficient.

### References
No direct references.

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated


rtibbles:
So glad I have good test coverage on this - will fix up the test issues.

rtibbles:
Have isolated the test error here to postgres sessions not getting disconnected, not with the primary behaviour. Will continue to investigate.

rtibbles:
I have updated this PR to separate out the Postgres speed up work which seems to be hitting some test failures (only on Travis, I fixed all the ones I could replicate locally), so that the work to support changes to the import export workflow can be merged sooner.

kolibri/core/content/utils/channel_import.py
```diff
@@ -115,9 +117,19 @@ def __init__(self, channel_id, channel_version=None, cancel_check=None):
...
-            schema_version=CONTENT_SCHEMA_VERSION, app_name=CONTENT_APP_NAME
-        )
+        if destination is None:
+            # If no destination is set then we are targeting the default database
+            self.destination = Bridge(
+                schema_version=CONTENT_SCHEMA_VERSION, app_name=CONTENT_APP_NAME
+            )
+        else:
+            # If a destination is set then pass that explicitly. At the moment, this only supports
+            # importing to an arbitrary SQLite file path.
```
ralphiee22, line 127: Assuming we don't need any SQLite specific code checks?

ralphiee22 (approved):
Haven't tested locally, but code looks good.

kolibri/core/content/utils/channel_import.py
```diff
@@ -469,7 +469,12 @@ def check_and_delete_existing_channel(self):
                         new_channel_version=self.channel_version,
                     )
                 )
-                self.delete_old_channel_data(existing_channel.root.tree_id)
+
+                root_node = self.destination.session.query(
```
ralphiee22, line 473: is a regression test needed for this?

kolibri/core/content/utils/channel_import.py
rtibbles: We might want to add some extra guards to ensure that this is only a SQLite file path, but this is more of a proof of concept at this point, as there is no practical way to access this code path.

kolibri/core/content/utils/channel_import.py
rtibbles: Possibly, I've opened a follow up issue here: https://github.com/learningequality/kolibri/issues/5960

PR #5854, opened 2019-08-06
Linting rule warning about using undefined translation strings

nucleogenesis:
### Summary

This is basically the flip side to https://github.com/learningequality/kolibri/pull/5790 - this linting rule will warn when we try to use a string that isn't defined.

Example:

```vue
<template>
    <div>
        <h1>{{ $tr('undefinedLabel') }}</h1>
    </div>
</template>

<script>
export default {
    name: 'my-component',
    $trs: {
        coachLabel: "Coach",
    }
}
</script>
```

In this case, there is an attempt to use a string by the key `undefinedLabel` - but it is not defined in $trs, so the linter will show an warning.

*Note that this will not address attempts to use common strings that are not defined.* So this works specifically at the level of an individual Vue component.

### Reviewer guidance

Add this rule to the .eslintrc file to activate it. Add a call to `$tr()` somewhere in a component passing an undefined string to it. `yarn lint-frontend` and see the expected error.

### References

Supplementary to https://github.com/learningequality/kolibri/pull/5790

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated


packages/eslint-plugin-kolibri/lib/utils.js
```diff
@@ -211,4 +211,33 @@ module.exports = {
...
+   * Report uses of undefined strings
+   */
+  reportUseOfUndefinedTranslation(context, definitions, uses) {
+    const definedStrings = definitions.map(prop => prop.name);
+    const badAttempts = uses.filter(prop => !definedStrings.includes(prop.value));
+
+    badAttempts.forEach(node => {
+      context.report({
+        node,
+        message: `Attempt to use undefined string ${node.value}`,
```
jonboiser: Maybe just something more matter-of-fact like `Message not defined in $trs: "<string_id>"` and string_id also wrapped in quotes with printed out.

packages/eslint-plugin-kolibri/lib/utils.js
```diff
@@ -211,4 +211,33 @@ module.exports = {
...
+  /**
+   * Report unused translation definitions.
+   */
+  reportUnusedTranslations(context, definitions, uses) {
+    const unused = definitions.filter(prop => !uses.includes(prop.name));
+
+    unused.forEach(prop => {
+      context.report({
+        node: prop.node,
+        message: `Unused translation defined for ${prop.node.name}`,
```
jonboiser: Have it match the style of the other errors, like `Unused message found in $trs: "<name>"`

jonboiser:
The 'unused' rule assumes that an ID appears as a value in `$trs`, it's being used. Screenshot shows no errors in my editor.

Screen Shot 2019-08-07 at 2 13 13 PM

packages/eslint-plugin-kolibri/lib/rules/vue-no-unused-translations.js
```diff
@@ -0,0 +1,101 @@
...
+        }
+      },
+    },
+    // This will include any defined Identifiers (eg, coachesLabel or submitAction) that is
+    // found inside of an array or as the value assigned in an object. This covers the case
+    // where a string may be used dynamically or in an iterator.
+    {
+      ObjectExpression(node) {
+        node.properties.forEach(prop => {
+          if (prop.value.value) {
```
jonboiser: You need to make these guards a lot safer, since this rule was getting some 'undefined is not an object' errors on some of my files. I replaced all the guards with `lodash/get` and it seemed to fix it.

e.g.

```
if(get(value, ['value', 'value']) {...}
```

packages/eslint-plugin-kolibri/lib/rules/vue-no-unused-translations.js
```diff
@@ -0,0 +1,101 @@
...
+    },
+  };
+
+  const scriptVisitor = Object.assign(
+    {},
+    // Check for a use within an $tr() function - a 'true use' of the string.
+    {
+      'CallExpression[callee.type="MemberExpression"]'(node) {
+        if (node.callee.property.name == $TR_FUNCTION && node.arguments.length) {
+          node.arguments.forEach(arg => usedStrings.push(arg.value));
```
jonboiser: Guard against `arg` being undefined here

packages/eslint-plugin-kolibri/lib/rules/vue-no-unused-translations.js
```diff
@@ -0,0 +1,101 @@
...
+      }
+    })
+  );
+
+  const templateVisitor = Object.assign(
+    {},
+    {
+      'CallExpression[callee.type="Identifier"]'(node) {
+        if (node.callee.name == $TR_FUNCTION && node.arguments.length) {
+          node.arguments.forEach(arg => Boolean(arg.value) && usedStrings.push(arg.value));
```
jonboiser: Use an if-guard instead of short circuiting here

jonboiser:
In this PR, could you add the two new rules to `eslintrc`. We'll want to have them enabled by default.

nucleogenesis:
Travis is failing because of the issues living in the code. Should we clean up to get the linter passing and keep it in eslintrc?

jonboiser:
Are there a lot of them? If it's not too many, we can put in eslint-disable-next-line comments in the mean time.

jonboiser:
There's some strings we don't want to delete because of crossComponentTranslator, but if you're sure that a certain clean up area is safe, let's go for it. (But handle the crossComponentTranslator case with the eslint-disable comments).

nucleogenesis:
> crossComponentTranslator

I previously removed all calls to crossComponentTranslator. The only instance of it is its definition. A part of the strings refactor (https://github.com/learningequality/kolibri/pull/5718) was to remove all uses of it, so any place where it was previously used to define strings - those strings were instead just defined on the component directly rather than using crossComponentTranslator.

> Are there a lot of them? If it's not too many, we can put in eslint-disable-next-line comments in the mean time.

There are about 30-40 individual instances across about a dozen files. 

In any case - if we're going to merge this with the rule enabled in `.eslintrc` then we ought to push a set of fixes in quick succession right?

The same goes for the flip-side of this PR (https://github.com/learningequality/kolibri/pull/5790) (which has far more instances that need to be fixed).

indirectlylit:
> In any case - if we're going to merge this with the rule enabled in .eslintrc then we ought to push a set of fixes in quick succession right?
> 
> The same goes for the flip-side of this PR (#5790) (which has far more instances that need to be fixed).

Yes, generally in the past we've combined the fixes and linter updates into a single PR (separate commits)

rtibbles:
In the interests of getting code more rapidly into develop, I merged #5790 without the corresponding fixes. I think it is reasonable to do the same here if this is good to go. The follow up PR can activate the linting and resolve the issues.

nucleogenesis:
I updated the error messages for both of the linting rules I added. With an approval it should be ready to merge. 

I will follow up with a PR applying fixes for the issues both rules found once merged.

rtibbles (approved):
Messaging comments addressed!

indirectlylit:
Looks like the tests are failing .. as intended :)

Issue #5786, opened 2019-07-17
check_debian_user is run unconditionally, even when running CLI via pip or PEX


jamalex:
### Observed behavior

Trying to run a kolibri command inside a virtualenv using a pip-installed kolibri version.

Because I separately also have the DEB installed, when [`check_debian_user`](https://github.com/learningequality/kolibri/pull/3934/files#diff-8f9ca278df35a8b43572c4b10dafe306R22) is called, it displays the ["You are running this command as the user" message](https://github.com/learningequality/kolibri/pull/3934/files#diff-acc9016e41b9367b7e8b63511df4786aR25) and requires user input in order to continue. As I'm running this in a headless mode, via the Python `sh` module, this blocks the script.

### Expected behavior

1. I would expect this message to only be displayed if I'm actually running the Debian-installed version of `kolibri`, not when I'm inside a virtualenv.
2. Additionally, when this message otherwise _would_ still be displayed, I would want to be able to override it with something like `--noinput`.

CC @lyw07

lyw07:
Thanks for reporting this issue! If we would like to display this image only when running the deb installer, I assume we will move the sanity check out of the kolibri source code but putting it into kolibri-installer-Debian source code?

jamalex:
Or perhaps it could live in Kolibri, but we could add in better mechanisms to detect what method is being used to call the CLI (or run the server). So the Debian CLI entry point (and system service) could set an envvar, as could the PEX CLI entry point, etc, before passing to the Kolibri CLI code. This would also be useful for our tracking of what methods people are using to install/run Kolibri, as right now we only report what OS it is.

lyw07:
yes that sounds better than putting the logic back to kolibri-installer-debian. I will try to see if there's a way for us to figure out how users install/run kolibri.

rtibbles:
First thing would be to check if this function solves the issue: https://github.com/learningequality/kolibri/blob/develop/kolibri/utils/server.py#L491

If not, would be good to fix the behaviour for this function if possible, and then use it to conditionally trigger the check debian user function.

lyw07:
thank you! I didn't realize that we have that function now. I will try using the function to see if it works in this case :)

lyw07:
I started to work on this but have some questions regarding the `--noinput` flag, and it would be great if anyone can clarify this:

[...]

PR #5859, opened 2019-08-07
Check kolibri user only when running Debian installer

lyw07:
### Summary

This PR checks the kolibri user only when user is running the Debian package. 
If the user runs as the user who is not the kolibri user defined in the Debian config, a user prompt will show up in the beginning to ask if the user would like to continue to run as the current user or not.
If the user decides not to run as the current user, KOLIBRI_HOME directory will be deleted (at this time, KOLIBRI_HOME directory should only contain `logs` folder and `kolibri_settings.json` file).

A flag `--no-input` is also added so that users can start kolibri with `kolibri start --no-input`, and the program will start without asking for user input.

### Reviewer guidance

1. Install the Debian package from this PR, configure the kolibri user to be `user A`
2. start kolibri with `user B` -- should be able to see the user prompt
3. start kolibri with `user B` and command `kolibri start --no-input` -- should not see the user prompt
3. get the pex file from this PR, run it with `user B` -- should not see the user prompt

### References

#5786 

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/utils/cli.py
+++ b/kolibri/utils/cli.py
@@ -128,12 +128,12 @@ def get_debug_param():
         return debug_option.default
 
 
-initialize_params = [
-    debug_option,
+base_params = [debug_option, noinput_option]
+
+initialize_params = base_params + [
     settings_option,
     pythonpath_option,
     skip_update_option,
-    noinput_option,
 ]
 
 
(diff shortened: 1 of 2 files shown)


kolibri/utils/cli.py
```diff
@@ -178,6 +182,9 @@ def __init__(self, *args, **kwargs):
         super(KolibriDjangoCommand, self).__init__(*args, **kwargs)
 
     def invoke(self, ctx):
+        # Check if the current user is the kolibri user when running kolibri from Debian installer.
```
rtibbles, line 188: I think we probably want to add this to both the `KolibriCommand` above and the `KolibriDjangoCommand`, as both presume the ability to access the `KOLIBRI_HOME` dir.

Can probably do a small refactor and have `base_params` and then `initialize_params` as additional to the `base_params` here.

lyw07: thank you! I will add it to KolibriCommand as well.

sorry I'm a bit confused about `base_params` you mentioned here. What is this for?

lyw07: for `base_params`, do you mean `debug_option` and `noinput_option`?

rtibbles: Yes - the two args that are used in both cases (I didn't bother defining a list of them before because there was only one).

lyw07: makes sense to me. thank you! I will create `base_params` for those two options.

rtibbles (commented):
Code looks right to me - @jamalex could you test to see if this fixes the issue you encountered?

rtibbles (approved):

PR #5861, opened 2019-08-08
Fix Sign Up API when non-default facility is selected

### LINKED ISSUE #5860: new accounts are seemingly created in the wrong facility
### Observed behavior

Accounts created in one facility seem to actually get created in another facility.

### Expected behavior

Facility should be correct. Might be related to #5812, so we should address that issue at the same time.

### User-facing consequences

Users cannot log in

### Errors and logs

…

### Steps to reproduce

* Navigate to https://mexico.unwomendemo.learningequality.org
* Select facility 'Nueva York'
* Go to the account creation page
* Re-select facility 'Nueva York' (#5812) and create a test account
* Log out
* Navigate back to https://mexico.unwomendemo.learningequality.org
* Select facility 'Nueva York'
* Try to log in - it doesn't work
* Reload page
* Select facility 'No pertenezco a ningún centro'
* Try to log in - it works

### Context

0.12.7

jonboiser:
### Summary

1. Fixes `SignUpViewset` to not ignore a non-default facility ID that's passed in
1. Make existing test for this case stronger, and refactor

**3: remove facility dropdown from Sign Up Page**

Picking a Facility is mandatory when landing on Sign Up / In page in a multiple facility server. (Cancel button is removed, and keyboard ESC doesn't work). Pop-up will show if user goes directly to Sign Up Page, does not if they get there via Sign In Page

Screen Shot 2019-08-08 at 2 33 45 PM

When facility is chosen, it is displayed as text

Screen Shot 2019-08-08 at 2 38 17 PM

### Reviewer guidance

1. You can create a multi-facility server by going to `/api/auth/facility` and creating a new facility. The SignIn/Up flow will now reflect #5860 better
1. Create Learner accounts for the different facilities. Check that the two users are indeed in different facilities with the admin accounts, or other means

### References

Fixes #5860 	
…

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated


kolibri/core/auth/test/test_api.py
```diff
@@ -612,72 +611,69 @@ def setUp(self):
...
+            {
+                "username": "bob",
+                "password": DUMMY_PASSWORD,
+                "facility": other_facility.id,
+            }
+        )
         self.assertEqual(response.status_code, status.HTTP_201_CREATED)
-        self.assertEqual(len(models.FacilityUser.objects.all()), 2)
+        self.assertEqual(
+            len(models.FacilityUser.objects.filter(facility=self.facility.id)), 1
```
ralphiee22: minor, but instead of `len` can do `models.FacilityUser.objects.filter(facility=self.facility.id).count()` here.

kolibri/plugins/user/assets/src/views/SignInPage/FacilityModal.vue
```diff
@@ -3,9 +3,7 @@
   <KModal
     :title="$tr('facilitySelectionModalHeader')"
     :submitText="$tr('submitFacilitySelectionButtonPrompt')"
-    :cancelText="$tr('close')"
```
ralphiee22, line 6: don't need the translated string below

kolibri/core/auth/test/test_api.py
ralphiee22: also should have a separate test case just indicating its testing passing in the `facility` param in the request

ralphiee22 (approved):
Minor notes, but tested change works as expected 👍

kolibri/plugins/user/assets/src/views/SignInPage/FacilityModal.vue
jonboiser: It’s being used somewhere else via cross component translator. This should be cleaned up on develop though

kolibri/core/auth/test/test_api.py
jonboiser: What do you mean? Just post one new user with a non-default facility?

ralphiee22: yes, and just put it in its own test, like `test_create_user_with_facility_param`.

PR #5835, opened 2019-07-30
Remove any Kolibri python dependency for building frontend assets

rtibbles:
### Summary
* Previous work has tried to make frontend asset building an exportable, independent library that can be seamlessly imported into external plugin projects without significant overhead.
* This has so far failed, still requiring the `kolibri` package to be pip installed, and even then that not really working most of the time.
* This PR resolves this by removing any dependency on `kolibri` itself when doing frontend asset building.
* It mostly achieves this by putting all build configuration into `webpack.config.js` inside the plugin folder.
* While most of the build time configuration is now stored in Javascript, getting the version of the plugin to use at build time, the file system path for the Python module path specified plugin, and the correct path for locale files is still most easily calculated using Python, so there is still some Python code doing work here.
* This also requires that the plugin (or Kolibri) is locally available for import, in order to infer version information - would like feedback on this approach.
* The WebpackBundleHook now just needs the `unique_slug` which is in common between the Python code and the `webpack.config.js`
* As further cleanup, this PR removes all of the unnecessary boilerplate, and defaults to using the `webpack_asset` template tag to load a specific SPA's main module into its page. We should add the sync hooks/async hooks back in as they are needed and to the extent that we want SPAs to be pluggable.
* Also updates a few kolibri-tools dependencies and makes vue-template-compiler a hard dependency again (but with a more flexible version requirement).

### Reviewer guidance
Does it build?
Does it build right?

----

### Contributor Checklist

PR process:

- [x] PR has the correct target branch and milestone
- [x] PR has 'needs review' or 'work-in-progress' label
- [x] If PR is ready for review, a reviewer has been added. (Don't use 'Assignees')
- [ ] If this is an important user-facing change, PR or related issue has a 'changelog' label
- [ ] If this includes an internal dependency change, a link to the diff is provided

Testing:

- [x] Contributor has fully tested the PR manually
- [ ] If there are any front-end changes, before/after screenshots are included
- [ ] Critical user journeys are covered by Gherkin stories
- [ ] Critical and brittle code paths are covered by unit tests

### Reviewer Checklist

- Automated test coverage is satisfactory
- PR is fully functional
- PR has been tested for [accessibility regressions](http://kolibri-dev.readthedocs.io/en/develop/manual_testing.html#accessibility-a11y-testing)
- External dependency files were updated if necessary (`yarn` and `pip`)
- Documentation is updated
- Contributor is in AUTHORS.md

PR #5864, opened 2019-08-10
Independent build

indirectlylit:
### Summary

migrated from #5835 because I accidentally closed that one and can't reopen

* Previous work has tried to make frontend asset building an exportable, independent library that can be seamlessly imported into external plugin projects without significant overhead.
* This has so far failed, still requiring the `kolibri` package to be pip installed, and even then that not really working most of the time.
* This PR resolves this by removing any dependency on `kolibri` itself when doing frontend asset building.
* It mostly achieves this by putting all build configuration into `webpack.config.js` inside the plugin folder.
* While most of the build time configuration is now stored in Javascript, getting the version of the plugin to use at build time, the file system path for the Python module path specified plugin, and the correct path for locale files is still most easily calculated using Python, so there is still some Python code doing work here.
* This also requires that the plugin (or Kolibri) is locally available for import, in order to infer version information - would like feedback on this approach.
* The WebpackBundleHook now just needs the `unique_slug` which is in common between the Python code and the `webpack.config.js`
* As further cleanup, this PR removes all of the unnecessary boilerplate, and defaults to using the `webpack_asset` template tag to load a specific SPA's main module into its page. We should add the sync hooks/async hooks back in as they are needed and to the extent that we want SPAs to be pluggable.
* Also updates a few kolibri-tools dependencies and makes vue-template-compiler a hard dependency again (but with a more flexible version requirement).

### Reviewer guidance
Does it build?
Does it build right?

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated


indirectlylit (commented):
to avoid abusing the webpack config schema, let's maybe rename the files to something like “pluginBuildConfig.js” and make the structure something like:

```js
module.exports = [
  {
    unique_plugin_id: 'default_frontend',
    webpack_config: {
      entry: './assets/src/core-app',
      output: {
        library: kolibriName,
      },      
    }
  },
  {
    unique_plugin_id: 'user_agent',
    webpack_config: {
      entry: './assets/src/userAgentCheck.js',
    }
  },
];
```


(Side note, I've always found the term 'slug' to be a bit jargony. if possible, would love to just call it an "id")

The other item was handling `__pycache__`, empty dirs, etc and testing on a few versions of python due to package enumeration weirdness.

other than that, this looks awesome! love all the deleted files. great job

rtibbles (approved):

indirectlylit:
sweet

PR #5873, opened 2019-08-13
Update CHANGELOG and VERSION for 0.12.8 

jonboiser:
### Summary

Updates VERSION and CHANGELOG for 0.12.8 **FINAL** release (So don't merge it until we're at that point).

- [x] Correct branch

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,16 @@
 
 List of the most important changes for each release.
 
+## 0.12.8
+
+### Changed or fixed
+
+- Fixed an issue where users creating accounts for themselves would be placed in the wrong facility
+- Fixed an issue where, occasionally, images would be missing from Khan Academy exercises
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,16 @@
 
 List of the most important changes for each release.
 
+## 0.12.8
+
+### Changed or fixed
+
+- Fixed an issue where users creating accounts for themselves would be placed in the wrong facility
+- Fixed an issue where, occasionally, images would be missing from Khan Academy exercises
+- Fixed an issue where the "Usage and Privacy" modal would not close when clicking the "Close" button
```
indirectlylit: would be nice to avoid repeating the phrase "Fixed an issue where", mostly for style reasons.

could do something like

- Fix account creation when multiple facilities are on a device
- Prevent images from being occasionally missing in Khan Academy exercises
- Fixed a regression related to the "Close" button in the "Usage and Privacy" modal

PR #5886, opened 2019-08-19
Fix i18n linting errors

nucleogenesis:
### Summary

This PR fixes the linting errors discovered by the linters created in https://github.com/learningequality/kolibri/pull/5854 and https://github.com/learningequality/kolibri/pull/5790

A few files were defined in a way that would not be found by the linter and the rule has been disabled specifically for the blocks of code affected or for the whole file as needed.

Additionally, the plugins are now enabled in `kolibri-tools/.eslintrc`.

Additional _bugfix_ thanks to @jonboiser - previously when a defined string matched it's key exactly, the linter acted as if it were a "use" - and thus would not show the error. This has been resolved.

### Reviewer guidance

Please review the code where strings have been removed and ensure that there is no place in that file where the string is used.

*Note - the `crossComponentTranslator()` function is no longer in use, so there is no concern that another component will try to use any of the strings that were removed.*

Additionally, if you have eslint in your editor of choice, try testing that these work by defining and not using a string or vice versa, use a string like `$tr('undefinedString')`.

### References

https://github.com/learningequality/kolibri/pull/5790
https://github.com/learningequality/kolibri/pull/5854

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated


kolibri/plugins/user/assets/src/views/UserProfileSideNavEntry.vue
```diff
@@ -1,5 +1,5 @@
 <template>
-
+  <!-- eslint-disable kolibri/vue-no-undefined-string-uses -->
```
jonboiser: what's wrong with this file?

kolibri/plugins/coach/assets/src/views/plan/CreateExamPage/index.vue
```diff
@@ -494,9 +494,6 @@
         'The max number of questions based on the exercises you selected is {maxQuestionsFromSelection}. Select more exercises to reach {inputNumQuestions} questions, or lower the number of questions to {maxQuestionsFromSelection}.',
       noneSelected: 'No exercises are selected',
       // TODO: Interpolate strings correctly
-      added: "Added '{item}'",
```
indirectlylit, line 497: let's delete the comment above, too:

```js
// TODO: Interpolate strings correctly
```

kolibri/plugins/device_management/assets/src/views/DeviceManagementSideNavEntry.vue
```diff
@@ -1,5 +1,5 @@
 <template>
-
+  <!-- eslint-disable kolibri/vue-no-undefined-string-uses -->
```
indirectlylit: could we also add a comment indicating why this is necessary?

kolibri/plugins/user/assets/src/views/SignInPage/index.vue
```diff
@@ -493,18 +493,10 @@
     },
     $trs: {
       createAccountAction: 'Create an account',
-      kolibri: 'Kolibri',
       poweredByKolibri: 'Powered by Kolibri',
       whatsThis: "What's this?",
       oidcGenericExplanation:
         'Kolibri is an e-learning platform. You can also use your Kolibri account to log in to some third-party applications.',
-      oidcSpecificExplanation:
-        "You were sent here from the application '{app_name}'. Kolibri is an e-learning platform, and you can also use your Kolibri account to access '{app_name}'.",
```
indirectlylit, line 502: let's leave this string in for now in case we need it eventually

kolibri/plugins/user/assets/src/views/UserProfileSideNavEntry.vue
indirectlylit: here too, recommend leaving a comment why this is necessary

indirectlylit (commented):
cool!

kolibri/plugins/user/assets/src/views/UserProfileSideNavEntry.vue
nucleogenesis: The linter wasn't pulling $trs defined on a couple of components that were defined as variables and then exported later on - it was only finding the definitions in components defined as part of the `export default` declaration.

I've refactored it to fix this and have removed these comments.

cc @indirectlylit @jonboiser

kolibri/plugins/user/assets/src/views/SignInPage/index.vue
nucleogenesis: Not sure why it doesn't say this is outdated, but I've added it back along with a comment disabling the rule for that line only and a comment noting that we are keeping the unused string in case we use it in the future. @indirectlylit

kolibri/plugins/coach/assets/src/views/plan/CreateExamPage/index.vue
nucleogenesis: Again - I'm not sure why this isn't `outdated` but I've removed this comment in a recent commit. @indirectlylit

jonboiser:
If crossComponentTranslator is no longer used, can we delete it, so we're not tempted to use it?

If we need to use it as a hack, we can still kind of re-implement it as-needed using the Translator class.

jonboiser (approved):

indirectlylit:
> If crossComponentTranslator is no longer used, can we delete it, so we're not tempted to use it?
> 
> If we need to use it as a hack, we can still kind of re-implement it as-needed using the Translator class.

One disadvantage of re-implementing it as necessary using the `Translator` class is that it would make "cleanup" a bit harder. The nice thing about `crossComponentTranslator` is it's easy to search for.

indirectlylit:
maybe we should add a `logging.warning` to it? e.g.

```js
export function crossComponentTranslator(Component) {
  if (process.env.NODE_ENV !== 'production') {
    logging.warning('TODO: remove crossComponentTranslator');
  }
  return new Translator(Component.name, Component.$trs);
}
```

PR #5888, opened 2019-08-20
Bootstrap 0.13 changelog

benjaoming:
### Summary

There was a PR #5494 that fixed a perhaps serious bug. Wanted to check if the "changelog" label applied to that PR would capture the change, but I rather don't think it will, unless we start tracking changes in CHANGELOG.md

### Reviewer guidance

Is the text concise?

### References

https://github.com/learningequality/kolibri-server/issues/20

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 List of the most important changes for each release.
 
+## 0.13.0 (unreleased)
+
+### Changed or fixed
+
+- Fixed: Unclean shutdowns on very large databases, due to prolonged database cleanup
+
 ## 0.12.8
 
 ### Changed or fixed


indirectlylit:
the text looks fine to me. I'm a little confused though: is this change actually associated with Kolibri `0.13.0`? Or will it also affect the `0.12.x` versions because it's in the `kolibri-server` project?

indirectlylit:
ref: https://github.com/learningequality/kolibri-server/releases

It looks like we've been tagging releases, but not posting the changes as "github releases":

image

jredrejo:
> the text looks fine to me. I'm a little confused though: is this change actually associated with Kolibri `0.13.0`? Or will it also affect the `0.12.x` versions because it's in the `kolibri-server` project?

No, kolibri-server can't fix the database lock problem, it's a fix that will be applied when kolibri updates to 0.13.0. It's just the issue was reported on kolibri-server and investigation took me to see the problem was in 0.12.x but was solved in current develop.

jredrejo:
> ref: https://github.com/learningequality/kolibri-server/releases
> 
> It looks like we've been tagging releases, but not posting the changes as "github releases":
> 
@indirectlylit  I did try to mimic the way kolibri did releases but I didn't spend time to finish the whole process, would you mind guiding me to know what are all the needed steps to do it correctly?

jredrejo (approved):
I'd say the wording is correct. The problem is a bit deeper: there is a task that was done whenever kolibri started and now it's only done when there is a database migration. That task took a long time and left the database locked so vacuum didn't start.
I don't think all those details need to be added to the changelog, just wanted to share them with you.

Btw: the problem will persist when upgrading from different versions of kolibri. It's going to happen only once. After the upgrade is finished , next starts won't have it. I wonder if it's worth spending time in kolibri trying to ensure the db upgrade closes big databases correctly, but that's a different story.

intelliant01:
> Btw: the problem will persist when upgrading from different versions of kolibri. It's going to happen only once. After the upgrade is finished , next starts won't have it. I wonder if it's worth spending time in kolibri trying to ensure the db upgrade closes big databases correctly, but that's a different story.

Just to record an observation - I have seen the database lock issue with a
large db on starts even after a version upgrade.
I shall specially watch out for this when testing with the next release.

benjaoming:
@indirectlylit @jredrejo I don't think it's necessary to have Github releases for `kolibri-server` since technically it's not released there. The git tags would seem sufficient?

indirectlylit:
> @indirectlylit I did try to mimic the way kolibri did releases but I didn't spend time to finish the whole process, would you mind guiding me to know what are all the needed steps to do it correctly?

Sure - it's super easy. When you [draft a release using the Github UI](https://help.github.com/en/articles/creating-releases), it will automatically create the git tag for you. Then you can drop the (markdown-formatted) changes into the description.

> @indirectlylit @jredrejo I don't think it's necessary to have Github releases for kolibri-server since technically it's not released there. The git tags would seem sufficient?

None of our products are actually released through Github.

We use Github releases to associate summaries of a range of changes with specific git tags. Using a tag alone doesn't provide enough context.

See e.g.:

* https://github.com/learningequality/kolibri-installer-windows/releases
* https://github.com/learningequality/kolibri-installer-android/releases
* https://github.com/learningequality/kolibri/releases

We should be doing this for the Debian package also.

indirectlylit (approved):

PR #5887, opened 2019-08-20
Plugin data bootstrapping and module namespacing

rtibbles:
### Summary
* Refactors bootstrapping template tags that bring Python code into the frontend to use a standard `plugin_data` API.
* Makes various dependency updates for polyfills.
* Namespaces frontend module bundles by the kolibri plugins they are a part of

### Reviewer guidance
Does everything still work as intended? Is the code clear and comprehensible?

https://github.com/learningequality/kolibri-exercise-perseus-plugin/compare/v1.2.1...v1.2.2a1

----

### Contributor Checklist

PR process:

- [x] PR has the correct target branch and milestone
- [x] PR has 'needs review' or 'work-in-progress' label
- [x] If PR is ready for review, a reviewer has been added. (Don't use 'Assignees')
- [ ] If this is an important user-facing change, PR or related issue has a 'changelog' label
- [ ] If this includes an internal dependency change, a link to the diff is provided

Testing:

- [x] Contributor has fully tested the PR manually
- [ ] If there are any front-end changes, before/after screenshots are included
- [ ] Critical user journeys are covered by Gherkin stories
- [x] Critical and brittle code paths are covered by unit tests

### Reviewer Checklist

- Automated test coverage is satisfactory
- PR is fully functional
- PR has been tested for [accessibility regressions](http://kolibri-dev.readthedocs.io/en/develop/manual_testing.html#accessibility-a11y-testing)
- External dependency files were updated if necessary (`yarn` and `pip`)
- Documentation is updated
- Contributor is in AUTHORS.md

PR #5896, opened 2019-08-22
Content renderer refactor

rtibbles:
### Summary
* Changes content renderers to use `presets` instead of `kind/extension` combinations for compatibility/consistency with Kolibri Studio.
* Extracts the `ContentRenderer` component and associated logic into `KContentRenderer` and puts it into the `kolibri-components` package.
* Collapses the content renderer API so that it is entirely defined by the content renderer mixin - the parent component and all the subcomponents now share the same API.
* Allows content renderers to take an `itemData` prop to directly load JSON content (useful for live preview of slideshows and perseus exercises).
* Does some final pluginData cleanup by removing the `kolibri_set_urls` tag and associating it with the core frontend asset instead.
* In the course of moving the plugin mediator into `kolibri-components` (required in order to handle loading of content renderers in e.g. Kolibri Studio), simplifies and flattens the dependency hierarchy for `index.js` to make it clearer how the core app is constructed.
* Removes `initSession` callback from the Content Renderer lifecycle - simplifies this by only instantiating the content renderer when the session has already been initialized.

### Reviewer guidance
* Does everything still work as intended?
* Does this actually simplify things?
* Is the content renderer API (as defined by the props in the mixin) too complicated?

### References
Dependent on #5887
Updates the perseus renderer for these changes:
https://github.com/learningequality/kolibri-exercise-perseus-plugin/compare/v1.2.2a1...v1.2.2a2

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated


kolibri/core/assets/src/core-app/apiSpec.js
```diff
@@ -156,7 +157,8 @@ export default {
     },
     router,
     mixins: {
-      contentRendererMixin,
+      responsiveWindowMixin,
+      responsiveElementMixin,
```
indirectlylit, line 161: My initial thought is that maybe the import paths should be more like `kolibri.components.*`, which would be an alias for `kolibri-components/src/*`. For example:

```js
import KResponsiveElementMixin from 'kolibri.components.KResponsiveElementMixin';
```

(I'm also thinking maying I should just flatten out the `/src` directory from `kolibri-components` since we've decided not to provide a built version...)

kolibri/core/assets/src/core-app/index.js
```diff
@@ -1,58 +1,100 @@
...
-require('../styles/main.scss');
 
 // monitor input modality
-const trackInputModality = require('kolibri-components/src/styles/trackInputModality').default;
-
 trackInputModality();
 
 // monitor page visibility
-require('./monitorPageVisibility');
+monitorPageVisibility();
```
indirectlylit: Perhaps just inline the logic to reduce indirection?

```js
document.addEventListener('visibilitychange', () => { store.dispatch('setPageVisibility') });
```

kolibri/core/assets/src/core-app/index.js
```diff
@@ -1,58 +1,100 @@
...
+import KThemePlugin from 'kolibri-components/src/KThemePlugin';
+import heartbeat from 'kolibri.heartbeat';
+import KContentPlugin from 'kolibri-components/src/content/KContentPlugin';
+import keenOptions from '../keen-config/options.json';
+import { i18nSetup, languageDirection } from '../utils/i18n';
+import ContentRendererLoadingComponent from '../views/ContentRenderer/ContentRendererLoading';
+import ContentRendererErrorComponent from '../views/ContentRenderer/ContentRendererError';
+import apiSpec from './apiSpec';
+import monitorPageVisibility from './monitorPageVisibility';
+import getPluginData from 'kolibri.utils.getPluginData';
```
indirectlylit: I find it confusing that we mix api spec and normal path style imports. It's not clear to me when to use one versus the other. I feel like we should standardize if possible

packages/kolibri-components/src/content/KContentPlugin.js
```diff
@@ -0,0 +1,44 @@
...
+    {
+      languageDirection,
+      ContentRendererLoadingComponent = {
+        render(h) {
+          return h('p', '-----');
+        },
+      },
+      ContentRendererErrorComponent = {
+        render(h) {
+          return h('p', 'xxxxx');
```
indirectlylit: what are these? could we include some comments

packages/kolibri-components/src/content/KContentPlugin.js
```diff
@@ -0,0 +1,44 @@
...
+        },
+      },
+      facade,
```
indirectlylit: This could probably just be called `coreAppGlobal` or something?

If I'm understanding correctly, the fact that the API is implemented using a 'facade' pattern is an internal implementation detail of the core app which should not be exposed to the API users.

kolibri/core/kolibri_plugin.py
```diff
@@ -2,14 +2,160 @@
...
+            # so this was the easiest way to inject it.
+        ).replace("__placeholder__", js_name)
+        return [
+            mark_safe(
+                """<script type="text/javascript">"""
+                + jsmin(js)
+                + """
+            {js_name}.__staticUrl = '{static_url}';
+            {js_name}.__mediaUrl = '{media_url}';
+            {js_name}.__contentUrl = '{content_url}';
```
indirectlylit, line 60: I'm a little confused about the relationship between this and pluginData().urls

kolibri/core/assets/src/core-app/urls.js
```diff
@@ -4,12 +4,17 @@
...
+  const urlObject = new URL(url, base);
+  return urlObject.href;
+}
 
 const urls = {
   setUp() {
-    Object.assign(this, global.kolibriUrls);
-    this.__staticURL = global.staticUrl;
-    this.__mediaURL = global.mediaUrl;
+    Object.assign(this, getPluginData().urls);
```
indirectlylit: This is a bit confusing because it's not clear what's actually being added to `this`.

kolibri/core/assets/src/core-app/urls.js
```diff
@@ -28,6 +33,13 @@ const urls = {
     const urlObject = new URL(url, base);
     return urlObject.href;
   },
+  storageUrl(fileId, extension, embeddedFilePath = '') {
+    const filename = `${fileId}.${extension}`;
+    if (['perseus', 'zip', 'h5p'].includes(extension)) {
+      return this['kolibri:core:zipcontent'](filename, embeddedFilePath);
+    }
+    return generateUrl(this.__contentURL, `${filename[0]}/${filename[1]}/${filename}`);
```
indirectlylit: might just reference `getPluginData().urls.__contentURL` directly? Copying it to `this` seems unnecessary

kolibri/core/assets/src/utils/branding.js
```diff
@@ -0,0 +1,43 @@
...
+  },
+  sideNav: {
+    topLogo: {
+      src: null,
+      alt: null,
+      style: null,
+    },
+    showKolibriFooterLogo: null,
+  },
+};
```
indirectlylit: cool

kolibri/core/assets/src/views/ContentRenderer/ContentRendererLoading.vue
```diff
@@ -0,0 +1,16 @@
...
+</template>
+
+
+<script>
+
+  export default {
+    name: 'ContentRendererLoading',
+  };
+
+</script>
```
indirectlylit: couldn't `ContentRendererError` and `ContentRendererLoading` just be private sub-components of `KContentRenderer`?

Also, `ContentRendererLoading` seems maybe unnecessary since it doesn't add anything to `KCircularLoader`?

kolibri/core/assets/src/views/language-switcher/mixin.js
```diff
@@ -1,11 +1,12 @@
 import { redirectBrowser } from 'kolibri.utils.browser';
 import { availableLanguages, currentLanguage } from 'kolibri.utils.i18n';
 import { httpClient } from 'kolibri.client';
+import urls from 'kolibri.urls';
 
 export default {
   methods: {
     switchLanguage(code) {
-      const path = this.Kolibri.urls['kolibri:core:set_language']();
+      const path = urls['kolibri:core:set_language']();
```
indirectlylit, line 9: could this be accessed through plugin data instead of aliasing through api spec?

kolibri/core/content/hooks.py
```diff
@@ -101,11 +39,11 @@ def render_to_page_load_async_html(self):
         # text direction than the interface due to the text direction of content
         urls = [chunk["url"] for chunk in self.bundle]
         tags = self.frontend_message_tag() + [
-            '<script>{kolibri_name}.registerContentRenderer("{bundle}", ["{urls}"], {content_types});</script>'.format(
-                kolibri_name=KOLIBRI_CORE_JS_NAME,
-                bundle=self.bundle_id,
+            '<script>{kolibri_name}.registerContentRenderer("{bundle}", ["{urls}"], {presets});</script>'.format(
+                kolibri_name="kolibriCoreAppGlobal",
```
indirectlylit, line 43: :)

kolibri/plugins/learn/assets/src/views/AssessmentWrapper/index.vue
```diff
@@ -456,15 +456,6 @@ oriented data synchronization.
...
       },
-      sessionInitialized() {
-        if (this.isUserLoggedIn) {
-          this.callInitMasteryLog();
-        } else {
-          this.createDummyMasteryLog();
-        }
-        this.nextQuestion();
-        this.$emit('sessionInitialized');
-      },
```
indirectlylit, line 467: were these drive-by changes or possibly a mistake? Seems unrelated to the rest of the PR

indirectlylit (commented):
looking good!

One of my main take-aways is that the new `getPluginData()` function is a bit confusing, because it's not super clear what the schema looks like or where the data gets set.

It also seems that the majority of cases in here actually refer to "core" data, not "plugin" data. Presumably a plugin can both access core's data and its own data? Was there an example of the latter which I missed?

kolibri/core/assets/src/core-app/apiSpec.js
rtibbles: Happy to structure it in whatever way seems best.

kolibri/core/assets/src/core-app/index.js
rtibbles: Yes.

kolibri/core/assets/src/core-app/index.js
rtibbles: Generally I use API spec style imports for anything exposed on the core API.

`getPluginData` is a little weird, because I set it up to be referenced using the core API path, but it is actually imported at build time (because it needs build time information to properly function).

packages/kolibri-components/src/content/KContentPlugin.js
rtibbles: Sure - these are just dummy error components.

packages/kolibri-components/src/content/KContentPlugin.js
rtibbles: Fair enough - also, this is only provided for use inside Kolibri, as the `undefined` behaviour below for the `facade` argument is required when using this outside of Kolibri.

kolibri/core/kolibri_plugin.py
rtibbles: These lines are adding additional urls to the exported URL object that is generated by the `django-js-reverse` code that we are leveraging above - it doesn't export a static, media, or content url, so we add those in specifically.

kolibri/core/assets/src/core-app/urls.js
rtibbles: Mostly the `django-js-reverse` url definitions, will add comments.

kolibri/core/assets/src/core-app/urls.js
rtibbles: Fair.

rtibbles: It just came long for free with the other Urls during the `Object.assign` which does need to happen to expose the functions directly on the `urls` export.

kolibri/core/assets/src/utils/branding.js
rtibbles: I feel like I have a better idea of what the API for this now is too!

kolibri/core/assets/src/views/ContentRenderer/ContentRendererLoading.vue
rtibbles: I have them separately to be able to use them as part of the async definition pattern https://vuejs.org/v2/guide/components-dynamic-async.html#Handling-Loading-State - they are used outside of  `KContentRenderer` to render the loading state while renderer is loaded asynchronously and to show the error state if it fails to load.

I could probably define the loading component inline, as it is only used in the loading state handling, but I needed to set the props on the `KCircularLoader` - I could have used Vue.extend to extend the `KCircularLoader` component and set the default prop value to what I wanted here, but this seemed more explicit.

The final reason that the loading and error components are injected, rather than being part of the exported content renderer is to allow for an alternative (e.g. vuetify) loader to be used in place in Studio, or in the case of the error component for a different wrapper component (not UiAlert, which I believe we don't export) and different interface text.

kolibri/core/assets/src/views/language-switcher/mixin.js
rtibbles: It could - but `pluginData` is scoped to the specific frontend bundle you are in, so it would only be accessible in this way inside the default frontend bundle. It seems easier and more consistent to continue to use the `kolibri.urls` import throughout Kolibri.

kolibri/plugins/learn/assets/src/views/AssessmentWrapper/index.vue
rtibbles: These were intentional - part of the changes I made to the ContentRenderer was to remove the session initialization part, and just make creating the component itself contingent on that happening (as it was almost entirely a logging/tracking concern).

kolibri/core/assets/src/views/ContentRenderer/ContentRendererLoading.vue
indirectlylit: I'm still not clear on why they live in `core` rather than `kolibri-components` though? I would have expected something like

```
src/
  ...
  KContentRenderer/
    index.js
    ContentRendererError.vue
    ContentRendererLoading.vue
  ...
```

> The final reason ... is to allow for an alternative (e.g. vuetify) loader to be used in place in Studio, or in the case of the error component for a different wrapper component (not UiAlert, which I believe we don't export) and different interface text.

In theory at least our K loaders and the V loaders should look interchangeable, and we'll eventually expose a KUIAlert. In practice .. not quite there yet

rtibbles: Right - because they were not both directly exportable, I just decided to go for the injection instead, can also add them as the 'default' but seems odd to add a default that does not work out of the box (in the case of the error component).

kolibri/core/assets/src/core-app/apiSpec.js
rtibbles: Have not changed in the scope of this PR - let's revisit when we look at the overall core API spec.

nucleogenesis:
@rtibbles 

Note - all errors found in Chromium on Debian

### Videos
Are fine.

### HTML5 Apps

HTML5 apps seem to work just fine - although the HTML5 app shows a console error that doesn't appear to break anything referring to a missing prop `defaultFile` - then an error saying `defualtFile` is already defined as a prop. These errors occur in that order one after the other.

Additionally this error follows those two:

```
Refused to load the script 'https://apis.google.com/js/platform.js' because it violates the following Content Security Policy directive: "default-src 'self' 'unsafe-inline' 'unsafe-eval' data: blob: http://localhost:8000". Note that 'script-src-elem' was not explicitly set, so 'default-src' is used as a fallback.
```

### Slideshow Issue

When testing a slideshow I'm getting the following error and the content doesn't load:

```
[Vue warn]: Invalid prop: custom validator check failed for prop "icon".

found in

---> <KIcon> at packages/kolibri-components/src/KIcon.vue
       <KLabeledIcon> at packages/kolibri-components/src/KLabeledIcon.vue
         <KFixedGridItem> at packages/kolibri-components/src/grids/KFixedGridItem.vue
           <KGridItem> at packages/kolibri-components/src/grids/KGridItem.vue
             <KFixedGrid> at packages/kolibri-components/src/grids/KFixedGrid.vue
               <KGrid> at packages/kolibri-components/src/grids/KGrid.vue
                 <PageHeader> at kolibri/plugins/learn/assets/src/views/PageHeader.vue
                   <KPageContainer> at packages/kolibri-components/src/KPageContainer.vue
                     <ContentPage> at kolibri/plugins/learn/assets/src/views/ContentPage.vue
                       <CoreBase> at kolibri/core/assets/src/views/CoreBase/index.vue
                         <Root>
```

### PDF & Epub Issue

I'm not able to view ePubs or PDFs either. It spins for a while then eventually shows that Kolibri cannot render this content. I tried an ePub from OpenStax and I tried loading a PDF that I just imported from TESSA via Studio as well as the PDF in my Learning Equality Slideshow Demo channel. It appears to be timing out while trying to load the content.

```
[Vue warn]: Failed to resolve async component: function () { return ({
            /* Check the Kolibri core app for a content renderer module that is able to
             * handle the rendering of the current content node.
             */
            component: this$1.retrieveContentRenderer(preset),
            // A component to use while the async component is loading
            loading: Vue.options.components['ContentRendererLoading'],
            // A component to use if the load fails
            error: Vue.options.components['ContentRendererError'],
            // Delay before showing the loading component.
            delay: 0,
            // The error component will be displayed if a timeout is
            // provided and exceeded.
            timeout: 30000,
          }); }
Reason: timeout (30000ms)
```

kolibri/core/assets/src/core-app/index.js
```diff
@@ -12,19 +16,45 @@ import trackInputModality from 'kolibri-components/src/styles/trackInputModality
...
+import ContentRendererErrorComponent from '../views/ContentRenderer/ContentRendererError';
+import apiSpec from './apiSpec';
+import plugin_data from 'plugin_data';
 // Do this before any async imports to ensure that public paths
 // are set correctly
 urls.setUp();
 
+// Shim window.location.origin for IE.
+if (!window.location.origin) {
+  window.location.origin = `${window.location.protocol}//${window.location.hostname}${
```
DXCanas, line 38: 👌

rtibbles:
> Additionally this error follows those two:
> `Refused to load the script 'https://apis.google.com/js/platform.js' because it violates the following Content Security Policy directive: "default-src 'self' 'unsafe-inline' 'unsafe-eval' data: blob: http://localhost:8000". Note that 'script-src-elem' was not explicitly set, so 'default-src' is used as a fallback.`

This error should appear, so that's fine!

packages/kolibri-components/src/content/KContentRenderer.js
```diff
@@ -0,0 +1,45 @@
...
+      this.registerContentActivity();
+      return null;
+    },
+  },
+  render: function(createElement) {
+    if (this.canRenderContent(this.defaultItemPreset)) {
+      const listeners = {
+        ...this.$listeners,
+      };
+      contentRendererMixin.interactionEvents.forEach(event => {
```
DXCanas, line 29: [Brings me back.](https://github.com/DXCanas/studio/blob/uppy-integration/contentcuration/contentcuration/static/js/edit_channel/FileUpload/index.vue#L207)

My understanding of the [Vue documentation for this](https://vuejs.org/v2/api/#vm-listeners) was that it could be either a function or an array. I'm guessing it's probably for if you have multiple functions assigned to the same event?


Anyway, the [`$on` documentation](https://vuejs.org/v2/api/#vm-on) (which I'm assuming will eventually be used? Meh?) doesn't seem to cover having an array passed in as the second argument. Worth a look?

packages/kolibri-components/src/content/mixin.js
```diff
@@ -0,0 +1,167 @@
...
+      type: Object,
+      default: null,
+      validator: fileValidator,
+    },
+    // As an alternative to passing a file object to set the state of the
+    // content renderer, can also pass raw itemData (which will be parsed by
+    // the renderer if there are no files or file object).
+    // The type could depend on the renderer, so we enforce nothing here
+    // except a null default.
+    itemData: {
```
DXCanas, line 89: 🧐

> Allows content renderers to take an itemData prop to directly load JSON content (useful for live preview of slideshows and perseus exercises).


Is my understanding that this is only taken advantage of by the slideshow component atm correct?

packages/kolibri-components/src/content/KContentRenderer.js
```diff
@@ -0,0 +1,45 @@
+import { RENDERER_SUFFIX } from './constants';
+import contentRendererMixin from './mixin';
```
DXCanas, line 2: > Collapses the content renderer API so that it is entirely defined by the content renderer mixin - the parent component and all the subcomponents now share the same API.

I think I might be missing something - this is the only place I'm seeing it imported now. How are subcomponents using it?

kolibri/plugins/learn/assets/src/views/AssessmentWrapper/index.vue
```diff
@@ -168,10 +164,6 @@ oriented data synchronization.
         type: Object,
         default: () => {},
       },
-      initSession: {
```
DXCanas, line 171: 😮

packages/kolibri-components/src/content/mixin.js
```diff
@@ -0,0 +1,167 @@
...
+      validator: languageValidator,
+    },
+    showCorrectAnswer: {
+      type: Boolean,
+      default: false,
+    },
+  },
+  computed: {
+    defaultItemPreset() {
+      return this.defaultFile
```
DXCanas, line 132: 🎉

DXCanas (changes requested):
> Changes content renderers to use presets instead of kind/extension combinations for compatibility/consistency with Kolibri Studio.

While I'm celebrating this interoperability, I'm wondering if we should (later) revisit our choice of words (across projects) here? I think it may be the reason the word `kind` was chosen for Kolibri in the first place?

But otherwise: Looks awesome! Not entirely sure about the one change I flagged. Please double check me!

packages/kolibri-components/src/content/mixin.js
rtibbles: And perseus.

packages/kolibri-components/src/content/KContentRenderer.js
rtibbles: It also gets imported in the mediator, and then gets injected into the component definition.

packages/kolibri-components/src/content/KContentRenderer.js
rtibbles: I think this is permissible, because this told me so: https://forum.vuejs.org/t/render-functions-multiple-event-handlers-on-same-event-solved/13755

rtibbles:
@nucleogenesis I replicated every error except the epub rendering issues, which seemed fine when I used epubs from Global Digital Library. As you said it wasn't working for epubs on Open Stax, which uses PDFs, I am assuming that it was just another instance of the PDF issue.

All rendering issues should now be resolved.

packages/kolibri-components/src/content/KContentRenderer.js
rtibbles: I think this:
https://github.com/vuejs/vue/blob/52719ccab8fccffbdf497b96d3731dc86f04c1ce/flow/component.js#L35
and this:
https://github.com/vuejs/vue/blob/df4af4bd1906b9f23b62816142fdfbd6336d3d2f/types/options.d.ts#L142

Are sufficient for me to feel sure about this - basically that the `$listeners` object (which is what the `on` value gets set to, is an object with a string key, and then a function or array of functions as the value.

Your comment did make me realize that the code above does not handle the situation in which the value on the `$listeners` object is already an array, so I have updated the code accordingly.

DXCanas (approved):

PR #5919, opened 2019-09-04
Default to zero-width without slot or label

bjester:
### Summary
The `KLabeledIcon` component has been updated to default to the zero-width character `&#8203;` when the label would be otherwise empty. This seems to fix the rendering of the HTML tag such that its height is consistent with its height when it has text.

#### Before
screenshot-localhost-8080-2019 09 04-10-05-01

#### After
screenshot-localhost-8080-2019 09 04-10-07-19

### Reviewer guidance
The easiest way to reproduce and test this is to use the [user CSV import functionality](https://kolibri.readthedocs.io/en/latest/advanced.html#import-users-from-a-csv-file), importing a user with an empty name.

### References
https://github.com/learningequality/kolibri/issues/5678

----

### Contributor Checklist

PR process:

- [X] If this includes an internal dependency change, a link to the diff is provided

Testing:

- [X] Critical user journeys are covered by Gherkin stories

### Reviewer Checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/packages/kolibri-components/src/KLabeledIcon.vue
+++ b/packages/kolibri-components/src/KLabeledIcon.vue
@@ -9,9 +9,9 @@
     <div class="label">
       <!-- nest slot inside span to get alignment and flow correct for mixed RLT/LTR -->
       <span dir="auto">
-        <slot>
-          {{ label }}
-        </slot>
+        <!-- Use zero-width space when empty -->
+        <slot v-if="!labelEmpty">{{ label }}</slot>
+        <template v-else>&#8203;</template>
       </span>
     </div>
   </span>
@@ -38,6 +38,16 @@
       label: {
         type: String,
         required: false,
+      },
+    },
+    computed: {
+      labelEmpty() {
+        const defaultSlot =
+          'default' in this.$slots && this.$slots.default.length
+            ? this.$slots.default[0].text
+            : null;
+
+        return !defaultSlot && !this.label;
       },
     },
   };
--- a/packages/kolibri-components/src/KLabeledIcon.vue
+++ b/packages/kolibri-components/src/KLabeledIcon.vue
@@ -42,12 +42,20 @@
     },
     computed: {
       labelEmpty() {
-        const defaultSlot =
-          'default' in this.$slots && this.$slots.default.length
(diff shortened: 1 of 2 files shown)


bjester:
@jonboiser I removed the "needs review" label for now. I thought there was just weird flaky issue in the build, but seems legit so will try to fix.

jonboiser (approved):
Looks good to me!

jonboiser:
Looks good to me, although I think the full name-less user might be the only case this happens. We should probably not allow this to happen in the CSV import command.

packages/kolibri-components/src/KLabeledIcon.vue
```diff
@@ -40,6 +40,24 @@
...
+        }
+
+        const defaultSlot = this.$slots.default[0];
+        return !(
+          defaultSlot.text ||
+          defaultSlot.tag ||
+          (defaultSlot.children && defaultSlot.children.length)
+        );
+      },
+    },
```
indirectlylit, line 60: I think this implementation could have been a bit simpler. Something like:

template:

```html
<slot>{{ computedLabel }}</slot>
```

computed prop:

```js
computedLabel() {
  const zeroWidth = "\u200b";
  return this.label ? this.label : zeroWidth;
},
```

bjester: Unfortunately I don't think that works for *empty* passed into the slot. I'm pretty sure I tried it. Those two different methods, label vs slot, made this more tricky than I thought.

jonboiser: I don't think there's any usage of KLabledIcon with nothing in the default slot. I think if somebody tries to use this component without a label prop or anything in the slot, we could log a warning since it's probably a mistaken usage.

bjester: Well, for clarification, if `someVar` in my example below turns out to be empty, Devon's suggestion would not end up working.
```
<KLabeledIcon>{{someVar}}</KLabeledIcon>
```
So I wanted this to be consist with an empty variable was passed to the `label` prop.

PR #5924, opened 2019-09-05
Use the specific kolibri host for access control origin header.

rtibbles:
### Summary
Due to a bug in Microsoft Edge, resource loading inside some HTML5 apps is currently broken because of the way that we set `Access-Control-Allow-Origin` headers to a wild card. It appears that this bug is still extant after several years: https://thisinterestsme.com/microsoft-edge-cors-bug/

This PR fixes this issue by setting a specific `Access-Control-Allow-Origin` header - to that of the current Kolibri host. This is good because HTML5 apps should only be loading things from Kolibri host anyway!

### Reviewer guidance
Do Funza apps work in Microsoft Edge?
Do HTML5 apps work properly with this and Kolibri-server?

I have manually tested this behind an nginx proxy and it worked fine, I have tested it on Microsoft Edge, but not specifically with Funza.

### References
cc @jamalex and @jredrejo as this has implications for our nginx configs - in fact, it makes the explicit overwriting of `Access-Control-Allow-Origin` headers in our nginx configs unnecessary, as I am now properly generating the host name from the request.

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/core/content/decorators.py
+++ b/kolibri/core/content/decorators.py
@@ -1,9 +1,3 @@
-from six.moves.urllib.parse import urlparse
-from six.moves.urllib.parse import urlunparse
-
-from kolibri.utils.conf import OPTIONS
-
-
 def add_security_headers(some_func):
     """
     Decorator for adding security headers to zipcontent endpoints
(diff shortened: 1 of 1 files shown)


radinamatic:
Seeing the "black box" @lauradanforth described in both FF & Edge (asset installed on Windows 10) 😕

rtibbles:
Interesting - there is a resource being loaded there that is from the Internet.

rtibbles:
Seems like this doesn't solve the issue reported, so going to close this PR.

PR #5937, opened 2019-09-12
kolibri cli returns exit code 1 if not parameter is passed

### LINKED ISSUE #5930: Running command "kolibri" (no args) gives return status 0
# Observed behavior

Edit: Original issue removed. Running "kolibri" and getting no output was due to the "auto_cd" option on ZSH being switched on implicitly by an Oh My ZSH plugin.

Only remaining issue now is that non-0 return code is potentially a bad idea in case kolibri is invoked with its arguments truncated by mistake.

### Expected behavior

Status code 1 and usage instructions.

### User-facing consequences

~Not being able to figure out whether to use `--help` `help` or `-h` or `man kolibri` etc.~

### Errors and logs
n/a

### Steps to reproduce

Run `kolibri`

### Context

Current `develop` branch

jredrejo:
### Summary
When `kolibri` cli is run without any options it retuns the help system and exit code `1`

### Reviewer guidance
1. Run `kolibri`without any option
2. It should return kolibri help
3. Executing `echo $?` should return `1` 

1. Run `kolibri` with any valid option (ex. `kolibri stop`)
2. Executing `echo $?` should return `0` 

### References
Closes: #5930 

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/utils/cli.py
+++ b/kolibri/utils/cli.py
@@ -335,15 +335,20 @@ You can also run: kolibri COMMAND --help
 """
 
 
-@click.group(help=main_help)
+@click.group(invoke_without_command=True, help=main_help)
+@click.pass_context
 @click.version_option(version=kolibri.__version__)
-def main():
+def main(ctx):
     """
     Kolibri's main function.
 
     Utility functions should be callable for unit testing purposes, but remember
     to use main() for integration tests in order to test the argument API.
     """
+    if ctx.invoked_subcommand is None:
+        click.echo(ctx.get_help())
+        ctx.exit(1)
+
     signal.signal(signal.SIGINT, signal.SIG_DFL)
 
 


benjaoming (approved):
Thanks @jredrejo - the new behavior is nice and explicit! :+1:

rtibbles:
Is this usual behaviour for help text? I can't check on my computer right now, but does just `grep` or `git` return an exit code of 0 or 1?

I looked at the original issue here and couldn't quite follow why this change was needed.

jredrejo:
@rtibbles you are asking the same question I did in the original issue. I checked it and @benjaoming was right. All the apps I could test return a non-zero error code when they show help after being executed without any parameter. For your examples, git returns `1` while grep returns `2`.

rtibbles:
Righto, sounds like we should file an upstream issue with Click too!

benjaoming:
Great idea @rtibbles - hereby reported, as I couldn't find any existing issue.

PR #5946, opened 2019-09-19
Refactor csv export to no longer require drf-csv.

rtibbles:
### Summary
* Since we now generate CSV exports on the backend, we no longer need to serve CSVs from API endpoints directly
* Remove the reliance on djangorestframework-csv that we used to use to do this, and simplify the generation of the CSVs

### Reviewer guidance
This seemed to have no effect on performance, or on output (except that I added more labels for headers), please verify, and make sure my code is readable.

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated


kolibri/core/logger/management/commands/exportlogs.py
```diff
@@ -43,81 +26,40 @@ def add_arguments(self, parser):
...
             default="session",
-            choices=["summary", "session"],
+            choices=classes_info.keys(),
             help='Log type to be exported. Valid values are "session" and "summary".',
         )
         parser.add_argument(
             "-w",
             "--overwrite",
-            action="store",
+            action="store_true",
```
jredrejo, line 35: should not it be 'store_false' ?
This makes `kolibri manage exportlogs --overwrite=true` raise the error
`kolibri manage exportlogs: error: argument -w/--overwrite: ignored explicit argument 'true'`
and having the same error if `--overwrite=flase` is used

kolibri/core/logger/test/test_api.py
```diff
@@ -582,56 +584,36 @@ def setUp(self):
...
-        self.client.login(
-            username=self.admin.username,
-            password=DUMMY_PASSWORD,
-            facility=self.facility,
-        )
+    def test_csv_download(self):
         expected_count = ContentSummaryLog.objects.count()
-        data_mock.__iter__.return_value = ContentSummaryLog.objects.all()
-        get_queryset_mock.return_value = ContentSummaryLog.objects
+        _, filepath = tempfile.mkstemp(suffix=".csv")
```
jredrejo, line 589: I remember I had to change this to a mock because creating tmp files where causing problem when running tests in Travis. Even using the `tmp_path` fixture (http://doc.pytest.org/en/latest/tmpdir.html) it failed. Do you know why this is not happening now?

jredrejo (changes requested):
Overall, it's a nice optimization . I love deleting code !
There 's an small bug, but other than that I don't see any problem. It's a clear improvement.

kolibri/core/logger/management/commands/exportlogs.py
rtibbles: Oh, this was an intentional change because I assumed the intention here was to just have `--overwrite` turn on the overwrite option. This is the way that I normally see binary flags done, without having to pass an argument - `store_true` means that if the flag is present, it will be set to `True`, and then without it, it defaults to `False`.

So the correct invocation now is `kolibri manage exportlogs --overwrite`.

kolibri/core/logger/test/test_api.py
rtibbles: Not sure - never had any issues with temp files in travis tests before.

jredrejo: I hate having such a bad memory but I do remember I had to use mocking because there were cases where the tmp files could not be created (My first implementation was exactly like the one you did it here but tests didn't pass in Travis). Maybe it was due to some previous Travis setup and the problem now is gone. I shoud have taken notes of it.

jredrejo (approved):

PR #5971, opened 2019-10-08
Change implementation of Fixed/Scrolling header, re-enable touch/momentum scroll

### LINKED ISSUE #2279: ios momentum scrolling not working

### LINKED ISSUE #3709: Momentum scrolling on PDF renderer
On a mobile device, it feels weird not to have momentum scrolling, compared to the rest of the app.

Version: v0.9.2
Browser: Firefox

### LINKED ISSUE #4827: make app bar wait a moment before re-appearing when scrolling up
### Observed behavior

feedback from @khangmach:

> Adding a slight delay before the appbar appears again upon scroll-up. The appbar seems to appear immediately, which may be weird in a lot of cases since not every slight scroll up motion may be an intention to actually scroll up. Sometimes the hand could fidget or something.

### Expected behavior

As implemented, the app bar "hangs out" immediately offscreen when it's hidden, always available to make an appearance. Instead, it should be a bit further away

Specifically, when hidden I think `top: -1 * APP_BAR_HEIGHT`. This could be adjusted to something like `top: -1.5 * APP_BAR_HEIGHT`

### User-facing consequences

app bar appears more often then it should

### Steps to reproduce

on mobile:

* scroll down
* scroll up very slightly

app bar immediately begins to re-appear

### Context

0.12

### LINKED ISSUE #5326: app bar will overshoot and detach from top of window on scroll
### Observed behavior

when scrolling quickly, it's possible for the app bar to detach from the top of the screen:

overshoot

note that when implementing this, we intentionally tried to avoid having any javascript that would re-position the header on every scroll event (e.g. by setting the `y` location as a transform).

The current implementation tries to be clever and "pins" the app bar to the page, allowing them to move upwards and downwards together:

scroll

It does this by flipping between `position: fixed` and `position: absolute`: https://github.com/learningequality/kolibri/blob/develop/kolibri/core/assets/src/views/CoreBase/ScrollingHeader.vue#L77-L85

Probably what we want to do is to flip between the two `position` modes, and instead animate just the hiding/showing portions with javascript.

### Expected behavior

app bar should never be lower than its height

### User-facing consequences

things look a little bit broken

### Errors and logs

…

### Steps to reproduce

scroll down and then scroll up quickly

### Context

0.12.2

### LINKED ISSUE #5558: shifting layout, perhaps due to floating app bar or scrollbar behavior
### Observed behavior

observed by @khangmach during a bug bash:

flicker2

### Expected behavior

layout should choose a state and stick with it

### User-facing consequences

user sees the UI jitter a little

### Steps to reproduce

likely has something to do with the exact screensize

### Context

0.12.3

chrome on mac

jonboiser:
### Summary
1. Refactor CoreBase and root CSS to make 'body' the main scrollable element
1. Add `  -webkit-overflow-scrolling: touch;` in select areas to make certain sub-elements touch scrollable (PDF/HTML5/EPUB renderer, KModal, SubNav, CoreTable)
1. Update the scroll position history stuff to listen to window scroll events and measure `window.pageYOffset`
1. Fix issue where global Aphrodite StyleSheet would be overwritten
1. Add resets for buttons in SubNav to make appearance compatible between mobile browsers.
1. Simplify implementaion of ScrollingHeader to hopefully improve animation performance and prevent the 'detaching-when-fast-scrolling' behavior observed before.

Demo:

demo

### Reviewer guidance

1. Try out the app in different small-screen mobile devices. View the app in places with a variety of layouts.

Some details about how the scrolling header is implemented:

1. The threshold for the scroll distance needed to hide the header depends on the size of the content. E.g. shorter pages need less travel than longer pages (e.g. the faclility > user page, channel page).
1. It takes 2x as much travel scrolling up to unhide the header than it does to scroll down and hide it (to mitigate #4827, as well the rebounding behavior some mobile devices have when you try to scroll past the end of a page).
1. If you pause for a second when nudging the content a small distance under the threshold for (un)hiding the header (also to mitigate #4827). This prevents situations where you might scroll up to 80% of the scroll threshold, and then toggle the header by continuing the same direction but a shorter distance.
1. If you arrive at the very top of the page, the header will reappear.
1. If you click a link on Page X when the header is hidden, then go back to Page X, the header will be shown again, but the saved scroll position will be such that you are scrolled further down by exactly the pixel height of the header (which might be slightly unpleasant). This seems to have something to do with how the browser's history/scroll position API works.

### References

Fixes #5326 
Fixes #5558 (maybe, except in Windows where the scroll bar is displayed in a more intrusive way on the very edge of the page)
Fixes #4827 
Fixes #2279 
Fixes #3709

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated


nucleogenesis:
All testing done on iPhone XS Max iOS 13 Chrome & Safari. 

**#1**

> If you click a link on Page X when the header is hidden, then go back to Page X, the header will be shown again, but the saved scroll position will be such that you are scrolled further down by exactly the pixel height of the header (which might be slightly unpleasant). This seems to have something to do with how the browser's history/scroll position API works.

In iOS 13 Chrome - I did not get this effect.
In iOS 13 Safari - I actually am scrolled _up_ a bit rather than down. 

**#2**

My scroll position is preserved only if I use the browser's back button. If I use a link to go back, I am taken to the top of the page.

Example video: https://drive.google.com/file/d/1pNAmWRNvaMGMOQVr_JL_dOFTwmEw_crS/view?usp=sharing

- Go to Coach Class Home
- Scroll Down to "Class Activity"
- Click "View All"
- If you use the browser's back button, you will return to your previous position.
- If you use the link at the top of the content "<- Class Home" then your scroll position is to the top of the page.

I didn't note any other issues or bugs. Everything looked great.

kolibri/core/assets/test/views/CoreBase/ScrollingHeader.spec.js
```diff
@@ -19,126 +19,4 @@ describe('Scrolling header movement logic', () => {
     const wrapper = createWrapper();
     expect(wrapper.exists()).toBe(true);
   });
-  it('should be pinned to the top by default', () => {
```
nucleogenesis, line 22: Should we write a new suite of tests for this feature? Is there a clear way to accomplish automated testing for it?

nucleogenesis (commented):
One other thought is whether or not we should update or add a new Gherkin story for this feature. All-in-all, everything looked and worked great in my testing. 

I suspect it would be worth also having people test in Android devices as well.

kolibri/core/assets/test/views/CoreBase/ScrollingHeader.spec.js
jonboiser: Yeah, I can add some basic tests for this. We can't really simulate actual scrolling, but do something similar to the old tests where pass in different props and make sure the component is doing the right thing in each case.

radinamatic:
@jonboiser I checked out your branch and it's running from:

```
(kolibri) didivm@didivm-VirtualBox:~/LE/kolibri$ git status
HEAD detached at jonbosier/scrollbar-delay
nothing to commit, working tree clean
(kolibri) didivm@didivm-VirtualBox:~/LE/kolibri$ git reset --hard jonbosier/scrollbar-delay 
HEAD is now at 24fec66e8 Add Gherkin story for momentum scrolling
(kolibri) didivm@didivm-VirtualBox:~/LE/kolibri$ 
```

Server is running fine in the in my dev VM, I can sign in as admin and a learner in both Firefox and Chrome:

LEDev1910 (start)  Running  - Oracle VM VirtualBox_009


However, when I try to access from other devices in the LAN (I had 2 mobiles and a table ready to test the scrolling behavior), all I can see is the flapping bird :confused: 

Even when I try to connect from the browser in the host machine (which for VirtualBox counts as just another device in the same LAN, as it has a different IP from the guest), Kolibri does not load and the console displays `ReferenceError: kolibriCoreAppGlobal is not defined` error.

Kolibri - Mozilla Firefox_008

Am I missing something?

radinamatic:
Disregard the above, I did not do the needful to make the dev env available in the LAN :woman_facepalming:, it's working fine when I install the DEB asset, continuing testing!

radinamatic:
Installed the DEB asset on Ubuntu 16.04 and tested on 3 different mobile clients:
- Safari on an old iPhone 5
- Chrome on moto g(6) phone (Android 9)
- Firefox on Huawei tablet (Android 8)

I did not get to fully test all the angles, so will continue, but overall the scrolling and header behavior appears smooth on all devices.

One thing I was not sure about is the position of the fullscreen and zoom buttons, as they seem to overlap the header while scrolling, and that was a bit confusing. 

Recordings here for [iPhone](https://drive.google.com/open?id=1yRO4tHfjigMQIC5mFD6sj8IH_DT-zIct), and [Android tablet](https://drive.google.com/open?id=1GUCaOIBzNhBpVIwq05i_DK2gwlkIEQlW).

radinamatic:
Fullscreen and zoom buttons on PDF renderer are now on the correct place! :+1: 

Kudos for the nice momentum scrolling, this behavior is much smoother compared to what we had so far. I tried to replicate all the steps and issues that went into this PR, and even though I might not have the detail-oriented-eye of somebody used to watch the smooth movements on iOS devices all day, this looks like a great improvement.

There is still a bit of a bouncing effect when scrolling to the bottom, but from what I see is limited to iPhone/iOS, no sign of it on Android devices. 

As for the app bar reappearing too fast when scrolling up, one thing noticed while nudging is that mobile browsers make the address bar disappear when scrolling down, and as soon as you start scrolling up it reappears - when that happens there's a better chance that the next nudge up will make our app bar appear too. Not sure if there's much we can do about it.

Tested on:
- Safari, Chrome and Firefox on an old iPhone 5
- Firefox and Chrome on moto g(6) phone (Android 9)
- Firefox and Chrome on Huawei tablet (Android 8)
- no actual physical Windows mobile device, but I did test on a Windows client with mobile device simulators in both Chrome and Firefox.

radinamatic (approved):
Looks very good to me, maybe @indirectlylit or @khangmach would want to take a final peek? :slightly_smiling_face:

khangmach:
Much smoother. Awesome work

kolibri/core/assets/src/views/CoreBase/ScrollingHeader.vue
```diff
@@ -287,16 +134,27 @@
...
-  }
 
-  .ease {
-    transition-timing-function: ease-in;
-    transition-duration: 0.1s;
-    transition-property: transform;
+    // Use different transition timing functions depending on direction
+    // to maintain some kind of symmetry
+    &.dir-up {
+      transition: top 0.25s ease-in;
```
indirectlylit: recommend animating with `translate` instead of `top` for performance:

https://www.paulirish.com/2012/why-moving-elements-with-translate-is-better-than-posabs-topleft/

Also, rather than `ease-in` and `ease-out`, we can consider using Material's curves:

| name | curve |
|--|--|
| Standard easing | `cubic-bezier(0.4, 0.0, 0.2, 1);` |
| Decelerate easing | `cubic-bezier(0.0, 0.0, 0.2, 1);` |
| Accelerate easing | `cubic-bezier(0.4, 0.0, 1, 1);`|

indirectlylit:
reviewing now - please wait to merge for a little bit, thanks!

kolibri/core/assets/src/styles/definitions.scss
```diff
@@ -43,3 +43,9 @@
   backface-visibility: hidden;
   perspective: 10000px;
 }
+
+// Use this mixin in select containers where the content is likely going to be
+// very large and mobile users may benefit from a smooth touch scroll experience.
+%momentum-scroll {
+  -webkit-overflow-scrolling: touch;
```
indirectlylit, line 50: cool!

kolibri/core/assets/src/views/CoreBase/ScrollingHeader.vue
```diff
@@ -14,265 +10,116 @@
...
-        if (this.pinned) {
-          return this.offset;
-        }
-        return this.offset - this.scrollPosition;
+      scrollThreshold() {
+        // Scroll thresholds are relative to how much you can scroll, since
+        // users may also modulate their scrolling based on how much content they see.
+        // Also, to mitigate overscroll rebound and other reasons,
+        // the upward threshold is set higher than the downward one.
+        const downThresh = Math.round(this.mainWrapperScrollHeight * 0.05);
```
indirectlylit: recommend putting a maximum value on this, just in case a page is crazy long

indirectlylit (approved):
looks great! just a couple minor comments

kolibri/core/assets/src/views/CoreBase/ScrollingHeader.vue
jonboiser: Which curves would replace ease-in and ease-out?

indirectlylit: whoops, forgot to link to the source: https://material.io/design/motion/speed.html#easing

They recommend:

* decelerate (ease out) for incoming (`dir-down`)
* accelerate (ease in) for outgoing (`dir-up`).

indirectlylit:
I'm going to go ahead and merge this - if we encounter future issues, we'll open new PRs. Great job!

PR #5972, opened 2019-10-08
Refactor js modules

indirectlylit:
### Summary

Been working on decoupling the shared component library, which lead me down a rabbit hole of trying to decouple fonts, which lead to some refactoring of user-agent stuff and other utility modules.

This PR should have no functional changes

### Reviewer guidance

* renamed plugin from `user_agent` to `frontend_head_assets`
* renamed `plugin_data` to `plugin_data_src` to prevent mistakenly importing it
* tried re-organizing browser info, redirecting, font loading, and other logic

feedback is welcome

### References

N/A

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated


kolibri/core/kolibri_plugin.py
```diff
@@ -96,7 +96,17 @@ def render_to_page_load_sync_html(self):
...
+        static_root = static("assets/fonts/noto-full")
+        full_file = "{}.{}.{}.css?v={}"
         return {
+            "fullCSSFileModern": full_file.format(
+                static_root, language_code, "modern", kolibri.__version__
+            ),
+            "fullCSSFileBasic": full_file.format(
+                static_root, language_code, "basic", kolibri.__version__
+            ),
+            "unsupportedUrl": reverse("kolibri:core:unsupported"),
```
rtibbles: This needs to be associated with the bundle that it was originally associated with - this variable is used by the browser minimum requirements check.

rtibbles (dismissed):
One thing that I think has to be changed, other changes seem reasonable.

The main thing I would want to check is that tree-shaking is happening as intended, otherwise, this could bloat the `user_agent`/`frontend_head_requirements` with `Screenfull` as an unnecessary dependency.

indirectlylit:
Ready for re-review.

> This needs to be associated with the bundle that it was originally associated with - this variable is used by the browser minimum requirements check.

Good catch - fixed.

> The main thing I would want to check is that tree-shaking is happening as intended, otherwise, this could bloat the user_agent/frontend_head_requirements with Screenfull as an unnecessary dependency.

You're right, `Screenfull` was being added to both `default_frontend` and `frontend_head_assets`. I've updated the PR to fix this.

Follow-up notes here: https://github.com/learningequality/kolibri/pull/6241

indirectlylit:
will need to integrate changes from https://github.com/learningequality/kolibri/pull/6270

nucleogenesis (approved):
Very strange - it looks good to me in the code review but when I went to smoke test it I cannot build the frontend assets.

Everywhere that you replaced `kolibri.utils.browser` with `kolibri.utils.browserInfo` except `CoreFullscreen.vue` isn't being pulled into my local repo for some reason. My HEAD is pointing to the right commit and I've git fetched everything a few times but those changes are not reflected in my local code.

While that is a strange issue indeed - the changes look good. When I grep-replace my code with the changes that are certainly in this commit, everything passes a smoke test.

indirectlylit:
thanks - due to botched merge

PR #5978, opened 2019-10-16
Quiz activation flow updates

nucleogenesis:
### Summary
Updates to the following:
- Quiz activation / closing / report visibility design updates.
- New Quiz Reports Header
- A new display of time since certain changes. Includes new specific ElapsedTime type component to accommodate Coach reports needs.
- Add new fields to Exam - date_created, date_archived, date_activated - as well as logic that updates those timestamps based on when their related actions occurred.
- New Quiz Status component, used for quiz state management by Coach
- New logic for Learners' ability to see quizzes based on their state.
- Updates to Coach Quiz Reports to better reflect quiz statuses
- Only provide Average Score values for quizzes that were completed _while the quiz was opened_. 
- Any started quizzes are converted to "completed" when a quiz is closed - unanswered questions are counted as wrong.
- Include a tooltip for all "Average Score" labels explaining that it only accounts for "completed" quizzes.
- CoreBase now allows override of `#main` width to accommodate Quiz > Report width design requirements for large displays.
- Misc other design updates.

### Reviewer guidance

**Gherkin Stories**
- See the updated Gherkin stories to start. Ensure they work accordingly.

**Design Overview*

Use this [Figma clickable prototype](https://www.figma.com/proto/TcSczscKh4zLTypFJh2Or4/0.13-Lesson-%26-quiz-updates?node-id=123%3A1363&viewport=1649%2C14%2C0.26439332962036133&scaling=min-zoom) for a simulation.

[Review the Figma page here.](https://www.figma.com/file/TcSczscKh4zLTypFJh2Or4/0.13-Lesson-%26-quiz-updates?node-id=100%3A515)

Frames 6.2, 7.*, 8 and 9 include design updates that should be reflected in the PR.

**Holistic Overview**
- Create a new class, add learners, assign a coach.
- Login as coach in one browser, have another browser open to use as learners (incognito, separate browser application).

__As Coach__
- Create a group or two and add learners.
- Now - in the Coach > Plan > Quizzes table, click to view a quiz's details, see that this page includes the sidebar on larger screens and is 1440px wide.
- Create multiple quizzes with some easy questions (you'll want to get some right at some point). Make sure that you go to "Edit Details" for some and assign them to group(s).

__Coach Plan Quiz Detail__
- Go to Coach > Plan > Quizzes and click a quiz link to see the detail page.
- See the header with the "All Quizzes" link back, "Quiz Title" and beneath the quiz title, see the amount of time since the quiz was created. The "Options" dropdown is also now a flat button.
- See the left sidebar with an *Open Quiz* button
- As the monitor goes to a mobile size, see that the boxes stack on top of one another
- Click "Open Quiz" - cancel the modal window and see no change.
- Click "Open Quiz" and confirm the modal, see now that the button is "Close Quiz" and has a "Opened X seconds ago" label.
- Keep this window open in a tab.

__Learner - Take The Quiz__
- In a new tab, go to the Coach > Reports > Quizzes
- See that your quizzes that are not open have "Open Quiz" buttons
- See that the quiz you already opened has the "Close Quiz" button.
- In your Learner browser window, log in as a user who is assigned to your opened quiz
- Go to your class and see that you have one quiz to take. You cannot see the other quizzes right now.
- Start the quiz, answer a few questions, DO NOT SUBMIT.
- See in the Coach Browser, that the Progess column is updated to say 1 has started and _no value_ in the Average Score (also, test the little (!) tooltip by Average Score)
- Login as a different Learner, take that quiz, submit it
- See in the Coach Browser that 1 Completed, 1 started and that there is an Average Score visible now.
- Go back to your Coach > Plan > Quizzes > Quiz Detail page for this quiz - refresh - see that there is now an "Average Score" visible. Also - the "Opened N minutes ago" and "Created N minutes ago" should update their times.

__Close Quiz__
- As the Coach, go ahead and click "Close Quiz" - test that the modal's "Cancel" works as expected (doing nothing) - but then click "Close Quiz" again and confirm the modal.
- See "Quiz Closed" and "N seconds ago" beneath it. 
- Also see "Quiz report visible to learners" and a switch set to "Yes"
- Go back to the Coach > Reports > Quizzes table and see that there are now 2 Completed quizzes and that the Average score has changed to account for the previously not completed quiz's answers being counted as wrong. The average score WILL NOT count scores for learners who didn't even start the quiz.

_Learner - Once Quiz is Completed_
- In Learn > Class - learner should see "Complete n%" on the cards of any quizzes that were completed.

__ Toggle Report Visibility __
- Go back to the Learner Browser - see that you can view your quiz report.
- As coach, toggle the "Report visible to learners" switch (see the snackbar saying it is not visible).
- Go back to the Learner Browser (refresh) and no longer see that you can view your quiz report.
- As coach, toggle the "Report visible to learners" to "yes" again, go back to Learner, refresh, and see that you can see your quiz report again.
- In the Learner browser, logout and then log in as a learner who was assigned that quiz, but who did not take the quiz, and you should not see that quiz. It is "Visible to learners" but since you did not start or complete the quiz, you have no report.

**General and Other Considerations**
- Open / Close / Toggle Visibility controls exist in multiple places - test them all:
-- Coach > Reports > Quizzes table
-- Coach > Reports > Quizzes > Quiz Detail sidebar
-- Coach > Plan > Quizzes table
-- Coach > Plan > Quizzes > Quiz Detail sidebar
- You should see a snackbar for every status change. From "newly created" to Open, Open to Close, Visible to not and vice-versa.
- All time-elapsed text items should be correct. Never see more than 59 seconds, 59 minutes, 23 hours.
- When you go to a Coach > Reports > Quizzes > Quiz Detailed Report you should see how many questions completed quizzes have answered in small text beneath the Completed badge.
- The *Recipients* label should always reflect the assigned sets of learners and the *Question order* should reflect the quiz's setting properly.
- The "Edit quiz details" page should no longer have "Active/Inactive" settings - but only the Recipients and Title - everything here should work as expected.
- Many strings that have been added also have context added in the code, may be worth reviewing for clarity.

### References

[Figma clickable prototype](https://www.figma.com/proto/TcSczscKh4zLTypFJh2Or4/0.13-Lesson-%26-quiz-updates?node-id=123%3A1363&viewport=1649%2C14%2C0.26439332962036133&scaling=min-zoom)

[Notion Project Page](https://www.notion.so/learningequality/Lesson-quiz-activation-flow-updates-738d20ced8f84d03a6bbb44e6922494e)

[Figma Design Page](https://www.figma.com/file/TcSczscKh4zLTypFJh2Or4/0.13-Lesson-%26-quiz-updates?node-id=100%3A515)

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated


integration_testing/features/coach/coach-activate-quiz.feature
```diff
@@ -3,29 +3,46 @@ Feature: Coach activates and deactivates quizzes
...
-      Then I see the *Edit quiz details for '<quiz>'* page
-      When I select *Active* under *Status*
-        And I click *Save changes* button
-      Then the page reloads
-#        And I see the snackbar notification *Changes to quiz saved* # No snackbar
-        And I see the quiz *Status* is *Active*
+  Scenario: Change the quiz status to *Open*
+    Given that quiz <quiz> has never been set to the status *Open*
+      When I am on *Coach - '<class>' > Plan > Quizzes > '<quiz>'* page
+      Then I see the *Open Quiz* button on the left sidebar
```
radinamatic: Given that *sidebar* appears only on certain browser widths, I'd eliminate that part.
``Then I see the *Open quiz* button``

quiz-sidebar

integration_testing/features/coach/coach-class-reports-quizzes.feature
```diff
@@ -68,8 +68,10 @@ Feature: Coach reviews learner reports for quizzes
   Scenario: A learner has completed a quiz
     When a learner has completed the <quiz>
     Then their *Progress* column says *Completed*
+      And their *Progress* column says *N questions answered* or *All questions answered*
```
radinamatic, line 71: Can't comment on line 65 above, but it should read `Then the learner's *Progress* column displays *X of Y questions answered*`.

radinamatic:
1.  Gherkin scenarios OK, left just a couple of comments.

2. What happened to the [`Delete quiz`](https://kolibri.readthedocs.io/en/latest/coach.html#delete-quiz) option? Did we eliminate it intentionally? cc @jtamiace 

3. > On a large monitor, note that the width of the content is 1440px

    Seems more like 1000px to me... :confused:  Or am I looking at the wrong place?

    width

4. I still believe labeling as `Completed` a quiz where they did not even answer a single question will be confusing to the learner:

    completed-or-not

5. Message `You have no quizzes assigned` should appear on learner's home for a class, when a quiz is not visible any more.

    LEDev1911n  Running  - Oracle VM VirtualBox_022

6. > Never see more than ... 23 hours.

    Will have to come back tomorrow to re-check :stuck_out_tongue: 

    >  how many questions completed quizzes have answered

    Had to re-read this a couple of times :joy:

kolibri/plugins/coach/assets/src/views/common/AverageScoreTooltip.vue
```diff
@@ -0,0 +1,33 @@
...
+</template>
+
+
+<script>
+
+  export default {
+    name: 'AverageScoreTooltip',
+    $trs: {
+      visibleToLearnersTooltipMessage: {
+        message: 'Calculated only from quizzes that were completed',
```
radinamatic, line 26: Suggest to change the wording here to `Calculated only from quizzes labeled as 'Completed'`. 

Since we are trying to avoid introducing a new status, in my mind this could potentially clarify the concept that even the quizzes which are _started_, but not **submitted** by learners themselves (that is, learners did not actively and intentionally _completed_ them), get **labeled** as _Completed_ once the coach closes the quiz.

kolibri/plugins/coach/assets/src/views/common/AverageScoreTooltip.vue
```diff
@@ -0,0 +1,33 @@
...
+        context:
+          'A message for a tooltip that shows when the user hovers over a small icon that tells the user that the Average Quiz Score is calculated only using quizzes that learners completed.',
```
radinamatic, line 28: In line with the above, change to: `Tooltip message that displays when user hovers the small '!' icon for the 'Average score' value. Indicates that the average score is calculated only from quizzes labeled as 'Completed'.`

kolibri/plugins/coach/assets/src/views/common/QuizStatus.vue
```diff
@@ -0,0 +1,248 @@
...
+          'The label for a switch that will toggle whether or not learners can view their quiz report.',
+      },
+      questionOrderLabel: {
+        message: 'Question order',
+        context: 'A label for the place where the question order is shown.',
+      },
+      yesLabel: {
+        message: 'Yes',
+        context: "Used to indicate to the user when a switch element has a true or 'on' value.",
+      },
```
radinamatic: Indicates the value of the switch element turned 'on', and the quiz report is visible for the learners.

radinamatic: Shouldn't there be a `No` value here, instead of `Yes`? Toggle is turned off... :confused: 

LEDev1911n  Running  - Oracle VM VirtualBox_023

`<rant-on>`
_I really, **really**, dislike toggles... Nothing but confusion_ :unamused: 
`<rant-off>`

kolibri/plugins/coach/assets/src/views/common/QuizStatus.vue
```diff
@@ -0,0 +1,248 @@
...
+        context:
+          'The label for a switch that will toggle whether or not learners can view their quiz report.',
+      },
+      questionOrderLabel: {
+        message: 'Question order',
+        context: 'A label for the place where the question order is shown.',
+      },
+      yesLabel: {
+        message: 'Yes',
+        context: "Used to indicate to the user when a switch element has a true or 'on' value.",
```
radinamatic: `Indicates the value of the switch element turned 'on', when the quiz report is visible for the learners.`

kolibri/plugins/coach/assets/src/views/common/StatusElapsedTime.vue
```diff
@@ -0,0 +1,219 @@
...
+        context:
+          'Indicates that an item was opened a number of hours prior to the current time, but the time is always less than one day ago',
+      },
+      openedDaysAgo: {
+        message: 'Opened {days} {days, plural, one {day} other {days}} ago',
+        context: 'Indicates that an item was opened a number of days prior to the current date.',
+      },
+      secondsAgo: {
+        message: '{seconds} {seconds, plural, one {second} other {seconds}} ago',
+        context: 'Indicates that something occured seconds prior to the current time.',
```
radinamatic: Missing one `r` in `occurred` (with 3 more instances below).

radinamatic:
These snackbars are missing something...

snackbar

kolibri/plugins/coach/assets/src/views/common/QuizStatus.vue
radinamatic: Shouldn't there be a `No` value/message here, instead of `Yes`? When toggle is turned off... :confused: 

cc @jtamiace 

LEDev1911n  Running  - Oracle VM VirtualBox_023

`<rant-on>`
_I really, **really**, dislike toggles... Nothing but confusion_ :unamused: 
`<rant-off>`

radinamatic (changes requested):
Great progress so far, a few things will need some touches! :+1: 

I commend you for all the effort in adding the context to strings :100: :heart:

kolibri/plugins/coach/assets/src/views/common/QuizStatus.vue
nucleogenesis: @radinamatic I asked @jtamiace and @khangmach about this while working on it and the standard seems to be that the visual color change largely plays a role in indicating status.

The "Yes" is there to indicate which side of the switch is the "Yes" position - I presume to cover cases of color-blindness. I initially thought that it should be a visual representation of the current value - but that's not the typical use of the switch.

integration_testing/features/coach/coach-class-reports-quizzes.feature
nucleogenesis: I updated this - it won't show as resolved on the PR because the comment is on a different line - but please check it out and if there is an issue please let me know @radinamatic

kolibri/plugins/coach/assets/src/views/common/QuizStatus.vue
indirectlylit: >  I initially thought that it should be a visual representation of the current value

That is correct. From the design:

image

kolibri/plugins/coach/assets/src/views/common/AverageScoreTooltip.vue
indirectlylit: I think this change might make things harder to understand just because it's a more complicated sentence.

Recommend simply: "Calculated only from completed quizzes". This should be clear that it just refers to the label on the quiz, and not the learner's activity.

kolibri/plugins/coach/assets/src/views/common/QuizStatus.vue
radinamatic: Ok, so there must be a `No` value/message visible to learners when the toggle is OFF, as in the image above, that's a relief :relieved:

nucleogenesis:
@radinamatic 

> On a large monitor, note that the width of the content is 1440px

> Seems more like 1000px to me... confused Or am I looking at the wrong place?

Yep! Sorry - I was not clear. This is only on the Coach > Plan > Quizzes > Quiz Detail page. I updated the reviewer guidance accordingly.

> Message You have no quizzes assigned should appear on learner's home for a class, when a quiz is not visible any more.

Fixed.

> What happened to the Delete quiz option? Did we eliminate it intentionally? cc @jtamiace

I actually removed this by mistake along with the Copy option. The are both now in Plan > Quizzes > Quiz Details dropdown in the new header at the top. Glad you noticed! Thanks!

> I still believe labeling as Completed a quiz where they did not even answer a single question will be confusing to the learner:

The way it will work is - if they `Started` the quiz but didn't answer, then the quiz is closed, then the quiz becomes `Complete` and they get a 0%. The logic is based on the state - however it could be changed so that 0 attempted answers counts the same as `Not Started` for the purposes of marking all of the quizzes complete when the quiz is closed.

@jtamiace thoughts?

I'll just note that if they never `Started` the quiz when the quiz is closed, then it's hidden from them and they don't get a score at all.

kolibri/plugins/coach/assets/src/views/common/AverageScoreTooltip.vue
nucleogenesis: @radinamatic I'm okay with this - shall we get more input from @jtamiace @lauradanforth ?

kolibri/plugins/coach/assets/src/views/common/StatusElapsedTime.vue
nucleogenesis: Thanks! Fixed.

kolibri/plugins/coach/assets/src/views/common/QuizStatus.vue
nucleogenesis: @indirectlylit @radinamatic 

I mentioned this to @jtamiace in a Slack DM but should have probably opened it in #design - anyway, here is her message on the topic:

> think we came to the conclusion that the label should always say “yes,” and leave the toggle as the thing that communicates current system status to the user
there are issues with the label being dynamic that outlined here https://www.nngroup.com/articles/toggle-switch-guidelines/
this descriptor is ambiguous: because the word Off appears on the right of the switch and the toggle dot is on the left, users can also interpret the state of the toggle as on (that is, away from the label Off).

radinamatic: To me this is still equally and utterly confusing: two different visually perceivable UI element states display the same "label":..?

confusing

Compounding issue is that fusing the two possible actions (open/close quiz; toggle report visibility) in the same column, and eliminating the value of the column header, makes it inaccessible.

jtamiace:
> I still believe labeling as Completed a quiz where they did not even answer a single question will be confusing to the learner:

> The way it will work is - if they Started the quiz but didn't answer, then the quiz is closed, then the quiz becomes Complete and they get a 0%. The logic is based on the state - however it could be changed so that 0 attempted answers counts the same as Not Started for the purposes of marking all of the quizzes complete when the quiz is closed.

I think what we were talking about is that if the learner did not make any attempts on questions, they would not see the report at all when the quiz is closed, their status would still be "has not started" (or what that string currently is) and they would not have a score (would just be a "--")

integration_testing/features/coach/coach-activate-quiz.feature
jtamiace: I notice the icon is `ic_error` next to average score - can this be changed to `ic_info`?

jtamiace:
Sorry I realize that I didn't really understand the question. Does the state change to "Started" once they click into the quiz, or after they interact with a question (but not necessarily inputting an answer)

I do think that 0 attempted answers should result in the same state as "Not started" once the quiz is closed

indirectlylit:
On the coach side we addressed this confusion by adding more nuance below "completed", e.g. "0 of 6 questions answered". Could we not simply add that information to the learner side, also?

image

I'd far prefer this than to start redefining "completion" depending on whether it's the learner or coach looking at it. That sort of inconsistency would almost be sure to bite us later.

(incidentally, maybe we should be showing the score to the learner in that view, also)

nucleogenesis:
> These snackbars are missing something...
> 
> (image redacted for brevity)

@radinamatic 

Somebody (me) mistakenly removed those a while back - this isn't an issue with this PR, but it is fixed in https://github.com/learningequality/kolibri/pull/5998

indirectlylit:
After chatting with Richard and Khang, we [came to this conclusion](https://www.notion.so/learningequality/We-will-add-the-score-to-the-learner-facing-quiz-cards-a037aaf76ab74df9a71a1ba3cfd9a5a9):

If we add the score to the learner-facing quiz cards, it should help give them a sense of the intended meaning that "the quiz is completed", rather than the unintended "you completed the quiz":

image

khangmach:
Yup, and to just add on a bit, the idea is that the coach indicator of "0 of 10 questions answered" is a helpful nuance for the coach to understand the details of a student's completion. For a learner however, a score indicator would similarly convey the same nuance of their own completion.

radinamatic:
In understand the concerns around redefining the status of _Completed_, taking it for good now, let's see if the experience on the ground is nuanced enough! :+1:

kolibri/plugins/coach/assets/src/views/common/AverageScoreTooltip.vue
radinamatic: @nucleogenesis Agreed with @indirectlylit above, could we change to `Calculated only from completed quizzes`?

radinamatic (approved):
Looking good, just one small change in the tooltip message! :tada: 

Great job, team!!! :heart:

kolibri/plugins/user/assets/src/views/UserIndex.vue
```diff
@@ -4,6 +4,7 @@
     v-bind="immersiveProperties"
     :appBarTitle="appBarTitle"
     :fullScreen="pageName === PageNames.SIGN_IN"
+    maxMainWidth="100%"
```
indirectlylit: Recommend removing this, and instead, having `fullScreen=true` imply that there is no max-width

kolibri/core/assets/src/views/CoreBase/index.vue
```diff
@@ -266,6 +273,11 @@
         type: Boolean,
         default: false,
       },
+      maxMainWidth: {
+        type: String,
+        required: false,
+        default: '1000px',
+      },
```
indirectlylit: I would recommend updating this to something like:

```javascript
      maxContentWidth: {
        type: Number,
        required: false,
        default: 1000,
      },
```

Then, have it only applied when `fullScreen` is `false`.

kolibri/plugins/coach/assets/src/views/common/AverageScoreTooltip.vue
```diff
@@ -0,0 +1,33 @@
+<template>
+
+  <div ref="icon" style="display:inline;">
+    <KIcon
+      icon="error"
+      :color="$themeTokens.primary"
+      style="position: relative; top:4px; left: 4px;"
```
indirectlylit:
```suggestion
      style="position: relative; top: 4px; left: 4px;"
```

kolibri/plugins/coach/assets/src/views/common/QuizHeader.vue
```diff
@@ -0,0 +1,82 @@
+<template>
+
+  <KPageContainer style="padding-top:16px;">
```
indirectlylit:
```suggestion
  <KPageContainer style="padding-top: 16px;">
```

kolibri/plugins/coach/assets/src/views/common/QuizHeader.vue
```diff
@@ -0,0 +1,82 @@
...
+  };
+
+</script>
+
+
+<style lang="scss" scoped>
+
+  .stats {
+    margin-right: 16px;
+  }
```
indirectlylit: appears to be unused

kolibri/plugins/coach/assets/src/views/common/QuizHeader.vue
```diff
@@ -0,0 +1,82 @@
...
+
+  h1 {
```
indirectlylit: Generally recommend against using tag selectors like `h1`, and instead defining a new class:

https://kolibri-dev.readthedocs.io/en/develop/frontend_architecture/conventions.html#styling-anti-patterns

In very simple components like this it's generally fine, but over time with larger components it can make maintainability harder.

kolibri/plugins/coach/assets/src/views/common/QuizHeader.vue
```diff
@@ -0,0 +1,82 @@
...
+  <KPageContainer style="padding-top:16px;">
+    <BackLink
+      :to="backlink"
+      :text="backlinkLabel"
+    />
+
+    <!-- Cheating to get the same layout effect but not
+         using a backlink...
+    -->
+    <BackLinkWithOptions>
```
indirectlylit: this component is called `BackLinkWithOptions` but it doesn't have a back link. This may have been an issue that pre-existed this PR when we were mistakenly aligning the options button with back links instead of headers.

Maybe it should be refactored to `HeaderWithOptions`

kolibri/plugins/coach/assets/src/views/common/QuizStatus.vue
```diff
@@ -0,0 +1,237 @@
+<template>
+
+  <KPageContainer>
+    <KButton
+      v-if="!exam.active && !exam.archive"
+      :primary="true"
+      :text="coachString('openQuizLabel')"
+      type="button"
+      style="margin-left: 0; margin-top: 1rem; margin-bottom: 0;"
```
indirectlylit:
```css
margin-top: 1rem;
```

In general, spacing is defined in multiples of `8px` to conform with the Material baseline grid

kolibri/plugins/coach/assets/src/views/common/QuizStatus.vue
```diff
@@ -0,0 +1,237 @@
...
+        <Score :value="avgScore" />
+      </KGridItem>
+      <KGridItem class="status-label">
+        {{ $tr('questionOrderLabel') }}
+      </KGridItem>
+      <KGridItem>
+        {{ orderDescriptionString }}
+      </KGridItem>
+
+    </KGrid>
```
indirectlylit, line 159: there are some strange things going on with the grid layouts here. For example:

image

Maybe we can have a quick meeting to talk about what you're trying to do and how we can do with `KGrid` or otherwise

kolibri/plugins/coach/assets/src/views/common/QuizStatus.vue
```diff
@@ -0,0 +1,237 @@
...
+    />
+    <div v-if="exam.active && !exam.archive">
+      <KButton
+        :text="coachString('closeQuizLabel')"
+        type="submit"
+        style="margin-left: 0; margin-top: 1rem; margin-bottom: 0;"
+        :appearanceOverrides="cancelStyleOverrides"
+        @click="showCancellationModal = true"
+      />
+      <StatusElapsedTime :date="examDateOpened" actionType="opened" />
```
indirectlylit: image

Could use `8px` vertical margin between the elapsed time and the button

kolibri/plugins/coach/assets/src/views/plan/QuizSummaryPage/index.vue
```diff
@@ -5,66 +5,51 @@
...
-
-    <KPageContainer v-if="!loading">
-      <section>
-        <BackLinkWithOptions>
-          <BackLink
-            slot="backlink"
-            :to="$router.getRoute('EXAMS')"
-            :text="coachString('allQuizzesLabel')"
-          />
+    <KGrid gutter="16">
```
indirectlylit, line 12: Recommend setting consistent vertical and horizontal gutters:

image

Can do this by adding a new `topMargin` prop to `KPageContainer`:

https://github.com/learningequality/kolibri/blob/develop/packages/kolibri-components/src/KPageContainer.vue#L49

kolibri/plugins/coach/assets/src/views/reports/ReportsQuizLearnerListPage.vue
```diff
@@ -125,16 +123,35 @@
...
 
 </script>
 
 
-<style lang="scss" scoped></style>
+<style lang="scss" scoped>
+
+  .small-answered-count {
+    display: block;
+    margin-left: 1.75rem;
```
indirectlylit: spacing in `8px` increments

kolibri/plugins/coach/assets/src/views/reports/ReportsQuizLearnerListPage.vue
```diff
@@ -125,16 +123,35 @@
...
+    font-size: small;
+    color: gray;
```
indirectlylit: We should avoid hard-coding any colors - they should all come from the theme.

In this case, the proper color is probably `$themeTokens.annotation`

kolibri/plugins/coach/assets/src/views/common/QuizStatus.vue
```diff
@@ -0,0 +1,237 @@
...
+    },
+    computed: {
+      orderDescriptionString() {
+        return this.exam.learners_see_fixed_order
+          ? this.coachString('orderFixedLabel')
+          : this.coachString('orderRandomLabel');
+      },
+      cancelStyleOverrides() {
+        return {
+          color: '#fff',
```
indirectlylit: Avoid hard-coding colors. Here the correct color is probably `$themeTokens.textInverted`

kolibri/plugins/coach/assets/src/views/common/QuizStatusModal.vue
```diff
@@ -0,0 +1,37 @@
...
+      },
+      modalDetail: {
+        type: String,
+        required: false,
+        default: '',
+      },
+    },
+  };
+
+</script>
```
indirectlylit: This component feels like a lot of extra complexity just to set default `continue` and `cancel` labels on the buttons. Maybe consider just using a standard `KModal`?

kolibri/plugins/learn/assets/src/views/classes/AssignedExamsCards.vue
```diff
@@ -5,12 +5,12 @@
       <h2>
         {{ $tr('examsHeader') }}
       </h2>
-      <p v-if="exams.length===0">
+      <p v-if="visibleExams.length===0">
```
indirectlylit:
```suggestion
      <p v-if="!visibleExams.length">
```

packages/kolibri-components/src/KSwitch.vue
```diff
@@ -0,0 +1,255 @@
...
+  .k-switch--color-primary {
+    &.is-checked:not(.is-disabled) {
+      .k-switch-track {
+        background-color: #a5d6a7;
+      }
+      .k-switch-thumb {
+        background-color: #4caf50;
+      }
+      .k-switch-focus-ring {
+        background-color: #a5d6a7;
```
indirectlylit, line 261: colors should be drawn from theme

indirectlylit (commented):
Looking good!

I haven't tested in depth yet, so consider this a preliminary review.

Make sure spacing is in `8px` increments, and that all colors are drawn from the theme and not hard-coded.

One thing that I'd generally caution against is building too many reusable components until we're absolutely sure that it will be a net win in the long-term.

Folks on the team know this refrain of mine. Two articles I often refer to are:

* https://www.sandimetz.com/blog/2016/1/20/the-wrong-abstraction
* https://blog.codinghorror.com/rule-of-three/

kolibri/plugins/coach/assets/src/views/common/QuizHeader.vue
nucleogenesis: Agreed but it might be easier to do it in https://github.com/learningequality/kolibri/pull/5996 (which branches from this branch) where I continue to refactor the same component similarly in Lessons - just a little bit easier to do it in one sweep on that branch than to do it here and fix it there during rebase?

kolibri/plugins/coach/assets/src/views/common/QuizStatus.vue
nucleogenesis: What browser did you see that in?

I'm approaching using KGrid and it's `:layout12` prop + `span: n` similarly to how I've used other grid-based systems wherein I assume the default is `:layout12="{ span: 12 }"` - which would make a single `<KGridItem>` effectively its own row - which in my testing worked as expected.

So - the "Report visible to learners" is span 8 and the KSwitch is span 4 - the two items in my mind making a single 12-column row.

I used KGrid expecting clean vertical spacing between each of the items and in my testing on Chromium/FF at all sizes things worked well - but if you're seeing the issues like above with the switch falling off the right edge I'll need to devise a better approach.

Would love some thoughts and insights in a call for sure @indirectlylit

kolibri/plugins/coach/assets/src/views/reports/ReportsQuizLearnerListPage.vue
nucleogenesis: This is the closest that I can get to the designs using 8px increments (24px).
8px-increment

The design calls for the text to align with the label in the labeled icon above it:
design-increments

The px value for this would be 28px (3.5 x 8px).

Should we aim to instead ensure that KLabeledIcon spaces itself in 8px increments? The icon is 15.75px in my browser - then the margin is in `em` https://github.com/learningequality/kolibri/blob/11ae8d25a104945c88c65706ab830da26022973c/packages/kolibri-components/src/KLabeledIcon.vue#L82 - but if that was spaced itself in 8px increments, then spacing the subtext using them would probably be acheivable

kolibri/plugins/coach/assets/src/views/plan/QuizSummaryPage/index.vue
nucleogenesis: Good eye @indirectlylit - I've added the prop and used the same default that was in the CSS on KPageContainer - so now I'm overriding it to use the same value as my gutters.

kolibri/plugins/coach/assets/src/views/common/QuizStatus.vue
indirectlylit: resolved in https://www.notion.so/learningequality/Final-behaviors-of-switch-toggles-in-0-13-lesson-quiz-updates-dec8f9feeaa54e6cb5122605b0a9b32f

- Switches no longer have labels in either lessons or quizzes, resulting in a slight change in layout
- Removed the switch toggle for report visibility from the quiz report table, so there is only a "Status" column that show actions and status: "Open quiz," "Close quiz," "Quiz closed"
    - Reason: confusing to have "open/close quiz" & "show/hide report" as they have overlaps in status visibility. There was also no clear user need to display the report switch in the table so we decided to return to that issue later if there is feedback on it.

kolibri/plugins/coach/assets/src/views/reports/ReportsQuizLearnerListPage.vue
indirectlylit: Got it, thanks for the info! My original note was incorrect.

some additional context:

* The `8px` baseline grid is a rule of thumb, not a requirement
* In the case of `KLabeledIcon`, we used `em` rather than `px` so that the component can be used with different font sizes, e.g. it works in both `<p>` and `<h1>`

Since you're styling the text to match the behavior of `KLabeledIcon`, your original strategy is correct.

Would you mind adding comments to the CSS flagging why these two places use `em`s?

E.g. `ReportsQuizLearnerListPage.vue`:

```css
  .small-answered-count {
    display: block;
    margin-left: 1.75rem; /* matches KLabeledIcon */
```

and `KLabeledIcon.vue`

```css
  .label {
    display: block;
    margin-left: 1.925em; /* scale with parent font size */
  }
```

kolibri/plugins/coach/assets/src/views/common/QuizHeader.vue
indirectlylit: that's fine!

kolibri/plugins/coach/assets/src/views/common/QuizStatus.vue
indirectlylit: I tested again and was not able to reproduce this behavior

indirectlylit:
let me know when ready for re-review

indirectlylit:
@nucleogenesis in the interest of keeping `develop` as recent as possible, I'm going to go ahead and merge this in. Let's address any theming or other follow up subsequent PRs.

PR #5984, opened 2019-10-23
Remote and disk file availability utilities

rtibbles:
### Summary
Adds:
* Public endpoint for querying the checksums of files available for a particular channel
* Utility function for generating the path to a public endpoint for checksums available
* Utility function for querying this endpoint and returning the checksums that can be imported for a particular channel
* Utility function for reading a drive location and returning all the files available there for a particular channel
* Tests for the latter two functions
* Integrates this into the importability annotation used in the ContentNodeGranular endpoint

Notes:
* Caches the latter two functions to prevent repeated requests and disk scans.

### Reviewer guidance
Do the tests describe the behaviour of these functions sufficiently?

### References
This is a necessary prerequisite of any functionality that seeks to show the user which files are able to be imported either from a P2P import or a local disk import.

https://www.notion.so/learningequality/Post-import-content-management-c1bba77a25ce438692ed6bc9b9f0d30c?p=fd7369c799c645e0bbaf5c74f5d297bb

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/core/content/utils/file_availability.py
+++ b/kolibri/core/content/utils/file_availability.py
@@ -15,30 +15,27 @@ from kolibri.core.content.utils.paths import get_file_checksums_url
 checksum_regex = re.compile("^([a-f0-9]{32})$")
 
 
-def _filter_checksums_by_channel(checksums, channel_id):
-    """
-    Helper function to filter checksums received to what is actually used by our
-    copy of the channel. Used by both module methods, as for a disk based import
-    there could be many files present that are not part of the channel, and for
-    remote import, the remote channel might be a different version and hence
-    have different local files associated with it.
-    """
-    return list(
-        set(
-            LocalFile.objects.filter(
-                files__contentnode__channel_id=channel_id
-            ).values_list("id", flat=True)
-        ).intersection(set(checksums))
+def generate_checksum_integer_mask(checksums, available_checksums):
+    return sum(
+        int(checksum in available_checksums) << i
+        for i, checksum in enumerate(checksums)
     )
 
 
-def _generate_mask(integer_mask):
+def _generate_mask_from_integer(integer_mask):
     while integer_mask:
         yield bool(integer_mask % 2)
         integer_mask /= 2
 
 
 def get_available_checksums_from_remote(channel_id, baseurl):
+    """
+    The current implementation prioritizes minimising requests to the remote server.
+    In order to achieve this, it caches based on the baseurl and the channel_id.
+    Also, it POSTs the complete list of non-supplementary files to the rmeote endpoint,
+    and thus can keep this representation cached regardless of how the availability on
+    the local server has changed in the interim.
+    """
     CACHE_KEY = "PEER_AVAILABLE_CHECKSUMS_{baseurl}_{channel_id}".format(
         baseurl=baseurl, channel_id=channel_id
     )
@@ -46,7 +43,7 @@ def get_available_checksums_from_remote(channel_id, baseurl):
 
         channel_checksums = (
             LocalFile.objects.filter(
-                available=False, files__contentnode__channel_id=channel_id
+                files__contentnode__channel_id=channel_id, files__supplementary=False
             )
             .values_list("id", flat=True)
             .distinct()
@@ -68,8 +65,10 @@ def get_available_checksums_from_remote(channel_id, baseurl):
                 integer_mask = int(response.content)
 
                 # Filter to avoid passing in bad checksums
-                checksums = list(
-                    compress(channel_checksums, _generate_mask(integer_mask))
+                checksums = set(
+                    compress(
+                        channel_checksums, _generate_mask_from_integer(integer_mask)
+                    )
                 )
                 cache.set(CACHE_KEY, checksums, 3600)
             except (ValueError, TypeError):
@@ -104,7 +103,11 @@ def get_available_checksums_from_disk(channel_id, basepath):
             cache.set(PER_DISK_CACHE_KEY, disk_checksums, 3600)
         else:
             disk_checksums = cache.get(PER_DISK_CACHE_KEY)
-        checksums = _filter_checksums_by_channel(disk_checksums, channel_id)
+        checksums = set(
+            LocalFile.objects.filter(
+                files__contentnode__channel_id=channel_id
+            ).values_list("id", flat=True)
+        ).intersection(set(disk_checksums))
         cache.set(PER_DISK_PER_CHANNEL_CACHE_KEY, checksums, 3600)
     else:
         checksums = cache.get(PER_DISK_PER_CHANNEL_CACHE_KEY)
--- a/kolibri/core/content/utils/file_availability.py
+++ b/kolibri/core/content/utils/file_availability.py
@@ -1,11 +1,11 @@
 import json
 import os
 import re
+from itertools import compress
 
 import requests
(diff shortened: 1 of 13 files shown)


indirectlylit:
refs https://www.notion.so/learningequality/fd7369c799c645e0bbaf5c74f5d297bb

kolibri/core/content/utils/file_availability.py
```diff
@@ -0,0 +1,111 @@
...
+        # Do something if we got a successful return
+        if response.status_code == 200:
+            try:
+                integer_mask = int(response.content)
+
+                # Filter to avoid passing in bad checksums
+                checksums = list(
+                    compress(channel_checksums, _generate_mask(integer_mask))
+                )
+                cache.set(CACHE_KEY, checksums, 3600)
```
jamalex, line 73: Instead of time, could we have this get invalidated when the server's CONTENT_CACHE_KEY (or whatever it's called) changes? So, retrieve that key at the start and use it as part of this local cache key?

kolibri/core/public/api.py
```diff
@@ -117,3 +123,34 @@ def get_public_channel_lookup(request, version, identifier):
...
+        checksums = json.loads(data.decode("utf-8"))
+        available_checksums = set(
+            LocalFile.objects.filter(available=True, id__in=checksums)
+            .values_list("id", flat=True)
+            .distinct()
+        )
+        return HttpResponse(
+            sum(
+                int(checksum in available_checksums) << i
+                for i, checksum in enumerate(checksums)
```
jamalex: Perhaps put this in a helper function alongside `_generate_mask`, so their logic is clearly defined in parallel?

kolibri/core/content/utils/file_availability.py
rtibbles: See the new comment for the function to see where I landed on this. I think reducing requests is a reasonable strategy for now.

kolibri/core/public/api.py
rtibbles: Done

rtibbles:
Behaviour if you import from yourself - unavailable topics are hidden, but it doesn't limit the `total_resources` of topics currently.

importing_from_yourself

rtibbles:
N.B. I did an implementation to add correct `total_resource` counts for the import source as well and while it surprised me by not complaining of too many SQLITE variables, it upped the response time on a production server for the contents of the `Math` topic on Khan Academy from 10s with this implementation, to 2 minutes (when all the Khan Academy files were available).

It's not impossible that this could be brought down to an acceptable level with optimization.

radinamatic:
So the first series of test was done by installing the DEB asset from some 7 hours ago (built most probably after [this commit](https://github.com/learningequality/kolibri/pull/5984/commits/14463570cf2a29c70aea36d0aaebe44d5df66952)) on Ubuntu 16.04 VM. 

1. Sometimes the `There was a problem loading this page...` lingers even after the channel DB is finished importing from the attached USB. It displays the drive name correctly when the page is reloaded, but looks bad and confusing.

   Ubuntu16 04 (start)  Running  - Oracle VM VirtualBox_157

2. If I understand correctly, the code in this PR does not calculate how many resources of the given channel are *actually present* on the attached drive, and the **Total size** value still displays **all** the resources in the channel DB, whether present on the local drive or not...?

   At this point there were actually only 4 KA EN resources already imported through another channel, and **On your device** value in the screenshot seems correct. When I attached the `TCWORLD15` USB drive which had a portion of the KA EN channel (version 14), the first confusing thing was the value of **14 of 21833 resources on your device**, compared to the above 4.

   Ubuntu16 04 (start)  Running  - Oracle VM VirtualBox_159

   Turns out that was because one of them was present in 9 (!!!) different locations :unamused: 
Not sure if this is something in the scope of this PR (is it?), or the current content management refactor, just noting that it is confusing and making testing a bit challenging. If need be, I'll open another issue.

   Ubuntu16 04 (start)  Running  - Oracle VM VirtualBox_162

3. Then I decided to _import_ one of the subtopics (**Math > Pre-algebra**) that I imagined _was not really_ on the USB drive (I probably never check that particular checkbox), and when I checked it, it reported 562 resources and 600 MB as **selected**. 

   Ubuntu16 04 (start)  Running  - Oracle VM VirtualBox_163

   Again, not sure if in the scope of this PR, but confusing especially since it took 5 minutes to report importing only 4 files, of which I ended finding just 2 in the **Math > Pre-algebra > ...** folder when browsing on Learn:

   Ubuntu16 04 (start)  Running  - Oracle VM VirtualBox_164

4. I can't say if local import was much _slower then before_, because browsing topics of the channel like KA EN in this workflow has always been slow, but even going back to **Math** when one is seeing its subtopics can take up to 20 seconds ([screenscast here](https://drive.google.com/open?id=1Z-jkcBOP-gx9gl-Ee-_6BV_PCMhqic_N), too big for GH), which **is** admittedly slow. 

5. Finally, import from local network was unsuccessful too. LAN address is recognized, list of channels from the peer device is correctly loaded, when `Select topics` button is pressed it apparently starts to load the DB, but breaks soon after that and displays the `Channel not found` error.  

   import-local-asset

   Tried with several different channels, plus adding the `https://develop.studio.learningequality.org` as a local address and it always broke on the same point. As a final check on the same host I fired up the VM where I was testing the 0.12.9b2 for VF-IS, and it imported correctly from channels located on the peer device. 

6. Since in the meantime you committed more stuff, I checked out your `rtibbles:importability_functionality` branch and run the devserver: peer import also breaks at the same point as on the VM where Kolibri from the (slightly older) DEB asset was installed, and import from the same attached USB was even slower (browsing the KA EN channel tree was interminable, I even gave up recording the GIF).

[Zipped `KOLIBRI_DATA` folder](https://drive.google.com/open?id=1Zel62AGPkb2_r2dQfoE-4_OLSG5ymn8c) from my `TCWORLD15` USB drive, if you want to test on your side.

rtibbles:
I am not sure what in this PR would have caused any breakage at any point - although the interminable slowness would definitely be contributed to by this PR.

All of the points you make about the total resources are true - as I note above, I tried implementing that but it made an endpoint you already described as 'interminably slow' even slower, so it seems that might be beyond reach here.

rtibbles:
OK - looks like there are definitely some 500s coming from the endpoint that I've been messing with. Could you post the `kolibri.log` here so that I can debug?

radinamatic:
> that might be beyond reach here

Suspected that much...

Logs are here [kolibri-2019-11-10.txt.zip](https://github.com/learningequality/kolibri/files/3829790/kolibri-2019-11-10.txt.zip), should have posted them above, sorry!

rtibbles:
Thanks for the logs, the errors make sense now.

radinamatic:
Update on the most recent tests on this performed earlier today: 

- local address import is now working correctly when importing from peer running the same dev branch (I understand that after the proper stuff has been baked in, peer import will be available independent of the version that the peer is running)

- import from USB is working, slower on devserver compared to production, but that is to be expected.

Jumping on #6049 to continue poking!

rtibbles:
Note that peer import from an older peer should be working with this PR - it's just not possible to tell what is available and what is not.

indirectlylit (approved):
I have not tested. However, in the interest of having `develop` as current as possible, I'd like to merge now and have issues surface earlier

PR #5985, opened 2019-10-23
Bug fixes and other code cleanup for Device plugin

jonboiser:
### Summary

Some bug and tech debt fixes to incorporate before working on bigger features in the Device plugin:

1. Remove the dynamic component from DeviceIndex and replace with `router-view`s
1. Take advantage of previous refactor to introduce `withAuthMessage` higher order component that will show `AuthMessage` if the user is not authorized for a given page (and will show the desired component if they are). Also, made sure all routes in Device plugin have proper authorization.
1. Remove `SELECT_CONTENT_TOPIC` route from the router and extend the `SELECT_CONTENT` route to accept an additional `?node_id` query param to use while navigating a channel in import/export workflow. This is intended to simplify extending the router to handle more new workflows that will rely on a similar URL scheme.
1. Clean up test suite (rename i.e. `component-name.spec` to `ComponentName.spec`, create reusable mock router, replace test-only CSS classes with `data-test` attrs)
1. Fix bugs in Device Info Channel, Manage Content Page, and Manage Permissions Page.

### Reviewer guidance

1. Do different things in Device plugin and look for obvious bugs.
1. Try doing things while not authenticated to test out

### References

…

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated


kolibri/plugins/device/assets/src/views/ManageContentPage/ChannelListItem.vue
```diff
@@ -37,7 +37,7 @@
     </template>
 
     <template slot="meta">
-      <div v-if="inImportMode && onDevice" class="spec-ref-on-device">
+      <div v-if="inImportMode && onDevice" data-test="on-device">
```
indirectlylit, line 40: Interesting! I like this strategy much better than using a `class` with the `spec-ref-` prefix.

If we're going to start using this convention, I'd slightly prefer a more explicit name like `data-spec-ref` or `data-test-ref` because initially I thought `data-test` was leftover debugging code.

PR #5990, opened 2019-10-24
Give default full name when creating a user with createsuperuser or importusers command

### LINKED ISSUE #5923: 'importusers' and “createsuperuser” management commands should not allow empty 'full name' field
### Observed behavior

The `kolibri manage importusers` command allows empty 'full name' fields when importing from CSV  files and defaults them to the empty string. This allows Facility Users to be made who have an empty full name, which is not possible inside the Kolibri UI, and can cause layout issues.

### Expected behavior

`importusers` doesn't allow empty 'full name' fields.

### User-facing consequences

…

### Context

Kolibri 0.13

jonboiser:
# Summary

1. When using `kolibri manage createsuperuser`, create FacilityUser model with `full_name` equal to username
1. Do the same using `kolibri manage importuser`, when full_name column is either missing or has an empty string
1. Add tests for this feature
1. Refactor tests (noticed that by not calling setup_device() so often, it reduces the run time to 1/3 of original). also added a cleanup to remove tempfiles created by importusers.

### Reviewer guidance

1. Run `createsuperuser` and `importusers` without full names and check that the resulting user accounts have their full name defaulted to username in the app.

### References

Fixes #5923
----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/core/auth/management/commands/importusers.py
+++ b/kolibri/core/auth/management/commands/importusers.py
@@ -119,9 +119,10 @@ def create_user(user, default_facility=None):
 
     except FacilityUser.DoesNotExist:
         password = user.get("password", DEFAULT_PASSWORD) or DEFAULT_PASSWORD
+        full_name = user.get("full_name", "") or username
         try:
             new_user = FacilityUser.objects.create_user(
-                full_name=user.get("full_name", ""),
+                full_name=full_name,
                 username=username,
                 facility=facility,
                 password=password,
--- a/kolibri/core/auth/test/test_user_import.py
+++ b/kolibri/core/auth/test/test_user_import.py
@@ -4,6 +4,7 @@ from __future__ import unicode_literals
 
 import csv
 import tempfile
+import os
 
 from django.core.management import call_command
 from django.core.management.base import CommandError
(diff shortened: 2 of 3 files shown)


kolibri/core/auth/test/test_user_import.py
```diff
@@ -136,77 +137,85 @@ def test_create_user_not_exist_bad_username(self):
 ]
 
 
+def test_device_not_setup(self):
```
rtibbles: Best to just wrap this in another test case class (you can do standalone functions for tests, but the form is a bit different).
```
class UserImportNoSetupTestCase(TestCase):
    def test_device_not_setup(self):
```

rtibbles (dismissed):
One change needed, then good to go.

kolibri/core/auth/management/commands/importusers.py
```diff
@@ -119,9 +119,10 @@ def create_user(user, default_facility=None):
 
     except FacilityUser.DoesNotExist:
         password = user.get("password", DEFAULT_PASSWORD) or DEFAULT_PASSWORD
+        full_name = user.get("full_name", "") or username
```
indirectlylit, line 122: shouldn't this just be

```python
full_name = user.get("full_name", username)
```

I don't really understand the use of the `or` in this line or the `password` line above. If the point is to avoid empty strings, I think it's much more explicit to use a ternary statement:

```python
full_name = user.get("full_name") if user.get("full_name") else username
```

(Statements with `or` in them look like they return booleans to me, not a way of choosing default values)

rtibbles: The point is to always set a `full_name` regardless of whether it has been passed in as blank, similarly for password. If it has been passed in as blank, it still sets the default password.

Doing `user.get("full_name")` will throw an `AttributeError` if `user` does not have this field, whereas having the default value will prevent this.

As such, to achieve the goal of having a non-empty `full_name` and `password` regardless of input, the code that is here is best - the other alternative would be:

`full_name = user.get("full_name", "") if user.get("full_name", "") else username`

rtibbles: This is a nice summary of how boolean operators map to conditional expressions: https://docs.python.org/3/library/stdtypes.html#boolean-operations-and-or-not

indirectlylit: Maybe using `or` as a way of choosing defaults is more idomatic than I thought? Not sure.

While we're nerding out, I think this one is actually my favorite, though certainly the most verbose:

```python
full_name = user.get("full_name")
if not full_name:
    full_name = username
```

rtibbles: My sense is that `or` is quite Pythonic, at the very least.

rtibbles: > Doing user.get("full_name") will throw an AttributeError if user does not have this field, whereas having the default value will prevent this.

As @indirectlylit kindly pointed out to me, this is incorrect. It will return `None`.

indirectlylit: at least [some people on the internet](https://stackoverflow.com/a/13710667) consider `value or default` to be an anti-pattern, but who knows...

my current favorite given all of the above:

```python
full_name = user.get("full_name") if user.get("full_name") else username
```

rtibbles: My personal dislike for the conditional expression/ternary is the repeated evaluation - although given that it is a dict getter, this is not particularly important in this instance.

indirectlylit (approved):
feedback addressed

PR #5994, opened 2019-10-26
update ifcfg to 0.19

indirectlylit:
Any reason that upgrading to ifcfg 0.19 for kolibri 0.12.9 would be risky?

https://github.com/ftao/python-ifcfg/releases


benjaoming (approved):
Nope, none at all, they might also be useful for people with VMs :)

PR #5995, opened 2019-10-26
changelog for 0.12.9

indirectlylit:
Does this look correct?

Any typos?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,23 @@
 
 List of the most important changes for each release.
 
+## 0.12.9
+
+### Internationalization and localization
+
+- Added Korean
+
+### Changed or fixed
+
+- Database vacuum now works correctly
+- Fixes related to network detection
+
+### Added
+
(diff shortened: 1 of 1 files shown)


lyw07 (approved):
Looks good to me! just wondering about one small thing - does the order of Internationalization and localization, Changed or fixed, Added matter here?

indirectlylit:
>  does the order of Internationalization and localization, Changed or fixed, Added matter here?

These headers are a fairly recent addition. Do you have any recommendations?

lyw07:
Oh I was asking just because our previous releases seem to list them in the order like:
```
Changed or fixed
Added
Internationalization and localization
```
so I was wondering if we have a convention or not. If we don't, then everything looks good to me!

indirectlylit:
good catch! I've updated the ordering to be consistent

lyw07:
thank you for addressing that! please feel free to merge it if you think everything is good to go :)

PR #6005, opened 2019-10-29
Add svg slot for KCheckbox

ralphiee22:
### Summary

Need to be able to add an svg icon for KDP, so added an optional slot to the `KCheckbox` component. Either use the slot or the label, but not both.

**_Before:_**
Screen Shot 2019-10-29 at 4 06 14 PM

**_After:_**
Screen Shot 2019-10-29 at 4 01 30 PM

### Reviewer guidance

Add within any `KCheckbox` and remove `label` attribute.
```
              <mat-svg
                class="svg"
                category="communication"
                name="business"
              />
              <span class="text">some text
              </span>
``` 

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/packages/kolibri-components/src/KCheckbox.vue
+++ b/packages/kolibri-components/src/KCheckbox.vue
@@ -43,12 +43,10 @@
 
       </div>
 
-      <div v-if="$slots.svg" class="k-checkbox-icon">
-        <slot name="svg"></slot>
-      </div>
+      <slot v-if="$slots.default"></slot>
 
       <label
-        v-if="label"
+        v-else-if="label && !$slots.default"
         dir="auto"
         class="k-checkbox-label"
         :for="id"
@@ -78,7 +76,7 @@
        */
       label: {
         type: String,
-        required: true,
+        required: false,
       },
       /**
        * Whether to show label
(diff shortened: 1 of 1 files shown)


indirectlylit:
I would expect this to be implemented using [`KLabeledIcon`](https://github.com/learningequality/kolibri/blob/release-v0.12.x/kolibri/core/assets/src/views/icons/KLabeledIcon.vue)

I'd also prefer not to put it in 0.12.9 unless we absolutely need to.

indirectlylit (changes requested):
This PR updates a core kolibri component (the checkbox) with new functionality (adding an icon config option).

Because this is a fundamental design pattern which would get introduced to the design system, I think we should be a bit more careful about the decision and make sure it's the right one. While it's not obviously wrong, I do have some reservations about having a checkbox and an icon directly adjacent to each other, because it makes it a little noisy and less clear what is clickable.

As an interim solution, I propose giving the checkbox a more generic slot which allows you to replace the entire label with whatever you want, rather than baking the icon functionality into the checkbox component itself. This mirrors what we do with `KButton` as well.

This provides us the immediate flexibility we need, and we can work in the future with @jtamiace and @khangmach later to decide if adding icons to `KCheckbox` components is a pattern that we expect to happen more often.

indirectlylit:
FYI, merging the latest `develop` will fix the failing test

indirectlylit (approved):

PR #6007, opened 2019-10-30
Vendor iceqube with scheduler

rtibbles:
### Summary
Vendors iceqube with these updates:
* RQ equivalent API
* RQ scheduler compatible scheduler module

Updates periodic background thread tasks to use the scheduler instead of independent threads.

### Reviewer guidance
Do all background tasks still execute as expected (ping back, vacuum).

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated


kolibri/core/tasks/iceqube/__init__.py
```diff
@@ -0,0 +1,7 @@
+from kolibri.core.tasks.iceqube.queue import Queue
+from kolibri.core.tasks.iceqube.worker import Worker
```
indirectlylit: now that it's vendored, could we give the module a more meaningful name? Currently it's not clear what role `iceqube` plays and how it relates to the sibling code:

image

kolibri/core/tasks/iceqube/__init__.py
```diff
@@ -0,0 +1,7 @@
...
+
+
+__version__ = "0.4"
```
indirectlylit: still necessary?

kolibri/core/tasks/iceqube/classes.py
```diff
@@ -0,0 +1,172 @@
...
+            return self.progress
+
+    def __repr__(self):
+        return "<Job id: {id} state: {state} progress: {p}/{total} func: {func}>".format(
+            id=self.job_id,
+            state=self.state,
+            func=self.func,
+            p=self.progress,
+            total=self.total_progress,
+        )
```
indirectlylit: "classes.py" is a bit non-descriptive for a file name

kolibri/core/tasks/iceqube/utils.py
```diff
@@ -0,0 +1,117 @@
...
+
+def import_stringified_func(funcstring):
+    """
+    Import a string that represents a module and function, e.g. {module}.{funcname}.
+
+    Given a function f, import_stringified_func(stringify_func(f)) will return the same function.
+    :param funcstring: String to try to import
+    :return: callable
+    """
+    assert isinstance(funcstring, str)
```
indirectlylit: Would be [slightly preferable](https://stackoverflow.com/a/20076295) to `raise` a custom error rather than using `assert`

indirectlylit (commented):
exciting stuff!

kolibri/core/deviceadmin/utils.py
```diff
@@ -179,3 +181,37 @@ def search_latest(search_root, fallback_version):
...
+        else:
+            logger.info("Sqlite database Vacuum finished.")
+
+
+def schedule_vacuum():
+    current_dt = local_now()
+    vacuum_time = current_dt.replace(hour=3, minute=0, second=0, microsecond=0)
+    if vacuum_time < current_dt:
+        # If it is past 3AM, change the day to tomorrow.
+        vacuum_time = vacuum_time.replace(day=vacuum_time.day + 1)
```
lyw07: Got an error when the date today is the last day of the month. (i.e. Oct 31st)
Here is part of the error traceback:
```
File "kolibri/utils/server.py", line 90, in run_services
    schedule_vacuum()
  File "kolibri/core/deviceadmin/utils.py", line 217, in schedule_vacuum
    vacuum_time = vacuum_time.replace(day=vacuum_time.day + 1)
ValueError: day is out of range for month
```

lyw07 (dismissed):

kolibri/core/analytics/management/commands/ping.py
```diff
@@ -70,91 +49,11 @@ def handle(self, *args, **options):
...
-        ]
-        facilities = [extract_facility_statistics(f) for f in Facility.objects.all()]
-        data = {"pi": pingback_id, "c": channels, "f": facilities}
-        logger.debug("Statistics data: {}".format(data))
-        jsondata = dump_zipped_json(data)
-        response = requests.post(url, data=jsondata, timeout=60)
-        response.raise_for_status()
-        return json.loads(response.content.decode() or "{}")
+                ping_once(started, server)
+            except Exception as e:
```
lyw07, line 56: sorry I'm just curious here - when we ping once, we will raise the exception. when we schedule a ping, the exceptions will be caught and only warning logs will be shown. Is this intended? Thank you!!

kolibri/core/tasks/iceqube/__init__.py
rtibbles: Sure - can refactor, I mostly kept it as is, as there had been some suggestion that we might want to allow for it to be unvendored at some point in the future, but that would be as easy to undo.

kolibri/core/tasks/iceqube/__init__.py
rtibbles: No.

kolibri/core/tasks/iceqube/classes.py
rtibbles: But... it has the classes!

Can address as part of the renaming.

kolibri/core/deviceadmin/utils.py
rtibbles: Ah, always good to test on an edge case day, nice catch!

kolibri/core/tasks/iceqube/utils.py
rtibbles: Yes - for the stringify_func method above also.

kolibri/core/analytics/management/commands/ping.py
rtibbles: Yes that's the intended behaviour.

lyw07: thanks for clarifying!

kolibri/core/deviceadmin/utils.py
rtibbles: This should be working correctly now, using a timedelta.

rtibbles:
Have addressed review comments - fixed the bug that @lyw07 found, and melted iceqube into kolibri to address @indirectlylit's concerns.

indirectlylit (approved):
this is fine to merge, pending an update to automatically upgrade DBs for developers

PR #6019, opened 2019-11-05
commands to upload and download glossary file

indirectlylit:
### Summary

We've moved the contents of the [glossary Google doc](https://docs.google.com/document/d/1_pctX5MdTblQlQumow0CCu28qOaE88BW72iw4qqs9yA/edit) to [crowdin's glossary](https://crowdin.com/project/kolibri/settings#glossary).

The risk here is that we don't have it checked in to git, which means that we don't have a reliable "source of truth". (Crowdin tends to be unpredictable at times, so it's better to have it stored in code.)

This PR adds two commands:

* `make i18n-upload-glossary`
* `make i18n-download-glossary`

It also adds a new file, `glossary.tbx`, which will serve as the source of truth for our glossary moving forward.

This PR also does some cleanup and refactoring of the `crowdin.py` file, mainly migrating to `click` from `argparse`

### Reviewer guidance

Updates to the glossary have some similarities with context, but also some important differences:

* The glossary exists for the whole crowdin project, not per-branch
* Uploading the glossary merges with the existing one, it doesn't replace it
* Arbitrary new glossary entries can be added by translators and editors, in both English and other languages

At this point I propose that we do _not_ include glossary updates as a standard part of our release process and string freeze, and instead do manual, intentional uploads and downloads as we see fit.

### References

related to the string refactor project: https://www.notion.so/learningequality/String-refactor-e79dd8c431ef4a1084142ebf5e81ccab

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated


radinamatic:
> At this point I propose that we do not include glossary updates as a standard part of our release process and string freeze, and instead do manual, intentional uploads and downloads as we see fit.

Sounds sensible, agreed!

`TBX` is the _proper_ format for glossary type of resources, but one concern I might have is that the XML structure is complex and not easily readable and/or editable locally. In the past I tried to tinker with it in the OxygenXML app (one of the best I know for working with XML files), and it wasn't an easy sailing.

Crowdin allows for upload and download of glossaries in CSV and XLSX formats, along with the TBX:

Kolibri - Learning Platform settings - Crowdin - Mozilla Firefox_145
 
Could we add the option to keep the CSV version of the glossary together with the TBX. _That_ one will be much easier to edit and clean up locally.

indirectlylit:
> Crowdin allows for upload and download of glossaries in CSV and XLSX formats, along with the TBX

Unfortunately this is not actually true. I tried using CSV first for exactly the reasons you describe, but Crowdin only allows upload – not download – through the API.

radinamatic:
> Crowdin only allows upload – not download – through the API

Figures... :unamused: 

Ok, since we are already keeping pulling the trigger on this manually, when we see fit, could we at least add the option to do the upload of the CSV version (that **is** allowed via API) too? So when we do want to upload a _cleaned_ or new version of the glossary, we grab the CSV by manual download, do the local editing with it, and then use the `make i18n-upload-glossary` to push it to Crowdin? 

Both new commands in this PR worked well for me otherwise, I managed to download the glossary, added a new term in the TBX, uploaded back, changed something on Crowdin, downloaded again! :+1: 

Let me know if the suggestion above is doable, and can be added in the scope of this PR.

indirectlylit:
> So when we do want to upload a cleaned or new version of the glossary, we grab the CSV by manual download, do the local editing with it, and then use the make i18n-upload-glossary to push it to Crowdin?

If we're already downloading the CSV manually, couldn't we just as easily upload it manually? I'm curious when editing by CSV would be preferable to editing in the Crowdin web UI?

My sense is that for the most part:

* Adding new glossary entries and deleting old ones should be done in Crowdin
* Edits to existing entries can be done in either Crowdin or the XML file

radinamatic:
Fair enough, I remember for some reason I preferred editing the CSV locally when I did the most recent upload, and I/we can still do it if necessary.

radinamatic (approved):
LGTM! :+1:

indirectlylit:
sounds good. For now, we can treat this functionality mostly as a way to back up and restore the state of the crowdin glossary, versus a tool for actual editing.

indirectlylit:
(I'll wait for @nucleogenesis to do a code review and then we can merge)

nucleogenesis (approved):
Looks great @indirectlylit - I really appreciate moving the URL definitions to the parts of the file where they're relevant.

PR #6020, opened 2019-11-05
double dose of dumps and updated language files

### LINKED ISSUE #5981: Uncaught SyntaxError: missing ) after argument list
### Observed behavior

Error on load in console, seems to be related to code in `hooks.py`:

```python
    def frontend_message_tag(self):
        if self.frontend_messages():
            return [
                '<script>{kolibri_name}.registerLanguageAssets("{bundle}", "{lang_code}", JSON.parse("{messages}"));</script>'.format(
                    kolibri_name="kolibriCoreAppGlobal",
                    bundle=self.unique_id,
                    lang_code=get_language(),
                    messages=json.dumps(
                        self.frontend_messages(),
                        separators=(",", ":"),
                        ensure_ascii=False,
                    )
                    .replace("'", "\\'")
                    .replace('"', '\\"'),
                )
            ]
        else:
            return []
```

Might also be preventing language assets from being loaded?

### Expected behavior

No error

### User-facing consequences

No translations on device page

### Errors and logs

```
Uncaught SyntaxError: missing ) after argument list
```

### Steps to reproduce

load device page

### Context

0.13.0 (develop)

indirectlylit:
### Summary

* replaces manual escape logic with another pass through `json.dumps`
* updates language files with most recent export from crowdin, and new names from modules

### Reviewer guidance

does this solve the problems?

### References

fixes #5981

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated


nucleogenesis:
@indirectlylit It removes the errors on my screen - but I am not getting translations when I change my language.

I don't see the `registerLanguageAssets` at all in the page.

indirectlylit:
Looks like this might be related to https://github.com/learningequality/kolibri/pull/5887

This code is looking for `kolibri.plugins.user.app-messages.json` but the actual file name is `user_module-messages.json`:

https://github.com/learningequality/kolibri/blob/5b7b0e4e2ae8c5edc81869db4cb70036eb69aa05/kolibri/core/webpack/hooks.py#L206-L218

cc @rtibbles thoughts?

nucleogenesis:
@indirectlylit @rtibbles Ahh this will be fixed after we go through the translation process again. The files will be named like the `kolibri.plugins.user.app-messages.json` after we do the i18n process through Crowdin.

You can see their names-to-be in the `locale/en/LC_MESSAGES` folder - the names of the CSV files are what will be uploaded to Crowdin, downloaded and then converted to the properly namespaced JSON files that will ultimately contain the translations.

rtibbles:
Yeah - I had posed this question when my PR doing this renaming was originally open, and we agreed that we could wait for us to reintroduce the translation files under the new names, so I didn't rename them.

indirectlylit:
So to be clear, I should pull the latest from crowdin and add them to this PR?

build_tools/i18n/crowdin.py
```diff
@@ -326,8 +326,18 @@ def _locale_data_from_csv(file_data):
...
+        json[row["Identifier"]] = row.get("Translation") or row.get(None)
+
+        # hack - sometimes it's a list?
+        if isinstance(json[row["Identifier"]], list):
+            json[row["Identifier"]] = json[row["Identifier"]][0]
+
+        # hack - sometimes it's None?
+        if not json[row["Identifier"]]:
+            json[row["Identifier"]] = ''
+
```
indirectlylit: @nucleogenesis I'm not sure why, but I had to add the hacks above in order to get things working

indirectlylit (commented):
@nucleogenesis I've run the crowdin scripts and updated all the JSON files. Language switching seems to work now

kolibri/core/webpack/hooks.py
```diff
@@ -262,17 +262,20 @@ def js_and_css_tags(self):
...
                     bundle=self.unique_id,
                     lang_code=get_language(),
                     messages=json.dumps(
-                        self.frontend_messages(),
-                        separators=(",", ":"),
-                        ensure_ascii=False,
-                    )
-                    .replace("'", "\\'")
-                    .replace('"', '\\"'),
+                        json.dumps(
```
rtibbles, line 273: At first I was confused as to why this works with only a singe `JSON.parse` on the frontend side of it.

Then I realized that the double dumps not only works, but is actually conceptually correct.

The first `json.dumps` just makes the Python object suitable for inlining in Javascript code as a valid Javascript object (previously we were kind of hacking that by putting it in quotation marks).

The second `json.dumps` takes the JSON string and makes is suitable for inlining in Javascript code as a valid Javascript string (which can then be properly fed to `JSON.parse`).

rtibbles (commented):
Code makes sense, but have not tested.

rtibbles:
Needs black linting, FYI.

kolibri/core/webpack/hooks.py
indirectlylit: Yeah, I enjoyed figuring this one out and the elegance of the solution

indirectlylit: If we wanted to be correct & safe everywhere, we should avoid

```python
"<script>console.log('{}')</script>".format(output_string)
```

and always do

```python
"<script>console.log({})</script>".format(json.dumps(output_string))
```

created a follow-up issue here: https://github.com/learningequality/kolibri/issues/6044

build_tools/i18n/crowdin.py
```diff
@@ -326,8 +326,17 @@ def _locale_data_from_csv(file_data):
...
+        # Third index is the translation, but might have a key 'None'
+        json[row["Identifier"]] = row.get("Translation") or row.get(None)
+
+        # hack - sometimes it's a list?
+        if isinstance(json[row["Identifier"]], list):
+            json[row["Identifier"]] = json[row["Identifier"]][0]
+
+        # hack - sometimes it's None?
+        if not json[row["Identifier"]]:
+            json[row["Identifier"]] = ""
```
nucleogenesis: This is weird - but the row ought to be a dict where the keys are gathered by the header row of the CSV that was parsed on line ~ 314 `csv_data = list(row for row in csv.DictReader(f))` - if a key is None, then the CSV likely is missing a header.

nucleogenesis: Also I see no reason why it wouldn't be a list.

Did you preapprove all of the translations on whatever branch you were using for testing? Missing translations is my best guess but I'm not sure. I've never encountered this while testing with a fully translated branch though.

nucleogenesis: And - if that's the case - it might be weird to have incomplete translations as well.

> So to be clear, I should pull the latest from crowdin and add them to this PR?

Would it have been easier (albeit a bit tedious) to write a script to rename all of the files in the locale directory to avoid pulling incomplete translations into the repo for now (assuming that's the case).

indirectlylit: reverted the change, re-ran everything, and now .. it works?

Issue #6008, opened 2019-10-30
Facility admin pages very slow with thousands of users


indirectlylit:
### Observed behavior

Reports that on a server with ~8000 users, facility pages are sometimes giving errors and timeouts are occurring:

![image](https://user-images.githubusercontent.com/2367265/67884531-3be75200-fb03-11e9-840e-596f5c8b5d3b.png)

### Expected behavior

The user interface should remain responsive and not cause user-facing errors, even for very high numbers of users (to be defined).

We may also want to provide guidance to users on how to user Postgresql rather than SQLite

### User-facing consequences

Inability to administer server.

### Errors and logs

None on the backend. Waiting for info re: client

### Steps to reproduce

Presumably, create a server with tens of thousands of users and interact with the Facility tab

### Context

EC2 from AWS with the following specs:

* T3 Large:  2vCPU, 8 GB of RAM
* Runs on 2.5GHX Intel Processor

Kolibri 0.12.8

KevinTripp:
Here is the chrome dev tools view of the page loading. It is one call that is getting a 504 error. 

![image](https://user-images.githubusercontent.com/2286110/67886957-c5594d00-fb20-11e9-92fd-d2f2a0c817c4.png)

Request URL: https://kolibri.connectedcorrections.com/api/auth/classroom/?parent=79f960e7d7afd0754776f9fcb04fe035&1572459020932=1572459020932
Request Method: GET
Status Code: 504 
Remote Address: 107.23.10.241:443
Referrer Policy: no-referrer-when-downgrade

KevinTripp:
@indirectlylit This is blocking our rollout of Kolibri, even though we are working on a few workarounds.  What is the status of this issue?

indirectlylit:
Hi @KevinTripp –

Sorry, we haven't had a chance to triage this issue yet. I'll update where once we've investigated the scope of a mitigating change.

KevinTripp:
@indirectlylit Thanks, I appreciate the update.

Is it possible to create classes and modify what students are part of a class via the api?

rtibbles:
Hi @KevinTripp this is possible - I am seeing if we can do a workaround here within the scope of 0.12.9 to avoid having to interact with the API directly.

What is the approximate number of classes that is being displayed on this page, so I can ensure that I am testing against the appropriate scale? Also, do you have a sense of how many coaches are assigned to each classroom on average?

rtibbles:
[...]

PR #6026, opened 2019-11-06
Don't serialize collection parent on roles

rtibbles:
### Summary
Provides a roughly 10x speedup for the classroom API endpoint
Does this by removing the unnecessary `collection_parent` field from the role serializer
The role serializer gets nested inside the FacilityUserSerializer that is used to return the coach data
Removes the only reference to the `collection_parent` field in the frontend, and replaces it with a bulk delete of all roles for that user (which is what the frontend filtering and multiple requests are also trying to achieve).

### Reviewer guidance
Does creating a user with a role still work?
Does editing a user's role from something to a learner still work?

### References
Should provide some relief for #6008 

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/core/auth/serializers.py
+++ b/kolibri/core/auth/serializers.py
@@ -16,17 +16,9 @@ from kolibri.core import error_constants
 
 
 class RoleSerializer(serializers.ModelSerializer):
-    collection_parent = serializers.SerializerMethodField()
-
     class Meta:
         model = Role
-        fields = ("id", "kind", "collection", "user", "collection_parent")
-
-    def get_collection_parent(self, instance):
-        if instance.collection.parent is not None:
-            return instance.collection.parent.id
-        else:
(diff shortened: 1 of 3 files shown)


indirectlylit (approved):
as far as I can tell, this removes the only reference to `collection_parent`

LGTM

radinamatic:
Asset installed on Ubuntu 16.04, creating users with all roles and reverting them back to learners is successful. All users can sign in, navigate and see their own profiles correctly, **before and after** the change.

Ubuntu16 04 (start)  Running  - Oracle VM VirtualBox_147


@rtibbles Is the manual check on Windows also needed?

radinamatic (approved):
Switching roles from and back to `learner` works without issues! :+1:

PR #6133, opened 2019-11-24
Remove language cookie

### LINKED ISSUE #5794: Django language cookie persistently overrides device language setting
### Observed behavior

When a super admin sets the device language - for example, to Spanish - and a user later uses the language selection modal to personalize to different language, for example, French - then the browser that the user used to select French will not automatically revert to using the device language.

The translation logic prioritizes the language code in the URL first, then sessions and cookies, then the device language setting. However, we are not clearing that cookie regularly and a user must enter an Incognito browser to access Kolibri in the device settings language.

### Expected behavior

At some point, the device settings language ought to take priority. Using a user's language of choice that they personally set in the browser for an extended period of time makes sense - however - it ought to be safe to assume that the device admin has set a language that best serves their and their learners' needs for the most part.

Ideally, we would have a timed refresh where a browser's cookie expires and the browser defaults to the device language.

### User-facing consequences

Users may need to manually adjust their language settings  between sessions, particularly in a shared context.

rtibbles:
### Summary
* Removes any use of the language cookie
* Renames the csrf token to be kolibri specific

### Reviewer guidance
Does the cookie now not get set and have no effect?

### References
Fixes #5794 - only removes the language cookie, as I think the language setting per user would need additional UI elements.

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated


indirectlylit (changes requested):
this seems to break language switching for me: 

image

> I think the language setting per user would need additional UI elements

I don't think so? Basically this would be implicitly set whenever someone uses the 'switch language' function. I can think of two flows:

1. user is on the sign in page
2. switch to another language
3. sign in (at this point the current language is associated with the account)

alternatively:

1. user is on the sign in page
2. sign in
3. switch to another language  (at this point the current language is associated with the account)

rtibbles:
I think doing it implicitly without the user option to explicitly set this would have almost no advantage over the current cookie based system, with the added issue that there is no way to clear the setting.

Error seems to be CSRF related, does it persist after a full page refresh? (this PR updates the csrf cookie name also)

rtibbles:
Oh, I need to change the token reference in the frontend as well, hence the breakage.

indirectlylit:
> I think doing it implicitly without the user option to explicitly set this would have almost no advantage over the current cookie based system, with the added issue that there is no way to clear the setting.

I can't speak to the implementation details. However here's a scenario to consider...

Let's say that an admin sets up Kolibri at a kiosk in a predominantly Spanish-speaking area, and sets the default language to Spanish...

* **French Learner** logs in and sets their language to French. They log out, and maybe the log-in page stays in French for a period of time (e.g. 1 hour)
* Later, **Spanish Learner** shows up at the kiosk and log-in page should be in Spanish. They log in, and everything stays in Spanish
* Later, **French Learner** comes back and the log-in page is still in Spanish. They log in using the Spanish page, and everything switches to French because it remembers the preference



> Error seems to be CSRF related, does it persist after a full page refresh? 

yes it does

rtibbles:
I understand the scenario - the issue is that this causes an implicit setting to become predominant over all other mechanisms to change the language - which is the same reason that we had added the device setting and then made it much more impactful in effect, and are now removing the cookie, in order to demote the impact of implicit settings.

Having the user language setting entirely implicit would make the original issue that was filed even more intractable, as not even clearing cookies would revert to the device setting.

indirectlylit:
Hm, I think we have two different interpretations of the issue, which says:

> the browser that the user used to select French will not automatically revert to using the device language
> ...
> Ideally, we would have a timed refresh where a browser's cookie expires and the browser defaults to the device language.
> ...
> Users may need to manually adjust their language settings between sessions, particularly in a shared context.

I take that to mean that, for shared browsers, the device setting should take precedence for anonymous users and users who have not yet chosen a language.

However, choosing a language is an explicit, intentional action, and should not be overwritten by admin device settings:

image

rtibbles:
The action is explicit, but the fact that it has this permanent effect would be implicit in this scheme (because the current setting is displayed to neither user nor admin) and it would be irrevocable (because there is no mechanism to unset it).

indirectlylit:
I don't really understand what you mean ... the current setting is displayed to the user, and they can change it at any time:

image

The only thing they can't currently do is say "use server default", which seems like something of an edge case.

rtibbles:
The fact that this is associated with a user is implicit, and it is a facility user setting that is not inspectable or editable by an admin.

indirectlylit:
Ah, I see what you're saying: admins would expect to be able to see and edit it in the user editor?

image

rtibbles:
Yes, and in parallel it seems appropriate that it would be in the user profile editor for users also (even if the language switcher is a shortcut for this).

indirectlylit (approved):
This appears to be an improvement with regard to #5794. Also, the CSRF token rename may also help with https://github.com/iiab/iiab/issues/1532

I think there are still some subtle follow-up issues for multiple users sharing a single client-side device and using multiple languages. I'll try to document those in follow-up issues.

PR #6160, opened 2019-11-27
Add extra logic to ensure H5P doesn't initialize before ready [donotmerge]

jamalex:
### Summary

H5P code was added around 0.12.0, and Hashi then changed approaches to a different style of async in 0.12.4, which seems to have broken the way H5P loads.

The H5P bootstrapping code runs first, and watches for the page to have finished loading (i.e. all the dependency scripts) before trying to start the app. When loaded by hashi, the page is already loaded, so it jumps the gun. This PR adds in an extra check so it... won't.

Slack context: https://learningequality.slack.com/archives/C0LK8QS9J/p1574708012271500

### Reviewer guidance

For reviewing, focus on the 2nd commit: https://github.com/learningequality/kolibri/commit/f8babd4822e42b5eecd4b7b46c0010d70cf40b8c

NOTE: this will likely be replaced by something @rtibbles does with Hashi itself on Monday. Assuming that works, this can be shuttered and closed.

### DIFF
--- a/kolibri/core/content/static/assets/h5p-standalone-dist/js/finished-loading.js
+++ b/kolibri/core/content/static/assets/h5p-standalone-dist/js/finished-loading.js
@@ -1,0 +1,1 @@
+window.hashiFinishedLoading = true;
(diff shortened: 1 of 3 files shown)


indirectlylit:
ref: https://github.com/learningequality/kolibri/issues/6211

ivanistheone:
superseded by hashi fixes in https://github.com/learningequality/kolibri/pull/6268

PR #6195, opened 2019-12-03
Parallel task execution for UI blocking tasks

### LINKED ISSUE #6178: channel listings will not load while a task is in progress
### Observed behavior

Channel listing get stuck while a task is in progress:

image

image

### Expected behavior

Operations that are not visible to the user in the task manager should never be blocked by tasks which are visible to the user.

### User-facing consequences

The user cannot continue to queue up new content while imports are in progress

### Steps to reproduce

1. start a long-running import
2. try to import from a new channel that has never been viewed before

### Context

0.13.0 beta 3

rtibbles:
### Summary
* Adds a priority queue for UI blocking tasks
* Adds a separate worker process with 3 workers to process UI blocking tasks
* Adds a DB write lock to prevent async tasks fighting over the database

### Reviewer guidance
Does the code make sense?
Does channel metadata import happen quickly in spite of queued content import jobs?

### References
Fixes #6178 

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/core/auth/management/commands/sync.py
+++ b/kolibri/core/auth/management/commands/sync.py
@@ -14,6 +14,7 @@ from ..utils import get_dataset_id
 from kolibri.core.auth.constants.morango_scope_definitions import FULL_FACILITY
 from kolibri.core.auth.management.utils import get_facility
 from kolibri.core.tasks.management.commands.base import AsyncCommand
+from kolibri.core.tasks.utils import db_task_write_lock
 from kolibri.utils import conf
 
 DATA_PORTAL_SYNCING_BASE_URL = conf.OPTIONS["Urls"]["DATA_PORTAL_SYNCING_BASE_URL"]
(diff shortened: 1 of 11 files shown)


kolibri/core/auth/management/commands/sync.py
```diff
@@ -158,13 +159,15 @@ def handle_async(self, *args, **options):
 
         # pull from server and push our own data to server
         if not no_pull:
-            sync_client.initiate_pull(Filter(dataset_id))
+            with db_task_write_lock:
+                sync_client.initiate_pull(Filter(dataset_id))
         if not no_push:
             sync_client.initiate_push(Filter(dataset_id))
```
ralphiee22: I think we need to wrap this as well because we serialize as part of a pull or push.

kolibri/core/content/management/commands/importchannel.py
```diff
@@ -174,12 +175,16 @@ def _start_file_transfer(  # noqa: C901
                                 .exclude(kind=content_kinds.TOPIC)
                                 .values_list("id", flat=True)
                             )
-                            import_ran = import_channel_by_id(
-                                channel_id, self.is_cancelled
-                            )
+                            with db_task_write_lock:
```
ralphiee22, line 178: any reason not to group the two statements into a single lock?

rtibbles: We could, but I was trying to make it more granular to prevent any task hogging the DB for too long.

indirectlylit (approved):
this looks good to me, and works well

some more general questions...

* is there any systematic way to know that we're wrapping every necessary location with a lock (e.g. @ralphiee22's note) and also
* how do we ensure future devs do this?
* what behavior will we expect to see if we missed one?

rtibbles:
> is there any systematic way to know that we're wrapping every necessary location with a lock (e.g. @ralphiee22's note) and also

Currently, no. What we would need to do is create a standard test that we can resuse for async tasks that detects DB writes outside of the lock context.

> how do we ensure future devs do this?

This would have to be in code review, to ensure this test was added for every async task.

> what behavior will we expect to see if we missed one?

Sometimes we might see database locked operational errors reported.

Issue #6170, opened 2019-11-29
task progress percentage can be a very long negative number in scientific notation during import


mrpau-richard:
### Observed behavior
When I import this channel `nosov-pagif` random data is shown briefly

### Expected behavior

It should display the correct progress of the channel import.

### User-facing consequences

confusion and broken-looking behavior

### Errors and logs
![temp](https://user-images.githubusercontent.com/8663934/69861980-6ed45f80-12d4-11ea-9cc4-abbc87fcf4ff.gif)

### Steps to reproduce

…

### Context

 * Kolibri version: Kolibri 0.13 b3
 * Operating system: Windows 7
 * Browser: Firefox(70.0.1)

radinamatic:
Not sure if I would describe this as a **broken** progress bar... 🤔 

Channel is small, so the bar appears just for a moment, but I also noticed this strange sequence of numbers when importing. Was that what you were referring to @mrpau-richard?

![2019-12-01_3-55-06](https://user-images.githubusercontent.com/1457929/69919241-30c57000-147b-11ea-86de-2f11a298a69e.jpg)

cc @nucleogenesis

indirectlylit:
updated description and title

lyw07:
this is because cloudflare compresses json files, and for compressed files `content-length` is not available in the header. The file size is set to 1e+100 which caused the issue when calculating the progress.
this issue should be fixed by https://github.com/learningequality/kolibri/pull/6198

PR #6198, opened 2019-12-03
set file total_size when file is compressed

lyw07:
### Summary

When a file is compressed either on Cloudflare (e.g. json files) or Google Cloud Storage (content databases in the future), the content-length of the file would not be available in the response headers and thus the total_size of the file would be set to 1e100.

However, click's progress bar requires the file size value to be an integer. https://github.com/learningequality/kolibri/blob/release-v0.13.x/kolibri/core/tasks/management/commands/base.py#L32
The workaround is to get the value of `X-Goog-Stored-Content-Length` when `content-length` is not available in the headers.

When we do remote peer-to-peer import, and a file is compressed through nginx, the content-length of the file would not be available either. I don't know other ways of getting the size of the response except calculating the length of the response content (Please feel free to comment in the PR if there's a better way to handle this). If we continue to set the file size to `1e+100` in the case that file is gzipped, progress bar will have weird behavior when importing via webUI, and an error will occur when importing via CLI.

### Reviewer guidance

Test case 1:
1. run kolibri 0.13.0 beta
2. import channel `7ad6cdc532ea4e73a43a4c47b942c2a7` (this database file has been compressed) through `kolibri manage importchannel network 7ad6cdc532ea4e73a43a4c47b942c2a7` and see that there's an error saying it needs Int instead of Float.
3. run the kolibri installer from this PR
4. import the channel again through CLI and see that the channel database is imported successfully.
(no need to import content here)

Test case 2:
1. run kolibri 0.13.0 beta
2. import the content of the channel `nosov-pagif`
3. see that the progress bar has a weird large value
4. run the kolibri installer from this PR
5. import the content of the channel `nosov-pagif` and see that the progress bar works correctly

Test case 3:
1. run kolibri 0.13.0 beta
2. set up peer-to-peer import from kolibribeta.learningequality.org
3. import some html5 apps from CK-12
4. see that the progress bar has a weird large value
5. run the kolibri installer from this PR
6. set up peer-to-peer import from kolibribeta.learningequality.org
7. import html5 apps from CK-12 and see that the progress bar works correctly

### References

https://github.com/learningequality/kolibri/issues/6170

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/core/content/utils/transfer.py
+++ b/kolibri/core/content/utils/transfer.py
@@ -160,21 +160,19 @@ class FileDownload(Transfer):
         # initiate the download, check for status errors, and calculate download size
         self.response = self.session.get(self.source, stream=True, timeout=self.timeout)
         self.response.raise_for_status()
-        try:
-            self.total_size = int(self.response.headers["content-length"])
-        except Exception:
+
+        content_length = self.response.headers.get("content-length")
+        gcs_content_length = self.response.headers.get("X-Goog-Stored-Content-Length")
+        if content_length:
+            self.total_size = int(content_length)
+        elif gcs_content_length:
             # When a compressed file is saved on Google Cloud Storage,
             # content-length is not available in the header,
             # but we can use X-Goog-Stored-Content-Length.
-            gcs_content_length = self.response.headers.get(
-                "X-Goog-Stored-Content-Length"
-            )
-            if gcs_content_length:
-                self.total_size = int(gcs_content_length)
-            else:
-                # HACK: set the total_size very large so downloads are not considered "corrupted"
-                # in importcontent._start_file_transfer
-                self.total_size = 1e100
+            self.total_size = int(gcs_content_length)
+        else:
+            # Get size of response content when file is compressed through nginx.
+            self.total_size = len(self.response.content)
 
         self.started = True
 
--- a/kolibri/core/content/utils/transfer.py
+++ b/kolibri/core/content/utils/transfer.py
@@ -162,7 +162,7 @@ class FileDownload(Transfer):
         self.response.raise_for_status()
 
         try:
-            self.total_size = self.response.headers["content-length"]
+            self.total_size = int(self.response.headers["content-length"])
         except KeyError:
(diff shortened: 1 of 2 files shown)


lyw07:
realize that the progress bar will have issues not only when the files are compressed online. will address this in this PR as well.

kolibri/core/content/utils/transfer.py
```diff
@@ -163,9 +163,18 @@ def start(self):
...
+            # When a compressed file is saved on Google Cloud Storage,
+            # content-length is not available in the header,
+            # but we can use X-Goog-Stored-Content-Length.
+            gcs_content_length = self.response.headers.get(
+                "X-Goog-Stored-Content-Length"
+            )
+            if gcs_content_length:
+                self.total_size = int(gcs_content_length)
+            else:
+                # HACK: set the total_size very large so downloads are not considered "corrupted"
```
lyw07: @jonboiser Do you recall what would lead to this `corrupted` file result? I saw that in your peer-to-peer import PR, you added this hack to help resolve the issue of the corrupted files during peer-to-peer import. But I'm not able to reproduce it by running peer-to-peer locally.
I'm trying to do this because setting size to 1e100 now has some issue with the progress bar. Thank you!

jonboiser: The `self.total_size = 1e100` hack was to make sure that field was set to very large number just in case the peer server was not setting its 'content-length' headers due to some kind of misconfiguration. It needed to be a large number since Kolibri would assume that file size being less than the 'content-length' header implied the file was corrupt.

jonboiser: I forget how we first found this bug. I think if you try peer to peer with different localhost ports, it's fine. So maybe we tried doing peer to peer with either a kolibribeta server or kolibri studio itself, not sure.

lyw07: thank you for the clarification!

indirectlylit (approved):
looks good to me! Tested the scenarios and everything works well

PR #6207, opened 2019-12-04
Updated peer discovery functionality

micahscopes:
### Summary
This PR completes the initial implementation of 0.13 peer discovery functionality.  It includes ~two~ three big changes from the previous zeroconf work:
- Retooling the `NetworkLocation` data structures and API to handle both static and dynamic locations using a common interface that is compatible with the import wizard as is.
- An update to the front-end implementation to match designs
- A gentle connection availability checker that is thoroughly unit tested and uses cached storage

Still to-do:
- end to end tests in a tox environment
- content availability stats in a public endpoint (preferably cached)
- some UI transition cleanup

Before:
image

After:
image

### Reviewer guidance
There are some Gherkin stories you can look at.

You can start lots of little Kolibri's on your computer using stuff like:
```
KOLIBRI_HOME=~/.kolibriLocalPeer12 kolibri start --port 8285
KOLIBRI_HOME=~/.kolibriLocalPeer932 kolibri start --port 9191
```
Try starting and stopping them and see if your peer list updates correctly.

### References
[notion docs](https://www.notion.so/learningequality/Auto-discovery-of-P2P-import-sources-4b5199f75def4c93982a155f15fd6a9b)

### Contributor Checklist

PR process:

Testing:

Todo:
- end to end tests in a tox environment
- content availability stats in a public endpoint (preferably cached)
- some UI transition cleanup

### Reviewer Checklist

- PR is fully functional
- Documentation is updated


kolibri/core/discovery/models.py
```diff
@@ -10,6 +12,14 @@ class NetworkLocation(models.Model):
     which can be used to sync content or data.
     """
 
+    class Meta:
+        ordering = ["added"]
+
+    # for statically added network locations: `id` will be a random UUID
+    # for dynamically discovered devices: `id` will be the device's `instance_id`
+    id = models.CharField(
+        primary_key=True, max_length=36, default=uuid.uuid4, editable=False
```
micahscopes, line 29: The ID has been modified so that this table can be used for both static (saved) and dynamic (discovered) network resources.

There are two proxy interfaces to the table for each of these use cases.

The initial reason for keeping one table was to avoid making unnecessary modifications to the channel import wizard front end.  But I kinda like how it turned out.

kolibri/core/discovery/models.py
```diff
@@ -10,6 +12,14 @@ class NetworkLocation(models.Model):
     which can be used to sync content or data.
     """
 
+    class Meta:
+        ordering = ["added"]
```
micahscopes, line 24: because this model's ID has been changed to a uuid string, we need a consistent way to sort `NetworkLocation` objects

kolibri/core/discovery/models.py
```diff
@@ -37,4 +49,48 @@ def available(self):
             self.save()
             return True
         except NetworkClientError:
+            if self.dynamic:
+                self.delete()
```
micahscopes: Dynamic `NetworkObjects` are automatically pruned when they're discovered to be unavailable.

kolibri/core/discovery/models.py
```diff
@@ -37,4 +49,48 @@ def available(self):
...
             return False
+
+
+class StaticNetworkLocationManager(models.Manager):
```
micahscopes, line 65: this explicit `Manager` might be unnecessary for the static case, but it's included for consistency

kolibri/core/discovery/models.py
```diff
@@ -37,4 +49,48 @@ def available(self):
...
+
+    def save(self, *args, **kwargs):
+        self.dynamic = False
+        return super(StaticNetworkLocation, self).save(*args, **kwargs)
+
+
+class DynamicNetworkLocationManager(models.Manager):
+    def get_queryset(self):
+        queryset = super(DynamicNetworkLocationManager, self).get_queryset()
+        return queryset.filter(dynamic=True)
```
micahscopes: at some point, this could be extended to dynamically filter out `DynamicNetworkLocation` objects with a redundant `StaticNetworkLocation` in existence (i.e. one with the same `instance_id`)

kolibri/core/discovery/utils/network/search.py
```diff
@@ -130,31 +145,75 @@ def remove_service(self, zeroconf, type, name):
             del self.instances[id]
 
 
-def get_available_instances(timeout=2, include_local=True):
+cache = Cache(os.path.join(KOLIBRI_HOME, "zeroconf_cache"))
```
micahscopes: using `diskcache` for now, although it might be nice to use redis if it's available (to avoid writes on raspberry pi sd cards, for example.

kolibri/core/discovery/utils/network/search.py
```diff
@@ -130,31 +145,75 @@ def remove_service(self, zeroconf, type, name):
...
+
+
+def gather_instance_data(instance):
```
micahscopes: this function is no longer used and should be deleted!

kolibri/core/discovery/utils/network/search.py
```diff
@@ -130,31 +145,75 @@ def remove_service(self, zeroconf, type, name):
...
+    endpoint_path = "/api/content/channel/?available=true"
+    network_client = NetworkClient(instance.get("base_url"))
+    instance["data"]["channels"] = network_client.get(endpoint_path).json()
+
+
+def find_peer_instances():
```
micahscopes: this is a raw interface to the zeroconf listener (with no debouncing or caching)

kolibri/core/discovery/utils/network/search.py
```diff
@@ -130,31 +145,75 @@ def remove_service(self, zeroconf, type, name):
...
-        instance["self"] = (
-            ZEROCONF_STATE["service"] and ZEROCONF_STATE["service"].id == instance["id"]
-        )
-        instances.append(instance)
-    return instances
+
+    return list(ZEROCONF_STATE["listener"].instances.values())
+
+
+def get_available_instances(timeout=2, include_local=True):
```
micahscopes: This is the main, higher level interface to the zeroconf listener

kolibri/core/discovery/utils/network/search.py
```diff
@@ -130,31 +145,75 @@ def remove_service(self, zeroconf, type, name):
...
+    """Find peer Kolibri instances, check their availability then add them to the database.
+    Returns either a `DynamicNetworkLocation` queryset or a list of DynamicNetworkLocation objects."""
+
+    discoveries_are_fresh = cache.get(ZEROCONF_DISCOVERIES_ARE_FRESH)
```
micahscopes: This `ZEROCONF_DISCOVERIES_ARE_FRESH` flag tells other potential zeroconf listeners (maybe even in other threads) not to spam the network if we've just recently scanned it.

kolibri/core/discovery/utils/network/search.py
```diff
@@ -130,31 +145,75 @@ def remove_service(self, zeroconf, type, name):
...
+                    id=instance.get("data").get("instance_id"),
+                )
+
+            except NetworkLocationNotFound:
+                logger.info(
+                    "The device with id %s could no longer be reached" % instance["id"]
+                )
+
+        cache.set(ZEROCONF_DISCOVERIES_ARE_FRESH, True, ZEROCONF_MIN_ALLOWED_REFRESH)
+        return DynamicNetworkLocation.objects
```
micahscopes: I was doing some fancy stuff to accumulate available instances and return a list of `DynamicNetworkLocation` objects known to be available, but ran into some caching confusions.

This function could probably be refactored to not return anything (and renamed).

kolibri/core/discovery/utils/network/search.py
```diff
@@ -130,31 +145,75 @@ def remove_service(self, zeroconf, type, name):
...
+                continue
+
+            if instance["self"]:
+                continue
+
+            try:
+                (
+                    network_location,
+                    created,
+                ) = DynamicNetworkLocation.objects.update_or_create(
```
micahscopes: here's where we cache the discovered peers... they will magically be in the same format as the manually saved ones, so that the content import wizard doesn't have to know that these dynamically discovered peers are any different from manually saved ones.

kolibri/deployment/default/settings/base.py
```diff
@@ -286,7 +286,7 @@
 }
 
 # System warnings to disable
-# see https://docs.djangoproject.com/en/1.9/ref/settings/#silenced-system-checks
+# see https://Kdocs.djangoproject.com/en/1.9/ref/settings/#silenced-system-checks
```
micahscopes: I don't understand this :thinking:

kolibri/core/discovery/utils/network/search.py
```diff
@@ -130,31 +145,75 @@ def remove_service(self, zeroconf, type, name):
...
+    cache.set(ZEROCONF_SERVICE_ID, short_id)
     if ZEROCONF_STATE["service"] is not None:
         unregister_zeroconf_service()
-    logger.info("Registering ourselves to zeroconf network with id '%s'..." % id)
-    data = {"version": kolibri.VERSION}
-    ZEROCONF_STATE["service"] = KolibriZeroconfService(id=id, port=port, data=data)
+    logger.info("Registering ourselves to zeroconf network with id '%s'..." % short_id)
+    data = {
+        "version": kolibri.VERSION,
+        "instance_id": id,
```
micahscopes: Here we're passing the full `instance_id` through zeroconf.  It will be used as the id for the `DynamicNetworkLocation` object, ultimately to make it easier to reference in the content import wizard.

kolibri/plugins/device/assets/src/views/ManageContentPage/SelectNetworkAddressModal/api.js
```diff
@@ -43,34 +47,23 @@ export function fetchAddresses(withChannelId = '') {
   });
 }
 
-export function fetchDevices(withChannelId = '') {
```
micahscopes, line 46: This is abstracted into three nearly identical interfaces for fetching static, dynamic and generic `NetworkLocationResource`s

kolibri/core/discovery/utils/network/search.py
rtibbles: Also for Cloud Kolibri we have seen issues when the underlying file system is a virtualised POSIX file system on top of a Windows file system.

My currently open PR #6195 has a similar issue - I think we can probably address both with a compatability layer similar to the cross process cache that uses Redis when available.

kolibri/core/discovery/utils/network/search.py
```diff
@@ -130,31 +144,69 @@ def remove_service(self, zeroconf, type, name):
...
-        instance["self"] = (
-            ZEROCONF_STATE["service"] and ZEROCONF_STATE["service"].id == instance["id"]
-        )
-        instances.append(instance)
-    return instances
+
+    return list(ZEROCONF_STATE["listener"].instances.values())
+
+
+def run_peer_discovery(timeout=2, include_local=True):
```
micahscopes: This is the high level interface to peer discovery.  It handles caching and is debounced in order to prevent spamming the network.

The debounce time is currently set to 5 seconds (in `settings.py` via `ZEROCONF_MIN_ALLOWED_REFRESH`)

kolibri/core/discovery/utils/network/search.py
rtibbles: This is great,excellent separation of concerns between the frontend and the backend.

kolibri/core/discovery/api.py
```diff
@@ -1,21 +1,32 @@
...
 
 
-class NetworkSearchViewSet(viewsets.ViewSet):
-    permission_classes = (UserHasAnyDevicePermissions,)
+class DynamicNetworkLocationViewSet(viewsets.ModelViewSet):
+    permission_classes = (CanManageContent,)
+    serializer_class = NetworkLocationSerializer
+    queryset = DynamicNetworkLocation.objects
 
     def list(self, request):
```
micahscopes: This endpoint is the main interface to discovered locations.  It both runs peer discovery and serializes the current cache of `DynamicNetworkLocation` objects.  During serialization, it will be calling `DynamicNetworkLocation.available`, which will automatically delete unavailable locations.  So it also prunes unavailable locations as well.

kolibri/core/discovery/api.py
```diff
@@ -1,21 +1,32 @@
...
-from kolibri.core.device.permissions import UserHasAnyDevicePermissions
-from kolibri.core.discovery.utils.network.search import get_available_instances
+from kolibri.core.discovery.utils.network.search import run_peer_discovery
 
 
 class NetworkLocationViewSet(viewsets.ModelViewSet):
     permission_classes = (CanManageContent,)
     serializer_class = NetworkLocationSerializer
-    queryset = NetworkLocation.objects.all()
+    queryset = NetworkLocation.objects
```
rtibbles: Does this work? `NetworkLocation.objects` is a Manager instance, not a queryset.

kolibri/core/discovery/models.py
rtibbles: Does any code depend on this being a UUID? Pre-existing `NetworkLocation` objects will have an integer id (which I assume will be turned into a string by the migration), so they would not be UUIDs.

kolibri/core/discovery/test/test_network_search.py
```diff
@@ -151,13 +151,6 @@ def test_irreconcilable_naming_conflict(self):
...
 
-    def test_excluding_local(self):
-        initialize_zeroconf_listener()
-        register_zeroconf_service(MOCK_PORT, MOCK_ID)
-        assert len(get_available_instances()) == 1
-        assert len(get_available_instances(include_local=False)) == 0
-        unregister_zeroconf_service()
-
     def tearDown(self):
         unregister_zeroconf_service()
```
rtibbles, line 168: If you are getting contamination between tests, then you might need to call an explicit:
`call_command("flush", interactive=False)`
in the `tearDown` method because of the `TransactionTestCase`.

kolibri/core/discovery/utils/network/search.py
rtibbles: Looking at the way you are using it, I think the cross process cache wrapper that is already implemented would handle this? (it will use redis when available, but otherwise will default to the `diskcache` based caching layer).

kolibri/core/discovery/utils/network/search.py
rtibbles: As a rule we have tended to avoid adding non-Django settings to the settings files, and left things that need to be configured to things specified in `options.py`.

It seems like the main motivation here is to have differential behaviour between prod and develop - but I think it might be easier to just have that modified during dev time when someone is working on it.

I'd suggest just setting it within this file as `5` for now. If we want to make it configurable we can add it to `options.py`.

kolibri/utils/cli.py
```diff
@@ -579,12 +579,18 @@ def status():
...
+    type=int,
+    help="Port on which to run Kolibri services",
+)
 @click.option(
     "--background/--foreground",
     default=True,
     help="Run Kolibri services as a background task",
 )
-def services(background):
+def services(port, background):
```
rtibbles, line 593: `kolibri-server` is not currently using the `services` command so this shouldn't cause any issues immediately. @jredrejo when we do switch `kolibri-server` to using the services command, would it be consistent with how everything is setup to explicitly tell the services command what port Kolibri is being served on?

kolibri/utils/cli.py
```diff
@@ -579,12 +579,18 @@ def status():
 
 
 @main.command(cls=KolibriDjangoCommand, help="Start worker processes")
+@click.option(
+    "--port",
+    default=OPTIONS["Deployment"]["HTTP_PORT"],
+    type=int,
+    help="Port on which to run Kolibri services",
```
rtibbles, line 586: Note that this is incorrect - this should be `Port on which Kolibri is being served` - it's an informative setting, not one that is forcing behaviour :)

rtibbles (commented):
A few questions and things that could be tweaked. Mostly focused on the backend, but didn't see anything that raised my eyebrows on the frontend.

kolibri/core/discovery/test/test_network_search.py
micahscopes: I took this functionality out of the function being tested ... I'd planned on writing some tests for the caching function, but didn't get to it yet. So right now I don't think `TransactionTestCase` is even being used!

kolibri/core/discovery/api.py
micahscopes: It does work, not sure why it didn't seem to work the other way... I'll have to double check on it after deleting pyc files

kolibri/core/discovery/models.py
micahscopes: It kind of does. it's possible that a single kolibri instance (with a single instance ID) could be saved several times with different addresses (for example in moving between different network environments, or over various network interfaces).  Whereas with the dynamic addresses, we want them to be in a 1:1 relationship with `instance_id`, so this seemed like an okay way to support both in the same table.

Another option would be to introduce a more complex constraint for uniqueness and a more complex query system for dynamic vs. static interfaces. But that seemed unnecessary for now, this does the job as long as nobody starts saving static locations with their `instance_id` for an ID.

Is it too hacky?

kolibri/utils/cli.py
micahscopes: I see, so the port is not a prescription

kolibri/core/discovery/models.py
micahscopes: I should clarify, it's fine if the migrated saved location entries have numeric IDs, they just need to be unique with respect to `instance_id`s. Having instance ID's as a private key is reserved for dynamic entries.

Can you think of a simpler way?

kolibri/plugins/device/assets/src/views/ManageContentPage/SelectNetworkAddressModal/SelectAddressForm.vue
```diff
@@ -9,68 +9,90 @@
...
     @cancel="$emit('cancel')"
   >
-    <UiAlert
-      v-if="uiAlertProps"
-      v-show="showUiAlerts"
-      :type="uiAlertProps.type"
-      :dismissible="false"
-    >
-      {{ uiAlertProps.text }}
+    <template v-slot>
```
indirectlylit: is this intended to be set up to be a slot? I don't see it used as such anywhere

indirectlylit:
I'm not quite understanding the behavior I'm seeing here, but something seems a bit off:

* alternating between two states
* disabled radio buttons referring to two different 'localhost' instances

toggle

kolibri/utils/cli.py
rtibbles: It's prescriptive when it is passed to the `kolibri start` command, but `kolibri services`, does not run the server itself, and so can't set the port.

indirectlylit:
Seeing some strange behaviors importing from a local source.

In this case I'm running the devserver (UI shown) on `8000` and the pex on `8080`

| | |
|--|--|
|adding|image|

kolibri/plugins/device/assets/src/views/ManageContentPage/SelectNetworkAddressModal/SelectAddressForm.vue
```diff
@@ -9,68 +9,90 @@
...
     </template>
 
-    <template v-if="false">
-      {{ $tr('searchingText') }}
-    </template>
+    <KFixedGrid slot="actions" class="actions" numCols="4">
+      <KFixedGridItem span="1">
+        <transition name="fade">
+          <div v-if="discoveringPeers" class="searching-indicator">
+            <KCircularLoader :size="16" :stroke="9" />{{ $tr('searchingText') }}
```
indirectlylit: We have a `KLabeledIcon` component for consistent spacing in situations like this. Also lowered the stroke width a bit.

```suggestion
            <KLabeledIcon>
              <KCircularLoader slot="icon" :size="16" :stroke="6" />
              {{ $tr('searchingText') }}
            </KLabeledIcon>
```

| before | after |
|--|--|
| image | image|

indirectlylit (dismissed):
looking great!

Added a few notes in the comments

kolibri/core/discovery/models.py
micahscopes: FYI, I'm addressing this issue by making a better defined interface for `DynamicNetworkAddress`'s.

integration_testing/features/super-admin/super-admin-adding-a-new-network-address.feature
```diff
@@ -0,0 +1,16 @@
+# START testing this scenario with a FRESH DB (make a copy of the current if you want to reuse it later), and make sure to have a stable connection available.
+
+Background:
+  Given I am signed into Kolibri as a <superuser> or <devie admin>
```
jamalex: devie -> device

integration_testing/features/super-admin/super-admin-auto-discovering-new-network-addresses.feature
```diff
@@ -0,0 +1,18 @@
+# START testing this scenario with a FRESH DB (make a copy of the current if you want to reuse it later), and make sure to have a stable connection available.
+
+Background:
+  Given I am signed into Kolibri as a <superuser> or <devie admin>
```
jamalex: device

integration_testing/features/super-admin/super-admin-auto-discovering-new-network-addresses.feature
```diff
@@ -0,0 +1,18 @@
...
+Feature: Successful auto-discovery of new address
+  Admin's device is able to automatically discover and connect to nearby Kolibri peers
+
+  Scenario: Connecting to a Kolibri peer with content
+     Given that there are Kolibri peers around me
+      And the peers each have content available
+  	 When I open the 'select network address' modal and wait a few seconds
+  	 Then I should see a loading spinner accompanied with the word 'searching...'
+  	   And I should see a list of found local Kolibri peers below all the manually entered network addresses
+       And each peer should display their IP address and peer ID
```
jamalex: And port?

integration_testing/features/super-admin/super-admin-forgetting-a-manually-added-address.feature
```diff
@@ -0,0 +1,14 @@
+# START testing this scenario with a FRESH DB (make a copy of the current if you want to reuse it later), and make sure to have a stable connection available.
+
+Background:
+  Given I am signed into Kolibri as a <superuser> or <devie admin>
```
jamalex: device

integration_testing/features/super-admin/super-admin-no-peers-automatically-found-when-no-peers-around.feature
```diff
@@ -0,0 +1,17 @@
+# START testing this scenario with a FRESH DB (make a copy of the current if you want to reuse it later), and make sure to have a stable connection available.
+
+Background:
+  Given I am signed into Kolibri as a <superuser> or <devie admin>
```
jamalex: device

kolibri/core/discovery/utils/network/search.py
jamalex: >It handles caching and is debounced in order to prevent spamming the network.

If this is always called from a single process, there wouldn't be any network spamming, though... Listing the instances the listener has been tracking is a completely local, passive, in-memory operation.

kolibri/core/discovery/models.py
jamalex: The migration issue that @rtibbles mentions is a concern here (for existing data), as it could break upgrades if there are existing saved NetworkLocations with numeric IDs.

The need for maintaining uniqueness could be addressed by switching away from adding the DynamicNetworkLocation model, to a more transient cached list of peers -- including a general suggestion about this as a top-level comment.

kolibri/core/discovery/utils/network/search.py
jamalex: (calling from multiple processes, on the other hand, would initialize a fresh listener for each, which adds a bit of network traffic as it pings to ask who's out there when it first starts up -- my top-level comment will make a suggestion for avoiding this completely)

jamalex (changes requested):
Awesome work! I'm not fit to comment on the frontend stuff, so just some suggestions on the backend pieces:

The flow I'm seeing here is:

DynamicNetworkLocationViewSet.list -> run_peer_discovery -> get_peer_instances -> initialize_zeroconf_listener

which means that in a multi-process environment (e.g. uwsgi), we're creating a bunch of parallel zeroconf network listeners (one per process), and needing to have that 3 second lag when requesting the peer list, if that specific process hasn't previously been initialized as a listener.

The desired flow, to minimize overhead/duplication and increase responsiveness (basically, scan response would be instantaneous) is:

1. DynamicNetworkLocationViewSet.list just retrieves the list of DynamicNetworkLocations from the database. It doesn't call anything on the zeroconf side. The database here is essentially serving as our multiprocess-enabled peer cache.
2. The service code that initializes the listener would be started once at server startup, and would then keep the peer cache up to date, adding a peer to the cache when it is discovered, and removing from the cache when one leaves the network. When it starts up, it should flush out the cache and start from scratch.

This then removes any need for debouncing, additional caching, wait times, etc. It should "just work" (tm).

It may also be overkill to use a Django model as the cross-process cache, if all we need to store is a JSON blob that enumerates the peers, and can then be served up to the client code -- could use whatever the "cross process cache wrapper" is that @rtibbles mentions. That way it can leverage Redis if available as well. And avoids the complexities of deduplication/uniqueness/UUID pk's.

Let me know if it would be useful to talk this through -- can jump on a quick call!

kolibri/core/discovery/models.py
micahscopes: Thanks for this feedback!  The latest push uses `diskcache` instead (for now).

kolibri/core/discovery/models.py
micahscopes: this is no longer relevant

kolibri/core/discovery/utils/network/search.py
micahscopes: This is all taken care of now!  It works great to just let zeroconf handle all that.

jamalex:
Updates from discussions:
- Updating to have dynamic locations _not_ saved to the `NetworkLocation` model would have cascading effects throughout the content import process, as all the frontend code and API's throughout that process assume they're working with an instance of that model (doing lookups by `peer_id`). So, we concluded that it'd be best to revert to using the model as the cache, rather than the diskcache.
- The migration isn't a concern in practice, as it's not an actual UUID, just a Char, so it just turns the old int IDs into short strings, which works. @micahscopes will test across sqlite/postgres.

Longer-term (tech debt): The content import process shouldn't assume that it's working with a statically saved network location. Instead of peer_id, we can likely just use base_url in most cases. That will then allow a lot of things to be cleaned up and simplified.

micahscopes:
FYI, in addition to adding as many of the suggested changes as I could, I also made a pretty nice caching connection checker at `kolibri.core.discovery.utils.network.connections`

Here's what it does:
- gentle connection checking through some exhaustive yet hopefully still readable logic
- keeps track of attempted connections and received device info
- stores these exclusively by base_url in two CrossProcessCache objects
- presents all this information via a single interface called `check_connection(base_url)`

It has some pretty comprehensive unit tests to go with it!

kolibri/core/discovery/utils/network/search.py
```diff
@@ -106,54 +106,60 @@ def add_service(self, zeroconf, type, name):
         id = _id_from_name(name)
         ip = socket.inet_ntoa(info.address)
 
-        self.instances[id] = {
+        base_url = "http://{ip}:{port}/".format(ip=ip, port=info.port)
+        is_self = (
+            id == ZEROCONF_STATE["service"].id if ZEROCONF_STATE["service"] else False
+        )
```
indirectlylit: minor style note - can simplify to:

```python
is_self = ZEROCONF_STATE.get("service") == id
```

kolibri/core/discovery/models.py
```diff
@@ -42,42 +39,15 @@ class Meta:
...
-            available_recently and self.last_available > self.last_unavailable
-        )
-
-        if is_available:
-            return True
-        elif unavailable_recently:
-            return False
-        else:
-            return True if self.ping() else False
+        return True if check_connection(self.base_url) else False
```
jamalex: Could probably just be `bool(check_connection(self.base_url))` (depending on which feels cleaner to you)

kolibri/core/discovery/utils/network/connections.py
```diff
@@ -0,0 +1,109 @@
...
+from .client import NetworkClient
+from .urls import parse_address_into_components
+from kolibri.core.utils.cache import CrossProcessCache
+
+
+class Nothing:
+    def __nonzero__(self):
+        return False
+
+    __bool__ = __nonzero__  # this is for python3
```
indirectlylit: is thins something that can't be done with `None`?

kolibri/core/discovery/utils/network/search.py
```diff
@@ -107,7 +107,9 @@ def add_service(self, zeroconf, type, name):
         ip = socket.inet_ntoa(info.address)
 
         base_url = "http://{ip}:{port}/".format(ip=ip, port=info.port)
-        is_self = id == ZEROCONF_STATE["service"].id
+        is_self = (
+            id == ZEROCONF_STATE["service"].id if ZEROCONF_STATE["service"] else False
```
jamalex: Might be easier for a/this human to parse if there were parens around `id == ZEROCONF_STATE["service"].id`

indirectlylit (commented):
still testing. 

However when running, I get this error on startup:

```
Exception in thread zeroconf-ServiceBrowser_Kolibri._sub._http._tcp.local.:
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/threading.py", line 801, in __bootstrap_inner
    self.run()
  File "/Users/d/.pex/install/kolibri-0.13.0b4.dev0+git.7.g1e965215-py2.py3-none-any.whl.1c96f71514c7e69b0d22c18963272db7a4f3d938/kolibri-0.13.0b4.dev0+git.7.g1e965215-py2.py3-none-any.whl/kolibri/dist/zeroconf.py", line 1442, in run
    handler(self.zc)
  File "/Users/d/.pex/install/kolibri-0.13.0b4.dev0+git.7.g1e965215-py2.py3-none-any.whl.1c96f71514c7e69b0d22c18963272db7a4f3d938/kolibri-0.13.0b4.dev0+git.7.g1e965215-py2.py3-none-any.whl/kolibri/dist/zeroconf.py", line 1383, in <lambda>
    state_change=state_change,
  File "/Users/d/.pex/install/kolibri-0.13.0b4.dev0+git.7.g1e965215-py2.py3-none-any.whl.1c96f71514c7e69b0d22c18963272db7a4f3d938/kolibri-0.13.0b4.dev0+git.7.g1e965215-py2.py3-none-any.whl/kolibri/dist/zeroconf.py", line 1279, in fire
    h(**kwargs)
  File "/Users/d/.pex/install/kolibri-0.13.0b4.dev0+git.7.g1e965215-py2.py3-none-any.whl.1c96f71514c7e69b0d22c18963272db7a4f3d938/kolibri-0.13.0b4.dev0+git.7.g1e965215-py2.py3-none-any.whl/kolibri/dist/zeroconf.py", line 1352, in on_change
    listener.add_service(*args)
  File "/Users/d/.pex/install/kolibri-0.13.0b4.dev0+git.7.g1e965215-py2.py3-none-any.whl.1c96f71514c7e69b0d22c18963272db7a4f3d938/kolibri-0.13.0b4.dev0+git.7.g1e965215-py2.py3-none-any.whl/kolibri/core/discovery/utils/network/search.py", line 134, in add_service
    dict(base_url=base_url, **device_info), id=id,
  File "/Users/d/.pex/install/kolibri-0.13.0b4.dev0+git.7.g1e965215-py2.py3-none-any.whl.1c96f71514c7e69b0d22c18963272db7a4f3d938/kolibri-0.13.0b4.dev0+git.7.g1e965215-py2.py3-none-any.whl/kolibri/dist/django/db/models/manager.py", line 85, in manager_method
    return getattr(self.get_queryset(), name)(*args, **kwargs)
  File "/Users/d/.pex/install/kolibri-0.13.0b4.dev0+git.7.g1e965215-py2.py3-none-any.whl.1c96f71514c7e69b0d22c18963272db7a4f3d938/kolibri-0.13.0b4.dev0+git.7.g1e965215-py2.py3-none-any.whl/kolibri/dist/django/db/models/query.py", line 476, in update_or_create
    lookup, params = self._extract_model_params(defaults, **kwargs)
  File "/Users/d/.pex/install/kolibri-0.13.0b4.dev0+git.7.g1e965215-py2.py3-none-any.whl.1c96f71514c7e69b0d22c18963272db7a4f3d938/kolibri-0.13.0b4.dev0+git.7.g1e965215-py2.py3-none-any.whl/kolibri/dist/django/db/models/query.py", line 537, in _extract_model_params
    "', '".join(sorted(invalid_params)),
FieldError: Invalid field name(s) for model DynamicNetworkLocation: 'version'.
```

kolibri/core/discovery/utils/network/search.py
```diff
@@ -175,6 +177,11 @@ def initialize_zeroconf_listener():
         SERVICE_TYPE, ZEROCONF_STATE["listener"]
     )
 
+    DynamicNetworkLocation.objects.purge()
```
jamalex: May be better to put this purge up before we initialize the new listener, since that spins off a new thread, and we don't want race conditions with new stuff being cleared.

jamalex: Also: does `purge` benefit much from having its own method, or would `DynamicNetworkLocation.objects.all().delete()` be sufficient?

kolibri/deployment/default/settings/dev.py
```diff
@@ -12,9 +12,9 @@
 
 DEVELOPER_MODE = True
 
-# Create a dummy cache for each cache
+# Create a memcache for each cache
 CACHES = {
-    key: {"BACKEND": "django.core.cache.backends.dummy.DummyCache"}
+    key: {"BACKEND": "django.core.cache.backends.locmem.LocMemCache"}
```
jamalex, line 17: Flagging for @rtibbles for Monday

kolibri/core/discovery/utils/network/search.py
micahscopes: I have a fix for this!

kolibri/core/discovery/utils/network/connections.py
micahscopes: I have some updates and tests for this to make sure these survive pickling, as they'll be stored in a cache.  Essentially the reason for this class is to make the connection caching logic more readable, as it's hard enough to follow already.   Basically it let me store different kinds of falsey types in the cache as a way of keeping track different kinds of failures.  The alternative would be to store these values as strings and then make lots of long complicated boolean logic statements in the caching logic with a bunch of string comparison.  That would have given me confusion overload.  Example:
```
INVALID_DEVICE_INFO = Nothing("invalid device info")
FAILED_TO_CONNECT = Nothing("failed to connect")

# these are different
assert INVALID_DEVICE_INFO != FAILED_TO_CONNECT

# but both are falsey, so you can stuff like this:
result = INVALID_DEVICE_INFO

if result:
    return result  # <---- this will never happen
elif result == FAILED_TO_CONNECT:
    print("failed to connect!")
   # ... this won't happen either
elif result == INVALID_DEVICE_INFO:
    print("invalid device info!")
    # now we can handle this specific kind of lack of a value correctly

```

jamalex (approved):
Looks good to me! Added a few optional suggestions, and a spot for @rtibbles to check.

Manual testing looked good -- only issues I saw seemed not to be specific to the P2P stuff (resource size counts being off sometimes, and occasional zombie process happening when I CTRL-C a `kolibri start --foreground` server -- can't reliably replicate the latter, but worth watching for)

indirectlylit:
> occasional zombie process happening 

I'm seeing this also, though it might predate this PR. Filed an issue here: https://github.com/learningequality/kolibri/issues/6245

indirectlylit:
merging in so we can evaluate in `develop` - let's address any outstanding issues in follow-up PRs this week

PR #6241, opened 2019-12-07
webpack tree shaking

indirectlylit:
### Summary

while investigating tree shaking in https://github.com/learningequality/kolibri/pull/5972 attempted to apply optimizations from https://webpack.js.org/guides/tree-shaking/

### Reviewer guidance

Note that tree shaking might not actually be occurring as we expected. According to [the docs](https://webpack.js.org/guides/tree-shaking/) we could add either `sideEffects`, `usedExports`, or both.

After running this, most bundles stayed the same size but these two got a bit lower:

| bundle | size before | size after |
|--|--|--|
|`kolibri.core.default_frontend` | 1.82 MB| 1.56 MB |
| `kolibri.plugins.epub_viewer.main` |695 KB | 687 KB|

However, it's not clear to me whether this change is safe - specifically because we don't know whether our modules do in fact have side effects or not...

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/packages/kolibri-tools/lib/bundleStats.js
+++ b/packages/kolibri-tools/lib/bundleStats.js
@@ -26,7 +26,7 @@ function buildWebpack(data, index, startCallback, doneCallback, options) {
   return compiler;
 }
 
-const { data, index, start } = getEnvVars();
+const { data, index, options, start } = getEnvVars();
 
 function build() {
   buildWebpack(
@@ -37,7 +37,8 @@ function build() {
     },
     () => {
       process.send('done');
-    }
+    },
+    options
   );
 }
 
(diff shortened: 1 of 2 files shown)


indirectlylit:
@rtibbles curious what you think about this. Seems a bit risky for 260KB?

rtibbles:
Yeah, I'd only want to do this if we had a linting rule that flagged execution in module level code - we could then explicitly flag when it is OK for code to have side effects in the module scope, such as the entry points for modules.

One place I can imagine this being an issue would be e.g. when we use module level scoped variables in Vue component definitions as a way to have locally defined variables without defining them on the component.

PR #6214, opened 2019-12-04
check log location only when version<0.12.4

lyw07:
### Summary

This PR is to run the function that checks the log file location only when the version is lower than 0.12.4.

### Reviewer guidance

Run a version lower than 0.12.3 and check that the log files have been moved to the new location.

### References

https://github.com/learningequality/kolibri/issues/5532

----

### Contributor Checklist

PR process:

- [X] PR has the correct target branch and milestone
- [X] PR has 'needs review' or 'work-in-progress' label
- [X] If PR is ready for review, a reviewer has been added. (Don't use 'Assignees')
- [ ] If this is an important user-facing change, PR or related issue has a 'changelog' label
- [ ] If this includes an internal dependency change, a link to the diff is provided

Testing:

- [X] Contributor has fully tested the PR manually
- [ ] If there are any front-end changes, before/after screenshots are included
- [ ] Critical user journeys are covered by Gherkin stories
- [ ] Critical and brittle code paths are covered by unit tests

### Reviewer Checklist

- Automated test coverage is satisfactory
- PR is fully functional
- PR has been tested for [accessibility regressions](http://kolibri-dev.readthedocs.io/en/develop/manual_testing.html#accessibility-a11y-testing)
- External dependency files were updated if necessary (`yarn` and `pip`)
- Documentation is updated
- Contributor is in AUTHORS.md

PR #6253, opened 2019-12-09
Revert "check log location only when version<0.12.4"

rtibbles:
Reverts learningequality/kolibri#6214


rtibbles:
Ref: https://github.com/learningequality/kolibri/pull/6214#issuecomment-563309525

indirectlylit (approved):

Issue #6247, opened 2019-12-09
Database image is malformed, and zombie attacks


indirectlylit:
### Observed behavior

error during import

### Expected behavior

no errors

### User-facing consequences

unknown

### Errors and logs

```
ERROR    An error occurred during channel import: (sqlite3.DatabaseError) database disk image is malformed [SQL: "SELECT name FROM sqlite_master WHERE type='table' ORDER BY name"] (Background on this error at: http://sqlalche.me/e/4xp6)
ERROR    Job 06e506288730459588547b07cec1499e raised an exception: Traceback (most recent call last):
  File "/Users/d/Projects/le/k013x/kolibri/core/tasks/worker.py", line 69, in handle_finished_future
    result = future.result()
  File "/Users/d/Projects/le/k013x/.venv/lib/python2.7/site-packages/concurrent/futures/_base.py", line 422, in result
    return self.__get_result()
  File "/Users/d/Projects/le/k013x/.venv/lib/python2.7/site-packages/concurrent/futures/thread.py", line 62, in run
    result = self.fn(*self.args, **self.kwargs)
  File "/Users/d/Projects/le/k013x/kolibri/core/tasks/worker.py", line 222, in wrap
    raise e
DatabaseError: (sqlite3.DatabaseError) database disk image is malformed [SQL: "SELECT name FROM sqlite_master WHERE type='table' ORDER BY name"] (Background on this error at: http://sqlalche.me/e/4xp6)
```

### Steps to reproduce

unknown

### Context

0.13.0 beta4

rtibbles:
My suspicion is that this is most likely caused by zombie processes (also the pickle issue https://github.com/learningequality/kolibri/issues/6238 seems like it could be another symptom, with py2/3 crossover).

benjaoming:
We also have seen an issue with `dbbackup` during an upgrade in 0.12.8: #6201 

As Richard notes: This was produced in a development, any chance that virtualenvs have switched between Python 2.7 and 3.x on the same `.kolibri` data?

indirectlylit:
yes, it is possible

indirectlylit:
Similarly reported by @jamalex during the bug bash:

https://www.notion.so/learningequality/Database-disk-image-is-malformed-when-importing-channel-created-and-recently-published-on-Studio-e668a9dfd5f24ab693295a06311a1cc1

> Restarting the server fixed this (`sudo service kolibri restart`).
> 
[...]

Issue #6201, opened 2019-12-03
Better exception message for malformed db (especially in initialize => dbbackup)


benjaoming:
### Observed behavior

An issue was reported here: https://community.learningequality.org/t/service-fails-at-startup/1337

We can see that `initialize` is called during an upgrade which leads to `dbbackup` producing the malformed db error outside of where it's normally anticipated.

### Expected behavior

1. A better failure message pointing to our updated troubleshooting section: https://kolibri.readthedocs.io/en/latest/manage.html#troubleshoot-database-issues
1. Catching the malformed db error in `dbbackup` or `initialize`.

### User-facing consequences

Bad advice making it harder to troubleshoot the issue.

### Errors and logs

```
Nov 28 15:15:50 raspberrypi kolibri[1390]: File “/usr/lib/python3/dist-packages/kolibri/utils/cli.py”,
Nov 28 15:15:50 raspberrypi kolibri[1390]: backup = dbbackup(version)
Nov 28 15:15:50 raspberrypi kolibri[1390]: File "/usr/lib/python3/dist-packages/kolibri/core/deviceadmi
Nov 28 15:15:50 raspberrypi kolibri[1390]: for line in db.connections[“default”].connection.iterdump(
Nov 28 15:15:50 raspberrypi kolibri[1390]: File “/usr/lib/python3.7/sqlite3/dump.py”, line 56, in _iter
Nov 28 15:15:50 raspberrypi kolibri[1390]: for row in query_res:
Nov 28 15:15:50 raspberrypi kolibri[1390]: sqlite3.DatabaseError: database disk image is malformed
Nov 28 15:15:51 raspberrypi systemd[1]: kolibri.service: Control process exited, code=exited, status=1/FA
Nov 28 15:15:51 raspberrypi systemd[1]: kolibri.service: Failed with result ‘exit-code’.
Nov 28 15:15:51 raspberrypi systemd[1]: Failed to start LSB: kolibri daemon, an offline education platfor
lines 1-15/15 (END)
```

### Steps to reproduce

Not sure!

### Context

Not sure, but definitely 0.13 relevant.

radinamatic:
Updated documentation link for [the steps to repair a malformed database](https://kolibri.readthedocs.io/en/latest/manage/get_support.html#malformed-db).

rtibbles:
Documentation was updated, and we now attempt to do automatic recovery.

PR #6273, opened 2019-12-10
Verify that database backend isn't loaded when invoking status and stop

benjaoming:
### Summary

CC: @rtibbles @jredrejo @jamalex @ralphiee22 

Let's ensure that `stop` and `status` do not cause the database to be loaded since this can happen during upgrades.

In this PR, I propose that a simple solution could be to inject an invalid setting for the database backend to catch any unnecessary access. It already fails, so we need to investigate how that can be altered -- or if it can't, I can propose different ways to fail upon database access during stop/status commands.

Also of interest: The `manage shell` command is important to be able to access despite database issues.

The risk of accessing the database during stop/status is that a different database schema is loaded or different versions of Python libraries are applied on a SQLite database file used by another process. This could be a reason for the malformed database issues.

### Reviewer guidance

* Do you know if the currently failing database backend loading can be avoided?

### References

#6247 
#6201 

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -60,6 +60,11 @@ matrix:
       sudo: true
       env:
         - TOX_ENV=py3.7
+    - python: 3.7
+      dist: xenial
+      sudo: true
+      env:
+        - TOX_ENV=illegaldbaccess
     - python: "2.7"
       env:
         - TOX_ENV=node10.x
--- a/kolibri/core/tasks/main.py
+++ b/kolibri/core/tasks/main.py
@@ -3,12 +3,32 @@ import os
 from sqlalchemy import create_engine
 from sqlalchemy import event
 from sqlalchemy import exc
+from sqlalchemy.engine.interfaces import Connectable
+from sqlalchemy.engine.interfaces import Dialect
 from sqlalchemy.pool import NullPool
 
 from kolibri.core.tasks.queue import Queue
 from kolibri.core.tasks.scheduler import Scheduler
 from kolibri.core.tasks.worker import Worker
 from kolibri.utils import conf
+
+
+class InvalidEngine(Dialect, Connectable):
+    def __init__(self, *args, **kwargs):
+        self.name = "invalid engine"
+        self.dispatch = self
+        self.checkout = self.invalid_property
+        self.connect = self.invalid_property
+        self._listen = self.ignored_property
+        self._adjust_fn_spec = self.ignored_property
+
+    def invalid_property(self, *args, **kwargs):
+        raise NotImplementedError("This engine is not supported")
+
+    def ignored_property(self, *args, **kwargs):
+        return None
+
+    invalid_property._adjust_fn_spec = ignored_property
 
 
 if conf.OPTIONS["Database"]["DATABASE_ENGINE"] == "sqlite":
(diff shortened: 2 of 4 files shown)


benjaoming:
Fixed: tox.ini cannot add an environment variable a single invocation of a command, needs it for the full test env. Therefore, added `dbaccess` to the test matrix.

rtibbles:
Looking at the test run, it seems like this will always fail, because it invokes start, which will always need to access the DB.

benjaoming:
> Looking at the test run, it seems like this will always fail, because it invokes start, which will always need to access the DB.

Ah no that should not be the case, I don't think it needs to start Kolibri before invoking status/stop.

benjaoming:
@rtibbles you can run this PR locally like this:

```
KOLIBRI_DATABASE_ENGINE="raise.importerror.if.loaded" kolibri status
```

This currently renders the following error which I find highly suspicious:

```
WARNING:root:No C Extensions available for this platform.

INFO     Option DATABASE_ENGINE in section [Database] being overridden by environment variable KOLIBRI_DATABASE_ENGINE
  File "/home/user/.virtualenvs/kolibri/bin/kolibri", line 11, in <module>
  File "/home/user/.virtualenvs/kolibri/lib/python3.6/site-packages/pkg_resources/__init__.py", line 489, in load_entry_point
    return get_distribution(dist).load_entry_point(group, name)
  File "/home/user/.virtualenvs/kolibri/lib/python3.6/site-packages/pkg_resources/__init__.py", line 2843, in load_entry_point
    return ep.load()
  File "/home/user/.virtualenvs/kolibri/lib/python3.6/site-packages/pkg_resources/__init__.py", line 2434, in load
    return self.resolve()
  File "/home/user/.virtualenvs/kolibri/lib/python3.6/site-packages/pkg_resources/__init__.py", line 2440, in resolve
    module = __import__(self.module_name, fromlist=['__name__'], level=0)
  File "/code/kolibri/kolibri/utils/cli.py", line 34, in <module>
    from kolibri.core.deviceadmin.utils import IncompatibleDatabase
  File "/code/kolibri/kolibri/core/deviceadmin/utils.py", line 13, in <module>
    from kolibri.core.tasks.main import scheduler
  File "/code/kolibri/kolibri/core/tasks/main.py", line 85, in <module>
    priority_queue = Queue(priority_queue_name, connection=connection)
  File "/code/kolibri/kolibri/core/tasks/queue.py", line 13, in __init__
    self.storage = Storage(connection)
  File "/code/kolibri/kolibri/core/tasks/storage.py", line 58, in __init__
    self.Base.metadata.create_all(self.engine)
  File "/home/user/.virtualenvs/kolibri/lib/python3.6/site-packages/sqlalchemy/sql/schema.py", line 4005, in create_all
    tables=tables)
  File "/home/user/.virtualenvs/kolibri/lib/python3.6/site-packages/sqlalchemy/engine/interfaces.py", line 1165, in _run_visitor
    raise NotImplementedError()
NotImplementedError
```

indirectlylit:
> The risk of accessing the database during stop/status is that a different database schema is loaded or different versions of Python libraries are applied on a SQLite database file used by another process. This could be a reason for the malformed database issues.

@benjaoming have you reproduced this scenario?

rtibbles:
The suspicious lines can probably be resolved by doing some lazy instantiation, but also it is not accessing the default DB.

rtibbles:
Superseded by #6281

PR #6281, opened 2019-12-11
Make objects that access the database lazy to prevent DB access only during import

rtibbles:
### Summary
Drawing inspiration from #6273 adds two tests to verify that neither the Django DB backend or the SQLAlchemy backends are accessing the database during the status and stop commands

Makes queue and scheduler object instantiation lazy to prevent inadvertent DB access.

### Reviewer guidance
Code make sense?

Note that I did not have to make any modifications to prevent access to the Django default DB.

### References
Supersedes #6273

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/utils/tests/test_cli_at_import.py
+++ b/kolibri/utils/tests/test_cli_at_import.py
@@ -1,5 +1,8 @@
 """
 Tests for `kolibri.utils.cli` module.
+These tests deliberately omit `@pytest.mark.django_db` from the tests,
+so that any attempt to access the Django database during the running
+of these cli methods will result in an error and test failure.
 """
 from __future__ import absolute_import
 from __future__ import print_function
--- a/kolibri/utils/tests/test_cli_at_import.py
+++ b/kolibri/utils/tests/test_cli_at_import.py
@@ -1,0 +1,35 @@
+"""
+Tests for `kolibri.utils.cli` module.
+"""
+from __future__ import absolute_import
+from __future__ import print_function
+
+from mock import patch
+
+
+@patch("sqlalchemy.create_engine")
+def test_status_no_db_access(create_engine_mock):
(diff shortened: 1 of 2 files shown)


benjaoming:
@rtibbles this is a great solution!

@indirectlylit - regarding https://github.com/learningequality/kolibri/pull/6273#issuecomment-564669557

> @benjaoming have you reproduced this scenario?

No I haven't - but I saw two cases where the malformed DB had occurred during an upgrade. These two CLI calls should not at all access the DB because of the implied multi-process risks.. so I am quite optimistic that we may have found something.

Especially also since the error occurs in new code? It's pertaining some load-time `scheduler` stuff - @rtibbles would you suspect that this is a new behavior?

benjaoming (changes requested):
In order to plug this hole once and for all, I was wondering if we can have the same test added for Django backend access?

We could also squash all of the various assertions in to one long test case? Calling `status`, `stop` and perhaps even `manage shell`?

benjaoming (approved):
@rtibbles gave me the tip that the lack of `@pytest.mark.djangodb` implicitly already guarantees that some of our tests don't hit the Django db backend :+1:

kolibri/utils/tests/test_cli_at_import.py
```diff
@@ -0,0 +1,35 @@
+"""
+Tests for `kolibri.utils.cli` module.
```
rtibbles, line 2:
```suggestion
Tests for `kolibri.utils.cli` module.
These tests deliberately omit `@pytest.mark.django_db` from the tests,
so that any attempt to access the Django database during the running
of these cli methods will result in an error and test failure.
```

PR #6292, opened 2019-12-12
closing db connections after the info is retrieved

### LINKED ISSUE #6245: occasional zombie processes related to unclean shutdown
### Observed behavior

observed by @jamalex and @indirectlylit during testing of zerofconf work

> occasional zombie process happening when I CTRL-C a `kolibri start --foreground` server

This caused me some very confusing issues when I had closed all my terminals and Kolibri was still running in the background

### Expected behavior

Clean and complete shutdown

### User-facing consequences

Could get very difficult-to-debug behaviors if zombie processes are accessing DBs

### Steps to reproduce

unknown

### Context

0.13.0 beta 4

jredrejo:
### Summary
Ensure to close the db connections when they are needed to avoid locks that can keep process stalled

### Reviewer guidance
are there any zombies around?

### References
Hopefully it will close #6245 

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/core/discovery/test/test_network_search.py
+++ b/kolibri/core/discovery/test/test_network_search.py
@@ -1,7 +1,7 @@
 import socket
 
 import mock
-from django.test import TestCase
+from django.test import TransactionTestCase
 from zeroconf import BadTypeInNameException
 from zeroconf import service_type_name
 from zeroconf import ServiceInfo
@@ -106,7 +106,7 @@ class MockZeroconf(Zeroconf):
     lambda: [MOCK_INTERFACE_IP],
 )
 @mock.patch("django.db.models.Manager.update_or_create")
-class TestNetworkSearch(TestCase):
+class TestNetworkSearch(TransactionTestCase):
     def test_initialize_zeroconf_listener(self, *mocks):
         assert ZEROCONF_STATE["listener"] is None
         initialize_zeroconf_listener()
--- a/kolibri/core/discovery/test/test_network_search.py
+++ b/kolibri/core/discovery/test/test_network_search.py
@@ -98,6 +98,10 @@ class MockZeroconf(Zeroconf):
 
 
 @mock.patch(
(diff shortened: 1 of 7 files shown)


micahscopes (approved):
This looks great, thanks for the catch @jredrejo :tada:

micahscopes:
I initially approved this but wanted @jamalex to take a look and make sure we are doing this in an optimal way, since the zeroconf listener functions might be called pretty often if the Kolibri instance is on a LAN full of other Kolibri instances.

(also, I tested this out locally and found unfortunately that it didn't expel the zombie kolibris)

jredrejo:
> I initially approved this but wanted @jamalex to take a look and make sure we are doing this in an optimal way, since the zeroconf listener functions might be called pretty often if the Kolibri instance is on a LAN full of other Kolibri instances.
> 
> (also, I tested this out locally and found unfortunately that it didn't expel the zombie kolibris)

It's good if more people checks this PR, I haven't looked deeply enough the zeroconf code and it's better to be sure nothing is broken.
However, keeping the db open is a very bad idea. It can't be the solution if the function is called very often. Caching or other alternatives like not using sqlite should be considered if that is the case.
Keeping a file based db open will produce many problems we have seen in the past:

- Zombies is the least important effect
- Journal grows forever, the connection keeps the db locked and writing is not possible, writting back the data in the journal never happens. Actually I remember asking to check the journal was kept in a previous bug bash session and that's how I saw today quickly these problems. If the db is correctly managed, no journal files should be present after a request is responsed.
- Data corruption while different process try to write in the db
- Performance degrades quickly...

So, this PR solves problems for sure. In case it p roduces others we should find out a different solution. 

Also, I could not see more zombies after applying this PR, but  I saw them immediately. Could you tell me how to reproduce them? There might be more problems producing them.

PR #6305, opened 2019-12-14
send demographic data in pingback

ralphiee22:
### Summary

Calculate demographic statistics around age and gender.
Send these statistics in the pingback to nutrition facts.

### Reviewer guidance

Can fill in demographic stats for users and observe the output of the pingback.
Is variance calculated correctly? do comments/queries make sense?

### References

https://www.notion.so/learningequality/Demographics-meta-data-6360e4fd277a44f2b809c0d89a7e9690

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated


kolibri/core/analytics/utils.py
```diff
@@ -62,6 +73,133 @@
...
+    queryset = FacilityUser.objects.all()
+    if learner:
+        roles_filter = Q(roles__isnull=True)
+    # handle stats at facility level
+    if dataset_id:
+        queryset = queryset.filter(roles_filter, dataset_id=dataset_id)
+    # handle stats at channel level
+    if channel_id:
+        queryset = queryset.filter(
+            roles_filter, contentsummarylog__channel_id=channel_id
```
rtibbles: Could you not just filter by the `roles_filter` once rather than doing it in each conditional?

kolibri/core/analytics/utils.py
```diff
@@ -62,6 +73,133 @@
...
+                birth_year_int=Cast("birth_year", output_field=IntegerField())
+            ).values_list("birth_year_int", flat=True)
+        )
+
+        demographic_stats = (
+            queryset
+            # convert birth year to integer
+            .annotate(
+                birth_year_int=Cast("birth_year", output_field=IntegerField())
+            ).aggregate(
```
rtibbles: In the future, we should update this to loop through all the genders specified in the queryset in order to generate an aggregate for each, in case we update the interface to allow for gender self selection.

kolibri/core/analytics/utils.py
```diff
@@ -62,6 +73,133 @@
...
+        )
+        # payload of birth year and gender statistics
+        stats = {
+            "bys": {
+                "a": demographic_stats["average_birth_year"],
+                "v": variance(list_of_birth_years),
+                "ts": demographic_stats["total_specified_birth_year"],
+                "d": demographic_stats["num_deferred_birth_year"],
+            },
+            "gs": {
```
rtibbles: With the comment above in mind, we would need to transmit the full gender description here, in order to not require a pre-specified gender on nutritionfacts - might be best to implement this part now and break the abbreviation convention here.

rtibbles (dismissed):
I think we should do some slight tweaks here and on the nutritionfacts side in order to future proof with user selected gender categories in mind.

kolibri/core/analytics/utils.py
```diff
@@ -62,6 +73,133 @@
...
+                    ),
+                    output_field=FloatField(),
+                ),
+            )
+        )
+        # payload of birth year and gender statistics
+        stats = {
+            "bys": {
+                "a": demographic_stats["average_birth_year"],
+                "v": variance(list_of_birth_years),
```
lyw07: have a question that's not related to the code implementation - just wondering why we decided to use variance instead of standard deviation here. thank you!

indirectlylit: It was chosen arbitrarily. Reporting standard deviation instead might be preferable because it's easier to interpret (same units as mean/average)

@rtibbles @ralphiee22 perhaps switch to SD?

rtibbles: SD seems fine to me.

lyw07: thank you for the explanation!

kolibri/core/analytics/test/test_utils.py
```diff
@@ -201,6 +216,21 @@ def test_extract_facility_statistics(self):
...
+            "dsl": {
+                "bys": {
+                    "a": mean(birth_year_list_learners),
+                    "sd": standard_deviation(birth_year_list_learners),
+                    "ts": 20,
+                    "d": 0,
+                },
+                "gs": {
+                    gender: {
+                        "count": FacilityUser.objects.filter(gender=gender).count()
```
jamalex: Do we need the extra level of embedding here, or could it just be `gender: FacilityUser.objects.filter(gender=gender).count()`?

kolibri/core/analytics/utils.py
```diff
@@ -62,6 +75,128 @@
...
+        # return population std
+        return round(math.sqrt(sum((d - m) ** 2 for d in data) / len(data)), 2)
+    return None
+
+
+def calculate_demographic_stats(dataset_id=None, channel_id=None, learner=True):
+    stats = {}
+    roles_filter = Q(roles__isnull=False)
+    if learner:
+        roles_filter = Q(roles__isnull=True)
```
jamalex: Could combine the 3 lines above into  `roles_filter = Q(roles__isnull=learner)`, and perhaps add a comment (to clarify the `learner` param).

kolibri/core/analytics/utils.py
```diff
@@ -62,6 +75,128 @@
...
+        queryset = queryset.filter(dataset_id=dataset_id)
+    # handle stats at channel level
+    if channel_id:
+        user_ids = (
+            queryset.filter(contentsummarylog__channel_id=channel_id)
+            .distinct()
+            .values_list("id", flat=True)
+        )
+        # pass distinct user_ids as subquery
+        queryset = FacilityUser.objects.filter(id__in=user_ids)
```
jamalex, line 105: What happened when you did this without the subquery? (i.e. just filtering `queryset` directly)

kolibri/core/analytics/utils.py
```diff
@@ -62,6 +75,128 @@
...
+
+    # calculate stats if there are USER_THRESHOLD users or more
+    if queryset.count() >= USER_THRESHOLD:
+        # get list of all birth years to calculate variance
+        list_of_birth_years = (
+            queryset.exclude(
+                Q(birth_year="")
+                | Q(birth_year=demographics.NOT_SPECIFIED)
+                | Q(birth_year=demographics.DEFERRED)
+            )
```
jamalex: Could be interesting to benchmark (at some point) against doing `birth_year__in=["", ...]` instead.

kolibri/core/analytics/utils.py
```diff
@@ -62,6 +75,128 @@
...
+            # convert birth year to integer
+            .annotate(
+                birth_year_int=Cast("birth_year", output_field=IntegerField())
+            ).values_list("birth_year_int", flat=True)
+        )
+
+        demographic_stats = (
+            queryset
```
jamalex: Do we have a sense of how this mega-query scales with larger databases? Would be good to be sure we're not going to lock up databases with a really long-running query before we ship.

kolibri/core/analytics/utils.py
```diff
@@ -62,6 +75,128 @@
...
+                    output_field=FloatField(),
+                ),
+            )
+        )
+        # calculate all gender counts
+        gender_counts = queryset.values("gender").annotate(count=Count("gender"))
+        # payload of birth year and gender statistics
+        stats = {
+            "bys": {
+                "a": demographic_stats["average_birth_year"],
```
jamalex: As we're already calculating the mean of the birth years inside `standard_deviation`, might be more efficient to get it from there (e.g. change to `mean_and_standard_deviation`) and then not need to also compute it at the DB level.

kolibri/core/analytics/utils.py
```diff
@@ -62,6 +75,128 @@
...
+                "sd": standard_deviation(list_of_birth_years),
+                "ts": demographic_stats["total_specified_birth_year"],
```
jamalex: Similarly, this is already calculated inside `standard_deviation`, so it would be more efficient to use that than compute it twice.

kolibri/core/analytics/utils.py
```diff
@@ -62,6 +75,128 @@
...
+        )
+
+        demographic_stats = (
+            queryset
+            # convert birth year to integer
+            .annotate(
+                birth_year_int=Cast("birth_year", output_field=IntegerField())
+            ).aggregate(
+                # count of deferred users
+                num_deferred_gender=Coalesce(
```
jamalex: Where does this get used?

kolibri/core/analytics/utils.py
jamalex: Based on my notes below, I think this query could largely (completely?) be bypassed, as we're computing these stats separately anyway and could just use them from there.

kolibri/core/analytics/utils.py
```diff
@@ -62,6 +75,128 @@
...
+        )
+        # calculate all gender counts
+        gender_counts = queryset.values("gender").annotate(count=Count("gender"))
+        # payload of birth year and gender statistics
+        stats = {
+            "bys": {
+                "a": demographic_stats["average_birth_year"],
+                "sd": standard_deviation(list_of_birth_years),
+                "ts": demographic_stats["total_specified_birth_year"],
+                "d": demographic_stats["num_deferred_birth_year"],
```
jamalex: We already have total numbers of learners/non-learners in the facility/channel stats, so this can be computed later by subtracting.

jamalex (changes requested):
Nice work! My comments are mainly about efficiency, and I think we could remove the big query and still get all the stats we need, if I'm reading it right (since it looks like we couldn't do stdev at the DB-level anyway?)

kolibri/core/analytics/utils.py
rtibbles: Deferred is different to 'not even asked' though - there could be user accounts that have never been accessed, and so have not been deferred.

jamalex: Ah, yeah -- but perhaps we could calculate that directly or with a simpler query, if it's the only thing needed from the above aggregation?

kolibri/core/analytics/test/test_utils.py
ralphiee22: This structure was in case we added some more statistics in the future.

kolibri/core/analytics/utils.py
ralphiee22: Before I was using `distinct`, because filtering by the related `ContentSummaryLog` would return redundant user models, which was also throwing the queries below off because it would distinct the birth years, so this is the workaround to getting unique users.

kolibri/core/analytics/utils.py
ralphiee22: residual code

kolibri/core/analytics/utils.py
jamalex: Note that the old code included both "deferred" **and** "not specified" in this count. I'm assuming we only want deferred, so I'll update the behavior.

kolibri/core/analytics/utils.py
jamalex: Result: same speed, but a bit less code.

jamalex:
I've posted my updates here, as a PR against @ralphiee22's branch:
https://github.com/ralphiee22/kolibri/pull/10

jamalex (approved):

kolibri/core/analytics/utils.py
indirectlylit: correct

Issue #6143, opened 2019-11-25
throttle class summary API requests from navigation events in coach


indirectlylit:
### Observed behavior

During 0.12.x, we decided to re-request class summary information "just in case" our diff-based polling didn't work to keep the coach reports up-to-date.

This should not actually be necessary, and incurs extra cost on both network and server CPU

### Expected behavior

After initially loading coach reports, all subsequent updates should be driven by just the polling for events.

### User-facing consequences

slower performance

### Errors and logs

none

### Steps to reproduce

navigate around coach and observe frequent requests to class summary endpoint

### Context

0.13.0 beta 1

indirectlylit:
It sounds like the risk of regressions if we were to do this is somewhat high, because there are fair number of changes that are not pushed through our event notification system. These include things like renames of items.

For this issue, recommend wrapping the call in [`throttle`](https://lodash.com/docs/#throttle) with a long time, e.g. 30 seconds. This will prevent a barrage of reloads during rapid navigation, while still giving opportunity to grab the latest data.

Longer-term, we should design "change events" more holistically, and allow all parts of the Kolibri app to subscribe to update events of certain data views based on some kinds of filters. (graphql?)

Issue #4811, opened 2019-01-28
reduce unnecessary calls in coach tools


indirectlylit:
### Observed behavior

see for example: https://github.com/learningequality/kolibri/pull/4797#discussion_r251530072

Many of the coach tools 'plan' pages are still using older vuex actions and state. However some of that is now redundant with the `classSummary` module, which is available in every coach page.

### Expected behavior

We should remove unnecessary API calls

### User-facing consequences

faster page load

### Context

0.12

Issue #5757, opened 2019-07-09
investigate a lighter-weight API for polling learner classroom status


indirectlylit:
### Observed behavior

In https://github.com/learningequality/kolibri/pull/5737 we added polling to refresh learners' classes.

### Expected behavior

We should investigate performance impact of this polling, and if necessary create a dedicated, lighter-weight API for the polling endpoint, or find ways to optimize the existing endpoint.

### User-facing consequences

potentially degraded performance

### Steps to reproduce

Load test with many users polling simultaneously

### Context

0.12.5

rtibbles:
One possibility would be to add an etag or last modified to the endpoint if we could calculate this efficiently.

PR #6319, opened 2019-12-15
[DRAFT] add a 15-second throttle to navigation-based refresh of class summary

indirectlylit:
### Summary

adds a throttle to navigation-based class summary refreshing to avoid hitting the expensive endpoint unnecessarily often

### Reviewer guidance

does anything about the mixture of  `lodash.throttle`, promises, and vuex actions look wrong?

### References

* fixes https://github.com/learningequality/kolibri/issues/6143
* fixes https://github.com/learningequality/kolibri/issues/4811
* refs https://github.com/learningequality/kolibri/issues/5757

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/plugins/coach/assets/src/modules/classSummary/index.js
+++ b/kolibri/plugins/coach/assets/src/modules/classSummary/index.js
@@ -2,6 +2,7 @@ import get from 'lodash/get';
 import set from 'lodash/set';
 import flatten from 'lodash/flatten';
 import find from 'lodash/find';
+import throttle from 'lodash/throttle';
 
 import Vue from 'kolibri.lib.vue';
 import ClassSummaryResource from '../../apiResources/classSummary';
@@ -423,5 +424,8 @@ export default {
     refreshClassSummary(store) {
       return store.dispatch('loadClassSummary', store.state.id);
     },
+    refreshClassSummaryThrottled: throttle(store => {
+      return store.dispatch('refreshClassSummary');
+    }, 15000),
   },
 };
--- a/kolibri/plugins/coach/assets/src/modules/pluginModule.js
+++ b/kolibri/plugins/coach/assets/src/modules/pluginModule.js
@@ -106,12 +106,11 @@ export default {
         ]).catch(error => {
           store.dispatch('handleError', error);
         });
-      } else {
-        // otherwise refresh but don't block
-        return store
-          .dispatch('classSummary/loadClassSummary', classId)
-          .catch(error => store.dispatch('handleApiError', error));
       }
+      // navigating within the same class, periodically refresh but don't block
+      return store
+        .dispatch('classSummary/refreshClassSummaryThrottled')
+        .catch(error => store.dispatch('handleApiError', error));
     },
   },
   modules: {


indirectlylit:
**do not merge**

I am noticing that this causes a regression, where quizzes do not show up immediately after creating them. @jonboiser I notice in many places we explicitly call `refreshClassSummary` after user actions. Is this a situation where we would need to audit for that sort of explicit refresh?

rtibbles:
We definitely do an explicit refresh when a lesson is created or updated, so we should be doing similarly for quizzes and groups. This would probably allow the navigation based refresh to be dropped all together. (Although it doesn't get around the issue of simultaneous editing).

indirectlylit:
> This would probably allow the navigation based refresh to be dropped all together. (Although it doesn't get around the issue of simultaneous editing).

We will need occasional updates to handle facility level changes to class roster until we implement event signals for those changes also

rtibbles:
Right - I had mentally included such things in facility editing :)

kolibri/plugins/coach/assets/src/modules/classSummary/index.js
```diff
@@ -423,5 +424,8 @@ export default {
     refreshClassSummary(store) {
       return store.dispatch('loadClassSummary', store.state.id);
     },
+    refreshClassSummaryThrottled: throttle(store => {
+      return store.dispatch('refreshClassSummary');
+    }, 15000),
```
rtibbles, line 429: I feel like invoking the throttle with `{leading: true, trailing: false}` produces the most easily understandable results here, but either will reduce the invocations.

kolibri/plugins/coach/assets/src/modules/pluginModule.js
```diff
@@ -106,12 +106,11 @@ export default {
...
           store.dispatch('handleError', error);
         });
-      } else {
-        // otherwise refresh but don't block
-        return store
-          .dispatch('classSummary/loadClassSummary', classId)
-          .catch(error => store.dispatch('handleApiError', error));
       }
+      // navigating within the same class, periodically refresh but don't block
+      return store
```
rtibbles, line 111: The other issue here is that returning the promise does seem to block navigation until the Promise is resolved.

kolibri/plugins/coach/assets/src/modules/classSummary/index.js
indirectlylit: I think the default is `{leading: true, trailing: true}`, which seemed best to me because it means you'll see recent changes as soon as possible after a navigation event, but also always get the correct information eventually, after a sequence of rapid navigation.

what do you mean by "the most easily understandable results"?

rtibbles: Just that the throttling behaviour you get is most easily understood (to me at least), in terms of which invocations are executed, and which skipped.

indirectlylit: Yeah, that's what I'd expect from `debounce` for sure. But for `throttle` I think leading and trailing is almost always more correct.

The function is called on the trailing edge _only_ if the original function was called more than once. It essentially ensures that all requests for new information are eventually accommodated, but batches of > 2 are collapsed to 2.

indirectlylit:
closing due to outstanding issues

Issue #4834, opened 2019-01-29
confusing learner flow through quiz


indirectlylit:
### Observed behavior

In this situation, there are 20 questions but only 7 are visible and there is no visible scrollbar:

> ![image](https://user-images.githubusercontent.com/2367265/51936399-f43d6980-23bc-11e9-952a-fc3ea73e8b99.png)

from @laurenlichtman:

> I’m on question 6 but the number of questions didn’t scroll on the right so it was unclear that there was more than 5 questions (besides that it says 5 out of 15 questions). I didn’t realize that I would need to scroll if I wanted to jump around the questions 

from @micahscopes, slightly related:

> Micah almost clicked “submit quiz” after the first question

### Expected behavior

quiz-taking process should be totally fool-proof

### User-facing consequences

learners may do worse on quizzes than they would have otherwise

### Steps to reproduce

take a quiz with 15 or 20 items in it

### Context

0.12

khangmach:
Without redesigning this entire interface (WIP/future milestone), simplest solution seems to be to just make the question list scrollbar visible upon page load.

laurenlichtman:
@khangmach I think that sounds like a really good solution!

indirectlylit:
this works for desktop, but tablets don't show a scrollbar until you start scrolling

rtibbles:
One thing I have implemented in the detail reports is to automatically scroll the side nav bar down as you click through (so that the item you are on will always be put as close to the middle of the side nav bar as possible). I think this would at least help to always show there are more questions to be completed.

rtibbles:
![scroll-side-nav](https://user-images.githubusercontent.com/1680573/52240989-644d6300-2887-11e9-9d8c-627f9c7344a7.gif)

indirectlylit:
I like that a lot

The sudden transition and the fact that all the questions are named so similarly makes it a little confusing though. It almost looks as though you clicked the bottom item, but the item two rows up got selected?

indirectlylit:
nice clap BTW :)

rtibbles:
Yeah - should probably update this to use the Element `scrollTo` method, and set `smooth` on the scrolling behaviour.

PR #6321, opened 2019-12-15
learner quiz view updates

indirectlylit:
### Summary

* uses real buttons instead of divs with click handlers, enabling keyboard navigation
* updated to use design system colors, spacing, and icon component
* prevent the entire component from reloading when the user is on the same quiz, which prevents scroll reset behavior
* when possible (supported by the browser), smoothly scroll questions into view as necessary

| before | after |
|--|--|
|||

### Reviewer guidance

This change is a bit more significant than I would really like at this stage, but it has some big improvements so let's give it test.

It also uses some hacks for a child component to get access to a parent component scrolling div which I don't really like

### References

fixes https://github.com/learningequality/kolibri/issues/4834

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/plugins/learn/assets/src/views/ExamPage/AnswerHistory.vue
+++ b/kolibri/plugins/learn/assets/src/views/ExamPage/AnswerHistory.vue
@@ -1,27 +1,28 @@
 <template>
 
-  <div :style="{ backgroundColor: $themeTokens.surface }">
+  <div>
     <ul class="history-list">
-      <template v-for="(question, index) in questions">
-        <li
-          :key="index"
-          :style="liStyle(index)"
+      <li
+        v-for="(question, index) in questions"
+        :key="index"
+        class="list-item"
+      >
+        <button
+          :class="$computedClass(liStyle(index))"
+          :disabled="questionNumber === index"
           class="clickable"
           @click="$emit('goToQuestion', index)"
         >
-          <svg class="item svg-item">
-            <circle
-              cx="32"
-              cy="32"
-              r="8"
-              :style="{ fill: isAnswered(question) ? 'purple' : 'lightgrey' }"
-            />
-          </svg>
-          <p class="item">
+          <KIcon
+            class="dot"
+            icon="dot"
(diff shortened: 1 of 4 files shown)


kolibri/plugins/learn/assets/src/views/ExamPage/AnswerHistory.vue
```diff
@@ -1,29 +1,29 @@
...
+    <li
+      v-for="(question, index) in questions"
+      :key="index"
+      :ref="`item-${index}`"
+      class="list-item"
+    >
+      <button
+        :class="buttonClass(index)"
+        :disabled="questionNumber === index"
+        class="clickable"
```
bjester, line 13: I think `:aria-selected="questionNumber === index"` would be great

kolibri/plugins/learn/assets/src/views/ExamPage/AnswerHistory.vue
```diff
@@ -32,29 +32,62 @@
...
+        const element = this.$refs[`item-${index}`][0];
+        if (element && element.scrollIntoView && this.wrapperComponentRefs.questionListWrapper) {
+          const container = this.wrapperComponentRefs.questionListWrapper.$el;
+          if (isAboveContainer(element, container)) {
+            element.scrollIntoView({ block: 'start', inline: 'nearest', behavior: 'smooth' });
+          } else if (isBelowContainer(element, container)) {
+            element.scrollIntoView({ block: 'end', inline: 'nearest', behavior: 'smooth' });
+          }
+        }
+      },
```
bjester, line 72: I think an alternate approach could be to emit another event for something like `questionRendered` (or perhaps just a `$nextTick` on the existing event) and then expose public methods of `isAboveContainer(container)`, `isBelowContainer(container)`, and `scrollIntoView(opts)`. It still would seem a little awkward because this component has a list of questions, but considering it accepts a `questionNumber` prop, I think exposing those methods which make decisions based on that prop could make sense.

bjester (approved):
I didn't find any regressions, and improves experience.

Issue #6314, opened 2019-12-15
newly-imported channels do not show without refresh and are not labeled


indirectlylit:
### Observed behavior

This issue is a combination of:

* a regression from https://github.com/learningequality/kolibri/issues/2030
* new functionality not-yet implemented

| step | screenshot |
|--|--|
| original channel listing |![image](https://user-images.githubusercontent.com/2367265/70857178-a0d1ee00-1e9e-11ea-8aa4-41f5accac722.png)|
| start a content import, and click "Back to channels" before it completes | ![image](https://user-images.githubusercontent.com/2367265/70857180-a4657500-1e9e-11ea-9e5b-c510a8a13975.png)|
| After import complete, the channel does not show up | ![image](https://user-images.githubusercontent.com/2367265/70857171-8e57b480-1e9e-11ea-8557-290879abeeba.png)|

### Expected behavior

* If an import related to a channel completes while viewing the channel list, the channel should move to or appear at the top of the channel list
* If an import related to a channel completes before viewing (or refreshing) the channel list, it should appear in its default position
* The 'New' label should appear next to any channel that has an 'import' task still visible in the task list.
* Clearing the task list using the button on the channel list page should clear 'New' labels, but _not_ change the channel order
* Navigating away from and back to the channel list moves channels to their default order

Original specs are very slightly different but the same in broad strokes:

https://www.figma.com/file/eJd6xQiNgebzzRWn2XDDS92R/0.13-content-import?node-id=472%3A0

![image](https://user-images.githubusercontent.com/2367265/70857283-3ae66600-1ea0-11ea-81f8-ac98ab49b203.png)

### User-facing consequences

This feature was important for https://github.com/learningequality/clearinghouse/issues/254

### Steps to reproduce

* Begin import of a new channel
* Navigate to 

### Context

0.13.0 beta 7

rtibbles:
Fixed in #6355

Issue #6229, opened 2019-12-05
diff stats are sometimes blank


indirectlylit:
### Observed behavior

note that diff stats are blank:

![image](https://user-images.githubusercontent.com/2367265/70260587-7321d200-1745-11ea-93e1-78a1dccdc8bd.png)

### Expected behavior

diff stats always show

### User-facing consequences

no information in UI

### Errors and logs

no errors in JS or Python

### Steps to reproduce

unknown. observed intermittently - refreshing made the numbers appear:

![image](https://user-images.githubusercontent.com/2367265/70260687-a9f7e800-1745-11ea-9fc8-34cbdae91b5b.png)

### Context

0.13.0 beta 3

jonboiser:
This seems like an edge case, that seems to happen when too many identical `channeldiffstats` requests (i.e. same channel and from same source) are triggered very quickly in a way that the client can't cancel or clean them up quickly enough. You can sort of reproduce this by going back and forth between the Manage Channel Page and this page very quickly and wait until this kind of console error appears:

```
  File "/Users/jon/.local/share/virtualenvs/kolibri-UH8N_Uf0/lib/python3.7/site-packages/sqlalchemy/engine/default.py", line 509, in do_execute
    cursor.execute(statement, parameters)
sqlalchemy.exc.OperationalError: (sqlite3.OperationalError) table content_contentnode_related already exists [SQL: '\nCREATE TABLE content_contentnode_related (\n\tid INTEGER NOT NULL, \n\tfrom_contentnode_id CHAR(32) NOT NULL, \n\tto_contentnode_id CHAR(32) NOT NULL, \n\tPRIMARY KEY (id), \n\tFOREIGN KEY(to_contentnode_id) REFERENCES content_contentnode (id), \n\tFOREIGN KEY(from_contentnode_id) REFERENCES content_contentnode (id)\n)\n\n'] (Background on this error at: http://sqlalche.me/e/e3q8)
```

From the Rest API, the response will be a FAILED task, so I guess we can show an AppError page in that case, which will prompt the user to refresh the page and hopefully get what they need.

indirectlylit:
> too many requests

This may be one way to trigger it, but during bug bash people were definitely seeing it on the "first try", and that's what I experienced also.

jonboiser:
[...]

Issue #6177, opened 2019-12-02
Task Bar counts failed tasks as "done" in progress bar, but not in the"x of y tasks complete" label


indirectlylit:
### Observed behavior

reported by @jtamiace during the 0.13.0 beta1 bug bash:

> 2 of 7 tasks failed, and the summary statement said '5 of 7 tasks complete'.
> 
> ![image](https://user-images.githubusercontent.com/2367265/69924866-48dcc580-1462-11ea-8bfb-e9802c31bf74.png)
> 
> ![image](https://user-images.githubusercontent.com/2367265/69924872-4e3a1000-1462-11ea-97b9-d4eb7f1ea483.png)

### Expected behavior

* Would be more accurate to say '7 of 7 tasks complete'.
* The task bar should also reflect this ratio, and _not_ incorporate actual number of resources or disk size

### User-facing consequences

confusing accounting of failed processes

### Steps to reproduce

start a number of imports, and cancel a few of them

### Context

0.13.0 beta 3

indirectlylit:
related: https://github.com/learningequality/kolibri/issues/6183

indirectlylit:
thanks for renaming

jonboiser:
@indirectlylit @jtamiace Should we count Completed + Canceled + Failed tasks as "Complete" for the purposes of this task bar? There's a utility function in the code that groups these together as "clearable" that we can use to count them.

All "clearable" tasks will contribute 1 whole unit to the progress bar. Any queued or running Tasks will contribute their percentage value to the progress bar and 0 to the label.

E.g. if you have 1 canceled task + 1 in-progress task at 50%:

Label: "1 of 2 Tasks complete"
Progress Bar: (100% + 50%) /2 = 75%

jtamiace:
I think counting completed, canceled, and failed as "complete" would solve the confusion.

radinamatic:
I fail to see _failed_ as _completed_, but it might be nitpicking in the current crunch... 😛

jonboiser:
There was some feedback from a bug bash that we could both make the message more specific about succeeded vs failed, and also divide the progress bar into different colored sections like we do on coach reports. (e.g. completed vs. canceled vs. failed).

radinamatic:
I agree that would be preferable as more specific and correct info for the user. I could maybe also see _cancelled_ as _completed_ because the _cancel_ action was instigated by the user (and it was, presumably successfully concluded), but _failed_ should fall in a separate category.

[...]

PR #6355, opened 2019-12-16
Only auto-delete individual tasks when they finish

### LINKED ISSUE #6140: tasks got cleared without triggering a 'clear task' action
### Observed behavior

Observed that my list of tasks got cleared after initiating a new import

### Expected behavior

tasks should not be cleared until explicitly done by the user

### User-facing consequences

loss of data: historical context

### Errors and logs

none

### Steps to reproduce

see notes below

### Context

0.13.0 beta 1

jonboiser:
### Summary

1. Updates the places where we clear out the whole queue after a task completes under the old only-one-task-at-a-time model with code to only clear out the task that was being watched for whatever reason. Fixes #6140 
1. ~Removes the lock icon on the SelectContentPage that used to be shown conditionally when a channel is unlisted (relies on a `channel.public` field that has been lost in recent changes)~
1. Add 'public' field to ChannelMetadataSerializer, so that the unlisted channel icon would show. This icon shows strictly when public = false, and not when public = null just in case the channel was never given an annotation.
1. Fixes code in NewChannelVersionPage that would short circuit a new CHANNELDIFFSTATS task if a similar looking task (but not the same kind of Task) of a different type was in the queue. Provides a solution to #6229 
1. Add channel list updating logic described in #6314

### Reviewer guidance

1. Can test the fix for #6140 by starting and finishing some tasks, then starting an import from Kolibri Studio (preferably a channel that you have not already downloaded the sqlite3 file for). This will not clear out the tasks from the Task Manager.
1. Test the fix for the channeldiffstats issue, by importing from a private channel on Studio. Don't clear out your queue. Then update the channel and re-view the upgrade page. Without this change, you should see the empty diffstats table. With this change, you should not see an empty table. This could be dependent on the order of things in the queue, so make sure the import task is towards the top.

### References

…

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/plugins/device/assets/src/views/ManageContentPage/ChannelPanel/WithSizeAndOptions.vue
+++ b/kolibri/plugins/device/assets/src/views/ManageContentPage/ChannelPanel/WithSizeAndOptions.vue
@@ -155,6 +155,7 @@
   }
 
   .private-icons {
+    position: relative;
     display: inline-block;
     margin-top: -3px;
     margin-bottom: 3px;
@@ -167,10 +168,17 @@
   }
 
   .new-label {
-    padding: 2px 5px 2px 4px;
+    position: absolute;
+    top: 2px;
+    padding: 2px 8px;
     margin-left: 8px;
     font-size: 14px;
+    font-weight: bold;
     border-radius: 2px;
+
+    .panel-sm & {
+      top: -2px;
+    }
   }
 
 </style>
--- a/kolibri/plugins/device/assets/src/views/ManageContentPage/ChannelPanel/WithSizeAndOptions.vue
+++ b/kolibri/plugins/device/assets/src/views/ManageContentPage/ChannelPanel/WithSizeAndOptions.vue
@@ -120,6 +120,16 @@
     padding: 16px 0;
   }
 
+  svg.lock-icon {
+    width: 24px;
+    height: 24px;
+
+    .panel-sm & {
+      width: 20px;
+      height: 20px;
+    }
+  }
+
   .col-2 {
     min-width: 80px;
     margin-right: 16px;
@@ -146,7 +156,14 @@
 
   .private-icons {
     display: inline-block;
+    margin-top: -3px;
+    margin-bottom: 3px;
     vertical-align: top;
+
+    .panel-sm & {
+      margin-top: -1px;
+      margin-bottom: 1px;
+    }
   }
 
   .new-label {
--- a/kolibri/plugins/device/assets/src/views/ManageContentPage/ChannelPanel/WithSizeAndOptions.vue
+++ b/kolibri/plugins/device/assets/src/views/ManageContentPage/ChannelPanel/WithSizeAndOptions.vue
@@ -5,7 +5,20 @@
     :class="{'panel-sm': windowIsSmall}"
     :style="{ borderTop: `1px solid ${$themePalette.grey.v_200}` }"
   >
-    <ChannelDetails :channel="channel" />
+    <ChannelDetails :channel="channel">
+
+      <template v-if="showNewLabel" v-slot:belowname>
+        <div class="private-icons">
+          <span
+            class="new-label"
+            :style="{
+              color: $themeTokens.textInverted,
+              backgroundColor: $themeTokens.success
+            }"
+          >{{ WithImportDetailsStrings.$tr('newLabel') }}</span>
+        </div>
+      </template>
+    </ChannelDetails>
 
     <div
       class="col-2"
@@ -35,7 +48,11 @@
   import responsiveWindowMixin from 'kolibri.coreVue.mixins.responsiveWindowMixin';
   import commonCoreStrings from 'kolibri.coreVue.mixins.commonCoreStrings';
   import bytesForHumans from 'kolibri.utils.bytesForHumans';
+  import { crossComponentTranslator } from 'kolibri.utils.i18n';
   import ChannelDetails from './ChannelDetails';
+  import WithImportDetails from './WithImportDetails';
+
+  const WithImportDetailsStrings = crossComponentTranslator(WithImportDetails);
 
   export default {
     name: 'WithSizeAndOptions',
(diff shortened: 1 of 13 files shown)


kolibri/plugins/device/assets/src/views/SelectContentPage/ChannelContentsSummary.vue
```diff
@@ -13,11 +13,12 @@
...
         <h1>
           <KLabeledIcon icon="channel" :label="channel.name" />
-          <KIcon
+          <!-- Commenting out this icon because we've lost the channel.public attribute -->
+          <!-- <KIcon
             ref="lockicon"
             class="lock-icon"
             icon="unlistedchannel"
-          />
+          /> -->
```
indirectlylit: @bjester do you have any info here? I think this implementation may have changed recently?

bjester: I'm not clear what this is. My work with public channel status was only in the backend.

jonboiser: The `public` field is only returned on the PublicChannelSerializer. The other ChannelSerializers do not seem to have the field as part of their output. I think we are able to show the lock icon properly at the Studio Channel Listing step, because we call the Public version of the channels API. In most other import/export workflows were are calling the ChannelMetadataViewset, which only shows local channels.

jonboiser: It's actually this change that removed the conditional too from this icon. https://github.com/jonboiser/kolibri/commit/4d9aa48f3a94d00acefa827e8fd2073bbd646d00#diff-9127258abd7abf8268a08ff621f35266

To show the icon correctly, we need to add the 'public' field to the Channel-list endpoint and bring back the v-if on the icon as well.

jonboiser: I'm noticing that in my content_channelmetadata table, all of the Channels have their `public` field equal to null. This feature is new to me, so I'm assuming at some point at import this field was supposed to be set to True or False?

bjester: On the Kolibri side, the `public` field is brand new. I believe that information was being used in a response from studio. If it's needed, I would suggest adding it.

bjester: Yeah it's `null` for channels imported prior to the migration since that's an annotation at time of import.

jonboiser: Oh yeah I'm seeing this, but it seems to label USB-imported channels as public = null by default. Or maybe because i exported these channels before this annotation.

kolibri/plugins/device/assets/src/views/ManageContentPage/TasksBar.vue
```diff
@@ -81,6 +75,14 @@
...
       },
     },
+    watch: {
+      doneTasks(val, oldVal) {
+        // Just refresh the channel list whenever anything finishes to get the latest version
+        if (val > oldVal) {
+          this.$store.dispatch('manageContent/refreshChannelList');
+        }
+      },
+    },
```
indirectlylit: It seems a bit odd to put this logic inside a watcher in a vue component. I would have expected it to live inside  the handler from the `refreshTaskList` action:

https://github.com/learningequality/kolibri/blob/aecc35cd267ea15d8e4e4bf30ecdcfcd90c29d92/kolibri/plugins/device/assets/src/modules/manageContent/actions/taskActions.js#L37-L47

Some of the computed props in this component might be implemented as vuex getters instead to help abstract away the task implementation details stuff.

related: https://github.com/learningequality/kolibri/issues/6177

jonboiser: I'd rather not put the refresh channel side effect in refreshTaskList, since we would potentially be updating the channel list in places other than the ManageContentPage. We'd need to think about how to handle sudden changes to the channel list like in Delete/Export flows.

jonboiser: In practice, I've found that having the channel list in vuex and globally shared is not always ideal and that each component should request and manage the channel list themselves for their own purposes. Under this structure, the ManageContentPage would still be watching the Task queue and refreshing its local copy of the channels.

kolibri/plugins/device/assets/src/views/ManageContentPage/ChannelPanel/WithSizeAndOptions.vue
```diff
@@ -110,4 +154,23 @@
...
+      margin-top: -1px;
+      margin-bottom: 1px;
+    }
+  }
+
+  .new-label {
+    padding: 2px 5px 2px 4px;
+    margin-left: 8px;
+    font-size: 14px;
+    border-radius: 2px;
```
indirectlylit, line 177: recommend some slight styling changes to this and the other 'New' label:

* bold
* 8px left and right padding
* align text bottom with row


| before | after |
|--|--|
| image|image|

```css
  .new-label {
    position: relative;
    top: 2px;
    padding: 2px 8px;
    margin-left: 8px;
    font-size: 14px;
    font-weight: bold;
    border-radius: 2px;
  }
```

indirectlylit (approved):
this looks great - nice job!

kolibri/plugins/device/assets/src/views/ManageContentPage/TasksBar.vue
indirectlylit: sounds good

indirectlylit:
also fixes https://github.com/learningequality/kolibri/issues/6365

rtibbles:
The build is stalled waiting for the MacOS build agent - merging.

Issue #6327, opened 2019-12-16
Perseus exercises appear above the BottomBar widget


indirectlylit:
### Observed behavior

reported by @jonboiser that exercises are over the bottom bar:

![image](https://user-images.githubusercontent.com/2367265/70873212-8a409b00-1f61-11ea-8424-3cebdb92d22d.png)

### Expected behavior

exercise below bottom bar

### User-facing consequences

broken UI

### Steps to reproduce

create or take a quiz with "place value when multiplying and dividing by 10"

### Context

0.13.0 beta 6

nucleogenesis:
I tried to replicate this without luck on the Kolibri beta server (beta7).

![perseus-does-not-overlay](https://user-images.githubusercontent.com/6356129/71132555-b308a080-21ac-11ea-8dc9-775165562e76.gif)

Try it out logging in as learner `queencreens` and taking the quiz in SUPER LE CLASS called "divide"

nucleogenesis:
Just kidding about the last comment - it appears this way in the Quiz Preview page.

It is fixed by making the z-index on the bottom bar 1062. Will make a PR.

rtibbles:
Fixed in #6396

PR #6396, opened 2019-12-19
Inilne style z-index magic number to avoid perseus overlay

nucleogenesis:
### Summary

Fixes issue noted here: 

https://github.com/learningequality/kolibri/issues/6327

### Reviewer guidance

Go to create a quiz, make sure that your page showing the quiz questions is sized so that it would go off the bottom of the screen.

The perseus renderer content will not overlay the bottom bar.

This issue isn't found in the quiz taking part as a learner, but only in quiz creation.

So I just added the style inline rather than making the whole bottom bar component have a wacky z-index.

### References

https://github.com/learningequality/kolibri/issues/6327

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/plugins/coach/assets/src/views/plan/CreateExamPage/CreateExamPreview.vue
+++ b/kolibri/plugins/coach/assets/src/views/plan/CreateExamPage/CreateExamPreview.vue
@@ -105,7 +105,7 @@
         :selectedExercises="selectedExercises"
       />
 
-      <BottomAppBar>
+      <BottomAppBar style="z-index: 1062;">
         <KRouterLink
           appearance="flat-button"
           :text="coreString('goBackAction')"


jonboiser:
Maybe to make the code a little nicer, could put this rule in a CSS class placed on the BottomAppBar and defined inside of CreateExamPreview? Also, are we sure perseus widgets don't have z-indices higher than 1061?

jonboiser (approved):

rtibbles:
Builds stalled on RPi step.

kolibri/plugins/coach/assets/src/views/plan/CreateExamPage/CreateExamPreview.vue
```diff
@@ -105,7 +105,7 @@
         :selectedExercises="selectedExercises"
       />
 
-      <BottomAppBar>
+      <BottomAppBar style="z-index: 1062;">
```
indirectlylit, line 108: where does this number magic come from?

This is fine for now, but generally, our z-indexes should map to Material elevations - https://material.io/design/environment/elevation.html#default-elevations

rtibbles:
I think we will want to do some materialisation of perseus z indices to properly handle this

rtibbles:
Alternatively, we can just chuck perseus inside an iframe and stop it leaking.

PR #6402, opened 2019-12-19
Allowing upload step to continue without pi image

DXCanas:
### Summary

Builds are _really_ slow. This change should let me block the pi pipeline without having this build fail. Once this is in, I can re-enable blocked-by-default on the pi pipeline.

----

### Contributor Checklist

PR process:

- [x] PR has the correct target branch and milestone
- [x] PR has 'needs review' or 'work-in-progress' label
- [x] If PR is ready for review, a reviewer has been added. (Don't use 'Assignees')
- [ ] If this is an important user-facing change, PR or related issue has a 'changelog' label
- [ ] If this includes an internal dependency change, a link to the diff is provided

Testing:

- [x] Contributor has fully tested the PR manually
- [x] If there are any front-end changes, before/after screenshots are included
- [ ] Critical user journeys are covered by Gherkin stories
- [ ] Critical and brittle code paths are covered by unit tests

### Reviewer Checklist

- Automated test coverage is satisfactory
- PR is fully functional
- PR has been tested for [accessibility regressions](http://kolibri-dev.readthedocs.io/en/develop/manual_testing.html#accessibility-a11y-testing)
- External dependency files were updated if necessary (`yarn` and `pip`)
- Documentation is updated
- Contributor is in AUTHORS.md

PR #6414, opened 2019-12-19
Preserve pigen image and container

jredrejo:
### Summary
Our building system prune all the docker images and containers after finishing.
When building the Raspberry Pi image, if the docker container is not kept, all the Debian system and packages are downloaded and installed once again in every run of the builder.
This PR keeps undeleted the image and container for the Raspberry Pi image. They should be deleted only when a new Debian distribuition for Raspberry is released, what happen every 2 years aproximately.

### Reviewer guidance
The pigen_work container and pi-gen images still exists in the building server after the image is built. 

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -239,8 +239,8 @@ i18n-upload-glossary:
 	python build_tools/i18n/crowdin.py upload-glossary
 
 docker-clean:
-	docker container prune -f
-	docker image prune -f
+	docker container prune --filter "label!=build=pigen" -f
+	docker image prune --filter "label!=build=pigen" -a -f
 
 docker-whl: writeversion docker-envlist
 	docker image build -t "learningequality/kolibri-whl" -f docker/build_whl.dockerfile .


DXCanas (approved):
This will work, but we need something a little more sustainable down the line. 

I still thinking blocking the build by default would be the way to go, as in #https://github.com/learningequality/kolibri/pull/6402

indirectlylit:
should this be targeted to 0.13?

DXCanas:
@indirectlylit I think that's where it'd be most helpful, yeah

indirectlylit:
needs a rebase

jredrejo:
I have changed the code to use actual labels that are applied in https://github.com/learningequality/pi-gen/blob/master/Dockerfile#L2
Previous method did not work because neither the container nor the image were labeled. It seems Docker does not consider the image name a label nor provide options to filter by the image or container names.

lyw07:
for the current build failure, I think https://github.com/learningequality/kolibri/pull/6402 can fix the issue?

DXCanas:
@lyw07 #6402 was only helpful if we want to block the build from buildking every time. I'm thinking we should maybe hold off on doing that until after release. This build should pass with the latest changes, though.

lyw07:
thank you for the info! I will merge this PR after the build passes

PR #6446, opened 2020-01-09
Maintain channel order across list views

### LINKED ISSUE #6366: maintain order of channels across lists
### Observed behavior

Switching between channel list and import views is a bit confusing because orders are similar but slightly different.

| channel list | import channel list |
|--|--|
| localhost_8000_en_device_ (1)|localhost_8000_en_device_|

### Expected behavior

Expect order on both pages to be consistent. In practice, this would mean bubbling any on-device channels to the top of the import list.

### User-facing consequences

on-device channels look like they went missing, but really just moved down in the list

### Context

0.13.0 beta 7

bransgithub:
### Summary
 * Moves installed channels to top of the list on the import view. Fixes #6366 

Before Fix:

| Channel View        | Import View  |
| ------------- |:-------------:| 
|     |  |

After Fix:

| Channel View        | Import View  |
| ------------- |:-------------:| 
|       |  |

This also handles channels imported via token:

And once the unlisted channel is installed:

| Channel View        | Import View  |
| ------------- |:-------------:| 
|   |  |

If the unlisted channel is already installed and we try to import it again: 

### Reviewer guidance

…

### References

…

----

### Contributor Checklist

PR process:

- ~~[ ] If this is an important user-facing change, PR or related issue has a 'changelog' label~~
- ~~[ ] If this includes an internal dependency change, a link to the diff is provided~~

Testing:

- ~~[ ] Critical user journeys are covered by Gherkin stories~~

### Reviewer Checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/plugins/device/assets/src/views/AvailableChannelsPage/index.vue
+++ b/kolibri/plugins/device/assets/src/views/AvailableChannelsPage/index.vue
@@ -170,8 +170,21 @@
         'transferType',
       ]),
       allChannels() {
+        const uninstalledChannels = this.availableChannels.filter(uninstalledChannel => {
+          return !this.installedChannelsWithResources.find(
+            ({ id }) => id === uninstalledChannel.id
+          );
+        });
+
         // Need to de-duplicate channels in case user enters same token twice, etc.
-        return uniqBy([...this.newUnlistedChannels, ...this.availableChannels], 'id');
+        return uniqBy(
+          [
+            ...this.newUnlistedChannels,
+            ...this.installedChannelsWithResources,
+            ...uninstalledChannels,
+          ],
+          'id'
+        );
       },
       multipleMode() {
         const { multiple } = this.$route.query;
--- a/kolibri/plugins/device/assets/test/views/AvailableChannelsPage.spec.js
+++ b/kolibri/plugins/device/assets/test/views/AvailableChannelsPage.spec.js
@@ -105,9 +105,9 @@ describe('availableChannelsPage', () => {
     const channels = channelListItems();
     const channelNProps = n => channels.at(n).props();
     expect(channelNProps(0).onDevice).toEqual(true);
-    expect(channelNProps(1).onDevice).toEqual(false);
+    expect(channelNProps(1).onDevice).toEqual(true);
     expect(channelNProps(2).onDevice).toEqual(false);
-    expect(channelNProps(3).onDevice).toEqual(true);
+    expect(channelNProps(3).onDevice).toEqual(false);
   });
 
   it('IN LOCALIMPORT/REMOTEIMPORT, with no filters, all appear', () => {
@@ -151,7 +151,7 @@ describe('availableChannelsPage', () => {
     filter.vm.model = 'bir ch';
     await wrapper.vm.$nextTick();
     expect(filterComponent().vm.titleFilter).toEqual('bir ch');
-    testChannelVisibility(wrapper, [false, true, false, false]);
+    testChannelVisibility(wrapper, [false, false, true, false]);
   });
 
   it('with both filters, the correct channels appear', async () => {
@@ -163,7 +163,7 @@ describe('availableChannelsPage', () => {
     await wrapper.vm.$nextTick();
     lFilter.vm.selection = { label: 'German', value: 'de' };
     await wrapper.vm.$nextTick();
-    testChannelVisibility(wrapper, [false, false, true, false]);
+    testChannelVisibility(wrapper, [false, false, false, true]);
   });
 
   it('the "select" link goes to the correct place', () => {


jonboiser:
@bransgithub Thanks for this. Could you rebase your code to the `origin/release-v0.13.x` branch? The commit history for this PR should only have commit `f926895` ("Maintain channel order between lists").

bransgithub:
> @bransgithub Thanks for this. Could you rebase your code to the `origin/release-v0.13.x` branch? The commit history for this PR should only have commit `f926895` ("Maintain channel order between lists").

@jonboiser Thanks - done! 

By the way, is there a way I can set the label/reviewer? It seems like I can't without write access, but if there's a way I'd love to know!

jonboiser:
Only members in the Learning Equality Organization can set labels and reviewers.

Could you test an additional scenario (installing unlisted channels)? To do that, click that "Import with token" button on the top of the "Import view" screen. Enter this string in the modal `nosov-pagif`, which should send you to a channel called "Slides, slides, slides" (I think).

Could you verify that your PR handles these kinds of channels as well?

I'll need to take a look at your code. I'm thinking there might be a way to implement the same solution in less code by taking advantage of some utility functions or other JS language features.

kolibri/plugins/device/assets/src/views/AvailableChannelsPage/index.vue
```diff
@@ -171,7 +173,28 @@
       ]),
       allChannels() {
         // Need to de-duplicate channels in case user enters same token twice, etc.
```
jonboiser: If the intent is to return something like `[...sortedInstalledChannels, ...otherChannels]`, maybe what you could do is compute the `otherChannels` array using something like `lodash/differenceBy` or `Array.prototype.filter` to remove the installed channels from `this.availableChannels` to produce `otherChannels`

kolibri/plugins/device/assets/src/views/AvailableChannelsPage/index.vue
```diff
@@ -210,6 +233,19 @@
       },
     },
     watch: {
+      installedChannelsWithResources: {
```
jonboiser: I'm thinking this might not be necessary to fix #6366. I'm worried that channels might move around unexpectedly if other Tasks get started by other admins, or by the same user.

Let's just make sure that more "permanent" ordering on the "Channel View" is maintained in the "Import View".

kolibri/plugins/device/assets/src/views/AvailableChannelsPage/index.vue
bransgithub: Yup, agreed. I don't know if I can directly use `lodash/differenceBy` since the elements between `sortedInstalledChannels` and `otherChannels` aren't the exact same, but I've cleaned it up to make use of `Array.prototype.filter`.

kolibri/plugins/device/assets/src/views/AvailableChannelsPage/index.vue
bransgithub: That makes sense. I've removed the unnecessary sections!

bransgithub:
I've addressed your suggestions and confirmed that it works with an installed, 'unlisted' channel (images included in original PR comment). Thanks!

kolibri/plugins/device/assets/src/views/AvailableChannelsPage/index.vue
```diff
@@ -171,7 +171,18 @@
       ]),
       allChannels() {
         // Need to de-duplicate channels in case user enters same token twice, etc.
-        return uniqBy([...this.newUnlistedChannels, ...this.availableChannels], 'id');
+        const allPossibleChannels = uniqBy(
```
jonboiser: I'm not sure this is quite correct yet. We actually want `this.newUnlistedChannels` to appear at the beginning of the list. Here's the scenario that you might not have seen yet (it's a brand new feature in Kolibri)

1. Go to the Import View.
1. Click the "Select Entire Channels Instead" link at the top (will put a checkbox next to all the channels).
1. Click "Import with token" and use the same token I gave you in a comment.
1. When you click "submit", 1) "slides, slides, slides" channel should be at the top, 2) the other installed channels should come afterwards (note if "slides" was already installed, don't show again, then 3) the other uninstalled channels.

jonboiser: So I think this is very close, just make sure that `...this.newUnlistedChannels` is at the beginning of the returned array, `installedChannelsWithResourecs` is in the middle (with no duplicates of unlisted channels), and the filtered array is at the very end :+1:

bransgithub: Ah, thanks for shedding some light on that! I've moved `this.newUnlistedChannels` to the top of the list (image appended to original comment). 

> if "slides" was already installed, don't show again

Does this mean that you don't expect "slides" to be shown in the 'import' view if it was already installed? If so, from what I can tell it's included in `this.installedChannelsWithResources` and I'm not sure if that should be the case then.

jonboiser: If you have these channels installed

1. Public Channel 1
2. Public Channel 2
3. Unlisted Channel 

Then the import view should show

1. Public Channel 1
2. Public Channel 2
3. Unlisted Channel 
4. (the other not-installed channels)

But if you enter the token for unlisted channel, it should show

3. Unlisted Channel (now at the top, and not repeated below)
1. Public Channel 1
2. Public Channel 2
4. (the other not-installed channels)

bransgithub: Thanks for the clarification! I believe those cases are all handled now; I've added another screenshot to the original post.

jonboiser:
Looks good from the screenshots! Can you rebase this on the latest of `origin release-v0.13.x` to fix the failing buildkite check. I'll give this a manual test on Monday. Thanks!

bransgithub:
Sounds like a plan. Happy Friday!

jonboiser (approved):

PR #6470, opened 2020-01-16
enable byte-range resuming of file imports

lyw07:
### Summary

This PR is to enable byte-range resuming of file imports so that when the internet connection is lost in the middle of importing a large file, retrying file import will start from where it stops earlier instead of starting from the beginning of the file. I use `Range` property in the request header to achieve this.

However, not all the files will be enabled with byte-range resuming. Files compressed with gzip (either done by us on Google Cloud Storage or done by Cloudflare), e.g. large channel database files, do not support range requests.

Here are some major changes I make in this PR:
- Byte-range resuming of importing files that are not compressed with gzip
- Move the retry logic from the commands python files to transfer.py to clean up the code and avoid keeping open new file handlers every 30 seconds
- Pass `is_cancelled()` to Transfer objects and raise `TransferCanceled` exception when users cancel the file transfer instead of calling `is_cancelled()` every iteration of file transfer. This is mainly to make code in `importchannel.py` simpler.
- Handle a case about retry import when internet connection is lost at the beginning of `start()`, where iteration of file content has not started yet, which we are not currently handling
- Edit and add tests based on the changes above

### Reviewer guidance

I've done manual testing under the following circumstances:
- [x] channel import
- [x] channel import no internet/resume
- [x] content network import
- [x] content network import cancel
- [x] content network import no internet cancel
- [x] content network import no internet/resume
- [x] content local import
- [x] content local import cancel
- [x] content local import skip
- [x] content export
- [x] content export cancel

please feel free to let me know if you have any questions! Thank you!

### References

https://www.notion.so/learningequality/Byte-range-resuming-of-file-imports-0755383c8810422291881c6a211780ed

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/core/content/management/commands/exportcontent.py
+++ b/kolibri/core/content/management/commands/exportcontent.py
@@ -99,43 +99,42 @@ class Command(AsyncCommand):
                 if self.is_cancelled():
                     break
 
-                filename = f.get_filename()
-
-                try:
-                    srcpath = paths.get_content_storage_file_path(filename)
-                    dest = paths.get_content_storage_file_path(
-                        filename, datafolder=data_dir
-                    )
-                except InvalidStorageFilenameError:
-                    # If any files have an invalid storage file name, don't export them.
-                    overall_progress_update(f.file_size)
-                    continue
-
-                # if the file already exists, add its size to our overall progress, and skip
-                if os.path.isfile(dest) and os.path.getsize(dest) == f.file_size:
-                    overall_progress_update(f.file_size)
-                    continue
-
-                copy = transfer.FileCopy(srcpath, dest)
-
-                with copy:
-
-                    with self.start_progress(
-                        total=copy.total_size
-                    ) as file_cp_progress_update:
-
-                        for chunk in copy:
-                            if self.is_cancelled():
-                                copy.cancel()
-                                break
-                            length = len(chunk)
-                            overall_progress_update(length)
-                            file_cp_progress_update(length)
-                        else:
-                            exported_files.append(dest)
+                dest = self.export_file(f, data_dir, overall_progress_update)
+                if dest:
+                    exported_files.append(dest)
 
             if self.is_cancelled():
                 # Cancelled, clean up any already downloading files.
                 for dest in exported_files:
                     os.remove(dest)
                 self.cancel()
+
+    def export_file(self, f, data_dir, overall_progress_update):
+        filename = f.get_filename()
+
+        try:
+            srcpath = paths.get_content_storage_file_path(filename)
+            dest = paths.get_content_storage_file_path(filename, datafolder=data_dir)
+        except InvalidStorageFilenameError:
+            # If any files have an invalid storage file name, don't export them.
+            overall_progress_update(f.file_size)
+            return
+
+        # if the file already exists, add its size to our overall progress, and skip
+        if os.path.isfile(dest) and os.path.getsize(dest) == f.file_size:
+            overall_progress_update(f.file_size)
+            return
+
+        copy = transfer.FileCopy(srcpath, dest, cancel_check=self.is_cancelled)
+
+        with copy, self.start_progress(
+            total=copy.total_size
+        ) as file_cp_progress_update:
+            try:
+                for chunk in copy:
+                    length = len(chunk)
+                    overall_progress_update(length)
+                    file_cp_progress_update(length)
+            except transfer.TransferCanceled:
+                return
+        return dest
--- a/kolibri/core/content/utils/transfer.py
+++ b/kolibri/core/content/utils/transfer.py
@@ -233,9 +233,12 @@ class FileDownload(Transfer):
             # When internet connection is lost at the beginning of start(),
             # self.response does not get an assigned value
             if hasattr(self, "response"):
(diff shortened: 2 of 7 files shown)


ralphiee22 (approved):
This looks awesome!

indirectlylit:
nice job @lyw07 - this looks quite sophisticated

kolibri/core/content/utils/transfer.py
```diff
@@ -149,17 +153,26 @@ def __init__(self, *args, **kwargs):
...
+            )
+            self.response.raise_for_status()
+        except Exception as e:
+            retry = retry_import(e)
+            if not retry:
+                raise
+            # Catch exceptions to check if we should resume file downloading
+            self.resume()
+            if self.cancel_check():
+                self._kill_gracefully()
```
jamalex, line 175: Should this cancel check also happen in the case of successful `get`? (i.e. be de-indented) -- or is there a reason not to?

kolibri/core/content/utils/transfer.py
```diff
@@ -184,16 +197,77 @@ def __iter__(self):
...
+            )
+            sleep(30)
+
+            byte_range_resume = None
+            # When internet connection is lost at the beginning of start(),
+            # self.response does not get an assigned value
+            if hasattr(self, "response"):
+                # Use Content-Length header to check if range requests are supported
+                # For example, range requests are not supported on compressed files
+                byte_range_resume = self.response.headers.get("content-length", None)
```
jamalex: Isn't the standard server header `accept-ranges: bytes`? There could be other proxies or server setups that prevent byte range requests, not sure which header is a more reliable indicator -- but per the spec, I'm assuming this would be.

kolibri/core/content/utils/transfer.py
lyw07: @jamalex yes, the cancel check would happen as well when `get` is successful, but in a bit later time here: https://github.com/learningequality/kolibri/pull/6470/files#diff-4d578307b2a82fe54231716093334b43R200. If a user hits `cancel` button after the `get` is successful, it will not cancel the process right away, but will do so before we start to actually import the content, since the `next()` function is called right after `start()`. Do you have any suggestions if there is a better way to handle this case? Thank you!

kolibri/core/content/management/commands/exportcontent.py
```diff
@@ -99,43 +99,42 @@ def handle_async(self, *args, **options):
...
+            dest = paths.get_content_storage_file_path(filename, datafolder=data_dir)
+        except InvalidStorageFilenameError:
+            # If any files have an invalid storage file name, don't export them.
+            overall_progress_update(f.file_size)
+            return
+
+        # if the file already exists, add its size to our overall progress, and skip
+        if os.path.isfile(dest) and os.path.getsize(dest) == f.file_size:
+            overall_progress_update(f.file_size)
+            return
```
jamalex, line 126: What if the temporary transfer file exists, but is smaller? Can we "resume" the export, too?

jamalex (approved):
Added a couple of questions (which may imply suggestions), but this looks great!

kolibri/core/content/utils/transfer.py
jamalex: That sounds good!

kolibri/core/content/utils/transfer.py
lyw07: ah I was using the header `accept-ranges: bytes` previously, but it didn't work well when the files are the ones we compressed on google cloud storage. In this case, somehow cloudflare thinks it's accepting range requests and sends the header `accept-ranges: bytes`. But when I set the header with `{"Range": "bytes={}-"}`, it errors out by saying the link does not support range request. So instead I used `content-length` because it seems like gzip files do not support range requests and thus do not have the `content-length` header. 
Or maybe I can set both conditions in this case?

kolibri/core/content/management/commands/exportcontent.py
lyw07: I think we normally remove the temporary transfer file when export does not succeed. If a temporary transfer file exists, I assume the case could be a power shutoff etc. 
I feel like we can resume the export too in that case but that can be addressed in another PR?

kolibri/core/content/utils/transfer.py
lyw07: thank you!

kolibri/core/content/utils/transfer.py
jamalex: Ok, makes sense. Sounds like they're not respecting the standard, so what you're doing is safer.

>Or maybe I can set both conditions in this case?

Yeah, that could be good. Only try to resume if both headers are set.

kolibri/core/content/management/commands/exportcontent.py
jamalex: Yeah, true, it would only be in something like a poweroff. And in that case stuff can get corrupted with USBs anyway, so maybe safest to just remove like you do here. No change needed.

kolibri/core/content/utils/transfer.py
lyw07: thank you! I will update the code with that :)

PR #6481, opened 2020-01-20
Show Python version in device info tab

### LINKED ISSUE #6462: Display the Python version where Kolibri server is running
### Observed behavior
During the testing it's good to know on what Python version i'm running on the Kolibri server, This also helpful when it comes on reporting and debugging an issue.

Some operating Systems had a multiple pre installed Python version, and it's difficult to pinpoint on what version that Kolibri use to run.

### Expected behavior
I expect to display the Python version at the Device info page or at the kolibri benchmark command output.

### User-facing consequences

…

### Errors and logs

…

### Steps to reproduce

…

### Context

…

jredrejo:
### Summary
As requested by some users, knowing the Python version the server is running might be useful. This small change shows it in the Device info tab

### Reviewer guidance
Python version should be shown correctly in the device info tab

### References
Fixes #6462 

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/core/device/api.py
+++ b/kolibri/core/device/api.py
@@ -1,3 +1,5 @@
+from sys import version_info
+
 from django.conf import settings
 from django.http.response import HttpResponseBadRequest
 from morango.models import InstanceIDModel
@@ -103,6 +105,9 @@ class DeviceInfoView(views.APIView):
         # Returns the named timezone for the server (the time above only includes the offset)
         info["server_timezone"] = settings.TIME_ZONE
         info["installer"] = installation_type()
+        info["python_version"] = "{major}.{minor}".format(
+            major=version_info.major, minor=version_info.minor
+        )
 
         return Response(info)
 
--- a/kolibri/core/device/api.py
+++ b/kolibri/core/device/api.py
@@ -105,10 +105,9 @@ class DeviceInfoView(views.APIView):
         # Returns the named timezone for the server (the time above only includes the offset)
         info["server_timezone"] = settings.TIME_ZONE
         info["installer"] = installation_type()
-        info["python_version"] = "{major}.{minor}".format(
-            major=version_info.major, minor=version_info.minor
+        info["python_version"] = "{major}.{minor}.{micro}".format(
+            major=version_info.major, minor=version_info.minor, micro=version_info.micro
         )
-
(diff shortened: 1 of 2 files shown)


kolibri/core/device/api.py
```diff
@@ -103,6 +105,9 @@ def get(self, request, format=None):
         # Returns the named timezone for the server (the time above only includes the offset)
         info["server_timezone"] = settings.TIME_ZONE
         info["installer"] = installation_type()
+        info["python_version"] = "{major}.{minor}".format(
+            major=version_info.major, minor=version_info.minor
+        )
```
benjaoming, line 110: I think we should have the entire contents of version_info here.

Maybe like this?

``` 
>>> from sys import version_info
>>> print(".".join(map(str, version_info)[:3]))
2.7.17
```

benjaoming (changes requested):
Great, just think we need the patch, too - from experience, we had 2.7 issues where it mattered which patch release you used on Windows IIRC.

kolibri/core/device/api.py
jredrejo: Cool, thanks, I didn't know the micro part could be useful. I have added it (not using map, I prefered a more explicit  expression, but with the same result)

jredrejo: Done, with explicit info

benjaoming: I wasn't sure of `micro` would be a persistent property, so I used `map(str, version_info)[:3]`, but it might not be so?

jredrejo: Well, we can't be sure `version_info` will be persistent either, if the Python API (https://docs.python.org/3/library/sys.html#sys.version_info)  changes we are done for sure :)
I prefer to follow the Zen of Python:
```
import this
...
Explicit is better than implicit.
Simple is better than complex.
...
```

benjaoming: Thanks, I hadn't looked it up, it's because I know these components as major, minor and patch. I never heard "micro" before :)

benjaoming (approved):
Cheers! :)

indirectlylit (approved):
looks good

image

Issue #6480, opened 2020-01-20
Migration using ALTER and CREATE/UPDATE in one transaction


benjaoming:
### Observed behavior

Upgrades are failing for Postgres in case data is updated after schema alterations in this migration:

https://github.com/learningequality/kolibri/blob/30dcfe13713c21c582cfbb8e91fb375f9f13737a/kolibri/core/device/migrations/0005_auto_20191203_0951.py

Failure is `Cannot ALTER TABLE device_devicesettings because it has pending trigger events`

More info: https://stackoverflow.com/a/12838113

### Secondary issue

We have a test environment run on Postgres, but we are not testing our migrations with data, so we cannot know if they actually work. How do we solve this?

Gherkin?

### Expected behavior

User can use 0.12, add something to `device_devicesettings` and upgrade Kolibri afterwards.

### User-facing consequences

TBC

### Errors and logs

```
TBC
```

### Steps to reproduce

n/a

### Context

Kolibri 0.12 upgrade to 0.13.0

ralphiee22:
My PR fixes the issue, but still leaves open the question how to catch these types of errors in the future. A non-automated solution could be gherkins as suggested by @benjaoming, but we should flag them for "risky" migrations. Another one could be more stringent review of PRs that add migrations. As we learn about these types of migration issues and learn how to deal with them, we could create a template of things to check/test for in migrations.

rtibbles:
I think the simplest automated solution would be to make a custom linting rule that ensures that `RunPython` operations never occur anywhere but as the last element of migrations (in the case where there are operations other than `RunPython`)? That should catch the majority of cases, I think.

benjaoming:
> Another one could be more stringent review of PRs that add migrations. 

Yes, I have often fantasized about adding linting on migrations, as I don't like when they're called `0123_auto_some_timestamp` - names should be verbose etc.

Data migrations and schema migrations should not be mixed, would be great to catch this in linting as well, really good idea @rtibbles ! :+1:

indirectlylit:
created follow-up issues related to linting:

* https://github.com/learningequality/kolibri/issues/6514 (`RunPython` behavior)
* https://github.com/learningequality/kolibri/issues/6515 (migration naming)

rtibbles:
Fixed in #6488

cpauya:
I confirm that #6488 fixed the issue: **specifically the exception during upgrade from v0.12.2 to v0.13.0 using PostgreSQL-9.5**.  

[...]

PR #6488, opened 2020-01-21
Move RunPython to a separate migration and transaction

ralphiee22:
### Summary

We separate the schema migration from the data migration for allowing guest access. For devices that have already ran the migrations prior to this change, we check if the corresponding schema migration has been applied (removing `guest_access` field from `facility_dataset`). If it has been applied, we skip the data migration because the field has already been removed, and the logic relies on that field still existing. 

### Reviewer guidance

Any possible problems with the simple approach?

### References

Fixes https://github.com/learningequality/kolibri/issues/6480.

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/core/device/migrations/0006_migrate_guest_access.py
+++ b/kolibri/core/device/migrations/0006_migrate_guest_access.py
@@ -1,0 +1,64 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.core.exceptions import ObjectDoesNotExist
+from django.db import migrations
+from django.db import models
+from django.db.migrations.recorder import MigrationRecorder
+
+
+def migrate_allow_guest_access(apps, schema_editor):
+    # if this migration has been run, drop out because logic depends on a field that no longer exists
+    if MigrationRecorder.Migration.objects.filter(
+        app="kolibriauth",
+        name="0017_remove_facilitydataset_allow_guest_access",
+        applied__isnull=False,
+    ).exists():
+        return
+    FacilityDataset = apps.get_model("kolibriauth", "FacilityDataset")
+    DeviceSettings = apps.get_model("device", "DeviceSettings")
+
+    try:
+        default_facility = DeviceSettings.objects.get().default_facility
+        allow_guest_access = default_facility.dataset.allow_guest_access
+    except ObjectDoesNotExist:
+        allow_guest_access = (
+            FacilityDataset.objects.filter(allow_guest_access=False).count() <= 0
+        )
+
+    DeviceSettings.objects.update(allow_guest_access=allow_guest_access)
+
+
+def revert_allow_guest_access(apps, schema_editor):
+    # if this migration has been run, drop out because logic depends on a field that no longer exists
+    if MigrationRecorder.Migration.objects.filter(
+        app="kolibriauth",
+        name="0017_remove_facilitydataset_allow_guest_access",
+        applied__isnull=False,
+    ).exists():
+        return
+
+    FacilityDataset = apps.get_model("kolibriauth", "FacilityDataset")
+    DeviceSettings = apps.get_model("device", "DeviceSettings")
+
+    allow_guest_access = (
+        DeviceSettings.objects.filter(allow_guest_access=False).count() <= 0
+    )
+
+    try:
+        default_facility = DeviceSettings.objects.get().default_facility
+        default_facility.dataset.allow_guest_access = allow_guest_access
+        default_facility.dataset.save()
+    except ObjectDoesNotExist:
+        FacilityDataset.objects.update(allow_guest_access=allow_guest_access)
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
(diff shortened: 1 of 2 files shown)


benjaoming:
I was wondering if having several transactions in one migration is a good idea? Does it already happen? Wouldn't that make the migration none-atomic?

ralphiee22:
After testing the behavior, if `RunPython` fails, it appears that the whole migration is rolled back (the sub-transactions are in a larger transaction?). 
Also interesting https://docs.djangoproject.com/en/1.11/howto/writing-migrations/#non-atomic-migrations

benjaoming:
@ralphiee22 I tried skimming the docs and I found in 2.2 docs that it seems like RunPython will likely still be problematic? I haven't read its definition in source code.



> By default, RunPython will run its contents inside a transaction on databases that do not support DDL transactions (for example, MySQL and Oracle). This should be safe, but may cause a crash if you attempt to use the schema_editor provided on these backends; in this case, pass atomic=False to the RunPython operation.
> 
> On databases that do support DDL transactions (SQLite and PostgreSQL), RunPython operations do not have any transactions automatically added besides the transactions created for each migration. Thus, on PostgreSQL, for example, you should avoid combining schema changes and RunPython operations in the same migration or you may hit errors like OperationalError: cannot ALTER TABLE "mytable" because it has pending trigger events.

https://docs.djangoproject.com/en/2.2/ref/migration-operations/#runpython

ralphiee22:
yeah I glossed over those docs as well, and was a bit confused by what they are saying and what I'm seeing. 
Anyways to mitigate all concerns, it would be best to separate out the schema and data migrations.

benjaoming:
I have had discussions about this previously, and I have this stance: It is okay to edit migrations. You can find other takes on this :)

Now, if we do edit `0005_auto_20191203_0951.py` and remove the `RunPython` part, we would have to add a new conditional data migration - one that takes into account that it may have been run previously (before `0005_auto_20191203_0951.py` was updated)

What would you think about that?

ralphiee22:
Is there any harm in just executing the `RunPython` again in a `0006` migration, regardless if it has run before or not? just trying to understand why we need the conditional check.

benjaoming:
No, the field `allow_guest_access`that is used to check previous settings has been removed here:

https://github.com/learningequality/kolibri/blob/develop/kolibri/core/auth/migrations/0017_remove_facilitydataset_allow_guest_access.py

Here is the definition of the RunPython:

```
def migrate_allow_guest_access(apps, schema_editor):
    FacilityDataset = apps.get_model("kolibriauth", "FacilityDataset")
    DeviceSettings = apps.get_model("device", "DeviceSettings")
    try:
        # Firstly we try to get the default facility's settings
        default_facility = DeviceSettings.objects.get().default_facility
        allow_guest_access = default_facility.dataset.allow_guest_access
    except ObjectDoesNotExist:
        # If it isn't set, we just find out if any facilities had guest access disabled
        # ...if no one disabled it, we will allow it
        allow_guest_access = (
            FacilityDataset.objects.filter(allow_guest_access=False).count() <= 0
        )
    DeviceSettings.objects.update(allow_guest_access=allow_guest_access)
```

If you run this twice, and `allow_guest_access` had been set manually in the meantime, it would get overwritten.

One good thing: For blank installations running all the migrations (or some squashed future version), all these decisions don't matter.

ralphiee22:
ahh right, didn't see that.
> Now, if we do edit 0005_auto_20191203_0951.py and remove the RunPython part, we would have to add a new conditional data migration - one that takes into account that it may have been run previously (before 0005_auto_20191203_0951.py was updated)

This approach sounds good moving forward.

kolibri/core/device/migrations/0006_migrate_guest_access.py
```diff
@@ -0,0 +1,64 @@
...
+from django.db import models
+from django.db.migrations.recorder import MigrationRecorder
+
+
+def migrate_allow_guest_access(apps, schema_editor):
+    # if this migration has been run, drop out because logic depends on a field that no longer exists
+    if MigrationRecorder.Migration.objects.filter(
+        app="kolibriauth",
+        name="0017_remove_facilitydataset_allow_guest_access",
+        applied__isnull=False,
```
benjaoming, line 15: Why `__isnull=False`? Could it be better with a `.exclude(applied=True)` ?

benjaoming (approved):
Great solution! Would be nice with a confirmation from @mrpau-dev (currently reproducing the issue with Postgres) that it works, so would suggest to keep the PR open until then.

kolibri/core/device/migrations/0006_migrate_guest_access.py
ralphiee22: `applied` is actually a datetime field. It is not null when the migration has been applied.

benjaoming: Ah of course! Thanks!

PR #6490, opened 2020-01-22
Maintain channel order when importing new resources

### LINKED ISSUE #6376: Importing more content (or deleting resources) to a channel undoes its original position in channel list
### Observed behavior

~Similar to #5292, if you import more content to a channel, it will always be moved to the bottom of the Device > Channel page even if you had given it a different order (or didn't reorder it but just imported it in the past at some point). Because a lot of places in the app use the same order, we also see #5292~

Edit: This has been resolved for Kolibri when using Sqlite, but when using Postgresl, what seems to happen is that the newly modified channel goes to the _top_ of the list. It's because the channel in this case falls into this branch:

https://github.com/learningequality/kolibri/blob/0ee38838e5fc74d1f85b15c867f53ec4c261448c/kolibri/core/content/utils/annotation.py#L479-L482

~A somewhat mitigating factor is that newly updated channels will temporarily be moved to the top of the channel and have a green "New" label attached to it, so it will be easy to find immediately after the import.~

### Expected behavior

Updating a channel doesn't move it to the ~"bottom"~ top of the channel list.

### User-facing consequences

1. Users will need to move the channel again
1. Users might not immediately realize the channel moved 

### Errors and logs

…

### Steps to reproduce
1. Run Kolibri using Postgresql as DB
1. Run a similar scenario as described above

### Context
Kolibri 0.13

bransgithub:
### Summary

Fixes #6376 

Currently, when importing new resources for an already-installed channel, the affected channel is moved to the bottom of the list (after tasks are cleared and the page is refreshed). This fix affects the `calculate_next_order()` method, which should now only set the channel as the last item on the list if it is a new, uninstalled channel. This means any existing/installed channels will maintain their position on the list.

**Before**

The current behavior, starting at the available channels page:

We will import new resources for "Guía...", a channel that is already installed. We can see that it is moved to the top temporarily, after which it then shows up at the bottom.

| "New" channel is at the top | After task is cleared/page refreshed, it's moved to the bottom 
|------|-------|
|   |  |

**After**

We'll start with the following channel list:

4  after_orig

And add new resources for HP LIFE. Now, HP LIFE should keep its original position:

| "New" channel is at the top | After task is cleared/page refreshed, channel keeps its position
|------|-------|
| 5  after_add_new | 6  after_refresh_fixed |

And for good measure, here's what happens when an entirely new/uninstalled channel is imported:

| "New" channel is at the top | "New" channel is at the bottom 
|------|-------|
|   |  |

### Reviewer guidance

`calculate_next_order()` is used in a couple different places; here are the top-level 'callers' from what I could tell:
- `importcontent.py/_transfer()`
- `scanforcontent.py/handle()`
- `upgrade.py/import_external_content_dbs()`
- `deletecontent.py/delete_metadata()`

I don't believe this will alter any of their expected behavior, but I think this is worth confirming.

### References

…

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/core/content/utils/annotation.py
+++ b/kolibri/core/content/utils/annotation.py
@@ -478,6 +478,8 @@ def calculate_next_order(channel, model=ChannelMetadata):
     latest_order = model.objects.latest("order").order
     if latest_order is None:
         channel.order = 1
-    else:
+
+    if channel.order is None or channel.order == 0:
         channel.order = latest_order + 1
+
     channel.save()


indirectlylit:
Thanks @bransgithub!

FYI there is at least one test that might need updating:

```
=================================== FAILURES ===================================
___________________ ChannelOrderTestCase.test_channel_order ____________________
self = <kolibri.core.content.test.test_data_migrations.ChannelOrderTestCase testMethod=test_channel_order>
    def test_channel_order(self):
        ChannelMetadata = self.apps.get_model("content", "ChannelMetadata")
        pos = 1
        for channel in ChannelMetadata.objects.all():
>           self.assertEqual(channel.order, pos)
E           AssertionError: 0 != 1
kolibri/core/content/test/test_data_migrations.py:114: AssertionError
```

For info on testing, see https://kolibri-dev.readthedocs.io/en/develop/getting_started.html#automated-testing

bransgithub:
Ah, good catch! Hopefully it's good now; thanks!

kolibri/core/content/utils/annotation.py
```diff
@@ -478,6 +478,8 @@ def calculate_next_order(channel, model=ChannelMetadata):
     latest_order = model.objects.latest("order").order
     if latest_order is None:
         channel.order = 1
-    else:
+
+    if channel.order is None or channel.order == 0:
```
jonboiser, line 482: I think I understand the first part of this guard (order is None), but what case does `channel.order == 0` account for?

bransgithub: I found in some cases, specifically with some tests*, that channel order is actually set to 0 instead of None. I thought this would be safer since I wasn't 100% sure if channel order was set to 0 in other locations as well.

_*In `test_data_migrations.py/ChannelOrderTestCase/test_channel_order`, the channels returned from `ChannelMetadata.objects` all have an order of 0._

jonboiser: Based on the model impelementation [here](https://github.com/learningequality/kolibri/blob/a1e53e541bfd8f0b4c55ef2d8faf2222039a6ab6/kolibri/core/content/models.py#L280), the default for the order field is 0, so that explains that. And the tests seem to assume that the order is strictly positive (>0).

The field can also be blank, so that's why it _could_ be None, if there is some server code that actually would erase the order field and make it None.

bransgithub: That makes sense!

I did find that when importing an entirely new channel, the channel used in `annotation.py` (`set_channel_metadata_fields()` and subsequently `calculate_next_order()`) did return an order of None - so I think the order field is being erased somewhere during this process, like you said.

ralphiee22:
Ultimately, this PR guards against what `channel.order` should not be, which is `None` or `0`. Also guards updating a channel which already has `order` set.

ralphiee22 (approved):
Thanks for your contribution!

PR #6494, opened 2020-01-24
Less brittle upload step

DXCanas:
### Summary

Having the download step download the entire `dist` folder (where everything is uploaded) rather than naming each file explicitly.

Allows for the step to succeed despite previous steps not having happened (as is the model for our block-by-default triggered builds).

Also provides relief for the issues we occasionally face from the rPi image, as it somtimes fails silently. That being the case, no zip file is uploaded to Builkdite and the download portion of the GCS upload step fails. 

Going back to relase 0.13 as that's when rPi was introduced.

### Reviewer guidance

Builds should pass. Down the line, they should pass regardless of whether we decided to build the pi image (and any other triggered build) or not.

### References

Aims to fix the same issue (and a bit more) as #6402

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/.buildkite/setup_and_upload_artifact.sh
+++ b/.buildkite/setup_and_upload_artifact.sh
@@ -32,13 +32,7 @@ fi
 PYTHON_CMD="$PYTHON_PATH .buildkite/upload_artifacts.py"
 echo "Now excuting  upload artifacts script..."
 mkdir -p dist
-buildkite-agent artifact download 'dist/*.pex' dist/
-buildkite-agent artifact download 'dist/*.whl' dist/
-buildkite-agent artifact download 'dist/*.tar.gz' dist/
-buildkite-agent artifact download 'dist/*.deb' dist/
-buildkite-agent artifact download 'dist/*.exe' dist/
-# buildkite-agent artifact download 'dist/*.dmg' dist/
-buildkite-agent artifact download 'dist/*.zip' dist/
+buildkite-agent artifact download 'dist/*' dist/
 
 {
     buildkite-agent artifact download '*.exe' dist/ --step "Sign Windows installer"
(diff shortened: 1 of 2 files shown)


DXCanas:
Might we have to backport this to previous kolibri versions? like 0.11 and 0.10?

.buildkite/setup_and_upload_artifact.sh
```diff
@@ -32,13 +32,7 @@ fi
...
 echo "Now excuting  upload artifacts script..."
 mkdir -p dist
-buildkite-agent artifact download 'dist/*.pex' dist/
-buildkite-agent artifact download 'dist/*.whl' dist/
-buildkite-agent artifact download 'dist/*.tar.gz' dist/
-buildkite-agent artifact download 'dist/*.deb' dist/
-buildkite-agent artifact download 'dist/*.exe' dist/
-# buildkite-agent artifact download 'dist/*.dmg' dist/
-buildkite-agent artifact download 'dist/*.zip' dist/
+buildkite-agent artifact download 'dist/*' dist/
```
aronasorman, line 35: What are the disadvantages of doing it this way?

aronasorman: Because all I see are advantages

lyw07 (approved):
LGTM!

.buildkite/setup_and_upload_artifact.sh
DXCanas: I guess it's less explicit. And, because it's less specific, there is a chance you'll have an incomplete list of files to upload (which is kind of the point anyway).

PR #6504, opened 2020-01-30
Fix blank AppBar title on Learner Quiz Pages

### LINKED ISSUE #6323: after refreshing a quiz, app bar title disappears
### Observed behavior

quiz title disappears on refresh or direct navigation to URL:

### Expected behavior

quiz title should always be visible

### User-facing consequences

important context missing from page

### Steps to reproduce

* as a learner, start taking a quiz
* refresh

### Context

0.13.0 beta 7

bransgithub:
### Summary

Fixes #6323 

Currently, refreshing a quiz page or navigating to it directly results in a blank app bar title:

before

By fetching the LearnerClassroomResource based on the classId param, we can set the classroom name/title whenever the page loads:

after

### Reviewer guidance

…

### References

…

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/plugins/learn/assets/src/modules/examViewer/handlers.js
+++ b/kolibri/plugins/learn/assets/src/modules/examViewer/handlers.js
@@ -1,5 +1,6 @@
 import {
   ContentNodeResource,
+  ClassroomResource,
   ExamResource,
   ExamLogResource,
   ExamAttemptLogResource,
@@ -32,6 +33,7 @@ export function showExam(store, params, alreadyOnQuiz) {
     store.commit('CORE_SET_PAGE_LOADING', false);
   } else {
     const promises = [
+      ClassroomResource.fetchModel({ id: classId }),
       ExamResource.fetchModel({ id: examId }),
       ExamLogResource.fetchCollection({ getParams: examParams }),
       ExamAttemptLogResource.fetchCollection({ getParams: examParams }),
@@ -39,7 +41,9 @@ export function showExam(store, params, alreadyOnQuiz) {
     ];
     ConditionalPromise.all(promises).only(
       samePageCheckGenerator(store),
-      ([exam, examLogs, examAttemptLogs]) => {
+      ([classroom, exam, examLogs, examAttemptLogs]) => {
+        store.commit('classAssignments/SET_CURRENT_CLASSROOM', classroom);
+
         // Local copy of exam attempt logs
         const attemptLogs = {};
 


kolibri/plugins/learn/assets/src/modules/examViewer/handlers.js
```diff
@@ -32,14 +33,17 @@ export function showExam(store, params, alreadyOnQuiz) {
     store.commit('CORE_SET_PAGE_LOADING', false);
   } else {
     const promises = [
```
jonboiser, line 35: could you try using the `ClassroomResource`? You can import it from `kolibri.resources`. If you only want the name of the class, this is a lighter-weight Viewset, since it doesn't retrieve the lessons and quizzes too.

Another optimization would be to check to see if the `currentClassroom` is defined before making the API call to ClassroomResource, potentially saving a network call.

jonboiser: Just confirmed that `ClassroomResource.fetchModel({ id: classId })` works fine at line 36.

rtibbles: Yeah, rather than requiring another request - wouldn't this data be available from the class_summary_api call as well?

jonboiser: I don't think so, since this is the Learner Quiz viewer, not the Coach quiz viewer (where Class Summary is available)

rtibbles: Oh sorry - completely missed that. Disregard!

podenborg: Thanks for the advice everyone, and sorry for the delay. I've changed the handlers to use `ClassroomResource` instead of `LearnerClassroomResource`, and it seems to work great.

As for this suggestion:
> Another optimization would be to check to see if the currentClassroom is defined before making the API call to ClassroomResource, potentially saving a network call.

I definitely agree that this would save an API call, but I'm not sure if the way I'd do it is clean enough. 

Given that the API calls are effectively wrapped into `promises` and called altogether via `ConditionalPromise.all(promises).only`, would it be better readability-wise to have a separate check for `currentClassroom` >> Make a separate call to `ClassroomResource` >> Make the aforementioned `ConditionalPromise` call for the rest of the APIs?

bransgithub: Thanks for the advice everyone, and sorry for the delay! I've changed `LearnerClassroomResource` to `ClassroomResource` and it seems to work great. 

As for checking for the existence of `classroomResource` before making a call to the API, I'm not sure if the way I'd implement it would be clean enough. 

It looks like the API calls are wrapped into `promises` and then  effectively 'batch called' with `ConditionalPromise.all(promises).only()`. I think something like this would work: Check if `classroomResource` exists >> Make call to `ClassroomResource` if it doesn't >> Make the rest of the 'batch' calls - but I'm not entirely sure if the extra handful of lines I'd add would be worth the single API call - any advice one way or the other would be great!

indirectlylit:
hi @bransgithub - I think you need to rebase this in order for it to build. There might also be a front-end linting issue

jonboiser (approved):

indirectlylit:
thank you!

Issue #6482, opened 2020-01-20
Automatically recover from Malformed Database on disposable SQLite dbs


benjaoming:
# Observed behavior

A report has come on a Malformed Database exception for `~/.kolibri/process_cache/cache.db`. This should of course not happen, but it seems inevitable that it occurs in multi-process environments and/or because of unclean shutdowns.

## Manual work-around

The manual fix is to delete the database files, since they are disposable:

* `~/.kolibri/process_cache/cache.db`
* `~/.kolibri/job_storage.sqlite3`
* `~/.kolibri/notifications.sqlite3`

On another note, why is it called `cache.db` when the other databases are consistently called `.sq
slite3`?

## Automatic work-around proposals

1. Catch these exceptions and delete the database, prompting for a restart of Kolibri
1. Add a sanity check, opening the database and deleting it upon failure
1. Add a new management command that refreshes these smaller disposable databases

### Expected behavior

I think we can take this in two steps: Fail gracefully and add user guidance, possibly automatically purging a disposable database if harmless.

### User-facing consequences

Kolibri doesn't start in rare cases

### Errors and logs

```
$ kolibri status
Traceback (most recent call last):
  File "/usr/bin/kolibri", line 11, in <module>
    load_entry_point('kolibri==0.13.0', 'console_scripts', 'kolibri')()
  File "/usr/lib/python3/dist-packages/pkg_resources/__init__.py", line 542, in load_entry_point
    return get_distribution(dist).load_entry_point(group, name)
  File "/usr/lib/python3/dist-packages/pkg_resources/__init__.py", line 2569, in load_entry_point
    return ep.load()
  File "/usr/lib/python3/dist-packages/pkg_resources/__init__.py", line 2229, in load
    return self.resolve()
  File "/usr/lib/python3/dist-packages/pkg_resources/__init__.py", line 2235, in resolve
    module = __import__(self.module_name, fromlist=['__name__'], level=0)
  File "/usr/lib/python3/dist-packages/kolibri/utils/cli.py", line 28, in <module>
    from . import server
  File "/usr/lib/python3/dist-packages/kolibri/utils/server.py", line 17, in <module>
    from kolibri.core.deviceadmin.utils import schedule_vacuum
  File "/usr/lib/python3/dist-packages/kolibri/core/deviceadmin/utils.py", line 13, in <module>
    from kolibri.core.tasks.main import scheduler
[...]

Issue #6502, opened 2020-01-30
Tasks don't complete in Device - Channels module


intelliant01:
### Observed behavior

Tasks are no longer completing after an abrupt shutdown (power to the server unplugged) while an update task was ongoing.
The tasks were completing and had even cleared them prior to the abrupt shutdown.

### Expected behavior

Tasks should complete.

### User-facing consequences

Can't accomplish any task in the Device -> Channels module.

### Errors and logs

Herein are attached the logs of the last few days ever since the problem started.
[kolibri-2020-01-25.txt](https://github.com/learningequality/kolibri/files/4133387/kolibri-2020-01-25.txt)
[kolibri-2020-01-26.txt](https://github.com/learningequality/kolibri/files/4133388/kolibri-2020-01-26.txt)
[kolibri-2020-01-27.txt](https://github.com/learningequality/kolibri/files/4133389/kolibri-2020-01-27.txt)
[kolibri-2020-01-23.txt](https://github.com/learningequality/kolibri/files/4133390/kolibri-2020-01-23.txt)
[kolibri-2020-01-24.txt](https://github.com/learningequality/kolibri/files/4133391/kolibri-2020-01-24.txt)

### Steps to reproduce

1. Start a channel content update task
2. Simulate an abrupt shutdown of Kolibri while the task is still underway (Unplug the device)
3. Start device and Kolibri
4. No new tasks shall complete thereafter

### Context

Kolibri server - 0.3.6
Kolibri - 0.13.0
Ubuntu 18.04.3

indirectlylit:
Thanks for the report @intelliant01!

I'm a bit confused though. I believe the expected behavior for both 0.12.x and 0.13.0 is that any transfers that are outstanding during a hard shutdown will "fail" and need to be manually re-initiated.

Would you mind clarifying a couple things?

> Tasks are no longer completing after an abrupt shutdown (power to the server unplugged) while an update task was ongoing.
> The tasks were completing and had even cleared them prior to the abrupt shutdown.

Are you saying that in 0.12.x, content transfers would resume after a hard shutdown of Kolibri? I just tested on 0.12.9, and the behavior was that the transfer had disappeared completely, which I believe is the same behavior as in 0.13.0.

> Can't accomplish any task in the Device -> Channels module.

Are you saying that after a hard shutdown, content transfers working altogether and cannot be re-initiated?

intelliant01:
[...]

PR #6511, opened 2020-02-01
Fix cache db corruption

jredrejo:
### Summary
This PR tries to avoid corruption problems in the sqlite cache.db file, using two approaches:
1. Ensures the db is closed after every use, to avoid leaving a journal file open that could be written from different threads
2. In case the db is corrupted when opening it, deletes it and recreates i t

### Reviewer guidance
Does the code make sense?

### References
Relates: #6482 
Relates: #6195 

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/deployment/default/cache.py
+++ b/kolibri/deployment/default/cache.py
@@ -1,6 +1,8 @@
 import copy
 import os
+import shutil
 import sys
+from sqlite3 import DatabaseError
 
 from diskcache import Cache
 
@@ -12,9 +14,13 @@ cache_options = OPTIONS["Cache"]
 pickle_protocol = OPTIONS["Python"]["PICKLE_PROTOCOL"]
 
 diskcache_location = os.path.join(KOLIBRI_HOME, "process_cache")
-
-diskcache_cache = Cache(diskcache_location, disk_pickle_protocol=pickle_protocol)
-
+try:
+    diskcache_cache = Cache(diskcache_location, disk_pickle_protocol=pickle_protocol)
+except DatabaseError:
+    shutil.rmtree(diskcache_location, ignore_errors=True)
+    os.mkdir(diskcache_location)
+    diskcache_cache = Cache(diskcache_location, disk_pickle_protocol=pickle_protocol)
+diskcache_cache.close()
 # Default to LocMemCache, as it has the simplest configuration
 default_cache = {
     "BACKEND": "django.core.cache.backends.locmem.LocMemCache",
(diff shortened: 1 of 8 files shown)


indirectlylit:
@jredrejo do you think there's any chance this might also relate to #6502?

jredrejo:
> @jredrejo do you think there's any chance this might also relate to #6502?

I don't think so, looking at his logs, I only can find a network error while importing the channel. In case there is a corruption problem it is not shown in the logs, what is strange. Anyway, the corruption could exist either in the job_storage database or in the same channel content database, but not in cache.db.

P.S: I have tried several times to reproduce his problem, shutting down kolibri in the middle of the import process and it always recovered it correctly after starting it again , so I have not been able to reproduce it :(

radinamatic:
@jredrejo Just checking if this needs any addition in the user docs...?

jredrejo:
> @jredrejo Just checking if this needs any addition in the user docs...?

No need, all the process is done automatically by the backend. There's not any user interaction.

rtibbles:
> Could we use a thread Rlock instead of a diskcache Rlock for threads not using diskcache at all? @rtibbles If not we should document somehow or find a way to close the diskcache database whenever db_task_write_lock is used.

I used this in preference to `threading` RLock to provide cross-process safety. It's being invoked as a context using `with`, so there should be very limited circumstances under which it does not perform properly cleanup.

rtibbles:
If it's not properly closing it - it might be good to make a wrapper function that includes the DB closing, and then we just use that instead of invoking RLock directly.

jredrejo:
> If it's not properly closing it - it might be good to make a wrapper function that includes the DB closing, and then we just use that instead of invoking RLock directly.

@rtibbles  done!

kolibri/deployment/default/cache.py
```diff
@@ -12,9 +14,13 @@
 pickle_protocol = OPTIONS["Python"]["PICKLE_PROTOCOL"]
 
 diskcache_location = os.path.join(KOLIBRI_HOME, "process_cache")
-
-diskcache_cache = Cache(diskcache_location, disk_pickle_protocol=pickle_protocol)
-
+try:
+    diskcache_cache = Cache(diskcache_location, disk_pickle_protocol=pickle_protocol)
+except DatabaseError:
+    shutil.rmtree(diskcache_location, ignore_errors=True)
```
lyw07, line 20: would it be necessary to call `diskcache_cache.close()` close the connection of the old Cache before or after removing the old cache.db? Thank you!

jredrejo: The old db is removed only  it if it's not able to open it because it's corrupted, so it does not need to be closed.

lyw07: thank you for the explanation!

lyw07 (approved):
tested on my machine and saw that the old cache db was removed when DatabaseError was raised. 👍

PR #6523, opened 2020-02-06
Fix corrupted sqlite databases

jredrejo:
### Summary
This PR tries to automate the repairing and, if repairing is not possible, creating again sqlite damaged databases.
It deals both with Django ORM connections (for `notifications.sqlite3`) and with SQLAlchemy connections (for `job_storage.sqlite3`).

Ideally these procedures should never been executed. Whenever we find in the logs they have been executed we should find out the reason and fix it if it's due to some code problem.

### Reviewer guidance
Does the code make sense?
Are there other places in the code where we should enclose db connections in a `try...catch` to capture `DatabaseError` exceptions and call the `repair_sqlite_db` function?

### References
Relates: #6482

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/core/sqlite/utils.py
+++ b/kolibri/core/sqlite/utils.py
@@ -23,16 +23,20 @@ def common_clean(db_name, db_file):
 def regenerate_database(connection):
     # procedure to create from scratch a sqlite database when using Django ORM
     from django.db.migrations.recorder import MigrationRecorder
+
     connection.close()
-    common_clean(connection.alias, connection.get_connection_params()['database'])
+    common_clean(connection.alias, connection.get_connection_params()["database"])
     if connection.alias == "notifications_db":
         logger.error("Regenerating {}".format(connection.alias))
         # delete the db migrations and run them again
         connection_migrations = MigrationRecorder(connection).Migration
-        connection_migrations.objects.filter(app='notifications').delete()
+        connection_migrations.objects.filter(app="notifications").delete()
         call_command(
-            "migrate", interactive=False, verbosity=False,
-            app_label="notifications", database="notifications_db"
+            "migrate",
+            interactive=False,
+            verbosity=False,
+            app_label="notifications",
+            database="notifications_db",
         )
         call_command("migrate", interactive=False, verbosity=False)
 
@@ -52,16 +56,17 @@ def repair_sqlite_db(connection):
     else:
         orm = "django"
         conn_name = connection.alias
-        original_path = connection.get_connection_params()['database']
+        original_path = connection.get_connection_params()["database"]
 
-    fname = "{con}_{dtm}.dump".format(con=conn_name,
-                                      dtm=datetime.now().strftime("%Y-%m-%d_%H-%M-%S"))
+    fname = "{con}_{dtm}.dump".format(
+        con=conn_name, dtm=datetime.now().strftime("%Y-%m-%d_%H-%M-%S")
+    )
     if not os.path.exists(dest_folder):
         os.makedirs(dest_folder)
     backup_path = os.path.join(dest_folder, fname)
     copyfile(original_path, backup_path)
 
-    if orm == 'sqlalchemy':
+    if orm == "sqlalchemy":
         # Remove current file, it will be automatically regenerated
         common_clean(conn_name, original_path)
         logger.error("Regenerating {}".format(connection.name))
--- a/kolibri/core/sqlite/utils.py
+++ b/kolibri/core/sqlite/utils.py
@@ -13,9 +13,9 @@ logger = logging.getLogger(__name__)
 
 
 def common_clean(db_name, db_file):
+    # let's remove the damaged db files
     if settings.DATABASES["default"]["ENGINE"] != "django.db.backends.sqlite3":
         return
-    # let's remove the old files
     os.remove(db_file)
     logger.error("{} is corrupted".format(db_name))
 
--- a/kolibri/core/sqlite/utils.py
+++ b/kolibri/core/sqlite/utils.py
@@ -1,0 +1,89 @@
+import io
+import logging
+import os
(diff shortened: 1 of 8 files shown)


kolibri/core/tasks/scheduler.py
```diff
@@ -60,7 +61,10 @@ def __init__(self, queue=None, connection=None):
             if connection is None:
                 connection = self.queue.storage.engine
         elif connection:
-            self.queue = queue(connection=connection)
```
rtibbles: I think it might be better to do this check inside `main.py` can check that the database is working at the initialization of the connection: https://github.com/learningequality/kolibri/blob/release-v0.13.x/kolibri/core/tasks/main.py#L44

kolibri/core/sqlite/utils.py
```diff
@@ -0,0 +1,94 @@
...
+    fixed_db_path = "{}.2".format(original_path)
+    with io.open(fixed_db_path, **KWARGS_IO_WRITE) as f:
+        # If the connection hasn't been opened yet, then open it
+        try:
+            for line in connection.connection.iterdump():
+                f.write(line)
+            connection.close()
+            copyfile(fixed_db_path, original_path)
+            # let's check if the tables are there:
+            cursor = connection.cursor()
```
lyw07, line 86: I notice that here, instead of executing the sql command in the next line, it goes back to the function https://github.com/learningequality/kolibri/pull/6523/files#diff-0d228a202c863487208a91d0c31b1614R36 again.

jredrejo: @lyw07  I can't reproduce what you said, could you pass me the corrupted db you're using to test?

lyw07: sorry! I guess I was adding too many logging messages and somewhere I messed up the code 😐I tested again and it works now

lyw07:
I'm sorry I'm not too sure what condition should apply to this as you mentioned here:
> Are there other places in the code where we should enclose db connections in a try...catch to capture DatabaseError exceptions and call the repair_sqlite_db function?

Should we check the connection every time we connect to the notifications db? I'm wondering if a call to save the notification would be applicable here: https://github.com/learningequality/kolibri/blob/release-v0.13.x/kolibri/core/notifications/api.py#L86. Thank you!

kolibri/core/tasks/scheduler.py
jredrejo: @rtibbles done

jredrejo:
> > Are there other places in the code where we should enclose db connections in a try...catch to capture DatabaseError exceptions and call the repair_sqlite_db function?
> 
> Should we check the connection every time we connect to the notifications db? I'm wondering if a call to save the notification would be applicable here: https://github.com/learningequality/kolibri/blob/release-v0.13.x/kolibri/core/notifications/api.py#L86. Thank you!

@lyw07  I  have done multiple checks and found two more points where it has to be added, the one you mention can be caught before in `notifications/tasks.py` . I have not been able to find more possible points where the db is accessed.

lyw07:
@jredrejo Hi! sorry I thought I made a comment yesterday but it turns out I didn't... I just have one last question - Do you know if this (https://github.com/learningequality/kolibri/blob/release-v0.13.x/kolibri/core/exams/models.py#L120) has been handled by the existing code? I'm not too sure about this so it would be great if you can help me with it here. Thank you so much!

jredrejo:
> @jredrejo Hi! sorry I thought I made a comment yesterday but it turns out I didn't... I just have one last question - Do you know if this (https://github.com/learningequality/kolibri/blob/release-v0.13.x/kolibri/core/exams/models.py#L120) has been handled by the existing code? I'm not too sure about this so it would be great if you can help me with it here. Thank you so much!

Yes, the only way to delete an exam is going through the /coach/#/uuid/plan/lessons  . Every time you pass to /coach the db is tested and regenerated.

kolibri/plugins/coach/api.py
```diff
@@ -190,9 +195,12 @@ def list(self, request, *args, **kwargs):
...
         # Use super on the parent class to prevent an infinite recursion.
-        response = super(viewsets.ReadOnlyModelViewSet, self).list(
-            request, *args, **kwargs
-        )
+        try:
+            response = super(viewsets.ReadOnlyModelViewSet, self).list(
+                request, *args, **kwargs
+            )
+        except DatabaseError:
+            repair_sqlite_db(connections["notifications_db"])
```
lyw07, line 205: similarly, do you think we should return an empty response here?

kolibri/plugins/coach/api.py
```diff
@@ -180,6 +183,8 @@ def get_queryset(self):
                 )
             except (LearnerProgressNotification.DoesNotExist):
                 return []
+            except DatabaseError:
+                repair_sqlite_db(connections["notifications_db"])
```
lyw07, line 188: do you think we need to return [] here?

jredrejo: yes, I have added

kolibri/plugins/coach/api.py
jredrejo: No, here the function does not returns, it has to continue, anyway I have seen another kind of error if the db file does not even exist and has added it to the exceptions.

kolibri/plugins/coach/api.py
lyw07: oh I guess you were trying to return [] after `repair_sqlite_db`?

jredrejo: :man_facepalming: yes, of course. I should not code at some hours in the day...

lyw07 (approved):
LGTM! Thank you, José!

Issue #6443, opened 2020-01-08
Select resources text broken on IE11


mrpau-richard:
### Observed behavior
Select resources text did not display the full length on IE11 browser.

### Expected behavior

…

### User-facing consequences

…

### Errors and logs
![Screen Shot 2020-01-08 at 11 59 20 AM](https://user-images.githubusercontent.com/8663934/71949583-0780c800-320f-11ea-95c0-beb016b5d95a.png)

### Steps to reproduce

…

### Context

Tell us about your environment, including:
 * Kolibri version: 0.13.0
 * Operating system: Windows 7
 * Browser: IE11

jeepurs:
I may have seen a related bug, in the form of the button hanging off the border of the expected square.

UN Women country server (Mexico): https://mexico-kolibri.mylearningpathway.org
Browser: Chrome 77.0.3865.75
Version: v0.12

I see this bug when I open developer tools and the window is reduced:

![image](https://user-images.githubusercontent.com/5150483/73578756-47655480-4435-11ea-8204-1e937c0676fa.png)

jonboiser:
Fixed in #6506 by giving the "select resource" button more room in the card, so it can grow into the card instead of into the right overflow area.

radinamatic:
Apparently the cards in other languages are more greedy, and the above wasn't _enough_ room. 

![Win7 (new start)  Running  - Oracle VM VirtualBox_303](https://user-images.githubusercontent.com/1457929/74160994-9160ee00-4c1e-11ea-8701-ce57b68bcd4b.png)

jonboiser:
@nucleogenesis I think maybe using the flex-grow for that cell approach might be best for this button if there's no optimal width for this element.

jonboiser:
@radinamatic Do you think french is has the longest message for this particular button? We can just keep allocating more space in the min-width rule for the button to accommodate french.

radinamatic:
I checked all the languages, @jonboiser, and French and Brazilian Portuguese are the only ones that are cut out.

 
![Win7 (new start)  Running  - Oracle VM VirtualBox_308](https://user-images.githubusercontent.com/1457929/74260748-5e8c2800-4cfa-11ea-9c45-664759ac1642.png)

jonboiser:
@radinamatic Thanks for going through all the languages. I think the initial solution was the least invasive, but I'm going to have to make a bigger change to the layout so it works for all languages.

PR #6537, opened 2020-02-11
Fix appearance of long label (in certain locales) and remove disabling for "select resources" buttons. 

### LINKED ISSUE #6528: Cannot import a second channel from Studio in IE11
### Observed behavior
Noticed while testing the 0.13.1b1.

When using IE11 it seems to be possible to import resources only from one public channel on Studio, as when you make a second attempt and press the `Select resources` button for any other  channel, nothing happens (user would expect to see the channel topic tree loading at this point).

- It is still possible to import unlisted channels using the token.
- If user stops and restarts Kolibri, they are now able to import, but only once.
- It is possible to import resources correctly from more than one channel in FF & Chrome. 

### Expected behavior
User needs to be able to import from more than one channel.

### User-facing consequences
Impossibility to import content on IE11.

### Errors and logs
Nothing visible in the console or network tab

### Steps to reproduce
1. Open Kolibri in IE11 on Windows 7.
2. Import a resource from one channel from Studio
3. When the task is finished, try importing from another channel.

### Context
 * Kolibri version: 0.13.1b1
 * Operating system: Windows 7
 * Browser: IE11

I'll install on Windows 10 and see if I can replicate on Edge.

cc @jonboiser @indirectlylit @lyw07

jonboiser:
### Summary

1. Implements a locale-independent fix for #6443 by making all three major columns in the relevant card component direct siblings of each other
. This ensures the "select resources" button isn't pushed off to the right.

Screenshot in Portuguese
Screen Shot 2020-02-11 at 12 59 58 PM

2. Removes logic that would disable the "Selected resource" button if any "active" tasks where in the task queue, which should fix #6528 , where in this button was functionally disabled, but was not styled at such.

3. Deletes some dead CSS, and the now-dead code for the activeTaskList getter
### Reviewer guidance

1. In IE11 and other browsers, check to see the "Selected resource" button is not cut off on the right edge for different locales.
1. In IE11 specifically, check to see that the importing resources workflow is not blocked at the Channel list when there are active tasks in the queue.

### References

…

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/plugins/device/assets/src/modules/manageContent/index.js
+++ b/kolibri/plugins/device/assets/src/modules/manageContent/index.js
@@ -98,15 +98,6 @@ export default {
         return null;
       };
     },
-    // Tasks that are active, complete, or failed.
-    // Canceling and canceled tasks are filtered here
-    // to hide them from users, but still let us clean
-    // them up when finished.
-    activeTaskList(state) {
-      return state.taskList.filter(
-        task => task.status !== TaskStatuses.CANCELING && task.status !== TaskStatuses.CANCELED
-      );
-    },
     managedTasks(state) {
       // Tasks that we want to show in the task manager - ignore channel metadata tasks here.
       return state.taskList.filter(
--- a/kolibri/plugins/device/assets/src/views/ManageContentPage/ChannelPanel/ChannelDetails.vue
+++ b/kolibri/plugins/device/assets/src/views/ManageContentPage/ChannelPanel/ChannelDetails.vue
@@ -47,6 +47,8 @@
         <slot name="belowdescription"></slot>
       </div>
     </div>
+
+    <slot name="append"></slot>
   </div>
 
 </template>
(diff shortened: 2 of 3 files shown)


radinamatic:
Looking good when I checkout your branch in my dev env, even when I try with the longer string `Sélectionnez les ressources` (we may change that in 0.14). 

Selection_321

I'll install the `exe` asset on Windows VM next to confirm and approve.

radinamatic:
Hm, it does not seem to be working well in IE11/W7  :confused: 

Peek 2020-02-12 02-09

Import works OK though, even when there are completed but not cleared tasks in the queue :+1: 


Width in Firefox looks OK:

ff

jonboiser:
@radinamatic Does the layout change to a columnar mobile layout if you keep narrowing the window on either IE or firefox?

radinamatic:
Eventually it does, but not really smooth transitions... That's IE for you, I guess :unamused: 

IE

jonboiser:
Yeah, the cards don't look great even on Chrome at that in-between width. I think the responsive layout should probably kick in at the "medium" screen breakpoint to address the general lack of responsiveness at between-medium-and-small range.

At least the label doesn't get cut at the larger window width.

indirectlylit:
let's open a subsequent issue to improve the responsiveness of this screen

radinamatic (approved):
Button is functional, let's deal with the width on IE and window responsiveness in subsequent PRs.

Issue #6541, opened 2020-02-12
unable to see channel tree and import content from disk with postgresql


lyw07:
### Observed behavior

When I'm trying to import channel from studio/disk, I'm not able to see the detailed channel tree, although I'm still able to import the entire channel by clicking `Select all`.

<img width="1211" alt="Screen Shot 2020-02-12 at 10 32 20 AM" src="https://user-images.githubusercontent.com/19424916/74365661-5fb56780-4d83-11ea-97e6-8161e5f7dae0.png">

I can confirm that this is not an issue in 0.12.9. It doesn't exist in 0.13.0 with sqlite either.

Another issue is that I'm unable to import content from disk.

### Expected behavior

Users should be able to see what content is available in the channel

### User-facing consequences

Users don't know what content is available

### Errors and logs

For the second error:
```
ERROR    Internal Server Error: /device/api/importexportsizeview
Traceback (most recent call last):
  File "/Users/lingyiwang/.venvs/test/lib/python2.7/site-packages/kolibri/dist/django/core/handlers/exception.py", line 41, in inner
    response = get_response(request)
  File "/Users/lingyiwang/.venvs/test/lib/python2.7/site-packages/kolibri/dist/django/core/handlers/base.py", line 187, in _get_response
    response = self.process_exception_by_middleware(e, request)
  File "/Users/lingyiwang/.venvs/test/lib/python2.7/site-packages/kolibri/dist/django/core/handlers/base.py", line 185, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/Users/lingyiwang/.venvs/test/lib/python2.7/site-packages/kolibri/dist/django/views/decorators/csrf.py", line 58, in wrapped_view
    return view_func(*args, **kwargs)
  File "/Users/lingyiwang/.venvs/test/lib/python2.7/site-packages/kolibri/dist/django/views/generic/base.py", line 68, in view
    return self.dispatch(request, *args, **kwargs)
  File "/Users/lingyiwang/.venvs/test/lib/python2.7/site-packages/kolibri/dist/rest_framework/views.py", line 495, in dispatch
    response = self.handle_exception(exc)
  File "/Users/lingyiwang/.venvs/test/lib/python2.7/site-packages/kolibri/dist/rest_framework/views.py", line 455, in handle_exception
    self.raise_uncaught_exception(exc)
[...]

PR #6544, opened 2020-02-12
convert contentnode ids to hex strings in channel stats during import

lyw07:
### Summary

This PR is to fix two issues:
1. when Kolibri connects to postgresql, the ids stored in the variable `channel_stats` are UUID objects. So when we try to search the contentnode id's hex string in `channel_stats` during import, it thinks the contentnode id doesn't exist in the dictionary, thus making `num_coach_contents`, `coach_content`, `total_resources` and `importable` all False or 0 regardless their actual value.

2. users are unable to import content from disk with postgresql. This is because `UUID(identifier, version=4)` is creating a UUID object on top of the UUID object `identifier`.

### Reviewer guidance

please check if channel import works when connecting to postgresql

### References

fixes https://github.com/learningequality/kolibri/issues/6541

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/core/content/serializers.py
+++ b/kolibri/core/content/serializers.py
@@ -11,6 +11,7 @@ from kolibri.core.content.models import ContentNode
 from kolibri.core.content.models import File
 from kolibri.core.content.models import Language
 from kolibri.core.fields import create_timezonestamp
+from django.db import connection
 
 
 class DynamicFieldsModelSerializer(serializers.ModelSerializer):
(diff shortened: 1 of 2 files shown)


kolibri/core/content/serializers.py
```diff
@@ -437,7 +438,16 @@ class Meta:
 
     @property
     def channel_stats(self):
-        return self.context["channel_stats"]
+        channel_stats = self.context["channel_stats"]
+        if connection.vendor == "postgresql":
```
jonboiser: maybe for a later date, but I've grepped this line `connection.vendor == "postgresql"` and it appears in 10 different places. maybe we can extract it to a utility later.

kolibri/core/content/serializers.py
```diff
@@ -437,7 +438,16 @@ class Meta:
...
+            # Convert channel IDs to hex strings since they are stored as UUID
+            # objects when Kolibri connects to a PostgreSQL database
+            stats_convert_id = {}
+            for key in channel_stats.keys():
```
jonboiser, line 446: is `channel_stats` always guaranteed to be a Dict? is there ever a chance it's a None?

lyw07: ah that's right! thank you for pointing this out. I will change that part

jonboiser (approved):

PR #6559, opened 2020-02-13
Allow None in validate_language_id

### LINKED ISSUE #6557: cannot change device language to 'browser default'
### Observed behavior

Trying to set the default device language as 'Browser default' does not work:

image

### Expected behavior

Setting should work or be removed

### User-facing consequences

broken behavior in device settings

### Errors and logs

* Request URL: `http://localhost:8000/api/device/devicesettings/`
* Request Method: `PATCH`
* Status Code: `400 Bad Request`
* Response: `language_id`

### Steps to reproduce

Try to set language to browser default

### Context

0.13.1 beta 2
mac
devserver

jonboiser:
Fixes `language_id` validator for the `DeviceSettingViewset` by allowing `None` as a value.

Fixes #6557

### DIFF
--- a/kolibri/core/device/serializers.py
+++ b/kolibri/core/device/serializers.py
@@ -38,7 +38,7 @@ class DeviceSerializerMixin(object):
         """
         Check that the language_id is supported by Kolibri
         """
-        if not check_for_language(language_id):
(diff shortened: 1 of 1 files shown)


indirectlylit (approved):
tested and it allows the setting to be saved

PR #6403, opened 2019-12-19
Improve startup sanity checks

rtibbles:
### Summary
Adds a DB migration sanity check
Moves the content directory sanity check to happen before any command that invokes Django

### Reviewer guidance
Follow the steps outlined https://github.com/learningequality/kolibri/issues/5877#issuecomment-566479886

Have marked as draft as I think this is mildly risky and not completely needed for 0.13.0.

### References
Fixes #5877 

----

### Contributor Checklist

PR process:

- [x] PR has the correct target branch and milestone
- [x] PR has 'needs review' or 'work-in-progress' label
- [x] If PR is ready for review, a reviewer has been added. (Don't use 'Assignees')
- [ ] If this is an important user-facing change, PR or related issue has a 'changelog' label
- [ ] If this includes an internal dependency change, a link to the diff is provided

Testing:

- [ ] Contributor has fully tested the PR manually
- [ ] If there are any front-end changes, before/after screenshots are included
- [ ] Critical user journeys are covered by Gherkin stories
- [ ] Critical and brittle code paths are covered by unit tests

### Reviewer Checklist

- Automated test coverage is satisfactory
- PR is fully functional
- PR has been tested for [accessibility regressions](http://kolibri-dev.readthedocs.io/en/develop/manual_testing.html#accessibility-a11y-testing)
- External dependency files were updated if necessary (`yarn` and `pip`)
- Documentation is updated
- Contributor is in AUTHORS.md

Issue #6197, opened 2019-12-03
Make CLI commands accessible as pure Python API


kollivier:
### Observed behavior

There seems to be some strange behavior for click where it shuts down the server automatically when there is not an active, user-created terminal running. This makes us unable to start and stop the Kolibri server from a GUI application such as the Mac app, and likely Android as well.

While there is a workaround we can use to sidestep click and call the Python functions directly,  this requires us to manually know how to initialize Kolibri and do so before calling each command, which may change over time and may eventually require different apps for different Kolibri versions.

This also means we cannot switch the apps to running a downloaded Kolibri pex file, as the problem occurs with the pex file as well. We need to extract the source and put it on the path to run it. 

It would be good to have a way to run the startup and shutdown process independently from click commands, either as pure python functions or an alternate cli mode. 

### Expected behavior

Starting Kolibri from within a GUI application by calling the user-facing startup commands works.

### User-facing consequences

Until this is resolved, users will be unable to update their Kolibri version without also updating the app.

### Errors and logs

…

### Steps to reproduce

Call `kolibri.utils.cli.main` or `kolibri.pex start --foreground` from inside a GUI application not running from the terminal.

### Context

Kolibri 0.13.0b3
macOS 10.14.5 (Mojave)
Embedded Safari (Not sure if version aligns with actual browser version)

PR #6580, opened 2020-02-18
Refactor migration check and add regression tests

benjaoming:
### Summary

Changes:

* Some assumptions about what "checks" should do are a bit unintuitive to me when our checks are performing migrations all by themselves.
* I also reworked the `initialize()` function to be easier to read as a sequential piece of logic.
* Added two regression tests.

Reading the code in sequential order, I also find it a bit weird, why we only check and auto-run migrations after checking if the version has been bumped (which runs migrations). I would like to have a comment on that behavior and include it in the code. CC: @rtibbles 

Since the changes in 0.13.1 already went through some testing, I'm happy to postpone this change.

### Reviewer guidance

I would like a review that tries to spot additional oddities that we might fix? I mean, now that `initialize()` is assembled into one function, is it easier to see what is happening?

### References

#6403 

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/utils/cli.py
+++ b/kolibri/utils/cli.py
@@ -32,7 +32,6 @@ from .sanity_checks import check_database_is_migrated
 from .sanity_checks import check_default_options_exist
 from .sanity_checks import check_log_file_location
 from .sanity_checks import check_other_kolibri_running
-from .sanity_checks import migrate_databases
 from .system import become_daemon
 from kolibri.core.deviceadmin.utils import IncompatibleDatabase
 from kolibri.core.upgrade import matches_version
@@ -51,6 +50,9 @@ from kolibri.utils.conf import KOLIBRI_HOME
 from kolibri.utils.conf import LOG_ROOT
 from kolibri.utils.conf import OPTIONS
 from kolibri.utils.logger import get_base_logging_config
+from kolibri.utils.sanity_checks import check_django_stack_ready
+from kolibri.utils.sanity_checks import DatabaseInaccessible
+from kolibri.utils.sanity_checks import DatabaseNotMigrated
 
 
 logger = logging.getLogger(__name__)
@@ -88,6 +90,21 @@ def should_back_up(kolibri_version, version_file_contents):
         # And the new version is not a dev version
         and "dev" not in kolibri_version
     )
+
+
+def conditional_backup(kolibri_version, version_file_contents):
+    if should_back_up(kolibri_version, version_file_contents):
+        # Non-dev version change, make a backup no matter what.
+        from kolibri.core.deviceadmin.utils import dbbackup
+
+        try:
+            backup = dbbackup(version_file_contents)
+            logger.info("Backed up database to: {path}".format(path=backup))
+        except IncompatibleDatabase:
+            logger.warning(
+                "Skipped automatic database backup, not compatible with "
+                "this DB engine."
+            )
 
 
 def get_version():
@@ -157,6 +174,22 @@ initialize_params = base_params + [
     pythonpath_option,
     skip_update_option,
 ]
+
+
+def _migrate_databases():
+    """
+    Internal help function:
+
+    Try to migrate all active databases. This should not be called unless Django has
+    been initialized.
+    """
+    from django.conf import settings
+
+    for database in settings.DATABASES:
+        call_command("migrate", interactive=False, database=database)
+
+    # load morango fixtures needed for certificate related operations
+    call_command("loaddata", "scopedefinitions")
 
 
 def get_initialize_params():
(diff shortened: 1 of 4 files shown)


kolibri/utils/cli.py
```diff
@@ -333,6 +357,29 @@ def initialize(skip_update=False):
         # our Kolibri version based update logic.
         run_plugin_updates()
 
+    check_content_directory_exists_and_writable()
+
+    if not skip_update:
```
lyw07: I'm wondering if this code block can be combined with the one above here: https://github.com/learningequality/kolibri/pull/6580/files#diff-8f9ca278df35a8b43572c4b10dafe306R355 as they have the same condition. Thank you!

rtibbles:
I will need to review this at further length - as I am fairly sure there were significant reasons for the separation of operations that I had put here.

benjaoming:
> I will need to review this at further length - as I am fairly sure there were significant reasons for the separation of operations that I had put here.

When I was running pre-commit hooks, I wondered if they had been separated in order to avoid a McCabe complexity check failing on `initialize`?

kolibri/utils/cli.py
benjaoming: That could make a lot of sense for readability, thanks!

rtibbles:
I think at least part of the motivation was separating out the checks that initialization that needs Django versus the initialization that doesn't - so possible that your changes do fit in with that.

I think it would be better to target these changes to develop, and do some more aggressive refactor, where we separate out the general Kolibri programmatic API from the purely CLI API, as per this issue https://github.com/learningequality/kolibri/issues/6197

benjaoming:
Makes sense to retarget for develop, not sure this needs to wait for a refactor as changes should not be affected by a refactor with undecorated `cli` functions. Perhaps this just makes the refactor easier as it starts splitting out some of the logic from `KolibriDjangoCommand.invoke`.

benjaoming:
Github had issues when this PR was last rebased. Trying force push to run Travis again.

kolibri/utils/cli.py
indirectlylit: To reduce indentation and make the function easier to understand, I usually prefer to "bail early" rather than nesting logic inside conditionals. Concretely:

```python
def initialize(skip_update=False):

    # ...always do this stuff...

    if skip_update:
        return

    # ...conditionally do this stuff...
```

benjaoming: @indirectlylit that's a good structure :+1: it's a bit hard to make the `initialize()` function prettier - there are several of the checks and upgrade behaviors where the order and conditions seem sensitive. I can't tell for sure though, it requires a lot of pondering for each line.

I've gathered it a bit and moved the content directory check now.

benjaoming: I hope the change is okay now, if I clean up `initalize()` more, I fear introducing regressions in the regression test PR :)

indirectlylit (approved):
I don't see any obvious issues with this, and I like the change where we move the DB migration logic outside of the sanity checks.

Shall we merge in and file follow-up issues if necessary?

kolibri/utils/cli.py
```diff
@@ -232,13 +259,12 @@ def __init__(self, *args, **kwargs):
         super(KolibriDjangoCommand, self).__init__(*args, **kwargs)
 
     def invoke(self, ctx):
+        """
+        Initialize Kolibri and run sanity checks.
+        """
         # Check if the current user is the kolibri user when running kolibri from Debian installer.
```
rtibbles: This comment should be moved along with the code it describes.

rtibbles:
This seems fine to merge, the only immediate thing that stands out is the orphaned comment!

kolibri/utils/cli.py
rtibbles:
```suggestion
```

benjaoming:
Fixing that!

kolibri/utils/cli.py
```diff
@@ -274,13 +300,16 @@ def _setup_django(debug):
...
 
 
-def initialize(skip_update=False):
+def initialize(skip_update=False):  # noqa: max-complexity=12
     """
     Currently, always called before running commands. This may change in case
     commands that conflict with this behavior show up.
     """
+    setup_logging(debug=get_debug_param())
     params = get_initialize_params()
```
rtibbles, line 307:
```suggestion
    # Check if the current user is the kolibri user when running kolibri from Debian installer.
```

benjaoming: This should be fixed... also seems Github's suggestion system isn't working that well, at least here it displays as if you are suggesting to insert an orphaned comment instead of a necessary function call :)

kolibri/utils/cli.py
benjaoming: Thanks for spotting that, fixed and pushed

benjaoming:
Thanks for reviewing - I think all feedback has been addressed!

Issue #5409, opened 2019-04-23
Pre-commit hooks interfering with developer workflows


bjester:
### Observed behavior
Talking with @indirectlylit, some observed issues are that the existing pre-commit hooks:
- are very slow
- have issues with GUIs due to virtual envs
- interfere with in-progress commit dev workflows
- maintain rule definitions separately from Travis & dev tooling

### Expected behavior

Ideally, pre-commit hooks:
- [ ] should run as fast as possible #6583
- [ ] should support developer tools used by the team, like Git GUIs
- [ ] should not interfere with in-progress commit dev workflows
- [ ] should operate on a consistent set of rule definitions across Travis and other dev tooling #6583

### User-facing consequences

None

### Issue resolution paths
Obtain feedback and buy-in from contributors regarding pre-commit hooks and move forward with either:
1. Removing pre-commit hooks to focus on increasing the hardiness of Travis and other dev tooling for protecting code quality
2. Reworking pre-commit hooks to resolve issues in the current implementation affecting developer workflows

### References
https://github.com/learningequality/kolibri/pull/5405
https://learningequality.slack.com/archives/CB37UM23A/p1555759651088400

benjaoming:
> maintain rule definitions separately from Travis & dev tooling

Deviations between the two is a bug that should be addressed. Pre-commit can be run in CI as well as from a local environment.

Pre-commit hooks can be removed. If you do this, only CI will lint your code.

benjaoming:
I expanded the `exclude` pattern in #6583 - `pre-commit` only runs on files that are changed in commits, but I'm wondering if this is a pattern that should be expanded gradually in the same manner to achieve the speed perceived as lacking.

indirectlylit:
closing given work in #6583.

Let's re-open a new issue if necessary

benjaoming:
Yes, if there are further issues, please separate out the feedback like this:

* Speed: Report speed issues on a per-hook basis so it's possible to understand if a particular hook has issues
* Dev tools: Please report exactly the GUIs that are causing issues so it can be investigated
* In-commit workflow: Pre-commit hooks are designed to interfere with developer's workflows when they try to commit something they shouldn't. If there are false positives, this would be an issue with that particular hook that should be fixed.

indirectlylit:
This is not an urgent issue, but  I migrated the last point here: https://github.com/learningequality/kolibri/issues/6607

[...]

PR #6583, opened 2020-02-19
Align local pre-commit with CI-based linting (Tox and Travis)

benjaoming:
# Summary

* Removes frontendlint and pythonblack tox entries (they are now part of "pythonlint3")
* ...also removes said entries from Travis, hence quicker
* All calls to pre-commit now happen through the same configuration
* Caching on travis potentially improved

Does anyone else get frustrated when their code breaks a CI linting check because black isn't in pre-commit?

Having it in pre-commit is much faster, as it just lints changed files.

### Reviewer guidance

~I didn't care to pin the black version. Does it matter?~

Edit: We use Black without pinning in tox.ini as well, so we always run the latest version.

### References

This partly solves #5409

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -65,3 +65,8 @@ repos:
         kolibri/core/logger/migrations/0003_auto_20170531_1140\.py|
         kolibri/core/logger/migrations/0005_auto_20180514_1419\.py|
       )$
+-   repo: https://github.com/psf/black
+    rev: stable
+    hooks:
+    - id: black
+#      language_version: python2.7
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -69,4 +69,5 @@ repos:
     rev: stable
     hooks:
     - id: black
+      args: [--exclude, '/(\.git|\.tox|\.venv|build|static|dist|node_modules)/']
 #      language_version: python2.7
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -69,5 +69,4 @@ repos:
     rev: stable
     hooks:
     - id: black
-      args: [--exclude, '/(\.git|\.tox|\.venv|build|static|dist|node_modules)/']
-#      language_version: python2.7
+      args: [--exclude, '/(\.git|\.tox|\.venv|build|static|dist|node_modules|kolibripip.pex)/']
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,3 +1,4 @@
+exclude: (\.git/|\.tox/|\.venv/|build/|static/|dist/|node_modules/|kolibripip\.pex)
 repos:
 -   repo: git://github.com/pre-commit/pre-commit-hooks
     rev: v2.0.0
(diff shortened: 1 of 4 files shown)


tox.ini
```diff
@@ -164,14 +154,6 @@ commands =
...
 commands = {[node_base]commands}
 
-[testenv:frontendlint]
-whitelist_externals = {[node_base]whitelist_externals}
-commands =
-    yarn
-    # Node-sass gets mardy if we don't do this.
-    npm rebuild node-sass
-    yarn run lint-frontend
-
```
benjaoming, line 174: CC: @indirectlylit @rtibbles this seems adequate to the pre-commit invocation?

benjaoming:
This is waiting for the other PRs to be merged, only then can we get the green lights from Travis :cloud: :traffic_light:

.pre-commit-config.yaml
```diff
@@ -23,3 +24,7 @@ repos:
         language: system
         files: \.(js|vue|scss|css)$
         exclude: '^kolibri/core/static/assets/fonts/.+?\.css$|^kolibri/core/content/static/assets/'
+-   repo: https://github.com/psf/black
+    rev: stable
+    hooks:
+    - id: black
```
indirectlylit, line 72: For Python2.7 development this might break. One option is to specify a Python version:

```
    hooks:
      - id: black
        language_version: python3.6
```

One issue is that this requires Python == 3.6. See previous conversation here:  https://github.com/psf/black/issues/585#issuecomment-478365434 (This is what led to the creation of https://github.com/learningequality/black-fmt)

You made this poll

image

So it looks like we're now in a position to require >= 3.6 after all!

However, requiring a specific version (==3.6) is less desirable than requiring >= 3.6, and ranges can't be specified in pre-commit `language_version`.

Perhaps there's a way to disable black when developing in 2.7, and just rely on tox?

indirectlylit: Ah, looks like we already had this conversation: https://github.com/learningequality/black-fmt/issues/3

indirectlylit: Hm I forgot you made this poll

image

So perhaps we're now in a position to require >= 3.6 after all!

benjaoming: I think it doesn't need be specified. It will install black for the provided python version. pre-commit creates its own virtual environment for the hooks and runs them independently from the external (virtual) environment.

I just tried it from a Python 2.7 environment, and a Python 3 environment was created instead. 

https://pre-commit.com/#overriding-language-version

indirectlylit (approved):
I'm a bit confused how it "knows" to use a py3 venv for black - e.g. where would this be specified?

However I tested with a py2.7 environment and confirmed that it seems to work

benjaoming:
> I'm a bit confused how it "knows" to use a py3 venv for black - e.g. where would this be specified?

I think it's in the hooks configuration:

https://github.com/psf/black/blob/master/.pre-commit-hooks.yaml

PR #6591, opened 2020-02-21
Add serializable isolation db connection for postgres

ralphiee22:
### Summary

SQLite and Postgres have different default [read isolation levels](https://www.geeksforgeeks.org/transaction-isolation-levels-dbms/). SQLite defaults to the "safest" option, which is "serializable", whereas postgres defaults to "read committed". This means that if a transaction reads some data, then in parallel another transaction that writes some data completes, and then the first transaction reads more data, there may be inconsistencies in the two sets of data the first transaction reads -- i.e. it's not a reliable "snapshot" of the database at a particular point in time (c.f. "non-repeatable reads" and "phantom reads").

At the time we built Morango, Kolibri only supported SQlite, which is what we used for our more intensive manual concurrency testing. When we added support for Postgres, we had assumed that it would be at least as robust on this front as SQLite. This type of edge-case mass concurrency testing is very difficult to capture in automated tests, so we don't currently have coverage.

On KDP, this led to some issues where data wasn't being deserialized from the underlying datastore into the Kolibri app models. No data was lost, but it wasn't showing up in the app layer. This was because the deserialization process loops over all the models and then at the very end clears their "dirty" bits. If new records are written into the datastore from an incoming sync in parallel, they may accidentally also have their "dirty bit" cleared, and hence not be deserialized in future batches.

This PR switches the Postgres connection that Morango uses for its large snapshot transactions to have read isolation level of "SERIALIZABLE", which matches SQLite.  

### Reviewer guidance

…

### References

…

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/deployment/default/settings/base.py
+++ b/kolibri/deployment/default/settings/base.py
@@ -161,7 +161,7 @@ elif conf.OPTIONS["Database"]["DATABASE_ENGINE"] == "postgres":
             "HOST": conf.OPTIONS["Database"]["DATABASE_HOST"],
             "PORT": conf.OPTIONS["Database"]["DATABASE_PORT"],
         },
-        "default_serializable": {
+        "default-serializable": {
             "ENGINE": "django.db.backends.postgresql",
             "NAME": conf.OPTIONS["Database"]["DATABASE_NAME"],
             "PASSWORD": conf.OPTIONS["Database"]["DATABASE_PASSWORD"],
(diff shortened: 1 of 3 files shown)


indirectlylit:
> [DO NOT MERGE YET]

FYI there's a new feature to create un-mergeable [draft pull requests](https://github.blog/2019-02-14-introducing-draft-pull-requests/).

Unfortunately, it is not possible to turn an existing PR into a draft

indirectlylit (approved):
image

PR #6611, opened 2020-03-01
OperationalError not caught, retry moved

benjaoming:
### Summary

The former behavior meant that the second retry would not catching an `OperationalError`, crashing Kolibri. The intention was that 6 retries would be done (60 seconds of retries), instead only 1 retry would be done after 10 seconds.

### Reviewer guidance

Add it for a 0.13.x patch release or wait?

### References

None, this was just random code reading.

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/deployment/default/wsgi.py
+++ b/kolibri/deployment/default/wsgi.py
@@ -23,8 +23,20 @@ while not application and tries_remaining:
     try:
         application = get_wsgi_application()
     except OperationalError:
-        # an OperationalError happens when sqlite vacuum is being executed. the db is locked
+        # An OperationalError happens when sqlite vacuum is being
+        # executed. the db is locked
+        print(
+            "Database assumed to be undergoing a VACUUM, retrying again in {} seconds...".format(
+                interval
+            )
+        )
         tries_remaining -= 1
         time.sleep(interval)
+
 if not application:
-    print("Could not start Kolibri")
+    print(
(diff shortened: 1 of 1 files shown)


jredrejo:
hi @benjaoming I see your point if this were a normal code placed in the middle of an application but, in this case I still like the code as it is: it's a last chance to start the uwsgi gateway after three tries. If it fails this last time, the application will not start and having the error traceback in the logs is a needed thing, thus it should not be cached in the `try..except `

benjaoming:
@jredrejo for that to work, the indentation would have to be removed. The "last try" is nested inside the `except` block.

jredrejo:
> @jredrejo for that to work, the indentation would have to be removed. The "last try" is nested inside the `except` block.

that's how it has to be, if you take it out of the `except` block it will be called when no errors happen. This last try should be done after an error has happened without solution.

benjaoming:
:confused:

jredrejo (approved):

benjaoming:
@jredrejo added an extra print statement that seemed nice, so the program doesn't seem unresponsive for 60 seconds.

Overall, this change expands the retry logic from 10 seconds to 60 seconds, in case that plays a role.

Adding a changelog on this because start-up behavior can be affected for larger databases.

indirectlylit (approved):
looks good

Issue #6301, opened 2019-12-13
Max call stack errors when closing a modal whilst the disconnected screen is on


jonboiser:
### Observed behavior

1. Open any modal in Kolibri
1. Stop the server
1. The disconnected overlay will show up
1. Close the modal you opened in first step
1. See errors in console

This might be because the KModal is using some part of the DOM that the disconnected overlay is using and might be causing some bottomless recursion

This also makes forms in the modal unusable when disconnected overlay is showing. See the "Select Network Location > New Address" form

![Screen Shot 2019-12-13 at 10 58 01 AM](https://user-images.githubusercontent.com/10248067/70824796-db049800-1d97-11ea-8ae9-be4eb6bb297e.png)

### Expected behavior

1. No errors

### User-facing consequences

…

### Errors and logs

…

### Context
Kolibri 0.13.beta6

jonboiser:
This seems to happen because both CoreSnackbar and KModal create an event listener for the 'focus' event that seem to cause them to go back and forth between one another's handler.

indirectlylit:
should the the disconnected overlay be over the modal?

jonboiser:
The two backdrops (GlobalSnackbar and KModal) actually have the same z-index of 24 (and I'm guessing the modal itself must be a little higher). So one way to mitigate this might be to put the GlobalSnackbar's backdrop higher than KModal's.

PR #6618, opened 2020-03-02
Fix Snackbar overlay not covering any open modals, plus keyboard a11y fixes

jonboiser:
### Summary

1. Addresses #6301 by simply making the Snackbar overlay div one z-index level higher than the KModal overlay div.
1. Adds a keyboard-focus outline to the snackbar action button and fixes an issue where focus would be trapped on the snackbar div, making the action button inaccessible when the snackbar backdrop was applied.

### Reviewer guidance

1. Test out the scenario in #6301 (e.g. open a modal and turn off the server). Make sure things still work when using keyboard only as well as mouse.
1. To test out the keyboard accessibility in other context, try accessing the snackbar while editing lessons (since the snackbar will have an "undo" action) when you remove items.

### References

…

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/packages/kolibri-components/src/KModal.vue
+++ b/packages/kolibri-components/src/KModal.vue
@@ -274,23 +274,18 @@
         this.$refs.modal.focus();
       },
       focusElementTest(event) {
-        // switching apps - not relevant
-        if (event.target === window) {
-          return;
-        }
-        // not sure when this would be true
-        if (!this.$refs.modal) {
-          return;
-        }
-        // addresses https://github.com/learningequality/kolibri/issues/3824
-        if (
-          event.target === this.$refs.modal ||
-          this.$refs.modal.contains(event.target.activeElement)
-        ) {
+        const { target } = event;
+        const noopOnFocus =
+          target.id === 'coresnackbar' || // keyboard focusing on snackbar
+          target === window || // switching apps
+          !this.$refs.modal || // if $refs.modal isn't available
+          target === this.$refs.modal || // addresses #3824
+          this.$refs.modal.contains(target.activeElement);
+        if (noopOnFocus) {
           return;
         }
         // focus has escaped the modal - put it back!
-        if (!this.$refs.modal.contains(event.target)) {
+        if (!this.$refs.modal.contains(target)) {
           this.focusModal();
         }
       },
@@ -312,7 +307,8 @@
     position: fixed;
     top: 0;
     left: 0;
-    z-index: 24;
+    // z-index needs to be one less than .snackbar-backdrop
+    z-index: 15;
     width: 100%;
     height: 100%;
     background: rgba(0, 0, 0, 0.7);
--- a/packages/kolibri-components/src/KModal.vue
+++ b/packages/kolibri-components/src/KModal.vue
@@ -276,12 +276,17 @@
       focusElementTest(event) {
         const { target } = event;
         const noopOnFocus =
-          target.id === 'coresnackbar' || // keyboard focusing on snackbar
           target === window || // switching apps
           !this.$refs.modal || // if $refs.modal isn't available
           target === this.$refs.modal || // addresses #3824
           this.$refs.modal.contains(target.activeElement);
         if (noopOnFocus) {
+          return;
+        }
+        // Fixes possible infinite recursion when disconnection snackbars appear
+        // along with KModal (#6301)
+        const $coreSnackbar = document.getElementById('coresnackbar');
+        if ($coreSnackbar && $coreSnackbar.contains(event.target)) {
           return;
         }
         // focus has escaped the modal - put it back!
--- a/kolibri/core/assets/src/views/CoreSnackbar/KeenUiSnackbar.vue
+++ b/kolibri/core/assets/src/views/CoreSnackbar/KeenUiSnackbar.vue
@@ -21,6 +21,7 @@
       <button
         v-if="action"
         class="ui-snackbar-action-button"
+        :class="buttonClass"
         :style="{ color: $themeTokens.textInverted }"
         @click.stop="onActionClick"
       >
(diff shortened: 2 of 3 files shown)


bjester:
When I click the `Try Now` button on the offline snackbar, I still get tons of these errors: `RangeError: Maximum call stack size exceeded.`

jonboiser:
@bjester Which browser are you using, and which modal did you open before disconnecting the server?

bjester:
@jonboiser Chrome 80.0.3987.122 and I was using the confirmation modal that appears when deleting downloaded content nodes.

jonboiser:
Hm, just tried the test scenario in the same area of the app and I don't get the errors. Maybe it's a Linux thing. @nucleogenesis Could you see if something similar happens for you?

jonboiser:
If anything, the culprit might be cross-browser/OS differences in how the DOM or Event API is implemented, which is crucial to the `noopOnFocus` variable that prevents the focus/refocusing infinite recursion

bjester:
@jonboiser Yeah it must be related to some browser differences, because I don't experience that error on Firefox.

jonboiser:
Could you put a debugger at this line https://github.com/jonboiser/kolibri/blob/1bfd842b6fcd326c184ad3a91855a22e8e10e673/packages/kolibri-components/src/KModal.vue#L278

and see what element the `event.target` is when you get the error?

bjester:
`event.target` is `window`

jonboiser:
That's weird. Maybe the check `event.target === window` doesn't actually work?

bjester:
That does work, `e.target === window`. I should probably say that I'm using the pex from your PR build here too, in case that is somehow broken.

jonboiser:
Okay, I can finally reproduce this on the PEX 🤷‍♂

packages/kolibri-components/src/KModal.vue
```diff
@@ -312,7 +307,8 @@
     position: fixed;
     top: 0;
     left: 0;
-    z-index: 24;
+    // z-index needs to be one less than .snackbar-backdrop
+    z-index: 15;
```
bjester, line 316: I know @indirectlylit has informed me multiple times that elevations should align with [MD elevations](https://material.io/design/environment/elevation.html#default-elevations). It says snackbars are 6 and dialogs are 24. That doesn't really help but I don't know how to resolve the discrepancy.

jonboiser:
Okay, IDK why it works differently with the pex, but I think what's happening is that the focus event is propagating down to the "Try again" button, which is not caught by my `noopOnFocus` thing. So I'm going to add yet another condition that will tell KModal to stop trying to focus itself if focus is given to the snackbar or any of its children via

```
const noopOnFocus = ... || 
event.target.id === "coresnackbar" ||		
document.getElementById("coresnackbar").contains(event.target)
```

packages/kolibri-components/src/KModal.vue
jonboiser: Hm, I think elevations aren't really z-indices, but actually correspond to actual elevations in display points and usually affected with box-shadow rules (plus obviously having the right order of magnitude of z-index). I guess it's just a coincidence that the z-indices match up with the elevations.

jonboiser:
@bjester @nucleogenesis Besides fixing the guard in KModal I made a change to the snackbar to use a slighty lighter approach to trapping the focus, kind of like what I did for the CoreMenu since it reacting to focus events over all of window was giving me problems. Could you give it another try with the PEX?

bjester:
Editing content, I'm able to access the snackbar and its action button using the keyboard. The issue I experienced previously is also fixed. So the entire snackbar should still receive focus, but should allow tabbing to the action button if there is one?

bjester:
@jonboiser In Chrome, the focus trapping worked fine when editing content, but in Firefox it isn't working correctly. The snackbar doesn't get focus

jonboiser:
> So the entire snackbar should still receive focus, but should allow tabbing to the action button if there is one?

That's how it was initially implemented (with tabindex=0 on the snackbar div), but on the keyboard modality, this ends up showing the focus ring on the whole thing. I think ideally, I would only want to tab into the action button if the snackbar has one (and not tab the snackbar around it), otherwise the whole snackbar has no tabindex.

jonboiser:
Okay, I made another change that will hopefully help with FF. 

1. I added another focus trap underneath the action button, which will be active whether there is a backdrop or not. I personally found that the ability to tab out of the app and into the browser chrome (into the dev tools or who knows where) was sort of confusing anyway. I tested in safari, chrome, ff with the disconnect modal and don't think i can remove the focus from the snackbar with mouse or keyboard.
1. I removed the focus outlining from the snackbar div. So you can tab into it with the keyboard, but it won't receive an outline, tab again to focus the button.

packages/kolibri-components/src/KModal.vue
indirectlylit: It's not a coincidence, no - but it's also not a Material spec

We chose by convention to align Kolibri z-indexes with Material elevations to avoid the situation where people choose arbitrary numbers just to get something above or below something else.

I think it's fine to have exceptions to that convention as long as we label them as-such, but in general let's try to stick to matching MD elevations.

bjester:
I still can't get focus on the snackbar in FF :/

jonboiser:
I think the issue on FF might be related to the DND widget on the lesson's page. I experience the same thing where no matter where I click on it/drag it/etc, the focus is immediately thrown to the top of the page I have to race to tab into the snackbar. (But weirdly this is only on the PEX installer, seems fine running the devserver).

I think this is the only part of the app where this is an issue (and should probably be changed because it's not good for a11y). But overall it seems this PR fixes the original disconnect snackbar + modal issue.

bjester (approved):

PR #6619, opened 2020-03-03
Enhance logging: Middleware work-around for missing cherrypy.access log entries, improve --debug

benjaoming:
### Summary

`cherrypy.access` does not show access to Django views, neither in the command line nor in written logs. This has been verified in an independent test case from Kolibri and [reported upstream](https://github.com/cherrypy/cherrypy/issues/1845).

This does not fix a regression but fixes something that by all accounts seem to have always been broken -- and for some reason we just started living with it?

Fixes https://github.com/learningequality/kolibri/issues/4049

#### Introduces important logging fixes:

* [x] A work-around for missing log calls from normal views. Before, only static requests were logged. Bug reported upstream: https://github.com/cherrypy/cherrypy/issues/1845
* [x] Log the name of the log instead of a module name. So for instance, `cli` becomes `kolibri.utils.cli`.
* [x] Reduce start-up noise from scheduler
* [x] Sets `DEBUG` level from `--debug` flag.
* [x] Also sets `settings.DEBUG=True` in Django settings (this was implemented through a broken behavior before)
* [x] Downgrades a special case of noisy HTTP 200 responses on `/api/tasks/tasks` to DEBUG instead of INFO

#### Benefits

* We can troubleshoot directly from logged outputs much much easier now, as we won't have to guess from the fetched static items and their HTTP referers.
* Log analysis from for instance goaccess.io will be very interesting to look at moving forwards

#### Questions

* ~Should this perhaps log to `cherrypy.access.django` ? It would make it easy to create a filtered access log without all the noise of static files.~ No - let's keep log analysis simple by having everything logged to `cherrypy.access`

#### Example

Log snippet from `kolibri start --foreground`:

```
INFO     Running Kolibri
INFO     Kolibri running on:

	http://127.0.0.1:8080/
	http://192.168.1.138:8080/

INFO     Starting Kolibri 0.13.2.dev0+git.20200303135956
INFO     ENGINE Listening for SIGTERM.
INFO     ENGINE Listening for SIGHUP.
INFO     ENGINE Listening for SIGUSR1.
INFO     ENGINE Listening for SIGINT.
INFO     ENGINE Bus STARTING
INFO     Starting scheduler workers.
INFO     Registering ourselves to zeroconf network with id '327e9b668c5dbdfd96c1f9f48c6290b5'...
INFO     ENGINE Serving on http://0.0.0.0:8080
INFO     ENGINE Bus STARTED
INFO     Ping succeeded! (response: {'id': 650475, 'messages': []})
[...]

PR #6620, opened 2020-03-03
Clean up cherrypy config / instantiation

benjaoming:
### Summary

While debugging cherrypy logging, it became clear to me that we have some confusing elements in the way it's set up:

* [x] We should use a conventional config pattern, namely using the `.update()` method on the config dict
* [x] Do not create a new `Server()` - it's the same as the default one `cherrypy.server`, so it's a waste of CPU cycles.
* [ ] Make the effects of `environment: "production"` explicit
* [ ] Fix 2 cherrypy warnings (that are hidden because of `environment: "production"`)

### Reviewer guidance

Flagging this work, in case there are other efforts or issues wrt. cherrypy

### References

n/a

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/utils/server.py
+++ b/kolibri/utils/server.py
@@ -239,19 +239,21 @@ def configure_http_server(port):
         config={"/": {"tools.caching.on": False}},
     )
 
-    # Instantiate a new server object
-    server = cherrypy._cpserver.Server()
-
     # Configure the server
-    server.socket_host = LISTEN_ADDRESS
-    server.socket_port = port
-    server.thread_pool = conf.OPTIONS["Server"]["CHERRYPY_THREAD_POOL"]
-    server.socket_timeout = conf.OPTIONS["Server"]["CHERRYPY_SOCKET_TIMEOUT"]
-    server.accepted_queue_size = conf.OPTIONS["Server"]["CHERRYPY_QUEUE_SIZE"]
-    server.accepted_queue_timeout = conf.OPTIONS["Server"]["CHERRYPY_QUEUE_TIMEOUT"]
-
+    cherrypy.config.update(
+        {
+            "server.socket_host": LISTEN_ADDRESS,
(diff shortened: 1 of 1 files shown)


jredrejo (commented):
I don't have anything against this, but I don't see any benefit on the change either. Looking at https://github.com/cherrypy/cherrypy/blob/master/cherrypy/__init__.py there's not any difference between the current code and the new code.

I see interesting the `production` environment setup. I see at https://github.com/cherrypy/cherrypy/blob/master/cherrypy/_cpconfig.py#L199-L205 that several configs are affected. This could be either good or bad, depending on the usage we do of them. I am aware of autoreloading, but I don't know if the effect of the others has been considered. It seems to me that some of the information we are trying to get in other PR (#6619) like tracebacks in the logs will be hidden with this option.

indirectlylit:
@benjaoming perhaps submit your changes as individual PRs to look at the changes independently

benjaoming:
>  It seems to me that some of the information we are trying to get in other PR (#6619) like tracebacks in the logs will be hidden with this option.

@jredrejo that's indeed why I have worked with these changes alongside each other. The implicit settings of `environment: production` have been really annoying thus-far, and I would like to replace it with all its inferred settings explicitly set one-by-one.

@indirectlylit - I'll finish the PR, and if there is need to separate the changes, I'll do that :+1:

benjaoming:
Can finish this PR when #6619 is merged.

indirectlylit:
#6619 is merged now. Most of the changes in here are stylistic so those are easy to merge.

The main question is whether we want to remove `"environment": "production"`. I don't understand the pros and cons.

benjaoming:
Had git-lfs issues and deleted the remote branch by mistake, can't reopen even though I have a new remote branch with the same name :shrug:

PR #6629, opened 2020-03-06
Batching and less DB intensive descendant queries for availability annotation

### LINKED ISSUE #6602: content import stuck at 100%
### Observed behavior

possibly related: https://github.com/learningequality/kolibri/issues/6502

We have reports of this behavior from the forums:

* [Thread 1443](https://community.learningequality.org/t/import-via-online-error-100-downloaded-but-still-incomplete-error/1443) and [thread 1400](https://community.learningequality.org/t/downloading-task-is-not-completing-its-stuck-at-100/1400)
* [Thread 1456](https://community.learningequality.org/t/cant-sign-in-on-kolibri/1456)

There was also this similar prior report with a few users chiming in, and we thought it had been addressed: [Thread 1396](https://community.learningequality.org/t/import-export-issues/1396/6)

Kolibri not completing

### Expected behavior

Imports work as expected

### User-facing consequences

Cannot import content

### Errors and logs

Example log file: [kolibri.txt](https://github.com/learningequality/kolibri/files/4257287/kolibri.txt)

Some examples of 500 errors people are seeing:

```
{
“request”: {
“path”: “/api/auth/session/”,
“entity”: “”,
“params”: {
“1582667895147”: 1582667895147
},
“headers”: {
“Content-Type”: “application/json”,
“Accept”: “application/json, application/json;q=0.8, text/plain;q=0.5, /;q=0.2”,
“X-CSRFToken”: “WC66F25F0mmnmTMSQH3axSySIbe7jkzM0xUc3K3eCZYCOo84sHzl7PJSKJMHnjwQ”
},
“method”: “POST”,
“canceled”: false
},
“url”: “/api/auth/session/?1582667895147=1582667895147”,
“raw”: {},
“status”: {
“code”: 500,
“text”: “Internal Server Error”
},
“headers”: {
“Content-Length”: “27”,
“Content-Type”: “text/html”,
“Date”: “Tue, 25 Feb 2020 21:59:59 GMT”,
“Server”: “0.0.0.0”,
“Vary”: “Cookie”,
“X-Frame-Options”: “SAMEORIGIN”
},
“entity”: “Server Error (500)”
}
```

```
{
“request”: {
“path”: “/api/device/deviceinfo/”,
“params”: {
“1582203680411”: 1582203680411
},
“headers”: {
“Content-Type”: “application/json”,
“Accept”: “application/json, application/json;q=0.8, text/plain;q=0.5, /;q=0.2”,
“X-CSRFToken”: “jpqvkxHGzPKqHz8NaKpqVjDz4qAXeKpSeWbhyiJn78pjGOuyq97AC2OvdsZw5jEo”
},
“method”: “GET”,
“canceled”: false
},
“url”: “/api/device/deviceinfo/?1582203680411=1582203680411”,
“raw”: {},
“status”: {
“code”: 500,
“text”: “Internal Server Error”
},
“headers”: {
“Content-Length”: “27”,
“Content-Type”: “text/html”,
“Date”: “Thu, 20 Feb 2020 13:03:08 GMT”,
“Server”: “0.0.0.0”,
“Vary”: “Cookie”,
“X-Frame-Options”: “SAMEORIGIN”
},
“entity”: “Server Error (500)”
}
```

```
{
“request”: {
“path”: “/api/device/deviceinfo/”,
“params”: {
“1582203680411”: 1582203680411
},
“headers”: {
“Content-Type”: “application/json”,
“Accept”: “application/json, application/json;q=0.8, text/plain;q=0.5, /;q=0.2”,
“X-CSRFToken”: “jpqvkxHGzPKqHz8NaKpqVjDz4qAXeKpSeWbhyiJn78pjGOuyq97AC2OvdsZw5jEo”
},
“method”: “GET”,
“canceled”: false
},
“url”: “/api/device/deviceinfo/?1582203680411=1582203680411”,
“raw”: {},
“status”: {
“code”: 500,
“text”: “Internal Server Error”
},
“headers”: {
“Content-Length”: “27”,
“Content-Type”: “text/html”,
“Date”: “Thu, 20 Feb 2020 13:03:08 GMT”,
“Server”: “0.0.0.0”,
“Vary”: “Cookie”,
“X-Frame-Options”: “SAMEORIGIN”
},
“entity”: “Server Error (500)”}
```

### Steps to reproduce

…

### Context

At least one user was using  v0.13.1 on Windows 10

rtibbles:
### Summary
0.13 subtly changed the concept of availability to connote user selection of what should be visible.
This required changing how we annotate the availability of content after an update, to preserve this.
This introduced a need to get all currently available nodes and annotate them and their descendants only.
The initial implementation of this resulted in highly intensive DB queries that caused a seemingly unending update request on the DB.
This PR resolves this by:
* copying a strategy from Django MPTT for doing the descendant queries
* batching updates so that larger channels are handled in smaller chunks, and that we are never applying too many constraints at once
* Treating topic nodes and non-topic nodes differently, as non-topic nodes have no descendants, so we do not need the same constraints.

As a bonus, it adds a little extra to the progress tracking so that the progress does not stall at 100% when annotating availability.

### Reviewer guidance
Note - I was able to replicate this in the shell by effectively making a high number of granular selections, this may help for testing this in the interface.

Does content import work as expected?
Is the code readable and the comments clear?

### References
Hopefully fixes #6602

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/core/content/utils/annotation.py
+++ b/kolibri/core/content/utils/annotation.py
@@ -646,6 +646,17 @@ def recurse_annotation_up_tree(channel_id):
     bridge.end()
 
 
+def calculate_dummy_progress_for_annotation(node_ids, exclude_node_ids, total_progress):
+    num_annotation_constraints = len(node_ids or []) + len(exclude_node_ids or [])
+
+    # Calculate a percentage of the total progress to denote to annotation
+    # between 1 and 10
+    annotation_proportion = min(10, max(1, int(num_annotation_constraints / 500)))
+
+    # Create some progress proportional to annotation task
+    return annotation_proportion * total_progress / (100 - annotation_proportion)
+
+
 def propagate_forced_localfile_removal(localfiles):
     files = File.objects.filter(supplementary=False, local_file__in=localfiles)
     ContentNode.objects.filter(files__in=files).update(available=False)
--- a/kolibri/core/content/utils/annotation.py
+++ b/kolibri/core/content/utils/annotation.py
@@ -3,13 +3,13 @@ import logging
 import os
 from itertools import groupby
 from math import ceil
-from math import floor
 
 from django.db.models import Sum
 from le_utils.constants import content_kinds
 from sqlalchemy import and_
 from sqlalchemy import cast
 from sqlalchemy import exists
+from sqlalchemy import false
 from sqlalchemy import func
 from sqlalchemy import Integer
 from sqlalchemy import or_
@@ -34,82 +34,20 @@ CONTENT_APP_NAME = KolibriContentConfig.label
 CHUNKSIZE = 10000
 
 
-def _MPTT_descendant_ids_statement(
-    connection, ContentNodeTable, node_ids, min_boundary, max_boundary
-):
+def _generate_MPTT_descendants_statement(mptt_values, ContentNodeTable):
     """
-    This function is modified from:
-    https://github.com/django-mptt/django-mptt/blob/master/mptt/managers.py#L66
+    This logic is modified from:
+    https://github.com/django-mptt/django-mptt/blob/38d46c26ca362c471b097ab96a3616b9b20fb883/mptt/managers.py#L137
     in order to render the result as a SQL Alchemy expression that we can use
     in other queries.
     """
-
-    # First get the relevant MPTT values from the database for the specified node_ids
-    # for topic nodes in the specified lft/rght range.
-    mptt_values = connection.execute(
-        select(
-            [
-                ContentNodeTable.c.tree_id,
-                ContentNodeTable.c.parent_id,
-                ContentNodeTable.c.lft,
-                ContentNodeTable.c.rght,
-            ]
-        )
-        .order_by(
-            ContentNodeTable.c.tree_id,
-            ContentNodeTable.c.parent_id,
-            ContentNodeTable.c.lft,
-        )
-        .where(
-            and_(
-                filter_by_uuids(ContentNodeTable.c.id, node_ids),
-                # Also filter by the boundary conditions
-                # We are only interested in nodes that are ancestors of
-                # the nodes in the range, but they could be ancestors of any node
-                # in this range, so we filter the lft value by being less than
-                # or equal to the max_boundary, and the rght value by being
-                # greater than or equal to the min_boundary.
-                ContentNodeTable.c.lft <= max_boundary,
-                ContentNodeTable.c.rght >= min_boundary,
-                # Only select values for descendant constraints from topics
-                ContentNodeTable.c.kind == content_kinds.TOPIC,
-            )
-        )
-    ).fetchall()
-
-    # Now we fetch a list of non-topic ids from the specified node ids
-    # that match the specified tree boundary ranges
-    non_topic_node_ids = map(
-        lambda x: x[0],
-        connection.execute(
-            select([ContentNodeTable.c.id]).where(
-                and_(
-                    filter_by_uuids(ContentNodeTable.c.id, node_ids),
-                    # Also filter by the boundary conditions
-                    # We are only interested in non-topic nodes that
-                    # are inside the range
-                    ContentNodeTable.c.rght >= min_boundary,
-                    ContentNodeTable.c.rght <= max_boundary,
-                    # Produce an id list for non topics
-                    ContentNodeTable.c.kind != content_kinds.TOPIC,
-                )
-            )
-        ).fetchall(),
-    )
-
-    or_queries = []
-
-    # If we have any node ids that are for not topics, then we add an explicit query
-    # to match against those node ids
-    if non_topic_node_ids:
-        or_queries.append(filter_by_uuids(ContentNodeTable.c.id, non_topic_node_ids))
+    queries = []
 
     # Group the resultant mptt data by tree_id and parent_id,
     # this will allow us to consolidate contiguous siblings to reduce
     # the total number of constraints.
     # This logic is verbatim from Django MPTT, only the query construction
     # has been translated from Django Q statements to SQL Alchemy and_ statements.
-
     for group in groupby(
         mptt_values,
         key=lambda n: (
@@ -135,7 +73,7 @@ def _MPTT_descendant_ids_statement(
                     min_max["max"] = max_val
                 next_lft = rght + 1
             elif lft != next_lft:
-                or_queries.append(
+                queries.append(
                     and_(
                         ContentNodeTable.c.tree_id == tree,
                         ContentNodeTable.c.lft >= min_max["min"],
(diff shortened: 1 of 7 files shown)


kolibri/core/content/utils/annotation.py
```diff
@@ -30,109 +34,368 @@
...
+                    # Produce an id list for non topics
+                    ContentNodeTable.c.kind != content_kinds.TOPIC,
+                )
+            )
+        ).fetchall(),
     )
 
+    or_queries = []
+
+    # If we have any node ids that are for not topics, then we add an explicit query
```
indirectlylit: > for not topics

typo?

jonboiser:
Import with 46 includes/11 excludes went through. Not sure if that pushes the limit, but it did have to batch the annotation into 9 pieces:

```
INFO     Setting availability to True of 2523 LocalFile objects based on passed in checksums
INFO     Setting availability of non-topic ContentNode objects based on LocalFile availability in 9 batches
INFO     Annotating ContentNode objects with children for 5 levels
INFO     Annotating ContentNode objects with children for level 5
INFO     Annotating ContentNode objects with children for level 4
INFO     Annotating ContentNode objects with children for level 3
INFO     Annotating ContentNode objects with children for level 2
INFO     Annotating ContentNode objects with children for level 1
```

Not sure if the Task Metadata code was affected, but this is the first time i've seen a task complete but not hit 100% of the resources:

```
193 of 194 resources (202 MB of 202 MB) imported

```

kolibri/core/content/utils/annotation.py
```diff
@@ -30,109 +34,368 @@
...
+                )
+                min_max = {"min": min_val, "max": max_val}
+                next_lft = rght + 1
+        or_queries.append(
+            and_(
+                ContentNodeTable.c.tree_id == tree,
+                ContentNodeTable.c.lft >= min_max["min"],
+                ContentNodeTable.c.rght <= min_max["max"],
+            )
+        )
```
indirectlylit: Could move this code into its own `_named_function` to keep things a bit tighter

kolibri/core/content/utils/annotation.py
```diff
@@ -30,109 +34,368 @@
...
+
+    if not or_queries:
+        # No constraints that apply in this range, so therefore this query should always
+        # evaluate to False, because nothing can match it.
+        return select([ContentNodeTable.c.id]).where(False == True)  # noqa E712
```
indirectlylit: Weird :)

could this be

```python
        return select([ContentNodeTable.c.id]).where(False)
```

or maybe

```python
        return select([ContentNodeTable.c.id]).filter(sqlalchemy.sql.false())
```

kolibri/core/content/utils/annotation.py
```diff
@@ -30,109 +34,368 @@
...
+    # To chunk the tree, we first find the full extent of the tree - this gives the
+    # highest rght value for this channel.
+    max_rght = connection.execute(
+        select([func.max(ContentNodeTable.c.rght)]).where(
+            ContentNodeTable.c.channel_id == channel_id,
+        )
+    ).scalar()
+
+    # Start a counter for the while loop
+    i = 0
```
indirectlylit: might help to initialize `i` to the min value?

kolibri/core/content/utils/annotation.py
```diff
@@ -30,109 +34,368 @@
...
-            filter_by_uuids(node_include.c.id, node_ids),
-            ContentNodeTable.c.lft >= node_include.c.lft,
-            ContentNodeTable.c.lft <= node_include.c.rght,
-            ContentNodeTable.c.tree_id == node_include.c.tree_id,
+def _MPTT_descendant_ids_statement(
+    connection, ContentNodeTable, node_ids, min_boundary, max_boundary
+):
+    """
+    This function is modified from:
+    https://github.com/django-mptt/django-mptt/blob/master/mptt/managers.py#L66
```
jonboiser: maybe put the permalink to this line here instead
```suggestion
    https://github.com/django-mptt/django-mptt/blob/38d46c26ca362c471b097ab96a3616b9b20fb883/mptt/managers.py#L66
```

kolibri/core/content/utils/annotation.py
```diff
@@ -30,109 +34,368 @@
...
+        select([func.max(ContentNodeTable.c.rght)]).where(
+            ContentNodeTable.c.channel_id == channel_id,
+        )
+    ).scalar()
+
+    # Start a counter for the while loop
+    i = 0
+
+    # Count the total number of constraints
+    constraint_count = len(node_ids or []) + len(exclude_node_ids or [])
```
indirectlylit, line 230: Do we need to support these being `None`? The method would be simplified if the signature was just

```python
def set_leaf_nodes_invisible(channel_id, node_ids=[], exclude_node_ids=[]):
    # ...
```

but maybe there's a need for it?

kolibri/core/content/utils/annotation.py
```diff
@@ -30,109 +34,368 @@
...
+
+    # Aim for a constraint per batch count of about 250 on average
+    # This means that there will be at most 750 parameters from the constraints
+    # and should therefore also limit the overall SQL expression size.
+
+    dynamic_chunksize = max(
+        1, min(CHUNKSIZE, floor(250 * max_rght / (constraint_count or 1)))
+    )
```
indirectlylit: Can `max_rght` be less than or equal to `0`?

If not, I think this is the same as

```python
min(CHUNKSIZE, ceil(250 * max_rght / (constraint_count or 1)))
```

kolibri/core/content/utils/annotation.py
```diff
@@ -30,109 +34,368 @@
...
 
-    if node_ids:
-        node_ids_statement = _MPTT_descendant_ids_statement(ContentNodeTable, node_ids)
-        update_statement = update_statement.where(
-            ContentNodeTable.c.id.in_(node_ids_statement)
+    logger.info(
+        "Removing availability of non-topic ContentNode objects in {} batches".format(
+            int(ceil(max_rght / dynamic_chunksize))
```
indirectlylit:
```suggestion
        "Removing availability of non-topic ContentNode objects in {} batches of {}".format(
            int(ceil(max_rght / dynamic_chunksize)), dynamic_chunksize
```

kolibri/core/content/utils/annotation.py
```diff
@@ -30,109 +34,368 @@
...
         )
+    )
+
+    while i * dynamic_chunksize < max_rght:
```
indirectlylit: could pull the contents of this loop out into its own `_named_function`

indirectlylit (approved):
looks good. I have not tested yet

After merging in I'd also like to cut a beta and get testing from both end users and @mrpau-dev 

Left some non-blocking code style comments. Some of those in `set_leaf_nodes_invisible` would apply to `set_leaf_node_availability_from_local_file_availability` also

kolibri/core/content/utils/annotation.py
rtibbles: Possibly - I was looking around and couldn't find a definitive answer in docs.

kolibri/core/content/utils/annotation.py
rtibbles: It's an integer counter, that I am then multiplying by the batch size.

kolibri/core/content/utils/annotation.py
rtibbles: It's `None` because: https://docs.quantifiedcode.com/python-anti-patterns/correctness/mutable_default_value_as_argument.html - but I think you have perhaps just temporarily forgotten this.

kolibri/core/content/utils/annotation.py
rtibbles: It shouldn't be, no, I was just trying to be systematic and perhaps overly cautious.

rtibbles:
There is a postgres test failure, so this might take a little while longer to fix.

kolibri/core/content/utils/annotation.py
jamalex: He knew it originally, but then the memory was modified as a side effect of another call to the same function! /s

kolibri/core/content/utils/annotation.py
```diff
@@ -30,109 +34,368 @@
...
+                ContentNodeTable.c.parent_id,
+                ContentNodeTable.c.lft,
+                ContentNodeTable.c.rght,
+            ]
         )
+        .order_by(
+            ContentNodeTable.c.tree_id,
+            ContentNodeTable.c.parent_id,
+            ContentNodeTable.c.lft,
+        )
```
jamalex: The ordering may not be necessary for our purposes here, and so it may be faster to omit.

kolibri/core/content/utils/annotation.py
```diff
@@ -30,109 +34,368 @@
...
+            and_(
+                filter_by_uuids(ContentNodeTable.c.id, node_ids),
+                # Also filter by the boundary conditions
+                # We are only interested in nodes that are ancestors of
+                # the nodes in the range, but they could be ancestors of any node
+                # in this range, so we filter the lft value by being less than
+                # or equal to the max_boundary, and the rght value by being
+                # greater than or equal to the min_boundary.
+                ContentNodeTable.c.lft <= max_boundary,
+                ContentNodeTable.c.rght >= min_boundary,
```
jamalex: I haven't read the rest of the code yet to fully understand how this gets used, but my initial thought: This will include descendants of nodes in between two nodes in node_ids, which aren't in fact descendants of any of the nodes in node_ids. In the referenced [MPTT source](https://github.com/django-mptt/django-mptt/blob/master/mptt/managers.py#L167), it build a composite OR query, rather than joining into one big min/max range.

kolibri/core/content/utils/annotation.py
jamalex: "non topics"
"non-topic nodes"
"hot topics"?

kolibri/core/content/utils/annotation.py
jamalex: Ah, I guess perhaps it's needed for the groupby below.

kolibri/core/content/utils/annotation.py
jamalex: Ah, I'd been interpreting these differently (and you replicate the actual MPTT stuff below). I see that these are instead for the batching purposes. Clarifying that in the comments above may be useful, since you reference this being a port of the MPTT code, but that aspect isn't done by MPTT, so the parallel comparison becomes harder.

kolibri/core/content/utils/annotation.py
jamalex: https://stackoverflow.com/a/12837029 says you can do something like
`session.query(MyObject).filter(sqlalchemy.sql.false())`

kolibri/core/content/utils/annotation.py
```diff
@@ -30,109 +34,368 @@
...
 
+    # This statement defines the update condition for the contentnode
+    # running exists on this (as it is used below) will produce either
+    # True, in the case when the contentnode has the required files
+    # available for rendering, or False otherwise.
     contentnode_statement = (
-        select([FileTable.c.contentnode_id])
-        .select_from(
+        # We could select any property here, as it's the exist that matters.
+        select([FileTable.c.contentnode_id]).select_from(
```
jamalex: Note that you can also just do things like `SELECT 1 FROM kolibriauth_facilityuser;` (maybe faster...?)

jamalex: Example here: https://stackoverflow.com/questions/7646173/sqlalchemy-exists-for-query#comment65330050_13336408

jamalex (approved):
Whew! Lots going on here! Not sure I wrapped my head around all the details, but will defer to the tests for correctness. Don't think I saw any blockers, mostly just some comments reflecting my confusion that could feed into commenting. Thanks!

rtibbles:
Thanks for the reviews, given the postgres test failures, I will address and integrate the comments prior to merge as a little more work is needed.

kolibri/core/content/utils/annotation.py
indirectlylit: > The first time that the function is called [without an argument], Python creates a persistent list. 

Wow, I had no idea that was the case!

I assume then that another thing I've done in the past has not been doing what I thought:

```python
def func(param=calc_default_value()):
```

kolibri/core/content/utils/annotation.py
rtibbles: Have tried to separate out the parts ported from MPTT and parts that aren't, so hopefully that will help.

PR #6634, opened 2020-03-07
Slim down the PF pipeline

DXCanas:
Only want a debian installer for pf build.

Still needs to bring in PF plugin.

### Summary

…

### Reviewer guidance

…

### References

…

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -7,40 +7,9 @@ steps:
 
   - wait
 
-  - trigger: "kolibri-macos"
-    label: ":mac:"
-    async: true
-    build:
-      message: "${BUILDKITE_MESSAGE}"
-      env:
-        LE_TRIGGERED_FROM_BUILD_ID: "${BUILDKITE_BUILD_ID}"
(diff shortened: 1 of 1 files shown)


indirectlylit:
> Still needs to bring in PF plugin.

For the short-term, we can probably just embed the plugin within this code base

indirectlylit:
I'm fine to merge this, but just trying to understand why it's necessary? We only need debian it's true, but what are the negative consequences of leaving these in?

(My only concern is that making these changes will make it a bit harder to merge this branch into upstream develop eventually)

DXCanas:
I figured we shouldn't build (and eventually maintain) all the other installers if we don't need them, and I wasn't clear on the intention to merge these changes back to develop. And if we're baking the plugin into the codebase, an entirely new pipeline sounds like it just isn't the right solution.  

I can think of a few other ways to do this that would be lighter touch. Leaning in that direction now.

PR #6637, opened 2020-03-08
Add option to customize sandbox.

rtibbles:
### Summary
Adds a Kolibri option to customize the sandbox attribute in the HTML5 viewer plugin.
Adds sanitization of the option to only pass through allowed options.
Sets the sandbox attribute of the iframe in the HTML5 viewer plugin to the option.

### Reviewer guidance
Set e.g. this in `options.ini`:
```
[HTML5]
SANDBOX = allow-scripts allow-same-origin allow-bourne allow-forms
```
This should result in the sandbox attribute of the iframe in the HTML5 viewer being set to:
`allow-scripts allow-same-origin allow-forms`

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/plugins/html5_viewer/kolibri_plugin.py
+++ b/kolibri/plugins/html5_viewer/kolibri_plugin.py
@@ -21,4 +21,4 @@ class HTML5AppAsset(content_hooks.ContentRendererHook):
 
     @property
     def plugin_data(self):
-        return {"sandbox": OPTIONS["HTML5"]["SANDBOX"]}
+        return {"html5_sandbox_tokens": OPTIONS["HTML5"]["SANDBOX"]}
--- a/kolibri/plugins/html5_viewer/kolibri_plugin.py
+++ b/kolibri/plugins/html5_viewer/kolibri_plugin.py
@@ -7,13 +7,18 @@ from le_utils.constants import format_presets
 from kolibri.core.content import hooks as content_hooks
 from kolibri.plugins import KolibriPluginBase
 from kolibri.plugins.hooks import register_hook
+from kolibri.utils.conf import OPTIONS
 
 
 class HTML5AppPlugin(KolibriPluginBase):
-    pass
+    kolibri_options = "options"
 
 
 @register_hook
 class HTML5AppAsset(content_hooks.ContentRendererHook):
     bundle_id = "main"
     presets = (format_presets.HTML5_ZIP,)
+
+    @property
+    def plugin_data(self):
+        return {"sandbox": OPTIONS["HTML5"]["SANDBOX"]}
--- a/kolibri/plugins/html5_viewer/options.py
+++ b/kolibri/plugins/html5_viewer/options.py
@@ -1,0 +1,42 @@
+# Source: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/iframe#attr-sandbox
+allowable_sandbox_tokens = set(
+    [
+        "allow-downloads-without-user-activation",
+        "allow-forms",
+        "allow-modals",
+        "allow-orientation-lock",
+        "allow-pointer-lock",
+        "allow-popups",
+        "allow-popups-to-escape-sandbox",
+        "allow-presentation",
+        "allow-same-origin",
+        "allow-scripts",
+        "allow-storage-access-by-user-activation ",
+        "allow-top-navigation",
+        "allow-top-navigation-by-user-activation",
+    ]
+)
+
+
+def clean_sandbox(sandbox_string):
+    """
+    Clean up sandbox string to ensure it only contains valid items.
+    """
+    sandbox_tokens = [
+        token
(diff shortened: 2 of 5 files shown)


kollivier:
Awesome, thanks! I'll start testing the content with this PR tomorrow!

rtibbles:
Oh, looks like this is causing a weird recursion issue - will look into that!

rtibbles:
This commit seemed to be the minimal fix to prevent the recursion issue, and it doesn't change the behaviour, just adds the filtering earlier.

kolibri/plugins/html5_viewer/options.py
```diff
@@ -0,0 +1,42 @@
...
+def clean_sandbox(sandbox_string):
+    """
+    Clean up sandbox string to ensure it only contains valid items.
+    """
+    sandbox_tokens = [
+        token
+        for token in sandbox_string.split(" ")
+        if token in allowable_sandbox_tokens
+    ]
+    return " ".join(sandbox_tokens)
```
jamalex, line 42: Maybe show a warning if something unrecognized is excluded, to help with catching typos?

kolibri/plugins/html5_viewer/kolibri_plugin.py
```diff
@@ -7,13 +7,18 @@
...
 
 
 @register_hook
 class HTML5AppAsset(content_hooks.ContentRendererHook):
     bundle_id = "main"
     presets = (format_presets.HTML5_ZIP,)
+
+    @property
+    def plugin_data(self):
+        return {"sandbox": OPTIONS["HTML5"]["SANDBOX"]}
```
jamalex: Should we perhaps also add this to the H5P renderer?

jamalex: And if so, what's the cleanest way to share this stuff across plugins without copypasta?

kolibri/plugins/html5_viewer/options.py
```diff
@@ -0,0 +1,42 @@
...
+option_spec = {
+    "HTML5": {
+        "SANDBOX": {
+            "type": "string",
+            "default": "allow-scripts",
+            "envvars": ("KOLIBRI_HTML5_SANDBOX",),
+            "clean": clean_sandbox,
+        }
+    }
+}
```
jamalex, line 54: Rather than a default, should we maybe always add "allow-scripts" back in on top of whatever they specify? Otherwise Hashi will break anyway, right?

jamalex (commented):
Super cool! Nice to see here how options.ini can tie into plugin data and be exposed on the frontend.

Added a couple of comments. Will leave approval to @kollivier's manual testing.

kolibri/plugins/html5_viewer/kolibri_plugin.py
indirectlylit: Perhaps we should give this a more unique/searchable name, e.g. 

```suggestion
        return {"html5_sandbox_type": OPTIONS["HTML5"]["SANDBOX"]}
```

I'll update the PR and merge

rtibbles: I would rather `html5_sandbox_tokens` - also, I think @kollivier is adding another commit to this PR, so would rather not mess with that.

rtibbles: I think no, for now - and we also need to work out how we share other code between the two, so I'd rather work that out before doing any duplication.

rtibbles:
Also, I think it might be wise to merge https://github.com/learningequality/kolibri/pull/6637/commits/a5f8ca3fb82201623c5de8f3ee5128ee9fedeade to 0.13.x, as it is a fairly major bug with the plugin API - I'll cherry pick it to a PR there.

kolibri/plugins/html5_viewer/options.py
rtibbles: Addressed!

kolibri/plugins/html5_viewer/kolibri_plugin.py
rtibbles: Updated.

kolibri/plugins/html5_viewer/options.py
kollivier: My vote would be to leave the options fully customizable, just because it gives the user full control. While I can't think of a reason a user might want to block scripts from running, I think if someone did want that behavior they should be able to do so. (And blocking Hashi in that case would be expected behavior - I suppose we could avoid injecting Hashi in that case even.)

kollivier (approved):
Tested with and without options set, and it works great! :)

kollivier:
Going to merge in so that we can work on getting a demo server with fixes in before Thursday.

PR #6638, opened 2020-03-09
Proof of concept pure Python websocket implementation for coach notifications

rtibbles:
### Summary
Vendors https://github.com/Pithikos/python-websocket-server in order to have a pure Python websocket server started alongside the regular Kolibri server.

Adds a very simple notifications broadcast when learner progress notifications are created.

Listens for these notifications via websocket in the coach app.

### Reviewer guidance
What do you think?

### References
I chose to vendor https://github.com/Pithikos/python-websocket-server as it is a single file of code, the pypi package is out of date, and it appears to be sporadically maintained. There are several open PRs that we could probably look to for inspiration for improvements.

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/core/notifications/api.py
+++ b/kolibri/core/notifications/api.py
@@ -4,6 +4,7 @@ from django.db.models import Count
 from django.db.models import Sum
 from django.db.models import When
 from le_utils.constants import content_kinds
+from rest_framework.renderers import JSONRenderer
 
 from .models import HelpReason
 from .models import LearnerProgressNotification
@@ -15,6 +16,8 @@ from kolibri.core.exams.models import ExamAssignment
 from kolibri.core.lessons.models import Lesson
 from kolibri.core.logger.models import AttemptLog
 from kolibri.core.logger.models import ContentSummaryLog
+from kolibri.plugins.coach.serializers import LearnerNotificationSerializer
+from kolibri.utils.websocket import SERVER
 
 
 @memoize
@@ -87,6 +90,8 @@ def save_notifications(notifications):
     with transaction.atomic():
         for notification in notifications:
             notification.save()
+    data = LearnerNotificationSerializer(notifications, many=True).data
+    SERVER.send_message_to_all(JSONRenderer().render(data))
 
 
 def create_notification(
--- a/kolibri/plugins/coach/assets/src/modules/coachNotifications/index.js
+++ b/kolibri/plugins/coach/assets/src/modules/coachNotifications/index.js
@@ -43,12 +43,12 @@ export default {
         .then(data => {
           store.commit('SET_NOTIFICATIONS', data.results);
           if (!store.state.poller) {
-            store.dispatch('startingPolling', { coachesPolling: data.coaches_polling });
+            store.dispatch('startWebSocket', { coachesPolling: data.coaches_polling });
           }
         })
         .catch(() => {
           if (!store.state.poller) {
-            store.dispatch('startingPolling', { coachesPolling: 0 });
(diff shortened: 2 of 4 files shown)


indirectlylit:
Nice, I love how small the change is! Not sure why we never explored this direction previously.

One thing I've experienced when using websockets in the past is that there's sometimes complexity in socket management and "garbage collection" - the server needs to watch the pool of open connections and check for various error and client states to make sure that doesn't keep growing.

The other complexity if we wanted to start building on this would eventually be around tying in to Django's existing semantics: authentication, subscriptions to particular model changes, etc. Those are the kinds of things that [Django Channels](https://channels.readthedocs.io/en/latest/introduction.html) implement.

However I do like the idea of starting with something much simpler and purpose-driven rather than trying to async all the things.

jredrejo:
I love the idea too, and, as @indirectlylit  says, the code changes are so small that looks perfect, 
but I am concerned by some things:
1. I'd like to have a test to see how performant this is, it uses Python SocketServer library behind the scenes, and Python sockets are famous for being slow. I'd bet they are fast enough for this use case but better to test it. At least adding some test as the `test_text_message_stress_bursts` at  https://github.com/Pithikos/python-websocket-server/blob/master/tests/test_text_messages.py would be interesting.
2. It uses threads, I'd like to test it under uwsgi to see if they work properly
3. Also we should see how this code work on different OS, as SocketServer is very  low level that's another thing to test.

So, I would not merge this until these three kind of tests are done, it's a feature too important to mess it now. 
Also, thinking of the future, I'd prefer to drop Python 2 support and move to a more robust and tested websockets solution as Django Channels (if tests are fine we can use this in the meantime, I just wonder if the effort is worth)

rtibbles:
Yeah, all the concerns are definitely warranted here. This implementation would not work if we were using uwsgi, as it is relying on there being a process-local reference to websocket server available.

In the uwsgi case this would fail, as the server would not be started in that process, so we would definitely need to explore an alternate mechanism for making this happen.

It may be that this PR should never be merged, that's why it's a draft - and I agree with all the precautionary measures before we even think about putting it into mainline.

jredrejo:
> Yeah, all the concerns are definitely warranted here. This implementation would not work if we were using uwsgi, as it is relying on there being a process-local reference to websocket server available.
> 
> In the uwsgi case this would fail, as the server would not be started in that process, so we would definitely need to explore an alternate mechanism for making this happen.

That should be solvable using 
```
    location /ws/ {
        proxy_pass http://localhost:9080; 
        proxy_http_version 1.1;
        proxy_set_header Upgrade $http_upgrade;
        proxy_set_header Connection "upgrade";
    }
```

in the nginx configuration, but we'd need to replace
` const socket = new WebSocket('ws://127.0.0.1:9080'); `
by 
`const socket = new WebSocket('ws://server_ip/ws/'); `
in the js part. If this can be configured, then, uwsgi should not be a problem.

rtibbles:
Oh - I wasn't concerned about that, but yes, we would need that!

It was more that inside the background log saving process it is calling `SERVER.send_message_to_all` which in the uwsgi case would not be a reference to the actual server which is running in the separate `services` process - so we might need to send these messages via a Socket instead...

indirectlylit:
closing to clean up PRs.

If you think this is something we might want to build off of later, feel free to add this to a new LE branch `learningequality:websockets_proto`

rtibbles:
OK, have rebased it and pushed to the LE branch.

PR #6645, opened 2020-03-10
improved webpack error reporting

indirectlylit:
### Summary

Current webpack errors were swallowing some useful info and being treated as "unhandled" exceptions, when they could be handled.

| Before | After |
|--|--|
| image | image |

### Reviewer guidance

Does this look like the right way to implement things?

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/core/webpack/hooks.py
+++ b/kolibri/core/webpack/hooks.py
@@ -105,11 +105,13 @@ class WebpackBundleHook(hooks.KolibriHook):
 
         if not stats_file_content or getattr(settings, "DEVELOPER_MODE", False):
 
-            if not os.path.exists(self._stats_file):
-                raise WebpackError("Missing stats file: '{}'".format(self._stats_file))
+            STATS_ERR = "Error accessing stats file '{}': {}"
 
-            with io.open(self._stats_file, mode="r", encoding="utf-8") as f:
-                stats = json.load(f)
+            try:
+                with io.open(self._stats_file, mode="r", encoding="utf-8") as f:
+                    stats = json.load(f)
+            except IOError as e:
+                raise WebpackError(STATS_ERR.format(self._stats_file, e))
 
             if getattr(settings, "DEVELOPER_MODE", False):
                 timeout = 0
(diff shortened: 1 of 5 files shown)


kolibri/core/webpack/middleware.py
```diff
@@ -0,0 +1,28 @@
...
+
+from .hooks import WebpackError
+
+
+logger = logging.getLogger(__name__)
+
+
+class WebpackErrorHandler:
+    def __init__(self, get_response):
+        self.get_response = get_response
```
rtibbles, line 13: Could we get this to only run if DEVELOPER_MODE is set to `True`?

Can do this by simply raising a `MiddlewareNotUsed` exception during the `__init__` if not: https://docs.djangoproject.com/en/1.11/topics/http/middleware/#marking-middleware-as-unused

Alternatively, we can only add this in the dev settings, not in the production settings.

kolibri/core/webpack/hooks.py
```diff
@@ -103,40 +101,43 @@ def _stats_file_content(self):
...
-                )
-            )
+        stats_file_content = caches[CACHE_NAMESPACE].get(cache_key)
+
+        if not stats_file_content or getattr(settings, "DEVELOPER_MODE", False):
+
+            if not os.path.exists(self._stats_file):
+                raise WebpackError("Missing stats file: '{}'".format(self._stats_file))
+
+            with io.open(self._stats_file, mode="r", encoding="utf-8") as f:
```
rtibbles: This seems like it will generate an unhandled `IOError` if the stats file doesn't exist? Can we reinstate the try...except just around this block?

rtibbles: Nvm, just saw the above.

rtibbles: Although trying and catching is more Pythonic :)

rtibbles (changes requested):
I think moving the middleware only to dev settings is preferable, if this is needed in production, something has gone incredibly wrong.

indirectlylit:
> I think moving the middleware only to dev settings is preferable, if this is needed in production, something has gone incredibly wrong.

Okay - yeah agreed. The middleware was basically a no-op so I didn't bother but can do

kolibri/core/webpack/hooks.py
indirectlylit: yeah, will change to wrap individual lines

The main thing I was doing here was removing the try...catch around the _entire_ block, because it seemed to be re-catching `WebpackError` thrown inside which was causing super confusing behaviors

rtibbles: Yup, that was definitely insufficiently granular.

indirectlylit:
updated based on feedback - thanks

rtibbles (approved):

PR #6649, opened 2020-03-11
working on pf plugin

indirectlylit:
### Summary

Initial work on creating a pf plugin which will handle:

* setting theme custom Profuturo theme
* setting Profuturo-specific sandbox settings for the SCORM/HTML5 apps
* setting options to enable experimental user import functionality

The premise is that this work should enable us to create custom builds and live demos from the `pf-mvp` branch over the coming weeks. Once we're further along, we can move the plugin to its own repo and merge `pf-mvp` into the mainline

### Reviewer guidance

1.Is there a cleaner way of handling this in our temporary feature branch? Currently I have temporarily modified `INTERNAL_PLUGINS` and `default_plugins.txt`
2. I don't quite understand how plugins are intended to override option values, specifically so we can set the `sandbox`. It seems like we don't distinguish between _defining_ options and _setting_ options. When I try to simply use `options_spec`, it complains that the `HTML5.SANDBOX` option was already being set by another plugin. My sense is we might need to modify the plugin options handling to support this?

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated


kolibri/plugins/hooks.py
```diff
@@ -291,7 +291,7 @@ def add_hook_to_class_registry(cls, hook):
             and cls._registered_hooks
             and hook not in cls.registered_hooks
         ):
-            raise RuntimeError(
+            logger.error(
```
indirectlylit: FYI @rtibbles I switched this because we were running into this issue: 

https://learningequality.slack.com/archives/CR0LK9W80/p1583968360115800?thread_ts=1583967561.115300&cid=CR0LK9W80

However I think now it's triggering this:

```
=================================== FAILURES ===================================
___________________ test_register_multiple_hook_single_hook ____________________
valid_single_hook = <class 'test.kolibri_plugin.SingleHook'>
    def test_register_multiple_hook_single_hook(valid_single_hook):
        valid_single_hook = hooks.register_hook(valid_single_hook)
        valid_single_hook.add_hook_to_registries()
    
        class OtherSingleHook(SingleHookAbstract):
            pass
    
        OtherSingleHook.__module__ = "test.kolibri_plugin"
        try:
            OtherSingleHook = hooks.register_hook(OtherSingleHook)
            OtherSingleHook.add_hook_to_registries()
            pytest.fail(
>               "Allowed a hook single instance hook to be registered more than once"
            )
E           Failed: Allowed a hook single instance hook to be registered more than once
kolibri/plugins/utils/test/test_hooks.py:80: Failed
----------------------------- Captured stderr call -----------------------------
ERROR    Attempted to register more than one instance of <class 'test.kolibri_plugin.OtherSingleHook'>
```

curious about your thoughts

rtibbles: The issue was that we were not catching all errors that could occur during plugin initialization when trying to disable a plugin, which left the user in a catch 22. I've now shifted Kolibri to do a more 'self healing' kind of approach, and just aggressively disabling plugins that don't initialize, while writing loud error messages about this. Let me know if this is too far in the other direction.

rtibbles:
Have:
* updated the behaviour when trying to initialize plugins to aggressively disable ones that error out
* separated out option default overriding from specifying extensions to the options spec - this means there should be no ambiguity over the ordering
* set the more lax default for the sandbox attribute using the new options default overriding

indirectlylit:
thanks for these changes @rtibbles  - looks good

PR #6652, opened 2020-03-11
Html5parse tests

rtibbles:
### Summary
* Adds tests for maintaing doctype in HTML parsing in zipcontent
* Adds tests for extra slash robustness in file path parsing in zipcontent
* Tweaks behaviour for doctype to pass tests
* Tweaks behaviour for extra slash robustness to pass tests

### Reviewer guidance
Does everything still work as expected for the target HTML5 apps?

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/core/content/views.py
+++ b/kolibri/core/content/views.py
@@ -290,7 +290,11 @@ class ZipContentView(View):
         zipped_path = get_path_or_404(zipped_filename)
 
         # Sometimes due to URL concatenation, we get URLs with double-slashes in them, like //path/to/file.html.
-        # Normalize the path by converting those double-slashes to a single slash.
+        # If it is a leading double slash (i.e. /zipcontent/filename.zip//file.html) then it will get passed
+        # to zipcontent as a single leading slash - detect this and remove if present.
+        if embedded_filepath.startswith("/"):
+            embedded_filepath = embedded_filepath[1:]
+        # Normalize the path by converting double-slashes occurring later in the path to a single slash.
         embedded_filepath = embedded_filepath.replace("//", "/")
 
         # if client has a cached version, use that (we can safely assume nothing has changed, due to MD5)
--- a/kolibri/core/content/views.py
+++ b/kolibri/core/content/views.py
@@ -8,7 +8,6 @@ import json
 import logging
 import mimetypes
 import os
-import re  # please forgive me
 import zipfile
 from collections import OrderedDict
 from xml.etree.ElementTree import SubElement
@@ -115,11 +114,7 @@ def replace_script(parent, script):
 
 def parse_html(content):
     try:
-        # We need the reference to the HTMLParser later to get the document encoding,
-        # so we can't use the html5lib.parse convenience function here.
-        tree_builder = html5lib.treebuilders.getTreeBuilder("etree")
-        parser = html5lib.HTMLParser(tree_builder, namespaceHTMLElements=False)
-        document = parser.parse(content)
+        document = html5lib.parse(content, namespaceHTMLElements=False)
 
         if not document:
             # Could not parse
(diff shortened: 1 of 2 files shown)


kolibri/core/content/views.py
```diff
@@ -290,7 +290,11 @@ def get(self, request, zipped_filename, embedded_filepath):
         zipped_path = get_path_or_404(zipped_filename)
 
         # Sometimes due to URL concatenation, we get URLs with double-slashes in them, like //path/to/file.html.
-        # Normalize the path by converting those double-slashes to a single slash.
+        # If it is a leading double slash (i.e. /zipcontent/filename.zip//file.html) then it will get passed
+        # to zipcontent as a single leading slash - detect this and remove if present.
+        if embedded_filepath.startswith("/"):
+            embedded_filepath = embedded_filepath[1:]
```
indirectlylit: This is confusing. The comment refers to a leading double slash, but

* the example shows a double slash later in the path
* the test looks for a leading single slash

kolibri/core/content/views.py
```diff
@@ -140,23 +135,28 @@ def parse_html(content):
...
+            # By HTML Spec if doctype is included, it must be the first thing
+            # in the document, so it has to be the first child node of the document
+            doctype_node = tree.childNodes[0]
+
+            # Check that this node is in fact a doctype node
+            if doctype_node.nodeType == doctype_node.DOCUMENT_TYPE_NODE:
+                # render to a string by calling the toxml method
+                # toxml uses single quotes by default, replace with ""
+                doctype = doctype_node.toxml().replace("'", '"')
+        except (html5lib.html5parser.ParseError, IndexError):
```
kollivier: It's worth noting that this will introduce a regression if we get an error other than these two errors, as it will not only fail to get the doctype, but the failure will result in a 500 and not render anything at all, instead of the quirky HTML that it serves today.

kolibri/core/content/views.py
kollivier: Maybe the comment can be improved, but the reason we need to handle this special case is because `embedded_filename` has already had the leading slash removed by Django as part of its built-in URL parsing. (See the `/zipcontent` URLs formatting string for more details) Hence for the URL `/zipcontent/filename.zip//file.html`, when `get` is called, `embedded_filename` will have the value `/file.html`, and thus will not be a relative path, but it also won't be a double-slash so it won't be caught by the original handling.

kollivier (changes requested):
I feel pretty uneasy about failing to load HTML content if we get some unexpected error when trying to get the doctype, as all HTML5 content in current Kolibri versions render without a doctype so we know that most content renders reasonably well even with the wrong doctype.

kolibri/core/content/views.py
indirectlylit: Ah, I see. I didn't look closely enough at it to notice `filename.zip` in the middle

kolibri/core/content/views.py
rtibbles: I am struggling to understand what other errors you think might be thrown - even the parser error catch is a bit of overkill, as that error would already have been thrown by the outer parse invocation if there was something fundamentally broken.

As the parser has already run by this point without triggering an exception, the only place an error could occur is `doctype_node = tree.childNodes[0]` if the tree happens to be completely empty. Even this is very unlikely as passing any string to `parser_dom.parse` always produces at least one child node.

kolibri/core/content/views.py
rtibbles: I am happy to accept suggestions for how to rephrase the comment :)

kolibri/core/content/views.py
kollivier: If there is one thing that life and my many years of coding has taught me, it's that I can't anticipate every possible problem that could happen, and also that code changes over time and so code that may not have thrown a certain error before may start throwing it later (or may start to error out with the introduction of new content unlike our current content base).

A lot of times, throwing a hard error is the best you can do, because you would need to know the details of the error in order to know what to do about it. In this case, though, we do know that it will render the page without the doctype, and in the case of most of our existing HTML5 apps, we know they will also work. So what is the argument against allowing them to continue loading in the face of an unexpected error?

kolibri/core/content/views.py
indirectlylit: The comment assumes knowledge of the url pattern defined externally in `urls.py`, but by the time it gets here that's already been parsed out. Perhaps:

```suggestion
        # If the embedded file path has a leading double slash (i.e. //file.html) it should get passed
        # to zipcontent with a single leading slash instead.
        if embedded_filepath.startswith("/"):
            embedded_filepath = embedded_filepath[1:]
```

The method signature could also be elaborated upon. For example:

```python
    def get(self, request, zipped_filename, embedded_filepath):
        """
        Handles GET requests and serves a static file from within the zip file.

        zipped_filename is the name of a zip file, and embedded_filepath is a
        path to something within that zip file.
        """
```

indirectlylit: I'm also surprised based on the comment that the code isn't this instead:

```python
        if embedded_filepath.startswith("//"):  # double slash
            embedded_filepath = embedded_filepath[1:]
```

And ... also surprised that this case isn't handled by the line below :

```python
# Normalize the path by converting double-slashes occurring later in the path to a single slash.
embedded_filepath = embedded_filepath.replace("//", "/")
```

kollivier: `embedded_filepath` will not start with `//`, because we tell Django that the URL is formatted as `/zipcontent/<zipped_filename>/<embedded_filepath>`. Notice the `/` in between `<zipped_filename>` and `<embedded_filepath>`. So if the URL is:

`/zipcontent/1234.zip//filename.html`

then when `get` is called, `zipped_filename = "1234.zip"` and `embedded_filepath = "/filename.html"` (not `//filename.html`). 

 As `embedded_filepath` should be a relative path to the file, and `/filename.html` is an absolute path starting from the root of a filesystem, we want to strip that `/` from the URL.

indirectlylit: ah OK - thanks for clarifying. So there is no leading double slash! In that case, I would write:

```suggestion
        # If the embedded file path has a leading slash (e.g. /file.html), it should get passed
        # to zipcontent with no leading slash.
        if embedded_filepath.startswith("/"):
            embedded_filepath = embedded_filepath[1:]
```

kollivier (approved):

PR #6670, opened 2020-03-19
Merge develop -> design system branch

indirectlylit:
Also does some linting

in resolving merge conflicts, some changes were lost and will need to be cherrypicked into the other repo. see: https://github.com/learningequality/kolibri-design-system/issues/12

cc @jonboiser


indirectlylit:
in resolving merge conflicts, some changes were lost and will need to be cherrypicked into the other repo. see: https://github.com/learningequality/kolibri-design-system/issues/12

indirectlylit:
self-merging even though some things are broken - let's try to address in subsequent PRs

PR #6672, opened 2020-03-19
Track AnonymousUser in ContentSessionLog

nucleogenesis:
### Summary

- In custom Auth middleware, create a session for AnonymousUser

- Add CharField `anonymous_session_id` on ContentSessionLog model.

- `max_length=40` chosen based on what I found in Django for the DB field for `session_key` ([link here](https://github.com/django/django/blob/b9cf764be62e77b4777b3a75ec256f6209a57671/django/contrib/sessions/migrations/0001_initial.py#L15))

- Ensure in the ContentSessionLogSerializer that when we don't have a
user, then we set the new `anonymous_session_id` with the value of the
AnonymousUser's session `session_key`.

Result is that anon user's behavior is tracked in the ContentSessionLog

### Reviewer guidance

- Open the sqlite db in a viewer and check out the `logger_contentsessionlog` table.
- Check out some content as a guest - refresh your DB and see that there are records for that user.
- Ensure that `progress` and `time_spent` and so forth are updated as expected. `user` will be `NULL` and `anonymous_session_id` will have a value.
- Do the same while logged in and see that the `anonymous_session_id` is `NULL` but `user` is not.

Consider any other places where this could possibly create issues?

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated


indirectlylit (commented):
Thanks! couple quick notes:

* this is targeted at `develop` and it should be 0.12 I believe?
* got some failing tests - didn't look closely

Also, I was thinking maybe it's simpler to just set `session_id` unconditionally, rather than making a new special concept of an `anonymous_session_id`? I recall we discussed `anonymous_session_id` being mutually exclusive with `user`, but I don't remember to what end

nucleogenesis:
@indirectlylit Test suite didn't have issue locally - but I made a change to address what I saw in Travis.

Also always set the session_id on the log - also removed anonymous labeling. Hopefully Travis is appeased and this is good to go.

kolibri/core/auth/middleware.py
```diff
@@ -64,3 +65,6 @@ def process_request(self, request):
             "'kolibri.core.auth.middleware.CustomAuthenticationMiddleware'."
         )
         request.user = SimpleLazyObject(lambda: _get_user(request))
+        # Ensure that we create a session for the AnonymousUser
+        if not request.session.get("anonymous_session_id"):
+            request.session.__setitem__("anonymous_session_id", str(uuid.uuid4()))
```
jamalex: I think this will still be coupled to the Django session. I believe it'll need to go through separate cookies, akin to:
https://stackoverflow.com/questions/1622793/django-cookies-how-can-i-set-them/5575578#5575578
(so it may need to go on `process_response`, instead, so you can set the cookie?)

nucleogenesis:
@jamalex 

Alrighty - got it working with COOKIES rather than sessions.

I thought initially that our sessions were working through cookies and that the previous implementation was effectively the same. However is the concern that attaching it to the session a la `request.session.__setitem__()` would persist it server-side whereas using the COOKIES like you mentioned would let the key transient server-side and persisted only client-side?

rtibbles:
This implementation seems to diverge from what we had previously discussed. Maybe I am missing the interim discussion.

In our previous conversation, I had suggested just using the Django session key for the anonymous id. So reading from `request.session.session_key` and setting that. This gets automatically cycled at login, and will change whenever the session expires.

The current implementation seems to be doing anonymous user tracking across multiple sessions (due to the glorious 100 year cookie), which is not what we had discussed previously.

nucleogenesis:
@rtibbles That's what I did in my first implementation - but @jamalex suggested the use of cookies per the latest commit. I think perhaps Devon and I misremembered the exact way that Jamie intended on implementing this and tracking the data we want to track.

kolibri/core/auth/middleware.py
```diff
@@ -64,3 +66,14 @@ def process_request(self, request):
             "'kolibri.core.auth.middleware.CustomAuthenticationMiddleware'."
         )
         request.user = SimpleLazyObject(lambda: _get_user(request))
+
+    def process_response(self, request, response):
```
rtibbles: Might be simpler to add this only in the session endpoint - we are always guaranteed to hit it, and it means we only do this once every four minutes, and not on every request to the server.

Can add this cookie setting here instead: https://github.com/learningequality/kolibri/blob/release-v0.12.x/kolibri/core/auth/api.py#L379

rtibbles: Note - my ulterior motive for preferring this is that modifying the session during a request makes Django automatically add `Vary: Cookies` header to any request, which makes it much harder to cache, so if we can restrict these kind of things only to user specific requests, it is preferable!

nucleogenesis: @rtibbles I'll give that a look in the AM tomorrow. I'm a bit confused since I'm using the `HttpResponse.set_cookie()` method for setting the cookie on the rather than setting it on the `session` object using `__setitem__()`.

That point in the code you reference has a session object and @jamalex wants to use a cookie of our own making instead so that it will never expire.

jamalex:
>I recall we discussed anonymous_session_id being mutually exclusive with user, but I don't remember to what end

@indirectlylit I know we discussed this again recently, but not sure whether that was before or after your comment. The reason we only set it on anonymous logs was for privacy, to avoid being able to correlate anonymous usage in a browser with logged in use in the same browser. It may not be a concern in practice, but we were being cautious. From a nuanced data analysis perspective, always storing it is preferred.

>I thought initially that our sessions were working through cookies and that the previous implementation was effectively the same. However is the concern that attaching it to the session a la request.session.__setitem__() would persist it server-side whereas using the COOKIES like you mentioned would let the key transient server-side and persisted only client-side?

@nucleogenesis Whether the session data is stored in cookies or server-side, setting information on the session is scoped to that session, and wouldn't persist (cleared when session changes).

>The current implementation seems to be doing anonymous user tracking across multiple sessions (due to the glorious 100 year cookie), which is not what we had discussed previously.

@rtibbles That is the desired behavior. This is to replace the visitorid stuff that we currently do in nginx, which is aimed at counting unique anonymous visits, and measuring things like retention. We don't want to double-count it as two users when the same anonymous browser visits on two separate occasions, so this needs to persist beyond the scope of a Django session.

kolibri/core/auth/middleware.py
jamalex: While this doesn't modify the session, as @nucleogenesis notes, so it may or may not have an effect on caching, moving this out of a middleware and into the session endpoint would also help by reducing the overhead somewhat -- even though this is a very lightweight middleware (no IO, just CPU), it is still adding a layer to the onion for every single request Django is processing.

jamalex:
I think part of what makes this confusing (in terms of intention) is also its naming as a "session" ID, when really it's a "visitor" ID or "browser" ID. Worth renaming?

kolibri/core/auth/middleware.py
rtibbles: One update you would need to make is to turn the `get_session` method into a `get_session_response` method, so that you return the `Response` object from the method.

With this, you can use `response.set_cookie` method to set the cookie in the anonymous case.

nucleogenesis:
@jamalex @rtibbles 

Pending Travis - the latest changes:

- `visitor_id` instead of `anonymous_session_id`
- `get_session` -> `get_session_response` where I set the cookie if it doesn't exist already.

rtibbles:
My other concern with the 100 year cookie is that I had previously set a multi year cache expiry and it caused issues on Windows for being too big a number.
https://github.com/learningequality/kolibri/blob/develop/kolibri/core/content/api.py#L72

I don't know if the same concern might apply to cookies as to caching, but it seems sensible to limit to a year, rather than 100.

nucleogenesis:
> My other concern with the 100 year cookie is that I had previously set a multi year cache expiry and it caused issues on Windows for being too big a number.
> https://github.com/learningequality/kolibri/blob/develop/kolibri/core/content/api.py#L72
> 
> I don't know if the same concern might apply to cookies as to caching, but it seems sensible to limit to a year, rather than 100.

@rtibbles If we want it to be persistent eternally - we could update the expiry if it exists?

```
# visitor_cookie_expiry set to 1 year in future

if not request.COOKIES.get("visitor_id"):
  response.set_cookie(
    "visitor_id", visitor_id, expires=visitor_cookie_expiry
  )
else:
  response.set_cookie(
    "visitor_id", request.COOKIES.get("visitor_id"), expires=visitor_cookie_expiry
  )
```

rtibbles:
Yeah, I think updating it is fine, especially as we are not doing it on every request now.

indirectlylit:
@nucleogenesis one linting error: 

```
kolibri/core/auth/middleware.py:1:1: F401 'datetime.timedelta' imported but unused
```

lyw07:
as requested by @indirectlylit, I tested this PR locally, and it works as described in the reviewer guidance 🎉
but I'm not sure if AnonymousUser information is expected to be seen in the content session log csv files generated by users through `Facility --> Data`, as I'm not seeing the anonymous user data in the generated csv files. Are we only including the non-anonymous user data in the csv files? Thank you!

rtibbles (approved):

nucleogenesis:
Thanks @lyw07 !

PR #6674, opened 2020-03-19
Bulkimport

jredrejo:
### Summary
Creation of users, roles, classrooms and users assignments to these classrooms after importing a CSV file, following the schema related at https://docs.google.com/document/d/1CobblEL--x7iyolahdY5f2PgGnTdvJetskIESyfDXP0/edit

### Reviewer guidance
Using the command:

```
usage: kolibri manage bulkimportusers 
                                      [--no-color] [--facility FACILITY]
                                      [--dryrun] [--delete] [--userid USERID]
                                      filepath
```
Example:
` kolibri manage bulkimportusers ./mock.csv --delete --dryrun`
Try to import a csv, and check all the new users, classes and their learners, admins and coaches are correctly assigned.

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated


kolibri/core/assets/src/api-resources/task.js
```diff
@@ -111,6 +111,19 @@ export default new Resource({
...
+   * @param {string} params.filepath - session or summary
+   * @param {string} params.dryrun - validate objects but don't write in the db
+   * @param {string} params.delete - delete users not included in the csv
+   *                                 and clear not included classrooms
+   * @returns {Promise}
+   *
+   */
+  bulkimportusers(params) {
+    return this.postListEndpoint('bulkimportusers', pickBy(params));
+  },
```
indirectlylit: Maybe `import_users_from_csv` instead of `bulkimportusers`? ("bulk" seems implied, and maybe good to specify this is a CSV import in case we add other options")

kolibri/core/auth/management/commands/bulkimportusers.py
```diff
@@ -0,0 +1,587 @@
...
+    "Username",
+    "Password",
+    "Full name",
+    "User type",
+    "Identifier",
+    "Birth year",
+    "Gender",
+    "Enrolled in",
+    "Assigned to",
+)
```
indirectlylit, line 40: Where are these headers actually used? Are they displayed to end users or used for validation?

See also: https://github.com/learningequality/kolibri/issues/6681

kolibri/core/auth/management/commands/bulkimportusers.py
```diff
@@ -0,0 +1,587 @@
...
+    """
+
+    def checker(v):
+        if int(v) < min or int(v) > max:
+            raise ValueError(v)
+
+    return checker
+
+
+def value_length(length, null=False):
```
indirectlylit: I found the expression `if null` a little confusing.

Maybe

```suggestion
def value_length(length, allow_null=False):
```

which would make it:

```python
if allow_null and v is None:
```

kolibri/core/auth/management/commands/bulkimportusers.py
```diff
@@ -0,0 +1,587 @@
...
+    """
+    Return a value check function which raises a ValueError if the supplied
+    value has a length greater than 'length'
+    If null is not True raises a ValueError if the supplied value is None.
+    """
+
+    def checker(v):
+        if null and v is None:
+            return checker
```
indirectlylit: I'm a bit confused what's happening here. Shouldn't this just be `return`, not `return checker`?

kolibri/core/auth/management/commands/bulkimportusers.py
```diff
@@ -0,0 +1,587 @@
...
+
+    def add_check(self, header_name, check, message):
+        """
+        Add a header check, i.e., check whether the header record is consistent
+        with the expected field names.
+        `header_name` - name of the header for the column to be checked
+        `check`- function to be used as validator of the values in the column
+        `message` - problem message to report if a value is not valid
+        """
+        self._checks.append((header_name, check, message))
```
indirectlylit, line 141: rather than a `message`, perhaps we should instead be creating validation codes? Then on the client side we can map code to internationalized messages?

Alternatively, we could do internationalization on the backend.

kolibri/core/auth/management/commands/bulkimportusers.py
```diff
@@ -0,0 +1,587 @@
...
+                        continue
+                    if classroom in group:
+                        group[classroom].append(username)
+                    else:
+                        group[classroom] = [
+                            username,
+                        ]
+            except AttributeError:
+                # there are not members of 'key'
+                pass
```
indirectlylit: Which lines are expected to throw `AttributeError`? Maybe the `try`...`except` can wrap a smaller number of lines?

Also, a `group` in Kolibri is a collection inside a class, so the variable name here is a bit confusing. Maybe `class_list`?

kolibri/core/auth/management/commands/bulkimportusers.py
```diff
@@ -0,0 +1,587 @@
...
+
+        # enrolled learners:
+        append_users(self.classrooms, "Enrolled in")
+
+        # assigned coaches
+        user_role = row.get("User type", "learner").lower()
```
indirectlylit: We shouldn't assume that the headers will be in English. Perhaps we can instead make constants which refer to column numbers, e.g.

```python
USER_TYPE_COL = 3
```

etc...?

kolibri/core/auth/management/commands/bulkimportusers.py
```diff
@@ -0,0 +1,587 @@
...
+        if user_role != "learner":
```
indirectlylit: flagging https://www.notion.so/learningequality/For-specifying-gender-user-role-and-other-enums-use-ALL_CAPS_IDS-8f65a14d514b4a02932e4142c79a17c0

(though we could also do a `string.lower()` on everything)

kolibri/core/auth/management/commands/bulkimportusers.py
```diff
@@ -0,0 +1,587 @@
...
+                        "field": header_name,
+                        "value": value,
+                    }
+                    error_flag = True
+                    yield error
+                except Exception as e:
+                    error = {
+                        "row": index + 1,
+                        "message": "Unexpected error [%s]: %s"
+                        % (e.__class__.__name__, e),
```
indirectlylit: very minor note: generally I'm a fan of `string.format()` rather than printf-style string formatting because it looks more pythonic

kolibri/plugins/facility/assets/src/constants.js
```diff
@@ -64,3 +65,13 @@ export const CSVGenerationStatuses = {
...
   AVAILABLE: 'AVAILABLE',
 };
+
+export const CSVImportStatuses = {
+  NOT_STARTED: 'NOTSTARTED',
+  VALIDATING: 'VALIDATING',
+  VALIDATED: 'VALIDATED',
+  SAVING: 'SAVING',
+  ERRORS_VALIDATING: 'ERRORSVALIDATING',
+  ERROR_SSAVING: 'ERRORSSAVING',
```
indirectlylit:
```suggestion
  VALIDATION_ERRORS: 'VALIDATION_ERRORS',
  SAVE_ERRORS: 'SAVE_ERRORS',
```

indirectlylit (commented):
Looking good! haven't tested yet but here's a quick code review

kolibri/core/assets/src/api-resources/task.js
jredrejo: ok, naming is one of the most difficult things :) I doubted among different options, because there's another importusers from csv command. Also this one validates, create classes, roles, etc. so that could be also part of the name. 
I have changed it as you say (but the actual command is still named bulikimportusers.py)

kolibri/core/auth/management/commands/bulkimportusers.py
jredrejo: Both things, they are in the specs doc, https://docs.google.com/document/d/1CobblEL--x7iyolahdY5f2PgGnTdvJetskIESyfDXP0/edit#heading=h.twkn8j1fjeiq

kolibri/core/auth/management/commands/bulkimportusers.py
jredrejo: thanks, makes sense

kolibri/core/auth/management/commands/bulkimportusers.py
jredrejo: no, it should return `checker` as validators are functions that are passed later, see function `csv_values_validation` on line 244. It works similarly to decorators. Examples:

```
        validator.add_check("Birth year", number_range(1900, 99999), "Not a valid year")
        validator.add_check("Username", value_length(125), "User name is too long")
```

kolibri/core/auth/management/commands/bulkimportusers.py
jredrejo: yes, I thought of it but I also thought this is a mvp and I prefered not to spend time on it now. I prefer to test the behaviour of the script asap. Same thing with the progress_update calculations that are not real yet.

kolibri/core/auth/management/commands/bulkimportusers.py
jredrejo: for the AttributeErrork, it's line 144. If there is not group with that key, split can't be executed. I will refactor it to apply the try only to that line.

About the name group, yes it makes sense in a kolibri context. I was thinking of it just as its meaning. I  have changed it as you say. I am afraid this kind of naming conflict might happen more times in the code.

kolibri/core/auth/management/commands/bulkimportusers.py
jredrejo: As Ivan mentioned in a thread, I think translating headers is quite dangerous. It's just a data type. I still remember how Microsoft regreted after translating VB for applications in the first Excel versions and had to roll it back later... I am not sure it's a good idea, because it will be really hard to validate the headers names, and to map them to our right field names.
In any case, if we decide to do it, I think this is something similar to the messages errors question: in case we do it it can be done later, after the code is tested and works correctly.

kolibri/core/auth/management/commands/bulkimportusers.py
jredrejo: yeah, I use the capitalization (imported from kolibri.core.auth.constants.demographics ) in the validation, this is just a quick trick, but I don't have any problem in using upper if you like it more ;)

kolibri/plugins/facility/assets/src/constants.js
jredrejo: thanks!

jredrejo:
> Looking good! haven't tested yet but here's a quick code review

@indirectlylit  Thank so much! I really appreciate your comments. I have applied most of your suggestions, and answered all the questions. There's only a couple of disagreement points we can discuss.

kolibri/core/auth/management/commands/bulkimportusers.py
indirectlylit: okay

kolibri/core/auth/management/commands/bulkimportusers.py
indirectlylit: I think translating headers is very different from translating enumerated constants because columns can be determined by column number independently of their name

indirectlylit: I guess the question relates to our ability to extend the import format later:

1. Do we add more columns but are forced to keep the order?
2. Can we add and remove columns as long as we keep the same names?

(1) is easier for i18n but (2) is better for long-term flexibility of the format

kolibri/core/assets/src/core-app/client.js
```diff
@@ -86,9 +86,19 @@ const client = options => {
       options.params[cacheBust] = cacheBust;
     }
   }
+  let mimeType = 'application/json';
+  // in case a File obj must be POSTed:
+  if (options.entity !== 'undefined' && options.method === 'POST') {
+    if (
+      Object.values(options.entity).filter(function(a) {
+        return typeof a === 'object';
+      }).length > 0
```
indirectlylit:
```suggestion
      Object.values(options.entity).filter(a => typeof a === 'object').length
```

kolibri/core/auth/management/commands/bulkimportusers.py
```diff
@@ -0,0 +1,614 @@
...
+        # assume the first argument defines the membership
+        members = args[0].lower()
+    else:
+        members = tuple(map(str.lower, args))
+
+    def checker(value):
+        if value.lower() not in members:
+            raise ValueError(value)
+
+    return checker
```
indirectlylit: This function is a bit confusing because of the `assert` and `if`/`else` branches. It's also only used twice, which seems to imply that it should really just be two functions with more explicit function signatures?

indirectlylit (commented):
hey sorry, I did not have time to work on this in detail. Just dropping a couple quick notes

kolibri/core/assets/src/core-app/client.js
jredrejo: ah, perfect, I thought we had the rule of not using ES6 syntax. I also prefer that one.

kolibri/core/auth/management/commands/bulkimportusers.py
jredrejo: ah, you're right, the assert was in my first iteration doing it aka TDD, I'll remove it. I's not needed anymore.
In any case, the if/else are for a different reason, enumeration could be a string (not applicable in our case but I did it as a general validator, in case you want to ensure a char is in, for example "ABCXY")

PR #6702, opened 2020-03-26
pin build dependencies

indirectlylit:
started causing breakage in 0.13, but we probably need to pin these as far back as possible and merge forward

### DIFF
--- a/requirements/build.txt
+++ b/requirements/build.txt
@@ -1,12 +1,12 @@
 # Requirements for building wheels
 # This does not depend on runtime stuff so we do not
 # include base.txt
-sphinx
-sphinx-intl
-sphinx-rtd-theme
-sphinx-autobuild
+sphinx==2.4.4
+sphinx-intl==1.0.0
+sphinx-rtd-theme==0.4.3
+sphinx-autobuild==0.7.1
 pex<1.3
 wheel
 setuptools>=2.2,<20.11          # needed to get pex working # pyup: ignore
-beautifulsoup4
+beautifulsoup4==4.8.2
 requests==2.18.4
--- a/requirements/build.txt
+++ b/requirements/build.txt
@@ -1,7 +1,7 @@
 # Requirements for building wheels
 # This does not depend on runtime stuff so we do not
 # include base.txt
-sphinx==2.4.4
+sphinx==1.8.5
 sphinx-intl==1.0.0
 sphinx-rtd-theme==0.4.3
 sphinx-autobuild==0.7.1
--- a/requirements/docs.txt
+++ b/requirements/docs.txt
@@ -1,5 +1,2 @@
 -r base.txt
-
-# These are for building the docs
-sphinx==1.7.4
-sphinx_rtd_theme
+-r build.txt


requirements/build.txt
```diff
@@ -1,12 +1,12 @@
 # Requirements for building wheels
 # This does not depend on runtime stuff so we do not
 # include base.txt
-sphinx
-sphinx-intl
-sphinx-rtd-theme
-sphinx-autobuild
+sphinx==2.4.4
```
lyw07: I think we need to pin it to 1.8.5 since starting 2.0.0, it dropped python 2.7 support. https://github.com/sphinx-doc/sphinx/blob/v2.0.0/CHANGES#L38

```suggestion
sphinx==1.8.5
```

indirectlylit: thank you

requirements/docs.txt
```diff
@@ -1,5 +1,2 @@
 -r base.txt
-
-# These are for building the docs
-sphinx==1.7.4
-sphinx_rtd_theme
+-r build.txt
```
benjaoming, line 2: Perhaps `-r docs.txt` in `build.txt` instead for DRY? That way, building docs (on RTD) can be kept with minimal deps.

Going forwards, that's the same behavior. Of course it doesn't really matter since we don't serve old versions of developer docs :man_shrugging: 

https://github.com/learningequality/kolibri/blob/develop/requirements/docs.txt

benjaoming: I know this is to fix old breakage. So the minimal required fix is the best one.

Why do we have docs dependencies in our build dependencies? These are only for developer docs.

* If it's to build docs locally for developers, it could be in `dev.txt`.
* If it's for RTD, it should be in `docs.txt`
* If it's for RTD and developers, then `dev.txt` should have `-r docs.txt`

benjaoming: Am wondering if the reasoning for this structure is an old vision of including user docs in the distributed version of Kolibri. But since then, a) the user docs moved out of Kolibri and b) developer docs were created to only be built for `develop` and not for legacy branches.

In this case, the minimal fix can be to move docs dependencies out of the build dependencies to reduce dependency-related breakage in the future.

indirectlylit: > Perhaps `-r docs.txt` in `build.txt` instead for DRY?

How is this not DRY? It now only defines `sphinx` in one place rather than two.

I did it this way because docs.txt includes base.txt, and build.txt says:

```
# This does not depend on runtime stuff so we do not
# include base.txt
```

> the minimal required fix is the best one

I'm not sure what you're suggesting - is this not the minimal fix already? All I've done is pin dependencies.

benjaoming: > I did it this way because docs.txt includes base.txt, and build.txt says:

Oh right, missed that one, then the minimal fix would be to pin the versions only in `build.txt` and not in `docs.txt` since they're used by RTD.

Opened up #6709 to figure out the mystery of why Sphinx ended up in build dependencies. I looked at git blame stuff, and I can see that requirements have been added in relation to PRs targeting `docs/`, so likely it was always a mistake having the dependencies 1) in `build.txt` and 2) specified twice.

indirectlylit:
closing in favor of #6709

PR #6705, opened 2020-03-27
Fix edit modals to navigate properly on save

### LINKED ISSUE #6673: user edit page should exit after saving
### Observed behavior

Saving data leaves the user on the current page and switches 'cancel' to 'close' which is unconventional in our app:

The same is true for editing user permissions in device. Here, there's an additional little "Saving..."  message that shouldn't be there:

### Expected behavior

Saving behavior should be the same as a normal modal

### User-facing consequences

Confusing, inconsistent UI. Initially I thought it was broken when the page didn't close

### Steps to reproduce

edit and save user

### Context

0.13.1

julianduque:
### Summary

This pull request fixes the issue https://github.com/learningequality/kolibri/issues/6673 adding the proper navigation capabilities to the following components:

- UserEditPage
- UserPermissionsPage
- ProfileEditPage

### References

* Fixes: https://github.com/learningequality/kolibri/issues/6673

### DIFF
--- a/kolibri/plugins/device/assets/src/views/UserPermissionsPage/index.vue
+++ b/kolibri/plugins/device/assets/src/views/UserPermissionsPage/index.vue
@@ -203,6 +203,7 @@
             this.createSnackbar(this.$tr('permissionChangeConfirmation'));
             this.saveProgress = SUCCESS;
             this.uiBlocked = false;
+            this.goBack();
           })
           .catch(() => {
             this.uiBlocked = false;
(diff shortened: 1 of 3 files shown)


jonboiser:
Thank you @julianduque !. 

In UserEditPage and ProfileEditPage, you can remove this ternary in `computed.cancelButtonText`, since the button should only say "Cancel" now. In both pages, you can actually also remove any code that sets `this.status = 'SUCCESS'` or checks `this.status === 'SUCCESS'` since that was only used to display "close" on the button.

julianduque:
@jonboiser cleaned up the code as you suggested

jonboiser (approved):
Looks good! Thank you!

PR #6706, opened 2020-03-27
Use proper branding and icon for User modal

### LINKED ISSUE #6657: 'new user' full-screen modal is implemented as an "end state" page
### Observed behavior

The "New User" modal has a black titlebar and a "Back" button:

image

### Expected behavior

This is a modal, so it should be ~a purple header and~ [EDIT: this was incorrect] a "Close" button

### User-facing consequences

inconsistenct UI

### Context

0.13.2

julianduque:
### Summary

Changes the colors of the appBar in Fullscreen mode to match branding.

### Reviewer guidance

Go to Facility > Users and Create or Edit a new User.

### References

* Fixes https://github.com/learningequality/kolibri/issues/6657


indirectlylit:
thank you for your contributions today! Feel free to add yourself to the [authors file](https://github.com/learningequality/kolibri/blob/develop/AUTHORS.md)

packages/kolibri-components/src/styles/colorsDefault.js
```diff
@@ -14,8 +14,8 @@ export const defaultTokenMapping = {
   fineLine: '#dedede',
   appBar: 'brand.primary.v_400',
   appBarDark: 'brand.primary.v_700',
-  appBarFullscreen: 'palette.grey.v_900',
-  appBarFullscreenDark: 'black',
+  appBarFullscreen: 'brand.primary.v_400',
+  appBarFullscreenDark: 'brand.primary.v_700',
```
indirectlylit: Note that there are two pairs of color tokens, `appBar` (which already uses `brand.primary`) and `appBarFullscreen` (which correctly uses dark gray and black). In order to change the color of a particular app bar, we would want to point it the other token, not change the definitions here in `colorsDefault.js`

However this is moot given the mistake in the original issue I mentioned above.

indirectlylit (dismissed):
Thank you for the contribution! 

Unfortunately I made an error in issue #6657 and the guidance there was incorrect. (I've updated the issue)

Summary: full screen modals like this one _do_ in fact have black headers, but this one should still be switched to have an `x` close button rather than a `<-` back button.

Sorry for the confusion. If the hackathon is over and you don't have time to follow up, we'll just close this.

packages/kolibri-components/src/styles/colorsDefault.js
julianduque: Ah makes sense, I'll update the PR. Any existing modal with that behavior? I want to make sure I'm fixing it the right way

julianduque: @indirectlylit done, I think I understand the problem now, reverted back to the previous colors and updated the immersivePage behavior for those modal pages.

indirectlylit (approved):
thank you!

PR #6708, opened 2020-03-27
Fix language sort order in device settings drop-down

### LINKED ISSUE #6640: Languages in device settings are not sorted
### Observed behavior

languages are not sorted:

image

### Expected behavior

sort order should be the same as the language switcher:

image

### User-facing consequences

harder to select a language

### Steps to reproduce

change device language setting

### Context

0.13.1

crcastle:
### Summary

The order of the languages in the device settings drop-down now matches the sort order in the user's change language modal.

The sort order is determined in both places by the `language_name` key in https://github.com/learningequality/kolibri/blob/develop/kolibri/locale/supported_languages.json.

Manual verification was performed running locally. `yarn run test` was run, but it reported that no test-covered code was modified.

Updated device settings drop-down:

User's change language modal:
image

Old device settings drop-down order:

### Reviewer guidance

Review by doing: Hamburger menu -> Device -> Settings -> Default language drop-down

Compare the order of languages in that drop-down to the order in top-right [user name] -> Change language modal. They did not match previously. The order in the Change language modal seemed correct, so that is the order used in both places now.

Two minor concerns:

* The code used to compare the languages for sorting is duplicated [here](https://github.com/learningequality/kolibri/blob/develop/kolibri/core/assets/src/views/language-switcher/mixin.js#L21-L35) and in the `index.vue` file I edited. I couldn't figure out a way to reuse the mixin cleanly.
* Super minor, but it might be a nice UX improvement if the currently selected language is always shown as the first item in the drop-down list. "Browser Default" is always shown as the first item. Currently selected language is shown alphabetically with all other languages.

### References

This fixes https://github.com/learningequality/kolibri/issues/6640

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated


crcastle:
Looking at the checklist items, I don't seem to be able to add reviewers, labels, or milestones to this PR (nor do I know who or what those should be).

kolibri/plugins/device/assets/src/views/DeviceSettingsPage/index.vue
```diff
@@ -89,7 +89,7 @@
 
 <script>
 
-  import mapValues from 'lodash/map';
+  import { map } from 'lodash';
```
indirectlylit: We don't currently have the `sideEffects` webpack 4 optimization turned on, so [in order to support tree shaking](https://stackoverflow.com/questions/35250500/correct-way-to-import-lodash), we need to continue import from `lodash/map`

However, renaming `mapValues` to `map` seems like a good idea

indirectlylit (commented):
Thanks for the contribution! One small change requested below if you have time.

> The code used to compare the languages for sorting is duplicated here and in the index.vue file I edited. I couldn't figure out a way to reuse the mixin cleanly.

Code duplication here isn't a huge deal because it's only used twice. However if you're interested in breaking it into a shared utility, let us know and we can give guidance.

> Super minor, but it might be a nice UX improvement if the currently selected language is always shown as the first item in the drop-down list. "Browser Default" is always shown as the first item. Currently selected language is shown alphabetically with all other languages.

Yes, that would also be more consistent with language modal. Feel free to make that change if you'd like.

> Looking at the checklist items, I don't seem to be able to add reviewers, labels, or milestones to this PR (nor do I know who or what those should be).

Thanks yes - the checklist is not currently well-written for external contributors. We've been meaning to improve that

indirectlylit:
Also, feel free to add yourself to the [authors file](https://github.com/learningequality/kolibri/blob/develop/AUTHORS.md)

kolibri/plugins/device/assets/src/views/DeviceSettingsPage/index.vue
crcastle: That makes sense @indirectlylit. Updated!

crcastle:
> Code duplication here isn't a huge deal because it's only used twice. However if you're interested in breaking it into a shared utility, let us know and we can give guidance.

Some guidance would be great, @indirectlylit. I'm fairly new to large Vue projects (and kolibri). Where would the shared utility live? And how would it be included where needed -- as a mixin or some other way? Thanks!

indirectlylit:
It's a little bit complicated due to Kolibri's plugin and webpack architecture.

* Make a new module `compareLanguages.js` which exports a single function in: [`core/assets/src/utils/`](https://github.com/learningequality/kolibri/tree/develop/kolibri/core/assets/src/utils)
* Import the `compareLanguages` function into [`core/assets/src/core-app/apiSpec.js`](https://github.com/learningequality/kolibri/blob/develop/kolibri/core/assets/src/core-app/apiSpec.js) and add it to `utils` in the main default export object.

With this update to `apiSpec`, you'll be able to import the function from anywhere using our special dot syntax:

```js
import compareLanguages from 'kolibri.utils.compareLanguages';
```

Then you can use the function in both `core/assets/src/views/language-switcher/mixin.js` and `plugins/device/assets/src/views/DeviceSettingsPage/index.vue`

crcastle:
Made some good progress on this @indirectlylit, but I have one question as you probably know how people use kolibri better than I do.

There are two concepts of a "selected language" in kolibri -- one for the user and one for the device. The way I've implemented it now, the user's selected language always appears at the top of a list of languages. If a device language is selected, that does not appear at the top. It appears in alphabetical order. Is that ok?

See below for how it would look if, for example, I've chosen Español as my user language but English as my device language.

Screen Shot 2020-04-05 at 9 51 53 PM

indirectlylit:
thanks @crcastle!

Yes, I think that behavior is entirely reasonable. For user's language, we currently push it to the top of the list, but I think it would have been nearly the same UX if it remained alphabetical, and I doubt any users will be affected by the inconsistency.

We'll try to review this week

indirectlylit (approved):
Looks good, thank you!

I pushed a merge commit to your branch - hopefully that will get the tests to pass

Issue #6610, opened 2020-03-01
Logged zeroconf ERROR on exit (edge case?)


benjaoming:
### Observed behavior

When exiting with `CTRL+C` from `kolibri start --foreground`, an error is logged in rare cases.

CC: @micahscopes - you might understand this quicker than me?

Ref: https://github.com/learningequality/kolibri/pull/6113

### Expected behavior

Exits cleanly

### User-facing consequences

None, currently we log this edge case. So either it's anticipated and should be silenced, or we should figure out why it happens?

### Errors and logs

```
ERROR:zeroconf:Unknown error, possibly benign: ValueError('list.remove(x): x not in list',)
Traceback (most recent call last):
  File "/home/benjamin/.virtualenvs/kolibri/lib/python3.6/site-packages/zeroconf.py", line 2089, in remove_listener
    self.listeners.remove(listener)
ValueError: list.remove(x): x not in list
```

### Steps to reproduce

Unfortunately, this was a rare incident, so I cannot reproduce it.

### Context

0.13.x branch

benjaoming:
Potentially related odd exit behaviors reported here: https://community.learningequality.org/t/stop-job-on-raspberry-pi-when-shutting-down/1482/9?

CC: @jamalex

benjaoming:
Here's another error, on 0.13.2.dev

```
INFO     ENGINE Bus STOPPING
INFO     ENGINE HTTP Server cherrypy._cpwsgi_server.CPWSGIServer(('0.0.0.0', 8080)) shut down
INFO     Kolibri instance 'bb83e66afbf6e41049f5fe0d5ff1a45f' has left the zeroconf network.
WARNING  JOBCHECKER shut down event received; closing.
WARNING  JOBCHECKER shut down event received; closing.
WARNING  SCHEDULECHECKER shut down event received; closing.
^CINFO     ENGINE Caught signal SIGINT.
INFO     ENGINE Bus STOPPING
INFO     ENGINE HTTP Server cherrypy._cpwsgi_server.CPWSGIServer(('0.0.0.0', 8080)) already shut down
INFO     ENGINE Bus STOPPED
INFO     ENGINE Bus EXITING
INFO     ENGINE Bus EXITED
ERROR    ENGINE Error in 'stop' listener <bound method ServicesPlugin.stop of <kolibri.utils.server.ServicesPlugin object at 0x7f50556a2128>>
Traceback (most recent call last):
[...]

PR #6710, opened 2020-03-28
KOLIBRI_ZEROCONF_DISABLE to disable zeroconf broadcaster and listener

benjaoming:
### Summary

A problem is reported in shutdown behavior of the Zeroconf service, so suggesting at the very least that it can be switched off -- both to test and to hot-fix potential issues.

https://community.learningequality.org/t/stop-job-on-raspberry-pi-when-shutting-down/1482/10

### Reviewer guidance

Can be released into 0.13.x?

### References

#6610

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/utils/options.py
+++ b/kolibri/utils/options.py
@@ -130,6 +130,11 @@ base_option_spec = {
             "default": False,
             "envvars": ("KOLIBRI_SERVER_PROFILE",),
         },
+        "ZEROCONF_DISABLE": {
+            "type": "boolean",
+            "default": False,
+            "envvars": ("KOLIBRI_ZEROCONF_DISABLE",),
+        },
     },
     "Paths": {
         "CONTENT_DIR": {
--- a/kolibri/utils/server.py
+++ b/kolibri/utils/server.py
@@ -98,23 +98,26 @@ class ServicesPlugin(SimplePlugin):
 
         scheduler.start_scheduler()
 
-        # Register the Kolibri zeroconf service so it will be discoverable on the network
-        from kolibri.core.discovery.utils.network.search import (
-            register_zeroconf_service,
-        )
+        if not conf.OPTIONS["Server"]["ZEROCONF_DISABLE"]:
+            # Register the Kolibri zeroconf service so it will be discoverable on the network
+            from kolibri.core.discovery.utils.network.search import (
+                register_zeroconf_service,
+            )
 
-        register_zeroconf_service(port=self.port)
+            register_zeroconf_service(port=self.port)
 
     def stop(self):
         scheduler.shutdown_scheduler()
         if self.workers is not None:
             for worker in self.workers:
                 worker.shutdown()
-        from kolibri.core.discovery.utils.network.search import (
-            unregister_zeroconf_service,
-        )
 
-        unregister_zeroconf_service()
+        if not conf.OPTIONS["Server"]["ZEROCONF_DISABLE"]:
+            from kolibri.core.discovery.utils.network.search import (
+                unregister_zeroconf_service,
+            )
+
+            unregister_zeroconf_service()
 
         if self.workers is not None:
             for worker in self.workers:
--- a/kolibri/utils/server.py
+++ b/kolibri/utils/server.py
@@ -99,7 +99,7 @@ class ServicesPlugin(SimplePlugin):
         scheduler.start_scheduler()
 
         if not conf.OPTIONS["Server"]["ZEROCONF_DISABLE"]:
-            # Register the Kolibri zeroconf service so it will be discoverable on the network
+            # Kolibri zeroconf service: Periodically scans network and broadcasts self
             from kolibri.core.discovery.utils.network.search import (
                 register_zeroconf_service,
             )
--- a/test/do_mock_provisioning.py
+++ b/test/do_mock_provisioning.py
@@ -38,12 +38,12 @@ while time() < now + timeout and status_code != 201:
     try:
         response = requests.post(url, json=data)
         status_code = response.status_code
-    except requests.exceptions.RequestException:
-        pass
+    except requests.exceptions.RequestException as e:
+        print("Exception requesting provisioning API: {}".format(e))
 
 if status_code == 201:
     print("success!")
     exit(0)
 else:
-    print("failed with status %i" % status_code)
+    print("failed with status {}".format(status_code))
     exit(1)


rtibbles:
Not quite sure what this has to do with the referenced (closed) issue - except that they are both to do with zeroconf.

Reading the attached community forum post, it seems like the issue is reproducible when a facility has not yet been provisioned? Is that an accurate description?

benjaoming:
@rtibbles 
> Not quite sure what this has to do with the referenced (closed) issue - except that they are both to do with zeroconf.

#6610 is the correct issue, but the same genre: Issues with shutting down Zeroconf threads.

This is intended as a "hand brake" for current and future issues. And for simple configurability.

> Reading the attached community forum post, it seems like the issue is reproducible when a facility has not yet been provisioned? Is that an accurate description?

I tried with and without Facilities a few times and didn't see a pattern.

The current failure on Travis is suspicious - I have a local failure with `tox -e startup_provision_shutdown_without_zombies` which is also present on `release-v0.13.x`.

indirectlylit:
note, tests failing with 

```
  File "test/do_mock_provisioning.py", line 48, in <module>
    print("failed with status %i" % status_code)
TypeError: %i format: a number is required, not NoneType
```

As I've expressed in other threads, I'm generally opposed to adding configuration options without a strong reason to do so because every config option comes with long-term costs and commits us to supporting it forever.

I do see some value in this but not sure it's enough to warrant increasing the options.ini surface area at this time. If we do choose to add it, it probably shouldn't be called `KOLIBRI_ZEROCONF_DISABLE` because that's an implementation detail, not a user-facing concept.

benjaoming:
I'm not sure what else to call this than `ZEROCONF_DISABLE` -- if it can get more user facing, please add a suggestion and let's do that :+1:

benjaoming:
> As I've expressed in other threads, I'm generally opposed to adding configuration options without a strong reason to do so

Zeroconf has caused at least more than one shutdown issue. Currently, it has its own specially dedicated Travis environment because of this, and despite that, we still have issues with unclean shutdowns.

Creating an option to disable it makes sense to me: It's not going to be disabled by default so any further issues should still be seen and fixed.

As I said, this is a handbrake -- but if needed, it's also a configuration option to switch off something that people might not want or need? Say, if we connect Kolibri to a very large network, and it's not intended for the whole network.

benjaoming:
Had to figure out that the tox test failed because it spawns a daemon that logs exceptional crashes in the temporary environment's log files.

indirectlylit:
Is this disabling broadcasting, the ability to find other devices, or both?

benjaoming:
I read the comment and went with that, unfortunately it isn't accurate: So the option should not be nested in `OPTIONS["Server"]`, as it will also disable discovery of other zeroconf peers.

```
        # Register the Kolibri zeroconf service so it will be discoverable on the network
        from kolibri.core.discovery.utils.network.search import (
            register_zeroconf_service,
        )
```

This also disables the discovery service (part of `register_zeroconf_service`)

In either way, the intention is to disable Zeroconf as a handbrake or configurable option, such that the service doesn't get stuck, and I think both the broadcaster and the listener can be at fault, haven't studied those parts

indirectlylit:
`options.py` represents a de-facto, documented API. We shouldn't be committing to any new options that increase the configuration surface area but do not address a long-term user need. Every option in there will eventually need:

* user-facing documentation
* gherkin stories and integration testing during every release
* indefinite support or clear paths to deprecation

There may well be a concrete need to add this in the future, and if so we can re-visit at that time. Having a concrete need will also help us figure out a better name and categorization for the setting, rather than something which mentioning “zeroconf” which should be meaningless to end-users.

With regard for the need for a “handbrake”:

* If we are having issues with zeroconf and shutdown, let’s fix those issues and have people upgrade rather than advising people to turn off zeroconf
* If turning off zeroconf is useful for debugging these issues, we can switch it off using an environment variable that is _not_ in `options.ini`, so it’s clear that it’s not a documented feature

benjaoming:
@indirectlylit the implementation for disabling zeroconf doesn't have to be through options, but can also be added elsewhere, you can add such suggestions in the PR, and it can be addressed.

> let’s fix those issues and have people upgrade rather than advising people to turn off zeroconf

A handbrake is of value because there are issues that you sometimes don't anticipate and then you have to advice to switch it off. Unclean shutdown issues lead to database corruption, hence the elevated need.

indirectlylit:
> Unclean shutdown issues lead to database corruption, hence the elevated need.

Are you saying that the zeroconf process has direct database access? If that's true and we're pulling the plug, that's definitely problematic. However if it's not accessing the DB directly there shouldn't be any risk of corruption, correct?

benjaoming:
@indirectlylit yes, it accesses `kolibri.core.discovery.models`.

Since `unregister_zeroconf_service` isn't the last call (not sure if blocking) in `ServicesPlugin.stop()`, the problem cannot be isolated to just the zeroconf threads' database connections. I'm also not sure how CherryPy shuts things down or if it might have other things waiting for the `ServicesPlugin` to exit.

I agree that fixing the fundamental problem would be great -- and that this is not the ideal way to disable zeroconf. I think it's more ideal to write a plugin (so it becomes pluggable!) that writes to `kolibri.core.discovery.models` -- that pattern can be used to write more discovery-type plugins. But it's not the easy handbrake I was looking for :)

benjaoming:
This was a dead-end. It doesn't serve the original purpose of handbrake, and I'm not sure why stepping through Kolibri's call frames had shown it to be stuck in zeroconf, because disabling it doesn't make any difference.

* Tried the build artifact .deb in a virtual machine and disabling zeroconf does not make the 1:30 shutdown freezes go away
* Original reporter has started troubleshooting deeper in systemd and found other reasons why Kolibri might be shutting down uncleanly _or_ shutting down correctly but leaving behind some kind of unclean state that makes systemd stall.

This consistently happens on my 18.04 clean VM, and disabling Kolibri makes it consistently go away.

VirtualBox_Ubuntu 18 04 kolibri_09_04_2020_16_26_26

Opened https://github.com/learningequality/kolibri-installer-debian/issues/90 -- need more diagnostics to fix it, but opened the issue in kolibri-installer-debian because of the systemd theme.

benjaoming:
Comment from @JamesKainer https://community.learningequality.org/t/slow-system-shutdown-on-linux-systems-systemd-service/1482/18

> After installing your zeroconf disable build I’m happy to say that the pi shuts down pretty much instantly (5s). There’s no 23s delay and no 90s delay. Thanks for adding that option to test! I’m going to do another fresh install with the initd change and then set up the facility fully to see if that gets rid of the delay. If it doesn’t then I think that’s pretty good confirmation of the 23s delay being zeroconf.

PR #6721, opened 2020-03-30
Improve performance of facility class admin pages

rtibbles:
### Summary
The `HierarchyRelationsFilter` produces complex queries that are not performant. This results in very slow queries to the point of timing out in large class sizes when querying the facility user endpoint with a `member_of` query parameter.

This changes how the `member_of` query parameter filters users, by instead looking for a direct membership. This means that users will no longer show as members of a parent collection by virtue of being a member of the child collection. Practically, this makes no difference, as we do not allow such things to be set through the interface, but we should eventually make this necessary in order to enforce this properly in the code.

### Reviewer guidance
Load the Facility admin page for a specific class - make sure it loads, and is accurate.

### References
Query summary before the change:
Screenshot from 2020-03-30 15-48-24

Query summary after the change:
Screenshot from 2020-03-30 15-48-36

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/core/auth/api.py
+++ b/kolibri/core/auth/api.py
@@ -33,20 +33,20 @@ from rest_framework.response import Response
 from .constants import collection_kinds
 from .constants import role_kinds
 from .filters import HierarchyRelationsFilter
+from .models import AdHocGroup
 from .models import Classroom
 from .models import Collection
 from .models import Facility
 from .models import FacilityDataset
 from .models import FacilityUser
-from .models import AdHocGroup
 from .models import LearnerGroup
 from .models import Membership
 from .models import Role
+from .serializers import AdHocGroupSerializer
 from .serializers import ClassroomSerializer
 from .serializers import FacilityDatasetSerializer
 from .serializers import FacilitySerializer
 from .serializers import FacilityUserSerializer
-from .serializers import AdHocGroupSerializer
 from .serializers import LearnerGroupSerializer
 from .serializers import MembershipSerializer
 from .serializers import PublicFacilitySerializer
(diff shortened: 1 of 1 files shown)


indirectlylit (dismissed):
This update causes a regression where the users tab becomes empty:

| after | before |
|--|--|
| image |image|

Maybe it would help to have two filters, e.g 

* `filter_member_of` which handles recursion
* `filter_direct_member_of` which is used where we know we only need direct children

rtibbles:
Thanks for testing, I obviously missed something here.

> Maybe it would help to have two filters

We could - the main question is whether this is helpful going forwards, as we never allow indirect membership in practice (so perhaps we should outlaw it and enforce denormalized membership?)

indirectlylit:
From what I can tell, membership in a facility is _not_ stored in the Membership table, but is instead stored implicitly by the Facility attribute on the User model, so there is some inconsistency there.

I don't know what the right answer is, but I do believe that we may eventually want to surface KDP collection concepts above facility (like Organization) within Kolibri, so that's a consideration when deciding whether all membership concepts should be denormalized.

One other thought: `HierarchyRelationsFilter` is used in many places besides here. Should we update this comment with warnings about the efficiency?

https://github.com/learningequality/kolibri/blob/18ddeea5a9b1e08a4fa0e1c66c612db0aef40f37/kolibri/core/auth/filters.py#L9-L16

indirectlylit:
removing 0.13.2 milestone

rtibbles:
Updated to fix issue - now filtering by collection membership or facility.

Performance remains the same.

indirectlylit (approved):
this looks good to me.

Before releasing 0.13.3, we can do a sanity-check integration test pass on coach and facility functionality with just a single platform

PR #6764, opened 2020-04-05
Remove direct dependency on ifcfg and use zeroconf.get_all_addresses

jamalex:
As discussed on Slack.

### DIFF
--- a/kolibri/utils/server.py
+++ b/kolibri/utils/server.py
@@ -5,10 +5,10 @@ from subprocess import CalledProcessError
 from subprocess import check_output
 
 import cherrypy
-import ifcfg
 import requests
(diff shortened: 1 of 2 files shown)


kollivier (approved):
💯

benjaoming:
@jamalex @kollivier For 0.14, we could add a patch for ifcfg for Android support and release it through ifcfg (we co-maintain that project).

indirectlylit:
> As discussed on Slack.

ref: https://learningequality.slack.com/archives/C5XMC1EAV/p1586039733091800

benjaoming:
Slack discussion doesn't note why Android support couldn't be added to ifcfg.

PR #6777, opened 2020-04-09
Update Sign In flow for devices with multiple facilities

### LINKED ISSUE #5875: validation message is shown immediately on auto-focus of multi-facility account creation page
### Observed behavior

minor issue with auto-focus after showing facility selection modal:

* on the sign in page, no validation is initially displayed (correct)
* on create account page, validation is immediately displayed

| sign in | create account |
|--|--|
| image | image|
| image | image |

### Expected behavior

no validation is initially displayed

### User-facing consequences

slightly aggressive validation

### Steps to reproduce

* Create a second facility, e.g. using `http://localhost:8000/api/auth/facility/`
* navigate to account creation page `http://localhost:8000/en/user/#/create_account`
* refresh and choose a facility
* note that validation message is shown immediately

### Context

0.12.8

nucleogenesis:
### Summary

Revision of login page and flow for better multiple facility handling.

Fixes #5875 

### Reviewer guidance

Design Notes:

- "Go back" will have arrow when KDS is merged in
- Button styles will also need fiddling but not much when KDS is merged in

You'll likely need to test this by creating one or a few new `KOLIBRI_HOME` folders. Before running the server, set `KOLIBRI_HOME=/nonexistant/path` and run the server. I should note that I had an issue where the first attempt at starting failed but I think that might be local - will test that separately anyway. If you have that issue, just run the server again and it should be good.

**Single Facility**

When you only have one facility - there is one design change to test.

- During Setup Wizard, select the "Personal" type of installation. 
- On the sign in page - nothing changes.

- On a new KOLIBRI_HOME - do setup wizard and select Formal, give it a name
- On the sign in page, you should see "Sign into FACILITY NAME"

- Do that again for a Non-Formal installation.
- On the sign in page, you should see "Sign into FACILITY NAME"

When you have multiple facilities, the following all should be true:

> **Note!!!**
> Sign in as your superuser and go to http://localhost:8000/api/auth/facility/ to create a new facility

- Follows the following design flow: https://www.figma.com/file/LM0YUkbWAOIerZ3EEg0FN6bS/P2P-Sync?node-id=605%3A1769
- When you are on the page prompting you to create a new account or sign in - you are taken to the "Select a facility" page.
- You can go back from that page to the previous page.

> **Note!!!!!!**
> Once you've selected a facility, you will not be taken back to the sign-in-or-sign-up prompt page with the two buttons. You need to remove the Local Storage value for 'facilityId' to see that page again.

- When you don't have a cookie set - you will always be taken first to the 'sign-in-or-sign-up' prompt page. Test this by trying to go to http://localhost:8000/en/user/#/signin once you've removed your LocalStorage.

- Test creating a user for a specific facility - then make sure you cannot log in as that user on a different facility and _can_ login as that user on the proper facility.

- Test that you can change your facility from the sign in page.

### References

https://www.notion.so/learningequality/Technical-design-90b9dac261ab4e97bd68b0840cd42a25

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated


kolibri/plugins/user/assets/src/views/SignInPage.vue
```diff
@@ -2,192 +2,100 @@
...
-            </p>
-            <div slot="options">
-              <component :is="component" v-for="component in loginOptions" :key="component.name" />
-            </div>
-            <p
-              v-if="showGuestAccess"
-              class="guest small-text"
+    <AuthBase>
+      <div
+        v-if="hasMultipleFacilities && selectedFacility.name"
```
jonboiser: This "Sign into 'Facility name'" should be enabled for a single-facility device when the `DeviceSetting.preset` is formal/nonformal. Right? I tried to take a shortcut by modifying the DeviceSetting model directly instead of going through the setup wizard

kolibri/plugins/user/assets/src/views/AuthBase.vue
```diff
@@ -0,0 +1,331 @@
...
+      logoText() {
+        return this.$kolibriBranding.signIn.title
+          ? this.$kolibriBranding.signIn.title
+          : this.coreString('kolibriLabel');
+      },
+      oidcProviderFlow() {
+        return plugin_data.oidcProviderEnabled && this.nextParam;
+      },
+      showGuestAccess() {
+        return this.facilityConfig.allow_guest_access && !this.oidcProviderFlow;
```
jonboiser: Hm, is this 'guest access' setting on both Device and Facility settings? I think it might be (it's "Allow users to access resouces..." on the Device Settings page). For this page, we only want to show this link if the Device Setting for guest access is on.

I think we might have left the facility field intact by accident. I see now that it's not something that is configurable for Facilities on the UI anymore.

kolibri/plugins/user/assets/src/views/SignUpPage.vue
```diff
@@ -40,12 +40,12 @@
             :disabled="busy"
           />
 
-          <template v-if="currentFacility">
+          <template v-if="selectedFacility.name">
```
jonboiser, line 43: Need to add a "Sign in with an existing account" escape link at the bottom of this form. (See last step of Scenario 1 on Figma)

kolibri/plugins/user/assets/src/views/UserIndex.vue
```diff
@@ -37,12 +41,18 @@
     mixins: [commonCoreStrings],
     computed: {
       ...mapState(['pageName']),
+      ...mapGetters(['facilities']),
       immersiveProperties() {
         if (this.pageName === PageNames.SIGN_UP) {
           if (!this.$route.query.step) {
+            const backRoute =
```
jonboiser, line 48: One fix for Navigation: 

If you are this point

Screen Shot 2020-04-09 at 3 19 44 PM

And you click "Create Account"

When you click the "X" exit link

Then you return to the SignIn page. (Currently it takes you to the signin/signup page).

kolibri/plugins/user/assets/src/views/SignInPage.vue
nucleogenesis: `selectedFacility` will only have a `name` if it's Formal/Nonformal

However using the preset makes sense - will need to see where that lives in the Store - I only looked for it on the Facility dataset properties

kolibri/plugins/user/assets/src/views/FacilitySelect.vue
```diff
@@ -0,0 +1,105 @@
...
+      },
+      label() {
+        return this.$route.query.next === PageNames.SIGN_UP
+          ? this.$tr('canSignUpForFacilityLabel')
+          : this.$tr('selectFacilityLabel');
+      },
+    },
+    methods: {
+      setFacility(facilityId) {
+        console.log(`setting facId ${facilityId}`);
```
jonboiser: Should remove this console log

kolibri/plugins/user/assets/src/views/AuthBase.vue
nucleogenesis: Ahh okay good to know. That's something that was there before but good to change it now since I'm working on the code. I'll update it to use the device setting here

kolibri/plugins/user/assets/src/views/FacilitySelect.vue
nucleogenesis: I wonder why linting didn't get mad about that, thanks!

kolibri/plugins/user/assets/src/views/SignInPage.vue
```diff
@@ -2,192 +2,100 @@
...
+                :style="suggestionStyle(i)"
+                @mousedown.native="fillUsername(suggestion)"
               />
-            </p>
+            </ul>
           </div>
+        </transition>
+        <transition name="textbox">
+          <KTextbox
+            v-if="needPasswordField"
```
jonboiser, line 62: The `learner_can_login_with_no_password` doesn't seem to be reflected when I switch between facilities with differing settings. Maybe the facility settings need to be reloaded when you switch?

nucleogenesis: This was a great catch thanks JB!

kolibri/plugins/user/assets/src/views/SignInPage.vue
nucleogenesis: That `preset` is set for the Facility. I added it to the `FacilitySerializer` and check for that here instead now.

nucleogenesis: Actually - since this is not shown for an Informal facility - wouldn't we want a separate string altogether here in that case?

If all three facilities presets are possible on a single device - then we still want to show the user the ability to change their facility, right?

I'll raise this in the peer-sync Slack as well - just wanted to note it here.

kolibri/plugins/user/assets/src/views/AuthBase.vue
nucleogenesis: I have this fixed and ready to commit but the `GET devicesettings` endpoint permissions need to be changed. I've noted this in #kolibri-dev to see if anybody objects to opening the permissions there and with general approval I'll push it.

kolibri/plugins/user/assets/src/views/FacilitySelect.vue
```diff
@@ -0,0 +1,106 @@
...
+            {{ facility.name }}
+          </KButton>
+        </div>
+      </div>
+      <div v-if="facilityList['disabled'].length" class="disabled-facilities">
+        <p>{{ $tr('askAdminForAccountLabel') }}</p>
+        <div
+          v-for="facility in facilityList['disabled']"
+          :key="facility.id"
+          class="disabled facility-name"
```
jonboiser: I don't think this `.disabled` class is needed. It should display as unstyled text even without it.

kolibri/plugins/user/assets/src/views/SignInPage.vue
```diff
@@ -2,192 +2,100 @@
...
-            </p>
-            <div slot="options">
-              <component :is="component" v-for="component in loginOptions" :key="component.name" />
-            </div>
-            <p
-              v-if="showGuestAccess"
-              class="guest small-text"
+    <AuthBase>
+      <div
+        v-if="hasMultipleFacilities"
```
jonboiser: I think the logic should be 

1. Show the "Sign in..." prompt whenever there are more than 1 facilities _or_ this is a non-personal instance.
2. Show the "change" link only if there are 1+ facilities.

jonboiser (approved):
Looks good to me!

Issue #6786, opened 2020-04-14
creating/updating a user clears class assignment for other users


indirectlylit:
### Observed behavior

Reported by PF:

> When you assign a new user from a CSV (for example, a student) to a class, all other users who were previously assigned to that class disappear, and only the imported new user is assigned.

They clarified that the users who were removed from classes were _not_ in the uploaded CSV.

### Expected behavior

Assigning users to classes should have no impact on existing users, unless those users are in the CSV and have had their assignment altered there

### User-facing consequences

Unexpected loss of data

### Errors and logs

unknown

### Steps to reproduce

I haven't reproduced locally yet.

Unable to reproduce on the demo server due to https://github.com/learningequality/kolibri/issues/6787

### Context

PF branch

PR #6791, opened 2020-04-15
Don't delete users in previous classes when updating

jredrejo:
### Summary
Don't clear memberships when updating classes. 
If the delete option is set, the users will have been deleted previously. If the option is not set, classes should keep their users

### Reviewer guidance

1. Create a classroom
2. Add users to the classroom
3. Export the data  in a csv file
4. Edit the csv file to add a new user to any of the classrooms
5. Import the modified csv file

Check classrooms keep their previous users and have the new users added.

### References

Fixes: #6786

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/core/auth/management/commands/bulkimportusers.py
+++ b/kolibri/core/auth/management/commands/bulkimportusers.py
@@ -656,9 +656,7 @@ class Command(AsyncCommand):
                     Classroom.objects.create(
                         name=classroom.name, parent=classroom.parent
                     )
-                # clear users from classes to be updated:
-                update_classes = [c.id for c in db_update_classes]
-                Membership.objects.filter(collection__in=update_classes).delete()
+
                 self.add_classes_memberships(
                     classes, users_data, db_new_classes + db_update_classes
                 )


kolibri/core/auth/management/commands/bulkimportusers.py
```diff
@@ -656,9 +656,7 @@ def handle_async(self, *args, **options):
                     Classroom.objects.create(
                         name=classroom.name, parent=classroom.parent
                     )
-                # clear users from classes to be updated:
-                update_classes = [c.id for c in db_update_classes]
-                Membership.objects.filter(collection__in=update_classes).delete()
+
```
indirectlylit, line 659: two questions, just to double check:

1. will this still handle the scenario where:
    * a learner is enrolled
    * export a CSV
    * edit the learner in the CSV to have a blank enrollment column
    * import CSV
    * the learner should be removed from the class
2. does this behavior affect both learner enrollment and coach assignment?

jredrejo: 1. No, the learner won't be removed, that's adding another level of complexity. If needed we can handle it, but I'd do it in a different PR, as it's a total different history.
2. Yes, both assignment and enrollment are managed the same way

indirectlylit (approved):

PR #6805, opened 2020-04-24
Optimize static file loading

rtibbles:
### Summary
* Runs collectstatic once from initialize if any updates have occurred.
* Always runs collectstatic with the `clear` argument to clear the files before collecting
* Runs a capability check to see if we can generate symlinks
* If so, runs collectstatic with `link` option to use symlinks instead of copies to reduce file duplication.

### Reviewer guidance
Does everything still seem to work OK? Does everything work OK on Windows?

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/plugins/utils/__init__.py
+++ b/kolibri/plugins/utils/__init__.py
@@ -361,9 +361,6 @@ def run_plugin_updates():
                 ", ".join(config["UPDATED_PLUGINS"])
             )
         )
-        logger.info("Copying updated static files")
-        # Collect any static files
-        call_command("collectstatic", interactive=False, verbosity=0)
 
         update_manager = PluginUpdateManager(config["UPDATED_PLUGINS"])
 
@@ -375,6 +372,8 @@ def run_plugin_updates():
                     ", ".join(update_manager.errors)
                 )
             )
+        return True
+    return False
 
 
 def autoremove_unavailable_plugins():
--- a/kolibri/utils/cli.py
+++ b/kolibri/utils/cli.py
@@ -50,6 +50,7 @@ from kolibri.utils.conf import KOLIBRI_HOME
 from kolibri.utils.conf import LOG_ROOT
 from kolibri.utils.conf import OPTIONS
 from kolibri.utils.logger import get_base_logging_config
+from kolibri.utils.system import CAN_USE_SYMLINKS
 
 
 logger = logging.getLogger(__name__)
@@ -295,7 +296,9 @@ def initialize(skip_update=False):
 
     version = get_version()
 
-    if version_updated(kolibri.__version__, version):
+    updated = version_updated(kolibri.__version__, version)
+
+    if updated:
         check_plugin_config_file_location(version)
         # Reset the enabled plugins to the defaults
         # This needs to be run before dbbackup because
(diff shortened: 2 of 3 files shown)


kolibri/plugins/utils/__init__.py
```diff
@@ -375,6 +372,8 @@ def run_plugin_updates():
                     ", ".join(update_manager.errors)
                 )
             )
+        return True
+    return False
```
indirectlylit, line 376: minor stylistic note on code that pre-dated this PR: I usually like to bail early out of functions so the bulk of the logic isn't nested in a conditional:

```python
def run_plugin_updates():
    if not config["UPDATED_PLUGINS"]:
        return False

    logger.info(
        "Detected updates to plugins: {}".format(
            ", ".join(config["UPDATED_PLUGINS"])
        )
    )
    update_manager = PluginUpdateManager(config["UPDATED_PLUGINS"])
    update_manager.update_plugins()
    if update_manager.errors:
        raise PluginUpdateException(
            "{} failed to update, please restart Kolibri".format(
                ", ".join(update_manager.errors)
            )
        )
    return True
```

The same strategy can be used to flatten the function below:

```python
    if skip_update:
        return

    # Run any plugin specific updates here in case they were missed by
    # our Kolibri version based update logic.
    updated = run_plugin_updates() or updated

    if not updated:
        return

    logger.info("Copying updated static files")
    # Collect any static files
    collectstatic_options = {
        "interactive": False,
        "verbosity": 0,
        "clear": True,
    }
    if CAN_USE_SYMLINKS:
        collectstatic_options["link"] = True
    call_command("collectstatic", **collectstatic_options)
```

rtibbles: I tried to do a minimal diff here to prevent merge conflicts with the changes that have happened in develop. Happy to make further updates once this gets there.

indirectlylit:
@mrpau-dev would you mind testing this PR?

cpauya:
All good so far!  Tested these on Firefox v75.0 under Win7 VM:

* Import content
    * radina channel
    * TESSA
* Users
    * Class, assign coaches, enroll learners
* Coach class
    * plan topics, plan quiz
* Learn
    * Class (Lesson, Quiz), Channels, Take quiz
Screen Shot 2020-04-29 at 5 34 01 PM


Or are there other features we should test in particular?

indirectlylit (approved):
that works for me if it sounds like sufficient testing for you @rtibbles

rtibbles:
I think the only other thing we'd want to check is that the symlinking of files doesn't mess with nginx in kolibri-server, but this is probably best done as a QA pass after merge?

indirectlylit:
perhaps @jredrejo should review?

jamalex:
Main thing I raised that would be good to test is switching KOLIBRI_HOME to be on another filesystem (than where the Kolibri Python package has been installed), especially if that's a different filesystem type (e.g. FAT/NTFS) and making sure that works too.

indirectlylit:
thanks - tracked that task here: https://github.com/learningequality/kolibri/issues/6839

Issue #6794, opened 2020-04-15
When only one language option is hiding "below the fold", maybe just show directly?


jamalex:
### Observed behavior

See the list at the bottom of the screenshot:
![image](https://user-images.githubusercontent.com/618416/79396636-35be2400-7f31-11ea-9aa4-4b07a04c27ac.png)

### Expected behavior

It would take the same amount of space to show the hidden language directly as it does to show the button to unveil it.

indirectlylit:
seems like this will be useful for PF MVP so will do this in 0.13.x

PR #6806, opened 2020-04-24
Configurable languages

rtibbles:
### Summary
* Adds options.ini LANGUAGES option under Deployment to configure available languages
* Adds special keywords: `kolibri-supported`, `kolibri-all` as options that will prefill a list of languages
* Adds custom validation to verify valid languages
* Allows list or a single language entry
* Moves `supported_languages.json` to `language_info.json`
* Adds a new explicit list of support intl language codes in `kolibri.utils.i18n` to define the officially supported languages
* Gives validation error if any of the keywords would result in empty language lists
* Sets the Django language setting based on the first listed language
* Add special processing for the `supported` and `all` keywords to set "en" as the first listed language to maintain previous behaviour

### Reviewer guidance
Does the option setting work?

Does language switching still work as expected?

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated


kolibri/utils/options.py
```diff
@@ -49,6 +53,70 @@ def calculate_thread_pool():
...
+            raise VdtValueError(BETA_LANGUAGES)
+        return value
+    return [value]
+
+
+def language_list(value):
+    """
+    Check that the supplied value is a list of languages,
+    or a single language, or a special shortcut parameter.
+    """
```
indirectlylit: Is this just a check, or does it do some kind of transformation?

It's not really clear to me what is "allowed" or what the API here is

kolibri/utils/options.py
```diff
@@ -49,6 +53,70 @@ def calculate_thread_pool():
...
+BETA_LANGUAGES = "beta"
+
+
+def _set_en_first(value):
+    if "en" in value:
+        value.insert(0, value.pop(value.index("en")))
+
+
+def _coerce_language_value(value):
+    value = str(value)
```
indirectlylit: coerce from what to what?

indirectlylit (commented):
looks good, haven't tested. Does this setting have an impact on the ordering as shown to end-users? There's some nuance here, especially with how we have certain "priority" languages that appear independent of alphabetical order in the sign-in page

kolibri/utils/options.py
rtibbles: The library we are using does call it a check, but yes, it is also cleaning and changing the value to a standard output format.

Can add better comments.

Any ideas on how to best document the allowable values? (I think this would be helpful if we ever want to allow a UI to set these, as it would be useful to introspect them for the allowable values)

kolibri/utils/options.py
rtibbles: Pulled this out to avoid complexity complaints from linting, will add some comments.

indirectlylit (approved):
alright!

kolibri/core/assets/src/views/language-switcher/LanguageSwitcherList.vue
```diff
@@ -51,6 +52,18 @@
...
+
+  const language_priorities = {};
+
+  prioritized_languages.forEach((lang, index) => {
+    language_priorities[lang] = index + 1;
+  });
+
+  const selectableLanguages = Object.values(availableLanguages).filter(
+    lang => lang.id !== currentLanguage
+  );
```
indirectlylit: `availableLanguages` isn't properly initialized until the component has mounted, so this doesn't currently work on refresh

kolibri/core/assets/src/views/language-switcher/LanguageSwitcherList.vue
```diff
@@ -73,14 +86,28 @@
...
-        return prioritized_languages
-          .filter(lang => availableLanguages[lang] !== undefined)
-          .filter(lang => lang !== currentLanguage)
-          .map(lang => availableLanguages[lang])
-          .slice(0, this.numVisibleLanguages)
-          .sort(this.compareLanguages);
+        let buttonLanguages = selectableLanguages;
+        if (buttonLanguages.length > this.numVisibleLanguages + 1) {
+          buttonLanguages = buttonLanguages
+            .sort((a, b) => {
```
indirectlylit: Seeing some unexpected behaviors here that I'm looking into. Might be because `.sort` is in-place, and references `selectableLanguages`. Will investigate

indirectlylit (commented):
thanks! looking into a couple issues now

indirectlylit:
I'll push some updates to your branch and merge

indirectlylit:
I'm not quite sure what the intended sort behavior is: to completely respect the order as defined in the priority language array, or to use the `compareLanguages` function after having moved priority languages to the front of the list.

I do think there is a valid use-case for plugins to specify both the priority languages _and_ the order they're shown in. We may want to add that as a setting in the future.

However for the default case, I think it makes the most sense to keep the visible buttons sorted using the standard `compareLanguages` function, as it is currently in 0.13. I switched it back to the old behavior in 979d7a6c5c5991384b024ebbb73b92c50a5e2fee

indirectlylit:
fixes https://github.com/learningequality/kolibri/issues/6794

Issue #6821, opened 2020-04-29
PF-related CSV import updates


indirectlylit:
## Coach identifiers

The identifiers currently used are not consistent with our user-facing terminology. In order to address this, let's update as follows:

* `COACH` -> `FACILITY_COACH`
* `ASSIGNABLE_COACH` -> `CLASS_COACH`

## Password behavior

1. Should be a required field. Empty fields should throw an error
2. On export, output an asterisk (`*`) instead of an empty field
3. On import, leave unchanged in the case of an asterisk. Otherwise, update with the value.
4. Remove "If it's not provided the system will set it to 'kolibri'" behavior

## Update users issue

This is still an issue: https://github.com/learningequality/kolibri/issues/6787

## Handling of class lists

Class names in comma-separated lists should be case-insensitive, and have any leading or trailing whitespace trimmed.

The following two values should be treated identically:

```
My class,another class
```

and

```
My class,  AnotheR ClasS
```

## NEW: do not export super admins

Request that super admins not be included in exported users

### Context

latest `pf-mvp` branch

jredrejo:
@indirectlylit For the COACH identifiers, technically it's a trivial change, however, it's going to break the consitency with the rest of the code in Kolibri, using https://github.com/learningequality/kolibri/blob/develop/kolibri/core/auth/constants/role_kinds.py

jredrejo:
@indirectlylit about the password behaviour. Currently when a password value is not in the csv, it sets "kolibri" when creating an user because passwords can't be empty when creating an user. Don't you think that's better than triggering an error and skipping the user if it's empty?

If not, when creating an user, should `*` be replaced by a default password or trigger an error too?

indirectlylit:
> For the COACH identifiers, technically it's a trivial change, however, it's going to break the consistency with the rest of the code in Kolibri

These are user-facing strings, so there is no requirement that they are consistent with Kolibri code. They must instead be consistent with user-facing documentation and UI.

indirectlylit:
> @indirectlylit about the password behaviour. Currently when a password value is not in the csv, it sets "kolibri" when creating an user because passwords can't be empty when creating an user. Don't you think that's better than triggering an error and skipping the user if it's empty?
> 
> If not, when creating an user, should * be replaced by a default password or trigger an error too?

My intuition is that setting the password to the word "kolibri" seems arbitrary and unexpected.

I think we should tailor the errors to the condition to help people learn the behavior:

| scenario | condition | error message |
[...]

PR #6822, opened 2020-04-29
various PF updates

indirectlylit:
### Summary

Adds 60px to HTML5 app. @kollivier could you check whether this prevents the double-scrollbar issue?

@jredrejo could you please briefly check the items below?

| Summary | Before | After |
|--|--|--|
| Limit language options |  image | image |
| fix theme |  image | image |
| update success message |  image| image|
| update CSV reference | image|image|
| update export framing & layout |  image|image|

### Reviewer guidance

recommend reviewing and merging before addressing other issues (cc @jredrejo)

----

### Contributor Checklist

PR process:

- [x] PR has the correct target branch and milestone
- [x] PR has 'needs review' or 'work-in-progress' label
- [x] If PR is ready for review, a reviewer has been added. (Don't use 'Assignees')
- [ ] If this is an important user-facing change, PR or related issue has a 'changelog' label
- [ ] If this includes an internal dependency change, a link to the diff is provided

Testing:

- [x] Contributor has fully tested the PR manually
- [x] If there are any front-end changes, before/after screenshots are included
- [ ] Critical user journeys are covered by Gherkin stories
- [ ] Critical and brittle code paths are covered by unit tests

### Reviewer Checklist

- Automated test coverage is satisfactory
- PR is fully functional
- PR has been tested for [accessibility regressions](http://kolibri-dev.readthedocs.io/en/develop/manual_testing.html#accessibility-a11y-testing)
- External dependency files were updated if necessary (`yarn` and `pip`)
- Documentation is updated
- Contributor is in AUTHORS.md

PR #6826, opened 2020-04-29
PF-related CSV import updates

jredrejo:
### Summary
Several changes to the bulk import/export management commands to apply received feedback:

## Coach identifiers
`FACILITY_COACH` instead of `COACH` 
`CLASS_COACH` instead of `ASSIGNABLE_COACH`

## Password behavior
1. On export, output an asterisk (`*`) instead of an empty field
2. On import, it is a required field. Empty fields should throw an error
3. On import, when updating, leave unchanged in the case of an asterisk. Otherwise, update with the value.
4. Removed "If it's not provided the system will set it to 'kolibri'" behavior (this makes impossible to import an exported csv in a different facility)

## Exporting roles
Super admins are not included when exporting users

## Handling of class lists
Class names in comma-separated lists are case-insensitive, and have any leading or trailing whitespace trimmed.

### Reviewer guidance
Check tests pass and all the requirements are fulfilled

### References

Closes: #6821

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/core/auth/management/commands/bulkexportusers.py
+++ b/kolibri/core/auth/management/commands/bulkexportusers.py
@@ -144,6 +144,8 @@ def csv_file_generator(facility, filepath, overwrite=True):
         )
 
         for item in query.values(*db_columns):
+            if item["kind"] == role_kinds.ADMIN:
+                continue
             if item["username"] not in usernames:
                 item["password"] = "*"
                 writer.writerow(map_output(item))
--- a/kolibri/core/auth/management/commands/bulkexportusers.py
+++ b/kolibri/core/auth/management/commands/bulkexportusers.py
@@ -145,6 +145,7 @@ def csv_file_generator(facility, filepath, overwrite=True):
 
         for item in query.values(*db_columns):
             if item["username"] not in usernames:
+                item["password"] = "*"
                 writer.writerow(map_output(item))
                 usernames.add(item["username"])
             yield item
--- a/kolibri/core/auth/management/commands/bulkexportusers.py
+++ b/kolibri/core/auth/management/commands/bulkexportusers.py
@@ -19,6 +19,7 @@ from .bulkimportusers import FILE_WRITE_ERROR
 from .bulkimportusers import MESSAGES
 from .bulkimportusers import NO_FACILITY
 from kolibri.core.auth.constants import role_kinds
+from kolibri.core.auth.constants.demographics import DEFERRED
 from kolibri.core.auth.models import Classroom
 from kolibri.core.auth.models import Facility
 from kolibri.core.auth.models import FacilityUser
@@ -27,7 +28,6 @@ from kolibri.core.query import GroupConcatSubquery
 from kolibri.core.tasks.management.commands.base import AsyncCommand
 from kolibri.core.tasks.utils import get_current_job
 from kolibri.utils import conf
-
 
 try:
     FileNotFoundError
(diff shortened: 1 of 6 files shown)


kolibri/plugins/facility/assets/src/views/ImportCsvPage/InfoModal.vue
```diff
@@ -42,8 +42,8 @@
             <div>{{ $tr('possibleValues') }}</div>
             <ul>
               <li><code>ADMIN</code></li>
-              <li><code>COACH</code></li>
-              <li><code>ASSIGNABLE_COACH</code></li>
+              <li><code>FACILITY_COACH</code></li>
+              <li><code>CLASS_COACH</code></li>
```
indirectlylit: note I already am making this change in another PR so you can remove this to avoid a conflict:

https://github.com/learningequality/kolibri/pull/6822/files#diff-f27530fa36d3478dddd8bfc53dd093b7R57-R58

kolibri/core/auth/management/commands/bulkexportusers.py
```diff
@@ -66,14 +66,14 @@
...
-    role_kinds.ASSIGNABLE_COACH: "ASSIGNABLE_COACH",
+    role_kinds.COACH: "FACILITY_COACH",
+    role_kinds.ASSIGNABLE_COACH: "CLASS_COACH",
 }
 
 
 def not_specified(field, obj):
     val = obj[field]
-    return None if val == "NOT_SPECIFIED" else val
+    return None if (val == "NOT_SPECIFIED" or val == DEFERRED) else val
```
indirectlylit, line 76: Shouldn't `NOT_SPECIFIED` and `DEFERRED` be treated as separate values?

indirectlylit (approved):
looking good. I know it's still a WIP but I tested and it reset my password to an asterisk :)

kolibri/plugins/facility/assets/src/views/ImportCsvPage/InfoModal.vue
jredrejo: I have done a rebase to sync your changes

kolibri/core/auth/management/commands/bulkexportusers.py
jredrejo: Honestly, I don't know. Our code is not consistent on this. My impression is that `NOT_SPECIFIED` was used in the past and `DEFERRED` is being used recently. In any case, when exporting it seems that not returning any value in the cvs is a good policy because none of both will have a real meaning for the user (specially if the are not English spoken).

indirectlylit: they are different concepts:

* "Not specified" means the user intentionally chose to not to provide an answer
* "Deferred" means the user hasn't made a decision yet

I agree that an empty value should be a synonym for `NOT_SPECIFIED`

In general, if we export a CSV, we should be able to re-import it and no data would change. By treating these value the same, we would switch users from `DEFERRED` to `NOT_SPECIFIED`

indirectlylit (dismissed):
Still doesn't handle importing with asterisks: it actually changes my password to an asterisk

indirectlylit:
@jredrejo I'm confused - now I don't see any correct behavior:

* User types are still `COACH` and `ASSIGNABLE_COACH`
* Asterisks are not outputed for passwords on export
    image
* On import, an asterisk in the password field actually sets the user's password to be an asterisk:
    image

jredrejo:
> @jredrejo I'm confused - now I don't see any correct behavior:
> 
It's normal, this is not finished, it's still a WIP and only part of the export part is done, import is not done yet.

jredrejo:
> @jredrejo I'm confused - now I don't see any correct behavior:
> 
> * User types are still `COACH` and `ASSIGNABLE_COACH`
> * Asterisks are not outputed for passwords on export
>   image
> * On import, an asterisk in the password field actually sets the user's password to be an asterisk:
>   image

There is one last thing that concerns me about this behaviour: it breaks the possibility of exporting a csv and importing it later in another facility to clone its configuration. The exported csv has to be edited to change the password asterisks by something else. If that's fine, no problem.

indirectlylit:
> it breaks the possibility of exporting a csv and importing it later in another facility to clone its configuration

Yes, this was actually intentional. If someone wants to clone a facility, we have alternate mechanisms of doing that through Morango which keep IDs intact

indirectlylit (approved):
merging now - will test and file subsequent issues if necessary

PR #6864, opened 2020-05-08
Initial implementation of app plugin.

rtibbles:
### Summary
Initial implementation of app plugin for Android, MacOS etc.to toggle on app specific behaviour
Add `/app/api/initialize/<token>` endpoint that allows for initialization of the app view and setting the cookie.
Has an optional `next` GET param that will redirect to a specific location.

### Reviewer guidance
Is having a registry the right way to conditionalize the behaviour depending on the OS?

Get the token for the URL from Kolibri shell:
```
from kolibri.core.device.models import DeviceAppKey
DeviceAppKey.get_app_key()
```
Put it in the above endpoint - you should get a cookie set `app_key_cookie`.

Try again with `?next=/learn` appended, you should be redirected to learn.

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated


kolibri/core/assets/src/utils/appCapabilities.js
```diff
@@ -0,0 +1,7 @@
+import plugin_data from 'plugin_data';
+
+const appCapabilities = plugin_data.appCapabilities || {};
+
+export default {
+  launchIntent: appCapabilities.launch_intent,
+};
```
jamalex, line 43: I've been juggling in my head the need to define the set of supported capabilities centrally in `core` vs in a more P2P fashion between plugins (pub/sub sort of model). Was there a strong reason to prefer this direction?

Similarly, flagging to myself to check elsewhere to see how a plugin would handle the case of it wanting to use a capability that's not even defined in core (gracefully) -- with an eye to supporting external/public plugins down the road.

kolibri/core/assets/src/utils/browser.js
```diff
@@ -37,4 +38,10 @@ export function isEmbeddedWebView() {
   return isEmbedded || isAndroidWebView();
 }
 
+export function isAppView() {
+  const isLocal = window.location.host !== '127.0.0.1' || window.location.host !== 'localhost';
```
jamalex: Isn't this... a tautology? Do we need to remove the `!`'s?

kolibri/core/assets/src/utils/browser.js
```diff
@@ -37,4 +38,10 @@ export function isEmbeddedWebView() {
...
+  const hasWebViewCookie = Boolean(cookiejs.get('app_key_cookie'));
```
jamalex: Rather than just the cookie being defined, I think checking that it's valid would be ideal.

Is that something we could add to the session endpoint? Have it include a boolean attribute like `secureAppContext`? Then we don't need to do anything at all with the cookie in the frontend (can just use the same function in the backend as we do in the permission check), and that boolean should be available to the frontend through the session data, right?

kolibri/core/assets/src/utils/browser.js
jamalex: Although I think I'd prefer not having this constraint anyway -- if the token is secure, this doesn't add security, and just makes remote debugging harder.

kolibri/core/assets/src/utils/browser.js
```diff
@@ -37,4 +38,10 @@ export function isEmbeddedWebView() {
...
+  return isEmbeddedWebView() && isLocal && hasWebViewCookie;
```
jamalex: Similarly to the localhost note above, I think it'd probably be best not having this be contingent upon the `isEmbeddedWebView` directly, as that will make debugging things like the click-to-login stuff easier.

jamalex: (there's a distinction between functionality that only cares about being in an app-level secure context, e.g. the click-to-login, and functionality that is only possible at all in the app context itself, e.g. the whatsapp sharing -- the latter can check both `secureAppContext` and `isAndroidWebView`)

jamalex: (although even the latter wouldn't need to be hidden in the UI based on isAndroidWebView -- what really matters for that is whether the _server_ is running in an Android app service context with the intent-based sharing capability)

kolibri/plugins/app/api.py
```diff
@@ -0,0 +1,77 @@
...
+from kolibri.plugins.app.utils import LAUNCH_INTENT
+
+
+class FromSameDevicePermission(BasePermission):
+    """
+    Allow only users on the same device as the server
+    """
+
+    def has_permission(self, request, view):
+        return request.META.get("REMOTE_ADDR") == "127.0.0.1"
```
jamalex: If we are going to keep a "local only" constraint, do we need to add "localhost" here too? Or does it always resolve the client address into an IP?

kolibri/plugins/app/utils.py
```diff
@@ -0,0 +1,42 @@
...
+    @property
+    def capabilities(self):
+        if self.enabled:
+            return {key: (key in self._capabilities) for key in CAPABILITES}
+        return {key: False for key in CAPABILITES}
+
+    def launch_intent(self, filename, message):
+        if LAUNCH_INTENT not in self._capabilities:
+            raise NotImplementedError(
+                "Launch intents is not supported on this platform"
```
jamalex: *is -> are
or
*Launch -> Launching

kolibri/plugins/app/api.py
```diff
@@ -0,0 +1,77 @@
...
+class FromAppViewPermission(BasePermission):
+    def has_permission(self, request, view):
+        return request.COOKIES.get(APP_KEY_COOKIE_NAME) == DeviceAppKey.get_app_key()
+
+
+class AppCommandsViewset(ViewSet):
+
+    permission_classes = (FromSameDevicePermission, FromAppViewPermission)
+
+    if LAUNCH_INTENT in interface:
```
jamalex: Conditional method existence :exploding_head:

kolibri/core/assets/src/utils/browser.js
jamalex: Seems to me like there are 3 distinct things, which can factor in differently for each feature:
1. Is it a secure app context?
2. Are we inside a webview?
3. Does the capability exist?

In production, we'd never have 1 without 2, so the main reason for allowing 1 without 2 is for development/debugging.

kolibri/plugins/learn/assets/src/views/ContentPage.vue
```diff
@@ -328,6 +327,20 @@
           params: { id },
         };
       },
+      launchIntent() {
+        return client.post({
+          path: urls['kolibri:kolibri.plugins.app:appcommands-launch_intent'](),
```
jamalex: I feel like maybe these URLs should somehow "go through" (be generated by) something more centralized in the "capabilities" system, rather than being somewhat "hard-coded" here. Just trying to think about how to keep the cross-plugin interfaces as clean/robust as possible. Just flagging in case you might have ideas for that.

kolibri/plugins/app/api.py
```diff
@@ -0,0 +1,77 @@
...
+        if app_key != token:
+            raise PermissionDenied("You have provided an invalid token")
+        redirect_url = request.GET.get("next", "/")
+        # Copied and modified from https://github.com/django/django/blob/stable/1.11.x/django/views/i18n.py#L40
+        if (redirect_url or not request.is_ajax()) and not is_safe_url(
+            url=redirect_url,
+            allowed_hosts={request.get_host()},
+            require_https=request.is_secure(),
+        ):
+            redirect_url = request.META.get("HTTP_REFERER")
```
jamalex, line 53: For our current purposes, don't think we need the HTTP_REFERER fallback, but I guess it doesn't hurt.

jamalex (changes requested):
A few suggestions -- can chat to brainstorm if useful!

kolibri/plugins/learn/assets/src/views/ContentPage.vue
```diff
@@ -141,7 +139,10 @@
   import { ContentNodeKinds } from 'kolibri.coreVue.vuex.constants';
   import CoachContentLabel from 'kolibri.coreVue.components.CoachContentLabel';
   import DownloadButton from 'kolibri.coreVue.components.DownloadButton';
-  import { isEmbeddedWebView, isAndroidWebView } from 'kolibri.utils.browser';
+  import client from 'kolibri.client';
+  import urls from 'kolibri.urls';
+  import { isEmbeddedWebView, isAppView } from 'kolibri.utils.browser';
+  import { launchIntent } from 'kolibri.utils.appCapabilities';
```
jamalex: To the note above about what happens if a capability isn't defined, I'm assuming this would error if it's not defined?

kolibri/core/assets/src/utils/appCapabilities.js
rtibbles: Mostly because there wasn't a clear need for it to be pluggable, so it seemed to make it more complicated.

In this case because we are exporting an object, attempting to check an undefined capability would return a falsy value.

kolibri/core/assets/src/utils/browser.js
rtibbles: Sure, can remove - I had initially added this before the shibboleth existence checking.

kolibri/core/assets/src/utils/browser.js
rtibbles: Yeah, can add it to the session certainly - in theory could then also prevent the cookie being read on the client side so that it is less leakable.

kolibri/core/assets/src/utils/browser.js
rtibbles: Sure, can remove the extra condition!

kolibri/plugins/app/api.py
rtibbles: It always resolves to the IP, yup.

kolibri/plugins/app/utils.py
rtibbles: or "Launch intents" :) will update

kolibri/plugins/app/api.py
rtibbles: Don't even expose the endpoint if you don't need to. SECURE!

kolibri/plugins/learn/assets/src/views/ContentPage.vue
rtibbles: Yeah, can make a helper function for this.

kolibri/plugins/learn/assets/src/views/ContentPage.vue
rtibbles: It would just be 'undefined'.

kolibri/core/assets/src/utils/appCapabilities.js
```diff
@@ -0,0 +1,16 @@
+import urls from 'kolibri.urls';
+import plugin_data from 'plugin_data';
+
+const appCapabilities = plugin_data.appCapabilities || {};
+
+export const can = {
+  launchIntent: appCapabilities.launch_intent,
```
indirectlylit: is this capability really for generic launching of intents?

I would have expected one of:

* `share_file`
* `launch_android_intent`

or something more specific

kolibri/core/assets/src/utils/appCapabilities.js
```diff
@@ -0,0 +1,16 @@
...
+};
+
+export const actionUrls = {
+  get launchIntent() {
+    return urls[`kolibri:kolibri.plugins.app:appcommands-launch_intent`]
+      ? urls[`kolibri:kolibri.plugins.app:appcommands-launch_intent`]()
+      : null;
+  },
+};
```
indirectlylit: This is fine for now, but I don't know that `actionUrls` is quite the interface I would have expected `appCapabilities` to expose. I might have expected more of a set of Vuex actions or just plain functions that get called. For example, maybe something like:



```js
export const capabilityHandlers = {
  shareFile(filename, message) {
    const url = urls[`kolibri:kolibri.plugins.app:appcommands-share_file`];
    if (!url) {
      logger.warn("not implemented };
      return;
    }
    return client.post({
      path: url,
      method: 'POST',
      params: { filename, message },
    });
};
```

kolibri/core/assets/src/utils/appCapabilities.js
jamalex: Ok, yeah, as long as it's got "graceful degradation" then that seems fine.

kolibri/plugins/app/api.py
jamalex: If we're removing the check on the frontend, we should remove it here too, and just be shibboleth-based.

kolibri/core/assets/src/utils/appCapabilities.js
jamalex: Yeah, there could be other sorts of intents at some point, so my vote would be for something like "share_file".

kolibri/core/assets/src/utils/appCapabilities.js
jamalex: Yeah, exposing functions that then handle things internally seems like a nice, clean interface (minimal surface area, for easy migration paths).

kolibri/core/assets/src/utils/browser.js
```diff
@@ -39,9 +39,9 @@ export function isEmbeddedWebView() {
 }
 
 export function isAppView() {
```
jamalex: I think renaming this to `isAppContext` as well, for consistency, would be good.

kolibri/core/device/utils.py
```diff
@@ -75,3 +77,22 @@ def provision_device(**kwargs):
     device_settings, _ = DeviceSettings.objects.get_or_create(defaults=kwargs)
     device_settings.is_provisioned = True
     device_settings.save()
+
+
+def valid_app_key(app_key):
+    from .models import DeviceAppKey
```
jamalex, line 83: Does this file get imported somewhere pre-app-loading, or why do these imports need to be local?

kolibri/core/device/utils.py
```diff
@@ -75,3 +77,22 @@ def provision_device(**kwargs):
...
+    return APP_KEY_COOKIE_NAME in request.COOKIES and valid_app_key(
+        request.COOKIES.get(APP_KEY_COOKIE_NAME)
+    )
+
+
+def set_app_key_on_response(response):
+    from .models import DeviceAppKey
+
+    response.set_cookie(APP_KEY_COOKIE_NAME, DeviceAppKey.get_app_key())
+    return response
```
jamalex: Based on the function name, might be cleaner if it did the operation to the response but didn't return the response.

kolibri/core/assets/src/utils/appCapabilities.js
rtibbles: Fine by me, I was following the name of the function in the app-wrapper, so let's update that there too.

kolibri/core/assets/src/utils/appCapabilities.js
rtibbles: Sure, fine by me. Will update.

kolibri/core/device/utils.py
rtibbles: To prevent circular imports.

kolibri/core/assets/src/core-app/apiSpec.js
```diff
@@ -101,6 +101,7 @@ import CatchErrors from '../utils/CatchErrors';
 import UiIconButton from '../views/KeenUiIconButton.vue';
 import UiToolbar from '../views/KeenUiToolbar.vue';
 import shuffled from '../utils/shuffled';
+import appCapabilities from '../utils/appCapabilities';
```
jredrejo, line 104: appCapabilities is not exporting a `default` function or const, thus this is not working

jredrejo (dismissed):

kolibri/core/assets/src/core-app/apiSpec.js
rtibbles: Ah yep, I changed that in response to review comments. Am reverting back to having a default based on review above.

kolibri/core/device/utils.py
rtibbles: Updated.

kolibri/core/assets/src/core-app/apiSpec.js
rtibbles: Sorry, this should now be fixed!

kolibri/core/assets/src/utils/browser.js
rtibbles: Done

rtibbles:
Have addressed all review comments and manually tested as far as is possible.

Have confirmed that:
* The appContext getter is returning `true` when you have previously visited the initialize endpoint with the  correct shibboleth
* The app still loads and functions correctly.

Cannot confirm that the `shareFile` API is working as intended, as it requires it to have been registered on the backend during runtime.

kolibri/core/assets/src/utils/appCapabilities.js
jamalex: I had like 7 different names in there for different functions trying different approaches to the sharing until I finally got one to work reliably. Will be migrating it into the app plugin anyway, and will give it consistent naming there.

jamalex (approved):

kolibri/plugins/app/api.py
```diff
@@ -32,25 +23,23 @@ def has_permission(self, request, view):
 
 class AppCommandsViewset(ViewSet):
 
-    permission_classes = (FromSameDevicePermission, FromAppViewPermission)
+    permission_classes = (FromAppViewPermission,)
```
jamalex: Should we also make this `FromAppContextPermission`, for full consistency?

jamalex: (or just `AppContextPermission`)

rtibbles: Sure

PR #6835, opened 2020-04-30
integrate PF functionality into develop

indirectlylit:
### Summary

* resolves conflicts merging the `pf-mvp` branch into `develop`
  * KIcon.vue
  * FacilityIndex.vue
  * DataPage/index.vue
  * facility/assets/src/routes.js
  * deployment/default/settings/base.py
  * tasks/api.py
  * cli.py
* remove pf-specific functionality

### Reviewer guidance

everyone on the reviewer list has changes that were affected. Please check parts of the code relating to your work

### References

https://github.com/learningequality/kolibri/pull/6667

----

### Contributor Checklist

PR process:

- [ ] PR has the correct target branch and milestone
- [ ] PR has 'needs review' or 'work-in-progress' label
- [ ] If PR is ready for review, a reviewer has been added. (Don't use 'Assignees')
- [ ] If this is an important user-facing change, PR or related issue has a 'changelog' label
- [ ] If this includes an internal dependency change, a link to the diff is provided

Testing:

- [ ] Contributor has fully tested the PR manually
- [ ] If there are any front-end changes, before/after screenshots are included
- [ ] Critical user journeys are covered by Gherkin stories
- [ ] Critical and brittle code paths are covered by unit tests

### Reviewer Checklist

- Automated test coverage is satisfactory
- PR is fully functional
- PR has been tested for [accessibility regressions](http://kolibri-dev.readthedocs.io/en/develop/manual_testing.html#accessibility-a11y-testing)
- External dependency files were updated if necessary (`yarn` and `pip`)
- Documentation is updated
- Contributor is in AUTHORS.md

PR #6865, opened 2020-05-11
CONTENT_FALLBACK_DIRS option along with cherrypy/devserver support.

jamalex:
### Summary

For Flatpak-based preloading of Endless content, as well as future SD card-based static content distribution for Android, this PR adds a `CONTENT_FALLBACK_DIRS` option that allows for additional search paths to be specified beyond the standard, primary `CONTENT_DIR`.

Content is only downloaded to the primary content location, but a file is found in one of the fallback directories instead, that path will be used by the server, and the file will be considered available.

`CONTENT_FALLBACK_DIRS` is a semicolon-separated list of paths. Those paths should each contain subfolders called "storage" and "databases", in the same form as the `<KOLIBRI_HOME>/content` directory (or other `CONTENT_DIR` path).

### Reviewer guidance

Run Kolibri with the environment variable `CONTENT_FALLBACK_DIRS` set to a path (or semi-colon-delimited list of paths) containing Kolibri content. Note that you then need to run `kolibri manage scanforcontent` to pick up any new content.

Note: will not currently work properly with kolibri-server or behind any of our standard nginx proxy configs, as those have the /static prefix hardcoded to a single lookup path.

### TODO

- [ ] Add automated tests for primary expected behavior
- [ ] Test with nginx and see if it's possible to provide easy nginx compatibility guidance
- [x] Test various interactions with content scanning and availability updating (e.g. update `get_channel_ids_for_content_database_dir` and references to it)

### Contributor Checklist

PR process:

Testing:

- [ ] **Critical and brittle code paths are covered by unit tests**

### Reviewer Checklist

- PR is fully functional
- Documentation is updated


kolibri/utils/options.py
```diff
@@ -239,7 +244,7 @@ def language_list(value):
             "type": "integer",
             "default": 2,
             "envvars": ("KOLIBRI_PICKLE_PROTOCOL",),
-        },
```
rtibbles, line 242: I don't think our linter enforces trailing commas, but seems an odd deletion.

kolibri/utils/options.py
```diff
@@ -198,7 +198,12 @@ def language_list(value):
             "type": "string",
             "default": "content",
             "envvars": ("KOLIBRI_CONTENT_DIR",),
-        }
+        },
+        "CONTENT_FALLBACK_DIRS": {
+            "type": "string",
```
rtibbles: We should do proper validation here so that this setting returns a tuple or list - this will also allow us to validate that the paths passed to us are valid. See `language_list` above for an example on how to do this.

kolibri/utils/server.py
```diff
@@ -227,16 +275,15 @@ def configure_http_server(port):
...
-        section="/", dir=paths.get_content_dir_path()
-    )
-
-    url_path_prefix = conf.OPTIONS["Deployment"]["URL_PATH_PREFIX"]
-
+    CONTENT_ROOT = "/" + paths.get_content_url(
+        conf.OPTIONS["Deployment"]["URL_PATH_PREFIX"]
+    ).lstrip("/")
+    content_dirs = [paths.get_content_dir_path()] + paths.get_content_fallback_paths()
+    dispatcher = MultiStaticDispatcher(content_dirs)
```
rtibbles, line 285: In the case where the fallback paths are not used, how much overhead does the multistatic dispatcher add compared to how cherrypy normally handles static files?

One way to be conservative about this would be to only use the MultiStaticDispatcher in the case where there are fallback content paths.

rtibbles (dismissed):
Validation can be improved - one flag about potential impact on performance.

kolibri/utils/options.py
jamalex: Seems there's [some debate](https://github.com/psf/black/issues/274) in the python-black project on trailing commas in singleton dicts. I'm guessing my version of black is from before the March release when this behavior was finally changed...

kolibri/utils/server.py
jamalex: It's still using the same cherrypy static handlers. The additional code that's executed per file request is:
https://github.com/learningequality/kolibri/pull/6865/files/78ae3149620cc3e98743ca87968e6557529b95c0#diff-601cde52caf17fa9d77c64ca9e73e927R239-R248

In the case of no fallback content paths, this loop will only be executed once, but it does do an extra `os.path.exists` call in there, which wouldn't be needed in the "single handler" case. Actually, the simplest way to "handle" that might be to just short-circuit the `find_handler` method if there's only one candidate handler anyway.

rtibbles:
Should we target this to the `app-support` branch, btw?

kollivier:
I took this for a spin locally and the fallback paths are working for me with a fallback path on an external hard drive. The main issue was that even with smaller channels, scanforcontent took a couple minutes to run on a fairly fast machine. 

@jamalex For cases where we know in advance what content will be in the fallback dir(s), can we extend the preseeded Kolibri trick to run scanforcontent so it can scan in the channels and load them into the Kolibri db beforehand? (I'm assuming the scanforcontent time hit is mostly due to a need to import things from the channel db into a Kolibri installation db?)

indirectlylit:
> Should we target this to the app-support branch, btw?

The "click to log in" and password-free changes for Android will be going in `app-support`. However I believe the Endless timeline might be a bit accelerated compared to that branch.

If `app-support` is currently stable enough for Endless to use and we are waiting on the Android changes, I would say **yes**: this change should go there and Endless should release builds from that branch.

Otherwise we can merge here and have the MrPau team do an integration test pass.

In either case, we'll need Gherkin stories for these changes in order for them to test it eventually.

kollivier:
I appear to have gotten this working, along with porting the preseeded Kolibri stuff to Mac so I can test Kolibri while keeping all the big channels on my external drive. :) Have to say that just loading right into a Kolibri with plenty of content after initial setup is a nice experience.

jredrejo:
For `kolibri-server`, the `CONTENT_FALLBACK_DIRS` can be removed/inserted in every restart of the server with a modification here. https://github.com/learningequality/kolibri-server/blob/master/kolibri_server_setup.py#L159
I'll work on it after this is approved

kolibri/utils/options.py
jamalex: Cool, thx -- done [here](https://github.com/learningequality/kolibri/pull/6865/commits/dd39641184627fd5b5bc7da9297c3071a6953097).

I didn't want to enforce path string formats using a regex or whatnot, because that can be hairy, from what I read (with Unicode, etc). Nor do I want to enforce that they already exist...

jamalex:
@kollivier:

>For cases where we know in advance what content will be in the fallback dir(s), can we extend the preseeded Kolibri trick to run scanforcontent so it can scan in the channels and load them into the Kolibri db beforehand?

We could, in principle. Except I can't imagine too many scenarios where we would know specifically what content would be there at install time. Except for full OS images, and in that case the simplest would be to start Kolibri, stop Kolibri, and then deprovision (with data left in place).

>(I'm assuming the scanforcontent time hit is mostly due to a need to import things from the channel db into a Kolibri installation db?)

It's hard to say. That's likely to be a decent portion of it (and if the external drive is non-SSD and/or over slower USB, then that could be a bottleneck for the DB reads). The other part would be annotation, if you have a lot of new content available in the fallback dirs. But for that, it shouldn't be much slower than if the content were on your primary disk/content path. Could you time the stages based on log outputs when you run `scanforcontent`?

jamalex:
> In either case, we'll need Gherkin stories for these changes in order for them to test it eventually.

Opened a tech debt issue here:
https://github.com/learningequality/kolibri/issues/6904

kolibri/utils/options.py
rtibbles: Ah, it was the latter that I was more interested in - but the idea here is to allow paths to come into being without a restart of Kolibri?

jamalex:
Build failure looks to be a result of the same thing as https://github.com/learningequality/kolibri/pull/6835#issuecomment-624350987

kolibri/utils/options.py
jamalex: The idea is that, for example: if a removable drive isn't currently plugged in/loaded yet, it doesn't block Kolibri from starting. Or that there could be a list of "potential" paths in here, but we shouldn't blow up just because one doesn't exist (just like PATH, PYTHONPATH, etc, are tolerant to nonexistent paths).

PR #6870, opened 2020-05-12
Don't use diskcache when we are using Redis

rtibbles:
### Summary
* Vendor diskcache RLock to use django cache transparently.
* Don't recreate diskcache at import.
* Only recreate diskcache when completely necessary.

### Reviewer guidance
Does kolibri still start?
Does importing a channel database still work?

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/core/tasks/utils.py
+++ b/kolibri/core/tasks/utils.py
@@ -1,12 +1,16 @@
 import importlib
 import logging
+import os
 import time
 import uuid
 
-from diskcache import RLock
+try:
+    from thread import get_ident
+except ImportError:
+    from threading import get_ident
 
 from kolibri.core.tasks import compat
-from kolibri.deployment.default.cache import diskcache_cache
+from kolibri.core.utils.cache import get_process_cache
 
 
 # An object on which to store data about the current job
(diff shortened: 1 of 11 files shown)


kolibri/core/tasks/utils.py
```diff
@@ -122,12 +126,52 @@ def shutdown(self):
...
+        pid_tid = "{}-{}".format(pid, tid)
+
+        value, count = self._cache.get(self._key, default=(None, 0))
+        is_owned = pid_tid == value and count > 0
+        assert is_owned, "cannot release un-acquired lock"
+        self._cache.set(self._key, (value, count - 1), self._expire)
+
         # RLOCK leaves the db connection open after releasing the lock
         # Let's ensure it's correctly closed
         self._cache.close()
```
jredrejo, line 168: If I understand this correctly, the lock is a key entry into the Django cache, that would be usually either redis or memory. If that's the case we don't need to close this , as there's not database involved.

kolibri/deployment/default/cache.py
```diff
@@ -16,19 +16,21 @@
...
+                diskcache_location, disk_pickle_protocol=pickle_protocol
+            )
+        except DatabaseError:
+            shutil.rmtree(diskcache_location, ignore_errors=True)
+            os.mkdir(diskcache_location)
+            diskcache_cache = Cache(
+                diskcache_location, disk_pickle_protocol=pickle_protocol
+            )
+        diskcache_cache.clear()
+        diskcache_cache.close()
```
jredrejo, line 32: same as the previous question, not needed if we are not writing on the /.kolibri/process_cache/cache.db database.

kolibri/deployment/default/cache.py
```diff
@@ -16,19 +16,21 @@
...
-    diskcache_cache.close()
-    return diskcache_cache
-
-
-try:
-    diskcache_cache = Cache(diskcache_location, disk_pickle_protocol=pickle_protocol)
-    diskcache_cache.close()
-except DatabaseError:
-    diskcache_cache = recreate_cache()
+def recreate_diskcache():
```
jredrejo, line 19: Do we really need this? are we using `.kolibri/process_cache/cache.db` anywhere?

jredrejo (commented):
Code looks good, however, testing it I still have some doubts on the usage of `.kolibri/process_cache/cache.db`. 
I have ran the code with no issues, so after the `cache.db` usage is cleared it looks good to me.
However a good bunch of tests (48) are broken with these changes.

kolibri/deployment/default/cache.py
rtibbles: Yes - it is being used as the `"process_cache"` in Kolibri, in the case when we are not using Redis. Without this, then the RLock would not work cross process without Redis.

jredrejo: I see the use of the database  in `CrossProcessCache` being used as a cache. 

However, for RLock, we're using `db_task_write_lock` and this one does not need the database after the changes this PR does.

Do we still need `CrossProcessCache` ? can't we use the `default` cache for the same task?

rtibbles: If Redis is active, we are just using the default cache, but without Redis, the default cache is the InMemCache, which is not cross process. We want our database write lock to be active across processes, so this line: https://github.com/learningequality/kolibri/pull/6870/files#diff-f4bf89eeeed59f411ac09bca78e9e5dcR6 returns the cross process cache if it exists (i.e. when Redis is not available) and otherwise falls back to the default cache.

jredrejo: Yes,  that makes totally sense. We should have either `redis` or `DiskCache` in CrossProcessCache for cache that need to be shared between processes.

But my doubts come for the fact that I have double checked that caches["process_cache"] is using InMemCache when running, for example, `kolibri manage importchannel network c401c87e3cf652128fd3efaa5b3774f2`

That's probably the reason why so many tests are failing. Something is not working properly and this code is not using DiskCache for process_cache

rtibbles: Ah, ok, I must have messed something up somewhere. Will fix locally!

rtibbles: Note that the issue above was due to the fact we overwrite all caches in the dev settings to LocMemCache - I have updated both the test and dev settings to preserve the process_cache when it is present, so that this is less confusing for testing and development.

jredrejo (approved):
After the latest commit it looks good to me. I'd only double check if the `self._cache.close() `sentences are still needed as I mentioned in my comments. I think they are not needed anymore with this PR, but they don't produce any damage anyway.

PR #6923, opened 2020-05-25
Facility settings updates for app suport

jredrejo:
### Summary
Changes to the facility settings for app support:
image
image

Change password does not appear, depending on the previous settings:
image

If "`Require password for learners`" is not set, password input fields are not shown when self-creating an account, and the user creation works:
image

Also, in the form where the admin creates the accounts, if the account is for a Learner, the password is hidden depending on the previous settings:
image

Doubts:

1. When creating an user without a password, the database will have a "NOT_SPECIFIED" value in the pasword field, This must be taken into account in some of the login scenarios @nucleogenesis 
image

2. Password input fields are not shown when self-creating an account, but they are still shown in the page where the admin can create an account,  @riddhiavlani  should be removed from there too?

### Reviewer guidance
Check the application behaves accordingly to the design
Check tests pass

### References
https://docs.google.com/presentation/d/1yCURJOW2azPsnxxxQrcXEKbzTj8CxF7HCu8FdVnp6UE/edit
Slides 14-29

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated


riddhiavlani:
@jredrejo, yes you are right. The password input field shouldn't appear for the admin/super admin while creating learner accounts if the 'password free login' has been enabled. Thanks for pointing that out! cc @jtamiace @indirectlylit

kolibri/plugins/user/assets/src/views/SignUpPage.vue
```diff
@@ -30,15 +30,16 @@
...
-            ref="passwordTextbox"
-            autocomplete="new-password"
-            :value.sync="password"
-            :isValid.sync="passwordValid"
-            :shouldValidate="formSubmitted"
-            :disabled="busy"
-          />
+          <template v-if="showPasswordInput">
+            <PasswordTextbox
+              ref="passwordTextbox"
```
indirectlylit, line 35: Why wrap in a `template`? `v-if` can be applied directly to `PasswordTextbox`

kolibri/plugins/user/assets/src/views/SignUpPage.vue
```diff
@@ -189,6 +190,10 @@
         // query is before hash
         return getUrlParameter('next');
       },
+      showPasswordInput() {
+        this.setEmptyPassword();
```
indirectlylit, line 194: `computed` props are meant to be read-only. It's not good to call a `method` which modifies `data` from within one.

A more conventional strategy would be to define a new computed prop like:

```js
passwordToSave() {
  if (this.facilityConfig.learner_can_login_with_no_password && this.password === '') {
    return 'NOT_SPECIFIED'
  }
  return this.password;
},
```

That said, I don't actually think the front-end should be setting the `NOT_SPECIFIED` string at all. I would expect this to be handled completely by the backend, because the string needs to bypass Django's normal hashing and salting mechanisms.

kolibri/plugins/facility/assets/src/views/UserCreatePage.vue
```diff
@@ -185,10 +186,29 @@
...
+        this.showPasswordInput = !this.facilityConfig.learner_can_login_with_no_password;
+        this.updatePasswordValues();
+      });
     },
     methods: {
+      ...mapActions(['getFacilityConfig']),
+      updatePasswordValues() {
+        if (!this.showPasswordInput) {
+          this.passwordValid = true;
+          if (this.password === '') this.password = 'NOT_SPECIFIED';
```
indirectlylit, line 209: will this work as expected? I would imagine if we're setting this string client-side, it's going to get hashed and salted by Django which is not the intended behavior

indirectlylit (commented):
I may merge in to get the strings uploaded, but I have some concerns about the implementation noted inline below

indirectlylit:
merging in for string freeze

kolibri/plugins/user/assets/src/views/SignUpPage.vue
jredrejo: > `computed` props are meant to be read-only. It's not good to call a `method` which modifies `data` from within one.
> 
> A more conventional strategy would be to define a new computed prop like:
> 
> ```js
> passwordToSave() {
>   if (this.facilityConfig.learner_can_login_with_no_password && this.password === '') {
>     return 'NOT_SPECIFIED'
>   }
>   return this.password;
> },
> ```
> 

Totally agree, thank you. I will change it.


> That said, I don't actually think the front-end should be setting the `NOT_SPECIFIED` string at all. I would expect this to be handled completely by the backend, because the string needs to bypass Django's normal hashing and salting mechanisms.

The front-end must send something, either'`NOT_SPECIFIED`, an empty string, or another keyword. It's a question of deciding what we want to do, and it's an open question I left in the PR description to be decided between the people involved in this. We need a convention for the backend to know that this is an user without password.
Said that: the backend will work fine in the part where it  manages the self-creation of the user. However, when handling the form where the admin creates users there's a pre-validation in the backend that requires a non-empty password. Code will be simpler and will require less changes using any keyword like `NOT_SPECIFIED` than sending an empty string.

kolibri/plugins/facility/assets/src/views/UserCreatePage.vue
jredrejo: no, if you look at the backend code changes, when it receives `NOT_SPECIFIED` as password, it's not hashed, but written in plain text in the db

PR #6929, opened 2020-05-27
Remove docker-clean recipe.

DXCanas:
### Summary

*These changes are intended to be propagated all the way up into current develop*.

With the introduction of a share-nothing multi-machine "cluster" of build agents, we can no longer assume that the cleanup step will run on the same machine that built the prior steps - the jobs regularly alternate between builders.

Further, a job (buildkite step) isn't really meant for that kind of cleanup anyway. It assumes that the source code is necessary, and pulls the entire repo. This is actually a pretty time consuming step when the network is bogged down due to our use of `git-lfs`.

For now, I've enabled an system-level hook to run some docker cleanup commands (look for the step in buildkite logs. Punctuated by a docker emoji) on every single job, on both our current agents. Even for the jobs that don't use Docker. This is heavy handed, but doesn't take very long. 

Conditionalizing the cleanup to only run when necessary is  possible, and I've locally drafted some ways to do that, but will require many commits/PR's across several repos. Most likely, the cleanup will become part of the pipelines' codebase so that cleanup can be customized per-build. This will probably happen eventually, but I don't think it's a priority right now.

### Reviewer guidance

Make sure the cleanup job is running in logs. I've verified that it's doing what's expected, but there are some pretty bad consequences down the road if it's not running.

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -1,7 +1,6 @@
 steps:
   - label: Build Python packages
     command:
-     - make docker-clean
      - mkdir -p dist
      - .buildkite/build_whl.sh
 
@@ -22,9 +21,6 @@ steps:
     if: build.tag != null
 
   - wait
-
-  - label: Cleaning up
-    command: make docker-clean
 
   - block: "Test .debs?"
   - label: Test on Trusty, Xenial, Bionic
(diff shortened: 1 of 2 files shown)


DXCanas:
The buildkite issues in this PR are the same ones that have been inconsistently plaguing the kolibri build for 0.10 - sometimes, for whatever reason, yarn packages don't resolve the way they're expected to.

No idea what to do about the Travis errors, though. @rtibbles

rtibbles:
Looks like 0.11 was before we upgraded to NodeJS 10 - so I think that's probably the cause here.

DXCanas:
@rtibbles How hard is the fix? Worth it to resolve here?

rtibbles:
Could upgrade to node 10 in theory, but I imagine there was more to it than that when we did it. So I'd rather we only did it if we needed to release.

DXCanas:
Seems this break is outside the scope of this PR - we can fix it down the line if necessary. For now, I'll merge. If subsequent versions break, I'll fix the break here and escalate the update with fix back up to newer versions.

PR #6955, opened 2020-05-29
Remote access device settings changes

jredrejo:
### Summary
A new setting to be used when the application is executed inside an app that will allow the user to disable access from other browsers to Kolibri.
image

This setting will be taken into account only when kolibri runs as an app and will be False by default, so other browsers won't be able to login into Kolibri unless the user allows it.

Changes in the login screen:
image

This combination of settings:
image
makes this screen appear when someone uses a browser to see content resources:
image

This PR also moves the Django REST `api-auth`/ urls to the dev urls so they can't be used in the standard appliction to avoid their use as a backdoor.

### Reviewer guidance
Do these screenshots match the design?
Using the browser, can you access to the server contents somehow?

### References
It uses #6864 to detect if Kolibri is in an app context.

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated


radinamatic:
Maybe we should not hide the warnings about potential security concerns and reveal them only on hover? :thinking:

jredrejo:
> Maybe we should not hide the warnings about potential security concerns and reveal them only on hover?

Well, this is only for the Android app and will be disabled by default. To enable it you will see the warning. So this design looks logical to me in order to avoid converting every option in a paragraph.
In any case, @riddhiavlani and @jtamiace can decide if they want to change it. From a dev perspective, it does not involve any technical problem changing it.

radinamatic:
>  in order to avoid converting every option in a paragraph

I can understand this concern, but stacking a security warning in something that requires a user to hover does not sound like a good practice. 

Would it be possible to display the warning as a paragraph below **ONLY** when the user checks/enables this option?

I also have questions regarding the wording of both the checkbox label, and the alert message:

1. Is there a difference between _from a browser_ and _over the network_? When this option is checked, those _others_ will have to use browsers in the local network to access the Kolibri app on that Android device, correct? Or is there another way to connect to it that we need to warn about? 

    I suggest the following rephrase: 
`Allow others in the network to access Kolibri on this device using a browser`

2. _Password free login_ is not the terminology we use in Kolibri UI. I suggest the following rephrase:

   `If learners are allowed to sign in with no password on this device, enabling this may be a potential security concern.`

cc @indirectlylit

jredrejo:
@radinamatic can you move your suggestions about the strings to the audit doc https://docs.google.com/document/d/1Us8qJxX1on_rEdE2pYqxWGhQezAy545xE_jVF330MKA/edit?usp=drivesdk 
I will implement the decissions taken there.
Thanks!

indirectlylit (approved):
looks like we have a little more iteration on the strings to do from the google doc, but this is looking good.

only potential issue i see is that now i'm blocked from logging in! 

image

jredrejo:
@riddhiavlani  I have updated the code with the changes in the strings, screenshots have been updated too.

radinamatic:
> only potential issue i see is that now i'm blocked from logging in!

Welcome to the club! :joy: 

José has a recipe for this :wink:

jredrejo:
> > only potential issue i see is that now i'm blocked from logging in!
> 
> Welcome to the club! 
> 
> José has a recipe for this

There are two ways, maybe for you the easiest is editing the database and resetting the `allow_other_browsers_to_connect` value in the `device_devicesettings` table. This is the fish: it will work until you set it again..
The other method requires setting the cookie that makes your browser think it's in an app webview, it's a bit longer, but that's the fishing and will work forever.

PR #6956, opened 2020-05-29
Convert keys from UUIDs to strings when channel stats are created, not accessed.

rtibbles:
### Summary
Importing content on postgresql can lead to timeouts because we are coercing all the keys of the channel_stats dict from a UUID to a string for each content node in a request.

This PR solves this by instead coercing the id of each individual content node when the channel stats are created.

### Reviewer guidance
Question - should this be rebased and targeted to 0.13.x instead?

### References
Follow up from #6544

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/core/content/serializers.py
+++ b/kolibri/core/content/serializers.py
@@ -11,7 +11,6 @@ from kolibri.core.content.models import ContentNode
 from kolibri.core.content.models import File
 from kolibri.core.content.models import Language
 from kolibri.core.fields import create_timezonestamp
-from uuid import UUID
 
 
 class DynamicFieldsModelSerializer(serializers.ModelSerializer):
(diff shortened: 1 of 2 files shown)


kolibri/core/content/serializers.py
```diff
@@ -438,24 +445,13 @@ class Meta:
...
-            channel_stats = stats_convert_id
-        return channel_stats
+        return self.context["channel_stats"]
 
     def get_total_resources(self, instance):
         # channel_stats is None for export
         if self.channel_stats is None:
             return instance.on_device_resources
-        return self.channel_stats.get(instance.id, {"total_resources": 0})[
+        return self.channel_stats.get(coerce_key(instance.id), {"total_resources": 0})[
```
lyw07: I tested the PR but one thing I found is that the error indicated in https://github.com/learningequality/kolibri/pull/6544 comes back because in `channel_stats` the ids are stored as UUIDs when connecting to postgresql. But `instance.id` is a string, and `coerce_key(instance.id)` is then also a string.

since we don't want to iterate over `channel_stats` to check that, maybe we can get one item from `channel_stats` to see whether the ids stored inside are UUIDs or strings and then make changes based on that result? - just my two cents though...

please feel free to let me know if there's any debug information you need. thank you!

lyw07 (approved):
Tested, and it works well! Thank you for fixing this issue! 🎉

PR #6969, opened 2020-06-01
Drop rest js. Use Axios.

rtibbles:
### Summary
Drops use of the no longer supported rest js.
Uses Axios instead.
Migrates all interceptors and other rest js specific code to Axios.

### Reviewer guidance
Does everything still work as expected - particularly places where the kolibri core client is used directly?

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated


jonboiser:
I found a couple of places that might be still using the old rest.js API. e.g.

https://github.com/rtibbles/kolibri/blob/c3f1921ea27f350d74cea2884efcf8bfd8f317b2/kolibri/core/assets/src/utils/appCapabilities.js#L36-L42 (uses `path` param instead of `url`)

Other files:
* https://github.com/rtibbles/kolibri/blob/c3f1921ea27f350d74cea2884efcf8bfd8f317b2/kolibri/plugins/device/assets/src/modules/deviceInfo/handlers.js#L13-L15
* https://github.com/rtibbles/kolibri/blob/c3f1921ea27f350d74cea2884efcf8bfd8f317b2/kolibri/plugins/device/assets/src/modules/deviceInfo/index.js#L25-L27

rtibbles:
Will update - also worth noting that I added code to allow the path attribute, and change it to url when invoked, so nothing should currently break, but maybe I should also add a deprecation warning so that we can catch anywhere else that it is still being used.

rtibbles:
Removed them and added a deprecation warning to the `path` backwards compatibility piece.

kolibri/plugins/device/assets/src/modules/deviceInfo/index.js
```diff
@@ -23,7 +23,7 @@ export default {
   actions: {
     updateDeviceName(store, name) {
       return client({
-        path: urls['kolibri:core:devicename'](),
+        url: urls['kolibri:core:devicename'](),
         method: 'PATCH',
```
jonboiser, line 27: On the axios docs, they use lower case for the method values. Does this still work and is axios case-sensitive?

```suggestion
        method: 'patch',
```

jonboiser: I think it might be nicer to use the methods axios provides just to make it a little more transparent e.g. `client.patch(...)`

rtibbles: It doesn't appear to be case sensitive - all the requests I tested were upper case, and all worked.

I would like us to eventually transition to just using the Axios methods directly, but wanted to limit the possibility of breakage first, then do a more aggressive refactor later.

kolibri/core/package.json
```diff
@@ -19,7 +20,6 @@
     "lodash": "^4.17.4",
     "loglevel": "^1.4.1",
     "markdown-it": "^8.3.1",
-    "rest": "^1.3.2",
```
jonboiser, line 22: `js-cookie` dependency now also looks unused, and the `kolibri-core-for-export/package.json` needs to be updated as well.

jonboiser:
More 'rest'-looking API stuff here

https://github.com/learningequality/kolibri/blob/c3f1921ea27f350d74cea2884efcf8bfd8f317b2/kolibri/core/assets/src/views/AppError/index.vue#L94-L96 (trying to access `status.code` from the Error object in vuex instead of `error.response.status`

kolibri/core/assets/test/heartbeat.spec.js
```diff
@@ -1,3 +1,4 @@
+import mock from 'xhr-mock';
```
jonboiser, line 1: I'd be interested in using these library here to mock the `Resource` class instead of what I've done with the `JestMockResource` thing. I think it's an improvement because it would actually move the test coverage of the tests using the mocks down a little further down to the DRF layer, esp. if we use more specific URLs). If possible, it might be an improvement to actually use the relevant URLs for the heartbeat API instead of what I think is the a catch-all `/.*`

kolibri/plugins/coach/assets/src/modules/pluginModule.js
```diff
@@ -77,11 +77,7 @@ export default {
...
       logging.error(errorObject);
-      if (
-        errorObject.status &&
-        errorObject.status.code &&
-        authErrorCodes.includes(errorObject.status.code)
-      ) {
+      if (errorObject.response.status && authErrorCodes.includes(errorObject.response.status)) {
         store.dispatch('handleApiError', '');
       } else {
         store.dispatch('handleApiError', errorObject);
```
jonboiser, line 83: We need to update `AppError.vue` to handle the new shape of `errorObject`

kolibri/core/assets/src/views/CoreBase/index.vue
```diff
@@ -314,7 +314,7 @@
       },
       notAuthorized() {
         // catch "not authorized" error, display AuthMessage
-        if (this.error && this.error.status && this.error.status.code == 403) {
```
jonboiser, line 317: When going to an invalid resource ID on Learn, `this.error` isn't an object, but a string `"Error: Request failed with status code 404"`

jonboiser (changes requested):
Some changes needed:

1. Fix the 405 errors on login
1. Update the package.json files to remove `js-cookie`, and make the 'kolibri-for-export'.. and core versions of package.json consistent
1. Fix issue of 404s on Learn page Content not being objects, but strings
1. Update `AppError.vue` to reflect new shape of error objects.

kolibri/core/assets/test/heartbeat.spec.js
rtibbles: Yeah, I think that would be good overall - we could probably do a minor refactor on your `JestMockResource` class to just use the xhr-mock under the hood.

kolibri/plugins/coach/assets/src/modules/pluginModule.js
rtibbles: Good catch, thanks.

kolibri/core/assets/src/views/CoreBase/index.vue
rtibbles: Seems that this was a result of Axios errors actually being Javascript Error objects, so it was just getting stringified rather than JSONified.

rtibbles:
All issues addressed.

> make the 'kolibri-for-export'.. and core versions of package.json consistent

Noting that the main reason for doing this is to drop dependencies from yarn.lock - it only normally gets committed when the package gets released and updated.

jonboiser (approved):

indirectlylit:
I'm seeing lots of these warnings lately. Could it be related to this change?

```
client.js:105 [WARN: kolibri/core/assets/src/core-app/client.js] entity is deprecated for accessing response data, please use the data key instead
```

rtibbles:
Yes - I have noticed this in my recent work on the device page too - I think the task polling code needs to be updated.

rtibbles:
In good news, at least it's just showing deprecation warnings and isn't broken!

indirectlylit:
mind opening a follow-up issue?

PR #6973, opened 2020-06-02
Initial implementation of action oriented logging API.

rtibbles:
### Summary
Opening for feedback. Implements first pass based on this proposal: https://docs.google.com/document/d/12NQkizkzSQvX4tNMV3oSdqtyzrF00GHU8nNd5birYZU/edit

### Reviewer guidance
Feedback on the division of endpoints, and the API of each endpoint.

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/core/logger/api.py
+++ b/kolibri/core/logger/api.py
@@ -1,6 +1,7 @@
 import logging
 
 from django.core.exceptions import ObjectDoesNotExist
+from django.db import transaction
 from django.db.models.query import F
 from django.db.utils import IntegrityError
 from django.http import Http404
@@ -8,9 +9,12 @@ from django_filters import ModelChoiceFilter
 from django_filters.rest_framework import CharFilter
 from django_filters.rest_framework import DjangoFilterBackend
 from django_filters.rest_framework import FilterSet
+from le_utils.constants import exercises
 from rest_framework import filters
+from rest_framework import serializers
 from rest_framework import status
 from rest_framework import viewsets
+from rest_framework.decorators import action
 from rest_framework.exceptions import ValidationError
 from rest_framework.response import Response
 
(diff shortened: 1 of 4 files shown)


kolibri/core/logger/api.py
```diff
@@ -39,11 +43,374 @@
...
+            return output
+
+    @logger_action
+    def update_attempt(self, request, content_id, context):
+        serializer = UpdateAttemptSerializer(data=request.data)
+        serializer.is_valid(raise_exception=True)
+        attempt_id = serializer.validated_data["attempt_id"]
+        end_timestamp = serializer.validated_data["end_timestamp"]
+
+        complete = serializer.validated_data["complete"]
```
indirectlylit, line 383: Even though these "actions" are not strictly RESTful from a DB-server perspective, it might be worth exploring making the API a bit more RESTful from a server-client perspective?

For example, we might expose an "attempt" resource, which can take either `POST` to create a new one, or `PUT`/`PATCH` to update it. Under the hood the creation also needs to update some existing rows in certain tables, but abstractly it's still creating a new attempt list, initialized with the first attempt.

In other words, the HTTP API could expose a RESTful logger API with resources that don't strictly map to any particular DB table.

indirectlylit (commented):
I was wondering whether this API will apply to new assessment types that don't conform to our existing mastery models and perseus exercises?

Also wondering if this might be an opportunity to make our names a bit more explicit? I've always found the names "attempt", "session", and "summary" to be vague and confusing.

kolibri/core/logger/api.py
rtibbles: Yes, they could definitely be two restful resources here: `session` and `attempt` with a create and update method for each. I was unclear whether that added any particular value to the API or not.

rtibbles:
`summary` doesn't get exposed here at all, so I don't think we need to worry about that.
`session` - what alternatives would you suggest? I just took a quick look through SCORM, and the only similar thing I found was... `session` (used in the key `session_time`

For `attempt` I could imagine us using `itemattempt` or `iteminteraction`?

indirectlylit:
simply adding descriptors would help distinguish from other similar concepts, such as a "user session" or an "attempt at taking a quiz".

So maybe:

* `content_interaction_session`
* `exercise_item_attempt`

kolibri/core/logger/api.py
indirectlylit: When you call this an "action oriented" API, is that basically the same thing as an RPC API, or is there another nuance there?

I think there are some clear advantages of making it RESTful, and also a few disadvantages:

Pros:

* HTTP verbs are well-defined
* Caching and idempotency rules can be easier to reason about
* xAPI is RESTful

Cons:

* Harder to design because we need define new abstract resources
* SCORM is not RESTful

rtibbles:
Going to close this for now, as I think the final implementation will be impacted by current planning work.

indirectlylit:
Sounds good. I think conceptually we mainly need to revisit 'context'?

rtibbles:
Yeah, I think it's possible to reuse most of this code, but maybe with a little bit of a refactor, so I shan't be throwing this branch away.

Context - and also the use of the four distinct actions, I think we can probably lighten the ontology.

PR #6974, opened 2020-06-02
Add back and update .mo/.po files and other i18n utilities

jonboiser:
Here's what I did:

1. Restored the entire `./kolibri/locale` directory from `indirectlylit/i18n-pf`, then discarded all the modifications to JSON files. Effectively, this added back the po files.
1. Cherry-picked the "copy over django.po..." commit from that same branch
1. Ran `make i18n-update`, which generated the `.mo` files and updated the noto CSS files


indirectlylit (approved):
installer works well

PR #7001, opened 2020-06-06
Scorm hashi bugs

rtibbles:
### Summary
Fixes two small issues that I noticed while using hashi after the SCORM updates:
* When SCORM data is undefined, the SCORM shim will error because it is trying to access an undefined key path
* An option had an extra space inside it, meaning it would not pass validation

### Reviewer guidance
* Check that non-SCORM html5 apps do not give an error trying to access a property on undefined as currently happens

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/packages/hashi/src/SCORM.js
+++ b/packages/hashi/src/SCORM.js
@@ -613,7 +613,7 @@ export default class SCORM extends BaseShim {
   }
 
   __calculateProgress() {
-    const score = this.data.cmi.core.score;
+    const score = getByKeyPath(this.data, 'cmi.core.score', SCHEMA, self.userData);
     if (score) {
       // If min and max are not set, raw will be a value in the range 0-100. Source:
       // https://support.scorm.com/hc/en-us/articles/206166466-cmi-score-raw-whole-numbers-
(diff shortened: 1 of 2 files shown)


packages/hashi/src/SCORM.js
```diff
@@ -622,8 +622,8 @@ export default class SCORM extends BaseShim {
       const raw = Number(isNaN(score.raw) ? min : score.raw);
       return Math.max(Math.min((raw - min) / (max - min), 1), 0);
     }
-    const lessonStatus = this.data.cmi.core.lesson_status;
-    if (statusProgressMap.hasOwnProperty(lessonStatus)) {
+    const lessonStatus = getByKeyPath(this.data, 'cmi.core.lesson_status', SCHEMA, self.userData);
+    if (Object.prototype.hasOwnProperty.call(statusProgressMap, lessonStatus)) {
```
kollivier, line 626: Just curious about why this change was made. e.g. how `Object.prototype.hasOwnProperty.call` differs in behavior from `statusProgressMap.hasOwnProperty`. Is this to handle the case where `lessonStatus` is undefined? If so, any reason not to check it for undefined first?

kollivier (approved):
Good catch, tested and the problem appears to be fixed!

kollivier:
Also, since pf-mvp changes have been merged to develop, we need to make sure we apply this fix there as well.

packages/hashi/src/SCORM.js
indirectlylit: This was flagged by our linter which has this rule: https://eslint.org/docs/rules/no-prototype-builtins

I made the same change the last time I merged `pf-mvp` -> `develop`

rtibbles: Yeah - I originally made this change on develop, so @indirectlylit's linting fix was caught up in my stash. It seemed simpler to retain this change than to undo it when I reapplied it to pf-mvp.

PR #7015, opened 2020-06-09
Bootstrap remote access data into page

### LINKED ISSUE #6998: 'remote access' permission should be bootstrapped, not loaded via API
### Observed behavior

reported by @rtibbles - we currently load this setting async:

```js
export function getRemoteAccessPermission(store) {
  return client({
    method: 'POST',
    path: urls['kolibri:kolibri.plugins.device:allowremoteaccess'](),
  }).then(response => {
    const data = response.entity;
    store.commit('SET_REMOTE_BROWSER_PERMISSION', data.allowed);
  });
}
```

This causes the 'access not allowed' message to briefly display:

### Expected behavior

setting should be bootstrapped into the page using `plugin_data`, e.g.:

```python
    @property
    def plugin_data(self):
        return {
            "appCapabilities": interface.capabilities,
            "contentCacheKey": ContentCacheKey.get_cache_key(),
            "languageGlobals": self.language_globals(),
            "oidcProviderEnabled": OIDCProviderHook.is_enabled(),
            "kolibriTheme": ThemeHook.get_theme(),
        }
```

### User-facing consequences

confusing load behavior

### Context

`app-support` branch

### LINKED ISSUE #7012: default 'network access' setting value
### Observed behavior

After upgrading, all access is blocked:

image

### Expected behavior

The value should:

* default not to restrict for upgrades from 0.13.x
* default not to restrict for non-app context
* default to restrict for app-contexts

### User-facing consequences

users are locked out of Kolibri after upgrading

### Steps to reproduce

run 0.13, then switch to `app-support` branch on the same DB

### Context

0.14

rtibbles:
### Summary
* Remove API based bootstrapping.
* Set default conditional on app plugin being enabled.
* Prevent login outside of app-context when remote access is enabled

### Reviewer guidance
Does a non-app context properly get blocked by the remote access setting?

Should we change the wording of remote access to make clear this is for non-app access? If not, we should probably change the restrictions to be based on IP address (127.0.0.1 rather than app-context).

### References
Fixes #6998, resolves #7012 

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated


kolibri/core/auth/api.py
```diff
@@ -494,6 +495,14 @@ def create(self, request):
         password = request.data.get("password", "")
         facility_id = request.data.get("facility", None)
 
+        if not allow_other_browsers_to_connect() and not valid_app_key_on_request(
```
jredrejo: Just a simple optimization for the cases where the app plugin is not enabled:
```
+from kolibri.core.device.models import app_is_enabled
 
 
 class KolibriAuthPermissionsFilter(filters.BaseFilterBackend):
@@ -495,7 +496,7 @@ class SessionViewSet(viewsets.ViewSet):
         password = request.data.get("password", "")
         facility_id = request.data.get("facility", None)
 
-        if not allow_other_browsers_to_connect() and not valid_app_key_on_request(
+        if app_is_enabled() and not allow_other_browsers_to_connect() and not valid_app_key_on_request(

```

jredrejo (approved):
I have just left a small optimization that could be applied to the code. Appart from t hat, everything is fine from my point of view.
My only concern is that we don't have tests for this  behaviour and it's complex enough to deserve them (big problems to the users can appear if it does not work properly). If you don't add them here, please, create an issue so we don't forget them.

Thanks.

kolibri/core/auth/api.py
rtibbles: Yep, I had also just noticed this - important for making sure this setting is a noop for non-app Kolibris!

rtibbles: Updated!

rtibbles:
Follow up issue here: https://github.com/learningequality/kolibri/issues/7019

PR #7027, opened 2020-06-10
More work on Device > Facility page + Setup Wizard + misc fixes

jonboiser:
1. Adds WIP facility import/discovery backends on the Setup Wizard
1. Integrates Device > Facility Page to existing KDP Sync Backend
1. Various fixes


kolibri/core/discovery/api.py
```diff
@@ -20,3 +22,18 @@ class DynamicNetworkLocationViewSet(NetworkLocationViewSet):
...
+
+
+class NetworkLocationFacilitiesView(viewsets.GenericViewSet):
+    """
+    Given a NetworkLocation ID, returns a list of Facilities that are on
+    that NetworkLocation, for the purposes of syncing
+    """
+
+    def retrieve(self, request, pk=None):
+        base_url = "http://192.168.1.8:8000/"
```
jredrejo, line 34: Using a hardcoded ip is correct?

jonboiser: No, this endpoint is still being worked on. It will use the `DynamicNetworkLocation` model's `baseurl` field to construct the `/api/public/v1/facility` endpoint for a specific peer.

This hardcoded URL is just my second computer's IP on the LAN, which I'm using for testing.

PR #7029, opened 2020-06-11
Add method to get initialize url.

rtibbles:
### Summary
* Adds get_initialize_url method to app interface API
* Has optional argument for next url
* Returns the URL relative to the host (protocol, host name and port must be supplied)

### Reviewer guidance
Can you generate the URL?

Note that I did not add the protocol hostname, and port, because the port can be dynamically changed by the `--port` argument, so it is best left to the invoker.

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/plugins/app/utils.py
+++ b/kolibri/plugins/app/utils.py
@@ -1,3 +1,5 @@
+from django.core.urlresolvers import reverse
+
 from kolibri.plugins.app.kolibri_plugin import App
 from kolibri.plugins.registry import registered_plugins
 
@@ -20,6 +22,19 @@ class AppInterface(object):
         for capability in CAPABILITES:
             if capability in kwargs:
                 self._capabilities[capability] = kwargs[capability]
+
+    def get_initialize_url(self, next_url=None):
+        if not self.enabled:
+            raise RuntimeError("App plugin is not enabled")
+        # Import here to prevent a circular import
+        from kolibri.core.device.models import DeviceAppKey
+
+        url = reverse(
+            "kolibri:kolibri.plugins.app:initialize", args=(DeviceAppKey.get_app_key(),)
+        )
+        if next_url is None:
+            return url
+        return url + "?next={}".format(next_url)
 
     @property
     def enabled(self):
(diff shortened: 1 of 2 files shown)


indirectlylit:
what functionality is this in support of, and how should I test it?

rtibbles:
Basically, it's just a single place for the Android (or any other app) wrapper to get the URL that it should initially navigate to in order to load Kolibri.

The URL is optional, in case it has saved state of the last URL visited and wants to restore that state after a restart.

rtibbles:
To test it - I would just check that giving it a URL like `/ar/learn/#/channels/` gives a URL that when you load it first goes to initialize and then redirects you to the URL you supplied.

kolibri/plugins/app/utils.py
```diff
@@ -21,6 +23,19 @@ def register(self, **kwargs):
...
                 self._capabilities[capability] = kwargs[capability]
 
+    def get_initialize_url(self, next_url=None):
+        if not self.enabled:
+            raise RuntimeError("App plugin is not enabled")
+        # Import here to prevent a circular import
+        from kolibri.core.device.models import DeviceAppKey
+
+        url = reverse(
+            "kolibri:kolibri.plugins.app:initialize", args=(DeviceAppKey.get_app_key(),)
```
kollivier, line 33: `get_app_key()` returns a UUID type, but `reverse` is expecting a string, so we need to convert when passing as an arg. I tried to push a fix for this myself but got a repo authentication error.

kollivier (changes requested):
PR works great with the aforementioned conversion, once that is in I'll approve this for merging, thanks! :)

kolibri/plugins/app/utils.py
rtibbles: Oh - that's odd. It should really always return a string. Will look into it.

rtibbles:
Got a weird test environment error, so have rebased too for good measure.

kollivier (approved):
LGTM!

PR #7033, opened 2020-06-11
Facility deletion management command

bjester:
### Summary

- Adds a management command for facility deletion to remove a facility entirely from the device
- Adds a `facilitytask` endpoint for initiating a facility deletion task
- Explicitly added the `on_delete=CASCADE` to the models since it in future versions of Django, it requires an explicit action it instead of assuming cascade
- Updated `@action` decorators with preferred alternative since it is deprecated

### Reviewer guidance

A nice to have feature was to have the command report progress. Relying on CASCADE deletions works just fine for most of the models, but also doesn't give much insight into that progress, as far as I know. This currently goes through all the models, counts them, and then deletes them; through this procedure it can report progress since it's handling the models individually.

Some feedback on whether we care about any of that would be great. It was a nice to have that didn't seem too difficult, but could get stale in the future if adding more related models, although still completely deleting the facility because of the cascade.

### References

[Notion page](https://www.notion.so/learningequality/Backend-Admins-can-delete-a-Facility-and-all-its-data-from-a-Device-13-pts-0-14-dev-fac-UC6-2277c6a5cf3940ecaf822f3e0d25bb38)

----

### Contributor Checklist

PR process:

Testing:

- [ ] ~If there are any front-end changes, before/after screenshots are included~

### Reviewer Checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/core/auth/management/commands/deletefacility.py
+++ b/kolibri/core/auth/management/commands/deletefacility.py
@@ -1,0 +1,341 @@
+import logging
+from contextlib import contextmanager
+
+from django.contrib.admin.models import LogEntry
+from django.core.exceptions import FieldError
+from django.core.management.base import CommandError
+from django.db import transaction
+from django.db.models import Q
+from morango.models import Buffer
+from morango.models import Certificate
+from morango.models import RecordMaxCounter
+from morango.models import RecordMaxCounterBuffer
+from morango.models import Store
+from morango.models import SyncSession
+from morango.models import TransferSession
+from morango.registry import syncable_models
+
+from kolibri.core.analytics.models import PingbackNotificationDismissed
+from kolibri.core.auth.constants.morango_sync import PROFILE_FACILITY_DATA
+from kolibri.core.auth.management.utils import DisablePostDeleteSignal
+from kolibri.core.auth.management.utils import get_facility
+from kolibri.core.auth.models import AdHocGroup
+from kolibri.core.auth.models import Classroom
+from kolibri.core.auth.models import Collection
+from kolibri.core.auth.models import FacilityDataset
+from kolibri.core.auth.models import FacilityUser
+from kolibri.core.auth.models import LearnerGroup
+from kolibri.core.auth.models import Membership
+from kolibri.core.auth.models import Role
+from kolibri.core.auth.utils import confirm_or_exit
+from kolibri.core.device.models import DevicePermissions
+from kolibri.core.exams.models import Exam
+from kolibri.core.exams.models import ExamAssignment
+from kolibri.core.lessons.models import Lesson
+from kolibri.core.lessons.models import LessonAssignment
+from kolibri.core.logger.models import AttemptLog
+from kolibri.core.logger.models import ContentSessionLog
+from kolibri.core.logger.models import ContentSummaryLog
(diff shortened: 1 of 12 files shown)


jonboiser (approved):

kolibri/core/auth/management/utils.py
```diff
@@ -22,6 +23,21 @@
...
+        post_delete.receivers = []
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        post_delete.receivers = self.receivers
+        self.receivers = None
+
+
 def _interactive_client_facility_selection():
     facilities = Facility.objects.all().order_by("name")
     message = "Please choose a facility to sync:\n"
```
radinamatic, line 43: @bjester Is the use of `sync` here intentional? This command should delete the facility, so why _sync_...? :thinking:

bjester: @radinamatic Oh good catch! I didn't notice it said "sync". I'll make that dynamic

bjester: Updated here https://github.com/learningequality/kolibri/pull/7187/commits/a41de91333175f594d4bf2d7ce1adaf0a7ed7bd5

kolibri/core/auth/management/commands/deletefacility.py
```diff
@@ -0,0 +1,341 @@
...
+
+    :type things: list
+    :type size: int
+    """
+    for i in range(0, len(things), size):
+        yield things[i : i + size]
+
+
+class Command(AsyncCommand):
+    help = "This command initiates the deletion process for a facility and all of it's related data."
```
jamalex, line 138: *its

kolibri/core/auth/management/commands/deletefacility.py
```diff
@@ -0,0 +1,341 @@
...
+        if not noninteractive:
+            # ensure the user REALLY wants to do this!
+            confirm_or_exit(
+                "Are you sure you wish to permanently delete this facility? This will DELETE ALL DATA FOR THE FACILITY."
+            )
+            confirm_or_exit(
+                "ARE YOU SURE? If you do this, there is no way to recover the facility data on this device."
+            )
+
+        # everything should get cascade deleted from the facility, but we'll check anyway
```
jamalex, line 175: Sounds good. The main thing that I could imagine might be less likely to cascade delete would be anonymous ContentSessionLogs, which don't FK onto a user. But they would still have an FK onto the FacilityDataset (as would all of these), so it _should_ all get cascade deleted, yeah.

kolibri/core/auth/management/commands/deletefacility.py
```diff
@@ -0,0 +1,341 @@
...
+            total_deleted = 0
+
+            # run the counting step
+            with self.start_progress(
+                total=delete_group.group_count()
+            ) as update_progress:
+                update_progress(increment=0, message="Counting database objects")
+                total_count = delete_group.count(update_progress)
+
+            # no the deleting step
```
jamalex, line 203: *now?

jamalex (commented):
Sorry for not reviewing before -- looks great! Just added a couple comment typo notes and a random comment. Nothing needing action.

Issue #6785, opened 2020-04-14
etag not accounting for plugin_data


nucleogenesis:
### Observed behavior

When we generate etags we are not accounting for plugin_data that is bootstrapped in `kolibri_plugin.py` files. 

etag generation happens here: https://github.com/learningequality/kolibri/blob/3a6bdb2aeeb6c7360558cee4ecd93ff93dab8274/kolibri/core/decorators.py#L319

### Expected behavior

When data necessary to front end rendering (ie, plugin_data) changes, then the etag for that resource should be generated anew.

### User-facing consequences

Changes not being reflected due to incorrect HTTP 304 responses are received when the plugin_data has changed but is not generating a new etag accordingly.

### Errors and logs

…

### Steps to reproduce

…

### Context

…

rtibbles:
Fixed in #7107

PR #7107, opened 2020-06-19
Cache etags 

nucleogenesis:
### Summary

After further discussion with @rtibbles this PR has changed to be removing the etag decorator we were looking at altogether.

-- below this is the original comment --

Updated the etag generation decorator function so that it also uses the rendered HTML which includes everything in `plugin_data`.

Is there a better or easier way to get `plugin_data` only?

I hate that I call the view_func twice and feel like that could be a problem - I'll come back to it.

### Reviewer guidance

- Load a page and check the root request URL in your devtools network tab.
- Check its etag 
- Go to the `kolibri_plugin.py` file for the plugin associated with the URL you're testing. Add a key to the `plugin_data`
- Reload the page and see that the same resource's etag has changed
- Reload again and see that it has not changed

### References

https://github.com/learningequality/kolibri/issues/6785

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/core/decorators.py
+++ b/kolibri/core/decorators.py
@@ -345,7 +345,7 @@ def cache_no_user_data(view_func):
         setattr(_response, "response", None)
 
         request = args[0]
-        etag = cache.get(request.path)
+        etag = cache.get(CACHE_KEY_TEMPLATE.format(request.path))
 
         # Doing this here - will also be the same in inner_func
         # required to delete the session for this to work as expected
--- a/kolibri/core/decorators.py
+++ b/kolibri/core/decorators.py
@@ -329,6 +329,7 @@ def cache_no_user_data(view_func):
     """
 
     CACHE_TIMEOUT = 15
+    CACHE_KEY_TEMPLATE = "SPA_ETAG_CACHE_{}"
     _response = local()
 
     def render_and_cache(response, cache_key):
@@ -353,7 +354,7 @@ def cache_no_user_data(view_func):
         if not etag:
             response = view_func(*args, **kwargs)
             setattr(_response, "response", response)
-            etag = render_and_cache(response, request.path)
+            etag = render_and_cache(response, CACHE_KEY_TEMPLATE.format(request.path))
         return etag
 
     @etag(calculate_spa_etag)
@@ -364,7 +365,7 @@ def cache_no_user_data(view_func):
         if not response:
             response = view_func(*args, **kwargs)
 
-        render_and_cache(response, request.path)
+        render_and_cache(response, CACHE_KEY_TEMPLATE.format(request.path))
         patch_response_headers(response, cache_timeout=CACHE_TIMEOUT)
         response["Vary"] = "accept-encoding, accept"
         return response
--- a/kolibri/core/decorators.py
+++ b/kolibri/core/decorators.py
@@ -7,7 +7,9 @@ from __future__ import unicode_literals
 
 import hashlib
 import sys
+from threading import local
 
+from django.core.cache import cache
 from django.utils.cache import patch_response_headers
 from django.views.decorators.http import etag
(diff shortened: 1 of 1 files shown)


nucleogenesis:
I'll add that I've gone through a bit and not seen that calling `view_func` twice on the request is causing any issues. I was concerned that it was going to duplicate SQL queries by calling whatever method/function is handling the view rendering.

kolibri/core/decorators.py
```diff
@@ -336,10 +328,20 @@ def cache_no_user_data(view_func):
     on a per user basis.
     """
 
-    @etag(calculate_spa_etag)
-    def inner_func(*args, **kwargs):
+    def calculate_spa_etag(*args, **kwargs):
```
rtibbles, line 343: My main concern with this approach is that if we are just rendering the entire page anyway, we may as well dispense with the etag and just return the page.

That said, I think that generating the etag from the page contents is definitely the most definitive measure.

At this point, we should take a step back and think about what we are aiming to achieve by the caching of the HTML pages here.

* By caching HTML pages, we reduce server load in two ways:
1. Once a page has been generated, for the next _CACHE_TIMEOUT_ seconds the server can respond with the cached page
2. Once a client has received a page, for the next _CACHE_TIMEOUT_ seconds the client will access the cached copy from its browser cache. When the cache has expired, then an etag can be used. With this solution, the etag no longer saves any server resources.

* We have deliberately set the _CACHE_TIMEOUT_ for these pages to a very low value (15s), this means that most of the value is coming from the server side caching, and the etag. The reason for this low value is to allow for changes in the state used to render the page to propagate quickly to clients (i.e. if the content cache key or theme changes).

We could probably recover some of the benefit of the etag here, by caching the etag value (I know this sounds a little bit crazy) - but this would probably only help if we did a pre-emptive recalculation strategy. i.e. if the etag is requested, but the cache timeout on the etag has not expired yet (but may expire soon) then we would kick off a background recalculation of the etag, rather than do it synchronously.

nucleogenesis:
As @rtibbles pointed out - the original solution renders the view anyway. Ultimately Richard's suggestion is to remove the etag and to address optimization along these lines separately.

So this PR has changed to be removing the etag altogether.

indirectlylit:
What is the effect of `CACHE_TIMEOUT`?

1. throttle to prevent regenerating etags more often than `CACHE_TIMEOUT`, even if DB has changed
2. etag is regenerated on every request, but it gets thrown out after `CACHE_TIMEOUT`


Just noting that even if etags are regenerated on every request and render the full page, they still save network bandwidth by returning a `304 Not Modified` with no body

kolibri/core/decorators.py
```diff
@@ -336,12 +328,44 @@ def cache_no_user_data(view_func):
...
+        ).hexdigest()
+        cache.set(cache_key, etag, CACHE_TIMEOUT)
+        return etag
+
+    def calculate_spa_etag(*args, **kwargs):
+        # Clear the local thread 'response' property
+        setattr(_response, "response", None)
+
+        request = args[0]
+        etag = cache.get(request.path)
```
rtibbles: Thought of one more thing in review:

I don't think it is huge likely, but might be good to prefix the request path with something to prevent accidental cache key collisions if something else is using the path:

So we can add this in the local scope:
`CACHE_KEY_TEMPLATE = "SPA_ETAG_CACHE_{}"`

Then to generate the cache key:
`cache_key = CACHE_KEY_TEMPLATE.format(request.path)`

rtibbles: Missed using the cache key template here :)

rtibbles (approved):
Code looks good - have manually tested and confirmed that I get a 304, even if I wait > 15 secs, so the regeneration of the etag is working properly.

PR #7109, opened 2020-06-22
latest strings

indirectlylit:
latest i18n strings


jonboiser (changes requested):
Discard the `.mo` file deletions 😄 . The `acg_UG` and `ff_CM` `.mo` files are showing a 100% deletion in this diff.

indirectlylit:
Good catch thanks.

This was done with just running the standard scripts, so need to investigate why this happened. Are you aware of this happening previously?

jonboiser:
Some kind of `mo`-deletion has happened on many of the recent i18n updates (adding new languages/messages/etc). My theory has been that these mo files get deleted as part of a cleanup step that might be shared between different `i18n` `make` commands.

indirectlylit:
@jonboiser as far as I can tell, these files are still there and not deleted. For example:

The diff for `ff_CM` shows:

image

I checked the file sizes before and after:

* 1,865 bytes before change
* 1,860 bytes after change

I believe the "100%" might actually be a binary file similarity measure [as described here](https://stackoverflow.com/questions/1046276/what-does-the-message-rewrite-90-after-a-git-commit-mean/1585059) but having trouble confirming this locally

indirectlylit:
Note, pushed more strings so the specific values have changed

jonboiser (approved):

Issue #7127, opened 2020-06-22
issues with facility import functionality


indirectlylit:
### Observed behavior

The import process has been stuck on 99% for about 10 minutes:

![image](https://user-images.githubusercontent.com/2367265/85341047-23fe7d00-b49c-11ea-8c28-c7401af854b0.png)

During that time I got a few dozen warnings like this:

```
WARNING  Validation error for FacilityUser with id 6572fd4d5e34e2efcb482a1737e376f7: {'password': ['This field cannot be blank.']}
```

A few dozen like this:

```
WARNING  Validation error for UserSessionLog with id 2aafbc857703fed2e431de94a113b0a7: ['FacilityUser matching query does not exist.']
```

And about 100,000 entries like this:

```
WARNING  Validation error for ContentSummaryLog with id 22714685a88c63c4e8c2f12fa7de7343: ['FacilityUser matching query does not exist.']
```

![2020-06-22 15 26 02](https://user-images.githubusercontent.com/2367265/85341855-1518ca00-b49e-11ea-82bb-2dc4e6b08976.gif)

Then, it hung at 100% for a couple minutes, while strangely saying step "4 of 7":

![image](https://user-images.githubusercontent.com/2367265/85342091-9d976a80-b49e-11ea-9822-1c6bf18292ae.png)

Waited longer and it did eventually finish:

![image](https://user-images.githubusercontent.com/2367265/85341524-32996400-b49d-11ea-87a8-22df71302071.png)

After this, I tried logging out and then back in using an admin from the newly-imported facility. This seemed to work, but when I went to the Coach page, I saw this:

![image](https://user-images.githubusercontent.com/2367265/85341657-8d32c000-b49d-11ea-8201-c79babfbcd71.png)

### Expected behavior

1. progress states and step numbers during import better-reflects the time remaining. If necessary we should switch to "indeterminate" progress bar and remove the percentage number.
2. after import, able to access coach reports

### User-facing consequences

significant frustration using data syncing

### Errors and logs

none apart from those shown above

### Steps to reproduce

import `kolibri-beta` demo facility to a local Kolibri instance

### Context

0.14.0 beta 3

bjester:
I believe we'll need to make the sync an operation that can't be canceled. We've started seeing messages like this when it's been cancelled before.

bjester:
[...]

PR #7187, opened 2020-06-25
Facility command and task improvements

bjester:
### Summary

- Adds new `NamespacedCacheProxy` to enable efficient clearing of cache keys related to Facility dataset data
- Clears namespaced dataset cache after deleting a facility and before syncing
- Updates the default chunk size for facility tasks to be `200` and makes them not cancellable
- Updates `sync` command to not aggregate progress tracking total
- Updates frontend to make progress indeterminate for certain facility task stages
- Updates `morango` to `0.5.1` which has [deserialization improvements](https://github.com/learningequality/morango/pull/84)

### Reviewer guidance

- From the CLI:
  1) Set up another Kolibri with some _real_ facility data
  2) On a second Kolibri instance, with fresh database, issue the `kolibri manage sync` command pulling from the first
  3) Verify facility data in Kolibri
  4) On the second instance, issue the `kolibri manage deletefacility` command deleting the facility
  5) On a second instance, issue the `kolibri manage sync` command pulling again to pull the same facility
  6) No warnings or errors should occur
- From the frontend:
  1) Initiate a sync with the Kolibri bet server
  2) Verify all facility import/sync tasks show indeterminate progress when not sending or receiving data

### References

https://github.com/learningequality/morango/pull/84
https://github.com/learningequality/kolibri/issues/7127

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/core/utils/cache.py
+++ b/kolibri/core/utils/cache.py
@@ -28,8 +28,8 @@ class DiskCacheRLock(object):
     https://github.com/grantjenks/python-diskcache/blob/2d1f43ea2be4c82a430d245de6260c3e18059ba1/diskcache/recipes.py
     """
 
-    def __init__(self, cache, key, expire=None):
-        self._cache = cache
+    def __init__(self, key, expire=None):
+        self._cache = process_cache
         self._key = key
         self._expire = expire
 
@@ -82,9 +82,7 @@ class NamespacedCacheProxy(BaseCache):
         params.update(KEY_PREFIX=namespace)
         super(NamespacedCacheProxy, self).__init__(params)
         self.cache = cache
-        self._lock = DiskCacheRLock(
-            process_cache, "namespaced_cache_{}".format(namespace)
-        )
+        self._lock = DiskCacheRLock("namespaced_cache_{}".format(namespace))
 
     def _get_keys(self):
         """
--- a/kolibri/core/utils/cache.py
+++ b/kolibri/core/utils/cache.py
@@ -1,6 +1,8 @@
 from django.core.cache import caches
 from django.core.cache import InvalidCacheBackendError
+from django.core.cache.backends.base import BaseCache
 from django.utils.functional import SimpleLazyObject
+from django.utils.synch import RWLock
 
 
 def __get_process_cache():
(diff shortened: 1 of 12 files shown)


kolibri/core/utils/cache.py
```diff
@@ -11,3 +13,85 @@ def __get_process_cache():
 
 
 process_cache = SimpleLazyObject(__get_process_cache)
+_namespace_locks = {}
+
+
+class NamespacedCacheProxy(BaseCache):
```
bjester, line 71: @rtibbles I created this proxy class to generically handle tracking cache keys and clearing them.

kolibri/core/utils/cache.py
```diff
@@ -11,3 +13,85 @@ def __get_process_cache():
...
+        """
+        :rtype: list
+        """
+        return self.cache.get("__KEYS__", default=[])
+
+    def _set_keys(self, keys):
+        """
+        :type keys: list
+        """
+        self.cache.set("__KEYS__", keys)
```
rtibbles: Shouldn't the `__KEYS__` be prefixed here by the namespace? Otherwise if we reuse this class elsewhere, the different caches will clear each other's keys?

kolibri/core/utils/cache.py
```diff
@@ -11,3 +13,85 @@ def __get_process_cache():
...
+
+    def __init__(self, cache, namespace, **params):
+        """
+        :type cache: BaseCache
+        :type namespace: str
+        """
+        params.update(KEY_PREFIX=namespace)
+        super(NamespacedCacheProxy, self).__init__(params)
+        self.cache = cache
+        self._lock = _namespace_locks.setdefault(namespace, RWLock())
```
rtibbles: These objects are process local - if this is running in a job on e.g. Kolibri-server, then this would be separate from main server process (which is running in uwsgi), and when we finally get iceqube to properly use multiprocessing, then multiple jobs would not respect each other's locks here.

Would this be a problem for the kind of locking you are trying to do here?

rtibbles (commented):
Two potential issues with the new namespaced cache class.

kolibri/core/utils/cache.py
bjester: Oh good catch

kolibri/core/utils/cache.py
bjester: Ah yeah that's true, so I should use the disk lock class?

rtibbles: Yeah, probably best to use that, using the `process_cache`, as with the `db_write_lock`:
https://github.com/learningequality/kolibri/blob/develop/kolibri/core/tasks/utils.py#L175

rtibbles: (this way it uses Redis in Redis deploys, but otherwise falls back to disk cache)

Issue #7003, opened 2020-06-07
Links to Content on Linux Fail in 0.13.0, 0.13.2, 0.13.3


nobicycle:
### Observed behavior
$ sudo -u kolibri ls -l /opt/kolibri/content
lrwxrwxrwx 1 root root 40 Jun  8 01:33 /opt/kolibri/content -> /mnt/System/Applications/Kolibri/content

$ sudo -u kolibri ls /opt/kolibri/content
databases  storage

No content is visible in the browser.
In the past, after stopping Kolibri, creating the link and starting Kolibri, ... Kolibri would re-establish its database and all would work fine. No Longer.
…

### Expected behavior
Content visible in the browser
…

### Errors and logs
[kolibri.txt](https://github.com/learningequality/kolibri/files/4742623/kolibri.txt)
[daemon.txt](https://github.com/learningequality/kolibri/files/4742625/daemon.txt)

### Context
 * Kolibri version 0.13.3, 0.13.2, 0.13.0
 * Arch Linux
 * Firefox
Python 3.6

jredrejo:
Hi @nobicycle, in order to reproduce it:
- is /mnt in a diferent disk volume?
- are you using apache or nginx?

nobicycle:
@jredrejo 
Yes, different disk and different partition.

I don't understand the relevance of the question about web server - the browser is simply accessing the internal Kolibri webserver.

BTW This was broken in the past. 
https://github.com/learningequality/kolibri/issues/4227

>  benjaoming commented
>  symlinks are very powerful mechanisms for distributing massive amounts of contents over several partitions and devices. Also in the case that the database is desired to run on a small, fast drive, while content is stored on a larger, slower drive. Therefore, I think we should keep up the effort to support it, although I agree that now that we have movedirectory, this is obviously cleaner and safer to communicate to normal users.
>

jredrejo:
> @jredrejo
> Yes, different disk and different partition.
> 
> I don't understand the relevance of the question about web server - the browser is simply accessing the internal Kolibri webserver.
> 
No, when using apache or nginx, the standard configuration uses it to serve content resources. They are not served by the internal webserver. Neither apache nor nginx can follow symlinks that are in different system partitions. So, this is a very important point.

That was the reason to add a new option, `STATIC_USE_SYMLINKS`,  in #6925. 
Please, add it to  your options.ini file using this format:

```
[Deployment]
STATIC_USE_SYMLINKS = False
```

[...]

PR #7217, opened 2020-06-30
Make static symlinking check more robust.

rtibbles:
### Summary
* Makes symlink checking more robust, to check for actually symlinking between the two locations required.

### Reviewer guidance
Does this properly skip symlinking of static files, when trying to copy static files between two different storage devices?

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/utils/system.py
+++ b/kolibri/utils/system.py
@@ -20,14 +20,12 @@ import logging
 import os
 import signal
 import sys
-import tempfile
 import time
 
 import six
 from django.db import connections
 
 from .conf import KOLIBRI_HOME
-from .conf import OPTIONS
 from kolibri.utils.android import on_android
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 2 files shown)


jredrejo:
@rtibbles do you think this could solve #7003 ?

rtibbles:
I am not sure - the main issues raised there seem to be about content, not static files. I must admit, reading the issue, I still can't quite understand what has happened to prevent symlinking from working for content between 0.12.8 and 0.13.x

kolibri/utils/system.py
```diff
@@ -239,27 +237,22 @@ def become_daemon(**kwargs):
...
+    between two locations
     return True if it succeeds, return False otherwise.
     """
-    # If STATIC_USE_SYMLINKS has been set to False, return False directly
-    if not OPTIONS["Deployment"]["STATIC_USE_SYMLINKS"]:
-        return False
 
-    fd, temp_target = tempfile.mkstemp()
-    temp_pathname = temp_target + ".lnk"
+    temp_pathname = os.path.join(destination_folder, os.path.basename(source_file))
```
lyw07, line 247: one thing I notice is that the `temp_pathname` may not be the actual symlink we make in the KOLIBRI_HOME directory.

For example, if `source_file` is `xxx/kolibri/core/static/assets/logo.ico`, then os.path.basename is `logo.ico`, which means the `temp_pathname` would be `~/.kolibri/static/logo.ico`. However, I guess what we want is `~/.kolibri/static/assets/logo.ico`? Or does this not matter?

rtibbles: I did this deliberately to avoid making a bunch of child directories in the general case - with the assumption that if we can symlink to the parent directory, then we can symlink to any child directories.

Can't think of any case where that might not be the case.

lyw07: ah sounds good! this makes sense to me!

lyw07 (approved):
LGTM!

lyw07:
sorry I forgot to ask,
> Does this properly skip symlinking of static files, when trying to copy static files between two different storage devices

do we consider my laptop's filesystem and a USB drive to be two different storage devices? Should static files not be symlinked under this case?

rtibbles:
I think so? I am not actually sure what the limits are.

lyw07:
oh I just tested - it creates the symlinks in this case. I guess as long as my filesystem has access to the other storage device, it should be fine?

rtibbles:
Yeah, my guess is that if you had a USB that was formatted as FAT32, then it might fail, but that the pre-check should catch that.

Issue #7155, opened 2020-06-24
'import users' page needs kbuttongroup


indirectlylit:
### Observed behavior

missing padding

![image](https://user-images.githubusercontent.com/2367265/85486384-62b93380-b57f-11ea-8458-7f5cd333884f.png)

### Context

0.14.0 beta 3

PR #7223, opened 2020-06-30
Fix is_learner attribute and admin login

### LINKED ISSUE #7199: SignInPage does not implement the "optional learner password" facility setting
### Observed behavior

Screenshot 1: 

I try to login as a learner `l`, but am brought to this second "Create a password" page. (Also see other issues pointed out in screenshot)

image

Screenshot 2:

I try to login as superuser `jb`. I get stuck at the sign-in page

image

### User-facing consequences

…

### Errors and logs

…

### Steps to reproduce

…

### Context
0.14 beta 6

jredrejo:
### Summary

Fixes the two issues described at #7199, all happening when the facility setup allowing users to login without password is set:

- Self-created users are requested to add a password when trying to login
- admin login becomes impossible

### Reviewer guidance
In the facility settings, allow the users to create their own accounts and to login without password.
1. Try to login as an admin
2. Create a new user and try to login using it without password

### References
Closes #7199 , #7155 

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated


kolibri/core/auth/api.py
```diff
@@ -218,7 +218,7 @@ def users_for_facilities(self, request):
 
         users = (
             FacilityUser.objects.filter(facility__in=facility_ids)
-            .annotate(is_learner=Exists(Role.objects.filter(user=OuterRef("id"))))
+            .annotate(is_learner=~Exists(Role.objects.filter(user=OuterRef("id"))))
```
rtibbles: That's an important character!

kolibri/plugins/user/assets/src/views/SignInPage.vue
```diff
@@ -282,11 +282,12 @@
       ...mapState(['facilityId']), // backend's default facility on load
       ...mapState('signIn', ['hasMultipleFacilities', 'usersForSelectedFacilities']),
       ...mapState({
+        passwordMissing: state => state.core.loginError === LoginErrors.PASSWORD_MISSING,
         invalidCredentials: state => state.core.loginError === LoginErrors.INVALID_CREDENTIALS,
         busy: state => state.core.signInBusy,
       }),
       simpleSignIn() {
-        return this.facilityConfig.learner_can_login_with_no_password;
+        return this.facilityConfig.learner_can_login_with_no_password && !this.passwordMissing;
```
rtibbles: To clarify here - this is being toggled off in the case that a non-learner has tried to login without the password, and then the backend has reported that the password is needed?

rtibbles (commented):
Just one question, but it seems to make sense to me, have not manually tested.

kolibri/plugins/user/assets/src/views/SignInPage.vue
jredrejo: I have moved it to the computed property to decide if the password must be shown, to be more semantic (if that can be said in English)

rtibbles: Yep - perfectly semantic use of semantic for me :)

rtibbles:
Apologies, @jredrejo - I was testing this locally, and things were behaving very much not as I expected, so I built on top of your fixes.

nucleogenesis:
The response speed of clicking a button for a user who needs to make a password is very short (< 1s) but also very noticeable - to the point that on slower hardware it could be > 1s. If I put "Slow 3G" in dev tools it takes ~1-2s.

The wait isn't a problem - but it might be worth having a loading transition.

slow-response-when-selecting-user

nucleogenesis:
@rtibbles It may be worth addressing the issues this PR is targeting during @jredrejo and my overhaul based on the new designs from Riddhi.

I've encountered a few weird issues - all of which seem to stem from how the "create a password" page is handled.

IF you click a username on the User List Flow, 
AND that user has `NOT_SPECIFIED` a password
AND the setting is that you must have a password
THEN the learner is taken to the "create a new password" page
BUT THEN the learner is logged in regardless of what happens on the "create a new password" page. Refresh the page - you're logged in. Go back to "select a user" and refresh - you're logged in as whatever user you selected in the first place.

This shouldn't be possible because they are required to have a password.

Additionally, if you've gotten into that "Create a password" page, the rest of the flow breaks. Go back and change your facility and it doesn't change your facility - keeps you on the current facility ("Sign in to 'Facility 1'" - even when you selected to go to Facility 2) - but lists no users when the user list is supposed to show. If you refresh - you're logged in as whatever user you selected to get you to "Create a password" in the first place - but once you log out you're in the facility you selected before it broke (meaning the value is at least being saved in the local storage).

Here is a gif showing some of what I'm talking about.

screencast-localhost_8000-2020 07 01-14_55_39

nucleogenesis (changes requested):
Left some comments and illustrative gifs

rtibbles:
Issues highlighted in review should be addressed!

nucleogenesis (approved):
LGTM

Thanks @rtibbles - @jredrejo and I should be smooth sailing from here.

PR #7226, opened 2020-06-30
Fix invalid JSON during morango sync

bjester:
### Summary

Updates the content negotiation class to use default behavior for morango.

### Reviewer guidance

- Needs a GZIP enabled peer
- Pull a facility first, make changes, then try to sync with the peer you pulled from (it should push)

### References

…

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/core/negotiation.py
+++ b/kolibri/core/negotiation.py
@@ -1,21 +1,29 @@
-from rest_framework.negotiation import BaseContentNegotiation
+from rest_framework.negotiation import DefaultContentNegotiation
 from rest_framework.parsers import JSONParser
 from rest_framework.parsers import MultiPartParser
 from rest_framework.renderers import JSONRenderer
 
 
-class LimitContentNegotiation(BaseContentNegotiation):
+class LimitContentNegotiation(DefaultContentNegotiation):
     def select_parser(self, request, parsers):
         """
         Always return JSONParser unless a 'multipart/form-data' content type is sent
+        or this is for morango
         """
+
+        if request.path.startswith("/api/morango"):
+            return super(LimitContentNegotiation, self).select_parser(request, parsers)
         if "multipart" in request.content_type:
             return MultiPartParser()
         return JSONParser()
 
-    def select_renderer(self, request, renderers, format_suffix):
+    def select_renderer(self, request, renderers, format_suffix=None):
         """
-        Always return JSONRenderer
+        Always return JSONRenderer unless for morango
         """
+        if request.path.startswith("/api/morango"):
+            return super(LimitContentNegotiation, self).select_renderer(
+                request, renderers, format_suffix=None
+            )
         renderer = JSONRenderer()
         return (renderer, renderer.media_type)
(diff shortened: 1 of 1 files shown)


kolibri/core/negotiation.py
```diff
@@ -1,21 +1,29 @@
...
         """
+
+        if request.path.startswith("/api/morango"):
+            return super(LimitContentNegotiation, self).select_parser(request, parsers)
         if "multipart" in request.content_type:
             return MultiPartParser()
         return JSONParser()
 
-    def select_renderer(self, request, renderers, format_suffix):
+    def select_renderer(self, request, renderers, format_suffix=None):
```
rtibbles, line 6: Does it need the renderer? Because it's gzipping the output too?

kolibri/core/negotiation.py
```diff
@@ -1,21 +1,29 @@
...
 
 
-class LimitContentNegotiation(BaseContentNegotiation):
+class LimitContentNegotiation(DefaultContentNegotiation):
     def select_parser(self, request, parsers):
         """
         Always return JSONParser unless a 'multipart/form-data' content type is sent
+        or this is for morango
         """
+
```
rtibbles: Note, I think we can just remove the `select_parser` override here, and just default to the `DefaultContentNegotiation` parser strategy - it's only the renderer that really needs to be overridden for performance reasons.

bjester: I can make this change

kolibri/core/negotiation.py
bjester: Hmm, I don't think it's needed.

bjester:
@jredrejo Could you verify that removing the `select_parser` override works as expected with whatever prompted your multipart additions?

kolibri/core/negotiation.py
rtibbles: So, do we need the special case for morango in the renderer, or just need the parsing change above?

kolibri/core/negotiation.py
```diff
@@ -1,21 +1,15 @@
...
-        if "multipart" in request.content_type:
-            return MultiPartParser()
-        return JSONParser()
-
-    def select_renderer(self, request, renderers, format_suffix):
-        """
-        Always return JSONRenderer
+        Always return JSONRenderer unless for morango
         """
+        if request.path.startswith("/api/morango"):
```
jamalex: Is there a more robust way to get this URL? Maybe using a `reverse` of some sort? There's a chance Kolibri has been mounted at a different subpath (@rtibbles knows about that process) -- though I'm not sure whether Morango actually supports being mounted like that (whether peers will then discover it correctly, etc).

bjester: Ah, I didn't know that was a known use case. Okay, I think we could just remove this. I was getting another JSON decode error which I can no longer reproduce.

rtibbles (approved):

Issue #7125, opened 2020-06-22
'database is locked' error while import/exporting facility and trying to view device info page


indirectlylit:
### Observed behavior

While running facility import or export, trying to access the device info page throws a 500 error:

![image](https://user-images.githubusercontent.com/2367265/85340258-a0905c00-b49a-11ea-8eb4-dcb4f452fe74.png)

```
{
[...]

PR #7251, opened 2020-07-06
Separate morango pull/push clients with better write locking and better cancellation handling

### LINKED ISSUE #7127: issues with facility import functionality

bjester:
### Summary

This PR integrates the latest morango changes in [this PR](https://github.com/learningequality/morango/pull/85), and only uses the `db_write_task_lock` when data is queuing or dequeuing locally. This also attempts to address cancellation during the stages of a sync task, since we should only allow cancellation when we can expect that the cancellation will occur as expected. For instance, attempting to cancel during a remote dequeuing shouldn't be allowed, as we don't have that control over the remote's processing.

Also addresses the issues with blank passwords as shown in https://github.com/learningequality/kolibri/issues/7127 by setting any blank passwords to `NOT_SPECIFIED` when deserializing.

No cancel button shown when locally integrating received data, or aka local dequeuing. Also progress is indeterminate.
Screenshot from 2020-07-07 14-17-57

### Reviewer guidance

- Utilize the UI to schedule a sync
- Use the management command to sync a facility
- Ensure cancelling during transfer works as expected
- Cancel button shouldn't appear when the task is in a uncancellable stage

### References

Depends on: https://github.com/learningequality/morango/pull/85
Related: https://github.com/learningequality/kolibri/issues/7125
Resolves: https://github.com/learningequality/kolibri/issues/7127

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/core/auth/management/commands/sync.py
+++ b/kolibri/core/auth/management/commands/sync.py
@@ -21,6 +21,7 @@ from kolibri.core.auth.constants.morango_sync import State
 from kolibri.core.auth.management.utils import get_facility
 from kolibri.core.auth.management.utils import run_once
 from kolibri.core.auth.models import dataset_cache
+from kolibri.core.tasks.exceptions import UserCancelledError
 from kolibri.core.tasks.management.commands.base import AsyncCommand
 from kolibri.core.tasks.utils import db_task_write_lock
 from kolibri.utils import conf
@@ -72,7 +73,7 @@ class Command(AsyncCommand):
         )
         # parser.add_argument("--scope-id", type=str, default=FULL_FACILITY)
 
-    def handle_async(self, *args, **options):
+    def handle_async(self, *args, **options):  # noqa C901
 
         (
             baseurl,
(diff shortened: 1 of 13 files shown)


kolibri/core/auth/management/commands/sync.py
```diff
@@ -70,7 +73,7 @@ def add_arguments(self, parser):
         )
         # parser.add_argument("--scope-id", type=str, default=FULL_FACILITY)
 
-    def handle_async(self, *args, **options):
+    def handle_async(self, *args, **options):  # noqa C901
```
rtibbles, line 76: Ah `#noqa C901` I know you well...

kolibri/core/auth/management/commands/sync.py
```diff
@@ -172,103 +175,181 @@ def handle_async(self, *args, **options):
...
         logger.info("Syncing has been completed.")
 
-    def _setup_pull_progress_tracking(self, sync_client, noninteractive):
+    @contextmanager
+    def _lock(self):
+        cancellable = False
+        # job can't be cancelled while locked
+        if self.job:
+            cancellable = self.job.cancellable
+            self.job.cancellable = False
```
rtibbles: Only thing to flag here is that this will not change the value of cancellable transmitted to the frontend (as this will not get saved to the task DB) - as long as this is only needed within the scope of the task, and is not intended to signal to the user that the task is not currently cancellable, this will be fine.

kolibri/core/tasks/api.py
```diff
@@ -835,6 +839,20 @@ def channeldiffstats(self, request):
 class FacilityTasksViewSet(BaseViewSet):
     queues = [facility_queue]
 
+    def _job_to_response(self, job):
+        response = super(FacilityTasksViewSet, self)._job_to_response(job)
```
rtibbles: Ah, yes - you do want it to change in the frontend. I'd recommend reverting these changes in job serialization, and we should make the change in cancellable on the job get saved to the task db.

rtibbles (commented):
Could simplify some of this slightly by saving the change in cancellable. This will need a bit of enhancement of the job API to add a save method, so I think we can skip this for now, and clean this up in later refactor.

Overall approach looks sound.

kolibri/core/tasks/api.py
bjester: Ya, so should I save the job in the task same as I would in the api here?

bjester: Ah well, I guess it gets enqueued.

rtibbles: Yeah - the reason I rolled back on just saving the cancellable status on the job is that we don't currently have any mechanism for doing that save except for changes to the extra_metadata.

rtibbles (approved):
Looks good to me - have not done any manual testing, but the code changes look right. Let's merge this to do bug bashing on it.

PR #7254, opened 2020-07-07
Serve zipcontent and hashi from a different origin to avoid null origin shenanigans

rtibbles:
### Summary
* Serve zipcontent endpoint from a different port (and hence another origin)
* Serve a wrapper HTML page that executes the majority of the Hashi code from that same port
* The HTML5 viewer code points to the wrapper HTML page inside a sandboxed iframe - because it is on a different origin, its origin does not get set to null
* That page loads and waits for data to be asynchronously loaded in from Kolibri proper via the postMessage API
* When the data has loaded, it creates another iframe inside itself (unsandboxed)
* Inside that iframe, the content page is loaded, which has a small script injected in it to reach out into the parent, get the preloaded data, and setup the storage shims
* The rest of the page then executes as normal

* Also turns on autoreload for cherrypy server and uses that by preference for development

### Reviewer guidance
Does zip content render as expected? Does SCORM content properly propagated data out? Does localstorage get persisted?

Open questions:
* Is creating a separate WSGI application the right way to go?
* Will this setup be ok for kolibri-server?
* Should we be serving static files on the alternate WSGI application so that we can pool resources across H5P items?
* Is the caching appropriate?
* Should we be pre-unzipping our H5P, ZIP, and Perseus files to avoid doing this on the fly all the time?
* Should we also then pre-process our HTML files to do the injection of the Hashi script block?

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated


kollivier:
@rtibbles For the open questions, which should be resolved before landing this PR? I assume the unzip questions would be part of a future PR?

rtibbles:
Yeah, I think the unzip would be part of a future PR - the other ones should probably be resolved before landing this (and there may be other blocking questions I have missed).

radinamatic:
**EXE asset tested on Windows 7**

Video & audio, PDF & ePUB files load without issues, exercises fail in all browsers, and HTML5 content is mixed bag: all types display well on Chrome, some on FF, and none on IE11.

[logs.zip](https://github.com/learningequality/kolibri/files/5463129/logs.zip)

### Exercises :x: 

Chrome | Firefox | IE11
------------ | ------------- | -------------
Win7 (start)  Running  - Oracle VM VirtualBox_052 | Win7 (start)  Running  - Oracle VM VirtualBox_053 | Win7 (start)  Running  - Oracle VM VirtualBox_054

### HTML5

type | Chrome :heavy_check_mark: | Firefox :exclamation:  | IE11 :x: 
------ | ------------ | ------------- | -------------
ASB | :heavy_check_mark: Win7 (start)  Running  - Oracle VM VirtualBox_034 | Win7 (start)  Running  - Oracle VM VirtualBox_033 | Win7 (start)  Running  - Oracle VM VirtualBox_035
PhET | :heavy_check_mark: Win7 (start)  Running  - Oracle VM VirtualBox_036 | PhET-FF | Win7 (start)  Running  - Oracle VM VirtualBox_037
Flexbox | :heavy_check_mark: Win7 (start)  Running  - Oracle VM VirtualBox_039 | :heavy_check_mark: Win7 (start)  Running  - Oracle VM VirtualBox_041 | Win7 (start)  Running  - Oracle VM VirtualBox_045
PLIX | :heavy_check_mark: | :heavy_check_mark: | plix-ie
Blockly | :heavy_check_mark: | :heavy_check_mark: | Win7 (start)  Running  - Oracle VM VirtualBox_046
SCORM | :heavy_check_mark: Win7 (start)  Running  - Oracle VM VirtualBox_043 | Win7 (start)  Running  - Oracle VM VirtualBox_048 | Win7 (start)  Running  - Oracle VM VirtualBox_047
H5P | :heavy_check_mark: Win7 (start)  Running  - Oracle VM VirtualBox_051 | :heavy_check_mark: Win7 (start)  Running  - Oracle VM VirtualBox_050 | Win7 (start)  Running  - Oracle VM VirtualBox_049

radinamatic:
Using Safari on the macOS Mojave VM to browse content on the Windows 7 server

### SCORM :heavy_check_mark: 

Mojave (start)  Running  - Oracle VM VirtualBox_056

radinamatic:
Video & audio files do not play on Safari 12.0.3, but probably not due to the server asset, as the same error occurs with `kolibri-beta` site.

Mojave (start)  Running  - Oracle VM VirtualBox_057

kollivier:
@radinamatic Yeah, that video problem looks like a codec issue. Safari is more picky about which video codecs it allows, so we need to be re-encoding videos when they don't use H.264/AAC. Could you file that issue on the PhET repo?

radinamatic:
* PDF & EPUB load withouth issues
* HTML5 all load but half have similar issues in Safari as we saw in Firefox on Windows, SCORM included

### HTML5 

type |   | error 
------ | ------ | -------------
ASB | :x:  | Mojave (start)  Running  - Oracle VM VirtualBox_064
PhET | :x:  | Mojave (start)  Running  - Oracle VM VirtualBox_065
Flexbox | :heavy_check_mark: |
PLIX | :heavy_check_mark: | 
Blockly | :heavy_check_mark: | 
SCORM | :x: | Mojave (start)  Running  - Oracle VM VirtualBox_066
H5P | :heavy_check_mark: |

rtibbles:
On reflection, I think this may be a good opportunity to stop using the zipcontent endpoint for perseus files. As we have more control over the render in the perseus case, we can just load the whole file and parse it (and looking at our perseus plugin code, it's going to require a bit of a rewrite to make it compatible with this, so it will need work either way).

rtibbles:
This should fix the HTML5 issues for IE11, Firefox and Safari. Still doing follow up to fix the exercise rendering.

radinamatic:
I wanted to re-test, but the Buildkite is consistently failing, so no assets for me... :confused:

rtibbles:
Yeah, sorry - I updated the perseus plugin dependency before the new version had been pushed to pypi (I was unable to publish it myself due to auth issues).

radinamatic:
NP, ping me when the Perseus plugin is on pypi!

rtibbles:
Test failure seems to be due to a Py3 only piece of code, will fix. Buildkite is now building though for testing.

kollivier:
I'm seeing strangeness with the autoresizing. When checking African Storybooks in the Kolibri QA channel, enabling fullscreen and then exiting caused me to end up with this:

image

I also want to note that we can't control how the HTML5 content responds to resizes, and I ran into a number of issues like this with the ProFuturo content when trying to make autoresizing work there. 

We have the ability to specify explicit height and width for the iframe, which would solve the problem for Feed the Monster, and since we're planning to move to always fullscreen (at which point, this autoresizing code will be removed anyway), I don't know if it's worth trying to solve the autoresize problem as part of this PR.

rtibbles:
Yeah, I think the issue you are seeing is because there's no associated shrinking the other way when the container has to get smaller.

Happy to pull it out of this PR, but I think it's still worth pursuing (and is something that H5P handles internally as well through similar means).

kollivier:
Yeah, I might suggest we pull it out for now. If we do want to keep pursuing this, I think we should just do it as a subsequent PR. There's quite a lot to review here already! :) 

I am curious as to how H5P does it, though. That's one more thing about their implementation I want to look into!

rtibbles:
It's not terribly different to what I have done, but without the maxing, I think: https://github.com/h5p/h5p-php-library/blob/1a644dda1134b1ce60635db8d0ab5f09caaa03e4/js/h5p.js#L254

rtibbles:
Have backed out the resizing code.

kollivier:
With the latest code, after entering and exiting fullscreen mode, the inner iframe appears to have the height and width attributes set to the fullscreen values. The width is actually specified twice, both as a CSS attribute and as an iframe attribute.  The CSS attribute (set to 100%) appears to take precedence, and so the width correctly adheres to the parent iframe's width. Is the hardcoding pixel values for the inner iframe's width and height intentional?

kollivier:
Also, if I add `height: 100%` to the inner iframe's CSS, content no longer seems to create that second scrollbar. (PLiX content is an example of that.)

radinamatic:
New asset on Windows 7, HTML files display well in all 3 browsers, fullscreen included :+1: :100: 

Exercises work in Firefox & Chrome, but not in IE11.

Win7 (start)  Running  - Oracle VM VirtualBox_084

New **Unity** content is still in the works it seems... It crashes Chrome, loads in Firefox but looks empty, and not compatible with IE11.

Chrome | Firefox | IE11
--- | --- | ---
Win7 (start)  Running  - Oracle VM VirtualBox_080 | Win7 (start)  Running  - Oracle VM VirtualBox_079 | Win7 (start)  Running  - Oracle VM VirtualBox_085


[logs.zip](https://github.com/learningequality/kolibri/files/5535695/logs.zip)

rtibbles:
I fear the issues you are seeing for the Unity content may be beyond our control - @kollivier any thoughts on how we can handle these failure cases better?

kollivier:
@radinamatic On Chrome, did you have the dev tools open when loading the content? Curious if it loads okay if the dev tools aren't running. I've been able to run it on Chrome on my admittedly well-resourced OS X machine, so this may be an example of where HTML5 content can have some enhanced system requirements beyond the base Kolibri requirements.

Reviewing the rest of the HTML content, the only other thing I noticed was that the SCORM content was having issues where it appeared it would only mark the content completed if I completed the entire module in one session. If I left the content and came back and completed it, it would continue to show it as in progress. I tried the content in 0.14.3, and it actually gave me the points while waiting on the loading screen, suggesting that the timer-based completion is running on SCORM as well. Will have to dig into that more.

kollivier:
@rtibbles I think the IE 11 "this content is not supported" message is probably a good approach. The Firefox issue with loading the script blob could be an issue with the zipcontent URL handling though, perhaps we're not setting the right content type for blob URLs?

In general, this might be another argument for having some way of marking requirements and warning if the client doesn't meet them, with a "you may have problems, are you sure you want to try loading this?" message ala the IE 11 one.

radinamatic:
> On Chrome, did you have the dev tools open when loading the content? Curious if it loads okay if the dev tools aren't running.

@kollivier Chrome crashes even without the open dev tools. This Win VM has 2 CPUs and 8GBs of RAM, so not totally low-resource system.

Unity-Chrome-Win7

On another hand, if I browse the server running on the Win guest from the browsers on my Ubuntu host (8 cores, 32 GB RAM), they both load Unity without issues.

Chrome | Firefox
--- | ---
Feed the Monster - Kolibri QA Channel - Kolibri - Google Chrome_088 | Feed the Monster - Kolibri QA Channel - Kolibri - Mozilla Firefox_087

kollivier:
@radinamatic It's strange, in that the Win Resource Monitor doesn't show memory maxing out or anything, but it does spike quickly before the crash. It may not be directly a memory issue, but maybe a lower resource machine just can't keep up with what the Unity web app is trying to do. (Though I agree that's not really a low resource machine. VM graphics drivers could be a factor as well.)

Anyway, the changes in the PR seem to be doing what they should to allow Unity to load (with the possible exception of the blob URL handling on Firefox), so I think we can treat complete Unity support as a separate issue. I'll see if we can find an alternate Unity game to try it with - it could be that the game is just resource intensive and written in a way that doesn't work well with the WebGL export.

radinamatic:
Testing Unity on macOS Mojave (Quad core, 8GB RAM) VM as a client browsing the server on Win 7 VM, and only Safari has issues loading. 

Chrome | Firefox | Safari
--- | --- | ---
Mojave (start)  Running  - Oracle VM VirtualBox_092 | Mojave (start)  Running  - Oracle VM VirtualBox_091 | Mojave (start)  Running  - Oracle VM VirtualBox_090

kolibri/core/content/apps.py
```diff
@@ -15,4 +18,10 @@ def ready(self):
         from kolibri.core.content.utils.sqlalchemybridge import prepare_bases
         from .signals import cascade_delete_node  # noqa: F401
 
+        # Do this to prevent import of broken Windows filetype registry that makes guesstype not work.
+        # https://www.thecodingforums.com/threads/mimetypes-guess_type-broken-in-windows-on-py2-7-and-python-3-x.952693/
+        mimetypes.init(
```
rtibbles: Hrm, realizing that this may be the wrong place for this, as we are mostly depending on it outside of the Django lifecycle, should revisit.

kolibri/core/content/zip_wsgi.py
```diff
@@ -0,0 +1,448 @@
+from __future__ import absolute_import
+from __future__ import print_function
+from __future__ import unicode_literals
+
+import codecs
+import json
+import logging
+import mimetypes
```
rtibbles, line 8: I think this is the only place the mimetype sniffing gets used, so should restore it here.

kolibri/core/content/zip_wsgi.py
```diff
@@ -0,0 +1,448 @@
...
+    response["Access-Control-Allow-Origin"] = "*"
+    response["Access-Control-Allow-Methods"] = "GET, OPTIONS"
+    requested_headers = request.META.get("HTTP_ACCESS_CONTROL_REQUEST_HEADERS", "")
+    if requested_headers:
+        response["Access-Control-Allow-Headers"] = requested_headers
+    # restrict CSP to only allow resources to be loaded from self, to prevent info leakage
+    # (e.g. via passing user info out as GET parameters to an attacker's server), or inadvertent data usage
+    response[
+        "Content-Security-Policy"
+    ] = "default-src 'self' 'unsafe-inline' 'unsafe-eval' data: blob:"
```
rtibbles, line 57: @kollivier we are adding the `blob:` protocol here in the CSP, so I suspect that the Firefox error is not the one causing the issue, but rather related to the Firefox debugger being unable to load the script source when it is a blob.

Similar issue filed here: https://bugzilla.mozilla.org/show_bug.cgi?id=1338558

Not sure what is causing the failed rendering though.

rtibbles (commented):
Just taking a quick look through.

rtibbles:
Have addressed my own comments, updated the exercise renderer not to have an IE11 incompatible regex, and simplified the inner iframe styling. Hopefully it should all be working as intended now!

radinamatic:
Exercises good to go in IE11 with the  the newest asset! :+1: 



This is a different Win 7 VM (same not-so-low resources) but Unity is still not fully loading neither in FF nor Chrome. I guess we're going to approach it separately.

radinamatic:
Out of curiosity, I tested Unity on Edge/Win10 VM, all seem good! :+1: 

Win10 (start)  Running  - Oracle VM VirtualBox_093

rtibbles:
Yeah, I think we can describe this PR as making Unity 'possible' but not making it work completely.

kolibri/core/content/zip_wsgi.py
kollivier: The Unity app works for me on the latest Firefox on OS X, so could be a platform-specific or resource issue. Definitely not a general issue with this code at least.

kollivier (approved):
Okay, things are looking pretty good now! I didn't do a thorough code review of the moved code to check exactly what changed, but the new code looks good (w00t indexeddb!) and H5P and a bunch of HTML5 content is working without regressions, so I think it's solid enough that we can get this in and keep poking around for regressions by trying various HTML5 contents.

kolibri/core/content/zip_wsgi.py
jredrejo, line 57: We need it to be:   
```
 response[
        "Content-Security-Policy"
    ] = "frame-src 'self' *;default-src 'self' 'unsafe-inline' 'unsafe-eval' data: blob:"

```
for it to work with nginx, i.e. `frame-src 'self' *; `must be inserted too

jredrejo (changes requested):
One change in the SCP should be added + templates are causing problems in h5p applications:
```

/zipcontent/45794c2b98973152711cd147fe1658a6.h5p/

  File "/usr/lib/python3/dist-packages/kolibri/dist/django/template/utils.py", line 65, in __getitem__
    return self._engines[alias]
KeyError: 'django'

During handling of the above exception, another exception occurred:

  File "/usr/lib/python3/dist-packages/kolibri/dist/cheroot/wsgi.py", line 386, in __call__
    return app(environ, start_response)
  File "./kolibri/core/content/zip_wsgi.py", line 437, in zip_content_view
    response = generate_zip_content_response(environ)
  File "./kolibri/core/content/zip_wsgi.py", line 428, in generate_zip_content_response
    response = _zip_content_from_request(request)
  File "./kolibri/core/content/zip_wsgi.py", line 410, in _zip_content_from_request
    response = get_embedded_file(zipped_path, zipped_filename, embedded_filepath)
  File "./kolibri/core/content/zip_wsgi.py", line 307, in get_embedded_file
    return get_h5p(zf, embedded_filepath)
  File "./kolibri/core/content/zip_wsgi.py", line 260, in get_h5p
    template = loader.get_template("content/h5p.html")
  File "/usr/lib/python3/dist-packages/kolibri/dist/django/template/loader.py", line 18, in get_template
    engines = _engine_list(using)
  File "/usr/lib/python3/dist-packages/kolibri/dist/django/template/loader.py", line 72, in _engine_list
    return engines.all() if using is None else [engines[using]]
  File "/usr/lib/python3/dist-packages/kolibri/dist/django/template/utils.py", line 89, in all
    return [self[alias] for alias in self]
  File "/usr/lib/python3/dist-packages/kolibri/dist/django/template/utils.py", line 89, in <listcomp>
    return [self[alias] for alias in self]
  File "/usr/lib/python3/dist-packages/kolibri/dist/django/template/utils.py", line 80, in __getitem__
    engine = engine_cls(params)
  File "/usr/lib/python3/dist-packages/kolibri/dist/django/template/backends/django.py", line 30, in __init__
    options['libraries'] = self.get_templatetag_libraries(libraries)
  File "/usr/lib/python3/dist-packages/kolibri/dist/django/template/backends/django.py", line 48, in get_templatetag_libraries
    libraries = get_installed_libraries()
  File "/usr/lib/python3/dist-packages/kolibri/dist/django/template/backends/django.py", line 103, in get_installed_libraries
    for app_config in apps.get_app_configs())
  File "/usr/lib/python3/dist-packages/kolibri/dist/django/apps/registry.py", line 138, in get_app_configs
    self.check_apps_ready()
  File "/usr/lib/python3/dist-packages/kolibri/dist/django/apps/registry.py", line 125, in check_apps_ready
    raise AppRegistryNotReady("Apps aren't loaded yet.")
django.core.exceptions.AppRegistryNotReady: Apps aren't loaded yet.

```

All the rest of html5 seem to work, but seing some errors in the console, mainly with the zipcontent parts:

image

In scorm, I also can see other kind of errors:
image

In Funza there are problems with the audio and localStorage:
image


and..... Unity works!

rtibbles:
Weird, wondering if some of these errors might be caused by the fact that our iframe sandbox property is still configurable. Given that the point of this PR is to make that no longer necessary as a workaround, I think I'll remove that and just set it as a fixed value again. I say this as I see `allow-same-origin` is not set on an iframe somewhere, but that is now the default.

kolibri/core/content/zip_wsgi.py
rtibbles: I was able to replicate the error you were seeing that was resolved by this by setting this in options.ini:
```
[HTML5]
SANDBOX = allow-scripts
```

I am going to remove this option completely and set it to the required value that we want to support, as I think it is causing several of the issues here (but not the template one!).

jredrejo (approved):
After the latest commit everything works properly in all the html5 applications, including h5p.

There's a recurring error message appearing in this latest format related with a svg icon, but it does not affect to the correct operation of the application:
image

I only doubt if we will have firewall problems in some cases, but that's something a sysadmin should be able to fix if he's setup a firewall.

Great work!

rtibbles:
Actually looking at the error, I think this is an issue with trying to display an icon for the `H5P` content kind - nothing to do with my PR :)

Issue #7147, opened 2020-06-24
errors when trying to explore without an account


indirectlylit:
### Observed behavior

Reported by @jonboiser, @jayoshih, @jtamiace, @radinamatic 

* https://www.notion.so/learningequality/Get-the-401-error-pop-up-after-clicking-Explore-without-account-1a40ad6d83af4d089c0c009958ae98a8
* https://www.notion.so/learningequality/Got-403s-when-I-tried-to-explore-without-an-account-84d10b3edb2c43ba80ad83d41c7960ca

However I can not reproduce on the Kolibri beta server

### Context

0.14.0 beta 3

micahscopes:
I've managed to reproduce this and am working on it now. 

The authorization popup only appears if I navigate to `/en/learn/#/classes` in an incognito window.  It's because in the incognito window, I've not authenticated to the `/learn/api/learnerclassroom` endpoint, which returns a 401 response with a `www-authenticate` header (as required by the standard).  The browser sees the 401 + `www-authenticate` header and decides to show a popup.

The two ways to solve this would be:
- send a 403/forbidden response instead
- don't send `www-authenticate` headers (a common hack)

micahscopes:
Ooh, I've found a third possibility, which is to send a `X-Requested-With: XMLHttpRequest` header with the request, which should suppress the `www-authenticate` header in the response` (https://stackoverflow.com/a/20221330)

jonboiser:
Users shouldn't be taken to the `/learn/classes` page when they click "Explore with an account" is that what's happening to you?

To avoid the situation where an unauthenticated user is going to `/classes`, we should be redirecting them away to the `/learn/topics` page like I want to do in #7154 (for authenticated users not part of any classes).

rtibbles:
Yeah - this sounds like a follow on consequence of #7154, but also agree we should not be sending the `www-authenticate` header, as it's not relevant in this context.

micahscopes:
When I click "Explore without an account", I'm directed to `/learn/topics`.  So I guess I haven't reproduced the exact behavior described here.

That said, it seems that the 401 popup could be considered to be a secondary issue.  Navigating to any page that makes a request to `/learn/api/learnerclassroom` will produce an authentication dialog because of the 401 response on `/learn/api/learnerclassroom`.

[...]

PR #7257, opened 2020-07-07
Suppress login popup on unauthenticated xhr requests in development

micahscopes:
### Summary
`BasicAuthentication` is used as a default authentication class in development.  Because it returns HTTP code 401 with a `www-authenticate` header, this causes the browser to show a login popup any time an unauthenticated request is made, even if that request is made in the background.

However, in production deployments, `BasicAuthentication` is not used, so the popups don't happen.

This caused some confusion in debugging #7147.

This change implements a middleware for the behavior described [here](https://stackoverflow.com/a/20221330), where `www-authenticate` headers are removed for all requests with the header `X-Requested-With: XMLHttpRequest`.

Additionally, the `X-Requested-With: XMLHttpRequest` header is set on all requests from Kolibri's default HTTP client. 

Currently the middleware is only enabled for development settings, but it could also be enabled in production if we wanted to use the 401 response code more intentionally.

### Reviewer guidance
Steps to test:
1. Start Kolibri with development settings
2. Open an incognito window
3. Navigate to `http://localhost:8000/en/learn/#/classes`
4. Make sure there's no login popup!

### References
- https://github.com/learningequality/kolibri/issues/7147
- https://stackoverflow.com/a/20221330

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/core/auth/middleware.py
+++ b/kolibri/core/auth/middleware.py
@@ -64,3 +64,14 @@ class CustomAuthenticationMiddleware(AuthenticationMiddleware):
             "'kolibri.core.auth.middleware.CustomAuthenticationMiddleware'."
         )
         request.user = SimpleLazyObject(lambda: _get_user(request))
+
+
+class XhrPreventLoginPromptMiddleware(object):
+    def __init__(self, get_response):
+        self.get_response = get_response
+
+    def __call__(self, request):
+        response = self.get_response(request)
+        if response and response.status_code == 401 and request.is_ajax():
+            del response["WWW-Authenticate"]
+        return response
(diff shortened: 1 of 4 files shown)


kolibri/core/auth/middleware.py
```diff
@@ -64,3 +64,14 @@ def process_request(self, request):
             "'kolibri.core.auth.middleware.CustomAuthenticationMiddleware'."
         )
         request.user = SimpleLazyObject(lambda: _get_user(request))
+
+
+class XhrPreventLoginPromptMiddleware(object):
+    def __init__(self, get_response):
+        self.get_response = get_response
+
+    def __call__(self, request):
```
jonboiser, line 83: Could you write some comments explaining why we delete this header?

jonboiser (approved):

PR #7275, opened 2020-07-08
Add a database_only mode in scanforcontent

dylanmccall:
## Summary

This change adds a `--database-only` argument to the `scanforcontent` management command. It allows us to use the command's existing smarts to conditionally import a new channel database, but skipping the step which sets content visibility based on detected files. It is useful for the content extensions work we are doing in the Kolibri flatpak.

### Reviewer guidance

I am unsure if this is the best place for this functionality - it just happened to be the smallest diff for the task at hand. (I had considered changing `importchannel` to understand when a channel database is already reachable on Kolibri's home dir or set of extra content directories, but `scanforcontent` already does that nicely). Please let me know if this looks sane or if there's a better place for it.

### References

Content import in Kolibri flatpak: learningequality/org.learningequality.Kolibri#7

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/core/content/management/commands/scanforcontent.py
+++ b/kolibri/core/content/management/commands/scanforcontent.py
@@ -1,4 +1,5 @@
 import logging
+from argparse import SUPPRESS
 
 from django.core.management.base import BaseCommand
 from sqlalchemy.exc import DatabaseError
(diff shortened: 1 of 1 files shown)


indirectlylit:
@dylanmccall would you mind re-targeting to the 0.14.x branch?

dylanmccall:
@indirectlylit okay, done!

indirectlylit:
thank you!

kollivier:
@dylanmccall Sorry it has taken time for us to look into this! Could you clarify whether this fix is a performance improvement (since in your case you know the files on disk are available), or did you run into an issue when calling `scanforcontent` without this in your setup?

Also, I might suggest using a flag name like `--skip-annotations` to be more explicit about what happens when you set the flag.

dylanmccall:
Sure, that sounds like a _much_ better name :) I'll change that.

The issue with using `scanforcontent` _without_ this is it ends up going through content files on disk and marks content nodes as available based on that, but for content extensions we already have a metadata file which lists the content nodes we want to show; we don't need to do that step. What we _do_ need `scanforcontent` to do is to discover the channel database. After doing that, we can use `importcontent` with the correct `--node-ids` and `--exclude-node-ids`. So it's a performance optimization, and it also (as importantly) solves a problem where Kolibri would end up showing unexpected content nodes after loading an extension.

kollivier:
@dylanmccall Okay, great, thanks for the detailed explanation! I think this is okay to merge, but it might be good to first add a comment in the code about it being related to content extensions (or maybe put a link to this PR?) so that people who come across it and wonder about it have some way to determine the context.

dylanmccall:
Oh yeah, people might be a little confused by that. I added a comment now. Thank you!

kollivier (approved):
Great, thanks! I think we're good to go with this!

PR #7276, opened 2020-07-09
Let the taskrunner use multiprocessing.

rtibbles:
### Summary
Fixes a very longstanding issue where the async task runner doesn't actually use multiprocessing when available.
Runs taskrunner test suites against both threading and multiprocessing for coverage.
Adds an option to run the task workers with multiprocessing if so desired.

### Reviewer guidance
Tests now seem to be running solidly, and it runs against both threading and multiprocessing across all versions.

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated


rtibbles:
Looks like the test flakiness reproduces with abundance on Travis. Going to set this as a draft for now.

kolibri/core/tasks/compat.py
```diff
@@ -1,19 +1,28 @@
...
+    from multiprocessing import synchronize  # noqa
 
     # Proxy Process to Thread to allow seamless substitution
     from multiprocessing import Process as Thread  # noqa
     from multiprocessing import Event  # noqa
 
     # any variable is local to a process, so this is
     # just a dummy
-    local = object
+    class local(object):
```
aronasorman, line 15: Interesting, what's the reason for the additional class definition here?

kolibri/core/tasks/compat.py
```diff
@@ -1,19 +1,28 @@
+from kolibri.utils.conf import OPTIONS
+
 try:
+    if not OPTIONS["Tasks"]["WORKER_MULTIPROCESSING"]:
```
aronasorman: I think a clearer setting and env var is something like `USE_WORKER_MULTIPROCESSING`

aronasorman (commented):
Looks good, just some minor comments. Will test in a bit

kolibri/core/tasks/compat.py
rtibbles: I think you can't actually directly instantiate `object` because it's an ABC.

rtibbles: Weird - seems to work fine locally though, possible that it caused issues in some Python versions.

kolibri/core/tasks/compat.py
rtibbles: Updated.

aronasorman:
Ran a local test:
- can confirm that kolibri in worker multiprocessing mode can still import sucessfully
- confirmed that adding this setting creates a new subprocess:
```
# multiprocessing worker mode
#+RESULTS:
: bash───python─┬─python
:               └─312*[{python}]

# vs in non-multiprocessing

#+RESULTS:
: bash───kolibri-0.15.0.───311*[{kolibri-0.15.0.}]

```

aronasorman (approved):

PR #7278, opened 2020-07-09
Remove vue-test-utils deprecated methods

### LINKED ISSUE #7215: Fix vue-test-utils deprecation errors
`vue-test-utils` in changing their API again. The most common errors right now are these:

```
      [vue-test-utils]: finding components with `find` is deprecated and will be removed in the next major version. Use `findComponent` instead.
      [vue-test-utils]: contains is deprecated and will be removed in the next major version. Use `wrapper.find`, `wrapper.findComponent` or `wrapper.get` instead.
      [vue-test-utils]: isVueInstance is deprecated and will be removed in the next major version.
```

In the `spec.js` files, replace all calls to `wrapper.find` with `wrapper.findComponent` (and whatever other recommendations they make) and it should clear these up.

MisRob:
### Summary

This PR removes all methods that will be removed in vue-test-utils v1.0. There are some issues I'd like to point out - please have a look at code comments below.

Maybe we can try to upgrade to v1 as @jonboiser mentioned in #7215 though taking into account how many files are affected by this PR I'd rather do that in a different PR.

### References

Closes #7215 
[Upgrade documentation](https://vue-test-utils.vuejs.org/upgrading-to-v1/)

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated


kolibri/plugins/coach/assets/src/views/common/notifications/__test__/ActivityList.spec.js
```diff
@@ -197,12 +197,6 @@ describe('ActivityList component', () => {
     });
     const { wrapper } = makeWrapper({});
 
-    wrapper.setMethods({
```
MisRob, line 200: This seems to be unnecessary as tests pass without it and I haven't found a place in tests and code where it would be needed. I just wanted to highlight it just in case I've missed something.

kolibri/plugins/coach/assets/src/views/common/notifications/__test__/NotificationCard.spec.js
```diff
@@ -1,6 +1,8 @@
 import { mount, RouterLinkStub } from '@vue/test-utils';
 import NotificationCard from '../NotificationCard';
 
+NotificationCard.methods.getRoute = x => x;
```
MisRob, line 4: vue-test-utils v1 doesn't support overwriting methods when mounted. The same is true for `setMethods`. To solve this issue I used [this strategy](https://github.com/learningequality/kolibri/blob/13445ab3e443d381154b5cde44b644cf2bc5a0d1/kolibri/plugins/user/assets/test/views/profile-page.spec.js#L9-L11) for now though I'd like to open a follow-up to try to get rid of overwriting methods in this way in the whole codebase. It should be noted that it might be complicated at some places (see [this conversation](https://learningequality.slack.com/archives/CB37UM23A/p1594049471466000) to get more details) though maybe we can try to at least get rid of as much occurrences as possible.

kolibri/plugins/device/assets/test/views/SelectTransferSourceModal.spec.js
```diff
@@ -6,8 +6,12 @@ function makeWrapper(options) {
   const wrapper = shallowMount(SelectTransferSourceModal, {
     ...options,
     stubs: {
-      SelectImportSourceModal: '<div data-test="select-import-source"></div>',
-      SelectDriveModal: '<div data-test="select-drive"></div>',
+      SelectImportSourceModal: {
+        template: '<div data-test="select-import-source"></div>',
```
MisRob, line 10: This is not documented in the migration guide linked in the PR description but v1 doesn't support string stubs anymore - you can check deprecation notice at the very bottom of [mount documentation](https://vue-test-utils.vuejs.org/api/mount.html).

kolibri/plugins/user/assets/test/views/sign-up-page.spec.js
```diff
@@ -20,12 +20,6 @@ function makeWrapper() {
   return mount(SignUpPage, {
     store,
     router,
-    methods: {
```
MisRob, line 23: Again, tests were passing without problems when I removed this so I haven't replaced it though wanted to point it out.

packages/kolibri-tools/package.json
```diff
@@ -17,6 +17,7 @@
     "@babel/core": "^7.9.6",
     "@babel/plugin-transform-runtime": "^7.9.6",
     "@babel/preset-env": "^7.9.6",
+    "@testing-library/jest-dom": "^5.11.0",
```
MisRob, line 20: I installed this package to be able to get rid of `isVisible` as recommended [here](https://vue-test-utils.vuejs.org/upgrading-to-v1/#isvisible). I am not sure what's our packages release process - do we need to raise `kolibri-tools` version?

By the way the library [looks very useful](https://github.com/testing-library/jest-dom#table-of-contents) - if there's interest maybe we can check it out and start using it for our tests. I can mention it on dev meeting.

kolibri/core/assets/test/views/app-bar.spec.js
```diff
@@ -53,23 +53,23 @@ describe('app bar component', () => {
...
     it('should be shown if userMenuDropdownIsOpen is true', async () => {
       const wrapper = createWrapper(undefined, { userMenuDropdownIsOpen: true });
-      expect(wrapper.contains(CoreMenu)).toBe(true);
+      expect(wrapper.findComponent(CoreMenu).element).toBeTruthy();
     });
     it('should show the language modal link if no components are added and user is not logged in', () => {
       expect(navComponents).toHaveLength(0);
       const wrapper = createWrapper();
-      expect(wrapper.contains(CoreMenuOption)).toBe(true);
+      expect(wrapper.findComponent(CoreMenuOption)).toBeTruthy();
```
jonboiser: Does this one also need to have `.element`?

jonboiser: Similar question for line 107

kolibri/plugins/setup_wizard/assets/test/views/FacilityPermissionsForm.spec.js
```diff
@@ -26,17 +26,17 @@ describe('FacilityPermissionsForm', () => {
   it('"non-formal" option is selected by default and facility name textbox is visible', () => {
     const { els } = makeWrapper();
     expect(els.nonFormalRadioButton().vm.isChecked).toEqual(true);
-    expect(els.nonFormalTextbox().isVisible()).toEqual(true);
+    expect(els.nonFormalTextbox().element).toBeVisible();
```
jonboiser, line 29: Is this `toBeVisible()` assertion provided by Jest?

packages/kolibri-tools/package.json
jonboiser: Yeah, this is a great addition! I don't believe we need to re-publish `kolibri-tools` if we're only using `testing-library` in Kolibri.

kolibri/plugins/coach/assets/src/views/common/notifications/__test__/NotificationCard.spec.js
jonboiser: Yeah, I think most of the time I've done this is out of desperation and deactivate a method that has some kind AJAX side effect. So for most of these we might be able to use the xhr-mock library that was added recently. But for this particular case, I think it's okay to do it occasionally with the intention of mocking the function's dependencies at some point.

jonboiser:
@MisRob I updated `app-bar.spec.js` to make sure we were testing the `.element` property for truthiness.

I also upgraded test utils locally to 1.0.2 and everything looks good. We can do a separate PR to update the package after this is merged.

jonboiser (approved):

jonboiser:
Oddly enough, we've had the test-utils pinned to 1.0.3 in the `yarn.lock` file for awhile, which is possibly why we've been seeing these deprecations errors. I'm just going to directly commit a small change to the yarn.lock and package.json to reflect this.

Issue #6613, opened 2020-03-02
Automated setup of Kolibri


kill9zombie:
Hello

The setup wizard is great if you've got a couple of instances to do.  As we have more  instances, we're trying to automate the setup:

- [x] Language
- [x] Facility Name
- [ ] No guest access
- [x] Admins have to create the accounts
- [ ] Require passwords on learner accounts
- [x] Automate setting up the super admin

So far we can set the language, facility name, learner signup/admins have to create accounts and super admin with the `kolibri manage provisiondevice` command:

`kolibri manage provisiondevice --facility Foo --superusername alice --superuserpassword alice --preset formal --language_id en --noinput`

We pretty much want the formal [config preset](https://github.com/learningequality/kolibri/blob/develop/kolibri/core/auth/constants/facility_configuration_presets.json), but with `learner_can_login_with_no_password` set to `false`.  We'd also want to set the previous `allow_guest_access` setting to `false`.

Is there a way to set device settings from the command line?  If so, we might just need another config preset.

### Context

Kolibri 0.13.0
Running in a docker container FROM ubuntu:18.04

The reason we're trying to set it up this way is because we're using the Kolibri [OpenID Connect plugin](https://github.com/learningequality/kolibri-oidc-client-plugin).  All the users are held in the OpenID Connect provider, so we always want people to login.

jonboiser:
Hi @kill9zombie , unfortunately the `provisiondevice` command is limited to just setting one of the three presets.

I think it might be possible to customize your presets on the command line by entering the `kolibri shell` and modifying the `kolibri.core.auth.models.FacilityDataset` model directly to get the two settings customized the way you want:

https://github.com/learningequality/kolibri/blob/7634c69dd178349c881da60e514b811dc100a1d3/kolibri/core/auth/models.py#L96-L129

Here's a [stackoverflow article](https://stackoverflow.com/questions/16853649/how-to-execute-a-python-script-from-the-django-shell) on how to pipe a customized script into the django shell. Let me know if this basic approach works for you or if you need additional help.

In the kolibri shell REPL, here are some commands that work

```python
from kolibri.core.auth.models import FacilityDataset

facility = FacilityDataset.objects.all()[0]
# do NOT allow access with no password
facility.learner_can_login_with_no_password = False
# 
facility.save()
```

Similarly, the `allow_guest_access` setting is on a different model `kolibri.core.device.models.DeviceSettings`

[...]

Issue #7281, opened 2020-07-09
server doesn't enforce internally-consistent settings


indirectlylit:
### Observed behavior

Certain values of facility and device settings are incompatible with each other. For example. you cannot have:

```
learner_can_login_with_no_password: true
learner_can_edit_password: true
```

because learners cannot edit their password when they don't have a password.

This behavior is enforced on the client-side but not the server-side, e.g. during device provisioning.

### Expected behavior

Server-side model validation, checked by unit tests, should ensure that server state is always internally consistent.

### Context

0.14.0 beta 8

ashmeet13:
Hi, can I pick this up? 
From what I understand going through the code - this [model class](https://github.com/learningequality/kolibri/blob/c655c7dcb03064d5764d7066a52999699c01178b/kolibri/core/auth/models.py#L98-L130) requires a clean() function that would validate such case?

New to Kolibri. Would really like if someone could guide me!

jonboiser:
@ashmeet13 I think implementing the `clean` method would work for the example in the original post. You might also need to modify some code on the API layer to call `clean` before saving. For example, this code here for `PATCHING` the setting does not do a  `clean` before `save`.

https://github.com/learningequality/kolibri/blob/10733c9a1fb0fba9122411e5a259d7e7fa0b82e8/kolibri/core/device/api.py#L151

I think we still need to figure out the full set of "inconsistent" settings, but you're free to get started with the one in the original post where a `DeviceSettings` model is in an invalid state if both
```
learner_can_login_with_no_password: true
learner_can_edit_password: true
```

fields are true.

ashmeet13:
Got it - I think for now I will start implementing the `clean()` function for

[kolibri/core/auth/models.py - FacilityDataset Model](https://github.com/learningequality/kolibri/blob/release-v0.14.x/kolibri/core/auth/models.py#L98)

This should solve the problem for the case in the original post.

rtibbles:
Fixed in #7856

PR #7282, opened 2020-07-09
parameters to device provisioning command for arbitrary settings

indirectlylit:
### Summary

* allow arbitrary settings to be passed in as JSON files
* added warning that the API for this command is liable to change in the future

### Reviewer guidance

You can specify JSON files that will take precedence over other settings.

For example:

```bash
kolibri manage provisiondevice --skip-update --device_settings ./ds.json --facility_settings ./fs.json
```

where _ds.json_ is:

```json
{
  "allow_guest_access": false
}
```

and _fs.json_ is:

```json
{
  "learner_can_login_with_no_password": false
}
```

### References

fixes https://github.com/learningequality/kolibri/issues/6613

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/core/device/management/commands/provisiondevice.py
+++ b/kolibri/core/device/management/commands/provisiondevice.py
@@ -1,15 +1,11 @@
 import logging
 
 from django.conf import settings
-from django.core.management.base import BaseCommand
-from django.core.management.base import CommandError
+from django.core.management.base import BaseCommand, CommandError
 from django.db import transaction
 from django.utils import six
-
-from kolibri.core.auth.constants.facility_presets import mappings
-from kolibri.core.auth.constants.facility_presets import presets
-from kolibri.core.auth.models import Facility
-from kolibri.core.auth.models import FacilityUser
+from kolibri.core.auth.constants.facility_presets import mappings, presets
+from kolibri.core.auth.models import Facility, FacilityUser
 from kolibri.core.device.utils import provision_device
 
 logger = logging.getLogger(__name__)
--- a/kolibri/core/device/management/commands/provisiondevice.py
+++ b/kolibri/core/device/management/commands/provisiondevice.py
@@ -1,4 +1,6 @@
+import json
 import logging
+import os
 
 from django.conf import settings
 from django.core.management.base import BaseCommand, CommandError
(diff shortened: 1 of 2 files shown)


indirectlylit:
ready for review. I believe the test failures are unrelated

kolibri/core/device/management/commands/provisiondevice.py
```diff
@@ -15,6 +17,45 @@
 logger = logging.getLogger(__name__)
 
 
+def _check_setting(name, available, msg):
+    if name not in available:
```
jonboiser, line 21: Does this check if the value is always a boolean (I think all of the settings are only Booleans, except the device `name`)

I think this should get caught at least when you `.save` the model, though, but it might show a big error message and traceback.

indirectlylit: The `landing_page` setting also takes an enumerated list of strings, not a bool.

At this point I intentionally omitted value validation because I didn't want to duplicate logic with internal model validation.

There's also even more complex logic because of inter-relationships between settings. Opened this related issue: https://github.com/learningequality/kolibri/issues/7281

indirectlylit:
@kill9zombie this is ready for your testing

jonboiser:
Just noticed a couple of UX things that should be fixed:

1. The first yes/no prompt should separate the options like `[y/n]` to help with readability

https://github.com/learningequality/kolibri/blob/84a540b7852f26a818d52803c253d0749ce09869/kolibri/core/device/management/commands/provisiondevice.py#L30-L35

2. The formal/informal/nonformal preset prompt should be formatted the same way. It might be nice if the user could just enter the first letter e.g. (`(f)ormal/(i)nformal/(n)onformal`) or somethign like that.

3. I ran the `provisiondevice` command twice, and it almost seemed like a way to create a second Facility. However, what happens is that it creates a new facility, but the superuser you make doesn't get linked to it (it gets linked to the default facility). Idk if the command should fail if the device is already provisioned to avoid this potential confusion.

indirectlylit:
I totally agree with the above - and there are other things that I noticed also. Was a bit worried about scope creep and possibly breaking scripts that rely on the existing behaviors of `provisiondevice`.

@rtibbles @jamalex curious what your vision is for the future of this command. My sense is that we might want to start from scratch with a cleaner API and more consistent behaviors?

rtibbles:
I would be fine with adding a deprecation warning to the existing command and starting afresh.

indirectlylit:
@jonboiser added a warning that this API is liable to change. This will allow us to start from scratch

a80a4be2dbfe395695488e5e0cff7c73703356d8

jonboiser (approved):

PR #7285, opened 2020-07-09
fix whitespace problems

micahscopes:
### Summary

I had committed a linting violation and with this PR I attempt to redeem myself.


jonboiser (approved):

rtibbles:
Suggest merging once linting passes, as other tests will still fail.

PR #7295, opened 2020-07-14
force utf8 decoding for SQL Alchemy types

aronasorman:
### Summary

This forces SQL Alchemy to interpret every column received from postgres and sqlite as UTF8. 

Somehow despite the encoding environment variables being set as UTF8, SQL Alchemy still interprets the values sent as ascii. This causes encoding issues on the profuturo server when some exceptions contain
non-ascii characters.

NOTE: don't merge yet, since we need to fully test this, we need some build artifacts deployed to profuturo first.

…

### Reviewer guidance

To check, once profuturo is deployed, import some new test channels.

If there are no errors, create some new tasks and then manually add some UTF8 strings.

…

### References
Permanently solves [this](https://sentry.io/organizations/learningequality/issues/1771408928/?environment=profuturo-prod&project=1378215&referrer=alert_email) sentry error.

This fix was suggested by [this](https://stackoverflow.com/questions/11972363/sqlalchemy-and-unicodedecodeerror) stack overflow answer.

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated


rtibbles:
Note that the linked SO post is specific to MySQL.

I think this error is Postgresql specific, and would be best applied using this fix here: https://stackoverflow.com/a/14788796

Would also be good to cover this with a simple test - I assume the issue is having a unicode string inside the extra_metadata on a job, so that should be relatively simple to replicate in a unit test.

Lastly, to properly test this, I realize that we only override Django settings for our postgres test suite, whereas what we really should be doing is overriding our options so that the iceqube storage backend is properly tested too.

aronasorman:
Hmm you're right, it might be best to create tests for this first. Especially since I'm unsure if that suggested SO solution is also the answer.

The basic problem here is that SQL Alchemy interprets whatever we get from the database as ascii instead of UTF-8 when it decodes. That SO answer seems like it's telling postgres what its encoding should be, which is already correct.

Anyway, more testing would help elucidate what the correct answer here is.

aronasorman:
After talking with @rtibbles, here's a summary of what we need to do to get this PR merged:

We need to add a new test that replicates this issue, and then the appropriate fix to address that test. However along the way we also need to change the test infrastructrure to actually test iceqube with postgres. Right now iceqube actually doesn't connect to postgres, and still continues to use sqlite.

We need to do the following changes:
- [x] deprecate the `postgres_test.py` setting and move the `TEST` values found there to `base.py`
- [x] move the credentials found in `postgres_test.py` to env variables found in `tox.ini`'s postgres section
- [x] create a new conditional in the `defaultbackend` function of `test_storage.py` which actually sets up postgres based on CONF values
- [ ] actually write the test that writes, and then reads, UTF-8 metadata on the `jobs` table.

@rtibbles let me know if I missed anything here!

rtibbles:
Seems to cover the bases!

indirectlylit:
This should be merged from 0.13 -> ProFuturo -> 0.14 -> Develop

indirectlylit:
adding 0.14 milestone so we don't lose track of it

kollivier:
Okay, I've added code to stringify exception and traceback objects, and also to log them and nullify if there are any errors encoding them to utf-8. Tests ensure this works, but of course I was unable to replicate the failure, so can't 100% confirm this will resolve the problem.

kolibri/core/tasks/storage.py
```diff
@@ -233,6 +233,27 @@ def mark_job_as_failed(self, job_id, exception, traceback):
...
+                exception = str(exception)
+                exception.encode('utf-8')
+        except:
+            exception = None
+        try:
+            if traceback:
+                logger.warning("Job had traceback: {}".format(traceback))
+                traceback = str(traceback)
+                traceback.encode('utf-8')
+        except:
```
aronasorman, line 254: Hey, can you change this to `except Exception`? Leaving it like this catches more than just real errors, since python likes to use exceptions for regular control flow

kolibri/core/tasks/test/test_storage.py
```diff
@@ -6,8 +7,9 @@
 
 from kolibri.core.tasks.job import Job
 from kolibri.core.tasks.job import State
-from kolibri.core.tasks.storage import Storage
+from kolibri.core.tasks.storage import Storage, ORMJob
```
rtibbles, line 10: Don't see this extra import being used anywhere, and linting won't allow this.

tox.ini
```diff
@@ -112,17 +112,20 @@ passenv = TOX_ENV
...
-    postgres: python3.5
+    postgres: python2.7
 deps =
     -r{toxinidir}/requirements/test.txt
     -r{toxinidir}/requirements/base.txt
     -r{toxinidir}/requirements/cext.txt
     -r{toxinidir}/requirements/postgres.txt
 commands =
-    py.test {posargs:--cov=kolibri --color=no}
+    py.test {posargs:--cov=kolibri --color=no} kolibri/core/tasks/test/test_storage.py -k test_can_save_and_read_utf8_metadata --pdb
```
rtibbles, line 128: We probably don't want to merge this change :)

tox.ini
```diff
@@ -112,17 +112,20 @@ passenv = TOX_ENV
...
     KOLIBRI_HOME = {envtmpdir}/.kolibri
-    DJANGO_SETTINGS_MODULE = kolibri.deployment.default.settings.postgres_test
+    KOLIBRI_DATABASE_ENGINE = postgres
+    KOLIBRI_DATABASE_USER = postgres
+    KOLIBRI_DATABASE_NAME = {env:POSTGRES_DB:default}
+    KOLIBRI_DATABASE_PASSWORD = ""
     KOLIBRI_RUN_MODE = tox
 basepython =
-    postgres: python3.5
+    postgres: python2.7
```
rtibbles, line 121: If switching back to 2.7 does not help with replication, I'd rather revert this change, as if someone can install postgres, they can install a newer version of Python - hopefully...

rtibbles (commented):
Just some things to reduce the scope of the change slightly.

kolibri/core/tasks/storage.py
kollivier: Could you give an example of the usage you're speaking about? i.e. a case where this code would throw but that we wouldn't want the except handling to run?

rtibbles:
Superseded by https://github.com/learningequality/kolibri/pull/7975

PR #7300, opened 2020-07-14
Allow configuration of Redis through Kolibri options

bjester:
### Summary

- Adds options to configure Redis through Kolibri's `options.ini`
- Adds startup check that determines if Redis is sensibly configured
- Use's the Redis Lock implementation when Redis is the cache backend, which uses the `SET NX` command
- Removes vendored version of `RLock` class
- Adds [connection pooling](https://django-redis-cache.readthedocs.io/en/latest/advanced_configuration.html#connection-pool) and options to adjust pooling settings

### Reviewer guidance

- Ensure Kolibri works as expected without Redis backend
- Update `options.ini` to use Redis and ensure:
  1. Kolibri runs as expected
  2. Messages are shown during startup indicating the Redis configuration is being overwritten
  3. Verify changes will are written to the `redis.conf` when `CACHE_REDIS_USE_CONF` is `False`
  4. Verify warning messages appear when maxmemory settings in `redis.conf` are defaults and when `CACHE_REDIS_USE_CONF` is `True`

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/core/apps.py
+++ b/kolibri/core/apps.py
@@ -10,11 +10,11 @@ from django.db.utils import DatabaseError
 from kolibri.core.sqlite.pragmas import CONNECTION_PRAGMAS
 from kolibri.core.sqlite.pragmas import START_PRAGMAS
 from kolibri.core.sqlite.utils import repair_sqlite_db
+from kolibri.core.utils.cache import process_cache
 from kolibri.core.utils.cache import RedisSettingsHelper
 from kolibri.utils.conf import OPTIONS
 
 logger = logging.getLogger(__name__)
-CACHE_OPTIONS = OPTIONS["Cache"]
 
 
 class KolibriCoreConfig(AppConfig):
@@ -34,9 +34,7 @@ class KolibriCoreConfig(AppConfig):
                 settings=os.environ["DJANGO_SETTINGS_MODULE"]
             )
         )
-        # When configured to use Redis, make sure we have sensible Redis settings
-        if CACHE_OPTIONS["CACHE_BACKEND"] == "redis":
-            self.check_redis_settings()
+        self.check_redis_settings()
 
     @staticmethod
     def activate_pragmas_per_connection(sender, connection, **kwargs):
@@ -91,30 +89,28 @@ class KolibriCoreConfig(AppConfig):
             cursor.execute(START_PRAGMAS)
             connection.close()
 
+    # flake8: noqa: C901
     @staticmethod
     def check_redis_settings():
         """
         Check that Redis settings are sensible, and use the lower level Redis client to make updates
         if we are configured to do so, and if we should, otherwise make some logging noise.
         """
-        from kolibri.core.utils.cache import process_cache
-
-        config_use_conf = CACHE_OPTIONS["CACHE_REDIS_USE_CONF"]
-        config_maxmemory = CACHE_OPTIONS["CACHE_REDIS_MAXMEMORY"]
-        config_maxmemory_policy = CACHE_OPTIONS["CACHE_REDIS_MAXMEMORY_POLICY"]
+        if OPTIONS["Cache"]["CACHE_BACKEND"] != "redis":
+            return
+        config_use_conf = OPTIONS["Cache"]["CACHE_REDIS_USE_CONF"]
+        config_maxmemory = OPTIONS["Cache"]["CACHE_REDIS_MAXMEMORY"]
+        config_maxmemory_policy = OPTIONS["Cache"]["CACHE_REDIS_MAXMEMORY_POLICY"]
 
         try:
             # see redis_cache.backends.single.RedisCache
-            client = process_cache.get_master_client()
-            helper = RedisSettingsHelper(client)
-            config_warn = False
+            helper = RedisSettingsHelper(process_cache.get_master_client())
 
             # default setting is "0", or no limit
             maxmemory = helper.get_maxmemory()
             if config_use_conf:
                 if maxmemory == 0:
                     logger.warning("Redis is configured without a maximum memory size.")
-                    config_warn = True
             elif maxmemory != config_maxmemory:
                 helper.set_maxmemory(config_maxmemory)
 
@@ -126,14 +122,13 @@ class KolibriCoreConfig(AppConfig):
                         "Redis is configured without a maximum memory policy. Using Redis with "
                         "Kolibri, the following is suggested: maxmemory-policy allkeys-lru"
                     )
-                    config_warn = True
             elif maxmemory_policy != config_maxmemory_policy:
                 helper.set_maxmemory_policy(config_maxmemory_policy)
 
             # should only change something when `config_use_conf` is also `False`
-            helper.save()
-
-            if config_warn:
+            if not config_use_conf:
+                helper.save()
+            elif maxmemory == 0 or maxmemory_policy == "noeviction":
                 logger.warning(
                     "Please see Redis configuration documentation for details: "
                     "https://redis.io/topics/config"
--- a/kolibri/core/apps.py
+++ b/kolibri/core/apps.py
@@ -10,8 +10,11 @@ from django.db.utils import DatabaseError
 from kolibri.core.sqlite.pragmas import CONNECTION_PRAGMAS
 from kolibri.core.sqlite.pragmas import START_PRAGMAS
 from kolibri.core.sqlite.utils import repair_sqlite_db
+from kolibri.core.utils.cache import RedisSettingsHelper
+from kolibri.utils.conf import OPTIONS
 
 logger = logging.getLogger(__name__)
+CACHE_OPTIONS = OPTIONS["Cache"]
 
 
 class KolibriCoreConfig(AppConfig):
@@ -31,6 +34,9 @@ class KolibriCoreConfig(AppConfig):
                 settings=os.environ["DJANGO_SETTINGS_MODULE"]
             )
         )
+        # When configured to use Redis, make sure we have sensible Redis settings
+        if CACHE_OPTIONS["CACHE_BACKEND"] == "redis":
+            self.check_redis_settings()
 
     @staticmethod
     def activate_pragmas_per_connection(sender, connection, **kwargs):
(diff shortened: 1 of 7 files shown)


kolibri/core/apps.py
```diff
@@ -10,8 +10,11 @@
 from kolibri.core.sqlite.pragmas import CONNECTION_PRAGMAS
 from kolibri.core.sqlite.pragmas import START_PRAGMAS
 from kolibri.core.sqlite.utils import repair_sqlite_db
+from kolibri.core.utils.cache import RedisSettingsHelper
+from kolibri.utils.conf import OPTIONS
 
 logger = logging.getLogger(__name__)
+CACHE_OPTIONS = OPTIONS["Cache"]
```
rtibbles: It's probably OK, but for safety, I'd recommend doing this inside the `check_redis_settings` method.

For context, `OPTIONS` is a lazily evaluated object that relies on the Kolibri plugin registry having been initialized, this should have happened before Django starts importing apps into its app registry, but something may go oddly with import order that might cause this module to be imported earlier.

kolibri/core/apps.py
```diff
@@ -31,6 +34,9 @@ def ready(self):
                 settings=os.environ["DJANGO_SETTINGS_MODULE"]
             )
         )
+        # When configured to use Redis, make sure we have sensible Redis settings
+        if CACHE_OPTIONS["CACHE_BACKEND"] == "redis":
```
rtibbles: Can probably just always call this method, and put the conditional check inside the method itself?

kolibri/core/apps.py
```diff
@@ -84,3 +90,54 @@ def activate_pragmas_on_start():
...
             cursor.execute(START_PRAGMAS)
             connection.close()
+
+    @staticmethod
+    def check_redis_settings():
+        """
+        Check that Redis settings are sensible, and use the lower level Redis client to make updates
+        if we are configured to do so, and if we should, otherwise make some logging noise.
+        """
+        from kolibri.core.utils.cache import process_cache
```
rtibbles: Did this throw an app registry not ready error if imported outside of the scope here?

kolibri/core/apps.py
```diff
@@ -84,3 +90,54 @@ def activate_pragmas_on_start():
...
+                if maxmemory_policy == "noeviction":
+                    logger.warning(
+                        "Redis is configured without a maximum memory policy. Using Redis with "
+                        "Kolibri, the following is suggested: maxmemory-policy allkeys-lru"
+                    )
+                    config_warn = True
+            elif maxmemory_policy != config_maxmemory_policy:
+                helper.set_maxmemory_policy(config_maxmemory_policy)
+
+            # should only change something when `config_use_conf` is also `False`
```
rtibbles, line 127: Could add a conditional on `not config_use_conf` here just to be sure?

rtibbles (commented):
Everything makes sense to me - a couple of comments.

kolibri/core/apps.py
bjester: Good call, I'll make that change.

kolibri/core/apps.py
bjester: I'll update this to go along with other `CACHE_OPTIONS` change

kolibri/core/apps.py
bjester: Yeah, doesn't hurt.

kolibri/core/apps.py
bjester: No, I didn't run into that, but I also didn't try. It seemed perhaps a little safer overall, but I guess if it doesn't break with it outside, we'll be okay.

kolibri/core/apps.py
bjester: Sigh, moving this inside now makes the function "too complex" :expressionless:

rtibbles: @bjester: Hey linter, how do you feel about this code?
Linter: It's complicated!

bjester: I reduced it as much as I could, which wasn't enough, so I added the comment to ignore it.

bjester:
@rtibbles I just added connection pooling. I had seen a message from you on slack about it. Tested it and everything still seems good!

kolibri/core/apps.py
```diff
@@ -84,3 +88,51 @@ def activate_pragmas_on_start():
...
+        try:
+            # see redis_cache.backends.single.RedisCache
+            helper = RedisSettingsHelper(process_cache.get_master_client())
+
+            # default setting is "0", or no limit
+            maxmemory = helper.get_maxmemory()
+            if config_use_conf:
+                if maxmemory == 0:
+                    logger.warning("Redis is configured without a maximum memory size.")
+            elif maxmemory != config_maxmemory:
```
jredrejo: I don't think this is a good idea: to change the `maxmemory` parameter of redis, the user should have requested it explictly.
The fact that the user wants to use redis as Cache does not justify changing the maxmemory value if the user does not set it.
The user might be using redis for other purposes and Kolibri should not change its configuration without the explicit config parameter. 
I think the above if should replace `config_use_conf` by `config_maxmemory`

kolibri/core/apps.py
```diff
@@ -84,3 +88,51 @@ def activate_pragmas_on_start():
...
+                helper.set_maxmemory(config_maxmemory)
+
+            # default setting is "noeviction"
+            maxmemory_policy = helper.get_maxmemory_policy()
+            if config_use_conf:
```
jredrejo: same as above, it should be `config_maxmemory_policy` instead of `config_use_conf`. We should not change the redis maxmemory policy unless it's explicitly requested.

jredrejo (changes requested):
Tested and it all works fine, including release of database connections.
My only concern is this code is changing `maxmemory` and `maxmemory_policy` without the user consent

bjester:
@jredrejo Okay so how does flipping the default `use_conf` value to `True` sound?

rtibbles:
@jredrejo for context, I had urged defaulting to using our Kolibri settings for redis for users who were using Redis solely for Kolibri. Would be nice if we could either turn this on by default in `kolibri-server` or provide some very strong suggestion that they do so!

jredrejo:
> @jredrejo for context, I had urged defaulting to using our Kolibri settings for redis for users who were using Redis solely for Kolibri. Would be nice if we could either turn this on by default in `kolibri-server` or provide some very strong suggestion that they do so!

Yes, `kolibri-server` currently turns it on by default. After this PR is merged I think it would be good that kolibri-server also sets a `maxmemory` setting to be a fraction (maybe one fourth) of the free memory of the device if the device is a low memory device. The rationale is to use this setting to avoid problems in low memory devices, like RPi, and not interfere in the  user settings if the device has enough  memory. I think less than 2 Gb of RAM should make kolibri-server to set the maxmemory option, but any other suggestions are welcome.


> @jredrejo Okay so how does flipping the default `use_conf` value to `True` sound?

I don't quite understand what you mean. In my opinion, it does not matter if the user decides to use redis for caching or not. After enabling redis in Kolibri, we should not change redis settings unless the user decides it. That's why I think in those two `if`s the right condition would be checking, instead of wether redis is selected, if the corresponding setting has been set by the user.

rtibbles:
Hrm, I wonder if what we should do instead is not encode a default into the options - and just let it be nullable? Then we only use it in the case it has actually been set, then we can drop the `use_conf` option?

bjester:
> I don't quite understand what you mean. In my opinion, it does not matter if the user decides to use redis for caching or not. After enabling redis in Kolibri, we should not change redis settings unless the user decides it. That's why I think in those two ifs the right condition would be checking, instead of wether redis is selected, if the corresponding setting has been set by the user.

The `use_conf` option does not enable/disable Redis for Kolibri, it is meant to signify whether the `redis.conf` should take precedence over the settings in the options file.

But as @rtibbles pointed out, if we do that we could drop that `use_conf` config option and make the other options nullable so it only changes something if settings are set for the other values. I think that's a good approach.

jredrejo (approved):
Everything looks good to me, thanks !

jredrejo:
@bjester I have found a late problem:
If the maxmemory is set after kolibri has been used, this error could  happen:
cherrypy.process.wspbus.ChannelFailures: ResponseError("OOM command not allowed when used memory > 'maxmemory'.",)

in such case, kolibri will never start unless the option is changed in the option.ini. I am not sure of what would be better here, https://ma.ttias.be/redis-oom-command-not-allowed-used-memory-maxmemory/ proposes several options but I don't really like any of them.

This is the traceback, at least we need to capture the error and allow kolibri continue loading:
```

ERROR    ENGINE Error in 'start' listener <bound method ServicesPlugin.start of <kolibri.utils.server.ServicesPlugin object at 0x7ff308573748>>
  File "/datos/le/mio/kolibri/venv/lib/python3.6/site-packages/cherrypy/process/wspbus.py", line 230, in publish
    output.append(listener(*args, **kwargs))
  File "/datos/le/mio/kolibri/kolibri/utils/server.py", line 111, in start
    register_zeroconf_service(port=self.port)
  File "/datos/le/mio/kolibri/kolibri/core/discovery/utils/network/search.py", line 188, in register_zeroconf_service
    device_info = get_device_info()
  File "/datos/le/mio/kolibri/kolibri/core/public/utils.py", line 14, in get_device_info
    device_name = kolibri.core.device.models.DeviceSettings.objects.get().name
  File "/datos/le/mio/kolibri/kolibri/core/device/models.py", line 43, in get
    cache.set(DEVICE_SETTINGS_CACHE_KEY, model, 600)
  File "/datos/le/mio/kolibri/venv/lib/python3.6/site-packages/redis_cache/backends/base.py", line 30, in wrapped
    return method(self, client, key, *args, **kwargs)
  File "/datos/le/mio/kolibri/venv/lib/python3.6/site-packages/redis_cache/backends/base.py", line 274, in set
    result = self._set(client, key, self.prep_value(value), timeout, _add_only=False)
  File "/datos/le/mio/kolibri/venv/lib/python3.6/site-packages/redis_cache/backends/base.py", line 267, in _set
    return client.set(key, value, nx=_add_only, ex=timeout)
  File "/datos/le/mio/kolibri/venv/lib/python3.6/site-packages/redis/client.py", line 1451, in set
    return self.execute_command('SET', *pieces)
  File "/datos/le/mio/kolibri/venv/lib/python3.6/site-packages/redis/client.py", line 775, in execute_command
    return self.parse_response(connection, command_name, **options)
  File "/datos/le/mio/kolibri/venv/lib/python3.6/site-packages/redis/client.py", line 789, in parse_response
    response = connection.read_response()
  File "/datos/le/mio/kolibri/venv/lib/python3.6/site-packages/redis/connection.py", line 642, in read_response
    raise response
redis.exceptions.ResponseError: OOM command not allowed when used memory > 'maxmemory'.


```

This was the redis server state:

```
127.0.0.1:6379> config get maxmemory-policy
1) "maxmemory-policy"
2) "allkeys-lru"

127.0.0.1:6379> config get maxmemory
1) "maxmemory"
2) "500"

127.0.0.1:6379> info memory
# Memory
used_memory:839768
used_memory_human:820.09K
used_memory_rss:4124672
used_memory_rss_human:3.93M
used_memory_peak:882392
used_memory_peak_human:861.71K
used_memory_peak_perc:95.17%
used_memory_overhead:832134
used_memory_startup:782504
used_memory_dataset:7634
used_memory_dataset_perc:13.33%
total_system_memory:16626339840
total_system_memory_human:15.48G
used_memory_lua:41984
used_memory_lua_human:41.00K
maxmemory:500
maxmemory_human:500B
maxmemory_policy:allkeys-lru
mem_fragmentation_ratio:4.91
mem_allocator:jemalloc-3.6.0
active_defrag_running:0
lazyfree_pending_objects:0

```

Setting 
CACHE_REDIS_MAXMEMORY = 870000
solves it, any value lower than ` used_memory_peak_human:861.71K ` breaks kolibri
The command 
`memory stats` can also be used to fetch the "_total.allocated_" memory

bjester:
@jredrejo Good catch. I think this should be considered a fatal error, but I think capturing it and displaying options to resolve it would be great.

PR #7306, opened 2020-07-14
Sign In Refactor

nucleogenesis:
### Summary

**Whats new:**

- New and improved designs thanks to @riddhiavlani 
- Make better use of vuex for managing state across the user plugin
- Simplify and clean up template logic
- Break a few bits out into their own components - add comment TODOs to do more extraction for a next pass when time allows
- Make better use of vue-router to encapsulate routing logic

**Whats not new:**

- Some stuff is still kind of hacky and gross and needs some additional restructuring that I will come back to in a future cycle.

**Whats going to be fixed elsewhere**

- Icon styling/spacing/underlining/etc on the backlinks (related: https://github.com/learningequality/kolibri-design-system/issues/69) - I know I've fixed this problem before but it seems to have cropped back up so I'll be working it out in KDS to be sure it's good across the board.

### Reviewer guidance

It feels impossible to describe every possibility... but the Figma linked below is a good start. Walk through each scenario.

You'll want a KOLIBRI_HOME with a single facility and another with multiple facilities.

It helps to have one of each kind of facility on your multiple-facility one - then some extras of each kind... name them fittingly for your testing purposes..

Open your db in sqlitebrowser so that you can change facility types and change learners' passwords to `NOT_SPECIFIED` to test scenarios as needed.

--- Test as many combinations of the following as you can muster ---

- Each facility type (nonformal, formal, informal)
- Devices with multiple facilities and a device with one facility (of any or each of the types)
- Devices with setting disallowing guest access
- Facilities with settings disallowing learner login without password after the learners already had passwords (they should just be logged in with a username or clicking from the UsersList, where relevant)
- Facilities with settings where learners are created without a password (in the DB the value is `NOT_SPECIFIED` for the learner) - the learner ought to be asked to create a password and then will be logged in. Log out and log back in with the new password.
- Facilities where learners can and cannot create accounts.
- Verify that the right strings are showing in the right places where applicable.
- Seriously do every possible routing situation you can think of. Including trying to go to /device without logging in - then clicking the "Sign in" link on that AuthMessage page - then signing in - then you should be redirected appropriately depending on your user's permissions and so on.
- RTL
- Create an account in every kind of facility that you can. Create one with "no password needed" and one with "password needed to login" for learners.
- Mess with the validation messaging - do the messages you see make sense when you see them?

### References

[Designs](https://www.figma.com/file/LI2L6gUncmCDDtaDECcbY6/Log-in-updates?node-id=43%3A3)

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated


kolibri/plugins/user/assets/src/views/SignInPage/index.vue
```diff
@@ -347,7 +349,38 @@
...
+      },
+    },
+    // Clear the username when entering the route.
+    // username may be held over if you select a user from UsersList
+    // then change to a facility that doesn't use UsersList
+    //
+    // TODO: If we want to clear the username whenever we switch facilities,
+    // then we can remove the `beforeRouteEnter` and use the commented out
+    // `beforeRouteLeave` below
+    beforeRouteEnter(to, from, next) {
```
nucleogenesis, line 378: Pretty sure this clears it when we switch facilities because we clear it onRouteEnter and we change facilities on a different route - will clarify this section's comments later.

kolibri/plugins/user/assets/src/routes.js
```diff
@@ -1,91 +1,120 @@
...
       } else {
-        // AUTH_SELECT for multiple facilities device
+        // If Multiple Facilities but we've not stored a facilityId in localstorage
+        // then we go to the AuthSelect route
         if (store.getters.facilities.length > 1 && !store.state.facilityId) {
-          router.replace({ name: PageNames.AUTH_SELECT });
+          next(router.getRoute(ComponentMap.AUTH_SELECT));
         } else {
-          router.replace({ name: PageNames.SIGN_IN });
+          next(router.getRoute(ComponentMap.SIGN_IN));
```
indirectlylit, line 37: Is the `ComponentMap` necessary? I think you can reference these directly like:

```suggestion
          next(router.getRoute(SignInPage.name));
```

indirectlylit (commented):
I'm a bit confused about the `whereToNext` logic. Will need to wrap my head around that.

nucleogenesis:
> I'm a bit confused about the `whereToNext` logic. Will need to wrap my head around that.

The FacilitySelect component requires a route object so it knows where to go after you select your facility. You can only go to SignInPage or SignUpPage routes.

So in AuthSelect (where you have no Facility selected yet):

```
signInRoute() {
        const whereToNext = this.$router.getRoute(ComponentMap.SIGN_IN);
        return { ...this.facilitySelectRoute, params: { whereToNext } };
      },
```

This gets the route to the FacilitySelect and then adds the `whereToNext` as a param - which is a neat (new to me) feature of the vue-router that you can allow props to be passed as params with the `props: true` flag on the route definition:

https://github.com/learningequality/kolibri/blob/02bded3ebc0be4bacaff59308c9eacc9e33aa6ae/kolibri/plugins/user/assets/src/routes.js#L88-L105

---

Edit: I will note that this pattern runs into an issue if you ever end up routing to yourself... as in:

- I'm in FacilitySelect with `whereToNext` pointing to SIGN_IN
- I redirect to `SignUpRoute` - where in the `beforeEnter()` method, I apply some logic that says "go to FacilitySelect where `whereToNext` points to SIGN_UP (instead of SIGN_IN)" - the reason is because the FacilitySelect component displays differently depending on its `whereToNext` route.
- So the router acts like "I am on the FacilitySelect route... and I ended up on FacilitySelect route... so I will change nothing even if the params which are passed as props change"

I found that out while trying to do exactly what the example says... but then I rechecked the design and realized that it wasn't necessary. I found some leads on using a `:key` on the router-view but I'm happy I didn't go too far down the rabbit hole before re-checking the designs

kolibri/plugins/user/assets/src/routes.js
nucleogenesis: In routes.js it would make sense since they're all imported in that file anyway, but the ComponentMap comes in handy within components as well.

I guess the difference is that `import { ComponentMap } from './constants'` in a bunch of files is one import per file - but if I needed to get the route for two or more components from one place then I'd have to import them all, even if they're not used in the template but only as places to be routed to.

kolibri/core/assets/src/views/AuthMessage.vue
```diff
@@ -84,7 +84,7 @@
           return '/';
         } else {
           const currentURL = window.encodeURIComponent(window.location.href);
-          return `${this.userPluginUrl()}#signin?redirect=${currentURL}`;
+          return `${this.userPluginUrl()}?redirect=${currentURL}`;
```
nucleogenesis, line 87: Routing from `/` will get us where we need to be

kolibri/plugins/user/assets/src/modules/pluginModule.js
```diff
@@ -30,6 +31,7 @@ export default {
     },
     setFacilityId(store, { facilityId }) {
       store.commit('SET_FACILITY_ID', facilityId);
+      store.dispatch('getFacilityConfig', facilityId);
```
rtibbles: Should return here so that we can wait on resolution of this async behaviour from the wrapping action.

kolibri/plugins/user/assets/src/modules/pluginModule.js
```diff
@@ -39,7 +41,7 @@ export default {
       if (selectedFacility) {
         return selectedFacility;
       } else {
-        return getters.facilities.find(f => (f.id = getters.currentFacilityId)) || null;
+        return getters.facilities.find(f => f.id == getters.currentFacilityId) || null;
```
rtibbles: That would certainly explain some weird behaviour - this should also be `===`

kolibri/plugins/user/assets/src/views/AuthBase.vue
```diff
@@ -198,10 +198,11 @@
         return this.facilityConfig.learner_can_sign_up;
       },
       signUpPage() {
+        let signUpRoute = this.$router.getRoute(ComponentMap.SIGN_UP);
```
rtibbles: This can be a `const` instead of a `let` as you are not reassigning the value anywhere in the scope.

kolibri/plugins/user/assets/src/views/SignInPage/SignInHeading.vue
```diff
@@ -0,0 +1,72 @@
...
+      showPasswordForm: {
+        type: Boolean,
+        required: true,
+      },
+    },
+    computed: {
+      ...mapGetters(['selectedFacility']),
+      ...mapState('signIn', ['username']),
+      strings() {
+        // Gross
```
rtibbles, line 55: Hear hear!

kolibri/plugins/user/assets/src/views/SignInPage/index.vue
```diff
@@ -1,35 +1,157 @@
...
 
-  <div class="fh">
-
+  <CoreBase
+    :immersivePage="false"
+    :immersivePagePrimary="false"
+    :fullScreen="true"
+  >
     <AuthBase>
+      <div v-if="!needsToCreatePassword">
```
rtibbles, line 13: This makes logical sense, but the fact that the conditional here is so disconnected from the div where we have the create a password logic made it hard for me to read. Don't think refactoring is helpful right now, but maybe just a comment here as well as at the `v-else`, maybe something like:

`Unless we know the user needs to create a password enter this div block for the main flow - see the v-else below for the create password flow`

rtibbles (commented):
Code changes make sense to me, need to do manual testing.

The one thing I don't understand is why we are putting the username and password in vuex state - they are only being referenced inside the `SignInPage` component as far as I can see, so not seeing the value of this transition.

rtibbles:
Tested:
Multi-facility:
* Selecting different facilities
* Logging into a facility with simplified login
* Logging into a facility without simplified login
* Logging into a user that needed to set a password

Single-facility:
* Logging into a facility without simplified login

All tested scenarios above work well.

One edge case I have noticed in the multi-facility case: Perhaps we should filter by facilities that actually have users? I have a bunch of facilities that have none.

kolibri/plugins/user/assets/src/views/SignInPage/index.vue
nucleogenesis: I think that should be it’s own route - but in a next pass. I was taking longer than I wanted so SignInPage is mostly a monolith for now. 

Will add some clarification in comments when I fix the comment issue

nucleogenesis:
> The one thing I don't understand is why we are putting the username and password in vuex state - they are only being referenced inside the SignInPage component as far as I can see, so not seeing the value of this transition.

Yeah you're definitely right. I did it early on expecting to need that information in various places but now that it's all shaken out that's unnecessary for sure.

I won't do that in this PR though for the sake of time and getting this into bug bash, but can make a follow up issue to clean that up.

Issue #7270, opened 2020-07-08
extremely large database size when sync attempts are made but don’t complete


lyw07:
### Observed behavior

Reported by a user:

When sync attempts are made but don’t complete, the table `morango_buffer` in the database took 8.2 GB.

### Expected behavior

According to Blaine, the table `morango_buffer` should be automatically cleared at the end of the sync

### User-facing consequences

Users have an extremely large database

### Errors and logs

…

### Steps to reproduce

…

### Context

Kolibri version: 0.14.0b8

intelliant01:
Additional information: 

Sync is not working with the database shared with @lyw07 on Kolibri 0.13.3 or later despite several attempts. 

I had cleared `morango_store, morango_recordmaxcounter, morango_recordmaxcounterbuffer, morango_syncsession, morango_transfersession` tables and got no error during Kolibri usage or its start. 

But still the sync process never progresses when another server is trying to pull for over 8 hours. It always shows that connection has been successfully established and the sync has started.  

Next reverted the database to the state where only `morango_buffer` is cleared and tried the sync. But the sync process is still just at `Syncing has been initiated (this may take a while)...`

cyberorg:
> Next reverted the database to the state where only `morango_buffer` is cleared and tried the sync. But the sync process is still just at `Syncing has been initiated (this may take a while)...`

redis-cli FLUSHALL

Running this will get it going forward

bjester:
@intelliant01 The sync process can take a significant amount of time, as the message says. With changes coming in 0.14, we've added progress tracking that will provide better visibility on the sync. Using a smaller chunk size may help speed up the sync, depending on the machines and networks involved. You can adjust it with the `--chunk-size` parameter; the default is 500. Clearing the `morango_buffers` table on both the local and remote Kolibri instances, while no sync is running, is currently advisable if there have been several unsuccessful sync attempts.

cyberorg:
Using beta11.

I was able to complete the sync only after clearing out morango_buffer as well as morango_transfersession, if morango_transfersession was not cleared then sync failed after downloading some data with error about already existing morango_transfersess_sync_session_id_0455b5bd_fk_morango_s(don't remember exact error).

After the successful sync and multiple reruns of sync, the morango_buffer is still full of data.

"kolibri manage flushsyncsessions" from https://github.com/learningequality/kolibri/pull/7313 clears out morango_buffer.

[...]

PR #7313, opened 2020-07-15
Add command to flush sync session data

bjester:
### Summary

When sync sessions do not complete properly, they can leave lots of data in the Morango buffers. This is a command that will easily clean those tables.

### Reviewer guidance

Run a few syncs, cancelling them after some has transferred, then run this command.

### References

Helps: https://github.com/learningequality/kolibri/issues/7270

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/core/auth/management/commands/flushsyncsessions.py
+++ b/kolibri/core/auth/management/commands/flushsyncsessions.py
@@ -1,14 +1,11 @@
 import logging
-from contextlib import contextmanager
 
-from django.db import transaction
 from django.db.models import Q
 from morango.models import Buffer
 from morango.models import RecordMaxCounterBuffer
 from morango.models import SyncSession
 from morango.models import TransferSession
 
-from kolibri.core.auth.management.utils import DisablePostDeleteSignal
 from kolibri.core.auth.management.utils import GroupDeletion
 from kolibri.core.auth.utils import confirm_or_exit
 from kolibri.core.tasks.management.commands.base import AsyncCommand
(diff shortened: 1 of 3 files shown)


kolibri/core/auth/management/commands/flushsyncsessions.py
```diff
@@ -0,0 +1,67 @@
...
+                transfer_session_id__in=transfer_sessions.values_list("pk", flat=True)
+            )
+            groups.append(
+                GroupDeletion(
+                    sync_session_id,
+                    querysets=[
+                        RecordMaxCounterBuffer.objects.filter(transfer_session_filter),
+                        Buffer.objects.filter(transfer_session_filter),
+                        transfer_sessions,
+                        SyncSession.objects.filter(id=sync_session_id),
```
lyw07: When we clear the syncsession table, I saw that the facility would be shown as "Never synced" although it was successfully synced before. Is it possible that we only delete the sync session that failed but not the ones that succeeded? or we can just leave them in the table since compared to buffer, there are not a lot of them?

bjester: Hmm, yeah I think keeping only the `SyncSession` records should work. It is a little awkward because we would normally have `TransferSession` records pointing to it. The `Buffer` then references the `TransferSession`. But of course, we should be able to remove all but the `SyncSessions`. I'll have to find where that's being determined to double check that would work; if you know, that would be awesome.

lyw07: oh yes, I agree with you. I guess this would be easiest approach if it would work. Otherwise, we could check to see the transfer sessions from Buffer and get the failed sync session ids from them? Or maybe we could call the flushsyncsessions command in the kolibri syncing logic with the specific failed/cancelled sync session id (sorry not sure if this is achievable as I'm not familiar with the sync command)?

lyw07 (approved):
Tested locally, and looks good to me!

PR #7316, opened 2020-07-15
Clean up c extension code and add pip cache for it

### LINKED ISSUE #4532: Make the warning "No C extensions available" an info-level log entry
### Observed behavior

The warning "No C Extensions available" when running kolibri makes users confused. It would be better if we hide it from most end-users

Please see the discussion here: https://learningequality.slack.com/archives/CB37UM23A/p1542131204155300

### Expected behavior

Users should not see the warning. It should be an `info` level event instead

### User-facing consequences

Users are confused about this warning.

### Errors and logs

```
<WARNING:root>:No C Extensions available for this platform.
```

### Steps to reproduce

Start kolibri in platforms such as Mac OS.

lyw07:
### Summary

This PR is to clean up c extension code as pip version has been upgraded and some packages now have py2/3 compatible and cross-platform wheel file so we can just install the binaries directly.
- remove kolibripip.pex which was the customized build of pip to bypass some download and installation restrictions in the old pip versions
- directly install all the packages through pip instead of downloading the wheel first and searching through the METADATA inside dist-info folder to find the dependencies
- for packages with abi3 tag, we are going to install them to different python version folders. Specifically, for cryptography 2.3, it has cp34 abi3 tag, which is expanded to cp34m, cp35m, cp36m, and cp37m folders. This is because `cffi`, one of the dependencies of cryptography, does not use this abi3 tag.
- update the windows cryptography python version needed to 3.6 since we are use python 3.6 for our windows installer
- also since only pip version greater than 19.3.1 is able to download abi3 tag successfully, exit the program if pip version does not match this criteria
- make the warning of having no c extensions available for the platform as an info-level logging message. Fixes https://github.com/learningequality/kolibri/issues/4532

This PR also sets up a cache directory, using docker volume so we can save time when downloading these wheel files and have the script run in a more stable way. We use https://www.piwheels.org/ to download precompiled wheel files for arm linux. However, the website is not always stable and causes build failure sometimes. By adding a docker volume serving as the pip cache for the c extension, we can mitigate the issue coming from piwheels. @DXCanas , do you mind checking this part so it doesn't affect your build pipeline work? Changes related to this are in `Makefile` and `docker/build_whl.dockerfile`.

Please feel free to let me know if you have any questions! Thank you!

### Reviewer guidance

Please check whether we have installed different platform-specific cryptography packages to `kolibri/dist/cext`. Thank you!
I have tested the PR on a docker container running locally.
The result folder is
```
kolibri/dist/cext/
|-- cp27
|   |-- Linux
|   |   |-- cp27m
|   |   `-- cp27mu
|   `-- Windows
|       `-- x86
|-- cp34
|   |-- Linux
|   |   |-- armv6l
|   |   |-- armv7l
|   |   |-- i686
|   |   `-- x86_64
|   `-- Windows
|       `-- x86
|-- cp35
|   |-- Linux
|   |   |-- armv6l
|   |   |-- armv7l
|   |   |-- i686
|   |   `-- x86_64
|   `-- Windows
|       `-- x86
|-- cp36
|   |-- Linux
|   |   |-- i686
|   |   `-- x86_64
|   `-- Windows
|       |-- AMD64
|       `-- x86
`-- cp37
    |-- Linux
    |   |-- armv6l
    |   |-- armv7l
    |   |-- i686
    |   `-- x86_64
    `-- Windows
        `-- x86
```

### References

https://github.com/learningequality/kolibri/issues/4532

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated


.gitignore
```diff
@@ -131,3 +131,6 @@ SQLITE_MAX_VARIABLE_NUMBER.cache
 
 # don't ignore the TTF font for H5P
 !h5p-core-18.ttf
+
+# static c extensions cache folder
+cext_cache
```
DXCanas, line 136: Can you describe why this was needed? I would think this dir is only every used in docker

lyw07: ah yes it's because when running in Travis and locally, the path `/cext_cache` that we use in docker is not writable, so I have to change it to the path `cext_cache` in that case. It's from this function: https://github.com/learningequality/kolibri/blob/4a564d952636b8137d70da6c6ca6d33a70072c5c/build_tools/install_cexts.py#L231

DXCanas: I'm... not sure I understand. But I don't work with Travis very much. If it's building in Travis "locally", why redefine the cache directory? Would we run into a similar problem if someone were to try and build this locally, without Docker?

I _think_ I found there error here:
https://travis-ci.org/github/learningequality/kolibri/jobs/708571303

lyw07: ah that one was the different error. It was because this script requires a pip version > 19.3.1. The code resolving that issue is here:
https://github.com/learningequality/kolibri/blob/4a564d952636b8137d70da6c6ca6d33a70072c5c/tox.ini#L86-L88
and
https://github.com/learningequality/kolibri/blob/4a564d952636b8137d70da6c6ca6d33a70072c5c/build_tools/install_cexts.py#L257-L264

The error related to the path `/cext_cache` is here: https://travis-ci.org/github/learningequality/kolibri/jobs/708533742#L975
And yes, you are running the script locally or in Travis, and not running as the root user, you will get a permission error when trying to create the path `/cext_cache`. So when that happens, we redefine the path to be `kolibri/cext_cache`

rtibbles (approved):
All outstanding questions have been answered - if there are any issues here, it will be best tested from builds coming out of the pipeline, so we'll merge this now to ensure that it gets more exposure through use.

PR #7333, opened 2020-07-20
Setup Wizard UX and Bug Fixes (wrong facility preset, "stuck" import tasks, saving superuser credentials)

### LINKED ISSUE #7303: Setup Wizard -> Personal -- Creates the facility as "nonformal" - should be "informal"
### Observed behavior

I created a new KOLIBRI_HOME and went through the setup wizard and selected "Personal" to test the Sign In Page work I'm doing for the Personal version.

When I got through with the Setup Wizard, it was created as `nonformal` rather than `informal`. I checked in Vuex and the database and it was created incorrectly.

### Expected behavior

In Setup Wizard:

Personal -> creates an `informal` facility
Advanced -> gives user the option to create a `formal` or `nonformal` facility

### User-facing consequences

Features we have for "Personal" devices/facilities will not be displayed properly.

### Context

0.14-beta9
Debian 10
Chrome

jonboiser:
### Summary

Some updates to the Setup Wizard

1. Fixes faulty polling logic when displaying FacilityImport task progress
1. "remembers" the the super user/admin credentials provided at the beginning of the facility import flow at the end when creating a device super admin. 
1. Fixes a bug where the `facility.dataset.preset` value was not being saved. Fixed in REST and CLI interfaces for device provisioning

### Reviewer guidance

1. Is it still possible to get "stuck" in the facility import step?
1. Once a device is provisioned, does the UI reflect the preset + settings you chose?
1. Is it possible to put the Device Superadmin credentials step in a bad state in terms of the "remembered" credentials from the previous step (providing facility/superuser credentials prior to importing from a peer)

### References

Fixes #7303 

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/plugins/setup_wizard/assets/src/views/importFacility/LoadingTaskPage.vue
+++ b/kolibri/plugins/setup_wizard/assets/src/views/importFacility/LoadingTaskPage.vue
@@ -74,21 +74,18 @@
       },
     },
     beforeMount() {
-      this.startPolling();
+      this.pollTask();
     },
     methods: {
-      startPolling() {
+      pollTask() {
         SetupTasksResource.fetchCollection({ force: true }).then(tasks => {
           this.loadingTask = {
             ...tasks[0],
             facility_name: this.facilityName,
           };
         });
-        if (this.loadingTask.status) {
-          return;
-        }
         setTimeout(() => {
-          this.startPolling();
+          this.pollTask();
         }, 2000);
       },
       retryImport() {
@@ -104,9 +101,6 @@
           })
           .catch(error => {
             this.$store.dispatch('handleApiError', error);
-          })
-          .then(() => {
-            this.startPolling();
           });
       },
       cancelTask() {
--- a/kolibri/plugins/setup_wizard/assets/src/views/importFacility/SelectFacilityForm.vue
+++ b/kolibri/plugins/setup_wizard/assets/src/views/importFacility/SelectFacilityForm.vue
@@ -143,7 +143,12 @@
         const $credentialsForm = this.$refs.credentialsForm;
         if ($credentialsForm) {
           // The form makes the call to the startpeerfacilityimport endpoint
-          return $credentialsForm.startImport();
+          return $credentialsForm.startImport().then(() => {
+            return {
+              username: $credentialsForm.username,
+              password: $credentialsForm.password,
+            };
+          });
         } else {
           return Promise.resolve(false);
         }
--- a/kolibri/plugins/setup_wizard/assets/src/views/importFacility/SelectFacilityForm.vue
+++ b/kolibri/plugins/setup_wizard/assets/src/views/importFacility/SelectFacilityForm.vue
@@ -143,11 +143,15 @@
         const $credentialsForm = this.$refs.credentialsForm;
         if ($credentialsForm) {
           // The form makes the call to the startpeerfacilityimport endpoint
-          return $credentialsForm.startImport().then(() => {
-            return {
-              username: $credentialsForm.username,
-              password: $credentialsForm.password,
-            };
+          return $credentialsForm.startImport().then(importStarted => {
+            if (importStarted) {
+              return {
+                username: $credentialsForm.username,
+                password: $credentialsForm.password,
+              };
+            } else {
+              return false;
+            }
           });
         } else {
           return Promise.resolve(false);
--- a/kolibri/plugins/setup_wizard/assets/src/views/importFacility/SelectSuperAdminAccountForm.vue
+++ b/kolibri/plugins/setup_wizard/assets/src/views/importFacility/SelectSuperAdminAccountForm.vue
@@ -127,7 +127,14 @@
         return FacilityImportResource.facilityadmins()
           .then(admins => {
             this.facilityAdmins = [...admins];
-            this.selected = { ...this.dropdownOptions[0] };
+            // NOTE: We don't have the facility user ID on hand to disambiguate if
+            // a duplicate username is used
+            this.selected =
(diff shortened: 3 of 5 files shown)


kolibri/plugins/setup_wizard/assets/src/views/importFacility/LoadingTaskPage.vue
```diff
@@ -74,21 +74,18 @@
...
     methods: {
-      startPolling() {
+      pollTask() {
         SetupTasksResource.fetchCollection({ force: true }).then(tasks => {
           this.loadingTask = {
             ...tasks[0],
             facility_name: this.facilityName,
           };
         });
-        if (this.loadingTask.status) {
```
jonboiser, line 87: This was the wrong condition, it should have said something like "If status is completed/canceled/failed". Just removing it here, and letting this page poll the endpoint unconditionally.

kolibri/plugins/setup_wizard/assets/src/views/importFacility/SelectFacilityForm.vue
```diff
@@ -143,7 +143,12 @@
         const $credentialsForm = this.$refs.credentialsForm;
         if ($credentialsForm) {
           // The form makes the call to the startpeerfacilityimport endpoint
-          return $credentialsForm.startImport();
+          return $credentialsForm.startImport().then(() => {
```
nucleogenesis: At this point `POST http://localhost:8000/api/tasks/facilitytasks/startpeerfacilityimport/` returns 401 when I pass bad credentials:

`{"detail":"401 Client Error: Unauthorized for url: https://kolibri-beta.learningequality.org/api/morango/v1/certificates/"}`

Then it just gets stuck and I have to refresh the page.

screenshot-localhost_8000-2020 07 21-12_13_33

kolibri/plugins/setup_wizard/assets/src/views/importFacility/SelectSuperAdminAccountForm.vue
```diff
@@ -138,6 +145,9 @@
         this.shouldValidate = false;
         if (this.$refs.password) {
           this.$refs.password.resetAndFocus();
+          if (this.selected.label === this.facility.username) {
```
nucleogenesis: I'm guessing that this is whereabouts this issue would be found in the code.

I get here and the username is defaulted to the username I used in the first place (`learningequality`) to start import and the password was autofilled.

If I go from `learningequality` (I'm testing using the beta server) to `micah` and then go back to `learningequality` - the password field is still filled with the `learningequality` password.

But - if you go:

`learningequality` -> `Create new super admin` -> `learningequality` --- then the password is no longer filled out. If I then go from there to `micah` and back to `learningequality` then the password is filled again.

Seems that this `resetFormAndRefocus()` method isn't being called whenever you switch away from the Create a new super admin form.

nucleogenesis (changes requested):
A few bugs are noted in comments elsewhere and here are a couple more I found:

---

There is a weird bug in the general flow in that a user can create multiple facilities within the Setup Wizard and never even know that's what they're doing.

Definitely an edge case but it isn't hard to run into it:

- Import a facility in the Advanced Setup.
- Get to the page for "sign in as super admin" of the setup wizard.
- Refresh your page, close your browser tab by mistake, etc... you start from the beginning of the Setup Wizard now.
- Go back through and import the same facility - or a new one - and it will be imported as well!

If I'm redirected to "Step 1" of Setup Wizard I (as a user) assume that I am starting from scratch. 

Also, once you get through importing your additional facilit(y/ies) through the setup wizard and log in as the super admin properly - the last page (Responsibilities...) doesn't work when you click "Next".

```
500 ERROR:
MultipleObjectsReturned at /setup_wizard/api/facilityimport/provisiondevice/
get() returned more than one Facility -- it returned 2!

Request Method: POST
Request URL: http://localhost:8000/setup_wizard/api/facilityimport/provisiondevice/
```

---

You are unable to click the "Cancel" button during import for the import task. If this is intentional, I'd suggest hiding it.

---

I also was not able to use my browser's back button once I got to the "Enter super admin credentials" after importing. 

I don't like the idea of hijacking someone's expected browser functionality. I think that if we don't want someone to be able to click back then we should hit them with a browser confirm dialog saying whatever bad things might happen.

Ideally, they won't have bad things happen if they click it, though. They should basically just be taken to the previous meaningful step where they can make a decision about how to setup their device.

This may involve, to some degree, caching or otherwise not committing imported data to the DB until the whole process is completed - finding an alternative to this may also help avoid the issue I listed at the top of this comment.

Not sure what of the above fits into the scope of this PR or our expectations for 0.14, but if they cannot be addressed now let me know and I'll make some Github issues for them.

kolibri/plugins/setup_wizard/assets/src/views/importFacility/SelectFacilityForm.vue
jonboiser: Cool, I found what was causing this 👍

jonboiser:
@nucleogenesis I made this issue #7333  to capture your larger comment. I think it's an okay defect to have for now because it's pretty edge-casey plus easy to undo using the new facility management functionality.

> You are unable to click the "Cancel" button during import for the import task. If this is intentional, I'd suggest hiding it.

This one I'm not sure about. The "cancel" button should be triggering a cancel task API call. In my experience the import happens too fast that either it finishes or passes the point of no return before I click it.

jonboiser:
@nucleogenesis I added fixes for

1. Handling bad credentials at the pre-import authorization step
1. Handling the transition between "create new..." and the saved admin credentials

PR #7335, opened 2020-07-20
Demo server banner & Sign in page routing updates

### LINKED ISSUE #7176: demo banner does not play nicely with login page updates
### Observed behavior

banner state is not preserved and doesn't show at all on some pages

### Expected behavior

* banner should be visible on all sign-in pages
* banner 'closed' state should be persisted
* scroll position should not change

### Steps to reproduce

see beta site

### Context

0.14.0 beta 3

nucleogenesis:
### Summary
 

1. Fixes issue where Demo Server stopped working
2. Fixes scroll jumping issue in User auth flow.
3. Also sneaking in some quick clean up on SignIn state management and fixing a bug with the "Create Account" button I saw where it wouldn't be disabled whenever the form is submitted for Sign In.

### Reviewer guidance

There are 1-3 interesting pages, depending on if you are in a multiple-facility device or not. This really only affects multiple-facility, but **please test a single facility** device as well to be safe.

1. AUTH_SELECT: The page where it says "Sign in if you have an existing account"
2. FACILITY_SELECT: Listing the facilities you may choose from
3. SIGN_IN: Where you sign in with your username and password and where a Learner who needs to create a password may do so.

Below, when I will say "navigate back and forth" those are the pages I'm referring to.

---

Enable the demo_server plugin:
`kolibri plugin enable kolibri.plugins.demo_server`

Then run `kolibri plugin list` to make sure you see  `kolibri.plugins.demo_server         ENABLED`

In an Incognito / Private Browsing / etc window, go to log into your multi-facility device.

Go back and forth throughout the process to make sure:

1. You see the Demo Server banner.
2. When you navigate through the form in any direction, your scroll position does not change. The banner remains opened or closed while you navigate through.
3. You can use back/forward on your browser without losing scroll position.

KNOWN ISSUES
- Going back from Create Account to the previous page is not 100% consistent for me at times. The last commit intends to make that more reliable. I think the UX in that case isn't the worst thing in the world. This PR fixes the really ugly jumping noted in the relevant issue.
- Demo Banner Open/Close state will not persist if you go to Create Account and then back - it always is open on page load.

### References

Fixes https://github.com/learningequality/kolibri/issues/7176

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/core/assets/src/state/modules/core/getters.js
+++ b/kolibri/core/assets/src/state/modules/core/getters.js
@@ -30,10 +30,6 @@ export function pageSessionId(state) {
   return state.pageSessionId;
 }
 
-export function demoBannerVisible(state, getters, rootState) {
-  return state.demoBannerVisible && rootState.pageName === 'SIGN_IN';
-}
-
 export function allowAccess(state, getters, rootState, rootGetters) {
   return state.allowRemoteAccess || rootGetters.isAppContext;
 }
--- a/kolibri/core/assets/src/state/modules/core/mutations.js
+++ b/kolibri/core/assets/src/state/modules/core/mutations.js
@@ -36,7 +36,4 @@ export default {
   CORE_SET_PAGE_VISIBILITY(state, visible) {
     state.pageVisible = visible;
   },
-  SET_CORE_BANNER_VISIBLE(state) {
-    state.demoBannerVisible = true;
-  },
 };
--- a/kolibri/core/assets/src/views/CoreBase/index.vue
+++ b/kolibri/core/assets/src/views/CoreBase/index.vue
@@ -67,7 +67,7 @@
       :class="fullScreen ? 'scrolling-pane' : 'content'"
       :style="contentStyles"
     >
-      <CoreBanner v-if="demoBannerVisible">
+      <CoreBanner v-if="coreBannerComponent && showDemoBanner">
         <template slot-scope="props">
           <component :is="coreBannerComponent" :bannerClosed="props.bannerClosed" />
         </template>
@@ -279,6 +279,11 @@
         required: false,
         default: 1000,
       },
+      showDemoBanner: {
+        type: Boolean,
+        default: false,
+        required: false,
+      },
     },
     data() {
       return {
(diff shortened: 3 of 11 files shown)


jonboiser:
Some things on the SIGN_IN step: ~username isn't being saved~ + spacing on back link. 

CleanShot 2020-07-20 at 16 38 47@2x

~Idk if this is in the design, but I think the case where all facilities disallow learner account sesf-service needs to be handled~
CleanShot 2020-07-20 at 16 41 03@2x

jonboiser:
Some minor stuff:

1. ~The click target for the back link is larger than I expected, it seems to be the full block element containing the link.~
CleanShot 2020-07-20 at 16 51 55@2x

2. This seems to be random, but it seems like you can trigger a submit-like event that validates the password field from pressing the back link. Clicking the "sign in" button triggers the validation always

CleanShot 2020-07-20 at 16 52 14@2x

radinamatic:
> The click target for the back link is larger than I expected, it seems to be the full block element containing the link.

:100: :heavy_plus_sign: for making this smaller so it does not occupy the full width of the element. Noticed it during the Beta12 bugbash, and it's especially painful on small screens (mobile phones) where hitting the target is much more challenging and prone to misses.

(:pray: to @jonboiser, saved me from reporting the issue)

nucleogenesis:
Thanks @jonboiser 

Should have these all fixed early on today including the spacing stuff (which I think will need some KDS tweaks maybe).

nucleogenesis:
@jonboiser the image saying "clicked here and got validation" is a side effect of how our form elements handle validation which is basically "has this been focused and then blurred?".

You arrive to the password field autofocused, so if you click away with an invalid password (empty) then you'll see the validation.

nucleogenesis:
@jonboiser Everything should be resolved except the arrows on the backlinks. I'm going to do that separately because I think there are going to be changes to other parts of the code unrelated to this PR's purpose.

jonboiser:
> @jonboiser the image saying "clicked here and got validation" is a side effect of how our form elements handle validation which is basically "has this been focused and then blurred?".
> 
> 
> 
> You arrive to the password field autofocused, so if you click away with an invalid password (empty) then you'll see the validation.

The issue I was having is that the validation was blocking me from going back when clicking the link maybe 5% of the time. Is there some kind of guard that keeps the user from going forward _and_ backwards if the password is empty?

jonboiser (approved):
Looks good!

micahscopes (commented):
Tested this and it works great on my end!

PR #7357, opened 2020-07-22
fix superuser creation in provisiondevice command

### LINKED ISSUE #7320: provisiondevice creates a superuser to the wrong facility
### Observed behavior

When there is an existing facility on Kolibri instance, running `provisiondevice` command would set up a superuser on the existing facility instead of the new facility it is supposed to created in.
I believe this is because we didn't pass in `facility` when running `create_superuser`: https://github.com/learningequality/kolibri/blob/release-v0.14.x/kolibri/core/device/management/commands/provisiondevice.py#L87

### Expected behavior

superuser should be created in the newly created facility

### User-facing consequences

users are unable to log in to the facility that is created through `provisiondevice` command

### Errors and logs

…

### Steps to reproduce

1. set up kolibri with one facility
2. run `provisiondevice` to set up another facility
3. notice that the facility is created successfully, but the superuser is in the old facility

### Context

Kolibri version: release-0.14.x

lyw07:
### Summary

This PR is to fix the superuser creation when running `provisiondevice` command. It includes
- pass the new facility to `create_superuser` function so we create the superuser in the new facility instead of the default facility
- exit the program when user gets asked about `What do you wish to name your facility?` and user input is `n` 
- directly call `FacilityUser.objects.create_superuser` and catch the exception if the superuser has existed instead of checking the existence of the superuser first. This is because we check the existence of the superuser in the function `FacilityUser.objects.create_superuser()` already. We can avoid making the same query twice in this case.
https://github.com/learningequality/kolibri/blob/release-v0.14.x/kolibri/core/auth/models.py#L595-L596
- when creating the superuser in the new facility, instead of checking whether there are similar usernames using `username__icontains=username`, we check the exact name using `username__iexact=username` in the `FacilityUser.objects.create_superuser()` function (linked above)
This can avoid the issue that when user `ad` exists, we cannot create user `a` as it's similar to `ad`.
(If there is a specific reason that we check similar names instead of the exact name, please feel free to let me know :) thank you!)

### Reviewer guidance

run `kolibri start` and set up facility in the web ui
run `kolibri manage provisiondevice`
please check if running `provisiondevice` would create a superuser on the new facility instead of the default facility

### References

Fixes #7320

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/core/device/management/commands/provisiondevice.py
+++ b/kolibri/core/device/management/commands/provisiondevice.py
@@ -1,8 +1,10 @@
 import json
 import logging
 import os
+import sys
 
 from django.conf import settings
+from django.core.exceptions import ValidationError
 from django.core.management.base import BaseCommand
 from django.core.management.base import CommandError
 from django.db import transaction
@@ -77,6 +79,8 @@ def create_facility(facility_name=None, preset=None, interactive=False):
             facility_name = get_user_response(
                 "What do you wish to name your facility? "
             )
+        else:
+            sys.exit(1)
 
     if facility_name:
         facility, created = Facility.objects.get_or_create(name=facility_name)
@@ -123,7 +127,7 @@ def update_facility_settings(facility, new_settings):
     logger.info("Facility settings updated with {}".format(new_settings))
 
 
-def create_superuser(username=None, password=None, interactive=False):
+def create_superuser(username=None, password=None, interactive=False, facility=None):
     if username is None and interactive:
         username = get_user_response("Enter a username for the super user: ")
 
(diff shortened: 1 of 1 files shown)


jonboiser (approved):
The code looks good to me!

kolibri/core/device/management/commands/provisiondevice.py
```diff
@@ -134,15 +138,17 @@ def create_superuser(username=None, password=None, interactive=False):
...
+        try:
+            FacilityUser.objects.create_superuser(username, password, facility=facility)
             logger.info(
-                "Superuser created with username {username}.".format(username=username)
+                "Superuser created with username {username} in facility {facility}.".format(
+                    username=username, facility=facility
+                )
             )
-        else:
+        except ValidationError:
```
jonboiser, line 148: Is it possible to run `provisiondevice` and run into this situation? I thought it would create a new empty Facility, so there wouldn't be any username collisions.

lyw07: yeah I think it could be, when the user somehow uses `provisiondevice` to create new superuser on existing facility instead of creating a new facility?

jonboiser: Ohh, that's weird. Wasn't there a different command just for making superusers? I think this command is now overloaded to handle both facility creation/provisioning/superusers, so that makes sense to do that check.

rtibbles: There is a `createsuperuser` command that is a django builtin - which I think only creates a superuser on the default facility for the device.

jonboiser: I think a better UX would be to reloop back to let the user pick a different name, but we can add that later.

jonboiser: Oh, I keep forgetting that it's a django built-in command. Yeah, I can see a future version of `provisiondevice` being more of a generic CLI tool for doing stuff you would do on the Setup Wizard + Device pages.

PR #7371, opened 2020-07-23
Set Redis memory policy first and warn about forced cache size decrease

bjester:
### Summary

When Redis has used more memory than the setting configured through Kolibri, it could fail if the policy was `noeviction`. So we'll set the policy first and then warn about the cache size change.

### Reviewer guidance

1. Setup Kolibri to use Redis with generous `maxmemory`
2.  Generate a significant amount of data in Redis
3. Stop Kolibri
3. Configure `maxmemory` of half of `$ redis-cli INFO MEMORY`
4. Start Kolibri
5. Ensure you see a warning about it decreasing
6. Ensure the `maxmemory` setting was configured

### References

Resolves: https://github.com/learningequality/kolibri/pull/7300#issuecomment-663105412

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/core/apps.py
+++ b/kolibri/core/apps.py
@@ -7,6 +7,7 @@ from django.apps import AppConfig
 from django.db.backends.signals import connection_created
 from django.db.utils import DatabaseError
 
+from kolibri.core.logger.utils.data import bytes_for_humans
 from kolibri.core.sqlite.pragmas import CONNECTION_PRAGMAS
 from kolibri.core.sqlite.pragmas import START_PRAGMAS
 from kolibri.core.sqlite.utils import repair_sqlite_db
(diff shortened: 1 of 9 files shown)


indirectlylit (approved):
I don't know this code well enough to give it a proper review, but the tests pass and I think it would be valuable to merge in before the next beta.

@jredrejo if you could still test this, that would be much appreciated

kolibri/core/apps.py
```diff
@@ -124,13 +119,29 @@ def check_redis_settings():
...
 
-            # should only change something when `config_use_conf` is also `False`
-            if helper.changed:
-                helper.save()
-            elif maxmemory == 0 or maxmemory_policy == "noeviction":
+            # default setting is "0", or no limit
+            maxmemory = helper.get_maxmemory()
+            if config_maxmemory > 0 and maxmemory != config_maxmemory:
+                used_memory = helper.get_used_memory()
+                if maxmemory < used_memory:
```
jredrejo, line 126: should not this be `config_maxmemory` ?

kolibri/core/apps.py
```diff
@@ -124,13 +119,29 @@ def check_redis_settings():
...
+                    logger.warning(
+                        "Redis was using {used_memory} before setting `maxmemory` configuration of {config_memory}".format(
+                            used_memory=bytes_for_humans(used_memory),
+                            config_memory=bytes_for_humans(config_maxmemory),
+                        )
+                    )
+
+                helper.set_maxmemory(config_maxmemory)
```
jredrejo, line 134: should not this be after an `else` clausule, so we don't set the new config_maxmemory if we know it's gonna cause problems because it's less than the current used memory?

kolibri/core/apps.py
bjester: Ah yes, not sure how I didn't catch that. Thanks!

kolibri/core/apps.py
bjester: No it's fine because setting the policy first will not trigger the issue you encountered.

PR #7393, opened 2020-07-28
Update permissions for NetworkLocationViewSet et al. so Admins can use p2p sync

### LINKED ISSUE #7380: Non-superuser Facility admins can't use the p2p sync workflow on the Facility > Data Page
### Observed behavior

Non superuser admins do not have the permissions to get NetworkAddresses → Sync Modal does not work for them. Originally this was a superuser permission since it was tied exclusively with content import/export

See in the picture that the call to the endpoint is returning a 403 code -> most likely insufficient permissions

image

### Expected behavior

Non-superuser admins can access this modal. 

Hopefully the task endpoints aren't also locked down like this...

### User-facing consequences

…

### Errors and logs

…

### Steps to reproduce

…

### Context

…

jonboiser:
### Summary

**Main changes**
1. Modifies permissions on `FacilityTaskViewset`, `NetworkLocationViewSet` and `NetworkLocationFacilitiesViewset` to be accessible from Setup Wizard, Device > Facilities page, _and_ Facility > Data (fixes #7380)
1. Cherry picks a commit from @bjester, fixing `permission_classes` on the method-based views.
1. Adds more specific permissions for different methods on the `FacilityTaskViewset`.  (e.g. methods that should be superuser-only were previously available to the lower "can manage content" permission)
1. Fixes bug in Task Manager where device name/ID metadata was not being displayed

**Bonus fixes**
1. Fixes bug in Device > Facilities > Task Manager where button group buttons had a breaking space between them
1. Adds a "failure" event to the SyncInterface's Sync Modal to handle errors that might happen before the sync task is started.

### Reviewer guidance

1. (1) is implemented by creating new Permissions classes that try to be custom tailored to how we want different endpoints are accessible to different roles. Does the implementation of the class include everybody we want to include and exclude no one we don't want to exclude.
1. Will all types of tasks that have a device (import + sync) get the right device metadata on the Task manager?

### References

…

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/core/assets/src/mixins/commonSyncElements.js
+++ b/kolibri/core/assets/src/mixins/commonSyncElements.js
@@ -86,8 +86,10 @@ export default {
       });
     },
     startPeerImportTask(data) {
-      const { facility, facility_name, baseurl, username, password } = data;
+      const { facility, facility_name, baseurl, username, password, device_name, device_id } = data;
       return FacilityTaskResource.startpeerfacilityimport({
+        device_name,
+        device_id,
         facility,
         facility_name,
         baseurl,
--- a/kolibri/core/assets/src/views/sync/FacilityAdminCredentialsForm.vue
+++ b/kolibri/core/assets/src/views/sync/FacilityAdminCredentialsForm.vue
@@ -100,7 +100,6 @@
       startImport() {
         this.shouldValidate = true;
         if (this.formIsValid) {
-          // return Promise.resolve(true);
           return this.startPeerImportTask({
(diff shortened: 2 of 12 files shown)


micahscopes (approved):
I tested this according to @jonboiser's specifications and verified that:
- Facility-Sync works for superusers + Admins,
- Facility Deletion + Import only works for Superusers
- KDP Bulk sync only works for Superusers

PR #7401, opened 2020-07-28
Use `yarn run lint-frontend` script in `frontend-lint` pre-commit hook

### LINKED ISSUE #7032: linting between pre-commit and package.json is inconsistent
### Observed behavior

mentioned by @rtibbles:

> our frontend linting commands in our package.json don't match the way we do linting in our pre-commit hooks/travis linting. Would be good to update those if possible.
> 
> There was an update you had made to the build tools Javascript which was getting linted by travis, but doesn't get linted by default by our package.json commands.

### Expected behavior

linting when running devserver should be consistent with  `pre-commit`, which is used by travis

### User-facing consequences

devs think PRs are good but they turn out not to be

### Steps to reproduce

make a change to build tools. This will be linted by `pre-commit` but not the devserver

### Context

0.14

jonboiser:
### Summary

1. Simplifies the `frontend-lint` pre-commit hook to simply run the `yarn run frontend-lint` script. This makes the output less verbose when it fails.
1. Renames the hook to `lint-frontend` to match the yarn script.

With the current implementation, whenever the hook fails, the command with the expanded `files` regex will appear in the Travis logs

```
$ /Users/jon/Github/kolibri/node_modules/.bin/kolibri-tools lint --write packages/eslint-plugin-kolibri/lib/rules/vue-filename-and-component-name-match.js kolibri/plugins/coach/assets/src/views/plan/LessonSummaryPage/ResourceListTable.vue kolibri/plugins/media_player/assets/src/views/MediaPlayerCaptions/transcriptMenuItem.js packages/kolibri-tools/lib/webpack.config.base.js kolibri/core/assets/src/views/ProgressBar.vue kolibri/plugins/epub_viewer/assets/src/views/NextButton.vue kolibri/plugins/epub_viewer/assets/src/views/TocButton.vue kolibri/core/assets/src/api-resources/channel.js kolibri/core/assets/test/progress-icon.spec.js kolibri/plugins/learn/assets/src/views/ChannelCardGroupGrid.vue kolibri/plugins/coach/assets/src/views/plan/PlanHeader.vue kolibri/plugins/device/assets/src/modules/managePermissions/index.js 
... 50 or so more lines...

<The actual errors>
```

By running the command as is, the only thing logged is the original script being run in node, which does not expand the glob into the terminal. This has the benefits:

1. It's easier to see the problem in the Travis logs (not only is a single `kolibri-tools lint` call pretty verbose, for some reason it's logged 3 times)
1. The pre-commit hook matches the `yarn run frontend-lint` script most people are using, so it's a little more predictable and consistent.

### Reviewer guidance

Does this still basically work the same? Does the "include" glob in the script match the original `files` field. Similarly the `--ignore` glob match the `exclude` regex?

I added one commit at the end to test that it will cause the tox lint env to fail.

### References

…

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -17,7 +17,7 @@ repos:
         language_version: python2.7
 -   repo: local
     hooks:
-    -   id: frontend-lint
+    -   id: lint-frontend
         name: Linting of JS, Vue, SCSS and CSS files
         description: This hook handles all frontend linting for Kolibri
         entry: yarn run lint-frontend
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -20,10 +20,9 @@ repos:
     -   id: frontend-lint
         name: Linting of JS, Vue, SCSS and CSS files
         description: This hook handles all frontend linting for Kolibri
-        entry: yarn run kolibri-tools lint --write
+        entry: yarn run lint-frontend
         language: system
-        files: \.(js|vue|scss|css)$
-        exclude: '^kolibri/core/static/assets/fonts/.+?\.css$|^kolibri/core/content/static/assets/'
+        pass_filenames: false
 - repo: local
   hooks:
   - id: no-auto-migrations
--- a/package.json
+++ b/package.json
@@ -29,7 +29,7 @@
     "bundle-stats": "kolibri-tools build stats --file ./build_tools/build_plugins.txt",
     "clean": "kolibri-tools build clean --file ./build_tools/build_plugins.txt",
     "preinstall": "node ./packages/kolibri-tools/lib/npm_deprecation_warning.js",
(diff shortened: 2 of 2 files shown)


jonboiser:
https://travis-ci.org/github/learningequality/kolibri/jobs/712739100#L1081

Yup the logs at the pre commit step are less verbose now

rtibbles:
Note this also fixes https://github.com/learningequality/kolibri/issues/7032 (but removes linting from the build scripts).

Can you expand `yarn run frontend-lint` to also lint the `build_tools` folder (and everything inside it) as well, then I think this will completely resolve the above issue without reducing the linting.

jonboiser:
@rtibbles I added the `build_tools` dir, so the js file in there is now covered by the `lint-frontend` script.

rtibbles (approved):
As long as Travis says it's ok, it's ok by me!

PR #7412, opened 2020-07-30
Update pre-commit linting to be selective again and write.

rtibbles:
### Summary
* Recent updates to linting, #7401, updated the pre-commit to use `yarn frontend-lint`
* This doesn't write changes to the files, so does not make auto formatting updates when possible
* This also lints every file that matches the glob pattern, so can take a lot of time
* This PR restricts the linting again and reinstates the write behaviour in the pre-commit hook

### Reviewer guidance

Does linting all files work? Does linting individual files work?

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -22,7 +22,6 @@ repos:
         description: This hook handles all frontend linting for Kolibri
         entry: yarn run lint-frontend:format
         language: system
-        files: \.(js|vue|scss|css)$
         pass_filenames: false
 - repo: local
   hooks:
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -20,8 +20,9 @@ repos:
     -   id: lint-frontend
         name: Linting of JS, Vue, SCSS and CSS files
         description: This hook handles all frontend linting for Kolibri
-        entry: yarn run lint-frontend
+        entry: yarn run lint-frontend:format
         language: system
+        files: \.(js|vue|scss|css)$
         pass_filenames: false
 - repo: local
   hooks:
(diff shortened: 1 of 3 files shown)


.pre-commit-config.yaml
```diff
@@ -20,8 +20,9 @@ repos:
     -   id: lint-frontend
         name: Linting of JS, Vue, SCSS and CSS files
         description: This hook handles all frontend linting for Kolibri
-        entry: yarn run lint-frontend
+        entry: yarn run lint-frontend:format
         language: system
+        files: \.(js|vue|scss|css)$
```
jonboiser: I thought that by not setting `files` and having `pass_filesnames: false`, that the hook would just run `entry` once like we normally do when we use the npm script.

Does this change end up running `entry` for every file matching the pattern, like `yarn run lint-frontend:format` (x 1000 times)?

rtibbles: Possibly, in which case I should activate `pass_filenames`.

Noting that I did this because after merging the earlier PR, I noticed that my pre-commit step was taking a very long time, even with Python files which shouldn't be touched by this hook at all.

Will investigate further here.

jonboiser: This is the config we _don't_ want. I ran the hook with `verbose : true` to see what was happening

```
    hooks:
    -   id: lint-frontend
        name: Linting of JS, Vue, SCSS and CSS files
        entry: yarn run lint-frontend:format
        language: system
        files: \.(js|vue|scss|css)$
        pass_filenames: true <- will call a `yarn run lint-frontend --write <expanded regex> for each file passing regex
        verbose: true
```

But I don't think adding the `files` field changes anything as long as `pass_filenames: false`. It just calls `lint-frontend:format` once, regardless of the `files` regex

jonboiser:
```suggestion
```

jonboiser (approved):

PR #7425, opened 2020-08-02
Update tzlocal to fix python 3.8 syntax warnings

arky:
This patch upgrade the tzlocal package so to prevent issues on machines running latest python 3.8 like Ubuntu LTS and Rasbian 

This minor patch resolves one of the errors during the Kolibri install as described in the bug https://github.com/learningequality/kolibri-server/issues/78


jredrejo:
Closing it to make it compatible with 0.14. New PR created at  #7428 
Thanks @arky

PR #7432, opened 2020-08-04
Prevent runtime error in UserType.js if FacilityUser has an incomplete set of Roles

### LINKED ISSUE #7419: facility user list: Cannot read property 'kind' of undefined
### Observed behavior

facility page is stuck here:

image

### Errors and logs

```
UserType.js:12 Uncaught (in promise) TypeError: Cannot read property 'kind' of undefined
    at UserType (UserType.js:12)
    at _userState (mappers.js:16)
    at Array.map (<anonymous>)
    at handlers.js:17
    at conditionalPromise.js:36
```

### Steps to reproduce

Unknown but I can repro locally. See me for the details

### Context

0.14.0 beta 14

jonboiser:
### Summary

Adds some handling to this `UserType` utility if some potentially-faulty assumptions about `FacilityUsers` and `Roles` are not true. Namely, in the unusual situation where a User has a `COACH` role for a Classroom-level `Collection`, but not a corresponding `ASSIGNABLE_COACH` role for the Facility, then we still display that user as an assignable coach on the Facility > Users page.

And more importantly, the whole page won't crash.

If there is some other unknown way a user somehow has a Role for a different facility, then there is a catch-all branch to just annotate them as a Learner.

I'll follow up with an issue to implement this business rule (you must have an `ASSIGNABLE` coach role if you have a Classroom-level `COACH` role) on the services layer so we can be more consistent and remove some of the code on the front-end (e.g. `UserType.js` and `updateFacilityLevelRoles` in `userManagement/utils.js`) that tries to implement it.

### Reviewer guidance

1. I re-created the scenario of having a Classroom-`COACH` role without the Facility-`ASSIGNABLE_COACH` role by making a coach, then deleting their `ASSIGNABLE_COACH` role in the DB. This should cause this page to crash, per #7419. With this PR, it will not crash and display the user as a "Coach".
1. Any other situation where the user has `ADMIN` or `ASSIGNABLE_COACH` for a Classroom or a completely different Facility would not make sense in normal usage, but should be handled by displaying the user as a Learner.

### References

…

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/core/assets/src/utils/UserType.js
+++ b/kolibri/core/assets/src/utils/UserType.js
@@ -9,5 +9,19 @@ export default function UserType(userObject) {
   }
 
   // get first role associated with this facility
-  return userObject.roles.find(role => role.collection === userObject.facility).kind;
+  const firstRole = userObject.roles.find(role => role.collection === userObject.facility);
+
+  if (firstRole) {
+    return firstRole.kind;
+  } else {
+    // HACK check for edge case where the User has a Classroom-level COACH role in
+    // spite of not having a Facility-level ASSIGNABLE_COACH role
+    const coachRole = userObject.roles.find(role => role.kind === UserKinds.COACH);
+    if (coachRole) {
+      return UserKinds.ASSIGNABLE_COACH;
+    } else {
+      // If all else fails, display user as a LEARNER
+      return UserKinds.LEARNER;
+    }
+  }
 }


rtibbles (approved):
Adds appropriate safe guards, hard to see how this could continue to produce an error without some horrifyingly scrambled data.

PR #7433, opened 2020-08-04
Pass filenames to only lint changed files in pre-commit.

rtibbles:
### Summary
Pre-commit is still linting all files, rather than just changed files.

Makes pre-commit pass filenames again to make this happen.

### Reviewer guidance
Does Travis still pass?

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -22,6 +22,7 @@ repos:
         description: This hook handles all frontend linting for Kolibri
         entry: yarn run lint-frontend:format
         language: system
+        files: \.(js|vue|scss|css)$
 - repo: local
   hooks:
   - id: no-auto-migrations
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -22,7 +22,6 @@ repos:
         description: This hook handles all frontend linting for Kolibri
         entry: yarn run lint-frontend:format
         language: system
-        pass_filenames: false
 - repo: local
   hooks:
   - id: no-auto-migrations


jonboiser:
How are you running `pre-commit`, or how is it run on Travis?

When I run `pre-commit run --all-files` and add `verbose: true` to that hook, it makes it seem as if it this is equivalent to running `yarn run lint-frontend:format` once, with the current config.

jonboiser:
Oh, I see. If you use `pre-commit` after each commit you're not running it with the `--all-files` flag like it's used in Travis.

jonboiser:
I think my assumptions about how this all worked in #7401 are wrong. It's more like if you want to use `pre-commit` as part of your git workflow, you want the hook that just runs the changed files into the `kolibri-tools lint` command.

But in the context of Travis CI, I would prefer to have _no_ filenames passed into the command so that the logs that result during a failure are less verbose. I don't know if you can have both of these.

.pre-commit-config.yaml
```diff
@@ -22,7 +22,6 @@ repos:
         description: This hook handles all frontend linting for Kolibri
         entry: yarn run lint-frontend:format
         language: system
```
jonboiser, line 24: Need to add this regex back if we're passing in filenames
```suggestion
        language: system
        files: \.(js|vue|scss|css)$
```

jonboiser: I think we need to go back to the drawing board to figure out how to have cleaner logs in Travis when this hook fails.

rtibbles:
Yeah - although I think we don't get as messy output from our other hooks, right? So if it's just the linting command that is problematic, we can fix that up to try to make something less messy.

Do you have a specific example of the messy output that we could look at more to try to make this better?

jonboiser:
Something like this is what I was trying to fix earlier:

https://travis-ci.org/github/learningequality/kolibri/jobs/711456164#L1082

At Line 1082 it logs the full command, which goes on for a few screenfuls of file paths, then finally to the actual linting error.

indirectlylit:
relevant recent PR: https://github.com/learningequality/kolibri/pull/6583

rtibbles:
Hrm - do other linting hooks suffer from the same issue, or is it just ours?

jonboiser (approved):

jonboiser:
The 'import order' output is very concise when it "fails". Maybe we can see what is different about these community plugins to see if we can make the linting hook less verbose. But I'll merge this for now to improve people's git workflows.

```
Trim Trailing Whitespace.................................................Passed
Flake8...................................................................Passed
Check Yaml...........................................(no files to check)Skipped
Check for added large files..............................................Passed
Debug Statements (Python)................................................Passed
Fix End of Files.........................................................Passed
Reorder python imports...................................................Failed
hookid: reorder-python-imports

Files were modified by this hook. Additional output:

Reordering imports in kolibri/core/api.py <- that's it, it doesnt output anything else

Linting of JS, Vue, SCSS and CSS files...................................
```

Issue #7283, opened 2020-07-09
setting doesn't get disabled in valid state


indirectlylit:
### Observed behavior

The second checkbox is disabled in whatever state the user left it in:

![2020-07-09 12 05 06](https://user-images.githubusercontent.com/2367265/87086977-2c1e2280-c1e7-11ea-8754-2c7d36a3be3a.gif)

### Expected behavior

The checkbox should be disabled in the only state valid: learners cannot edit their password when they don't have a password

### Steps to reproduce

see facility settings

### Context

0.14.0 beta 8

alexMet:
@jonboiser This issue has been resolved and it works in the "develop" branch, but not in the "release-v0.14.x".

jonboiser:
Oh yeah, we should have closed this out! I think I'll have to cherry-pick this one to release-v0.14.x as well so it gets in the next release.

jonboiser:
@alexMet Have you checked out some of the issues in the [0.14.4 milestone](https://github.com/learningequality/kolibri/milestone/60)? Most of them are frontend issues similar to the ones you've fixed recently.

alexMet:
@jonboiser Oh sorry, i didn't see that you mentioned me. I was working on #7475 and i found the bug, but i don't know how to fix it. So i picked the other issues, that we are talking about.

PR #7446, opened 2020-08-06
pr/7283 setting doesn't get disabled in valid state

mdctleo:
### Summary
setting doesn't get disabled in valid state #7283 

Aug-06-2020 01-03-15

### Reviewer Checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/plugins/facility/assets/src/views/FacilityConfigPage/index.vue
+++ b/kolibri/plugins/facility/assets/src/views/FacilityConfigPage/index.vue
@@ -55,7 +55,12 @@
                 :key="setting"
                 :label="$tr('learnerNeedPasswordToLogin')"
                 :checked="!settings['learner_can_login_with_no_password']"
-                @change="toggleSetting('learner_can_login_with_no_password')"
+                @change="() => {
+                  toggleSetting('learner_can_login_with_no_password')
+                  settings['learner_can_edit_password'] &&
+                    !settings['learner_can_login_with_no_password'] ?
+                      toggleSetting('learner_can_edit_password') : null
+                }"
               />
               <KCheckbox
                 :key="setting + 'learner_can_edit_password'"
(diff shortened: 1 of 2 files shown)


mdctleo:
I can't seem to change label to work-in-progress for this pull request, perhaps due to permission

indirectlylit:
@mdctleo I've added the work-in-progress label

thanks for your contribution, and let us know when it's ready for review

mdctleo:
@indirectlylit I noticed that the original issue is tagged with frontend:dev, is backend validation necessary for this change? Thanks!

mdctleo:
@indirectlylit bumping this for review

kolibri/plugins/facility/assets/src/views/FacilityConfigPage/index.vue
```diff
@@ -55,13 +55,19 @@
                 :key="setting"
                 :label="$tr('learnerNeedPasswordToLogin')"
                 :checked="!settings['learner_can_login_with_no_password']"
-                @change="toggleSetting('learner_can_login_with_no_password')"
+                @change="() => {
+                  toggleSetting('learner_can_login_with_no_password')
+                  settings['learner_can_edit_password'] &&
+                    !settings['learner_can_login_with_no_password'] ?
+                      toggleSetting('learner_can_edit_password') : null
+                }"
```
indirectlylit: let's move this to be a component method in the `script` section so we don't pack multiple lines of JS into a `template` attribute

indirectlylit (commented):
Thanks @mdctleo

By the way if you aren't in there, feel free to add yourself to https://github.com/learningequality/kolibri/blob/release-v0.14.x/AUTHORS.md

>  I noticed that the original issue is tagged with frontend:dev, is backend validation necessary for this change?

The tags are definitely not strict.

If you'd like to adding back-end validation (and maybe even a unit test) in addition, that would be great! It's always better to enforce consistency in the API when possible. However the system should also not crash if inconsistent values somehow got set in an older API or DB.

kolibri/plugins/facility/assets/src/views/FacilityConfigPage/index.vue
mdctleo: @indirectlylit thanks for the feedback! I think the backend validation is mentioned here #7281, I will take a look at it in a separate pr.

jonboiser (approved):

PR #7472, opened 2020-08-12
Add `KOptionalText` component to replace `<KEmptyPlaceholder v-else >` pattern

jonboiser:
### Summary

`KEmptyPlaceholder` is always used with a `v-else` directive next to another text block.

This `KOptionalText` component will either take a `text` prop or a `template` in the default slot and then either display the text in the prop/slot or the `KEmptyPlacehlder` if it's an empty string.

### Reviewer guidance

1. Does this refactor still work for all the current usages of `KEmptyPlaceholder`. 
1. Is it extensible for any future usages of this pattern.
1. Can we/should we make it smarter so we can be a little looser about accepting `text` props that are null or undefined, rather than preferring it strictly be a string?

### References

…

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/core/assets/src/core-app/index.js
+++ b/kolibri/core/assets/src/core-app/index.js
@@ -22,6 +22,7 @@ import KThemePlugin from 'kolibri-design-system/lib/KThemePlugin';
 import heartbeat from 'kolibri.heartbeat';
 import KContentPlugin from 'kolibri-design-system/lib/content/KContentPlugin';
 import KSelect from '../views/KSelect';
+import KOptionalText from '../views/KOptionalText';
 import { i18nSetup, languageDirection } from '../utils/i18n';
 import ContentRendererErrorComponent from '../views/ContentRenderer/ContentRendererError';
 import apiSpec from './apiSpec';
@@ -86,6 +87,7 @@ Vue.use(KContentPlugin, {
 });
 
 Vue.component('KSelect', KSelect);
+Vue.component('KOptionalText', KOptionalText);
 
 // Start the heartbeat polling here, as any URL needs should be set by now
 heartbeat.startPolling();
--- a/kolibri/core/assets/src/views/ElapsedTime.vue
+++ b/kolibri/core/assets/src/views/ElapsedTime.vue
@@ -1,6 +1,6 @@
 <template>
 
-  <KOptionalText :text="data ? $formatRelative(ceilingDate, { now }) : ''" />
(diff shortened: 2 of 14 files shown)


indirectlylit:
this is cool.

* Are there instances of `KEmptyPlaceholder` that _don't_ follow this pattern?
* If not, should we just replace `KEmptyPlaceholder`?
* Should this be in the design system?

jonboiser:
> * Are there instances of `KEmptyPlaceholder` that _don't_ follow this pattern?

Nah, `KEmptyPlaceholder` is never just used by itself

> * If not, should we just replace `KEmptyPlaceholder`?

This makes sense to me. I think `KOptionalText` better represents what we're trying to do (show text OR a "--", but never just a blank space). If we replace it, I would prefer to have a smarter version than what's here, where it knows how to handle all kinds of falsy values of the `text` prop or the content of the default slot and can also handle things like styling.

> * Should this be in the design system?

The smarter version I describe above should be in the design system, I think.

nucleogenesis:
This is rad - will love to see a smarter version in KDS!

nucleogenesis (commented):
This LGTM - but I have a few thoughts. I found this in my "Review Requested" lol from a long time ago but gave it another review as if it were fresh :-D 

I don't really see a need for slots here. Slots seem most useful when rendering another component or other HTML within another component. To use them to pass a string around seems off. The places where you use slots could be done more cleanly IMO using a computed property.

Perhaps we can remove the `type` on the prop - but instead use a custom validator that checks if the input is a string, prints a warning if it isn't that says "we're putting KEmtpyPlaceholder in this component where you gave us not a string" but then we can accept `null` or `undefined` values as well.

Anyway - great work - happy to chat more about it synchronously as well.

indirectlylit:
This is also an interesting case study for the KDS release process:

* minor release – OK to add `KOptionalText` because it doesn't break existing API
* major release – OK to deprecate `KEmptyPlaceholder` and subsume functionality into `KOptionalText` – possibly with a deprecation warning rather than total breakage?

nucleogenesis:
KOptionalText implements KEmptyPlaceholder it just gives a component that people can use with a better pattern so we wouldn't deprecate it I don't think - although I wouldn't mind making it kind of like a private internal KDS implementation detail anyway.

indirectlylit:
migrated to https://github.com/learningequality/kolibri-design-system/issues/285

PR #7486, opened 2020-08-19
Provision device automatically using automatic_provision.json

dylanmccall:
### Summary

This makes it easier to create a ready-to-use Kolibri instance with
special configuration, such as adding a default admin user and changing
the landing page. For example, a wrapper application could prepopulate
KOLIBRI_HOME with the following _automatic_provision.json_:

    {
        "facility": "Example",
        "superusername" : "admin",
        "superuserpassword" : "admin",
        "preset" : "nonformal",
        "facility_settings": {},
        "device_settings": {
            "landing_page" : "learn"
        }
    }

This way, we skip Kolibri's initial setup screen and have a default working facility out of the box. This is useful for an environment where we aren't concerned about how someone sets up Kolibri; we're just using it to browse existing content offline, although leaving the door open to create an account and access more functionality.

### Reviewer guidance

I am uncertain if this is the best place for this functionality. It is a bit of a hack to support a feature we need for Endless's _USB Internet_ project and I think it is generally useful (as well as much easier to deal with if we do it in Kolibri itself). Some second opinions about the notion of doing this would be appreciated, as well as any pointers for how to make it more robust or easier to maintain.

### References

None at the moment.

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/utils/cli.py
+++ b/kolibri/utils/cli.py
@@ -11,10 +11,12 @@ from __future__ import print_function
 from __future__ import unicode_literals
 
 import importlib
+import json
 import logging
 import os
 import signal
 import sys
+from collections import Mapping
 from sqlite3 import DatabaseError as SQLite3DatabaseError
 
 import click
@@ -30,6 +32,7 @@ from . import sanity_checks
 from . import server
 from .debian_check import check_debian_user
 from .system import become_daemon
+from kolibri.core.device.utils import device_provisioned
 from kolibri.core.deviceadmin.utils import IncompatibleDatabase
 from kolibri.core.upgrade import matches_version
 from kolibri.core.upgrade import run_upgrades
@@ -53,6 +56,11 @@ from kolibri.utils.sanity_checks import DatabaseNotMigrated
 from kolibri.utils.system import symlink_capability_check
 
 
+try:
+    FileNotFoundError
+except NameError:
+    FileNotFoundError = IOError
+
 logger = logging.getLogger(__name__)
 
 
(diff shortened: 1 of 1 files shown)


kolibri/utils/cli.py
```diff
@@ -190,6 +200,31 @@ def _migrate_databases():
...
+        options = None
+
+    if isinstance(options, Mapping):
+        options.setdefault("superusername", None)
+        options.setdefault("superuserpassword", None)
+        options.setdefault("preset", "nonformal")
+        options.setdefault("language_id", None)
+        options.setdefault("facility_settings", {})
+        options.setdefault("device_settings", {})
+        call_command("provisiondevice", interactive=False, **options)
```
indirectlylit, line 225: maybe `provisiondevice.json` should be deleted after success, because presumably it becomes a "dead" file after provisioning?

kolibri/utils/cli.py
```diff
@@ -63,6 +71,8 @@
 # https://github.com/PythonCharmers/python-future/issues/22
 click.disable_unicode_literals_warning = True
 
+provisiondevice_file = os.path.join(KOLIBRI_HOME, "provisiondevice.json")
```
indirectlylit: maybe `automatic_provision_values.json`?

indirectlylit (commented):
haven't tested, but looks good to me conceptually!

indirectlylit:
somewhat related PR: **parameters to device provisioning command for arbitrary settings #7282**

kolibri/utils/cli.py
dylanmccall: Sure! I added that :)

kolibri/utils/cli.py
dylanmccall: Okay, I renamed it to `automatic_provision.json`, which does indeed sound a bit clearer about what it'll do.

dylanmccall:
This is (probably) superseded by kolibri-installer-gnome#9 and kolibri-installer-mac#65.

PR #7489, opened 2020-08-25
pr/7473 Login flow looks like it accepts any username despite the error validation

### LINKED ISSUE #7473: Login flow looks like it accepts any username despite the error validation
### Observed behavior
I am able to enter any kind of special characters I want for the username and am allowed to press the "NEXT" button which makes it look like my username was acccepted

character bug

Screen Shot 2020-08-12 at 1 01 43 PM

### Expected behavior
I should not be able to press press the next button until I resolved the error

### User-facing consequences
Misleading error messaging, weird characters entered.

### Steps to reproduce
- Go to login form
- Enter anything for the username

### Context
Tell us about your environment, including:
 * Kolibri version: kolibribeta
 * Operating system: MacOS
 * Browser: Chrome, firefox

mdctleo:
Issue #7473 

Now next button is disabled when entering invalid characters

Screen Shot 2020-08-24 at 7 06 16 PM

### DIFF
--- a/kolibri/plugins/user/assets/src/views/SignInPage/index.vue
+++ b/kolibri/plugins/user/assets/src/views/SignInPage/index.vue
@@ -91,7 +91,7 @@
               class="login-btn"
               :text="$tr('nextLabel')"
               :primary="true"
-              :disabled="busy"
+              :disabled="!isNextValid"
               @click="signIn"
             />
           </div>
@@ -334,6 +334,9 @@
         return (
           this.hasMultipleFacilities || get(this.selectedFacility, 'dataset.preset') !== 'informal'
         );
+      },
+      isNextValid() {
+        return !this.busy && validateUsername(this.username);
       },
     },
     watch: {
--- a/kolibri/plugins/user/assets/src/views/SignInPage/index.vue
+++ b/kolibri/plugins/user/assets/src/views/SignInPage/index.vue
@@ -336,7 +336,7 @@
         );
       },
       isNextButtonEnabled() {
-        return !this.busy && validateUsername(this.username);
+        return !this.busy && this.username !== '' && validateUsername(this.username);
       },
     },
     watch: {
@@ -498,6 +498,10 @@
         this.showDropdown = false;
       },
       signIn() {
+        if (!this.isNextButtonEnabled) {
+          return;
+        }
+
         this.busy = true;
 
         const sessionPayload = {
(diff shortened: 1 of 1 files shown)


kolibri/plugins/user/assets/src/views/SignInPage/index.vue
```diff
@@ -91,7 +91,7 @@
               class="login-btn"
               :text="$tr('nextLabel')"
               :primary="true"
-              :disabled="busy"
+              :disabled="!isNextValid"
```
jonboiser: The button should be disabled if `busy` or `!isNextValid`
```suggestion
              :disabled="busy || !isNextValid"
```

kolibri/plugins/user/assets/src/views/SignInPage/index.vue
```diff
@@ -335,6 +335,9 @@
           this.hasMultipleFacilities || get(this.selectedFacility, 'dataset.preset') !== 'informal'
         );
       },
+      isNextValid() {
```
jonboiser: Related to the above comment, I think I would just keep the `busy` part out of this property, or maybe rename it so it's more general, something  to the effect of "next button is enabled"

kolibri/plugins/user/assets/src/views/SignInPage/index.vue
```diff
@@ -91,7 +91,7 @@
               class="login-btn"
               :text="$tr('nextLabel')"
               :primary="true"
-              :disabled="busy"
+              :disabled="!isNextButtonEnabled"
```
jonboiser, line 94: You'll need to put a guard in the `signIn` method as well to prevent a submit event on the kwyboard. Even though the "next" button is disabled, I can still press "Enter" on my keyboard and move forward with an invalid username like `jb!!`

kolibri/plugins/user/assets/src/views/SignInPage/index.vue
```diff
@@ -484,6 +487,10 @@
           this.$refs.username.focus();
         }
       },
+      handleUsernameInput() {
```
jonboiser, line 490: I added this so that the `invalidText` under the textbox would appear immediately, since the only other way would be to blur the textbox now that the form and next button are disabled..

jonboiser (approved):
Looks good! Thank you!

kolibri/plugins/user/assets/src/views/SignInPage/index.vue
mdctleo: alright make sense, thank you for the help!

PR #6004, opened 2019-10-29
Fix python black issues in develop

bjester:
### Summary

A new version of `black` was released yesterday. This makes fixes for incorrectly formatted code for the new version.
https://github.com/psf/black/releases/tag/19.10b0

### Reviewer guidance

- Review the changes
- Does the `pythonblack` suite complete successfully?

----

### Contributor Checklist

PR process:

- [ ] PR has the correct target branch and milestone
- [ ] PR has 'needs review' or 'work-in-progress' label
- [ ] If PR is ready for review, a reviewer has been added. (Don't use 'Assignees')
- [ ] If this is an important user-facing change, PR or related issue has a 'changelog' label
- [ ] If this includes an internal dependency change, a link to the diff is provided

Testing:

- [ ] Contributor has fully tested the PR manually
- [ ] If there are any front-end changes, before/after screenshots are included
- [ ] Critical user journeys are covered by Gherkin stories
- [ ] Critical and brittle code paths are covered by unit tests

### Reviewer Checklist

- Automated test coverage is satisfactory
- PR is fully functional
- PR has been tested for [accessibility regressions](http://kolibri-dev.readthedocs.io/en/develop/manual_testing.html#accessibility-a11y-testing)
- External dependency files were updated if necessary (`yarn` and `pip`)
- Documentation is updated
- Contributor is in AUTHORS.md

PR #7530, opened 2020-09-24
Update pre-commit config for `black` hook, remove `black-fmt` dependency

jonboiser:
### Summary

I was noticing that running the `black` CLI didn't match what was happening when running `pre-commit`, and I think it's because the `repo` field for the `black` hook should be a `git://` address.

This changes that config, and on subsequent `pre-commit run --all-files`, it downloads the stable version and applies the formatting as of version 20 or so of `black`.

This also removes the npm scripts for `fmt-backend` as its use of `black-fmt` is not consistent with the pre-commit behavior and might be confusing.

### Reviewer guidance

…

### References

…

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/.buildkite/create_integration_testing_worksheet.py
+++ b/.buildkite/create_integration_testing_worksheet.py
@@ -114,9 +114,9 @@ def create_artifact(str_arg):
 
 def fetch_feature_files():
     """
-        Fetch all the .features scenarios at the Integration testing directory
-        The order of the scenarios at the spreadsheet will based on the ORDER.txt
-        This will return a list and a count of .features scenarios
+    Fetch all the .features scenarios at the Integration testing directory
+    The order of the scenarios at the spreadsheet will based on the ORDER.txt
+    This will return a list and a count of .features scenarios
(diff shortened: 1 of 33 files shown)


indirectlylit (approved):
We've run into issues before where black gets upgraded in one place (e.g. on a dev's machine) and not in another (e.g. on CI). See for example: https://github.com/learningequality/kolibri/pull/6004

Here's another related PR: https://github.com/learningequality/kolibri/pull/6583

This change is fine with me, but just flagging that as long as black isn't pinned, there's a chance we might continue to run into these kinds of incompatible-formatting issues...

jonboiser:
Yeah, let's see how this works out. I can follow up on this by adding a pinned version of `black` to `requirements/dev.txt` and making sure it's the same tag in the pre-commit config.

rtibbles:
Would be nice if we could do linting commits that did not 'shift the blame'.

PR #7531, opened 2020-09-24
Apply pool_pre_ping changes to 0.12.x

aronasorman:
Applying #7522 changes to 0.12.x so we can deploy it sooner for Cloud Kolibri, which uses 0.12.


rtibbles:
This is missing the second commit from that PR - so you may get build failures as a result.

aronasorman:
Ah thanks @rtibbles! Added it now

rtibbles:
Ugh - linting rules are different between branches, so this has failed because of that. Just need to run `yarn run fmt-backend` on it.

rtibbles (approved):

PR #7535, opened 2020-09-29
Add snackbar notifications when saving facility settings

### LINKED ISSUE #7134: 'save' notification in facility settings should use snackbar
### Observed behavior

Here we use an old-style notification:

image

### Expected behavior

Should use a pop-up, transient snackbar like user-creation:

image

### User-facing consequences

* users on small screens might not see notification if they are scrolled too far down
* ui alert shifts screen contents
* inconsistent style of user feedback

### Context

0.14.0 beta 3

alexMet:
### Summary

Replace the facility settings old-style alert with snackbar notifications.

Screenshot_2020-09-30 Configure Facility - Kolibri

### Reviewer guidance

I haven't accounted for pageLoadFailure. I don't know how to handle that case.

### References

#7134 

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/plugins/facility/assets/src/modules/facilityConfig/actions.js
+++ b/kolibri/plugins/facility/assets/src/modules/facilityConfig/actions.js
@@ -1,7 +1,5 @@
 import { FacilityDatasetResource, FacilityResource } from 'kolibri.resources';
 
-import { notificationTypes } from '../../constants';
-
 export function saveFacilityName(store, payload) {
   return FacilityResource.saveModel({
     id: payload.id,
@@ -25,7 +23,6 @@ export function saveFacilityName(store, payload) {
 }
 
 export function saveFacilityConfig(store) {
-  store.commit('CONFIG_PAGE_NOTIFY', null);
   const { facilityDatasetId, settings } = store.state;
   const resourceRequests = [
     FacilityDatasetResource.saveModel({
@@ -33,15 +30,9 @@ export function saveFacilityConfig(store) {
       data: settings,
     }),
   ];
-  return Promise.all(resourceRequests)
-    .then(function onSuccess() {
-      store.commit('CONFIG_PAGE_NOTIFY', notificationTypes.SAVE_SUCCESS);
-      store.commit('CONFIG_PAGE_COPY_SETTINGS');
-    })
-    .catch(function onFailure() {
-      store.commit('CONFIG_PAGE_NOTIFY', notificationTypes.SAVE_FAILURE);
-      store.commit('CONFIG_PAGE_UNDO_SETTINGS_CHANGE');
-    });
+  return Promise.all(resourceRequests).then(function onSuccess() {
+    store.commit('CONFIG_PAGE_COPY_SETTINGS');
+  });
 }
 
 export function resetFacilityConfig(store) {
--- a/kolibri/plugins/facility/assets/src/views/FacilityConfigPage/ConfigPageNotifications.vue
+++ b/kolibri/plugins/facility/assets/src/views/FacilityConfigPage/ConfigPageNotifications.vue
@@ -1,58 +1,16 @@
-<template>
-
-  <div>
-    <UiAlert
-      v-if="notification === notificationTypes.PAGELOAD_FAILURE"
-      type="error"
-      @dismiss="dismiss()"
-    >
-      {{ $tr('pageloadFailure') }}
-    </UiAlert>
-
-    <UiAlert
-      v-if="notification === notificationTypes.SAVE_SUCCESS"
-      type="success"
-      @dismiss="dismiss()"
-    >
-      {{ $tr('saveSuccess') }}
-    </UiAlert>
-
-    <UiAlert
-      v-if="notification === notificationTypes.SAVE_FAILURE"
-      type="error"
-      @dismiss="dismiss()"
-    >
-      {{ $tr('saveFailure') }}
-    </UiAlert>
-  </div>
-
-</template>
-
-
 <script>
 
-  import UiAlert from 'kolibri-design-system/lib/keen/UiAlert';
-  import { notificationTypes } from '../../constants';
+  // TODO: This component is kept for $trs strings, delete on 0.15 version
+  // Strings should be moved to FacilityConfigPage/index.vue and translations updated
 
   export default {
     name: 'ConfigPageNotifications',
-    components: { UiAlert },
-    props: {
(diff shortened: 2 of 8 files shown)


kolibri/plugins/facility/assets/src/views/FacilityConfigPage/ConfigPageNotifications.vue
```diff
@@ -1,58 +1,16 @@
...
-  </div>
-
-</template>
-
-
 <script>
 
-  import UiAlert from 'kolibri-design-system/lib/keen/UiAlert';
-  import { notificationTypes } from '../../constants';
+  // TODO: This component is kept for $trs strings, delete on 0.15 version
```
jonboiser, line 3: Nice, this is exactly what we've had to do a number of times because it's hard to move translated messages around 👍

kolibri/plugins/facility/assets/src/modules/facilityConfig/actions.js
```diff
@@ -25,23 +23,16 @@ export function saveFacilityName(store, payload) {
...
-  return Promise.all(resourceRequests)
-    .then(function onSuccess() {
-      store.commit('CONFIG_PAGE_NOTIFY', notificationTypes.SAVE_SUCCESS);
-      store.commit('CONFIG_PAGE_COPY_SETTINGS');
-    })
-    .catch(function onFailure() {
-      store.commit('CONFIG_PAGE_NOTIFY', notificationTypes.SAVE_FAILURE);
-      store.commit('CONFIG_PAGE_UNDO_SETTINGS_CHANGE');
-    });
+  return Promise.all(resourceRequests).then(function onSuccess() {
```
jonboiser, line 33: The test failure [here](https://travis-ci.org/github/learningequality/kolibri/jobs/731420697#L7768) might be because you moved this logic to `FacilityConfigPage`. Do you think you'd be able to update the test (`facilityConfig.spec.js`). To run the Jest test runner, run the command `yarn run test` in the console.

kolibri/plugins/facility/assets/src/views/FacilityConfigPage/index.vue
```diff
@@ -218,23 +216,30 @@
...
         }
       },
-      dismissNotification() {
-        this.$store.commit('facilityConfig/CONFIG_PAGE_NOTIFY', null);
+      updateSettings(action) {
+        this.$store
+          .dispatch(action)
+          .then(() => {
+            // eslint-disable-next-line kolibri/vue-no-undefined-string-uses
+            this.createSnackbar(this.NotificationsStrings.$tr('saveSuccess'));
```
jonboiser, line 224: I think this change might be causing the test failure [here](https://travis-ci.org/github/learningequality/kolibri/jobs/731420697#L7788), because we're calling the `$store.dispatch` mock in the test a second time here in `createSnackbar` (under the hood, since the `createSnackbar` method created by `mapActions(['createSnackbar'])`is a shorthand for `store.dispatch('createSnackbar', args)`

jonboiser: I think it would be fine to update the assertion that the mock was called twice, and add a new assertion to say the mock was called with the snackbar action.

kolibri/plugins/facility/assets/src/modules/facilityConfig/actions.js
alexMet: The reason the test fails, is because the Promise throws an exception, when it fails and we can't catch it, because the component handles the exception. But if we catch the exception here, then we won't be able to call the snackbar in the component.

kolibri/plugins/facility/assets/test/state/facilityConfig.spec.js
```diff
@@ -139,15 +138,13 @@ describe('facility config page actions', () => {
       return store.dispatch('facilityConfig/saveFacilityConfig').then(() => {
         expect(DatasetStub.getModel).toHaveBeenCalledWith(1000, {});
         expect(saveStub).toHaveBeenCalledWith(expect.objectContaining(expectedRequest), false);
-        expect(store.state.facilityConfig.notification).toEqual('SAVE_SUCCESS');
       });
     });
 
     it('when save fails', () => {
       const saveStub = DatasetStub.__getModelSaveReturns('heck no', true);
       return store.dispatch('facilityConfig/saveFacilityConfig').then(() => {
```
jonboiser: This test needs to continue off the `catch` because the action doesn't handle errors anymore.

```suggestion
      return store.dispatch('facilityConfig/saveFacilityConfig').catch(() => {
```

jonboiser:
I added a commit to fix the tests

kolibri/plugins/facility/assets/src/views/FacilityConfigPage/index.vue
```diff
@@ -218,23 +216,30 @@
...
+            this.copySettings();
+          })
+          .catch(() => {
+            // eslint-disable-next-line kolibri/vue-no-undefined-string-uses
+            this.createSnackbar(this.NotificationsStrings.$tr('saveFailure'));
+            this.$store.commit('CONFIG_PAGE_UNDO_SETTINGS_CHANGE');
```
jonboiser: This needs to include the `facilityConfig` namespace, since the action this line was taken from was part of the namespaced `facilityConfig` module.

```suggestion
            this.$store.commit('facilityConfig/CONFIG_PAGE_UNDO_SETTINGS_CHANGE');
```

kolibri/plugins/facility/assets/src/views/FacilityConfigPage/index.vue
```diff
@@ -218,23 +216,30 @@
...
+          });
       },
       resetToDefaultSettings() {
         this.showModal = false;
-        this.$store.dispatch('facilityConfig/resetFacilityConfig').then(() => {
-          this.copySettings();
```
jonboiser, line 227: We need to keep this call to `copySettings` somewhere inside of `updateSettings`. This creates a copy of the new settings object in order to do a "dirty check" if the user changes things again after saving.

jonboiser (approved):
Looks good! I added two commits to fix the tests and fix a bug with the UNDO mutation missing its namespace.

alexMet:
Looks good! Thank you and I' m sorry but i was stuck.

PR #7543, opened 2020-10-05
Fix grouping of boolean exprs for `isAndroidWebView`

### LINKED ISSUE #7538: Export Logs, User Import and Export missing after Facility Sync
### Observed behavior

Under Facility Data, the option for Export usage Logs as well as Import and export users is missing once a facility has been sync. 
Screen Shot 2020-09-30 at 2 19 53 PM

### Expected behavior

Expected behavior, I think it should still show Export Usage data and Import and Export users option even after a sync.

### User-facing consequences

For example, if Kolibri on a Pi is sent to students and when it comes back performs sync with a centralized instance of Kolibri, there still needs to be a way to import/export users as well as the ability to look at logs. I am able to generate the logs from cli, however, I am not sure about importing/exporting users using cli. 

### Steps to reproduce

Perform facility sync, once the sync has happened the option for exporting logs and users data management pieces are no longer there.  

### Context

Kolibri Version: 0.14.3

jonboiser:
Fixes #7538, where the user and activity log sections of the Facility Data page would not be shown on Safari.

### DIFF
--- a/kolibri/core/assets/src/utils/browserInfo.js
+++ b/kolibri/core/assets/src/utils/browserInfo.js
@@ -185,7 +185,7 @@ export const userAgent =
  */
 const isAndroid = /Android/.test(userAgent);
 export const isAndroidWebView =
-  (isAndroid && /wv/.test(userAgent)) || /Version\/\d+\.\d+/.test(userAgent);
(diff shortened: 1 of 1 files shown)


kollivier (approved):
LGTM, thanks! :)

PR #7550, opened 2020-10-08
Remove i18n tech debt from 0.14 

### LINKED ISSUE #7086: Remove crossComponentTranslator in FacilitiesTasksPage
### Observed behavior

Use of `crossComponentTranslator` to get strings in FacilityTasksPage after string freeze.

### Expected behavior

Define strings in the component they're used in or decide if the used strings should be in a core file.

### User-facing consequences

None but early on in my time at LE I eradicated `crossComponentTranslator` usage across the app and I will not continue to suffer its existence when we enter our next round of translations.

jonboiser:
### Summary

1. This removes every current usage of `crossComponentTranslator` by either moving messages around or into new or existing `common[Plugin]strings` mixins (new ones were made for "User" and "Device" plugins).

### Reviewer guidance

Ideally, try to find each of these messages in the actual app and make sure they are shown correctly (in English, since the messages have new namespaces).

If that's too much, the refactor is pretty mechanical, so just check for typos or things like that.

### References

Fixes #7086
----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated


kolibri/core/assets/src/mixins/commonCoreStrings.js
```diff
@@ -117,6 +117,12 @@ export const coreStrings = createTranslator('CommonCoreStrings', {
     context:
       "\nCould also be translated as \"View information about providing identifier\"\n\nAll 'AriaLabel' type of messages are providing additional context to the screen-reader users. \n\nIn this case the screen-reader will announce the message to the user indicating that they can access more information and examples about the 'Identifier' through the 'i' icon.",
   },
+
+  // Content activity
+  notStartedLabel: {
+    message: 'Not started',
+    context: 'For content that has not been viewed or engaged with yet',
```
radinamatic, line 124: I know this is context and not the user string, but still remembering that we should steer away from the term `content` :slightly_smiling_face: 

Maybe `Used to indicate a resource that learners did not yet open and engage with`?

jonboiser: Is "resource" the most generic term? I would probably even pick something more generic than "content" because "Not started" is a status for things like Exams and Lessons. This string was extracted as a status indicator for an Exam and I think I just wanted to give it a generic category as a common string.

nucleogenesis (approved):
LGTM! Thanks @jonboiser

kolibri/core/assets/src/mixins/commonCoreStrings.js
indirectlylit: In the coach tools we actually have distinct labels for each type to ensure that the gender agreement is correct. For example:

Lesson not started:

https://github.com/learningequality/kolibri/blob/4cf820b14386aecc228fecff64c847bad407cbb1/kolibri/plugins/coach/assets/src/views/common/status/statusStrings.js#L122

Exercise not started:

https://github.com/learningequality/kolibri/blob/4cf820b14386aecc228fecff64c847bad407cbb1/kolibri/plugins/coach/assets/src/views/common/status/statusStrings.js#L66

Also for:

* Question
* Quiz
* Resource

PR #7552, opened 2020-10-09
Fix new lesson creation title false error

### LINKED ISSUE #6593: Error ("A lesson with this name...") message while naming a new lesson
### Observed behavior

Each time I create a new lesson, I get this message 'A lesson with this name already exists' below the 'Title' text field. It disappears immediately though- within a couple of seconds
image (5)

### Expected behavior

The message shouldn't have appeared since there are no lessons with that name

### User-facing consequences

…

### Errors and logs

…

### Steps to reproduce

…

### Context

Kolibri 0.13.0
OS: Windows
Browser: Chrome

alexMet:
### Summary

Fixed the incorrect error on new lesson creation.

simplescreenrecorder

Also, added an error catch if somehow a lesson gets created with the same name while you are creating your new lesson.

simplescreenrecorder-(2)

### References

#6593 

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/plugins/coach/assets/src/views/plan/LessonCreationPage/index.vue
+++ b/kolibri/plugins/coach/assets/src/views/plan/LessonCreationPage/index.vue
@@ -36,6 +36,8 @@
 <script>
 
   import { crossComponentTranslator } from 'kolibri.utils.i18n';
+  import { ERROR_CONSTANTS } from 'kolibri.coreVue.vuex.constants';
+  import CatchErrors from 'kolibri.utils.CatchErrors';
   import commonCoreStrings from 'kolibri.coreVue.mixins.commonCoreStrings';
   import AssignmentDetailsModal from '../assignments/AssignmentDetailsModal';
   import LessonRootPage from '../LessonsRootPage';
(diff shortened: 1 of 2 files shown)


jonboiser:
This is great! We've been wanting to resolve this bug for a really long time! 👍

jonboiser (approved):
Nice! I didn't think the solution would be so simple 👍

alexMet:
@jonboiser I also thought it wouldn't be so simple. But, i don't really get the difference between the "disabled" and the "formIsSubmitted" variables though, in "AssignmentDetailsModal.vue".

jonboiser:
Ah, this is a pattern we sometimes use for forms. Normally the order of events is this

1. User clicks "submit" or keypresses "enter"
1. `formIsSubmitted` is set to true, which usually triggers invalidation errors like ("this field cannot be blank") to appear, but we do not make a request yet to the server.
1. If the form data is valid, then we send it to the server
1. Then we set `disabled` to true so various buttons and form inputs will appear as disabled.
1. If the request succeeds, then we usually close out the form and show a notification.
1. If the request fails, then we sometimes set both `disabled` and `formIsSubmitted` to `false` so the user can try again.

So it's sort of like `formIsSubmitted` is a misnomer because the form has not really been submitted yet; it's more like "the form is about to be submitted". In other parts of the code I have used the name and idea of `formShouldValidate`, which might be a little more accurate.

alexMet:
Oh i see, thank you very much, for taking the time to explain it to me!

PR #7561, opened 2020-10-14
Fix case change for facility name

### LINKED ISSUE #7493: Change of case in facility name is not allowed
### Observed behavior

Creating facility name from `kolibri manage provisiondevice` always results in facility name being created in lower case. (will create another issue for this)
Thereafter trying to edit it via GUI results in such a duplicate facility name error as shown in the screenshot below -

image

There is a hack to achieve the change by giving the facility a different name earlier and then eventually giving the desired name herein.

### Expected behavior

Changing case of a facility name should be allowed and not result in the duplicate name error as above.

### User-facing consequences

Inconvenient to use.

### Errors and logs

N/A

### Steps to reproduce

Edit facility name from Facility - Settings and then only change case in the name.

### Context

Kolibri 0.14.3 deb

alexMet:
### Summary

Fixes case changing for facility name.

simplescreenrecorder

### References

Fixes #7493 

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/plugins/facility/assets/src/views/FacilityConfigPage/EditFacilityNameModal.vue
+++ b/kolibri/plugins/facility/assets/src/views/FacilityConfigPage/EditFacilityNameModal.vue
@@ -36,6 +36,10 @@
     name: 'EditFacilityNameModal',
     mixins: [commonCoreStrings],
     props: {
+      facilityId: {
+        type: String,
+        required: true,
+      },
       facilityName: {
         type: String,
         required: true,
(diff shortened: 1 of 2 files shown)


jonboiser (changes requested):
This looks good, but could you make one change at this line

https://github.com/learningequality/kolibri/pull/7561/files#diff-ee5f0b1e2b649ee99cd8efdf30af63a949c2961ff128b9f9d900375a607df5cbR14

by removing the `.trim` from `v-model.trim="name"`? This will prevent the textbox from trimming whitespace when users are editing the name (which can be jarring)

alexMet:
@jonboiser I removed it from `v-model` and i added in the check inside of `nameIsInvalidText`. Because, otherwise you could submit the form with whitespace and an error would occur.

jonboiser:
> I removed it from `v-model` and i added in the check inside of `nameIsInvalidText`. Because, otherwise you could submit the form with whitespace and an error would occur.

Great call! I think this issue #7565 that was just reported is because we did not use `v-model.trim` but _also_ didn't do a later check of the trimmed input.

jonboiser (approved):

Issue #6560, opened 2020-02-14
In postgres, importing changing channel content moves channel to top of list


indirectlylit:
### Observed behavior

postgres-specific follow-up from https://github.com/learningequality/kolibri/issues/6376

From @jonboiser 

> Unforunately, this unwanted re-ordering issue is still happening when using Postgresql, except that the channel goes to the top of the list.

### Expected behavior

Updating a channel doesn't change its position

### User-facing consequences

* Users will need to move the channel again
* Users might not immediately realize the channel moved

### Steps to reproduce

* run with postgres
* import, adjust position, import more

### Context

0.13.1 beta 2

paulbusse:
I found the root cause for this one. The query 
```python
    latest_order = model.objects.latest("order").order 
    if latest_order is None: 
         channel.order = 1
```
is indeed the problem as pointed out to me by @jonboiser. `latest` is implemented as an `order by` which in Postgres puts the NULL values first (see [Postgres docs](https://www.postgresql.org/docs/13/queries-order.html)). As channels are created with a NULL in the `order` field, this query will always return NULL when an actual order is assigned to the channel.

PR follows later this week after testing is complete.

indirectlylit:
great, thanks!

PR #7619, opened 2020-10-26
Fixing #6560: In postgres, importing changing channel content moves channel to top of list

paulbusse:
### Summary
kolibri/plugins/device/assets/src/ManageContentPage/index.vue
  - Not (re-)sorting the channels before showing them on the page
  - remove the sortedChannels procedure
  - the channel panel uses the installedChannelsWithResources instead of sortedChannels
  - import SortBy no longer needed

But compared to the problem statement in the issue, based on the proposed fix, this is not Postgres related, The order resolves itself if the "new" content, is no longer new.

### Reviewer guidance
Follow the description in the ticket.
The changes are local to the file mentioned

### References
* #6560
* [Screenshot](https://drive.google.com/file/d/1gw5y3Bxq78pDTAIuA35o4mpiD3t324kb/view?usp=sharing)

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/plugins/device/assets/src/views/ManageContentPage/index.vue
+++ b/kolibri/plugins/device/assets/src/views/ManageContentPage/index.vue
@@ -37,7 +37,7 @@
 
       <div class="channels-list">
         <ChannelPanel
-          v-for="channel in sortedChannels"
+          v-for="channel in installedChannelsWithResources"
           :key="channel.id"
           :channel="channel"
           :disabled="channelIsBeingDeleted(channel.id)"
@@ -67,7 +67,6 @@
 
   import find from 'lodash/find';
   import get from 'lodash/get';
-  import sortBy from 'lodash/sortBy';
   import { mapState, mapGetters, mapActions } from 'vuex';
   import commonCoreStrings from 'kolibri.coreVue.mixins.commonCoreStrings';
   import { TaskResource } from 'kolibri.resources';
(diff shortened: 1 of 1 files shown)


jonboiser:
Hi @paulbusse, I traced the root cause of the channel ordering issue in #6376. In postgres, there is something about this query

```
 latest_order = model.objects.latest("order").order 
 if latest_order is None: 
     channel.order = 1 
```

that doesn't work properly. So I don't think the solution will be the front-end-only change in this PR. 

Also, your changes remove a special sorting logic that our apps does where it moves Channels that are being updated to the top. This sorting is not backed by the DB table behind the channels, but changes with a different DB table related to Tasks (e.g. tasks related importing, exporting, deleting channel contents), and we need to keep this special sorting.

paulbusse:
Hi @jonboiser, 
First of all I should have read through all the referenced material. Sorry for that.

I used this approach because the `ChannelMetadataSerializer` does not return the `order` field. So the frontend is on its own to cannot use that field to determine the order of the channels but relies on the order in the response.

I've seen a similar behavior as described in PR #6490, using Postgres. The difference being that in Postgres a newly added channel did not go to the very bottom. But after changing the channel it would go to the top and, when the task is clear and the page is refreshed, back into position.

PR #7632, opened 2020-10-27
Disable form on submit for leaner enrollment and coach assignment

### LINKED ISSUE #7347: When enrolling learners to the class, there is no visual feedback while the request is being made
### Observed behavior

For the Learner-Enrollment and Coach-Assignment workflows, there might be some delay before the request suceeds and the user is redirected back to the Facility > Class page. Some visual feedback that something is happening might be helpful.

### Expected behavior

After clicking the "confirm" button, while waiting for a response from the server:

1. Disable the "Confirm" button
1. Disable the checkboxes next to each users name

If the request succeeds, then we return to the higher-level page.
If the request fails, we stay on this page and show an error, and re-enable all of the controls.

From: https://kolibri-beta.learningequality.org/en/facility/#/a3f606630690237bbe94fe2a1a850af4/classes/0ca0280c91ccb9a880d922914226557f/coach-assignment

image

### Steps to reproduce

Try to enroll some learners/coaches to a class, possibly while throttling connection speed to create a longer delay.

### Context

…

alexMet:
### Summary

Disable the form, all the checkboxes and the confirm button, after submitting, for learner enrollment and coach assignment. Show a generic error message ("Changes not saved") when the API throws an error.

simplescreenrecorder

### Reviewer guidance

Should the error handling do something else, besides only showing a message?

### References

#7347 

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/plugins/facility/assets/src/modules/classAssignMembers/actions.js
+++ b/kolibri/plugins/facility/assets/src/modules/classAssignMembers/actions.js
@@ -11,8 +11,6 @@ export function enrollLearnersInClass(store, { classId, users }) {
       collection: classId,
       user: userId,
     })),
-  }).catch(err => {
-    store.dispatch('handleApiError', err, { root: true });
   });
 }
 
@@ -26,7 +24,5 @@ export function assignCoachesToClass(store, { classId, coaches }) {
       user: userId,
       kind: UserKinds.COACH,
     })),
-  }).catch(err => {
-    store.dispatch('handleApiError', err, { root: true });
   });
 }
--- a/kolibri/plugins/facility/assets/src/views/ClassEnrollForm.vue
+++ b/kolibri/plugins/facility/assets/src/views/ClassEnrollForm.vue
@@ -12,6 +12,7 @@
           v-model="selectedUsers"
           :users="items"
           :selectable="true"
+          :disabled="disabled"
           :emptyMessage="emptyMessageForItems(items, filterInput)"
         />
       </template>
@@ -22,7 +23,7 @@
         :text="coreString('confirmAction')"
         :primary="true"
         type="submit"
-        :disabled="selectedUsers.length === 0"
+        :disabled="disabled || selectedUsers.length === 0"
       />
     </div>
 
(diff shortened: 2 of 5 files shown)


marcellamaki:
Hi Alex - this looks good. I see your question "Should the error handling do something else, besides only showing a message?" I noticed that you moved the error handling from action.js and added the snackbar. Did you move the catch only for the snackbar, or did you have an idea or suggestion of something else the error handling could do?

alexMet:
Hi @marcellamaki. Yes, i moved the catch so i can re-enable the form and then show the error message. I didn't had an idea or suggestion, i just thought, if the API throws an error, it should be reported. But, since it is an offline platform, i don't know how you get error reports or if you have an error handling pattern, so i thought i should ask. :)

jonboiser:
@alexMet, for errors like this that we can anticipate at design time, we tend to inform the user in a "graceful" way like how've you implemented it with the "Changes not saved" pop-up notification. Although this does make the failure a little mysterious and hard for the user to debug (so normally they'd have to report it to GH or the forum and we would need to do some detective work).

In other places where we did not anticipate a specific error and design a way to handle it in the UI, we use this Vuex action `handleApiError` that brings up a whole page Error page `AppError.vue`, with diagnostic and contact info.

marcellamaki (approved):
Thanks again @alexMet - this looks good!

kolibri/plugins/facility/assets/src/modules/classAssignMembers/actions.js
```diff
@@ -11,8 +11,6 @@ export function enrollLearnersInClass(store, { classId, users }) {
       collection: classId,
       user: userId,
     })),
-  }).catch(err => {
-    store.dispatch('handleApiError', err, { root: true });
```
jonboiser, line 15: Yeah, so now when this operation fails, you won't see the whole-page `AppError` component, but just the snackbar notification, which is what I recommended in the original issue.

alexMet:
@jonboiser That is what i though, but i wanted to ask, just to be sure. Thank you very much for taking the time explaining it to me. :)

PR #7640, opened 2020-10-28
Add 30 second TTL to ProcessLock redis keys

aronasorman:
This PR adds a 30 second TTL to the redis keys created by the ProcessLock object. This can be overriden through a new option called `OPTIONS["Cache"]["CACHE_LOCK_TTL"]`.

As an extra, we also pin `drf-yasg` to a specific version so it remains compatible with our pinned drf version.

### DIFF
--- a/kolibri/core/utils/cache.py
+++ b/kolibri/core/utils/cache.py
@@ -26,12 +26,12 @@ class ProcessLock(object):
     def __init__(self, key, expire=None):
         """
         :param key: The lock key
-        :param expire: The cache key expiration in seconds
+        :param expire: The cache key expiration in seconds (defaults to the CACHE_LOCK_TTL option if not set)
         :type key: str
         :type expire: int
         """
         self.key = key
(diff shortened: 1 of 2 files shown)


rtibbles:
Maybe better targeted at 0.14.x?

Also note that this commit on that branch does the drf-yasg pinning: https://github.com/learningequality/kolibri/commit/e72c6f737cdb8e4198a4380620001d064e7a3e18

aronasorman:
Oof looks like i based this on develop. I'll do some branch surgery to get the changelist pristine.

aronasorman:
Changed the target branch and removed the commit adding a pin to `drf-yasg`.

rtibbles (approved):
This seems reasonable to me - the only other things I can think we would do to mitigate this is to just try to defer entirely to DB locks to handle this, and add arbitrary wait times instead of this extra locking layer. But that would be a much bigger change.

Issue #7641, opened 2020-10-29
Unable to tell from search results which exercises are for coaches


MindfulQA:
### Observed behavior

Exercises in search results do not include a description and do not indicate whether they are coach exercises or not.

![Screen Shot 2020-10-28 at 8 56 01 PM](https://user-images.githubusercontent.com/73143721/97523866-80990980-1960-11eb-85a9-f2187b3a2ab5.png)

…

### Expected behavior

Would expect to be able to view an exercise's title, description, and whether the exercise is for coaches from the search results page.
…

### User-facing consequences

The user is unable to read a description of an exercise, or easily identify which exercises are for coaches.
…

### Errors and logs

N/A
…

### Steps to reproduce

1. Open Kolibiri Mac desktop app and log in to Coach account
2. From the class home page, click on an existing lesson (or create a new one if there is not one already existing)
3. The lesson should open under the "Reports" tab. From the lesson title box, click the "options" drop-down menu and click "manage resources."
4. Enter a search term in the search bar and click the magnifying glass
6. Filter the search results using the "Type" drop down by clicking "Exercise"
7. Observe that the results show the title of the exercise, but not a description or coach indicator.

Note for testing - issue occurs in scenario in line 149:

https://github.com/learningequality/kolibri/blob/release-v0.14.x/integration_testing/features/coach/coach-create-lesson.feature 

…

### Context

macOS Catalina 12.15.7 - Kolibri version 0.14

…

nucleogenesis:
Hey there @MindfulQA - I wonder if the wording of the story is unclear. There should not be any indication of content _not_ being coach content. We only show a blue icon when the content is coach content, but show nothing if the content isn't.

I tested and saw the icon by searching for "Perseus" and opening the "KA Perseus" topic folder where the "Make 10" content correctly shows the blue coach content icon.

MindfulQA:
Hey @nucleogenesis - thanks! That part makes sense - it may have been that we were doing many different searches, and assumed that at least some of them would have been coach content. But there was no coach content indicator for any of the ones we were looking at. Sounds like that might have been a coincidence based on searches we did.

PR #7654, opened 2020-10-30
Clarification in lesson creation gherkin

nucleogenesis:
### Summary

While trying to replicate https://github.com/learningequality/kolibri/issues/7641 I noticed the line was a little ambiguous that the reader may think that there should be an indication that there are _no_ coach contents. 

### Reviewer guidance

Does the added line clarify as intended? Is there a better way to get the idea across?

### References

https://github.com/learningequality/kolibri/issues/7641


radinamatic (approved):
Thank you, this should make the scenario clearer :+1:

PR #7655, opened 2020-11-02
Merge `0.14.4-weekly6` into `develop`

jonboiser:


requirements/dev.txt
```diff
@@ -6,7 +6,7 @@ tox==3.1.3
 django-debug-panel==0.8.3
 django-debug-toolbar==1.9.1
 sphinx-autobuild
-drf-yasg
+drf-yasg==1.17.1
```
manuq, line 9: :+1: This is needed right now to setup Kolibri for development. Today I went through the getting started guide and got an error on this `drf-yasg` dependency. I was about to open a PR, glad to find one already.

rtibbles (approved):

requirements/dev.txt
jonboiser: @manuq, were you planning to work on an issue? If, so you will most likely want to be using the `release-v0.14.x` as your base. Because of how we do our releases, that branch is where we target changes that will fix issues in any milestone like `0.14.4` or `0.14.5`

manuq: @jonboiser ah good to know. I'm doing some research for Endless, no issue in particular. I was following the [getting started guide](https://kolibri-dev.readthedocs.io/en/develop/getting_started.html) from scratch today, because my git checkout was outdated.

PR #7660, opened 2020-11-04
Keyboard navigation focus outline

### LINKED ISSUE #7651: keyboard focus outline is not visible on cards in Learn page
### Observed behavior

In this video [from this link](https://kolibri-beta.learningequality.org/en/learn/#/classes), I shift-tab until I am focused on the skip-link, hit enter, hit tab once, then am focused on the first card "Jessica's class" (I can confirm this because I keypress "Enter" and am taken to the right page). Yet, there is no green outline on the card. 

(Notice that some elements on the top nav aren't outlined either while back tabbing)

CleanShot 2020-10-29 at 11 42 56

(This happens in a lot of other areas, but fixing this in Learn will probably fix it globally)

### Expected behavior

Cards (which are anchor tags under the hood) get a green focus outline while in the keyboard modality.

### User-facing consequences

It's hard for keyboard-only users to know where they are focusing.

### Errors and logs

…

### Steps to reproduce

…

### Context

…

alexMet:
## Summary

Added keyboard focus outline to Channel card, Content cards and core menu links, removed the extra tab key press after the Navbar.

Removed an extra tab key press in the sign in page.

Before:

After:

### References

#7651 

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/core/assets/src/views/SideNav.vue
+++ b/kolibri/core/assets/src/views/SideNav.vue
@@ -1,6 +1,12 @@
 <template>
 
-  <div ref="sideNav" class="side-nav-wrapper" tabindex="0" @keyup.esc="toggleNav">
+  <div
+    v-show="navShown"
+    ref="sideNav"
+    class="side-nav-wrapper"
+    tabindex="0"
+    @keyup.esc="toggleNav"
+  >
     <transition name="side-nav">
       <div
         v-show="navShown"
--- a/kolibri/core/assets/src/views/CoreMenu/CoreMenuOption.vue
+++ b/kolibri/core/assets/src/views/CoreMenu/CoreMenuOption.vue
@@ -62,6 +62,7 @@
             ':hover': {
               backgroundColor: this.$themeBrand.primary.v_100,
             },
+            ':focus': this.$coreOutline,
           };
         }
         return {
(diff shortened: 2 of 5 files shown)


jonboiser:
@alexMet , this is great! The focus outline was something we used to get without having to add classes and styles to individual elements, but your solution here is something we can apply to other parts of the code until we figure out how to fix that more global issue.

alexMet:
@jonboiser It took me a while to figure out how this all works. Yes, this is a quick fix, a global one would be much better. I tried changing `<router-link>` to `<KRouterLink>` since the latter always handles the outline, but that didn't work. :)

jonboiser:
> I tried changing `<router-link>` to `<KRouterLink>` since the latter always handles the outline, but that didn't work. :)

Oh, really? Why didn't that work?

alexMet:
The outline works fine, but couldn't make the content inside the card show.

jonboiser:
Ah, okay. I think `KRouterLink` usually works best with simple text only links, but I'll need to double-check if that's true or if there's something special about the Classes card links.

kolibri/core/assets/src/views/SideNav.vue
```diff
@@ -1,6 +1,12 @@
 <template>
 
-  <div ref="sideNav" class="side-nav-wrapper" tabindex="0" @keyup.esc="toggleNav">
+  <div
+    v-show="navShown"
```
jonboiser, line 4: What does this `v-show` do?

alexMet: In a previous issue, I added the tabindex in order to close the side menu with the escape button. Because of that, you have to press the tab key 2 times in order to go from the navbar to the main content. So with the navShow the div can be hidden until you open the menu.

PR #7671, opened 2020-11-06
i18n -> kolibri-tools for cross product i18n

nucleogenesis:
### Summary

Allows any package importing `kolibri-tools` full access to Kolibri's i18n suite of tools.

`crowdin.py` 

- New required ENV VAR `CROWDIN_LOGIN` - must be set alongside the already required `CROWDIN_API_KEY`. 
- Many relatively simple changes to account for Perseus strings only when they exist and *should* exist (which is only when we're running in Kolibri)
- Automatically find the proper `locale` path for where to get CSV files using `os.getcwd()` - so this must be run at the project's root. 

`Makefile`

- Issue with i18n commands going into my virtual env (likely others' as well) fixed by `cd` into the `kolibri` dir before running

`ExtractMessages.js`

- Hard-coded project paths means future projects must be added to this file and configured there.
- Comments in the file serve as basic guides for adding a new project - the process ought to be relatively simple as you just need to define and account for a few paths to where you want the strings extracted from.
- Uses globs to find all JS and Vue files. Parses them down to their respective JS scripts.
- Provides error messages that ought to help future debugging if something goes haywire. Where things can break, I tried to add messages giving hints about where it broke and why.
- Takes command line arguments, including a couple of helpful debugging / dev commands and one for determining which project we're running on
- Accounts for messages defined where the key is the value of a different variable - even if that variable is defined in another file and/or it is an Object.

```
import Constants from '../Constants';
import { SlideshowConstant } from "shared/OtherStrings";

const ExampleVariable = "exampleVariable";
const SomeObject = {
    foo: "bar",
};

export default {
    name: "ExampleComponent",
    ........
    $trs: {
        [Constants.video]: "Video",    // Constants.video refers to the string value "video"
        [SlideshowConstant]: "Slideshows",    // SlideshowConstant refers to the string value "slideshow"
        [ExampleVariable]: "Well here's an example of a variable defined locally, which will work as well",
        [SomeObject.foo]: "Baz",
    },
    ........
}
```

This would ultimately end up extracting the following strings:

```
ExampleComponent.video: "Video",
ExampleComponent.slideshow: "Slideshows",
ExampleComponent.exampleVariable: "Well here's an example of a variable defined locally, which will work as well",
ExampleComponent.bar: "Baz",
```

`General i18n changes`

- Moved everything i18n from `build_tools` and `kolibri-tools` into `packages/kolibri-tools/lib/i18n` and updated all paths accordingly.

`Other notes`

- String profiling can still be made to work - but only through Webpack. The path globbing used in `ExtractMessages` could go a long way toward getting the profiling working - but the profile code would only be needed by Studio if/when they want to move strings into Common files in a future translation process.
- I think there are still some artifacts of the i18n in the `cli.js` and frankly I'm not entirely sure right now if they're necessary for other processes - such as serving the translated JSON files. I'll dig further into this tomorrow but if you have insights please don't hesitate!

## HOW TO TEST THIS

Test this in Studio and Kolibri

### Kolibri

1. Set your necessary environment variables `CROWDIN_PROJECT` - must be one of `kolibri` or `kolibri-studio` - this is derived from Crowdin itself. Then your `CROWDIN_LOGIN` - which is your Crowdin username.

2. `make i18n-upload branch=makeupabranch`

Everything should go smoothly and you should see your branch on Crowdin.

### Studio

0a. Check out the branch on this PR: https://github.com/learningequality/studio/pull/2520 `nucleogenesis/i18n`
0b. If you have not already, `cd path/to/kolibri/packages/kolibri-tools` then `yarn link` - which allows you to link to `kolibri-tools`
1. `cd back/to/studio` and `yarn link kolibri-tools`
2. Your `CROWDIN_PROJECT` should be `kolibri-studio`
3. `make i18n-upload branch=makeupabranch`

Everything should go smoothly and you should see your branch on Crowdin.


rtibbles:
> I have gotten into it with yarn several times about the package versions of vue-template-compiler and things to do with eslint. The version of eslint we use is apparently different between kolibri and kolibri-tools - but I'm not really sure.

This happened previously with `vue-template-compiler` when @micahscopes was using `kolibri-tools` for linting Studio. I have no idea how the version of `eslint` could be different between Kolibri and `kolibri-tools` because we only use `eslint` in `kolibri-tools` and then kolibri uses it from there...

Makefile
```diff
@@ -183,7 +183,7 @@ pex: writeversion
 	ls dist/*.whl | while read whlfile; do pex $$whlfile --disable-cache -o dist/kolibri-`cat kolibri/VERSION | sed 's/+/_/g'`.pex -m kolibri --python-shebang=/usr/bin/python; done
 
 i18n-extract-backend:
-	python -m kolibri manage makemessages -- -l en --ignore 'node_modules/*' --ignore 'kolibri/dist/*'
+	python -m kolibri manage makemessages -- -l en --ignore 'node_modules/*' --ignore 'kolibri/dist/*' --ignore 'venv/*'
```
indirectlylit: > ` --ignore 'venv/*'`

is this general, or dev environment specific? (for example, my pipenv virtual environment is in `.venv`)

Makefile
```diff
@@ -199,30 +199,30 @@ i18n-django-compilemessages:
 	cd kolibri && PYTHONPATH="..:$$PYTHONPATH" python -m kolibri manage compilemessages
 
 i18n-upload: i18n-extract
-	python build_tools/i18n/crowdin.py upload-sources ${branch}
+	python packages/kolibri-tools/lib/i18n/crowdin.py upload-sources ${branch}
```
indirectlylit, line 202: We might want to deprecate these makefile commands, and instead document using the crowdin.py script directly. That would help avoid duplication that we're seeing in Studio and would begin to see other places:

https://github.com/learningequality/studio/pull/2520/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R73-R126

Having these aliases in the makefile was something of a historic artifact and I don't think there's a good reason to have them here

packages/kolibri-tools/lib/i18n/utils.py
```diff
@@ -6,22 +6,74 @@
...
+                os.pardir,
+                os.pardir,
+                os.pardir,
+                "{}-locale".format(PROJECT_NAME),
+            )
+        )
+    if not os.path.exists(locale_path):
+        os.makedirs(locale_path)
+
+    return locale_path
```
indirectlylit: this seems fairly brittle, and I'm not entirely sure what it's supposed to do

docs/i18n.rst
```diff
@@ -225,7 +225,12 @@ First, you'll need to have GNU ``gettext`` available on your path. You may be ab
 
 .. note:: If you install ``gettext`` on Mac with Homebrew, you may need to add the binary to your path manually
 
-Next, ensure you have an environment variable ``CROWDIN_API_KEY`` set to the Learning Equality organization `account API key <https://support.crowdin.com/api/api-integration-setup/>`__.
+Next, ensure you have the following environment variables set:
+
+* ``CROWDIN_API_KEY`` - your `account API key <https://support.crowdin.com/api/api-integration-setup/>`.
+* ``CROWDIN_LOGIN`` - your `Crowdin username <https://crowdin.com/settings#account>`.
+* ``CROWDIN_PROJECT`` - (optional, default: "kolibri") the "Project Identifier" on Crowdin that you want to work with. You can find this on Crowdin in the project's settings in the "API & Webhooks" section. 
+* ``CROWDIN_LOCALE_ABSOLUTE_PATH`` - (optional) set where you want the generated files to appear (eg "kolibri/locale" or "contentcuration/locale").
```
indirectlylit: > (optional) set where you want the generated files to appear (eg "kolibri/locale" or "contentcuration/locale")

optional, when?

package.json
```diff
@@ -13,9 +13,10 @@
     "test-jest-cov": "yarn run test-jest --coverage",
     "coverage": "yarn run test-jest-cov",
     "build": "kolibri-tools build prod --file ./build_tools/build_plugins.txt && yarn run hashi-build",
-    "makemessages": "kolibri-tools build i18n --file ./build_tools/build_plugins.txt",
+    "makemessages": "NODE_ENV=i18n-extract kolibri-tools build i18n --file ./build_tools/build_plugins.txt",
+    "profilemessages": "NODE_ENV=i18n-profile kolibri-tools build i18n --file ./build_tools/build_plugins.txt",
     "combineprofiles": "node ./packages/kolibri-tools/lib/combineStringProfiles.js ./kolibri/locale/en/LC_MESSAGES/profiles/",
-    "transfercontext": "node ./packages/kolibri-tools/lib/i18nSyncContext.js run && yarn lint-frontend:format",
+    "transfercontext": "APP_NAME=kolibri node ./packages/kolibri-tools/lib/i18nSyncContext.js run && yarn lint-frontend:format",
```
indirectlylit: Rather than controlling `kolibri-tools build` behavior with env vars and setting them like this, maybe we should instead have different commands for each action.

If necessary we can still still provide need a generic, env-var-driven `build` wrapper command

packages/kolibri-tools/lib/i18n/fonts.py
```diff
@@ -43,6 +43,14 @@
     )
 )
 
+if not os.path.exists(OUTPUT_PATH):
+    print(
```
indirectlylit: maybe `logger.warning`

packages/kolibri-tools/lib/i18n/utils.py
```diff
@@ -6,22 +6,74 @@
...
+    )
+    PERSEUS_SOURCE_PATH = os.path.join(PERSEUS_LOCALE_PATH, "en", "LC_MESSAGES")
+    if not os.path.exists(PERSEUS_SOURCE_PATH):
+        os.makedirs(PERSEUS_SOURCE_PATH)
+    if not os.path.exists(PERSEUS_LOCALE_PATH):
+        os.makedirs(PERSEUS_LOCALE_PATH)
+except:
+    PERSEUS_LOCALE_PATH = None
+    PERSEUS_SOURCE_PATH = None
+    pass
```
indirectlylit: a lot of this complexity would go away with a vendored perseus @rtibbles

Makefile
nucleogenesis: Considering - `--include` (or equivalent) or `--ignore=*/site-packages/*`

docs/i18n.rst
nucleogenesis: Will improve with clarifying error messages. Not optional.

package.json
rtibbles: Yeah, I'm not a big fan of the env vars here, would rather add more command line flags, or additional commands.

I think in general, we could pull the `i18n` command out into its own top level command in `kolibri-tools`, it was only in `build` historically because of the use of the webpack pipeline to do string extraction. I've not looked over the PR in detail yet, but my assumption was we were dropping that for pure AST parsing.

packages/kolibri-tools/lib/i18n/utils.py
rtibbles: I'll get back to you on that...

Makefile
rtibbles: Could make it simpler by adding a `cd` into the kolibri dir, and then you don't need to ignore these top level folders.

Can also specify `--extension=html,py`, but by default it should ignore JS files anyway.

packages/kolibri-tools/lib/cli.js
```diff
@@ -141,7 +141,7 @@ program
     const buildModule = {
       [modes.PROD]: 'production.js',
       [modes.DEV]: 'webpackdevserver.js',
-      [modes.I18N]: 'i18n.js',
+      [modes.I18N]: 'i18n/index.js',
```
rtibbles, line 144: As I suggested above, I think we can pull this out into a top level `i18n` command alongside `build`, `lint`, and `test`.

packages/kolibri-tools/lib/i18n/ExtractStrings.js
```diff
@@ -50,11 +47,11 @@ ExtractStrings.prototype.apply = function(compiler) {
   var self = this;
   // Only do this in non-production mode, as otherwise the module detection code
   // does not work for .vue files.
-  if (process.env.NODE_ENV !== 'production') {
+  if (process.env.NODE_ENV === 'i18n-extract') {
```
rtibbles: I think this is superfluous - only had this flag here previously because it was being invoked inside the webpack lifecycle, and in production mode we were unable to parse the AST of the Vue files properly, so we just got blank messages.

packages/kolibri-tools/lib/i18n/ExtractStrings.js
```diff
@@ -361,10 +359,11 @@ ExtractStrings.prototype.apply = function(compiler) {
 };
 
 ExtractStrings.prototype.writeOutput = function(messageExport) {
+  const writePath =`${this.messageDir}/locale/CSV_FILES/en/LC_MESSAGES`;
```
rtibbles: Is this still being used as a webpack plugin, or has it been separated? I'd much prefer the latter and I think it would make things less complicated overall.

packages/kolibri-tools/lib/i18n/utils.py
rtibbles: Yeah, I am not quite sure what this is meant to be doing or why `node_modules` is particularly meaningful here, but using that as a sign post feels like a code smell.

packages/kolibri-tools/lib/i18n/vue-intl-locale-data.js
```diff
@@ -0,0 +1,37 @@
+/*
+ * This is an auto-generated file, any manual edits will be overridden.
+ *
+ * To regenerate, see instructions here:
+ *   https://kolibri-dev.readthedocs.io/en/develop/references/i18n.html
+ *
+ * This file was generated by frontend_build/src/intl_code_gen.js
```
rtibbles, line 7: We should probably update this comment :)

rtibbles (commented):
I had assumed from our conversations that we were going to completely back the i18n code out of the webpack life cycle. Given that Studio does not currently rely on webpack for this, that seemed like a good direction to me, but I'm not seeing those changes here. I think it should generally simplify a lot of the code too, as it's just a matter of doing a glob search for files, and then passing them to be AST parsed.

packages/kolibri-tools/lib/i18n/utils.py
nucleogenesis: Yeah this stinks to high heaven. I'm going to clean this all up a bit and make things dependent on env vars entirely to avoid the kind of confusion I ran into while working on it :-D

packages/kolibri-tools/lib/i18n/ExtractStrings.js
nucleogenesis: I only added this (and the same kind of thing in the ProfileStrings.js file) so that we don't run the entire string extraction every time you start the dev server or `yarn build`. See my bottom-most comment about why I didn't extract the logic into direct AST parsing.

packages/kolibri-tools/lib/i18n/vue-intl-locale-data.js
nucleogenesis: > I think it should generally simplify a lot of the code too, as it's just a matter of doing a glob search for files, and then passing them to be AST parsed.

Unfortunately, how the string extraction works doesn't use the same AST parser as I used in the i18nSyncTransfer work (where I already took the glob approach). I thought it was a great idea before I started working - but as I got into it I quickly realized that it would be biting off more than is necessary to chew.

The AST parsers have minute details about them that are different - they call a node differently or structure the node metadata differently or they require that you use whatever weird specific options you need to accommodate features you're using or not using - so it isn't as easy as just using the existing ExtractStrings.js code and replacing webpack's build system w/ a different one.

Which is why I figured that adding some kind of condition to when/how that plugin is actually run would basically have the same ultimate outcome: working code with a single command (package.json script anyway) that extracts/profiles the messages.

packages/kolibri-tools/lib/i18n/ExtractStrings.js
nucleogenesis: Explained below why I didn't pull it out of webpack

packages/kolibri-tools/lib/i18n/vue-intl-locale-data.js
nucleogenesis: One thing I should add is that I went this route so that I could get it up and running sooner than later so that Radina can test the languages.

I'll need to test https://github.com/learningequality/studio/pull/2520 to see if it runs/builds without needing to actually use the newest `kolibri-tools` that is in this repo. I believe that the everything i18n that isn't to do with extraction/profiling/context transfer/Crowdin (aka `kolibri-tools`) is entirely local to Studio so if the translation JSONs are committed then the server should run.

In that case - then I can feel less pressure and put a big pretty webpack-less and env var-less bow on our cross-product i18n sometime between now and Thanksgiving.

rtibbles: >  how the string extraction works doesn't use the same AST parser as I used in the i18nSyncTransfer

Ooof... sorry I should have caught this when you first did this work. Are you using esprima rather than espree? I had very specific issues with esprima that led me to use espree for the string extraction previously, so I guess I assumed you would face the same issues with esprima too...

packages/kolibri-tools/lib/i18n/language_info.json
```diff
@@ -0,0 +1,184 @@
+[
```
rtibbles, line 1: Note that this is now living in two places, and we definitely need it in production in Kolibri.

packages/kolibri-tools/lib/i18n/ExtractMessages.js
```diff
@@ -0,0 +1,804 @@
...
+ *    are not being extracted properly, etc
+ *
+ * This block of code will setup paths specific to each supported
+ * project and will define how their files are processed. These paths
+ * will be used to determine which globs are processed
+ *
+ */
+
+// NP - Be sure you select a name for your project here and add it to this array
+const SUPPORTED_PROJECTS = ['kolibri', 'studio'];
```
rtibbles, line 44: I don't like this at all, and it means it would be difficult to use this tooling for Kolibri plugins.

We should either allow the file paths to be configured (for use in Studio and use cases we don't anticipate) or to leverage the Kolibri plugin build machinery that already exists to get these paths (which includes specifying the locale path on a per plugin basis).

My main reason for pushing back on this strongly, is that I think if this is just for Kolibri and Studio, we should not bother making generalized code and should just copy paste across the repos. My intention was that this tooling would be useful for Kolibri plugins also and hence more generally applicable.

rtibbles (changes requested):
I think there's probably some more detail to go into here, but my skim of the extraction AST parsing code looks reasonable to me.

The main blocker is that we have to generate JSON files on a per frontend bundle basis, not per plugin. This is because different Javascript assets may be injected into different SPAs (for example the side nav bundles are used on every SPA and need strings).

To avoid having to load all the strings for every plugin on every page, we should reinstate the per-bundle JSON generation. We can maintain per plugin CSV generation for translator convenience.

So, we can use our plugin parsing machinery that we use for code building, and use this to also generate the file paths we search in for string extraction.

In the process of this, we can also extract the specific webpack configurations for each plugin (which will allow us to properly traverse aliased import paths). We can then also use the webpack configuration to discover the entry points of each of our frontend bundles, and extract the appropriate strings for each of those into the relevant bundle JSON from our translated CSVs.

This means that we still get to upload strings from unimported files, because we are extracting strings via globbing, but we are only loading used strings at runtime. Further, it means that if someone has been a bit naughty and has imported strings from another plugin, then we can allow at JSON generation time for that string to be duplicated into the JSON files for multiple frontend bundles, meaning we won't get caught out by cross component translation issues.

packages/kolibri-tools/lib/i18n/ExtractMessages.js
nucleogenesis: Right on - on our next pass I look forward to hashing this out together. If using the build machinery still basically gets us those glob paths and we read the files in and make ASTs from the string of the file then what I'd consider the most complicated parts of the script won't require much change.

I did try to give relatively easy ways to extend this process to other projects - but it was based on an incorrect assumption about how I should break down the strings in Kolibri with regard to the plugins within each of the `kolibri/plugins/*`.

Thanks for the thorough breakdown here!

packages/kolibri-tools/lib/i18n/language_info.json
nucleogenesis: What do you think of this commit (in a branch off of this one): https://github.com/learningequality/kolibri/commit/edc7c76cbc4b527f34c0cf723843c543d372b387

rtibbles (approved):
Merging for now as we need this in kolibri-tools for Studio.

@nucleogenesis will file follow up issues to ensure we can continue to properly process strings for 0.15.

MBKayro:
hi @nucleogenesis,
if i have these error, this is mean idont have access to Crowdin of my account or the official learningeqality ?

raise HTTPError(http_error_msg, response=self)
requests.exceptions.HTTPError: 403 Client Error: Forbidden for url: https://api.crowdin.com/api/project/kolibri/add-directory?account-key=AHIjSxwXDVnF5kthppluCkyQbJEOai9d&login=MBKayro&name=makeupabranch&is_branch=1&json
make: *** [Makefile:202 : i18n-upload] Erreur 1

MBKayro:
hi @nucleogenesis,

is these error related to my personnel account of CROWDIN or not allowed to access to CROWDIN of learningequality

/models.py", line 940, in raise_for_status
    raise HTTPError(http_error_msg, response=self)
requests.exceptions.HTTPError: 403 Client Error: Forbidden for url: https://api.crowdin.com/api/project/kolibri/add-directory?account-key=275b31fc2b980aed3a079e6250583619&login=MBKayro&name=makeupabranch&is_branch=1&json
make: *** [Makefile:202 : i18n-upload] Erreur 1

PR #7687, opened 2020-11-11
Better conditional checks

rtibbles:
### Summary
Uses the skip duplicate actions workflow to skip checks in a way that is compatible with the github required checks interface.

### Reviewer guidance
Do the required checks that don't get run on this PR not block merging?


jonboiser (approved):
Looks good!

PR #7688, opened 2020-11-12
Fixing #6560: In postgres, importing changing channel content moves channel to top of list

paulbusse:
### Summary
The root cause is explained [here](https://github.com/learningequality/kolibri/issues/6560#issuecomment-725359351).
- replaced `latest` by `max`, should be more db independent
- rewrote the function to prevent the query from being executed on every call to this function.

BTW. To get rid of the bad orders of you have to reorder them once after the patch is installed.

### Reviewer guidance
Following tests have been executed:
* adding new channels on existing channel list. (Postgres)
* adding a channel on a new kolibri set up. (Postgres)
* re-order the channels and added channels after that. 
* Ran the pytest suite (for sqlite compatibility)
If there are more test I need to run let me know.

No migration is provided.

### References
* #6560
* #7615 (I close this pull request: my previous attempt)

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/core/content/utils/annotation.py
+++ b/kolibri/core/content/utils/annotation.py
@@ -4,6 +4,7 @@ import os
 from itertools import groupby
 from math import ceil
 
+from django.db.models import Max
 from django.db.models import Sum
 from le_utils.constants import content_kinds
 from sqlalchemy import and_
@@ -749,11 +750,10 @@ def calculate_included_languages(channel):
 
 
 def calculate_next_order(channel, model=ChannelMetadata):
-    latest_order = model.objects.latest("order").order
-    if latest_order is None:
-        channel.order = 1
-
     if channel.order is None or channel.order == 0:
-        channel.order = latest_order + 1
+        max_order = model.objects.aggregate(Max("order")).get("order__max", 0)
+        if max_order is None:
+            max_order = 0
+        channel.order = max_order + 1
 
     channel.save()


kolibri/core/content/utils/annotation.py
```diff
@@ -749,11 +750,10 @@ def calculate_included_languages(channel):
 
 
 def calculate_next_order(channel, model=ChannelMetadata):
-    latest_order = model.objects.latest("order").order
```
jredrejo, line 752: I think the only line that must be changed is this one, just replacing it by 
`latest_order = model.objects.filter(order__isnull=False).latest("order").order`
would solve the issue without more changes needed.
@paulbusse do you see any inconvenience in doing just this change?
If you don't see any issue in it, I think it's better than doing the new aggregations this PR introduces

jredrejo: bah, @paulbusse  forget my previous comment, your solution is better for the case where no channels have been added yet.
Please, move the PR to be targeted at release-v0.14.x instead of develop and it's fine to be merged. Sorry for the noise.

paulbusse: @jredrejo I kinda made the same journey :-)

Also reading [this](https://learnsql.com/blog/how-to-order-rows-with-nulls/) means that `latest` on a column with `NULL` values is not trustworthy. Django should standardize the usage of NULL values in the generated query, e.g. by adding `NULLS LAST`. `MAX` must, per the definition in the standard, skip over `NULL` values. (Just my €0.02)

Unless I'm overlooking something I'm not able to change the target of the PR.

jredrejo: > @jredrejo I kinda made the same journey :-)
> 
> Also reading [this](https://learnsql.com/blog/how-to-order-rows-with-nulls/) means that `latest` on a column with `NULL` values is not trustworthy. Django should standardize the usage of NULL values in the generated query, e.g. by adding `NULLS LAST`. `MAX` must, per the definition in the standard, skip over `NULL` values. (Just my €0.02)
> 
> Unless I'm overlooking something I'm not able to change the target of the PR.

I have changed the target in the GH UI, but the PR includes, logically, all the commits that are different between develop and  release-v0.14.x. If you can, the steps I follow in these cases are:
- From the issue6560b branch: `git checkout -b keep_commit`
- From  the release-v0.14.x:
```
 git branch -D  issue6560b
 git checkout -b issue6560b
 git cherry-pick id_of_your_commit
 git push -f 
 git branch -D keep_commit
```
If you have problems, don't worry, leave it as it is and we'll cherry-pick the commit from develop to   release-v0.14.x

paulbusse: I tried, but never done this before.

I got as far as the `push`. But that one gives an error 
```
fatal: The current branch issue6560b has no upstream branch.
To push the current branch and set the remote as upstream, use

    git push --set-upstream origin issue6560b
```
you forgot the `-f `to force the change ;)


(PS. The first command is `git branch -D  issue6560b`, I assumed)

yes
Don't worry. We'll cherry pick it.

jredrejo (approved):

PR #7691, opened 2020-11-13
Add new watched prop to consolidate logic in 'disabled'

### LINKED ISSUE #7376: "Continue" button on Select Address Modal is not disabled when there are no selectable options
### Observed behavior
From @micahscopes comment https://github.com/learningequality/kolibri/pull/7350#pullrequestreview-453915802

88273510-e6119600-cc9f-11ea-9f46-f28af05eb646

In this video, the "Continue" button is enabled at the peer-discovery phase even though there are no enabled Devices/Addresses and no zeroconf peers.

This might be because the one static address "8001" is selected, but is offline, so not really usable.

### Expected behavior

1. The "Continue" button is disabled if there are no online peers
1. Disabled addresses radio buttons are _not selected_

image

### User-facing consequences

…

### Errors and logs

…

### Steps to reproduce

…

### Context

…

sairina:
#### Description of Change
- Adds new watch property to "watch" for changes in `addresses()`
- Creates new list to hold address ids that are either available or have content
- Adds a check for logic in `submitDisabled()` so that 'Continue' button will stay disabled
- Updates template for `disabled` attributes

#### Manual verification steps performed
- Tested with server that is offline, servers that are always available (beta and demo), and self server. See screenshots below.

1. If server is unreachable, it is disabled and radio button is not selected.
2. If server is reachable or has content, radio button for first available is selected and user can switch between available servers.
3. Continue button is clickable.

Screen Shot 2020-11-13 at 1 45 49 PM

1. If server is unreachable, server is disabled, radio button is not selected.
2. 'Continue' button is also disabled.

Screen Shot 2020-11-13 at 1 46 07 PM

…

### Reviewer guidance
- Please test with additional online/offline servers to verify

…

### References
Fixes #7376 
…

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/core/assets/src/views/sync/SelectAddressModalGroup/SelectAddressForm.vue
+++ b/kolibri/core/assets/src/views/sync/SelectAddressModalGroup/SelectAddressForm.vue
@@ -247,7 +247,7 @@
     watch: {
       addresses(addresses) {
         this.availableAddressIds = addresses
-          .filter(address => address.available || address.hasContent)
+          .filter(address => address.available)
           .map(address => address.id);
         this.resetSelectedAddress();
       },
@@ -275,9 +275,6 @@
             this.savedAddresses = addresses;
             this.stage = this.Stages.FETCHING_SUCCESSFUL;
             this.savedAddressesInitiallyFetched = true;
-            if (this.savedAddresses.find(({ id }) => this.selectedId === id)) {
-              this.selectedAddressId = this.selectedId;
-            }
           })
           .catch(() => {
             this.stage = this.Stages.FETCHING_FAILED;
@@ -285,7 +282,9 @@
       },
       resetSelectedAddress() {
         if (this.availableAddressIds.length !== 0) {
-          this.selectedAddressId = this.availableAddressIds[0];
+          const selectedId = this.selectedId ? this.selectedId : this.selectedAddressId;
+          this.selectedAddressId =
+            this.availableAddressIds.find(id => selectedId === id) || this.availableAddressIds[0];
         } else {
           this.selectedAddressId = '';
         }
--- a/kolibri/core/assets/src/views/sync/SelectAddressModalGroup/SelectAddressForm.vue
+++ b/kolibri/core/assets/src/views/sync/SelectAddressModalGroup/SelectAddressForm.vue
@@ -45,7 +45,7 @@
             :value="a.id"
             :label="a.nickname"
             :description="a.base_url"
-            :disabled="formDisabled || !a.available || !a.hasContent"
+            :disabled="formDisabled || !isAddressAvailable(a.id)"
           />
           <KButton
             v-if="!hideSavedAddresses"
@@ -69,7 +69,7 @@
             :value="d.instance_id"
             :label="formatNameAndId(d.device_name, d.id)"
             :description="d.base_url"
-            :disabled="formDisabled || !d.available || discoveryFailed || !d.hasContent"
+            :disabled="formDisabled || discoveryFailed || !isAddressAvailable(d.id)"
           />
         </div>
       </template>
@@ -166,6 +166,7 @@
     },
     data() {
       return {
+        availableAddressIds: [],
         savedAddresses: [],
         savedAddressesInitiallyFetched: false,
         discoveredAddresses: [],
@@ -190,6 +191,11 @@
       addresses() {
         return this.savedAddresses.concat(this.discoveredAddresses);
       },
+      isAddressAvailable() {
+        return function(addressId) {
+          return Boolean(this.availableAddressIds.find(id => id === addressId));
+        };
+      },
       initialFetchingComplete() {
         return this.savedAddressesInitiallyFetched && this.discoveredAddressesInitiallyFetched;
       },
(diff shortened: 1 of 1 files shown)


kolibri/core/assets/src/views/sync/SelectAddressModalGroup/SelectAddressForm.vue
```diff
@@ -237,6 +244,14 @@
         return null;
       },
     },
+    watch: {
+      addresses(addresses) {
+        this.availableAddressIds = addresses
+          .filter(address => address.available || address.hasContent)
```
micahscopes: One small thing here, I haven't tested it yet, but I think this should be `a.available && a.hasContent`, since `!(!a.available || !a.hasContent)` is the same as `a.available && a.hasContent`.

In other words, we want to consider the address to be "available" if it both has `a.available == true` and `a.hasContent == true`.

kolibri/core/assets/src/views/sync/SelectAddressModalGroup/SelectAddressForm.vue
```diff
@@ -237,6 +244,14 @@
...
+          .map(address => address.id);
+        this.resetSelectedAddress();
```
micahscopes, line 252: Ordinarily, it might make sense to have `this.availableAddressIds` be a computed property, but because you're also calling `this.resetSelectedAddress()` every time `this.addresses` is updated, I think a watcher is the right move.

jonboiser:
I thing I found a possible regression or pre-existing bug (if it's pre-existing, we can file separate issue). I:

1. Added a computer on the network as a "static" address using "add new address". So it appeared twice on the "Select network address" modal (once as a "static" address, and again as a "dynamic" address)
1. I then turned off Kolibri on that computer.
1. The "dynamic" address below the divider disappeared, but the "static" address did not become disabled as expected.

kolibri/core/assets/src/views/sync/SelectAddressModalGroup/SelectAddressForm.vue
```diff
@@ -270,9 +284,8 @@
           });
       },
       resetSelectedAddress() {
-        const availableAddress = find(this.addresses, { available: true });
-        if (availableAddress) {
-          this.selectedAddressId = availableAddress.id;
+        if (this.availableAddressIds.length !== 0) {
+          this.selectedAddressId = this.availableAddressIds[0];
```
micahscopes: Nice consolidation!  Much easier to read this way

kolibri/core/assets/src/views/sync/SelectAddressModalGroup/SelectAddressForm.vue
```diff
@@ -258,7 +273,6 @@
         return fetchStaticAddresses(this.fetchAddressArgs)
           .then(addresses => {
             this.savedAddresses = addresses;
-            this.resetSelectedAddress();
             this.stage = this.Stages.FETCHING_SUCCESSFUL;
             this.savedAddressesInitiallyFetched = true;
             if (this.savedAddresses.find(({ id }) => this.selectedId === id)) {
```
micahscopes: The line below this one seems to restore `this.selectedAddressId` after `this.resetSelectedAddress` just reset it in line 261.  With your change, `resetSelectedAddress` in the `addresses` watcher could get called _after_, which might result in the user's selection getting lost.  If that's the case, it could potentially make sense to retool the `resetSelectedAddress` function? Or else to restore the selected address in the watcher?

kolibri/core/assets/src/views/sync/SelectAddressModalGroup/SelectAddressForm.vue
jonboiser: Actually, the `hasContent` condition doesn't need to hold if you're only syncing Facility data. Maybe it's okay if the address is merely available. If the user is trying to import content, they will be taken to a page with no channels, which is okay.

jonboiser: I think it's okay to just remove the `|| address.hasContent` in the predicate, because I don't want this modal to block us from syncing with devices that might only have Facility/User data but no content.

micahscopes (changes requested):
Hey!  It's looking good.  The logic consolidation is easier to read.  I didn't test it yet, cause I gotta jump off for the day, but I saw a couple of things worth noting.  There's one small logic thing I'm pretty sure needs changing [here](https://github.com/learningequality/kolibri/pull/7691/files#r523283711) (looks like @jonboiser followed up with some more info on that), and another thing [here](https://github.com/learningequality/kolibri/pull/7691#discussion_r523288526) that I'm not so sure about without testing, but that looks like it could use some further investigation.

Regardless, it looks like you're off to a great start here!

micahscopes (approved):
Hey, the updates look good to me.  I also tested manually, including the case for reappearing devices, and as far as I can tell, everything is working as intended.  Bonus: when I shut down a peer instance of Kolibri, waited a bit then started it back up again, my selection was preserved, so I think your changes did the trick.

Great work!

sairina:
Creating a `availableAddressIds` list allowed us to consolidate some of the logic for the `disabled` attribute because we were able to create a new computed property (`isAddressAvailable`) where that logic could live. 

Because the watcher calls `resetSelectedAddress` after the `availableAddressIds` list is updated, we had to make sure that the user choice for `selectedId` would not get overwritten. @bjester refactored `resetSelectedAddress` to make sure that wouldn't happen.

Issue #7709, opened 2020-11-25
kolibri-tools needs improved message extraction for Kolibri plugins


nucleogenesis:
Per the last comments on: https://github.com/learningequality/kolibri/pull/7671

1. We need to update the `kolibri-tools` package to properly handle the plugin paths in Kolibri. The PR was merged and `kolibri-tools` subsequently release to accommodate the changes in that PR which make the i18n tools available cross-product.

2. Context transfer has hard-coded kolibri paths. Should require relatively light refactor to accommodate Studio as well

PR #7710, opened 2020-11-25
kolibri-tools bug fix & packages published bumped versions

nucleogenesis:
Bumps packages versions based on latest npm release for each. 

There were some Kolibri-specific relative paths in some of the i18n code ported to `kolibri-tools` that you wouldn't run into if you were `yarn linking` the package because symlinks. Those are fixed here.

I got all screwed up with the package publishing the first couple times - then since I was already in 3 versions I might as well make one with the latest fix and point Studio there.

### DIFF
--- a/kolibri/core/package.json
+++ b/kolibri/core/package.json
@@ -33,7 +33,7 @@
   },
   "devDependencies": {
     "eslint-plugin-import": "^2.20.0",
-    "kolibri-tools": "0.13.0-dev.10",
+    "kolibri-tools": "0.14.5-dev.4",
     "responselike": "^1.0.2"
   }
 }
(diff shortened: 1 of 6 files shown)


kolibri/utils/i18n.py
```diff
@@ -27,7 +27,10 @@ def get_installed_app_locale_path(appname):
 
 def _get_language_info():
     file_path = os.path.join(
-        os.path.dirname(kolibri.__file__), "locale", "language_info.json"
```
rtibbles: This path resolution will fail in production, so let's just leave this with the duplicated language_info.json files until we've sorted out how to handle that.

rtibbles (changes requested):
Let's revert the language_info.json deduplication, as it will break runtime production Kolibri straight away.

nucleogenesis:
Noting that we'll be fixing this when we circle back to this and resolve https://github.com/learningequality/kolibri/issues/7709 - closing for now.

PR #7714, opened 2020-12-04
Get port number for zeroconf from cherrypy.server

dylanmccall:
### Summary

With this change, it is possible to set `KOLIBRI_HTTP_PORT=0`, or pass Kolibri a socket through socket activation, and have it detect the assigned port number. This affects flathub/org.learningequality.Kolibri#16.

### Reviewer guidance

I added a callback mechanism to the start command. This is really useful for the Kolibri front-end apps (https://github.com/learningequality/kolibri-installer-mac / https://github.com/learningequality/kolibri-installer-gnome) since it means the server will tell us the moment it is listening, and what URL it can be reached at. I am not sure if the way I did that is very kosher here since I was really just cobbling it together until it worked. Guidance on how to do this better (even if it's more complicated) would be appreciated.

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/utils/cli.py
+++ b/kolibri/utils/cli.py
@@ -502,6 +502,26 @@ def start(port, background):
     Start the server on given port.
     """
 
+    start_with_ready_cb(port, background, ready_cb=_on_ready)
+
+
+def _on_ready(urls, bind_addr=None, bind_port=None):
+    if not urls:
+        logger.error(
+            "Could not detect an IP address that Kolibri binds to, but try "
+            "opening up the following addresses:\n"
+        )
+        urls = [
+            "http://{}:{}".format(ip, bind_port) for ip in ("localhost", "127.0.0.1")
+        ]
+    else:
+        logger.info("Kolibri running on:\n")
+    for addr in urls:
+        sys.stderr.write("\t{}\n".format(addr))
+    sys.stderr.write("\n")
+
+
+def start_with_ready_cb(port, background, ready_cb=None):
     # Check if there is an options.ini file exist inside the KOLIBRI_HOME folder
     sanity_checks.check_default_options_exist()
 
@@ -528,25 +548,6 @@ def start(port, background):
     else:
         logger.info("Running Kolibri as background process")
 
-    if serve_http:
-
-        __, urls = server.get_urls(listen_port=port)
-        if not urls:
-            logger.error(
-                "Could not detect an IP address that Kolibri binds to, but try "
-                "opening up the following addresses:\n"
-            )
-            urls = [
-                "http://{}:{}".format(ip, port) for ip in ("localhost", "127.0.0.1")
-            ]
-        else:
-            logger.info("Kolibri running on:\n")
-        for addr in urls:
-            sys.stderr.write("\t{}\n".format(addr))
-        sys.stderr.write("\n")
-    else:
-        logger.info("Starting Kolibri background workers")
-
     # Daemonize at this point, no more user output is needed
     if background:
 
@@ -564,7 +565,7 @@ def start(port, background):
 
         become_daemon(**kwargs)
 
-    server.start(port=port, serve_http=serve_http)
+    server.start(port=port, serve_http=serve_http, ready_cb=ready_cb)
 
 
 @main.command(cls=KolibriCommand, help="Stop the Kolibri process")
--- a/kolibri/utils/server.py
+++ b/kolibri/utils/server.py
@@ -80,12 +80,11 @@ class NotRunning(Exception):
 
 
 class ServicesPlugin(SimplePlugin):
-    def __init__(self, bus, port):
+    def __init__(self, bus):
         self.bus = bus
-        self.port = port
         self.workers = None
 
-    def start(self):
+    def start(self, *args, **kwargs):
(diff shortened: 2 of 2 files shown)


rtibbles:
This seems like a reasonable approach - and it does make sense to defer the zeroconf registration until the server is ready to serve, as otherwise requests initiated based on zeroconf recognition will error out.

This also makes even more sense if we actually exposed a good chunk of the CLI API on the `kolibri.` namespace as a main entry point API as per: #6197 (not asking to resolve that in the scope of here, but just to give context).

Would also mean that apps would no longer have to poll a status endpoint to get this info.

dylanmccall:
Oh, that's a good ticket to know about. I was just thinking about how doing that would help to make the weird callbacks thing here less weird since we could do a bit of subclassing instead - and perhaps use the same mechanism to split zeroconf setup a bit more  from the cherrypy server init. (But of course doing all of that would definitely be scope creep for this branch :b).

kolibri/utils/cli.py
```diff
@@ -15,6 +15,7 @@
 import os
 import signal
 import sys
+from functools import partial
```
danigm: This is imported but not used. The linter is failing:
https://github.com/learningequality/kolibri/pull/7714/checks?check_run_id=1646492052

kolibri/utils/cli.py
```diff
@@ -502,6 +503,26 @@ def start(port, background):
...
+
+
+def _on_ready(urls):
+    if not urls:
+        logger.error(
+            "Could not detect an IP address that Kolibri binds to, but try "
+            "opening up the following addresses:\n"
+        )
+        urls = [
+            "http://{}:{}".format(ip, bind_port) for ip in ("localhost", "127.0.0.1")
```
danigm, line 515: `bind_port` is not defined. Maybe the partial was imported to do that and you forgot it?

Something like:
```
start_with_ready_cb(port, background, ready_cb=partial(_on_ready, port))

def _on_ready(bind_port, urls):
```

danigm (changes requested):

rtibbles:
Seems that running the command `make staticdeps-cext` is failing in this PR, not entirely sure what the cause of that is. Again, the approach looks sound, and I have no concerns with this being merged, assuming that it passes manual testing and the check failures are addressed.

kolibri/utils/cli.py
dylanmccall: Thanks for spotting this! Indeed, I ended up breaking some things when I reorganized this yesterday :) I changed server.py to pass `bind_port` and `bind_addr` to the `on_ready` callback, and that should take care of it.

dylanmccall:
Hmm, looks like the failure in `make staticdeps-cext` is happening elsewhere as well. For instance: https://github.com/learningequality/kolibri/runs/1645602601. (Some of that is skipped in the release-0.14.x branch, I think). I wonder if it's a pypi thing? Pip is dropping (has dropped?) support for Python 2.7 this month, so that could be related, although I'm puzzled how that would result in the particular errors it is giving us.

rtibbles:
Hrm, pip 21 is where they will drop python 2.7 support and that's not out yet, so I don't think that's the issue just yet.

rtibbles:
Hrm, I am replicating some of the errors locally - it seems to be something to do with non-ARM-Linux architecture Python 3.4 cryptography libraries.

rtibbles:
This error is happening in the buildkite build also - but seemingly only for builds happening on develop. Possible it is happening because of adding in Python 3.9 support?

kolibri/utils/server.py
```diff
@@ -395,6 +387,24 @@ def handler(signum, frame):
...
+
+    # Write the PID file again, in case the port number has changed
+    _write_pid_file(PID_FILE, port=bind_port)
+
+    # Register the Kolibri zeroconf service so it will be discoverable on the network
+    from kolibri.core.discovery.utils.network.search import (
+        register_zeroconf_service,
+    )
+
+    register_zeroconf_service(port=bind_port)
```
benjaoming, line 402: Hi LE friends! Just passing by to see what new stuff is happening with Kolibri :heart: 

I was wondering a bit about this change since CherryPy isn't the only HTTP server that can be in front of Kolibri. So if the Zeroconf service should be started together with the other background services, then the old location is better?

rtibbles: Hrm - that's a good point, this would mess up use of zeroconf behind an nginx/uwsgi deployment for example. Good catch.

There is the slight advantage here that in the cherrypy instance it only registers zeroconf once the port is actually available, so we may need to go for a two pronged approach.

rtibbles:
@dylanmccall I've done a PR that should address the issue that @benjaoming raised, here: https://github.com/learningequality/kolibri/pull/8026 - this will make the zeroconf listener register on the set port if the server is not being run by Kolibri.

dylanmccall:
Oh, neat! Sorry I didn't get around to following up on this, but looks like the change I'm proposing here kind of belongs over there anyway :) Thanks for letting me know, @rtibbles. I'll keep an eye on #8026.

rtibbles:
Closing in favour of #8026

PR #7718, opened 2020-12-07
Pin ubuntu to 18.04 for tox tests for Python3.4 compatibility.

rtibbles:
### Summary
* Python setup action requires ubuntu 18.04 for Python34
* We were using ubuntu-latest which recently updated to 20.04
* This pins the tox action to ubuntu 18.04

### Reviewer guidance
* Do tests run?


rtibbles:
Looks like the appropriate tests didn't actually run on this PR, but they ran on my fork and confirm the fix: https://github.com/rtibbles/kolibri/runs/1512876215?check_suite_focus=true

PR #7720, opened 2020-12-11
create Freedesktop Desktop file and channel icon when importing channel

dnohales:
This is a draft, don't merge until we discuss all the TODOs and the motivation in general.

### Summary

This PR add support for creating desktop launchers compatible with [Freedesktop's Desktop Entry Specificatioin](https://specifications.freedesktop.org/desktop-entry-spec/). This basically creates the folder `KOLIBRI_HOME/content/share` to add the desktop files and export the channel icons there. Then the OS will add that directory to its `XDG_DATA_DIRS` env variable.

The motivation is to have a better desktop integration for Kolibri, in this particular case, for Linux (or Freedesktop based DE), having Kolibri channels as desktop launchers.

### TODOs

- [ ] Discuss support for something like this in different platforms (is it wanted?). Right now, one particularity of this is that no user are affect unless the OS opt-in by adding the new directory to `XDG_DATA_DIRS`.
- [ ] Using channel thumbnail for launcher icon may be not suitable for desktop launchers due to the different restrictions they had compared to the desktop specification. We could mitigate that either by:
    - Adding restrictions to the thumbnail (make existing channels that don't fit the restrictions broken)
    - Creating a new field for channels, being `icon`, fallback to `thumbnail`.
    - Crop the thumbnail the best we can when exporting the launcher icon (thinking of libraries like [autocrop](https://github.com/leblancfg/autocrop)), this is compatible with the previous item.

### Reviewer guidance

To test these changes you have to add `KOLIBRI_HOME/content/share` to `XDG_DATA_DIRS`. This can be achieved by, for example, in a modern `systemd` based system, by creating the following executable file in `/usr/lib/systemd/user-environment-generators/61-kolibri`:

```
#!/bin/bash

# This will work for Flatpak Kolibri, change it if required
XDG_DATA_DIRS="$HOME/.var/app/org.learningequality.Kolibri/data/kolibri/share:${XDG_DATA_DIRS:-/usr/local/share:/usr/share}"
echo "XDG_DATA_DIRS=$XDG_DATA_DIRS"
```

Then reboot the computer and add/delete channels to see the launchers.

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/core/content/management/commands/deletecontent.py
+++ b/kolibri/core/content/management/commands/deletecontent.py
@@ -8,6 +8,7 @@ from kolibri.core.content.models import ChannelMetadata
 from kolibri.core.content.models import LocalFile
 from kolibri.core.content.utils.annotation import propagate_forced_localfile_removal
 from kolibri.core.content.utils.annotation import set_content_invisible
+from kolibri.core.content.utils.freedesktop import delete_channel_freedesktop_files
 from kolibri.core.content.utils.import_export_content import get_import_export_data
 from kolibri.core.content.utils.importability_annotation import clear_channel_stats
 from kolibri.core.content.utils.paths import get_content_database_file_path
@@ -50,6 +51,7 @@ def delete_metadata(channel, node_ids, exclude_node_ids, force_delete):
 
     if delete_all_metadata:
         logger.info("Deleting all channel metadata")
+        delete_channel_freedesktop_files(channel.id)
         with db_task_write_lock:
             channel.delete_content_tree_and_files()
 
--- a/kolibri/core/content/management/commands/importcontent.py
+++ b/kolibri/core/content/management/commands/importcontent.py
@@ -7,6 +7,7 @@ from django.core.management.base import CommandError
 from le_utils.constants import content_kinds
 
 from ...utils import annotation
+from ...utils import freedesktop
 from ...utils import paths
 from ...utils import transfer
 from kolibri.core.content.errors import InvalidStorageFilenameError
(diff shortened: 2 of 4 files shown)


rtibbles:
> Discuss support for something like this in different platforms (is it wanted?). Right now, one particularity of this is that no user are affect unless the OS opt-in by adding the new directory to XDG_DATA_DIRS.

I think the main thing to be mindful of here is that this is very much 'app' specific functionality. We have created a specific plugin for app specific code, and also created an interface for different apps to register functionality against that. I think this might be best exposed as a 'channel launcher' capability in the app plugin, that specific apps can then leverage in order to do whatever they need to do.

> Using channel thumbnail for launcher icon may be not suitable for desktop launchers due to the different restrictions they had compared to the desktop specification. We could mitigate that either by:
Adding restrictions to the thumbnail (make existing channels that don't fit the restrictions broken)
Creating a new field for channels, being icon, fallback to thumbnail.
Crop the thumbnail the best we can when exporting the launcher icon (thinking of libraries like autocrop), this is compatible with the previous item.

One advantage of making this implemented specifically in different app platforms is that we could do the third option, but do it flexibly in a way that suits the specific platform for the app rather than trying to enforce some sort of one size fits all (different platforms will prefer e.g. circular versus rounded square icons).

dylanmccall:
I opened a new pull request which has this idea but as a standalone Kolibri plugin, in #7838.

rtibbles:
Closing as stale.

PR #7724, opened 2020-12-12
Customize session middleware to allow exempting views from session middleware.

rtibbles:
### Summary
* It is possible for heavily cached endpoints to include session expiry headers
* This means that a response with those session expiry headers could be stored in a long term cache
* This will then cause anyone accessing those endpoints to be immediately logged out
* This fixes that by exempting long running caches from session middleware
* To prevent cache collisions for recommendation endpoints, caching is removed from them

…

### Reviewer guidance
- Login to the devserver
- Change `SESSION_COOKIE_AGE` in base settings to 1
- Request the channelmetadata endpoint /api/content/channel/
- Note that the set-cookie header is not set

…

### DIFF
--- a/kolibri/core/auth/middleware.py
+++ b/kolibri/core/auth/middleware.py
@@ -2,6 +2,7 @@ from django.apps import apps
 from django.conf import settings
 from django.contrib.auth import get_user
 from django.contrib.auth.middleware import AuthenticationMiddleware
+from django.contrib.sessions.middleware import SessionMiddleware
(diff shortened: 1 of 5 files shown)


jamalex (approved):
I do worry a bit about performance hits from removing caching on the "popular" endpoint in particular, as it has a ton of data it queries (and isn't user-specific), but it doesn't look easy to do something specific to that one. Let's revisit before shipping to production, at least.

PR #7728, opened 2020-12-23
Restrict API for device info for non-superadmin users

### LINKED ISSUE #6650: non-superadmins with "can manage content" permission should only see the Device > Channels tab (and not Info, Settings)
### Observed behavior

As a learner with permission to manage content, I can still see the device info and settings tabs:

image

### Expected behavior

* Non-super-admins should only see the content tab
* the API that returns device info should also be protected by super admin permission

### User-facing consequences

Currently:

* Info tab leaks privileged information
* Settings tab is locked down but confusing because the user is in fact logged in

| Info | settings |
|--|--|
|image |image|

### Steps to reproduce

* give a learner permission to manage content
* log in and view device tab

### Context

0.13.x

sairina:
### Summary

1. Added check to remove privileged information from `Device Info` endpoint for non-superadmin users: `/api/device/deviceinfo/` now only returns the free space available. Superadmin users still can see all device info.
2. For frontend, added restricted, "unauthorized" message for non-superadmins in case there is a frontend issue, similar to: image

…

### Reviewer guidance

**Check that the endpoint is still reachable and returns 'free space' by the following users:**

1. Coach
2. Learner that can manage content
3. Admin

**Check that the endpoint is still reachable and returns all device info by the following user:**
1. Superadmin

…

### References

Addresses #6650 

…

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/core/device/api.py
+++ b/kolibri/core/device/api.py
@@ -113,6 +113,19 @@ class DeviceInfoView(views.APIView):
         info["python_version"] = "{major}.{minor}.{micro}".format(
             major=version_info.major, minor=version_info.minor, micro=version_info.micro
         )
+
+        if not request.user.is_superuser:
+            # If user is not superuser, return just free space available
+            info.pop("version")
+            info.pop("urls")
+            info.pop("database_path")
+            info.pop("device_id")
+            info.pop("os")
+            info.pop("server_time")
+            info.pop("server_timezone")
+            info.pop("installer")
+            info.pop("python_version")
+
         return Response(info)
 
 
--- a/kolibri/core/device/api.py
+++ b/kolibri/core/device/api.py
@@ -115,16 +115,19 @@ class DeviceInfoView(views.APIView):
         )
 
         if not request.user.is_superuser:
-            # If user is not superuser, return just free space available
-            info.pop("version")
-            info.pop("urls")
-            info.pop("database_path")
-            info.pop("device_id")
-            info.pop("os")
-            info.pop("server_time")
-            info.pop("server_timezone")
-            info.pop("installer")
-            info.pop("python_version")
+            # If user is not superuser, return just free space available and kolibri version
+            keys_to_remove = [
+                "urls",
+                "database_path",
+                "device_id",
+                "os",
+                "server_time",
+                "server_timezone",
+                "installer",
+                "python_version",
+            ]
+            for key in keys_to_remove:
+                info.pop(key)
 
         return Response(info)
 
--- a/kolibri/core/device/api.py
+++ b/kolibri/core/device/api.py
@@ -127,7 +127,7 @@ class DeviceInfoView(views.APIView):
                 "python_version",
             ]
             for key in keys_to_remove:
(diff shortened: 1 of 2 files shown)


kolibri/core/device/api.py
```diff
@@ -113,6 +113,19 @@ def get(self, request, format=None):
         info["python_version"] = "{major}.{minor}.{micro}".format(
             major=version_info.major, minor=version_info.minor, micro=version_info.micro
         )
+
+        if not request.user.is_superuser:
+            # If user is not superuser, return just free space available
+            info.pop("version")
```
jredrejo: why is `version` restricted? we show it even in the kolibri login page

jredrejo (changes requested):
Hi @sairina I think your code works properly, but in order to be more legible I'd use this approach to remove multiple keys from the dict:
https://www.kite.com/python/answers/how-to-remove-multiple-keys-from-a-dictionary-in-python#use-del 
You see in a single line what keys are going to be removed

kolibri/core/device/api.py
sairina: Thanks for catching that! I updated it. Are there any additional keys that non-superadmins need access to?

jredrejo: honestly, I think the rest of the keys are not important for non-superadmins, urls might be but the superadmin might want not to show all of them, so I think it's good as it is

kolibri/core/device/api.py
```diff
@@ -113,6 +113,22 @@ def get(self, request, format=None):
...
+                "database_path",
+                "device_id",
+                "os",
+                "server_time",
+                "server_timezone",
+                "installer",
+                "python_version",
+            ]
+            for key in keys_to_remove:
+                info.pop(key)
```
jredrejo: I know I'm being nitpicking but if you are not going to use the retrieved value, `del` is recommended instead of `pop`

jredrejo (approved):

kolibri/core/device/api.py
sairina: Not nitpicking - helpful for me to know! I'll fix.

sairina: Fixed

kolibri/core/device/api.py
```diff
@@ -113,6 +113,22 @@ def get(self, request, format=None):
...
             major=version_info.major, minor=version_info.minor, micro=version_info.micro
         )
+
+        if not request.user.is_superuser:
+            # If user is not superuser, return just free space available and kolibri version
+            keys_to_remove = [
+                "urls",
+                "database_path",
+                "device_id",
+                "os",
```
jamalex, line 123: Note that `device_id` (`instance_id`) and `os` are in [/api/public/info/](https://kolibri-demo.learningequality.org/api/public/info/), as they are needed for identification (and showing OS icons) with peer discovery. So we don't currently treat those as private info. But those are accessed through a different endpoint, so this won't affect it.

PR #7731, opened 2021-01-05
Add `clearable` property to `task` API payload

### LINKED ISSUE #7704: Move task-is-clearable flag to the API layer, instead of using frontend code to derive this from `task.status`
This is a bit of tech debt that we run into when developing UIs for presenting and managing tasks

There is this piece of JS that is used in a lot of places to determine whether a task is clearable

https://github.com/learningequality/kolibri/blob/577c10e9670a10dbc1955dffc6e7dfe6098bbe57/kolibri/plugins/device/assets/src/constants.js#L58-L61

I think this concept of "clearability" belongs on the API layer instead not just because it is so ubiquitous in the UI (and could save us from having to import this utility function all over the place), but because "clearability" is tied to whether a Task ID is a valid ID for the `/api/tasks/cleartask(s)` endpoints. 

This could probably be implemented by adding some code to this function here:

https://github.com/learningequality/kolibri/blob/release-v0.14.x/kolibri/core/tasks/api.py#L1222-L1244

Then most/all of the instances where we call `isTaskClearble(task)` could be replaced by simply reading `task.is_clearable` etc.

jonboiser:
### Summary

1. Moves the concept of a task being "clearable" from the front-end via the `taskIsClearable` utility to the backend by adding a `task.clearable` property to the serialized tasks
2. Updates front-end code to directly access `task.clearable` rather than using the utility.

### Reviewer guidance

Check that task-related UIs still work correctly (e.g. facility importing/syncing, content importing, CSV uploading, etc.)

### References

Fixes #7704 

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/core/tasks/api.py
+++ b/kolibri/core/tasks/api.py
@@ -1229,6 +1229,7 @@ def _job_to_response(job):
             "progress": [],
             "id": None,
             "cancellable": False,
+            "clearable": False,
         }
     else:
         output = {
@@ -1238,6 +1239,7 @@ def _job_to_response(job):
             "percentage": job.percentage_progress,
             "id": job.job_id,
             "cancellable": job.cancellable,
+            "clearable": job.state in [State.FAILED, State.CANCELED, State.COMPLETED],
         }
         output.update(job.extra_metadata)
         return output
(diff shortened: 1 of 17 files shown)


rtibbles (commented):
Code read through gets the OK from me.

PR #7775, opened 2021-01-30
Update progress using summary spent_time, not session

### LINKED ISSUE #7764: Resources are not completed, even after waiting for 53 minutes
### Observed behavior

In Chile UNW servers, after upgrading from 0.12.9 to 0.14.3 they have seen that after a long time waiting for the resources to be completed they are not marked

### Expected behavior
Completion time should be within the times set in the algorithm (5 minutes in apps, 30 seconds/page in pdf files)

### User-facing consequences
Users can't progress in their learning path

### Errors and logs
Provided by one partner:

image

### Steps to reproduce
In Kolibri 0.14.6

1. Install a channel having resources of kind "App" (html5)
2. Go to one of the resources that are of kind: "app"
3. Begin to see it, wait for a couple of minutes and then, go back to the channel page, exiting the resource
4. Enter again in the same resource, contentsummarylog will restart the completion percentage, while the time_spent accumulates correctly

Checked this does not happen in 0.12.9 nor 0.13.3

One more thing:
In 0.14.6 the resource has to be open for 300 seconds to be marked as completed, while in 0.12.9 & 0.13.3 only needs 120 seconds

Some logs:
0.14.6
```
  http://localhost:8080/api/logger/contentsummarylog/7b518585aadc3cc92f34645b409affef/
{"start_timestamp":"2021-01-26T17:53:04.885Z","end_timestamp":"2021-01-26T17:55:04.992Z","progress":0.4000666666666667,"time_spent":120}
{"start_timestamp":"2021-01-26T17:53:04.885Z","end_timestamp":"2021-01-26T17:56:03.934Z","progress":0.5500933333333333,"time_spent":178.942}

<--- back to the channel list and again to the resource
{"end_timestamp":"2021-01-26T17:56:53.521Z","time_spent":179.39700000000002}
{"end_timestamp":"2021-01-26T17:58:54.008Z","progress":0.37312666666666666,"time_spent":299.884}
{"end_timestamp":"2021-01-26T18:00:43.237Z","progress":0.78039,"time_spent":408.942}
{"end_timestamp":"2021-01-26T18:01:29.738Z","progress":0.8804066666666667,"time_spent":455.443}

<--- back to the channel list and again to the resource
{"end_timestamp":"2021-01-26T18:01:56.559Z","time_spent":455.938}
{"end_timestamp":"2021-01-26T18:04:01.177Z","progress":0.41037666666666667,"time_spent":580.556}
```

0.13.3
```
{"id":"f493826a2b76f524730704d6cab2406a","user":"1855464ef46ec21092d537bfb6059894","content_id":"a3de31cb87415435bf9d7233cfa97852","channel_id":"5472ce04133b499db5368ff35dc82b5c","start_timestamp":"2021-01-26T21:02:58.425000+01:00","currentmasterylog":null,"end_timestamp":"2021-01-26T21:02:58.425000+01:00","completion_timestamp":null,"time_spent":0.0,"progress":0.0,"kind":"html5","extra_fields":{}}
{"time_spent":0.446,"start_timestamp":"2021-01-26T20:02:58.425Z","extra_fields":{"contentState":{"cookie":{"byPath":{},"rootCookies":{}},"localStorage":{}}},"end_timestamp":"2021-01-26T20:02:58.975Z"}
{"time_spent":30.039,"progress":0.15005000000000002,"start_timestamp":"2021-01-26T20:02:58.425Z","end_timestamp":"2021-01-26T20:03:28.568Z"}

<--- back to the channel list and again to the resource
{"time_spent":30.468,"end_timestamp":"2021-01-26T20:04:36.792Z"}
{"time_spent":90.039,"progress":0.6502066666666667,"end_timestamp":"2021-01-26T20:05:36.363Z"}
{"time_spent":120.039,"progress":1,"end_timestamp":"2021-01-26T20:06:06.363Z","completion_timestamp":"2021-01-26T20:06:06.414Z"}
```

### Context
UNW Chile server
No available information on the user browser or OS but the attached screenshot looks like Chrome on W10

jredrejo:
### Summary

Users are complaining because spending long time (more than 50 minutes in some cases) in some resources they don't get it marked as completed. The reason was the user spends the time entering in the resource in different moments.

This PR changes the criteria to update the summary log in several renderers.
Renderers that use time as a completion criteria were restarting the completion percentage everytime the user goes back to the resource. 

Users should see its summary progress percentage increase when continuing reading/viewing a previous resource, so the completion is ended in the right completion time. 

### Reviewer guidance
In any channel enter into a pdf, epub or  html5 app.
Check that exiting/entering several times in the resource does not avoid it to be completed.

### References

Closes: #7764 

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/plugins/h5p_viewer/assets/src/views/Html5AppRendererIndex.vue
+++ b/kolibri/plugins/h5p_viewer/assets/src/views/Html5AppRendererIndex.vue
@@ -35,6 +35,7 @@
 
 <script>
 
+  import { mapGetters } from 'vuex';
   import { now } from 'kolibri.utils.serverClock';
   import UiIconButton from 'kolibri-design-system/lib/keen/UiIconButton';
   import CoreFullscreen from 'kolibri.coreVue.components.CoreFullscreen';
@@ -53,6 +54,7 @@
       };
     },
     computed: {
+      ...mapGetters(['summaryTimeSpent']),
       name() {
         return nameSpace;
       },
@@ -67,7 +69,6 @@
       });
       this.hashi.initialize((this.extraFields && this.extraFields.contentState) || {});
       this.$emit('startTracking');
-      this.startTime = now();
       this.pollProgress();
     },
     beforeDestroy() {
@@ -78,8 +79,8 @@
     },
     methods: {
       recordProgress() {
-        const totalTime = now() - this.startTime;
-        this.$emit('addProgress', Math.max(0, totalTime / 300000));
+        const totalTime = this.summaryTimeSpent * 1000;
+        this.$emit('updateProgress', Math.max(0, totalTime / 300000));
         this.pollProgress();
       },
       pollProgress() {
--- a/kolibri/plugins/h5p_viewer/assets/src/views/Html5AppRendererIndex.vue
+++ b/kolibri/plugins/h5p_viewer/assets/src/views/Html5AppRendererIndex.vue
@@ -79,7 +79,7 @@
     methods: {
       recordProgress() {
         const totalTime = now() - this.startTime;
-        this.$emit('updateProgress', Math.max(0, totalTime / 300000));
+        this.$emit('addProgress', Math.max(0, totalTime / 300000));
         this.pollProgress();
       },
       pollProgress() {
--- a/kolibri/core/assets/src/state/modules/logging.js
+++ b/kolibri/core/assets/src/state/modules/logging.js
@@ -17,6 +17,9 @@ export default {
     sessionTimeSpent(state) {
       return state.session.time_spent;
     },
+    summaryTimeSpent(state) {
+      return state.summary.time_spent;
+    },
   },
   mutations: {
     SET_LOGGING_SUMMARY_STATE(state, summaryState) {
--- a/kolibri/plugins/epub_viewer/assets/src/views/EpubRendererIndex.vue
+++ b/kolibri/plugins/epub_viewer/assets/src/views/EpubRendererIndex.vue
@@ -452,7 +452,7 @@
     },
     beforeDestroy() {
       this.updateContentState();
-      this.updateProgress();
+      this.addProgress();
       this.$emit('stopTracking');
       window.removeEventListener('mousedown', this.handleMouseDown, { passive: true });
       clearInterval(this.updateContentStateInterval);
@@ -461,9 +461,9 @@
       delete global.ePub;
     },
     methods: {
-      updateProgress() {
+      addProgress() {
         if (this.locations.length > 0) {
-          this.$emit('updateProgress', this.sessionTimeSpent / this.expectedTimeToRead);
+          this.$emit('addProgress', this.sessionTimeSpent / this.expectedTimeToRead);
         }
       },
       handleReadyRendition() {
@@ -480,7 +480,7 @@
         this.book.locations.generate(LOCATIONS_INTERVAL).then(locations => {
           this.locations = locations;
           this.$emit('startTracking');
-          this.updateContentStateInterval = setInterval(this.updateProgress, 30000);
+          this.updateContentStateInterval = setInterval(this.addProgress, 30000);
 
           // Update current location, .currentLocation() can return Promise or value
           Promise.resolve()
--- a/kolibri/plugins/epub_viewer/assets/src/views/EpubRendererIndex.vue
+++ b/kolibri/plugins/epub_viewer/assets/src/views/EpubRendererIndex.vue
@@ -226,7 +226,7 @@
       };
     },
     computed: {
-      ...mapGetters(['sessionTimeSpent']),
+      ...mapGetters(['summaryTimeSpent']),
       isAtStart() {
         return get(this.rendition, 'location.atStart', false);
       },
@@ -452,7 +452,7 @@
     },
     beforeDestroy() {
       this.updateContentState();
-      this.addProgress();
+      this.updateProgress();
       this.$emit('stopTracking');
       window.removeEventListener('mousedown', this.handleMouseDown, { passive: true });
       clearInterval(this.updateContentStateInterval);
@@ -461,9 +461,9 @@
       delete global.ePub;
     },
     methods: {
-      addProgress() {
+      updateProgress() {
         if (this.locations.length > 0) {
-          this.$emit('addProgress', this.sessionTimeSpent / this.expectedTimeToRead);
+          this.$emit('updateProgress', this.summaryTimeSpent / this.expectedTimeToRead);
         }
       },
       handleReadyRendition() {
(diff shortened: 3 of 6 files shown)


kollivier:
@jredrejo This code appears to re-introduce the bug with progress calculation that we were originally trying to fix. Can you confirm that with this PR, the HTML5 renderer marks a node as 100% complete in 5 minutes, as opposed to in 90 seconds, as happened previously? `addProgress` adds to the saved progress value, so it should be getting the percent changed since the last `addProgress` call, but it seems in this PR the intent is to set total progress completed instead.

Also, were the PDF, ePub and Slideshow renderers all affected by the same problem?

Looking at the code, I think the issue might simply be that the HTML5 renderer isn't using `sessionTimeSpent` when calculating the progress, which is why progress calculation starts from 0 on each load.

jredrejo:
> @jredrejo This code appears to re-introduce the bug with progress calculation that we were originally trying to fix.

Sorry, I was not aware of a previous bug

> Can you confirm that with this PR, the HTML5 renderer marks a node as 100% complete in 5 minutes, as opposed to in 90 seconds, as happened previously? `addProgress` adds to the saved progress value, so it should be getting the percent changed since the last `addProgress` call, but it seems in this PR the intent is to set total progress completed instead.
> 

No, you're right `addProgress` is also wrong, but `updateProgress` in the way is coded is wrong too if used the way they were.

> Also, were the PDF, ePub and Slideshow renderers all affected by the same problem?
> 
PDF & ePub are for sure, for Slideshow I'm not sure as I am not sure of the intended completion rule, but I think it's not affected looking at the code

> Looking at the code, I think the issue might simply be that the HTML5 renderer isn't using `sessionTimeSpent` when calculating the progress, which is why progress calculation starts from 0 on each load.

no , sessionTimeSpent is restarted every time the user returns to the resource. That's the source problem for the epub & pdf renderers.

There are a couple of possible fixes:

1.  use summaryTimeSpent for the calculations
2. change the update method so it does take into account increments in summary

As 2 could cause problems with other code, I've implemented 1

rtibbles:
I'm also realizing that there's an issue with some of our renderers - as they should not be accessing getter state inside them, because that will cause weird behaviour when loaded in a non-content page context. Don't think we should fix that in the course of this PR, but something we need to resolve at some point.

kolibri/plugins/h5p_viewer/assets/src/views/Html5AppRendererIndex.vue
```diff
@@ -78,7 +79,7 @@
     },
     methods: {
       recordProgress() {
-        const totalTime = now() - this.startTime;
```
rtibbles, line 81: This is actually the correct way for a renderer to calculate the time spent - we should be using this in the other renderers, not using vuex getters.

jredrejo: This calculates the time spent in the current session, and it's fine for that purpose.
However this time was being used to calculate the completion percentage and it's the reason why users are complaining of spending 50 minutes in the resource without being able to complete it.
If the user reads a pdf or see some html5 content in different days because he does not complete the whole time in a row, using totalTime will make the user unable to get the completion percentage if it's restarted everytime he begins because the previous spent time is not taken into account to calculate the completion percentage.

For example, in a pdf with 40 pages, completion time will be 40 x 30 = 1200 seconds. If the user reads 30 pages spending 15 minutes, and returns the next day to continue, all the previous time is discarded if we don't use the previous time spent (avaialable in the vuex getter) to calculate the percentage, so he might be reading the document forever without getting the completion time if he never spends those 1200 seconds (20 min) in a single session.

rtibbles: Right - that's why we should be using `addProgress` not `updateProgress` - we should be adding to the total progress achieved from the progress in this session. If `addProgress` is not doing that properly, then that's a bug in `addProgress`.

If I read for 2 minutes on day one, and that gives me 12% completion.
On day two I read for 3 minutes and that gives me 18% completion.

If we add the progress, it should give a total of 30% completion, without any need to refer to to the previous day's progress.

kollivier: To use `addProgress`, we need to change the renderers to start passing the percent of progress since the last `addProgress` call was made. 

Right now, most renderers are sending the progress since session start instead, so when `addProgress` is used for updates, it leads to updates like this (with 2% progress each X seconds):

progress += 2%
progress += 4%
progress += 6%
progress += 8%
progress += 10%

So after five progress updates, instead of progress being the expected 10%, it is actually 30%.

rtibbles: Right - because most of the renderers were originally using `updateProgress` so were not doing the incremental changes, so we need to update the implementation to handle that.

rtibbles: (or at least they were coded that way - I think it may have just been totally wrong originally)

kollivier: Originally, there was only `updateProgress`, and some renderers (mainly video and audio) used it to add to progress, while others used it like a setter. 

I suspect it turned out this way because people are naturally inclined to code progress as a percent of completion rather than calculating a differential, and the term `update` usually implies setting rather than incrementing or otherwise combining the old and new value.

To address this, I created `addProgress` as an attempt to support both approaches and clarify the differences between them:

https://github.com/learningequality/kolibri/commit/95b6783abfbe919cf5e0a70b9491e1c84c7f8730

I think there are two separate issues here:

1) fixing the progress calculation bug, and
2) refactoring renderers to work in non-content contexts (I'm not familiar with this, and since many renderers use `sessionTimeSpent` getter, it appears to be a long-standing issue.)

Are you proposing that (2) should be fixed as part of this work? Otherwise, I would think the next action would be to create an issue for (2) and not restrict use of state getters in this PR as many renderers are already doing this.

rtibbles: Yeah, I'd be fine with just fixing 1) - I just don't want the fix to add to 2) in the process if avoidable :)

jredrejo: ok, so back to the first idea this PR had: 
- Inside a session we add progresss not update to avoid resetting previous session time
- The addProgress function in the store should be changed to avoid accumulating percentages, as Kevin commented previously currently, the renderers are sending the whole session percentage, causing a wng calculation.

But  taking into account that most of the renderers are doing it incorrectly, probably since ever, they all have to be modified.

jonboiser (approved):
Tested this out manually with some PDFs and ASB HTML5s and I can see the progress getting updated at a more reasonable rate.

PR #7658, opened 2020-11-03
Fixing issue #6892: collect and report aggregate stats about client browsers through telemetry

paulbusse:
### Summary
The change collects the HTTP user agent in `(kolibri/core/auth/api.py)SessionViewset.get_session_response` and passes it on to `UserSessionLog.update_log`. The latter function uses the `ua_parser` module to parse the HTTP user agent string into the requested string. and saves it to the database.

The data is summarized in extract_facility_statistics.

Other changes:
* `ua_parser 0.10.0` was added to the dependencies
* `logger/migrations/0008_usersessionlog_device_info.py` was generated
* `test_extract_facility_statistic` was adapted to add the new field

### Reviewer guidance
* the `update_log` definition has been adapted. I could only find one call to the function. So I did not default the new argument.
* `ua_parser` is very fault resistant, I used that in the code. I don't check for `None` on the families
* I used a multiline statement to generate the device_info string, that may be against coding guidelines.

I tested this using all available browsers (chrome, edge, firefox, comodo dragon) on all available os's (windows 10, ubuntu). The pingback is not tested.

### References
* #6892 

----

### Contributor Checklist

PR process:

- [x] PR has the correct target branch and milestone
- [ ] PR has 'needs review' or 'work-in-progress' label
- [ ] If PR is ready for review, a reviewer has been added. (Don't use 'Assignees')
- [ ] If this is an important user-facing change, PR or related issue has a 'changelog' label
- [ ] If this includes an internal dependency change, a link to the diff is provided

Testing:

- [ ] Contributor has fully tested the PR manually
- [ ] If there are any front-end changes, before/after screenshots are included
- [ ] Critical user journeys are covered by Gherkin stories
- [ ] Critical and brittle code paths are covered by unit tests

### Reviewer Checklist

- Automated test coverage is satisfactory
- PR is fully functional
- PR has been tested for [accessibility regressions](http://kolibri-dev.readthedocs.io/en/develop/manual_testing.html#accessibility-a11y-testing)
- External dependency files were updated if necessary (`yarn` and `pip`)
- Documentation is updated
- Contributor is in AUTHORS.md

PR #7781, opened 2021-02-03
Set null=True and blank=True for UserSessionLog device info

rtibbles:
## Summary
* Rectifies an oversight in review where the device_info was not allowed to be set to null and blank and only blank
* This would not have resulted in migration failures for historic data, but might be problematic for different versions of Kolibri syncing with each other

## References
Follow up to #7658

----

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated


jonboiser (approved):
Code looks good and DB structure looks good after running migrations

PR #7811, opened 2021-02-24
Require default values for all optional Vue props

apurva-modi:
Fixed some of the linting errors that pop up while enabling 'vue/require-default-prop'

### DIFF
--- a/kolibri/core/assets/src/views/CoreBase/index.vue
+++ b/kolibri/core/assets/src/views/CoreBase/index.vue
@@ -245,7 +245,8 @@
       // link to where the 'back' button should go
       immersivePageRoute: {
         type: Object,
-        required: true,
+        required: false,
+        default: null,
       },
       // determines the color, primary being the classic kolibri appbar color
       immersivePagePrimary: {
--- a/kolibri/core/assets/src/views/CoreBase/index.vue
+++ b/kolibri/core/assets/src/views/CoreBase/index.vue
@@ -220,15 +220,15 @@
       },
       authorizedRole: {
         type: String,
-        required: false,
+        default: null,
       },
       authorizationErrorHeader: {
         type: String,
-        required: false,
+        default: null,
       },
       authorizationErrorDetails: {
         type: String,
-        required: false,
+        default: null,
       },
       // IMMERSIVE-SPECIFIC
       immersivePage: {
@@ -245,7 +245,7 @@
       // link to where the 'back' button should go
       immersivePageRoute: {
         type: Object,
-        required: false,
+        required: true,
       },
       // determines the color, primary being the classic kolibri appbar color
       immersivePagePrimary: {
--- a/kolibri/core/assets/src/views/CoreMenu/CoreMenuDivider.vue
+++ b/kolibri/core/assets/src/views/CoreMenu/CoreMenuDivider.vue
@@ -1,0 +1,33 @@
+<template>
+
+  <li>
+    <span
+      class="divider"
+      :style="{ borderTop: `solid 1px ${$themeTokens.fineLine}` }"
+    >
+    </span>
+  </li>
+
+</template>
+
+
+<script>
+
+  export default {
+    name: 'CoreMenuDivider',
+  };
+
+</script>
+
+
+<style lang="scss" scoped>
+
+  .divider {
+    display: block;
+    min-width: 100%;
+    height: 1px;
+    margin: 8px 0;
+    overflow-y: hidden;
+  }
+
+</style>
--- a/kolibri/core/assets/src/views/CoreMenu/CoreMenuOption.vue
+++ b/kolibri/core/assets/src/views/CoreMenu/CoreMenuOption.vue
@@ -42,11 +42,26 @@
   export default {
     name: 'CoreMenuOption',
     props: {
-      isDivider: Boolean,
-      label: String,
-      link: String,
-      secondaryText: String,
-      icon: String,
+      isDivider: {
+        type: Boolean,
+        default: false,
+      },
+      label: {
+        type: String,
+        default: null,
+      },
+      link: {
+        type: String,
+        default: null,
+      },
+      secondaryText: {
+        type: String,
+        default: null,
+      },
+      icon: {
+        type: String,
+        default: null,
+      },
     },
     inject: ['showActive'],
     computed: {
--- a/kolibri/core/assets/src/views/CoreMenu/CoreMenuOption.vue
+++ b/kolibri/core/assets/src/views/CoreMenu/CoreMenuOption.vue
@@ -1,14 +1,7 @@
 <template>
 
   <li>
-    <span
-      v-if="isDivider"
-      class="divider"
-      :style="{ borderTop: `solid 1px ${$themeTokens.fineLine}` }"
-    >
-    </span>
     <a
-      v-else
       :href="link"
       class="core-menu-option"
       role="menuitem"
@@ -42,13 +35,9 @@
   export default {
     name: 'CoreMenuOption',
     props: {
-      isDivider: {
-        type: Boolean,
-        default: false,
-      },
       label: {
         type: String,
-        default: null,
+        required: true,
       },
       link: {
         type: String,
@@ -60,7 +49,7 @@
       },
       icon: {
         type: String,
-        default: null,
+        required: true,
       },
     },
     inject: ['showActive'],
@@ -125,14 +114,6 @@
     &:hover {
       outline-offset: -1px; // override global styles
     }
-  }
-
-  .divider {
-    display: block;
-    min-width: 100%;
-    height: 1px;
-    margin: 8px 0;
-    overflow-y: hidden;
   }
 
 </style>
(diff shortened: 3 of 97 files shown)


packages/kolibri-tools/.eslintrc.js
```diff
@@ -85,7 +85,7 @@ module.exports = {
       },
     ],
     'vue/attribute-hyphenation': [ERROR, 'never'],
-    'vue/require-default-prop': OFF,
+    'vue/require-default-prop': ERROR,
```
jonboiser, line 88: You need to revert this line for now until all of the `vue/require-default-prop` errors are fixed. This is because the remaining errors will cause the "GitHub Actions / All file linting" check to fail.
```suggestion
    'vue/require-default-prop': OFF,
```

apurva-modi: I amended the commit as per you request. If rest is upto your liking I can go ahead and fix all the lint errors.

jonboiser: Ok, sounds good. You can fix all of these `'vue/require-default-prop'` errors in this PR, then convert it from a draft to "ready for review" when you're done.

kolibri/core/assets/src/views/CoreBase/index.vue
```diff
@@ -245,7 +245,7 @@
       // link to where the 'back' button should go
       immersivePageRoute: {
         type: Object,
-        required: false,
+        required: true,
```
jonboiser: I don't think this should be required, since it's only needed if the `immersivePage` prop is true. You will potentially see a lot of prop validation errors on non-immersive pages.

kolibri/core/assets/src/views/CoreMenu/CoreMenuOption.vue
```diff
@@ -42,11 +42,26 @@
   export default {
     name: 'CoreMenuOption',
     props: {
-      isDivider: Boolean,
-      label: String,
-      link: String,
-      secondaryText: String,
-      icon: String,
+      isDivider: {
```
jonboiser: I think this is pointing out that the `isDivider = true` is a special case and should be separated out as its own component because its inclusion forces us to make `label` and `icon` optional props, when in the more common usage, they are required.

kolibri/core/assets/src/views/KSelect/KeenUiSelect.vue
```diff
@@ -211,13 +214,22 @@
           return [];
         },
       },
-      placeholder: String,
-      icon: String,
+      placeholder: {
+        type: String,
+        default: null,
```
jonboiser:
```suggestion
        default: '',
```

Since this might actually put a string literal `"null"` in some situations.

kolibri/core/assets/src/views/KSelect/KeenUiSelect.vue
```diff
@@ -242,7 +254,10 @@
         type: String,
         default: 'Search',
       },
-      filter: Function,
+      filter: {
+        type: Function,
+        default: () => null,
```
jonboiser:
```suggestion
        default: null,
```

See line 341 of this component. This might cause a bug if you provide a default implementation of `filter` as a function.

kolibri/plugins/coach/assets/src/views/plan/LessonSummaryPage/ManageLessonModals.vue
```diff
@@ -47,6 +47,7 @@
       // Should be 'COPY' or 'DELETE'
       currentAction: {
         type: String,
+        default: null,
```
jonboiser: Can you improve this? Make it a required prop and then add a validator function. That will make the comment above unnecessary.

apurva-modi: okay!

kolibri/plugins/device/assets/src/views/ManageContentPage/ChannelPanel/ChannelDetails.vue
```diff
@@ -70,11 +70,12 @@
     props: {
       channel: {
         type: Object,
+        default: () => {},
```
jonboiser:
```suggestion
        default: () => ({}),
```

kolibri/plugins/device/assets/src/views/ManageContentPage/ChannelPanel/WithCheckbox.vue
```diff
@@ -50,11 +50,12 @@
     props: {
       channel: {
         type: Object,
+        default: () => {},
```
jonboiser:
```suggestion
        default: () => ({}),
```

kolibri/plugins/device/assets/src/views/ManageContentPage/FilteredChannelListContainer.vue
```diff
@@ -73,14 +73,15 @@
...
+        default: () => [],
       },
       selectAllCheckbox: {
         type: Boolean,
         default: false,
       },
       selectedChannels: {
         type: Array,
-        required: false,
+        default: () => null,
```
jonboiser:
```suggestion
        default: () => [],
```

This also needs to be an array

kolibri/plugins/media_player/assets/src/views/MediaPlayerTranscript/TranscriptCue.vue
```diff
@@ -43,18 +43,27 @@
   export default {
     name: 'TranscriptCue',
     props: {
-      cue: Object,
-      mediaDuration: Number,
+      cue: {
+        type: Object,
+        default: () => {},
```
jonboiser:
```suggestion
        default: () => ({}),
```

In general, the `default` needs to be a function that returns an empty object. Otherwise `cue` will be defaulted to `undefined` if it is using `() => {}`

kolibri/core/assets/src/views/CoreMenu/CoreMenuDivider.vue
```diff
@@ -0,0 +1,50 @@
...
+  };
+
+</script>
+
+
+<style lang="scss" scoped>
+
+  @import '~kolibri-design-system/lib/styles/definitions';
+
+  .core-menu-option {
```
jonboiser: you can delete this entire class since it's not used. also the `@import`, since i don't see any imports used in the `.divider` class

kolibri/core/assets/src/views/CoreMenu/CoreMenuOption.vue
```diff
@@ -42,11 +35,22 @@
   export default {
     name: 'CoreMenuOption',
     props: {
-      isDivider: Boolean,
-      label: String,
-      link: String,
-      secondaryText: String,
-      icon: String,
+      label: {
```
jonboiser, line 38: make `label` and `icon` required

kolibri/plugins/learn/assets/src/views/MasteredSnackbars/index.vue
```diff
@@ -105,11 +105,11 @@
     props: {
       nextContent: {
         type: Object,
-        required: false,
+        default: () => {},
```
jonboiser:
```suggestion
        default: () => ({}),
```

kolibri/plugins/learn/assets/src/views/MasteredSnackbars/index.vue
```diff
@@ -105,11 +105,11 @@
...
       },
       nextContentLink: {
         type: Object,
-        required: false,
+        default: () => {},
```
jonboiser:
```suggestion
        default: () => ({}),
```

kolibri/plugins/device/assets/src/views/ManageContentPage/SelectionBottomBar.vue
```diff
@@ -59,19 +59,21 @@
       },
       resourceCounts: {
         type: Object,
-        required: false,
+        default: () => {},
```
jonboiser:
```suggestion
        default: () => ({}),
```

kolibri/core/assets/src/views/CoreMenu/CoreMenuOption.vue
```diff
@@ -42,11 +35,22 @@
...
+        type: String,
```
jonboiser, line 39:
```suggestion
        type: String,
        required: true,
```

kolibri/core/assets/src/views/CoreMenu/CoreMenuOption.vue
```diff
@@ -42,11 +35,22 @@
...
+        default: null,
```
jonboiser:
```suggestion
        type: String,
        required: true,
```

kolibri/core/assets/src/views/CoreMenu/CoreMenuOption.vue
```diff
@@ -42,11 +35,22 @@
...
+      },
+      secondaryText: {
+        type: String,
+        default: null,
+      },
+      icon: {
+        type: String,
+        default: null,
```
jonboiser:
```suggestion
        type: String,
        required: true,
```

jonboiser (approved):

PR #7826, opened 2021-03-09
Extract NewPasswordPage from SignInPage, and make new route

### LINKED ISSUE #7321: Sign In Page Improvements Needed (Follow up to #7306)
### Observed behavior

This issue is to track and manage the fact that I am intentionally leaving some things for post-release in the Sign In refactor done on #7306 

- ~username/password don't need to be in Vuex~ Fixed in https://github.com/learningequality/kolibri/pull/7335
- SignInPage is monolithic and its parts ought to be broken out. Password create & sign in probably should be its own route.

### Expected behavior

Better code so the next person who dives into the SignIn part of user has less mental overhead.

### User-facing consequences

None.

jonboiser:
### Summary

1. Adds a new `NewPasswordPage`, which is assigned to a new route `/set-password`
2. Adds a test for the new component
3. In the test environment, adds a `global.flushPromises` utility, following some advice in [these docs](https://vue-test-utils.vuejs.org/guides/testing-async-components.html)
4. Refactors some tests to remove the hacky multiple calls to `nextTick`.

### Reviewer guidance

1. Does the Sign-In page still work in "learner has to update password when facility changes password requirement setting from not-required to required" flow?
2. Is the handling for edge cases good enough? e.g. when a user goes directly to a `/set-password` URL and the learner has already set their password, or the `username` query param is invalid.

### References

Closes #7321

(We should write separate issues for the other suggested refactors, since they are a little bit more complicated).

This might add some tech debt because the error states just redirect back to the sign-in page. But this is not a regression since these errors were unhandled originally. We would need some specs on how to be more informative to the user when these errors occur.

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated
edit

### DIFF
--- a/kolibri/plugins/coach/assets/src/views/common/notifications/__test__/ActivityList.spec.js
+++ b/kolibri/plugins/coach/assets/src/views/common/notifications/__test__/ActivityList.spec.js
@@ -72,8 +72,7 @@ function makeWrapper(options) {
 describe('ActivityList component', () => {
   it('on first render, calls the notification resource', async () => {
     const { fetchMock, wrapper } = makeWrapper({});
-    await wrapper.vm.$nextTick();
-    await wrapper.vm.$nextTick();
+    await global.flushPromises();
     expect(fetchMock).toHaveBeenCalled();
     expect(wrapper.vm.moreResults).toBe(false);
   });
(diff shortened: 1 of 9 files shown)


kolibri/plugins/coach/assets/src/views/common/notifications/__test__/ActivityList.spec.js
```diff
@@ -72,8 +72,7 @@ function makeWrapper(options) {
 describe('ActivityList component', () => {
   it('on first render, calls the notification resource', async () => {
     const { fetchMock, wrapper } = makeWrapper({});
-    await wrapper.vm.$nextTick();
-    await wrapper.vm.$nextTick();
+    await global.flushPromises();
```
nucleogenesis, line 75: _Swanky_

kolibri/plugins/user/assets/src/views/SignInPage/index.vue
```diff
@@ -148,56 +148,6 @@
         @userSelected="setSelectedUsername"
       />
     </div>
-
-    <!-- TODO: This can be its own separate component -->
-    <!--
-        Learner was created without a password, but now must create one.
-        This ought to be routed separately.
-      -->
```
nucleogenesis, line 156: From TODO to TODONE :fist_oncoming:

nucleogenesis (approved):
Looks good to me. Thanks for this!

Once the red x below is a green check this should be good to go!

PR #7856, opened 2021-03-15
Enforce server-consistent settings

ashmeet13:
### Summary
Added a function for validating consistency in FacilityDataset model. Learners cannot have the following setting - 
```
learner_can_login_with_no_password: true
learner_can_edit_password: true
```
This has been implemented from the client side already. This PR brings the consistency to server side logic

### Reviewer guidance
A unit test has been added in `test_datasets.py` to test the case of both these values being set to `True`.
A database `IntegrityError` would be raised if this case is reached.

### References
Resolves the example in #7281
There could be more inconsistencies yet to be identified.   

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/core/auth/models.py
+++ b/kolibri/core/auth/models.py
@@ -139,11 +139,7 @@ class FacilityDataset(FacilityDataSyncableModel):
             return "FacilityDataset (no associated Facility)"
 
     def save(self, *args, **kwargs):
-        try:
-            self.ensure_compatibility()
-        except IncompatibleDeviceSettingError as e:
-            raise IncompatibleDeviceSettingError(str(e))
-
+        self.ensure_compatibility()
         super(FacilityDataset, self).save(*args, **kwargs)
 
     def ensure_compatibility(self, *args, **kwargs):
--- a/kolibri/core/auth/models.py
+++ b/kolibri/core/auth/models.py
@@ -48,6 +48,7 @@ from .constants import facility_presets
 from .constants import morango_sync
 from .constants import role_kinds
 from .constants import user_kinds
+from .errors import IncompatibleDeviceSetting
 from .errors import InvalidRoleKind
 from .errors import UserDoesNotHaveRoleError
 from .errors import UserHasRoleOnlyIndirectlyThroughHierarchyError
@@ -136,6 +137,23 @@ class FacilityDataset(FacilityDataSyncableModel):
             )
         else:
             return "FacilityDataset (no associated Facility)"
+
+    def save(self, *args, **kwargs):
+        try:
+            self.ensure_compatibility()
+        except IncompatibleDeviceSetting as e:
+            raise IntegrityError(str(e))
+
+        super(FacilityDataset, self).save(*args, **kwargs)
+
+    def ensure_compatibility(self, *args, **kwargs):
+        if self.learner_can_login_with_no_password and self.learner_can_edit_password:
+            raise IncompatibleDeviceSetting(
+                "Device Settings [learner_can_login_with_no_password={}] & [learner_can_edit_password={}] values incompatible.".format(
+                    self.learner_can_login_with_no_password,
+                    self.learner_can_edit_password,
+                )
+            )
 
     def calculate_source_id(self):
         # if we don't already have a source ID, get one by generating a new root certificate, and using its ID
--- a/kolibri/core/auth/models.py
+++ b/kolibri/core/auth/models.py
@@ -48,7 +48,7 @@ from .constants import facility_presets
 from .constants import morango_sync
 from .constants import role_kinds
 from .constants import user_kinds
-from .errors import IncompatibleDeviceSetting
+from .errors import IncompatibleDeviceSettingError
 from .errors import InvalidRoleKind
 from .errors import UserDoesNotHaveRoleError
 from .errors import UserHasRoleOnlyIndirectlyThroughHierarchyError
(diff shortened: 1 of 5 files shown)


jonboiser:
Thanks, this looks good! It should not be possible from the front-end (but might be through something like cURL), but with this validation, you can crash the server by trying to update these facility settings through the REST API. For example (replace the UUID with your own facility ID and the `user:pass` with your own credentials):

```
## Incomptaible Device Settings
curl -X "PATCH" "http://localhost:8000/api/auth/facilitydataset/684094eaecc970e5d8acacab348c706b/" \
     -H 'Content-Type: application/json; charset=utf-8' \
     -u 'user:pass' \
     -d $'{
  "learner_can_login_with_no_password": "true",
  "learner_can_edit_password": "true"
}'

```

Will result in a 500 error because of these new validations.

It would be good to handle these more gracefully as a `400 Bad Request` error. I believe this can probably be done on the `FacilityDatasetSerializer`, and might be able to leverage the new validations you add.

https://www.django-rest-framework.org/api-guide/serializers/#validation

jonboiser (changes requested):
See comment below. When updating `DeviceSettings` through the rest API, return a 400 error instead of not handling the `IntegrityErrored` raised by this new validation code.

ashmeet13:
I have added the validation for `FacilityDatasetSerializer` but was not able to test it with the method you suggested. I tried the following command - 
```
curl -X "PATCH" "http://localhost:8000/api/auth/facilitydataset/fc395a304ffa9c28718481419e5f868d/" \
     -H 'Content-Type: application/json; charset=utf-8' \
     -u 'superuser:password' \
     -d $'{
  "learner_can_login_with_no_password": "true",
  "learner_can_edit_password": "true"
}'
``` 
and kept getting the following error - ` [{"id":"NOT_FOUND","metadata":{"view":"Facility Dataset Instance"}}]`

Could not understand how to add a test case for this too in `test_api.py`. Could you please point towards how exactly should a test for this be done?

jonboiser:
Can you make sure `fc395a304ffa9c28718481419e5f868d` is a valid facility dataset ID? You can change the `PATCH` to `GET` to see if you also get a 404.

For the test, you can add the new test to this `TestCase` class and use the unit tests within as examples. https://github.com/learningequality/kolibri/blob/develop/kolibri/core/auth/test/test_api.py#L889

ashmeet13:
I tried again after resetting my db.

I wasn't sure how exactly to get my facility id - I used the following URL `127.0.0.1:8000/api/auth/facility/` to get the following output
```
[
    {
        "id": "63de7f49d2c7aab8c305b24f0268810d",
        "name": "Home Facility Facility",
        "num_users": 1,
        "num_classrooms": 0,
        "last_synced": null,
        "dataset": {
            "id": "589f4421f58bdcb87edbfdb14f57fd01",
            "learner_can_edit_username": true,
            . . .
     }
]
```
Using `63de7f49d2c7aab8c305b24f0268810d` I tried both `GET` and `PATCH` and received the same error. With `GET` the logs also gave me a 404  
`INFO     127.0.0.1 - - "GET /api/auth/facilitydataset/63de7f49d2c7aab8c305b24f0268810d/" 404 0 "" "curl/7.58.0"`

jonboiser:
You should be using the `dataset` ID in your request URL. So use `589f4421f58bdcb87edbfdb14f57fd01` instead

ashmeet13:
Thank you for the help. Was able to test it using cURL.
I have tested it both through cURL and have added a unit test for the serializer validate.

kolibri/core/auth/test/test_api.py
```diff
@@ -987,6 +987,21 @@ def post_resetsettings():
             response = post_resetsettings()
             self.assertDictContainsSubset(mappings[setting], response.data)
 
+    def test_validation_for_incompatible_settings(self):
```
rtibbles: There is another case here, where the two values are set in sequential requests. I would suggest making a second test case which does two requests, the first one setting `"learner_can_login_with_no_password": "true",` and the second setting `"learner_can_edit_password": "true",` if the validation is working correctly, then this should produce a 400.

I think it will probably produce a 500 because of an unhandled IntegrityError however.

kolibri/core/auth/models.py
```diff
@@ -137,6 +138,23 @@ def __str__(self):
         else:
             return "FacilityDataset (no associated Facility)"
 
+    def save(self, *args, **kwargs):
+        try:
```
rtibbles: I think better to just let this specific error be raised, so that we can distinguish it from an `IntegrityError` which will generally indicate some sort of database level constraint violation, which is not the case here.

kolibri/core/auth/serializers.py
```diff
@@ -84,6 +84,21 @@ class Meta:
             "preset",
         )
 
+    def validate(self, attrs):
```
rtibbles: Because of the second test case I suggested below, this validation will pass in cases where it should not.

I think it will be more effective (and DRYer) to not use the DRF Serializer validation here, but rather to override the serializer save method, and catch the `IncompatibleDeviceSetting` error that I have suggested should be raised by the model save method.

This would look something like:
```
def save(self, **kwargs):
    try:
        return super(FacilityDatasetSerializer, self).save(**kwargs)
    except IncompatibleDeviceError:
        raise serializers.ValidationError(....
```

rtibbles (changes requested):
This will not guard against all cases, and will need some modifications that I suggest in order to pass tests for these.

kolibri/core/auth/test/test_api.py
ashmeet13: I missed this. My bad. I have added the test case. Also added another case where if only `"learner_can_login_with_no_password": "true"` is set, a 400 is checked for since `"learner_can_edit_password": "true"` is a default value.

kolibri/core/auth/models.py
ashmeet13: Got it. I have made this change.

kolibri/core/auth/serializers.py
ashmeet13: Yup. Catching the error in `save` makes it simpler. Thank you for the help.

kolibri/core/auth/models.py
rtibbles: Can remove the extra 'try/except' clause here, and just call `self.ensure_compatibility()` and let the error raise from there - no need to reraise it.

kolibri/core/auth/test/test_datasets.py
```diff
@@ -71,3 +71,9 @@ def test_dataset_representation(self):
         )
         new_dataset = FacilityDataset.objects.create()
         self.assertEqual(str(new_dataset), "FacilityDataset (no associated Facility)")
+
+    def test_dataset_incompatible_setting(self):
+        with self.assertRaises(IntegrityError):
```
rtibbles: You have one failing test, as this needs to be updated to the `IncompatibleDeviceSettingError` now!

rtibbles (changes requested):
Just two more things to clean up, but looking good over all!

ashmeet13:
Done. My bad for not running the tests before pushing.

ashmeet13:
@jonboiser @rtibbles could you please review this PR.
Thank you.

rtibbles:
Not completely your fault, as we currently have an issue with two of our CI checks failing!

rtibbles (approved):
Great work, thanks for the contribution!

PR #7865, opened 2021-03-16
Fix #7386: Make reusable NewBadge component.

### LINKED ISSUE #7386: create a reusable 'New' badge component
### Observed behavior

as of https://github.com/learningequality/kolibri/pull/7368 the 'New' label is used in 3 places, with duplicated styling and strings:

> image

### Expected behavior

badge should be be a re-usable core component

### Context

0.14.0

YashJipkate:
### Summary

Extract code and logic of the new badge to a reusable component - NewBadge.

### Reviewer guidance

The element is used in three places:
- `NewChannelVersionBanner.vue`
- `WithSizeAndOptions.vue`
- `WithImportDetails.vue`

### References

Fixes #7386.

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated


kolibri/plugins/device/assets/src/views/ManageContentPage/NewBadge.vue
```diff
@@ -0,0 +1,40 @@
...
+        type: String,
+        required: true,
+      },
+    },
+  };
+
+</script>
+
+
+<style>
```
jonboiser:
```suggestion
<style lang="scss" scoped>
```

Could you convert these to scoped styles? You will also see that the `NewBadge` component is still inheriting styles from the parent components. If you make them scoped, they lose the styling:

image

jonboiser (changes requested):

kolibri/plugins/device/assets/src/views/ManageContentPage/NewBadge.vue
YashJipkate: Done!

jonboiser: It's not enough to just change the scoping. You'll need to transfer the styles from the `new-badge` classes into the component, and also make sure that the styling is responsive. For example, in `WithSizeAndOptionslvue`, the `.new-label.panel-sm` styles would have been applied when the window is narrow before you extracted `NewBadge`. With this PR, that no longer happens.

jonboiser: It might be good enough to just add another `new-label` class to the `NewBadge` component like this `<NewBadge class="new-label">`. Then both the styles from the parent and from within the component will be applied.

YashJipkate: I am sorry but I'm not able to reproduce this. I am getting a perfectly aligned component (even without the scoping).
Screenshot from 2021-03-17 01-58-38

The `.new-label.panel-sm` style applies too.
Screenshot from 2021-03-17 01-58-50

Is there something I am missing?

kolibri/plugins/device/assets/src/views/ManageContentPage/NewBadge.vue
```diff
@@ -0,0 +1,40 @@
...
+      },
+    },
+  };
+
+</script>
+
+
+<style lang="scss" scoped>
+
+  .new-label {
```
indirectlylit: This `new-label` class definition has moved _inside_ the `NewBadge` component but has conflicting definitions that remain in parent components.

Classes used on the root node of a child component are a subtle case where vue's scoped styles mechanism breaks down, and it's possible for class definitions to collide with between child and parent component styles.

See https://vue-loader.vuejs.org/guide/scoped-css.html#child-component-root-elements

Therefore, it's important that the class name used inside the badge component is different than class names defined and applied from outside the badge.

indirectlylit (dismissed):

kolibri/plugins/device/assets/src/views/ManageContentPage/NewBadge.vue
YashJipkate: Hi @indirectlylit!
Thanks for the information, it was new to me.
I have only put the common stylings - the 'badge specific' styles like padding and font size - in the `NewBadge` component, as should be the case. The stylings for its positioning relative to the external components are defined in the parent components. Hence the same class name won't create conflicts but compliment each other. 

But I agree with you, the parent and child shouldn't be this tightly coupled. I have changed the class name of the component. Please take a look.
Thanks!

indirectlylit: thank you! that addresses my concern.

I haven't tested locally and will let @jonboiser or others give the final OK to merge

jonboiser (approved):
looks good, thanks!

PR #7866, opened 2021-03-16
Added validation for question_id UUID in QuestionSourceSerializer

### LINKED ISSUE #3094: Provide validation for `Exam.question_sources` JSON field when creating a new Exam model
I noticed that the `Exam` has a JSON field for `question_sources`, and has a doc string explaining what the shape of `question_sources` should be:

https://github.com/learningequality/kolibri/blob/3b4c56f3409f1f1133eb73ebdad72729749a37b9/kolibri/core/exams/models.py#L31-L39

But the `ExamSerializer` does not seem to validate `question_sources` when POSTing a new Exam. The exercise should exist before it is attached to an exam. (in case of a race condition where a question is deleted before the exam is created).

Maybe we should validate `question_sources` and other un-typed JSON fields in the API.

What are some strategies for doing this?

apurva-modi:
### Summary

The serializer was already checking exercise_id for valid hex field, so I added the same for the question_id in the QuestionSourceSerializer and also added the test cases to validate the change made.

…

### Reviewer guidance
Test cases can be run to test the change.
kolibri manage test -- kolibri.core.exams

…

----

### Contributor Checklist

PR process:

Testing:

### DIFF
--- a/kolibri/core/exams/serializers.py
+++ b/kolibri/core/exams/serializers.py
@@ -28,7 +28,8 @@ class NestedCollectionSerializer(ModelSerializer):
 
 class QuestionSourceSerializer(Serializer):
     exercise_id = HexUUIDField(format="hex")
-    question_id = CharField()
+    # V0 need not have question_id that is why required=False
+    question_id = HexUUIDField(format="hex", required=False)
     title = CharField()
     counter_in_exercise = IntegerField()
 
--- a/kolibri/core/exams/test/test_exam_api.py
+++ b/kolibri/core/exams/test/test_exam_api.py
@@ -2,6 +2,8 @@ from __future__ import absolute_import
 from __future__ import print_function
 from __future__ import unicode_literals
 
+import uuid
+
 from django.core.urlresolvers import reverse
 from rest_framework import status
 from rest_framework.test import APITestCase
@@ -217,8 +219,8 @@ class ExamAPITestCase(APITestCase):
         exam["title"] = title
         exam["question_sources"].append(
             {
-                "exercise_id": "e1e1e1e1e1e1e1e1e1e1e1e1e1e1e1e1",
-                "question_id": "q1",
+                "exercise_id": uuid.uuid4().hex,
+                "question_id": uuid.uuid4().hex,
                 "title": "Title",
                 # missing 'counter_in_exercise'
             }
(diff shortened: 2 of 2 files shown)


kolibri/core/exams/test/test_exam_api.py
```diff
@@ -233,6 +233,20 @@ def test_exam_with_invalid_exercise_id(self):
         exam["question_sources"].append(
             {
                 "exercise_id": "e1",
+                "question_id": "e1e1e1e1e1e1e1e1e1e1e1e1e1e1e1e1",
```
jonboiser: For the valid UUIDs in the fixtures, can you use a realistic UUID? Either create one via the `uuid` module or generate and copy-paste it into the code.

kolibri/core/exams/serializers.py
```diff
@@ -28,7 +28,7 @@ class Meta:
 
 class QuestionSourceSerializer(Serializer):
     exercise_id = HexUUIDField(format="hex")
-    question_id = CharField()
+    question_id = HexUUIDField(format="hex", required=False)
```
jonboiser, line 32: Is it really possible to have an optional `question_id`?

apurva-modi: [because of the V0](https://github.com/learningequality/kolibri/blob/dc6a62c9f4579d3f0ffbbe6692c5b5a51d6550e0/kolibri/core/exams/models.py#L63):
        JSON array describing exercise nodes this exam draws questions from,
        how many from each, and the node titles at the time of exam creation:
       `[
            {
                "exercise_id": <exercise_pk>,
                "number_of_questions": 6,
                "title": <exercise_title>
            },
            ...
        ]`

jonboiser: I see! Could you add a comment above the field explaining that?

kolibri/core/exams/test/test_exam_api.py
```diff
@@ -233,6 +235,20 @@ def test_exam_with_invalid_exercise_id(self):
...
+        )
+        response = self.post_new_exam(exam)
+        self.assertEqual(response.status_code, 400)
+
+    def test_exam_with_invalid_question_id(self):
+        self.login_as_admin()
+        exam = self.make_basic_exam()
+        exam["question_sources"].append(
+            {
+                "exercise_id": "e1e1e1e1e1e1e1e1e1e1e1e1e1e1e1e1",
```
jonboiser: Also replace these valid `exercise_id`s with realistic UUID values

kolibri/core/exams/serializers.py
rtibbles: I think we should actually enforce this - whenever a V0 quiz is loaded into the frontend, it should be converted into a V1 quiz by the frontend code, so if anyone edits a V0 quiz, it will be resaved as a V1 quiz, so validation should only be needed for V1.

kolibri/core/exams/test/test_exam_api.py
apurva-modi: okay

kolibri/core/exams/serializers.py
apurva-modi: I'll do that

jonboiser (approved):

PR #7883, opened 2021-03-17
Add Boolean field `is_leaf`  to serializers to replace common "leaf or branch" checks in frontend

### LINKED ISSUE #7802: Add something like `is_leaf` Boolean field to serialized ContentNodes to replace common "leaf or branch" checks in JS code
In the `ContentNodeViewSet` (and other related viewsets), it would be nice to have a field like `is_leaf` that returns `true` whenever the node is a resource/leaf and `false` if it is a topic or a channel.

There are dozens of places in the JS code where we have to check if this is the case based on the `kind` field that could be simplified if we had this field.

image

kmrinal19:
### Summary

This pull request incorporates the following changes:
* Add a Boolean field `is_leaf` in the backend serializers to indicate whether a node is a resource/leaf.
* Use field `is_leaf` to replace common "leaf or branch" checks in the frontend

### References
Fixes #7802

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated


kolibri/core/content/serializers.py
```diff
@@ -34,6 +34,7 @@ class ChannelMetadataSerializer(serializers.ModelSerializer):
     lang_name = serializers.SerializerMethodField()
     available = serializers.SerializerMethodField()
     num_coach_contents = serializers.IntegerField(source="root.num_coach_contents")
+    is_leaf = serializers.SerializerMethodField()
```
jonboiser: Don't include `is_leaf` on serializers for `ChannelMetadata` or anything that's not a ContentNode. On the frontend, we may use the `ContentCard` for lots of non-ContentNode things (which we should probably move away from), but we know channels and exams, etc. are not leaf nodes already.

kolibri/plugins/learn/assets/src/views/ChannelCard.vue
```diff
@@ -119,7 +122,7 @@
     computed: {
       ...mapGetters(['isLearner', 'isUserLoggedIn']),
       isTopic() {
-        return this.kind === ContentNodeKinds.TOPIC || this.kind === ContentNodeKinds.CHANNEL;
```
jonboiser, line 122: Like I said, I don't think it's necessary to put an `is_topic` field on a serialized Channel object. I'm not sure, but this `isTopic` field might be unnecessary, if the `channels` array in the `contents` prop is all channels. If this is the case, we can just say that `isTopic` is a constant function returning `true`. It's possible that we left this code when copy-pasting it from something else 🤷 

https://github.com/learningequality/kolibri/blob/5815bd7d04f368e3e967341f518e4b5c1ee55357/kolibri/plugins/learn/assets/src/views/ChannelsPage.vue#L8-L14

kolibri/plugins/learn/assets/src/views/TopicsPage.vue
```diff
@@ -137,9 +136,8 @@
       },
     },
     methods: {
-      genContentLink(id, kind) {
-        const routeName =
-          kind === ContentNodeKinds.TOPIC ? PageNames.TOPICS_TOPIC : PageNames.TOPICS_CONTENT;
+      genContentLink(id, isLeaf) {
+        const routeName = !isLeaf ? PageNames.TOPICS_TOPIC : PageNames.TOPICS_CONTENT;
```
jonboiser:
```suggestion
        const routeName = isLeaf ? PageNames.TOPICS_CONTENT : PageNames.TOPICS_TOPIC;
```
to make the `true` case first

jonboiser (changes requested):
Looks good! But Please investigate if we can remove the `is_leaf` field from `ChannelMetadataSerializer` and just hard code the `isLeaf = true` prop on the front-end for cards showing channels.

kolibri/core/logger/serializers.py
```diff
@@ -27,6 +27,7 @@
 class ContentSessionLogSerializer(KolibriModelSerializer):
 
     extra_fields = serializers.JSONField(default="{}")
+    is_leaf = serializers.SerializerMethodField()
```
jonboiser: I don't think `is_leaf` is required for this serializer.

rtibbles: Yes, none of the changes in this file should be kept.

kmrinal19: Sorry, my bad. I must have left those while refactoring.

jonboiser (approved):
Looks good, thank you!

I tested this out manually browsing resources on the Learn Page and also in the import/export workflow on the Device Page.

PR #7887, opened 2021-03-17
Infer dataset change

### LINKED ISSUE #7859: Update Lesson and Exam model infer_dataset method to use the collection rather than assigned_by foreign key
When Lesson and Exam models were first created, their `infer_dataset` method referred to the user that it was assigned by using the `assigned_by` ForeignKey.

In a scenario where only a limited subset of a facility's data is synchronized to a device, such as in a single learner syncing scenario, this would cause Lessons and Quizzes to remain forever serialized in the Morango store, as the data for the assigning user would be absent, as outside the partition of data that would be synced for a single user.

In order to ensure that Lesson and Exam models can be properly deserialized, their `infer_dataset` method should instead rely on the `collection` ForeignKey, as the full Collection hierarchy of the Facility will be shared during a single user sync.

udithprabhu:
### Summary
Changed infet_dataset from Exam and Lesson from created_by to collection

…

### References
Fixes https://github.com/learningequality/kolibri/issues/7859

…

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/core/auth/test/test_datasets.py
+++ b/kolibri/core/auth/test/test_datasets.py
@@ -108,7 +108,7 @@ class FacilityDatasetTestCase(TestCase):
             self.lesson.cached_related_dataset_lookup("created_by"),
         )
         self.assertEqual(self.exam.infer_dataset(), self.facility.dataset.id)
-   
+
     def test_dataset_incompatible_setting(self):
         with self.assertRaises(IncompatibleDeviceSettingError):
             FacilityDataset.objects.create(
--- a/kolibri/core/auth/test/test_datasets.py
+++ b/kolibri/core/auth/test/test_datasets.py
@@ -9,6 +9,8 @@ from ..models import Facility
 from ..models import FacilityDataset
 from ..models import FacilityUser
 from ..models import LearnerGroup
+from kolibri.core.exams.models import Exam
+from kolibri.core.lessons.models import Lesson
 
 
 class FacilityDatasetTestCase(TestCase):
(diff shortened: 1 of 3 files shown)


jamalex:
Thanks for the contribution, @udithprabhu!

As you noted in the issue, a simple test case could be very helpful to add here. I would suggest perhaps adding it as a new testing class [in here](https://github.com/learningequality/kolibri/blob/release-v0.14.x/kolibri/core/auth/test/test_datasets.py). You'll need to create a Facility, a FacilityUser (linked to the Facility), a Classroom (linked to the Facility), and an Exam (linked to the Classroom and FacilityUser), and then you can call `cached_related_dataset_lookup` with both fields to compare them, as you say.

udithprabhu:
> Thanks for the contribution, @udithprabhu!
> 
> As you noted in the issue, a simple test case could be very helpful to add here. I would suggest perhaps adding it as a new testing class [in here](https://github.com/learningequality/kolibri/blob/release-v0.14.x/kolibri/core/auth/test/test_datasets.py). You'll need to create a Facility, a FacilityUser (linked to the Facility), a Classroom (linked to the Facility), and an Exam (linked to the Classroom and FacilityUser), and then you can call `cached_related_dataset_lookup` with both fields to compare them, as you say.

I have made the changes as per your suggestions. I have listed all the conditions that I could think of. If I have missed any, please let me know.

jamalex (approved):
This looks great, thank you so much!

The failing test seems unrelated to these changes, so it shouldn't be an issue, but I'll wait for someone to confirm that before we do the final merge.

PR #7905, opened 2021-03-23
Add `/api/tasks/restarttask` endpoint

### LINKED ISSUE #7348: "Restarting" a Task normally means deleting it and triggering a new one with same parameters -> UI consequences
### Observed behavior

I don't believe there's a way to simply restart a Task once it's in the CANCELED/FAILED status. So to implement the "retry" feature for failed facility syncs/imports, we actually delete the original one and start a new one with data that is tacked on to the task metadata.

Having to do things this way has a couple of consequences for the UI code and actual UX:

1. "Restarting" a task with our standard "poll for tasks" framework means 1) delete it, 2) see it disappear, 3) start a new one, 4) see an identical-looking task reappear
1. The API calls to restart a task require the same full payload, as opposed to a more minimal one with just the TaskID. This is why we do not allow restarting Facility Import tasks, because this would require us tacking on plaintext credentials onto metadata to use the same approach

### Expected behavior

If it doesn't already exist, it would be nice to have a simpler interface to restart a Task and reduce the process from 2 steps (delete + create) to just 1 (restart).

And concretely, this would mean that the Facility Import/Sync Task Manager UIs in the Facility Page and Setup Wizard wouldn't have the disappear-then-reappear behavior it does now.

### Steps to reproduce

…

### Context

…

ashmeet13:
### Summary
Currently, the TaskAPI backend does not support restarting a task. To restart as it is now - we have to cancel/clear the task and then initiate it again. This PR adds - 
1. The backend logic to support restart for a `FAILED or CANCELLED` state task
2. Endpoint to restart a task via a `POST` request containing the `task_id` to restart.

…

### Reviewer guidance
I have added unit tests for the backend logic that is added for `queue.py` to restart a task.
I was not able to write the unit test for the API endpoint. Could someone guide me here?

…

### References
Fixes #7348 

…

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/core/tasks/queue.py
+++ b/kolibri/core/tasks/queue.py
@@ -66,6 +66,22 @@ class Queue(object):
         job_id = self.storage.enqueue_job(job, self.name)
         return job_id
 
+    def restart_job(self, job_id):
+        """
+        Given a job_id, restart the job for that id. A job will only be restarted if
+        in CANCELED or FAILED state.
+
+        This will create a new job with a new job_id.
+        """
+
+        old_job = self.fetch_job(job_id)
+        if old_job.state in [State.CANCELED, State.FAILED]:
+            self.clear_job(job_id)
+            job = Job(old_job)
+            return self.enqueue(job)
+        else:
+            return None
+
     def cancel(self, job_id):
         """
         Mark a job as canceling, and let the worker pick this up to initiate
--- a/kolibri/core/tasks/queue.py
+++ b/kolibri/core/tasks/queue.py
@@ -74,7 +74,6 @@ class Queue(object):
 
         This will create a new job with a new job_id.
         """
-
         old_job = self.fetch_job(job_id)
         if old_job.state in [State.CANCELED, State.FAILED]:
             self.clear_job(job_id)
--- a/kolibri/core/tasks/queue.py
+++ b/kolibri/core/tasks/queue.py
@@ -75,6 +75,7 @@ class Queue(object):
         """
 
         old_job = self.fetch_job(job_id)
+        print(old_job.state)
         if old_job.state in [State.CANCELED, State.FAILED]:
             self.clear_job(job_id)
             job = Job(old_job)
--- a/kolibri/core/tasks/queue.py
+++ b/kolibri/core/tasks/queue.py
@@ -1,3 +1,4 @@
+from kolibri.core.tasks.exceptions import JobNotRestartable
 from kolibri.core.tasks.job import Job
 from kolibri.core.tasks.job import State
 from kolibri.core.tasks.storage import Storage
@@ -75,13 +76,14 @@ class Queue(object):
         """
 
         old_job = self.fetch_job(job_id)
-        print(old_job.state)
         if old_job.state in [State.CANCELED, State.FAILED]:
             self.clear_job(job_id)
             job = Job(old_job)
             return self.enqueue(job)
         else:
-            return None
+            raise JobNotRestartable(
+                "Cannot restart job with state={}".format(old_job.state)
+            )
 
     def cancel(self, job_id):
         """
--- a/kolibri/core/tasks/test/test_api.py
+++ b/kolibri/core/tasks/test/test_api.py
@@ -118,6 +118,16 @@ class TaskAPITestCase(BaseAPITestCase):
             for i in [4, 5, 6]:
                 assert_clearable(i, True)
 
+    def test_restart_task(self, queue_mock):
+        queue_mock.restart_job.return_value = 2
+        queue_mock.restart_job.return_value = fake_job(state=State.QUEUED)
+
+        response = self.client.post(
+            reverse("kolibri:core:task-restarttask"), {"task_id": "1"}, format="json"
+        )
+
+        self.assertEqual(response.data, {})  # TODO fix response as recieved.
+
 
 class TaskAPIPermissionsTestCase(APITestCase):
     def setUp(self):
--- a/kolibri/core/tasks/test/test_api.py
+++ b/kolibri/core/tasks/test/test_api.py
@@ -120,15 +120,23 @@ class TaskAPITestCase(BaseAPITestCase):
 
     def test_restart_task(self, queue_mock):
         queue_mock.restart_job.return_value = 2
-        queue_mock.restart_job.return_value = fake_job(state=State.QUEUED)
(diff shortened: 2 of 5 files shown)


kolibri/core/tasks/queue.py
```diff
@@ -66,6 +66,23 @@ def update_progress(progress, total_progress, stage=""):
...
+        """
+
+        old_job = self.fetch_job(job_id)
+        print(old_job.state)
+        if old_job.state in [State.CANCELED, State.FAILED]:
+            self.clear_job(job_id)
+            job = Job(old_job)
+            return self.enqueue(job)
+        else:
+            return None
```
rtibbles: Feels like we should raise some sort of error here if someone is trying to restart a job that was not cancelled or did not fail?

We can then handle that error in the API endpoint and return a `ValidationError`.

kolibri/core/tasks/test/test_job_running.py
```diff
@@ -301,3 +314,60 @@ def test_can_cancel_a_job_that_updates_progress(self, inmem_queue):
             assert time_spent < 5
         # and hopefully it's canceled by this point
         assert job.state == State.CANCELED
+
```
rtibbles, line 318: Would be good to add tests for:

`test_successful_job_cannot_restart` `test_queued_job_cannot_restart` too!

rtibbles (changes requested):
A couple of suggestions for improvement, but this is a very solid implementation already, good work!

kolibri/core/tasks/test/test_job_running.py
ashmeet13: Added these. The tests checks for the `JobNotRestartable` exception.

kolibri/core/tasks/queue.py
ashmeet13: Yup. Wasn't sure about this. Added this with a new exception `JobNotRestartable`

ashmeet13:
@rtibbles could you please guide me on how to write the tests for API? I couldn't get `Mock` to work for me.
There is a test that I have added that is failing as it is now. Could you have a look please? Here is the [permalink](https://github.com/learningequality/kolibri/pull/7905/files#diff-dc9e99b60e8975b122438010bb0d2e8c3d2fc358dcba63f043eaaf4e9d15fc69R121-R132)

kolibri/core/tasks/test/test_api.py
```diff
@@ -118,6 +118,18 @@ def assert_clearable(index, expected):
             for i in [4, 5, 6]:
                 assert_clearable(i, True)
 
+    def test_restart_task(self, queue_mock):
+        queue_mock.restart_job.return_value = 2
+        queue_mock.restart_job.return_value = fake_job(state=State.QUEUED)
```
rtibbles: Pass in `id=2` here, and get rid of the line above. That will force your fake job to have an id of 2, so your assertion at the end of the test should now be true.

rtibbles (approved):
Looking good - left a comment on how to update the API test, once that's fixed, I think this is good to merge!

kolibri/core/tasks/test/test_api.py
ashmeet13: I tried that. Still have the error.
There are two queue functions that I need to `patch` here - 
1. `restart_job`
2. `fetch_job`

Here is how they are used in the implementation `restarttask` endpoint in the api - 
```
            try:
                task_id = _queue.restart_job(request.data["task_id"])
                resp = _job_to_response(_queue.fetch_job(task_id))
                break
```

`restart_job` should return an integer with value 2
`fetch_job` should return a fake job with State=Queue, id=2

Error is at the stage where `restart_job` raises `JobNotFound`. Hence it isn't being patched.

rtibbles: Oh, sorry, I understand what you are trying to do now.

In that case, I think the only error here is that the second line should be:
```
queue_mock.fetch_job.return_value = fake_job(state=State.QUEUED, id=2)
```

Because at the moment you are only mocking `restart_job` and not `fetch_job` too.

ashmeet13: Oh my bad for that. But it still fails even after updating it to - 
```
queue_mock.restart_job.return_value = 2
queue_mock.fetch_job.return_value = fake_job(state=State.QUEUED, id=2)
```

rtibbles: Ah yes, you are right - I just checked this out locally. The issue here is that the `@patch("kolibri.core.tasks.api.queue")` decorator on the TestCase does not seem to be mocking anything - the other tests in this test case all pass, as they all pass in the case that a job is not found.

I think the issue is due to the fact that the TaskViewset `queues` property is set to a list of `[queue, priority_queue]` because this is defined at module parse time, the mocking fails to intercept this, because at the time that `[queue, priority_queue]` is defined the references are to the original, non-mocked objects.

The easiest way to make this work would probably be to turn the `queues` property of the TaskViewset into:
```
@property
def queues(self):
    return [queue, priority_queue]
```

This change along with passing `job_id=2` instead of `id=2` (my mistake), causes the tests to now pass.

ashmeet13: I see. This was something new to me. Thank you for the help @rtibbles!
The test seems to be fixed.

rtibbles: Yeah, understanding when and how mocks actually apply can be a little bit difficult to wrap one's head around sometimes!

jonboiser:
@ashmeet13 This is really great, thanks! Here it is in action when restarting a channel import task. I did find a potential bug, where I could not trigger a restart a second time, so I'll file a bug for that. Let me know if you have any ideas on why this might be happening. You might notice that when I retry the request after it was restarted, I don't get validation errors.

In any case, we will soon try to integrate this into the app and add the ability to restart more types of Tasks!

CleanShot 2021-03-26 at 14 22 24

jonboiser:
Ah, I think I see what's going on. The new Task actually has a new ID. I wonder if there's a way to let the new task keep the ID of the old one.

rtibbles:
I suggested implementing it as a task with a new id, as a simpler way to copy over the arguments, but without having to replicate the previous state metadata.

I assume if the frontend just continues to read the current task id, this shouldn't be an issue though right? Just because you're POSTing and not updating the body.

PR #7924, opened 2021-03-25
Update zip_content options to better reflect deployment needs.

rtibbles:
### Summary
* Post merge review by @jamalex of #7254 showed an issue with the way that the zip content options were being used
* Specifically ZIP_CONTENT_PORT as a setting should be reserved for the port that Kolibri serves its zipcontent server on
* This PR then changes the other option to ZIP_CONTENT_ORIGIN, which can either be the port that a proxied alternate origin server is served on, or a full origin value
* This then updates the interface code between backend and frontend to either pass the port, the proxied port, or the full origin, depending on how it is configured.

### Reviewer guidance
Do the port options make sense?
Does the documentation make sense?
Does HTML5 content still serve? (I have locally confirmed that it does)

### References
Follow up from #7254 

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/core/content/test/test_zipcontent.py
+++ b/kolibri/core/content/test/test_zipcontent.py
@@ -99,9 +99,16 @@ class ZipContentTestCase(TestCase):
         return generate_zip_content_response(self.environ)
 
     def test_zip_file_url_reversal(self):
+        from kolibri.utils.conf import OPTIONS
+
+        path_prefix = OPTIONS["Deployment"]["ZIP_CONTENT_URL_PATH_PREFIX"]
+
+        if path_prefix != "/":
+            path_prefix = "/" + path_prefix
         file = LocalFile(id=self.hash, extension=self.extension, available=True)
         self.assertEqual(
-            file.get_storage_url(), "/zipcontent/{}/".format(self.filename)
+            file.get_storage_url(),
+            "{}zipcontent/{}/".format(path_prefix, self.filename),
         )
 
     def test_non_zip_file_url_reversal(self):
@@ -301,6 +308,11 @@ class ZipContentTestCase(TestCase):
         self.assertEqual(response.status_code, 405)
 
 
+@override_option("Deployment", "ZIP_CONTENT_URL_PATH_PREFIX", "prefix_test/")
+class UrlPrefixZipContentTestCase(ZipContentTestCase):
+    pass
+
+
 class HashiViewTestCase(TestCase):
     def setUp(self):
         self.environ = {}
@@ -359,3 +371,8 @@ class HashiViewTestCase(TestCase):
     def test_delete_not_allowed(self):
         response = self._get_hashi(REQUEST_METHOD="DELETE")
         self.assertEqual(response.status_code, 405)
+
+
+@override_option("Deployment", "ZIP_CONTENT_URL_PATH_PREFIX", "prefix_test/")
+class UrlPrefixHashiViewTestCase(HashiViewTestCase):
+    pass
--- a/kolibri/core/content/utils/paths.py
+++ b/kolibri/core/content/utils/paths.py
@@ -253,8 +253,16 @@ HASHI = "hashi/"
 ZIPCONTENT = "zipcontent/"
 
 
+def zip_content_path_prefix():
+    path_prefix = conf.OPTIONS["Deployment"]["ZIP_CONTENT_URL_PATH_PREFIX"]
+
+    if path_prefix != "/":
+        path_prefix = "/" + path_prefix
+    return path_prefix
+
+
 def get_zip_content_base_path():
-    return "{}{}".format(conf.OPTIONS["Deployment"]["URL_PATH_PREFIX"], ZIPCONTENT)
+    return "{}{}".format(zip_content_path_prefix(), ZIPCONTENT)
 
 
 HASHI_FILENAME = None
@@ -277,13 +285,15 @@ def get_hashi_html_filename():
 
 
 def get_hashi_base_path():
-    return "{}{}".format(conf.OPTIONS["Deployment"]["URL_PATH_PREFIX"], HASHI)
+    return "{}{}".format(zip_content_path_prefix(), HASHI)
 
 
 def get_hashi_path():
-    return "{}{}{}".format(
-        conf.OPTIONS["Deployment"]["URL_PATH_PREFIX"], HASHI, get_hashi_html_filename()
-    )
+    return "{}{}{}".format(zip_content_path_prefix(), HASHI, get_hashi_html_filename())
+
+
+def zip_content_static_root():
+    return urljoin(zip_content_path_prefix(), "static/")
 
 
 def get_content_storage_file_url(filename):
--- a/kolibri/core/content/zip_wsgi.py
+++ b/kolibri/core/content/zip_wsgi.py
@@ -71,7 +71,7 @@ def django_response_to_wsgi(response, environ, start_response):
 
 template_engine = Engine(
     dirs=[os.path.join(os.path.dirname(__file__), "./templates/content")],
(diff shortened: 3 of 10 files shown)


jamalex:
Still reviewing, and will test, but some more general notes in the interim:

Not specific to this PR, but more broadly: we currently only half-heartedly support the `URL_PATH_PREFIX` option across Kolibri. To avoid further perpetuating that lack of support here, I would recommend changing this to "BASE_URL" instead of "ORIGIN".

Not something that was added in this PR, but we should change the port to something more obscure than "8888" -- which I know is the default port for Resilio Sync's web UI, amongst other things... like "8293" or something.

rtibbles:
> "BASE_URL" instead of "ORIGIN"

Yes, that would be more accurate - especially as I did allow it to be part of the returned path in the validation.

Any thoughts on it being allowed to short hand as just an alternate port? Felt useful for proxies.

jamalex:
> Any thoughts on it being allowed to short hand as just an alternate port? Felt useful for proxies.

I understood the motivation for it, given a) having two settings that are potentially competing, or at least need clear precedence defined, is messy, and b) we can't/shouldn't just require that it be a fully defined URL in all cases, especially with proxies where the host might change. So, while it feels a bit awkward, I didn't have a better suggestion.

jamalex:
Another side thing that's not specific to this PR: The addition of these new port settings makes "`HTTP_PORT`" feel a bit... underspecified -- since _all_ of these are HTTP ports! But that poor option has been renamed so many times I don't know that it could handle another.

rtibbles:
Upon trying to implement this - it feels like it might actually be better to give a parallel `ZIP_CONTENT_URL_PATH_PREFIX` to the regular `URL_PATH_PREFIX`, and keep `ZIP_CONTENT_ORIGIN` as purely origin focused. Otherwise it would be tricky to expose say just a custom port (while not changing the domain/IP) and have a path prefix on the URL.

kolibri/core/content/test/test_zipcontent.py
```diff
@@ -99,9 +99,16 @@ def _get_file(self, file_name, base_url=None, **kwargs):
         return generate_zip_content_response(self.environ)
 
     def test_zip_file_url_reversal(self):
+        from kolibri.utils.conf import OPTIONS
+
+        path_prefix = OPTIONS["Deployment"]["ZIP_CONTENT_URL_PATH_PREFIX"]
+
+        if path_prefix != "/":
+            path_prefix = "/" + path_prefix
```
jamalex, line 107: A pattern I sometimes use that would be even more tolerant/robust here is: `"/" + path_prefix.strip("/") + "/"`

kolibri/core/content/test/test_zipcontent.py
```diff
@@ -99,9 +99,16 @@ def _get_file(self, file_name, base_url=None, **kwargs):
         return generate_zip_content_response(self.environ)
 
     def test_zip_file_url_reversal(self):
+        from kolibri.utils.conf import OPTIONS
+
+        path_prefix = OPTIONS["Deployment"]["ZIP_CONTENT_URL_PATH_PREFIX"]
```
jamalex, line 104: Since this option isn't being overridden using [the decorator](https://github.com/learningequality/kolibri/pull/7924/files#diff-38ed1052b92c27b26b69d8a9d87f366c31634293db3f9b0452293f6748a07d8dR36), this will be directly pulling from the local options.ini file (or envvars), which makes the test less predictable. I might be missing something, though.

kolibri/core/content/test/test_zipcontent.py
```diff
@@ -301,6 +308,11 @@ def test_delete_not_allowed(self):
         self.assertEqual(response.status_code, 405)
 
 
+@override_option("Deployment", "ZIP_CONTENT_URL_PATH_PREFIX", "prefix_test/")
+class UrlPrefixZipContentTestCase(ZipContentTestCase):
+    pass
```
jamalex, line 313: Is this test just testing that the option exists?

kolibri/core/content/test/test_zipcontent.py
```diff
@@ -359,3 +371,8 @@ def test_patch_not_allowed(self):
     def test_delete_not_allowed(self):
         response = self._get_hashi(REQUEST_METHOD="DELETE")
         self.assertEqual(response.status_code, 405)
+
+
+@override_option("Deployment", "ZIP_CONTENT_URL_PATH_PREFIX", "prefix_test/")
+class UrlPrefixHashiViewTestCase(HashiViewTestCase):
+    pass
```
jamalex, line 378: I guess not, since otherwise this would just be testing the same thing. Stubs?

kolibri/core/content/utils/paths.py
```diff
@@ -253,8 +253,16 @@ def get_file_checksums_url(channel_id, baseurl, version="1"):
 ZIPCONTENT = "zipcontent/"
 
 
+def zip_content_path_prefix():
+    path_prefix = conf.OPTIONS["Deployment"]["ZIP_CONTENT_URL_PATH_PREFIX"]
+
+    if path_prefix != "/":
+        path_prefix = "/" + path_prefix
```
jamalex, line 260: See note above re: possibly more robust sanitization.

kolibri/core/content/zip_wsgi.py
```diff
@@ -419,4 +419,6 @@ def get_application():
         get_zip_content_base_path(): zip_content_view,
     }
 
+    print(path_map)
+
```
jamalex: To remove?

kolibri/utils/options.py
```diff
@@ -45,9 +45,25 @@
 HTTP_PORT
 RUN_MODE
 URL_PATH_PREFIX
+    Serve Kolibri from a subpath under the main domain. Used when serving multiple applications from
+    the same origin. This option is not heavily tested, but is provided for user convenience.
 LANGUAGES
-ZIP_CONTENT_HOST
+ZIP_CONTENT_URL_PATH_PREFIX
+    The zip content equivalent of URL_PATH_PREFIX - allows all zip content URLs to be prefixed with
+    a fixed path.
```
jamalex: It wasn't clear to me from reading this whether it was affecting how it was being served, or how the frontend was pointing to it, or both. From reading the code, I think it's maybe both? Which is probably OK, since there's not much of a use case for having a different prefix on the other server than locally (seeing as if the origin is changed, presumably the local zip server wouldn't even be used). But it might be good to be clearer about how it's used.

jamalex (commented):
Looks good, just a few suggestions so far. I'm prepping to test it locally, as well.

kolibri/core/content/test/test_zipcontent.py
rtibbles: The `.strip` already happens during option validation.

kolibri/core/content/test/test_zipcontent.py
rtibbles: Tests run in an encapsulated `KOLIBRI_HOME` so this should be safe, it won't be pulling from any previously defined local state.

kolibri/core/content/test/test_zipcontent.py
rtibbles: No, it's rerunning the whole test case with the option set.

kolibri/core/content/zip_wsgi.py
rtibbles: __Debug__

kolibri/utils/options.py
rtibbles: It's both, yes - although in the case where you've set the origin option too, the fact that it is now serving from the other endpoint is a moot point.

rtibbles:
Updated based on latest feedback, and rebased to latest develop.

jamalex (approved):
Perfect

PR #7928, opened 2021-03-26
Updates to memberships, roles, and permissions handling and validation

### LINKED ISSUE #6993: Unexisting role is being used and created in create_dummy_facility_data test helper
### Observed behavior
There is an useful helper called `create_dummy_facility_data`, https://github.com/learningequality/kolibri/blob/develop/kolibri/core/auth/test/helpers.py#L32 that can be used in tests that need a facility with classes and users created. The test creates users with all the roles Kolibri has and classes with them enrolled or assigned.

However, this function is using a non-existing role: `classadmin` in several parts of the code, creating users and assigning them to the classes:
https://github.com/learningequality/kolibri/blob/develop/kolibri/core/auth/test/helpers.py#L75-L80
https://github.com/learningequality/kolibri/blob/develop/kolibri/core/auth/test/helpers.py#L123

### Expected behavior
To avoid confussion the code using or creating this kind of users should be removed

### Context
all versions of kolibri
It only affects to developers, not end users.

### LINKED ISSUE #7299: issues creating lessons and quizzes in very large facilities
### Observed behavior

Reported by a user with many thousands of users in their facility:

> We have just come across an issue when working in Kolibri as an admin. Creating a lesson or quiz doesn't work. When trying to create a lesson, we can fill out the details, but then when we hit save we get stuck on the page and disconnected from the server. Same behavior when creating a quiz, it gets stuck on the creation page.

image

image

Users page takes 13-15 seconds to load:

image

### Expected behavior

Operations should be fast even for thousands of users in a facility

### User-facing consequences

difficulty scaling

### Steps to reproduce

Contact @rtibbles or @indirectlylit for details from the user.

@rtibbles believes that this is an issue with the `HierarchicalRelationsFilter`: https://learningequality.slack.com/archives/CB37UM23A/p1594754290117000

### Context

0.13.x

### LINKED ISSUE #7442: Queries generated by `KolibriAuthPermissionsFilter` can't be used as subqueries
This is a followup on #7431.

### Observed behavior
When logged into Kolibri as a superadmin of an imported facility, a certain API request was failing.  According to @rtibbles, this was because it tried to filter on a queryset with a query generated by `KolibriAuthPermissionsFilter`.  The workaround was to [avoid relying on using that query as a subquery](https://github.com/learningequality/kolibri/pull/7431/commits/00a6256437334799a610c30aff1048df3ff84eef), and instead pass in a list of ids, which just so happened to be conveniently available.

### Expected behavior
`KolibriAuthPermissionsFilter` should generate querysets that work with ordinary Django queryset tools.

### User-facing consequences
This could send developers on wild goose chases, causing them to behave superstitiously or get disheartened because they're not quite sure what's going on!

### Steps to reproduce
1. Checkout https://github.com/rtibbles/kolibri/commit/edab8e390f96774112e898787df588bddc3f2653
2. Make sure the correct python/javascript libraries installed in your virtual environment, since this may be an old version of Kolibri by the time you read this
3. Start Kolibri with the home folder attached here: [kolibri-home-with-missing-data.tar.gz](https://github.com/learningequality/kolibri/files/5025442/kolibri-home-with-missing-data.tar.gz)
4. Log into "REMOTE FACILITY" with the username "micah" and the password "micah"
5. Try using this endpoint: http://localhost:8000/api/lessons/lesson/?collection=2fde259776840852390f963a452fb06c
- You should get a 500 error here
6. You can put a breakpoint inside of the `if` statement [here](https://github.com/learningequality/kolibri/blob/50f6510f3c7e1cc30d68ce1af58192ecd4c1608a/kolibri/core/lessons/viewsets.py#L70) and do `print(queryset.query)`
7. Now logout and log back in to "home facility" with username "micah" + password "micah"
8. Try the same endpoint and inspect the `queryset.query` again.  You should see that a different query this time, and the request should work fine.

### LINKED ISSUE #7842: Remove HierarchyRelationsFilter and enforce explicit permissions model
### Observed behavior
Our current role permissions and membership model allows for nested membership, and role permissions.

In practice, we require explicit membership of any collection. Each user is assigned to a specific Facility, and we only allow learners to be in LearnerGroups or AdHocGroups via the frontend interface if they are already a member of the relevant Classroom.

Through our user interface we only allow users to be Admins or Coaches at the Facility level, and at the Classroom level we only allow Coaches.

We do not allow assignment of role permissions to any other collections through the interface.

In order to support the more general permissions model, we use the HiearchyRelationsFilter to determine many permissions and membership checks. However, we have seen multiple cases where the HierarchyRelationsFilter causes significant performance issues in Facilities with a modestly (100+) sized collection tree (c.f. #6721, #7823).

### Expected behavior
Permissions and membership checks should only rely on the explicitly used membership and role assignment that we use.

We should enforce this pattern, rather than simply relying on it to be put in place by convention through the UI, so we should add validation to the serializers and the models where relevant.

We should remove HierarchyRelationsFilter in the process.

### User-facing consequences
Performance issues

rtibbles:
### Summary
* Completely removes use of HierarchyRelationsFilter in favour of filtering based on fixed membership and role structure reflective of current use
* Updates `readable_by_user_filter` to take only a user argument, and to return Django `Q` objects to avoid use of queryset union and intersection, which prevents the querysets being further used in subqueries
* Enforces current collection hierarchy
* Enforces current role structure - admins, coaches, assignable coaches at the facility level, coaches only at the Classroom level, no roles for other collections
* Enforces current membership structure - only classroom members can be assigned as members of child collections
* Adds and updates tests
* Updates documentation of the auth module

### Reviewer guidance
Does everything still work as expected in facility management?
Do the documentation updates fully capture all the changes?
Are any of the enforcement measures (like removing classroom coach roles when classroom assignable coach is removed) too much?
Should we always add the `assignable coach` role at the facility level if it doesn't exist for the user when they are assigned as a class coach?
Does this have the anticipated improvements for performance during permission checking/filtering?

### References
Fixes #6993
Fixes #7299
Fixes #7442
Fixes #7842

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated


docs/backend_architecture/uap/concepts_and_definitions.rst
```diff
@@ -14,13 +14,11 @@ all the data associated with a particular Facility are referred to as a
 Users
 -----
 
-There are two kinds of users: ``FacilityUser`` and ``DeviceOwner``. A
+The user model is ``FacilityUser``. A
```
indirectlylit:
```suggestion
Kolibri's users are instances of the ``FacilityUser`` model, which derives from Django's ``AbstractBaseUser``. A user
```

docs/backend_architecture/uap/concepts_and_definitions.rst
```diff
@@ -14,13 +14,11 @@ all the data associated with a particular Facility are referred to as a
...
 ``FacilityUser`` is associated with a particular ``Facility``, and the user's
-account and data may be synchronized across multiple devices. A
-``DeviceOwner`` account is not associated with a particular ``Facility``, but
-is specific to one device, and is never synchronized across multiple devices.
-A ``DeviceOwner`` is like a superuser, and has permissions to modify any data
-on her own device, whereas a ``FacilityUser`` only has permissions for some
+account and data may be synchronized across multiple devices.
+A ``FacilityUser`` may be made into a superuser, with permissions to modify any data
+on her own device, however, normally a ``FacilityUser`` only has permissions for some
```
indirectlylit:
```suggestion
on her own device. However, normally a ``FacilityUser`` only has permissions for some
```

docs/backend_architecture/uap/concepts_and_definitions.rst
```diff
@@ -44,13 +42,12 @@ Class A contains two LearnerGroups, Group Q and Group R.
...
 
-A ``FacilityUser`` (but not a ``DeviceOwner``) can be marked as a member of a
-``Collection`` through a ``Membership`` object. Being a member of a Collection
-also means being a member of all the Collections above that Collection in the
-hierarchy. Thus, in the illustration below, Alice is directly associated with
-Group Q through a ``Membership`` object, which makes her a member of Group Q.
-As Group Q is contained within Class A, which is contained within Facility X,
-she is also implicitly a member of both those collections.
+A ``FacilityUser`` can be marked as a member of a ``Collection`` through a ``Membership`` object.
+Being a member of a Collection, requires first being a member of
```
indirectlylit:
```suggestion
Being a member of a Collection requires first being a member of
```

indirectlylit: the capitalization of some words seems a little odd. Maybe either format words as code if referring to the class (`Collection`) or lower-case them if referring to the concept (collection)

docs/backend_architecture/uap/concepts_and_definitions.rst
```diff
@@ -65,8 +62,9 @@ Roles
 Another way in which a ``FacilityUser`` can be associated with a particular
 ``Collection`` is through a ``Role`` object, which grants the user a role with
 respect to the ``Collection`` and all the collections below it. A ``Role``
-object also stores the "kind" of the role (currently, one of "admin" or
-"coach"), which affects what permissions the user gains through the ``Role``.
+object also stores the "kind" of the role (currently, one of "admin",
```
indirectlylit:
```suggestion
object stores the "kind" of the role (currently, one of "admin",
```

docs/backend_architecture/uap/concepts_and_definitions.rst
```diff
@@ -14,13 +14,11 @@ all the data associated with a particular Facility are referred to as a
 Users
 -----
 
-There are two kinds of users: ``FacilityUser`` and ``DeviceOwner``. A
+The user model is ``FacilityUser``. A
 ``FacilityUser`` is associated with a particular ``Facility``, and the user's
-account and data may be synchronized across multiple devices. A
-``DeviceOwner`` account is not associated with a particular ``Facility``, but
-is specific to one device, and is never synchronized across multiple devices.
-A ``DeviceOwner`` is like a superuser, and has permissions to modify any data
```
indirectlylit, line 22: Separate from this PR, but flagging that `DeviceOwner` is still referenced in other places:

https://kolibri-dev.readthedocs.io/en/develop/backend_architecture/uap/models.html#models

docs/backend_architecture/uap/concepts_and_definitions.rst
```diff
@@ -78,9 +76,13 @@ marked with kind "coach", which we can informally read as "Bob is a coach for
 Class A". We consider user roles to be "downward-transitive" (meaning if you
 have a role for a collection, you also have that role for descendents of that
 collection). Thus, in our example, we can say that "Bob is also a coach for
-Group Q". Furthermore, as Alice is a member of Group Q, we can say that "Bob
+Group Q". Furthermore, as Alice is a member of Class A, we can say that "Bob
 is a coach for Alice".
 
+To simplify the role structure, a user can be assigned as an admin, coach, or assignable coach
+at the Facility level, but only as a coach at the Classroom level. Users cannot
+be assigned roles at the AdHocGroup/LearnerGroup level.
```
indirectlylit:
```suggestion
A user can be assigned certain roles for different collection types:

* ``Facility`` collections: admin, coach, or assignable coach roles
* ``Classroom`` collections: coach roles 
* ``LearnerGroup`` and ``AdHocGroup`` collections: no roles
```

Is it assignable coach or regular coach for `Classroom`?

docs/backend_architecture/uap/implementation.rst
```diff
@@ -58,112 +58,14 @@ Efficient hierarchy calculations
 In order to make decisions about whether a user has a certain permission for
 an object, we need an efficient way to retrieve the set of roles the user has
 in relation to that object. This involves traversing the Role table,
-Collection hierarchy, and possibly the Membership table, but we can delegate
-most of the work to the database engine (and leverage efficient hierarchy
-lookups afforded by MPTT). The following algorithms and explanations will
-refer to the naming in the following diagram:
+Collection hierarchy, and possibly the Membership table. Because we require
+explicit representation of membership at any level in the hierarchy, we can rely
```
indirectlylit:
```suggestion
explicit representation of membership at each level in the hierarchy, we can rely
```

docs/backend_architecture/uap/implementation.rst
```diff
@@ -210,8 +111,8 @@ the following overridable methods:
   - ``user_can_update_object``
   - ``user_can_delete_object``
 - The queryset-filtering ``readable_by_user_filter`` method, which takes in a
-  queryset and returns a queryset filtered down to just objects that should be
-  readable by the user.
+  user and returns a Django Q object that can be used to filter to just
```
indirectlylit:
```suggestion
  user and returns a Django ``Q`` object that can be used to filter to just
```

indirectlylit (commented):
Added some notes on the documentation.

Seems like a good move to me to align database constraints more closely with our use in practice. Presumably we can relax the constraints later if the need arises.

I do think it could be useful to get some anonymized sample data from a range of implementations to test migrations in case somehow data exists in the wild that violates constraints currently imposed by the UI.

docs/backend_architecture/uap/concepts_and_definitions.rst
rtibbles: Classroom is coach only, assignable coach is only at the facility level.

rtibbles:
@indirectlylit that seems difficult to achieve in practice. If there is a concern that this may be the case, moving the errors out of the model save methods and just into the API endpoints/serializers would let us continue to have enforcement, but not break any syncing of existing data.

indirectlylit:
> If there is a concern that this may be the case

Yeah, I don't have a good sense of the risk apart from knowing that sometimes we've had unexpected database states in e.g. Studio. Alternatively as long as migrations fail early and loudly so people can report issues, that seems acceptable as well.

rtibbles:
I think it may be worth adding either a data migration or an upgrade step to make sure that the database is in a compatible state.

jredrejo (approved):
Tested it deeply and everything works fine. I really love the documentation updates! I have also checked it fixes the mentioned issues. Thanks!

PR #7942, opened 2021-04-01
Make `window.kolibri` object available to Hashi iframe context

marcellamaki:
This is very much a draft/WIP PR that is working to establish the foundation for a kolibri object and the core MVP functions. 

Some functions are fully working, some are WIP with elements of pseudo code, as I need a more well-thought-out strategy for doing manual QA in some places. The new data-fetching code in `mainClient.js` should most likely be relocated to another file - maybe a `handlers.js`, or within the custom channels vue file which is not yet finished (TBD). 

I am also not sure if `__navigateTo()` is doing _enough_, and if it's not, what exactly needs to be adjusted/added.

### DIFF
--- a/packages/hashi/src/mainClient.js
+++ b/packages/hashi/src/mainClient.js
@@ -3,6 +3,7 @@ import LocalStorage from './localStorage';
 import Cookie from './cookie';
 import SCORM from './SCORM';
 import { events, nameSpace } from './hashiBase';
+import Kolibri from './kolibri';
 
 /*
  * This is the main entry point for interacting with the Hashi library.
@@ -22,6 +23,7 @@ export default class MainClient {
       cookie: new Cookie(this.mediator),
       SCORM: new SCORM(this.mediator),
     };
+    this.kolibri = new Kolibri(this.mediator);
     this.now = now;
     this.ready = false;
     this.contentNamespace = null;
--- a/packages/hashi/src/mainClient.js
+++ b/packages/hashi/src/mainClient.js
@@ -31,14 +31,6 @@ export default class MainClient {
     this.contentNamespace = null;
     this.startUrl = null;
     this.__setData = this.__setData.bind(this);
-
-    // this.mediator.registerMessageHandler({
-    //   nameSpace,
-    //   event: events.GETCONTENT,
-    //   callback: () => {
-    //     // this.mediator.sendMessage({ nameSpace, event: events.REQUESTINGDATA, data: true });
-    //   },
-    // });
   }
   initialize(contentState, userData, startUrl, contentNamespace) {
     /*
@@ -135,20 +127,25 @@ export default class MainClient {
 
     // if filtering by optional params
     if (message.options) {
-      let hash = window.location.hash.split('/');
-      let id = hash[hash.length - 1];
-      ContentNodeResource.fetchCollection({ id }).then(contentNodes => {
-        if (contentNodes) {
-          message.status = 'success';
-        } else {
-          message.status = 'failure';
-        }
+      let getParams = {};
+      let options = message.options;
+      if (options.parent && options.parent == 'self') {
+        // need to fetch this value when this
+        // function is move to a location that has access
+        // to `content`
+        // getParams.parent = rootNode;
+      } else if (options.parent) {
+        getParams.parent = options.parent;
+      }
+      options.ids ? (getParams.ids = options.ids) : null;
+      options.page ? (getParams.page = options.page) : null;
+      options.pageSize ? (getParams.ids = options.pageSize) : null;
+      ContentNodeResource.fetchCollection({ getParams }).then(contentNodes => {
+        contentNodes ? (message.status = 'success') : (message.status = 'failure');
         let response = {};
         response.page = message.options.page ? message.options.page : 1;
         response.pageSize = message.options.pageSize ? message.options.pageSize : 50;
-        console.log(response);
-        let results = contentNodes.filter(node => node.id == id);
-        response.results = results;
+        response.results = contentNodes;
         message.data = response;
         message.type = 'response';
         this.mediator.sendMessage({
@@ -179,7 +176,6 @@ export default class MainClient {
   }
 
   __navigateTo(message) {
-    console.log('navigating');
     let id = message.nodeId;
     ContentNodeResource.fetchModel({ id }).then(contentNode => {
       let routeBase, context;
@@ -202,7 +198,14 @@ export default class MainClient {
   }
 
   __getOrUpdateContext(message) {
-    console.log(message);
+    // to update context with the incoming context
+    if (message.context) {
+      router.push({ query: { context: message.context } }).catch(() => {});
+    } else {
+      // just return the existing query
+      const urlParams = new URLSearchParams(window.location.search);
+      return urlParams.has('context') ? urlParams.get('context') : null;
+    }
   }
 
   get data() {
--- a/packages/hashi/src/mainClient.js
+++ b/packages/hashi/src/mainClient.js
@@ -73,24 +73,15 @@ export default class MainClient {
     this.mediator.sendMessage({ nameSpace, event: events.READYCHECK, data: true });
 
     this.on(this.events.DATAREQUESTED, message => {
-      let id = message.id;
-      let status;
-      ContentNodeResource.fetchModel({ id }).then(contentNode => {
-        if (contentNode) {
-          status = 'success';
-        } else {
-          status = 'failure';
-        }
-        message.status = status;
-        message.data = contentNode;
-        message.type = 'response';
-        this.mediator.sendMessage({
-          nameSpace,
-          event: events.DATARETURNED,
-          data: message,
-        });
-        router.push({ query: { context: 'newContext' } }).catch(() => {});
-      });
+      this.__fetchContentData(message);
+    });
+
+    this.on(this.events.NAVIGATETO, message => {
+      this.__navigateTo(message);
+    });
+
+    this.on(this.events.CONTEXT, message => {
+      this.__getOrUpdateContext(message);
     });
   }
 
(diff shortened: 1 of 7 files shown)


packages/hashi/src/mainClient.js
```diff
@@ -104,6 +120,94 @@ export default class MainClient {
...
+      let options = message.options;
+      if (options.parent && options.parent == 'self') {
+        // need to fetch this value when this
+        // function is move to a location that has access
+        // to `content`
+        // getParams.parent = rootNode;
+      } else if (options.parent) {
+        getParams.parent = options.parent;
+      }
+      options.ids ? (getParams.ids = options.ids) : null;
```
rtibbles: It's purely personal preference, but I find this in-line ternary operator to conditionalize assignment a bit hard to read.

You could also achieve the effect you want here by using the lodash helper function `pick` https://lodash.com/docs/4.17.15#pick

```
const getParams = pick(options, ['ids' ,'page', 'pageSize', 'parent']);
if (getParams.parent === 'self') {
...
}
```

Pick will only assign defined properties.

packages/hashi/src/mainClient.js
```diff
@@ -104,6 +120,94 @@ export default class MainClient {
...
     });
   }
+
+  __fetchContentData(message) {
+    // based on the incoming information, get data
+    // from Kolibri to pass back to the iframe
+
+    // if filtering by optional params
+    if (message.options) {
+      let getParams = {};
```
rtibbles: Can use `const` for both `getParams` and `options` here as you are never reassigning their values. You've assigned it to an object, and even though you are changing the attributes of that object, you are not changing the referent of the variable.

packages/hashi/src/mainClient.js
```diff
@@ -1,8 +1,11 @@
+import { ContentNodeResource } from 'kolibri.resources';
```
rtibbles: Flagging that we will want to handle this via dependency injection in the long run - as we will have to use Hashi for previewing HTML5 apps accurately in Studio, so this will break.

packages/hashi/src/mainClient.js
```diff
@@ -104,6 +120,94 @@ export default class MainClient {
...
+        message.type = 'response';
+        this.mediator.sendMessage({
+          nameSpace,
+          event: events.DATARETURNED,
+          data: message,
+        });
+      });
+    }
+    // or, if getting by a specific id
+    else if (message.id) {
```
rtibbles: As there are no shared code paths between these two parts of the function - might be good to just break them up into two functions, you can then add the conditionalization when you setup the listener above:
```
    this.on(this.events.DATAREQUESTED, message => {
     if (message.dataType === 'Collection') {
      this.__fetchContentCollection(message);
    else if (message.dataType === 'Model')
      this.__fetchContentModel(message);
    }
    });
```

Also suggesting here to add an explicit data type - just to be explicit about what is requested.

packages/hashi/src/mainClient.js
```diff
@@ -104,6 +120,94 @@ export default class MainClient {
...
+          event: events.DATARETURNED,
+          data: message,
+        });
+      });
+    }
+  }
+
+  __navigateTo(message) {
+    let id = message.nodeId;
+    ContentNodeResource.fetchModel({ id }).then(contentNode => {
```
rtibbles: I think we should be encapsulating more of this logic in the dependency injection - but this is on the right track.

packages/hashi/src/mainClient.js
```diff
@@ -104,6 +120,94 @@ export default class MainClient {
...
+      let routeBase, context;
+      if (contentNode && contentNode.kind === 'topic') {
+        routeBase = '/topics/t';
+      } else if (contentNode) {
```
rtibbles: The conditionality is stronger than you have below - I have probably not explained this properly in the specs.

On navigateTo we should
* If TOPIC - always navigate to the TOPIC.
* If RESOURCE - if currently in a custom channel context launch the render overlay
* If RESOURCE - if not in a custom channel context - navigate to RESOURCE

packages/hashi/src/mainClient.js
```diff
@@ -104,6 +120,94 @@ export default class MainClient {
...
+        routeBase = '/topics/c';
+      }
+      if (!message.context) {
+        // if there is custom context, don't re-route
+        const path = `${routeBase}/${id}`;
+        router.push({ path: path, query: { context: context } }).catch(() => {});
```
rtibbles: To save implementation work by all of the custom channels apps - we can probably update the `node_id` property of the context ourselves here, in the case when we are showing the render overlay over the custom channel navigation.

packages/hashi/src/mainClient.js
```diff
@@ -104,6 +120,94 @@ export default class MainClient {
...
+      this.mediator.sendMessage({
+        nameSpace,
+        event: events.DATARETURNED,
+        data: message,
+      });
+    });
+  }
+
+  __getOrUpdateContext(message) {
+    // to update context with the incoming context
```
rtibbles: Note we need to do the URI encoding and decoding of the context here.

packages/hashi/src/mainClient.js
```diff
@@ -1,8 +1,11 @@
+import { ContentNodeResource } from 'kolibri.resources';
+import router from 'kolibri.coreVue.router';
```
rtibbles: The other general thing is that aside from the dependency injection, the `mainClient` doesn't really need to know or care about most of the stuff we are doing here - so we can actually put most of this code in the `kolibri.js` file - because it is being instantiated in the iframeClient and the mainClient, we can keep the code together for easy cross reference.

packages/hashi/src/mediator.js
```diff
@@ -50,6 +52,37 @@ class Mediator {
     this.remote.postMessage(message, '*');
   }
 
+  // a function to manage messages for kolibri.js,
+  // when most messages require a response, to minimize redundancy
+  sendMessageAwaitReply({ event, data, nameSpace }) {
```
rtibbles, line 57: This looks very familiar....

packages/hashi/src/mediator.js
```diff
@@ -50,6 +52,37 @@ class Mediator {
...
+    return new Promise((resolve, reject) => {
+      const msgId = uuidv4();
+      function handler(message) {
+        if (message.message_id === msgId && message.type === 'response') {
```
rtibbles, line 61: Once we've received the corresponding message, would be good to unregister the message handler for cleanup.

packages/hashi/src/mainClient.js
```diff
@@ -59,6 +63,18 @@ export default class MainClient {
       });
     });
     this.mediator.sendMessage({ nameSpace, event: events.READYCHECK, data: true });
+
+    this.on(this.events.DATAREQUESTED, message => {
+      this.__fetchContentData(message);
```
rtibbles: I am surprised you have not needed to call `.bind` on any of the handlers you have here.

rtibbles (commented):
All the basics seem to be in place. One thing I would strongly urge you to consider sooner rather than later is getting some automated tests in place - the existing hashi tests should give some guidance on how to best go about that.

packages/hashi/src/mediator.js
```diff
@@ -50,6 +52,38 @@ class Mediator {
...
+          if (message.status === 'success') {
+            console.log(message.data);
+            return resolve(message.data);
+          } else if (message.status === 'failure' && message.err) {
+            return reject(message.err);
+          }
+          // Otherwise something unspecified happened
+          return reject();
+        }
+        this.removeMessageHandler({
```
rtibbles: This is only removing the message handler in the case when the message *doesn't* match - this might cause weird behaviour. Should move this inside the `if` block as the first thing.

packages/hashi/test/kolibri.spec.js
```diff
@@ -0,0 +1,98 @@
...
+    });
+    it('should be called once', () => {
+      kolibri.shim.getContentById().then(() => {
+        expect(client).toHaveBeenCalledTimes(1);
+      });
+    });
+    it('should return a promise that resolves to a ContentNode object', () => {
+      expect(kolibri.shim.getContentById()).toBeInstanceOf(Promise);
+      kolibri.shim.getContentById().then(() => {
+        expect(client).toBeInstanceOf(Object);
```
rtibbles: We're not asserting anything about the returned data here, only about `client`.

Should return an argument from the function you're chaining off then, and make an assertion that it's the data from the response.

packages/hashi/test/kolibri.spec.js
```diff
@@ -0,0 +1,98 @@
...
+    });
+    it('should be called once', () => {
+      kolibri.shim.getContentByFilter().then(() => {
+        expect(client).toHaveBeenCalledTimes(1);
+      });
+    });
+    it('should return a promise that resolves to an array of metadata objects', () => {
+      expect(kolibri.shim.getContentByFilter()).toBeInstanceOf(Promise);
+      kolibri.shim.getContentByFilter().then(() => {
+        expect(client).toBeInstanceOf(Array);
```
rtibbles: As above

packages/hashi/test/kolibri.spec.js
```diff
@@ -0,0 +1,98 @@
...
+    });
+    it('should be called once', () => {
+      kolibri.shim.getContext().then(() => {
+        expect(client).toHaveBeenCalledTimes(1);
+      });
+    });
+    it('should return a promise that resolves to a context Object', () => {
+      expect(kolibri.shim.getContext()).toBeInstanceOf(Promise);
+      kolibri.shim.getContext().then(() => {
+        expect(client).toBeInstanceOf(Object);
```
rtibbles: And again!

packages/hashi/test/kolibri.spec.js
```diff
@@ -0,0 +1,89 @@
...
+    it('should return a promise that resolves to a ContentNode object', () => {
+      return kolibri.shim.getContentById().then(data => {
+        expect(data).toEqual(response);
+      });
+    });
+  });
+
+  describe('getContentByFilter method', () => {
+    beforeEach(function() {
+      response = [{ 1: [{ node: 'abc' }] }, { 2: [{ node: '123' }] }];
```
jonboiser: Should the `response` match the `typedef` for `PageResult` e.g. `{page, pageSize, results}`? Or does it get transformed inside the shim? If not, a more realistic test fixture is good documentation.

packages/hashi/test/kolibri.spec.js
```diff
@@ -0,0 +1,89 @@
...
+      kolibri.shim.getContentByFilter = jest.fn().mockResolvedValue(response);
+    });
+    it('should be called once', () => {
+      kolibri.shim.getContentByFilter().then(() => {
+        expect(kolibri.shim.getContentByFilter).toHaveBeenCalledTimes(1);
```
jonboiser: It would be a stronger test to spy on `self.mediator.sendMessageAwaitReply` and assert that it is called with the correct `event`, `data`, `namespace` args. Because this test only asserts that the mock was called and is equivalent to the the following code that will never fail even if the implementation of the shim changes:

```
const mock = jest.fn()
mock()
expect(mock).toHaveBeenCalled()
```

Because it does not actually utilize the implementation (since it's mocked), this test provides no coverage over the code.

jonboiser (approved):

PR #7947, opened 2021-04-02
Open CSV file with utf-8 encoding in Py3.

### LINKED ISSUE #7946: Exporting CSV file that contains Unicode text fails on Python 3 on Windows
### Observed behavior
User reports of CSV file export failing prompted me to add a test for this in a Windows environment in https://github.com/learningequality/kolibri/pull/7899.

When Unicode text should be inserted into the exported CSV, such as a Unicode username or full_name, then an error occurs on windows.

```
Error trying to write csv file: 'charmap' codec can't encode characters in position 29-35: character maps to <undefined>
```

### Expected behavior
CSV Export should work on all platforms.

This is fixed by this commit: https://github.com/learningequality/kolibri/pull/7899/commits/c088d08b3e6c4a0f4194a17062457cbd7ae34816

### User-facing consequences
Windows users are unable to export CSV files.

rtibbles:
### Summary
* We do not open files CSV files for writing with unicode encoding on Python 3
* This causes writing unicode to CSV files fail on Windows
* This fixes that.

### Reviewer Guidance
* Install on Windows
* Create user with unicode username and full name
* Login with user, interact with some content
* Log back in as admin or super admin and attempt to export session logs
* Verify that session logs export

### References
Cherry-picked commit from https://github.com/learningequality/kolibri/pull/7899/commits/c088d08b3e6c4a0f4194a17062457cbd7ae34816 that is part of https://github.com/learningequality/kolibri/pull/7899 that has the full test and fix for this as well as Windows tests.

Fixes #7946

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/core/logger/csv_export.py
+++ b/kolibri/core/logger/csv_export.py
@@ -154,7 +154,7 @@ def csv_file_generator(facility, log_type, filepath, overwrite=False):
     if sys.version_info[0] < 3:
         csv_file = io.open(filepath, "wb")
     else:
-        csv_file = io.open(filepath, "w", newline="")
+        csv_file = io.open(filepath, "w", newline="", encoding="utf-8")
 
     with csv_file as f:
         writer = csv.DictWriter(f, header_labels)


nucleogenesis:
@rtibbles if you're confident this fixes the issue with no room for regression (seems to be the case) then go ahead and merge - otherwise can you please provide some reviewer guidance here for whoever manually confirms this in Windows?

rtibbles:
Added  - note that I added another PR to address the build issue here: https://github.com/learningequality/kolibri/pull/7949

kolibri/core/logger/csv_export.py
```diff
@@ -154,7 +154,7 @@ def csv_file_generator(facility, log_type, filepath, overwrite=False):
     if sys.version_info[0] < 3:
         csv_file = io.open(filepath, "wb")
     else:
-        csv_file = io.open(filepath, "w", newline="")
+        csv_file = io.open(filepath, "w", newline="", encoding="utf-8")
```
jonboiser, line 157: I see code similar to this in places like here

https://github.com/learningequality/kolibri/blob/ab3b75dc5f5e2fa4df303ce3dba086913618f727/kolibri/core/auth/csv_utils.py#L189-L193

does the `encoding` need to be set there too?

rtibbles: You mean for the `sys.version_info[0] < 3` code path? No - I had originally added that and it broke stuff, I think it's not needed because it is opened in `'wb'` mode as opposed to just `'w'` - which again is due to weird differences between the csv module's behaviour in Python 2 vs 3.

pcenov:
@rtibbles and @jonboiser I was able to test and verify this change as fixed with the following Buildkite build: https://buildkite.com/organizations/learningequality/pipelines/kolibri-windows/builds/785/jobs/26f70c82-def5-4985-be68-064dca2294c4/artifacts/40245733-bd06-4a2d-8d6f-772af9fcf159
I'm able to export the session logs. Here's how the content looks when viewed with MS Excel with default settings on my PC:

As you can see for some reason the Chinese and Arabic characters are not displayed correctly.
However when I opened the same file with Google Sheets all characters are displayed as expected:

rtibbles:
Thanks @pcenov - yes, we've seen the Excel issue before when the CSV file is opened with the wrong encoding, unfortunately, I don't think there's a way to make Excel properly detect the encoding by default that I'm aware of.

Issue #6255, opened 2019-12-09
persist content task list across server restarts


indirectlylit:
### Observed behavior

Restarting the server clears the list of completed tasks

### Expected behavior

Completed task list should persist across restarts. Any tasks that were ongoing when the server restarted should show as either canceled or failed.

### User-facing consequences

users lose access to important historical information

### Steps to reproduce

* let a few tasks run to completion
* start a new task
* stop server
* start server

### Context

0.13.0 beta 3

jonboiser:
It looks like the job that underlies the task sticks around in the `job_storage.sqlite3` DB, but it doesn't get returned from the `/api/tasks` endpoint. I think that the job/task metadata doesn't actually get persisted in the job storage DB (but maybe it's stored in memory?), so it might not be serializable after the restart.

rtibbles:
It does get persisted in the database, but at the moment, we clear the job storage DB on server restart, so that's why it doesn't currently persist across restarts.

jonboiser:
I just tried this locally and there are some rows in the `jobs` and `scheduledjobs` tables that don't go away after I restart the webpack server.

I just assumed that they weren't being returned from `/api/task` endpoint, because the only thing these rows have is an `id` and no other displayable metadata.

vkWeb:
I looked into this & I think upon server restart, the jobs are getting cleared from the `job_storage.sqlite3` db. 

But I cannot find the code where this deletion of records is happening :thinking:

rtibbles:
Fixed in #7951

PR #7951, opened 2021-04-04
fix(tasks): tasks queue is cleared on server start

vkWeb:
### Summary
When the kolibri server starts, we were clearing the tasks queue which lead us to lose the historical information of completed, failed or cancelled tasks.

This PR fixes the issue by not letting that happen. 

### Reviewer guidance

1. Complete some tasks or cancel running tasks.
2. Restart the server.
3. See the task manager. 
4. The task list should have persisted.  

### References

Fixes https://github.com/learningequality/kolibri/issues/6255

----

### Contributor Checklist

PR process:

Testing:

### Reviewer Checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/utils/server.py
+++ b/kolibri/utils/server.py
@@ -20,7 +20,6 @@ from kolibri.core.content.utils import paths
 from kolibri.core.content.zip_wsgi import get_application
 from kolibri.core.deviceadmin.utils import schedule_vacuum
 from kolibri.core.tasks.main import initialize_workers
-from kolibri.core.tasks.main import queue
 from kolibri.core.tasks.main import scheduler
 from kolibri.utils import conf
 from kolibri.utils.android import on_android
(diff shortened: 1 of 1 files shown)


vkWeb:
cc @jonboiser, @rtibbles for review.

vkWeb:
I think we should clear the tasks from db when the user clears them from the frontend...? This will simplify frontend's logic as we'll just need to display jobs that get return from `GET api/tasks/tasks` (no filtering) and also it'll save us some memory. 

If we move forward with the current implementation then jobs will persist on db forever.

rtibbles (approved):
This fixes the issue as stated.

rtibbles:
Added a follow up issue to do the same for the task scheduler: https://github.com/learningequality/kolibri/issues/7954

PR #7959, opened 2021-04-06
fix(scheduler): scheduled tasks does not persist on server restart

### LINKED ISSUE #7954: Don't clear scheduled tasks at startup and ensure that the pingback scheduler does not add multiple pingbacks as a result
### Observed behavior
In line with the regular task queue, we clear scheduled tasks at startup.
#7951 changed this behaviour for the regular task queue, but not for the scheduled tasks.

### Expected behavior
We should be able to persist scheduled tasks across server restarts.
The 'schedule_pingback` function should not queue a new ping back in the scheduler if one already exists.

### User-facing consequences
If we implement scheduled tasks created by the user, they would be cleared every time the server restarted.

vkWeb:
## Summary

On #7951 we changed the behaviour for the regular task queue to stay persistent on server restarts, but not for the scheduled tasks. This PR achieves persistency for scheduled tasks. 

- Now, we should be able to persist scheduled tasks across server restarts.
- The `schedule_pingback` and `schedule_vaccum` functions should not queue in the scheduler if one already exists.

## References
Closes https://github.com/learningequality/kolibri/issues/7954

## Reviewer guidance
1. Clear all the records of scheduled_jobs table in job_storage.sqlite3 db to avoid any confusion during testing.
2. Start the server, now check the scheduled_jobs table, two jobs must be scheduled: `_ping` and `vacuum`.
3. Schedule a new job.
4. Restart the server, now all the scheduled jobs (`_ping`, `vacuum` and your own job) must have persisted.
5. Also, make sure that _ping and vacuum aren't queued again on server restarts.

----

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/core/tasks/scheduler.py
+++ b/kolibri/core/tasks/scheduler.py
@@ -250,23 +250,3 @@ class Scheduler(StorageMixin):
 
     def _now(self):
         return local_now()
-
-    def is_func_scheduled(self, func):
-        """
-        Returns True if the given func is scheduled for execution else False.
-
-        :type func: a callable or str.
-        :param func: the func to be checked whether scheduled or not.
-        """
-        if callable(func):
-            func = stringify_func(func)
-
-        if not isinstance(func, str):
-            raise TypeError("func must be a callable or str type.")
-
-        all_scheduled_jobs = self.get_jobs()
-        for scheduled_job in all_scheduled_jobs:
-            if scheduled_job.func == func:
-                return True
-
-        return False
--- a/kolibri/core/tasks/scheduler.py
+++ b/kolibri/core/tasks/scheduler.py
@@ -16,6 +16,7 @@ from kolibri.core.tasks.job import State
 from kolibri.core.tasks.queue import Queue
 from kolibri.core.tasks.storage import StorageMixin
 from kolibri.core.tasks.utils import InfiniteLoopThread
+from kolibri.core.tasks.utils import stringify_func
 from kolibri.utils.conf import OPTIONS
 from kolibri.utils.time_utils import local_now
 from kolibri.utils.time_utils import naive_utc_datetime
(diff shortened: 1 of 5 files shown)


vkWeb:
cc @rtibbles for review.

kolibri/utils/server.py
```diff
@@ -86,22 +85,29 @@ def __init__(self, bus, port):
...
     def start(self):
-        # Initialize the iceqube scheduler to handle scheduled tasks
-        scheduler.clear_scheduler()
-
+        # If pinging is not disabled by the environment
         if not conf.OPTIONS["Deployment"]["DISABLE_PING"]:
 
-            # schedule the pingback job
-            from kolibri.core.analytics.utils import schedule_ping
+            # schedule the pingback job if not already scheduled
```
rtibbles: I think I'd prefer to put this check inside the `schedule_ping` function rather than in here.

kolibri/core/tasks/scheduler.py
```diff
@@ -249,3 +250,23 @@ def _ns_query(self, session):
 
     def _now(self):
         return local_now()
+
+    def is_func_scheduled(self, func):
```
rtibbles: I think there's a simpler solution here.

When we schedule a task: https://github.com/learningequality/kolibri/blob/develop/kolibri/core/tasks/scheduler.py#L143 we are able to pass in arbitrary args and kwargs for the job constructor.

If we update the job constructor to allow a supplied job_id to be passed in:
https://github.com/learningequality/kolibri/blob/develop/kolibri/core/tasks/job.py#L90:
`self.job_id = kwargs.get("job_id", uuid.uuid4().hex)`

Then we can just choose a constant job_id for the scheduled pingback task and just check that it exists.

This also has the virtue that we can schedule additional tasks for the `_ping` if we want.

rtibbles (changes requested):
Can simplify approach and do some cleanup.

Would also be a good to add some tests of this behaviour.

kolibri/core/tasks/scheduler.py
vkWeb: Wow :heart:, your approach is so simple!  

I just have one nit thought, don't you think assigning constant job ids might become an issue when jobs like ping are in good numbers...?

kolibri/utils/server.py
vkWeb: Yes, I'll do that.

kolibri/core/tasks/scheduler.py
rtibbles: We should expect that the `schedule_ping` function just ensures that there is a single ping scheduled - if other `_ping` jobs need to be scheduled, then they can not pass the constant job_id, and so will be assigned a random uuid.

vkWeb:
@rtibbles sir approach has been simplified as you suggested. I'm new to mock so I'm learning it, tests will be pushed by tomorrow.

rtibbles (approved):
This works as expected in local testing, and has the added benefit of not repeating the ping back on repeated restarts of the server - it persists the previous success and will ping again when the time is right!

PR #7932, opened 2021-03-29
When toggling Demo Banner, do not send focus to a new location outside of the sign-in page forms

apurva-modi:
Recorded the previous element that holds focus before the close button is clicked and transfer focus to that element once the header is dismissed

7661-demo

…

### Contributor Checklist

PR process:

- [x] PR has the correct target branch and milestone
- [ ] PR has 'needs review' or 'work-in-progress' label
- [x] If PR is ready for review, a reviewer has been added. (Don't use 'Assignees')
- [ ] If this is an important user-facing change, PR or related issue has a 'changelog' label
- [ ] If this includes an internal dependency change, a link to the diff is provided

Testing:

- [x] Contributor has fully tested the PR manually
- [x] If there are any front-end changes, before/after screenshots are included
- [ ] Critical user journeys are covered by Gherkin stories
- [ ] Critical and brittle code paths are covered by unit tests

### Reviewer Checklist

- Automated test coverage is satisfactory
- PR is fully functional
- PR has been tested for [accessibility regressions](http://kolibri-dev.readthedocs.io/en/develop/manual_testing.html#accessibility-a11y-testing)
- External dependency files were updated if necessary (`yarn` and `pip`)
- Documentation is updated
- Contributor is in AUTHORS.md

PR #7973, opened 2021-04-07
Cherry-pick #7932 and remove admin credentials from demo banner

jonboiser:
## Summary

1. Cherry-picks #7932 to 0.14.7
2. Removes admin credentials from the demo banner copy


pcenov:
Tested and verified at https://kolibri-beta.learningequality.org/en/user/#/signin - the admin credentials are removed from the demo banner.

pcenov:
@jonboiser I just noticed that now that the admin credentials have been removed we should probably change the text above the credentials as well which is incorrect because it says: "Explore any of the three primary user types:"

@radinamatic

Issue #7974, opened 2021-04-07
focus outline can appear around elements when hovering, even outside of keyboard modality


jonboiser:
## Observed behavior

These are some screenshots I made browsing https://kolibri-dev.learningequality.org, which is using kolibri-design-system 0.2.2b1.

These screenshots were created while using the mouse exclusively (i.e. never entering into keyboard modality, at least intentionally by tabbing into elements).

**On video player (hovering)**

![CleanShot 2021-04-07 at 12 09 52](https://user-images.githubusercontent.com/10248067/113938693-cb406c80-97af-11eb-9f8c-5e3fcac8b2f6.png)

**Learn search bar**

![CleanShot 2021-04-07 at 12 09 31](https://user-images.githubusercontent.com/10248067/113938735-d7c4c500-97af-11eb-9ee0-78bfd4d062c5.png)

**Under top nav while side nav is open**
![CleanShot 2021-04-07 at 12 06 22](https://user-images.githubusercontent.com/10248067/113938750-e0b59680-97af-11eb-9368-89f6ea8fbe37.png)

**Core menu while hovering**

![CleanShot 2021-04-07 at 12 05 27](https://user-images.githubusercontent.com/10248067/113938794-ef03b280-97af-11eb-9fa5-b30e89778a93.gif)

## Expected behavior

In all of these screenshots, there should not be any outline

## Steps to reproduce

Go to the various parts of the app in Chrome and you will see the outline when hovering with mouse (or sometimes without having to hover)

Priyaraj17:
Hey, We just need to add 'outline: none' in the css files. I will get started ASAP.

jonboiser:
> Hey, We just need to add 'outline: none' in the css files. I will get started ASAP.

@Priyaraj17 I don't think the solution is simple as that. We want the green outline to show when using the keyboard "tab" button to navigate the UI. Try navigating Kolibri right now with just the keyboard to see what I mean.

So any change to the CSS that will remove that is unacceptable (i.e. setting `outline: none` globally).

Priyaraj17:
Hey, I have fixed a few issues and now I will be working on the core menu and top nav when side nav is open.

Priyaraj17:
@jonboiser I was working on 2 issues simultaneously. I have solved the issues here. Shall I reference a pull request that I have made while solving a different issue? I believe it will resolve this issue too.

Priyaraj17:
[...]

PR #7983, opened 2021-04-08
Validate usernames during sign-in flow + fix bug in facility settings page

### LINKED ISSUE #7602: Sign in page needs validation for incorrect usernames
### Observed behavior

When password is required, there is no validation for the username field before the password is asked for

Screen Shot 2020-10-20 at 12 36 30 PM

### Expected behavior

If the user enters an invalid username for the current facility, they should see a validation message for an incorrect username

### User-facing consequences

The user may make multiple attempts at signing in and experience frustration before realizing their username was wrong in the first place

### Steps to reproduce

1. Enable the facility setting that requires passwords, or sign in as a coach/admin
2. Go to sign in page
3. Enter an invalid username
4. Click "Next"

### Context

Kolibri 0.14.4.dev0
MacOS
Chrome

### LINKED ISSUE #7977: Facility settings page makes invalid request when unchecking "Require password for learners" option
## Observed behavior

1. Start with "Require password..." and "Allow learners to edit..." both checked
2. Uncheck "Require password..."
3. Press "Save changes"
4. An error occurs because of the validation implemented in #7856. "Allow learners..." looks disabled, but is actually still set to `true` on the client side.

CleanShot 2021-04-07 at 18 02 45

## Expected behavior

The scenario above works, and sends a valid request to the facility dataset endpoint.

## Steps to reproduce

See OB section

## Context

K 0.15

jonboiser:
## Summary

1. Fixes #7977 
2. Fixes #7602 
3. Renames the keypress handler for the username textbox and changes it to always call `event.preventDefault` when pressing "Enter". This is because it was causing the form to be submitted twice when using the keyboard.
4. Changes the username + password error messages in the sign in page
5. Improves the error UX by automatically selecting the invalid text inputs so they can be edited more easily.

Here is a video of the sign-in flow. I use three methods to interact with the username + password fields: 1) keyboard enter in input, 2) keyboard input on submit button, 3) keyboard to input text, but mouse to click button

CleanShot 2021-04-08 at 16 30 49

## Reviewer guidance

1. Carefully review the code fixing the facility settings page + test it manually.
2. Test sign-in page and look for any ways to bypass the username validation.
3. Consider the changes to the keypress handler I made. 

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated


kolibri/plugins/user/assets/src/views/SignInPage/index.vue
```diff
@@ -483,7 +511,14 @@
       },
     },
     $trs: {
-      signInError: 'Incorrect username or password',
+      incorrectPasswordError: {
+        message: 'Incorect password',
```
jtamiace:
```suggestion
        message: 'Incorrect password',
```

kolibri/plugins/user/assets/src/apiResource.js
```diff
@@ -1,5 +1,26 @@
...
+    },
+  })
+    .then(response => {
+      return response.data.username_exists;
+    })
+    .catch(() => {
+      // Just fail silently
+      return false;
+    });
+}
```
nucleogenesis, line 26: This logic seems useful in several places in the app - when users create their own account or when admins create users. I wonder if this logic already exists and maybe we should create a follow-up issue to see if we need to DRY up implementations of fetching that endpoint - or we might see if the UX of those parts of the app can likewise be improved.

nucleogenesis (changes requested):
The keyboard focus outline shows on hover or any kind of focus. It should only show up when you're using keyboard navigation to get there, I believe. @radinamatic thoughts?

https://user-images.githubusercontent.com/6356129/114410560-52029a00-9b60-11eb-8477-46290f1bf8ad.mp4

radinamatic:
> It should only show up when you're using keyboard navigation to get there

Visual focus outline should not appear on hovering, but I don't see a problem if it appears on mouse click into an input field, as often that thin flickering cursor line indicating that the input field is, in fact, in focus, is difficult to see, and having the outline around the field helps with that.

The only accessibility blocker is when the outline is NOT present while using the keyboard to navigate with the <kdb>Tab</kdb> key. 

However, I was not able to trigger that outline of the _Username_ field neither by hover, nor by actually navigating into it with the <kdb>Tab</kdb> key. I can see the outline when all other elements come into focus, but zilch on the input field... 😕 

https://user-images.githubusercontent.com/1457929/114490870-dfec8c80-9c15-11eb-9fd1-38b0c0de90f3.mp4

Tried on both Firefox and Chrome on my dev Ubuntu 18.04 VM, and also by connecting to the dev server from a Chrome on a Windows device in the LAN. Am I missing something?

jonboiser:
> The keyboard focus outline shows on hover or any kind of focus. It should only show up when you're using keyboard navigation to get there, I believe. @radinamatic thoughts?

@nucleogenesis I think this is part of #7974, but I didn't add it to the issue since I could only reproduce it locally, and not on the Kolibri-Dev demo server.

kolibri/plugins/user/assets/src/apiResource.js
jonboiser: > when users create their own account or when admins create users.

This is currently handled as an error when POST or PATCHINGing to the `facilityuser` endpoint. So if you try to create a new user as an admin, you will get a "username exists" error _after_ the request is made to the API.

As a new feature, we can check against the `usernameexists` endpoint to invalidate usernames as they are being typed.

nucleogenesis (approved):
Ah - the bug I noted that made me request changes is indeed not due to this PR - approved!

PR #7985, opened 2021-04-09
First pass at MVP single-user syncing backend

### LINKED ISSUE #7858: Create proof of concept single user syncing to test for edge cases and possible issues
To ensure that single user syncing behaviour is feasible and properly scoped, first create a proof of concept that handles the following data set:

* Single facility
* Two classrooms
* Multiple users in each classroom
* Multiple learner groups in each classroom
* Lessons and quizzes in each classroom, both assigned and unassigned
* Learner to be synced is a member of at least two groups
* At least 2 lessons and quizzes are assigned to each of the groups in the learner's classroom
* At least 2 lessons and quizzes are assigned to the learner's classroom

Conduct syncing and identify any errors, or when data gets stuck in the store as not able to deserialize due to missing foreign keys.

jamalex:
## Summary

First pass at #7858, adding the `--user-id` parameter to the `sync` management command, to initiate single-user syncing.

Some fun facts:
- No changes were needed to Morango, or to the "server side" of Kolibri. Just some modifications to the `sync` command itself and to its supporting utils functions. This means that the current code is actually backwards-compatible with older versions of Kolibri running as the "server".
- Given the top-level partition they're under, Lessons, Exams, and their Assignments are not in fact currently included in a single-user sync. So we'll be building a side-channel syncing method for that instead, which was our fallback. That does mean that when syncing against older Kolibri versions, to a single-user device, these models won't be included. For that reason, we may not actually want to allow it to be back-compatible, to avoid confusion.

## References

- Closes #7858.

## Reviewer guidance

I tested this by creating two Kolibri data directories, one to act as the server, the other as the client:

```
export KOLIBRI_HOME=~/.singleusersyncserver
yarn run python-devserver
# and then used the UI to provision it, and add a user
```

```
export KOLIBRI_HOME=~/.singleusersyncclient
# the following values are based on the facility ID and user ID from the other server
kolibri manage sync --user-id=71f1b6a9c53e79f379bd96a6464a3025 --facility=d1cf3dfad3c01583100467336d7b3693 --baseurl=http://127.0.0.1:8000
# when it asks for credentials, you can put in either the superuser or the learner account that were created
```

@rtibbles is working on updating/fixing the automated syncing tests to cover this.

----

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/core/auth/management/commands/sync.py
+++ b/kolibri/core/auth/management/commands/sync.py
@@ -70,7 +70,7 @@ class Command(AsyncCommand):
             help="password of superuser or facility admin on server we are syncing with",
         )
         parser.add_argument(
-            "--user-id",
+            "--user",
             type=str,
             help="for single-user syncing, the user ID of the account to be synced",
         )
@@ -100,7 +100,7 @@ class Command(AsyncCommand):
             options["chunk_size"],
             options["username"],
             options["password"],
-            options["user_id"],
+            options["user"],
             options["no_push"],
             options["no_pull"],
             options["noninteractive"],
--- a/kolibri/core/auth/management/commands/sync.py
+++ b/kolibri/core/auth/management/commands/sync.py
@@ -1,6 +1,7 @@
 import json
 import logging
 import math
+import re
 from contextlib import contextmanager
 
 from django.core.management import call_command
@@ -14,6 +15,8 @@ from ..utils import create_superuser_and_provision_device
 from ..utils import get_baseurl
 from ..utils import get_client_and_server_certs
 from ..utils import get_dataset_id
+from ..utils import get_single_user_sync_filter
+from ..utils import provision_single_user_device
 from kolibri.core.auth.constants.morango_sync import PROFILE_FACILITY_DATA
 from kolibri.core.auth.constants.morango_sync import ScopeDefinitions
 from kolibri.core.auth.constants.morango_sync import State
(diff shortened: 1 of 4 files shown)


indirectlylit:
> Given the top-level partition they're under, Lessons, Exams, and their Assignments are not in fact currently included in a single-user sync.

Sounds like this is new since [the last discussion](https://www.notion.so/learningequality/Review-proposed-0-15-scope-e6c51fd08a13435a8df3485c1f05b5dc)? I'd really hoped to avoid side channels if possible

kolibri/core/auth/management/commands/sync.py
```diff
@@ -59,12 +62,17 @@ def add_arguments(self, parser):
...
+            help="username of superuser or facility admin on server we are syncing with",
         )
         parser.add_argument(
             "--password",
             type=str,
-            help="password of superuser on server we are syncing with",
+            help="password of superuser or facility admin on server we are syncing with",
+        )
+        parser.add_argument(
+            "--user-id",
```
rtibbles: For consistency with the `--facility` argument, could we change this just to `--user`?

indirectlylit:
notes from conversation: 

https://www.notion.so/learningequality/Syncing-of-lessons-and-quizzes-beaf4711881f4077b949a37d962143d9

kolibri/core/auth/management/commands/sync.py
jamalex: Main thing to disambiguate it from is `--username` (which is used for authentication, vs this one, which is specifying a) that it's a single-user sync, and b) which user to sync)

jamalex: image

jamalex: 👍

rtibbles (approved):

rtibbles:
MacOS Build agent appears to be waiting forever.

PR #7989, opened 2021-04-09
Frontend only H5P Rendering and xAPI progress tracking integration

### LINKED ISSUE #7776: Video can't be forwarded in h5p_viewer in Google Chrome
### Observed behavior
When using Chrome to watch a video that's inside a h5p container, when clicking on the video timeline to move to a different point the video always goes back to the beginning. THere's no way to jump to a new position

Using Firefox (both normal and developer version) this feature works correctly.

### Expected behavior
The video timeline should allow the user to move to different parts of the video

### User-facing consequences
No way to go back or forward when watching a video

### Errors and logs
In the browser console:

`:8080/zipcontent/612a7a2941596b60ef89709e150f8cc3.h5p/content/videos/files-5fb7f16f211b2.mp4#tmp:1 Failed to load resource: net::ERR_CONTENT_LENGTH_MISMATCH
`

As a side note, this error is also shown in the browser console:
```
pluginMediator.js:122 Kolibri Modules: kolibri.plugins.media_player.main registered
vue.runtime.esm.js:1888 TypeError: Cannot read property 'icon' of undefined
    at Sub.svgIconComponent (index.vue:62)
    at Watcher.get (vue.runtime.esm.js:4479)
    at Watcher.evaluate (vue.runtime.esm.js:4584)
    at Sub.svgIconComponent (vue.runtime.esm.js:4836)
    at Sub.<anonymous> (index.vue?648c:1)
    at Sub.Vue._render (vue.runtime.esm.js:3548)
    at Sub.updateComponent (vue.runtime.esm.js:4066)
    at Watcher.get (vue.runtime.esm.js:4479)
    at new Watcher (vue.runtime.esm.js:4468)
    at el (vue.runtime.esm.js:4073)
logError @ vue.runtime.esm.js:1888
```

This second error is not important: if the resource has not an icon it appears in the console. Theorically it should not be related to the forward/backward problem, but just in case.

### Steps to reproduce
Load content from a channel with a h5p video and test it 

### Context
Kolibri 0.14.6
h5p plugin enabled with  `kolibri plugin enable kolibri.plugins.h5p_viewer`
Browsers: Google chrome Versión 88.0.4324.96 (Build oficial) (64 bits) 
Firefox 84.0.2
Firefox developer 83.0b7
OS: Ubuntu 20.04

### LINKED ISSUE #8127: Update kolibri-tools eslint parsing to handle webpack dynamic `import` syntax
## Observed behavior
Currently we still have to use `require.ensure` for dynamic import syntax in webpack. This is because our eslint parser does not recognize the `import` keyword as being valid to use except at the top level of the module.

## Expected behavior
Can use dynamic `import` syntax

rtibbles:
## Summary

# H5P
* Currently, H5P rendering is executed as a Python reflection of https://github.com/tunapanda/h5p-standalone/
* This has the distinct advantage of working, and also reflecting the basic mechanism of how H5P is served on PHP
* However, because of design choices in how H5P is constructed, it tends to rely on hundreds of files to render a single content item, leading to an enormous number of simultaneous requests to the backend.
* It's not clear how well this will scale going forward, and is also not how the H5P PHP backend handles things
* The H5P PHP backend, which is the mostly widely used and well supported (https://github.com/h5p/h5p-php-library) relies on some PHP precaching of H5P assets - creating concatenated Javascript and CSS files that get served for each H5P content item. This allows a large reduction in the number of static assets served.
* We have recently implemented an entirely client side rendering flow for our own internal .perseus files (https://github.com/learningequality/kolibri-exercise-perseus-plugin/blob/release-v1.3.x/kolibri_exercise_perseus_plugin/assets/src/views/PerseusRendererIndex.vue#L423) - as our .perseus files and .h5p files are both just differently named zip files, a similar approach is taken here.
* Instead of using the zipcontent endpoint to load individual assets, the entire H5P file is preloaded to the frontend, and then unzipped, parsed, and each file turned into a Blob and a URL for that blob generated. (this is the same mechanism that [EpubJS](https://github.com/futurepress/epub.js/) uses to render Epubs in the frontend).
* We mirror the path substitution done in the PHP implementation for CSS concatenation, to ensure that all relatively referenced assets in CSS files are instead referenced by their new Blob URLs.
* For the user defined contents referenced in the H5P `content/content.json` we shim the `H5P.getPath` method to do a lookup into our own internal directory so we can return the Blob URLs.
* We also shim the H5P ContentType class to override the `getLibraryFilePath` which allows us to return our blob URLs for files inside the content widget library folders.

# xAPI
* H5P uses xAPI to log progress information and updates. In order to have a full integration for H5P, this PR also implements an xAPI statement validation layer, and a Javascript only API for xAPI.
* This does open up the possibility for HTML5 apps in general to use xAPI, but avoids the full blown Learning Record Store implementation that would be required to support the deprecated [xAPI Launch spec,](http://adlnet.github.io/xapi-launch/) and the newer [CMI5 Launch specification](https://github.com/AICC/CMI-5_Spec_Current/blob/quartz/cmi5_spec.md#content_launch)
* In line with our SCORM implementation, this adds validation and tests for that validation for xAPI statements
* It then adds a simple xAPI object accessible inside the inner iframe window, with an interface inspired by [xAPIWrapper](https://adlnet.github.io/xAPIWrapper/) but using Promises, and terminating in a local object store, while relying on general purpose hashi mechanisms to propagate it into saved state
* It then applies heuristics based on the [CMI5 vocabulary](https://github.com/AICC/CMI-5_Spec_Current/blob/quartz/cmi5_spec.md#xapi_data_model) and the verbs that are defined for use in H5P, in order to calculate progress
* It then enhances how Hashi reports progress to use either the xAPI calculated progress or the SCORM calculated progress

# zipcontent
* Serves all Hashi and H5P from static HTML files to avoid the complexity of creating paths for them on the WSGI server
* Hosts `zipcontent` and `static` files under the `content` path, to allow static versions of the alt origin server to be served seamlessly

Net result:

H5P loads with a single network request - albeit a longer one for large files.
H5P now reports progress to Kolibri.

Outstanding questions:
* What should we use for activity URIs for Kolibri content nodes? A URI based on a domain we own, see below.
* What should we use to identify Kolibri users to xAPI? An email address - see below.
* Should we define the `platform` property of the xAPI `context` object? No, this will add unnecessary bloat

Possible things that are still broken:
* I have not done an exhaustive check of H5P content, but have checked that all of the ones in @radinamatic's QA channel do indeed work
* I may have been too rigorous, or not rigorous enough in some of my validation work on the xAPI statements

TODOs:
* [x] Use zipcontent endpoint for the H5P file for getLibraryPath
* [x] Resizing the iframe so that the H5P does not inner scroll
* [x] Update User identifier to use a kolibri specific fake email
* [x] Update H5P static file paths to handle cache busting when we upgrade
* [x] Don't use contentNameSpace for H5P content identifier - just use a dummy value
* [x] Dummy activity id for xAPI statements
* [x] Strip out data before emitting from iframe
* [x] ~Reconstitute when bringing back in~ Not bothering with this as H5P never reads this data back in.
* [x] Prevent resize from emitting same value twice
* [x] Code split to load xAPI and H5P code on demand to avoid adding ~200KB to the Hashi bundle on every load just for H5P and xAPI
* [x] Concatenate all JS and CSS to prevent execution order issues
* [x] Fix residual display bugs for Course Presentation and Chart types
* [x] Handle errors during H5P rendering (e.g. Find the Words type breaks on IE11 due to internal incompatibilities in the H5P renderer code) and report an error to flag for the user

## References
Fixes #7776
Fixes #8127 

## Reviewer guidance
Is the code sufficiently well commented for posterity?
Where could more direct quotation of specification documents be helpful, could any of the foregoing essay be useful in the code?
Does H5P work as expected?
Do H5P activities record progress as expected?

----

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated


jamalex:
>The only thing we can't intercept is if a library calls the top level H5P.getLibraryPath method, and then tries to manually construct the path. The only work around I can think for this is for us to parse H5P files at import and keep a library of these files as static/media files somewhere, and have getLibraryPath point to that endpoint.

Could we not use the zipcontent endpoint as a fallback in this case?

>What should we use for activity URIs for Kolibri content nodes?

My suggestion would be something like `https://content.<somedomainweown>/<content_id>`.

- We want to use `content_id` rather than `node_id` because when we're logging interactions we want those to be scoped to the "identity" of the content itself, regardless of where it lives, with things like `node_id` and `channel_id` perhaps being included in the "context" fields.
- I don't think there's a strict requirement that these URLs serve anything in particular (or even exist), but having them on a domain we control means we could later choose to display something (or add a redirect) as useful.

>What should we use to identify Kolibri users to xAPI?

I'd still suggest (as we discussed briefly) something like a `mailto:<user_id>@<somedomainwecontrol>` (or the hashed version of that). Similarly, we don't need to have these emails be functional out of the gate, but we could choose to do something with them later.

jredrejo:
Tested with normal h5p files and h5p videos and checked this PR fixes #7776

MBKayro:
The latest update fix a lot of problems, tested with ricecooker and H5P work perfectly and fast

jamalex (approved):
@rtibbles walked us through the riskier bits on a call, and we were unable to... trip him up, so this seems good to me! The proof will be in some thorough manual testing pudding.

kolibri/core/content/utils/paths.py
```diff
@@ -277,7 +277,7 @@ def zip_content_path_prefix():
 
     if path_prefix != "/":
         path_prefix = "/" + path_prefix
-    return path_prefix
+    return get_content_url(path_prefix)
```
MBKayro: i have this error : 
Uncaught (in promise) Error: Ajax error for /content/storage/6/b/6b843f1b8e7e37300fc3bce3dfa81706.h5p : 404 Not Found

and

Validation failed for 'download_url' in  
Object { id: Getter & Setter, contentnode: Getter & Setter, priority: Getter & Setter, preset: Getter & Setter, supplementary: Getter & Setter, thumbnail: Getter & Setter, checksum: Getter & Setter, available: Getter & Setter, file_size: Getter & Setter, extension: Getter & Setter, … }
mixin.js:61

rtibbles: I've not seen the first issue - and that's precisely where the H5P file should be, so that's a little odd.

The second is a residual prop on the renderer mixin that is no longer being passed in, nothing to do with this work.

MBKayro: this is what i see:

in a fresh installation, i have the first error to not load H5P in front-end, however the file missing is actually in content.

However, when i use      "return path_prefix" i can load h5p normaly, but when i switch to new "return get_content_url(path_prefix)" the H5P load and no error popup. so maybe it need a fresh content (.kolibri) to see error or maybe something i do wrong

rtibbles: Thanks @MBKayro - the change I made was too aggressive for the result I was looking for, have updated it.

MBKayro:
a lot of progress, and happy for the speed to load h5p, still to test in android device. Thanks @rtibbles for these great work

MBKayro:
it ill be great to send the "Total Score" of H5P activity to coach reports :)

radinamatic:
H5P looking good on all browsers! 💯 

Two detailed noticed: warning about a deprecated H5P library, and a weird interaction where the content renderer section gets expanded as the user progresses through the activity and is pushing the rest of the page-container sections below, leaving empty space (happens in all the browsers).

IE11 | Chrome
---|---
Win7 (start)  Running  - Oracle VM VirtualBox_059 | Win7 (start)  Running  - Oracle VM VirtualBox_061

rtibbles:
The deprecation warning is expected.

The ever expanding content is not.

radinamatic:
> The deprecation warning is expected.

😎 

> The ever expanding content is not.

It's not _ever expanding_, just enough to be wasteful!

radinamatic:
Another comment that comes to mind: one of these resources took more than 15 seconds to load, but if full load is by design (if I understand correctly), is there a way for us to display that _loading_ status to the user?

radinamatic:
Yey for the loader, and much less wasted space!

Looks like IE11 is willing to load a video from YT, even though it's not available online, while Firefox & Chrome get blocked in the loading the player... 🤔 

IE11 | Firefox | Chrome
--- | --- | ---
load-ie11 | Win7 (start)  Running  - Oracle VM VirtualBox_066 | Win7 (start)  Running  - Oracle VM VirtualBox_067

A part from this video hickup, let's wait for @pcenov to chime in about testing on a wider variety of H5P resources, before we merge!

pcenov:
@radinamatic I tested the rendering of the H5P resources in IE11 and Chrome on Windows 7 by importing the H5P Samples channel and did not observe any IE11 specific issues.
However I stumbled upon multiple issues that are valid for all browsers such as:
1. Some of the resources like Chart, Collage and Course Presentation do not load at all:


2. The first card of the Dialog Cards resource doesn't load until I click the Next arrow:


3. The Virtual Tour (360) doesn't seem to work in both browsers and in Chrome it doesn't even load:


4. I'm seeing various errors in the Console such as:

`TypeError: Cannot read property 'icon' of undefined`
`Error: Kolibri/core/assets/src/views/ContentRenderer/filePresetStrings.js] Download translation string not defined for 'h5p'`
`Uncaught TypeError: Cannot read property 'on' of undefined`
`Unable to find constructor for: H5P.Chart 1.2`

5. Occasional 400 error:

pcenov:
@radina I am able to install [the latest build](https://buildkite.com/organizations/learningequality/pipelines/kolibri-windows/builds/1326/jobs/14f7d3b9-6010-4d59-87b3-323835175519/artifacts/5f4ba2ca-47b3-4a6a-a33c-501ce89839b2) on Windows 7, run it in IE11, import the 43 H5P resources from the H5P Samples channel and see most of them.
The following resources still won’t load: Chart, Collage, Course Presentation, Essay, Find Multiple Hotspots, Find the Words, Image Hotspots, Image Juxtaposition, Impressive Presentation, Single Choice Set
The Dialog Cards and the Virtual Tour (360) issues are still present.

MBKayro:
hi @pcenov  i have the same problem, however i think  that the channel must be updated with the latest H5P plugins. i generate new H5P content and upload it with recicooker and use it to test and most of H5P work perfectly like presentation ....

rtibbles:
In my local testing, I had previously seen issues with several of the H5P content types, but had managed to eliminate them.

I think there may be some timing/performance issues happening where in some machines things are happening in a different order, and this is causing things to present differently.

@MBKayro - yes, I imagine upgrading the content might resolve many issues, but I am also keen to make it as backwards compatible as possible, as, unlike an online instance, Kolibri will not be able to frequently update the H5P packages.

radinamatic:
H5P | Browser | Screenshot
--- | ---- | ---
Accordion | - FF ❌</br>   - Chr. ✔️ </br> - IE11 ✔️  | Win7 (start)  Running  - Oracle VM VirtualBox_086
Advanced Fill the Blanks | - FF ❌</br>   - Chr. ✔️ </br> - IE11 ✔️  | Win7 (start)  Running  - Oracle VM VirtualBox_087
Arithmetic Quiz</br> </br> **note the differences between Chrome & IE11** | - FF ❌</br>   - Chr. ✔️ </br> - IE11 ✔️  | Win7 (start)  Running  - Oracle VM VirtualBox_088
Chart | - FF ❌</br>   - Chr. ❌ </br> - IE11 ❌  | Win7 (start)  Running  - Oracle VM VirtualBox_089
Collage | - FF ❌</br>   - Chr. ✔️ </br> - IE11 ❌  | Win7 (start)  Running  - Oracle VM VirtualBox_090
Course Presentation | - FF ✔️</br>   - Chr. ✔️ </br> - IE11 ❌  | Win7 (start)  Running  - Oracle VM VirtualBox_091

I'll continue, mixed bag so far... 😕

radinamatic:
With the latest asset I found one activity (*Find the Words*) that fails in Firefox, and another (*Interactive Video*) that fails to load in both FF & Chrome on Windows 7.

FF | FF | Chrome
--- | --- | ---
Win7 (start)  Running  - Oracle VM VirtualBox_002 | Win7 (start)  Running  - Oracle VM VirtualBox_003 | Win7 (start)  Running  - Oracle VM VirtualBox_004

pcenov:
Hi @radinamatic 
Here’s what I found while testing the latest build on Windows 7 and Windows 10 https://buildkite.com/learningequality/kolibri-windows/builds/1379#5079953e-2c76-4117-a8cb-322562e1bb88 using Firefox, IE11, Edge and Chrome

1. The H5p resources are working significantly better than the previous time I checked. Here are the issues I was able to find:

WIN 7:

Firefox 90.0 (32-bit):

For some reason there are multiple H5P resources which are not displayed correctly in Firefox on Win 7:
Arithmetic Quiz, Chart, Drag the Words, Impressive Presentation, Personality Quiz (zoomed in too much), Single Choice Quiz








IE 11

Find the Words is stuck loading -- known issue, as discussed

The interactive Video resource is displayed only in IE11 --- known issue, as discussed

WIN 10

IE 11
The Impressive Presentation resource won’t load:


2. Loading of audio files in the Kamakalima channel - I did not experience any issues in any of the four browsers, the audio and resources were displayed correctly.

rtibbles:
> For some reason there are multiple H5P resources which are not displayed correctly in Firefox on Win 7:
Arithmetic Quiz, Chart, Drag the Words, Impressive Presentation, Personality Quiz (zoomed in too much), Single Choice Quiz

I think this must be to do with resizing logic. I will try reinstating the fix for this once more.

For the Timeline - the images not displaying is due to their being loaded from an external site, this is expected and a wontfix. IE11 is odd in that it lets the images through in spite of the content security policy.

> The Impressive Presentation resource won’t load:

My guess is that this is specific to IE11 and outside of our control. I verified this by attempting to render https://h5p.org/impressive-presentation (the source of the demo content) in BrowserStack IE11, it too failed to load, so this is a wontfix, as it's outside of our control.

radinamatic:
With the last asset from yesterday I was experiencing similar issues in Firefox as @pcenov above, but it would be inconsistent upon page reload, resource appearing squished or zoomed in, then loading fine again, then squished again... 

The latest asset from today managed to fix that for all resources minus the **Personality Quiz**:

first load | reload 1 | reload 2 | reload 3 | reload 4
--- | --- | --- | --- | ---
Win7 (start)  Running  - Oracle VM VirtualBox_022 | Win7 (start)  Running  - Oracle VM VirtualBox_023 | Win7 (start)  Running  - Oracle VM VirtualBox_024 | Win7 (start)  Running  - Oracle VM VirtualBox_026 | Win7 (start)  Running  - Oracle VM VirtualBox_025

A part from that, all the rest is good to go! 👏🏽

rtibbles:
I think I'm happy to leave "inconsistent scaling of background image on the Personality Quiz type" as a known issue...

radinamatic (approved):
> I think I'm happy to leave "inconsistent scaling of background image on the Personality Quiz type" as a known issue...

I wholeheartedly agree, approved! 🎉

rtibbles:
Failed MacOS Python test check appears to be a weird blip. Merging.

PR #7993, opened 2021-04-10
Make dataset_id caching thread local and only used in sync to avoid locks.

### LINKED ISSUE #7463: Lingering Redis locks from previous dirty shutdowns can cause locks during sign-in that result in errors or timeouts
### Observed behavior

When using Redis, if Kolibri has an open lock while it's forced to shut down, such that the lock isn't released, then Kolibri will run into issues acquiring locks when it is restarted.

### Expected behavior

- Ensure all appropriate defensive code is in place to prevent lingering locks in Redis
- Lingering locks should TBD, options for consideration:
  1. Prevent Kolibri from fully starting without informing the user. This may be appropriate because depending on the lingering lock type, the database could be in an incomplete state, so it would give the user a chance to make a backup. Then perhaps startup could proceed with confirmation from the user
  2. Add a management command to clear locks
  2. Always clear locks on startup
  3. ???

### User-facing consequences

Being unable to acquire locks, Kolibri requests may hang, timeout, or raise an error. Depending on cache efficiency, authentication may be impossible for some users.

### Errors and logs

```
ERROR 2020-08-10 17:18:56,981 django.request Internal Server Error: /api/auth/session/
  File "/usr/local/lib/python2.7/dist-packages/kolibri/dist/django/core/handlers/exception.py", line 41, in inner
  File "/usr/local/lib/python2.7/dist-packages/kolibri/dist/django/core/handlers/base.py", line 187, in _get_response
  File "/usr/local/lib/python2.7/dist-packages/kolibri/dist/django/core/handlers/base.py", line 185, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/usr/local/lib/python2.7/dist-packages/kolibri/dist/django/views/decorators/csrf.py", line 58, in wrapped_view
  File "/usr/local/lib/python2.7/dist-packages/kolibri/dist/rest_framework/viewsets.py", line 116, in view
    return self.dispatch(request, *args, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/kolibri/dist/django/utils/decorators.py", line 67, in _wrapper
    return bound_func(*args, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/kolibri/dist/django/utils/decorators.py", line 149, in _wrapped_view
    response = view_func(request, *args, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/kolibri/dist/django/utils/decorators.py", line 63, in bound_func
    return func.__get__(self, type(self))(*args2, **kwargs2)
  File "/usr/local/lib/python2.7/dist-packages/kolibri/dist/rest_framework/views.py", line 495, in dispatch
    response = self.handle_exception(exc)
  File "/usr/local/lib/python2.7/dist-packages/kolibri/dist/rest_framework/views.py", line 455, in handle_exception
    self.raise_uncaught_exception(exc)
  File "/usr/local/lib/python2.7/dist-packages/kolibri/dist/rest_framework/views.py", line 492, in dispatch
    response = handler(request, *args, **kwargs)
  File "./kolibri/core/auth/api.py", line 577, in create
    login(request, user)
  File "/usr/local/lib/python2.7/dist-packages/kolibri/dist/django/contrib/auth/__init__.py", line 161, in login
    user_logged_in.send(sender=user.__class__, request=request, user=user)
  File "/usr/local/lib/python2.7/dist-packages/kolibri/dist/django/dispatch/dispatcher.py", line 193, in send
    for receiver in self._live_receivers(sender)
  File "/usr/local/lib/python2.7/dist-packages/kolibri/dist/django/contrib/auth/models.py", line 25, in update_last_login
    user.save(update_fields=['last_login'])
  File "/usr/local/lib/python2.7/dist-packages/kolibri/dist/django/contrib/auth/base_user.py", line 80, in save
    super(AbstractBaseUser, self).save(*args, **kwargs)
  File "./kolibri/core/auth/models.py", line 219, in save
    self.ensure_dataset()
  File "./kolibri/core/auth/models.py", line 231, in ensure_dataset
    inferred_dataset_id = self.infer_dataset(*args, **kwargs)
  File "./kolibri/core/auth/models.py", line 694, in infer_dataset
    return self.cached_related_dataset_lookup("facility")
  File "./kolibri/core/auth/models.py", line 190, in cached_related_dataset_lookup
    dataset_id = dataset_cache.get(key)
  File "./kolibri/core/utils/cache.py", line 124, in get
    return self.cache.get(self.make_key(key), *args, **kwargs)
  File "./kolibri/core/utils/cache.py", line 70, in __exit__
    self.release()
  File "./kolibri/core/utils/cache.py", line 64, in release
    self._lock.release()
  File "/usr/local/lib/python2.7/dist-packages/kolibri/dist/redis/lock.py", line 224, in release
    raise LockError("Cannot release an unlocked lock")
LockError: Cannot release an unlocked lock
```

### Steps to reproduce

1. Set up Kolibri to use Redis
2. Open Kolibri shell, and run:
```
from kolibri.core.utils.cache import process_cache
lock = process_cache.lock(
    "namespaced_cache_dataset",
    timeout=2000,  # milliseconds
    sleep=0.01,  # seconds
    blocking_timeout=5,  # seconds
    thread_local=True,
)
lock.acquire()
```
3. Close the shell
4. Kolibri will encounter crippling issues

### Context

Kolibri 0.14

### LINKED ISSUE #7908: Make dataset_id caching thread local and opt-in
### Observed behavior
Issues like #7463 can occur with both Redis cache backends and diskcache cache backends. This is a result of our ProcessLock implementation being left permanently in an acquired state due either to disconnection from Redis, or unclean shutdown of the Kolibri instance.

This results in calls to `infer_dataset` being permanently held in a waiting state for the lock to release, but this will never happen as the process that originally acquired the lock no longer exists.

### Expected behavior
Unclean shutdowns or abrupt Redis disconnections should not result in Kolibri not being able to login, or write any other data to syncable models.

If we could work out how to decide that an acquired lock was no longer valid, we could attempt to remove it - however this seems like a slightly hard problem to solve, and if we go around removing acquired locks incorrectly, then we may end up thinking we've solved issues to do with concurrency, when we actually have not.

The simpler solution seems to be to not rely on a cache based lock for safe concurrent processing. https://github.com/learningequality/kolibri/issues/7891 proposes a solution in this vein for the `db_write_task_lock`, where it is suggested to defer to DB specific locking mechanisms in order to ensure transaction safety. This is a safe approach, but does not work for a situation where writing to a database is not the target of the locking.

For the case of our dataset_id caching, it was principally introduced in order to speed up the deserialization process in syncing (https://github.com/learningequality/kolibri/pull/4494). Given this, the same could be achieved by creating a thread local cache for these values, that is only used on demand.

In this way, in normal operation, calls to `cached_related_dataset_lookup` would not rely on a cache at all, and instead would defer to the database as the source of truth. This would ensure that in ordinary operation, dataset_id lookups always correspond to the latest committed changes in the database.

During the sync operation a thread local dict would be instantiated that would store the cached values and in this case `cached_related_dataset_lookup` would store any values onto this thread local dict.

This would mean that no lock is required, as the cache is only used and created during the runtime of the sync command.

### User-facing consequences
Prevent lingering locks locking logins.

rtibbles:
## Summary
* Prevents uncleared locks for cache operations by removing the use of locks
* Creates a thread local object for caching dataset_ids only when activated
* Activates and deactivates the object during sync

## References
Fixes #7908
Fixes #7463

## Reviewer guidance
I tested this by adding some logging statements to the kolibri/core/auth/models.py `AbstractFacilityDataModel` `cached_related_dataset_lookup` to log when a cached value was found and when one was not found.

During a very simple sync, first datasets were looked up twice without being found in the cache, but then multiple subsequent lookups were found in the cache object.

During repeated logins via the interface, no cached values were used, implying that the activation and deactivation is working as intended.

----

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/core/auth/management/commands/sync.py
+++ b/kolibri/core/auth/management/commands/sync.py
@@ -108,6 +108,7 @@ class Command(AsyncCommand):
             call_command("loaddata", "scopedefinitions")
 
         dataset_cache.clear()
+        dataset_cache.activate()
 
         # try to connect to server
         controller = MorangoProfileController(PROFILE_FACILITY_DATA)
@@ -204,6 +205,7 @@ class Command(AsyncCommand):
             self.job.extra_metadata.update(sync_state=State.COMPLETED)
             self.job.save_meta()
 
+        dataset_cache.deactivate()
         logger.info("Syncing has been completed.")
 
     @contextmanager
--- a/kolibri/core/auth/models.py
+++ b/kolibri/core/auth/models.py
@@ -23,13 +23,13 @@ from __future__ import print_function
 from __future__ import unicode_literals
 
 import logging
+from threading import local
 
 import six
 from django.contrib.auth.models import AbstractBaseUser
 from django.contrib.auth.models import AnonymousUser
 from django.contrib.auth.models import UserManager
 from django.core import validators
-from django.core.cache import cache
 from django.core.exceptions import ObjectDoesNotExist
 from django.core.exceptions import ValidationError
 from django.db import models
@@ -75,11 +75,37 @@ from kolibri.core.device.utils import get_device_setting
 from kolibri.core.device.utils import set_device_settings
 from kolibri.core.errors import KolibriValidationError
 from kolibri.core.fields import DateTimeTzField
-from kolibri.core.utils.cache import NamespacedCacheProxy
 from kolibri.utils.time_utils import local_now
 
 logger = logging.getLogger(__name__)
-dataset_cache = NamespacedCacheProxy(cache, "dataset")
+
+
+class DatasetCache(local):
+    def __init__(self):
+        self.deactivate()
+
+    def activate(self):
+        self._active = True
+
+    def deactivate(self):
+        self._active = False
+        self.clear()
+
+    def clear(self):
+        self._cache = {}
+
+    def get(self, key):
+        if self._active:
+            return self._cache.get(key)
+        return None
+
+    def set(self, key, dataset_id):
+        if self._active:
+            self._cache[key] = dataset_id
+        return None
+
+
+dataset_cache = DatasetCache()
 
 
 def _has_permissions_class(obj):
@@ -203,7 +229,7 @@ class AbstractFacilityDataModel(FacilityDataSyncableModel):
                 dataset_id = getattr(self, related_obj_name).dataset_id
             except ObjectDoesNotExist as e:
                 raise ValidationError(e)
-            dataset_cache.set(key, dataset_id, 60 * 10)
+            dataset_cache.set(key, dataset_id)
         return dataset_id
 
     def calculate_source_id(self):


bjester (approved):
Code changes make sense and look good.

pcenov:
Didn't stumble upon any issues while testing https://github.com/learningequality/kolibri/releases/tag/v0.14.7-rc4 today, @rtibbles please let me know if there's anything specific that I should watch for or test in relation to this.

jonboiser:
@rtibbles @pcenov This version has been deployed on https://kolibri-beta.learningequality.org. I was able to sign-in with no problems, but didn't run any facility syncs or anything like that beforehand. Is there something we can do to test it to confirm that the locks during sign-in were really avoided?

rtibbles:
I think the locks were left lingering due to pod restarts? So, I don't know that we have a reliable way of reproducing. The main concern for me was just making sure that it didn't cause any regressions in behaviour - and then we'd have to take the long view on whether it resolved the locks (although, I would be very surprised if it had not).

PR #8000, opened 2021-04-12
Ensure progress_fraction is propagated through resource api

### LINKED ISSUE #6112: learner progress not propagated all the way up topic tree
### Observed behavior

child topic:

image

parent topic:

image

### Expected behavior

progress is propagated all the way to the root

### User-facing consequences

missing waypoints on finding in-progress and completed resources

### Steps to reproduce

* go deep in a topic tree and make progress on an exercise
* navigate up the topic tree towards the root

### Context

0.13.0 alpha 3

nucleogenesis:
## Summary

In the Learn topicsRoot module's handler `showChannels` - I added some logic to gather the progress for the channels and merge the values into the store so that the display logic that is already in place for the ChannelCard to display the "In progress" icon has the `progress` value it is expecting on each Channel's node.

This won't happen if the user is not logged in - we skip the calls to `ContentNodeProgressResource.fetchCollection` in this case.

I lodash.chunk()ed the calls to that resource to avoid too complex queries, per @rtibbles notes about a similar bit of logic in `RecommendedSubpage.vue`.

## References

Fixes #6112 

## Reviewer guidance

- Check as a user who has progress
- Change to another logged in user who has different progress (see that it is different)
- Log out and view as guest - make sure you don't see any progress on the frontend and that you see no calls to `/api/content/contentnodeprogress`.

----

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/plugins/learn/assets/src/modules/topicsRoot/handlers.js
+++ b/kolibri/plugins/learn/assets/src/modules/topicsRoot/handlers.js
@@ -1,4 +1,6 @@
-import { ContentNodeResource } from 'kolibri.resources';
+import { ContentNodeProgressResource, ContentNodeResource } from 'kolibri.resources';
+import chunk from 'lodash/chunk';
+import merge from 'lodash/merge';
 import { PageNames } from '../../constants';
 import { _collectionState } from '../coreLearn/utils';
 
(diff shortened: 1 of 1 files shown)


kolibri/core/content/serializers.py
```diff
@@ -31,10 +31,20 @@ def __init__(self, *args, **kwargs):
 class ChannelMetadataSerializer(serializers.ModelSerializer):
     root = serializers.PrimaryKeyRelatedField(read_only=True)
     lang_code = serializers.SerializerMethodField()
+    progress_fraction = serializers.SerializerMethodField()
```
rtibbles: We cache this request *forever* in the frontend: https://github.com/learningequality/kolibri/blob/develop/kolibri/core/content/api.py#L127 - meaning that this will never get updated. Also, an upstream cache would cache this, regardless of the user.

This is why we moved `progress_faction` off the ContentNodeViewset data, and instead made the separate `ContentNodeProgressViewset`.

We should fetch this data in the frontend the same way that we do the other content node progress, and just do an addiitonal non-blocking call to get this (and then dynamically update progress to be non-null when it does get set).

```
ContentNodeProgressResource.fetchModel(channel.root, { force: true}).then....
```

rtibbles (changes requested):
Shouldn't update the channel data, as it is user agnostic. Can use the `root` property of the channel metadata to query ContentNodeProgress instead.

kolibri/core/content/serializers.py
rtibbles: Guessing that we probably want to do this for multiple channels at once, so can do `fetchCollection({ id__in: channels.map(c => c.root)});`

rtibbles: Example here of how it was done for the recommended page: https://github.com/learningequality/kolibri/blob/44bf1ed3591321f4b7ef7c4e5c6ceda8c01db7ae/kolibri/plugins/learn/assets/src/views/RecommendedSubpage.vue#L88

With some guards against sending too many ids at once (which is something else that caused issues previously).

rtibbles:
Also - when you revisit this - if you wanted to clean up `ContentNodeProgressViewset` and consolidate all the progress calculation  logic in there, I would be very happy.

jonboiser (approved):

PR #8004, opened 2021-04-13
Implementation to enforce convenient write locks on Postgres and Sqlite  

### LINKED ISSUE #7891: Replace db_task_write_lock with a solution that is specific to database backends and removes reliance on cache layers
### Observed behavior
We have seen various issues with our attempts to use both diskcache and Redis to store a write lock for tasks on the database. This includes the write lock being purged from Redis, and it not being properly turned off from diskcache.

### Expected behavior
We should be able to easily infer whether a task is currently blocked from writing to the database.

For SQLite, this is as simple as attempting to get any sort of write lock on the database within a transaction, and retrying if it fails to do so because of a timeout error.

An implementation of this that would be specific to SQLite could be as simple as this, as any attempt to write to SQLite will result in a write lock:
```
from django.db import transaction
from django.core.exceptions import OperationalError

@contextmanager
def db_task_write_lock():
    while True:
        try:
            with transaction.atomic():
                <some operation that writes to the DB to give a write lock>
                yield
                <revert the above write operation>
                break
            except OperationalError as e:
                if "database is locked" not in str(e):
                    raise e
```

For Postgresql, we would be better off using some sort of advisory lock (potentially copying some or all of the implementation by @bjester here: https://github.com/learningequality/studio/blob/hotfixes/contentcuration/contentcuration/db/advisory_lock.py).

apurva-modi:
## Summary
Created a simple context manager to implement database write lock using atomic transactions.
-  For Sqlite created dummy operation to execute the task.
-  For Postgres implemented advisory lock.

…

## References
Fix for #7891 

## Testing checklist

## PR process


kolibri/core/device/models.py
```diff
@@ -165,3 +165,7 @@ def get_app_key(cls):
             key = app_key.key
             cache.set(APP_KEY_CACHE_KEY, key, 5000)
         return key
+
+
+class DummyModel(models.Model):
```
rtibbles: Minor thing - but might be good to call this `SQLiteLock` to give a clearer indication of its purpose.

Another thing we can do is borrow from the implementations above and give the model this method:
```
    def save(self, *args, **kwargs):
        self.pk = 1
        super(DeviceAppKey, self).save(*args, **kwargs)
```
This means that there will only ever be 1 row in the database, which suits our purposes for a single valued lock.

Before merging, we'll also have to generate a Django migration for this new model.

kolibri/core/tasks/utils.py
```diff
@@ -121,3 +122,4 @@ def shutdown(self):
 
 
 db_task_write_lock = ProcessLock("db_task_write_lock")
```
rtibbles: We should just replace the original `db_task_write_lock` here with your new `DatabaseLock`.

kolibri/core/utils/lock.py
```diff
@@ -0,0 +1,87 @@
...
+
+    def __init__(self):
+        self.lock_id = settings.TASK_LOCK_ID or 1
+        self.db_type = connection.vendor
+        self.operation = self._get_operation()
+
+    def _wait_sqlite(self):
+        while True:
+            try:
+                self.operation.execute()
```
rtibbles: Can just directly do:
```
self.operation = DummyOperation()
self.operation.execute()
```

kolibri/core/utils/lock.py
```diff
@@ -0,0 +1,87 @@
...
+                break
+            except OperationalError as e:
+                if "database is locked" not in str(e):
+                    raise e
+
+    def _wait_postgresql(self):
+        self.operation.execute()
```
rtibbles: Likewise here:
```
self.operation = PostgresLock(self.lock_id)
self.operation.execute()
```

kolibri/core/utils/lock.py
```diff
@@ -0,0 +1,87 @@
...
+
+    def _wait_not_implemented(self):
+        raise NotImplementedError(
+            "kolibri.core.utils.cache.DatabaseLock not implemented for vendor {vendor}".format(
+                vendor=connection.vendor
+            )
+        )
+
+    def __enter__(self):
+        getattr(
```
rtibbles: Can just do a vendor check here:
```
if self.db_type == "sqlite":
    self._wait_sqlite()
elif self.db_type == "postgresql":
   self._wait_postgresql()
else:
    raise NotImplementedError(            "kolibri.core.utils.cache.DatabaseLock not implemented for vendor {vendor}".format(
                vendor=connection.vendor
            ))
```
This should make the code a bit more readable, as there's less indirection to follow the code paths with getting methods by format interpolation.

kolibri/core/utils/lock.py
```diff
@@ -0,0 +1,87 @@
...
+            "_{db_type}_operation".format(db_type=self.db_type),
+            self._not_implemented_operation,
+        )()
+
+    def __init__(self):
+        self.lock_id = settings.TASK_LOCK_ID or 1
+        self.db_type = connection.vendor
+        self.operation = self._get_operation()
+
+    def _wait_sqlite(self):
```
rtibbles: For both of these we need to wrap the operations in an atomic context, to ensure that the lock is persisted throughout the database queries carried out within the context manager.

In the PostgreSQL case, this means that we would not need to unlock the advisory lock, as it is automatically released once the transaction is closed. This also gives us a better guarantee that the lock will get cleaned up, as we can rely on PostgreSQL's transaction management to do this.

In the SQLite case, it means we only have to clean up our dummy model creation in the success case, as otherwise the transaction will be rolled back and the create not applied.

kolibri/core/utils/lock.py
```diff
@@ -0,0 +1,87 @@
...
+    def __enter__(self):
+        getattr(
+            self,
+            "_wait_{db_type}".format(db_type=self.db_type),
+            self._wait_not_implemented,
+        )()
+
+    def __exit__(self):
+        if self.operation:
+            self.operation.revert()
```
rtibbles: I think you could simplify the implementation here somewhat - the additional classes and methods are causing you to require more code than necessary.

With `from django.db import transaction` as an additional import, you could then consolidate most of this into the `__enter__` function to make it a bit less indirect.

```
def __enter__(self):
    if self.db_type == "sqlite":
        while True:
            try:
                with transaction.atomic():
                       obj = DummyModel.objects.create()
                       yield
                       obj.delete()
                break
            except OperationalError as e:
                if "database is locked" not in str(e):
                    raise e
    elif self.db_type == "postgresql":
        with transaction.atomic():
            query = "SELECT pg_advisory_lock({key}) AS lock;".format(key=self.lock_id)
            with connection.cursor() as c:
                c.execute(query)
            yield
    else:
        raise NotImplementedError("kolibri.core.utils.cache.DatabaseLock not implemented for vendor {vendor}".format(
            vendor=connection.vendor
        ))
```

Could also implement this all as a function using the `contextlib.contextmanager` decorator, with the `lock_id` as the argument to the function.

The important thing here is the `yield` keyword, which allows reusing the atomic transaction context manager within the overall context.

rtibbles (commented):
We absolutely need the atomic transaction wrapping to make this work, so that's a must have.

In terms of testing, testing this is going to be a little bit tricky, so I think we should cohack on this to get this out the door!

rtibbles:
The errors are in affected code paths, so we may have something to unpick here. The strangest thing is that the test we created and tested locally was failing too.

I'm rerunning the tests just to be sure.

apurva-modi:
I run tests before pushing.

rtibbles:
Yeah, something seems to be behaving differently in CI - I'll check it out in the morning!

rtibbles (approved):

kolibri/core/utils/lock.py
```diff
@@ -0,0 +1,62 @@
...
+                    operation.revert()
+                break
+            except OperationalError as e:
+                if "database is locked" not in str(e):
+                    raise e
+    elif connection.vendor == "postgresql":
+        with transaction.atomic():
+            operation = PostgresLock(key=lock_id)
+            operation.execute()
+            yield
```
bjester, line 56: I don't see `operation.revert()` here? That seems necessary because the type of advisory lock obtained is a session level lock, which should remain locked for the duration of the connection. I would suggest changing `pg_advisory_lock` to `pg_advisory_xact_lock` ([reference](https://www.postgresql.org/docs/9.1/functions-admin.html#FUNCTIONS-ADVISORY-LOCKS)). Note that for the transaction level lock functions (`xact`), there aren't corresponding `unlock` functions because it would be cleared by COMMIT or ROLLBACK of the transaction, including if the connection is closed.

rtibbles: Ah, thanks - that's my bad for misleading @apurva-modi here. I'll submit an update.

PR #8023, opened 2021-04-22
Consolidate WSGI configuration

### LINKED ISSUE #8020: Consolidate wsgi configuration in `kolibri.deployment.default.wsgi`
## Observed behavior
We currently set a lot of additional wsgi behaviour in `server.py` which  it would be useful for uwsgi (and other wsgi services) invoked versions of Kolibri. At present, the new static files handling, and alternate origin server are not exposed in a useful way for deployment.

## Expected behavior
The default Kolibri wsgi app should include all static file serving, including content and media files.
The default Kolibri alternate origin wsgi app should include all static file serving, and content serving.

This will require some extra subclassing of our `DjangoWhiteNoise` implementation, in order to allow the addition of lazily evaluated files from content or media files.

This needs to be done, bearing in mind the Whitenoise caveats about not adding global static file lazy handling - because if we were to do that, then for any request to the server we would end up checking the disk for the existence of a static file, before serving the request.

## User-facing consequences
Complexities for deployment.

rtibbles:
## Summary
* Moves WSGI application configuration out of `server.py` and into `deployment/default` directory
* Updates `wsgi.py` in that folder to return a complete WSGI server for Kolibri that serves the Django application and any static files
* Creates `alt_wsgi.py` in that folder to return a complete WSGI server for the alternate origin server, which serves the zipcontent endpoint and all other required static files
* Updates all static file serving to be done by whitenoise
* Does this by further subclassing whitenoise to:
  * allow dynamic finding of specific files
  * caching those files once found to allow future requests to be faster
  * using Django static files machinery to discover files in multiple directories by url

## References
Fixes #8020
Depends on #7924

## Reviewer guidance
Does file seeking still work for videos and audio?
Do all static and content files load as expected?

----

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/core/urls.py
+++ b/kolibri/core/urls.py
@@ -34,19 +34,15 @@ from __future__ import absolute_import
 from __future__ import print_function
 from __future__ import unicode_literals
 
-from django.conf import settings
 from django.conf.urls import include
 from django.conf.urls import url
-from django.conf.urls.static import static
 
 from .views import GuestRedirectView
 from .views import logout_view
 from .views import RootURLRedirectView
 from .views import set_language
-from .views import static_serve_with_fallbacks
 from .views import StatusCheckView
 from .views import UnsupportedBrowserView
-from kolibri.core.content.utils import paths
 from kolibri.core.device.translation import i18n_patterns
 from kolibri.plugins.utils.urls import get_urls as plugin_urls
 
@@ -73,10 +69,3 @@ core_urlpatterns = [
 urlpatterns = [url(r"", include(core_urlpatterns, namespace="core"))]
 
 urlpatterns += plugin_urls()
-
-urlpatterns += static(settings.MEDIA_URL, document_root=settings.MEDIA_ROOT)
-
-content_dirs = [paths.get_content_dir_path()] + paths.get_content_fallback_paths()
-urlpatterns += static(
-    paths.get_content_url("/"), view=static_serve_with_fallbacks(content_dirs)
-)
--- a/kolibri/deployment/default/alt_wsgi.py
+++ b/kolibri/deployment/default/alt_wsgi.py
@@ -7,7 +7,7 @@ import os
 import kolibri.core.content
 from kolibri.core.content.utils import paths
 from kolibri.core.content.zip_wsgi import get_application
-from kolibri.utils.django_whitenoise import DjangoWhiteNoise
+from kolibri.utils.kolibri_whitenoise import DynamicWhiteNoise
 
 os.environ.setdefault(
     "DJANGO_SETTINGS_MODULE", "kolibri.deployment.default.settings.base"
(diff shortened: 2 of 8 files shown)


jredrejo (commented):
1. As a standalone kolibri app, it works fine, all the changes work perfectly and we could even get rid of nginx if just python were multithreaded ....
2. As a wsgi application, ignoring the whitenoise setup, it works fine too
3. As a wsgi application trying to use the whitenoise setup it does not work. i.e. having `module =kolibri.deployment.default.alt_wsgi:alt_application `instead of `kolibri.deployment.default.wsgi:application` does not work

Theorically that's not important as running nginx + uwsgi makes whitenoise unneeded and the current wsgi setup should be enough, but I wonder if solving this would make sense

This is the traceback when trying to use alt_application:
```
  File "./kolibri/deployment/default/alt_wsgi.py", line 33, in <module>
    alt_application = generate_alt_wsgi_application()
  File "./kolibri/deployment/default/alt_wsgi.py", line 24, in generate_alt_wsgi_application
    return DjangoWhiteNoise(
  File "./kolibri/utils/django_whitenoise.py", line 60, in __init__
    self.add_files_from_finders()
  File "./kolibri/utils/django_whitenoise.py", line 96, in add_files_from_finders
    for finder in finders.get_finders():
  File "/datos/le/mio/kolibri/venv/lib/python3.8/site-packages/django/contrib/staticfiles/finders.py", line 264, in get_finders
    yield get_finder(finder_path)
  File "/datos/le/mio/kolibri/venv/lib/python3.8/site-packages/django/contrib/staticfiles/finders.py", line 277, in get_finder
    return Finder()
  File "/datos/le/mio/kolibri/venv/lib/python3.8/site-packages/django/contrib/staticfiles/finders.py", line 129, in __init__
    app_configs = apps.get_app_configs()
  File "/datos/le/mio/kolibri/venv/lib/python3.8/site-packages/django/apps/registry.py", line 138, in get_app_configs
    self.check_apps_ready()
  File "/datos/le/mio/kolibri/venv/lib/python3.8/site-packages/django/apps/registry.py", line 125, in check_apps_ready
    raise AppRegistryNotReady("Apps aren't loaded yet.")
django.core.exceptions.AppRegistryNotReady: Apps aren't loaded yet.
unable to load app 0 (mountpoint='') (callable not found or import error)
*** no app loaded. GAME OVER ***
VACUUM: unix socket /tmp/kolibri_uwsgi.sock removed.
```

rtibbles:
Ah, thanks @jredrejo this is an important catch, as I had previously been relying on the two servers running on the same process for the Django initialization, I'll dig into this more, as I think it should be possible to do something more targeted for the alt server.

rtibbles:
@jredrejo the alt wsgi server should now serve independently! Tested by importing and running from a Python shell using waitress.

rtibbles:
Note that both the regular application and the alt_application use whitenoise, and both are now required for running Kolibri.

application is still the main Kolibri WSGI application, alt_application is the WSGI application served on a different origin (usually different post on the same LISTEN_ADDRESS) that serves zip content in a safe way.

jredrejo (approved):
It works perfectly now. Thanks!
(It seems one test is failing in Windows due to a hardcoded path join, that should be fixed)

rtibbles:
Will fix - just glad we have windows tests now!

Issue #5754, opened 2019-07-09
auto-generate consistent environment variables for options


indirectlylit:
### Observed behavior

* We allow configuration to be set in both `options.ini` and as environment variables
* Environment variables usually follow the pattern `KOLIBRI_[option_name]`
* We've had a couple recent deviations from this pattern, e.g. SENTRY and OIDC
* The main reason we allowed vars to be specified separately from options was so that we could still handle some legacy envvars

https://github.com/learningequality/kolibri/blob/3a2356883278f79b4a7655a47c7fa6db70acc184/kolibri/utils/options.py#L50-L177

### Expected behavior

* In `options.py`, rather than specifying `envvars` we should specify `envvar_aliases`
* All envvars should be generated by prepending `KOLIBRI_` to the option name
* Most existing envvar definitions can be removed

### User-facing consequences

Inconsistent environment variable definitions

### Context

0.12.5

rtibbles:
Sounds good to me.

PR #8025, opened 2021-04-23
Add KOLIBRI_ENABLE_CUSTOM_CHANNEL_NAV - default True to options.py

### LINKED ISSUE #7830: Create temporary Learn Plugin options flag using options.ini mechanisms to disable custom channel navigation by default

nucleogenesis:
## Summary

Adds an option that will be picked up from the options.ini file or by env var called `DISABLE_CUSTOM_CHANNEL_NAV`.

It is not then used anywhere as (I believe) it is intended for future debugging and/or dev use during development of the new feature.

## References

Fixes #7830 

## Reviewer guidance

[See that I did what is said in the well written and clear documentation that I ought to have read](https://kolibri-dev.readthedocs.io/en/develop/backend_architecture/plugins.html#creating-a-plugin)

### DIFF
--- a/kolibri/plugins/learn/options.py
+++ b/kolibri/plugins/learn/options.py
@@ -1,9 +1,9 @@
 option_spec = {
-    "LEARN": {
-        "DISABLE_CUSTOM_CHANNEL_NAV": {
+    "Learn": {
+        "KOLIBRI_ENABLE_CUSTOM_CHANNEL_NAV": {
             "type": "boolean",
             "default": False,
-            "envvars": ("DISABLE_CUSTOM_CHANNEL_NAV",),
+            "envvars": ("KOLIBRI_ENABLE_CUSTOM_CHANNEL_NAV",),
         },
     },
 }
--- a/kolibri/plugins/learn/options.py
+++ b/kolibri/plugins/learn/options.py
@@ -1,0 +1,9 @@
+option_spec = {
+    "LEARN": {
+        "DISABLE_CUSTOM_CHANNEL_NAV": {
+            "type": "boolean",
+            "default": False,
+            "envvars": ("DISABLE_CUSTOM_CHANNEL_NAV",),
+        },
+    },
+}
--- a/kolibri/utils/options.py
+++ b/kolibri/utils/options.py
@@ -48,6 +48,7 @@ URL_PATH_PREFIX
 LANGUAGES
 ZIP_CONTENT_HOST
 ZIP_CONTENT_PORT
+DISABLE_CUSTOM_CHANNEL_NAV
 
 [Python]
 PICKLE_PROTOCOL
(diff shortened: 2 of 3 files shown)


kolibri/utils/options.py
```diff
@@ -367,6 +368,11 @@ def path_list(value):
             "default": 8888,
             "envvars": ("KOLIBRI_ZIP_CONTENT_PORT",),
         },
+        "DISABLE_CUSTOM_CHANNEL_NAV": {
```
rtibbles: As this is a learn plugin specific option, could we add it in, as an option there, rather to the global options? 

See here: https://github.com/learningequality/kolibri/blob/release-v0.14.x/kolibri/plugins/html5_viewer/kolibri_plugin.py#L14 and here for an example of this: https://github.com/learningequality/kolibri/blob/release-v0.14.x/kolibri/plugins/html5_viewer/options.py#L45

rtibbles (commented):
Let's scope this to the learn plugin.

Would also be helpful to expose this in the plugin data in Learn for ease of use: https://github.com/learningequality/kolibri/blob/release-v0.14.x/kolibri/plugins/learn/kolibri_plugin.py#L51

kolibri/plugins/learn/options.py
```diff
@@ -0,0 +1,9 @@
+option_spec = {
+    "LEARN": {
+        "DISABLE_CUSTOM_CHANNEL_NAV": {
+            "type": "boolean",
+            "default": False,
```
rtibbles, line 5: The issue asks for this to be disabled by default, so if the flag is going to be `DISABLE` then this should default to `True`!

kolibri/plugins/learn/options.py
```diff
@@ -0,0 +1,9 @@
+option_spec = {
+    "LEARN": {
```
rtibbles: If you're editing the default, can also make this Title Case, which seems to be our unwritten convention for options section names.

kolibri/plugins/learn/options.py
```diff
@@ -0,0 +1,9 @@
...
+        "DISABLE_CUSTOM_CHANNEL_NAV": {
+            "type": "boolean",
+            "default": False,
+            "envvars": ("DISABLE_CUSTOM_CHANNEL_NAV",),
```
rtibbles: We've also tended to prefix these with `KOLIBRI_` - there's an open issue to just generate these names automatically to avoid this hullabaloo: https://github.com/learningequality/kolibri/issues/5754

If we can just prefix `KOLIBRI_` here for now, would be good.

rtibbles (changes requested):
A couple more tweaks - sorry!

kolibri/plugins/learn/options.py
nucleogenesis: Made it into an `ENABLE` name instead

rtibbles (approved):

PR #8032, opened 2021-04-27
Custom channels renderer and context param

marcellamaki:
## Summary

This PR is for a custom renderer that displays in place of the normal Kolibri Topics page when the options flag is set to be a custom channel. 

## Reviewer guidance

You can do a manual review by navigating to the custom channel created by Richard. There are also some placeholder modals to test that the events emitted inside the iframe are able to communicate out properly to the main component. I am not clear on how routing should be handled in the overlay, so what's currently in place (just a computed value change to open or closed) may need to be updated when this is combined with JB's work, if the content id for the overlay that is open should be displayed in the URL. I also am not sure how to test out if the context object is working as intended and would appreciate feedback around how I've implemented it - I'm not totally sure what I thought of is necessarily the best way.

I will create a follow-up issue for writing tests once all of the front-end changes are incorporated.

HTML5-renderer

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated


kolibri/plugins/learn/assets/src/routes/index.js
```diff
@@ -86,6 +87,18 @@ export default [
       store.commit('CORE_SET_ERROR', null);
     },
   },
+  {
+    name: PageNames.CUSTOM_TOPICS_CHANNEL,
+    path: '/topics/:channel_id',
```
rtibbles: Having duplicate paths is an issue here - on my machine, this simply continues with the old behaviour.

I think it would be best to put the conditionality inside the main root channel component, and only have a single path here - this would also mean that you don't have to duplicate the handler above.

kolibri/plugins/learn/assets/src/views/CustomChannelRenderer.vue
```diff
@@ -0,0 +1,202 @@
+<template>
+
+  <div>
+    <KContentRenderer
```
rtibbles: My suggestion here was to copy the relevant parts of the `HTML5RendererIndex.vue` implementation into here, so that you have more direct access to the iframe (also in this way, there are now two Hashi instances, one in this component, the other inside the HTML5 renderer component).

kolibri/plugins/learn/assets/src/views/CustomChannelRenderer.vue
```diff
@@ -0,0 +1,202 @@
...
+        this.getOrUpdateContext(message);
+      });
+    },
+    methods: {
+      // helper functions for fetching data from kolibri
+      // called in mainClient.js
+
+      fetchContentCollection(message) {
+        const options = message.options;
+        const getParams = pick(options, ['ids', 'page', 'pageSize', 'parent']);
```
rtibbles: Note that the backend does not currently support pagination, so should just pick `ids` and `parent` for now.

kolibri/plugins/learn/assets/src/views/CustomChannelRenderer.vue
```diff
@@ -0,0 +1,202 @@
...
+          if (contentNode) {
+            message.status = 'success';
+          } else {
+            message.status = 'failure';
+          }
+          message.data = contentNode;
+          message.type = 'response';
+          self.hashi.mediator.sendMessage({
+            nameSpace: 'hashi',
+            event: 'kolibridatareturned',
```
rtibbles: Can reference this as `this.hashi.events.KOLIBRIDATARETURNED`

kolibri/plugins/learn/assets/src/views/LearnIndex.vue
```diff
@@ -136,6 +143,9 @@
       currentPage() {
         return pageNameToComponentMap[this.pageName] || null;
       },
+      currentPageIsCustomChannel() {
+        return this.pageName == PageNames.CUSTOM_TOPICS_CHANNEL ? true : false;
```
rtibbles: Ternary is unnecessary here.

kolibri/plugins/learn/assets/src/views/LearnIndex.vue
```diff
@@ -161,6 +171,15 @@
             immersivePageIcon: 'close',
           };
         }
+        if (this.pageName === PageNames.CUSTOM_TOPICS_CHANNEL) {
```
rtibbles: Can reuse your computed prop above here (and then if we change the implementation to not be contingent on page name as I think we will need to, then you don't have to encode twice).

packages/hashi/src/mediator.js
```diff
@@ -59,13 +59,12 @@ class Mediator {
...
         if (message.message_id === msgId && message.type === 'response') {
-          this.removeMessageHandler({
-            nameSpace,
-            event: 'datareturned',
-            callback: handler,
-          });
-          if (message.status === 'success') {
-            console.log(message.data);
+          if (message.status == 'success') {
+            // this.removeMessageHandler({
```
rtibbles: Any reason for this to be commented out?

kolibri/plugins/learn/assets/src/routes/index.js
jonboiser: Yeah, I think without this change, there might be problems when refreshing or going directly to the URL for a custom channel, since its custom-ness isn't encoded in the URL anywhere.

kolibri/plugins/learn/assets/src/views/ChannelsPage.vue
```diff
@@ -41,10 +41,19 @@
     },
     methods: {
       genChannelLink(channel_id) {
-        return {
-          name: PageNames.TOPICS_CHANNEL,
-          params: { channel_id },
-        };
+        let currentChannel = this.channels.find(channel => channel.id == channel_id);
+        if (currentChannel && currentChannel.options.modality == 'CUSTOM_NAVIGATION') {
```
jonboiser: We like using strict equality, since it can root out certain types of bugs.

```suggestion
        let currentChannel = this.channels.find(channel => channel.id === channel_id);
        if (currentChannel && currentChannel.options.modality === 'CUSTOM_NAVIGATION') {
```

kolibri/plugins/learn/assets/src/views/CustomChannelRenderer.vue
jonboiser: > so that you have more direct access to the iframe

Yeah, that would be really helpful if this component provided more access to the `iframe` and the "kolibri side" of the hashi, just like in `HTML5RendererIndex`. I believe this will make it easier to access the theming set within the custom channel HTML5 and pass it to the modal holding the ContentRenderer.

jonboiser: Actually, I see below the point that there are now two hashis in play here. So now I'm wondering which hashi I would need to access to get the theme!

kolibri/plugins/learn/assets/src/views/CustomChannelRenderer.vue
```diff
@@ -0,0 +1,202 @@
...
+      return {
+        currentContent: this.topic,
+        open: false,
+      };
+    },
+    computed: {
+      ...mapState('topicsTree', ['topic']),
+      context() {
+        const urlParams = new URLSearchParams(window.location.search);
+        const fetchedEncodedContext = urlParams.has('context')
```
jonboiser: Context is showing up like this in the URLS: `context=%5Bobject%20Object%5D`. You might need to `JSON.stringify` before passing it out. Similarly will need to parse after decoding it.

rtibbles: You can also retrieve the params from `this.$route.params` rather than the global location search.

marcellamaki:
Updated screengrab after refactoring: 
updated-custom-channels-gif

rtibbles:
Looks like something got lost in the refactor, seeing this:

```
TypeError: this.encodeContext is not a function
    at VueComponent.getOrUpdateContext (CustomContentRenderer.vue:164)
    at CustomContentRenderer.vue:90
    at mediator.js:26
    at Array.forEach (<anonymous>)
    at Mediator.handleMessage (mediator.js:24)
```

kolibri/plugins/learn/assets/src/views/CustomContentRenderer.vue
```diff
@@ -0,0 +1,223 @@
...
+            event: events.KOLIBRIDATARETURNED,
+            data: message,
+          });
+        });
+      },
+      getOrUpdateContext(message) {
+        // to update context with the incoming context
+        if (message.context) {
+          message.context.customChannel = message.context.customChannel || true;
+          const encodedContext = JSON.stringify(this.encodeContext(message.context));
```
rtibbles: I think this is the wrong way around - would want to stringify the context object first, then URI encode it (can also just do both inside the `encodeContext` method.

kolibri/plugins/learn/assets/src/views/CustomContentRenderer.vue
```diff
@@ -0,0 +1,223 @@
...
+
+      fetchContentCollection(message) {
+        const options = message.options;
+        const getParams = pick(options, ['ids', 'parent']);
+        if (options.parent && options.parent == 'self') {
+          getParams.parent = this.topic.id;
+        }
+        let self = this;
+        ContentNodeResource.fetchCollection({ getParams }).then(contentNodes => {
+          contentNodes ? (message.status = 'success') : (message.status = 'failure');
```
jonboiser: Move the ternary to the right of the assignment:
```suggestion
          message.status = contentNodes ? 'success' : 'failure';
```

kolibri/plugins/learn/assets/src/views/CustomContentRenderer.vue
```diff
@@ -0,0 +1,223 @@
...
+            data: message,
+          });
+        });
+      },
+
+      fetchContentModel(message) {
+        let id = message.id;
+        let self = this;
+        ContentNodeResource.fetchModel({ id }).then(contentNode => {
+          if (contentNode) {
```
jonboiser: For consistency, let's use the same style here and in `fetchContentCollection`. (either use a one-liner or an if-else in both functions)

kolibri/plugins/learn/assets/src/views/CustomContentRenderer.vue
```diff
@@ -0,0 +1,219 @@
...
+    },
+    mounted() {
+      this.hashi = new Hashi({ iframe: this.$refs.iframe, now });
+      const zipFile = this.topic.files.find(f => f.extension === 'zip');
+      this.hashi.initialize({}, {}, zipFile.storage_url, zipFile.checksum);
+      this.hashi.on('collectionrequested', message => {
+        this.fetchContentCollection(message);
+      });
+      this.hashi.on('modelrequested', message => {
+        this.fetchContentModel(message);
```
jonboiser: To avoid having to do `let self = this` inside the functions. (Helps out with the unused vue properties rule).

```suggestion
        this.fetchContentModel(message).bind(this);
```

jonboiser (approved):
looks good!

PR #8037, opened 2021-04-28
Updates all content renderers with time-based progress calculation and pages visited/total available

### LINKED ISSUE #214: Update KContentRenderer mixin to better support custom completion criteria
* Add `timeSpent` as an optional prop for the KContentRenderer mixin
* Add `duration` as nullable prop. If prop is null, then renderer must define 'durationDefault' data item to set a default value.
* Add reusable progress by time spent computed prop to KContentRenderer mixin
  * Simplifying audio and video progress tracking by just using clock time instead of media track time
  * Allowing for unified time based progress tracking for any content type, based on `duration` or `durationDefault`
* Add prop that presents a Boolean for 'always use duration based progress tracking' - `forceTimeBasedProgress` - which reads from the contentnode options `force_time_based_progress` flag.

### LINKED ISSUE #7970: Update content renderer progress behaviour
* Update all content renderers to use mixin time spent calculation when using time spent on a resource for progress calculation
* Make PDF, EPub and Slideshows use 'all pages/slides visited' as the default progress calculation method
* Allow PDF, EPub and Slideshows to revert to time based progress tracking based on the `forceTimeBasedProgress` from the KContentRenderer mixin (see https://github.com/learningequality/kolibri-design-system/issues/214)

sairina:
## Summary

1. Updated PDF, Epub, Slideshow renderers to use "pages/slides visited out of total pages/slides" for default progress calculation.
2. Gave PDF, Epub, Slideshow renderers the option to use time-based progress-tracking (via `forceTimeBasedProgress`).
3. Updated all content renderers with `defaultDuration` and `durationBasedProgress` computed props from the `KContentRenderer` mixin.
4. Added progress calculation tests for all renderers.

## References

- Resolves #7970  and resolves learningequality/kolibri-design-system#214
- Requires this additional PR to be linked to Kolibri from KDS: learningequality/kolibri-design-system#224

### PDF, Epub, Slideshow Directions
_(for Epub and slideshows, follow the same steps, but use Epub and slideshow files instead of PDF files)._
1. Sign in as a `superadmin` and go to the learner profile (Sidenav -> Learn)
2. Click on a channel that has a PDF

**Check that progress is being correctly calculated via number of pages/slides visited out of total possible pages/slides:**
Pre-requisite:
- Open `mixin.js` in KDS
- Make sure [line 145](https://github.com/sairina/kolibri-design-system/blob/kcontentrenderer/lib/content/mixin.js#L145) reads `return this.options.force_time_based_progress || false;`
3. Open a PDF file that has not yet been opened.
4. Read through the PDF file. As you scroll through the file, make sure that:

- The blue clock icon appears within ~10-20 seconds to indicate you have started making progress. (Note that for PDFs and EPubs, you do not have to change pages for an update to be logged, but you do have to move to the next slide for an update to be logged for a slideshow.
5. Return to the channel’s index.
6. Wait another 10-20 seconds, and refresh the channel index page; make sure that the blue progress bar hasn’t moved.
7. Continue to scroll through the PDF and periodically go back to the channel’s index to see if the blue progress bar has moved according to how many pages you have viewed.
8. Once all pages/slides have been visited, the blue clock icon should change to a gold star icon

**Check that progress is being correctly calculated as time spent:**
Pre-requisite:
- Open `mixin.js` in KDS
- CHANGE [line 145](https://github.com/sairina/kolibri-design-system/blob/kcontentrenderer/lib/content/mixin.js#L145) to read `return true;`
9. Open a short (2-3 page) PDF file that has not yet been opened.
10. Read through the PDF file. As you scroll through, make sure that:
- The blue clock icon appears within ~20 seconds for the PDF; you will not have to scroll to any other pages, but you will have to for the epub and slideshow.
11. Stay on the page and wait for about 90 seconds (this should be when the 2-3 page file reaches its “target time” to indicate that the learner should have completed reading through the PDF; for the epub, it will take longer, so choose a shorter file, and for the slideshow, you’ll have to wait on the page for 5 minutes - sorry, go make yourself some tea!). Check to see that the blue clock icon has turned into a gold star icon.

### HTML5
1. Sign in as a `superadmin` and go to the learner profile (Sidenav -> Learn)
2. Click on a channel that has an HTML5app
3. Check that progress is being correctly calculated as time spent
4. Open the HTML5 app.
5. As you scroll through, make sure that:

- The blue clock icon appears within ~20 seconds.
- Stay on the page and wait for 5 minutes. 
- Check to see that the blue clock icon has turned into a gold star icon.

### Media Player

1. Sign in as a `superadmin` and go to the learner profile (Sidenav -> Learn)
2. Click on a channel that has a media file
3. Check that progress is being correctly calculated as time spent
4. Open the file.
5. As you watch or listen, make sure that:

- The blue clock icon appears within ~10 seconds.
- Stay on the page and wait until the length of the file is done (e.g. if the video is 2:05, you have to be actively watching the video for that amount of time). 
- Check to see that the blue clock icon has turned into a gold star icon.

----

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/plugins/epub_viewer/assets/src/views/EpubRendererIndex.vue
+++ b/kolibri/plugins/epub_viewer/assets/src/views/EpubRendererIndex.vue
@@ -237,6 +237,12 @@
         }
         return null;
       },
+      visitedPages() {
+        if (this.extraFields && this.extraFields.contentState) {
+          return this.extraFields.contentState.visitedPages || [];
+        }
+        return [];
+      },
       epubURL() {
         return this.defaultFile.storage_url;
       },
@@ -469,8 +475,21 @@
     methods: {
       updateProgress() {
         if (this.locations.length > 0) {
-          this.$emit('updateProgress', this.durationBasedProgress);
+          if (this.forceTimeBasedProgress) {
+            // update progress using total time user has spent on the pdf
+            this.$emit('updateProgress', this.durationBasedProgress);
+          } else {
+            // update progress using number of pages seen out of available pages
+            this.$emit('updateProgress', this.visitedPages.length / this.locations.length);
+          }
         }
+      },
+      listVisitedPages(currentLocation) {
+        let visited = this.visitedPages;
+        if (!visited.includes(currentLocation)) {
+          visited.push(currentLocation);
+        }
+        return visited;
       },
       handleReadyRendition() {
         this.updateRenditionTheme(this.themeStyle);
@@ -684,6 +703,7 @@
         this.sliderValue = location.start.percentage * 100;
         this.updateCurrentSection(location.start);
         this.currentLocation = location.start.cfi;
+        this.listVisitedPages(this.currentLocation);
         this.updateContentState();
       },
       handleSliderChanged(newSliderValue) {
@@ -699,9 +719,13 @@
           contentState = {
             ...this.extraFields.contentState,
             savedLocation: this.currentLocation || this.savedLocation,
+            visitedPages: this.visitedPages,
           };
         } else {
-          contentState = { savedLocation: this.currentLocation || this.savedLocation };
+          contentState = {
+            savedLocation: this.currentLocation || this.savedLocation,
+            visitedPages: this.visitedPages,
+          };
         }
         this.$emit('updateContentState', contentState);
       },
--- a/kolibri/plugins/epub_viewer/assets/src/views/EpubRendererIndex.vue
+++ b/kolibri/plugins/epub_viewer/assets/src/views/EpubRendererIndex.vue
@@ -333,6 +333,14 @@
         const seconds = (numberOfWords * 60) / WORDS_PER_MINUTE;
         return seconds;
       },
+      /* eslint-disable kolibri/vue-no-unused-properties */
+      /**
+       * @public
+       */
+      defaultDuration() {
+        return this.duration || this.expectedTimeToRead;
+      },
+      /* eslint-enable kolibri/vue-no-unused-properties */
     },
     watch: {
       sideBarOpen(newSideBar, oldSideBar) {
@@ -462,7 +470,15 @@
     },
     methods: {
       updateProgress() {
+        console.log('locations', this.locations.length, this);
         if (this.locations.length > 0) {
+          console.log(
+            'updating progress',
+            this.timeSpent,
+            this.summaryTimeSpent,
+            this.expectedTimeToRead,
+            this.summaryTimeSpent / this.expectedTimeToRead
+          );
           this.$emit('updateProgress', this.summaryTimeSpent / this.expectedTimeToRead);
         }
       },
--- a/kolibri/plugins/epub_viewer/assets/src/views/EpubRendererIndex.vue
+++ b/kolibri/plugins/epub_viewer/assets/src/views/EpubRendererIndex.vue
@@ -708,6 +708,7 @@
         this.updateCurrentSection(location.start);
         this.currentLocation = location.start.cfi;
         this.storeVisitedPage(this.currentLocation);
+        this.updateProgress();
         this.updateContentState();
       },
       handleSliderChanged(newSliderValue) {
--- a/kolibri/plugins/epub_viewer/assets/src/views/EpubRendererIndex.vue
+++ b/kolibri/plugins/epub_viewer/assets/src/views/EpubRendererIndex.vue
@@ -221,6 +221,7 @@
         sliderValue: 0,
         scrolled: false,
         currentLocation: null,
+        visitedPages: [],
         updateContentStateInterval: null,
       };
     },
@@ -237,11 +238,16 @@
         }
         return null;
       },
-      visitedPages() {
-        if (this.extraFields && this.extraFields.contentState) {
-          return this.extraFields.contentState.visitedPages || [];
-        }
-        return [];
+      savedVisitedPages: {
+        get() {
+          if (this.extraFields && this.extraFields.contentState) {
+            return this.extraFields.contentState.savedVisitedPages;
+          }
+          return [];
+        },
+        set(value) {
+          this.visitedPages = value;
+        },
       },
       epubURL() {
         return this.defaultFile.storage_url;
@@ -330,19 +336,16 @@
       increaseFontSizeDisabled() {
         return this.fontSize === `${FONT_SIZE_MAX}px`;
       },
-      expectedTimeToRead() {
-        const WORDS_PER_MINUTE = 300;
-        const CHARS_PER_WORD = 10;
-        const numberOfWords = (this.locations.length * LOCATIONS_INTERVAL) / CHARS_PER_WORD;
-        const seconds = (numberOfWords * 60) / WORDS_PER_MINUTE;
-        return seconds;
-      },
       /* eslint-disable kolibri/vue-no-unused-properties */
       /**
        * @public
        */
       defaultDuration() {
-        return this.expectedTimeToRead;
+        const WORDS_PER_MINUTE = 300;
+        const CHARS_PER_WORD = 10;
+        const numberOfWords = (this.locations.length * LOCATIONS_INTERVAL) / CHARS_PER_WORD;
+        const seconds = (numberOfWords * 60) / WORDS_PER_MINUTE;
+        return seconds;
       },
       /* eslint-enable kolibri/vue-no-unused-properties */
     },
@@ -475,21 +478,21 @@
     methods: {
       updateProgress() {
         if (this.locations.length > 0) {
-          if (this.forceTimeBasedProgress) {
+          if (this.forceDurationBasedProgress) {
             // update progress using total time user has spent on the epub
             this.$emit('updateProgress', this.durationBasedProgress);
           } else {
             // update progress using number of pages seen out of available pages
-            this.$emit('updateProgress', this.visitedPages.length / this.locations.length);
+            this.$emit('updateProgress', this.savedVisitedPages.length / this.locations.length);
           }
         }
       },
-      listVisitedPages(currentLocation) {
-        let visited = this.visitedPages;
+      storeVisitedPage(currentLocation) {
+        let visited = this.savedVisitedPages;
         if (!visited.includes(currentLocation)) {
           visited.push(currentLocation);
         }
-        return visited;
+        this.savedVisitedPages = visited;
       },
       handleReadyRendition() {
         this.updateRenditionTheme(this.themeStyle);
@@ -703,7 +706,7 @@
         this.sliderValue = location.start.percentage * 100;
         this.updateCurrentSection(location.start);
         this.currentLocation = location.start.cfi;
-        this.listVisitedPages(this.currentLocation);
+        this.storeVisitedPage(this.currentLocation);
         this.updateContentState();
       },
       handleSliderChanged(newSliderValue) {
@@ -719,12 +722,12 @@
           contentState = {
             ...this.extraFields.contentState,
             savedLocation: this.currentLocation || this.savedLocation,
-            visitedPages: this.visitedPages,
+            savedVisitedPages: this.visitedPages || this.savedVisitedPages,
           };
         } else {
           contentState = {
             savedLocation: this.currentLocation || this.savedLocation,
-            visitedPages: this.visitedPages,
+            savedVisitedPages: this.visitedPages || this.savedVisitedPages,
           };
         }
         this.$emit('updateContentState', contentState);
--- a/kolibri/plugins/epub_viewer/assets/src/views/EpubRendererIndex.vue
+++ b/kolibri/plugins/epub_viewer/assets/src/views/EpubRendererIndex.vue
@@ -151,7 +151,6 @@
   import clamp from 'lodash/clamp';
   import Lockr from 'lockr';
   import FocusLock from 'vue-focus-lock';
-  import { mapGetters } from 'vuex';
   import CoreFullscreen from 'kolibri.coreVue.components.CoreFullscreen';
   import responsiveElementMixin from 'kolibri.coreVue.mixins.responsiveElementMixin';
   import responsiveWindowMixin from 'kolibri.coreVue.mixins.responsiveWindowMixin';
@@ -226,7 +225,6 @@
       };
     },
     computed: {
-      ...mapGetters(['summaryTimeSpent']),
       isAtStart() {
         return get(this.rendition, 'location.atStart', false);
       },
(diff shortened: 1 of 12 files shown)


kolibri/plugins/slideshow_viewer/assets/src/views/SlideshowRendererComponent.vue
```diff
@@ -121,6 +121,15 @@
       contentHeight: function() {
         return window.innerHeight * 0.7 + 'px';
       },
+      /* eslint-disable kolibri/vue-no-unused-properties */
+      /**
+       * @public
+       * Note: the default duration historically for slidshows has been 5 min
+       */
+      defaultDuration() {
+        return this.duration || 300;
```
rtibbles: I was assuming that `defaultDuration` could just return a single value, and then the mixin can worry about conditionalizing between `duration` and the `defaultDuration`.

kolibri/plugins/pdf_viewer/assets/src/views/PdfRendererIndex.vue
```diff
@@ -197,17 +200,14 @@
       this.currentLocation = this.savedLocation;
       this.showControls = true; // Ensures it shows on load even if we're scrolled
       const loadPdfPromise = PDFJSLib.getDocument(this.defaultFile.storage_url);
-
```
rtibbles, line 200: Looks like your IDE might be applying more stringent linting standards than ours - blank lines in code can be good for readability, so would be better not to remove them.

kolibri/plugins/learn/assets/src/views/ContentPage.vue
```diff
@@ -335,9 +335,9 @@
       markAsComplete() {
         this.wasIncomplete = false;
       },
-      genContentLink(id, isLeaf) {
+      genContentLink(id, kind) {
         return {
-          name: isLeaf ? PageNames.TOPICS_CONTENT : PageNames.TOPICS_TOPIC,
```
jonboiser: The `isLeaf` field is something we just added to the ContentNode API to avoid this `kind === TOPIC` check. Was it not working?

kolibri/plugins/media_player/assets/src/views/MediaPlayerIndex.vue
```diff
@@ -207,6 +215,7 @@
     },
     beforeDestroy() {
       clearInterval(this.updateContentStateInterval);
+      this.updateProgress();
```
rtibbles: Flagging this issue: https://github.com/learningequality/kolibri/issues/2355

I am not sure what the appropriate fix is for it, tbh, but imagine it would have something to do with a navigation guard in the learn ContentPage.

It might require public methods on the renderer components to do any cleanup.

rtibbles: (this is not a request to fix this, just flagging that this is doubling down on something that may be an issue!)

kolibri/plugins/slideshow_viewer/assets/src/views/SlideshowRendererComponent.vue
sairina: 🤦🏽‍♀️ You're right. The mixin already does that! Thanks for catching.

kolibri/plugins/media_player/assets/src/views/MediaPlayerIndex.vue
```diff
@@ -188,6 +188,14 @@
       transcriptVisible() {
         return this.transcript && !this.loading && this.captionTracks.length > 0;
       },
+      /* eslint-disable kolibri/vue-no-unused-properties */
+      /**
+       * @public
+       */
+      defaultDuration() {
+        return this.duration || this.player.duration();
```
bjester: Could `this.player` not be initialized by the time this is executed? The player is instantiated on `mount` and then `$nextTick`

kolibri/plugins/learn/assets/src/views/ContentPage.vue
sairina: I must have had a merge conflict and committed an older version of the code. I'll make sure it's fixed before the PR is finalized. Thanks for catching this!

kolibri/plugins/epub_viewer/assets/src/views/EpubRendererIndex.vue
```diff
@@ -333,6 +337,14 @@
         const seconds = (numberOfWords * 60) / WORDS_PER_MINUTE;
         return seconds;
       },
+      /* eslint-disable kolibri/vue-no-unused-properties */
+      /**
+       * @public
+       */
+      defaultDuration() {
+        return this.expectedTimeToRead;
```
rtibbles: Is `expectedTimeToRead` being used anywhere else now? Could just move its implementation into here if not.

kolibri/plugins/epub_viewer/assets/src/views/EpubRendererIndex.vue
```diff
@@ -463,9 +475,22 @@
...
+          if (this.forceTimeBasedProgress) {
+            // update progress using total time user has spent on the epub
+            this.$emit('updateProgress', this.durationBasedProgress);
+          } else {
+            // update progress using number of pages seen out of available pages
+            this.$emit('updateProgress', this.visitedPages.length / this.locations.length);
+          }
         }
       },
+      listVisitedPages(currentLocation) {
```
rtibbles: This method name doesn't quite match up with how you are using it - maybe something like `storeVisitedPage`? Unless I'm missing the point, but its primary purpose seems to be recording that a page has been visited.

kolibri/plugins/epub_viewer/assets/src/views/EpubRendererIndex.vue
```diff
@@ -678,6 +703,7 @@
         this.sliderValue = location.start.percentage * 100;
         this.updateCurrentSection(location.start);
         this.currentLocation = location.start.cfi;
+        this.listVisitedPages(this.currentLocation);
```
rtibbles: Hrm, the `listVisitedPages` method updates the returned value of the computed `this.visitedPages` property - this is different to setting `this.currentLocation` which is data, and so is local state that is being updated.

This means that when you then call `this.updateContentState()` the updates you have made to `this.visitedPages` will only be present because Vue has cached the output of the `visitedPages` computed property that you have then modified. If the computed property were to be recomputed for some reason, then your modifications would be lost.

Two possible solutions:

Do what currentLocation and savedLocation do - have one in data, and one that is a computed prop derived from extraFields.contentState.

Pass the `currentLocation` to `updatedContentState` and just update `visitedPages` on the `contentState`, and then rely on that being set into the `extraFields` prop from outside the component, to then update the computed prop `visitedPages`

kolibri/plugins/epub_viewer/assets/src/views/EpubRendererIndex.vue
```diff
@@ -239,6 +237,12 @@
         }
         return null;
       },
+      visitedPages() {
+        if (this.extraFields && this.extraFields.contentState) {
+          return this.extraFields.contentState.visitedPages || [];
+        }
+        return [];
```
rtibbles: May be a premature optimization, but we could implement this as an object, rather than an array.

This would mean when we set a new page on the visitedPages we'd just do:

`visitedPages[page] = true`

This means we don't need to do an O(n) lookup on an Array every time we check to see if the page is already in vistedPages.

kolibri/plugins/epub_viewer/assets/src/views/EpubRendererIndex.vue
```diff
@@ -463,9 +475,22 @@
     methods: {
       updateProgress() {
         if (this.locations.length > 0) {
-          this.$emit('updateProgress', this.summaryTimeSpent / this.expectedTimeToRead);
+          if (this.forceTimeBasedProgress) {
+            // update progress using total time user has spent on the epub
+            this.$emit('updateProgress', this.durationBasedProgress);
+          } else {
+            // update progress using number of pages seen out of available pages
+            this.$emit('updateProgress', this.visitedPages.length / this.locations.length);
```
rtibbles: If you did use my suggestion for visitedPages to be an object, you'd change this to `Object.keys(this.visitedPages).length`

kolibri/plugins/media_player/assets/src/views/MediaPlayerIndex.vue
```diff
@@ -379,8 +374,9 @@
       handleSeek() {
         // record progress before updating the times,
         // to capture any progress that happened pre-seeking
-        this.recordProgress();
-
+        if (!this.forceTimeBasedProgress) {
```
rtibbles: I don't think we need these `forceTimeBasedProgress` checks in this file - as we are only doing time based progress here.

kolibri/plugins/media_player/assets/src/views/MediaPlayerIndex.vue
```diff
@@ -421,11 +419,14 @@
         );
         this.progressStartingPoint = this.dummyTime;
       },
+      // An alternative to recordProgress, this updates tracking based on clock-time spent on media
+      updateProgress() {
```
rtibbles: Let's revert this for now, and just stick with recordProgress, as messing with this all seemed a bit more complex than we anticipated.

kolibri/plugins/pdf_viewer/assets/src/views/PdfRendererIndex.vue
```diff
@@ -151,9 +146,23 @@
         }
         return 0;
       },
+      visitedPages() {
```
rtibbles: My comments for the epub renderer also apply here!

kolibri/plugins/slideshow_viewer/assets/src/views/SlideshowRendererComponent.vue
```diff
@@ -106,6 +106,12 @@
       currentSlide() {
         return this.slides[this.currentSlideIndex];
       },
+      visitedSlides() {
```
rtibbles: visitedPages comments apply here too.

rtibbles (changes requested):
Main blocker here is modification of computed prop value in `visitedPages` and `visitedSlides` - the rest is just a bit of cleanup (or possibly my misapprehension).

kolibri/plugins/media_player/assets/src/views/MediaPlayerIndex.vue
sairina: Hmm, does this mean that media player won't be using `defaultDuration`, since it doesn't use `forceDurationBasedProgress`, and thus `updateProgress()`?

Essentially, just go back to what it originally was?

kolibri/plugins/epub_viewer/assets/src/views/EpubRendererIndex.vue
```diff
@@ -678,6 +707,7 @@
         this.sliderValue = location.start.percentage * 100;
         this.updateCurrentSection(location.start);
         this.currentLocation = location.start.cfi;
+        this.storeVisitedPage(this.currentLocation);
```
rtibbles, line 710: I feel like we could just call `updateProgress` here too. At the moment, it's not getting updated except on the 30 second timer, whereas if someone is quickly flicking through the pages, that won't be catching it as fast as this would.

kolibri/plugins/pdf_viewer/assets/src/views/PdfRendererIndex.vue
```diff
@@ -151,9 +147,28 @@
...
+      },
       fullscreenText() {
         return this.isInFullscreen ? this.$tr('exitFullscreen') : this.$tr('enterFullscreen');
       },
+      /* eslint-disable kolibri/vue-no-unused-properties */
+      /**
+       * @public
+       */
+      defaultDuration() {
+        return this.targetTime;
```
rtibbles: Is `targetTime` used anywhere else? Like in the EPub renderer we may be able to just rename it to `defaultDuration`.

kolibri/plugins/pdf_viewer/assets/src/views/PdfRendererIndex.vue
```diff
@@ -242,23 +254,22 @@
         }
         this.$emit('startTracking');
         this.updateContentStateInterval = setInterval(this.updateProgress, 30000);
-        // Automatically master after the targetTime, convert seconds -> milliseconds
-        this.timeout = setTimeout(this.updateProgress, this.targetTime * 1000);
+        if (this.forceDurationBasedProgress) {
```
rtibbles: I wonder if we even need this any more - we already have the polling for update progress above.

kolibri/plugins/pdf_viewer/assets/src/views/PdfRendererIndex.vue
```diff
@@ -305,6 +320,10 @@
         } else {
           // TODO: there is a miscalculation that causes a wrong position change on scale
           this.savePosition(this.calculatePosition());
+
+          // determine how many pages user has viewed/visited
+          let currentPage = parseInt(this.currentLocation * this.totalPages) + 1;
+          this.storeVisitedPage(currentPage);
```
rtibbles, line 319: Can also updateProgress here as well, similarly to the EPub Renderer.

kolibri/plugins/slideshow_viewer/assets/src/views/SlideshowRendererComponent.vue
```diff
@@ -201,10 +222,17 @@
         if (this.currentSlideIndex > this.highestViewedSlideIndex) {
           this.highestViewedSlideIndex = this.currentSlideIndex;
         }
+        this.storeVisitedSlide(this.currentSlideIndex);
+        this.updateContentState();
```
rtibbles, line 227: Let's also updateProgress here too.

rtibbles (changes requested):
A couple more tweaks for updateProgress!

kolibri/plugins/pdf_viewer/assets/src/views/PdfRendererIndex.vue
```diff
@@ -256,7 +253,7 @@
         this.updateContentStateInterval = setInterval(this.updateProgress, 30000);
         if (this.forceDurationBasedProgress) {
           // Automatically master after the targetTime, convert seconds -> milliseconds
-          this.timeout = setTimeout(this.updateProgress, this.targetTime * 1000);
+          // this.timeout = setTimeout(this.updateProgress, this.targetTime * 1000);
```
rtibbles: Can just delete the whole `if` statement rather than comment it out.

rtibbles (approved):

PR #8044, opened 2021-05-03
Cache user between requests as well as on requests.

rtibbles:
## Summary
All of our authenticated API requests use `request.user` for either permissions checking, or customizing the response content, or both.
We currently ensure that the user is only requested from the database once within the request cycle in the way that we handle it in our custom authentication/session middleware.
This PR adds caching by session key for a user, so that we maintain a representation of the user object keyed by their session, so multiple requests from the same session will use the cached user representation.

## Reviewer guidance
* Do authenticated requests have 1 less database query after the first time?
* Is the user being properly cached?

See here for where I have taken the session key lookup from: https://docs.djangoproject.com/en/1.11/_modules/django/contrib/auth/#get_user

----

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/core/auth/middleware.py
+++ b/kolibri/core/auth/middleware.py
@@ -43,12 +43,15 @@ def get_anonymous_user_model():
 
 def _get_user(request):
     if not hasattr(request, "_cached_user"):
-        user_id = _get_user_session_key(request)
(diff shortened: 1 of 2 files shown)


jredrejo (approved):
LGTM

PR #8049, opened 2021-05-03
Remove mptt from collections.

### LINKED ISSUE #7962: Remove use of MPTT in Collection hierarchy
## Observed behavior
Having to write MPTT fields has caused issues in the past including:
* Slow writes when creating AdHoc groups for large collection hierarchies
* Corrupted MPTT trees when doing data synchronization

Once https://github.com/learningequality/kolibri/pull/7928 is merged, we will no longer need MPTT fields to describe our collection hierarchy, as all role permissions are explicitly granted, and the collection hierarchy made concrete.

## Expected behavior
We should maintain a `parent` `ForeignKey` field in order to continue to encode the parent relationships between collections, but can drop all the other MPTT specific fields from the model. Simplest solution would be to change the base class from MPTTModel to a Django Model.

We should also ensure that we sort collections by name in the frontend to ensure that we have a consistent ordering, now that it will no longer be ordered by 'lft' (which roughly correlates to the order of creation).

## User-facing consequences
More scalable data.

rtibbles:
## Summary
* Removes the use of MPTT from our Collection hierarchy, deferring instead to parent foreign keys only

## References
Fixes #7962 

## Reviewer guidance
This is an irreversible change - I made some effort to ensure that running the migration backwards would actually work (without the custom changes I made, it would throw an integrity error as it would not set a value for the MPTT fields). However, it would still require a manual step of:
```
from kolibri.core.auth.models import Collection
Collection.objects.rebuild()
```
After the backwards migration to have things working again.

To verify that this change was worth it, I ran the following benchmark on develop vs this branch:
```
import time
from kolibri.core.auth.models import Classroom
from kolibri.core.auth.models import LearnerGroup
from kolibri.core.auth.models import AdHocGroup

f = Facility.objects.create(name="test")
classes = []
for j in range(0, 200):
    c = Classroom.objects.create(name=str(j), parent=f)
    classes.append(c)
    for k in range(0, 10):
        lg = LearnerGroup.objects.create(name=str(k), parent=c)

total_time = 0
for loop in range(0, 10):
    start = time.time()
    for m in range(0, 200):
        AdHocGroup.objects.create(name=str(m), parent=classes[0])

    total_time += (time.time() - start)

print("Adhoc groups in first created class", total_time / 10)


total_time = 0
for loop in range(0, 10):
    start = time.time()
    for m in range(0, 200):
        AdHocGroup.objects.create(name=str(m), parent=classes[-1])

    total_time += (time.time() - start)


print("Adhoc groups in last created class", total_time / 10)

total_time = 0
for loop in range(0, 10):
    start = time.time()
    for m in range(0, 200):
        AdHocGroup.objects.create(name=str(m), parent=classes[m])

    total_time += (time.time() - start)


print("Adhoc groups evenly distributed across all classes", total_time / 10)

f.delete()
```
The results were as follows:
```
Develop
Adhoc groups in first created class 1.6902011394500733
Adhoc groups in last created class 1.0079774379730224
Adhoc groups evenly distributed across all classes 2.476174759864807
```

```
No MPTT Collections
Adhoc groups in first created class 0.25468289852142334
Adhoc groups in last created class 0.2554821729660034
Adhoc groups evenly distributed across all classes 0.24961459636688232
```

In the worst case for develop (and the one that seems most ecologically plausible), this represents a 10x speedup for creation of new collections. In the best case for develop it is a 4x speedup. The other advantage is that this now represents a fixed time operation, where the size of the collection hierarchy no longer affects the speed of creation or deletion.

----

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated


jredrejo (approved):
Deeply tested , including sync with an older version of kolibri.
Everything works fine, no issues found, and code LGTM

rtibbles:
Thanks for the thorough testing (especially the older version sync), much appreciated!

PR #8051, opened 2021-05-04
On hover, add light-grey background to ItemProgressDisplay

### LINKED ISSUE #7662: On Coach > Class Home page, Lessons/Quizzes progress panels have no visual change on focus or hover
### Observed behavior

Screenshot from https://kolibri-beta.learningequality.org/en/coach/#/17f732bf12eca15a0cf6d846c3a1c8c4/home

image

When you hover over these items with a mouse or focus on them with a keyboard, there is no visual difference. (A related issue is #7651 where links in general don't get any visual indicator of keyboard focus).

### Expected behavior

On mouse-hover or keyboard-focus, there is some visual change to these items on the Coach > Class Home page.

**The actual visual change might need some decisions**, because the strategy used for flat, white buttons of changing the background from white to light grey will wash out other gray elements, like the progress bar and other text. 

I can't find the exact issue, but this is something that happened on the Individual Learner selector for assignments---with the wrong grey background, the grey text and disabled checkboxes would get completely obscured.

### User-facing consequences

The issue is less pronounced when using a mouse or touchpad, but for keyboard-only navigation, it's very difficult to know where you are without some visual assistance.

### Errors and logs

…

### Steps to reproduce

1. Go to a Coach > Class Home Page
1. mouse-hover or keyboard-focus one of these Quizzes/Lessons progress cards
1. The items do not change in appearance.

### Context

…

jonboiser:
## Summary

1. In the Coach Homepage, adds a light-grey background when hovering on `ItemProgressDisplay`.

This does not address #7974, although the `a` tags in this component are affected by this issue as well.

CleanShot 2021-05-03 at 14 19 53

## References

Fixes #7662

## Reviewer guidance

1. Does it work in all browsers?
2. Are my choices for padding good?
2. Is my padding/margin hack ok?

----

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated


radinamatic:
Do you plan to continue working on this...? 
The proposed fix just partially solves the issues you reported in #7662, as the bigger accessibility concern is that visual focus outline is still missing when navigating by keyboard.

jonboiser:
These links are just one of many components affected by https://github.com/learningequality/kolibri/issues/7974. The hope is that after this issue is resolved, they will get the outline back. Right now, what you can do is tab-focus the link, then hover over it with your mouse and it will get an outline, just like I described in #7974.

radinamatic:
Ok, I see it now, seems the case of crossed modalities... So the hope is that hover of this element will be indicated just by the background color change, and focus state would get just the green outline once #7974 is resolved?

nucleogenesis (commented):
I think that this is a weird looking way of adding a hover response. The left & right padding between the hovered container and the text inside is pretty small at 8px.

Personally, I'd favor a full-width and full-height hover background but would like @jtamiace or @khangmach 's thoughts in general on this fix.

https://user-images.githubusercontent.com/6356129/117373790-cc34ee80-ae80-11eb-8d1b-e5f6814aed70.mp4

---

Separately, with the solution you give I think we should increase the margin between the hovered items and the breaklines between them such that the hovered effect doesn't touch the line. It feels odd that the top item doesn't touch anything left & right & top - but it's flush on bottom then the middle item is flush on both sides then the bottom has a lot of whitespace beneath. I think adding a little whitespace between the breakline and the item could improve the visual separation of the elements as the hover styles go into effect.

jonboiser:
@nucleogenesis I added some extra padding around the links, so that the background doesn't touch the border.

CleanShot 2021-05-06 at 17 42 17

> Personally, I'd favor a full-width and full-height hover background but would like @jtamiace or @khangmach 's thoughts in general on this fix.

The problem I have with this is that I want the background to match the click target. If I make the background full-width, there will be a zone around the link that highlights when hovered, but isn't clickable. 

(There's actually a small zone where the opposite happens: it's clickable but not highlighted bc of the negative margins added by KGrid).

I think we could do this properly, but that would require negating the padding from `Block` + `BlockItem` so that `ItemProgressDisplay` (the link) could span the entire width of `Block`

nucleogenesis (approved):
That looks good to me! Totally get the issue w/ the hover target not matching the click target.

PR #8052, opened 2021-05-05
Python API

### LINKED ISSUE #6197: Make CLI commands accessible as pure Python API

rtibbles:
## Summary
* Consolidates non-CLI code in `kolibri.utils.main`
* Exposes core CLI methods as a Python API
* Adds a restart CLI function and API method

## References
Fixes #6197 

## Reviewer guidance
Test:
```
from kolibri.main import initialize
from kolibri.main import start
from kolibri.main import restart
from kolibri.main import enable_plugin
from kolibri.main import disable_plugin
```

Should we add anything else?
Is this the right place to expose these?

----

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/utils/cli.py
+++ b/kolibri/utils/cli.py
@@ -331,6 +331,18 @@ def services(port, background):
     server.start(port=port, zip_port=None, serve_http=False, background=background)
 
 
+@main.command(cls=KolibriCommand, help="Restart the Kolibri process")
+def restart():
+    """
+    Restarts the server if it is running
+    """
+    if server.restart():
+        logger.info("Kolibri has successfully restarted")
+        sys.exit(0)
+    logger.info("Kolibri has failed to restart - confirm that the server is running")
+    sys.exit(1)
+
+
 @main.command(
     cls=KolibriDjangoCommand,
     context_settings=dict(ignore_unknown_options=True, allow_extra_args=True),
--- a/kolibri/utils/cli.py
+++ b/kolibri/utils/cli.py
@@ -1,52 +1,26 @@
-"""
-Please make note ``kolibri.utils.cli`` has  a sensitive module import order.
-
-This deliberately imports ``kolibri.__init__`` before loading additional
-``kolibri.*`` modules.
-
-TODO: Investigate and challenge/explain this load order.
-"""
 from __future__ import absolute_import
 from __future__ import print_function
 from __future__ import unicode_literals
 
 import importlib
 import logging
-import os
 import signal
 import sys
-from sqlite3 import DatabaseError as SQLite3DatabaseError
 
 import click
-import django
-from django.core.management import call_command
 from django.core.management import execute_from_command_line
-from django.core.management.base import handle_default_options
-from django.db.utils import DatabaseError
 
 import kolibri
-from . import sanity_checks
-from . import server
-from .debian_check import check_debian_user
-from kolibri.core.deviceadmin.utils import IncompatibleDatabase
-from kolibri.core.upgrade import matches_version
-from kolibri.core.upgrade import run_upgrades
 from kolibri.plugins import config
 from kolibri.plugins import DEFAULT_PLUGINS
-from kolibri.plugins.utils import autoremove_unavailable_plugins
-from kolibri.plugins.utils import check_plugin_config_file_location
 from kolibri.plugins.utils import disable_plugin
-from kolibri.plugins.utils import enable_new_default_plugins
 from kolibri.plugins.utils import enable_plugin
 from kolibri.plugins.utils import iterate_plugins
-from kolibri.plugins.utils import run_plugin_updates
-from kolibri.utils.conf import KOLIBRI_HOME
-from kolibri.utils.conf import LOG_ROOT
+from kolibri.utils import server
 from kolibri.utils.conf import OPTIONS
-from kolibri.utils.logger import get_base_logging_config
-from kolibri.utils.sanity_checks import check_django_stack_ready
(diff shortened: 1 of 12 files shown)


kolibri/utils/cli.py
```diff
@@ -450,13 +225,13 @@ def main(ctx):
 @main.command(cls=KolibriDjangoCommand, help="Start the Kolibri process")
 @click.option(
     "--port",
-    default=OPTIONS["Deployment"]["HTTP_PORT"],
+    default=None,
```
jredrejo, line 228: Should we update https://kolibri.readthedocs.io/en/latest/access/index.html to comment the possibility of port autodiscovery using ---port=0 ?

kolibri/utils/main.py
```diff
@@ -0,0 +1,265 @@
...
+
+    except (DatabaseError, SQLite3DatabaseError) as e:
+        if "malformed" in str(e):
+            logger.error(
+                "Your database appears to be corrupted. If you encounter this,"
+                "please immediately back up all files in the .kolibri folder that"
+                "end in .sqlite3, .sqlite3-shm, .sqlite3-wal, or .log and then"
+                "contact Learning Equality. Thank you!"
+            )
+        raise
```
jredrejo, line 149: I love this strategy :)

kolibri/utils/main.py
```diff
@@ -0,0 +1,265 @@
...
+                "Tried to check that the database was accessible "
+                "and an error occurred: {}".format(e)
+            )
+            raise
+
+
+def update(old_version, new_version):
+    """
+    Called whenever a version change in kolibri is detected
+    """
```
jredrejo, line 237: In the previous version of this function, the comment had also this text:

> TODO: We should look at version numbers of external plugins, too!

is there a reason to remove it now? I don't see any differences affecting the plugins

kolibri/utils/server.py
```diff
@@ -580,13 +609,26 @@ def start(port=0, zip_port=0, serve_http=True, background=False):
         autoreloader = Autoreloader(bus)
         autoreloader.subscribe()
 
+    reload_plugin = RestartPlugin(bus)
```
jredrejo, line 612: I find `RestartPlugin` a confusing name. Reading plugins in the kolibri context does not bring magicbus to my mind at first

kolibri/utils/cli.py
```diff
@@ -469,6 +244,10 @@ def start(port, zip_port, background):
     """
     Start the server on given port.
     """
+    port = OPTIONS["Deployment"]["HTTP_PORT"] if port is None else port
+    zip_port = (
+        OPTIONS["Deployment"]["ZIP_CONTENT_PORT"] if zip_port is None else zip_port
```
jredrejo, line 249: As `OPTIONS["Deployment"]["ZIP_CONTENT_PORT"] `is 0 by default, the port number is not known if it's not set by the user. I think we should add this information to https://kolibri.readthedocs.io/en/latest/access/index.html so system administrators know they need to open this port in the firewall (thinking specially of the annoying messages coming from Windows on different ports everytime kolibri starts)

jredrejo (commented):
Overall it looks good, most of the changes are just code relocated to a more logical place or where it makes kolibri start logic better. The few important changes are fine for my  pov and after testing it I have found no issues.
Just a few comments, specially related to documentation for our users and our code.

kolibri/utils/cli.py
rtibbles: I think that would be good, yes!

kolibri/utils/main.py
rtibbles: No, except that it was addressed a long time ago, and I only just noticed that it is still here!

kolibri/utils/server.py
rtibbles: Can update - but there are lots of classes called Plugin in this file!

kolibri/utils/cli.py
rtibbles: Yes, definitely - I'll add a user docs label to the PR.

kolibri/utils/server.py
jredrejo: yes, and I feel the same in all the cases, but maybe it's too late to refactor all this stuff. They are actual plugins, even if they are not kolibri plugins. It's just made me think twice what I was checking becuase I'm quite biased on this field.

jredrejo (approved):

PR #8026, opened 2021-04-24
Remove cherrypy dependency

rtibbles:
## Summary
* Removes cherrypy as a dependency
* Replaces it with explicit use of cheroot and magicbus
* Cleans up unused dependencies
* Passes port to zeroconf listener registration once main server has completed serving
* Has fallback for services command where it will pass in the set port once the process has started running
* Consolidates all server start code in server.py
* Uses pid file to track server status using ProcessBus life cycle hooks
* Use pid file throughout start cycle rather than separate startup lock
* Simplifies the stop procedure
* Sets default zip port to 0 to allow it to be dynamically chosen
* Creates a port cache to record previously available ports - port discovery slows startup time, so this helps to mitigate this
* Directly connect server logging to kolibri logging

## References
Depends on #8023 
Fixes #8024
Supersedes #7714
This also contributes to fixing https://github.com/learningequality/kolibri-app/issues/74 in a most general way, by allowing apps to just specify port 0, and allowing a free port to be chosen for use.

## Reviewer guidance
Kolibri should run and serve as intended.
Appropriate logging happens for all events in the server life cycle.
Start Kolibri with `--port=0` and verify that the port bound is the one reported, and that zeroconf registers that port

----

## Testing checklist

- [x] Contributor has fully tested the PR manually
- [ ] If there are any front-end changes, before/after screenshots are included
- [x] Critical user journeys are covered by Gherkin stories
- [x] Critical and brittle code paths are covered by unit tests

## PR process

- [x] PR has the correct target branch and milestone
- [x] PR has 'needs review' or 'work-in-progress' label
- [x] If PR is ready for review, a reviewer has been added. (Don't use 'Assignees')
- [ ] If this is an important user-facing change, PR or related issue has a 'changelog' label
- [ ] If this includes an internal dependency change, a link to the diff is provided

## Reviewer checklist

- Automated test coverage is satisfactory
- PR is fully functional
- PR has been tested for [accessibility regressions](http://kolibri-dev.readthedocs.io/en/develop/manual_testing.html#accessibility-a11y-testing)
- External dependency files were updated if necessary (`yarn` and `pip`)
- Documentation is updated
- Contributor is in AUTHORS.md

PR #8057, opened 2021-05-06
Lock used ports to prevent their being picked up by the port cache.

### LINKED ISSUE #8056: kolibri.utils.server.PortCache can read $KOLIBRI_HOME/port_cache file in such a way as to block server startup
## Observed behavior

I tried to start up my devserver and kept getting errors that 8000 wasn't available. It only happened on a specific KOLIBRI_HOME but not on another.

I was digging around and noticed that the `$KOLIBIR_HOME/port_cache` file in the failing one [(gist with cat of options and port_cache files)](https://gist.github.com/nucleogenesis/0e267ece5ffb02c99619675132b954f2) had the values:

```
8000
XXXXX
```

But the succeeding one [(see gist of whole run w/ cat of options.ini and port_cache files)](https://gist.github.com/nucleogenesis/6f3d1f173972a2667624bd6212ae4c7c):

```
XXXXX
8000
```

## Expected behavior

PortCache should load/save/read/etc the port cache without blocking server startup.

## Steps to reproduce

In your $KOLIBRI_HOME directory, make sure the number `8000` is first in the `$KOLIBRI_HOME/port_cache` file and try running `yarn devserver`.

## Context

`develop`
Debian 10

rtibbles:
## Summary
* The port caching machinery introduced in #8026 had a significant bug where it would not take into account ports that were already occupied by currently running servers.
* This fixes that by pre-emptively registering any ports being used for servers with the port cache before WSGI servers are initialized.

## References
Fixes #8056

…

## Reviewer guidance
Set `8000` as the first value of the `port_cache` file in KOLIBRI_HOME dir.
Run `yarn run python-devserver` and confirm no errors occur

----

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/utils/server.py
+++ b/kolibri/utils/server.py
@@ -588,12 +588,13 @@ def _read_pid_file(filename):
     Reads a pid file and returns the contents. PID files have 1 or 2 lines;
      - first line is always the pid
      - second line is the port the server is listening on.
-     - third line is the status of the server process
+     - third line is the port the alternate origin server is listening on
+     - fourth line is the status of the server process
 
     :param filename: Path of PID to read
-    :return: (pid, port, status): with the PID in the file and the port number
-                          if it exists. If the port number doesn't exist, then
-                          port is None.
+    :return: (pid, port, zip_port, status): with the PID in the file, the port numbers
+                          if they exist. If the port number doesn't exist, then
+                          port is None. Lastly, the status code is returned.
     """
     if not os.path.isfile(PID_FILE):
         return None, None, None, STATUS_STOPPED
--- a/kolibri/utils/server.py
+++ b/kolibri/utils/server.py
@@ -128,7 +128,12 @@ def check_port_availability(host, port):
 class PortCache:
     def __init__(self):
         self.values = {}
+        self.occupied_ports = set()
         self.load()
+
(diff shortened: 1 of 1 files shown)


nucleogenesis (changes requested):
1) The `_read_pid_file` needs its docstring updated to note the `zip_port` is part of it's return value.

2) The server.pid file seems to be getting written wrong  - or not deleted during shut down maybe for some reason. It seems that if it doesn't delete the PID file so it thinks Kolibri is still running even though it very much isn't [gist with devserver output, ps aux to prove no kolibris are running](https://gist.github.com/nucleogenesis/97d967d39373dc6ab2de3d9778654fad)

This is the way the PID file kinda looks:

```
8973
8080
0
10
```

Noting that 10 is the status code for "Shutting Down"

nucleogenesis (approved):
@rtibbles 

Ignore the above - after `make clean`-ing my pyc files I cannot replicate this despite my best efforts. Always the server.pid file is killed between runs.

Oh - but don't ignore the bit about the docstring if you wouldn't mind quickly updating that before merging.

rtibbles:
Updated the _read_pid_file docstring.

PR #8070, opened 2021-05-10
MVP Implementation of QTI Multiple Choice and Text Entry

### LINKED ISSUE #13490: Complete a prototype for parsing a QTI Zip file and rendering a multiple choice question
❌ **This issue is not open for contribution. Visit <a href="https://learningequality.org/contributing-to-our-open-code-base/" target="_blank">Contributing guidelines</a>** to learn about the contributing process and how to find suitable issues.

## Requirements

- [x] extract the IMSCP manifest and parse
- [x] render the assessment from the manifest by the itemId prop
- [x] render a multiple choice question

### LINKED ISSUE #13491: Render simple free response QTI questions
❌ **This issue is not open for contribution. Visit <a href="https://learningequality.org/contributing-to-our-open-code-base/" target="_blank">Contributing guidelines</a>** to learn about the contributing process and how to find suitable issues.

Render the `<qti-text-entry-interaction>` element, inline.

## Requirements

A `TextEntryInteraction` component with the following props (with the equivalent from Pydantic in the comment):
```
    base: NumberProp(false), // Optional[int] = None
    stringIdentifier: QTIIdentifierProp(false), // Optional[QTIIdentifier] = None
    expectedLength: NonNegativeIntProp(false), // Optional[NonNegativeInt] = None
    patternMask: StringProp(false), // Optional[str] = None
    placeholderText: StringProp(false), // Optional[str] = None
    format: FormatProp(false) // Optional[Format] = None
```

Render a KTextbox.

If `base` is defined, it should have the `type="number", else the prop should not be passed.

The width of the KTextbox should be set according to the `expectedLength` if defined, but should not be used as a constraint.

If `patternMask` is provided, this should be interpreted as a regular expression that should be used to control the value of the `invalid` prop when matched against the value of the KTextbox.

We define format, but do not currently allow non-text inputs here, so we will not implement this as yet.

rtibbles:
## Summary
* Depends on https://github.com/learningequality/kolibri/pull/13533
* Adds qti viewer plugin
* Uses kolibri-zip to load QTI zip file, read metadata, and load assessment item according to itemId prop
* Adds support for the itemData prop
* Updates the SafeHTML component to allow registering of custom tags.
* Adds a dev only QTI sandbox to learn for testing QTI with interactive flagging and response editing
* Adds functions for coercing and validating QTI base data types.
* Adds common prop definition tooling and a type coersion composable - this is needed because of the use of HTML to drive props, meaning that frequently non-string data needs to be coerced to the proper value
* Implements rendering functionality for choice interactions
* Creates a helper class to handle QTI variable declaration parsing to create response variables that will collect user input
* Implements the public API for assessment type content viewers - primarily the "checkAnswer" method
* Adds initial text entry interaction implementation.
* Adds a brief README as an overview of the component architecture

## References
Fixes #13490
Fixes #13491

## Reviewer guidance
For initial (developer only) manual testing, load up Kolibri. Ensure that the qti_viewer plugin is enabled:
```
kolibri plugin enable kolibri.plugins.qti_viewer
```
Then navigate to: `http://127.0.0.1:8000/en/learn/#/qti_sandbox/`
Pressing the button on the top right:
image

Should open a side panel modal where you can traverse through an accordion based menu. The resources under "Multiple Choice", "Multiple Select", and "Text Entry" should be mostly implemented, although some of the advanced cases and shared vocabulary may not be completely implemented. We can revisit those when we're going for full QTI compliance!

Aside from testing, the main questions to ask during the code walkthrough are:

* Is the component hierarchy and usage clear? Is it obvious how and why data flows?
* Why are some components using render functions and others using templates?
* Is the useTypedProps composable the best way to handle this? It's implicit, and ends up requiring that all the props be wrapped in linting ignore rules because even the ones that are used are not directly accessed.
* Does the `props.js` module make sense? It may not be completely necessary now, but we will be repeating these prop definitions over and over again as we implement more of QTI.
* Is the test coverage in the values.spec and declarations.spec adequate? Is anything missing? Most of these tests were generated by claude-code and then modified by me.
* Is any of this implementation overkill for the MVP? Similarly to the Kolibri Studio PR that is the companion to this, I erred on the side of extreme data/type safety caution as a way of making sure I was doing things correctly.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -3,3 +3,6 @@
 *.ttf filter=lfs diff=lfs merge=lfs -text
 *.ttc filter=lfs diff=lfs merge=lfs -text
 *.woff filter=lfs diff=lfs merge=lfs -text
+# use Git LFS to track large JS fixture file
+# which is a vendored modified version of a file and doesn't need to track changes
+kolibri/plugins/qti_viewer/assets/src/components/__fixtures__/items.js filter=lfs diff=lfs merge=lfs -text
--- a/kolibri/__init__.py
+++ b/kolibri/__init__.py
@@ -35,6 +35,7 @@ INTERNAL_PLUGINS = [
     "kolibri.plugins.pdf_viewer",
     "kolibri.plugins.perseus_viewer",
     "kolibri.plugins.pwa",
+    "kolibri.plugins.qti_viewer",
     "kolibri.plugins.setup_wizard",
     "kolibri.plugins.slideshow_viewer",
     "kolibri.plugins.user_auth",
--- a/kolibri/plugins/learn/assets/src/routes/index.js
+++ b/kolibri/plugins/learn/assets/src/routes/index.js
@@ -13,6 +13,17 @@ import BookmarkPage from '../views/BookmarkPage.vue';
 import ExploreLibrariesPage from '../views/ExploreLibrariesPage';
 import classesRoutes from './classesRoutes';
 
+// Conditionally import QTI sandbox routes in non-production
+let qtiSandboxRoutes = [];
+if (process.env.NODE_ENV !== 'production') {
+  try {
+    const { sandboxRoutes } = require('../../../../qti_viewer/assets/src/sandbox');
+    qtiSandboxRoutes = sandboxRoutes;
+  } catch (e) {
+    // QTI viewer plugin may not be available
+  }
+}
+
 const { channelsMap, fetchChannels } = useChannels();
 const { isUserLoggedIn } = useUser();
 
@@ -173,6 +184,8 @@ export default [
       next();
     },
   },
+  // Add QTI sandbox routes in non-production
+  ...qtiSandboxRoutes,
   {
     path: '*',
     redirect: '/',
--- a/kolibri/plugins/qti_viewer/assets/src/README.md
+++ b/kolibri/plugins/qti_viewer/assets/src/README.md
@@ -1,0 +1,28 @@
+# QTI Viewer Code Architecture
+
+## Overall Architecture Approach
+
+The QTI Viewer follows a **hierarchical data flow architecture** using provide/inject pattern to allow two way data flow when there are clearly defined component ancestors and descendants, but we cannot depend on direct parent/child relationships to use props and events.
+
+## Provide/Inject Data Flow Pattern
+
+Each level provides specific concerns to its children:
+
+### QTIViewer
+- Provides assessment-level context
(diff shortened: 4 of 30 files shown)


kolibri/plugins/qti_viewer/assets/src/views/AssessmentItemRef.vue
```diff
@@ -0,0 +1,58 @@
...
+        assessmentItem: null,
+      };
+    },
+    computed: {
+      href() {
+        const relativeHref = this.dom && this.dom.getAttribute('href');
+        if (relativeHref) {
+          const path = new URL(
+            relativeHref,
+            new URL(this.filePath, 'http://b.b/')
```
nucleogenesis: Where does b.b come from here?

nucleogenesis (commented):
I looked at this before but realized I never commented. I gave it another pass and everything looks good to me.  This will be an excellent foundation to work from

kolibri/plugins/qti_viewer/assets/src/views/AssessmentItemRef.vue
rtibbles: This is the easiest way to do URL path concatenation, by using a dummy origin and then reading the path off the resultant URL object.

kolibri/plugins/qti_viewer/assets/src/views/AssessmentItemRef.vue
```diff
@@ -0,0 +1,65 @@
...
+      };
+    },
+    computed: {
+      href() {
+        const relativeHref = this.json && this.json['@href'];
+        if (relativeHref) {
+          return new URL(
+            relativeHref,
+            new URL(this.getFilePath(), 'http://b.b/')
+          ).pathname.substring(1);
```
indirectlylit: A little clarification

```suggestion
          // iterative URL path construction using dummy domain
          return new URL(
            relativeHref,
            new URL(this.getFilePath(), 'http://example.com/')
          ).pathname.substring(1);
```

rtibbles: I deliberately used a non-valid TLD to avoid any confusion, but yes.

marcellamaki (approved):
With the latest commit to extend the text entry field width so longer phrases/sentences are viewable, this seems ready to go for surveys. 

Manual QA with surveys confirms the learner survey, survey report, survey re-submit, coach survey assignment, and coach survey report functionality.

PR #8075, opened 2021-05-12
Kolibri stop for Windows

### LINKED ISSUE #7103: "Kolibri stop" command is not working in Windows
### Observed behavior

There's an error when I used `kolibri stop` in command prompt.
image

### Expected behavior

User can stop running the Kolibri using `kolibri stop` command.
### User-facing consequences

…

### Errors and logs

```
Microsoft Windows [Version 10.0.18362.836]
(c) 2019 Microsoft Corporation. All rights reserved.

C:\Users\Zech Enoch Legaspi>kolibri stop
WARNING:root:No C Extensions available for this platform.

[37mINFO     Initiating shutdown of Kolibri[0m
OSError: [WinError 87] The parameter is incorrect

The above exception was the direct cause of the following exception:

  File "c:\python36\lib\runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "c:\python36\lib\runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "C:\Python36\Scripts\kolibri.exe\__main__.py", line 9, in <module>
  File "c:\python36\lib\site-packages\kolibri\dist\click\core.py", line 764, in __call__
    return self.main(*args, **kwargs)
  File "c:\python36\lib\site-packages\kolibri\dist\click\core.py", line 717, in main
    rv = self.invoke(ctx)
  File "c:\python36\lib\site-packages\kolibri\dist\click\core.py", line 1137, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "c:\python36\lib\site-packages\kolibri\utils\cli.py", line 223, in invoke
    return super(KolibriCommand, self).invoke(ctx)
  File "c:\python36\lib\site-packages\kolibri\dist\click\core.py", line 956, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "c:\python36\lib\site-packages\kolibri\dist\click\core.py", line 555, in invoke
    return callback(*args, **kwargs)
  File "c:\python36\lib\site-packages\kolibri\utils\cli.py", line 557, in stop
    server.stop(pid=pid)
  File "c:\python36\lib\site-packages\kolibri\utils\server.py", line 169, in stop
    kill_pid(pid)
  File "c:\python36\lib\site-packages\kolibri\utils\system.py", line 102, in _windows_kill_pid
    _kill_pid(pid, signal.CTRL_C_EVENT)
  File "c:\python36\lib\site-packages\kolibri\utils\system.py", line 56, in _kill_pid
    os.kill(pid, softkill_signal_number)
SystemError: <built-in function kill> returned a result with an error set

```

### Steps to reproduce

…

### Context

* Kolibri version: 0.13.3 and 0.14.0 beta 2
 * Operating system: Windows 7 and 10
 * Browser: n/a

rtibbles:
## Summary
* Turns the restart flag into a general purpose command flag
* Uses that as the primary means for stopping the server
* Does this as sending signals on Windows is problematic
* When we do have to resort to os.kill when the process is unresponsive, catch any SystemErrors that arise

## References
Fixes https://github.com/learningequality/kolibri/issues/7103

## Reviewer guidance
Does `kolibri stop` still work? Does it work on Windows without error?

To test:
Open two terminals in Windows
In one run `kolibri start`
Wait for it to 'go to background'
In the other run `kolibri stop`
Verify that the terminal with start, then stops!

----

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/utils/server.py
+++ b/kolibri/utils/server.py
@@ -62,7 +62,7 @@ PROFILE_LOCK = os.path.join(conf.KOLIBRI_HOME, "server_profile.lock")
 PORT_CACHE = os.path.join(conf.KOLIBRI_HOME, "port_cache")
 
 # File used to send a state transition command to the server process
-PROCESS_CONTROL_FLAG = os.path.join(conf.KOLIBRI_HOME, "prcess_control.flag")
+PROCESS_CONTROL_FLAG = os.path.join(conf.KOLIBRI_HOME, "process_control.flag")
 
 # This is a special file with daemon activity. It logs ALL stderr output, some
 # might not have made it to the log file!
--- a/kolibri/utils/server.py
+++ b/kolibri/utils/server.py
@@ -62,7 +62,7 @@ PROFILE_LOCK = os.path.join(conf.KOLIBRI_HOME, "server_profile.lock")
 PORT_CACHE = os.path.join(conf.KOLIBRI_HOME, "port_cache")
 
 # File used to send a state transition command to the server process
-COMMAND_FLAG = os.path.join(conf.KOLIBRI_HOME, "command.flag")
+PROCESS_CONTROL_FLAG = os.path.join(conf.KOLIBRI_HOME, "prcess_control.flag")
 
 # This is a special file with daemon activity. It logs ALL stderr output, some
 # might not have made it to the log file!
@@ -406,14 +406,14 @@ class SignalHandler(BaseSignalHandler):
         self.process_pid = os.getpid()
 
 
-class CommandPlugin(Monitor):
+class ProcessControlPlugin(Monitor):
     def __init__(self, bus):
         self.mtime = self.get_mtime()
         Monitor.__init__(self, bus, self.run, 1)
 
     def get_mtime(self):
         try:
-            return os.stat(COMMAND_FLAG).st_mtime
+            return os.stat(PROCESS_CONTROL_FLAG).st_mtime
         except OSError:
             return 0
 
@@ -421,10 +421,12 @@ class CommandPlugin(Monitor):
         mtime = self.get_mtime()
         if mtime > self.mtime:
             # The file has been deleted or modified.
-            with open(COMMAND_FLAG, "r") as f:
+            with open(PROCESS_CONTROL_FLAG, "r") as f:
                 try:
                     command = f.read().strip()
                 except (IOError, OSError):
+                    # If the file does not exist, or there is
+                    # an error when reading the file, we just carry on.
                     command = ""
             if command == RESTART:
                 self.bus.log("Restarting server.")
@@ -459,7 +461,7 @@ def stop():
     if not pid:
         return status
 
-    with open(COMMAND_FLAG, "w") as f:
+    with open(PROCESS_CONTROL_FLAG, "w") as f:
         f.write(STOP)
     if not wait_for_status(STATUS_STOPPED, timeout=30):
         if pid_exists(pid):
@@ -621,7 +623,7 @@ def start(port=0, zip_port=0, serve_http=True, background=False):
         autoreloader = Autoreloader(bus)
         autoreloader.subscribe()
 
-    reload_plugin = CommandPlugin(bus)
+    reload_plugin = ProcessControlPlugin(bus)
     reload_plugin.subscribe()
 
     bus.graceful()
@@ -632,7 +634,7 @@ def start(port=0, zip_port=0, serve_http=True, background=False):
 
 
 def restart():
-    with open(COMMAND_FLAG, "w") as f:
+    with open(PROCESS_CONTROL_FLAG, "w") as f:
         f.write(RESTART)
     if not wait_for_status(STATUS_STOPPED):
         return False
--- a/kolibri/utils/server.py
+++ b/kolibri/utils/server.py
@@ -24,7 +24,6 @@ from zeroconf import get_all_addresses
 
 import kolibri
 from .system import become_daemon
-from .system import kill_pid
 from .system import pid_exists
 from kolibri.utils import conf
 from kolibri.utils.android import on_android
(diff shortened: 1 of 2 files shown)


kolibri/utils/server.py
```diff
@@ -62,8 +61,8 @@
 # File used to store previously available ports
 PORT_CACHE = os.path.join(conf.KOLIBRI_HOME, "port_cache")
 
-# File used to flag a restart is required
-RESET_FLAG = os.path.join(conf.KOLIBRI_HOME, "reset.flag")
+# File used to send a state transition command to the server process
+COMMAND_FLAG = os.path.join(conf.KOLIBRI_HOME, "command.flag")
```
micahscopes: For the uninitiated, calling this something like `PROCESS_CONTROL_FLAG` might be a little more clear, although that specific name is a little long winded. Before I understood the context I first associated "`COMMAND`" with the CLI interface.

kolibri/utils/server.py
```diff
@@ -403,29 +406,41 @@ def ENTER(self):
...
         mtime = self.get_mtime()
         if mtime > self.mtime:
             # The file has been deleted or modified.
-            self.bus.log("Restarting server.")
-            self.thread.cancel()
-            self.bus.restart()
+            with open(COMMAND_FLAG, "r") as f:
+                try:
+                    command = f.read().strip()
+                except (IOError, OSError):
```
micahscopes, line 427: I like the specificity of error handling here.  Could you add a comment about when these errors happen?  I'm assuming it's when the flag file doesn't exist?

kolibri/utils/server.py
```diff
@@ -609,7 +621,7 @@ def start(port=0, zip_port=0, serve_http=True, background=False):
         autoreloader = Autoreloader(bus)
         autoreloader.subscribe()
 
-    reload_plugin = RestartPlugin(bus)
+    reload_plugin = CommandPlugin(bus)
```
micahscopes: Again here, I think something like `ProcessControlPlugin` has a nice ring to it

kolibri/utils/server.py
```diff
@@ -444,29 +459,26 @@ def stop():
...
-            "Process wth pid %s still exists after soft kill signal; attempting a SIGKILL."
-            % pid
-        )
-        os.kill(pid, signal.SIGKILL)
-    if pid_exists(pid):
-        logging.error("Kolibri process has failed to shutdown")
-        return STATUS_UNCLEAN_SHUTDOWN
+            logger.debug("Process wth pid %s still exists; attempting a SIGKILL." % pid)
+            try:
+                os.kill(pid, signal.SIGKILL)
```
micahscopes, line 470: It looks like previously on POSIX systems, `SIGTERM` was used, but with this PR, `SIGKILL` will always be used.  Is the idea that the following lines should handle shutdown gracefully on POSIX systems, and if not a `SIGKILL` is fine?
```
self.bus.log("Stopping server.")
self.thread.cancel()
self.bus.transition("EXITED")
```

Do you see any risk to changing this for POSIX systems?

kolibri/utils/server.py
```diff
@@ -444,29 +459,26 @@ def stop():
...
+            except SystemError as e:
+                logger.debug(
```
micahscopes, line 472: I'm a little concerned that this could potentially cause reentrancy issues, since calls to `stderr` are reentrant, although I'm not totally certain that it would.

micahscopes (changes requested):
Overall it looks good!  I made some requests for minor readability changes and left a comment about a potential reentrancy issue.

kolibri/utils/server.py
rtibbles: Sounds good!

kolibri/utils/server.py
rtibbles: Yes, or the file exists but can't be read - in Python 3, `IOError` and `OSError` were merged I believe, but in Python 2 we have to catch both.

kolibri/utils/server.py
rtibbles: The idea is not to use signals at all (and instead use the file based flag) for cross-OS consistency. We revert to using SIGKILL in the extreme case where the Kolibri process is unresponsive to that.

kolibri/utils/server.py
rtibbles: The process it is killing is a different process to the one where the SystemError is being thrown, so I don't think this should be a concern.

micahscopes: Sounds good!

radinamatic:
All seems to be working as described! 👍🏽 

Win 7 | Win 10
--- | ---
win7-kolibri-stop | win10-kolibri-stop

radinamatic (approved):
LGTM! 💯

kolibri/utils/server.py
```diff
@@ -62,8 +61,8 @@
 # File used to store previously available ports
 PORT_CACHE = os.path.join(conf.KOLIBRI_HOME, "port_cache")
 
-# File used to flag a restart is required
-RESET_FLAG = os.path.join(conf.KOLIBRI_HOME, "reset.flag")
+# File used to send a state transition command to the server process
+PROCESS_CONTROL_FLAG = os.path.join(conf.KOLIBRI_HOME, "prcess_control.flag")
```
micahscopes: I assume you meant for the file to be called "process_control.flag"?

micahscopes (approved):
Just one more slight spelling slip!  But otherwise looks great

kolibri/utils/server.py
rtibbles:
```suggestion
PROCESS_CONTROL_FLAG = os.path.join(conf.KOLIBRI_HOME, "process_control.flag")
```

PR #8076, opened 2021-05-12
Catch ValueError to catch errant JSONDecodeErrors.

### LINKED ISSUE #7502: Catch ValueError during network discovery for when JSON parsing fails
### Observed behavior

When trying to import content channels (from local network or internet), no error message is produced even though the connection isn't established. (The "trying to connect" message stays on forever) This happens when there is a typo in the address. (I used kolibri-demo.learningequality.omg)  When there's no internet, an error message DOES get produced. 

### Expected behavior

An error message should've been produced.

### User-facing consequences

The user will not know how long they have to wait to see if the connection is being established.

### Errors and logs

…

### Steps to reproduce

Import content channels from local network or internet >input "kolibri-demo.learningequality.omg" or something else with a typo in it.

### Context

Kolibri version: 0.14.3
Operating system: Android 5.1
Browser: Android App
Device: BLU Advance 5.0

rtibbles:
## Summary
* Somehow inputting a non-existent URL for a peer on the Android or Mac App would result in a JSON decode error in the network client
* Instead of just throwing a 500 error, handle that exception and let the requests return

## References
Fixes #7502

## Reviewer guidance
Any concerns?

----

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/core/discovery/utils/network/client.py
+++ b/kolibri/core/discovery/utils/network/client.py
@@ -55,6 +55,10 @@ class NetworkClient(object):
                     return response.url.rstrip("/").replace("api/public/info", "")
             except (requests.RequestException) as e:
                 logger.info("Unable to connect: {}".format(e))
+            except ValueError:
+                logger.info(
+                    "Invalid JSON returned when attempting to connect to a remote server"
+                )
 
         # we weren't able to connect to any of the URL variations, so all we can do is throw
         raise errors.NetworkLocationNotFound()


jamalex:
Could you give an example of a "nonexistent URL"? Is it a path on a server that does exist, or an origin that shouldn't be responding at all?

rtibbles:
> origin that shouldn't be responding at all

In the original issue the URL entered was `kolibri-demo.learningequality.omg` - I could not replicate a `JSONDecodeError` with that locally, but the stack traces were fairly clear that that was the error being raised. And all we want to do is make sure we properly ignore things for which requests is unable to connect or decode.

jamalex (approved):
Ok, yeah, just wanted to make sure it wasn't something about Kolibri sending back something malformed, or it making a request to a bad path, that we should address at the root. Looks good.

PR #8082, opened 2021-05-12
Create mock for `kolibri.resources` that automocks every exported Resource class

jonboiser:
This simplifies the mocking of `kolibri.resources`. Now, if your test might call some resource, do this:

```
jest.mock('kolibri.resources')
```

This will import a mock that automocks every Resource class. Then, if you need to mock a specific Resource, import it, set its mock behavior (and remember to reset it at the end of the test)

```
import { FacilityUserResource } from 'kolibri.resources';

jest.mock('kolibri.resources')

describe('', () => {
  afterEach(() => {
    FacilityUserResource.fetchCollection.mockReset();
   })
  it('...', () => {
	FacilityUserResource.fetchCollection.mockResolvedValue([...]);
   })
})

```

Since this mock should handle most use cases, I also deleted the `jestMockResource` utility.


rtibbles (approved):
Nice, good cleanup! Tests still pass, so looking good.

PR #8085, opened 2021-05-13
KDP-friendly kolibri-tools version

nucleogenesis:
Ran `yarn publish-packages` on `develop` - successfully bumped all package versions and published them to npm.

Also made @rtibbles owner https://www.npmjs.com/package/browserslist-config-kolibri/access


packages/kolibri-core-for-export/package.json
```diff
@@ -34,6 +36,6 @@
     "vuex-router-sync": "^5.0.0"
   },
   "devDependencies": {
-    "kolibri-tools": "0.13.0-dev.10"
+    "kolibri-tools": "0.14.7",
```
jredrejo: this trailing comma is giving problems when doing `yarn install `

jredrejo (changes requested):

jredrejo (changes requested):
There was a small bug cause by one comma

nucleogenesis:
@jredrejo apologies I missed your review on this - just pushed a fix.

packages/publish.js
```diff
@@ -67,11 +67,11 @@ const currentCwd = process.cwd();
 process.chdir(path.resolve(__dirname, '..'));
 
 function publishCommand(workspace) {
-  execSync(`yarn workspace ${workspace} publish --new-version ${version} --tag ${tag}`, {
```
rtibbles: These three packages all need to be published - and the build is failing completely because of the hard coded version number here that doesn't match up to the version that has actually been generated for each of the packages.

nucleogenesis:
@rtibbles @jredrejo I thought I'd clean this up really quickly and ran the `packages/publish` script - which updated the package.json files - but when I go to NPM I don't see the versions.

So I ran it again and get this now:

`error Couldn't publish package: "https://registry.yarnpkg.com/kolibri-tools: You cannot publish over the previously published versions: 0.14.8-dev.1."`

I'm clearly missing something here... https://www.npmjs.com/package/kolibri-tools

packages/publish.js
```diff
@@ -72,6 +72,6 @@ function publishCommand(workspace) {
   });
 }
 
-['eslint-plugin-kolibri', 'kolibri', 'kolibri-tools'].forEach(publishCommand);
```
rtibbles: This still needs to be reverted - we need to publish each of these.

nucleogenesis: I'd just realized I missed that bit when rebasing to fix the conflicts. I think I'll just start from a fresh `develop` branch and force push it here

rtibbles:
> I'm clearly missing something here... https://www.npmjs.com/package/kolibri-tools

0.14.8-dev.1 is there as `next`. Because of the extant edits to the publishing command though, there is no updated version of the `eslint-plugin-kolibri` package, nor the `kolibri` package.

This PR also needs to be rebased as it has multiple merge conflicts.

packages/publish.js
nucleogenesis: Do we need to be publishing `browserslist-config-kolibri` ?

`error Package marked as private, not publishing.`

It is added to that array in a recent-ish commit

rtibbles: Yes, we would need to - I guess I forgot what 'private' means, and didn't unflag that...

nucleogenesis:
@rtibbles how weird - npm really showed only 0.14.7 as `next` until I came back to it this morning and refreshed. 

I cannot find or un-privatize that browserslist package on npm. If you can make that public I will re-run the command on the latest `develop` branch and force push on this branch.

rtibbles:
The browserlist-config-kolibri package has not been published to npm yet - we just need to remove this flag from the package.json https://github.com/learningequality/kolibri/blob/release-v0.15.x/packages/browserslist-config-kolibri/package.json#L9 to unflag it as private.

nucleogenesis:
So - it looks like `kolibri-components` is a requirement for the version of `kolibri-tools` that KDS points to.

I think we'll just need to update the commit we're pointing to for KDS in kolibri/core/package.json once https://github.com/learningequality/kolibri-design-system/pull/258 is merged.

nucleogenesis:
There is an additional issue that `kolibri-tools` cli expects there to be a `./setup.cfg` file in any implementing repository.

[This commit](https://github.com/learningequality/kolibri/commit/29791ea06e90f14f706e820b35e9e0a737a56837) introduces an issue now where the cli tools are unusable in `kolibri-design-system` when using the latest `kolibri-tools` that was published here.

I can create a separate issue for that - which will block the completion of this one as we'll want to rebase this onto those changes, publish again, then merge along with KDS#258

rtibbles:
> This commit introduces an issue now where the cli tools are unusable in kolibri-design-system when using the latest kolibri-tools that was published here.

Ah, whoops - should have made this optional.

indirectlylit:
migrated work to https://github.com/learningequality/kolibri/pull/8459

PR #8086, opened 2021-05-14
First updates to hashi documentation

marcellamaki:
At this point this is a pretty high-level overview, so would appreciate thoughts about 
- how clear this is/isn't 
- if it is serving its intended purpose well
- what more specific technical details would be useful to include

I think as this work continues we might need to continue to update the documentation, so my goal for this PR is to just get a foundation in place.

### DIFF
--- a/docs/frontend_architecture/HTML5_API.rst
+++ b/docs/frontend_architecture/HTML5_API.rst
@@ -17,6 +17,7 @@ Access the API from within an HTML5 app by using ``window.kolibri.[function]``
 Functions:
 
 .. code-block:: javascript
+
     /**
     * Type definition for Language metadata
     * @typedef {Object} Language
@@ -97,8 +98,8 @@ Functions:
     * Method to search for contentnodes on Kolibri and return
     * an array of matching metadata
     * @param {Object} options - The different options to search by
-    * @param {string=} options.keyword - search term for key word search
-    * @param {string=} options.under - id of topic to search under, or 'self'
+    * @param {string} [options.keyword] - search term for key word search
+    * @param {string} [options.under] - id of topic to search under, or 'self'
     * @return {Promise<PageResult>} - a Promise that resolves to an array of ContentNodes
     */
     searchContent(options)
--- a/docs/frontend_architecture/HTML5_API.rst
+++ b/docs/frontend_architecture/HTML5_API.rst
@@ -17,13 +17,70 @@ Access the API from within an HTML5 app by using ``window.kolibri.[function]``
 Functions:
 
 .. code-block:: javascript
+    /**
+    * Type definition for Language metadata
+    * @typedef {Object} Language
+    * @property {string} id - an IETF language tag
+    * @property {string} lang_code - the ISO 639‑1 language code
+    * @property {string} lang_subcode - the regional identifier
+    * @property {string} lang_name - the name of the language in that language
+    * @property {('ltr'|'rtl'|)} lang_direction - Direction of the language's script,
+    * top to bottom is not supported currently
+    */
+
+    /**
+    * Type definition for ContentNode metadata
+    * @typedef {Object} ContentNode
+    * @property {string} id - unique id of the ContentNode
+    * @property {string} channel_id - unique channel_id of the channel that the ContentNode is in
+    * @property {string} content_id - identifier that is common across all instances of this resource
+    * @property {string} title - A title that summarizes this ContentNode for the user
+    * @property {string} description - detailed description of the ContentNode
+    * @property {string} author - author of the ContentNode
+    * @property {string} thumbnail_url - URL for the thumbnail for this ContentNode,
+    * this may be any valid URL format including base64 encoded or blob URL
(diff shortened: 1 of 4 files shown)


packages/hashi/README.md
```diff
@@ -22,3 +22,104 @@ Step 1: Install Hashi package deps (run from Kolibri root)
 Step 2: Build Kolibri and hashi
 
 `yarn run build`
+
+
+
+Custom Channels
```
rtibbles: We should call this 'custom navigation' as this can happen in any topic, not just at the channel level.

packages/hashi/README.md
```diff
@@ -22,3 +22,104 @@ Step 1: Install Hashi package deps (run from Kolibri root)
...
+    Gets the current context object that is encoded in the URL
+- `getVersion()`
+    Gets the version of Kolibri (and thus API) that is available
+
+
+### Data Flow
+
+The HTML5 App, using window.kolibri, calls a Hashi Kolibri API function. The general API pattern is set up through promises, which allows us to manage asynchronous data requests.
+
+`sendMessageAwaitReply()`, which is defined in mediator.js, is the core of how these requests are managed. The function registers and removes message handlers, which in turn use the postMessage API to communicate either locally (to `window`) or to `remote`.
```
rtibbles: I think it is useful to document this, but I feel like documenting the APIs for someone developing against Hashi should be separated from documenting these implementation details for someone developing hashi.

rtibbles (changes requested):
The main thrust of the issue is to document hashi for HTML5 app developers, in this PR we seem to be doing both this, and also documenting hashi for developers using Hashi in Kolibri, and for people developing hashi. I think it would be good to separate these concerns by documenting the different tracks separately.

The README seems like a good place for the latter, but not for the HTML5 app developer audience.

docs/frontend_architecture/HTML5_API.rst
```diff
@@ -0,0 +1,93 @@
...
+
+Functions:
+
+.. code-block:: javascript
+
+    /*
+    * Method to query contentnodes from Kolibri and return
+    * an array of matching metadata
+    * @param {Object} options - The different options to filter by
+    * @param {string=} options.parent - id of the parent node to filter by, or 'self'
```
jonboiser: I just noticed that been using a different "optional parameter" syntax in my JSDoc strings for the theming. Personally, I like the style I use more, since it's emphasizing the argument name, rather than the type, e.g.
```suggestion
    * @param {string} [options.parent] - id of the parent node to filter by, or 'self'
```

docs/frontend_architecture/HTML5_API.rst
```diff
@@ -0,0 +1,93 @@
...
+    */
+    getContentById(id)
+
+    /*
+    * Method to search for contentnodes on Kolibri and return
+    * an array of matching metadata
+    * @param {Object} options - The different options to search by
+    * @param {string=} options.keyword - search term for key word search
+    * @param {string=} options.under - id of topic to search under, or 'self'
+    * @return {Promise<PageResult>} - a Promise that resolves to an array of ContentNodes
```
jonboiser, line 103: Could you add the `typedefs` for `PageResult` and other Object types in the docs as well?

marcellamaki:
@radinamatic - these were the original Gherkin scenarios drafted at the start of the project to help guide the outcomes. Please let me know if they are okay for now (in terms of how they are written - I know we will probably have to go over some of the functionality), or if any changes would be helpful. Thanks!

docs/frontend_architecture/HTML5_API.rst
```diff
@@ -0,0 +1,151 @@
...
+=================
+
+The purpose of the ``kolibri.js`` extension of our HTML5 API is to allow a sandboxed HTML5 app to safely request the main Kolibri application's data.
+
+External/partner product teams can create HTML5 applications that are fully embeddable within Kolibri and can read Kolibri content data, which they otherwise wouldn't be able to access. This opens up possibilities for creative ways in which learners can engage with content, because partners can create any type of app they want. The app could be something completely new, developed for a content source that we are adding to the platform, or it could be a branded, offline recreation of a partner's existing learning app that previously would not have been able to exist on Kolibri.
+
+When a user has permissions to access a custom channel, and they click on it in the main learn tab, rather than viewing "normal Kolibri," they will experience a full-screen HTML5 app. One "out-of-the-box" user interaction is the "navigateTo()" function, which opens  a modal that displays a content node. For other data fetching requests, the app, not Kolibri, has the responsibilty of determining what to do with that data.
```
radinamatic: I would suggest _italic_ formatting for what you've put between quotes, and pairs of backtics for ``code``:

> ...rather than viewing the _normal_ Kolibri, they will experience a full-screen HTML5 app. One _out-of-the-box_ user interaction is the ``navigateTo()`` function, which opens a modal that displays a content node.

rtibbles (approved):
So as not to let the perfect be the enemy of the good, let's merge this and iterate later.

PR #8090, opened 2021-05-18
Add utility to enable user migration

### LINKED ISSUE #8061: Allow for destructive teleportation and merging of FacilityUsers across different facilities
Currently, no Kolibri device possesses the capacity to engage in single user syncing. This has resulted in there being many devices, particularly app installation devices that have a standalone facility containing as few as 1 users.

In the future, even with single user syncing, this will continue to be an issue, as many users will wish to start using Kolibri in advance of being able to sync to a specific facility server device. They will therefore, quite reasonably, expect to be able to create a user account straight away and start having their progress logged and tracked.

Reconciling these facility users or later allowing them to join a facility has a bit of a blocker due to the way that we partition data by the facility they are in. We cannot therefore, simply update the facility foreign key and dataset foreign key and call it a day - the partitioning of the UUIDs of all the data associated with the user will still be based on the original facility.

In order to resolve this issue, we need to give the appearance to the end user of this seamless facility switching happening, but in a way that obeys the constraints of our data synchronization mechanisms.

As such, we will engage in destructive teleportation - on any particular local device that has the user we wish to move, and also the data about the facility we wish to move the user to, we will give the appearance of having moved the user by recreating all the data about the user in the new facility.

The are two separate scenarios to cover here:
1. There is no target user for merging in the target facility
2. There is a user in the target facility that we wish to merge the other user into

In scenario 1, we will simply read all the data about the FacilityUser, and create a new facility user with that data in the target facility. Once the facility user has been created, we will then copy all of the associated logging models to be recreated, targeting the new user.

No assignment data will be copied.

In scenario 2, we will simply discard any information from the original user, and assume that the information on the user targeted for the merge is canonical. We will then read all of the associated logging models and recreate them with the target user as the user foreign key.

In the case of a conflict in SummaryLog, we will take a max of the progress across the two Summary Logs, and sum all the time spent across all the related session logs. In the case of a conflict in the MasteryLog, we will mark 'complete' by a logical OR on the complete fields, and take the max of the level properties.

No assignment data will be copied.

In both instances once the copy has been successful, the original user and all their data will be hard deleted.

rtibbles:
## Summary
* Adds utility to merge one user into another

## References
Fixes #8061

## Reviewer guidance
Do the tests make sense?
Should we be stopping signups in the public and private viewsets when the facility doesn't allow signups?

----

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/core/auth/api.py
+++ b/kolibri/core/auth/api.py
@@ -24,9 +24,9 @@ from django.db.models import TextField
 from django.db.models import Value
 from django.db.models.functions import Cast
 from django.http import Http404
+from django.http import HttpResponseForbidden
 from django.utils.decorators import method_decorator
 from django.utils.timezone import now
-from django.views.decorators.csrf import csrf_exempt
 from django.views.decorators.csrf import ensure_csrf_cookie
 from django_filters.rest_framework import CharFilter
 from django_filters.rest_framework import DjangoFilterBackend
@@ -40,6 +40,7 @@ from rest_framework import status
 from rest_framework import views
 from rest_framework import viewsets
 from rest_framework.response import Response
+from rest_framework.serializers import ValidationError
 
 from .constants import collection_kinds
 from .constants import role_kinds
@@ -546,9 +547,7 @@ class SignUpViewSet(viewsets.ViewSet):
             "username": request.data.get("username", ""),
             "full_name": request.data.get("full_name", ""),
             "password": request.data.get("password", ""),
-            "facility": request.data.get(
-                "facility", Facility.get_default_facility().id
-            ),
+            "facility": request.data.get("facility"),
             "gender": request.data.get("gender", ""),
             "birth_year": request.data.get("birth_year", ""),
         }
@@ -556,9 +555,36 @@ class SignUpViewSet(viewsets.ViewSet):
     def create(self, request):
 
         data = self.extract_request_data(request)
+        facility_id = data["facility"]
+
+        if facility_id is None:
+            facility = Facility.get_default_facility()
+            data["facility"] = facility.id
+        else:
+            try:
+                facility = Facility.objects.select_related("dataset").get(
+                    id=facility_id
+                )
+            except Facility.DoesNotExist:
+                raise ValidationError(
+                    "Facility does not exist.",
+                    code=error_constants.FACILITY_DOES_NOT_EXIST,
+                )
+
+        if not facility.dataset.learner_can_sign_up:
+            return HttpResponseForbidden("Cannot sign up to this facility")
+
         # we validate the user's input, and if valid, login as user
         serialized_user = self.serializer_class(data=data)
         if serialized_user.is_valid(raise_exception=True):
+            if (
+                data["password"] == "NOT_SPECIFIED"
+                and not facility.dataset.learner_can_login_with_no_password
+            ):
+                raise ValidationError(
+                    "No password specified and it is required",
+                    code=error_constants.PASSWORD_NOT_SPECIFIED,
+                )
             serialized_user.save()
             if data["password"] != "NOT_SPECIFIED":
                 serialized_user.instance.set_password(data["password"])
@@ -569,29 +595,6 @@ class SignUpViewSet(viewsets.ViewSet):
                 facility=data["facility"],
             )
             login(request, authenticated_user)
-            return Response(serialized_user.data, status=status.HTTP_201_CREATED)
-
-
-@method_decorator(csrf_exempt, name="dispatch")
-class PublicSignUpViewSet(SignUpViewSet):
-    """
-    Identical to the SignUpViewset except that it also allows a hashed password
-    and does not login the user.
-    This endpoint is intended to allow a FacilityUser in a different facility
-    on another device to be cloned into a facility on this device, to facilitate
-    moving a user from one facility to another.
-    """
-
-    def create(self, request):
-
-        data = self.extract_request_data(request)
-        password_hashed = request.data.get("password_hashed", False)
-        serialized_user = self.serializer_class(data=data)
-        if serialized_user.is_valid(raise_exception=True):
-            serialized_user.save()
-            if data["password"] != "NOT_SPECIFIED" and not password_hashed:
-                serialized_user.instance.set_password(data["password"])
-                serialized_user.instance.save()
             return Response(serialized_user.data, status=status.HTTP_201_CREATED)
 
 
--- a/kolibri/core/auth/api.py
+++ b/kolibri/core/auth/api.py
@@ -26,6 +26,7 @@ from django.db.models.functions import Cast
 from django.http import Http404
 from django.utils.decorators import method_decorator
 from django.utils.timezone import now
+from django.views.decorators.csrf import csrf_exempt
 from django.views.decorators.csrf import ensure_csrf_cookie
 from django_filters.rest_framework import CharFilter
 from django_filters.rest_framework import DjangoFilterBackend
@@ -561,13 +562,36 @@ class SignUpViewSet(viewsets.ViewSet):
             serialized_user.save()
             if data["password"] != "NOT_SPECIFIED":
                 serialized_user.instance.set_password(data["password"])
-            serialized_user.instance.save()
+                serialized_user.instance.save()
             authenticated_user = authenticate(
                 username=data["username"],
                 password=data["password"],
                 facility=data["facility"],
             )
             login(request, authenticated_user)
+            return Response(serialized_user.data, status=status.HTTP_201_CREATED)
+
+
+@method_decorator(csrf_exempt, name="dispatch")
+class PublicSignUpViewSet(SignUpViewSet):
+    """
+    Identical to the SignUpViewset except that it also allows a hashed password
+    and does not login the user.
+    This endpoint is intended to allow a FacilityUser in a different facility
+    on another device to be cloned into a facility on this device, to facilitate
+    moving a user from one facility to another.
+    """
+
+    def create(self, request):
+
+        data = self.extract_request_data(request)
+        password_hashed = request.data.get("password_hashed", False)
+        serialized_user = self.serializer_class(data=data)
+        if serialized_user.is_valid(raise_exception=True):
+            serialized_user.save()
+            if data["password"] != "NOT_SPECIFIED" and not password_hashed:
+                serialized_user.instance.set_password(data["password"])
+                serialized_user.instance.save()
             return Response(serialized_user.data, status=status.HTTP_201_CREATED)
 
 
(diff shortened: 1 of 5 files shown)


kolibri/core/auth/utils.py
```diff
@@ -20,3 +26,134 @@ def create_adhoc_group_for_learners(classroom, learners):
...
+    ],
+}
+
+
+def _merge_user_models(source_user, target_user):
+    for f in user_fields:
+        source_value = getattr(source_user, f, None)
+        target_value = getattr(target_user, f, None)
+        if not target_value and source_value:
+            setattr(target_user, f, source_value)
```
jamalex, line 37: Might be good to add a comment indicating that if they both have a field specified, the target user's version will take precedence.

kolibri/core/auth/utils.py
```diff
@@ -20,3 +26,134 @@ def create_adhoc_group_for_learners(classroom, learners):
...
+        raise ValueError("Cannot merge a user with themselves")
+
+    dataset_cache.clear()
+    dataset_cache.activate()
+
+    _merge_user_models(source_user, target_user)
+
+    id_map = {}
+
+    def _merge_log_data(source_user, target_user, LogModel):
```
jamalex: This function is pretty high-density. A few judicious comments to section it out a bit could help readability.

kolibri/core/auth/utils.py
```diff
@@ -20,3 +26,134 @@ def create_adhoc_group_for_learners(classroom, learners):
...
+
+    _merge_log_data(source_user, target_user, UserSessionLog)
+
+    _merge_log_data(source_user, target_user, MasteryLog)
+
+    _merge_log_data(source_user, target_user, AttemptLog)
+
+    source_user.delete()
+
+    dataset_cache.deactivate()
```
jamalex: Would a context manager pattern be safer for this, in case an exception occurs, so it still does this as part of cleanup?

kolibri/core/auth/utils.py
```diff
@@ -20,3 +26,134 @@ def create_adhoc_group_for_learners(classroom, learners):
...
+
+    _merge_log_data(source_user, target_user, ContentSummaryLog)
+
+    _merge_log_data(source_user, target_user, UserSessionLog)
+
+    _merge_log_data(source_user, target_user, MasteryLog)
+
+    _merge_log_data(source_user, target_user, AttemptLog)
+
+    source_user.delete()
```
jamalex: Is this always the desired behavior? And do we want it to be a full propagated deletion, or just a removal from this device?

In the S2S case, we're going to remove the entire source facility after the migration happens. But probably cleanest to do as a "removal" rather than deletion. Otherwise it will keep all the data around in the Morango Store, just waiting for the day when it can excitedly tell another device that it should delete it too.

kolibri/core/auth/utils.py
```diff
@@ -20,3 +26,134 @@ def create_adhoc_group_for_learners(classroom, learners):
...
+            new_log = LogModel(user=target_user, **data)
+            new_log.id = new_log.calculate_uuid()
+            new_log.ensure_dataset()
+            log_map[log.id] = new_log.id
+            new_logs.append(new_log)
+        LogModel.objects.bulk_create(new_logs, batch_size=750)
+
+    _merge_log_data(source_user, target_user, ContentSessionLog)
+
+    _merge_log_data(source_user, target_user, ContentSummaryLog)
```
jamalex: This one has a very real potential for collisions, if the two users had engaged with the same piece of content. So it will probably need to have some special handling -- e.g. maybe an argument to `_merge_log_data` called `avoid_collisions` that does a lookup to see if records with these IDs already exist (and instead of a long string of separate foreignkey lookups, could batch it with `__in`).

It looks like something similar would be needed for MasteryLog, based on how it calculates its ID. And I guess we don't bring across ExamLog since the exam won't exist if it's a different facility?

While under normal circumstances those are the only ones that are likely to have collisions, if the same source user's data is merged into a target user twice (which could happen, if the source user had previously been synced to two different devices, etc), then we need to think about the session logs as well. There are two ways that could go:

- We do what you have here, which is not to copy across the `_morango_source_id`. Then the double-migration scenario would result in duplicates of all session logs being created, because the source ID for these models is random.
- Or, we include `_morango_source_id` in the fields to copy over for the session logs. They'll still end up with new unique UUIDs, but if they're ported over again (or, if they're ported back later to the original user), they'll now have the potential for colliding, in which case we wouldn't need an `avoid_collisions` argument, because we'd need to do a target-aware merge for _all_ the models.

kolibri/core/auth/utils.py
```diff
@@ -20,3 +26,134 @@ def create_adhoc_group_for_learners(classroom, learners):
...
+    UserSessionLog,
+]
+
+user_fields = [
+    "gender",
+    "birth_year",
+    "id_number",
+]
+
+log_fields = {
```
jamalex: Given how hard it can be to remember to add stuff to things like this when models change, I wonder whether a blocklist might be safer than an allowlist?

Or, we might even consider having both! Then the tests could assert that there are no fields on the models that have not been listed in either the allowlist or the blocklist, forcing someone adding a new field to determine which one it belongs in?

kolibri/core/auth/api.py
```diff
@@ -571,6 +572,29 @@ def create(self, request):
             return Response(serialized_user.data, status=status.HTTP_201_CREATED)
 
 
+@method_decorator(csrf_exempt, name="dispatch")
+class PublicSignUpViewSet(SignUpViewSet):
```
jamalex: I was going to comment that we should add verification here that "allow signups" is enabled on the facility, but then saw it was inherited... but then saw that we don't seem to do that validation on `SignUpViewSet` either? Does that setting only impact what shows up in the UI...??

The sending of the hashed password could be avoided by using a temporary password and then overwriting the hashed password once the user has been synced onto the local device, but if we're adding this new endpoint anyway, then I guess we might as well. And not creating the endpoint isn't ideal, as the other thing that's different here is the CSRF exemption (which removes the need for additional requests and code). And this being a public API endpoint means we'll keep it consistent over time. Though... we have it inheriting the `extract_request_data` method from the non-public viewset -- might be safest to have it just do username and password, and then set the other stuff after the sync has happened, so the API interface for this isn't changing over time.

jamalex: I'm thinking more about the sending of the hashed password... in order to do the subsequent CSR request to sync the newly created user over to the client, one will need to know the password of the user (for cases where we don't know the admin password). And if we're only working with the hashed password, we won't know the plaintext password, and the end user won't necessarily be in the loop for this process (e.g. for S2S reprovisioning), so we can't just ask them. That means the idea of using a temporary password during creation, then overwriting the hashed password locally after the sync has happened, may be the way to go?

(Side note: After peeking at the parent viewset, I was amused that if you sign up to Kolibri using the password "NOT_SPECIFIED", it creates the account but then errors out with a 500 trying to auto-sign you in...)

kolibri/core/auth/test/test_api.py
```diff
@@ -926,6 +927,100 @@ def test_sign_up_able_no_guest_access(self):
...
 
+class PublicSignUpTestCase(APITestCase):
+    @classmethod
+    def setUpTestData(cls):
+        cls.facility = FacilityFactory.create()
+        provision_device()
+
+    def post_to_sign_up(self, data):
+        return self.client.post(
+            reverse("kolibri:core:publicsignup-list"), data=data, format="json"
```
jamalex: Not sure what pattern we use elsewhere, but it might be wise to have tests of public endpoints actually be hardcoded to make requests to a fixed path, since URL reversal would mask any accidental undesired changes to the path.

(Alternatively, adding explicit tests that the URL reversal is equal to a particular path)

kolibri/core/auth/api.py
jamalex: (if we did end up keeping `password_hashed`, I note that I don't think I saw it being tested explicitly in the test suite)

jamalex (changes requested):
Great work! Teleportation is almost here!

Sorry for the verbose "think out loud" review style, but hope it's helpful!

kolibri/core/auth/utils.py
rtibbles: I am trying to remember why I didn't do that in the first place - I think it might have just been because we are only using this inside the sync task at the moment, and the cache is entirely thread local.

kolibri/core/auth/utils.py
rtibbles: Maybe a 'full removal' flag to choose which?

kolibri/core/auth/api.py
rtibbles: This is what prompted my question here:
> Should we be stopping signups in the public and private viewsets when the facility doesn't allow signups?

kolibri/core/auth/test/test_api.py
```diff
@@ -925,6 +925,36 @@ def test_sign_up_able_no_guest_access(self):
...
+    def test_no_sign_up_no_signups(self):
+        self.facility.dataset.learner_can_sign_up = False
+        self.facility.dataset.save()
+        response = self.post_to_sign_up(
+            {"username": "user", "password": DUMMY_PASSWORD}
+        )
+        self.assertEqual(response.status_code, status.HTTP_403_FORBIDDEN)
+        self.assertFalse(models.FacilityUser.objects.all())
+        self.facility.dataset.learner_can_sign_up = True
+        self.facility.dataset.save()
```
jamalex, line 937: Doesn't setup reset the data for each test anyway?

kolibri/core/auth/test/test_utils.py
```diff
@@ -50,3 +57,428 @@ class GetFacilityFailureTestCase(TestCase):
...
+    end_timestamp = datetime.datetime.now()
+    correct = False
+    time_spent = factory.LazyFunction(random.random)
+
+
+class TeleportUserTestCase(TestCase):
+    @classmethod
+    def setUpTestData(cls):
+        cls.facility = FacilityFactory.create()
+        i = 1
```
jamalex: I was unclear on why a variable is set for this -- was the original intention to loop?

jamalex (approved):
Looks good. Some little things with the tests that aren't blockers but could be addressed.

kolibri/core/auth/test/test_api.py
rtibbles: Nope - I think because it is using `setUpTestData`, but still not entirely sure why the transaction rollback isn't rectifying it.

kolibri/core/auth/test/test_utils.py
rtibbles: Uh, probably, yes.

PR #8093, opened 2021-05-19
In Facility task cards, only quote the facility name and not the ID fragment

### LINKED ISSUE #7123: name of facility is incorrectly wrapped in single-quotes
### Observed behavior

We use single-quotes to wrap user-generated text, like the name of facilities. We use parentheses to wrap the computer-generated IDs of facilities.

The task manager combines the two as `Sync '{facility} (`id`)'`

image

### Expected behavior

String should be `Sync '{facility}' (`id`)`

### User-facing consequences

potential confusion about what the actual name of the facility is

### Steps to reproduce

import a facility

### Context

0.14 beta 3

jonboiser:
## Summary

Fixes #7123

## Reviewer guidance

The tests should confirm the new behavior, but this can be manually confirmed by importing/syncing/deleting a facility in the app.

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/core/assets/src/mixins/taskStrings.js
+++ b/kolibri/core/assets/src/mixins/taskStrings.js
@@ -69,7 +69,7 @@ const taskStrings = createTranslator('TaskStrings', {
     context: 'Sync task status',
   },
   syncFacilityTaskLabel: {
-    message: "Sync '{facilityName}'",
+    message: 'Sync {facilityName}',
     context: 'Description of sync-facility task',
   },
   syncStepAndDescription: {
@@ -87,7 +87,7 @@ const taskStrings = createTranslator('TaskStrings', {
     context: 'Remove facility task status',
   },
   removeFacilityTaskLabel: {
-    message: "Remove '{facilityName}'",
+    message: 'Remove {facilityName}',
     context: 'Description of a remove-facility task',
   },
   removeFacilitySuccessStatus: {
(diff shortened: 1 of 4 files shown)


rtibbles (approved):

kolibri/core/assets/src/mixins/taskStrings.js
```diff
@@ -69,7 +69,7 @@ const taskStrings = createTranslator('TaskStrings', {
     context: 'Sync task status',
   },
   syncFacilityTaskLabel: {
-    message: "Sync '{facilityName}'",
+    message: 'Sync {facilityName}',
```
indirectlylit, line 72: This change and similar ones looks incorrect to me. From [the design system](https://design-system.learningequality.org/writing/#notes):

> Always wrap user-generated text in 'single quotes'.

indirectlylit (commented):
I think this PR may have taken away some single-quotes that were actually correct (cc @jonboiser @rtibbles)

kolibri/core/assets/src/mixins/taskStrings.js
jonboiser: `facilityName` is interpolated to something like `'My facility' (a1b2)`, so the full message should match what you're saying. Check out the tests.

indirectlylit: Ah, interesting – I didn't realize that `facilityNameWithId` was used inside other strings. I think that this pattern, while DRY, is technically incorrect for i18n strings because we're essentially using string concatenation to join two incomplete phrases.

A more technically correct string would include both inputs:

```
     message: `Sync '{facilityName}' ({id})`, 
```

My understanding is that this less-DRY format is easier for the translators to work with effectively.

(cc @radinamatic in case you have guidance and @jredrejo from the original change)

None of this is super critical – just wanted to flag for discussion and future work.

PR #8095, opened 2021-05-19
Add Bookmarks API

### LINKED ISSUE #8066: Add REST API for contentnode-bookmarks feature
## Use Cases

* Create/Delete bookmarks to ContentNodes
* View all bookmarked contentnodes as a list
* When a contentnode is bookmarked as a user, indicate that with a bookmark icon on its contentcard 
* When a contentnode is not bookmarked as a user, indicate that with an empty bookmark icon on its contentcard

There really isn't a use case for updating bookmarks or viewing an individual bookmark, so those abilities can be left out.

## Usage

Root URL

`/api/bookmarks`

### Create a bookmark

Request

```
POST /api/bookmarks/bookmarks
{
   "contentnode_id": "4d93d3e9-cd14-4d3f-8958-a7d24c3de807"
}
```

Response
```
{
   "contentnode_id": "4d93d3e9-cd14-4d3f-8958-a7d24c3de807",
   "is_bookmarked": true,
   "timestamp": "..."
}
```

### Delete a bookmark

Request
```
DELETE /api/bookmarks/bookmarks/4d93d3e9-cd14-4d3f-8958-a7d24c3de807
```

Response
```
{
   "contentnode_id": "4d93d3e9-cd14-4d3f-8958-a7d24c3de807",
   "is_bookmarked": false
}
```

### Retrieve all your bookmarks (get full ContentNode objects)

This is useful when the client needs to display full Content Cards for the bookmarks

```
GET /api/bookmarks/bookmarks
```

Response 
```
[
	{
		"contentnode_id": "4d93d3e9-cd14-4d3f-8958-a7d24c3de807",
		"title": "My bookmarked video",
		"description": "...",
		"thumbnail": "...",
		"is_leaf": false
	}
]
```

### Retrieve all your bookmarks (get full ContentNode objects)

This is a more minimal payload that can be used to annotate Content Nodes were from a different request. E.g., on the Learn Page, ContentNodes are requested separately. Then this list is used to mark the nodes that are already bookmarked.

The payload can be `Array<string>` or `Array<{contentnode_id}>` 

```
GET /api/bookmarks/bookmarks?ids_only=true
```

```
[
	"4d93d3e9-cd14-4d3f-8958-a7d24c3de807",
	"3f6a2357-9908-4ea4-a31d-c0f222d3cc8c",
	"5d4519ba-1b2a-47db-a530-bec11713d0b9",
	"38eb82d0-b5a1-4fe5-814e-67fb1aaf8828",
	"c97ba35f-57f7-4b12-9355-77b517dc143d",
	"af42d4fe-245e-476c-8511-b7e9914386cc",
	"07803f01-504a-42ee-b343-cf2189e66cc0",
	"f4dfdc9f-0509-484a-ac9e-971c4181c9e3"
]
```

nucleogenesis:
## Summary

Introduces a new module at `kolibri.core.bookmarks`.

Creates a new `Bookmark` model which associates a unique combination of `content_id` `channel_id` and `contentnode_id` with a `FacilityUser.id`. The uniqueness of the four values together is enforced at the Model level with `unique_together`.

Creates a `BookmarksViewSet(ValuesViewset)` class that implements the `create` `list` and `destroy` methods per the spec given in #8066 with consideration for @rtibbles comment on the necessary schema. Unimplemented HTTP verbs return 405 Method Not Allowed.

Users may delete, create and retrieve a list of their own Bookmarks. 

Includes tests for enforcing the protection against duplicates both at the Model and at the View level. In the Model, an `IntegrityError` will be raised and is something that should be found in testing. The client, however, will just get back the item they tried to create since it already exists - there is no error in trying to create a duplicate or to delete a non-existent Bookmark returned by the API.

### This does not make changes to the front-end

There is a massive refactor on deck for the content cards and this is laying some foundation for that work. As the content cards are developed they will have this available to use.

## References

Fixes #8066 

## Reviewer guidance

- [ ] Have I missed anything in the implementation details?
- [ ] Have I made proper use of `ValuesViewset` - is there a more ideal way to implement an API like this in Kolibri?

----

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated


kolibri/core/bookmarks/api.py
```diff
@@ -0,0 +1,74 @@
...
+        request.data.update({"facility_user": request.user})
+        obj, created = Bookmark.objects.get_or_create(**request.data)
+
+        serializer = BookmarksSerializer(obj)
+        return_status_code = HTTP_201_CREATED if created else HTTP_200_OK
+
+        return Response(serializer.data, status=return_status_code)
+
+    def destroy(self, request, pk):
+        try:
```
jonboiser, line 21: Does the actual destruction happen via the `ValuesViewset < ModelViewset`?

kolibri/core/bookmarks/api.py
```diff
@@ -0,0 +1,74 @@
...
+        return Response(serializer.data, status=HTTP_200_OK)
+
+    def create(self, request):
+        """
+        Overridden for two reasons:
+        1) Inject user into the data before creating it to avoid NOT NULL on facility_user
+        2) Provide meaningful HTTP responses accordingly, including 200 for duplicates
+        """
+        # POST requires these three to come from the client
+        required_keys = ["channel_id", "contentnode_id", "content_id"]
```
jonboiser: Is it possible to just require the `contentnode_id` (and make the others optional), since the others can be inferred from that ID? Also, this might be the only opportunity to do some validation that the content exists, since we don't have a ForeignKey relationship between Bookmarks and ContentNodes.

kolibri/core/bookmarks/serializers.py
```diff
@@ -0,0 +1,14 @@
+from rest_framework.serializers import ModelSerializer
+
+from .models import Bookmark
+
+
+class BookmarksSerializer(ModelSerializer):
```
rtibbles: I know this is not currently how we've done it, but I'd be perfectly happy to keep the serializer alongside the viewset class that it's used in, I don't see much value in how we have structured them right now.

kolibri/core/bookmarks/models.py
```diff
@@ -0,0 +1,30 @@
...
+
+import uuid
+
+from django.db import models
+from morango.models import UUIDField
+
+from kolibri.core.auth.models import FacilityUser
+
+
+class Bookmark(models.Model):
```
rtibbles: This needs to inherit from `from kolibri.core.auth.models import AbstractFacilityDataModel` in order to be syncable.

kolibri/core/bookmarks/models.py
```diff
@@ -0,0 +1,30 @@
...
+
+    id = UUIDField(primary_key=True, default=uuid.uuid4)
+    content_id = models.CharField(max_length=50, null=False, blank=False)
+    contentnode_id = models.CharField(max_length=50, null=False, blank=False)
+    channel_id = models.CharField(max_length=50, null=False, blank=False)
+    facility_user = models.ForeignKey(FacilityUser, blank=False)
+
+    morango_model_name = "bookmark"
+
```
rtibbles, line 21: Can use the BaseLogModel as a template for some of this, but it will be important to set the partition correctly in order to ensure single user syncing happens:

https://github.com/learningequality/kolibri/blob/develop/kolibri/core/logger/models.py#L73

Can use roughly the same partition function here, but no need to account for the anonymous case.

kolibri/core/bookmarks/models.py
```diff
@@ -0,0 +1,30 @@
...
+from kolibri.core.auth.models import FacilityUser
+
+
+class Bookmark(models.Model):
+
+    id = UUIDField(primary_key=True, default=uuid.uuid4)
+    content_id = models.CharField(max_length=50, null=False, blank=False)
+    contentnode_id = models.CharField(max_length=50, null=False, blank=False)
+    channel_id = models.CharField(max_length=50, null=False, blank=False)
+    facility_user = models.ForeignKey(FacilityUser, blank=False)
```
rtibbles: For most other models we have used `user` as the name of the foreign key to a facility user, would be good to keep this consistent.

kolibri/core/bookmarks/models.py
```diff
@@ -0,0 +1,30 @@
...
+from django.db import models
+from morango.models import UUIDField
+
+from kolibri.core.auth.models import FacilityUser
+
+
+class Bookmark(models.Model):
+
+    id = UUIDField(primary_key=True, default=uuid.uuid4)
+    content_id = models.CharField(max_length=50, null=False, blank=False)
```
rtibbles: Any reason not to have all of these as UUID fields - this will give another level of validation to ensure the values are correct.

kolibri/core/bookmarks/models.py
```diff
@@ -0,0 +1,30 @@
...
+
+from django.db import models
+from morango.models import UUIDField
+
+from kolibri.core.auth.models import FacilityUser
+
+
+class Bookmark(models.Model):
+
+    id = UUIDField(primary_key=True, default=uuid.uuid4)
```
rtibbles: Can skip the id field, as this is the default if you have inherited from `AbstractFacilityDataModel` above.

kolibri/core/bookmarks/api.py
```diff
@@ -0,0 +1,74 @@
...
+from .serializers import BookmarksSerializer
+from kolibri.core.api import ValuesViewset
+
+
+class BookmarksViewSet(ValuesViewset):
+    values = ("channel_id", "contentnode_id", "id", "content_id")
+    serializer_class = BookmarksSerializer
+    permission_classes = [IsAuthenticated]
+
+    def list(self, request):
```
rtibbles: You can avoid a lot of this reimplementation here by using the permissions system and permissioned filtering.

See here for the log model permissions: https://github.com/learningequality/kolibri/blob/develop/kolibri/core/logger/models.py#L61

Obviously would not need the anonymous permission here, and could possibly just use `IsOwn` until there's a use case for admins looking at other people's bookmarks.

You would set this on the model class, and then you could use: https://github.com/learningequality/kolibri/blob/develop/kolibri/core/logger/api.py#L129
```
    permission_classes = (KolibriAuthPermissions,)
    filter_backends = (KolibriAuthPermissionsFilter, DjangoFilterBackend)
```

In order to ensure the default filtering by user is happening.

kolibri/core/bookmarks/api.py
```diff
@@ -0,0 +1,74 @@
...
+            "channel_id", "content_id", "contentnode_id", "id"
+        )
+        serializer = BookmarksSerializer(bookmarks, many=True)
+
+        return Response(serializer.data, status=HTTP_200_OK)
+
+    def create(self, request):
+        """
+        Overridden for two reasons:
+        1) Inject user into the data before creating it to avoid NOT NULL on facility_user
```
rtibbles: You are reimplementing a lot of the logic for a serializer here - would be easier to overrride the `create` method of the serializer instead, and inject the user at that stage:
```
def create(self, validated_data):
    if "request" in self.context:
        user = self.context["request"].user
        validated_data["user"] = user
    try:
        return super(BookmarkSerializer, self).create(validated_data)
    except IntegrityError:
        return Bookmark.objects.get(**validated_data)
```

Alternatively, can just require that the user is submitted with the request, and lean on the permissions above to validate that the user creating the bookmark is allowed to do so (this then leaves open the possibility of an admin creating a bookmark for someone else).

kolibri/core/bookmarks/api.py
rtibbles: This seems reasonable, can simplify the serializer, do validation just that this specific node id exists, and then record all three values for future reference.

kolibri/core/bookmarks/api.py
rtibbles: Deferring to the permissions noted above will mean we can not overwrite this method - but for now, this is not actually deleting anything, no.

kolibri/core/bookmarks/api.py
```diff
@@ -0,0 +1,74 @@
...
+            return Response(
+                "Cannot delete Bookmark with id {} because it does not exist.",
+                status=HTTP_204_NO_CONTENT,
+            )
+        if bookmark.facility_user != request.user:
+            return Response(None, status=HTTP_403_FORBIDDEN)
+        else:
+            return Response({"success": True}, status=HTTP_204_NO_CONTENT)
+
+    def update(self, *args, **kwargs):
```
rtibbles: This is part of what I need to update for the values viewset - letting the base class not include methods, and only mixing in the ones that are actually required. I can update this once I've done that transfer from the Studio implementation.

kolibri/core/bookmarks/api.py
```diff
@@ -0,0 +1,74 @@
...
+                "Missing required: {}".format(", ".join(missing_keys)),
+                status=HTTP_400_BAD_REQUEST,
+            )
+
+        # Everything is now validated, let's make the bookmark and return a response
+        request.data.update({"facility_user": request.user})
+        obj, created = Bookmark.objects.get_or_create(**request.data)
+
+        serializer = BookmarksSerializer(obj)
+        return_status_code = HTTP_201_CREATED if created else HTTP_200_OK
```
rtibbles: If we don't really care about whether it's created or just already created, we could probably ignore this distinction - I added a suggestion for the serializer create method that would make this not needed.

rtibbles (changes requested):
A few things to update - can also simplify some things by using the permissions system and consolidating some logic in the serializer.

kolibri/core/bookmarks/api.py
nucleogenesis: No but I neither tested for that nor implemented it -_-

kolibri/core/bookmarks/migrations/0001_initial.py
```diff
@@ -0,0 +1,71 @@
...
+                        on_delete=django.db.models.deletion.CASCADE,
+                        to=settings.AUTH_USER_MODEL,
+                    ),
+                ),
+            ],
+        ),
+        migrations.AlterUniqueTogether(
+            name="bookmark",
+            unique_together=set(
+                [("user", "content_id", "contentnode_id", "channel_id")]
```
rtibbles: Need to update this migration as the unique together is now only on the user and the contentnode_id.

rtibbles (approved):

PR #8102, opened 2021-05-21
7921 klabeledicon related UI fixes

### LINKED ISSUE #7921: Fix widgets that are using KLabeledIcon in the Facility Page + Device Page
Here are some breakages in layouts using `KLabeledIcon` after 
https://github.com/learningequality/kolibri-design-system/pull/195

I think these can probably be fixed by moving the misaligned elements into the `label` slot.

Screenshots in Safari

**http://localhost:8000/en/facility/#/data**

Sync title alignment

**http://localhost:8000/en/facility/#/users**

Role label

**http://localhost:8000/es-419/device/#/content/channels/95a52b386f2c485cb97dd60901674a98**

Coach labels

image

**http://localhost:8000/es-419/device/#/content/manage_channel/a791922463b2587f9a8f3987efe8b7c9**

Unlisted channel (lock) icon

image

nucleogenesis:
## Summary

Fixes the misc visual issues noted in #7921 

A complication with `KLabeledIcon` is that it is currently set to `width: 100%` - for reasons. Those reasons are to do with how it acts in certain circumstances. Since this is only fixed with CSS in one place here right now, I think that it's fine and, in some cases, to be expected. 

## References

Fixes #7921 

## Reviewer guidance

Are the issues fixed - please try in Safari if you can.

----

### DIFF
--- a/kolibri/core/assets/src/views/sync/FacilityNameAndSyncStatus.vue
+++ b/kolibri/core/assets/src/views/sync/FacilityNameAndSyncStatus.vue
@@ -5,13 +5,15 @@
       <h2 class="name">
         <KLabeledIcon icon="facility">
           {{ facility.name }}
+          <template #iconAfter>
+            <KIcon
+              v-if="facility.dataset.registered"
+              ref="icon"
+              icon="registered"
+              :style="{ fill: $themePalette.green.v_500 }"
+            />
+          </template>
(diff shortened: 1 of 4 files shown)


jonboiser (approved):
Looks good! Hopefully that's all of them 🤞

PR #8117, opened 2021-05-26
Add vacuum for morango tables in Postgresql databases

### LINKED ISSUE #8080: Add Postgres support to current vacuum job, targeting key Morango tables
Our scheduled vacuum task [currently only works on sqlite](https://github.com/learningequality/kolibri/blob/develop/kolibri/core/deviceadmin/utils.py#L187). Particularly with syncing now becoming more common, and Postgres usage as well, we end up with table bloat in the temporary Morango tables.

There's a [Morango issue to be able to clear old data out of these tables](https://github.com/learningequality/morango/issues/104). We should, in parallel, add support for VACUUMing these specific tables in Postgres, and have the garbage collection command run first, followed by the vacuuming.

The specific tables to ensure are vacuumed are those for the `Buffer` and `RecordMaxCounterBuffer` models.

jredrejo:
## Summary
Performs vacuum on morango tables if the kolibri is using a Postgresql database
It also renames current `vacuumsqlite` command to be just `vacuum`

Also, for the case of morango tables growing too much, executing
`kolibri manage vacuum --full=true`
will do a full vacuum on the `morango_recordmaxcounterbuffer` and `morango_buffer` tables

## References
Closes: #8080 

## Reviewer guidance
Installing a postgresql db and setting options.ini to be something like this 
```
[Database]
DATABASE_ENGINE=postgres
DATABASE_HOST = "127.0.0.1"
DATABASE_USER = "learningequality"
DATABASE_PASSWORD = "kolibri"
DATABASE_NAME="kolibri"
```
(replacing the options with those that match your database)

run `kolibri manage vacuum `and check that no problems appear. 
In case `psql` is available, this command `SELECT pg_size_pretty( pg_total_relation_size('morango_buffer') ); `can be used to check the size of the tables before and after performing the vacuum

----

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/core/deviceadmin/utils.py
+++ b/kolibri/core/deviceadmin/utils.py
@@ -183,7 +183,7 @@ def search_latest(search_root, fallback_version):
         return os.path.join(search_root, newest)
 
 
-def perform_vacuum(database=db.DEFAULT_DB_ALIAS):
+def perform_vacuum(database=db.DEFAULT_DB_ALIAS, full=False):
     connection = db.connections[database]
     if connection.vendor == "sqlite":
         try:
@@ -207,14 +207,20 @@ def perform_vacuum(database=db.DEFAULT_DB_ALIAS):
         else:
             logger.info("Sqlite database Vacuum finished.")
     elif connection.vendor == "postgresql":
-        morango_models = [
-            m
-            for m in apps.get_models(include_auto_created=True)
-            if "morango.models" in str(m)
-        ]
+        if full:
+            morango_models = ("morango_recordmaxcounterbuffer", "morango_buffer")
+        else:
+            morango_models = [
+                m
+                for m in apps.get_models(include_auto_created=True)
+                if "morango.models" in str(m)
+            ]
         cursor = connection.cursor()
         for m in morango_models:
-            cursor.execute("vacuum analyze {};".format(m._meta.db_table))
+            if full:
+                cursor.execute("vacuum full analyze {};".format(m))
+            else:
+                cursor.execute("vacuum analyze {};".format(m._meta.db_table))
         connection.close()
 
 
--- a/kolibri/core/deviceadmin/utils.py
+++ b/kolibri/core/deviceadmin/utils.py
@@ -7,6 +7,7 @@ from datetime import datetime
 from datetime import timedelta
 
 from django import db
+from django.apps import apps
 from django.conf import settings
 
 import kolibri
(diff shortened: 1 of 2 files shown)


kolibri/core/deviceadmin/utils.py
```diff
@@ -205,6 +206,16 @@ def perform_vacuum(database=db.DEFAULT_DB_ALIAS):
             logger.error(new_msg)
         else:
             logger.info("Sqlite database Vacuum finished.")
+    elif connection.vendor == "postgresql":
+        morango_models = [
+            m
+            for m in apps.get_models(include_auto_created=True)
+            if "morango.models" in str(m)
+        ]
```
jamalex: Is the time it takes to vacuum a table proportional to the size of the table, or to how much junk space there is "to vacuum"?

The only Morango tables with much churn are the [`Buffer`](https://github.com/learningequality/morango/blob/master/morango/models/core.py#L412) and the [`RecordMaxCounterBuffer`](https://github.com/learningequality/morango/blob/master/morango/models/core.py#L520).

While it shouldn't hurt to vacuum the others as well, some of them (e.g. the `Store`) can be very large, but under normal circumstances not have any churn at all (they're added and updated, but never deleted), and since we're running this vacuuming automatically, we'll want to keep it as efficient as we can.

jredrejo: The time it will take will depend on how disorganized the tables are. Usually large tables are more disorganized than small ones but the size of the table is not an accurate rule of thumb. Such as this is, the size reduction won't be as big as executing a _full vacuum_, but it will reorganize the data to be performant. In any case with this implementation the tables are not locked while vacuum and the application will start normally. The user should not notice anything.

Without adding _full_ to the vacuum command this is a non-blocking command. I'm thinking of adding a `--full `option (to be false by default) to the command so in case we want to execute it in a cron, it would be a complete vacuum of the tables. That would produce table locks and will reduce much more the size of the tables, but it would not be executed when kolibri starts, only on demand by a command. 

What do you think? having a possiblity to run it under cron for a full vacuum would be useful for us? In that case, limiting it to the two tables you mention makes more sense as they won't be available while vacuum is executed.

rtibbles: Have discussed this with @jredrejo and it seems like his plan is good!

1) The non-full will run on all tables because it is non-locking
2) Will add a 'full' flag that will run only on those two tables because that will reclaim disk space for tables that grow and then shrink again.

jamalex: That sounds good, yeah -- for Cloud Kolibri and KDP we'll want to have that cron job running regularly to keep things in check (as there's a ton of churn on those tables).

jredrejo: Done

rtibbles (approved):

rtibbles:
Test failures are not related to this PR.

PR #8118, opened 2021-05-26
Add '@vue/composition-api' plugin; refactor SelectAddressForm with `setup()`; implement #7345

### LINKED ISSUE #7345: Make the last used address/device pre-selected in following imports
### Observed behavior
[Follow-up from](https://github.com/learningequality/kolibri/pull/7301#pullrequestreview-448553104) #7301 

When _Select network address_ modal is opened, it by default has pre-selected the last address that was added, even if the user has used a different one from the list since that last addition. 

### User-facing consequences
Better UX, less clicks especially if the import address is used frequently.

### Proposed Solution (What we estimated)
- [ ] Use `lockr` to store the last address used - then ensure that when the modal is opened the correct item is selected.

### Steps to reproduce
…

### Context

 * Kolibri version: 0.14

### LINKED ISSUE #7785: When importing Facility Data from peers, Network Locations that are not Kolibri instances should not be shown
### Observed behavior

Because code is shared between the Facility-import and Content-Import workflows, you see the same list of network locations. However, instances of Studio only have content and should only be shown in the Content-Import workflow.

image

### Expected behavior

1. Kolibri Instances are shown on both the Facility-import and Content-Import workflows
2. Studio Instances are only shown on the Content-Import workflow

It is expected as part of the estimate that we will separate the facility sync and content import/export flow - refactoring and cleaning up this part of the code in the process.

### Steps to reproduce

1. In a Content-Import workflow, add a URL like `hotfixes.studio.learningequality.org` as a new network location.
2. Enter the Facility-import workflow and you will see the Studio URL; select it.
3. You will not see any Facilities

### Context

0.14.6

jonboiser:
## Summary

1. Adds `@vue/composition-api@1.0.0-rc9` (seems fine to me and the final should be out sometime in Q3), so we can finally use it and take advantage of its benefits.
2. Adds `vueuse`, which is a large library of utility functions that work with the Vue composition api. 
2. Factors out all the network location code out of `SelectAddressForm` into two "usables": `useDynamicAddresses` and `useSavedAdddresses`, and moving the object exposed from the usables into `SelectAddressForm`'s `setup` function. Why is this cool? Because now all the code related to the two types of addresses can live in their own module, instead of being interlaced in lots of different parts of the Vue component. There are opportunities for cleaning up the API from the usables; I just wanted to do a direct 1-to-1 translation from Vue option -> usable without too much extra refactoring. This is also an opportunity to increase the test coverage of this code without having to worry about the extra complexities of unit testing the Component they live in.
3. Fixes a bug where if you add an address, then select a different address right after, then the radio button for the added address will be selected after the polling for the dynamic addresses.
4. Implements #7345 by using https://vueuse.org/core/usestorage/#usage, which lets you create a reactive proxy to a local storage value.
5. ~turns on the ESLint 'no-shadow' rule (which can cause a lot of confusion when you don't see it 😠 )~ Had to revert it bc there's a lot of this in the codebase (needs followup issue)

**Video of unreported bug that was fixed:**

CleanShot 2021-05-26 at 11 10 10

## References

Fixes #7345, and hopefully makes it easier to "fix" #7785 (which seems to be an issue with how static addresses are being filtered within `useSavedAddresses`)

## Reviewer guidance

1. Does the UI work
2. Review the code carefully. Within `setup` and inside the usable's implementation, notice how I need to access `ref.value`, but outside of these functions (e.g. inside the template or `options.computed`, I don't need to do that (since _those_ variables are now reactive proxies). Are there any places in the code where I overlook those caveats?

----

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated
Use Vue Composition API plugin

### DIFF
--- a/kolibri/core/assets/src/views/sync/SelectAddressModalGroup/useDynamicAddresses.js
+++ b/kolibri/core/assets/src/views/sync/SelectAddressModalGroup/useDynamicAddresses.js
@@ -5,6 +5,7 @@ import {
   onBeforeUnmount,
   getCurrentInstance,
 } from '@vue/composition-api';
+import { get, set, useIntervalFn } from '@vueuse/core';
 import { fetchDynamicAddresses } from './api';
 
 const Stages = Object.freeze({
@@ -17,7 +18,6 @@ export default function useDynamicAddresses(props) {
   const { fetchAddressArgs, discoverySpinnerTime } = props;
   const addresses = ref([]);
   const stage = ref('');
-  const intervalId = ref('');
   const discoveredAddressesInitiallyFetched = ref(false);
   const $parent = getCurrentInstance().proxy.$parent;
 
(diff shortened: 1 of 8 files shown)


nucleogenesis (commented):
Do you think there are any notable changes to how we'd approach testing the components? It might help with team adoption in general to see how testing is affected by this change even if it's just a matter of composition-friendly code being bundled into modules with smaller and more specific scope being naturally easier to test.

jonboiser:
> Do you think there are any notable changes to how we'd approach testing the components? It might help with team adoption in general to see how testing is affected by this change even if it's just a matter of composition-friendly code being bundled into modules with smaller and more specific scope being naturally easier to test.

I don't think anything really needs to change in terms of testing the component. You _could_ test the "usables" and the components separately (and mock usable stuff in the context of the component test), but that would be a weaker test, since the two specs could get out of sync.

But pragmatically, what I would do is write the "usable" first, give it a stable API, and give it high test coverage. Then integrate it into a component, but only cover a subset of the component specific code, especially if mocking the XHR stuff is a pain. Then together, you have pretty good test coverage and the gaps can be filled in over time.

kolibri/core/assets/src/views/sync/SelectAddressModalGroup/useDynamicAddresses.js
```diff
@@ -0,0 +1,81 @@
...
+  ref,
+  computed,
+  onBeforeMount,
+  onBeforeUnmount,
+  getCurrentInstance,
+} from '@vue/composition-api';
+import { get, set, useIntervalFn } from '@vueuse/core';
+import { fetchDynamicAddresses } from './api';
+
+const Stages = Object.freeze({
```
micahscopes, line 11: nice!  you learn something new every day, and today I learned about this

micahscopes (approved):
I think this is great.  I also tried it and it seemed to work.  I have a few points of feedback:
- I've longed for this kind of thing
  - Now that it's here, I realize how much of a shift it is
  - It'd be great to do almost a mini workshop hack session for people on the team to understand how/when to use this stuff and how to think about it, and also how to understand the "use" metaphor
- I had a hard time getting this branch building at first, and needed to try all kinds of things like `make clean`, switching to `develop` and back and deleting `node_modules`.  I don't think it has anything to do with this PR, but just a heads up that I ran into these issues.

PR #8121, opened 2021-05-27
Error properly on startup

### LINKED ISSUE #8119: Kolibri fails to handle errors during startup and instead, itself errors
## Observed behavior
When making changes to the zeroconf code that resulted in an error, a further error occurred during the handling of said error.

## Errors and logs
```
ERROR    Error in 'START_ERROR' listener <bound method ProcessBus.START_ERROR of <magicbus.process.ProcessBus object at 0x7fda2bd12390>>
  File "/home/richard/.virtualenvs/kolibri/lib/python2.7/site-packages/magicbus/base.py", line 273, in publish
    result = listener(*args, **kwargs)
  File "/home/richard/.virtualenvs/kolibri/lib/python2.7/site-packages/magicbus/process.py", line 104, in START_ERROR
    self.transition('EXITED')
  File "/home/richard/.virtualenvs/kolibri/lib/python2.7/site-packages/magicbus/base.py", line 186, in transition
    output.append(self._transition(next_state))
  File "/home/richard/.virtualenvs/kolibri/lib/python2.7/site-packages/magicbus/base.py", line 223, in _transition
    self._transition(self.errors[newstate], *sys.exc_info())
  File "/home/richard/.virtualenvs/kolibri/lib/python2.7/site-packages/magicbus/base.py", line 215, in _transition
    return self.publish(newstate, *args, **kwargs)
  File "/home/richard/.virtualenvs/kolibri/lib/python2.7/site-packages/magicbus/base.py", line 291, in publish
    raise exc
ChannelFailures: TypeError('set_pid_file() takes exactly 2 arguments (5 given)',)
```

## Expected behavior
Errors should be handled and not result in further errors during handling.

This seems to be a result of binding the `set_pid_file` method of `PIDPlugin` to every status in the `magicbus` status map. This includes the `START_ERROR`, `EXIT_ERROR` etc states - which have a different callback signature of the exception traceback. 

To fix this, we should only bind the function to handle the states that just set a status.

rtibbles:
## Summary
* Resolves a regression caused by #8026 whereby errors occurring during plugin life cycle events would then cause an error in their handling.
* Does this by letting the handler function take arbitrary *args, to match the signature of the magicbus error state handlers: https://github.com/cherrypy/magicbus/blob/main/magicbus/process.py#L101
* Simplifies service shutdown by only makes the call to shutdown the worker threads once, and makes it blocking
* Adds a handle_SIGINT method to give different logging when a SIGINT signal is called versus SIGTERM

## References
Fixes #8119 

## Reviewer guidance
Make sure Kolibri starts up properly and then reacts properly to CTRL-C, or throwing any error during a call to register the zero conf service, or start one of the worker threads.

Noting that I have observed recently that a command prompt will display and then further logging text will happen after the CTRL-C - I am at a loss to work out how to prevent this, as the way that magicbus handles signal handlers is the way that every StackOverflow post recommends to try to handle this, so I am at a loss as to how to resolve it.

----

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/utils/server.py
+++ b/kolibri/utils/server.py
@@ -277,17 +277,14 @@ class ServicesPlugin(SimplePlugin):
         register_zeroconf_service(port=port or self.port)
 
     def STOP(self):
-        from kolibri.core.tasks.main import scheduler
-
-        scheduler.shutdown_scheduler()
-        if self.workers is not None:
-            for worker in self.workers:
-                worker.shutdown()
         from kolibri.core.discovery.utils.network.search import (
             unregister_zeroconf_service,
         )
 
         unregister_zeroconf_service()
+        from kolibri.core.tasks.main import scheduler
+
+        scheduler.shutdown_scheduler()
 
         if self.workers is not None:
             for worker in self.workers:
--- a/kolibri/utils/server.py
+++ b/kolibri/utils/server.py
@@ -324,7 +324,7 @@ class PIDPlugin(SimplePlugin):
             handler.priority = 10
             self.bus.subscribe(bus_status, handler)
 
-    def set_pid_file(self, status):
+    def set_pid_file(self, status, *args):
         """
         Writes a PID file in the format Kolibri parses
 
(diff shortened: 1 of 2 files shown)


jredrejo (approved):
Tested without issues, and code LGTM

PR #8122, opened 2021-05-27
Zeroconf broadcast of SoUD status

### LINKED ISSUE #8077: Add facility device or SoUD flag to device info (for zeroconf broadcast)
For automatic background syncing (#7964), devices will need an efficient way to know what other devices and their associated datasets are on the local network. The most efficient way to facilitate this sharing of information is via the zeroconf broadcast itself, because it's multicast (i.e. many-to-many), rather than by requiring an additional HTTP request, which would lead to an n^2 barrage of 1:1 requests between every pair of devices on the network.

The simplest place to add this information is in the results of [`get_device_info()`](https://github.com/learningequality/kolibri/blob/develop/kolibri/core/public/utils.py#L9), which is what's called to get the info that's included in the zeroconf broadcasts. This has the beneficial side effect of also including this info in the response to the `/api/public/info/` endpoint, which can be used as a fallback when devices are connecting outside of a local network context, or when an address was added manually.

The proposed new field to add would be:

~`"facilities": {"<facility_id_1>": "*", "<facility_id_2>": ["<single_user_id_1>", "<single_user_id_2>"]}`~

`"soud": <boolean>`

rtibbles:
## Summary
* Creates a device setting for whether a device is a subset of users device or not
* Broadcasts that fact in the zeroconf device info
* Stores that fact in the NetworkLocation model
* Make sure that zeroconf service properties handling does not break when dealing with Boolean values

## References
Fixes #8077

## Reviewer guidance
Does the server start properly?
Is the subset of users devices flag set in the zeroconf information?
Does it get persisted into the NetworkLocation model when a DynamicNetworkLocation is added?

----

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated


jredrejo (commented):
Code works fine and another kolibri in the network running 0.14.5 detected this new kolibri instance with its SoUD field perfectly.


```
INFO     Kolibri instance '9ad9ee5e19f80c03774f7d3d29dffdc3' joined zeroconf network; service info: {'id': '9ad9ee5e19f80c03774f7d3d29dffdc3', 'ip': '192.168.0.44', 'local': False, 'port': 8000, 'host': '9ad9ee5e19f80c03774f7d3d29dffdc3.kolibri.local', 'base_url': 'http://192.168.0.44:8000/', 'self': False, 'application': 'kolibri', 'kolibri_version': '0.15.0.dev0+git.20210527155757', 'instance_id': '9ad9ee5e19f80c03774f7d3d29dffdc3', 'device_name': 'jose-xps', 'operating_system': 'Linux', 'subset_of_users_device': 'FALSE'}
```

jredrejo (approved):

PR #8126, opened 2021-05-31
Create API for a subset of users device to request permission to sync

### LINKED ISSUE #8060: Create utility and API endpoint for a subset of users device to request permission to sync with a full facility device
In order to prevent sync stampedes in cases where multiple subset of users devices join a network in quick succession where a full facility device is present, we need functionality to allow the SoUDs to request that they be able to sync.

Requirements
-------------------
* [ ] Create an endpoint that will:
  * [ ] Return an error code when accessed on an SoUD - should return status code 418
  * [ ] Accept a POST request to update its place in the queue:
    * [ ] the client can optionally specify an identifier it has already for a place
    * [ ] an identifier for the place in the queue if one was not sent with the request
    * [ ] signal if a sync should be initiated now
    * [ ] the keep alive interval at which the client should check in to maintain its place in the queue (the actual expiry of the queue place should be bigger than the signalled interval to allow some buffer)
* [ ] Create a utility function for the client to call which will:
  * [ ] Queue an async task to make a request to the above endpoint
  * [ ] In this async task, get the information from the server and:
    * [ ] If the server says "sync now" immediately queue a sync task for the server
    * [ ] if the server responds with an identifier and interval, schedule itself to run again in the future with that identifier as an argument, at the interval specified

jredrejo:
## Summary

This PR has two APIs, one for the  server to handle client requests and another for the clients to handle the response from the server.

### On the server
The API is accessed via a viewset that:

  * [ ] Return an error code when accessed on an SoUD - should return status code 418
  * [ ] Accept a POST request to update its place in the queue:
    * [ ] the client can optionally specify an identifier it has already for a place
    * [ ] an identifier for the place in the queue if one was not sent with the request
    * [ ] signal if a sync should be initiated now
    * [ ] the keep alive interval at which the client should check in to maintain its place in the queue (the actual expiry of the queue place should be bigger than the signalled interval to allow some buffer)

### On the client
There's a function called  `kolibri.core.public.utils.begin_request_soud_sync` that will begin the process described in #8060
  * [ ] Queue an async task to make a request to the above endpoint
  * [ ] In this async task, get the information from the server and:
    * [ ] If the server says "sync now" immediately queue a sync task for the server
    * [ ] if the server responds with an identifier and interval, schedule itself to run again in the future with that identifier as an argument, at the interval specified
    * [ ] In case of network connection problems with the server, it retries up to 10 times with different intervals (pseudo-randomized  to avoid multiple clients connecting at the same time) 
 
## References
Closes: #8060 

## Reviewer guidance
As this has been developed using TDD

- Do new tests make sense?
-  Are all the possible cases in the server covered by the tests?
-  Are all the possible cases in the client covered by the tests?
- Do tests pass?

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated


kolibri/core/public/utils.py
```diff
@@ -28,3 +42,84 @@ def get_device_info():
...
+        else response.content
+    )
+    server_response = json.loads(response_content or "{}")
+    if response.status_code == status.HTTP_404_NOT_FOUND:
+        return  # Request done to a server not owning this user's data
+
+    if response.status_code == status.HTTP_200_OK:
+        if server_response["action"] == SYNC:
+            # TODO: queue a sync task in the server, when this is developed
+            logger.info("Enqueuing a sync task for user {}".format(user))
```
rtibbles: This exists!

Just need to enqueue a sync task like this: https://github.com/learningequality/kolibri/blob/develop/kolibri/core/tasks/api.py#L944 but with the added `user` argument of the user id.

kolibri/core/public/api.py
```diff
@@ -146,3 +156,77 @@ def get_public_file_checksums(request, version):
...
+            allow_sync = False
+        return (allow_sync, data)
+
+    def create(self, request):
+        SyncQueue.clean_stale()  # first, ensure not expired devices are in the queue
+        device_info = get_device_info()
+        if device_info["subset_of_users_device"]:
+            content = {"I'm a Subset of users device": "Nothing to do here"}
+            # would love to use HTTP 418, but it's not fully usable in browsers
+            return Response(content, status=status.HTTP_404_NOT_FOUND)
```
rtibbles: Could we maybe do something other than 404 here, just to indicate that this was found, but this device is not to be synced with?

jredrejo: Replaced by a 400

kolibri/core/public/api.py
```diff
@@ -146,3 +156,77 @@ def get_public_file_checksums(request, version):
...
+            ).count()
+        return Response(queue)
+
+    def check_queue(self):
+        current_transfers = TransferSession.objects.filter(active=True).count()
+        if current_transfers <= MAX_CONCURRENT_SYNCS:
+            allow_sync = True
+            data = {"action": SYNC}
+        else:
+            polling = current_transfers + SyncQueue.objects.all().count()
```
rtibbles: Might be good to add a constant here for the 'waiting time' and use this as a multiplier, or even be calculating some sort of syncing estimate (we'd need to check in with @bjester and @jamalex to see if that would be feasible).

kolibri/core/public/api.py
```diff
@@ -146,3 +156,77 @@ def get_public_file_checksums(request, version):
...
+        facilities = Facility.objects.all()
+        queue = {}
+        for facility in facilities:
+            queue[facility.id] = SyncQueue.objects.filter(
+                user__facility=facility
+            ).count()
+        return Response(queue)
+
+    def check_queue(self):
+        current_transfers = TransferSession.objects.filter(active=True).count()
```
rtibbles, line 186: There doesn't seem to be any concept of ordering here - if a sync is not currently active, and you happen to have checked in, then you will be the one to sync next.

I think my assumption was that we would do a first in first out queue, and so you would only sync if no one else was ahead of you. This might be an argument for simply using an autoincrementing integer PK instead of a UUID, and we can use that for the ordering too.

The more complex solution might involve the server actively prioritizing which items to sync, in which case having a separate 'ordering' field may be better, so that the server could decide to queue up one sync before another.

Another alternative would be to use the datetime field as the primary ordering, could then also have a simple 'priority' flag to bump some syncs ahead of the other.

rtibbles (commented):
Looking solid - but I think we should be leveraging the ordering in the queue to determine who syncs when.

kolibri/core/public/api.py
jredrejo: According to our previous conversations, the requests will receive a SYNC response most of the times . This queue should be an extraordinary option for cases where the server has too many syncs on process (I set that limit on 10 in the code and we can discuss this limit but I don't think there's a way to estimate this number because the amount of data to be synced will vary a lot between cases of use, depending on how long the user has been unsynced and how much information the coach and the student have added in the meantime, what's impossible to estimate).

So, most of the times the request will not be queued. If the server is not in the limit, no matter if there are requests enqueued because waiting for those users to end their `keep_alive` time would be contradictory and would cause more problems of the kind we're trying to avoid.

For the rest of the cases: when an user updates his request and he can't sync yet. This is the only case where I think we could do some kind of change, returning a `keep_alive` time based on its position in the queue, ordered by the `datetime` field. I will do a proof of concept for it but I'm not sure it's worth if the queue is suppossed to be mostly empty or with few members.

jredrejo: Done in https://github.com/learningequality/kolibri/pull/8126/commits/3acc7e5de23631c66da72a90340dd6a1faf39b3b

kolibri/core/public/utils.py
jredrejo: It's done, needed to refactor part of the tasks/api.py to make it independent of the request but all tests (old and new) pass now

kolibri/core/public/api.py
```diff
@@ -146,3 +156,90 @@ def get_public_file_checksums(request, version):
         json.dumps({"id": error_constants.NOT_FOUND, "metadata": {"view": ""}}),
         content_type="application/json",
     )
+
+
+def position_in_queue(id):
+    qs = SyncQueue.objects.all().order_by("datetime").values("id")
```
rtibbles: Hopefully the queue is never that big so querying the whole queue wouldn't be an issue, but could possibly do this with a query to get the datetime for the id, and then just query the count of items that have a datetime less than that?

kolibri/core/public/api.py
```diff
@@ -146,3 +156,90 @@ def get_public_file_checksums(request, version):
...
+            ).count()
+        return Response(queue)
+
+    def check_queue(self, pk=None):
+        current_transfers = TransferSession.objects.filter(active=True).count()
+        if current_transfers <= MAX_CONCURRENT_SYNCS:
+            allow_sync = True
+            data = {"action": SYNC}
+        else:
+            polling = current_transfers + SyncQueue.objects.all().count()
```
rtibbles: Should we only do this if `pk is None`?

kolibri/core/public/api.py
```diff
@@ -146,3 +156,90 @@ def get_public_file_checksums(request, version):
...
+            data = {
+                "action": QUEUED,
+                "keep_alive": polling,
+            }
+            allow_sync = False
+        return (allow_sync, data)
+
+    def create(self, request):
+        SyncQueue.clean_stale()  # first, ensure not expired devices are in the queue
+        device_info = get_device_info()
```
rtibbles: Can also use `get_device_setting` utility here just to get the single value.

https://github.com/learningequality/kolibri/blob/814e00f9400a810986c76a47203c094fd6f3f42c/kolibri/core/device/utils.py#L22 (with a default of `False`)

Like it is used in the device info function: https://github.com/learningequality/kolibri/blob/develop/kolibri/core/public/utils.py#L16

rtibbles (approved):
All looks good to me, a couple of comments, and would be good to have some manual QA on the existing functionality that has been slightly refactored.

radinamatic:
DEB asset installed on Ubuntu 18.04 and 20.04 VMs, on both devices it failed to import facilities from Kolibri Beta, Kolibri Demo and UNW staging servers, but discovery, import and syncing between the 2 devices in the local network seems to be working without problems.

Logs:
[Ubuntu1804.zip](https://github.com/learningequality/kolibri/files/6659437/Ubuntu1804.zip)
[Ubuntu2004.zip](https://github.com/learningequality/kolibri/files/6659439/Ubuntu2004.zip)

VM | Beta | Demo | UNW staging
---|---|---|---
Ubuntu 18.04 | Ubuntu18 04 (start)  Running  - Oracle VM VirtualBox_018 | Ubuntu18 04 (start)  Running  - Oracle VM VirtualBox_019 | Ubuntu18 04 (start)  Running  - Oracle VM VirtualBox_023
Ubuntu 20.04 | Importing facility from Ubuntu 18.04  ✔️  Ubuntu20 04 (start)  Running  - Oracle VM VirtualBox_025 | - Importing facility from Beta and UNW ❌ <br> - Syncing facility with Ubuntu 18.04 ✔️  Ubuntu20 04 (start)  Running  - Oracle VM VirtualBox_030

pcenov:
@radinamatic on my end I tested the https://buildkite.com/learningequality/kolibri-windows/builds/1236#1db9de8f-42d4-4111-a187-81d031948eb7 build on Windows 7 and Windows 10 and again on both devices the import from Kolibri Beta and Kolibri Demo servers fails.
Discovery, import and syncing between the devices in the local network is working fine as pointed out by you.

Note that in both Windows 7 and Windows 10 when I start Kolibri it opens an incorrect port which I have to manually change to the correct one (http://127.0.0.1:8080):
Win 7: http://127.0.0.1:58451/
Win 10: http://127.0.0.1:49716/


Logs:

[Win7Logs.zip](https://github.com/learningequality/kolibri/files/6663417/Win7Logs.zip)
[Win10DemoImportFail.zip](https://github.com/learningequality/kolibri/files/6663421/Win10DemoImportFail.zip)
[Win10Logs.zip](https://github.com/learningequality/kolibri/files/6663422/Win10Logs.zip)
[Win10MultipleImportAndSyncTests.zip](https://github.com/learningequality/kolibri/files/6663424/Win10MultipleImportAndSyncTests.zip)

radinamatic:
@pcenov Thank you! 💯 

Windows launcher ports issue is still pending, yes. 
cc @jonboiser

pcenov:
Yet another note: both ZukVillage and Demo Facility imports are failing in Kolibri 14.7 as well:

kolibri/core/public/api.py
jredrejo: no, it should be done always. If pk is None , it's creating the element in the queue and we should provide a minimal keep_alive time. Originally it would be the length of the queue + numbers of syncs in progress, but after the conversation we had last Monday when we decided to enqueue if there's one sync in progress it should be something like 
`polling = minimal_ttl + SyncQueue.objects.all().count()`

rtibbles: Oh, my thought was that in the case that the `pk is not None` we throw this value away inside the `if` statement below, so there's no need to call this count unless the `pk is None`.

kolibri/core/public/api.py
jredrejo: I thought using get_device_info would be preferable as it considers the case when the device is not provisioned. If that can't ever happen, perfect, I'll change it.

rtibbles: Can just use the `default` kwarg to return False in that case - means we're only grabbing the info we need.

kolibri/core/public/api.py
jredrejo: I don't think performance will be a problem here, if the queue gets thousands of request we'll have to look for different problems . 
In any case, there's an `updated` field https://github.com/jredrejo/kolibri/blob/sync_endpoint/kolibri/core/device/models.py#L198 that's a float number and should be as quick as the id when sorting. Adding an index to `updated` and ordering by it should be more or less as performant as what you propose with less code (and likely less db queries) . Does it sounds good ?

kolibri/core/public/api.py
jredrejo: Done

kolibri/core/public/api.py
jredrejo: Ahh, I understood what you meant, done

kolibri/core/public/api.py
jredrejo: Done

kolibri/core/public/api.py
```diff
@@ -199,8 +209,8 @@ def check_queue(self, pk=None):
 
     def create(self, request):
         SyncQueue.clean_stale()  # first, ensure not expired devices are in the queue
-        device_info = get_device_info()
-        if device_info["subset_of_users_device"]:
+        is_SoUE = get_device_setting("subset_of_users_device", False)
```
rtibbles:
```suggestion
        is_SoUD = get_device_setting("subset_of_users_device", False)
```

kolibri/core/public/api.py
```diff
@@ -199,8 +209,8 @@ def check_queue(self, pk=None):
...
+        if is_SoUE:
```
rtibbles:
```suggestion
        if is_SoUD:
```

rtibbles:
We have confirmed that the issues are specific to issues on the servers, not to changes in this PR!

radinamatic:
Facility import and syncing in local network seems to work well between this build and Kolibri 0.14 (both on Ubuntu devices).

Ubuntu18 04 (start)  Running  - Oracle VM VirtualBox_032
`DistrictSchool16 (6a6e)` runs on Kolibri 0.14

radinamatic:
Facility import and sync also seems to work (no user facing errors) between this build and Kolibri 0.13. However, learner progress from the 0.15 device is not reflected on the 0.13 device after the sync, but the learner progress from the 0.13 device is displaying in the Coach dashboard on the 0.15 device.

The only error I can see in the 0.13 terminal during sync:
```
WARNING  Validation error for Lesson with id 2951b045cd25d754c7b5c8590754fcf3: ['FacilityUser matching query does not exist.']
WARNING  Validation error for Lesson with id 90ab284864ac9b9038e8a34755913465: ['FacilityUser matching query does not exist.']
WARNING  Validation error for Lesson with id 82d6fe57fe2ae7ee0ea2fd0f026f3b46: ['FacilityUser matching query does not exist.']
```


0.13 | 0.15
---|---
LEDev2104 (start)  Running  - Oracle VM VirtualBox_035 | Ubuntu18 04 (start)  Running  - Oracle VM VirtualBox_036

rtibbles:
Notifications don't currently get regenerated during sync, so this is expected behaviour.

radinamatic:
Maybe the screenshot is not really indicative, and I wasn't expecting the notification to appear in the Class activity pane, but in case of the successful sync, the Coach > Reports should display the same interactions, correct? Both devices have the same channel/resources available, so that should not be the issue.

rtibbles:
> Coach > Reports should display the same interactions

No, not exactly, as a lot of the reports also lean on the notifications for generating their data.

Issue #7525, opened 2020-09-23
Kolibri hangs loading content from Ciencia NASA (topics have thousands of child nodes)


dylanmccall:
### Observed behavior

The Ciencia NASA Kolibri channel has a topic node with a very large number of children. Attempting to view this content in Kolibri causes it to hang for a very long time (measured in minutes), blocking the UI. The content does eventually load, but not before I have given up.

### Expected behavior

Instead, this topic should load faster. It may be that it needs a smaller number of immediate descendants.

### User-facing consequences

The present situation renders a large portion of the channel inaccessible to end users.

### Steps to reproduce

First, import all content from the Ciencia NASA channel:

    kolibri manage importchannel network da53f90b1be25752a04682bbc353659f
    kolibri manage importcontent network da53f90b1be25752a04682bbc353659f

Next, open the Kolibri web interface and, in the Learn section, navigate to _Ciencia NASA / Ciencia Espacial_.

### Context

* Kolibri 0.14.3
* Linux
* Firefox, as well as the Kolibri GNOME app

indirectlylit:
see also: https://github.com/learningequality/kolibri/issues/3773 (add support for arbitrarily-sized topics)

jonboiser:
Yeah, the bottleneck seems to be on the client-side. In the profiler, it already takes 30 seconds just to create what I think is the virtual DOM representation of the list (lots of calls in vue to an `insert` method), then that is followed by several minutes of forced reflows and style updates (due to our use of a responsive window + element mixin that watches for resizing events) and mysterious gaps before the list actually appears.

![image](https://user-images.githubusercontent.com/10248067/94063276-b7499600-fd9c-11ea-8914-40dd57246694.png)

One possible way to address this is to put the whole list in a virtual window (using something like `vue-virtual-scroll-list`) which would probably fix the initial 30 second component initialization (which is already long) and possibly some of the other things.

indirectlylit:
> several minutes of forced reflows and style updates (due to our use of a responsive window + element mixin that watches for resizing events)

My guess is this might be attributable mostly to the responsive element mixin, more than the responsive window mixin. If this is causing layout thrashing, we should definitely address that.

Would you mind disabling that functionality and profiling again? You should be able to simply edit this file in `./node_modules/`:

https://github.com/learningequality/kolibri-design-system/blob/e013087ed119157cc8506e0985c7f1c5cf53312f/lib/KResponsiveElementMixin.js#L21-L36

Can change to:

[...]

PR #8138, opened 2021-06-02
Tree viewset for retrieving nested, paginated views of topic trees

### LINKED ISSUE #8108: Create API endpoint(s) for nested, paginated topic tree data retrieval
The new channel display for 0.15 requires showing multiple levels of the ContentNode hierarchy at once, and also for properly paginating the contents of these topics.

Using our current APIs this would result in requests to:
* Get the parent topic
* Get the child topics (up to the pagination limit - N)
* N requests to get the paginated children of each of the child topics

This issue proposes instead to create a single endpoint that returns all of this information in a single request. As this will be the primary means by which the topic tree is navigable, this will still have the virtue of being easily cacheable and reusable across multiple clients.

rtibbles:
## Summary
* To accommodate the nested, paginated topic tree display as outlined in designs for this feature: https://github.com/learningequality/kolibri-ecosystem/issues/7 this PR creates an API endpoint that allows for subsections of a topic tree to be queried in a single request, with the parent node, the paginated children, and the paginated grand children.
* The children and the children of each child are returned as a pagination object, containing the `results` Array, and `more` which is either `null` or an object with the parameters that can be passed to the same endpoint in order to query for more results.
* This PR also updates the ContentNodeResource to allow for simple access to the tree endpoint without having to define a new resource, and with a simplified caching mechanism that works across the TreeViewset and ContentNodeViewset requests.

## References
Fixes #8108

Contributes to the solution for #7525

I did a dummy implementation using the current topics page as a PoC of using this API, an animation is seen here, but I did not commit the changes for this:
viewmore

## Reviewer guidance
Does the API and its tests make sense?
Are the changes to the ContentNodeResource a good model for handling read only resources in the future?

----

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/core/assets/src/api-resources/contentNode.js
+++ b/kolibri/core/assets/src/api-resources/contentNode.js
@@ -1,5 +1,8 @@
+import isPlainObject from 'lodash/isPlainObject';
 import { Resource } from 'kolibri.lib.apiResource';
 import Store from 'kolibri.coreVue.vuex.store';
+import urls from 'kolibri.urls';
+import cloneDeep from '../cloneDeep';
 
 export default new Resource({
   name: 'contentnode',
@@ -33,5 +36,52 @@ export default new Resource({
   },
   fetchNextSteps(getParams) {
     return this.fetchDetailCollection('next_steps', Store.getters.currentUserId, getParams);
+  },
+  cache: {},
+  fetchModel({ id }) {
+    if (this.cache[id]) {
+      return Promise.resolve(this.cache[id]);
+    }
+    return this.client({ url: this.modelUrl(id) }).then(response => {
+      this.cacheData(response.data);
+      return response.data;
+    });
+  },
+  cacheData(data) {
+    if (Array.isArray(data)) {
+      for (let model of data) {
+        this.cacheData(model);
+      }
+    } else if (isPlainObject(data)) {
+      if (data[this.idKey]) {
+        this.cache[data[this.idKey]] = Object.assign(
+          this.cache[data[this.idKey]] || {},
+          cloneDeep(data)
+        );
+        if (data.children) {
+          this.cacheData(data.children);
+        }
+      } else if (data.results) {
+        for (let model of data.results) {
+          this.cacheData(model);
+        }
+      }
+    }
+  },
+  fetchCollection(params) {
+    return this.client({ url: this.collectionUrl(), params }).then(response => {
+      this.cacheData(response.data);
+      return response.data;
+    });
+  },
+  fetchTree(id, params) {
+    const url = urls['kolibri:core:contentnode_tree_detail'](id);
+    return this.client({ url, params }).then(response => {
+      this.cacheData(response.data);
+      return response.data;
+    });
+  },
+  fetchMoreTree({ id, params }) {
+    return this.fetchTree(id, params);
   },
 });
--- a/kolibri/core/assets/src/api-resources/contentNode.js
+++ b/kolibri/core/assets/src/api-resources/contentNode.js
@@ -4,6 +4,81 @@ import Store from 'kolibri.coreVue.vuex.store';
 import urls from 'kolibri.urls';
 import cloneDeep from '../cloneDeep';
 
+/**
+ * Type definition for Language metadata
+ * @typedef {Object} Language
+ * @property {string} id - an IETF language tag
+ * @property {string} lang_code - the ISO 639‑1 language code
+ * @property {string} lang_subcode - the regional identifier
+ * @property {string} lang_name - the name of the language in that language
+ * @property {('ltr'|'rtl'|)} lang_direction - Direction of the language's script,
+ * top to bottom is not supported currently
+ */
+
+/**
+ * Type definition for AssessmentMetadata
+ * @typedef {Object} AssessmentMetadata
+ * @property {string[]} assessment_item_ids - an array of ids for assessment items
+ * @property {number} number_of_assessments - the length of assessment_item_ids
+ * @property {Object} mastery_model - object describing the mastery criterion for finishing practice
+ * @property {boolean} randomize - whether to randomize the order of assessments
+ * @property {boolean} is_manipulable - Whether this assessment can be programmatically updated
+ */
+
+/**
+ * Type definition for File
+ * @typedef {Object} File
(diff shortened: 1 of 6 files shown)


kolibri/core/assets/src/api-resources/contentNode.js
```diff
@@ -34,4 +37,51 @@ export default new Resource({
...
+    });
+  },
+  fetchTree(id, params) {
+    const url = urls['kolibri:core:contentnode_tree_detail'](id);
+    return this.client({ url, params }).then(response => {
+      this.cacheData(response.data);
+      return response.data;
+    });
+  },
+  fetchMoreTree({ id, params }) {
```
nucleogenesis, line 174: > The children and the children of each child are returned as a pagination object, containing the results Array, and more which is either null or an object with the parameters that can be passed to the same endpoint in order to query for more results.

This would be a great bit of knowledge to put in a comment somewhere around here I think. My initial question after looking at the code was "what do I pass on here" but then I read the description.

kolibri/core/assets/src/api-resources/contentNode.js
```diff
@@ -34,4 +37,51 @@ export default new Resource({
   fetchNextSteps(getParams) {
     return this.fetchDetailCollection('next_steps', Store.getters.currentUserId, getParams);
   },
+  cache: {},
+  fetchModel({ id }) {
```
nucleogenesis, line 116: Could you add some comments explaining why you've overridden the caching mechanism specifically for this child of `Resource`? It might help in case there is need for debugging here in the future to know _why_ it was done this way instead of using the default implementation in `Resource`.

nucleogenesis (dismissed):
Just a couple requests for comments in the `contentNode.js` otherwise it looks good to me.

kolibri/core/assets/src/api-resources/contentNode.js
rtibbles: Sounds good, yep - I can also add some more documentation on the API side as well.

kolibri/core/assets/src/api-resources/contentNode.js
rtibbles: Can do - also, any thoughts on this implementation (it's essentially a simplified form of the existing implementation, but removing support for dynamic fields loading, and allowing for easier sharing of data for caching across endpoints)?

nucleogenesis: After another pass - one thing I'm not seeing is where the cache is actually going to be read from.

The default implementation of `Resource.fetchCollection` calls  `Resource.getCollection` which does all of the cache checking business.

These implementations of `ContentNodeResource.fetch(Collection | Tree | MoreTree)` never reach into the cache before hitting the API.

Otherwise, this is certainly a far more easily grokked set of functions than all of the cache internals of `Resource`.

rtibbles: `fetchModel` and `fetchCollection` are the most commonly used entrypoints to the fetching behaviour. One big change here is that this is no longer trying to do any caching of collection fetching behaviour - instead relying on browser cache to prevent API requests hitting the server in cases where the cache headers are set.

The only thing this really does is to make sure if we do `fetchModel` after a `fetchCollection` that contained that specific model, we don't do another request to the server just to get that one model that we already have all the data for.

So, yeah, in many ways the cache is doing less - and I think it's appropriate for it to do so, especially when combined with good caching headers on the backend.

nucleogenesis (commented):
EDIT: - With [this branch](https://github.com/rtibbles/kolibri/commit/4517d90930728e826d888e749ff08741d1f60258) off of this PR's branch checked out I saw completely solid response times and payloads :+1:

---
 
I'm also getting 30+ second loading times and every payload coming back is precisely 16.7MB.

I can't even view the response from the server because it freezes the browser tab.

image

kolibri/core/assets/src/api-resources/contentNode.js
nucleogenesis: Related to the comment I left about payload sizes, I also noted that there is not `content-encoding` header on the responses - but the client is ready to accept. Are we not gzipping our responses?

rtibbles: Not for API responses generally, no - we did have it enabled in CherryPy prior to 0.12.x and it was a source of considerable slowdown.

kolibri/core/content/api.py
```diff
@@ -686,6 +724,167 @@ def resume(self, request, **kwargs):
...
+                # its parent.
+                if len(gc_by_parent[gc["parent_id"]]) < self.page_size:
+                    gc_by_parent[gc["parent_id"]].append(gc["id"])
+                    yield gc["id"]
+
+    def retrieve(self, request, **kwargs):
+        """
+        A nested, paginated representation of the children and grandchildren of a specific node
+
+        :param request: request object
```
jredrejo, line 819: shouldn't this doc add the request might have `depth` & `lft__gt`?

kolibri/core/content/test/test_content_app.py
```diff
@@ -233,13 +234,154 @@ def test_related_filter(self):
...
+
+    def test_contentnode_tree_long(self):
+        builder = ChannelBuilder(levels=2, num_children=30)
+        builder.insert_into_default_db()
+        content.ContentNode.objects.all().update(available=True)
+        root = content.ContentNode.objects.get(id=builder.root_node["id"])
+        response = self.client.get(
+            reverse("kolibri:core:contentnode_tree-detail", kwargs={"pk": root.id})
+        )
+        self._recurse_and_assert([response.data], [root])
```
jredrejo, line 400: I see here tests to check `more` and `lft__gt` are correctly returned when calling the endpoint but I don't see the `depth` and `lft__gt ` parameters of the endpoint being tested (unless I get lost in the recursion)

jredrejo (commented):
Just a couple of minor details. Overall it looks great

kolibri/core/content/api.py
rtibbles: Yes - I have been trying to look how best to document query params so they are picked up by our API explorer as well, will update this.

jredrejo (approved):
Thank you, I see it good to go.

PR #8142, opened 2021-06-07
Add task decorators and task APIs for functions registered via decorators

vkWeb:
## Summary

> ### This summary is outdated. Please refer to `kolibri.core.tasks.job.RegisteredJob`'s docstring for updated info.

This PR adds two task decorators:

- `task.register(job_id=None, validator=None, permission=None, priority=task.Priority.REGULAR)`
    
    This decorator when applied registers function to `JobRegistry.REGISTERED_JOBS` dict which will be queried when client 
    submits a method via API endpoint.
    
- `task.config(group=None, cancellable=False, track_progress=False)`

    This decorator affects only if first `task.register` is applied otherwise it raises an exception. This allows modification of task 
    attributes.

Next, I've added three methods available only on jobs registered via decorator. Suppose `add` is defined as:

```python
@task.register
def add(x, y): 
    return x + y
```  
- `add.task.enqueue_at(dt, interval=0, repeat=0).initiate(65, 85)`
    It's very readable. `add` will be scheduled based on `dt, interval, repeat` with arguments 65 and 85.

- `add.task.enqueue_in(delta_t, interval=0, repeat=0).initiate(89, 26)`
    `add` will be scheduled based on `delta_t, interval, repeat` with arguments 89 and 26.

- `add.task.initiate(56, 24)`
    `add` will be enqueued with arguments 56 and 24.

The `task.register` decorator and the api methods have been applied to `_ping` job for example showcase and to manually test everything.

**To dos for this PR:**
- improve implementation based on reviews
- add unit tests (extremely important)    

## Reviewer guidance

Apply decorators to functions, play with the api and see if everything's working as expected.

----

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated


vkWeb:
cc @rtibbles for review.

vkWeb:
@rtibbles I was taking a fresh look at the code today morning, it came to me that we can reduce some (unnecessary...?) code and simplify more by exposing `add.task.initiate` as:

```
add.task.initiate(enq_in=False, enq_at=False, time=None, interval=0, repeat=0, *args, **kwargs)
```
Going with the above method will let us remove `enqueue_in` and `enqueue_at` & related params variables. These feel kind of extra code to me. What do you think? What method should we go with, `add.task.enqueue_in(..).initiate(..)` one or the above one?

vkWeb:
Marking this as ready for review since most of the work is done. 

We just need to finalize which of the `initiate` api we should be going forward with, once that's done, I'll refactor it and add tests for it. Then this will be ready to merge after addressing review comments (if any).

kolibri/core/tasks/test/test_decorators.py
```diff
@@ -0,0 +1,91 @@
...
+
+
+@pytest.fixture
+def registered_jobs():
+    JobRegistry.REGISTERED_JOBS.clear()
+    yield JobRegistry.REGISTERED_JOBS
+    JobRegistry.REGISTERED_JOBS.clear()
+
+
+class TestTaskDecorators(object):
```
jonboiser: Could you make this a [django `TestCase` ](https://docs.djangoproject.com/en/3.2/topics/testing/overview/)and use the assertion methods they provided? (e.g. `self.assertEqual`)? This is just to be consistent with the rest of the test suite; plus you might need/want some of the features provided to the test runner.

vkWeb: Yeap, good idea. I've pushed commits that addresses this.

kolibri/core/tasks/decorators.py
```diff
@@ -0,0 +1,108 @@
...
+                self.config,
+                group=group,
+                cancellable=cancellable,
+                track_progress=track_progress,
+            )
+
+        funcstring = stringify_func(func)
+
+        try:
+            registered_job = JobRegistry.REGISTERED_JOBS[funcstring]
```
rtibbles: Given that we are modifying the registered task in place I am not sure I understand the motivation of separate 'register' and 'config' decorators.

My first assumption when I read the description of the API was that this would allow the same registered job to be exposed with different configurations - but here we are updating the configuration in place of the registered job.

With this in mind, it seems simpler to just put all the kwargs for the config decorator into the register decorator.

kolibri/core/analytics/utils.py
```diff
@@ -508,13 +511,7 @@ def schedule_ping(
...
-            started,
-            _ping,
-            interval=interval * 60,
-            repeat=None,
-            started=started,
-            server=server,
-            checkrate=checkrate,
-            job_id=job_id,
+        _ping.task.job_id = job_id
+        _ping.task.enqueue_at(started, interval=interval * 60, repeat=None).initiate(
```
rtibbles: I understand the motivation for wanting to separate out the args and kwargs for `enqueue_at` here - but I think the additional `initiate` is a slightly odd syntax. If we want to clearly delineate the arguments, we can change it so that `enqueue_at` takes explicit `args` and `kwargs` keyword arguments, which then get passed to the invoked task as `*args, **kwargs`.

kolibri/core/tasks/job.py
```diff
@@ -39,6 +61,7 @@ class State(object):
     should be set with the function's return value.
     """
 
+    PENDING = "PENDING"
```
rtibbles, line 64: It seems like this would only be the case before the job gets saved to the task database (after that it's at least SCHEDULED or QUEUED) - when is it useful to track that a job is PENDING?

kolibri/core/tasks/job.py
```diff
@@ -222,3 +256,118 @@ def __repr__(self):
...
+        """
+        self.enqueue_at_params = {
+            "dt": specific_time,
+            "interval": interval,
+            "repeat": repeat,
+        }
+        self.enqueue_in_params.clear()
+        return self
+
+    def initiate(self, *args, **kwargs):
```
rtibbles: I think this can be refactored into a 'create job object' method - then we have three methods:
`enqueue`, `enqueue_in`, `enqueue_at` that call this to generate the Job object, then each one calls the appropriate part of the if statement below.

This also means that you don't have to track state on `self` which is a little dangerous if multiple parallel threads are spawning tasks, as they might share memory and leak state to each other. By just calling the scheduler or queue method in each method, the arguments are scoped inside the method.

kolibri/core/tasks/job.py
```diff
@@ -48,6 +71,22 @@ class State(object):
...
 
 
+class Priority(object):
+    """
+    This class defines the priority levels and their corresponding string values.
+
+    REGULAR priority is for tasks that can wait for some time before it actually
+    starts executing. Tasks that are tracked on task manager should use this priority.
+
+    HIGH priority is for tasks that want execution as soon as possible. Tasks that
```
rtibbles, line 81: The description is correct - but the examples less so. The primary one here is channel metadata import.

kolibri/core/tasks/job.py
```diff
@@ -149,20 +195,8 @@ def __init__(self, func, *args, **kwargs):
         self.args = args
         self.kwargs = kwargs
         self.result = None
-
         self.storage = None
-
-        if callable(func):
```
rtibbles, line 155: Any reason to remove this code? Seems like harmless defensive programming to me.

kolibri/core/tasks/utils.py
```diff
@@ -18,13 +18,16 @@ def get_current_job():
 
 
 def stringify_func(func):
-    if not callable(func):
```
rtibbles, line 21: Oh, disregard my comment above, you're just consolidating logic.

kolibri/utils/main.py
```diff
@@ -15,6 +15,7 @@
 
 import kolibri
 from kolibri.core.deviceadmin.exceptions import IncompatibleDatabase
+from kolibri.core.tasks.decorators import import_task_modules_frm_django_apps
```
rtibbles: Typo? `from` not `frm`?

rtibbles (commented):
Some ideas for how the API and code could be a little cleaner, and a couple of questions.

kolibri/core/tasks/decorators.py
vkWeb: The main motivation behind separation of `register` and `config` decorator was to separate "configurable" attributes from other attributes. Every kwarg in one decorator will make that decorator too longish and kind of hard to read, don't you think?

kolibri/core/analytics/utils.py
vkWeb: That's actually a great idea! Commits landing right now.

kolibri/core/tasks/job.py
vkWeb: Actually, when I was manually testing, I sometimes felt unclear if the job has been actually QUEUED or not so to make that explicit I added PENDING.

kolibri/core/tasks/job.py
vkWeb: Superb! This will be a good design.

kolibri/core/tasks/job.py
vkWeb: Commits landing!

kolibri/utils/main.py
vkWeb: Commits landing!

kolibri/core/tasks/decorators.py
rtibbles: I think there are two alternatives:

1. Just be ok with lots of kwargs - this feels OK, especially as it has the advantage of the function parameters being explicit.
2. Add a `config` kwarg that accepts a dict of the configurable attributes - this clearly delineates these attributes as 'configurable' but you lose the function parameter introspectability to know what arguments can be passed.

I think because I don't really see the kwargs being passed to `register` and `config` right now as being that different to each other (for example, if a task is cancellable is not really configurable, and more a matter of whether the internal implementation of the task properly handles when a cancel event is raised) my preference is for what I suggested above, which is 1.

It's possible we may have a clearer need for separating configuration from registration in the future, but for now, it seems unnecessary.

kolibri/core/tasks/job.py
rtibbles: OK, if there was ambiguity that this helps clear up, that's fine.

kolibri/core/tasks/decorators.py
vkWeb: Okay, so let us move forward with 1 as said above. 

For `POST /api/tasks/`, I added `@task.permission_classes([])` as a separate decorator, I think we will be fine with that or do we want to keep `permission_classes` as a kwarg in register? I will say separation will be much better in this case to allow for easy readability of permissions especially when bitwise operators are used.

kolibri/core/tasks/job.py
vkWeb: Yup, PENDING is just to keep things explicit internally.

kolibri/core/tasks/decorators.py
rtibbles: I think to maintain simplicity, just adding it as another kwarg to register for now would be fine - we can then see if it produces monstrous code and we need to rethink, but I think it is mostly just a matter of code formatting, so with `black` running on the code, I think it should be readable.

vkWeb: Okay, let's go with only register. One to rule them all! Commits landing asap.

vkWeb:
I've refactored the API and decorators based on the feedback. This new design and code feels very robust. Thanks @rtibbles for the amazing ideas! 

This is ready to merge.

cc @rtibbles @jonboiser.

kolibri/core/tasks/decorators.py
```diff
@@ -0,0 +1,81 @@
...
+from kolibri.core.tasks.job import JobRegistry
+from kolibri.core.tasks.job import Priority
+from kolibri.core.tasks.job import RegisteredJob
+from kolibri.core.tasks.utils import stringify_func
+
+
+logger = logging.getLogger(__name__)
+
+
+class _TaskDecorators(object):
```
rtibbles: I think it would be good to just make `register` a top level module function, rather than using the class here.

rtibbles (changes requested):
One last thing to clean up.

kolibri/core/tasks/decorators.py
vkWeb: I personally like decorators of form `object.action` (that's why I left the class there :sweat_smile:). 

Anyway, since we only have one decorator let us get rid of class. 

In future if we think of a second decorator then we'll get back to class please! :laughing:

rtibbles (approved):
Very clean!

PR #8165, opened 2021-06-21
Initiate auto-syncing from zeroconf

### LINKED ISSUE #8157: Initiate auto-syncing from zeroconf listeners
For a device that has a subset of users, when a new device joins the network, that is not a subset of users device, attempt to initiate an automatic sync with that device.

Could achieve this by adding behaviour directly in zeroconf, or potentially by adding signals to the DynamicNetworkLocation model.

jredrejo:
## Summary

For a device that has a subset of users, when a new device joins the network, that is not a subset of users device, attempt to initiate an automatic sync with that device.

## References
Closes: #8157

## Reviewer guidance
All the api handshake seems to work fine, tested between three devices: 
1. A previous version of  kolibri
2. A kolibri server that's not a SoUD 
3. A kolibri server that's a SoUD and has some users synced from the previous server

After starting 1, 2, 3 and stopping 2 the algorithm increases the timeouts to request , what is correct.
After starting 2 again, the algorithm waits to have a permission to request, respecting the order in the queue

However, I have watched the queue in server 2 is always keeping one `TransferSession` active, so this seems to be a never-to-sync situation. The session marked below never ends:
image

As we changed the `MAX_CONCURRENT_SYNCS` [variable to be 1](https://github.com/learningequality/kolibri/blob/develop/kolibri/core/public/api.py#L33), syncing is never going to happen while one session  is stalled

To avoid this problem, and following @bjester adviced, the server is not giving permission to sync if there has not been any syncing activity in the last five minutes. After that changed (in a commit in this PR) syncing is triggered automatically by zeroconf and it seems to work fine between these two servers (having server 3 two `FacilityUsers`):
image

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/core/discovery/utils/network/search.py
+++ b/kolibri/core/discovery/utils/network/search.py
@@ -12,7 +12,10 @@ from zeroconf import ServiceInfo
 from zeroconf import USE_IP_OF_OUTGOING_INTERFACE
 from zeroconf import Zeroconf
 
+from kolibri.core.auth.models import FacilityUser
+from kolibri.core.device.utils import get_device_setting
 from kolibri.core.discovery.models import DynamicNetworkLocation
+from kolibri.core.public.utils import begin_request_soud_sync
 from kolibri.core.public.utils import get_device_info
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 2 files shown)


rtibbles (approved):
This looks simple to me and the manual testing looks excellent.

If we need to tweak it, doing so should be simple going forwards.

PR #8172, opened 2021-06-22
Implement swipe-to-open/close gestures for Side Navigation

### LINKED ISSUE #6899: swipe from left to open side nav on mobile
### Observed behavior

Click-to-open on mobile, reported by @jtamiace 

A previous work-in-progress PR was opened some time ago here: https://github.com/learningequality/kolibri/pull/8172
- Please use this PR as reference for your starting point. 
- The acceptance criteria that still needs to be met in the previous PR is that **this feature should work for both LTR and RTL language support.** 

### Expected behavior

| Would be nice to open this menu when swiping from the left edge of the screen |
|--| 
| image|

### User-facing consequences

Less efficient use on touch-screens

### Context

0.13

jonboiser:
## Summary

1. Uses https://vueuse.org/core/useswipe/#useswipe to listen to TouchEvents (so this should only work on touch screens, and not with mice) and open/close the SideNav when swiping right/left.

*Screenshot from iOS simulator*

(this can also be reproduced using Chrome mobile device simulator)
CleanShot 2021-06-22 at 11 19 50

## References

Fixes #6899 

## Reviewer guidance

The options for `useSwipe`, like the threshold, as well as the size of the "swipe zone" for opening the menu may need to be tuned. Do you have any suggestions?

I also made a decision not to place a guard around the swipe-left-to-close gesture for large devices. This means that something like a MS Surface user could use the gesture.

Also, this does not account for RTL languages (might need have a follow up issue). The `isRTL` prop should be accessible using https://v3.vuejs.org/api/composition-api.html#getcurrentinstance within the `setup()` function like this

```
const isRTL = getCurrentInstance().proxy.isRtl
```

In an RTL layout, the swipe directions need to be reversed, and the swipe zone on CoreBase needs to be moved to the right edge of the screen.

----

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/core/assets/src/views/CoreBase/index.vue
+++ b/kolibri/core/assets/src/views/CoreBase/index.vue
@@ -8,6 +8,9 @@
 
     <div v-if="blockDoubleClicks" class="click-mask"></div>
 
+    <!-- Invisible div on the very edge that enables the swipe-to-open-menu gesture -->
+    <div v-if="windowIsSmall" ref="swipeZone" class="swipe-zone"></div>
+
     <ScrollingHeader
       :scrollPosition="scrollPosition"
       :alwaysVisible="fixedAppBar"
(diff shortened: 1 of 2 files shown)


rtibbles:
Seems like we'd need to address the RTL aspects before merging this.

marcellamaki:
Closing out and updating the original issue to refer to this this as a starting point.

PR #8177, opened 2021-06-25
Return learning_activities and duration from contentnode endpoints.

rtibbles:
## Summary
* Returns inferred learning_activities based on content kinds
* Returns null duration value until information is available in the database

## Reviewer guidance
Does `learning_activities` return an appropriate list?
Does `duration` return null?

----

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/core/content/api.py
+++ b/kolibri/core/content/api.py
@@ -227,6 +227,22 @@ def map_file(file):
     return file
 
 
+kind_activity_map = {
+    content_kinds.EXERCISE: "practice",
+    content_kinds.VIDEO: "watch",
+    content_kinds.AUDIO: "listen",
+    content_kinds.DOCUMENT: "read",
+    content_kinds.HTML5: "explore",
+}
+
+
+def infer_learning_activity(node):
+    activity = kind_activity_map.get(node["kind"])
+    if activity:
+        return [activity]
+    return []
+
+
 class BaseContentNodeMixin(object):
     """
     A base mixin for viewsets that need to return the same format of data
@@ -258,6 +274,11 @@ class BaseContentNodeMixin(object):
         "rght",
         "tree_id",
     )
+
+    field_map = {
+        "learning_activities": infer_learning_activity,
+        "duration": lambda x: None,
+    }
 
     def get_queryset(self):
         return models.ContentNode.objects.filter(available=True)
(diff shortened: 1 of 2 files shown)


MisRob:
Thank you for helping with this @rtibbles. The code looks good to me. Locally, I checked for a request with this URL: http://127.0.0.1:8000/api/content/contentnode/09298be9288d50949d9cf3deb20b1f12/next_content/?contentCacheKey=1622526562

and am not getting the new data in the response:
```
{
  "kind":"video",
  "id":"5e43d269693d5c69a4c06029067df8ce",
  "title":"Capital",
  "thumbnail":"/content/storage/7/b/7b38117db02542649056a3d3c8631808.png",
  "is_leaf":true
}
```

Are these updates supposed to make it work for the `next_content` endpoint too?

If yes, could the issue be caused by `BaseContentNodeMixin` not inheriting from `ValuesViewset`? I am not sure how exactly all this works but in other occurrences in the codebase, it seems that we use `field_map` only in classes having a connection to `ValuesViewset`.

MisRob:
Or could I have problems with some caching, maybe? I see some cache key in the request URL `contentCacheKey=1622526562`

kolibri/core/content/api.py
```diff
@@ -227,6 +227,22 @@ def map_file(file):
     return file
 
 
+kind_activity_map = {
```
MisRob, line 230: Is this the final backend definition? If yes, we'll need to change [frontend constants](https://github.com/learningequality/kolibri/blob/develop/kolibri/core/assets/src/constants.js#L32) to lower case. I can open a PR for that.

rtibbles: All of our other constants for content are lower case, so this seemed consistent.

rtibbles: Also, if we could change the name of the map from `LearningActivityKinds` to just `LearningActivities` it would sit better with me - want to avoid confusing them with the `ContentNodeKinds`.

rtibbles:
If you check your network tab, are you hitting the browser cache?

To avoid this happening in production, it might behoove us to trigger an update to the content cache key on update.

MisRob:
It was the browser cache, I'm sorry.

MisRob (approved):

kolibri/core/content/api.py
MisRob: Resolved in https://github.com/learningequality/kolibri/pull/8180

PR #8183, opened 2021-07-01
Sync resumption for SoUD syncs and cleanup handling

### LINKED ISSUE #8019: Sync command should support resuming a `SyncSession` through ID argument
## Background
The `sync` management command doesn't currently support resuming a sync session, so if synchronization becomes interrupted for any reason, you must start from scratch to complete the transfer. 

## User story
As a user who is using Kolibri's `sync` command in connectivity challenged environments, I need to be able resume a sync if it is interrupted. Syncing large amounts of data can take some time, especially on slower connections, and if that becomes interrupted, I'm faced with having to restart the entire process which leads to a loss of time and perhaps increased cost from the data transfer.

## Deliverables
- The `sync` command should update its task state with the `SyncSession` ID when it has created a new session to allow for resuming a specific interrupted sync task through the UI
- The `sync` command should accept a `--syncsession_id` argument and should pass it to [`resume_sync_session`](https://github.com/learningequality/morango/issues/103)

bjester:
## Summary

- Requires `morango>=0.6.5`
- Refactors sync command logic into base class
- Updates existing `sync` command to inherit new base class and to output the `SyncSession` ID for resuming
- Adds new `resumesync` command that inherits new base class and accepts a `--id` parameter for a `SyncSession` to resume
- Refactors SoUD syncing to keep session alive, and handles cleanup on disconnect on client instance and server
- Integrates changes from https://github.com/learningequality/python-zeroconf-py2compat/pull/12
- Refactors Zeroconf handling to stop restarting it and use new `Zeroconf.update_interfaces`

## References

Resolves: https://github.com/learningequality/kolibri/issues/8019

### Situations that touch areas affected by this PR:
- Network connects/disconnects
- Other Kolibri instances appearing/disappearing from the network
- Setting up Kolibri (as learner-only or importing facility)
- Kolibri starts/stops
- Syncing to networked Kolibris
- Syncing to KDP
- Learner-only device recurring syncing

### Scenario: Full facility
1. Connect your device to a network/VPN with another Kolibri instance with a full facility
2. Start a fresh install of Kolibri
3. In the Kolibri setup wizard, choose Advanced > Import all data from an existing facility
4. Ensure the other Kolibri instance is discovered on the network
5. Import the facility from the other instance
6. Finish setting up Kolibri
7. Disconnect from the network with the other Kolibri instance
8. From Device > Facilities, choose <kbd>SYNC</kbd> for the facility you imported, and select "Local network or internet"
9. Verify that the other Kolibri instance is not shown as available, and keep the "Select network address" modal open
10. Reconnect to the network with the other Kolibri instance
11. Verify that the other Kolibri instance appears within 60 seconds
12. Complete syncing with the instance

### Scenario: Learner-only
1. Connect your device to a network/VPN with another Kolibri 0.15 instance with a full facility
2. Start a fresh install of Kolibri
3. In the Kolibri setup wizard, choose Advanced > Import one or more user accounts from an existing facility
4. Ensure the other Kolibri instance is discovered on the network
5. Choose a facility and import a learner from it
6. Finish setting up Kolibri
7. Disconnect from the network with the other Kolibri instance
8. Wait a few minutes
9. Reconnect to the network with the other Kolibri instance
10. Verify that your device (eventually) syncs the instance again

----

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated


bjester:
Unmarked this as ready for review as it seems I need to refactor the restarting of the Zerconf service in order to trigger those events

rtibbles:
> Unmarked this as ready for review as it seems I need to refactor the restarting of the Zerconf service in order to trigger those events

Possible that we could also modify the zeroconf interface a little, if we want to make the way we do the restart a little less blunt - if that would help.

bjester:
> > Unmarked this as ready for review as it seems I need to refactor the restarting of the Zerconf service in order to trigger those events
> 
> Possible that we could also modify the zeroconf interface a little, if we want to make the way we do the restart a little less blunt - if that would help.

@rtibbles Yeah should be fairly straightforward

kolibri/core/public/utils.py
```diff
@@ -194,6 +291,20 @@ def begin_request_soud_sync(server, user):
...
 
 
+def stop_request_soud_sync(server, user):
+    """
+    Cleanup steps to stop SoUD syncing
+    """
+    info = get_device_info()
+    if not info["subset_of_users_device"]:
+        # this does not make sense unless this is a SoUD
+        logger.warn("Only Subsets of Users Devices can do this")
```
jredrejo: In a server used by an external SoUD this appears repeated several times every time the server starts and stops.
This is the list of calls taking the server to this point:
```

[20]   /datos/le/mio/kolibri/kolibri/core/discovery/utils/network/search.py(111)partial_unsubscribe()
-> self.remove_instance(other_instance)
[21]   /datos/le/mio/kolibri/kolibri/core/discovery/utils/network/search.py(153)remove_instance()
-> stop_request_soud_sync(server=instance.base_url, user=user_id)
[22] > /datos/le/mio/kolibri/kolibri/core/public/utils.py(299)stop_request_soud_sync()
-> info = get_device_info()

```

jredrejo (commented):
I have tested it and hadn't found any problem with these changes.
Only the warning messages appearing multiple times on a not SoUD server that has been used to sync several SoUD:
```
INFO     Kolibri instance 'd34fa0e29c06817d732e4f7b81f3e2aa' joined zeroconf network; device info: {'application': 'kolibri', 'kolibri_version': '0.15.0a5.dev0+git.18.g94c233af', 'instance_id': 'd34fa0e29c06817d732e4f7b81f3e2aa', 'device_name': 'jose-xps', 'operating_system': 'Linux', 'subset_of_users_device': False}
WARNING  Only Subsets of Users Devices can do automated SoUD syncing.
WARNING  Only Subsets of Users Devices can do automated SoUD syncing.
WARNING  Only Subsets of Users Devices can do automated SoUD syncing.
WARNING  Only Subsets of Users Devices can do automated SoUD syncing.
WARNING  Only Subsets of Users Devices can do automated SoUD syncing.
WARNING  Only Subsets of Users Devices can do automated SoUD syncing.
WARNING  Only Subsets of Users Devices can do automated SoUD syncing.
WARNING  Only Subsets of Users Devices can do automated SoUD syncing.
WARNING  Only Subsets of Users Devices can do automated SoUD syncing.
WARNING  Only Subsets of Users Devices can do automated SoUD syncing.
WARNING  Only Subsets of Users Devices can do automated SoUD syncing.
WARNING  Only Subsets of Users Devices can do automated SoUD syncing.
WARNING  Only Subsets of Users Devices can do automated SoUD syncing.
WARNING  Only Subsets of Users Devices can do automated SoUD syncing.
WARNING  Only Subsets of Users Devices can do automated SoUD syncing.
WARNING  Only Subsets of Users Devices can do automated SoUD syncing.
WARNING  Only Subsets of Users Devices can do automated SoUD syncing.
WARNING  Only Subsets of Users Devices can do automated SoUD syncing.
WARNING  Only Subsets of Users Devices can do automated SoUD syncing.
WARNING  Only Subsets of Users Devices can do automated SoUD syncing.
WARNING  Only Subsets of Users Devices can do automated SoUD syncing.
WARNING  Only Subsets of Users Devices can do this
WARNING  Only Subsets of Users Devices can do this
WARNING  Only Subsets of Users Devices can do this
WARNING  Only Subsets of Users Devices can do this
WARNING  Only Subsets of Users Devices can do this
WARNING  Only Subsets of Users Devices can do this
WARNING  Only Subsets of Users Devices can do this
WARNING  Only Subsets of Users Devices can do this
WARNING  Only Subsets of Users Devices can do this
WARNING  Only Subsets of Users Devices can do this
WARNING  Only Subsets of Users Devices can do this
WARNING  Only Subsets of Users Devices can do this
WARNING  Only Subsets of Users Devices can do this
WARNING  Only Subsets of Users Devices can do this
WARNING  Only Subsets of Users Devices can do this
WARNING  Only Subsets of Users Devices can do this
WARNING  Only Subsets of Users Devices can do this
WARNING  Only Subsets of Users Devices can do this
WARNING  Only Subsets of Users Devices can do this
WARNING  Only Subsets of Users Devices can do this
WARNING  Only Subsets of Users Devices can do this
INFO     Kolibri running on: http://172.17.0.1:10000/
```

bjester:
Thanks @jredrejo! That does seem to be a small regression, which I'll look into

kolibri/core/auth/management/utils.py
```diff
@@ -419,3 +462,330 @@ def delete(self, progress_updater, sleep=None):
...
+                    noninteractive,
+                    client_cert_scope.read_filter,
+                )
+                # and push our own data to server
+            if not no_push:
+                self._push(
+                    sync_session_client,
+                    noninteractive,
+                    client_cert_scope.write_filter,
+                )
```
jamalex, line 543: A couple potential consequences (depending on whether I'm missing some other pieces of this) of pulling the filters directly from the client certificate:
- This prevents the (otherwise valid and previously allowed) option of doing a single-user sync between two full-facility devices. While our current use cases/UI don't facilitate this option, it could be good to keep as an option for initiating via CLI.
- More importantly, I think this will also prevent doing a single-user sync the other direction (syncing a full-facility device as the "client" against a single-user device), which was something we explicitly aimed to allow (and is covered in the [ecosystem/integration tests](https://github.com/learningequality/kolibri/blob/release-v0.15.x/kolibri/core/auth/test/test_morango_integration.py#L585), which could be good to test locally with this PR as well -- they don't get run in the automatic tests, currently, as they're too slow; need to toggle on the `INTEGRATION_TEST` envvar).

jamalex (changes requested):
Took a peek at one specific piece, per a request, and added a comment there.

kolibri/core/auth/management/utils.py
bjester: Thanks @jamalex! I didn't realize that was a use case. I will fix it.

kolibri/core/discovery/utils/network/broadcast.py
```diff
@@ -0,0 +1,561 @@
...
+                # info backwards incompatible with older versions of Kolibri
+                val = TRUE if val else FALSE
+            properties[key] = json.dumps(val)
+
+        return ServiceInfo(
+            SERVICE_TYPE,
+            self.name,
+            server=self.server,
+            address=self.ip,
+            port=self.port or DEFAULT_PORT,
```
rtibbles, line 155: When would we not have the port information here? (If this is just copy pasted from the previous implementation, feel free to ignore).

rtibbles (commented):
Code read through makes sense to me!

kolibri/core/discovery/utils/network/broadcast.py
bjester: This does originate from the existing code, but I think you're right that we really don't need it

kolibri/core/auth/management/utils.py
bjester: I have made the changes and all of the morango integration tests pass

kolibri/core/public/utils.py
bjester: I have resolved this issue. Thanks @jredrejo

pcenov:
Hi @bjester, today I had a look at this [build](https://buildkite.com/learningequality/kolibri-python-package/builds/3864) on Windows, Ubuntu and Android and noticed that the auto sync is not functioning at all for the learner-only devices that I've setup on both Windows and Android using Ubuntu as a server.
Here are the logs and DB files in case you are not aware of that issue:
[WindowsLogsAndDB.zip](https://github.com/learningequality/kolibri/files/7347237/WindowsLogsAndDB.zip)
[UbuntuServerLogsAndDB.zip](https://github.com/learningequality/kolibri/files/7347238/UbuntuServerLogsAndDB.zip)
The other main scenario about syncing a full facility is functioning correctly. Let me know if you need any additional info.

bjester:
Thanks @pcenov, I'm aware of that and have been debugging that issue, but the cause is still unclear

pcenov:
Tested again today on the latest [Windows 10, Ubuntu and Android builds](https://buildkite.com/learningequality/kolibri-python-package/builds/3868) and can confirm that the sync functions correctly now.
I tried the scenarios listed in the description plus deleting the saved data on a learner-only device and restoring it to check whether it will sync back; disconnecting and reconnecting, deleting and adding resources and various other scenarios and didn't find any other issues.
I noticed that when I attempt to import a Facility or a channel from my local network I do see the Android learner-only device's URL as localhost (6b16) which seems incorrect but I don't know whether it's displayed intentionally or is a regression issue so bringing that to your attention as well.
@bjester @radinamatic

rtibbles (approved):
Nice - the cleanup of SoUD syncing jobs on unregistering is a nice addition too!

rtibbles:
Force merging as linting issue is resolved in https://github.com/learningequality/kolibri/pull/8513

Issue #8170, opened 2021-06-22
Create public facility user endpoint for single user syncing


rtibbles:
To support single user syncing setup, a public API endpoint is required that will:
* Return a user's information when that user has signed in as a learner
* Return a paginated list of users that can be selected for single user syncing when logged in as an admin
* Allow for search by username, fullname, classroom name, Identifier Number - check blackbox of facility admin searching

The form of the return data can be based on permissions based filtering - for ease, can always return paginated data, and simply return just the learner if a learner is logged in.

Authentication will be required for the endpoint, and should probably use HTTP Basic Auth in order to allow simple repeated queries without having to maintain a session.

jredrejo:
The first two points of this issue have been implemented in in  #8185 in this commit: https://github.com/learningequality/kolibri/pull/8185/commits/47b486067fc25bf0748c326f9f07e1dd4e0c9b0a

As BasicAuth does not allow to add the facility and two users might have the same username & password in different facilities, this has be modified to include the facility_id as a realm when authenticating using BasicAuth:
https://github.com/learningequality/kolibri/blob/a56d04850e9c765b50a94ec0b25c0279ed197dfb/kolibri/core/auth/api.py#L82-L106
I.E: username@facility_id should be the user

rtibbles:
Done!

PR #8185, opened 2021-07-05
Wizard state machine to implement single user syncing flow 

### LINKED ISSUE #8166: Implement single user syncing flow in the setup wizard
Implement designs/user flow described here:

https://www.figma.com/file/w3raYcGRUhhaVm1Of8Dcvo/0.15-Device-setup?node-id=142%3A20

as per https://github.com/learningequality/kolibri/issues/7933#issuecomment-865223151

jredrejo:
## Summary

- Refactored current wizard to use state machines 
Machine is defined in kolibri/plugins/setup_wizard/assets/src/wizardMachine.js
image

- Create a new Initialize Device screen, according to https://www.figma.com/file/w3raYcGRUhhaVm1Of8Dcvo/0.15-Device-setup?node-id=229%3A63
wizard2

Using admin credentials:
admin

### IMPORTANT NOTE:
One part of the proposed design (provision using an admin account) is not implemented yet. However, this PR touches so many files and the probability of conflicts in the short term is so high that I would like to see this merged as soon as possible, using another PR (if needed) to add this missing functionality. The most important and difficult part is done, so there's room to test and check if all the added code makes sense.

## References

Closes: #8166 

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated


rtibbles (commented):
Not sure I am completely understanding how exactly the state machine drives the routes, but the overall state machine does seem easy to grok.

My one suggestion would be to use provide/inject for passing the `service` object down to child components, rather than vuex state, as it does not need to be reactive.

https://vuejs.org/v2/api/#provide-inject

jredrejo:
> Not sure I am completely understanding how exactly the state machine drives the routes, but the overall state machine does seem easy to grok.
> 
> My one suggestion would be to use provide/inject for passing the `service` object down to child components, rather than vuex state, as it does not need to be reactive.
> 
> https://vuejs.org/v2/api/#provide-inject

Thank you @rtibbles . That was a great suggestion, I have done it and works perfectly

kolibri/core/assets/src/views/sync/SelectAddressModalGroup/api.js
```diff
@@ -27,6 +27,21 @@ function channelIsAvailableAtLocation(channelId, location) {
     });
 }
 
+function versionCompare(version) {
+  //check if version is >= 0.15.0
+  const v = version.split('.');
+  if (parseInt(v[0]) > 0) return true;
+  if (parseInt(v[1]) >= 15) return true;
+  return false;
+}
```
indirectlylit: may want to consider doing some basic validation of expected version strings and raising an exception if it doesn't match

kolibri/plugins/setup_wizard/assets/src/views/onboarding-forms/SetupMethod.vue
```diff
@@ -41,58 +60,107 @@
...
+            device_id: address.id,
+          },
+        });
+      },
     },
     $trs: {
-      setupMethodHeader: {
-        message: 'Create or import facility',
+      initializeDevice: {
+        message: 'Initialize device',
```
radinamatic: We changed the wording for the heading here, [check the latest on Figma](https://www.figma.com/file/w3raYcGRUhhaVm1Of8Dcvo?node-id=229:63#94924868):

Selection_005

radinamatic:
I know this is not fully implemented, so I'm not sure if the issues below are due to that.

EXE asset for this build has been installed on 2 VMs, Windows 7 with a working facility (learner, class), and on Windows 10 I attempted to go through importing a single user (firewall was off, devices can see each other):

Step 1 | Step 2
--- | ---
Win10 (start)  Running  - Oracle VM VirtualBox_001 | Win10 (start)  Running  - Oracle VM VirtualBox_002

```TypeError: can't access property "name", _vm.facility is undefined``` error in the console.

[logs.zip](https://github.com/learningequality/kolibri/files/6926836/logs.zip)

File: kolibri/core/device/permissions.py [hunk 3]
```diff
@@ -47,3 +47,18 @@ def has_permission(self, request, view):
 class IsSuperuser(DenyAll):
     def has_permission(self, request, view):
         return request.user.is_superuser
+
+
+class LODUserHasSyncPermissions(BasePermission):
+    """
+    When a device for LOD is provsioned, after the first user has been synced,
+    this user must be able to sync more users
+    """
```
indirectlylit: I see a lot of `LOD` references, which I assume stands for the user-facing concept we settled on, "Learn-only device"? I would personally recommend spelling out `learn_only_device` for clarity rather than using an acronym, even though it's more verbose.

In other places in this PR we also use `subset_of_users_device` and even `soud` so it would be good to tighten up our naming consistency.

(For future-proofing I still believe `limited_device` is the most useful technical name because it is not as specific but that never caught on)

kolibri/plugins/device/assets/src/views/WelcomeModal.vue
```diff
@@ -70,6 +80,14 @@
...
         context: 'Welcome message for user indicating that they need to import resources.',
       },
+      posLODWelcomeMessage1: {
+        message: 'The first thing you should do is import some channels to this device',
+        context: 'Welcome message for user which appears after provisioning a Learner Only Device.',
+      },
+      posLODWelcomeMessage2: {
+        message: `The user reports, lessons, and quizzes in  '{facilityName}' will not display properly until you import the resources associated with them.`,
+        context: 'Welcome message for user indicating that they need to import resources.',
+      },
```
indirectlylit: since message IDs are shown to translators, these should be spelled out using the user-facing terminology:

```suggestion
      learnOnlyDeviceWelcomeMessage1: {
        message: 'The first thing you should do is import some channels to this device',
        context: 'Welcome message for user which appears after provisioning a Learner Only Device.',
      },
      learnOnlyDeviceWelcomeMessage2: {
        message: `The user reports, lessons, and quizzes in  '{facilityName}' will not display properly until you import the resources associated with them.`,
        context: 'Welcome message for user indicating that they need to import resources.',
      },
```

jtamiace:
Great work José! 👏 Dropping in my requests for some small design improvements:

Increase the space between the option `Import all data from an existing facility` and `Learn-only device` by 16px:
Screen Shot 2021-08-03 at 2 31 34 PM

---

Include the Kolibri version number along with the localhost link in the caption line of the radio options here:
| Implementation  | Design |
| ------------- | ------------- |
| Screen Shot 2021-08-03 at 2 31 46 PM | Screen Shot 2021-08-03 at 2 36 34 PM |

---

The header here should read `Import individual user accounts` (user not plural), and the username shouldn't have a character counter, since that's only useful for creating user accounts:
| Implementation  | Design |
| ------------- | ------------- |
| Screen Shot 2021-08-03 at 2 32 28 PM | Screen Shot 2021-08-03 at 2 38 39 PM |

---

Finally, after clicking `Continue` - this modal should take the user to the `Device > Channels` page, not the Learn home page:
Screen Shot 2021-08-03 at 2 33 08 PM

kolibri/core/device/permissions.py
rtibbles: I think for dev facing code, I'd prefer we stick to `subset_of_users_device` which is then consistent with the previous references we have made to this. We can still call it a `Learner only device` in user facing strings.

kolibri/core/tasks/api.py
```diff
@@ -1101,6 +1111,76 @@ def startdeletefacility(self, request):
...
+    )
+    def startprovisionsoud(self, request):
+        baseurl = request.data.get("baseurl", None)
+        facility_id = request.data.get("facility_id", None)
+        username = request.data.get("username", None)
+        password = request.data.get("password", None)
+        user_id = request.data.get("user_id", None)
+        device_name = request.data.get("device_name", None)
+        if user_id is None:
+            auth_url = urljoin(baseurl, reverse("kolibri:core:session-list"))
```
rtibbles: I know I suggested this, but wonder if we should create a separate public endpoint for this for future stability?

I had suggested something similar here: https://github.com/learningequality/kolibri/issues/8170 using Basic Auth to allow return of either one user or multiple users, depending on permissions.

kolibri/core/tasks/api.py
```diff
@@ -1101,6 +1111,76 @@ def startdeletefacility(self, request):
...
         return Response(resp)
 
+    @decorators.action(
+        methods=["post"],
+        detail=False,
+        permission_classes=[
+            IsSuperuser | NotProvisionedCanPost | LODUserHasSyncPermissions
+        ],
+    )
+    def startprovisionsoud(self, request):
```
rtibbles: Not required, but this might be a place to try out the new tasks API.

An example is here: https://github.com/learningequality/kolibri/blob/release-v0.15.x/kolibri/core/content/tasks.py#L28

rtibbles (commented):
A few comments and suggestions, but this is looking solid. Interested to see the results of manual testing!

radinamatic:
Setting up a `learn-only-device` was successful when I used my dev VM on Ubuntu as a server, and the same Win 10 as a learner device. 

1. I find this screen a bit confusing, as 3 different terms (import, load, sync) seem to be used for what is essentially the same action. @jtamiace thoughts?
    
    Win10 (start)  Running  - Oracle VM VirtualBox_005


2. Once the setup is completed, the first imported learner is automatically signed in, but on the **Learn** page, the phrase `As an administrator, you can import channels` looks out of place and confusing. Administrator on a `learn-only-device`?

    Win10 (start)  Running  - Oracle VM VirtualBox_008

    Actually, I'd like to suggest the change of the wording for this link to something more more straightforward, like `Import channels on your device`. cc @jtamiace 

3. Keyboard navigation seems to be working well, but focus outline is uneven or cut out in some places.

    1 | 2
    --- | ---
    Win10 (start)  Running  - Oracle VM VirtualBox_006 | Win10 (start)  Running  - Oracle VM VirtualBox_009

pcenov:
I installed this [build](https://buildkite.com/learningequality/kolibri-windows/builds/1554) on 2 separate Windows 10 laptops (physical devices, not VMs) connected to the same WiFi network with disabled firewall and was not able to go through the Learn-only device flow as the URL of the server wouldn't show in the Select network address modal:



Also if I attempt to import all data from an existing facility where I can manually add the server URL I get a different 404 error and a Something went wrong screen:



Importing a facility from the Facilities tab is also not possible:



I also tried with a different build sent over to me by @radinamatic - kolibri-0.15.0a2.dev0+git.60.g56381baf-unsigned.exe and I'm experiencing the same issues.

Finally I'm facing the same issues when using Ubuntu and Win 10 virtual machines.

Logs: 
[logs.zip](https://github.com/learningequality/kolibri/files/6931952/logs.zip)

pcenov:
https://learningequality.testquality.com defect [D5](https://learningequality.testquality.com/project/12317/result/2461464/defect/8056).

kolibri/core/assets/src/views/sync/SelectAddressModalGroup/api.js
jredrejo: Done

kolibri/plugins/device/assets/src/views/WelcomeModal.vue
jredrejo: Done

jredrejo:
> Great work José!  Dropping in my requests for some small design improvements:
> 
> Increase the space between the  ....

Thank you @jtamiace ! So good your eye fixed my design blindness. I think I have completed all your suggestions. I have updated the animated gif from this PR summary , so you can check them there.

jredrejo:
Thanks @radinamatic , I'll comment inline:

> Setting up a `learn-only-device` was successful when I used my dev VM on Ubuntu as a server, and the same Win 10 as a learner device.
> 
> 1. I find this screen a bit confusing, as 3 different terms (import, load, sync) seem to be used for what is essentially the same action. @jtamiace thoughts?
>    Win10 (start)  Running  - Oracle VM VirtualBox_005

I see what you mean. I don't have any problem in changing those terms, For  me "Loading user.." might be "Importing user" but the other terms you have underlined come directly from the syncing task command. Changing them can be done but I think it should be discussed with @bjester  because they are real, when it says _syncing_ is because it's _syncing_, as this is a part of the provisioning.

> 2. Once the setup is completed, the first imported learner is automatically signed in, but on the **Learn** page, the phrase `As an administrator, you can import channels` looks out of place and confusing. Administrator on a `learn-only-device`?
>    Win10 (start)  Running  - Oracle VM VirtualBox_008
>    Actually, I'd like to suggest the change of the wording for this link to something more more straightforward, like `Import channels on your device`. cc @jtamiace

That's not part of this PR but I don't have any problem in changing it if needed.

> 3. Keyboard navigation seems to be working well, but focus outline is uneven or cut out in some places.
>    
>    
>    
>    1
>    2
>    
>    
>    
>    
>    Win10 (start)  Running  - Oracle VM VirtualBox_006
>    Win10 (start)  Running  - Oracle VM VirtualBox_009

I'll need you to explain to me how to reproduce that and what it exactly means

kolibri/plugins/setup_wizard/assets/src/views/onboarding-forms/SetupMethod.vue
jredrejo: Thanks, done, updated in the animated gif above.

jredrejo:
> I installed this [build](https://buildkite.com/learningequality/kolibri-windows/builds/1554) on 2 separate Windows 10 laptops (physical devices, not VMs) connected to the same WiFi network with disabled firewall and was not able to go through the Learn-only device flow as the URL of the server wouldn't show in the Select network address modal:
> 
> 
> 
> Also if I attempt to import all data from an existing facility where I can manually add the server URL I get a different 404 error and a Something went wrong screen:
> 
> 
> 




Thank you @pcenov  I think I have been able to reproduce the problem and should be fixed now

For the above comment you did, I haven't been able to reproduce it, but it looks like it's the same root problem. Please, test it when you can to confirm it.

> Importing a facility from the Facilities tab is also not possible:
> 
> 
> 
> I also tried with a different build sent over to me by @radinamatic - kolibri-0.15.0a2.dev0+git.60.g56381baf-unsigned.exe and I'm experiencing the same issues.
> 
> Finally I'm facing the same issues when using Ubuntu and Win 10 virtual machines.
> 
> Logs:
> [logs.zip](https://github.com/learningequality/kolibri/files/6931952/logs.zip)

jredrejo:
> I know this is not fully implemented, so I'm not sure if the issues below are due to that.
> 
> EXE asset for this build has been installed on 2 VMs, Windows 7 with a working facility (learner, class), and on Windows 10 I attempted to go through importing a single user (firewall was off, devices can see each other):
> 
> Step 1	Step 2
> Win10 (start) Running - Oracle VM VirtualBox_001	Win10 (start) Running - Oracle VM VirtualBox_002
> `TypeError: can't access property "name", _vm.facility is undefined` error in the console.
> 
> [logs.zip](https://github.com/learningequality/kolibri/files/6926836/logs.zip)

Thanks @radinamatic  That should be fixed  now. It would be great if you have a chance to test it again in Windows.

jtamiace:
> I find this screen a bit confusing, as 3 different terms (import, load, sync) seem to be used for what is essentially the same action.

@jredrejo @radinamatic I'm not sure if the implementation in this PR is using the same components as the task manager in `Device > Channels`, but we can remove almost all the lines except `Finished` and `'learner' from 'ZukVillage' successfully loaded to this device` to be consistent with both the designs and the current 0.14 implementation of facility import. I'd argue all those other details aren't that useful to know at this stage in the setup process. That way we'd at least remove the word 'Sync'. I think there was a reason we went "Loaded/Loading" instead of import but I can't quite remember or find any notes on why...

importuser

---

> Once the setup is completed, the first imported learner is automatically signed in, but on the Learn page, the phrase As an administrator, you can import channels looks out of place and confusing. Administrator on a learn-only-device?

> Actually, I'd like to suggest the change of the wording for this link to something more more straightforward, like Import channels on your device

Agree it should change, and also the action or statement here is conditional depending on who's logged in. Here's my suggestion, comparing what it is for super admins on 0.14 and proposal for learners on after this setup

Screen Shot 2021-08-05 at 12 47 08 PM

radinamatic:
> ...be consistent with both the designs and the current 0.14 implementation of facility import. I'd argue all those other details aren't that useful to know at this stage in the setup process. That way we'd at least remove the word 'Sync'. 

@jtamiace Yes, that does look much cleaner, let's go with that @jredrejo!

> action or statement here is conditional depending on who's logged in. Here's my suggestion, comparing what it is for super admins on 0.14 and proposal for learners on after this setup. 128411908-aaa5bf91-642e-4aa3-9991-54174571d9c1

Agreed with this proposal, although I would argue to change that `As administrator...` into straightforward `Import channels...` link even for the administrators, but we might deal with that outside the scope of this PR.

radinamatic:
@jredrejo I installed the most recent asset and repeated the scenario, with the only difference that Windows 10 was setup as a class server (with full facility, users, channels), and Windows 7 as a learn-only-device.

Installation on both went without issues, but after I select the Win 10 device to import from, and I input the learner username and password, nothing happens in the frontend, but there is a 500 error in the terminal at `/api/tasks/facilitytasks/startprovisionsoud/` 

Select device | Import learner
--- | ---
Win7 (start)  Running  - Oracle VM VirtualBox_001 | Win7 (start)  Running  - Oracle VM VirtualBox_003

[logs.zip](https://github.com/learningequality/kolibri/files/6942880/logs.zip) Windows 7 (learn-only)

@pcenov Please try to test this using both EXE and DEB assets, to see if this is a a Windows-only issue.

pcenov:
Hi @jredrejo and @radinamatic, I tested both the latest [Windows and Debian](https://buildkite.com/learningequality/kolibri-python-package/builds/3292) builds and I'm facing exactly the same issues as before. I tested on both 2 real Windows laoptops and 2 Ubuntu VMs and was not able to go neither through the learner-only import as the device URL is not discovered (maybe there should be an option to manually add the URL) or through existing facility import where after I manually add the URL and click Continue I see a 'Something went wrong' error.

Logs:
[UbuntuLogs.zip](https://github.com/learningequality/kolibri/files/6944586/UbuntuLogs.zip)
[WindowsLogs.zip](https://github.com/learningequality/kolibri/files/6944583/WindowsLogs.zip)

Ubuntu VM screenshots:



Windows laptop:

kolibri/core/tasks/api.py
jredrejo: Done in https://github.com/learningequality/kolibri/pull/8185/commits/47b486067fc25bf0748c326f9f07e1dd4e0c9b0a

jredrejo: Adapted BasicAuth to include the facility_id as a realm when login https://github.com/learningequality/kolibri/blob/a56d04850e9c765b50a94ec0b25c0279ed197dfb/kolibri/core/auth/api.py#L82-L106

kolibri/core/tasks/api.py
jredrejo: Done in https://github.com/learningequality/kolibri/pull/8185/commits/1f5634cf30b3bab33acff3ea1f6f37d67a73f780

pcenov:
Hi @jredrejo, here are the screenshots and log files from today's testing on this [build](https://buildkite.com/learningequality/kolibri-python-package/builds/3297)
Logs:
[UbuntuLogs.zip](https://github.com/learningequality/kolibri/files/6953795/UbuntuLogs.zip)
[Win10Logs.zip](https://github.com/learningequality/kolibri/files/6953796/Win10Logs.zip)
Screenshots:


I'm getting the 500 error in 3 different test setups - importing a learner from Windows 10 to another Windows 10 physical device; importing a learner from Ubuntu VM to another Ubuntu VM; importing a learner from Ubuntu VM to Windows 10 physical device.

radinamatic:
Still getting `/api/soudtasks/` 500 errors on the [latest build](https://buildkite.com/learningequality/kolibri-windows/builds/1579):
* Windows 10 was setup as a class server (with full facility, users, channels), and Windows 7 as a learn-only-device. 
* After I input the individual learner username and password to import, nothing happens in the frontend, but there is a 500 error in the terminal at `/setup_wizard/api/soudtasks/` 

Learner 1 | Learner 2
--- | ---
Win7 (start)  Running  - Oracle VM VirtualBox_006 | Win7 (start)  Running  - Oracle VM VirtualBox_007

[logs-Win10.zip](https://github.com/learningequality/kolibri/files/6961851/logs-Win10.zip) Class server
[logs-Win7.zip](https://github.com/learningequality/kolibri/files/6961852/logs-Win7.zip) Learn-only

radinamatic:
@jredrejo Great job, both @pcenov and I managed to import individual users today! 💯  🎉  

Two additional pending things that would be good to address before we merge this ([both mentioned by @jtamiace above](https://github.com/learningequality/kolibri/pull/8185#issuecomment-893747371))

Delete these lines from step 3. | `FINISH` should forward to Device > Channels tab _*before*_ opening the 'Welcome to Kolibri!' modal
--- | --- 
Win7 (start)  Running  - Oracle VM VirtualBox_009 | Win7 (start)  Running  - Oracle VM VirtualBox_010

indirectlylit:
Just flagging that I see `lod` 101 times and `soud` 35 times. Are these synonyms?

I understand that at this point the ship might have sailed, and it may be too much effort to standardize terminology.

jredrejo:
> Just flagging that I see `lod` 101 times and `soud` 35 times. Are these synonyms?
> 
> I understand that at this point the ship might have sailed, and it may be too much effort to standardize terminology.

SoUD = Subsets of Users Device
LOD = Learner Only Device

In practice they are synonyms (unless a SoUD is used only by a coach)

This PR has taken several weeks to be done. In the beginning it was SoUD and there are already some APIs using this naming to start the sync process. Then I was told that LOD was the new name, then someone told SoUD is better for devs, then a poll in Slack has been done to decide the name.... Honestly, I don't care. If someone takes a final decission a quick search and replace can be done in this PR. In any case, as I said above, SoUD is already in use in some other parts of the code and likely LOD too.

indirectlylit:
yeah sorry to keep bringing it up. for the sake of future devs I would love to just finally pick one and stick with it, but given our limited capacity maybe this is one of those things we just live with

kolibri/core/auth/api.py
```diff
@@ -74,6 +79,33 @@
 from kolibri.plugins.app.utils import interface
 
 
+class BasicAuthenticationFacility(BasicAuthentication):
```
bjester: Morango [allows something similar](https://github.com/learningequality/morango/blob/master/morango/api/permissions.py#L13) which I also needed to take advantage of by passing the facility ID.

rtibbles: Oh, I was unaware of the morango implementation when @jredrejo and I discussed this. Would it be worth just reusing that?

bjester: The Morango class is a little more generic/flexible which doesn't necessarily make it super clear that we're passing through additional parameters, but I think that also makes it less likely to change and little more future-proof, so reusing it probably wouldn't hurt!

kolibri/plugins/setup_wizard/assets/src/views/importLODUsers/LoadingTaskPage.vue
```diff
@@ -0,0 +1,191 @@
...
+            this.loadingTask = {
+              ...soudTasks[0],
+              facility_name: this.facility.name,
+              full_name: this.fullName(soudTasks[0].id),
+              device_id: this.state.value.device.id,
+            };
+            if (this.loadingTask.status === TaskStatuses.COMPLETED) this.finishedTask();
+          } else this.isPolling = false;
+        });
+        if (this.isPolling) {
```
bjester, line 119: Should this perhaps wait for the previous `fetchCollection` to complete before "queuing" the next poll attempt?

bjester (commented):
Looks pretty good to me! Lots of QA and review already. Only comments are that you could potentially re-use Morango's `BasicMultiArgumentAuthentication` class with a small refactor, and a question about when the frontend triggers the next poll attempt.

kolibri/plugins/setup_wizard/assets/src/views/importLODUsers/LoadingTaskPage.vue
jredrejo: I'm not sure of this because I'm not using `catch` in case an error  happens. If there were errors, polling would stop. I've prefered to ignore possible errors in this point.

kolibri/core/auth/api.py
jredrejo: I think the facility ID is needed because there might be these scenarios in one device:
one user has been copied from one facility to another, having seme credentials in both facilities
or by casuality someone  has chosen the same username and password in two different facilities 

How do you know which user is trying to use the authentication?

bjester: @jredrejo You're correct, it needs the facility ID for authentication. The point that I was making is that Morango has a class that allows for passing arbitrary additional parameters, like the facility ID, just less explicitly than you have coded here. So it's possible we could simply use that class instead.

kolibri/plugins/setup_wizard/assets/src/views/importLODUsers/LoadingTaskPage.vue
rtibbles: I think we can probably do some follow up clean up more generally, and maybe have a reusable helper resource to replace or augment `TaskResource` that can consolidate some of the polling/fetching/task initiation behaviour?

jredrejo: yep, but I think that should be a reason for a different issue/PR after this is merged. That's going to be used in multiple places

rtibbles: Definitely not in scope for this PR.

kolibri/core/auth/api.py
jredrejo: @bjester  @rtibbles  done, but beware that I've had to fix some issues that were important in kolibri and have raised when using this authentication. They are at #8298 for a separate aproval, but cherry-picked here or authentication didn't work

jredrejo: Thanks again @bjester  for your help, everything is working now using `BasicMultiArgumentAuthentication` from Morango

bjester (approved):
Fantastic work, José!

kolibri/plugins/setup_wizard/api.py
```diff
@@ -115,18 +127,49 @@ def provisiondevice(self, request):
...
     startdeletefacility = property()
+
+
+class SetupWizardSoUDTaskView(TasksViewSet):
+    """
+    Needed because TasksViewSet permissions don't allow
+    fetch list of task or create task from a provisioning device
+    """
+
+    permission_classes = [HasPermissionDuringSetup | HasPermissionDuringLODSetup]
```
vkWeb, line 175: @jredrejo a flag - we will not need to inherit `TasksViewSet` after merging https://github.com/learningequality/kolibri/pull/8303.

vkWeb: The reason I'm very hesitant to inheriting `TasksViewSet` is because the whole purpose of the task api refactor is to prevent such inheritance :)

kolibri/plugins/setup_wizard/tasks.py
```diff
@@ -0,0 +1,149 @@
...
+    ],
+)
+def startprovisionsoud(
+    username=None,
+    password=None,
+    baseurl=None,
+    facility_id=None,
+    user_id=None,
+    extra_metadata={},
+):
```
vkWeb, line 144: Right now, the task function don't get passed `extra_metadata` as an argument because it gets popped off by the `Job` constructor. 

@rtibbles I think we should not pop the `extra_metadata`? What do you say?

kolibri/plugins/setup_wizard/api.py
```diff
@@ -115,18 +127,49 @@ def provisiondevice(self, request):
...
+
 
 class SetupWizardFacilityImportTaskView(FacilityTasksViewSet):
     """
     An open version of FacilityTasksViewSet for the purposes of managing the
     import-facility task during setup
     """
 
-    permission_classes = (HasPermissionDuringSetup,)
+    permission_classes = [HasPermissionDuringSetup | HasPermissionDuringLODSetup]
```
vkWeb, line 158: We are inheriting `FacilityTasksViewSet` so that we can send request to `create` API during setup process, right? Did I get it correct? 

If yes then we will not need to inherit `FacilityTasksViewSet` after merging #8303.

vkWeb (commented):
Some task API related thoughts.

kolibri/plugins/setup_wizard/api.py
jredrejo: Yep, thanks, @vkWeb,   I bear it in mind. As soon as #8303 is merged I will be really happy to remove this subclassing but in the meantime we need this to be working to be able to test the sync processs. 
I'll try to prepare a PR this week with the changes needed so you can check it and we can merge it right after #8303 is merged too.

vkWeb: Sir, that sounds like a good plan!

kolibri/plugins/setup_wizard/tasks.py
rtibbles: `extra_metadata` is intended to be specifically for metadata that is tracked on the job, in addition to the progress metadata, so it really should just be concerned with tracking the job, which is why it gets popped.

rtibbles:
Have filed a follow up issue for the main extent behavioural part. Let's get this merged, as the PR is working, code is approved, everything else is just gravy.

jredrejo:
> @jredrejo Great job, both @pcenov and I managed to import individual users today!  
> 
> Two additional pending things that would be good to address before we merge this ([both mentioned by @jtamiace above](https://github.com/learningequality/kolibri/pull/8185#issuecomment-893747371))
> 
> Delete these lines from step 3.	`FINISH` should forward to Device > Channels tab __before__ opening the 'Welcome to Kolibri!' modal
> Win7 (start) Running - Oracle VM VirtualBox_009	Win7 (start) Running - Oracle VM VirtualBox_010

@radinamatic the second thing was done the day after the comment.
For the lines to be removed, It does not affect to functionality. I'd like to have this merged now for two reasons: it will allow us to test it better (hacking sessions included) and the longer this PR is unmerged, the biggest is the possibility of conflicting with code in Kolibri, due to its size and the amount of affected files.
Please, let's merge it and fill an issue to remove those lines afterwards.

kolibri/plugins/setup_wizard/tasks.py
vkWeb: Yes, that's alright. 

Then here the task function accesses `extra_metadata` which will always be an empty dict, we should shift that part to the validator and include `job_data` in the returning dict of validator.

PR #8186, opened 2021-07-06
Create API endpoint for Tasks backend

vkWeb:
## Summary

This PR adds `POST /api/tasks/` endpoint method. 

**Note:** I couldn't add tests as I'm moving out of my town this morning. Tests will be added after I return from the trip (most probably 11th July).

## Reviewer guidance

Does the endpoint handle bulk and normal tasks properly? Are permissions working as expected? 

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated


kolibri/core/tasks/api.py
```diff
@@ -314,11 +393,15 @@ def queues(self):
         return [queue, priority_queue]
 
     def default_permission_classes(self):
-        # exclusive permission for facility management
-        if self.action == "startexportlogcsv":
+        if self.action in ["list", "deletefinishedtasks"]:
```
rtibbles, line 406: Presumably this would all just go away when we transfer all these permissions over to the tasks themselves.

rtibbles:
One open question I am realizing is how we do permission checks for retrieving task data - we may need to filter the list of registered task types that the user has permission for, and then retrieve only the tasks that match that list?

kolibri/core/tasks/api.py
vkWeb: Yep that's true but for the time being this change fixes an issue we discussed on Slack.

vkWeb:
> One open question I am realizing is how we do permission checks for retrieving task data - we may need to filter the list of registered task types that the user has permission for, and then retrieve only the tasks that match that list?

As discussed on Slack, for now we'll be moving ahead with `CanManageContent` permission for all `/api/tasks/` endpoints. We plan to revisit this later when we get more insight on our new task submission workflow. 

<hr>

@rtibbles I've added very thorough tests. Now, this PR is ready for final review and merge! :tada:

kolibri/core/tasks/api_urls.py
```diff
@@ -1,10 +1,12 @@
 from rest_framework import routers
 
+from .api import BaseViewSet
 from .api import FacilityTasksViewSet
 from .api import TasksViewSet
 
 router = routers.SimpleRouter()
 router.register("tasks", TasksViewSet, base_name="task")
 router.register("facilitytasks", FacilityTasksViewSet, base_name="facilitytask")
+router.register("", BaseViewSet, base_name="managetask")
```
rtibbles: I would prefer not to use a blank base URL path here, it feels like an anti-pattern to me.

We can have this under `managetask` for now, but as we finish migrations of tasks to the new API, I would expect that `BaseViewSet` will be renamed `TasksViewSet` and that this will all be under `"tasks"` with a basename of `"tasks"`, as at that point, we can dispense with the other two endpoints!

kolibri/core/tasks/tasks.py
```diff
@@ -0,0 +1,37 @@
+"""
```
rtibbles: The task defined in here should live in `kolibri/core/content/tasks.py`

kolibri/core/tasks/job.py
```diff
@@ -357,28 +359,17 @@ def enqueue_at(self, datetime, interval=0, repeat=0, args=(), kwargs={}):
...
             *args,
             job_id=self.job_id,
             group=self.group,
             cancellable=self.cancellable,
             track_progress=self.track_progress,
+            extra_metadata=self.extra_metadata,
             **kwargs
         )
-
+        self.extra_metadata.clear()
```
rtibbles: Not understanding the purpose of `self.extra_metadata` in this file - also, if multiple jobs are being created, there's a risk we are sharing the extra_metadata dict among them.

Given that this `self.extra_metadata` is always an empty object, it's not doing anything additional over the default (and safer) setting of it inside the Job constructor: https://github.com/learningequality/kolibri/blob/develop/kolibri/core/tasks/job.py#L189

kolibri/core/tasks/main.py
```diff
@@ -133,6 +134,9 @@ def __queue():
     Priority.HIGH: priority_queue,
 }
 
+# This storage instance should be used to access job_storage db
+job_storage = Storage(connection=connection)
```
rtibbles: This may be the cause of the postgres errors you are seeing - make sure to follow the same lazy object pattern that the other objects defined in this module follow, rather than instantiating the `Storage` object in the module scope directly.

kolibri/core/tasks/api.py
```diff
@@ -207,8 +244,50 @@ def list(self, request):
...
+
+        API endpoint:
+            POST /api/tasks/
+
+        Request payload parameters:
+            - `task` (required): a string representing the dotted path to task function.
+            - all other key value pairs are passed to the validator if the
+              task function has one otherwise they are passed to the task function itself
+              as keyword args.
+
```
rtibbles, line 258: We don't need to address this in this PR, but one thought that occurs is that it would be useful to be able to schedule tasks through this endpoint as well.

Any thoughts on how we could extent the API to allow for that? Possibly optional `enqueue_at` and `enqueue_in` arguments?

rtibbles (changes requested):
A bit of cleanup to do, but this is looking very promising over all!

kolibri/core/tasks/api_urls.py
vkWeb: @rtibbles the only reason to use a blank base URL path was to have urlconf generated as `/api/tasks/` not `/api/tasks/tasks/`. Don't you think the former looks more beautiful?

kolibri/core/tasks/tasks.py
vkWeb: Okay, will do. 

Also, I've some questions regarding separation of tasks (related to breaking changes..). Will talk to you on Slack about those!

kolibri/core/tasks/job.py
vkWeb: `self.extrametadata` plays a very important role there.

If a task function wants to set extra metadata, it can do that in its validator by doing `task_func.extra_metadata={"some": "thing"}`. This `extra_metadata` is binded to the registered job instance's `extra_metadata`, which gets passed to the job constructor via `_ready_job`. 

Every job that is registered will have its own `RegisteredJob.extra_metadata` so there's no chance of metadata being shared wrongly between multiple jobs.

Have a look at the demo `tasks.py` module to see it practically.

kolibri/core/tasks/main.py
vkWeb: Ohhh! Will fix this! :tada: 

But sir, why is this the case though? I mean any specific reason for following lazy object initialization...?

kolibri/core/tasks/api.py
vkWeb: Yes, we can extent our API to make it do that. 

Not only `enqueue_at` and `enqueue_in` but also `repeat` & `interval`. 

So, here we have two choices of request payloads:

**1.** 
```
{
  "task": "....",
  "enqueue_at": "July 27 2021  2:00PM",
  "repeat": 5,
  "interval": 86400, (in seconds)
  ... (args that get passed to the task function)
}
```  
**2.**
```
{
  "task": "....",
  "enqueue_at": {
        "time": "July 27 2021  2:00PM",
        "repeat": 5,
        "interval": 86400, (in seconds)
       }
  ... (args that get passed to the task function)
}
```  

My choice is the second one as it offers zero chance of conflict with names of task function arguments. What's your say?

kolibri/core/tasks/api_urls.py
rtibbles: The general pattern we have for API URLs has <appname>/<endpoint> - so I'd rather keep with that, even if it looks repetitive in this case.

kolibri/core/tasks/job.py
rtibbles: I missed that in my initial review - but the way this is being used inside the validator is concerning.

In the last PR, there was some clean up to ensure that we were not using the registered task object as a transitory state object during task registration. This PR has reintroduced some of that concerning use again.

As a general rule, the validator functions should be pure functions without side effects - if we need to set the `extra_metadata` property - it seems that returning `extra_metadata` as a key on the dict that gets returned from the validator function is the appropriate way to do this, not by assigning it to the global registered task object.

Again, the issue here is if two tasks are being enqueued at the same time, with shared memory at play, then the use of the global object could lead either to the extra_metadata for one queuing being applied to another, or the extra metadata for one being cleared while another is being queued. It's a recipe for hard to narrow down bugs.

kolibri/core/tasks/main.py
rtibbles: I am not completely certain that this is the cause of your Postgresql failures - but the main reason for the lazy initialization is to allow objects from this module to be imported before any app initialization has happened, without fear of database connections being initialized.

kolibri/core/tasks/api.py
rtibbles: I do prefer the second option, yes - there are some other options, but this seems the most sensible.

kolibri/core/tasks/api_urls.py
vkWeb: Okay - so I'll remove the new endpoint that I've added. Since, `TasksViewSet` inherits from `BaseViewSet`, until `TasksViewSet` is alive it'll handle the endpoints. After migration, I'll remove the two viewsets that we have and will rename `BaseViewSet` to `TasksViewSet`.

kolibri/core/tasks/job.py
vkWeb: The issue I can see is not with two different tasks because they'll have their own registered job instance so no issues with shared memory there. But yes if the same task is being enqueued by two different threads, then there can be race conditions with this approach. Is my thinking correct sir @rtibbles?

kolibri/core/tasks/main.py
vkWeb: Got it!

kolibri/core/tasks/job.py
rtibbles: Yes, it's the potential sharing of state between two tasks of the same type that I am concerned about.

vkWeb: Will be fixed. :rocket: 

From now on, I'll keep this lesson in mind.

vkWeb:
@rtibbles ready for merge :rocket:

rtibbles (approved):
All comments addressed.

kolibri/core/tasks/api.py
```diff
@@ -207,8 +244,60 @@ def list(self, request):
...
+                extra_metadata = validator_result.get("extra_metadata")
+                if extra_metadata is not None and not isinstance(extra_metadata, dict):
+                    raise serializers.ValidationError(
+                        "In the dict returned by validator, 'extra_metadata' must be a dict."
+                    )
+
+                request_data = validator_result
+
+            job_id = registered_job.enqueue(**request_data)
```
jredrejo, line 294: `task` key should be removed from the `request_data`. With the current implementation the task function is forced to have an unused arg called `task`

rtibbles: Given that it's already been validated to have `"task"` in the data, can use:
```
funcstr = request_data.pop("task")
```

Above - using pop instead of get to address this!

vkWeb: Yes, this will be fixed in the upcoming PR that aims to decouple some tasks to their respective places.

jredrejo: hi , @vkWeb  I'm using the current implementation for a new feature. If you plan to do changes on the api it's important to know it.
There are a some things I have missed using this new tasks api:
1. Possibility to select a different queue (only normal or high are available)
2. A JS api (I've done my own but having a general to inherit would be great)
3. Possibility to change the permissions of some of the `TaskViewSet` functions. I have had to create a new class inheriting it just to be able to change the permissions to see the list of tasks.
4. I have read in the specs the possility to use `groupName` to group tasks but it's not implemented as far as I have seen

In any case, none of these things are fundamental and I've found workarounds for them, but having them would make a dev life easier. The general work done on the tasks is great and being able to register new tasks from a plugin is awesome. I have had to delete code from a previous implementation and that's my favourite refactoring ;)

vkWeb: Hi sir @jredrejo. I'm reallyyy happy to hear that! 

Right now, the Tasks API and the tasks backend are under active refactoring. In the span of next one month we can expect our tasks backend to give a much more delightful developer experience. I'll keep you updated on the developments via Slack!     

> Possibility to select a different queue (only normal or high are available).
> I have read in the specs the possility to use groupName to group tasks but it's not implemented as far as I have seen.

After implementing single worker pool, we'll have the ability to give tasks any queue we want, later on we can fetch/delete tasks by any particular queue. Moreover, `Queue` interface will be removed, queue will just be a string column on the job storage table. Then, `high` and `normal` priority levels will be used to decide which task should be fetched first by the workers.   

I've planned to implement single worker pool after I shift tasks from the `kolibri.core.tasks.api` monolith to their respective django apps. Once shifting is complete, we'll get the freedom to mould our tasks backend in whatever way we want without disturbing the working of tasks itself.

> A JS api (I've done my own but having a general to inherit would be great)

A JS API...? I'm not sure I understand..?

> Possibility to change the permissions of some of the TaskViewSet functions. I have had to create a new class inheriting it just to be able to change the permissions to see the list of tasks.

The `register_task` decorator has a `permission_classes` argument that'll allow flexible setting of permissions for any particular task. I would request you try not to inherit TaskViewSet, let us discuss the problem to see the pain that's forcing us to inherit.

vkWeb: @jredrejo just to flag, all the upcoming PRs related to tasks backend will target https://github.com/learningequality/kolibri/tree/tasks_api_refactor branch so that it don't cause trouble in the 0.15 release.

jredrejo: Hello again @vkWeb  I have been working on this in this PR  #8185 so all the code I'll mention here is there if you want to take a look

> > A JS api (I've done my own but having a general to inherit would be great)
> 
> A JS API...? I'm not sure I understand..?
> 

Something in https://github.com/learningequality/kolibri/blob/release-v0.15.x/kolibri/core/assets/src/api-resources/task.js to make all this new taks endpoints easier to be used from JS. I had to add in _kolibri/plugins/setup_wizard/assets/src/api.js_ some functions to handle it  (`SetupSoUDTasksResource` object)

> > Possibility to change the permissions of some of the TaskViewSet functions. I have had to create a new class inheriting it just to be able to change the permissions to see the list of tasks.
> 
> The `register_task` decorator has a `permission_classes` argument that'll allow flexible setting of permissions for any particular task. I would request you try not to inherit TaskViewSet, let us discuss the problem to see the pain that's forcing us to inherit.

If you take a look at #8185 you can find in _kolibri/plugins/setup_wizard/tasks.py_  the `startprovisionsoud` with its own permission_classes as you say. So far so good. 
The problem comes when the frontend needs to pull the tasks list to know if this task is in process. This code is being executed when provisioning kolibri after been installled, so there are not users yet and the permissions in `startprovisionsoud` work fine, but there's no way to get the list of the tasks from the frontend because `TasksViewSet` has a different permissions for `list`. 
I had to creeate  the class `SetupWizardSoUDTaskView(TasksViewSet) `at _kolibri/plugins/setup_wizard/api.py_ in #8185 to be able to use `SetupSoUDTasksResource.fetchCollection`  in _kolibri/plugins/setup_wizard/assets/src/views/importLODUsers/LoadingTaskPage.vue_ . If you know of a better solution to poll the tasks without inheriting `TaskViewSet` I am all ears :)

vkWeb: > Something in https://github.com/learningequality/kolibri/blob/release-v0.15.x/kolibri/core/assets/src/api-resources/task.js to make all this new taks endpoints easier to be used from JS. I had to add in kolibri/plugins/setup_wizard/assets/src/api.js some functions to handle it (SetupSoUDTasksResource object).

I have not touched the frontend code at all because my all focus has been on first refactoring the tasks backend.

> The problem comes when the frontend needs to pull the tasks list to know if this task is in process. This code is being executed when provisioning kolibri after been installled, so there are not users yet and the permissions in startprovisionsoud work fine, but there's no way to get the list of the tasks from the frontend because TasksViewSet has a different permissions for list.

Oh, I get it. This should be handled on the TasksViewSet level, we must not inherit it. 

Ummm, looks like @rtibbles initial suggestion will prove helpful now. His suggestion was to filter the tasks based on whether user has permission for it or not, if the user do have permission then that task gets included in the response.  

So, in our case of `startprovisionsoud`, if the user sending API request to `list` has permission `[IsSuperuser | NotProvisionedCanPost | LODUserHasSyncPermissions]` the `startprovisionsoud` will be included in the response list. 

This feels like a good general solution to this problem. What do you think sir @jredrejo?

jredrejo: > 
> Ummm, looks like @rtibbles initial suggestion will prove helpful now. His suggestion was to filter the tasks based on whether user has permission for it or not, if the user do have permission then that task gets included in the response.
> 
> So, in our case of `startprovisionsoud`, if the user sending API request to `list` has permission `[IsSuperuser | NotProvisionedCanPost | LODUserHasSyncPermissions]` the `startprovisionsoud` will be included in the response list.
> 
> This feels like a good general solution to this problem. What do you think sir @jredrejo?

it sounds perfect !

vkWeb: If you say I'll send a PR right away...?

Issue #8189, opened 2021-07-07
Add multiple learning activities tooltips to learning activity bar


MisRob:
## Observed behavior

We already have [a learning activity bar component](https://github.com/learningequality/kolibri/issues/8106) prepared and although it can display an icon for multiple learning activities, implementation is not full. It should also display corresponding tooltips [as designed in Figma](https://www.figma.com/file/uiZuCIqh2KYvBUfbCiJyyA/Hybrid-Learning?node-id=212%3A0):

![Screenshot from 2021-07-07 19-41-31](https://user-images.githubusercontent.com/13509191/124805108-5cfc7900-df5b-11eb-96f3-d07f4bbbdee4.png)

PR #8190, opened 2021-07-07
Add support for multiple learning activities icon

MisRob:
## Summary

I added support for `Array` type for `kind` property of `LearningActivityIcon` because backend API sends learning activities associated with a content node as an array:
```
learning_activities: ["watch"]
```
Because more learning activities can be returned, I also added logic that displays multiple activities icon

Screenshot from 2021-07-07 15-09-38

in case an array with more than one item is passed to `LearningActivityIcon` which reflects needs illustrated on hybrid learning designs.

I reflected these updates in `LearningActivityBar` component too. Although a correct icon will be displayed in the bar now for multiple learning activities, it's not a full implementation yet (see https://github.com/learningequality/kolibri/issues/8189).

----

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/plugins/learn/assets/src/views/LearningActivityBar.vue
+++ b/kolibri/plugins/learn/assets/src/views/LearningActivityBar.vue
@@ -4,9 +4,8 @@
     <KLabeledIcon :style="{ 'margin-top': '8px' }">
       <template #icon>
         <LearningActivityIcon
-          v-if="learningActivityKind"
           data-test="learningActivityIcon"
-          :kind="learningActivityKind"
+          :kind="learningActivities"
           :shaded="true"
         />
       </template>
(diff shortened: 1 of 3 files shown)


kolibri/plugins/learn/assets/src/views/LearningActivityBar.vue
```diff
@@ -120,13 +119,14 @@
         required: true,
       },
       /**
-       * Learning activity constant
+       * An array of one or more learning activities constants
        */
-      learningActivityKind: {
-        type: String,
+      learningActivities: {
+        type: Array,
```
MisRob, line 125: I didn't consider important to support both `Array` and `String` like for  `LearningActivityIcon` which is much more general component than `LearningActivityBar` so for the sake of simplicity, I've just changed `String` to `Array` here. Another reason is that I assume that we will simply pass data as received from API to the bar which is always an array, even when there is only one item.

marcellamaki (approved):
Looks good @MisRob!

PR #8196, opened 2021-07-09
Add env var to disable file based logging

### LINKED ISSUE #7979: Allow setting different logging modes via environment variable
## Observed behavior
We currently have a very fully specified logging configuration, which gets intellligently used both before the Django stack is started and after.

However, it always assumes that the log location should be to disk, which can cause problems for some deployment models.

## Expected behavior
Because logging is used by the options.ini parsing machinery in `options.py`, we can't use a configurable option in options.ini to configure logging. It therefore seems best to simply have an environment variable that can be specified to turn off file based logging here: https://github.com/learningequality/kolibri/blob/develop/kolibri/utils/logger.py#L195

So, all that would change is that the handler lists would be `['console']` in each case, rather than `['file', 'console', 'file_debug']`.

We should call the environment variable: `KOLIBRI_NO_FILE_BASED_LOGGING`.

## Context
This is most acutely needed for cloud deployments of Kolibri, where persistent file storage complicates deployment.

rtibbles:
## Summary
* Adds an environment variable which prevents file based logging
* Cleans up unused cherrypy log formatter
* Consolidates configuration environment variables into `env.py`
* Adds a CLI command to page through all available configuration environment variables with descriptions

## References
Fixes #7979

## Reviewer guidance
Run kolibri with and without `KOLIBRI_NO_FILE_BASED_LOGGING` env var set, see that with it set, nothing is written to logs.

Screenshot from 2021-07-13 13-00-13

----

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/utils/cli.py
+++ b/kolibri/utils/cli.py
@@ -449,3 +449,35 @@ def list():
             + " " * (max_len - len(plugin) + 4)
             + ("ENABLED" if plugin in config.ACTIVE_PLUGINS else "DISABLED")
         )
+
+
+@main.command(cls=KolibriGroupCommand, help="Configure Kolibri and enabled plugins")
+def configure():
+    pass
+
+
+def _get_env_vars():
+    """
+    Generator to iterate over all environment variables
+    """
+    template = "{envvar} - {description}\n\n"
+
+    from kolibri.utils.env import ENVIRONMENT_VARIABLES
+
+    for key, value in ENVIRONMENT_VARIABLES.items():
+        yield template.format(envvar=key, description=value.get("description", ""))
+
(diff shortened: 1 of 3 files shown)


jredrejo (commented):
I know this is not probably the first place in the code where this happens, but should not we centralize or, at least, document in a centralized document these env variables?
Something like https://github.com/learningequality/kolibri/blob/develop/kolibri/utils/options.py#L192 because there might be a point where we have conflicting variables or we don't even know how many we're using

jredrejo (approved):
This is great, thanks!

PR #8198, opened 2021-07-12
Implement automatic provisioning for new devices.

### LINKED ISSUE #6613: Automated setup of Kolibri

### LINKED ISSUE #7803: Move automatic device provisioning feature from apps into Kolibri
### Description

The Kolibri app has a feature where it can automatically provision a device if a file named `automatic_provision.json` exists in `KOLIBRI_HOME` and contains provisioning information. (This function exists in `src/kolibri_tools/utils.py` in the app repo.

It was decided that this functionality would be useful for non-app installations as well, and should be moved into Kolibri proper.

rtibbles:
## Summary
* Consolidates provisioning code
* Adds an option for the file path of an automatic provisioning file
* Creates a utility to read an automatic provisioning file and provision a device based on it
* On first upgrade, attempts to run automatic provisioning

## References
Fixes #7803
Fixes #6613

## Reviewer guidance
Does the file format make sense? Does it allow for future extensibility in ways that might cover, e.g. content configuration?

## TODO
- [ ] Add Gherkin story

----

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/core/device/management/commands/provisiondevice.py
+++ b/kolibri/core/device/management/commands/provisiondevice.py
@@ -9,7 +9,7 @@ from django.core.management.base import CommandError
 from django.utils import six
 
 from kolibri.core.auth.constants.facility_presets import presets
-from kolibri.core.auth.models import Facility
+from kolibri.core.device.utils import get_facility_by_name
 from kolibri.core.device.utils import setup_device_and_facility
 from kolibri.core.device.utils import validate_device_settings
 from kolibri.core.device.utils import validate_facility_settings
@@ -36,10 +36,9 @@ def json_file_contents(parser, arg):
         return parser.error("The file '{}' does not exist".format(arg))
     with open(arg, "r") as f:
         try:
-            output = json.load(f)
+            return json.load(f)
         except ValueError as e:  # Use ValueError rather than JSONDecodeError for Py2 compatibility
             return parser.error("The file '{}' is not valid JSON:\n{}".format(arg, e))
-    return output
 
 
 def get_all_user_input(facility_name, preset, language_id, username, password):
@@ -95,22 +94,10 @@ def validate_options(options):
             facility_name, preset, language_id, username, password
         )
 
-    facility = None
+    facility = get_facility_by_name(facility_name)
 
-    if facility_name:
-        facility_query = Facility.objects.filter(name__iexact=facility_name)
-
-        if facility_query.exists():
-            facility = facility_query.get()
-            logger.warn(
-                "Facility with name '{name}' already exists, not modifying preset.".format(
-                    name=facility.name
-                )
-            )
-    else:
-        facility = Facility.get_default_facility() or Facility.objects.first()
-        if not facility:
-            raise CommandError("No facility exists")
+    if not facility and not facility_name:
+        raise CommandError("No facility exists")
 
     try:
         device_settings = validate_device_settings(
--- a/kolibri/core/device/management/commands/provisiondevice.py
+++ b/kolibri/core/device/management/commands/provisiondevice.py
@@ -4,58 +4,17 @@ import os
 import sys
 
 from django.conf import settings
-from django.core.exceptions import ValidationError
 from django.core.management.base import BaseCommand
 from django.core.management.base import CommandError
-from django.db import transaction
 from django.utils import six
 
-from kolibri.core.auth.constants.facility_presets import mappings
 from kolibri.core.auth.constants.facility_presets import presets
 from kolibri.core.auth.models import Facility
-from kolibri.core.auth.models import FacilityUser
-from kolibri.core.device.utils import provision_device
+from kolibri.core.device.utils import setup_device_and_facility
+from kolibri.core.device.utils import validate_device_settings
+from kolibri.core.device.utils import validate_facility_settings
 
 logger = logging.getLogger(__name__)
-
-
-def _check_setting(name, available, msg):
-    if name not in available:
-        raise CommandError(msg.format(name))
-
-
-def check_facility_setting(name):
-    AVAILABLE_SETTINGS = [
-        "learner_can_edit_username",
-        "learner_can_edit_name",
-        "learner_can_edit_password",
-        "learner_can_sign_up",
-        "learner_can_delete_account",
-        "learner_can_login_with_no_password",
-        "show_download_button_in_learn",
-    ]
-    _check_setting(
-        name,
-        AVAILABLE_SETTINGS,
-        "'{}' is not a facility setting that can be changed by this command",
-    )
-
-
-def check_device_setting(name):
-    AVAILABLE_SETTINGS = [
-        "language_id",
-        "landing_page",
-        "allow_guest_access",
-        "allow_peer_unlisted_channel_import",
-        "allow_learner_unassigned_resource_access",
-        "name",
-        "allow_other_browsers_to_connect",
-    ]
-    _check_setting(
-        name,
-        AVAILABLE_SETTINGS,
-        "'{}' is not a device setting that can be changed by this command",
-    )
 
 
 def get_user_response(prompt, valid_answers=None, to_lower_case=True):
@@ -72,8 +31,19 @@ def get_user_response(prompt, valid_answers=None, to_lower_case=True):
 languages = dict(settings.LANGUAGES)
 
 
-def create_facility(facility_name=None, preset=None, interactive=False):
-    if facility_name is None and interactive:
+def json_file_contents(parser, arg):
+    if not os.path.exists(arg) or not os.path.isfile(arg):
+        return parser.error("The file '{}' does not exist".format(arg))
+    with open(arg, "r") as f:
+        try:
+            output = json.load(f)
+        except ValueError as e:  # Use ValueError rather than JSONDecodeError for Py2 compatibility
+            return parser.error("The file '{}' is not valid JSON:\n{}".format(arg, e))
+    return output
+
+
+def get_all_user_input(facility_name, preset, language_id, username, password):
+    if facility_name is None:
         answer = get_user_response(
             "Do you wish to create a facility? [y/n] ", ["y", "n"]
         )
@@ -83,6 +53,49 @@ def create_facility(facility_name=None, preset=None, interactive=False):
             )
         else:
             sys.exit(1)
+
+    if facility_name is not None and preset is None:
+        preset = get_user_response(
+            "Which preset do you wish to use? [{presets}]: ".format(
+                presets="/".join(presets.keys())
+            ),
+            valid_answers=presets,
+        )
+
+    if language_id is None:
+        language_id = get_user_response(
+            "Enter a default language code [{langs}]: ".format(
+                langs=",".join(languages.keys())
+            ),
+            valid_answers=languages,
+        )
+
+    if username is None:
+        username = get_user_response("Enter a username for the super user: ")
+
+    if password is None:
+        confirm = ""
+        while password != confirm:
+            password = get_user_response("Enter a password for the super user: ")
+            confirm = get_user_response("Confirm password for the super user: ")
+
+    return facility_name, preset, language_id, username, password
+
+
+def validate_options(options):
+    interactive = options["interactive"]
+    facility_name = options["facility"]
+    preset = options["preset"]
+    language_id = options["language_id"]
+    username = options["superusername"]
+    password = options["superuserpassword"]
+
+    if interactive:
+        facility_name, preset, language_id, username, password = get_all_user_input(
+            facility_name, preset, language_id, username, password
+        )
+
+    facility = None
 
     if facility_name:
         facility_query = Facility.objects.filter(name__iexact=facility_name)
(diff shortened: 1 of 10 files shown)


kolibri/core/device/management/commands/provisiondevice.py
```diff
@@ -73,8 +32,19 @@ def get_user_response(prompt, valid_answers=None, to_lower_case=True):
...
-    if facility_name is None and interactive:
+def json_file_contents(parser, arg):
+    if not os.path.exists(arg) or not os.path.isfile(arg):
+        return parser.error("The file '{}' does not exist".format(arg))
+    with open(arg, "r") as f:
+        try:
+            output = json.load(f)
+        except ValueError as e:  # Use ValueError rather than JSONDecodeError for Py2 compatibility
+            return parser.error("The file '{}' is not valid JSON:\n{}".format(arg, e))
+    return output
```
indirectlylit: super minor - I think this behaves the same without a temporary variable?

```suggestion
            return json.load(f)
        except ValueError as e:  # Use ValueError rather than JSONDecodeError for Py2 compatibility
            return parser.error("The file '{}' is not valid JSON:\n{}".format(arg, e))
```

kolibri/core/device/utils.py
```diff
@@ -141,3 +111,219 @@ def set_app_key_on_response(response):
...
+        },
+        "username": "superuser",
+        "password": "password"
+    }
+    All fields are optional.
+    """
+    from kolibri.core.auth.models import Facility
+
+    if device_provisioned() or not os.path.exists(file_path):
+        return
```
indirectlylit: for a function called `provision_from_file`, these conditions seem like they should be exceptions, not silently fail

kolibri/utils/options.py
```diff
@@ -361,6 +361,11 @@ def url_prefix(value):
             "default": "",
             "description": "Additional directories in which Kolibri will look for content files and content database files.",
         },
+        "AUTOMATIC_PROVISION_FILE": {
+            "type": "path",
+            "default": "automatic_provision.json",
```
indirectlylit: It's a little confusing that this is the default, but the file doesn't actually exist in the codebase. Two options I can think of:

1. make the default `null`
2. create an actual `automatic_provision.json` in the codebase, and use that to be the "source of truth" for default values. This would have the advantage that the it would serve as a self-documenting reference, which would be more likely to stay up-to-date than the "Expects a JSON file with the following format (example values supplied)" comment.

kolibri/core/device/utils.py
```diff
@@ -141,3 +111,219 @@ def set_app_key_on_response(response):
...
+                    )
+                )
+
+
+def provision_from_file(file_path):
+    """
+    Expects a JSON file with the following format (example values supplied):
+    {
+        "facility": "My Facility",
+        "preset": "formal",
```
indirectlylit, line 282: what's the meaning/intent of passing in a preset, if the preset is then overwritten by the specific settings below?

kolibri/core/device/utils.py
```diff
@@ -141,3 +111,219 @@ def set_app_key_on_response(response):
...
+            "language_id": "en",
+            "landing_page": "homepage",
+            "allow_guest_access": true,
+            "allow_peer_unlisted_channel_import": true,
+            "allow_learner_unassigned_resource_access": true,
+            "name": "My Device",
+            "allow_other_browsers_to_connect": true
+        },
+        "username": "superuser",
+        "password": "password"
```
indirectlylit: Slightly odd that `device_settings` and `facility_settings` have their own sub-objects, but these two don't.

E.g. 

```suggestion
        "super_user_credentials": {
            "username": "superuser",
            "password": "password"
        }
```

Do we have a principled way of deciding how deeply nested this file should be, and when nesting is appropriate? If not, we might want to consider imposing some rules.

kolibri/core/device/utils.py
```diff
@@ -141,3 +111,219 @@ def set_app_key_on_response(response):
...
+                        username=username, facility=facility
+                    )
+                )
+
+
+def provision_from_file(file_path):
+    """
+    Expects a JSON file with the following format (example values supplied):
+    {
+        "facility": "My Facility",
```
indirectlylit:
```suggestion
        "facility_name": "My Facility",
```

otherwise might be confused with facility ID

kolibri/core/device/utils.py
```diff
@@ -141,3 +111,219 @@ def set_app_key_on_response(response):
...
+        "preset": "formal",
+        "facility_settings": {
+            "learner_can_edit_username": true,
+            "learner_can_edit_name": true,
+            "learner_can_edit_password": true,
+            "learner_can_sign_up": true,
```
indirectlylit, line 287: In previous work on so-called ["device templates"](https://www.notion.so/learningequality/Data-and-content-architecture-proposal-8e03a6da67394c578d40336b7a7953bf#5a856f31bfcf49e29678a7bb5dec342c), I recall there may have been use-cases where we sometimes wanted these values to be fixed, and other times we wanted them to be treated as defaults with the expectation that users with sufficient permissions could modify them. I don't recall specific scenarios.

Here's the relevant snippet:

> Under the hood, management of certain device properties can be delegated to a facility user with 'sufficient' permission. The template can specify assign one of four device template permission delegation strategies to a particular device property:
> 
> 1. **No delegation:** the property is fixed at the default value.
> 2. **By user role:** the property can be modified by certain types of user (learner, coach, admin). For example, facility admins might have the ability to manage content, but not facilities or device settings.
> 3. **By user permissions:** for example, a particular user might have `can_manage_content=True`. We can add additional user properties like `can_manage_facilities` and `can_manage_device_settings`.
> 4. **By both user role and permissions.**
> 
> Note that delegation can be considered an implementation detail, and would not necessarily be exposed to users in full. It's more likely that we would define fewer, simpler device management options to handle the most common use-cases and expose those options to users.

This whole delegation system might not be exactly what we need, but I could imagine possibly having settings take _objects_ instead of raw values to allow for future extensibility.

For example we could do something like this to distinguish default from permanently fixed values:

```suggestion
            "learner_can_sign_up": { "value": true, "fixed": true }
```

kolibri/core/device/utils.py
```diff
@@ -141,3 +111,219 @@ def set_app_key_on_response(response):
...
+        AVAILABLE_SETTINGS,
+        "'{}' is not a device setting that can be changed by this command",
+    )
+
+
+def validate_facility_settings(new_settings):
+    # Override any settings passed in
+    for key in new_settings:
+        check_facility_setting(key)
+    return new_settings
```
indirectlylit, line 159: Mainly a conversational note below - not a change request.

One pattern I've started to use occasionally is a list of objects rather than a top-level object. E.g. rather than

```json
{
  "some_setting": "some_value",
  "another_setting": "another_value",
}
```

use 

```json
[
  { "name": "some_setting", "default": "some_value" },
  { "name": "another_setting", "default": "another_value" },
]
```

Benefits are that it's easier to process these files with `filter` and `map`, and it's natural to extend with new metadata.

indirectlylit (commented):
looks good! no blockers but a few questions and comments

kolibri/core/device/utils.py
rtibbles: Fair - it was just simpler to fail silently, but can catch the error where it's invoked instead.

kolibri/core/device/management/commands/provisiondevice.py
rtibbles: Yes, this was like that when I got here - some guy called @indirectlylit made the temporary variable... :)

kolibri/utils/options.py
rtibbles: The main reason for not having a file is that the default behaviour is *not* to automatically provision.

I think setting the default to empty string would be most explicit, and then forcing people leveraging this functionality to explicitly name the file they want to import from.

kolibri/core/device/utils.py
rtibbles: Both are optional.

kolibri/core/device/utils.py
rtibbles: I think this seems like a good change, yes.

kolibri/core/device/utils.py
rtibbles: Yeah, that was my instinct also, I had only not changed it because of a holdover from the provisiondevice management command.

kolibri/core/device/utils.py
rtibbles: I would be OK with keeping it simple for now, but allowing either of these in the future. Seems OK to have the shorthand for the defaults.

kolibri/core/device/utils.py
rtibbles: It feels unnecessarily verbose to me - and it's not so hard to do `pick` type behaviours with objects, and can always map on `.items()` if need both the keys and the values.

indirectlylit:
thanks for the updates!

kolibri/core/device/management/commands/provisiondevice.py
```diff
@@ -72,8 +31,18 @@ def get_user_response(prompt, valid_answers=None, to_lower_case=True):
 languages = dict(settings.LANGUAGES)
 
 
-def create_facility(facility_name=None, preset=None, interactive=False):
-    if facility_name is None and interactive:
+def json_file_contents(parser, arg):
+    if not os.path.exists(arg) or not os.path.isfile(arg):
+        return parser.error("The file '{}' does not exist".format(arg))
+    with open(arg, "r") as f:
```
jamalex, line 37: Probably not high priority to handle here, but I suppose there could in principle still be permission errors thrown at this point.

kolibri/core/device/test/test_device_provision.py
```diff
@@ -40,28 +46,23 @@ def test_create_facility_set_preset(self):
...
-        facility = create_facility(facility_name=default_facility.name)
-        self.assertEqual(default_facility, facility)
-
-    def test_create_super_user(self):
-        Facility.objects.create(name="Test")
-        provision_device()
-        create_superuser(username="test", password="test")
-        self.assertTrue(FacilityUser.objects.get(username="test").is_superuser)
+        self.assertIsNotNone(default_facility)
+        setup_device_and_facility(None, None, None, None, {}, None, None)
```
jamalex, line 50: image

kolibri/core/device/test/test_device_provision.py
```diff
@@ -104,3 +105,131 @@ def test_create_device_settings_default_facility(self):
...
+        default_facility = Facility.get_default_facility()
+        self.assertEqual(
+            DeviceSettings.objects.get().default_facility, default_facility
+        )
+
+
+class DeviceProvisionFileUnhappyTestCase(TestCase):
+    """
+    Tests for provisiondevice command.
+    """
```
jamalex, line 187: A mention of what makes this test case unhappy might be good here (doesn't need to be a full psychoanalysis, just some high-level platitudes).

kolibri/utils/main.py
```diff
@@ -150,6 +152,27 @@ def _setup_django():
...
+        check_plugin_config_file_location(version)
+        # Reset the enabled plugins to the defaults
+        # This needs to be run before dbbackup because
+        # dbbackup relies on settings.INSTALLED_APPS
+        enable_new_default_plugins()
+
+
+def _upgrades_after_django_setup(updated, version):
+    # On first ever run, attempt automatic provisioning
+    if not version and OPTIONS["Paths"]["AUTOMATIC_PROVISION_FILE"]:
```
jamalex: What if it's been run before but not yet provisioned? We may still want to support picking up a provisioning file added post-facto so we could handle that case (to allow for a workflow like running Kolibri once to generate the home directory, perhaps load content etc, then adding the conf, provisioning template, etc into there (and not needing to know to delete the .data_version or whatever file from there).

jamalex (approved):
Looks good! Added some optional suggestions.

jamalex (commented):
Looks good!

kolibri/core/device/test/test_device_provision.py
indirectlylit: 😆

PR #8202, opened 2021-07-13
UX for learner and coach/admin single user syncing

### LINKED ISSUE #8178: Create front end for automated syncing behaviors of individual learner syncing
See: https://app.zenhub.com/workspaces/kolibri-ecosystem-6047e525b8b9d900182e2d6f/issues/learningequality/kolibri/7934

### LINKED ISSUE #8179: Create front end for coach view of user sync statuses
See: https://app.zenhub.com/workspaces/kolibri-ecosystem-6047e525b8b9d900182e2d6f/issues/learningequality/kolibri/7999

### LINKED ISSUE #8187: Create API endpoint for fetching sync statuses of individual users and classes
This issue is the compliment to the front end issues #8178 and #8179, which create the front end displays for the learner's sync status (displayed in the AppBar dropdown menu) and the new coach/admin view of the sync status overview for a class. 

Data will need to be pulled from multiple places - both to determine if there is an active or failed sync, or to determine if the user is enqueued. Richard has suggested creating a new intermediary table that can be queried that contains the necessary data.

Returned data for each user should include 
- user id
- queue information
- transfer session information
- time of last sync

Outside of the scope of this issue is adding any enhancements/features that will facilitate transfer session reuse or other optimizations

marcellamaki:
## Summary
Fixes: #8187, closes #8178, resolves #8179 

Front end feature for coach/admin view of sync statuses and learner dropdown. ([Figma references](https://www.figma.com/file/MkIaNq2w3sx4Sh2uVOd35K/Single-user-syncing?node-id=11%3A643))
Creates corresponding endpoints and adds to resource layer

What is covered in this PR: 

- Create new API endpoint for User Sync Status
- The data returned from this endpoint should include the user, the last time the user synced (if they have synced before), if the user is actively syncing, or if the user is enqueued. 

- On the UX for learners, the sync status should be displayed for the learner in the dropdown menu (Figma: "Learner POV/changes to user dropdown")

Screen Shot 2021-07-16 at 7 07 31 PM

- On the UX for coaches, the coach should be able to open a page that contains a table of learners and see their statuses. If the coach has questions about the status options, they can view the informational modal. Coaches should also be able to access this table from the quiz report view 
(Figma: "Coach POV")

Screen Shot 2021-07-16 at 7 00 13 PM

Screen Shot 2021-07-16 at 7 00 21 PM

Screen Shot 2021-07-16 at 7 00 28 PM

Screen Shot 2021-07-16 at 7 21 28 PM

Outstanding TO DOs: 
- Add in new icons after Misha's PR is merged
- Finish new python tests and ensure all tests are passing

Follow up issues for: user name overflow, limited device messaging/"learn-only mode"

## Reviewer guidance
To test these changes end to end, use `kolibri shell` to create some UserSyncStatus objects in the database. Alternatively, you can use some fake data values in just the front end for reviewing only the UX/front-end changes (which for a front-end only review, would probably be simpler). To do this, you can add objects for: 

AppBar.vue (add user sync data for single learner to  `userSyncStatus`)
```
   data() {
      return {
        userMenuDropdownIsOpen: false,
        userSyncStatus: null,
        isPolling: false,
      };
    },
```

ClassLearnersListPage.vue (add class list data, an array of user sync objects) 
```
data: function() {
      return {
        displayTroubleshootModal: false,
        classSyncStatusList: [],
      };
    },
```

Each of these user objects should have the following keys `"id", "queued", "sync_session", "last_synced", "active",  "user", "user_id"`

----

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated


kolibri/core/device/api.py
```diff
@@ -164,3 +168,37 @@ def patch(self, request):
         settings.name = request.data["name"]
         settings.save()
         return Response({"name": settings.name})
+
+
+class UserSyncStatusViewSet(ReadOnlyValuesViewset):
```
rtibbles: These are confusingly named, my inference is that they are named the wrong way around, looking at the filtering going on in the get_queryset method.

kolibri/core/device/api.py
```diff
@@ -164,3 +168,37 @@ def patch(self, request):
...
+            classroom = Classroom.objects.get(pk=classroom_id)
+            classroom_users = FacilityUser.objects.filter(
+                memberships__collection=classroom
+            )
+            queryset = queryset.filter(user__in=classroom_users)
+        return queryset
+
+
+class ClassListSyncStatusViewSet(ReadOnlyValuesViewset):
+    values = (
```
rtibbles, line 202: As the only difference between these two Viewsets is filtering that is done, I'd recommend consolidating these and using a FilterSet to return different sets of data.

```
class SyncStatusFilter(FilterSet):

    member_of = ModelChoiceFilter(
        method="filter_member_of", queryset=Collection.objects.all()
    )

    def filter_member_of(self, queryset, name, value):
        return queryset.filter(Q(user__memberships__collection=value) | Q(user__facility=value))

    class Meta:
        model = UserSyncStatus
        fields = ["user", "member_of"]
```

Can then add to the viewset as:
```
filter_backends = (KolibriAuthPermissionsFilter, DjangoFilterBackend)
filter_class = SyncStatusFilter
```

We don't actually need `KolibriAuthPermissionsFilter` on there, but we should also add the permission based filtering to prevent unauthorized access.

rtibbles (commented):
Can use filters to prevent reimplementing the same endpoint multiple times!

kolibri/core/assets/src/views/AppBar.vue
```diff
@@ -82,6 +82,15 @@
                   :userType="getUserKind"
                 />
               </div>
+              <div v-if="getUserKind === 'learner'">
```
rtibbles, line 85: At some point (but not necessarily now) we should refactor the name of this getter to `userKind` - calling it `getUserKind` is more in line with our practice of getters that are actually functions (I had to double check that this was not expecting the id of a user as an argument).

kolibri/core/assets/src/views/AppBar.vue
```diff
@@ -208,11 +240,31 @@
         this.$emit('showLanguageModal');
         this.userMenuDropdownIsOpen = false;
       },
+      mapSyncStatusOptionToLearner() {
```
rtibbles: Any reason for this not to be a computed prop, rather than an invoked method?

kolibri/core/assets/src/views/AppBar.vue
```diff
@@ -208,11 +240,31 @@
...
+        if (this.userSyncStatus) {
```
rtibbles: Empty arrays are not falsey in Javascript, so I don't know when this would not be true - unless the API endpoint can return `null`?

kolibri/core/assets/src/views/AppBar.vue
```diff
@@ -153,13 +164,16 @@
     data() {
       return {
         userMenuDropdownIsOpen: false,
+        userSyncStatus: [],
```
rtibbles: Given that this is expected to be a plain object, using an empty array seems like an error here. Maybe `null` as you are doing a falsey check elsewhere?

kolibri/core/assets/src/views/CoreMenu/index.vue
```diff
@@ -127,7 +127,7 @@
 
   @import '~kolibri-design-system/lib/styles/definitions';
   .ui-menu-header {
-    padding: 1rem 1rem 1rem 50px;
+    padding: 1rem 1rem 1rem 1.2rem;
```
rtibbles, line 130: I feel like there might be a reason for this fixed pixel padding here, or does 1.2rem give the desired padding at all screen resolutions?

kolibri/plugins/learn/viewsets.py
```diff
@@ -186,3 +186,40 @@ def get_queryset(self):
                 lesson_assignments__collection__membership__user=self.request.user,
                 is_active=True,
             )
+
+
+class LearnerSyncStatusViewset(ReadOnlyValuesViewset):
```
rtibbles: I assume this will be deleted in favour of the general viewset above?

kolibri/plugins/coach/assets/src/views/plan/LessonSummaryPage/ManageLessonModals.vue
```diff
@@ -55,7 +55,6 @@
       },
     },
     computed: {
-      ...mapState(['classList']),
```
rtibbles: Is this state not required?

kolibri/plugins/coach/assets/src/views/ClassLearnersListPage.vue
```diff
@@ -0,0 +1,214 @@
...
+
+
+<script>
+
+  import CoreBase from 'kolibri.coreVue.components.CoreBase';
+  import CoreTable from 'kolibri.coreVue.components.CoreTable';
+  import commonCoreStrings from 'kolibri.coreVue.mixins.commonCoreStrings';
+  import { SyncStatus } from 'kolibri.coreVue.vuex.constants';
+  import { mapState, mapActions } from 'vuex';
+  import SyncStatusDisplay from '../../../../../core/assets/src/views/SyncStatusDisplay';
```
rtibbles, line 85: One slight issue with doing this import is that we will either duplicate the strings across the core and the coach plugin or the strings will not be included in the coach plugin (I have completely forgotten at this point which way the i18n machinery goes in this regard).

It might be _simpler_ (for a certain value of simple) to add these components to the core API and reference them in the same way as CoreBase and CoreTable.

kolibri/core/device/api.py
```diff
@@ -164,3 +172,48 @@ def patch(self, request):
...
+        "user",
+        "user_id",
+    )
+
+    def get_queryset(self):
+        return UserSyncStatus.objects.filter()
+
+    def annotate_queryset(self, queryset):
+
+        queryset = queryset.annotate(
```
rtibbles, line 216: These feel right, we can double check after merging to see if doing this or an explicit subquery produces a more performant query.

rtibbles (commented):
Some clean up and questions.

kolibri/core/assets/src/views/CoreMenu/index.vue
marcellamaki: I also suspect there might be a reason, but keeping it at `50px` now causes some very wonky alignment issues in the learner view. I've tests it in several places with different screen sizes and so far I don't see any issues with this change, but I definitely do not know the ins and outs of manual QA to know if there are _no_ problems. Also open to suggestions about how else I might set up the nested component here in a way that would keep me from making an update in `index.vue` - maybe @MisRob would have an idea...

marcellamaki:
I believe this is as ready for review as I am able to make it right now. I am having an issue with some of my tests not passing, and I would be happy to chat with either @MisRob or @rtibbles if you have any suggestions. I am 90% sure it has to do with async issues but....I can't quite figure out where I'm going wrong.

kolibri/plugins/coach/assets/src/views/reports/ReportsControls.vue
```diff
@@ -56,6 +61,14 @@
         return isEmbeddedWebView || this.disableExport;
       },
     },
+    methods: {
+      viewLearners() {
+        this.$router.push(this.$router.getRoute('ClassLearnersListPage'));
```
jredrejo: Same as above

kolibri/plugins/coach/assets/src/views/home/HomePage/OverviewBlock.vue
```diff
@@ -65,10 +76,16 @@
         return this.$router.getRoute('CoachClassListPage', {}, { facility_id });
       },
     },
+    methods: {
+      viewLearners() {
+        this.$router.push(this.$router.getRoute('ClassLearnersListPage'));
```
jredrejo: if you give a name to the route in kolibri/plugins/coach/assets/src/routes/index.js it would be simpler, `this.$router.push('CLASSLEARNERSLIST'),`

kolibri/plugins/coach/assets/src/views/home/HomePage/OverviewBlock.vue
```diff
@@ -35,6 +35,17 @@
           {{ $formatNumber(learnerNames.length) }}
         </template>
       </HeaderTableRow>
+      <HeaderTableRow v-if="learnerNames.length > 0">
```
jredrejo, line 38: `<HeaderTableRow v-if="learnerNames.length">` is enough if you're sure `learnerNames` is defined, is you need to check if learnerNames is not undefined:
`<HeaderTableRow v-if="!learnerNames || !learnerNames.length">`

kolibri/plugins/coach/assets/src/views/ClassLearnersListPage.vue
```diff
@@ -0,0 +1,218 @@
...
+        }
+        if (learnerSyncData) {
+          if (learnerSyncData.active) {
+            return SyncStatus.SYNCINGSYNCING;
+          } else if (learnerSyncData.queued) {
+            return SyncStatus.QUEUED;
+          } else if (learnerSyncData.last_synced) {
+            const currentDateTime = new Date();
+            const TimeDifference = learnerSyncData.last_synced - currentDateTime;
+            const diffMins = Math.round(((TimeDifference % 86400000) % 3600000) / 60000);
```
jredrejo: Same as above

kolibri/plugins/coach/assets/src/views/ClassLearnersListPage.vue
```diff
@@ -0,0 +1,218 @@
...
+        return SyncStatus.NOT_CONNECTED;
+      },
+      pollClassListSyncStatuses() {
+        this.fetchUserSyncStatus({ id: this.$route.params.classId }).then(status => {
+          this.classSyncStatusList = status;
+        });
+        if (this.isPolling) {
+          setTimeout(() => {
+            this.pollClassListSyncStatuses();
+          }, 10000);
```
jredrejo: maybe this polling interval should be a property instead, so a) it's changeable and b) it's more readable

kolibri/core/assets/src/views/AppBar.vue
```diff
@@ -170,15 +184,52 @@
...
+        if (this.userSyncStatus) {
+          if (this.userSyncStatus.active) {
+            return SyncStatus.SYNCINGSYNCING;
+          } else if (this.userSyncStatus.queued) {
+            return SyncStatus.QUEUED;
+          } else if (this.userSyncStatus.last_synced) {
+            const currentDateTime = new Date();
+            const TimeDifference = this.userSyncStatus.last_synced - currentDateTime;
+            const diffMins = Math.round(((TimeDifference % 86400000) % 3600000) / 60000);
+            if (diffMins < 60) {
```
jredrejo: would not be simpler something like:
`if (TimeDifference < 5184000000) // 5184000000 = 60 days in milliseconds`

kolibri/core/assets/src/views/SyncStatusDescription.vue
```diff
@@ -0,0 +1,50 @@
...
+  export default {
+    name: 'SyncStatusDescription',
+    props: {
+      syncStatus: {
+        type: String,
+        default: '',
+      },
+    },
+    computed: {
+      syncTextDisplayMap() {
```
jredrejo, line 23: Maybe I'm too nitpicky, but seeing a switch that can be so easily replaced by an object looks ugly to me:
```
const statusTranslations= {SyncStatus.RECENTLY_SYNCED:this.$tr('syncedDescription') , 
SyncStatus.QUEUED:this.$tr('queuedDescription'), ....}

return statusTranslations[this.syncStatus] || ''
```
looks better than a switch for me, but I can be wrong

kolibri/core/assets/src/views/SyncStatusDisplay.vue
```diff
@@ -0,0 +1,125 @@
...
+          // case SyncStatus.NOT_RECENTLY_SYNCED:
+          //   return this.$tr('notRecentlySynced');
+          case SyncStatus.NOT_CONNECTED:
+            return this.$tr('notConnected');
+          default:
+            return '';
+        }
+      },
+      syncIconDisplayMap() {
+        switch (this.syncStatus) {
```
jredrejo: same look and feel with the switch as above

kolibri/plugins/coach/assets/src/routes/index.js
```diff
@@ -64,6 +65,14 @@ export default [
       titleParts: ['activityLabel', 'CLASS_NAME'],
     },
   },
+
+  {
+    path: '/:classId/learners',
```
jredrejo, line 72: I'd add 
`name: 'CLASSLEARNERSLIST' `
so future references to this route are easier

kolibri/core/assets/src/views/AppBar.vue
```diff
@@ -170,15 +184,52 @@
       dropdownName() {
         return !hashedValuePattern.test(this.username) ? this.username : this.fullName;
       },
+      mapSyncStatusOptionToLearner() {
+        if (this.userSyncStatus) {
+          if (this.userSyncStatus.active) {
+            return SyncStatus.SYNCINGSYNCING;
```
MisRob: _Syncing syncing
still syncing
stiiill syncing and syncing... :)_

 `SyncStatus.SYNCING`

kolibri/core/assets/src/views/AppBar.vue
```diff
@@ -170,15 +184,52 @@
...
       window.addEventListener('click', this.handleWindowClick);
       this.$kolibriBranding = branding;
     },
+    mounted() {
+      this.isPolling = true;
+      this.pollUserSyncStatusTask(this.userId);
+    },
     beforeDestroy() {
       window.removeEventListener('click', this.handleWindowClick);
+      this.isPolling = false;
```
MisRob, line 218: _(Can be follow-up)_ I'd suggest using `clearTimeout` to stop polling similarly to https://github.com/learningequality/kolibri/blob/develop/kolibri/plugins/learn/assets/src/views/classes/ClassAssignmentsPage.vue#L49.

Depending on timing, the current implementation can cause one unnecessary call that seems to be unintentional. You can see it here, for example (please wait a while for the second call in the console):

peek


Generally speaking, such behavior has potential to cause unexpected errors - for example when a request would be made after a user logs out to an endpoint that requires authentication. In this case, it's doesn't break anything in this way. Maybe due to our multi-apps architecture (I assume that completely another bundle is loaded after being redirected to login)? So it's not a blocker but would be good to address at some point.

kolibri/core/assets/src/views/AppBar.vue
```diff
@@ -170,15 +184,52 @@
...
+      pollUserSyncStatusTask() {
+        this.fetchUserSyncStatus({ id: this.userId }).then(syncData => {
+          if (syncData && syncData[0]) {
+            this.userSyncStatus = syncData[0];
+          }
+        });
+        if (this.isPolling) {
+          setTimeout(() => {
+            this.pollUserSyncStatusTask();
+          }, 10000);
```
MisRob: _(Would be nice but no action needed)_ Might be good to save polling interval value to constants so we can locate it easily when we need to tweak it

kolibri/core/assets/src/views/SyncStatusDisplay.vue
```diff
@@ -0,0 +1,125 @@
...
+  import { SyncStatus } from 'kolibri.coreVue.vuex.constants';
+
+  export default {
+    name: 'SyncStatusDisplay',
+    props: {
+      syncStatus: {
+        type: String,
+        default: '',
+      },
+      displaySize: {
```
MisRob, line 33: When choosing between more predefined values, we usually add a validator to document it and also to check on those values during development. Here's one example https://github.com/learningequality/kolibri/blob/develop/kolibri/plugins/epub_viewer/assets/src/views/PreviousButton.vue#L25.

MisRob: _(Would be nice but no action needed)_ Is there any reason for not using simpler names for `displaySize` values? For example, `small` instead of `sync-status-small`.

kolibri/plugins/coach/assets/src/views/home/HomePage/OverviewBlock.vue
```diff
@@ -35,6 +35,17 @@
...
+        <template #key>
+        </template>
+        <template #value>
+          <KButton
```
MisRob: This element is navigating users to another page so I think that using `KRouterLink` would be more appropriate (https://design-system.learningequality.org/buttons/)

kolibri/plugins/coach/assets/src/views/reports/ReportsControls.vue
```diff
@@ -3,7 +3,12 @@
   <div v-show="!$isPrint" class="report-controls">
     <slot></slot>
     <div class="report-controls-buttons">
-
+      <KButton
```
MisRob: Similarly to the above, I'd expect `KRouterLink` here

kolibri/plugins/coach/assets/src/views/ClassLearnersListPage.vue
```diff
@@ -0,0 +1,218 @@
...
+        }
+        return options;
+      },
+    },
+    mounted() {
+      this.isPolling = true;
+      this.pollClassListSyncStatuses({ classroom_id: this.$route.params.classId });
+    },
+    beforeDestroy() {
+      this.isPolling = false;
```
MisRob, line 138: _(Can be follow-up)_ I'd recommend using `clearTimeout` here too for reasons explained earlier

kolibri/plugins/coach/assets/src/views/ClassLearnersListPage.vue
```diff
@@ -0,0 +1,218 @@
...
+        displayTroubleshootModal: false,
+        classSyncStatusList: [],
+      };
+    },
+    computed: {
+      ...mapState('classSummary', ['learnerMap']),
+      syncStatusOptions() {
+        let options = [];
+        for (const [value] of Object.entries(SyncStatus)) {
+          // skip displaying the "not recently synced" as a separate option, per Figma design
```
MisRob: Even though _"Not recently synced"_ is merged with _"Unable to sync"_ in the troubleshooting modal, according [to the designs](https://www.figma.com/file/MkIaNq2w3sx4Sh2uVOd35K/Single-user-syncing?node-id=57%3A1573) it seems that both should be still communicated in the label:

Screenshot from 2021-07-21 15-06-13

The current implementation shows only _"Unable to sync"_ so a bit more logic will be needed to clarify this point to users.

kolibri/plugins/coach/assets/src/views/ClassLearnersListPage.vue
```diff
@@ -0,0 +1,218 @@
...
+          value !== SyncStatus.NOT_RECENTLY_SYNCED ? options.push(value) : null;
+        }
+        return options;
+      },
+    },
+    mounted() {
+      this.isPolling = true;
+      this.pollClassListSyncStatuses({ classroom_id: this.$route.params.classId });
```
MisRob: `pollClassListSyncStatuses` method doesn't accept any parameters

kolibri/plugins/learn/api_urls.py
```diff
@@ -10,5 +10,8 @@
     r"learnerclassroom", LearnerClassroomViewset, base_name="learnerclassroom"
 )
 router.register(r"learnerlesson", LearnerLessonViewset, base_name="learnerlesson")
+router.register(
+    r"learnersyncstatus", LearnerLessonViewset, base_name="learnersyncstatus"
```
MisRob: I haven't noticed this URL being used

kolibri/plugins/coach/assets/src/views/ClassLearnersListPage.vue
```diff
@@ -0,0 +1,218 @@
...
+        let learnerSyncData;
+        if (this.classSyncStatusList) {
+          learnerSyncData = this.classSyncStatusList.filter(entry => {
+            return entry.user_id == learnerId;
+          });
+          learnerSyncData = learnerSyncData[learnerSyncData.length - 1];
+        }
+        if (learnerSyncData) {
+          if (learnerSyncData.active) {
+            return SyncStatus.SYNCINGSYNCING;
```
MisRob: `SyncStatus.SYNCING`

kolibri/plugins/coach/assets/src/views/ClassLearnersListPage.vue
```diff
@@ -0,0 +1,218 @@
...
+          } else if (learnerSyncData.queued) {
+            return SyncStatus.QUEUED;
+          } else if (learnerSyncData.last_synced) {
+            const currentDateTime = new Date();
+            const TimeDifference = learnerSyncData.last_synced - currentDateTime;
+            const diffMins = Math.round(((TimeDifference % 86400000) % 3600000) / 60000);
+            if (diffMins < 60) {
```
MisRob: _(Would be nice but no action needed)_  Even though there are not that many duplicates, here it might make sense to have calculations from line 163 to line 167 extracted to a helper function for the sake of UX consistency (so that we don't tweak `60` in this condition but forget to do the same thing in the `AppBar`).

kolibri/core/assets/src/state/modules/core/actions.js
```diff
@@ -897,3 +901,16 @@ export function notLoading(store) {
     });
   });
 }
+
+export function fetchUserSyncStatus(store, id) {
```
MisRob: _(Would be nice but no action needed)_ When reviewing, I was a bit confused that `id` can be both a user ID and a class ID. Might be helpful to document it here.

MisRob (commented):
Awesome work @marcellamaki :) Regarding frontend, I appreciate components architecture. I've pointed out several issues in code comments. Feel absolutely free to skip or open follow-ups for those that are marked correspondingly. I only skimmed through backend briefly. Thanks for the backend test suite - it helped me to understand the whole feature and also some frontend parts better. It was my first encounter with syncing so it was a lot of studying - thanks for the detailed PR description and all references. I haven't gotten around to testing with `kolibri shell`.

Regarding problems with tests, I'll have a look.

Please also see the following:

#### 1. Sync statutes are being polled even when I visit _"Learn"_ while not logged in. However, there is no information about the sync status in the user menu for a guest user. I assume that the only purpose of polling from `AppBar` is to fetch this information for UI or are there any more reasons? If this assumption is correct, I'd suggest removing it for guests.

Peek 2021-07-21 16-33

#### 2. When I navigate to the learners page from _"Reports"_ tab in _"Coach"_ and then click the back button, I expect to be returned to the "Reports" tab but I am taken to "Class Home" tab instead.

Peek 2021-07-21 17-06

kolibri/core/assets/src/views/CoreMenu/index.vue
MisRob: I'll have a look

MisRob: I assume that there indeed was a specific reason to have `50px` and that it was to make header text aligned with options. See the screenshot from `develop`:

Screenshot from 2021-07-22 10-27-05

According to the new designs for user syncing

Screenshot from 2021-07-22 10-35-43

this is to be deprecated, at least in some cases, so this change looks fine to me.

However, I am not sure if we're supposed to keep the previous layout when no sync information is visible in the menu, like for the admin. If yes, this change will cause the following regression:

Screenshot from 2021-07-22 10-33-43

I feel fine about using the new layout everywhere for the sake of consistency. Maybe let's see what designers think.

MisRob: In any case, it's not a blocker. I checked all places where we use `CoreMenu` and haven't noticed any other issues.

kolibri/core/assets/src/views/SyncStatusDisplay.vue
marcellamaki: No, that is a good suggestion! I have still not adjusted to file-scoped CSS and am used to having very long, descriptive class names needed to keep track of things in giant css files 😂 Here simpler is definitely better!

MisRob:
@marcellamaki @rtibbles I've briefly skimmed through the latest frontend updates and haven't noticed any issues

kolibri/core/device/test/test_api.py
```diff
@@ -462,3 +467,52 @@ def test_device_name_max_length(self):
...
+
+    def test_usersyncstatus_list(self):
+        response = self.client.get(reverse("kolibri:core:usersyncstatus-list"))
+        expected_count = UserSyncStatus.objects.count()
+        self.assertEqual(len(response.data), expected_count)
+
+    def test_user_sync_status_class_single_user_for_filter(self):
+        response = self.client.get(
+            reverse("kolibri:core:usersyncstatus-list"),
+            data={"user_id": self.user1.id},
```
rtibbles: Here you are filtering by `user_id` but the `FilterSet` for this endpoint specifies `user` as the parameter.

You should create another UserSyncStatus during your test setup for this test to be meaningful.

kolibri/core/assets/src/state/modules/core/actions.js
rtibbles: This comment is not true.

The backend expects either `user` to filter by a user_id or `member_of` to filter by a collection_id - this flexible use of `id` is doing no filtering at all.

rtibbles (changes requested):
Two critical changes needed.

kolibri/core/assets/src/state/modules/core/actions.js
```diff
@@ -897,3 +901,38 @@ export function notLoading(store) {
...
+
+export function fetchUserSyncStatus(store, params) {
+  // for fetching all users that are members of a particular classroom id
+  if (params.member_of) {
+    return UserSyncStatusResource.fetchCollection({
+      force: true,
+      getParams: { member_of: params.member_of },
+    }).then(
+      syncData => {
+        store.commit('SET_CORE_CHANNEL_LIST', _userSyncStatusState(syncData));
```
rtibbles: One last question here - why is this being stored using the `SET_CORE_CHANNEL_LIST` mutation? This seems like the wrong place to store this data.

I think we'd be better off not setting any store state in here, and just returning the data from the function.

Should also return the same data as we are storing if the `_userSyncStatusState` mapping function is important.

rtibbles (approved):

PR #8205, opened 2021-07-14
Add permissions to SyncStatus model.

rtibbles:
## Summary
* Adds very simple read only + same facility permissions to the sync status model

----

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/core/device/models.py
+++ b/kolibri/core/device/models.py
@@ -4,6 +4,7 @@ from uuid import uuid4
 
 from django.conf import settings
 from django.db import models
+from django.db.models import Q
 from morango.models import UUIDField
 from morango.models.core import SyncSession
 
@@ -11,6 +12,8 @@ from .utils import LANDING_PAGE_LEARN
 from .utils import LANDING_PAGE_SIGN_IN
 from kolibri.core.auth.models import Facility
 from kolibri.core.auth.models import FacilityUser
+from kolibri.core.auth.permissions.base import q_none
+from kolibri.core.auth.permissions.general import DenyAll
 from kolibri.core.utils.cache import process_cache as cache
 from kolibri.plugins.app.utils import interface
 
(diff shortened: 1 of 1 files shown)


kolibri/core/device/models.py
```diff
@@ -210,9 +213,29 @@ def clean_stale(cls, expire=180.0):
         cls.objects.filter(updated__lte=staled_time).delete()
 
 
+class SyncStatusPermissions(DenyAll):
+    """
+    By default set all permissions to read only, as these should
+    only be directly updated by internal Kolibri processes, not by users.
+    """
+
```
rtibbles:
```suggestion
    def user_can_read_object(self, user, obj):
        return True
```

kolibri/core/device/models.py
```diff
@@ -210,9 +213,34 @@ def clean_stale(cls, expire=180.0):
...
+    def user_can_read_object(self, user, obj):
+        if hasattr(user, "dataset"):
+            return obj.user.dataset_id == user.dataset_id
+        return False
+
+    def readable_by_user_filter(self, user):
+        """
+        For simplicity, let anyone from the same facility read the sync status
+        for now. In the future, we may want to restrict this, but it is not
+        sensitive information.
```
jamalex: The model contains a foreignkey onto SyncSession, which means it directly stores the sync session ID, which is used as the "session token" (like a browser session_id cookie) for purposes of syncing -- if the syncsession is still active, anybody with that ID can sync within the scope of that sync session, without further certificate verification.

I'm not seeing anywhere in the codebase yet that pulls this model in to display in the frontend, but I'd recommend that it not send the sync session ID as part of that, instead just looking up the relevant fields (e.g. last_activity timestamp) across the foreignkey. If that were enforced, then it would perhaps be OK to leave these permissions as-is. But the safest would be to only allow access for someone who's a coach or admin for the target user.

jamalex (changes requested):

jamalex (approved):
Cool, thanks!

PR #8213, opened 2021-07-16
Fix deprovision cache issue

rtibbles:
## Summary
* Add tests for device settings behaviours.
* Fix device settings cache persistence on deletion.

## References
Fixes issue after deprovisoining where device settings would persist in cache

## Reviewer guidance
Setup device
Deprovision
Try to set up device again

----

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/core/device/models.py
+++ b/kolibri/core/device/models.py
@@ -97,12 +97,14 @@ class DeviceSettings(models.Model):
     def save(self, *args, **kwargs):
         self.pk = 1
         self.full_clean()
-        super(DeviceSettings, self).save(*args, **kwargs)
+        out = super(DeviceSettings, self).save(*args, **kwargs)
         cache.set(DEVICE_SETTINGS_CACHE_KEY, self, 600)
+        return out
 
     def delete(self, *args, **kwargs):
-        super(DeviceSettings, self).delete(*args, **kwargs)
+        out = super(DeviceSettings, self).delete(*args, **kwargs)
         cache.delete(DEVICE_SETTINGS_CACHE_KEY)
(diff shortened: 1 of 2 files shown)


kolibri/core/device/models.py
```diff
@@ -93,6 +100,10 @@ def save(self, *args, **kwargs):
         super(DeviceSettings, self).save(*args, **kwargs)
         cache.set(DEVICE_SETTINGS_CACHE_KEY, self, 600)
 
+    def delete(self, *args, **kwargs):
+        super(DeviceSettings, self).delete(*args, **kwargs)
```
bjester: Parsing the obfuscated Manager code, it seems that perhaps this method isn't necessary? It looks like it should be utilizing the `delete` method on the QuerySet. Otherwise, it seems like this should return the result of `super.delete`.

rtibbles: Gotcha - can try removing this method and seeing if the tests still pass.

bjester: You might add one more test. I see `DeviceSettings.objects.all().delete()` which I think invokes the QuerySet method, but `DeviceSettings.objects.delete()` would go through the manager

bjester (approved):
Works for me!

kolibri/core/device/models.py
rtibbles: Added the manager test, it passed. Removed the model method, and it failed.

rtibbles:
Expected checks are only in develop atm. Merging!

PR #8219, opened 2021-07-22
Assignment handling within single-user syncing

### LINKED ISSUE #8079: Add side loading for single user sync of data that does not fall within single user scope
This issue shall serve as the tech spec for https://github.com/learningequality/kolibri/issues/8038

The partitions for Lessons, Quizzes, and their respective Assignments do not fall within the scope of an individual user syncing certificate, and hence need to be synced across to a SoUD through a parallel (non-Morango) mechanism. However, this can still happen within the context of a Morango sync session, and use the open sync session as a means of authentication.

The proposal is to add a new public API endpoint in Kolibri, `/api/public/assignment_sync/<user_id>`, that can handle `HEAD` requests (to check the server's current ETag value before sending a bunch of data), `GET` requests (to retrieve the current set of assigned resources), and `POST` requests (to send the current set of assigned resources).

When the SoUD is the client, initiating a sync with a full-facility server:

- After the "pull" phase of the sync, the client calculates its own ETag, then sends an ETag-conditional `GET` request to `/api/public/assignment_sync/<user_id>`.
- The server calculates its own ETag, and if it matches the client-provided value, returns a `304 Not Modified`.
- Otherwise, it returns the serialized data for the assigned resources. If this happens, the client removes all Lesson/Quiz assignments it has stored for the user, and creates the records that it received.

When the full-facility device is the client, initiating a sync with the SoUD:

- After the "push" phase of the sync, the client (full-facility device) calculates its own ETag, then sends a `HEAD` request to the SoUD's `/api/public/assignment_sync/<user_id>` to check its ETag.
- If they match, it continues merrily.
- If they don't match, it makes a `GET` request to the endpoint, then removes and replaces its local resources with the response.

All of these requests will include a header with the Sync Session ID. The server should validate that:
- The corresponding Sync Session exists and is still active. Otherwise, return a `401 Unauthorized`.
- The user_id is for a user that is within scope of the Sync Sessions certificates. Otherwise, return a `403 Forbidden`.
- When it is validating records received, ensure the partition is correct, and the ID matches the computed ID. Otherwise, return a `400 Bad Request`.

The proposed structure of the serialized payload is:
```
{
    "lessons": [{<lesson_1_serialization>}, {<lesson_2_serialization>}, ...],
    "quizzes": [{<quiz_1_serialization>}, {<quiz_2_serialization>}, ...]
}
```
There should be no need to sync the *Assignment objects themselves, as they are implicit in the lists of resources (these are all-inclusive lists of the resources currently assigned, so assignments can be deleted or created based on presence of a resource in here).

Note: when quiz/lesson resource and assignment model instances are created on a SoUD, they should be marked as "not dirty" to prevent unnecessary serialization (that could also cause issues if the device later synced the full facility). Similarly, the signal that tracks model deletion should be bypassed when they are removed.

jamalex:
## Summary

Fixes https://github.com/learningequality/kolibri/issues/8079.

## Reviewer guidance

There's a lot of code duplication between the logic for exams and for lessons, but they're different enough (e.g. in the naming of some attributes) that it seemed cleanest to just copy-paste and adapt.

I'll add some inline notes.

To run the test locally, I use:
```
INTEGRATION_TEST=True pytest kolibri/core/auth/test/test_morango_integration.py::EcosystemSingleUserAssignmentTestCase::test_single_user_assignment_sync --pdb --capture=no
```

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated


kolibri/core/auth/apps.py
```diff
@@ -13,3 +13,10 @@ class KolibriAuthConfig(AppConfig):
     def ready(self):
         from .signals import cascade_delete_membership  # noqa: F401
         from .signals import cascade_delete_user  # noqa: F401
+
+        from kolibri.core.auth.sync_event_hook_utils import (
+            register_sync_event_handlers,
+        )  # noqa: F401
+        from morango.api.viewsets import session_controller  # noqa: F401
+
+        register_sync_event_handlers(session_controller)
```
jamalex, line 22: I wasn't sure whether this was the best place to put this, but... I couldn't think of a better place. Suggestions welcome, or maybe this is ok.

kolibri/core/auth/management/commands/sync.py
```diff
@@ -475,11 +479,14 @@ def handler(transfer_session):
...
-                / float(transfer_session.records_total)
-            )
+            try:
+                progress = (
+                    100
+                    * transfer_session.records_transferred
+                    / float(transfer_session.records_total)
+                )
+            except ZeroDivisionError:
+                progress = 100
```
jamalex: Not sure whether this was fixed elsewhere, but I was encountering divzero errors in my tests, when no data was being synced.

kolibri/core/auth/sync_event_hook_utils.py
```diff
@@ -0,0 +1,69 @@
...
+        "context": context,
+    }
+
+
+def _pre_transfer_handler(context):
+    assert context is not None
+
+    kwargs = _extract_kwargs_from_context(context)
+
+    if isinstance(context, LocalSessionContext):
```
jamalex, line 52: This is to avoid having the hooks be called twice, once for local and once for remote context.

kolibri/core/auth/test/sync_utils.py
```diff
@@ -57,7 +57,7 @@ def manage(self, *args):
 
     def create_model(self, model, **kwargs):
         kwarg_text = ",".join(
-            '{key}=\\"{value}\\"'.format(key=key, value=value)
+            "{key}={value}".format(key=key, value=repr(value))
             for key, value in kwargs.items()
```
jamalex, line 61: Was getting some format issues in the piping -- this helped.

kolibri/core/exams/single_user_assignment_utils.py
```diff
@@ -0,0 +1,106 @@
+from .models import ExamAssignment
```
jamalex, line 1: This file is virtually identical to kolibri/core/lessons/single_user_assignment_utils.py -- in terms of the overall structure -- some field names and details are different.

I'd be happy to try to DRY it up, it just might make it a bit more "magical" (e.g. a function containing this logic that gets passed models and field names etc, and then the comments in here etc would need to be generalized as well -- might hurt readability.

kolibri/core/auth/hooks.py
```diff
@@ -0,0 +1,29 @@
+from kolibri.plugins.hooks import define_hook
+from kolibri.plugins.hooks import KolibriHook
+
+
+@define_hook
+class FacilityDataSyncHook(KolibriHook):
```
rtibbles, line 6: This could be separated out into a pre-hook and post-hook, which would then allow each method to be setup with an `@abc.abstractmethod` decorator, which then gives a clear requirement that the method be instantiated.

kolibri/core/exams/single_user_assignment_utils.py
```diff
@@ -0,0 +1,106 @@
...
+
+    # get a list of all syncable exams that still have an active assignment
+    to_update = syncableexams.filter(exam_id__in=assignments.values_list("exam_id"))
+
+    # get a list of all syncable exams that don't have an active assignment anymore
+    to_delete = syncableexams.exclude(exam_id__in=assignments.values_list("exam_id"))
+
+    # create new syncable exam objects for all new assignments
+    for assignment in to_create:
+        IndividualSyncableExam.objects.create(
```
rtibbles, line 26: Any reason not to do a bulk create here?

rtibbles (commented):
Code all makes sense to me - some potential for cleanup, and possibly some optimization?

rtibbles (approved):
Tests check out. Other issues can be followed up after merge.

kolibri/core/exams/single_user_assignment_utils.py
indirectlylit: agree with the copy/paste strategy for code maintainability

indirectlylit (commented):
exciting! nice work

PR #8224, opened 2021-07-27
I18n tools

### LINKED ISSUE #7709: kolibri-tools needs improved message extraction for Kolibri plugins

### LINKED ISSUE #8218: String context transfer from Crowdin fails for Vue files
## Observed behavior
@nucleogenesis did some work to fix this in #8048, but it might need some retouches.

Running `make i18n-download` apparently executes correctly, reports context as transferred, and errors out only at the last step (linter). However, when I check status, it seems as if only `PO`, `MO` and `JSON` and 2 `JS` files are modified, and not a single `Vue` file, which is where most of our strings and context should be.

## Errors and logs
Terminal output: 
[make i18n-download.txt](https://github.com/learningequality/kolibri/files/6850916/make.i18n-download.txt)

## Expected behavior
All the Vue files have the added context transferred from Crowdin.

## User-facing consequences
...

## Steps to reproduce
…

## Context
…

cc @rtibbles

rtibbles:
## Summary
* Cleans up and consolidates i18n code
* Make CSV to JSON conversion create separate JSON for each bundle
* Remove hard coding of project and path information
* Makes CLI more user friendly
* Makes JS CLI commands within kolibri-tools for all i18n code
* Consolidate language_info.json into a single file that gets packaged for export at publish
* Consolidates all AST parsing to use the babel parser, and uses consistent functions for detection of translation functionality
* Fixes context transfer code to pick up context from any downloaded CSV file
* Transfers context from Crowdin into source code
* Updates JSON message files based on currently used strings
* Adds an `i18n` entry to the `setup.cfg` to have shared configuration across Python and JS i18n tooling

## References
Fixes #7709
Fixes #8218

## Reviewer guidance
Confirm that different translation steps still function as expected.
Do the changes in the JSON file change commit make sense?
Is the context transfer complete (last commit)?
Does the breakdown of JS CLI commands make sense?
Do any docs needs to be updated? I think the workflow is still the same.

----

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated


packages/kolibri-tools/package.json
```diff
@@ -49,11 +50,13 @@
     "file-loader": "^2.0.0",
     "glob": "^7.1.2",
     "htmlhint": "^0.9.13",
+    "ini": "^2.0.0",
     "intl-messageformat-parser": "^1.3.0",
     "jest": "^24.5.0",
     "jest-serializer-vue": "^2.0.2",
     "json-loader": "^0.5.4",
     "launch-editor-middleware": "^2.2.1",
+    "lodash": "^4.17.21",
```
indirectlylit, line 59: Doing this would avoid adding yet another copy of lodash:

```suggestion
    "lodash": "^4.17.20",
```

ref: https://github.com/learningequality/kolibri/pull/8224/files#diff-51e4f558fae534656963876761c95b83b6ef5da5103c4adef6768219ed76c2de

radinamatic:
A part from the font generation error already mentioned in Slack, I am unable to see the latest string context changes pulled from the Crowdin when running the `make i18n-download branch=release-v0.14.x` locally on this branch.

For example, I've changed the context of the [`CoachClassListPage.noClassesDetailsForAdmin`](https://crowdin.com/translate/kolibri/3810/en-es#276404) string, project has been built successfully, CSVs downloaded but the context change is not being updated in the local CSV.

Cowdin | local CSV
--- | ---
Selection_022 | LEDev2104 (start)  Running  - Oracle VM VirtualBox_019

Terminal also reports one Vue file skipped:

```
Kolibri: Skipping /home/didivm/LE/kolibri/kolibri/plugins/coach/assets/src/views/common/LearnerExerciseReport.vue because $trs property was given an empty object
```

setup.cfg
```diff
@@ -40,3 +40,7 @@ exclude_lines =
 	raise NotImplementedError
 
 	if __name__ == .__main__.:
+
+[i18n]
+project = kolibri
+locale_data_folder = kolibri/locale
```
indirectlylit: could do something along the lines of...


```suggestion
[kolibri:i18n]
project = kolibri
locale_data_folder = kolibri/locale
```

where eventually we'd also have e.g.

```ini
[kolibri:i18n]
project = studio
locale_data_folder = contentcuration/locale
```

rtibbles:
Does the context make it into the source code? If Crowdin is not putting it into the CSV files, not sure what I can do about that - but it doesn't seem to be predictable which CSV files the context ends up in on export.


> Kolibri: Skipping /home/didivm/LE/kolibri/kolibri/plugins/coach/assets/src/views/common/LearnerExerciseReport.vue because $trs property was given an empty object

This is intentional, there are no strings, so it is skipping. This is more a warning of a pointless $trs object in the component than anything else.

rtibbles:
I reran the context generation, it also did not reflect the update in Crowdin.

The issue seems to be that the old context is present in some of the downloaded CSV files, while the new context is only present in one. I don't think our code has any way to decide which is the newest one.

radinamatic:
> Does the context make it into the source code?

If I understand correctly this is the Vue file where the context should have been updated, and no, it does not seem to make it in there:

LEDev2104 (start)  Running  - Oracle VM VirtualBox_020

I don't know the details how the command is updating the context when the CSV files are downloaded. Could it be we are not fetching correctly from Crowdin, or it needs more time (?) to update itself... 😕 

The most recent context is saved in the **source** CSV files, that can be downloaded manually from the **Settings > Files** page, and I can also download the CSV file in question (`kolibri.plugins.coach.app-messages.csv`) from the Crowdin **Editor** page in all the languages, both display the new context correctly:

[kolibri.plugins.coach.app-messages.csv](https://github.com/learningequality/kolibri/files/6889405/kolibri.plugins.coach.app-messages.csv)

However, when downloaded manually from the **Settings > Translations** page, the same CVS file inside any language folder of the ZIP, does not have the updated context.

radinamatic:
A part from the context issue, devserver builds & runs OK, displays locales correctly and I did not notice any other apparent problems. We need to have the new 0.15.x branch on Crowdin to translate the new strings for S2S, so the context is not a blocker at this moment. 

I'm going to approve this and if @indirectlylit does not have other suggestions, I vote we merge and come back to figure out the context update issue in subsequent PRs.

radinamatic (approved):
So great seeing all this context for user strings finally makes it into the code! 💯 
Thank you @rtibbles, and thank you @dufcrule!

indirectlylit:
>  if @indirectlylit does not have other suggestions

Just beginning my review now. If there's urgency to merge I can always file follow-up issues

packages/kolibri-tools/build_kolibri_tools.js
```diff
@@ -30,3 +30,17 @@ kolibriPkg.dependencies = kolibriCorePkg.dependencies;
 fs.writeFileSync(kolibriPackageFile, JSON.stringify(kolibriPkg, undefined, 2), {
   encoding: 'utf-8',
 });
+
+/*
+ * Step 4: Copy the kolibri language_info.json into the kolibri-tools package for use externally
+ */
+
+const languageInfo = require(path.resolve(__dirname, '../../kolibri/locale/language_info.json'));
```
indirectlylit, line 38: I thought that `kolibri-tools` was an independent package? This seems to still be coupled to it's current location in the Kolibri repo.

packages/kolibri-tools/lib/cli.js
```diff
@@ -369,29 +379,412 @@ program
...
+
+// I18N Message Handling
+program
+  .command('i18n')
+  .arguments('<mode>', 'Mode to run in, options are: extract, transfer, profile')
+  .option(
+    '--dry-run',
+    `Will only run this file up to the point where we start extracting strings
+                        I found it useful for debugging early on - may be useful for future iterations
+                        such as for adding new projects or other root-level changes in the code base`,
```
indirectlylit: Text in these options is hard to read when using `--help` in the terminal:

image

packages/kolibri-tools/lib/cli.js
```diff
@@ -369,29 +379,412 @@ program
...
+    filePath
+  )
+  .action(function(options) {
+    const intlCodeGen = require('./i18n/intl_code_gen');
+    intlCodeGen(options.outputDir, options.langInfo);
+  });
+
+// I18N Message Handling
+program
+  .command('i18n')
```
indirectlylit: With regard to `kolibri-tools i18n`, there are other more-specifically-named commands which makes this one a bit nebulous-looking:

```
Commands:
  build [options] <mode>      Build frontend assets for Kolibri frontend plugins
  lint [options] [files...]   Run linting on files or files matching glob patterns
  test [options]
  compress [files...]
  i18n-code-gen [options]
  i18n [options] <mode>
  i18n-csv-to-json [options]
  i18n-stats [options]
  i18n-profile [options]
```

Recommend choosing a more specific name and adding some more descriptions

image

packages/kolibri-tools/lib/cli.js
```diff
@@ -369,29 +379,412 @@ program
...
+    false
+  )
+  .option(
+    '--dump-extracted',
+    `Will dump extractedMessages to a json file in the root where you ran this script.
+                              The file will be timestamped for uniqueness.`,
+    false
+  )
+  .option(
+    '-pf , --pluginFile <pluginFile>',
```
indirectlylit: `-pf` is confusing because posix conventions would interpret it as `-p -f`

Names for `--` arguments are also usually lower-kebab-case rather than camel: `--plugin-file`. Similar note on other options in this file

packages/kolibri-tools/lib/cli.js
```diff
@@ -369,29 +379,412 @@ program
...
+  )
+  .option(
+    '--searchPath <searchPath>',
+    'Set path to search for files containing strings to be extracted'
+  )
+  .action(function(mode, options) {
+    const modes = {
+      EXTRACT: 'extract',
+      TRANSFER: 'transfer',
+    };
```
indirectlylit: maybe cleaner as separate CLI commands with shared underlying code rather than a single command

packages/kolibri-tools/lib/i18n/SyncContext.js
```diff
@@ -1,369 +1,224 @@
...
     });
-  });
+  }
+
+  if (dump) {
+    const timestamp = Math.floor(Date.now() / 1000);
+    writeSourceToFile(
+      `updatedContextFiles-${timestamp}.txt`,
+      updatedFiles.reduce((acc, fileObj) => {
+        return acc + fileObj.filePath + '\n' + fileObj.newFile + '\n\n';
```
indirectlylit: This predated your work, but mind expanding instances of `acc` to a more descriptive name?

packages/kolibri-tools/lib/cli.js
```diff
@@ -15,6 +17,12 @@ function list(val) {
   return val.split(',');
 }
 
+function filePath(val) {
+  return path.resolve(process.cwd(), val);
+}
```
indirectlylit, line 22: We haven't established this as a convention at LE, but I personally prefer to prefix internal module helper functions with an underscore

packages/kolibri-tools/lib/cli.js
```diff
@@ -369,29 +379,412 @@ program
...
+      );
+    }
+  });
 
-  for (const packageName of Object.keys(results.versions)) {
-    if (!results.versions[packageName].isSatisfied) {
-      let required = engines[packageName];
+// I18N CSV to JSON
+program
+  .command('i18n-csv-to-json')
```
indirectlylit: general comment on CLI command names:

rather than naming them based on file formats (which are somewhat arbitrary) I think it would be long-term more helpful to name them based on intended goal.

For example, instead of `i18n-csv-to-json` we might call it `i18n-crowdin-to-git` or something along those lines. (I realize that might not be quite correct here, but I'm still processing the code)

packages/kolibri-tools/lib/i18n/csvToJSON.js
```diff
@@ -0,0 +1,63 @@
...
+        if (definition) {
+          messages[msg] = definition['Translation'];
+        } else {
+          logging.error(`Could not find translation for message ${msg} in CSV files`);
+        }
+      }
+      if (!dryRun) {
+        const localeFolder = path.join(localeDataFolder, toLocale(intlCode), 'LC_MESSAGES');
+        mkdirp.sync(localeFolder);
+        const filename = path.join(localeFolder, name + '-messages.json');
```
indirectlylit: where ever these files end up, would be great to add a `readme.md` file that has a reference back to here for future i18n spelunkers

packages/kolibri-tools/lib/cli.js
```diff
@@ -369,29 +379,412 @@ program
...
+      );
+      program.command('i18n-csv-to-json').help();
+    }
+    const csvToJSON = require('./i18n/csvToJSON');
+    csvToJSON(options.dryRun, pathInfo, options.ignore, options.langInfo, options.localeDataFolder);
+  });
+
+// I18N Stats
+program
+  .command('i18n-stats')
```
indirectlylit: Flagging that this is now the second meaning of "stats" in this file – the original being a mode of the `build` command. This makes searching through the file more confusing.

packages/kolibri-tools/lib/i18n/utils.py
```diff
@@ -6,66 +6,25 @@
...
-
-SOURCE_PATH = os.path.join(LOCALE_PATH, "en", "LC_MESSAGES")
-
-if not os.path.exists(SOURCE_PATH):
-    os.makedirs(SOURCE_PATH)
+# Path to the kolibri locale language_info file, which we use if we are running
+# from inside the Kolibri repository.
+_KOLIBRI_LANGUAGE_INFO_PATH = os.path.join(
+    os.path.dirname(__file__), "../../../../kolibri/locale/language_info.json"
+)
```
indirectlylit, line 19: somewhat brittle coupling to details of existing repo structure. Perhaps should be pulled from `setup.cfg` or some other global per-project config?

indirectlylit (commented):
Great job – significant improvements here.

I haven't tested yet, just read through the commits. My comments below mostly pertain to longer-term concerns and nothing blocks merging to move forward with short-term updates.

I did not yet have a chance to test context transfers (cc @radinamatic) – if helpful I can take a look at this on Friday

packages/kolibri-tools/build_kolibri_tools.js
rtibbles: This is the build tools - it is built inside the Kolibri repo, and makes a copy of this into the package, so that it is indeed independent of the Kolibri repo at publish time. Similarly to how we bundle up the core API spec to be accessible outside of the kolibri repo.

packages/kolibri-tools/lib/cli.js
rtibbles: I am also happy to get rid of these options - I did not use the dry-run or dump options at all while testing.

packages/kolibri-tools/lib/cli.js
rtibbles: I could also just separate this out into two separate commands - I was trying to follow the `build` command at first in putting all the modes under one roof, but I think it's simpler to just split this into two commands.

packages/kolibri-tools/lib/cli.js
rtibbles: Quite a few popular Node CLI utilities use camelCase, which is why this was implemented like this previously.

See, e.g. jest: https://jestjs.io/docs/cli#options

(Webpack uses kebab-case, so it's not hard and fast)

packages/kolibri-tools/lib/cli.js
rtibbles: I concur.

packages/kolibri-tools/lib/i18n/SyncContext.js
rtibbles: **Acc**knowledged.

packages/kolibri-tools/lib/cli.js
rtibbles: I like that practice in Python, but less so in Javascript, where the definition of an internal function is based on whether it is exported.

packages/kolibri-tools/lib/i18n/utils.py
rtibbles: I did think about adding it there - mostly this is a workaround for us only keeping one copy of `language_info.json` in the repository (previously it was duplicated in the locale folder and in this folder), this allows us to only replicate this default file at build time of the package.

The reason I decided not to add it as a global configuration was because it should only need to be specified when our language_info.json is insufficient.

setup.cfg
rtibbles: Hrm, not a big fan of the `:`, but it seems like Python ini parsers that allow subsections (ConfigObj) do it in a completely different way to the ini file parser for nodejs that I'm using, so this might be simplest.

kolibri/core/assets/src/views/AppError/ReportErrorModal.vue
```diff
@@ -92,12 +92,14 @@
       forumUseTips: {
         message:
           'Search the community forum to see if others encountered similar issues. If unable to find anything, paste the error details below into a new forum post so we can rectify the error in a future version of Kolibri.',
+
```
indirectlylit: looks like the previous change might have introduced an extra newline in the output?

indirectlylit (approved):
awesome, thanks for incorporating my feedback

kolibri/core/assets/src/views/AppError/ReportErrorModal.vue
rtibbles: I honestly have no idea why this happened - I hadn't changed anything in the code. One possibility is that the script is not idempotent. I'll take another look.

rtibbles: Hrm, it does seem to be idempotent.

rtibbles: OK, narrowed down the issue - seems to have been a problem that happened only when 'updating' context that hadn't changed. So I've just stopped it doing that, which seems more appropriate anyway!

PR #8223, opened 2021-07-26
Notification generation during cleanup stage of sync

bjester:
## Summary

- Adds a new Morango operation to the cleanup stage for generating notifications post-sync
- Fixes key error for certain notifications associated with `ContentNode`s that may not exist
- Fixes null-constraint issue for new device info property
- Dependency change:
```diff
diff --git a/requirements/base.txt b/requirements/base.txt
index 00695874f..f3a887763 100644
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -16,7 +16,7 @@ le-utils==0.1.24
 kolibri_exercise_perseus_plugin==1.3.5
 jsonfield==2.0.2
 requests-toolbelt==0.8.0
-morango==0.5.5
+morango==0.6.1
 tzlocal==2.1
 pytz==2020.5
 python-dateutil==2.7.5

```

## References

Resolves: https://github.com/learningequality/kolibri/issues/8188

## Reviewer guidance

1) Connect to the VPN
2) Provision a fresh instance of Kolibri
2) Initiate a sync from the CLI for "Hack Session School": 
```bash
$ kolibri manage sync --baseurl "http://10.8.0.1:8080/" --username "learningequality" --password "learningequality"
```
4) Check notifications DB after sync finishes:
```bash
$ sqlite3 develop/notifications.sqlite3 
-- Loading resources from /home/bjester/.sqliterc
SQLite version 3.31.1 2020-01-27 19:55:54
Enter ".help" for usage hints.
sqlite> select count(*) from notifications_learnerprogressnotification ;
count(*)  
----------
120       
```

----

## Testing checklist

- [X] Contributor has fully tested the PR manually
- [ ] If there are any front-end changes, before/after screenshots are included
- [ ] Critical user journeys are covered by Gherkin stories
- [ ] Critical and brittle code paths are covered by unit tests

## PR process

- [X] PR has the correct target branch and milestone
- [X] PR has 'needs review' or 'work-in-progress' label
- [X] If PR is ready for review, a reviewer has been added. (Don't use 'Assignees')
- [ ] If this is an important user-facing change, PR or related issue has a 'changelog' label
- [X] If this includes an internal dependency change, a link to the diff is provided

## Reviewer checklist

- Automated test coverage is satisfactory
- PR is fully functional
- PR has been tested for [accessibility regressions](http://kolibri-dev.readthedocs.io/en/develop/manual_testing.html#accessibility-a11y-testing)
- External dependency files were updated if necessary (`yarn` and `pip`)
[...]

PR #8232, opened 2021-07-28
Fix check_device_info

jredrejo:
## Summary
This PR fixes a couple of regressions introduced yesterday:

1. Changes introduced in `check_device_info` have created a bug because `update` function for a `dict` returns always None, not the dict itself. The function needs to return the dict
2. Changes introduced in `_attempt_connections` have created a bug returning `None` for the `subset_of_users_device` instead of `False` when the key does not exist in the discovered kolibri server.

## References

References: #8223
References: #8222 

## Reviewer guidance

Trying to provision a new kolibri installation must be able to detect and enable other servers when "Import from other facility" is used.

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/core/discovery/utils/network/client.py
+++ b/kolibri/core/discovery/utils/network/client.py
@@ -55,7 +55,7 @@ class NetworkClient(object):
                     info = response.json()
                     self.info = {}
                     for key in device_info_keys.get(DEVICE_INFO_VERSION, []):
-                        self.info[key] = info.get(key)
+                        self.info[key] = info.get(key) or False
                     if self.info["application"] not in ["studio", "kolibri"]:
                         raise requests.RequestException(
                             "Server is not running Kolibri or Studio"
--- a/kolibri/core/discovery/utils/network/client.py
+++ b/kolibri/core/discovery/utils/network/client.py
@@ -9,6 +9,10 @@ from .urls import get_normalized_url_variations
 
 logger = logging.getLogger(__name__)
 
+device_info_defaults = {
+    "subset_of_users_device": False,
+}
+
 
 class NetworkClient(object):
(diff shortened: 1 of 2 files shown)


kolibri/core/discovery/utils/network/connections.py
```diff
@@ -33,7 +33,9 @@ def check_device_info(base_url):
     try:
         info = NetworkClient(base_url=base_url).info
         if info["application"] in ["studio", "kolibri"]:
-            return device_info_defaults.copy().update(info)
+            complete_info = device_info_defaults.copy()
```
rtibbles, line 32: Seems like it might be simpler just to move the `device_info_defaults` into `kolibri/core/discovery/utils/network/client.py` and apply the change there only. Can then just return `info` here.

kolibri/core/discovery/utils/network/client.py
```diff
@@ -55,7 +55,7 @@ def _attempt_connections(self, urls):
                     info = response.json()
                     self.info = {}
                     for key in device_info_keys.get(DEVICE_INFO_VERSION, []):
-                        self.info[key] = info.get(key)
+                        self.info[key] = info.get(key) or False
```
rtibbles: If we move the `devince_info_defaults` into this file, we can then do:
```
self.info[key] = info.get(key, device_info_defaults.get(key))
```

rtibbles (approved):
This will work but think we can simplify!

bjester:
Thanks @jredrejo! I'm not sure why I didn't experience issues with this update

kolibri/core/discovery/utils/network/client.py
jredrejo: Done

PR #8247, opened 2021-07-30
Sync notifications updates

rtibbles:
## Summary
* Create or update UserSyncStatus for non-SoUD devices when a user's request to sync is queued
* Create or update UserSyncStatus for non-SoUD devices when a user's request to sync is initiated
* Create or update UserSyncStatus in the pre-transfer of a sync, to log the SyncSession and flag that it is no longer queued
* Update UserSyncStatus endpoint to compute status on the backend rather than the frontend
* Remove unused fields from endpoint
* For SoUD devices if there are no non-SoUD devices to sync with return no data to flag as 'not connected'
* Add missing string for not recently synced status
* Use `ElapsedTime` component to reuse existing component, and to be closer to spec

## References
Follow up to #8202 

## Reviewer guidance
Does the backend now properly report sync status?

Does the frontend properly render all states?

The previous 'recently synced' time was 60 days - this has been reduced to 1 day in this implementation, but maybe still needs updating.

----

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/core/device/api.py
+++ b/kolibri/core/device/api.py
@@ -1,3 +1,4 @@
+from datetime import timedelta
 from sys import version_info
 
 from django.conf import settings
@@ -6,6 +7,7 @@ from django.db.models import Max
 from django.db.models import OuterRef
 from django.db.models.query import Q
 from django.http.response import HttpResponseBadRequest
+from django.utils import timezone
 from django_filters.rest_framework import DjangoFilterBackend
 from django_filters.rest_framework import FilterSet
 from django_filters.rest_framework import ModelChoiceFilter
@@ -31,6 +33,8 @@ from kolibri.core.auth.api import KolibriAuthPermissions
 from kolibri.core.auth.api import KolibriAuthPermissionsFilter
 from kolibri.core.auth.models import Collection
 from kolibri.core.content.permissions import CanManageContent
+from kolibri.core.device.utils import get_device_setting
+from kolibri.core.discovery.models import DynamicNetworkLocation
 from kolibri.utils.conf import OPTIONS
 from kolibri.utils.server import get_urls
 from kolibri.utils.server import installation_type
@@ -193,6 +197,28 @@ class SyncStatusFilter(FilterSet):
         fields = ["user", "member_of"]
 
 
+RECENTLY_SYNCED = "RECENTLY_SYNCED"
+SYNCING = "SYNCING"
+QUEUED = "QUEUED"
+NOT_RECENTLY_SYNCED = "NOT_RECENTLY_SYNCED"
+
+
+def map_status(status):
+    """
+    Summarize the current state of the sync into a constant for use by
+    the frontend.
+    """
+    if status["active"]:
+        return SYNCING
+    elif status["queued"]:
+        return QUEUED
+    elif status["last_synced"]:
+        if timezone.now() - status["last_synced"] < timedelta(hours=1):
+            return RECENTLY_SYNCED
+        else:
+            return NOT_RECENTLY_SYNCED
+
+
 class UserSyncStatusViewSet(ReadOnlyValuesViewset):
     permission_classes = (KolibriAuthPermissions,)
     filter_backends = (KolibriAuthPermissionsFilter, DjangoFilterBackend)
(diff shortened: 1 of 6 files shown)


marcellamaki (commented):
Code read-through looks good (so much cleaner than mine! 🙏)! Haven't done a manual QA review yet

kolibri/core/device/api.py
```diff
@@ -193,34 +197,67 @@ class Meta:
...
+    """
+    Summarize the current state of the sync into a constant for use by
+    the frontend.
+    """
+    if status["active"]:
+        return SYNCING
+    elif status["queued"]:
+        return QUEUED
+    elif status["last_synced"]:
+        if timezone.now() - status["last_synced"] < timedelta(hours=1):
```
jamalex: I'm thinking it might be good to have the time delta be a configurable value, since "recent" can vary by context.

kolibri/core/device/api.py
```diff
@@ -193,34 +197,67 @@ class Meta:
...
     )
 
+    field_map = {
+        "status": map_status,
+    }
+
     def get_queryset(self):
-        return UserSyncStatus.objects.filter()
+        # If this is a subset of users device, we should just return no data
+        # if there are no possible devices we could sync to.
```
jamalex, line 243: Does this then make it show no status at all? Would we not want it to show a "not connected" message in this case?

kolibri/core/device/kolibri_plugin.py
```diff
@@ -0,0 +1,22 @@
+from kolibri.core.auth.hooks import FacilityDataSyncHook
+from kolibri.plugins.hooks import register_hook
+
+
+@register_hook
+class SingleUserExamSyncHook(FacilityDataSyncHook):
```
jamalex: Shouldn't affect anything, but I think this name is copypasta.

kolibri/core/device/kolibri_plugin.py
```diff
@@ -0,0 +1,22 @@
...
+    def pre_transfer(
+        self,
+        dataset_id,
+        local_is_single_user,
+        remote_is_single_user,
+        single_user_id,
+        context,
+    ):
+        # if we're about to send data to a single-user device, prep the syncable exam assignments
+        if context.sync_session and single_user_id is not None:
```
jamalex, line 16: I don't think this condition matches the comment; the condition I think just says something like "if either I or the other side is a single-user device, and one of us is about to send data to the other".

To match the comment, I think it'd be:
`if context.is_producer and remote_is_single_user:`

kolibri/core/device/api.py
rtibbles: No - absence of status is interpreted by the frontend as 'not connected', so it does indeed do this!

kolibri/core/device/kolibri_plugin.py
rtibbles: :yum: will fix.

kolibri/core/device/kolibri_plugin.py
rtibbles: The comment is also left over pasta.

jamalex: 🍝

kolibri/core/device/api.py
marcellamaki: @jamalex - that's a great point and a good catch, because this is actually an arbitrary interval that I made up while I was coding. I had asked Khang if he had an interval in mind of what "not recently synced" meant, and as I recall, he didn't think there had been a decision about that, and then I completely forgot to follow up with folks... 🙃 So, perhaps there could be some discussion/options about that now.

rtibbles: I had wondered about the somewhat arbitrary '60 days' in milliseconds encoding in the frontend wrt to this. I assumed that 60 days was probably not what we wanted, but the exact time does seem to be context dependent.

kolibri/plugins/coach/assets/src/views/ClassLearnersListPage.vue
```diff
@@ -61,9 +61,9 @@
                 />
               </td>
               <td>
-                <span dir="auto">
-                  {{ $tr('lastSyncedTime', { time: mapLastSynctedTimeToLearner(learner.id) }) }}
-                </span>
+                <ElapsedTime
+                  :date="mapLastSyncedTimeToLearner(entry)"
```
marcellamaki: With fresh eyes, seems like `entry` was maybe pulled from some of my code in `mapLastSyncedTimeToLearner` that you later refactored and now is undefined... Maybe it should be `learner.id`?

rtibbles: You mean there might be a bug in the code I refactored but did not test? This seems possible...

rtibbles:
```suggestion
                  :date="mapLastSyncedTimeToLearner(learner.id)"
```

jamalex (approved):

PR #8260, opened 2021-08-03
Add conditional polling interval for learner sync status

marcellamaki:
Adds a check to see if a learner is in the queue. If so, updates the polling interval for checking the sync status to every 1 second.

### DIFF
--- a/kolibri/core/assets/src/views/AppBar.vue
+++ b/kolibri/core/assets/src/views/AppBar.vue
@@ -197,10 +197,6 @@
       window.addEventListener('click', this.handleWindowClick);
       this.$kolibriBranding = branding;
     },
-    mounted() {
-      this.isUserLoggedIn ? (this.isPolling = true) : null;
-      this.pollUserSyncStatusTask(this.userId);
-    },
     beforeDestroy() {
       window.removeEventListener('click', this.handleWindowClick);
       this.isPolling = false;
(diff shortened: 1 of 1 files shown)


kolibri/core/assets/src/views/AppBar.vue
```diff
@@ -219,6 +220,15 @@
           }, this.pollingInterval);
         }
       },
+      setPollingInterval(status) {
+        if (status === SyncStatus.QUEUED) {
+          // check more frequently for updates if the user is waiting to sync,
+          // so that the sync isn't missed
+          this.pollingInterval = 1000;
```
rtibbles, line 223: Occurs to me looking at this now, that technically `this.pollingInterval` doesn't need to be reactive, so we could just set it in `created` rather than in data. Doesn't need to be changed though!

kolibri/core/assets/src/views/AppBar.vue
```diff
@@ -219,6 +220,15 @@
...
+        } else {
+          this.pollingInterval = 10000;
```
rtibbles, line 225: Another thing we should maybe do, is prevent polling when the drop down menu is hidden. So perhaps conditionalizing polling on `userMenuDropdownIsOpen`.

rtibbles (approved):
Can merge as is, but with the more aggressive polling, we should maybe back off when it's not being displayed.

PR #8262, opened 2021-08-04
Fix soud intervals

### LINKED ISSUE #8236: Need to requeue for sync after a successful automatic single user sync
## Observed behavior
After a successful automatic single user sync, no syncs happen again until the server restarts or a new server joins the network.

## Expected behavior
Need to requeue for sync after the successful completion of the sync

jredrejo:
## Summary
1. Ensure SOUD sync happens at regular intervals
2. Ensure sync is not scheduled by zeroconf is a sync has been previously scheduled
3. Clean failed syncinc tasks and updates UserSyncStatus to avoid staled states
4. Whenever the server accepts a sync, it will send the syncing interval so the client will use it to schedule the new syncing.

## References
Closes: #8236 
Fixes "Possible that sync jobs can get duplicated for the same server when the server is restarted" from https://www.notion.so/learningequality/Single-User-Syncing-Hack-Session-60-mins-bfb609b7d7de47b1af1e73b6446e3cbd

## Reviewer guidance

Restart several times the server or the synced server and check everything is working fine

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/core/auth/management/commands/sync.py
+++ b/kolibri/core/auth/management/commands/sync.py
@@ -25,6 +25,7 @@ from kolibri.core.auth.management.utils import run_once
 from kolibri.core.auth.models import dataset_cache
 from kolibri.core.auth.sync_event_hook_utils import register_sync_event_handlers
 from kolibri.core.logger.utils.data import bytes_for_humans
+from kolibri.core.public.utils import schedule_new_sync
 from kolibri.core.tasks.exceptions import UserCancelledError
 from kolibri.core.tasks.management.commands.base import AsyncCommand
 from kolibri.core.utils.lock import db_lock
@@ -269,6 +270,7 @@ class Command(AsyncCommand):
             self.job.save_meta()
 
         dataset_cache.deactivate()
+        schedule_new_sync(baseurl, user_id)
         logger.info("Syncing has been completed.")
 
     @contextmanager
--- a/kolibri/core/auth/management/commands/sync.py
+++ b/kolibri/core/auth/management/commands/sync.py
@@ -82,6 +82,12 @@ class Command(AsyncCommand):
             action="store_true",
             help="do not create a facility and temporary superuser",
         )
+        parser.add_argument(
+            "--resync-interval",
+            type=int,
+            default=5,
+            help="Seconds to schedule a new sync",
+        )
         # parser.add_argument("--scope-id", type=str, default=FULL_FACILITY)
 
     def handle_async(self, *args, **options):  # noqa C901
@@ -97,6 +103,7 @@ class Command(AsyncCommand):
             no_pull,
             noninteractive,
             no_provision,
+            resync_interval,
         ) = (
             options["baseurl"],
             options["facility"],
@@ -108,6 +115,7 @@ class Command(AsyncCommand):
             options["no_pull"],
             options["noninteractive"],
             options["no_provision"],
+            options["resync_interval"],
         )
 
         PORTAL_SYNC = baseurl == DATA_PORTAL_SYNCING_BASE_URL
@@ -270,7 +278,8 @@ class Command(AsyncCommand):
             self.job.save_meta()
 
         dataset_cache.deactivate()
-        schedule_new_sync(baseurl, user_id)
+        if user_id:
+            schedule_new_sync(baseurl, user_id, resync_interval)
         logger.info("Syncing has been completed.")
 
     @contextmanager
--- a/kolibri/core/auth/management/commands/sync.py
+++ b/kolibri/core/auth/management/commands/sync.py
@@ -85,7 +85,7 @@ class Command(AsyncCommand):
         parser.add_argument(
             "--resync-interval",
             type=int,
-            default=5,
+            default=None,
             help="Seconds to schedule a new sync",
         )
         # parser.add_argument("--scope-id", type=str, default=FULL_FACILITY)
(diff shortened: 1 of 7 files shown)


kolibri/core/auth/management/commands/sync.py
```diff
@@ -269,6 +270,7 @@ def handle_async(self, *args, **options):  # noqa C901
             self.job.save_meta()
 
         dataset_cache.deactivate()
+        schedule_new_sync(baseurl, user_id)
```
rtibbles: We should definitely conditionalize this on `user_id` - but also wondering if we should be adding an extra argument to the command to flag that it's a scheduled sync too.

(ultimately, I think this would be better handled by refactoring the sync command into different functions so it can be called outside of the scope of the management command - but that's too much for now).

kolibri/utils/options.py
```diff
@@ -475,6 +475,15 @@ def url_prefix(value):
             """,
         }
     },
+    "SYNCING": {
```
rtibbles: This could also fit under `Deployment` - if we keep it here, the convention has been for the top level names to be Title case rather than all caps.

kolibri/core/public/utils.py
```diff
@@ -171,35 +210,51 @@ def request_soud_sync(server, user=None, queue_id=None, ttl=10):
...
-        UserSyncStatus.objects.update_or_create(user_id=user, defaults={"queued": True})
-
-        if server_response["action"] == SYNC:
-            job_id = startpeerusersync(server, user)
-            logger.info(
-                "Enqueuing a sync task for user {} in job {}".format(user, job_id)
+    UserSyncStatus.objects.update_or_create(user_id=user, defaults={"queued": True})
+
+    if server_response["action"] == SYNC:
+        SYNC_INTERVAL = OPTIONS["SYNCING"]["SYNC_INTERVAL"]
```
rtibbles: Hrm, it might be simpler to just always read this from the server response for now. Because the options file only gets read at startup, writing to the options file doesn't make any guarantees about this option being updated, so this feels a little bit brittle potentially? In practice it may not be an issue for apps because everything will be running in a single process, so the in memory update should work.

Guessing that the reason to want to persist it into options is for the rescheduling of the sync? This might be a place where having a sync interval argument for the sync command might be helpful, as if it's not None then that's an indicator we should be syncing again, and also means we don't have to rely on this to persist the desired sync interval.

(One edge case that comes to mind is if two different servers with different sync intervals are both syncing with a SoUD, then these values could collide with each other).

rtibbles (commented):
I think adding a `sync_interval` argument to the sync command to both flag that we need to schedule a new sync and also how long we should wait.

kolibri/core/auth/management/commands/sync.py
jredrejo: Done

kolibri/core/public/utils.py
jredrejo: I had to modify the `sync` command and the functions calling it to accept this new interval parameter that's now sent by the server everytime a sync is accepted.

kolibri/core/auth/management/commands/sync.py
```diff
@@ -82,6 +82,12 @@ def add_arguments(self, parser):
             action="store_true",
             help="do not create a facility and temporary superuser",
         )
+        parser.add_argument(
+            "--resync-interval",
+            type=int,
+            default=5,
```
rtibbles: I think this should default to `None` so that we only do the resync in the case that we are calling it from inside a scheduled job.

kolibri/core/auth/management/commands/sync.py
```diff
@@ -270,7 +278,8 @@ def handle_async(self, *args, **options):  # noqa C901
             self.job.save_meta()
 
         dataset_cache.deactivate()
-        schedule_new_sync(baseurl, user_id)
+        if user_id:
```
rtibbles: Can change this to `if user_id and if resync_interval:`

jredrejo: Thanks, makes sense, done.

kolibri/core/public/utils.py
```diff
@@ -171,35 +211,47 @@ def request_soud_sync(server, user=None, queue_id=None, ttl=10):
...
-            logger.info(
-                "Server busy, will try again in {} seconds with pk={}".format(
-                    time_alive, pk
-                )
-            )
+
+def schedule_new_sync(server, user, interval=OPTIONS["Deployment"]["SYNC_INTERVAL"]):
+    # reschedule the process for a new sync
+    dt = datetime.timedelta(seconds=interval)
+    job = Job(request_soud_sync, server, user)
```
rtibbles: One thing I realized as I was stopping and restarting my server for testing is that while we are deduplicating the actual sync jobs above, the scheduled checks are getting duplicated still here.

My scheduled jobs table is growing with each restart. Maybe an id of a hash of `server:user` might help to deduplicate?

jredrejo: Done. The number of jobs now is reasonable. One thing I noticed is that restarting the server makes zeroconf pick a different url between all the ips available in the server machine, so with the same server sometimes it creates new hashes. 

I can think of cases where one learner syncs with different servers of different coaches in the same network  so it seems the server must be in the hash.

rtibbles (approved):

Issue #8210, opened 2021-07-16
Coach Bookmarking functionality - quizzes 


marcellamaki:
Coaches should be able to select bookmarks from within the quiz creation feature. Within the quiz creation page, the coach should be able to navigate into the bookmarks through a new Bookmarks card added to the page, and then, see bookmarked resources that the coach can select through checkboxes to add to the resource.  (Note: in the codebase, "quizzes" are also called "exams")

<img width="921" alt="Screen Shot 2021-07-16 at 11 14 24 AM" src="https://user-images.githubusercontent.com/17235236/125970095-4adfb1bc-042e-4891-963d-764bdb2c4112.png">

```
Feature: Coaches can bookmark resources and topics 

       Scenario: Coaches can bookmark resources while in quiz creation flow
		Given that I am creating a quiz
			And I navigate to a list of exercises and topics
		When I click bookmark icon on the exercise or topic card
		Then I see a snackbar confirming the resource or topic was bookmarked
			And the resource or topic is added to my list of bookmarks

Feature: Coaches can select and use bookmarked resources in quiz creation
	
	Scenario: coaches add a bookmarked exercise to a quiz
		Given that I am creating a quiz
		When I browse the new bookmarks content tree
			And I see a list of bookmarks consisting of only exercises and topics
			And I find and select a bookmarked exercise via the checkbox
		Then I see a snackbar appear confirming the exercise was selected

	Scenario: coaches add a bookmarked topic to a quiz
		Given that I am creating a quiz
		When I browse the new bookmarks content tree
			And I see a list of bookmarks consisting of only exercises and topics
			And I find and select a bookmarked topic via the checkbox
		Then I see a snackbar appear confirming the topic was selected

	Scenario: coaches add a multiple bookmarked items to a quiz
		Given that I am creating a quiz
		When I browse the new bookmarks content tree
			And I find and select multiple bookmarked items via multiple checkboxes
		Then I see a snackbar appear confirming the items were selected
```

rtibbles:
This is complete.

Issue #8211, opened 2021-07-16
Selecting bookmarks within lesson management


marcellamaki:
Coaches should be able to select bookmarks from within the context of creating or managing lessons. Just like with quiz creation, the coach should be able to navigate into the bookmarks through a new Bookmarks card added to the page, and then, see bookmarked resources that the coach can select through checkboxes to add to the resource.

**Note from the designer on this interaction:** 
> lessons uses an auto-save feature per selection, so the user just has to “close” out

[Figma designs](https://www.figma.com/file/uiZuCIqh2KYvBUfbCiJyyA/Hybrid-Learning?node-id=121%3A1):
<img width="905" alt="Screen Shot 2021-07-16 at 11 24 54 AM" src="https://user-images.githubusercontent.com/17235236/125971508-79433872-5f97-44b7-b515-9125036fda70.png">

Corresponding gherkin scenarios: 
```
Feature: Coaches can bookmark resources and topics

       Scenario: Coaches can bookmark resources while in lesson resource management flow
		Given that I have created a lesson
			And I am adding resources to my lesson
			And I see a list of available resources to add
		When I click bookmark icon on a resource or topic card
		Then I see a snackbar confirming the resource or topic was bookmarked
			And the resource or topic is added to my list of bookmarks

Feature: Coaches can select and use bookmarked resources in lesson creation

	Scenario: coaches add a bookmarked resource to a lesson
		Given that I have created a lesson
			And I am on the lesson resource management page
		When I browse the new bookmarks content tree
			And I see a list of bookmarks consisting of only resources and navigable topics
			And I find and select a bookmarked resource via the checkbox
		Then I see a snackbar appear confirming the resource was added to my lesson

	Scenario: coaches add a multiple bookmarked resources to a lesson
		Given that I have created a lesson
			And I am on the lesson resource management page
		When I browse the new bookmarks content tree
			And I see a list of bookmarks consisting of only resources and navigable topics
			And I find and select multiple bookmarked resources via multiple checkboxes
		Then I see a snackbar appear confirming the resources were added to my lesson
```

rtibbles:
Flagging that all of this has been implemented, except the ability to bookmark within the lesson creation flow.

rtibbles:
We have previously intentionally restricted bookmarking only to resources - I think we should do the same here.

rtibbles:
With this in mind, this has already been implemented within the 0.18 work in develop:

[...]

PR #8272, opened 2021-08-11
Bookmarks feature

Evgeni998:
## Summary
Add a new functionality to see your bookmarks in the creation of new quiz menu (Coach -> Plan ->Quizzes -> New Quiz -> Create New Quiz) and also in the Manage lessons menu (Coach -> Plan -> Lessons -> Lesson -> Manage Resources). The functionality works fine when you open the bookmarks and if you try to go deeper into nested folders, but the `ResourceSelectionBreadcrumbs` component hasn't been implemented to work yet (in other words, you can't navigate into the different stages of the folder chain yet, I will attach a screenshot, to show you what I mean.) The bookmarks that are seen are dynamic (come from the API), and if you add a new bookmark it should show in the bookmark component.
**NOTE:** I have left the old route watcher logic in the `LessonResourceSelection` page because the logic is the same as in the `CreateExamPage` file, and I am using it as a template of how to make it work properly and change it later there as well
Screenshot from 2021-08-13 13-32-57 

Here are alson screenshots of the bookmark menu as it is right now

Screenshot from 2021-08-13 13-40-29
**-----------------------------------------------------------------------------------------------------------**
Screenshot from 2021-08-13 13-41-04

## References
Here are the issues that the PR is trying to close
https://github.com/learningequality/kolibri/issues/8210
https://github.com/learningequality/kolibri/issues/8211

## Reviewer guidance
You can test the changes by going to `Coach -> Plan ->Quizzes -> New Quiz -> Create New Quiz` and `Coach -> Plan -> Lessons -> Lesson -> Manage Resources`. You can also manually make a POST request to create a new bookmark and see it reflect on the current bookmark component.

----

## Testing checklist

## PR process


.bash_profile
```diff
@@ -0,0 +1 @@
+export KOLIBRI_RUN_MODE="dev"
```
indirectlylit: this should be in your `$HOME` folder, not checked into the repo

kolibri/plugins/coach/assets/src/views/plan/CreateExamPage/index.vue
```diff
@@ -159,13 +186,19 @@
   import pickBy from 'lodash/pickBy';
   import BottomAppBar from 'kolibri.coreVue.components.BottomAppBar';
   import commonCoreStrings from 'kolibri.coreVue.mixins.commonCoreStrings';
+  import {
+    BookmarksResource,
+    ContentNodeResource,
+  } from '../../../../../../../core/assets/src/api-resources/index';
```
indirectlylit: When importing from `core` to a plugin, use the `kolibri` namespace rather than relative paths. For example:

```js
import { ContentNodeResource } from 'kolibri.resources';
```

ref: https://kolibri-dev.readthedocs.io/en/develop/frontend_architecture/core.html

kolibri/plugins/coach/assets/src/views/plan/CreateExamPage/index.vue
```diff
@@ -81,44 +81,71 @@
...
+        />
+      </div>
+      <div v-if="showChannels">
+        <p>Select from bookmarks</p>
+        <div @click="lessonCardClicked">
+          <LessonContentCard
+            title="Bookmarks"
+            :link="{}"
+            kind="bookmark"
+            :description="this.bookmarks.length + ' resources'"
```
indirectlylit: All user-facing strings need to be localized.

For details, see https://kolibri-dev.readthedocs.io/en/develop/i18n.html

kolibri/core/assets/src/views/ContentIcon.vue
```diff
@@ -58,6 +58,13 @@
         :color="color"
         style="top:0;"
       />
+      <KIcon
+        v-if="is('bookmark')"
+        icon="bookmark"
+        :class="[colorClass]"
+        :color="color"
+        style="top:0;"
+      />
```
indirectlylit: Bookmarks are not [content nodes](https://kolibri-dev.readthedocs.io/en/develop/backend_architecture/content/concepts_and_definitions.html) (in the sense of Kolibri resources that come from Studio channels), so they should not be added to `ContentIcon` or `ContentNodeKinds`.

kolibri/plugins/coach/assets/src/views/plan/CreateExamPage/index.vue
```diff
@@ -359,18 +396,47 @@
...
     watch: {
+      $route(to, from) {
+        if (this.$route.params.topicId) {
+          ContentNodeResource.fetchCollection({
+            getParams: { parent: this.$route.params.topicId },
+          }).then(bookmarks => {
+            this.bookmarkNodes = bookmarks;
+          });
+        }
+      },
```
indirectlylit: I haven't inspected this in detail, but generally route handling is not performed in components using `watch: $route`. (`watch` in general is often a brittle tool). In most other places, routes populate vuex state which then drives component rendering. It's a bit more boilerplate but keeps things easier to reason about.

kolibri/plugins/coach/assets/src/views/plan/LessonResourceSelectionPage/index.vue
```diff
@@ -15,52 +15,84 @@
...
+          @moreresults="handleMoreResults"
+        />
+      </div>
+      <div v-else>
+        <div @click="lessonCardClicked">
+          <LessonContentCard
+            title="Bookmarks"
+            :link="{}"
+            kind="bookmark"
+            :description="this.bookmarks.length + ' resources'"
```
indirectlylit, line 47:
```suggestion
            :description="bookmarks.length + ' resources'"
```

indirectlylit (commented):
Hi @Evgeni998 thank you for your work so far on this! A few things caught my eye so I'm just leaving some quick notes inline.

Also, it's always appreciated to fill in the PR description template, especially with links to any mockups or specs you're following. That makes reviewing much easier.

thanks again!

kolibri/core/assets/src/views/ContentIcon.vue
marcellamaki: It seems like the bookmark icon doesn't fit neatly into any of our existing icon files, @Evgeni998, so I think it's okay to create a file just for this icon, unless @indirectlylit or someone else has a suggestion of a good place for it to live. 

You can look at `PermissionIcon.vue` or `LanguagesIcon.vue` for some references of how to do that in a file only contains one or two icon options.

kolibri/plugins/coach/assets/src/views/plan/CreateExamPage/index.vue
marcellamaki: It might be helpful to reference some of the other places where links are used in this file, like the [topicsLink](https://github.com/learningequality/kolibri/blob/b46ec1b5b0376dcd8fc86fb9628e32cb091042e5/kolibri/plugins/coach/assets/src/views/plan/CreateExamPage/index.vue#L575) as a starting point for getting familiar with our routing. It's taken me a while to wrap my head around how and where we are use route objects in Kolibri -- I still manage to get feedback about it every PR 😄 -- so maybe we can chat on Monday about this when I sign on!

kolibri/plugins/coach/assets/src/views/plan/CreateExamPage/index.vue
```diff
@@ -81,44 +81,71 @@
       </KGrid>
 
       <h2>{{ $tr('chooseExercises') }}</h2>
+      <div v-if="!showChannels">
+        <br>
+        <ContentCardList 
+          :contentList="bookmarkNodes"
+          :contentHasCheckbox="contentHasCheckbox"
+          :contentCardMessage="() => {return ''}"
```
marcellamaki: Props don't have to be bound, but can also be constant values. So to simplify this, you can just set the value to the empty string (or another constant value, in other places where you use this pattern). It could be rewritten as:
`contentCardMessage=''` 
(note that the : at the front of the prop has been removed)

marcellamaki (commented):
Nice work so far! I've added a few follow ups to Devon's comments and I'll set up a time that we can talk through the feedback, what you're still working on, and what you may have questions about.

kolibri/plugins/coach/assets/src/views/plan/CreateExamPage/index.vue
Evgeni998: Yes, that makes sense, that was also my first approach `contentCardMessage=""` but then i receive this error
Screenshot from 2021-08-16 15-00-18 
and that is why i passed it as a function.

kolibri/core/assets/src/views/ContentIcon.vue
indirectlylit: that sounds right to me

kolibri/plugins/coach/assets/src/views/plan/CreateExamPage/index.vue
```diff
@@ -81,44 +81,71 @@
       </KGrid>
 
       <h2>{{ $tr('chooseExercises') }}</h2>
+      <div v-if="!showChannels">
+        <br>
```
indirectlylit: generally it's preferable to use CSS rather than HTML for layout spacing (e.g. add a top margin)

kolibri/plugins/coach/assets/src/views/plan/CreateExamPage/index.vue
indirectlylit: yeah, looks like in this case the prop requires a function which returns the value:

https://github.com/learningequality/kolibri/blob/b46ec1b5b0376dcd8fc86fb9628e32cb091042e5/kolibri/plugins/coach/assets/src/views/plan/LessonResourceSelectionPage/ContentCardList.vue#L113-L117

rtibbles: Ooof - that seems like an idiosyncratic choice, but probably now is not the time to change it.

indirectlylit:
@Evgeni998 FYI once you've addressed some feedback from the PR, feel free to hit the 'resolve' button:

image

thanks!

.bash_profile
Evgeni998: Fixed

kolibri/plugins/coach/assets/src/views/plan/CreateExamPage/index.vue
Evgeni998: Changed. Thank you

kolibri/plugins/coach/assets/src/views/plan/CreateExamPage/index.vue
Evgeni998: Added localization

kolibri/core/assets/src/views/ContentIcon.vue
Evgeni998: I added the icon into new file `BookmarkIcon.vue`. Thanks

kolibri/plugins/coach/assets/src/views/plan/CreateExamPage/index.vue
Evgeni998: I changed the fetching logic to go into route handlers

marcellamaki:
Hi @Evgeni998 - you've made a lot of great updates on this, so thanks for your revisions. Overall, most of the functionality seems to be working on the front end and I think this is very close! I'm going to take a closer look at the code itself and provide some feedback and ask a few team members to take a look also, but the one thing I noticed with a quick manual QA is that single listed bookmarks on the don't seem to be added on check, whereas if you navigate into a bookmarked topic, the functionality is working as expected. Just wanted to raise this so you could take a look as folks continue to review today, so you have something to keep working on. Please feel free to reply in the comments if you have any questions. 

I've included a gif below for reference - you will see the first resource that I check, there is no snackbar indicator, and after closing the modal it doesn't show up on the list of resources, whereas the second selection, after navigating into a topic, works well!

bookmarks-single-resouce

kolibri/plugins/coach/assets/src/views/plan/CreateExamPage/index.vue
```diff
@@ -103,7 +103,7 @@
             :title="$tr('bookmarks')"
             :link="getBookmarksLink()"
             :kind="$tr('bookmark')"
-            :description="'0 resources'"
+            :description="this.bookmarksCount + $tr('resources')"
```
indirectlylit: We cannot use string concatenation like this to construct user-facing strings, because in some languages the ordering of the words might need to be reversed.

Instead, always inject values into the string as variables using ICU syntax:
https://kolibri-dev.readthedocs.io/en/develop/i18n.html#icu-message-syntax

kolibri/core/assets/src/constants.js
```diff
@@ -27,6 +27,7 @@ export const ContentNodeKinds = {
   CLASSROOM: 'CLASSROOM',
   ACTIVITY: 'ACTIVITY',
   SLIDESHOW: 'slideshow',
+  BOOKMARK: 'bookmark',
```
nucleogenesis, line 30: Bookmarks are not ContentNodes, but are only references to ContentNode so this should be removed. I don't see anywhere that you're using `ContentNodeKinds.BOOKMARK` so removal shouldn't impact anything else unless I'm missing the usage.

kolibri/plugins/coach/assets/src/modules/examCreation/handlers.js
```diff
@@ -118,6 +124,56 @@ export function showExamCreationTopicPage(store, params) {
     });
   });
 }
+export function showExamCreationBookamrksPage(store, params) {
```
nucleogenesis: typo here `Bookamrks`

kolibri/plugins/coach/assets/src/views/plan/LessonResourceSelectionPage/index.vue
```diff
@@ -290,6 +333,11 @@
           });
       }
     },
+    created() {
+      this.getBookmarks().then(() => {
+        this.getBookmarksData();
+      });
+    },
```
nucleogenesis: There is room for optimization here as we're hitting the DB two separate times when we could get the data in a single query. If this isn't in scope for this PR then we can tackle it in a follow-up issue. I wish I'd thought of it when working on #8095

kolibri/plugins/coach/assets/src/modules/examCreation/handlers.js
```diff
@@ -118,6 +124,56 @@ export function showExamCreationTopicPage(store, params) {
...
+    const loadRequirements = [topicNodePromise, childNodesPromise];
+
+    return Promise.all(loadRequirements).then(([topicNode, childNodes]) => {
+      return filterAndAnnotateContentList(childNodes).then(() => {
+        store.commit('SET_TOOLBAR_ROUTE', {
+          name: PageNames.EXAMS,
+        });
+        return showExamCreationPage(store, {
+          classId: params.classId,
+          bookmarksList: childNodes,
```
nucleogenesis, line 147: Where does the call to the Bookmarks API occur in this context? This looks like it will just assign all child nodes for the topic node to the `bookmarksList` key.

nucleogenesis: Seems like there is a missing call to the `getBookmarks` function below?

kolibri/plugins/coach/assets/src/modules/examCreation/handlers.js
```diff
@@ -118,6 +124,56 @@ export function showExamCreationTopicPage(store, params) {
...
+  return store.dispatch('loading').then(() => {
+    let dataArr = [];
+    getBookmarks()
+      .then(bookmarks => {
+        bookmarks.forEach(bookmark => {
+          ContentNodeResource.fetchModel({ id: bookmark.contentnode_id }).then(data => {
+            dataArr.push(data);
+          });
+        });
+      })
```
nucleogenesis: Similarly to a previous comment, it seems like we should consider making an API endpoint (possibly in a follow-up PR) on the Bookmarks API that returns a combination of Bookmarks and their associated ContentNodes.

In the absence of that optimization, this could avoid making an API request for every bookmark using `fetchCollection`  with `id__in` like:

```
const contentNodeIds = bookmarks.map(b => b.contentnode_id);
ContentNodeResource.fetchCollection({id__in: contentNodeIds})
    .then(/* and so on */)
```

Which should get you all of the ContentNode data with one request.

kolibri/plugins/coach/assets/src/modules/examCreation/handlers.js
```diff
@@ -118,6 +124,56 @@ export function showExamCreationTopicPage(store, params) {
...
+          pageName: PageNames.EXAM_CREATION_BOOKMARKS,
+          ancestors: [...topicNode.ancestors, topicNode],
+        });
+      });
+    });
+  });
+}
+export function showExamCreationAllBookmarks(store) {
+  return store.dispatch('loading').then(() => {
+    let dataArr = [];
```
nucleogenesis: You don't have to create a variable here outside of the scope of the Promise chain below, but rather can return from one `then` into the next.

```
Promise.resolve(1)
  .then(one => one + 1)
  .then(two => two + 1)  // The value of this function's `two` param is what the function above returned
  .then(three => {
  	return three + 1
   })
  .then(four => console.log(four)); // Outputs 4
```

So in the first `then(/* ... */)` you can gather and `return` the collection of bookmarks, which will be the parameter given to the next `then` clause.

kolibri/plugins/coach/assets/src/views/plan/LessonResourceSelectionPage/index.vue
```diff
@@ -298,6 +346,21 @@
         addToWorkingResources: 'ADD_TO_WORKING_RESOURCES',
         removeFromSelectedResources: 'REMOVE_FROM_WORKING_RESOURCES',
       }),
+      getBookmarksData() {
+        this.bookmarks.forEach(bookmark => {
+          ContentNodeResource.fetchModel({ id: bookmark.contentnode_id }).then(data => {
+            this.bookmarkNodes.push(data);
+          });
+        });
+      },
```
nucleogenesis: Another place where we can optimize against making a call for every Bookmark's ContentNode

nucleogenesis (changes requested):
Great work so far! I spotted a few places where we might run into making too many DB queries and posted a few questions.

kolibri/core/assets/src/constants.js
Evgeni998: I added this line because when I pass the prop kind in the `LessonContentCard `component in `CreateExamPage/index.vue` file in one case the `kind `is bookmark. in that case the received prop goes through validation
Screenshot 2021-08-24 180409
and that function goest to see if the string bookmarks is part of the `ContentNodeKinds` object. 
Screenshot 2021-08-24 180743
If i remove this line, i receive an error that check failed for kind
Screenshot 2021-08-24 180902
I think one option is to rework the `ValidateContentKind` function, but I might break something else

kolibri/plugins/coach/assets/src/modules/examCreation/handlers.js
Evgeni998: I'm getting the `childNodes` when I have opened a folder within the bookmarks menu and I'm going into nested folders. The first time I get the bookmarks is in this function
Screenshot 2021-08-27 143559
and after that the requests should be the same as in the content nodes from what I have seen so far. Maybe I could be mistaken. Let me know if I don't get this right.

kolibri/plugins/coach/assets/src/modules/examCreation/handlers.js
Evgeni998: Yes, this is a good point, fixed it. I made array of promises and then made the request. Thank you.
Screenshot 2021-08-27 180111

kolibri/core/assets/src/constants.js
nucleogenesis: I think that you could probably simplify altogether by replacing that use of `LessonContentCard` with some HTML there.

Since Bookmarks are not a "Kind" of ContentNode, I think it makes sense to create one-off code and styles rather than trying to slot the Bookmarks functionality into components like `ContentIcon` or `CardThumbnail` when really all you need is a simple card that says "Bookmarks", has a description and has the bookmark icon and that links properly to the list of Bookmarks.

So - in `kolibri/plugins/coach/assets/src/views/plan/LessonResourceSelectionPage/index.vue` where you use `LessonContentCard`, you ought to just use the same styles to create some raw HTML there, using `KRouterLink` to wrap everything.

One of the reasons is that the `kind` prop on `LessonContentCard` is supposed to map to a `KLabeledIcon` for the title which passes `kind` to `ContentCard` resulting in there being empty space to the left of the title. If there was an icon there, it'd be the same bookmark icon the user sees to the left, resulting in redundant icons.

kolibri/plugins/coach/assets/src/views/plan/CreateExamPage/index.vue
```diff
@@ -81,44 +81,68 @@
...
+        />
+      </div>
+      <div v-if="showChannels">
+        <p>Select from bookmarks</p>
+        <div @click="lessonCardClicked">
+          <LessonContentCard
+            :title="$tr('bookmarks')"
+            :link="getBookmarksLink()"
+            :kind="$tr('bookmark')"
+            :description="$tr('resources', { count: this.bookmarksCount })"
```
nucleogenesis: When there are no resources, this results in a string that just says `resources` so we should probably just hide this whole card if there are no bookmarks available - alternatively we could include a string that tells the user there are no bookmarks (thoughts @jtamiace ?)

kolibri/plugins/coach/assets/src/views/plan/CreateExamPage/index.vue
```diff
@@ -81,44 +81,68 @@
...
+          :contentIsIndeterminate="() => false"
+          @changeselectall="toggleTopicInWorkingResources"
+          @change_content_card="toggleSelected"
+          @moreresults="handleMoreResults"
+        />
+      </div>
+      <div v-if="showChannels">
+        <p>Select from bookmarks</p>
+        <div @click="lessonCardClicked">
+          <LessonContentCard
```
nucleogenesis: See my response to a previous comment about replacing this whole section with some ad hoc HTML to avoid having to tweak other components just for this one appearance of Bookmarks

kolibri/plugins/coach/assets/src/views/plan/CreateExamPage/index.vue
nucleogenesis: screenshot-localhost_8000-2021 08 27-11_13_23

kolibri/plugins/coach/assets/src/modules/lessonResources/handlers.js
```diff
@@ -132,7 +140,50 @@ export function showLessonResourceSelectionTopicPage(store, params) {
...
+  return BookmarksResource.fetchCollection()
+    .then(bookmarks => {
+      return bookmarks.map(bookmark => {
+        return ContentNodeResource.fetchModel({ id: bookmark.contentnode_id });
+      });
+    })
+    .then(bookmarkPromises => {
+      return Promise.all(bookmarkPromises);
+    });
+}
```
nucleogenesis: Thanks for reworking this but unfortunately the problem I was describing still exists.

If I have 40 Bookmarks, this will result in 41 requests. The first to get my Bookmarks, the other 40 to get each ContentNode one at a time from the API. In order to avoid the potential performance issues this could cause, we should make sure to limit the quantity of API requests whenever possible.

In this instance, you can rework it as follows:

```
return BookmarksResource.fetchCollection() 
    .then(bookmarks => bookmarks.map(bookmark => bookmark.contentnode_id))
    .then(contentNodeIds => {
            getParams: {
                ids: contentNodeIds, // This filters only the ids we want
            }
        });
    });
```

This is a bit obscure in terms of being a capability of the `fetchCollection` method and our API altogether. There is some [Django Rest Framework magic involved](https://www.django-rest-framework.org/api-guide/filtering/#djangofilterbackend) that allows us to filter queries against certain properties of the model using GET params. 

_Taking things a step further_ - we have run into issues previously where if there are too many ContentNodes being requested at once, the server will slow down or throw an error. In order to avoid this, we will want to use [lodash/chunk](https://lodash.com/docs/4.17.15#chunk) to break the IDs into lists of a certain size.

```
import chunk from "lodash/chunk"; // at the top
/* ... */
return BookmarksResource.fetchCollection() 
    .then(bookmarks => bookmarks.map(bookmark => bookmark.contentnode_id))
    .then(contentNodeIds => {
        const chunkedContentNodeIds = chunk(contentNodeIds, 50); // Breaking contentNodeIds into lists no more than 50 in length
        // Now we will create an array of promises, each of which queries for the 50-id chunk
        const fetchPromises = chunkedContentNodeIds.map(idsChunk => {
                getParams: {
                    ids: idsChunk, // This filters only the ids we want
                }
            });
        });
        return Promise.all(fetchPromises); // Avoids overloading in a single query, but better than 1 query per ID
    });
```

I know this is a lot - I'd be happy to chat about this further next week for sure! Especially if things don't work as expected with the above :-)

kolibri/plugins/coach/assets/src/views/plan/LessonResourceSelectionPage/index.vue
```diff
@@ -15,52 +15,76 @@
       <h1>
         {{ $tr('documentTitle', { lessonName: currentLesson.title }) }}
       </h1>
```
nucleogenesis: When showing Bookmarks, this results in an empty string `Manage resources in "` - will need to hide this when showing Bookmarks as the main listed content nodes in the component.

nucleogenesis (dismissed):
I left a couple more comments requesting some additional refactors. I would be happy to chat further next week.

A couple other notes:

1 - Overall, I'm happy to get this PR merged sooner than later and we can start tying up loose ends - one of which is that it will probably be best to paginate how we get the Bookmarks and may need some design guidance there (@jtamiace ) - all of this could be tackled in a follow up issue.

2 - This is one issue that is probably going to be a bit complex to work out, but currently the routing is a bit wonky when you start going into your Bookmarks.

- When viewing Bookmarks in Lesson creation, the back button goes to a broken version of the previous page.
- One cannot navigate back and forth between "Lesson Creation -> Bookmarks -> Bookmarked Topic -> Subtopics ..." 
- I think Bookmarks should have its own entire page component rather than piggy-backing on LessonResourceSelectionPage - this will probably make routing within the Bookmarks easier to accomplish too as we may want to introduce the pagination mentioned above as well.

@jtamiace - I'm not sure how much of - if any of the above should be applied in this PR vs a follow-up discussion and/or issue. It seems kind of ambiguous from the user's perspective where they are between searching for content and viewing their bookmarks as is.

jtamiace:
> One cannot navigate back and forth between "Lesson Creation -> Bookmarks -> Bookmarked Topic -> Subtopics ..."

@nucleogenesis follow up might be best for this because looking at the designs it's not really clear what the intended behavior should be when you click into a subtopic from bookmarks. Does the current implementation take the user out of the bookmarks and into "Channel > Topic > Subtopic"?

And for pagination, I've used this in the topic/category search UI that can also be used for bookmarks. I'm not exactly sure how many results per "page" would be best for performance though.

Screen Shot 2021-08-30 at 10 58 05 AM

marcellamaki (approved):
Per meeting between @Evgeni998 and @nucleogenesis and follow up conversations I've had with @nucleogenesis and @rtibbles, going to go ahead and merge this, and @nucleogenesis will create follow up issues as needed.

PR #8277, opened 2021-08-13
SoUD Status Indicator & Learn-only Notification & Block Facility/Coach 

### LINKED ISSUE #8263: Add limited device messaging
**Overview:**

Devices that have been affected by single user syncing need to have an indicator that denotes the device is in a “learn-only” mode, meaning coach and admin features are not available on that device. 

[Figma designs](https://www.figma.com/file/MkIaNq2w3sx4Sh2uVOd35K/Single-user-syncing?node-id=11%3A643): 
Screen Shot 2021-08-05 at 10 50 49 AM

Screen Shot 2021-08-05 at 10 52 00 AM
Screen Shot 2021-08-05 at 10 52 10 AM
Screen Shot 2021-08-05 at 10 52 33 AM

### LINKED ISSUE #8264: Make learner polling indicator conditional for only SoUD
For the new polling indicator on the dropdown menu, this should be updated to appear and poll only when the user is a learner who is logged in on a subset of users device. 

Screen Shot 2021-08-05 at 10 58 50 AM

nucleogenesis:
## Summary

- Bootstraps the `DeviceSettings.subset_of_users_device` setting into `plugin_data['isSubsetOfUersDevice']` 
- Conditionally shows the Status Indicator introduced in #8260 only when on a SoUD and logged in as Learner
- Conditonally shows a "learn only - admin and coach features not available" notice on the profile dropdown and side nav when on SoUD (to all kinds of users)
- Introduces two strings in the `LearnOnlyDeviceNotice` component @radinamatic 
- **(NEW)** Stifles Coach & Facility in their respective kolibri_plugin definitions by not loading the URLs when on SoUD.

If / when plugins can be disabled programmatically, this should be refactored out and the plugins should be properly disabled altogether.

_Learner's profile dropdown (with the notice)_
Profile-Dropdown-with-notice

_Learn-only notice in SideNav_
SideNav-with-notice

~This does not disable/hide or disallow routing for the Coach, Facility & Device sections but that can be done in a follow-up PR next week.~ _It does now_

## References

Fixes #8263 
Fixes #8264

## Reviewer guidance

- Set your DeviceSettings' first entry's `subset_of_users_device` to `1` in sqlitebrowser (or do it in the shell).
- Log in and see the notices in the SideNav and profile dropdown
- Does everything work when logged in as Coach? How about if you're accessing as a guest?

----

## Testing checklist

- [x] Critical user journeys are covered by Gherkin stories `features/learner/learner-learn-mode-only-devices.feature`

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated


kolibri/core/assets/src/mixins/nav-components.js
```diff
@@ -37,5 +37,15 @@ export default {
         return this.isLearner || this.isCoach || this.isAdmin || this.isSuperuser;
       }
     },
+    filterIfSoUD(isSoUD) {
+      // Returns a filter predicate depending on if on SoUD
+      if (!isSoUD) {
+        // Don't filter anything if not on SoUD
+        return () => true;
+      } else {
+        const hiddenNavItemRoles = [UserKinds.COACH, UserKinds.ADMIN];
```
rtibbles: Nice, I think this is the neatest way this could have been done.

Definitely agree with the Superuser being left there, as that's a device specific designation.

kolibri/core/assets/src/views/AppBar.vue
```diff
@@ -177,6 +187,9 @@
         fullName: state => state.core.session.full_name,
         userId: state => state.core.session.user_id,
       }),
+      userIsLearner() {
+        return this.getUserKind == UserKinds.LEARNER;
+      },
       menuOptions() {
         return navComponents
           .filter(component => component.section === NavComponentSections.ACCOUNT)
```
rtibbles, line 198: I was about to comment that we should add the filterBySoUD here as well, when I realized we could just do it at NavComponent registration:

https://github.com/learningequality/kolibri/blob/98bb185b73b247c0e774b229dd507fa6e74c4eaf/kolibri/core/assets/src/utils/navComponents.js#L39

Check for SoUD status in there ^ from plugin data, and just not add the component in the first place.

Probably good to do a logging.info when we do that so that future coders are not confused when their nav items don't show though!

kolibri/core/assets/src/views/AppBar.vue
```diff
@@ -168,6 +177,7 @@
         isPolling: false,
         // poll every 10 seconds
         pollingInterval: 10000,
+        isSubsetOfUsersDevice: plugin_data['isSubsetOfUsersDevice'],
```
marcellamaki, line 180: TIL that `plugin_data` is a thing that exists in Kolibri 🤯

kolibri/plugins/facility/assets/src/views/FacilityIndex.vue
```diff
@@ -112,6 +114,12 @@
         return false;
       },
     },
+    beforeMount() {
```
rtibbles: I think this is redundant with the URL changes you've made, so I'd rather just get rid of the change in this file.

Plus `isSubsetOfUsersDevice` is not defined for the plugin data for _this_ plugin.

rtibbles (changes requested):
Looks right - one thing that looks left over from a previous implementation.

kolibri/core/assets/test/views/app-bar.spec.js
```diff
@@ -66,11 +83,55 @@ describe('app bar component', () => {
...
+        let wrapper;
+        beforeEach(() => {
+          wrapper = createWrapper(undefined, { isSubsetOfUsersDevice: true }, true);
+        });
+
+        it('shows the Learn-only notice', () => {
+          expect(wrapper.find('[data-test="learnOnlyNotice"]').exists()).toBe(true);
+        });
+
+        describe('when signed in as a learner', () => {
```
marcellamaki: thank you for adding tests 😄

kolibri/plugins/facility/assets/src/views/FacilityIndex.vue
nucleogenesis: Thanks for the keen eye here! I reverted the CoachIndex changes but missed this one.

rtibbles (approved):

kolibri/core/assets/src/views/AppBar.vue
```diff
@@ -104,6 +104,12 @@
               <LogoutSideNavEntry v-if="isUserLoggedIn" />
             </template>
 
+            <template #footer>
+              <!-- Only show this when on a SoUD -->
+              <div v-if="isSubsetOfUsersDevice" class="role" data-test="learnOnlyNotice">
```
rtibbles: Oh, just noticed one other thing here - shouldn't we only show this to non-learners? Otherwise it's a slightly confusing notice "this is a learn only device, so you can only see the things you'd normally see!".

rtibbles (changes requested):
One more thing that I am not sure was in the spec but should be!

kolibri/core/assets/src/views/AppBar.vue
```diff
@@ -190,6 +190,12 @@
       userIsLearner() {
         return this.getUserKind == UserKinds.LEARNER;
       },
+      showSoudNotice() {
+        return (
+          this.isSubsetOfUsersDevice &&
+          [UserKinds.ADMIN, UserKinds.COACH].includes(this.getUserKind)
```
rtibbles: Can maybe just use `isAdmin || isCoach` from the getters here: https://github.com/learningequality/kolibri/blob/release-v0.15.x/kolibri/core/assets/src/state/modules/session.js#L18

That means we are showing the warning to superusers too, which seems relevant.

kolibri/core/assets/src/views/SideNav.vue
```diff
@@ -190,10 +191,17 @@
...
         privacyModalVisible: false,
-        isSubsetOfUsersDevice: plugin_data['isSubsetOfUsersDevice'],
+        isSubsetOfUsersDevice: plugin_data.isSubsetOfUsersDevice,
       };
     },
     computed: {
+      ...mapGetters(['getUserKind']),
+      showSoudNotice() {
+        return (
+          ![UserKinds.LEARNER, UserKinds.ANONYMOUS].includes(this.getUserKind) &&
```
rtibbles: Ditto!

rtibbles (commented):
Just one more thing...

kolibri/core/assets/src/views/AppBar.vue
nucleogenesis: :+1: updated!

rtibbles (approved):

PR #8298, opened 2021-08-14
fix bad order of the parameters to check role permissions

jredrejo:
## Summary

This fixes a bug in a permissions check that has been in our code base since Nov 2017.... it seems it was not important but with some sync permissions checks it's appeared now.

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/core/auth/models.py
+++ b/kolibri/core/auth/models.py
@@ -742,7 +742,7 @@ class FacilityUser(KolibriAbstractBaseUser, AbstractFacilityDataModel):
             target_user = FacilityUser.objects.get(id=scope_params.get("user_id"))
             if self == target_user:
                 return True
-            if self.has_role_for_user(target_user, role_kinds.ADMIN):
+            if self.has_role_for_user(role_kinds.ADMIN, target_user):
                 return True
             return False
         return False


kolibri/core/auth/backends.py
```diff
@@ -6,7 +6,7 @@
 from kolibri.core.auth.models import FacilityUser
 
 
-FACILITY_CREDENTIAL_KEY = "facility"
+FACILITY_CREDENTIAL_KEY = "facility_id"
```
rtibbles: I remember when @bjester first implemented this, he had it as `facility_id` and then changed it back. At the moment this will break login, as it is still passing `facility`.

If possible, I'd prefer we not include this change, and just pass `facility` for the facility id in the authentication parameters.

kolibri/core/auth/models.py
```diff
@@ -742,7 +742,7 @@ def has_morango_certificate_scope_permission(
             target_user = FacilityUser.objects.get(id=scope_params.get("user_id"))
             if self == target_user:
                 return True
-            if self.has_role_for_user(target_user, role_kinds.ADMIN):
+            if self.has_role_for_user(role_kinds.ADMIN, target_user):
```
rtibbles, line 745: I feel like this PR could be limited to just this change, and then the need to make the above changes removed by changing in this line: https://github.com/learningequality/kolibri/pull/8185/files#diff-c968516236d8c64ebcae34e7fb63ce29d0d217e6642e416de5386a14a362ceddR54 to `facility` instead of `facility_id`.

kolibri/core/auth/backends.py
jredrejo: in such case authentication must be changed because using `facility` just does not work. That's why I changed it

jredrejo: This part https://github.com/learningequality/kolibri/blob/release-v0.15.x/kolibri/core/auth/management/utils.py#L238-L249 does not work if `facility` is used, but it does using `facility_id`

kolibri/core/auth/models.py
jredrejo: Happy with it if we can change https://github.com/learningequality/kolibri/blob/release-v0.15.x/kolibri/core/auth/management/utils.py#L238-L249 to work with `facility`

kolibri/core/auth/backends.py
rtibbles: Ah yes, I think this code was the reason that @bjester originally had it set to `facility_id` so I'd like to understand why he switched it back!

bjester: @jredrejo There's an [integration test](https://github.com/learningequality/kolibri/blob/release-v0.15.x/kolibri/core/auth/test/test_morango_integration.py#L76) for this which verifies that code you linked does actually work as it was prior to your changes. I'm not sure what issue you're running into. If you look [here](https://github.com/learningequality/kolibri/blame/40dcc2408b86980e3d17720da476725c28792fd2/kolibri/core/auth/backends.py#L14), you can see that `FacilityUserBackend` was originally accepting a `facility` keyword argument. So changing `facility` to `facility_id` could be breaking more functionality, since support for `facility` was existing prior to any 0.15 change

jredrejo: commit removed!

rtibbles (approved):

bjester (approved):

PR #8299, opened 2021-08-15
Single worker pool for tasks

vkWeb:
## Summary

This PR implements a dynamic single worker pool for tasks. Dynamic as in `high workers` come into action only when `regular workers` are busy and we still have jobs with `high` priority. 

If there are less workers running than there are `regular workers`, we look first for jobs with `high` priority, if found one we run it else we look for jobs with `regular` priority, if found we run it. 

If all `regular workers` are busy, then the remaining workers only look for `high` priority jobs. If found one, we run it.

This algorithm will make sure `high` priority jobs don't need to wait. 

## Reviewer guidance

Is the single worker pool logic suitable for our tasks needs? Do you see any edge cases?

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/core/tasks/storage.py
+++ b/kolibri/core/tasks/storage.py
@@ -5,7 +5,6 @@ from copy import copy
 from sqlalchemy import Column
 from sqlalchemy import DateTime
 from sqlalchemy import func
-from sqlalchemy import Integer
 from sqlalchemy import or_
 from sqlalchemy import PickleType
 from sqlalchemy import String
@@ -14,6 +13,7 @@ from sqlalchemy.ext.declarative import declarative_base
 from sqlalchemy.orm import sessionmaker
 
 from kolibri.core.tasks.exceptions import JobNotFound
+from kolibri.core.tasks.job import Priority
 from kolibri.core.tasks.job import State
 from kolibri.utils.conf import OPTIONS
 
@@ -31,14 +31,14 @@ class ORMJob(Base):
 
     __tablename__ = "jobs"
 
-    # The hex UUID given to the job upon first creation
+    # The hex UUID given to the job upon first creation.
     id = Column(String, primary_key=True, autoincrement=False)
 
     # The job's state. Inflated here for easier querying to the job's state.
     state = Column(String, index=True)
 
-    # The job's order in the entire global queue of jobs.
-    queue_order = Column(Integer, autoincrement=True)
+    # The job's priority. Helps to decide which job to run next.
+    priority = Column(String, index=True)
 
     # The queue name passed to the client when the job is scheduled.
     queue = Column(String, index=True)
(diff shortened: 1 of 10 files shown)


vkWeb:
@rtibbles this PR changes the job storage database table, looks like our version upgrade check of `<=0.15.x` will recreate the tables?

rtibbles:
> will recreate the tables?

Yes, should do - so this should be fine!

vkWeb:
@rtibbles this is ready for review :rocket: I've added tests that cover important code updates. 

The test that used to fail on postgres from now on should not fail.

kolibri/core/tasks/worker.py
```diff
@@ -91,23 +87,22 @@ def start_job_checker(self):
...
+
         Returns: None
         """
-        try:
-            while len(self.future_job_mapping) < self.num_workers:
-                self.start_next_job()
-        except Empty:
-            logger.debug("No jobs to start.")
-        for job in self.storage.get_canceling_jobs(self.queues):
+        job_to_start = self.get_next_job()
```
rtibbles, line 101: The removal of the while loop here means that we will only ever queue one job during a single run of `check_jobs`.

I think it might be good to do something more along the lines of:

```
while job_to_start:
    self.start_next_job(job_to_start)
    job_to_start = self.get_next_job()
```

That way we will fill up the worker threads as fast as possible.

kolibri/core/tasks/storage.py
```diff
@@ -143,39 +147,71 @@ def mark_job_as_canceling(self, job_id):
         """
         self._update_job(job_id, State.CANCELING)
 
-    def get_next_queued_job(self, queues):
+    def get_next_queued_job(self, queues=None, priority_order=Priority.PriorityOrder):
```
rtibbles: Is there a use case for filtering by `queues` for the next queued job now? My assumption was that we would only be using the `queues` field for filtering jobs to return via the API?

rtibbles (commented):
One slight regression, one question.

kolibri/core/tasks/worker.py
vkWeb: Ummm, yes you are right. Adding the loop back.

kolibri/core/tasks/storage.py
vkWeb: No, there's no use case. Let us remove that?

rtibbles: Yes, it should simplify the code!

rtibbles (approved):

kolibri/core/tasks/storage.py
```diff
@@ -143,39 +147,68 @@ def mark_job_as_canceling(self, job_id):
...
-    def get_next_queued_job(self, queues):
+    def get_next_queued_job(self, priority_order=Priority.PriorityOrder):
+        """
+        Looks for the oldest queued job with priorities provided in `priority_order`. It
+        runs through the `priority_oder` list sequentially.
+
+        If `queues` is not None and a list of queue names, we only look in those `queues` else
+        we look irrespective of `queue`.
+
+        :param queues: the queue names we should look into.
```
vkWeb, line 158: @rtibbles code docs updated in task api permission PR.

PR #8300, opened 2021-08-16
Additional Gherkin scenarios for syncing and importing of facility after Wi-Fi network change

pcenov:


radinamatic (approved):
Looking good, we can organize the new scenarios on TQ as we please! 👍🏽 💯

PR #8303, opened 2021-08-17
Task APIs now behave based on registered job's permissions

vkWeb:
## Summary
This PR makes the `list`, `retrieve`, `create` and `restarttask` API work based on the registered job's permissions. This change was done to support use cases found on #8185. 

@jredrejo @rtibbles after we merge this PR, we'll be able to query the task API to list tasks only for a particular `queue`.  

Tests are being worked upon, will be pushed by today itself. 

## Reviewer guidance
Does this new permission based API suits our use cases? Any security considerations?  

----

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated


kolibri/core/tasks/api.py
```diff
@@ -230,38 +240,64 @@ def validate_create_req_data(self, request):
                 )
 
             # Check permissions the DRF way
```
rtibbles: Maybe move this comment inside the function where you moved the code too?

kolibri/core/tasks/api.py
```diff
@@ -197,16 +198,25 @@ def initial(self, request, *args, **kwargs):
...
-        return [CanManageContent]
+    def check_registered_job_permissions(self, request, registered_job):
+        """
+        Checks whether the `request` is allowed to proceed based on
+        `registered_job.permissions`.
+
+        Raises `PermissionDenied` if `request.user` is not allowed to proceed.
+        """
+        for permission in registered_job.permissions:
+            if not permission.has_permission(request, self):
```
rtibbles, line 224: One slight issue with just using the 'registered_job' and not the job object itself is that we might need more granular control.

For example, for the 'facility' queue jobs, we only want to allow a user to see jobs for their own facility (if they have appropriate permissions) and not for all facilities, to do this, we would have to introspect the Job object.

The superuser is of course the exception, as they can see all facility jobs across all facilities.

kolibri/core/tasks/job.py
```diff
@@ -277,26 +277,28 @@ def add(x, y):
         Look at each method's docstring for more info.
     """
 
+    from kolibri.core.tasks.queue import DEFAULT_QUEUE
+
     def __init__(
         self,
         func,
-        validator,
```
rtibbles, line 323: This seems to have been reverted from the previous PR?

rtibbles (commented):
Might need some tweaks to handle more fine grained permissions - possible that we could still maintain a common permissions framework between job creation and reading, by making it inspect the job object itself.

We might want to either use directly, or borrow heavily from the Kolibri Permissions here https://github.com/learningequality/kolibri/blob/release-v0.15.x/kolibri/core/auth/permissions/base.py#L15 (rather than using DRF permissions).

kolibri/core/tasks/api.py
vkWeb: @rtibbles do we have a DRF permission on the lines of `CanManageContentOfOwnFacility` or can we build one like we have here: https://github.com/learningequality/kolibri/blob/release-v0.15.x/kolibri/core/tasks/permissions.py...?

kolibri/core/tasks/api.py
vkWeb: No, not a great idea. That function has its own doc string that defines what its doing. 

Let this comment stay here for clarity of what's going on.

kolibri/core/tasks/job.py
vkWeb: No sir, actually I first removed the default values of the parameters but then I added them back so that instances of `RegisteredJob` can be created by passing only a part of the arguments. Might help in testing.

kolibri/core/tasks/api.py
rtibbles: My note was mostly just about the fact that the 'DRF way' refers to what is happening now inside the function that is being called.

kolibri/core/tasks/api.py
rtibbles: My thought was to do something similar to the `readable_by_user_filter` that we have on the Kolibri permissions classes: https://github.com/learningequality/kolibri/blob/b87f9da274a6a9218419cf9f858be695aa04304a/kolibri/core/auth/permissions/base.py#L57

We couldn't use it directly though, as it would have to be compatible with the SQLAlchemy interface we are using for querying jobs.

rtibbles: Or... because it depends on metadata inside the stored job object, it would probably have to do the filtering in Python anyway.

kolibri/core/tasks/job.py
rtibbles: Yes, I agree that having the defaults is helpful, I was just curious as you had previously had defaults, removed them, and now they were coming back in. I think this change is good, just wanted to make sure it was intentional.

kolibri/core/tasks/api.py
vkWeb: Resolved on Slack.

kolibri/core/tasks/api.py
```diff
@@ -290,7 +341,9 @@ def create(self, request):
 
                 request_data = validator_result
 
-            job_id = registered_job.enqueue(**request_data)
+            job_id = registered_job.enqueue(
+                facility_id=request.user.facility_id, **request_data
```
jredrejo: This is not working when is being executed from the setup wizard. In this case the request is done from an anonymous user that in the previous implementation has permission to execute this endpoint because it's in a non-provisioned device. 
Now we get this error and the wizard can not sync any user:
```
  File "/datos/le/mio/kolibri/kolibri/core/tasks/api.py", line 345, in create
    facility_id=request.user.facility_id, **request_data
AttributeError: 'KolibriAnonymousUser' object has no attribute 'facility_id'
```

jredrejo (dismissed):

kolibri/core/tasks/api.py
vkWeb: Looks like I should remove the `[IsAuthenticated]` permission class from the `permission_classes` and use `request.user.get(...)` instead of accessing the attribute directly.

vkWeb:
@rtibbles ready for merge.

vkWeb:
@rtibbles was just looking through open kolibri PRs. I saw this open. So thought I should ping you for review.

rtibbles (commented):
Code looks good, manual QA mostly checks out.

The only issue that I saw was that the pingback task was now being returned to the frontend during initial setup/provisioning when importing a facility from another device - as the frontend assumes that the first task returned is the only task returned, this was a slight issue.

It also seems like this is a relaxing of permissions - presumably the pingback task should not be returned when the device is unprovisioned?

vkWeb:
@rtibbles I've fixed it. This is ready for merge.

rtibbles (approved):

PR #8287, opened 2021-08-13
Spanish translation for the first batch of new 0.15 strings from Crowdin

radinamatic:
## Summary

Running `make i18n-download...` seems to have created a lot of changes in JSON files, not only for Spanish... 😕 

## References
…

## Reviewer guidance
Take look at the diff and the files, not sure what's the reason for all these changes.

----

## Testing checklist

- [X] Contributor has fully tested the PR manually
- [ ] If there are any front-end changes, before/after screenshots are included
- [ ] Critical user journeys are covered by Gherkin stories
- [ ] Critical and brittle code paths are covered by unit tests

## PR process

- [X] PR has the correct target branch and milestone
- [X] PR has 'needs review' or 'work-in-progress' label
- [X] If PR is ready for review, a reviewer has been added. (Don't use 'Assignees')
- [ ] If this is an important user-facing change, PR or related issue has a 'changelog' label
- [ ] If this includes an internal dependency change, a link to the diff is provided

## Reviewer checklist

- Automated test coverage is satisfactory
- PR is fully functional
- PR has been tested for [accessibility regressions](http://kolibri-dev.readthedocs.io/en/develop/manual_testing.html#accessibility-a11y-testing)
- External dependency files were updated if necessary (`yarn` and `pip`)
- Documentation is updated
- Contributor is in AUTHORS.md

PR #8319, opened 2021-08-18
0.15 strings ES translations from Crowdin

radinamatic:
## Summary
Running `make i18n-download...` creates a lot of changes in JSON files, not only for Spanish, but that is probably due to recent changes of the i18n workflows. No errors during the local manual testing.

## References
Supersedes #8287 

## Reviewer guidance
Take look at the diff and the files, test locally.

----

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated


rtibbles (approved):
Looking at the bn_BD json files as an example, all the diffs in string ids make sense to me. Looks like we've cleaned up some cruft, and are maybe including some messages that were missed before.

rtibbles (approved):

radinamatic:
Thank you!

@pcenov did a [critical workflows testing using Arabic UI](https://learningequality.testquality.com/project/12317/plan/21170/run/149932/result/2468485) this morning and found no regressions. We should be good 🙂

kolibri/plugins/setup_wizard/assets/src/views/importLODUsers/SelectFacilityForm.vue
```diff
@@ -69,7 +69,10 @@
       },
     },
     $trs: {
-      commaSeparatedPair: '{first}, {second}',
+      commaSeparatedPair: {
+        message: '{first}, {second}',
+        context: 'DO NOT TRANSLATE.',
+      },
```
indirectlylit, line 75: This string's ID and context guidance seem suspect.

[Not all languages use ASCII commas](https://stackoverflow.com/a/39286300) as list separators and even if they did in RTL languages it might need to be reordered.

kolibri/core/static/assets/fonts/noto-subset.ar.css
```diff
@@ -32,618 +32,613 @@ ABQCMgAKAkIAAAINAAACrQAeAtoAHgFXAAABDQAAAq0AHgLaAB4BlQAeAeYAHgH7AAACPgAA\
...
+AEYAAQAEAJUABABmAGUAcgAFAAwAFAAaACAAJgCUAAMAIwB8AI0AAgAEAI8AAgAGAJEAAgAI\
+AJMAAgAKAAQACgAQABYAHACMAAIABACOAAIABgCQAAIACACSAAIACgABAAMAAwBlAGYAAQAA\
+AAEACAABABQASQABAAAAAQAIAAEABgBSAAIAAQCzALsAAAAGAAAAAQAIAAMAAQASAAEAKgAA\
+AAEAAAAIAAIAAQD8AQQAAAABAAAAAQAIAAEABv/3AAIAAQEFAQ0AAAAEAAAAAQAIAAEANgAB\
+AAgABQAMABQAHAAiACgA+gADANsA3gD7AAMA2wDfAPcAAgDbAPgAAgDeAPkAAgDfAAEAAQDb\
+AAEAAAABAAgAAQAGABsAAQABAKA=') format('woff');
   font-style: normal;
   font-weight: normal;
-  unicode-range: U+20,U+30-39,U+41-5A,U+5F,U+61-69,U+6C-7A,U+621-63A,U+640-64A,U+660-661,U+669;
+  unicode-range: U+20,U+30-36,U+38-39,U+41-49,U+4B-5A,U+5F,U+61-69,U+6C-7A,U+621-63A,U+640-64A,U+660-661,U+669;
```
indirectlylit, line 627: @rtibbles my guess is that it's the unicode range update which triggered such a big diff in the css-embedded font, but still not completely sure.

One thing that's interesting is the unicode range is fairly sensitive to the _exact_ characters used. Here for example it changes `U+30-39` to `U+30-36,U+38-39`, essentially removing the number `7` from the subset font... if we wanted to reduce churn, we could add some "smoothing" to the subset logic to make it a bit less precise but more stable.

indirectlylit (commented):
looked through the diff and looks good!

kolibri/plugins/setup_wizard/assets/src/views/importLODUsers/SelectFacilityForm.vue
rtibbles: Yeah, there's a difference between "DO NOT TRANSLATE THE WORDS INSIDE {}" and "DO NOT TRANSLATE" full stop. I guess if the nuance of the first is being lost, that's why the second has been opted for?

kolibri/core/static/assets/fonts/noto-subset.ar.css
rtibbles: I don't think it's particularly worrissome - although it would be nice to have a content hash of these CSS files in the filename, as we've discussed previously!

kolibri/plugins/setup_wizard/assets/src/views/importLODUsers/SelectFacilityForm.vue
indirectlylit: "DO NOT TRANSLATE THE WORDS INSIDE {}" is true for most strings in crowdin

rtibbles: Except for the words inside {} inside {}!

PR #8327, opened 2021-08-18
Epub: when relocating to the last page, ensure we put the slider at 100

### LINKED ISSUE #8309: Investigate potential bug with progress/completion
There may be a bug in develop with progress/completion. It seems like there are mismatches between the progress that is displayed in the UI and the progress that is calculated, and that some resources cannot be successfully "completed". So far, it seems to be occurring in the epub content renderer, and possibly with videos. 

The purpose of this issue is to investigate and see if the issue is reproducible, and investigate where the problem might be happening, to create a detailed follow up issue.

image

nucleogenesis:
## Summary

Previously we always used the `location.start.percentage` value in the Epub renderer to set the slider's percentage. This ensures that when we are moving to a location whose `location.end.percentage` is 100, we set the slider value to 100 (completing the resource).

## References

Fixes #8309 

## Reviewer guidance

Does it fix it?

----

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/plugins/epub_viewer/assets/src/views/EpubRendererIndex.vue
+++ b/kolibri/plugins/epub_viewer/assets/src/views/EpubRendererIndex.vue
@@ -382,6 +382,7 @@
     created() {
       // Try to load the appropriate directional CSS for the particular content
       this.cssPromise = this.$options.contentModule.loadDirectionalCSS(this.contentDirection);
+      this.visitedPages = this.savedVisitedPages || {};
     },
     beforeMount() {
       global.ePub = Epub;
@@ -485,7 +486,7 @@
             // update progress using number of pages seen out of available pages
             this.$emit(
               'updateProgress',
-              Object.keys(this.visitedPages).length / this.locations.length
+              Object.keys(this.visitedPages || {}).length / this.locations.length
             );
           }
         }
@@ -704,7 +705,7 @@
         this.currentSection = this.getCurrentSection(currentLocationStart);
       },
       relocatedHandler(location) {
-        console.log(location);
+        //console.log(location);
         // Ensures that when we're on the last page, we set the slider value to 100
         // otherwise, we show the slider % using the start
         if (location.atEnd) {
--- a/kolibri/plugins/epub_viewer/assets/src/views/EpubRendererIndex.vue
+++ b/kolibri/plugins/epub_viewer/assets/src/views/EpubRendererIndex.vue
@@ -704,10 +704,11 @@
         this.currentSection = this.getCurrentSection(currentLocationStart);
       },
       relocatedHandler(location) {
+        console.log(location);
         // Ensures that when we're on the last page, we set the slider value to 100
         // otherwise, we show the slider % using the start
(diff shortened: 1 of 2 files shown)


rtibbles:
Flagging that resource completion is based on 'visiting every page' now, and not on the slider location being at 100 percent. So while this might be a useful fix, I'm not sure it will fix the underlying issue.

kolibri/plugins/epub_viewer/assets/src/views/EpubRendererIndex.vue
```diff
@@ -704,7 +704,13 @@
         this.currentSection = this.getCurrentSection(currentLocationStart);
       },
       relocatedHandler(location) {
-        this.sliderValue = location.start.percentage * 100;
+        // Ensures that when we're on the last page, we set the slider value to 100
+        // otherwise, we show the slider % using the start
+        if (location.end.percentage === 100) {
```
rtibbles: Should this be `== 100` or `== 1` - I am guessing based on the fact that we multiply it by `100` below that the `location.end.percentage` value is actually a fraction between 0 and 1?

nucleogenesis: Good catch. After looking at what I get in the `location` object, apparently it puts a `atEnd: true` bool flag on it when it's the last page, so I refactored to use that.

nucleogenesis:
> Flagging that resource completion is based on 'visiting every page' now, and not on the slider location being at 100 percent. So while this might be a useful fix, I'm not sure it will fix the underlying issue.

Figured it out :)

marcellamaki:
Hi @nucleogenesis - so I do think what you've described this fix as doing is accurate and working. My question is kind of the same as Richard's which is.... is there maybe something else going on with epub progress that deserves more investigation? I don't think there's any downside to merging this, but I did some more "realistic" manual QA that I usually do and actually read through an epub (and I picked one in French, and my French is terrible, so that it would be an authentic learning experience and not me skimming and clicking through 🙃). Despite actually fully engaging with it _twice_, it still shows as "in progress" and every time I open it, it comes up to 60% (and the corresponding page.) Again - this is sort of outside what you have described fixing here, but it does seem like we should attend to it. Maybe I can create a follow up issue for this, and add a bit more context?

rtibbles:
I think we can merge this, but not mark the issue as fixed, seems easiest.

marcellamaki (approved):

pcenov:
Hi everyone, today I had a look at this [build](https://buildkite.com/learningequality/kolibri-windows/builds/1773) and just want to add that:
1. The slider goes to 100% even before reaching the last page which is confusing from an end-user point of view.
2. Out of a number of epub resources that I have attempted to complete only one was marked as completed (Tinisiima-The_Ungrateful_Man_ePub_Document) but that happened while I was at about 80% of the book, which is again quite confusing. I couldn't achieve getting any of the other epub files to be marked as completed.




@radinamatic

PR #8333, opened 2021-08-19
Server sync queue

rtibbles:
## Summary
* Adds test to the sync queue endpoint to guarantee behaviour
* Consolidates logic so that behaviour is unified between the 'create' and 'update' flows
* Scales the `keep_alive` and `sync_interval` values based on how busy the server is, but with maximum values so that there is no runaway interval
* Ensures that the next person to queue is the one that is first in the queue, not just the first person to ask after the last sync finished
* Requires the user parameter to be added on both create and update, and confirms that the user in question is the user in the queue
* Adds more verbose logging to client side utilities
* Updates client side utilities to match new API endpoint signatures
* Ignores errored transfersessions when working out how many active syncs there are!

## Reviewer guidance
A bit hard to test without a few instances running with this code, but I think this makes the queue work as expected. The only slight issue might be that it will slow syncing down slightly because of the stricter enforcement of the queue order, but I don't think it will make it any slower than one person hogging the queue!

----

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/core/public/api.py
+++ b/kolibri/core/public/api.py
@@ -251,11 +251,13 @@ class SyncQueueViewSet(viewsets.ViewSet):
         else:
             # polling time at least HANDSHAKING_TIME seconds per position in the queue to
             # be greater than the time needed for the handshake part of the ssl protocol
+            # we add one to the zero based position, as if the position is zero and it
+            # got to here, it means the sync queue is currently full, so we need to wait
             # we make sure that it is never less than half of the stale queue time, to
             # prevent a malignant loop whereby very long queues cause clients to lose
             # their place in the queue by only polling again after their queue has
             # already expired!
-            polling = min(HANDSHAKING_TIME * pos, STALE_QUEUE_TIME / 2)
+            polling = min(HANDSHAKING_TIME * (pos + 1), STALE_QUEUE_TIME / 2)
             data = {
                 "action": QUEUED,
                 "keep_alive": polling,
--- a/kolibri/core/public/api.py
+++ b/kolibri/core/public/api.py
@@ -215,7 +215,7 @@ class SyncQueueViewSet(viewsets.ViewSet):
         # Make sure it is no longer than half the time by which we measure 'recently synced'
         # to make sure we are never letting syncing drift too far.
         sync_interval = min(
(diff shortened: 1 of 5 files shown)


kolibri/core/public/api.py
```diff
@@ -188,86 +188,101 @@ def list(self, request):
...
+        SyncQueue.clean_stale(expire=STALE_QUEUE_TIME)
+
+        # Calculate the total size of the queue to scale things
+        total_queue_size = SyncQueue.objects.count()
+
+        # Scale the sync_interval we send to clients based on the total number of
+        # queued single user sync requests.
+        # Make sure it is no longer than half the time by which we measure 'recently synced'
+        # to make sure we are never letting syncing drift too far.
+        sync_interval = min(
```
jredrejo, line 217: if the queue is empty this will return 0 as a sync interval, so further resyncs won't happen

jredrejo (changes requested):

kolibri/core/public/api.py
```diff
@@ -188,86 +188,101 @@ def list(self, request):
...
+            OPTIONS["Deployment"]["SYNC_INTERVAL"] * total_queue_size, DELAYED_SYNC / 2
```
rtibbles:
```suggestion
            OPTIONS["Deployment"]["SYNC_INTERVAL"] * (total_queue_size + 1), DELAYED_SYNC / 2
```

kolibri/core/public/test/test_api.py
```diff
@@ -258,3 +274,425 @@ def test_public_filter_unlisted(self):
...
+            )
+        response = self.client.put(
+            reverse("kolibri:core:syncqueue-detail", kwargs={"pk": queue.id}),
+            data={"user": self.learner.id},
+            format="json",
+        )
+        data = response.json()
+        self.assertEqual(response.status_code, 200)
+        self.assertEqual(
+            data["sync_interval"], OPTIONS["Deployment"]["SYNC_INTERVAL"] * 6
```
rtibbles:
```suggestion
            data["sync_interval"], OPTIONS["Deployment"]["SYNC_INTERVAL"] * 7
```

jredrejo (approved):

PR #8355, opened 2021-08-27
Add yarn license list checking to ensure no GPL only packages in our dependencies

### LINKED ISSUE #2321: Licenses of all bundled PyPI+NPM dependencies as overview
## Summary

This is something we should look at, but IMO it doesn't have to be now. We have no conflicts of licenses in our dependencies AFAIK, and there are no one immediately requesting an overview.

But it would be nice with an actual overview. The issue for instance arises in the debian package where I'd like to at least specify a list of relevant licenses.

It could be auto-generated and put in the docs by using the stuff that fetches PyPI license info:

https://github.com/learningequality/kolibri/blob/develop/test/conditional/test_licenses.sh

@indirectlylit were you also working on something for NPM?

rtibbles:
## Summary
* Adds checking of our JS dependencies to our license checking

## References
Fixes #2321

## Reviewer guidance
Does the regex look right? I tested it with these examples and it worked. matching only GPL by itself:
```
MIT OR GPL
GPL
GPL OR MIT
GPL or MIT
MIT or GPL
├─ (LGPL-2.0 or MIT)
```

----

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/test/conditional/test_licenses.sh
+++ b/test/conditional/test_licenses.sh
@@ -35,4 +35,11 @@ do
     fi
 done
 
+if yarn licenses list | grep -iP '(?<!OR )(?<!L)GPL(?! OR)'
+# If any output from the above, then we have found something, use && true to coerce to boolean
+then
+    echo "Problem! Incompatible license found in Javascript dependencies"
+    exit 14
+fi
+
 exit 0
--- a/.github/workflows/check_licenses.yml
+++ b/.github/workflows/check_licenses.yml
@@ -14,7 +14,7 @@ jobs:
         uses: fkirc/skip-duplicate-actions@master
         with:
           github_token: ${{ github.token }}
-          paths: '["requirements/*.txt"]'
+          paths: '["requirements/*.txt", "package.json", "**/package.json", "yarn.lock", "test/conditional/test_licenses.sh"]'
   licenses:
     name: Licenses check
     needs: pre_job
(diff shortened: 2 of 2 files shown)


test/conditional/test_licenses.sh
```diff
@@ -35,4 +35,13 @@ do
     fi
 done
 
+yarnlicense = `yarn licenses list | grep -iP '(?<!OR )(?<!L)GPL(?! OR)'`
+
+if yarnlicense && true
```
indirectlylit: what's the effect of `&& true`? maybe add a comment

indirectlylit (commented):
> Fixes #2321

Deep cut!

Looks like the test is being skipped. Why is it conditional?

image

test/conditional/test_licenses.sh
rtibbles: I _think_ it's coercing to a boolean - I copied it from above.

rtibbles:
> Why is it conditional?

It's conditional on the dependencies changing - should perhaps also be conditional on the bash file itself changing!

rtibbles:
Updated - also noticed it wasn't conditional on JS dependencies changing, so have updated that too.

test/conditional/test_licenses.sh
bjester: I think this is slightly different than the source where you copied it from, since the above `&& true` is apart of the `if` statement which won't trigger the script halt on a non-zero exit code with `set -e`. So if it were to fail, it wouldn't go any farther than line 38 because that isn't in the `if` test and `set -e`. You should get the following error if it actual goes past line 38 as it is:
```
./test/conditional/test_licenses.sh: line 41: yarnlicense: command not found
```

rtibbles: Yeah, I added spaces between `yarnlicenses=`... which made it interpret it not as a variable assignment, should be updated now.

bjester: @rtibbles Note that the error I shared is line 41, not line 38. When `yarn licenses list | grep -iP '(?<!OR )(?<!L)GPL(?! OR)'` is executed, if it doesn't find a match it will exit with code `1` which stops the script at line 38 and exits with the same code:
```
$ ./test/conditional/test_licenses.sh
Checking all requirements installed with pip, except docutils;nose;...
$ echo $?
1
```
If I substitute `GPL` in the regex for `MIT`, then it should match, which allows it to get to line 41, and thus:
```
$ ./test/conditional/test_licenses.sh
Checking all requirements installed with pip, except docutils;nose;...
./test/conditional/test_licenses.sh: line 41: yarnlicense: command not found
```

rtibbles: Ah, gotcha - yeah, that's what is causing the exit code 1 then!

rtibbles: I finally trial and errored my way to success.

Ran the script locally using MIT instead of GPL to confirm it does indeed give a non-zero exit code in that case.

bjester (approved):
:100:

PR #8367, opened 2021-08-31
Ensure the user will see the welcome modal after login

### LINKED ISSUE #8308: Clean up modal appearance and text in single user sync setup wizard flow
> I find this screen a bit confusing, as 3 different terms (import, load, sync) seem to be used for what is essentially the same action.

 I'm not sure if the implementation in this PR is using the same components as the task manager in `Device > Channels`, but we can remove almost all the lines except `Finished` and `'learner' from 'ZukVillage' successfully loaded to this device` to be consistent with both the designs and the current 0.14 implementation of facility import. I'd argue all those other details aren't that useful to know at this stage in the setup process. That way we'd at least remove the word 'Sync'. I think there was a reason we went "Loaded/Loading" instead of import but I can't quite remember or find any notes on why...

importuser

---

> Once the setup is completed, the first imported learner is automatically signed in, but on the Learn page, the phrase As an administrator, you can import channels looks out of place and confusing. Administrator on a learn-only-device?

> Actually, I'd like to suggest the change of the wording for this link to something more more straightforward, like Import channels on your device

Agree it should change, and also the action or statement here is conditional depending on who's logged in. Here's my suggestion, comparing what it is for super admins on 0.14 and proposal for learners on after this setup

Screen Shot 2021-08-05 at 12 47 08 PM

jredrejo:
## Summary

According to the proposed design, when finishing the setup wizard on a Learning Only Device, the modal window to recommend import channels will appear _after_ the user logins

## References
Closes: #8308 

## Reviewer guidance
Do a clean installation importing some user and check after the user logins, he's redirected to the device tab and sees the modal window
prueba

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/plugins/learn/assets/src/views/ContentUnavailablePage.vue
+++ b/kolibri/plugins/learn/assets/src/views/ContentUnavailablePage.vue
@@ -17,6 +17,8 @@
 
   import { mapGetters } from 'vuex';
   import urls from 'kolibri.urls';
+  import redirectBrowser from 'kolibri.utils.redirectBrowser';
+  import plugin_data from 'plugin_data';
 
   export default {
     name: 'ContentUnavailablePage',
@@ -25,6 +27,7 @@
         title: this.$tr('documentTitle'),
       };
     },
+
     computed: {
       ...mapGetters(['canManageContent', 'isLearner']),
       deviceContentUrl() {
@@ -38,6 +41,12 @@
       showLearnerText() {
         return this.isLearner && !this.canManageContent;
       },
+    },
+    beforeMount() {
+      if (plugin_data.isSubsetOfUsersDevice) {
+        const deviceContentUrl = urls['kolibri:kolibri.plugins.device:device_management'];
+        redirectBrowser(deviceContentUrl());
+      }
     },
     $trs: {
       header: {
--- a/kolibri/plugins/learn/assets/src/views/ContentUnavailablePage.vue
+++ b/kolibri/plugins/learn/assets/src/views/ContentUnavailablePage.vue
@@ -17,8 +17,6 @@
 
   import { mapGetters } from 'vuex';
   import urls from 'kolibri.urls';
-  import redirectBrowser from 'kolibri.utils.redirectBrowser';
-  import plugin_data from 'plugin_data';
 
   export default {
     name: 'ContentUnavailablePage',
@@ -41,12 +39,6 @@
       showLearnerText() {
         return this.isLearner && !this.canManageContent;
       },
-    },
-    beforeMount() {
-      if (plugin_data.isSubsetOfUsersDevice) {
-        const deviceContentUrl = urls['kolibri:kolibri.plugins.device:device_management'];
-        redirectBrowser(deviceContentUrl());
-      }
     },
     $trs: {
       header: {
--- a/kolibri/plugins/device/assets/src/views/DeviceIndex.vue
+++ b/kolibri/plugins/device/assets/src/views/DeviceIndex.vue
@@ -72,7 +72,10 @@
       ...mapState({ welcomeModalVisibleState: 'welcomeModalVisible' }),
       ...mapState('coreBase', ['appBarTitle']),
       welcomeModalVisible() {
-        return this.welcomeModalVisibleState && !window.sessionStorage.getItem(welcomeDimissalKey);
+        return (
+          this.welcomeModalVisibleState &&
+          window.sessionStorage.getItem(welcomeDimissalKey) !== 'true'
+        );
       },
       pageName() {
         return this.$route.name;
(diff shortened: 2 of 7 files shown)


jtamiace (approved):

kolibri/plugins/learn/assets/src/views/ContentUnavailablePage.vue
```diff
@@ -39,6 +42,12 @@
         return this.isLearner && !this.canManageContent;
       },
     },
+    beforeMount() {
+      if (plugin_data.isSubsetOfUsersDevice) {
+        const deviceContentUrl = urls['kolibri:kolibri.plugins.device:device_management'];
+        redirectBrowser(deviceContentUrl());
```
rtibbles: We should add two conditionals here:
```
if (this.canManageContent && deviceContentUrl) {
```

No point redirecting them if the user does not have content management permissions, and if the plugin is disabled, then the `deviceContentUrl` will be undefined.

rtibbles: Also, is this required? We redirect a user to the device page after initial setup, and we already have a URL for them to be navigate there.

rtibbles (changes requested):
I think we don't need the automatic redirect from the content unavailable page. I think the proposal in the issue was to update the text there to make it more straight forward, I think that would be the better change here.

kolibri/plugins/learn/assets/src/views/ContentUnavailablePage.vue
jredrejo: Honestly, I'm not satisfied with this solution either, but it was the only way I found to do it for the case where the setup wizard uses an admin account to sync the users.
In that case, after the wizard finishes, it shows the login screen (the wizard can't sign in with the first synced user because it does not know the user password), and after login the user is redirected to learn, not to device. 


As you said, the user is redirected to device after initial setup... if setup syncing is done by the users not but the admin.

jredrejo:
> I think we don't need the automatic redirect from the content unavailable page. I think the proposal in the issue was to update the text there to make it more straight forward, I think that would be the better change here.

I'd be very happy to remove it, please, read my above comment and let's confirm it

kolibri/plugins/learn/assets/src/views/ContentUnavailablePage.vue
indirectlylit: Confirmed that this PR shows the modal @jredrejo describes for learners that were imported using an admin account.

Seems reasonable to me to keep this with the introduction of the extra conditions, and preferably also adding a comment explaining why this is necessary

rtibbles: Yeah, I think I can live with this with the extra conditions and the comment - the explanation for why this is necessary helps!

jredrejo: There's another option, @jtamiace I'd like to know your opinion: the main problem comes from the fact that using the admin account to import users we don't know the users' passwords. When importing finishes the wizard logouts the admin user so any user can login and will be redirected to learn.

If the wizard does not logs out the admin, after the wizard finishes kolibri will be open under the admin account in the device section and all the code from `ContentUnavailablePage.vue `can be removed, it's not needed as first view of kolibri will be automatically in the device page . If being logged as an admin user after the wizard finishes it is a good option I can remove several lines of code because the wizard will finish always with the same behaviour, no matter who did the sync.

rtibbles: I am guessing that the automated call to logout might be at the root of this issue too: https://github.com/learningequality/kolibri/issues/8379 - so simplifying that code might be a good idea overall!

jtamiace: > If the wizard does not logs out the admin, after the wizard finishes kolibri will be open under the admin account in the device section and all the code from ContentUnavailablePage.vue can be removed,

@jredrejo streamlining the experience so the admin goes straight to Device is helpful, but would we still need to keep it in, in case they decide to explore other places of the app before they start importing content? Unless content import is the only thing a user can do post-setup, might be good to keep `ContentUnavailablePage.vue` in.

jredrejo:
> I think we don't need the automatic redirect from the content unavailable page. I think the proposal in the issue was to update the text there to make it more straight forward, I think that would be the better change here.

Ok, I've removed the redirection in the content unavailable page and set the wizard to do the same, no matter who's done it: an admin user or a normal user. So, if an admin does the sync he will stay logged and will be redirected to the device/channels page and see the model after finishing the import and will see exactly the same a normal user, i.e. the animated gif in this PR description.

rtibbles (approved):

PR #8370, opened 2021-08-31
Restart zeroconf after setup

### LINKED ISSUE #8344: Provisioning learn-only device in setup wizard doesn't resync until restart
## Observed behavior

1. Start a fresh instance of Kolibri
2. Go through the Advanced setup wizard, selecting "learn-only" device
3. Authenticate with the remote server using the learner credentials
4. Finish setup
5. Observe the Kolibri instance does not resync with the server
6. Restart Kolibri
7. Observe the Kolibri instance does resync with the server

## Errors and logs

N/A

## Expected behavior

Kolibri always makes recurring requests to sync with the server when provisioned as a SoUD

## User-facing consequences

Broken expectation that SoUD make recurring syncs with the server

## Context

Latest Kolibri `release-v0.15.x`
Ubuntu 20.04
Chrome

jredrejo:
## Summary

Create a new endpoint + an api resource to restart zerconf when the setup wizard ends after syncing one or more users in a SoUD

## References
Closes: #8344 
Depends on #8367 been merged

## Reviewer guidance
Create a new Kolibri device provisioning it importing users from another device.
Before this PR users were not synced unless the server is restarted. After this pr, syncing processes will start as soon as the setup is finished

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/plugins/setup_wizard/api.py
+++ b/kolibri/plugins/setup_wizard/api.py
@@ -173,3 +173,26 @@ class SetupWizardSoUDTaskView(TasksViewSet):
     """
 
     permission_classes = [HasPermissionDuringSetup | HasPermissionDuringLODSetup]
+
+
+class SetupWizardRestartZeroconf(ViewSet):
+    """
+    An utility endpoint to restart zeroconf after setup is finished
+    in case this is a SoUD
+    """
+
+    permission_classes = [HasPermissionDuringSetup | HasPermissionDuringLODSetup]
+
+    @decorators.action(methods=["post"], detail=False)
+    def restart(self, request):
+        import logging
+        from kolibri.core.discovery.utils.network.search import (
+            unregister_zeroconf_service,
+            initialize_zeroconf_listener,
+        )
+
+        logger = logging.getLogger(__name__)
+        unregister_zeroconf_service()
+        initialize_zeroconf_listener()
+        logger.info("Zeroconf has reinitialized")
+        return Response({})
--- a/kolibri/plugins/setup_wizard/assets/src/api.js
+++ b/kolibri/plugins/setup_wizard/assets/src/api.js
@@ -64,15 +64,12 @@ export const SetupSoUDTasksResource = new Resource({
 export const FinishSoUDSyncingResource = new Resource({
   name: 'restartzeroconf',
   namespace: 'kolibri.plugins.setup_wizard',
-  finish(store, admin) {
+  finish() {
     const welcomeDimissalKey = 'DEVICE_WELCOME_MODAL_DISMISSED';
     const device_url = urls['kolibri:kolibri.plugins.device:device_management'];
     window.sessionStorage.setItem(welcomeDimissalKey, false);
     this.postListEndpoint('restart');
-    if (admin) store.dispatch('kolibriLogout');
-    else {
-      redirectBrowser(device_url());
-    }
+    redirectBrowser(device_url());
     return '';
   },
 });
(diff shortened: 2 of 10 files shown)


indirectlylit:
will test once #8367 is resolved

kolibri/plugins/setup_wizard/assets/src/api.js
```diff
@@ -58,3 +60,16 @@ export const SetupSoUDTasksResource = new Resource({
...
+
+export const FinishSoUDSyncingResource = new Resource({
+  name: 'restartzeroconf',
+  namespace: 'kolibri.plugins.setup_wizard',
+  finish() {
+    const welcomeDimissalKey = 'DEVICE_WELCOME_MODAL_DISMISSED';
+    const device_url = urls['kolibri:kolibri.plugins.device:device_management'];
+    window.sessionStorage.setItem(welcomeDimissalKey, false);
+    this.postListEndpoint('restart');
+    redirectBrowser(device_url());
```
rtibbles: In case `device_url` is `undefined`, we should change this to:
```suggestion
    redirectBrowser(device_url ? device_url(): null);
```

jredrejo: @rtibbles  maybe this is a naive question, but how could `device_url` be null? are there any case where we are not loading the `device` plugin?

rtibbles: If a user has disabled the device plugin then this would be null. In general, I have preferred to enforce any coupling between our plugins to be loose, so that one does not cause errors because another is disabled.

jredrejo: ok, I never thought this would be an use case but, as usual, you're right. Done.

But... another naive question, shouldn't we add this guard to https://github.com/learningequality/kolibri/blob/release-v0.15.x/kolibri/core/assets/src/core-app/urls.js#L14 ? or making it fail is being used in the code?

rtibbles: I would like to add a guard that just returned a function for every URL, so that we could always invoke the function, and just return defined in that case, but the best way to achieve that would be with an [ES6 Proxy](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy) object, which is not available on IE11 and not polyfillable.

The function you are pointing to there is only used by the URL functions defined in our Javascript, rather than the JS generated by Django-js-reverse (we could in theory hack that template to do this, but it felt like a lot of work!).

kolibri/plugins/setup_wizard/api.py
```diff
@@ -173,3 +173,26 @@ class SetupWizardSoUDTaskView(TasksViewSet):
...
+    """
+    An utility endpoint to restart zeroconf after setup is finished
+    in case this is a SoUD
+    """
+
+    permission_classes = [HasPermissionDuringSetup | HasPermissionDuringLODSetup]
+
+    @decorators.action(methods=["post"], detail=False)
+    def restart(self, request):
+        import logging
```
rtibbles, line 188: Any reason not to do these imports and logging configuration in module scope?

rtibbles (commented):
The code changes here look good, one query, but nothing that blocks merging.

Have not tested manually.

kolibri/plugins/setup_wizard/api.py
jredrejo: just to avoid doing the import if the wizard does the other two kinds of setup instead of the SoUD provisioning

indirectlylit (approved):
confirmed that after this change, the devices begin auto-syncing after import

PR #8383, opened 2021-09-02
filter SoUD devices when scanning the network to import new facilities

### LINKED ISSUE #8276: Single user syncing - Getting a 403 error when attempting to Sync with a learner-only device as a Super admin
## Observed behavior
As a super admin I have the ability to manually sync facilities. However when I attempt to sync with a learner-only device I'm getting a 403 error without any explicit information what is going on.

## Expected behavior
The expected behavior is not clear and should be further discussed.

1. Install this [build](https://buildkite.com/learningequality/kolibri-windows/builds/1586#97b40f19-5fb4-4cd7-888e-69531fedca75) on 2 devices, set up an admin device and import a learn-only device.
2. As a super admin go to Device>Facilities and click the Sync button next to your facility
3. Select the learner-only device and attempt to sync with it (note that syncing with a fully imported facility is functioning correctly)

## Logs
[logs.zip](https://github.com/learningequality/kolibri/files/6975925/logs.zip)

## Usage Details
 - OS: Windows 10, Ubuntu
 - Browser: Chrome

jredrejo:
## Summary
When a superuser tries to import new facilities, subsets of users devices should not be shown nor found in the list of available devices to look for new facilites

## References
Closes: #8276 

## Reviewer guidance
1.  Setup a subset of users device (a.k.a. learn only device) a
2. Setup another kolibri server, both in your network

Before this PR:

- superuser of the kolibri server will see the SoUD when scanning facilities to import

After this PR:

 - superuser of the kolibri server will NOT see the SoUD when scanning facilities to import

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/core/assets/src/views/sync/SelectAddressModalGroup/api.js
+++ b/kolibri/core/assets/src/views/sync/SelectAddressModalGroup/api.js
@@ -48,26 +48,28 @@ function fetchAddressesForLOD(LocationResource = NetworkLocationResource) {
 }
 
 function fetchAddressesWithChannel(withChannelId = '', LocationResource = NetworkLocationResource) {
-  return LocationResource.fetchCollection({ force: true }).then(locations => {
-    // If channelId is provided, then we are in an 'import-more' workflow and disable
-    // locations that do not have the channel we are importing from.
-    if (withChannelId !== '') {
-      const locationsWithAvailabilityPromises = locations.map(location => {
-        // Need to wrap in normal promise, otherwise Promise.all will cause some of these
-        // to resolve as undefined
-        return new Promise(resolve => {
-          return channelIsAvailableAtLocation(withChannelId, location).then(isAvailable => {
-            resolve({ ...location, hasContent: isAvailable });
+  return LocationResource.fetchCollection({ force: true, getParams: { filterSoUD: true } }).then(
+    locations => {
+      // If channelId is provided, then we are in an 'import-more' workflow and disable
+      // locations that do not have the channel we are importing from.
+      if (withChannelId !== '') {
+        const locationsWithAvailabilityPromises = locations.map(location => {
+          // Need to wrap in normal promise, otherwise Promise.all will cause some of these
+          // to resolve as undefined
+          return new Promise(resolve => {
+            return channelIsAvailableAtLocation(withChannelId, location).then(isAvailable => {
+              resolve({ ...location, hasContent: isAvailable });
+            });
           });
         });
-      });
-      return Promise.all(locationsWithAvailabilityPromises);
-    }
+        return Promise.all(locationsWithAvailabilityPromises);
+      }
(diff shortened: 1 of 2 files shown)


kolibri/core/assets/src/views/sync/SelectAddressModalGroup/api.js
```diff
@@ -48,26 +48,28 @@ function fetchAddressesForLOD(LocationResource = NetworkLocationResource) {
 }
 
 function fetchAddressesWithChannel(withChannelId = '', LocationResource = NetworkLocationResource) {
-  return LocationResource.fetchCollection({ force: true }).then(locations => {
```
rtibbles: Do we need to do this filtering for channel import? It's only facility syncing that is problematic here.

kolibri/core/discovery/api.py
```diff
@@ -16,6 +16,16 @@ class NetworkLocationViewSet(viewsets.ModelViewSet):
     serializer_class = NetworkLocationSerializer
     queryset = NetworkLocation.objects.all()
 
+    def filter_queryset(self, queryset):
```
rtibbles: Could also use a FilterSet here with just `subset_of_users_device` as the only field.

rtibbles (commented):
A couple of questions.

kolibri/core/assets/src/views/sync/SelectAddressModalGroup/api.js
jredrejo: I thought we would prefer  not to use them for any sync work, but if you think it's fine to import from them, no problem. I've removed it

kolibri/core/discovery/api.py
jredrejo: oh, I should have thought of that, much simpler, of course. Done

kolibri/core/assets/src/views/sync/SelectAddressModalGroup/api.js
rtibbles: Yeah, I just think of situations where there are only other SoUD devices on a network but you still want to get content peer to peer.

kolibri/core/discovery/api.py
```diff
@@ -15,6 +16,10 @@ class NetworkLocationViewSet(viewsets.ModelViewSet):
     permission_classes = [NetworkLocationPermissions | NotProvisionedHasPermission]
     serializer_class = NetworkLocationSerializer
     queryset = NetworkLocation.objects.all()
+    filter_backends = [DjangoFilterBackend]
+    filterset_fields = [
```
rtibbles, line 20: And I had forgotten that for fields only filtering you can just do this instead of even defining a `FilterSet`!

rtibbles (approved):

PR #8385, opened 2021-09-02
Do not create accounts in Subset of users devices

### LINKED ISSUE #8274: Single User Syncing - A self created user's device status is always "Not connected to server"
## Observed behavior
After importing a learn-only device I can create an account for another learner user which is never synced.

## Expected behavior
The expected behavior should be further discussed as either the option to create accounts for learn-only devices should be disabled or there should be an option to sync such users.

1. Install this [build](https://buildkite.com/learningequality/kolibri-windows/builds/1586#97b40f19-5fb4-4cd7-888e-69531fedca75) and import a learn-only device.
2. Sign out with the imported learner
3. Create a new account by selecting the "Create an account" option
4. Click the username of the user to the top right corner to check the Device status

## Usage Details
 - OS: Windows 10
 - Browser: Chrome

jredrejo:
## Summary
In a subset of users device (aka Learner only device) creation of new accounts must not be allowed, no matter the facility settings.

## References
Closes: #8274 

## Reviewer guidance
Setup a subset of users device (a.k.a. learn only device) from a kolibri server facility that allows creation of new accounts by the uers

Before this PR:

- Sign in screen shows the "Create account" button

After this PR:

- Sign in screen does not show  the "Create account" button

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/plugins/user/assets/src/routes.js
+++ b/kolibri/plugins/user/assets/src/routes.js
@@ -10,6 +10,7 @@ import ProfileEditPage from './views/ProfileEditPage';
 import SignInPage from './views/SignInPage';
 import SignUpPage from './views/SignUpPage';
 import NewPasswordPage from './views/SignInPage/NewPasswordPage';
+import plugin_data from 'plugin_data';
 
 router.beforeEach((to, from, next) => {
   const profileRoutes = [ComponentMap.PROFILE, ComponentMap.PROFILE_EDIT];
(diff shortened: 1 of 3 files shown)


indirectlylit (commented):
cc @rtibbles I'm surprised by this decision.

If a facility allows users to create new accounts, I would have expected that they could do this from either the main server or an individual-user device. This would better support parents creating accounts for their children at home, for example.

If we're disabling this to save time, maybe we should open a follow-up issue to re-enable it later?

rtibbles:
> If we're disabling this to save time, maybe we should open a follow-up issue to re-enable it later?

This is blocked because it's a technical limitation - the user account has to originate from the full facility.

In order to allow this to work we'd have to have a more complex technical workflow behind the scenes that somewhat replicates what we have done for the 'account merging' for migrating 0.14 devices to learner only devices.

Happy for us to open a follow up issue, but as I noted in the issue being fixed here, this is the practicable short term fix for 0.15.0.

rtibbles (approved):

rtibbles:
Follow up issue here: https://github.com/learningequality/kolibri/issues/8400

indirectlylit:
sounds good!

PR #8314, opened 2021-08-17
Implementation of new LearnImmersiveLayout parent component for Content Renderer

marcellamaki:
## Summary

This PR creates a new `LearnImmersiveLayout` component that is based on CoreBase and is rendered instead of `CoreBase` for all pages that use the content renderer. This decision was made after discussion in Slack (with JB, @MisRob and @nucleogenesis) and based on the timeline for the hybrid learning work. 

The layout component is the parent for the new toolbars, the simplified content renderer layout, and the side informational panel.

## References

Fixes #8234 

### Figma references: 
Screen Shot 2021-08-17 at 12 34 32 PM

**Note:** The content renderer display was updated in the implementation following [a conversation on Slack that determined that the renderer should fill the viewport](https://learningequality.slack.com/archives/C023DNC5GCX/p1628543607006500)

### Implementation: 

**Accessing channel content**
content-renderer-channels

**Accessing class content**
content-renderer-classes

**Accessing practice activity**
content-renderer-practice
…

### Notes:  
- Some elements of the content renderer layout still need to be adjusted according to the new designs, including the Assessment Wrapper for practice activities (spacing issues can be seen in gif), and adjustments as needed to include metadata such as the progress bar, estimated time, etc. around the content. I have considered this out-of-scope of the issue and focused just on creating the parent component, which was a blocker.
- The LearningMasteryBar (the "secondary header bar") is still a WIP, with Sairina working on a draft PR #8158. I have likewise considered me taking over and finishing that work out of scope for this issue. (I am happy to pick it up as needed as a followup, however)
- I have left some commented-out code related to progress tracking in the `ContentPage` that is not used now (and therefore was causing many, many linting errors), but can be repurposes for #8112. (Note also that there is an issue opened for #8309)

### Manual QA: 
- Access the content renderer as both a learner and a coach through channels and classes
- Both the Content Page and the Lesson Resource Viewer should have the new layout and be working correctly
- The interactions on these pages in the toolbar are not set up yet and are out of scope for this issue

#### Any potential areas of concern that need extra testing?
- The `ContentPage` component has been updated significantly. Most of this has been a _reduction_ in functionality, but I do want to be sure that nothing has been taken out that needs to remain. 
- My biggest concerns are: 1) that there are other places that the content renderer is used that also need to be updated that I somehow have just missed entirely (I almost missed updating `LESSON_RESOURCE_VIEWER` 😄)  and 2) that I have caused too much chaos in `LearnIndex.vue` by making these changes

----

## Testing checklist

[...]

PR #8395, opened 2021-09-03
Initial pass for estimated time label in content page

### LINKED ISSUE #8111: Display estimated time on resources
Overview:
Resources should have a display of the estimated time to completion, if the metadata exists

Done when:
- Estimated time displays for any resource that has this metadata 
- Explanatory tooltip displays on hover

Beyond scope of this ticket:
- Setting up estimated time metadata

[Figma reference](https://www.figma.com/file/uiZuCIqh2KYvBUfbCiJyyA/Hybrid-Learning?node-id=517%3A46959):
Screen Shot 2021-05-25 at 5 10 53 PM

nucleogenesis:
## Summary

Adds estimated time to ContentPage under the content.

Currently uses a placeholder value - will show "100 minutes" on every content item until the metadata is available.

Also, the display is hidden when `process.env.NODE_ENV === "production"` for now.

https://user-images.githubusercontent.com/6356129/132058556-dd5839be-17d8-4bbb-8fca-f62d368d6395.mp4

## References

Fixes #8111 

## Reviewer guidance

@jtamiace does this look okay? It's a little closer to the text than in the Figma but implements the default.

----

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/plugins/learn/assets/src/views/ContentPage.vue
+++ b/kolibri/plugins/learn/assets/src/views/ContentPage.vue
@@ -63,6 +63,22 @@
     </template>
     <KCircularLoader v-else />
 
+    <p v-if="estimatedTime">
+      <!--
+        TODO This icon should be updated ot an alarm clock style
+        clock per the figma linked in #8111
+      -->
+      <KLabeledIcon
+        ref="estimatedTime"
+        class="estimated-time"
+        icon="schedule"
+        :label="$tr('estimatedTime', { estimatedTime })"
+      />
+      <KTooltip reference="estimatedTime" :refs="$refs">
+        {{ $tr("estimatedTimeTooltip") }}
+      </KTooltip>
+    </p>
+
     <!-- TODO consolidate this metadata table with coach/lessons -->
     <!-- eslint-disable-next-line vue/no-v-html -->
     <p dir="auto" v-html="description"></p>
@@ -213,6 +229,17 @@
         extraFields: state => state.core.logging.summary.extra_fields,
         fullName: state => state.core.session.full_name,
       }),
+      estimatedTime() {
+        /*
+          TODO: This should return something like `this.content.estimatedTime`
+          once the metadata is made available
+        */
+        if (process.env.NODE_ENV !== 'production') {
+          return 100;
+        } else {
+          return null;
+        }
+      },
       isTopic() {
         return this.content.kind === ContentNodeKinds.TOPIC;
       },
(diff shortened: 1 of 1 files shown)


kolibri/plugins/learn/assets/src/views/ContentPage.vue
```diff
@@ -396,6 +429,15 @@
         message: 'Share',
         context: 'Option to share a specific file from a learning resource.',
       },
+      estimatedTime: {
+        message: '{ estimatedTime } minutes',
```
indirectlylit, line 433: "minutes" would need a singular form (edge case)

```suggestion
        message: 'About {estimatedTime, number, integer} {estimatedTime, plural, one {minute} other {minutes}}',
```

indirectlylit (commented):
These time estimates will be very inaccurate, and I feel like the learner would benefit from understanding that.

Two ideas:

1. We could hedge the estimate with the word 'about': "About 30 minutes"
2. For estimates >= 60 minutes, round to the nearest half hour: "About 1.5 hours"

kolibri/plugins/learn/assets/src/views/ContentPage.vue
rtibbles: Could we not use a `short` `time` type here? https://formatjs.io/docs/core-concepts/icu-syntax#time-type

nucleogenesis: Heck yah @rtibbles - I'm going to write some tests for this and will make use of that.

@jtamiace - thoughts on Devon's suggestion about hedging?

The tooltip kind of does that hedging itself by calling the time the "suggested time to complete". 

Maybe we could consider just putting the label as the string with the time instead of using a tooltip. It isn't obvious that there is a tooltip there (no `(i)` icon to initiate it). There is plenty of space there to have a full string rather than an icon-labled string.

nucleogenesis:
:man_facepalming: I did this on the `release-v0.15.x` branch and not `develop` which is where #8314 was merged.

Going to end up closing this PR, but will wait until next week so we can resolve the above questions as they're relevant anyway.

radinamatic:
> Maybe we could consider just putting the label as the string with the time instead of using a tooltip. It isn't obvious that there is a tooltip there (no (i) icon to initiate it). There is plenty of space there to have a full string rather than an icon-labled string.

Tooltips are generally PITA for accessibility, so I fully support fully visible string instead! 👍🏽

PR #8401, opened 2021-09-06
Remove use of python ua-parser from the backend.

rtibbles:
## Summary
* The Python `ua-parser` module is exhaustive, and as a result requires at least a thousand regexes.
* Currently it initializes them all at module import, which causes anywhere up to a 5 second delay in initializing Kolibri on a highly specced dev machine
* We could defer import until after initialization, but that would still mean a very lengthy initialization process at some point
* To avoid this, instead just do all user agent parsing in the frontend using the `ua-parser-js` library (which is far more performant than other user agent libraries), and which we were already partially using in vendored form
* Removes vendoring, and imports this library
* Uses this for all browser and os info inspection from the user agent
* Explicitly sends this info with session requests to be recorded in the user session logs
* Changes all session requests to use POST or PUT
* Sets 'active' on login POSTs so that user session logs are created
* Removes unneeded SessionResource

## References

Removes major initialization bottleneck, reducing startup time from ~5s to < 0.1s.

Follow up from #7658 

## Reviewer guidance
Does login and subsequent pings to the session endpoint work?
Does appropriate data get recorded in the UserSessionLog?
Is startup performance on Android improved?

----

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated


indirectlylit:
> ~5s to < 0.1s.

Wow, that's huge!

Will review and test later this week

> Follow up from #7658

Since that was merged into `develop` after 0.14, presumably the slow startup time didn't hit most users (except I suppose S2S?)

rtibbles:
> Since that was merged into develop after 0.14, presumably the slow startup time didn't hit most users (except I suppose S2S?)

Yes, and that's why I targeted it to 0.15, so we could prevent this issue even there.

jredrejo (approved):
Code looks good, nothing to comment. 
All the functionalities work well without regressions.
Tested it in linux (dev), android and windows 10. I haven't measured the improvement when starting but data logs are now better, at least detecting the Linux version.

Before:
before

After:
after

Issue #4585, opened 2018-12-11
After a successful backup, remove older backups


benjaoming:
### Observed behavior

In relation to #4578 and the growth in database storage footprint in addition to some historical problems of accumulated garbage data and redundant backup files, we should have a management command to clean up the database.

### Expected behavior

After a successful database backup, delete all other database backups bar last completed backup.

### User-facing consequences

Large disk-space footprint can easily be recovered. UI can follow.

### Errors and logs

n/a

### Steps to reproduce

n/a

### Context

#4578 

CC: @jredrejo @indirectlylit @rtibbles @ralphiee22

rtibbles:
The only time we trigger a database backup is when we upgrade, so it's not unreasonable to only maintain the most recent backup.

Once backup has completed successfully here: https://github.com/learningequality/kolibri/blob/release-v0.15.x/kolibri/utils/main.py#L77

We should then cleanup database backups that are not the most recent backup.

Can add tests for the behaviour in main in the tests for that module: https://github.com/learningequality/kolibri/blob/release-v0.15.x/kolibri/utils/tests/test_main.py

rtibbles:
Fixed in #8402

PR #8402, opened 2021-09-07
Remove older backups

AtKristijan:
## Summary
 Get all backups from current version.
 Remove older backups and leave the newest one

## References
 * https://github.com/learningequality/kolibri/issues/4585

----

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/core/deviceadmin/utils.py
+++ b/kolibri/core/deviceadmin/utils.py
@@ -47,11 +47,11 @@ def get_backup_files():
     """
     default_path = default_backup_folder()
     backups = os.listdir(default_path)
-    prefix = "db-v{}"
+    prefix = "db-v"
     backups = filter(lambda f: f.endswith(".dump"), backups)
     backups = filter(lambda f: f.startswith(prefix), backups)
     backups = list(backups)
-    backups.sort()
+    backups.sort(reverse=True)
     return backups
 
 def get_dtm_from_backup_name(fname):
--- a/kolibri/core/deviceadmin/utils.py
+++ b/kolibri/core/deviceadmin/utils.py
@@ -41,6 +41,19 @@ def default_backup_folder():
     return os.path.join(KOLIBRI_HOME, "backups")
 
 
+def get_backup_files(kolibri_version):
+    """
+    Returns all backups from current kolibri version.
+    """
+    default_path = default_backup_folder()
+    backups = os.listdir(default_path)
+    prefix = "db-v{}".format(kolibri_version)
+    backups = filter(lambda f: f.endswith(".dump"), backups)
+    backups = filter(lambda f: f.startswith(prefix), backups)
+    backups = list(backups)
+    backups.sort()
+    return backups
+
 def get_dtm_from_backup_name(fname):
     """
     Returns the date time string from our automated backup filenames
--- a/kolibri/core/deviceadmin/utils.py
+++ b/kolibri/core/deviceadmin/utils.py
@@ -41,13 +41,13 @@ def default_backup_folder():
     return os.path.join(KOLIBRI_HOME, "backups")
 
 
-def get_backup_files(kolibri_version):
+def get_backup_files():
     """
     Returns all backups from current kolibri version.
     """
     default_path = default_backup_folder()
     backups = os.listdir(default_path)
-    prefix = "db-v{}".format(kolibri_version)
+    prefix = "db-v{}"
     backups = filter(lambda f: f.endswith(".dump"), backups)
     backups = filter(lambda f: f.startswith(prefix), backups)
     backups = list(backups)
--- a/kolibri/core/deviceadmin/utils.py
+++ b/kolibri/core/deviceadmin/utils.py
@@ -54,6 +54,7 @@ def get_backup_files():
     backups.sort(reverse=True)
     return backups
 
+
 def get_dtm_from_backup_name(fname):
     """
     Returns the date time string from our automated backup filenames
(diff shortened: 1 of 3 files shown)


kolibri/utils/tests/test_main.py
```diff
@@ -70,6 +70,35 @@ def test_version_updated():
...
+    if not os.path.exists(default_path):
+        os.mkdir(default_path)
+
+    main.conditional_backup("0.14.0","0.15.1")
+    main.conditional_backup("0.14.0","0.15.2")
+    main.conditional_backup("0.10.1","0.15.3")
+    main.conditional_backup("0.10.2","0.15.4")
+    backups = get_backup_files(fallback_version)
+
+    for backup in backups[1:]:
```
rtibbles: I am not completely understanding the file removal here - maybe a code comment to explain how this is contributing to the test?

kolibri/core/deviceadmin/utils.py
```diff
@@ -41,6 +41,19 @@ def default_backup_folder():
     return os.path.join(KOLIBRI_HOME, "backups")
 
 
+def get_backup_files(kolibri_version):
```
rtibbles: It's possible that we have backups from many previous versions, and most backups will be from previous versions. So when we are cleaning up backups, filtering to one specific Kolibri version is unlikely to save us much space.

Instead we probably want to get all the backups that are _not_ from this version, as that's most likely the backup we just created.

kolibri/utils/main.py
```diff
@@ -78,6 +78,10 @@ def conditional_backup(kolibri_version, version_file_contents):
 
         try:
             backup = dbbackup(version_file_contents)
+            backups = get_backup_files(kolibri_version)
```
rtibbles: As I commented above, I think the logic here is inverted - we should be getting all the backup files not for this version, as that will be the older backups.

kolibri/core/deviceadmin/utils.py
```diff
@@ -41,6 +41,19 @@ def default_backup_folder():
...
+    """
+    Returns all backups from current kolibri version.
+    """
+    default_path = default_backup_folder()
+    backups = os.listdir(default_path)
+    prefix = "db-v{}".format(kolibri_version)
+    backups = filter(lambda f: f.endswith(".dump"), backups)
+    backups = filter(lambda f: f.startswith(prefix), backups)
+    backups = list(backups)
+    backups.sort()
```
rtibbles: We might want to sort this backwards - as the most recent version/datetime in the filename will be the latest alphabetically.

rtibbles (commented):
I am not completely sure the functionality is doing precisely what we want, have not manually tested yet.

kolibri/core/deviceadmin/utils.py
```diff
@@ -41,6 +41,19 @@ def default_backup_folder():
     return os.path.join(KOLIBRI_HOME, "backups")
 
 
+def get_backup_files():
+    """
+    Returns all backups from current kolibri version.
+    """
+    default_path = default_backup_folder()
+    backups = os.listdir(default_path)
+    prefix = "db-v{}"
```
rtibbles: We're not looking for any specific version so we can drop the string format here.
```suggestion
    prefix = "db-v"
```
I think this is the reason that the test is failing, because it's effectively filtering out all the DB backups, because none of them start with the string literal `db-v{}`.

kolibri/utils/tests/test_main.py
```diff
@@ -70,6 +72,50 @@ def test_version_updated():
...
+    backup = dbbackup("0.13.2")
+    backup = dbbackup("0.13.3")
+    backup = dbbackup("0.13.4")
+    backup = dbbackup("0.13.5")
+
+    # calling function for conditional backup
+    main.conditional_backup("0.11.1","0.15.3")
+
+    backups_after = os.listdir(default_path)
+    prefix = "db-v{}".format(fallback_version)
```
rtibbles: This test is not making a strong enough assertion - it is only asserting that files starting with `fallback_version` in their prefix have been deleted.
```suggestion
    prefix = "db-v"
```

kolibri/utils/tests/test_main.py
```diff
@@ -70,6 +72,50 @@ def test_version_updated():
...
+    backups_after = filter(lambda f: f.endswith(".dump"), backups_after)
+    backups_after = filter(lambda f: f.startswith(prefix), backups_after)
+    backups_after = list(backups_after)
+
+    # checking if delete is working properly in the conditional_backup
+    assert len(backups_after) == 1
+    assert ("db-v0.15.3" in backups_after[0]) == True
```
rtibbles: Can just assert the statement, rather than compare to `True`.
```suggestion
    assert "db-v0.15.3" in backups_after[0]
```

rtibbles (changes requested):
Needs some more cleanup, but the basic logic is sound.

kolibri/utils/tests/test_main.py
```diff
@@ -71,6 +74,50 @@ def test_version_updated():
...
+def test_conditional_backup():
+    import os
+
+    """
+    Tests our db backup logic: conditional_backup, remove all backups
+    """
+    from kolibri.core.deviceadmin.utils import default_backup_folder
+
+    default_path = default_backup_folder()
+    if not os.path.exists(default_path):
```
rtibbles, line 91: I think the test is still failing because of the backup folder not being clean - in the case when it does exist, you may want to just make sure it is empty here. So force removal, and then just recreate it.

rtibbles (approved):
Tests good, functionality looks good!

PR #8410, opened 2021-09-08
Faster null import

rtibbles:
## Summary
* Uses values call to get all files to download instead of instantiating as models 3-4 x speed up (which should also improve performance for all imports)
* Gives a much earlier and simpler trapdoor for content import when remote content storage is being used
* Should give approximately 4-5x speedup overall for remote storage content import

## Reviewer guidance
Does content import and upgrade still work as expected?
Is it faster for remote storage import?

----

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/core/content/management/commands/deletecontent.py
+++ b/kolibri/core/content/management/commands/deletecontent.py
@@ -149,7 +149,7 @@ class Command(AsyncCommand):
             total=total_file_deletion_operations + additional_progress
         ) as progress_update:
 
-            for file in LocalFile.objects.delete_unused_files():
+            for _ in LocalFile.objects.delete_unused_files():
                 progress_update(1, progress_extra_data)
 
             with db_lock():
--- a/kolibri/core/content/management/commands/exportcontent.py
+++ b/kolibri/core/content/management/commands/exportcontent.py
@@ -7,6 +7,7 @@ from ...utils import paths
 from ...utils import transfer
 from kolibri.core.content.errors import InvalidStorageFilenameError
 from kolibri.core.content.utils.import_export_content import get_import_export_data
+from kolibri.core.content.utils.paths import get_content_file_name
 from kolibri.core.tasks.management.commands.base import AsyncCommand
 from kolibri.core.tasks.utils import get_current_job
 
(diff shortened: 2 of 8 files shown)


radinamatic:
I tested the asset from 2 commits in this build on Windows 10, and I see no issues with importing, but I'm not really noticing  speed improvements either. It takes 5 minutes to download 1GB of content from KA Spanish channel on Studio. Throttling to 4G is just slightly slower, so not sure if the FF is just not too precise with those figures.

Or maybe the speed increase is noticeable just for higher volume imports...? 🤔

rtibbles:
> Or maybe the speed increase is noticeable just for higher volume imports...? thinking

This may not have a noticeable increase in import speed when you are actually downloading content - but it should reduce the time that a task spends at 0% before file downloads even begin.

Probably only really noticeable on a very big channel like KA - where the download task will probably churn on 0% for quite some time.

When running it on a hosted Kolibri instance where we are not actually downloading the content, it should be a significant speed improvement.

radinamatic:
When I select the full KA EN US channel (54GB), it has been sitting for 6 minutes now like this, not even showing 0%... 😕 

Win10 (start)  Running  - Oracle VM VirtualBox_027

rtibbles:
Is that better or worse than before?

aronasorman:
@rtibbles I can test on a BCK instance either tomorrow or Monday!

radinamatic:
I honestly don't remember how long it took before on the full KA EN channel 😞

radinamatic:
It took about 8 minutes for 0% to appear and see something in the progress bar.
Process was also heavy on the VM resources (and it was not particularly _constrained_: 2 cores and 8MB RAM) - although that might be normal.

I cancelled the import task at 2% progress of downloaded resources, but the channel is not appearing neither in Device > Channels, nor in Learn > Channels.

1 | 2 | 3
--- | --- | ---
Workspace 1_002 | Win10 (start)  Running  - Oracle VM VirtualBox_030 | Win10 (start)  Running  - Oracle VM VirtualBox_031

[logs.zip](https://github.com/learningequality/kolibri/files/7139558/logs.zip)

rtibbles:
Yep, cancellation cancels the import process, it doesn't leave it part way done.

radinamatic:
Not even if there are 2 GB of resources locally? 😳

rtibbles:
Oh, I appear to be wrong - it should be annotating after cancellation.

bjester (approved):
Testers did not uncover regressions, looks good

PR #8372, opened 2021-09-01
Fix debug logging

rtibbles:
## Summary
* Premature options evaluation from various modules accessing OPTIONS in the module scope led to logging being initialized without the debug argument
* This removes the use of PICKLE PROTOCOL for the job storage, and hard codes it to 2 - ideally we will change this to a JSON field instead and remove use of pickling.
* Refactors the kolibri.core.tasks.compat module to expose functions that return constructed objects, to avoid evaluation of the OPTION in the module scope
* Turns the current_job_tracker into a SimpleLazyObject so it is only evaluated when accessed.
* Suppresses Django template DEBUG messages

## References
Fixes #7294

## Reviewer guidance
Does running the python devserver show debug statements in the console?
Does trying to navigate to a 404 under e.g. `/en/learn/topics.js` not show huge DEBUG logs as shown in #7294 ?

…

----

## Testing checklist

- [x] Contributor has fully tested the PR manually
- [ ] If there are any front-end changes, before/after screenshots are included
- [ ] Critical user journeys are covered by Gherkin stories
- [x] Critical and brittle code paths are covered by unit tests

## PR process

- [x] PR has the correct target branch and milestone
- [x] PR has 'needs review' or 'work-in-progress' label
- [x] If PR is ready for review, a reviewer has been added. (Don't use 'Assignees')
- [ ] If this is an important user-facing change, PR or related issue has a 'changelog' label
- [ ] If this includes an internal dependency change, a link to the diff is provided

## Reviewer checklist

- Automated test coverage is satisfactory
- PR is fully functional
- PR has been tested for [accessibility regressions](http://kolibri-dev.readthedocs.io/en/develop/manual_testing.html#accessibility-a11y-testing)
- External dependency files were updated if necessary (`yarn` and `pip`)
- Documentation is updated
- Contributor is in AUTHORS.md

PR #8415, opened 2021-09-09
Properly test that we do not prematurely initialize our OPTIONS

rtibbles:
## Summary
* Currently our test for no premature initialization of OPTIONS is undercut by pytest-django loading Django settings, which initialize Kolibri's OPTIONS object
* Get around this by making tests in our root test/ folder run without the pytest-django plugin
* Then move the cli import test into this folder

## References
Follow up to https://github.com/learningequality/kolibri/pull/8372 to try to prevent us getting into that state again!

## Reviewer guidance
Does the separation of tests make sense?

----

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated


tox.ini
```diff
@@ -27,7 +27,8 @@ deps =
 commands =
     sh -c 'kolibri manage makemigrations --check'
     # Run the actual tests
-    python -O -m pytest {posargs:--cov=kolibri --cov-report= --cov-append --color=no}
+    python -O -m pytest {posargs:--cov=kolibri --cov-report= --cov-append --color=no} kolibri
+    python -O -m pytest {posargs:--cov=kolibri --cov-report= --cov-append --color=no} -p no:django test
```
bjester, line 31: My initial concern was that the second execution would overwrite the code coverage, but seems that's what `--cov-append` addresses.

bjester (approved):

tox.ini
rtibbles: We're also not actually reporting our code coverage - it gets messed up because of our conditional test runs.

Issue #8191, opened 2021-07-07
Prompt user to download to prevent user list CSV export popup being blocked silently in Chrome


jamalex:
## Observed behavior

The download popup for user CSV exports is blocked by the browser, without notifying the user.

## Expected behavior

After clicking the "EXPORT" button here:
![image](https://user-images.githubusercontent.com/618416/124833436-0713db00-df33-11eb-8022-b8bbf38e6c8e.png)
a spinner and status appears:
![image](https://user-images.githubusercontent.com/618416/124833502-21e64f80-df33-11eb-8e48-be0522d9200d.png)
but when it completes, the spinner disappears and button reactivates, but no file download is triggered.

Turns out the browser was blocking the popup automatically:
![image](https://user-images.githubusercontent.com/618416/124833633-5528de80-df33-11eb-9922-82f60ffed103.png)

This doesn't happen when downloading the logging CSVs, even though [they all use `window.open`](https://github.com/learningequality/kolibri/search?q=window.open&type=code). My assumption is that this is because those are initiated by a user action (click), whereas the user list CSV export popup is triggered internally (by a state change).

## User-facing consequences

We received a report from a user that the feature was broken, and it took a long time to realize that the cause was the blocked popup.

## Steps to reproduce

Not sure whether this happens locally, or only on a deployed server at a URL with a domain name in it.

## Context

Kolibri: 0.14.7
Local OS: Windows 10
Browser: Chrome

rtibbles:
Seems like this should just be a matter of copying the interaction from the export logs page, where we have an explicit place to download once generation has completed.

khangmach:
Handled in #6582. Work should be coupled with the designs found there

rtibbles:
Fixed in #8419 - further work will be resolved in #6582.

Issue #6582, opened 2020-02-18
generating logs in the UI failed but gave no indication of failure


indirectlylit:
### Observed behavior

Reported in the [forums](https://community.learningequality.org/t/0-13-generating-data-on-session-logs-takes-so-long-doesnt-seem-to-finish/1387/2)

> I tried generating the sessions logs, and it has been more than a week now Kolibri trying to generate the logs. I have tried, shutting down Kolibri, restarting my machine, but even after all this, it does not stop the  generating process, I cant stop so it has been ongoing.
> 
> ![image](https://user-images.githubusercontent.com/2367265/74775832-562c7380-524b-11ea-95aa-bd9e0c0ef310.png)

The user was able to successfully export the logs from the command line using:

```bash
kolibri manage exportlogs --log-type summary
kolibri manage exportlogs --log-type session
```

### Expected behavior

Logs should be generated and the UI should be updated

### User-facing consequences

Cannot export logs

### Errors and logs

unknown - needs follow-up

### Steps to reproduce

unknown - needs follow-up

### Context

unknown - needs follow-up

rtibbles:
Looking at the task handling code for this page: https://github.com/learningequality/kolibri/blob/develop/kolibri/plugins/facility/assets/src/modules/manageCSV/actions.js#L67 it appears that there is no handling of task failure.

My assumption is that the export task has failed somehow (possibly due to the Windows unicode bug fixed here: https://github.com/learningequality/kolibri/pull/7947 and so no error is now reported to the frontend. The only surprising thing is that the erro persisted across restarts.

I think the solution here is to properly report task failures.

khangmach:
![slice.png](https://images.zenhubusercontent.com/6042a21c8000460e5640191f/28f77eb9-564c-475c-bfcc-d4a498a7c70e)

Here's a simple update to the log generation UI. 

Core adjustments:
- condensing "note" text into a single "learn more" modal with similar functionality to the ToS link below on the page
- Add retry functionality
- Add error handling for generation failing via unknown/general errors
- Change 'download' behavior of "Import and export users" section to match log export (related issue: #8191)

PR #8419, opened 2021-09-10
a way to handle pop-up blocker in mozilla

Evgeni998:
## Summary
Following up on the https://github.com/learningequality/kolibri/issues/8191  issue, After testing on Chrome and Firefox, when the user tries to export for the first time on Firefox it show a pop-up blocker

Screenshot 2021-09-10 124826

That blocker only happened on Firefox on the first try ( so I was testing on a private window ) and in Chrome it worked fine all the time. After reading some of the browser default settings, I see that there is a field block pop-up windows that is set to true as  default. When you manually allow the export to happen it adds that url in the exceptions tab and that's why the second and every seqential try it automatically downloads it. 

Screenshot 2021-09-10 125345

One solution that I first started implementing was to add a browser extension called [browserSettings](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/browserSettings) which allowed me to change the browser settings of the user and I could untick the blocker field. I thought that the solution was somewhat malicious and it's not up to me to control the settings of the user so I tried with a different approach. I read that the main reason for the pop-up blocker was that the window.open() didn't happen from user action but from a state change(which was async)

`The general rule is that popup blockers will engage if window.open or similar is invoked from javascript that is not invoked by direct user action. That is, you can call window.open in response to a button click without getting hit by the popup blocker, but if you put the same code in a timer event it will be blocked.`

So the only solution that I found suitable in this case was to first create a new window with window.open and then after the state changes to make that link download the file. One problem that still persists is that the window still remains open after the user exported the data and, so that is something to be worked on.

## Reviewer guidance

You can test the pop-up blocker by opening firefox in a private window and trying to export the data.

----

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/plugins/facility/assets/src/views/DataPage/ImportInterface/index.vue
+++ b/kolibri/plugins/facility/assets/src/views/DataPage/ImportInterface/index.vue
@@ -18,18 +18,24 @@
         @click="showInfoModal = true"
       />
     </p>
-    <p style="margin-top: 24px;">
+    <p style="margin-top: 24px; display:flex; flex-direction: column">
       <KRouterLink
         :text="$tr('import')"
         appearance="raised-button"
-        style="margin: 0 16px 0 0;"
+        style="width: max-content; margin: 0 16px 20px 0;"
         :to="$store.getters.facilityPageLinks.ImportCsvPage"
       />
       <KButton
-        :text="$tr('export')"
+        :text="$tr('downloadCSV')"
         appearance="raised-button"
-        style="margin: 0 16px 0 0;"
-        :disabled="isExporting"
+        style="width: max-content; margin: 0 16px 20px 0;"
+        :disabled="!exported"
+        @click="downloadCsv"
+      />
+      <KButton
+        appearance="basic-link"
+        :text="$tr('generateCSV')"
+        style="margin: 0px 8px 10px 0px"
         @click="exportCsv"
       />
       <DataPageTaskProgress v-if="isExporting">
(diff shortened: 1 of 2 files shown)


indirectlylit:
Hi @Evgeni998 - it wasn't very clear in #8191 but if you look in the comments, there's a reference to another issue #6582 which contains more information as a footnote.

The fix here is to update user export flow to match the session and summary log export flows. In particular, the log flows have separate user-initiated actions for generating and downloading:

> image

This is as opposed to the export user flow, which tries to combine them into a single 'export' action:

> image


So we'll update this to instead have separate actions for 'generate' and 'download':

> image

(Note that the specific user-facing strings above are just for example and we should iterate on those before finishing the PR.)

This will cause downloading to be the direct result of a user-initiated action and bypass popup blockers.

kolibri/plugins/facility/assets/src/views/DataPage/ImportInterface/index.vue
```diff
@@ -68,31 +74,31 @@
...
+            this.exportUsersFilename,
+            this.$store.getters.activeFacilityId
+          ),
+          '_blank'
+        );
+      },
     },
     $trs: {
+      generateCSV: {
+        message: 'Generate CSV file',
```
radinamatic, line 100: Is it possible to implement the string change that we have for logs?
`Generate CSV file` when there's none available, and `Generate a new CSV file` when there is an older one present.

kolibri/plugins/facility/assets/src/views/DataPage/ImportInterface/index.vue
```diff
@@ -114,8 +120,8 @@
         message: 'Import a CSV file to create and update users',
         context: 'Additional information about importing external spreadsheets.\n',
       },
-      export: {
-        message: 'Export',
+      downloadCSV: {
+        message: 'Download CSV',
```
radinamatic, line 124: **Download** breaks the connection with the wording in the surrounding strings, I suggest we go with something like `Export users (CSV)`.

rtibbles (approved):
Functionality confirmed working by @radinamatic, we'll merge and iterate on strings.

rtibbles:
Build issue is fixed in develop.

PR #8421, opened 2021-09-12
Remove all user strings with instances of 'active/inactive' 

### LINKED ISSUE #8420: Remove the obsolete user strings with 'active/inactive'
## Observed behavior
[Prior discussion on Slack](https://learningequality.slack.com/archives/CB37UM23A/p1631222446061300)
We currently use naming strategy of **Open/Closed** exam and **Visible to learners** for lessons. 

[Full list of the strings in question](https://www.notion.so/learningequality/833e2a916545468295b2b2933a8f0290?v=3e22809b9b374830a9ad5505c1b81021)

## Errors and logs
…

## Expected behavior
…

## User-facing consequences
…

## Steps to reproduce
…

## Context
…

radinamatic:
## Summary
For the time being I just commented out the relative pieces of code so the strings do not get extracted. Ideally we should delete them, but this should be enough to avoid them ending up on Crowdin.

My commenting out may not be the _prettiest_, but it seems to do the job, and at least I did not see any errors in the terminal or the browser console.

## References
Fixes #8420 

## Reviewer guidance
Test if pages:

- Reports > Lessons
- Reports > Quizzes
- Plan > Lessons
- Plan > Quizzes

work as expected.

----

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/plugins/coach/assets/src/views/common/LessonActive.vue
+++ b/kolibri/plugins/coach/assets/src/views/common/LessonActive.vue
@@ -28,15 +28,15 @@
         required: true,
       },
     },
-    computed: {
-      label() {
-        return this.active ? this.$tr('lessonActiveLabel') : this.$tr('lessonInactiveLabel');
-      },
-    },
-    $trs: {
-      lessonActiveLabel: 'Active',
-      lessonInactiveLabel: 'Inactive',
-    },
+    // computed: {
+    //   label() {
+    //     return this.active ? this.$tr('lessonActiveLabel') : this.$tr('lessonInactiveLabel');
+    //   },
+    // },
+    // $trs: {
+    //   lessonActiveLabel: 'Active',
+    //   lessonInactiveLabel: 'Inactive',
+    // },
   };
 
 </script>
--- a/kolibri/plugins/coach/assets/src/views/common/QuizActive.vue
+++ b/kolibri/plugins/coach/assets/src/views/common/QuizActive.vue
@@ -28,15 +28,15 @@
         required: true,
       },
     },
-    computed: {
-      label() {
-        return this.active ? this.$tr('activeQuizLabel') : this.$tr('inactiveQuizLabel');
-      },
-    },
-    $trs: {
-      activeQuizLabel: 'Active',
-      inactiveQuizLabel: 'Inactive',
-    },
+    // computed: {
+    //   label() {
+    //     return this.active ? this.$tr('activeQuizLabel') : this.$tr('inactiveQuizLabel');
+    //   },
+    // },
+    // $trs: {
+    //   activeQuizLabel: 'Active',
+    //   inactiveQuizLabel: 'Inactive',
+    // },
   };
 
 </script>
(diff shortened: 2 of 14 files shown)


kolibri/plugins/coach/assets/src/views/common/LessonActive.vue
```diff
@@ -28,15 +28,15 @@
...
-    computed: {
-      label() {
-        return this.active ? this.$tr('lessonActiveLabel') : this.$tr('lessonInactiveLabel');
-      },
-    },
-    $trs: {
-      lessonActiveLabel: 'Active',
-      lessonInactiveLabel: 'Inactive',
-    },
+    // computed: {
```
nucleogenesis, line 31: This `LessonActive` component seems to be used in `AssignmentSummary` (not sure where in the app that is exactly but maybe in Coach > Plan somewhere)

This component has one job which is to say "Active" or "Inactive" based on the given value. 

So this component should be refactored out of our app altogether or we should keep the strings to avoid them missing in `AssignmentSummary`.

kolibri/plugins/coach/assets/src/views/common/QuizActive.vue
```diff
@@ -28,15 +28,15 @@
...
-    computed: {
-      label() {
-        return this.active ? this.$tr('activeQuizLabel') : this.$tr('inactiveQuizLabel');
-      },
-    },
-    $trs: {
-      activeQuizLabel: 'Active',
-      inactiveQuizLabel: 'Inactive',
-    },
+    // computed: {
```
nucleogenesis, line 31: Same here as in the `LessonActive` changes. This is used in `ReportsGroupReportQuizHeader` currently so we'll need to account for that with the refactor by either adding or borrowing a newer string in place of this Active|Inactive one.

kolibri/plugins/coach/assets/src/views/common/LessonActive.vue
nucleogenesis: Confirming now that we can also just remove `AssignmentSummary` as it is an unused component.

kolibri/plugins/coach/assets/src/views/common/QuizActive.vue
nucleogenesis: screenshot-localhost_8000-2021 09 14-09_54_30

`ReportsGroupReportQuizQuestionListPage.vue` and `ReportsGroupReportQuizLearnerListPage.vue`

have these headers for "status" and "average score" - @jtamiace safe to just remove the Status?

Seems also that there may be a bug where the average score isn't showing that seems unrelated to this PR.

nucleogenesis (approved):
For the purposes of getting this in before String Freeze - go for it. I don't see anything here that would block that.

However, a follow-up issue should be made where a dev goes through and cleans up - not just the comments, but also to verify that all unnecessary functionality is removed - including one or more completely useless components.

I think that leaving the comments like so will be helpful in navigating that work so no need to remove them yet.

rtibbles (dismissed):
Holding this up for now until we can confirm that the strings we would want to update these to `open/closed` and `visible to learner` for quizzes/lessons are in a place where we can implement these changes after string freeze.

PR #8422, opened 2021-09-13
Extend custom content renderer API

danigm:
## Summary

Extend the custom content renderer javascript API to be able to do advanced content query and add a random filter to get the content randomly ordered.

## Testing checklist

- [*] Contributor has fully tested the PR manually

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/core/content/api.py
+++ b/kolibri/core/content/api.py
@@ -157,6 +157,10 @@ class ContentNodeFilter(IdFilter):
     parent__isnull = BooleanFilter(field_name="parent", lookup_expr="isnull")
     include_coach_content = BooleanFilter(method="filter_include_coach_content")
     contains_quiz = CharFilter(method="filter_contains_quiz")
+    # Use a CharFilter instead a BooleanFilter to be able to avoid cache using
+    # different values
+    random = CharFilter(method="filter_random")
+    only_content = BooleanFilter(method="filter_only_content")
 
     class Meta:
         model = models.ContentNode
@@ -175,6 +179,26 @@ class ContentNodeFilter(IdFilter):
             "kind_in",
             "contains_quiz",
         ]
+
+    def filter_only_content(self, queryset, name, value):
+        """
+        Return only content nodes
+
+        :param queryset: all content nodes for this channel
+        :return: content nodes without topics
+        """
+        return queryset.exclude(kind="topic")
+
+    def filter_random(self, queryset, name, value):
+        """
+        Return the content randomly ordered
+
+        :param queryset: all content nodes for this channel
+        :param value: a random value, it's not used in the query but it's used
+                      by the cache system, so same value could return same order
+        :return: content nodes randomly ordered
+        """
+        return queryset.order_by("?")
 
     def filter_kind(self, queryset, name, value):
         """
--- a/kolibri/plugins/learn/assets/src/views/ChannelRenderer/CustomContentRenderer.vue
+++ b/kolibri/plugins/learn/assets/src/views/ChannelRenderer/CustomContentRenderer.vue
@@ -117,6 +117,11 @@
           getParams: {
             ids: options.ids,
             parent: options.parent === 'self' ? this.topic.id : options.parent,
+            channel_id: this.topic.id,
+            page: options.page ? options.page : 1,
+            page_size: options.pageSize ? options.pageSize : 50,
+            only_content: options.only_content,
+            random: options.random,
           },
         })
           .then(contentNodes => {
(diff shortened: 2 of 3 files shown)


kolibri/plugins/learn/assets/src/views/ChannelRenderer/CustomContentRenderer.vue
```diff
@@ -117,6 +117,7 @@
           getParams: {
             ids: options.ids,
             parent: options.parent === 'self' ? this.topic.id : options.parent,
+            ...options.extra,
```
rtibbles: I'd rather have these continue to be explicit to keep the API documented, otherwise, we might look for where the `random` filter in the FilterSet is being used and realize we don't have any code references to it anywhere, and then delete it.

packages/hashi/src/kolibri.js
```diff
@@ -94,6 +94,7 @@ export default class Kolibri extends BaseShim {
        * @param {string[]} options.ids - an array of ids to filter by
        * @param {number} [options.page=1] - which page to return from the result set
        * @param {number} [options.pageSize=50] - the page size for pagination
+       * @param {Object} options.extra - extra parameters to pass to the fetchCollection call
```
rtibbles: Likewise here.

kolibri/core/content/api.py
```diff
@@ -176,6 +179,17 @@ class Meta:
...
+    def filter_random(self, queryset, name, value):
+        """
+        Return the content randomly ordered
+
+        :param queryset: all content nodes for this channel
+        :param value: a random value, it's not used in the query but it's used
+                      by the cache system, so same value could return same order
+        :return: content nodes randomly ordered
+        """
+        return queryset.order_by("?")
```
rtibbles, line 201: I assume there's no need of a guarantee of persistent randomization between different clients for the same 'seed' value here?

packages/hashi/src/kolibri.js
manuq: Yes I agree, all should be explicit and documented. No "extra" parameters.

kolibri/core/content/api.py
manuq: No, there is no need to fetch the same results passing the same seed value.

kolibri/plugins/learn/assets/src/views/ChannelRenderer/CustomContentRenderer.vue
manuq: Yes I agree. I see the way you @danigm found to pass extra parameters as a fantastic starting point, but it's a shortcut. All options should be explicit and well documented. Not directly accessing the backend from the frontend.

From the frontend, the API could be designed like this for fetching 5 random results using pagination:

```
options: {
  randomize_results: true,
  page_size: 5,
  page: 1,
}
```

Or add a separate API method to fetch N random content nodes, which is all we need. But that would only fit our current needs.

danigm:
Thanks for the review, I've updated the PR to a more specific solution instead of the generic one. I left the random filter, but also added another one to also exclude topic contents.

In this solution I've also added the pagination parameters to the query, because it was not used.

I've also limited the `fetchColletion` queries in this API to the channel nodes, to avoid leaking outside the sandbox. But I'm not sure if this should be a restriction or if it could be an optional parameter something like `channel: 'self'`.

packages/hashi/src/kolibri.js
```diff
@@ -94,6 +94,9 @@ export default class Kolibri extends BaseShim {
        * @param {string[]} options.ids - an array of ids to filter by
        * @param {number} [options.page=1] - which page to return from the result set
        * @param {number} [options.pageSize=50] - the page size for pagination
+       * @param {string} options.random - set to a random string to get the result randomly ordered
+       * @param {boolean} options.only_content - set to true to skip topics
```
manuq, line 98: All options so far are cameCase, so I would expect the same here.

packages/hashi/src/kolibri.js
```diff
@@ -94,6 +94,9 @@ export default class Kolibri extends BaseShim {
        * @param {string[]} options.ids - an array of ids to filter by
        * @param {number} [options.page=1] - which page to return from the result set
        * @param {number} [options.pageSize=50] - the page size for pagination
+       * @param {string} options.random - set to a random string to get the result randomly ordered
```
manuq, line 97: From the user of the API, it would be better to say "randomize: true" than passing a random string. Or is this option the random seed?

danigm: It's a random "seed". The API is caching every request, so we can't use a boolean here because in that case we'll get the same ordering on every request with the same parameters.

danigm:
I'm  closing this PR for now. We'll work on this on our fork and we can create the PR once it's ready for review.

PR #8423, opened 2021-09-13
Add suggested time component with strings

### LINKED ISSUE #8111: Display estimated time on resources

nucleogenesis:
## Summary

This is only to add strings and includes minimal functionality.

screenshot-localhost_8000-2021 09 13-14_09_17

Adding strings and component for "Suggested Time" frontend feature. 

Does not do anything except provide three possible strings.

1) A pluralized "N minutes" or "N minute" string (depending on if N is 1 or not).
2) A tooltip message "Suggested time to complete" which is to be shown on hover [pending feedback here](https://github.com/learningequality/kolibri/pull/8395#issuecomment-913896554) from @jtamiace 
3) If the decision is to opt for showing the label up front, there is a pluralized "Suggested time to complete: N minutes" string as well in case we do not make the decision before string freeze.

## References

Fixes #8111 
#8395 

## Reviewer guidance

Are these going to work? Do we need different strings for this use case?

Do we need to have it go "X hours Y minutes"?

----

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/plugins/learn/assets/src/views/SuggestedTime.vue
+++ b/kolibri/plugins/learn/assets/src/views/SuggestedTime.vue
@@ -1,13 +1,10 @@
 <template>
 
   <div class="time-context">
-    <div v-if="time > 0">
-      {{ $tr('shortTime', { time: 120 }) }}
-    </div>
-    <div v-if="time > 0">
-      {{ $tr('suggestedTimeLabel', { time: 120 }) }}
-    </div>
-    {{ $tr('suggestedTimeTooltip') }}
+    <span v-if="seconds > 0">
+      {{ $tr('suggestedTimeLabel') }}
+      <TimeDuration :seconds="seconds" :expand="true" />
+    </span>
   </div>
 
 </template>
@@ -15,29 +12,23 @@
 
 <script>
 
+  import TimeDuration from 'kolibri.coreVue.components.TimeDuration';
+
   export default {
     name: 'SuggestedTime',
+    components: { TimeDuration },
     props: {
       /* Time in ms */
-      time: {
+      seconds: {
         type: Number,
-        default: 0,
+        default: 8100,
       },
     },
     $trs: {
-      suggestedTimeTooltip: {
+      suggestedTimeLabel: {
         message: 'Suggested time to complete',
         context:
-          'A tooltip explaning what the clock icon and time indicates about the resource  they are viewing',
-      },
-      shortTime: {
-        message: '{time} {time, plural, one {minute} other {minutes}}',
-        context: 'Converts the time given into internationalized time. Nothing to translate here.',
-      },
-      suggestedTimeLabel: {
-        message: 'Suggested time to complete: {time} {time, plural, one {minute} other {minutes}}',
-        context:
-          "Indicates to the user the suggested time the resource should take to complete. The '# minutes' string is internationalized automatically",
+          'A label preceding a number of minutes that the current resource is estimated to take to complete',
       },
     },
   };
--- a/kolibri/plugins/learn/assets/src/views/SuggestedTime.vue
+++ b/kolibri/plugins/learn/assets/src/views/SuggestedTime.vue
@@ -1,0 +1,48 @@
+<template>
+
+  <div class="time-context">
+    <div v-if="time > 0">
+      {{ $tr('shortTime', { time: 120 }) }}
+    </div>
+    <div v-if="time > 0">
+      {{ $tr('suggestedTimeLabel', { time: 120 }) }}
+    </div>
+    {{ $tr('suggestedTimeTooltip') }}
+  </div>
+
+</template>
+
+
+<script>
+
+  export default {
+    name: 'SuggestedTime',
+    props: {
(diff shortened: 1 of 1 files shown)


kolibri/plugins/learn/assets/src/views/SuggestedTime.vue
```diff
@@ -0,0 +1,48 @@
...
+        default: 0,
+      },
+    },
+    $trs: {
+      suggestedTimeTooltip: {
+        message: 'Suggested time to complete',
+        context:
+          'A tooltip explaning what the clock icon and time indicates about the resource  they are viewing',
+      },
+      shortTime: {
```
radinamatic: @nucleogenesis Is there a way to reuse the existing `minute/minutes` string from the [TimeDuration](https://github.com/learningequality/kolibri/blob/develop/kolibri/core/assets/src/views/TimeDuration.vue#L42) component? 

If I'm understanding correctly, you could even combine the `suggestedTimeTooltip` message with the existing `minutes` from `TimeDuration` component to achieve the purpose of `sugestedTimeLabel`, and effectively need to stub just one string here 😉

kolibri/plugins/learn/assets/src/views/SuggestedTime.vue
```diff
@@ -0,0 +1,48 @@
...
+        message: '{time} {time, plural, one {minute} other {minutes}}',
```
rtibbles: These do need to be translated - you're not using any time related syntax here, this is just a simple `plural` message.

If we use this string and it is not translated, we will end up with `minutes` being displayed to end users.

rtibbles (changes requested):
I think using the `TimeDuration` component is the most stringly economical way of doing this - I wonder if we could just avoid the 'full' label that has the prefix and the TimeDuration text in it?

kolibri/plugins/learn/assets/src/views/SuggestedTime.vue
nucleogenesis: I noticed that and figured that when I come back to it (post strings-getting-in rush) then I'll conditionalize showing any strings at all on whether or not we have > 0 passed in

nucleogenesis: Ah - I already had it conditionalized.

@radinamatic - I think that TimeDuration will work the best but it isn't accurate - it rounds down so 2hrs 45min would show "2 hours". If @jtamiace and @rtibbles are okay with the way it rounds, then this component can be the "Suggested time to complete:" followed by the TimeDuration component.

radinamatic: Understood, thank you @nucleogenesis! 
That type of time rounding does not seem like a big concern to me, but will defer to @jtamiace & @rtibbles.

rtibbles: I think if it's an issue we can probably further conditionalize TimeDuration to round up rather than down?

jtamiace: It seems okay for now since it still gives a ballpark about expected length, but I'd hope accuracy is something we can improve in the future. 

Any issue I can foresee is around planning and scheduling around learning with limited time and the resource taking longer than they expected, but I can't imagine that causing too many issues because it's just an estimate and people will engage resources at their own discernment.

radinamatic: > further conditionalize TimeDuration to round up rather than down

And I totally misread that, and thought that 1:45h was rounded to 2h... 🤦🏽‍♀️ 
Yes, rounding up would be more accurate in this case!

rtibbles: > I'd hope accuracy is something we can improve in the future.

We can also conditionalize the TimeDuration component's threshold for when it starts displaying 'hours', so we could make it display whole minutes for up to 120 minutes, and only then start displaying hours. Once we get past 120 minutes, hopefully any time durations are very ballpark anyway.

nucleogenesis: I just pushed it up with only the "Suggested time to complete" string.

In follow up I can use TimeDuration to get things looking how we want - but at this point this should be a go for strings I think. Anybody feel free to force push anything you need to get this merged!

rtibbles: Thanks for the update - and apologies for brigading your tiny PR!

rtibbles (approved):

PR #8432, opened 2021-09-15
Don't require `label` and `icon` prop in `CoreMenuOption`

MisRob:
`CoreMenuOption` provides two ways of defining its content - one through props and one through the default slot. When I was creating `LearningActivityBar`, I needed [to use slot](https://github.com/learningequality/kolibri/blob/develop/kolibri/plugins/learn/assets/src/views/LearningActivityBar.vue#L68) to be able to define the icon color and I didn't notice that there are two errors in the console caused by `label` and `icon` props being required by `CoreMenuOption`:

Screenshot from 2021-09-15 14-30-18

This PR fixes these errors by making `label` and `icon` props optional so it works properly when `CoreMenuOption`'s content is provided through its default slot.

### DIFF
--- a/kolibri/core/assets/src/views/CoreMenu/CoreMenuOption.vue
+++ b/kolibri/core/assets/src/views/CoreMenu/CoreMenuOption.vue
@@ -12,13 +12,13 @@
     >
       <slot>
         <KLabeledIcon>
-          <template #icon>
+          <template v-if="icon" #icon>
             <KIcon
               :icon="icon"
               :class="$computedClass(optionIconStyle)"
             />
           </template>
-          <div>{{ label }}</div>
+          <div v-if="label">{{ label }}</div>
         </KLabeledIcon>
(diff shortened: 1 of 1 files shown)


kolibri/core/assets/src/views/CoreMenu/CoreMenuOption.vue
```diff
@@ -37,7 +37,8 @@
     props: {
       label: {
         type: String,
-        required: true,
+        required: false,
```
rtibbles, line 40: Maybe just for safety, we should conditionalize the `KIcon` component on `icon` being truthy, and the div for `{{ label }}` on the label being truthy?

MisRob: Yes that's a good point, will do

rtibbles (approved):

PR #8438, opened 2021-09-16
First pass at regression test and fix for SoUD FacilityUser sync conflict

jamalex:
## Summary

In Morango, when a model is modified on both sides, it results in a merge conflict, and one copy of the data wins (and the serialized copy of the one that didn't win is stored in case the app wants to resolve the conflict later).

We don't allow FacilityUsers to be synced back across a single-user connection -- they can only be read. That's why we don't allow users to edit their profile on a SoUD -- but that doesn't stop some backend machinery from still writing to the FacilityUser on a SoUD. In particular, we saw a case with Django writing to the `last_login` field on the `FacilityUser` that made it "dirty", and then when a `full_name` change was synced in from a full-facility device, it conflicted, and the tablet's copy won, so the data with the updated `full_name` was relegated to the `conflicting_serialized_data` field in the Store.

This PR ~prevents~ attempts to prevent changes made to a FacilityUser on a SoUD from causing a conflict with incoming changes. **The regression tests work as intended, but my hook doesn't currently seem to be getting picked up, so the tests are still failing. Ideas welcome around what I may be missing.**

## Reviewer guidance

I'm running the tests with:

```
INTEGRATION_TEST=true pytest kolibri/core/auth/test/test_morango_integration.py::EcosystemSingleUserAssignmentTestCase::test_facility_user_conflict_syncing_from_tablet --pdb --capture=no
```

and

```
INTEGRATION_TEST=true pytest kolibri/core/auth/test/test_morango_integration.py::EcosystemSingleUserAssignmentTestCase::test_facility_user_conflict_syncing_from_laptop --pdb --capture=no
```

----

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/core/auth/kolibri_plugin.py
+++ b/kolibri/core/auth/kolibri_plugin.py
@@ -1,0 +1,26 @@
+from kolibri.core.auth.hooks import FacilityDataSyncHook
+from kolibri.core.auth.models import FacilityUser
+from kolibri.plugins.hooks import register_hook
+
+
+@register_hook
+class SingleFacilityUserChangeClearingHook(FacilityDataSyncHook):
+    def pre_transfer(
+        self,
+        dataset_id,
+        local_is_single_user,
+        remote_is_single_user,
+        single_user_id,
+        context,
+    ):
+
+        # If this is a single-user device, make sure we ignore any changes
+        # that have been made to FacilityUsers, to avoid conflicts.
+        if local_is_single_user and single_user_id is not None:
+            try:
+                user = FacilityUser.objects.get(
+                    id=single_user_id, _morango_dirty_bit=True
+                )
+                user.save(update_dirty_bit_to=False)
+            except FacilityUser.DoesNotExist:
+                pass
--- a/kolibri/core/auth/test/sync_utils.py
+++ b/kolibri/core/auth/test/sync_utils.py
@@ -71,6 +71,17 @@ class KolibriServer(object):
             )
         )
 
+    def update_model(self, model, pk, **kwargs):
+        kwarg_text = json.dumps(kwargs, default=str)
+        self.pipe_shell(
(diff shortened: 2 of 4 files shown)


radinamatic:
I confirm that the change of learner's **Full Name** does get synced to the learn-only device, but could it be that it cannot sync on it's own, and needs an interaction with resources, to sort of _get admitted to the syncing train_? 🤔 

This is the scenario:

- Set up a learn-only device.
- Synced with server, class and lesson visible to learner.
- Disconnected the learn-only device from the network.
- Edited the **Full Name** of the learner on the server device.
- Reconnected the learn-only device and observed the sync on both: while devices synced twice, and I was refreshing the **Profile** page on the learn-only device, **Full Name** change was not visible on the learn-only device.
- Only after I interacted with a lesson, and another sync happened, I was able to see the new **Full Name** in the **Profile**.

[logs-and-DB-Windows10-learn-only.zip](https://github.com/learningequality/kolibri/files/7187985/logs-and-DB-Windows10-learn-only.zip)
[logs-and-DB-Ubuntu18-server.zip](https://github.com/learningequality/kolibri/files/7187986/logs-and-DB-Ubuntu18-server.zip)


new-full-name

radinamatic:
I repeated the scenario 5 times more, without engaging with resources on the learn-only device, and almost invariably the full name change becomes visible to the learner only after the 3rd or the 4th sync.

[DB-and-logs-Ubuntu18.zip](https://github.com/learningequality/kolibri/files/7188630/DB-and-logs-Ubuntu18.zip)
[DB-and-logs-Win10.zip](https://github.com/learningequality/kolibri/files/7188631/DB-and-logs-Win10.zip)


https://user-images.githubusercontent.com/1457929/133858061-7e4bc998-0865-457d-8940-afdc99cc0561.mp4

rtibbles (approved):
Updates look good, and fix confirmed in manual testing.

Issue #6574, opened 2020-02-15
Wrong resource deletion count in Tasks detail view


intelliant01:
### Observed behavior

Deleted 6 files + 1 topic folder from an imported channel (`zapok-dofil`) and it shows a very high and unrelated number of resources deleted.

Tried it twice on one server (Linux + Kolibri 0.13.1rc1) and in each case it gave an absurd count.

![wrong_delete_count](https://user-images.githubusercontent.com/2577060/74586253-c8167a00-500b-11ea-8468-f73c58a6c592.gif)

After second delete attempt, just to ratify if the previous number was a result of any pending cleanup -

![Task manager - incorrect resource count-2](https://user-images.githubusercontent.com/2577060/74586259-cd73c480-500b-11ea-978a-0ecaa4c0dba9.png)

Tried it on another as well with a different OS (Windows + Kolibri 0.13.0) and yet got the same issue.

![Task manager - incorrect resource count](https://user-images.githubusercontent.com/2577060/74586263-d8c6f000-500b-11ea-90bb-4a1b9aff3c58.png)

_Added later:_
Tried it with another channel (`lirub-laruk`) to be sure that it isn't an isolated issue w.r.t. a channel. Deleted 8 resources and there is no count shown -

![Task manager - incorrect resource count-3](https://user-images.githubusercontent.com/2577060/74586747-c9967100-5010-11ea-8d4a-82a678f8b647.png)
 

### Expected behavior

Count should be the number of resources a user actually deletes. Herein 6 or 7 if topic folder is counted as resource.

### User-facing consequences

Confusion. Anxiety that they may have deleted something extra unintentionally.

### Errors and logs

Shall provide if needed.

### Steps to reproduce

As evident in animation above.

### Context

Given above.
Referenced channels : `zapok-dofil`, `lirub-laruk`

ivanistheone:
Related issue: https://github.com/learningequality/kolibri/issues/6351

cpauya:
I have encountered this similar issue of (wrong size reported) while testing [Kolibri v0.13.2-beta1](https://github.com/learningequality/kolibri/releases/tag/v0.13.2-beta1) in an Ubuntu 16.04.6 VM

The imported KA-English channel is only 19GB but the delete task shows 54GB:
![kolibri-v0 13 2-b1--delete size wrong](https://user-images.githubusercontent.com/175580/78031700-7a28bc00-7396-11ea-912b-bf85c4cf9c93.gif)

[...]

PR #8441, opened 2021-09-16
Wrong resource deletion

AtKristijan:
## Summary

…

## References

…

## Reviewer guidance

…

----

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/core/content/management/commands/deletecontent.py
+++ b/kolibri/core/content/management/commands/deletecontent.py
@@ -5,7 +5,7 @@ from django.core.management.base import CommandError
 from django.db.models import Sum
 
 from kolibri.core.content.models import ChannelMetadata
-from kolibri.core.content.models import LocalFile
+from kolibri.core.content.models import LocalFile, ContentNode
 from kolibri.core.content.utils.annotation import propagate_forced_localfile_removal
 from kolibri.core.content.utils.annotation import reannotate_all_channels
 from kolibri.core.content.utils.annotation import set_content_invisible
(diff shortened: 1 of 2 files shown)


kolibri/core/content/management/commands/deletecontent.py
```diff
@@ -127,9 +127,20 @@ def handle_async(self, *args, **options):
         delete_all_metadata = delete_metadata(
             channel, node_ids, exclude_node_ids, force_delete
         )
-
+        nodes = ContentNode.objects.filter(id__in=node_ids).all()
```
rtibbles: Looks like these lines are unused.

kolibri/utils/server.py
```diff
@@ -657,8 +657,8 @@ def start(port=0, zip_port=0, serve_http=True, background=False):
     service_plugin.subscribe()
 
     # Setup zeroconf plugin
-    zeroconf_plugin = ZeroConfPlugin(bus, port)
```
rtibbles: Shouldn't commit this change, although I know it's needed for WSL!

kolibri/core/content/management/commands/deletecontent.py
```diff
@@ -138,13 +149,10 @@ def handle_async(self, *args, **options):
                 "file_size__sum", 0
             )
             job.extra_metadata["file_size"] = total_file_deletion_size
-            job.extra_metadata["total_resources"] = total_file_deletion_operations
+            job.extra_metadata["total_resources"] = total_resource_number
             job.save_meta()
-
```
rtibbles, line 143: Would be better to avoid flyby commits like this - seems like your IDE might be reformatting files when you save?

kolibri/core/content/management/commands/deletecontent.py
```diff
@@ -127,9 +127,20 @@ def handle_async(self, *args, **options):
         delete_all_metadata = delete_metadata(
             channel, node_ids, exclude_node_ids, force_delete
         )
-
+        nodes = ContentNode.objects.filter(id__in=node_ids).all()
+        lst_nodes = list(nodes)
         unused_files = LocalFile.objects.get_unused_files()
-
+        total_resource_number, unused_files_, total_bytes_to_transfer = get_import_export_data(
```
rtibbles: In some cases, the `delete_all_metadata` command is already invoking this function - it might be simpler to change what is happening inside that function to always invoke this, and then to return both the `delete_all_metadata` flag, and the `total_resource_number`.

One consideration for this is that this function takes a very long time, so if we can avoid calling it more than we need to, that is preferable.

rtibbles (changes requested):
Some cleanup needed, and potential consolidation of logic.

kolibri/utils/server.py
AtKristijan: This was pushed by mistake. I will fix tihs.

kolibri/core/content/management/commands/deletecontent.py
AtKristijan: I do not have auto reformatting on files when i save but i did reformatting. I wont use reformatting anymore on whole file.

kolibri/core/content/management/commands/deletecontent.py
AtKristijan: I will change it like this so we wont invoke that funciton 2 times

AtKristijan: I tried this but we need to change the logic inside because this `get_import_export_data` is only called when we have force_delete. So there is two possible ways that i can think about it. We can have small meeting and talk about it

rtibbles:
Should fix https://github.com/learningequality/kolibri/issues/6574

pcenov:
Tested and verified as fixed in this [build](https://buildkite.com/learningequality/kolibri-python-package/builds/3770) - the correct number of resources is displayed now after deleting some or all of them.
Fixes #6574

rtibbles (approved):

PR #8442, opened 2021-09-16
Create separate SQLite databases for independent data to increase write concurrency

rtibbles:
## Summary
* Splits out SyncQueue model into its own SQLite database
* Splits out NetworkLocation model and its proxy models into its own SQLite database
* Adds appropriate DBRouters for the above
* Consolidates names of these additional databases in a constants file
* Standardizes the names of the aliases so they can be used to generate the file on disk also
* Adds upgrade task to move historical NetworkLocation objects to the new database
* Generalizes generation of additional databases in Django settings

## Reviewer guidance
Does syncing and zeroconf work as expected?

Part of the motivation for standardizing the additional database aliases and names was so that when this is merged into current develop, we can extend the database preseeding to these other database files.

----

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated


kolibri/core/discovery/upgrade.py
```diff
@@ -0,0 +1,35 @@
...
+
+from kolibri.core.discovery.models import NetworkLocation
+from kolibri.core.upgrade import version_upgrade
+from kolibri.deployment.default.sqlite_db_names import NETWORK_LOCATION
+
+logger = logging.getLogger(__name__)
+
+
+@version_upgrade(old_version="<0.15.0")
+def move_network_location_entries():
```
jredrejo, line 17: As we're doing the upgrade (for the static locations only, I suppose) we could remove this table from the previous database as Django won't do it.
In new installations of kolibri the table won't exist anyway.

jredrejo (approved):
The code is good and I have tested it as much as I have could, without any issues. Maybe it's just my suggestion but it seemed both the server and the client were smoother, both syncing or even importing channels.

Otoh: Windows is again playing with the time and making tests fail.

kolibri/core/discovery/upgrade.py
rtibbles: Can give that a go, yes!

Issue #5719, opened 2019-06-27
refactor User Auth, Learn, Device, and Facility apps to use new Coach-style page indexes


indirectlylit:
### Observed behavior

In Kolibri, we use a thin wrapper around the Vue router which allows routes to be associated with handler functions in addition to components. For some context on the motivations of this design decision, see https://github.com/vuejs/vue-router/issues/958

The typical pattern we originally used for routing in an app was roughly that all routes have a `name` (taken from a list of string constants), a `path`, and a `handler` (usually calling a Vuex action responsible for loading any data necessary for rendering the page):

https://github.com/learningequality/kolibri/blob/9f6eca3b2f078e00a6b7edd6776aff6aab5d4cfd/kolibri/plugins/user/assets/src/routes.js#L21-L29

The `handler` would also be responsible for setting core Vuex `PageName` state. This would in turn be checked by a single index page responsible for the whole app, containing a single instance of `CoreBase`. Both the props passed to `CoreBase` and the page component displayed within `CoreBase` needed to be dynamically set based on the current `PageName`:

https://github.com/learningequality/kolibri/blob/9f6eca3b2f078e00a6b7edd6776aff6aab5d4cfd/kolibri/plugins/user/assets/src/views/UserIndex.vue#L1-L14

https://github.com/learningequality/kolibri/blob/9f6eca3b2f078e00a6b7edd6776aff6aab5d4cfd/kolibri/plugins/user/assets/src/views/UserIndex.vue#L25-L29

https://github.com/learningequality/kolibri/blob/9f6eca3b2f078e00a6b7edd6776aff6aab5d4cfd/kolibri/plugins/user/assets/src/views/UserIndex.vue#L46-L48

In more complex apps (such as Coach), this pushed a lot of complexity into a single Vue file and made it very hard to maintain.

### Proposed common routing pattern

In the new Coach implementation, routes are now associated with both a `handler` _and_ a `component`, and the `name` (if not specified) is drawn automatically from the component name:

https://github.com/learningequality/kolibri/blob/9f6eca3b2f078e00a6b7edd6776aff6aab5d4cfd/kolibri/plugins/coach/assets/src/routes/index.js#L49-L53

Additionally, we sometimes embed information about the page title directly in the route as a `meta` tag (let's discuss this and evaluate if we want to re-use the pattern).

In coach, the index page becomes trivial band contains no logic:

[...]

PR #8443, opened 2021-09-17
split user plugin

indirectlylit:
* Splits `user` app plugin into two plugins: `user_auth` and `user_profile`
* Switches to server-side redirect for auth
* Tighten up layout of first setup wizard page
* Updates URL for setup wizard to just `/setup`
* Adds new `PageRoot` core component and begins migration to that (more post-0.15) - ref #5719


kolibri/core/assets/src/views/PageRoot.vue
```diff
@@ -1,5 +1,6 @@
 <template>
```
rtibbles, line 1: Does this need to be put into the core API? It seems like boiler plate.

kolibri/plugins/user_profile/assets/src/views/commonUserStrings.js
```diff
@@ -0,0 +1,37 @@
+import { createTranslator } from 'kolibri.utils.i18n';
+
+const commonUserStrings = createTranslator('CommonUserPageStrings', {
```
rtibbles: Seems like this should work as the namespace and ids are remaining constant.

kolibri/plugins/user_auth/assets/src/views/UserAuthIndex.vue
```diff
@@ -88,11 +73,6 @@
       },
     },
     $trs: {
-      userProfileTitle: {
```
rtibbles, line 91: Did this message just get removed entirely?

kolibri/plugins/user_profile/assets/src/views/ProfileEditPage.vue
```diff
@@ -175,8 +189,7 @@
               if (this.isReferredFromLearnPage) {
                 this.navigateToLearnPage();
```
rtibbles, line 190: No need to do it in this PR, but can you file an issue to clean up and decouple this behaviour? This will break in the case that the learn plugin is disabled.

I think the preferred solution is to just redirect to the 'next' query param if it is set.

kolibri/plugins/user_profile/assets/src/views/getUrlParameter.js
```diff
@@ -0,0 +1,7 @@
+// https://davidwalsh.name/query-string-javascript
+export default function getUrlParameter(name) {
```
rtibbles: Not actually seeing where this is used in the profile plugin (although with my comment above, it would be, presumably).

rtibbles (commented):
Looks good code wise - a couple of questions, but nothing blocking.

pcenov:
Hi @indirectlylit I was asked by @radinamatic to test this PR and I'm happy to report that all main workflows are functioning correctly. 
At the same time I stumbled upon an issue with the setup wizard for which I'm not sure whether it's caused by any of the above changes and therefore I've reported it separately here: https://github.com/learningequality/kolibri/issues/8489

kolibri/core/assets/src/views/PageRoot.vue
indirectlylit: Indeed, it is in the core API!

https://github.com/learningequality/kolibri/pull/8443/files#diff-a2dd73b039d1c55598114bb71dc7ce6bd5fffd2e7ed8e921b218c49c53e99cc7

It's a step in the direction of this: https://github.com/learningequality/kolibri/issues/5719

Once 0.15 is cut, I'll apply the pattern to `user_auth` also

kolibri/plugins/user_auth/assets/src/views/UserAuthIndex.vue
indirectlylit: Yes, it was redundant with `coreString('profileLabel')`

kolibri/plugins/user_profile/assets/src/views/ProfileEditPage.vue
indirectlylit: I hear what you're saying, but would it actually break? presumably `isReferredFromLearnPage` would always be `false` if Learn is disabled.

kolibri/plugins/user_profile/assets/src/views/commonUserStrings.js
indirectlylit: hm this is wrong though - it contains strings for auth. Will investigate

kolibri/plugins/user_profile/assets/src/views/getUrlParameter.js
indirectlylit: yup this also looks to be only for auth

kolibri/plugins/user_profile/assets/src/views/ProfileEditPage.vue
rtibbles: Not if another plugin was also using `learn` as its URL. The prime example that comes to mind is someone forking our learn plugin but not changing the URL (this has actually happened, so I am not being fanciful).

indirectlylit: opened issue here: https://github.com/learningequality/kolibri/issues/8494

indirectlylit:
@pcenov thank you for testing!

@rtibbles I think this should be ready to go

rtibbles (approved):
Code looks good, QA checks out.

PR #8446, opened 2021-09-17
Handle 500+ status codes when requesting a SoUD sync

rtibbles:
## Summary
* Fixed some assertions in tests
* Added test for handling ConnectionError in request
* Fixed issue where using `requests.exceptions.ConnectionError` made Python report that it was not derived from `BaseException`
* Handle 500+ status codes from the queue endpoint by initiating ttl back off

## Reviewer guidance
Tests cover the changes?

----

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/core/public/utils.py
+++ b/kolibri/core/public/utils.py
@@ -9,6 +9,7 @@ from django.core.management import call_command
 from django.core.urlresolvers import reverse
 from django.utils import timezone
 from morango.models import InstanceIDModel
+from requests.exceptions import ConnectionError
 from rest_framework import status
 
 import kolibri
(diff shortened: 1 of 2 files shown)


kolibri/core/public/utils.py
```diff
@@ -222,8 +223,9 @@ def request_soud_sync(server, user, queue_id=None, ttl=4):
             # got a place in the queue to sync with this server, so we can be
             # more sure that the server is actually available.
             response = requests.put(server_url, json=data, timeout=30)
-
-    except requests.exceptions.ConnectionError:
+        if response.status_code >= status.HTTP_500_INTERNAL_SERVER_ERROR:
+            raise ConnectionError()
```
jamalex, line 227: I'm assuming there's a reason we don't want to use `raise_for_status` here, yeah? Are there cases where we want 4xx response statuses to be swallowed?

jamalex (approved):
I don't know the scheduling code well enough to say anything definitive around how the tests work, but the code change itself seems wise (though I added a comment about `raise_for_status` as a possible alternative pattern, noting that it raises `requests.exceptions.HTTPError` instead, so we'd then need to catch that too).

kolibri/core/public/utils.py
rtibbles: Yeah - any 4xx error codes indicate that the server has received and processed the request, and is not interested, so we just stop trying.

jredrejo (commented):
LGTM, thank you!

Issue #6316, opened 2019-12-15
migrate from pickle to a generic serialization protocol


indirectlylit:
### Observed behavior

Currently we persist information about tasks using pickle, but this has caused us some issues when python versions get mixed. This should not actually happen, but could cause issues in the future...

* important for https://github.com/learningequality/kolibri/issues/6255
* potentially a security risk
* slow performance
* may be problematic for multiprocessing

From @rtibbles 

> I am very open to abandoning pickle altogether - it would place more restrictions on the kind of args and kwargs we pass, but I don't think that's a bad thing.
> 
> I think the only reason it is the current mechanism is that it does get pickled eventually to get passed to the executor pool for execution

### Expected behavior

pickle-free task communication

### User-facing consequences

edge case issues

### Context

0.13.0 beta 7

benjaoming:
There's also some SQLalchemy schema dumping/loading for Kolibri channel import/export that might be of interest. It was already using protocol 2, so not of imminent concern.

* https://github.com/learningequality/kolibri/blob/develop/kolibri/core/content/utils/sqlalchemybridge.py
* https://github.com/learningequality/kolibri/blob/develop/kolibri/core/content/management/commands/generate_schema.py

rtibbles:
I don't count this as serialization - so I was not intending to imply we should changes these - just in task serialization etc.

vkWeb:
Does [`json`](https://docs.python.org/3/library/json.html) from the python's standard library make sense here? json will allow us to encode python object hierarchies to their string representation (which can then be stored in db). 

Tbh, I haven't used pickle and also I'm pretty new to kolibri, so I am not sure if this approach will make sense, but just sharing my thoughts...

EliKlein:
I think I'd like to start looking at this, but I'm also pretty new with Kolibri, so I want to make sure I'm looking in the right place. It should just be the Job class that needs to be serialized right? And then the ORMJob and ScheduledJob classes can have their "obj" column type changed and stuff, to finish it off, basically?

I'll check where else these are all used too, but I just wanna be sure I'm looking at the right thing to start with.

Also, json sounds like it might work, but I think I'm also gonna look for other options to be sure.

rtibbles:
Hi @EliKlein - apologies, I hadn't realized that @vkWeb was not assigned to this issue. He was planning to take this on - but if you want to take a stab at it, he is currently busy with some other tasks, so I'd welcome your contribution.

[...]

PR #8447, opened 2021-09-17
Use JSON serialization for Job into ORMJob

EliKlein:
## Summary

Serializing Job objects with JSON, in order to change a PickleType column to String.

## References

#6316
Jobs are now unable to accept args/kwargs that need to be kept as a specific type (e.g. if you wanted to pass one job as an argument to another job, you'd use its id, not an object representing it)
Jobs now do not save Exceptions as the same type they were initially raised (they become ErrorSavedWithJob)
Jobs now do not save results

The results might be able to be changed to be saved if needed, but Exception typing would be very difficult if possible, and type'd args/kwargs are basically not doable. I would otherwise say that this is necessary in order to abandon the PickleType columns.

Besides the restrictions outlined above, ORMJob and ScheduledJob behave essentially the same as before these changes. They still have a .obj property that references a Job object in python memory; it's just internally stored as a string as well, for when the ORM stops keeping track of those objects.

Most of the comments and such referencing pickling the Job column are altered or removed.

Finally, I unfortunately had to use exactly the same code in two spots (ORMJob and ScheduledJob). There might be a way to consolidate them into one spot, but that seems like something I'd need to know more about SQLAlchemy to do.

## Reviewer guidance

This still needs to run through pytest. I could not get pytest to run through all of the tests on my own machine, despite finding that the concerned tests pass individually (all of the ones that stall or break when run as a group, that is).

I don't believe this code needs any additional tests written for it, as it doesn't add functionality. I changed some old tests to work around the new expected restrictions I mentioned.

----

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated


EliKlein:
I suppose it has to have syntax that works with all versions of python then? I didn't know it doesn't work to use two sets of keyword arguments in python 2.

rtibbles:
Yes - we are still supporting Python 2, as about 40% of our user base that we get telemetry data from is still using Python 2.7, so it's still very much in demand, in spite of being end of life!

rtibbles:
> Finally, I unfortunately had to use exactly the same code in two spots (ORMJob and ScheduledJob). There might be a way to consolidate them into one spot, but that seems like something I'd need to know more about SQLAlchemy to do.

This seems fine to me - if there's a good way to consolidate it, I'll give some pointers, but not something to be overly concerned about.

EliKlein:
I think I can make a lot of these pass real quick, should I update the pull request now or wait for everything to settle?

rtibbles:
I probably won't review fully until Monday, so feel free to make updates!

EliKlein:
I'm getting a test failure (test_inactive_kolibri in test_dbbackup.py) with my most recent commit and I'm having trouble understanding the error. Everything else is passing.

kolibri/core/tasks/exceptions.py
```diff
@@ -15,3 +15,14 @@ class JobNotFound(Exception):
...
 class JobNotRestartable(Exception):
     pass
+
+
+class ErrorSavedWithJob(Exception):
+    def __init__(self, message, prior_type):
+        if message is None or message == "":
+            message = "This job had an error before it was saved to the database."
+        self.prior_type = prior_type
+        super().__init__(message)
```
rtibbles: For a Python 2 compatible super invocation, this needs to be called like this:
```suggestion
        super(ErrorSavedWithJob, self).__init__(message)
```

kolibri/core/tasks/test/taskrunner/test_job_running.py
```diff
@@ -115,9 +114,10 @@ class EventProxy(object):
     creation.
     """
 
-    def __init__(self, *args, **kwargs):
-        self.event_id = uuid.uuid4().hex
-        EVENT_PROXY_MAPPINGS[self.event_id] = Event(*args, **kwargs)
+    def __init__(self, event_id=uuid.uuid4().hex, *args, **kwargs):
```
rtibbles: Should set `event_id=None` here - as the default is evaluated in module scope, not at each invocation of the `__init__` function.

Can then, in the body of the `__init__` function do this:
```
if event_id is None:
    event_id = uuid.uuid4().hex
```

EliKlein:
That should do it then!

kolibri/core/analytics/utils.py
```diff
@@ -423,6 +423,9 @@ def perform_ping(started, server=DEFAULT_SERVER_URL):
 
     language = get_device_setting("language_id", "")
 
+    if not isinstance(started, datetime.datetime):
+        started = datetime.datetime.strptime(started)
+
```
vkWeb: We'll need to supply a second argument that tells the format of the date string to `strptime` like `strptime(started, "%d %m %y")` to work. 

It would be better to use dateutil's parser. Their parser tries best to figure out the format itself. 
```
from dateutil.parser import parse
started = parse(started)
```

kolibri/core/tasks/exceptions.py
```diff
@@ -15,3 +15,14 @@ class JobNotFound(Exception):
...
+
+class ErrorSavedWithJob(Exception):
+    def __init__(self, message, prior_type):
+        if message is None or message == "":
+            message = "This job had an error before it was saved to the database."
+        self.prior_type = prior_type
+        super(ErrorSavedWithJob, self).__init__(message)
+
+    def __str__(self):
+        return "{} Exception type: {}".format(self.message, self.prior_type)
```
vkWeb: I don't think we should be going ahead with a separate exception class to handle exception objects.

My recommendation would be:

Our tasks backend will save the exception's name and the message/traceback as string. Assertions for defensive programming and unit testing will then be normal string comparison, like: 
```
Kolibri plugins will have code like:

if job.exception == "TypeError"
    logger.debug("something wrong")  

Unit tests will have code like:
assert job.exception == "TypeError"
```

kolibri/core/tasks/job.py
```diff
@@ -156,9 +172,43 @@ def __getstate__(self):
...
+
+        if self.exception is not None:
+            working_dictionary["exception_type"] = type(self.exception).__name__
+            if self.exception.args[0] and isinstance(self.exception.args[0], str):
+                working_dictionary["exception_message"] = self.exception.args[0]
+
+        if self.state in [State.CANCELING, State.RUNNING]:
+            logger.debug(
+                "A task is currently in action. This is a bad time to save it to the database."
+            )
```
vkWeb: This is not true. There's no bad time to update the job database. So, we should remove this.

kolibri/core/tasks/job.py
```diff
@@ -156,9 +172,43 @@ def __getstate__(self):
...
-            "result",
         ]
-        return {key: self.__dict__[key] for key in keys if key in self.__dict__}
+        working_dictionary = {key: self.__dict__[key] for key in keys if key in self.__dict__}
+
+        if self.exception is not None:
+            working_dictionary["exception_type"] = type(self.exception).__name__
+            if self.exception.args[0] and isinstance(self.exception.args[0], str):
+                working_dictionary["exception_message"] = self.exception.args[0]
+
```
vkWeb: Following from my above recommendation:

We should be setting `self.exception` to `self.exception.__name__`.

```
if self.exception is not None and isinstance(self.exception, Exception):
    self.exception = type(self.exception).__name__
```

kolibri/core/tasks/job.py
```diff
@@ -185,15 +235,22 @@ def __init__(self, func, *args, **kwargs):
         if job_id is None:
             job_id = uuid.uuid4().hex
 
+        exception_type = kwargs.pop("exception_type", None)
+        if exception_type is None:
+            exception = None
+        else:
+            exception_message = kwargs.pop("exception_message", None)
+            exception = ErrorSavedWithJob(exception_message, exception_type)
+
```
vkWeb: Following from the above recommendation regarding not having a separate exception class: 

These all should be removed.

kolibri/core/tasks/job.py
```diff
@@ -185,15 +235,22 @@ def __init__(self, func, *args, **kwargs):
...
         self.job_id = job_id
         self.state = kwargs.pop("state", State.PENDING)
-        self.traceback = ""
-        self.exception = None
+        self.traceback = kwargs.pop("traceback", "")
+        self.exception = exception
```
vkWeb: `self.exception = kwargs.pop("exception", "")`.

kolibri/core/tasks/test/taskrunner/test_worker.py
```diff
@@ -67,7 +68,9 @@ def test_can_handle_unicode_exceptions(self, worker):
 
         returned_job = worker.storage.get_job(job_id)
         assert returned_job.state == "FAILED"
-        assert isinstance(returned_job.exception, TypeError)
+        assert isinstance(returned_job.exception, (TypeError, ErrorSavedWithJob))
+        if isinstance(returned_job.exception, ErrorSavedWithJob):
+            assert returned_job.exception.prior_type == "TypeError"
         assert returned_job.exception.args[0] == error_text
```
vkWeb: `assert returned_job.exception == "TypeError"`.

vkWeb (commented):
@EliKlein good job on this PR!  

I've written some thoughts on simplifying the approach. 
<hr>

The code that has been duplicated in two places will be best solved by having job `obj` in just one table `jobs` and the scheduling information on another table `scheduled_time`. A column on `jobs` table will foreign key into `scheduled_time` table if the job is SCHEDULED.

kolibri/core/analytics/utils.py
rtibbles: I think the issue here is that the JSON deserialization is not properly coercing objects back to their type - as we want to allow datetime objects to be serialized to JSON, we should be properly deserializing them from JSON too, and not passing them as string objects to functions that are expecting datetime objects.

We should be using this function to do the parsing: https://docs.djangoproject.com/en/1.11/ref/utils/#django.utils.dateparse.parse_datetime as that directly correlates with the DjangoJSONEncoder that is being used to dump the JSON in the first place.

We should be systematically deserializing the JSON to Python objects though, so that these implementation details are opaque to functions using this.

kolibri/core/tasks/exceptions.py
rtibbles: I think even simpler would be just to add a rule for serialization exceptions to JSON.

It would could store the import path for the Exception class, and any relevant details (note the traceback is important here). Then the job can reserialize and create the Exception class from the details in the JSON object.

kolibri/core/tasks/job.py
```diff
@@ -156,9 +172,43 @@ def __getstate__(self):
...
+        except TypeError as e:
+            # One or more of the keys is not JSON-serializable.
+            logger.debug("Job '{}' is not be saved: {}".format(self.job_id, str(e)))
+            string_result = None
+
+        return string_result
+
+    @classmethod
+    def from_json(cls, json_string):
+        working_dictionary = json.loads(json_string)
```
rtibbles, line 185: This is where we need to be doing more - we are using the DjangoJSONEncoder above to serialize some Python objects that otherwise will not be serialized by `json.dumps` - we should be doing the inverse here and properly deserializing everything.

Systematically extending `DjangoJSONEncoder` and an equivalent Python `JSONDecoder` object to handle these cases would allow this to be handled very systematically, and for most of the logic of this PR to be focused on this serialization and deserialization step - rather than having to reach into so many other parts of the code.

rtibbles (commented):
I think a lot of the complexity here could be reduced by focusing on consolidating serialization and deserialization logic.

I think the hybrid property could be removed, and we could just focus on having helper functions that the storage class uses for deserializing and serializing the job objects. For simplicty, that could just be methods on the job object, and we are sure to call those before setting data.

kolibri/core/analytics/utils.py
vkWeb: @rtibbles :100:.

EliKlein: If all of the serialization/deserialization needs to retain typings for things that can't be serialized (so far just dates and errors), does this mean we should have some sort of type-decoding dictionary, and an extra, reserved attribute in the serialized table that lists the type of each part that isn't a primitive?

I saw people warning against checking each value to see if it's a datetime because that's slow and can, in rare cases, yield false positives. Following that advice, there would be only two ways to do it; first, checking in each function that is expecting a datetime (or error); second, storing more data in the object, and using that data to decode any objects that were turned into just strings.

I'm kinda liking the idea of going back and doing the second one, it's just a bit harder initially.

EliKlein:
If I'm going to save exception types as module strings, I'd like to add `stringify_type` next to `stringify_func` and change `import_stringified_func` to just `import_stringified` since it's already not checking that the returned object is callable or anything.

Does that sound good? That was the end decision, for saving exceptions, right?

kolibri/core/analytics/utils.py
rtibbles: This is one reason why moving more of the Job data into separate ORMJob columns might be advantageous, as it would reduce the overhead for this kind of type checking.

If we had a top level `exception` column on the ORMJob table (that is nullable) then we don't need a general purpose parser to handle serialized `Exception` objects.

I think the main remaining issue then is that we use a `datetime` object as a function argument and we are expecting to be able to serialize/deserialize that. I think we could use a try/except to handle deserialization, using the Django utils parse datetime.

The Python JSONDecoder object allows passing `object_hooks` which try to do further deserialization, so, for example, we would try to coerce any strings to datetimes using the Django parse function, and just return the original string if it failed.

See the documentation here: https://docs.python.org/3/library/json.html#json.JSONDecoder and an example here: https://stackoverflow.com/a/10734224 (note they are using a different parsing function).

rtibbles:
> If I'm going to save exception types as module strings, I'd like to add stringify_type next to stringify_func and change import_stringified_func to just import_stringified since it's already not checking that the returned object is callable or anything.

Yes - this seems good. I've also added a comment that perhaps we can just move the exception to a top level column on the ORMJob table, so that we don't have to handle it generically when we are parsing - although if it's handled internally to the Job object, it may matter less.

kolibri/core/analytics/utils.py
vkWeb: @rtibbles I have a different opinion on this one.

I believe the task function should take care of parsing the datetime string. The reason being we can have in rare cases some false positives. I would say, we must establish that the arguments passed to the task function will all be valid-JSON types ONLY. So, then, it will be the job of the task function to take care of parsing the datetime string. 

This will be more performant and the task function will have 100% clarity on what's going on.

rtibbles: In that case, we should not use `DjangoJSONEncoder` and just use regular `json.dumps`, so as not to have any ambiguity. That means that the function initializing the pingback task would have to pre-serialize the argument.

I am fine with this if we want to be strict.

vkWeb: Yes `json.dumps` should be the way to go! :rocket:

vkWeb:
@rtibbles any specific reasons for creating exception objects out of their stringified form? I believe exception serves just one purpose: letting the client know what's been wrong. 

So, exception type as string and its full traceback as string should be what we all need...?

rtibbles:
> So, exception type as string and its full traceback as string should be what we all need...?

This could well be sufficient yes - I don't think we ever reraise the exception anywhere, so being able to recreate the exception may be too much.

Even just the traceback might be sufficient.

vkWeb:
@rtibbles yes we don't reraise it anywhere. So let us just keep it as string.

EliKlein:
One of the tests in `test_worker.py` ensures that errors are allowed to have unicode characters in their data. The test checks the message of the error (arg 0) to see that it is what's expected.

I've changed the test to just check that the expected message is within the traceback (since it will be at the end); but I thought I'd mention that it is (or was) "a" usage of the Exception as an object.

vkWeb:
@EliKlein yes - we do check whether the exception message can contain unicode chars or not but that doesn't feel to me like a use of exception as an object. 

The real use of exception as an object will be when we want to re-raise the exception after retrieving it from the database. Currently, we don't have such a use case so we are safe to serialize exception objects into strings.

EliKlein:
OK, This should be nearly done if not done.

vkWeb (approved):
@EliKlein Awesome work! :tada: 

There are just some linting issues that's failing the CI, can you please fix those? here's our docs on it: https://kolibri-dev.readthedocs.io/en/develop/getting_started.html#linting-and-auto-formatting.

@rtibbles apart from linting issues, this is LGTM from my side!

EliKlein:
Yarn doesn't seem to recognize the fmt-backend command for me. I see what it says it wants changed, I guess I can just do that?

rtibbles:
> Yarn doesn't seem to recognize the fmt-backend command for me. I see what it says it wants changed, I guess I can just do that?

Thanks - apologies, it seems our developer docs are out of date in this regard.

```
pre-commit run --all-files
```
is the easiest invocation.

kolibri/core/tasks/job.py
```diff
@@ -142,12 +143,26 @@ class Job(object):
     to the workers.
     """
 
-    def __getstate__(self):
+    def to_json(self):
+        """
+        Creates and returns a JSON-serialized string representing this Job.
+        This is how Job objects are persisted through restarts.
+
+        Note: Uses DjangoJSONEncoder to encode things. When using a type that's
```
rtibbles:
```suggestion
```

kolibri/core/tasks/job.py
```diff
@@ -142,12 +143,26 @@ class Job(object):
...
+        covered by it, the job's .func method needs to deal with that.
```
rtibbles:
```suggestion
```

kolibri/core/tasks/job.py
```diff
@@ -142,12 +143,26 @@ class Job(object):
...
+
```
rtibbles:
```suggestion
```

kolibri/core/tasks/job.py
```diff
@@ -156,9 +171,40 @@ def __getstate__(self):
             "args",
             "kwargs",
             "func",
-            "result",
```
rtibbles, line 159: We should save the result - it just should have to be JSON serializable too.

kolibri/core/tasks/job.py
```diff
@@ -156,9 +171,40 @@ def __getstate__(self):
...
         ]
-        return {key: self.__dict__[key] for key in keys if key in self.__dict__}
+
+        if isinstance(self.exception, Exception):
+            self.exception = type(self.exception).__name__
```
rtibbles: We shouldn't coerce the exception here - this is where the exception gets recorded into the ORM, so we should do the coersion there:

https://github.com/learningequality/kolibri/blob/6903afec0ca3ebdf65fbf0eb96ef5348ed9fc2d2/kolibri/core/tasks/storage.py#L243

kolibri/core/tasks/job.py
```diff
@@ -156,9 +171,40 @@ def __getstate__(self):
...
+
+        if self.result is not None:
```
rtibbles: See above.

kolibri/core/tasks/job.py
```diff
@@ -156,9 +171,40 @@ def __getstate__(self):
...
+
+        working_dictionary = {
+            key: self.__dict__[key] for key in keys if key in self.__dict__
+        }
+
+        try:
+            string_result = json.dumps(working_dictionary)
+        except TypeError as e:
+            # One or more of the keys is not JSON-serializable.
+            logger.debug("Job '{}' is not be saved: {}".format(self.job_id, str(e)))
```
rtibbles: I don't think we should fail silently here - we should raise an exception here to flag that this is not JSON serializable, we should not just carry on.

kolibri/core/tasks/job.py
rtibbles: These comments are no longer accurate, as we have dropped use of the DjangoJSONEncoder.

kolibri/core/tasks/job.py
```diff
@@ -142,12 +143,26 @@ class Job(object):
...
+    def to_json(self):
+        """
+        Creates and returns a JSON-serialized string representing this Job.
+        This is how Job objects are persisted through restarts.
+
+        Note: Uses DjangoJSONEncoder to encode things. When using a type that's
+        covered by it, the job's .func method needs to deal with that.
+
+        Information not saved:
+        - Results of the function, if it finished
```
rtibbles: This comment segment will need to be updated too.

kolibri/core/tasks/scheduler.py
```diff
@@ -172,16 +171,16 @@ def schedule(self, dt, func, interval=0, repeat=0, *args, **kwargs):
                 interval=interval,
                 repeat=repeat,
                 scheduled_time=naive_utc_datetime(dt),
-                obj=job,
```
rtibbles, line 175: Can just set `saved_job=job.to_json()` in here.

rtibbles (changes requested):
Blocking:
* Save `result`
* Don't fail silently, if a Job object doesn't save it's a big issue

Good to have:
* Don't coerce exception to string inside the Job object
* Update code comments.

rtibbles:
Thanks for all the work here @EliKlein - there's just a couple of things that I think are a blocker to merge, but this is very nearly there!

EliKlein:
I'm not running all of the tests here, but I ran a lot of them, including the most relevant sets, and they passed, so I think they'll probably all pass.

rtibbles (approved):

rtibbles:
Thank you @EliKlein - great work!

rtibbles:
If you would like to be added to our AUTHORS.md, let me know, I can update it (or feel free to make a follow up PR to do so!) https://github.com/learningequality/kolibri/blob/release-v0.15.x/AUTHORS.md

PR #8457, opened 2021-09-22
Remove precaching of Django static files.

rtibbles:
## Summary
* When we startup the Kolibri server, on my dev machine we spend about 0.4 seconds precaching all Django static files
* This removes that startup bottleneck by switching to a purely dynamic static file loading

## Reviewer guidance
This does not seem to affect static file serving performance, as after the first request, the file is cached just in the same way that it would have been previously.

Whitenoise does not do this by default, because it assumes it can wrap and serve any static file. We have modified whitenoise to only check for specific prefixes (like /static/) which reduces the search space for files.

Doing this change removes the startup bottleneck and seems to have no effect on siege performance for static files.
Before:
```
** SIEGE 4.0.4
** Preparing 25 concurrent users for battle.
The server is now under siege...
Lifting the server siege...
Transactions:		       44018 hits
Availability:		      100.00 %
Elapsed time:		       59.32 secs
Data transferred:	      217.70 MB
Response time:		        0.03 secs
Transaction rate:	      742.04 trans/sec
Throughput:		        3.67 MB/sec
Concurrency:		       24.94
Successful transactions:       44018
Longest transaction:	        0.19
Shortest transaction:	        0.01
```
After:
```
** SIEGE 4.0.4
** Preparing 25 concurrent users for battle.
The server is now under siege...
Lifting the server siege...
Transactions:		       44640 hits
Availability:		      100.00 %
Elapsed time:		       59.38 secs
Data transferred:	      220.78 MB
Response time:		        0.03 secs
Transaction rate:	      751.77 trans/sec
Throughput:		        3.72 MB/sec
Concurrency:		       24.94
Successful transactions:       44640
Longest transaction:	        0.13
Shortest transaction:	        0.02

```

----

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/deployment/default/wsgi.py
+++ b/kolibri/deployment/default/wsgi.py
@@ -15,7 +15,7 @@ from django.db.utils import OperationalError
 
 from kolibri.core.content.utils import paths
 from kolibri.utils import conf
-from kolibri.utils.kolibri_whitenoise import DjangoWhiteNoise
+from kolibri.utils.kolibri_whitenoise import DynamicWhiteNoise
 
 os.environ.setdefault(
     "DJANGO_SETTINGS_MODULE", "kolibri.deployment.default.settings.base"
@@ -30,7 +30,7 @@ def generate_wsgi_application():
     content_dirs = [paths.get_content_dir_path()] + paths.get_content_fallback_paths()
 
     # Mount static files
-    return DjangoWhiteNoise(
+    return DynamicWhiteNoise(
         django_application,
         static_prefix=settings.STATIC_URL,
         dynamic_locations=[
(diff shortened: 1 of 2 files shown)


jamalex (approved):
Looks good!

Not introduced here, but does this handle both the case where `static_prefix` ends with a trailing `/`, and where it doesn't?

rtibbles:
No - but that is disallowed by Django, so we should instead raise an error if it doesn't.

> It must end in a slash if set to a non-empty value.

https://docs.djangoproject.com/en/1.11/ref/settings/#std:setting-STATIC_URL

PR #8458, opened 2021-09-22
Renderer immersive styling updates

nucleogenesis:
## Summary

After seeing that the "estimated time" was going to be in the sidebar, I went and updated the styles to match what @jtamiace [shared on 8111](https://github.com/learningequality/kolibri/issues/8111#issuecomment-914703792) for the content renderers - lots of fiddly CSS stuff.

I hacked and slashed and need to update some test suites to account for the changes that I made to some of the structure.  Will get the tests passing ASAP.

| Content Type | Before | After | After Mobile |
|-|-|-|-
| EPUB | before-epub | epub-desktop-startedstatus | epub-mobile-just-completed-manually-with-modal |
| Exercise | before-exercise-rtl | exercise-desktop-coach-content-hints-shown |exercise-mobile-hints-shown |
| Exercise - Menu | exercise-context-menu-aligned-ltr | exercise-context-menu-aligned-rtl | No mobile - but just noting that the context menus are aligned in both RTL and LTR :) |
| HTML5 | before-html5 | html5-desktop-immersif | html5-mobile-not-optimized-for-so-small-methinks |
|PDF|before-pdf|pdf-desktop-so-immersive|pdf-mobile-immersive-yet-short|
| Video | before-video-rtl | video-desktop-just-completed-manually-with-modal | video-mobile-startedstatus|

## References

[The image in this comment](https://github.com/learningequality/kolibri/issues/8111#issuecomment-914703792)

## Reviewer guidance

This touches a lot of layout spacing things - so please try testing at various sizes on multiple devices if you can.

- General layout and spacing, alignment of top/bottom bars with content.
- Icon spacing & alignment
- Tooltip alignment & visibility
- Hints on exercises
- Test a bunch of Perseus exercises if you can - especially content NOT in the QA channel (the more diverse a selection of content the better please).

----

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated


kolibri/plugins/learn/assets/src/views/LearningActivityBar.vue
```diff
@@ -170,6 +183,12 @@
       };
     },
     computed: {
+      ...mapState({
```
MisRob: I'd suggest not adding dependencies to Vuex in smaller components that don't need it like this learning activity bar. I don't see a reason why not to use  `props` here that are made for that purpose, especially when the amount of information you're adding here is not huge nor complex. This will also simplify the test suite that will otherwise have tendencies to grow complicated even for very simple tests (you won't need `makeStore` and you won't need to know store structure to test a simple UI piece)

MisRob: Another good reason to avoid is that we're starting using Composition API and it'd be good to keep smaller building block components flexible enough to be able to work well no matter if their more complex parent components are built with Vuex or with Composition API.

nucleogenesis: Done and pushed - thanks for catching that Misha!

kolibri/core/assets/src/core-app/apiSpec.js
```diff
@@ -101,6 +101,9 @@ import InteractionList from '../views/InteractionList';
 import ExamReport from '../views/ExamReport';
 import TextTruncator from '../views/TextTruncator';
 import TimeDuration from '../views/TimeDuration';
+// TODO Migrate SuggestedTime into ../views/SuggestedTime. This is not to be done
```
rtibbles: We can do this! As long as the component name does not change, the string ids will be constant - and messages are extracted from any CSV file, not just the plugin specific one.

kolibri/plugins/perseus_viewer/assets/src/views/PerseusRendererIndex.vue
```diff
@@ -1,22 +1,16 @@
...
-        >
-          <div id="workarea" style="margin-left: 0px"></div>
-        </div>
-
+  <div
+    v-if="itemId || itemData"
+    class="bibliotron-exercise perseus-root"
+    :class="{ 'perseus-mobile': isMobile }"
+  >
+    <LessonMasteryBar
```
rtibbles, line 8: I _think_ this is the right place for this - although if it's something we would expose in, say, the QTI renderer too, then we might need to move it out of here in the future.

kolibri/plugins/pdf_viewer/assets/src/views/PdfRendererIndex.vue
```diff
@@ -407,15 +407,14 @@
...
   .pdf-renderer {
     @extend %momentum-scroll;
     @extend %dropshadow-2dp;
 
     position: relative;
-    height: 500px;
-    // This ensures that showing vs hiding the controls
-    // will not cover visible content below (ie, author name)
-    margin-bottom: $controls-height;
+    height: calc(100vh - #{$top-bar-height} - #{$controls-height});
```
rtibbles: Woot!

kolibri/plugins/learn/assets/src/views/classes/Hint.vue
```diff
@@ -1,59 +0,0 @@
-<template>
```
rtibbles, line 1: I wonder if we could just leave this as is, and instead remove the hint button inside the Perseus Renderer, and then just hook things up via props and events via the `LessonMasteryBar`?

That would leave us open for reusing this in say, the QTI renderer if needed.

rtibbles (commented):
Some suggestions for tweaks, but nothing blocking.

jtamiace:
I'm putting my comments in this deck because I don't want to overload this PR with screenshots. Hope this doesn't make things difficult to track @nucleogenesis - let me know if this isn't a helpful way to document requested changes. It's all looking great so far and feeling really nice to interact with 👏

https://docs.google.com/presentation/d/1eDQ4NFgD1suKOnAALWbV4L_3KX_VSQlHyDKFZRYD14M/edit#slide=id.p

nucleogenesis:
@jtamiace thank you for that! I went through and posted a few comments back to you with a few questions. Otherwise, I'll get started on updates.

kolibri/plugins/learn/assets/src/views/classes/Hint.vue
nucleogenesis: I'd rather not keep the cruft around in case it's maybe useful. The component isn't usable as-is and is only used in `LessonMasteryBar` so I don't see the need for a reusable component - particularly if the QTI Renderer ends up finding `LessonMasterBar` useful in its design, in which case the hint code will be reused as-is.

kolibri/plugins/perseus_viewer/assets/src/views/PerseusRendererIndex.vue
nucleogenesis: If we want it in the QTI renderer we can move it to Core and update the imports for sure if we keep that part of the design consistent between Perseus & QTI.

radinamatic:
The `View lesson resources` icon is not going to be permanent, correct?

I have no users, classes, lessons or quizzes on this facility (fresh one, viewing the resource from the Library as a super admin), so makes no sense to have the icon `View lesson resources` during the free content exploration... 🤔 

LEDev2104 (start)  Running  - Oracle VM VirtualBox_046

radinamatic:
Video transcript is not loading:

LEDev2104 (start)  Running  - Oracle VM VirtualBox_048

radinamatic:
Not strictly content renderer, but not sure of these missing padding on the Library page on tablet screen is related to the CSS changes...? (both FF & Chrome on Linux)

LEDev2104 (start)  Running  - Oracle VM VirtualBox_049


Idem for this jumbled sections while browsing channel on tablet screen 😕 

LEDev2104 (start)  Running  - Oracle VM VirtualBox_050

radinamatic:
Exercises generally look good, some glitches on smaller mobile resolutions:

1 | 2
--- | ---
exercise2 | exercise

radinamatic:
Btw, I was not able to access the above server running on my dev VM from any other device in my local network 🤔 

`Uncaught ReferenceError: kolibriCoreAppGlobal is not defined`

Kolibri — Mozilla Firefox_003

nucleogenesis:
> Not strictly content renderer, but not sure of these missing padding on the Library page on tablet screen is related to the CSS changes...? (both FF & Chrome on Linux)

This section isn't in scope for the PR - I think that @marcellamaki or @MisRob are working on those issues elsewhere (if not then they're upcoming).

radinamatic:
One last thing regarding the `Save to bookmark` button, @nucleogenesis, that I did not mention because I was not sure if it happens because the feature is not fully implemented: resource is correctly saved under Bookmarked when the button is pressed, but that _pressed_ state seems to stick even when the user opens another resource, so when the user hovers over the button, the label now says `Remove from bookmarks`, even though that second resource was never saved before.

marcellamaki:
@radinamatic - good eye on both the library page layout and the "save to bookmarks" button! I'm including fixes for those in two upcoming PRs and will definitely 👋 at you for testing them then to be sure they are good to go

rtibbles:
This bookmark stickiness was probably because of an incorrect assumption I had made about the internal state of the top bar being torn down between rendering content items, so the simplest fix would be to clear the bookmark data when the content being displayed is updated.

pcenov:
@radinamatic happy to report that I tested a number of resources in IE11 and also with RTL languages and did not stumble on any new issues other than the ones already reported. :) Noting that it's currently not possible to view any of the assigned resources when being logged in as a learner, so I was testing only at http://127.0.0.1:8080/en/learn/#/library

nucleogenesis:
Everything mentioned here (and to me on Slack by @jtamiace ) are resolved in my last commit (or I hope they are anyway).

Feedback points needed:

- isLessonContext - I didn't dig into "how do we know this" yet but have guesses. Not sure this is in scope here so we can follow up on it.
- I manually mess w/ some styling on the HTML tag in PDF renderer and it's hacky but works - so if you could just try to break a PDF that'd be awesome.


Separately, my "tool  tip stays on the page" fix broke them elsewhere so I removed it :( KTooltip needs some TLC in general

nucleogenesis:
Also, hints now will look decent in any language:

image

image

radinamatic:
@pcenov Could you give one more run with the latest asset from this PR to check responsiveness on the PDF and Perseus exercises, before we merge? Thank you!

pcenov:
Hi @radinamatic, here are the only 2 issues I was able to find regarding the PDFs and Perseus exercises
1. The menu with more options appears under the full screen header of the PDF files and the tooltip 'More options' should not be displayed in mobile as it remains visible as long as the menu is displayed.
2. The text of the tooltip message which is shown after tapping the info icon for the hint is never displayed in full size regardless of the resolution.

| PDF | Exercise
| --- | ---
|   |

nucleogenesis:
@pcenov thank you! Great catch on the PDF bar overlaying the menu!

I've fixed that problem and have finally found a way to hack that tooltip into place without messing anything else up.

https://user-images.githubusercontent.com/6356129/136110339-ec66efa5-25ae-4c0d-8f2c-2df3b7a7af2e.mp4

As for the tooltip sticking on mobile and when the menu is opened - I'm not sure the solution here and I don't think fixing it is in scope for this PR. The KTooltip component needs to be able to show on mobile because it is used in our info icon which can be clicked to show a tooltip elsewhere.

I won't say this is the best solution - but perhaps we could add a prop to KTooltip which says "don'tShowOnMobile" or something so that we can give devs that option where we will explicitly not show the tooltip in a mobile (or touch, probably) context.

That could be followed up on in Kolibri Design System the best, I think.

pcenov:
@nucleogenesis thank you as well - I've retested and verified that both issues are now fixed. Awesome!

radinamatic (approved):
Manual testing successful and passing, great job! 👍🏽 💯

Issue #8454, opened 2021-09-21
updates to naming and versioning of internal packages and tools


indirectlylit:
# Status

This issue is currently a draft and open for feedback.

# Current state

The Kolibri product ecosystem has a number of "internal" Node packages. These are currently tracked monorepo-style in the Kolibri repo, and some are used by other products. For example, [Studio](https://github.com/learningequality/studio/blob/92ef5585d9dd4f61d060c01437e233674258cad6/package.json#L84) and [KDS](https://github.com/learningequality/kolibri-design-system/blob/c078ef08d41ad4eb77795dc5a1c8b729918e7b0e/package.json#L44) both use `kolibri-tools`.

These packages are currently defined within the [`packages`](https://github.com/learningequality/kolibri/tree/a7d136171181893b394b89425c4ff6dcdd00403a/packages) directory:

| package name | description |
|--|--|
| [`kolibri-tools`](https://www.npmjs.com/package/kolibri-tools) | Utilities for building Kolibri plugins; (2) Linting code in Kolibri products; (3) i18n string handling |
| [`browserslist-config-kolibri`](https://www.npmjs.com/package/browserslist-config-kolibri) | Supported browsers config used by `kolibri-tools` |
| [`eslint-plugin-kolibri`](https://www.npmjs.com/package/eslint-plugin-kolibri) | Custom eslint rules used by `kolibri-tools` |
| `kolibri-core-for-export` a.k.a. [`kolibri`](https://www.npmjs.com/package/kolibri) | Exposes importable interfaces for `kolibri-tools` |
| `hashi` (unpublished) |  Untrusted JS sandbox with communication API  |

These are all implemented as [Yarn 1.0 workspaces](https://classic.yarnpkg.com/lang/en/docs/workspaces/), along with `kolibri/core` and `kolibri/plugins/*`:

https://github.com/learningequality/kolibri/blob/a7d136171181893b394b89425c4ff6dcdd00403a/package.json#L5-L9

Currently, all of these plugins except for Hashi are supposed track Kolibri's version number, and get re-published to NPM on a regular basis. This occurs with the `packages/publish.js` tool. In practice, this process occurs rarely and is not documented. Currently all NPM dependencies are `dev` versions.

Some additional notes:

* Kolibri depends on `kolibri-tools` as a workspace-based package using symlinks
* Kolibri depends on `hashi` both as a workspace-based package and also through custom webpack build tooling in hashi that targets Kolibri's static folder as the output directory
[...]

PR #8459, opened 2021-09-23
Upgrades KDS and dependencies

indirectlylit:
## Summary

* Upgrade vue from 2.6.11 -> 2.6.14 in line with KDS
* Remove counterproductive `kolibri-tools` references which were pinning to old versions rather than using yarn workspace symlinks
* Allow browserlist-config-kolibri to be published
* Upgrade private packages to 0.15.0-dev.4 (see follow-up issue #8454)
* Upgrade KDS to v1.0.0

## References

* https://github.com/learningequality/kolibri-design-system/releases/tag/v1.0.0
* supersedes #8085

## Reviewer guidance

Sanity-check to make sure things run as well as they did prior to this commit

----

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated


packages/kolibri-tools/package.json
```diff
@@ -101,9 +101,6 @@
     "node": "10.x",
     "npm": ">= 3"
   },
-  "optionalDependencies": {
-    "kolibri": "0.13.0-dev.10"
```
rtibbles, line 105: I think given that we want to get rid of this package, this is probably OK - but just being aware that outside of the `packages` workspace context, I can imagine this being an issue, as there's no indication that this should be installed.

rtibbles (changes requested):
Should clean up the code that does this version setting logic as well as the versions themselves:
https://github.com/learningequality/kolibri/blob/release-v0.15.x/packages/publish.js#L57

Aside from this, the asset has built fine and runs as expected.

packages/kolibri-tools/package.json
indirectlylit: why is it optional?

rtibbles: Because it's only needed for running tests, not for linting or building plugins.

indirectlylit:
updated with cleanup to _publish.js_

rtibbles (approved):

PR #5051, opened 2019-02-13
Add empty states to various coach tables and widgets

jonboiser:
### Summary

1. Extends CoreTable to accept an 'emptyMessage' prop, which it renders in a `<p>` depending on whether it detects any children in its tbody slot.
1. Updates all of the /coach reports pages that use CoreTable to show an appropriate empty message. (Non-coach pages using CoreTable were not updated)
1. Updates other list widgets to also show an empty message
1. Fixes a few tables that was using a "title" header under a column for user names.
1. Vertically align the table cells in CoachExamsPage Fixes #5046 (unless the misalignment happens anywhere else).
1. Fixes Lesson Status Icon always being active Fixes #5056
1. Remove the second Needs Help indicator on ItemProgress widget. Fixes #5059

screen shot 2019-02-13 at 2 20 35 pm

### Reviewer guidance

1. To manually test, go to as many coach pages as possible and make sure the empty message appears when there is no data, and _does not appear_ when there _is_ data.
1. In the code, make sure any of use `$tr('...')` looks correct (i.e. always has `coachStrings.$tr` and coachstrings is in scope.

### References

Fixes #5022
----

### Contributor Checklist

PR process:

- [x] PR has the correct target branch and milestone
- [x] PR has 'needs review' or 'work-in-progress' label
- [x] If PR is ready for review, a reviewer has been added. (Don't use 'Assignees')
- [ ] If this is an important user-facing change, PR or related issue has a 'changelog' label
- [ ] If this includes an internal dependency change, a link to the diff is provided

Testing:

- [ ] Contributor has fully tested the PR manually
- [ ] If there are any front-end changes, before/after screenshots are included
- [ ] Critical user journeys are covered by Gherkin stories
- [ ] Critical and brittle code paths are covered by unit tests

### Reviewer Checklist

- Automated test coverage is satisfactory
- PR is fully functional
- PR has been tested for [accessibility regressions](http://kolibri-dev.readthedocs.io/en/develop/manual_testing.html#accessibility-a11y-testing)
- External dependency files were updated if necessary (`yarn` and `pip`)
- Documentation is updated
- Contributor is in AUTHORS.rst

PR #8461, opened 2021-09-23
Refactor bookmarks cards

marcellamaki:
## Summary

This PR updates some of UI element conflicts that surfaced during string freeze on the Learner bookmarks page, with the main changes being: 
- The cards used on the bookmarks page are now responsive, with mobile-specific layouts
- Slightly updated "no bookmarks" page
- Basic tests added for bookmarking

Notes (out of scope)
- The "remove from bookmarks" feature works, but the "view information" to open the side panel is not included in this PR since that work is still in progress.
- After refactoring, I have not applied all of the card changes to the channel browsing. It doesn't break things horribly, but that work is also not _included_ as part of this PR

## References

[Figma specs](https://www.figma.com/file/uiZuCIqh2KYvBUfbCiJyyA/Hybrid-Learning?node-id=76%3A6407)

### No bookmarks 
Screen Shot 2021-09-23 at 10 53 33 AM

### Desktop view
Screen Shot 2021-09-23 at 11 02 26 AM

**Note: the channel thumbnail is going to be added as new metadata, so it is not visible yet and you just see the alt text (and relatedly, the channel title is not yet added, so the title is not being added to the alt text as of yet)**

### Various mobile previews (Chrome dev tools)
Screen Shot 2021-09-23 at 11 02 56 AM
Screen Shot 2021-09-23 at 11 02 45 AM

…

## Reviewer guidance

- As these are mostly UI changes, there are not many interactions to test (other than removing a bookmark)
- I would most appreciate feedback in terms of making the code more readable or suggestions of how to refactor or simplify some of it
- I'd also love feedback about if I am using KDS theming correctly, and if there are places where I should be using theme tokens where I have hardcoded values (just because...I am not sure if I know what all of the tokens I might use are, honestly)
- Relatedly, some places where it might be useful to have slightly more dynamic css and fewer specific pixel values (not KDS, but maybe moving some of this into styles={} or setting values in computed. 
- How does this look from a design point of view? What adjustments are still needed? (I am most specifically noting the mobile thumbnails, but am planning to swap out with @MisRob's work on the new <Thumbnail/> if that seems like a better option) 
- I updated some of the values in `ContentCard/card.scss` - is this going to cause a conflict elsewhere?
- Test suggestions/guidance always very welcome 🙏 
----

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/plugins/learn/assets/src/views/BookmarkPage.vue
+++ b/kolibri/plugins/learn/assets/src/views/BookmarkPage.vue
@@ -23,6 +23,7 @@
 
       <KButton
         v-if="more && !loading"
+        data-test="load-more-button"
         :text="coreString('viewMoreAction')"
         @click="loadMore"
       />
--- a/kolibri/plugins/learn/assets/src/views/BookmarkPage.vue
+++ b/kolibri/plugins/learn/assets/src/views/BookmarkPage.vue
@@ -2,43 +2,23 @@
 
   <div>
 
-    <h3>{{ $tr('bookmarksHeader') }}</h3>
-
-    <p v-if="!bookmarks.length && !loading">
+    <p v-if="!bookmarks.length && !loading" class="no-bookmarks">
       {{ $tr('noBookmarks') }}
     </p>
 
+
     <template v-else>
-      <ContentCard
-        v-for="(content, index) in bookmarks"
-        :key="content.id"
-        class="card"
-        :isMobile="windowIsSmall"
-        :title="content.title"
-        :thumbnail="content.thumbnail"
-        :kind="content.kind"
-        :isLeaf="content.is_leaf"
-        :progress="content.progress || 0"
-        :numCoachContents="content.num_coach_contents"
-        :link="genContentLink(content.id, content.is_leaf)"
-        :contentId="content.content_id"
-        :subtitle="bookmarkCreated(content.bookmark.created)"
-      >
-        <template #actions>
-          <KIconButton
-            icon="info"
-            :tooltip="infoText"
-            :ariaLabel="infoText"
-            @click="showResourcePanel(content)"
-          />
-          <KIconButton
-            icon="clear"
-            :tooltip="removeText"
-            :ariaLabel="removeText"
-            @click="removeBookmark(content, index)"
-          />
-        </template>
-      </ContentCard>
+      <h1 class="bookmarks-header">
+        {{ $tr('bookmarksHeader') }}
+      </h1>
+      <ContentCardGroupGrid
+        v-if="bookmarks.length"
+        :contents="bookmarks"
+        :genContentLink="genContentLink"
+        :cardViewStyle="windowIsSmall ? 'card' : 'list'"
+        numCols="1"
+        :footerIcons="footerIcons"
+      />
 
       <KButton
         v-if="more && !loading"
@@ -49,7 +29,6 @@
         v-else-if="loading"
         :delay="false"
       />
-
     </template>
   </div>
 
@@ -58,15 +37,15 @@
 
 <script>
 
-  import { mapActions } from 'vuex';
-  import client from 'kolibri.client';
-  import urls from 'kolibri.urls';
+  // import { mapActions } from 'vuex';
+  // import client from 'kolibri.client';
+  // import urls from 'kolibri.urls';
   import commonCoreStrings from 'kolibri.coreVue.mixins.commonCoreStrings';
   import responsiveWindowMixin from 'kolibri.coreVue.mixins.responsiveWindowMixin';
   import { ContentNodeResource } from 'kolibri.resources';
-  import { now } from 'kolibri.utils.serverClock';
   import genContentLink from '../utils/genContentLink';
-  import ContentCard from './ContentCard';
+  import { normalizeContentNode } from '../modules/coreLearn/utils.js';
+  import ContentCardGroupGrid from './ContentCardGroupGrid';
 
   export default {
     name: 'BookmarkPage',
@@ -76,7 +55,7 @@
       };
     },
     components: {
-      ContentCard,
+      ContentCardGroupGrid,
     },
     mixins: [commonCoreStrings, responsiveWindowMixin],
     data() {
@@ -84,32 +63,19 @@
         loading: true,
         bookmarks: [],
         more: null,
-        resourcePanelNode: null,
-        now: now(),
       };
     },
     computed: {
-      infoText() {
-        return this.coreString('viewInformation');
-      },
-      removeText() {
-        return this.coreString('removeFromBookmarks');
+      footerIcons() {
+        return { close: 'removeFromBookmarks', info: 'viewInformation' };
       },
     },
     created() {
       ContentNodeResource.fetchBookmarks({ params: { limit: 25 } }).then(data => {
         this.more = data.more;
-        this.bookmarks = data.results;
+        this.bookmarks = data.results.map(normalizeContentNode);
         this.loading = false;
       });
-    },
-    mounted() {
-      this.timer = setInterval(() => {
-        this.now = now();
-      }, 10000);
-    },
-    beforeDestroy() {
-      clearInterval(this.timer);
     },
     methods: {
       ...mapActions(['createSnackbar']),
@@ -119,27 +85,20 @@
           this.loading = true;
           ContentNodeResource.fetchBookmarks({ params: this.more }).then(data => {
             this.more = data.more;
-            this.bookmarks.push(...data.results);
+            this.bookmarks.push(...data.results.map(normalizeContentNode));
             this.loading = false;
           });
         }
       },
-      removeBookmark(bookmark, index) {
-        client({
-          method: 'delete',
-          url: urls['kolibri:core:bookmarks_delete_by_node_id'](bookmark.id),
-        }).then(() => {
-          this.bookmarks.pop(index);
-          this.createSnackbar(this.$tr('removedNotification'));
-        });
-      },
-      showResourcePanel(bookmark) {
-        this.resourcePanelNode = bookmark;
-      },
-      bookmarkCreated(date) {
-        const time = this.$formatRelative(date, { now: this.now });
-        return this.coreString('bookmarkedTimeAgoLabel', { time });
-      },
+      // removeBookmark(bookmark, index) {
+      //   client({
+      //     method: 'delete',
+      //     url: urls['kolibri:core:bookmarks_delete_by_node_id'](bookmark.id),
+      //   }).then(() => {
+      //     this.bookmarks.pop(index);
+      //     this.createSnackbar(this.$tr('removedNotification'));
+      //   });
+      // },
     },
     $trs: {
       bookmarksHeader: {
@@ -147,10 +106,10 @@
         context:
           "Header on the 'Bookmarks' page with the list of the resources user previously saved.",
       },
-      removedNotification: {
-        message: 'Removed from bookmarks',
-        context: 'Message indicating that a resource has been removed from the Bookmarks page.',
-      },
+      // removedNotification: {
+      //   message: 'Removed from bookmarks',
+      //   context: 'Message indicating that a resource has been removed from the Bookmarks page.',
+      // },
       noBookmarks: {
         message: 'You have no bookmarked resources',
         context: "Status message in the 'Bookmarks' page when user did not bookmark any resources.",
@@ -163,8 +122,19 @@
 
 <style lang="scss" scoped>
 
-  .card {
-    margin: 5px;
+  .no-bookmarks {
+    width: 100%;
+    height: 100%;
+    padding-top: 170px;
+    text-align: center;
+  }
+
+  .bookmarks-header {
+    margin-bottom: 34px;
+    font-size: 24px;
+    font-style: normal;
+    font-weight: normal;
+    line-height: 33px;
   }
 
 </style>
--- a/kolibri/plugins/learn/assets/src/views/BookmarkPage.vue
+++ b/kolibri/plugins/learn/assets/src/views/BookmarkPage.vue
@@ -1,37 +1,33 @@
 <template>
 
   <div>
-
-    <p v-if="!bookmarks.length && !loading" class="no-bookmarks">
+    <h1>
+      {{ $tr('bookmarksHeader') }}
+    </h1>
+    <p v-if="!bookmarks.length && !loading">
       {{ $tr('noBookmarks') }}
     </p>
 
+    <HybridLearningCardGrid
+      v-if="bookmarks.length"
+      :contents="bookmarks"
+      :genContentLink="genContentLink"
+      :cardViewStyle="windowIsSmall ? 'card' : 'list'"
+      numCols="1"
+      :footerIcons="footerIcons"
+      @removeFromBookmarks="removeFromBookmarks"
+    />
 
-    <template v-else>
-      <h1 class="bookmarks-header">
-        {{ $tr('bookmarksHeader') }}
-      </h1>
-      <ContentCardGroupGrid
-        v-if="bookmarks.length"
-        :contents="bookmarks"
-        :genContentLink="genContentLink"
-        :cardViewStyle="windowIsSmall ? 'card' : 'list'"
-        numCols="1"
-        :footerIcons="footerIcons"
-        @removeFromBookmarks="removeFromBookmarks"
-      />
-
-      <KButton
-        v-if="more && !loading"
-        data-test="load-more-button"
-        :text="coreString('viewMoreAction')"
-        @click="loadMore"
-      />
-      <KCircularLoader
-        v-else-if="loading"
-        :delay="false"
-      />
-    </template>
+    <KButton
+      v-if="more && !loading"
+      data-test="load-more-button"
+      :text="coreString('viewMoreAction')"
+      @click="loadMore"
+    />
+    <KCircularLoader
+      v-else-if="loading"
+      :delay="false"
+    />
   </div>
 
 </template>
@@ -48,7 +44,7 @@
   import urls from 'kolibri.urls';
   import { PageNames } from '../constants';
   import { normalizeContentNode } from '../modules/coreLearn/utils.js';
-  import ContentCardGroupGrid from './ContentCardGroupGrid';
+  import HybridLearningCardGrid from './HybridLearningCardGrid';
 
   export default {
     name: 'BookmarkPage',
@@ -58,7 +54,7 @@
       };
     },
     components: {
-      ContentCardGroupGrid,
+      HybridLearningCardGrid,
     },
     mixins: [commonCoreStrings, responsiveWindowMixin],
     data() {
@@ -70,13 +66,13 @@
     },
     computed: {
       footerIcons() {
-        return { close: 'removeFromBookmarks', info: 'viewInformation' };
+        return { info: 'viewInformation', close: 'removeFromBookmarks' };
       },
     },
     created() {
       ContentNodeResource.fetchBookmarks({ params: { limit: 25 } }).then(data => {
         this.more = data.more;
-        this.bookmarks = data.results.map(normalizeContentNode);
+        this.bookmarks = data.results ? data.results.map(normalizeContentNode) : [];
         this.loading = false;
       });
     },
@@ -123,23 +119,3 @@
   };
 
 </script>
-
-
-<style lang="scss" scoped>
-
-  .no-bookmarks {
-    width: 100%;
-    height: 100%;
-    padding-top: 170px;
-    text-align: center;
-  }
-
-  .bookmarks-header {
-    margin-bottom: 34px;
-    font-size: 24px;
-    font-style: normal;
-    font-weight: normal;
-    line-height: 33px;
-  }
-
-</style>
--- a/kolibri/plugins/learn/assets/src/views/BookmarkPage.vue
+++ b/kolibri/plugins/learn/assets/src/views/BookmarkPage.vue
@@ -89,13 +89,13 @@
           });
         }
       },
-      removeFromBookmarks(bookmark, index) {
+      removeFromBookmarks(bookmark) {
         if (bookmark) {
           client({
             method: 'delete',
             url: urls['kolibri:core:bookmarks_detail'](bookmark.id),
           }).then(() => {
-            this.bookmarks.pop(index);
+            this.bookmarks = this.bookmarks.filter(bm => bm.bookmark.id !== bookmark.id);
             this.createSnackbar(this.$tr('removedNotification'));
           });
         }
--- a/kolibri/plugins/learn/assets/src/views/BookmarkPage.vue
+++ b/kolibri/plugins/learn/assets/src/views/BookmarkPage.vue
@@ -18,6 +18,7 @@
         :cardViewStyle="windowIsSmall ? 'card' : 'list'"
         numCols="1"
         :footerIcons="footerIcons"
+        @removeFromBookmarks="removeFromBookmarks"
       />
 
       <KButton
@@ -37,13 +38,14 @@
 
 <script>
 
-  // import { mapActions } from 'vuex';
-  // import client from 'kolibri.client';
-  // import urls from 'kolibri.urls';
+  import { mapActions } from 'vuex';
   import commonCoreStrings from 'kolibri.coreVue.mixins.commonCoreStrings';
   import responsiveWindowMixin from 'kolibri.coreVue.mixins.responsiveWindowMixin';
   import { ContentNodeResource } from 'kolibri.resources';
   import genContentLink from '../utils/genContentLink';
+  import client from 'kolibri.client';
+  import urls from 'kolibri.urls';
+  import { PageNames } from '../constants';
   import { normalizeContentNode } from '../modules/coreLearn/utils.js';
   import ContentCardGroupGrid from './ContentCardGroupGrid';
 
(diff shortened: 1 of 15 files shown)


indirectlylit (commented):
Will review the code, but first two quick design questions - cc @jtamiace 

## Icon buttons

It's rare in Kolibri (versus Studio) to use purely icon buttons rather than text-based so I wanted to make sure this was intentional. This is partly because learners on tablets don't have tooltips, and partly because learners are less likely to know iconography conventions in contrast to Studio where we use icon buttons more liberally for "power users".

| deletion in bookmarks | deletion in coach UI |
|--|--|
| image | image |


If we do stick with icon buttons, note that we might be using the wrong one for 'delete':

* https://design-system.learningequality.org/icons/#token-clear
* https://design-system.learningequality.org/icons/#token-delete

## Empty states

We've made 'empty states' more consistent, and I think this design deviates slightly from previous directions:

* https://github.com/learningequality/kolibri/pull/5051
* https://github.com/learningequality/kolibri/pull/3550

I would have expected it to look more like:

image

Perhaps something to track in KDS?

jtamiace:
We should probably go with your suggestion for the empty state.

I'm not too worried about the icon button for removing the bookmark because it's used in the Learn interface for closing out of content and quiz renderers, so hopefully already an association with "making this go away." There's also a way to confirm what it does when they interact with it, and an undo action that if it was unintentional. 

Screen Shot 2021-09-24 at 1 58 23 PM

kolibri/plugins/learn/assets/src/views/BookmarkPage.vue
```diff
@@ -76,41 +57,28 @@
...
-        return this.coreString('removeFromBookmarks');
+      footerIcons() {
+        return { close: 'removeFromBookmarks', info: 'viewInformation' };
       },
     },
     created() {
       ContentNodeResource.fetchBookmarks({ params: { limit: 25 } }).then(data => {
         this.more = data.more;
-        this.bookmarks = data.results;
+        this.bookmarks = data.results.map(normalizeContentNode);
```
MisRob, line 78: Although it's possible that API returns an empty array in `data.results` and therefore `map` won't fail even when there are no bookmarks, I'd suggest adding a guard to check if `data.results` are truthy before trying to use `map` just in case it was `undefined` or `null`, just to be sure. Something like

```
this.bookmarks = data.results ? data.results.map(normalizeContentNode) : []
```

MisRob: Although it's possible that API returns an empty array in `data.results` and therefore `map` won't fail even when there are no bookmarks, I'd suggest adding a guard to check if `data.results` are truthy before trying to use `map` just in case it was `undefined` or `null`, just to be sure. Something like

```
this.bookmarks = data.results ? data.results.map(normalizeContentNode) : []
```

kolibri/plugins/learn/assets/src/views/BookmarkPage.vue
```diff
@@ -169,8 +132,19 @@
...
-  .card {
-    margin: 5px;
+  .no-bookmarks {
+    width: 100%;
+    height: 100%;
+    padding-top: 170px;
+    text-align: center;
+  }
+
+  .bookmarks-header {
```
MisRob: I am not sure how this would look like here on this page and am using this style only as a good example so @marcellamaki no changes may be needed here if results are not acceptable at this point.

Generally speaking, I'd rather use `h1` as is without this style even though it might mean that it won't be pixel perfect, and then eventually update our global styles for fonts, maybe even using a system based on `rem` units. For this reason, I've been trying to avoid updating font styles in hybrid learning work unless it looked really necessary. I am afraid that otherwise, this approach would lead to styling titles and texts on each page separately and will tend to end up with many inconsistencies across the whole app.

My suggestion would be to ignore tweaking fonts altogether (unless something looks very horrible without doing so) and having a GitHub issue for texts styles. I am not sure though about how our current system works and if changes to it could lead to problems in parts of the app that don't have this new hybrid learning look so we could also use that issue to think through this.

@indirectlylit @jtamiace any thoughts?

MisRob:
Are the left and right margins on one of your screenshots intentional?
Screenshot from 2021-09-27 07-05-37

Designs have the same left and right margin as the top margin (24px): https://www.figma.com/file/uiZuCIqh2KYvBUfbCiJyyA/Hybrid-Learning?node-id=130%3A11752

MisRob:
> I updated some of the values in ContentCard/card.scss - is this going to cause a conflict elsewhere?

I'd say that no one can say for sure so I'd suggest checking all places where `ContentCard` is used to see if there is some before/after difference. If not, no problem I guess :) If yes, then it's something to be included in the PR description (that's one of the purposes of _"If there are any front-end changes, before/after screenshots are included"_ in the PR checklist - it's not only for screenshots of new features but also for checking regressions)

If there are some regressions, we'll need to have a look if that particular part of the app is going to be refactored into hybrid learning overcoat before the next release or no and based on that, decide if we need to adjust `ContentCard` with some props or even having two separate cards components.

> (I am most specifically noting the mobile thumbnails, but am planning to swap out with @MisRob's work on the new if that seems like a better option)

I'd say this is the same question. It'd be ideal to use (and adjust if needed) our new thumbnails that were made for hybrid learning in general (`learn/assets/src/views/thumbnails`), however, I am not sure if it's welcome in all places where `ContentCardListViewItem` is currently used. If such an update shows to be complicated, I think it'd be fine to stay with `CardThumbnail` and follow up on thumbnails later. Seems to be part of a larger issue of having many cards and some of them somewhere in the middle of a migration to hybrid learning, so I wouldn't complicate it now.

kolibri/plugins/learn/assets/src/views/ContentCardGroupGrid.vue
```diff
@@ -1,33 +1,38 @@
...
     <ContentCardListViewItem
       v-for="content in contents"
-      v-else
       :key="content.id"
-      :channelThumbnail="setChannelThumbnail(content)"
-      :channelTitle="channelTitle(content)"
+      :channelThumbnail="content.channel_thumbnail"
+      :channelTitle="content.channel_thumbnail"
       :description="content.description"
+      activityLength="shortActivity"
```
MisRob: Is this a temporary value?

Regarding activity length, there is[ some logic prepared in `LearningActivityLabel` ](https://github.com/learningequality/kolibri/blob/develop/kolibri/plugins/learn/assets/src/views/cards/ResourceCard/LearningActivityLabel.vue#L106). So far it's been used to show duration or its textual representation on the home page. If it'd be useful, feel free to move some of this code from the component somewhere higher in the structure and use it for bookmarks.

kolibri/plugins/learn/assets/src/views/ContentCardListViewItem.vue
```diff
@@ -182,13 +231,39 @@
...
+        return this.pageName === PageNames.LIBRARY;
+      },
+      ceilingDate() {
+        if (this.createdDate > this.now) {
+          return this.now;
+        }
+        return this.createdDate;
+      },
+      bookmarkCreated() {
+        const time = this.$formatRelative(this.ceilingDate, { now: this.now });
```
MisRob: Will the time be correct if I have a page with this component open for a couple of minutes and then bookmark something? I haven't tested this myself yet though from what I see in code my guess would be that right now `this.now` would rather return time when this component was mounted and not necessarily when a bookmark was really created.

kolibri/plugins/learn/assets/test/bookmark-page.spec.js
```diff
@@ -0,0 +1,51 @@
...
+
+  it('smoke test', () => {
+    const wrapper = shallowMount(BookmarkPage);
+    expect(createdSpy).toHaveBeenCalled();
+    expect(wrapper.exists()).toBe(true);
+  });
+  describe('When the user clicks the remove from bookmarks icon', () => {
+    it('will make a call to remove the bookmark from the list of bookmarks', () => {
+      wrapper.findComponent({ name: 'ContentCardGroupGrid' }).vm.$emit('removeFromBookmarks');
+      expect(removeFromBookmarksSpy).toHaveBeenCalled();
```
MisRob: This doesn't test if a call was made. If there was an error in `removeFromBookmarks` method, this test would still pass even though the call wasn't made. That's one of the reasons why it's usually not recommended to test internal methods of components.

I'd recommend mocking `client` and checking that it's been called with correct `method` and `url` parameters.

kolibri/plugins/learn/assets/test/bookmark-page.spec.js
```diff
@@ -0,0 +1,51 @@
...
+    });
+  });
+
+  describe('When there are more bookmarks than in the default display', () => {
+    it('displays a load more button', () => {
+      expect(wrapper.find("[data-test='load-more-button']")).toBeTruthy();
+    });
+    it('clicking the load more button calls the load more function', () => {
+      wrapper.find("[data-test='load-more-button']").vm.$emit('click');
+      expect(loadMoreSpy).toHaveBeenCalled();
```
MisRob: Similar to the above, checking that some method inside of a component has been called doesn't say much about if things actually work. I'd recommend checking that `fetchBookmarks` on `ContentNodeResource` has been called with correct `params`.

kolibri/plugins/learn/assets/src/views/BookmarkPage.vue
jtamiace: I agree - didn't realize the header was a different style than other headers in the app. Plain h1 would be better for consistency, and I don't see a reason why the header on this page needs to be different than the others.

kolibri/plugins/learn/assets/src/views/ContentCardListViewItem.vue
marcellamaki: This card component is used for the bookmarks page itself, and so I am trying to figure out a way where it might be mounted and then have a bookmark created after without a re-navigation. I have tried testing this and so far it seems that the bookmarks time is reflecting accurately. But if you have a specific scenario in mind (or if someone else has one!), I would be happy to give it a try!

marcellamaki:
@jtamiace - I am wondering about how to proceed with the thumbnails and card spacing from a design point of view. I'll include a few screenshots below, and please let me know if you have questions. 

With the thumbnails, since we have a variety of image ratios, there are some constraints that are used to contain the image (i.e., it shrinks to fit its boundaries, rather than being cropped). This is less noticeable with the list card view, but the top margin of the card _is_ offset, see below:

Screen Shot 2021-09-28 at 10 52 10 AM

(full image container background - and alignment - visible in grey)

Screen Shot 2021-09-28 at 10 51 57 AM


But by keeping the width fixed, the difference in height becomes more noticeable on mobile especially. I can fill the background of the image's container as a light grey (which is something @MisRob has implemented with her new thumbnails work, although I don't think the component will be appropriate to use here but maybe can be added with a larger card refactor) which to me seems to keep the margin feel consistent, since there is a clear separation between "bonus space" of the thumbnail and the white of the margin. It also feels less "crisp" (I can't think of a better word - hopefully you will know what I am trying to get at)...

For example, in some various mobile views: 
| Without background |  With grey background  |
|---|---|
| Screen Shot 2021-09-28 at 10 42 57 AM | Screen Shot 2021-09-28 at 10 42 36 AM |
| Screen Shot 2021-09-28 at 10 42 50 AM | Screen Shot 2021-09-28 at 10 42 25 AM  |


So, I guess the question is, which option would you prefer that I use? Also, if @MisRob or @indirectlylit have alternative suggestions of how I might manage this, would be very open to hearing them!

MisRob:
Regarding the thumbnails, as I mentioned above I think it's completely fine to not use the new `ContentNodeThumbnail` if it's more complicated to use it rather than not for components that are sometimes used for new hybrid learning features and sometimes used for parts that we won't manage to migrate for the new release, if there are such places. However if possible I'd suggest mimicking its behavior as closely as possible - which means showing thumbnails in 16:9 ratio with light gray letterboxing (we agreed on this because 16:9 is the most common ratio) since I assume we'll migrate all cards eventually to render consistent thumbnails, is that right?

kolibri/plugins/learn/assets/src/views/ContentCardListViewItem.vue
MisRob: Thank you for checking it. From your explanation, it seems that it creates a hidden dependency from where this component is being used so even though things work for the bookmarks page, I'd recommend using something like [this removed implementation](https://github.com/learningequality/kolibri/pull/8461/files?authenticity_token=%2FOJGzRbCqz5GWJWPjneZGBH5FyeWWt%2BAvhmW9SXn9yzooCYichAcOdCfEN6t5btF%2BK1cQ72w9qRN22Fv%2FNwh8w%3D%3D&file-filters%5B%5D=.scss&file-filters%5B%5D=.vue#diff-96160db5aef4cbd2958957ee894db8c453a6b4dbfa372f557187af54855a1242L106-L113) so that we can be sure that it will work as expected no matter of scenario since this is a general component. Related to that, `ContentCardListViewItem` is used in `ContentCardGroupGrid` which seems to be imported into another four pages.

marcellamaki: Oh, now I see what you're saying. I will re-add the code for it to work as expected, and I also have in my note today to follow up on the usage of `ContentCardListViewItem` and the other places `ContentCardGroupGrid` is used, because I noticed that yesterday when I was working on something else. Thanks Misha!

MisRob: Sure, and regarding those components being used from both non-hybrid-learning and hybrid-learning, I hope we can find something simple to not block this PR way too much :) Just from the code, those various dependencies look to be "meeeh" :)) Let me know please if I can help with that somehow.

kolibri/plugins/learn/assets/src/views/BookmarkPage.vue
nucleogenesis: +1 for universal styles in typography. I feel like this should live in KDS both as guidance *and* as some sort of CSS or SCSS mixins that we can use to define consistent cross-product styles.

kolibri/plugins/learn/assets/src/views/ContentCard/CardThumbnail.vue
```diff
@@ -55,6 +70,28 @@
...
+  .chip-left {
+    position: absolute;
+    bottom: 20px;
+    left: 10px;
+    height: 34px;
+    padding: 8px;
+    font-size: 13px;
+    background-color: rgba(0, 0, 0, 0.7);
+    border-radius: 4px;
+  }
```
nucleogenesis: You don't have to account for both LTR and RTL with CSS classes in this case. We have some **_magic_** in our build system that automatically creates those styles. @indirectlylit (I think) may be best to describe how this works.

I just tested it - if you just create one class and style for LTR like so:

```
.chip {
    left: 10px;
    /* the rest of the styles */
}
```

Then change your language to Arabic and check the CSS on that `chip` class and you'll see that the left in your code is a right in the browser.

nucleogenesis: Separately, and this may be something to discuss in KDS as well, I received early on and have since held closely to the guidance that we should try to work within 8px increments wherever possible, 4 if needed, which is derived to some extent by Material guidelines.

I double checked KDS and see that we don't really have such guidelines, although it would be nice to have some. In any case, it's fairly nitpicky and looks fine as-is in the browser. Would love @indirectlylit and @MisRob 's thoughts w/ regard to whether we should aim to codify some aspect of this guidance in KDS or if it should just be a Kolibri thing (or if we are getting little/no benefit from the long-held norm).

I personally appreciate consistency in the code and have found trying to do everything at 8/16/24/etc px has been pretty painless in most cases.

nucleogenesis: One last note here is that I think that the height entry is unnecessary, you're expanding the container vertically with padding already and I didn't see any change when I removed it locally.

kolibri/plugins/learn/assets/src/views/ContentCardListViewItem.vue
```diff
@@ -272,30 +362,35 @@
...
 
-  .footer-icon {
+  .footer-left {
     display: block;
     float: right;
   }
+  .footer-right {
+    display: block;
+    float: left;
+  }
```
nucleogenesis: image

These icons ought to be flipped to the other side but the RTL version puts the icons under the thumbnail.

I'd suggest using flex here instead of floats. I think that floats aren't really supposed to be used for layout and that their primary purpose is to allow elements to flow around something that is stuck to one side. 

I think that something like this could work if you remove the footer-left/right styles and classes altogether. I tried it out quickly and seems to work well, but the `footerIcons` prop will need to be ordered so that the X is the outer-most icon. The `flex-end` automatically aligns everything to the correct side between RTL/LTR. Then the changes to `metadata-info-footer` align everything vertically when in mobile.

```
.footer {                     
  position: absolute;         
  bottom:0;                   
  display: flex;              
  justify-content: flex-end;  
  width: 100%;                
  padding: 16px;                         
}  

.metadata-info-footer {
  flex: auto;          
  align-self: center;  
  margin: 0;           
  font-size: 13px;     
}                               
                           
```

nucleogenesis (changes requested):
One additional item to note is that there is some usable whitespace in the mobile cards between the footer & the text. I think maybe an 8px vertical padding on the activity type (the icon & "watch/read/etc" bit) in mobile might give some good spacing between the thumbnail and the activity and the title.

Of course, items with small descriptions or no description are doomed to have some whitespace there, but this example is truncated. Maybe even using the standard font size could make use of the available space naturally without needing.

I quickly tried it out in devtools and a combination of these changes feel to maximize the use of available whitespace (but would need some testing, maybe tweaking for desktop vs mobile). 

```
.metadata-info {
    margin: 8px 0; // This won't look right on desktop though
}

.text {
    /* other .text styles */
    font-size: 16px;
}
```

image

marcellamaki:
**Updated notes on changes for devs**
- Per conversation several weeks ago with @MisRob and @bjester, I've done some refactoring to make separate "hybrid learning components" so we have greater reassurance that changes made in this PR will not cause regressions, and that any regressions will be known as regressions, rather than causing confusion of "what is in progress or in scope" 
- Ensured thumbnail uses 16:9 ratios with grey in alignment with other thumbnail work
- Adjusted some spacing and layouts to be more standard using pixel increments per @nucleogenesis, and other layout tweaks based on design guidance and consistent application of KDS principles

**Updated QA guidance for @radinamatic and @pcenov (and @jtamiace for design, if so inclined)**

Core functionality to test: 
- Bookmark page should be available and display either a selection of bookmarks or the "You have no bookmarked resources" message
- A user should be able to remove a bookmark by clicking the X on the bookmarks page
- A user should be able to add or remove a bookmark from within the content renderer. 
- [The bug Radina noted with the bookmark state incorrectly carrying over between bookmarks should now be fixed](https://github.com/learningequality/kolibri/pull/8458#issuecomment-931467746)

While there are a few adjustments that still need to be made to the cards, most of the card UI is "done" and UI/design review can include these points: 
- Cards should correspond to figma 
- Card should render correctly on mobile devices (with the correct mobile layout)
- Card content (aside from noted bugs) should be accurate 
- Bookmarked timestamp/date should display correctly

NOTES: 

Progress/Completion bug weirdness is popping up here too, so the progress bar may or may not be visible, based on what you are looking at! My recommendation right now is to put a pin in that part of the testing 📌 

Known TO-DOs are pointed out in the screenshot below (learning activity icon and string is missing and the placeholder is messing up some spacing, channel icon and channel title for alt text missing, activity length is a placeholder 'short activity')
Screen Shot 2021-10-12 at 5 23 09 PM

pcenov:
Hi @marcellamaki and @radinamatic 
I've tested this one today in Win 10 and Ubunto on several browsers + the Android build.
I should be noted that testing was hampered by the great number of known issues and especially the bookmark state carrying over. Also I'm not able to view any of the resources and therefore test the completion progress. The mobile view of the Library page is still a WIP.

Here is the test report:

|Functionality|Test Result|Comment|
|-|-|-|
|Display either a selection of bookmarks or the "You have no bookmarked resources" message|Pass|
|Remove a bookmark by clicking the X |Fail|https://github.com/learningequality/kolibri/issues/8497|
|Add or remove a bookmark from within the content renderer|Pass||
|Bookmark state incorrectly carrying over between bookmarks|Fail|The bookmark state is still carrying over|
|Bookmarked card should correspond to figma|Pass|aside from noted bugs|
|Bookmarked card should render correctly on mobile devices|Pass|--|
|Bookmarked card content (aside from noted bugs) should be accurate|Pass|aside from noted bugs|
|Bookmarked timestamp/date should display correctly|Pass|--|
|Progress/Completion|Not Tested|The feature cannot be tested with this build|
|RTL display of the bookmarked resources|Pass|-|

kolibri/plugins/learn/assets/src/views/HybridLearningContentCardListView.vue
```diff
@@ -0,0 +1,400 @@
...
+    align-self: center;
+    margin: 0;
+  }
+
+  .footer-icons {
+    position: absolute;
+    right: 16px;
+    bottom: 16px;
+    display: inline;
+  }
```
nucleogenesis: image

This is a problem that I tried to address during the vueification of Studio in KDS where the KIconButtons come through skewed. 

In lieu of that more sophisticated fix, we can fix this issue with some fairly simple CSS that jives with our material styles.

```
.footer-icons {
    .button {
        height: 32px!important;
        width: 32px!important;
    }
    svg {
        top: 4px;
     }
}
```

These styles may need to be applied inline to the KIconButton components (the `.button` one anyway) but I'd try it out in the `<style>` tags first.

If the styles don't show up, perhaps you'll need `/deep/` in front of `svg`.

marcellamaki:
Hi @pcenov - thank you for the QA given the challenges of known bugs. I've also realized one last code change I made to fix the icon I didn't push to github, I just had it locally, so I apologize for the oversight and for the difficulty you encountered while testing this. Hopefully as Jacob and I push though this last bit of frontend work and integration in the next two weeks, we will soon have much more complete (and less frustrating) features for you to test!

kolibri/plugins/learn/assets/src/views/BookmarkPage.vue
```diff
@@ -125,26 +89,21 @@
...
+      removeFromBookmarks(bookmark, index) {
+        if (bookmark) {
+          client({
+            method: 'delete',
+            url: urls['kolibri:core:bookmarks_detail'](bookmark.id),
+          }).then(() => {
+            this.bookmarks.pop(index);
+            this.createSnackbar(this.$tr('removedNotification'));
+          });
+        }
```
nucleogenesis: See in the video here that I click to remove "Intro to addition" but the "Color Vision" item is removed.

The correct bookmark, however, is actually removed when I refresh. The way this code pops the item off the list of bookmarks is the issue but the API call is working properly.

In Python, you can `pop(index)` to remove a specific element, but you cannot in JS. Rather `Array.prototype.pop()` only ever removes the last element from the array.

You can use `Array.prototype.splice` or `lodash/remove` to remove the element but this seems like it should work well enough - noting that the values in `this.bookmarks` are actually entire ContentNode entities with a key `bookmark` on them that points to the bookmark data. (Perhaps we can call `this.bookmarks` something that makes this clear like `this.bookmarkedContentNodes`).

`const bookmarks = this.bookmarks.filter(bm => bm.bookmark.id !== bookmark.id);`

https://user-images.githubusercontent.com/6356129/137220866-b0e6f81a-93e1-472d-b18d-139139236182.mp4

nucleogenesis (changes requested):
1) The bug Radina noted in the other PR is not fixed and doesn't feel in scope here (you've not changed the source of the issue). I mentioned this in #hybrid-learning on Slack that I fixed that toggle locally so that I could add bookmarks in order to test being able to delete them properly. 

[You can see my commit here](https://github.com/learningequality/kolibri/commit/0a9a59b64624ca2f98673b35a5c8ff5ccba7d366) - it's not fully tested but it let me add bookmarks where I couldn't initially.

Generally, this is coming together great. Looks good, feels good. Just a couple issues outstanding and then should be good to go!

kolibri/plugins/learn/assets/src/views/ContentCard/CardThumbnail.vue
indirectlylit: > Separately, and this may be something to discuss in KDS as well, I received early on and have since held closely to the guidance that we should try to work within 8px increments wherever possible, 4 if needed, which is derived to some extent by Material guidelines.
> 
> I double checked KDS and see that we don't really have such guidelines, although it would be nice to have some. In any case, it's fairly nitpicky and looks fine as-is in the browser. Would love @indirectlylit and @MisRob 's thoughts w/ regard to whether we should aim to codify some aspect of this guidance in KDS or if it should just be a Kolibri thing (or if we are getting little/no benefit from the long-held norm).
> 
> I personally appreciate consistency in the code and have found trying to do everything at 8/16/24/etc px has been pretty painless in most cases.

Sorry, just saw this message!

Thanks for flagging this @nucleogenesis. This is in fact already documented in the design system here:

https://design-system.learningequality.org/layout/#spacing

For more context, folks can read about "baseline grids" in the context of [Material Design](https://material.io/design/layout/spacing-methods.html#baseline-grid) or [in general](https://www.designsystems.com/space-grids-and-layouts/).

nucleogenesis:
@marcellamaki I finally (sorry) just got to get to test this and I'm not seeing any content in channels and the cards in the library page don't do anything so I cannot add any Bookmarks.

I tried rebasing the branch locally but still wasn't seeing any content. If this isn't a known issue resolved elsewhere, I'd hold off on merging before the bug bash.

marcellamaki:
@nucleogenesis - the issue you describe is a result of me doing a sub-par rebase (things to learn for the future!). It is fixed in [the subsequent PR for the Library Page](https://github.com/learningequality/kolibri/pull/8526). Obviously now we are not hurrying to get in before the bug bash, so no need to merge quickly, but just as an FYI. 

If you're able to access content from the home page, it _should_ open the content renderer correctly. (It is working locally for me on this branch). If that's a feasible workaround for this PR, and you can review what you need, that's great, but if not, just let me know and I can scoot some of the updates to the Library page back into this PR tomorrow so that it's a bit easier to work with. 

[ETA: I am just going to pull that code in in the morning so everything is easier to review!]

marcellamaki:
Superseded by #8548

Issue #8124, opened 2021-05-28
Re-evaluate usage of TextTruncator and optimize its performance


MisRob:
## Observed behavior

`TextTruncator` triggers layout reflows that on pages with many instances of `TextTruncator` can cause significant performance issues, as observed in #7525.

This is the first part of the profile (at some point my laptop runs out of resources and doesn't record anymore) of Learn section of  Ciencia Espacial topic containing 1151 sub-topics (which means 1151 of `TextTruncator` instances) of Ciencia NASA channel where the following bottlenecks can be observed:

![profile](https://user-images.githubusercontent.com/13509191/119797716-c445f880-beda-11eb-90db-c8d3d549a12d.png)

As can be seen from the following, for a page containing hundreds of `TextTruncator` instances, there is an enormous number of expensive style calculations and layouts/reflows run.

**Performance analysis for a single `TextTruncator` instance**:

**(1) `TextTruncator` ->  `mounted` -> `ResizeSensor` (`KResponsiveElementMixin`) -> `attachResizeEvent`**

![Screenshot from 2021-05-27 10-28-55](https://user-images.githubusercontent.com/13509191/119793026-82b34e80-bed6-11eb-97a4-870cadbab609.png)

| 1 style recalculation | 1 layout/reflow |
| -------------------------- | ------------------- |
| ![Screenshot from 2021-05-27 10-29-13](https://user-images.githubusercontent.com/13509191/119793041-86df6c00-bed6-11eb-83e6-5954e4fe95de.png) | ![Screenshot from 2021-05-27 10-29-37](https://user-images.githubusercontent.com/13509191/119793060-8ba42000-bed6-11eb-94e0-d6990dff4b65.png) |

**(2) `TextTruncator` -> `ResizeSensor` (`KResponsiveElementMixin`) -> `resetExpandShrink`**

![Screenshot from 2021-05-27 10-33-33](https://user-images.githubusercontent.com/13509191/119793753-2866bd80-bed7-11eb-9792-2dffe0a9a3c8.png)

| 1 style recalculation | 1 layout/reflow |
| -------------------------- | ------------------- |
[...]

Issue #5468, opened 2019-05-09
many text wrapping issues related to titles in cards


bjester:
### Observed behavior

Also relevant:

* https://github.com/learningequality/kolibri/issues/5464
* https://github.com/learningequality/kolibri/issues/4419

In Firefox, titles on content cards will break in the middle of a word, splitting the word in two pieces without indication that this has happened.
![Word Break](https://user-images.githubusercontent.com/819838/57481889-0a7a8180-7258-11e9-8c50-f1b3003adca4.png)

### Expected behavior

In the English language, I expect to see a hyphen appended to the first half of the word, indicating the word has been broken in half and the rest follows on the next line. Although, it may be advisable to change the wrapping controls to avoid this entirely, opting for a different approach if the title is too long for the area on the card.

### User-facing consequences

Users may misunderstand a content's title because longer words may be split in half with no indication of what's happening.

### Errors and logs

None.

### Steps to reproduce

Find or create content with lengthy titles. Screenshot is taken of content from channel with token `gojad-patot`

### Context

Kolibri - develop
Ubuntu 18.04
Firefox 66

jonboiser:
It seems that Firefox doesn't support the `word-break: break-word` CSS. So the fallback `word-break: break-all` is being used.

Maybe we can remove this fallback rule.

indirectlylit:
what was the problem with `word-break: normal`?

jonboiser:
I don't know. It seems like it doesn't try to do anything if it's `word-break: normal`
![Screen Shot 2019-05-10 at 1 32 50 PM](https://user-images.githubusercontent.com/10248067/57555095-264d5880-7328-11e9-9685-090d1293a483.png)

jonboiser:
Maybe we don't need a word-break rule at all, since there's supposed to be some JS that truncates long titles anyway.

indirectlylit:
the change was in PR #4439 re: issue #4419

| before | after |
|--|--|
| ![image](https://user-images.githubusercontent.com/2367265/57560825-48050a80-733d-11e9-8bf8-c0422876084a.png) | ![image](https://user-images.githubusercontent.com/2367265/57560831-4fc4af00-733d-11e9-9c68-0b13ac4b1396.png) |

That doesn't really seem like the right fix to the problem to me - we can't just break words across lines at arbitrary places.

[...]

Issue #5464, opened 2019-05-08
Edge case with long, unbroken title on content cards


jonboiser:
### Observed behavior
Related: https://github.com/learningequality/kolibri/issues/5468

These are some PDFs from the TESSA channel, which have as their title, a very long single word, like "namba_ya_moduli_1__kusoma_na_kuandika_kwa_madhumuni_mbalimbali_printable"

On the cards, this ends up being completely elided:

![Screen Shot 2019-05-08 at 3 10 45 PM](https://user-images.githubusercontent.com/10248067/57411883-b4e49d00-71a3-11e9-8147-d36a31fb6bdb.png)

Updated screenshot:

![image](https://user-images.githubusercontent.com/10248067/110861438-fbf5ba80-8272-11eb-9e1d-6f25a8d0f643.png)

### Expected behavior

An alternative way to show at least a little bit of these long titles.

The text should be truncated such that it does not run off the side, even if the text is one long un-spaced string. Hovering should show a tooltip with the whole title.

### User-facing consequences

…

### Errors and logs

…

### Steps to reproduce

…

### Context

v0.12 develop
…

radinamatic:
Still extant ([recent Slack conversation](https://learningequality.slack.com/archives/CB37UM23A/p1607656280385500)), not sure if we could up the priority @jonboiser?

http://35.227.71.104/en/learn/#/topics/c/991073b0312e4959aaa3118340c584a3
http://35.227.71.104/en/learn/#/topics/t/708fc2e73a0f4ab59dfeebfd1203896d

Issue #4419, opened 2018-10-14
Long words in title problematic in content cards


rtibbles:
### Observed behavior
@christianmemije reports:
Long card title get cut off inside cards, fonts popping in making text shift

### Expected behavior
Titles always get displayed in a length sensitive way.
Fonts should not pop in and make text move!

### User-facing consequences
Can't read titles.
Jerky visual experience.

### Steps to reproduce
Look at a content card with a long title.

### Context
* v0.11.0-beta1

indirectlylit:
I'm not reproducing in either Learner or Coach

christianmemije:
This happens for card with long words. We would need to add ` word-break: break-word;`
![localhost_8000_learn_](https://user-images.githubusercontent.com/7193975/47034550-87af9f00-d12c-11e8-9091-97684c018c8e.png)

christianmemije:
Overflowing text fixed in #4439

indirectlylit:
As [indicated here](https://github.com/learningequality/kolibri/issues/5468#issuecomment-491455064), the previous solution caused real words to break at arbitrary points.

One strategy might be to have the `TruncatedText` component gracefully handle overflow, e.g. like:

> ![image](https://user-images.githubusercontent.com/2367265/57562711-00d14680-734a-11e9-92df-0c96dc66c92a.png)

@rtibbles do you see the font movement again after #5479 was merged?

rtibbles:
If anyone else has time to confirm this is still fixed, I would appreciate it.

radinamatic:
On `0.12.y` dev branch I see *`...`* substituting the whole long title of the content card, on both FF & Chrome on Ubuntu 18.04. Not liking the tooltip strategy, it forces the user to hover to figure out what is inside.

![LEDev1905 (start)  Running  - Oracle VM VirtualBox_033](https://user-images.githubusercontent.com/1457929/57986311-a20b7b80-7a73-11e9-9681-947e60d2133d.png)

Words in long titles are still truncated arbitrarily in FF:

![LEDev1905 (start)  Running  - Oracle VM VirtualBox_034](https://user-images.githubusercontent.com/1457929/57986344-03cbe580-7a74-11e9-8305-63f435e068fc.png)

indirectlylit:
>  I see `...` substituting the whole long title of the content card, on both FF & Chrome on Ubuntu 18.04. Not liking the tooltip strategy, it forces the user to hover to figure out what is inside.

[...]

PR #8464, opened 2021-09-24
Experiment with pure CSS alternative to `TextTruncator` for simpler use cases

MisRob:
## Summary

Introduces `TextTruncatorCss` as a simpler pure CSS solution to our current JavaScript based `TextTruncator`. It is meant to be temporary or additional to the JS solution and for shorter texts like titles on cards.

### Motivation

While working on the new home page in the hybrid learning project, I need to truncate cards titles to two lines. We have some [performance issues with `TextTruncator`](https://github.com/learningequality/kolibri/issues/8124) that are caused by `ResizeSensor` and `shave.js` where `shave.js` seems to be even bigger bottleneck than `ResizeSensor`. To avoid risking introducing similar problems to new pages, I used this task as an opportunity to play with some CSS alternatives that could be potentially used for simpler use cases, at least until we decide on what to do with `TextTruncator` issues.

Although my primary focus was on cards since there can be many instances of them on one page and risk of performance problems is thus higher, I also thought that having the CSS solution available might be useful for other simple use-cases like the [one-line resource title in the learning activity bar](https://www.figma.com/file/uiZuCIqh2KYvBUfbCiJyyA/Hybrid-Learning?node-id=1944%3A83765) and similar.

### Solution

The component is built with a progressive enhancement approach, using three CSS truncation techniques. Please check the comments in the component for details.

## `TextTruncator` and `TextTruncatorCss` comparison

Tested on a small sample of texts resembling cards titles in Firefox, Chrome, and IE11. I tried to locate differences and take screenshots of `TextTruncatorCss` results that looked the worst.

### (1) One line

```javascript
<TextTruncatorCss
  :maxLines="1"
  text="Idiopathic pulmonary fibrosis - causes, symptoms, diagnosis, treatment, pathology"
/>

<TextTruncator
  :maxHeight="23"
  text="Idiopathic pulmonary fibrosis - causes, symptoms, diagnosis, treatment, pathology"
/>
```

`TextTruncator` and `TextTruncatorCss` seem to behave identically in all Firefox, Chrome, and IE11.

230px-css-firefox,chrome,ie

### (2) Multiple lines

```javascript
<TextTruncatorCss
  :maxLines="2"
  text="Idiopathic pulmonary fibrosis - causes, symptoms, diagnosis, treatment, pathology"
/>

<TextTruncator
  :maxHeight="46"
  text="Idiopathic pulmonary fibrosis - causes, symptoms, diagnosis, treatment, pathology"
/>
```

Similar behavior in Firefox and Chrome, however `TextTruncatorCss` seems to truncate by word rather than by character more often than `TextTruncator`. `TextTruncator` works well in IE 11. `TextTruncatorCss` is giving bad results in IE11 since in some cases, there can be a gap between ellipsis dots and the last character of a text. Seems to be a major problem with the CSS-based solution.

| | Width 200px - Firefox, Chrome | Width 230px - Firefox, Chrome | Width 230px - IE11 |
| -- | -- | -- | -- |
| `TextTruncatorCss` | 200px-css-firefox,chrome | 230px-css-firefox,chrome | 230px-css-ie |
| `TextTruncator` | 200px-js-firefox,chrome | 230px-js-firefox,chrome | 230px-js-ie |

### (3) Multiple lines - a long word with no breaks

```javascript
<TextTruncatorCss
  :maxLines="2"
  text="Idiopathicpulmonaryfibrosiscausessymptomsdiagnosistreatmentpathology"
/>

<TextTruncator
  :maxHeight="46"
  text="Idiopathicpulmonaryfibrosiscausessymptomsdiagnosistreatmentpathology"
/>
```

`TextTruncator` doesn't respect height settings and only shows one line in all Firefox, Chrome, and IE11. However, I guess that it's just a bug rather than the JS solution limitation. `TextTruncatorCss` works well in Firefox and Chrome, but has the same problem like in (2) in IE11.

|    | Firefox, Chrome | IE11 |
| -- | -- | -- |
| `TextTruncatorCss` | 03-css-firefox,chrome | 03-css-ie |
| `TextTruncator` | 03-js-firefox,chrome,ie | 03-js-ie |

## References

- [Figma designs for the home page](https://www.figma.com/file/uiZuCIqh2KYvBUfbCiJyyA/Hybrid-Learning?node-id=1788%3A6234)
- [`TextTruncator` performance issue](https://github.com/learningequality/kolibri/issues/8124)

## Comments

There's an issue in IE11 with three ellipsis dots not showing in some situations. It seems to be a font issue. The problem disappears after unchecking this rule

```css
.partial-fonts-loaded body, .partial-fonts-loaded button, .partial-fonts-loaded input, .partial-fonts-loaded select, .partial-fonts-loaded textarea {
	font-family: noto-subset,noto-common,sans-serif;
}
```

and falling back to `sans-serif`. I haven't included it in comparison because I saw it happen for both `TextTruncator` and `TextTruncatorCss` in some scenarios.

## Reviewer guidance

I moved this logic from the home page branch to get some feedback from the team if this could be an acceptable way forward for us, so please review the advantages and disadvantages of both approaches for shorter texts. For more background on reasons to use `TextTruncator`, please see @indirectlylit's [comment](https://github.com/learningequality/kolibri/issues/8124#issuecomment-923977618). An alternative solution for home page cards could be to use the existing `TextTruncator`. In that case, I'd suggest prioritizing refactoring of `TextTruncator` performance-wise.

----

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/core/assets/src/views/TextTruncatorCss.vue
+++ b/kolibri/core/assets/src/views/TextTruncatorCss.vue
@@ -1,0 +1,133 @@
+<template>
+
+  <!--
+    Text is wrapped in two `div`s to allow parent components adding
+    padding style directly on `<TextTruncatorCss>` component no matter
+    of what truncating technique is used. Otherwise adding padding directly
+    would break when using technique (B) since text that should be truncated
+    would show in padding area.
+  -->
+  <div>
+    <div :class="$computedClass(truncate)">
+      {{ text }}
+    </div>
+  </div>
+
+</template>
+
+
+<script>
+
+  /**
+   * Truncates text to a certain number of lines and adds "..."
+   *
+   * This is a pure CSS alternative to `TextTruncator` for simpler
+   * use cases like card titles since `TextTruncator` is causing performance
+   * issues (https://github.com/learningequality/kolibri/issues/8124)
+   * and thus its usage is currently not recommended for larger
+   * amounts of instances on one page.
+   *
+   * Compared to `TextTruncator` there are two disadvantages:
+   * - depending on length of words of the text, there might be a gap
+   *   between the last visible word and "..." in Internet Explorer
+   *   and old versions of some other browsers (see implementation comments)
+   * - it currently doesn't offer "View all"/"View less" functionality
+   */
+  export default {
+    name: 'TextTruncatorCss',
+    props: {
+      text: {
+        type: String,
+        required: true,
+      },
+      maxLines: {
+        type: Number,
+        required: false,
+        default: 1,
+      },
+      /**
+       * Text line height in rem.
+       * Used only for Internet Explorer fallback.
+       */
+      lineHeightIE: {
+        type: Number,
+        required: false,
+        default: 1.4,
+      },
+    },
+    computed: {
+      truncate() {
+        /*
+          (A)
+          For one line, use standard ellipsis text overflow
+          that works well for such scenario
+        */
+        if (this.maxLines === 1) {
+          return {
+            overflow: 'hidden',
+            textOverflow: 'ellipsis',
+            whiteSpace: 'nowrap',
+          };
+        }
+
+        // 54 is random number only to be able to define `supports` test condition
+        if ('CSS' in window && CSS.supports && CSS.supports('-webkit-line-clamp: 54')) {
+          /*
+            (B)
+            For multiple lines, use line clamp in browsers that support it
+            (https://developer.mozilla.org/en-US/docs/Web/CSS/-webkit-line-clamp)
+          */
+          return {
+            overflow: 'hidden',
+            display: '-webkit-box',
+            '-webkit-line-clamp': `${this.maxLines}`,
+            '-webkit-box-orient': 'vertical',
+            // needed to make line clamp work for very long word with no spaces
+            overflowWrap: 'break-word',
+          };
+        } else {
+          /*
+            (C)
+            Fallback for multiple lines in Internet Explorer and some older versions
+            of other browsers that don't support line clamp
+            (https://caniuse.com/mdn-css_properties_-webkit-line-clamp).
+            Calculate max height and add "..." in `::before` while covering it with
+            white rectangle defined in `::after` when text doesn't need to be truncated.
+            Adapted from https://hackingui.com/a-pure-css-solution-for-multiline-text-truncation/
+            and https://css-tricks.com/line-clampin/#the-hide-overflow-place-ellipsis-pure-css-way.
+          */
+          const ellipsisWidth = '1rem';
+          return {
+            overflow: 'hidden',
+            position: 'relative',
+            lineHeight: `${this.lineHeightIE}rem`,
+            maxHeight: `${this.maxLines * this.lineHeightIE}rem`,
+            // needed to make truncation work for very long word with no spaces
+            // `word-wrap` is a legacy name for `overflow-wrap` that needs to be used for IE
+            wordWrap: 'break-word',
+            // create space for "..."
+            paddingRight: ellipsisWidth,
+            marginRigth: `-${ellipsisWidth}`,
+            '::before': {
+              content: "'...'",
+              position: 'absolute',
+              right: 0,
+              bottom: 0,
+            },
+            // cover "..." with white rectangle when text
+            // doesn't need to be truncated
+            '::after': {
+              content: "''",
+              position: 'absolute',
+              right: 0,
+              width: ellipsisWidth,
+              height: '50%',
+              background: 'white',
+            },
+          };
+        }
+      },
+    },
+  };
+
+</script>
--- a/kolibri/core/assets/src/views/TextTruncatorCss.vue
+++ b/kolibri/core/assets/src/views/TextTruncatorCss.vue
@@ -19,19 +19,12 @@
 <script>
 
   /**
-   * Truncates text to a certain number of lines and adds "..."
+   * Truncates text to a certain number of lines
+   * and adds an ellipsis character "…"
    *
-   * This is a pure CSS alternative to `TextTruncator` for simpler
-   * use cases like card titles since `TextTruncator` is causing performance
-   * issues (https://github.com/learningequality/kolibri/issues/8124)
-   * and thus its usage is currently not recommended for larger
-   * amounts of instances on one page.
-   *
-   * Compared to `TextTruncator` there are two disadvantages:
-   * - depending on length of words of the text, there might be a gap
-   *   between the last visible word and "..." in Internet Explorer
-   *   and old versions of some other browsers (see implementation comments)
-   * - it currently doesn't offer "View all"/"View less" functionality
+   * Internet Explorer note:
+   * Depending on length of words of the text, there might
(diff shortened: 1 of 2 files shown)


kolibri/core/assets/src/views/TextTruncatorCss.vue
```diff
@@ -0,0 +1,133 @@
...
+            lineHeight: `${this.lineHeightIE}rem`,
+            maxHeight: `${this.maxLines * this.lineHeightIE}rem`,
+            // needed to make truncation work for very long word with no spaces
+            // `word-wrap` is a legacy name for `overflow-wrap` that needs to be used for IE
+            wordWrap: 'break-word',
+            // create space for "..."
+            paddingRight: ellipsisWidth,
+            marginRigth: `-${ellipsisWidth}`,
+            '::before': {
+              content: "'...'",
```
indirectlylit: Note that there is an actual [ellipsis character](https://www.compart.com/en/unicode/U+2026):

```suggestion
              content: "'…'",
```

MisRob:
Regarding a11y, I haven't tested it myself, but found [some information here](https://butterpep.com/line-clamp-overflow-ellipsis.html). For most techniques, it says that the whole text would be read. My assumption is that for shorter texts like resource titles, it wouldn't be a problem. @radina is that right?

indirectlylit (commented):
This looks great! The CSS implementation degrades gracefully on IE11 and otherwise the behaviors are quite similar, so I would support migrating to it entirely rather than maintaining two parallel implementations.

Shorter-term the main question is cost/risk of migration – if either are high, we might want to use both at least until 0.15. Longer-term once things are consolidated, we could even move it to the design system as `KTextTruncator`.

> Multiple lines - a long word with no breaks

The behavior for the CSS implementation looks good here. Flagging that this case has been a source of many bugs over the years, and warrants a careful eye:

* https://github.com/learningequality/kolibri/issues/5468
* https://github.com/learningequality/kolibri/issues/5464
* https://github.com/learningequality/kolibri/issues/4419


> There's an issue in IE11 with three ellipsis dots not showing in some situations. It seems to be a font issue.

Would you mind filing a separate issue for this? For some context, here is an overview of the intended behavior:

https://github.com/learningequality/kolibri/blob/6903afec0ca3ebdf65fbf0eb96ef5348ed9fc2d2/packages/kolibri-tools/lib/i18n/fonts.py#L589-L603

and the client-side font loading occurs here:

https://github.com/learningequality/kolibri/blob/6903afec0ca3ebdf65fbf0eb96ef5348ed9fc2d2/kolibri/core/assets/src/utils/setupAndLoadFonts.js

> Regarding a11y ... for most techniques, it says that the whole text would be read. 

I've always assumed that the best behavior would be to only read the visible text: if we've made a design decision to truncate it visually, that should probably apply to audible text too? That said, I don't think that either choice is "wrong" and both will result in a reasonable experience for all users.

If we wanted to get really fancy here, we could use `aria-hidden="true"` and make a separate off-screen copy of the text for screen readers and truncate that based on some metric like `maxLen = maxLines * 50`.

marcellamaki:
Hi @MisRob - thanks for working on this and layout out the comparison so clearly. I like the idea of moving towards a single option, and maybe even moving the component into KDS ultimately, as @indirectlylit suggests. 

As to @indirectlylit's point about migration - right now, it looks like the original `TextTruncator` is only used in 5 files, and I think we could swap this out fairly easily. If we choose to do that, I think it would be good to merge this promptly and swap out the components right away, so there is plenty of time for QA (especially seeing the edge cases with long text), and then we can revert or adjust as needed. To me, this seems simpler and a better use of time than refactoring the existing `TextTruncator` separately, especially if we plan on replacing it in the near future, as long as no one finds the IE11 spacing issues to be a problem.

radinamatic:
> I've always assumed that the best behavior would be to only read the visible text: if we've made a design decision to truncate it visually, that should probably apply to audible text too? That said, I don't think that either choice is "wrong" and both will result in a reasonable experience for all users.

I did not find a clear recommendation re. screenreader output _full text_ vs. _only the visible text_ for long card titles, but if we assume that the goal is to have comparable experience for **all** users, reading only the visible text should be kosher.

Text truncation is an issue for when the screenreader output is cut on some important actionable element like a button label, but I don't see it as a blocker for truncation of long sentences - all users can always click the card to open and read the full title sentence.

kolibri/core/assets/src/views/TextTruncatorCss.vue
MisRob: Thank you, updated

MisRob:
Thank you everyone for your feedback. I've just pushed all updates so this should be ready for the final review and eventually merge.

---

Regarding replacing `TextTruncator` completely, originally I had three areas of concern:

(1) gap between the last word and ellipsis in IE - according to the feedback, doesn't seem to be a blocker

(2) wasn't sure about what should be proper accessibility behavior - reading the feedback here it seems to be quite open and most importantly, with @indirectlylit's suggestion :

> we could use aria-hidden="true" and make a separate off-screen copy of the text for screen readers and truncate that based on some metric like maxLen = maxLines * 50

we should be able to tweak it as needed at any point in the future

(3) when compared to `TextTruncator` it doesn't have "View more/less" functionality - however, I checked all occurrences of `TextTruncator` and I didn't see `TextTruncator`'s `showViewMore` prop actually being used. If we need it at some point, we can something similar to `TextTruncatorCss`.

All of these issues seem to be resolved so I feel fine about migrating completely. I don't know if we have enough time to switch it now since as @indirectlylit pointed out, there are relatively tricky edge cases we need to be cautious about. However even if we get to it, I'd suggest keeping `TextTruncator` component around because it has "View more/less" strings - just in case they might be handy before the next release (e.g. it looks that if we get to adding "View more/less" to the new `TextTruncatorCss`, it might be helpful in @nucleogenesis 's [truncation in side panels](https://github.com/learningequality/kolibri/pull/8466/files#diff-125f7ddb4ac461cbed0012782df0034cca420ce8912e187e075e504eedbd6644R29))

MisRob:
@indirectlylit I am putting opening the font issue on my TODO's list - if that's fine I'd like to do it as soon as I open a PR with the finished home page that uses `TextTruncatorCss` for cards titles so I have simple and straightforward steps to reproduce for the issue (even though I saw it for the `TextTruncator` too it might not be relevant to report it when it seems we might stop using it soon and I discovered it only by using it temporarily in the new home page code where it won't be used so I wouldn't have good steps to reproduce right now)

marcellamaki (approved):
Nice work @MisRob!

indirectlylit:
hooray!

MisRob:
> I am putting opening the font issue on my TODO's list - if that's fine I'd like to do it as soon as I open a PR with the finished home page that uses TextTruncatorCss for cards titles so I have simple and straightforward steps to reproduce for the issue

@indirectlylit I wasn't able to reproduce this particular font issue [in the home page PR](https://github.com/learningequality/kolibri/pull/8504) anymore. There's another IE11 text ellipsis problem though that I'm reporting in the PR now.

indirectlylit:
thank you!

Issue #8107, opened 2021-05-25
Create new informational side panel for hybrid learning view


marcellamaki:
**Overview**:
The new side panel component is used to display resource information and lesson plan information. 

Done when:
- Side panel can be toggled open and closed (note: this can be a last `TODO` for when the menu bar is merged)
- Side panel has the appropriate new mobile-responsive breakpoints laid out in figma
- Side panel content is rendered either as information panel or as lesson/topic panel
- As an information panel: 
      1.  The resource title, description, subject, level, language, author, license holder,  and license details are visible and displayed in line with the figma design 
      2. Long descriptions should be truncated to fit within a "preview" 170px high, and have an option to toggle to read more (with scroll) and then read less
      3. The license details should default to being collapsed, and similarly have a toggle to display that metadata 
      4. There should be a button at the bottom of the panel to download resource, and this interaction should be implemented as part of the PR 
- As a topic/lesson panel:
   1. Each related (previous/next resource) should be listed in order 
   2. The current resource should be indicated as current with a shaded background (as indicated in design)
   3. Basic metadata of the title of the activity/resources and an icon should be displayed, with tooltips
   4. Empty topics should render with a "no more topics" message (see figma references)

**Beyond scope of this ticket**:
- Creating the "chips" seen at the top of the info panel
- Side panel implemented in other place in the code base
- Interactions to navigate between topics/resources in the side panel _(as this needs further clarification)_
- Display of metadata related to progress, completion and estimated time

**NOTE**: 
- Any metadata that is within the scope of this ticket but not available should be noted and raised for discussion to clarify the timeline of when it can be included. `TODOs` should be included in the code as well to keep track, and outstanding items should be referenced in the PR.

**Figma references**:
<img width="729" alt="Screen Shot 2021-05-25 at 10 36 11 AM" src="https://user-images.githubusercontent.com/17235236/119516778-09382680-bd45-11eb-8c18-d7c965d6c1df.png">
<img width="308" alt="Screen Shot 2021-05-25 at 10 31 36 AM" src="https://user-images.githubusercontent.com/17235236/119516509-cfffb680-bd44-11eb-818f-99513867d16c.png">
<img width="926" alt="Screen Shot 2021-05-25 at 10 31 28 AM" src="https://user-images.githubusercontent.com/17235236/119516510-cfffb680-bd44-11eb-80ff-87149e69ddb0.png">
[...]

PR #8267, opened 2021-08-09
Add a resource completion modal

MisRob:
## Summary

Removes mastery snackbars in favor of new responsive completion modal.

Note that "Next resources" section at the bottom of a resource page hasn't been removed intentionally as we don't have UI for navigating to the next resource after closing the completion modal yet.

|    |    |     |     |
| -- | -- | -- | -- |
| Screenshot from 2021-08-09 16-06-13 | Screenshot from 2021-08-09 16-06-31 | Screenshot from 2021-08-09 16-06-49 | Screenshot from 2021-08-09 16-29-10 |

## References

- Closes  #8109
- [Figma designs](https://www.figma.com/file/uiZuCIqh2KYvBUfbCiJyyA/Hybrid-Learning?node-id=2254%3A6) (note that there are some intentional deviations from designs that have emerged from discussions with Khang and @radinamatic )

## Reviewer guidance

- Test for smaller resolutions
- Test keyboard navigation
- Test RTL

**(A) Channel resource page**
1. Login
2. Go to "Learn" -> "Channels"
3. Navigate to a channel resource that you haven't completed yet
4. Finish the resource (e.g. by completing watching a video)
5. See that snackbars with points and a next resource don't appear any more and the modal is displayed instead
6. Interact with the modal - stay on the page, close the modal, go to the next resource, click one of the recommended resources 

**(B) Channel resource page for anonymous user**

Make sure you're logged out and follow (A) steps. In addition to that, check that the warning message is displayed instead of the point obtained on the modal after completing a resource.

**(C) Class resource page**
1. Login
2. Go to "Learn" -> "Classes"
3. Navigate to a class resource that you haven't completed yet
4. Finish the resource (e.g. by completing watching a video)
5. See that snackbars with points and a next resource don't appear any more and the modal is displayed instead
6. Interact with the modal - stay on the page, close the modal, go to the next resource, click one of the recommended resources 

----

## Testing checklist

- [x] Contributor has fully tested the PR manually
- [x] If there are any front-end changes, before/after screenshots are included
- [ ] Critical user journeys are covered by Gherkin stories
- [ ] Critical and brittle code paths are covered by unit tests

## PR process

- [x] PR has the correct target branch and milestone
- [x] PR has 'needs review' or 'work-in-progress' label
- [x] If PR is ready for review, a reviewer has been added. (Don't use 'Assignees')
- [ ] If this is an important user-facing change, PR or related issue has a 'changelog' label
[...]

Issue #3781, opened 2018-06-08
audit alignment of displayed user data and content


indirectlylit:
### Observed behavior

Sometimes we use `text-align` on user data and content. This over-rides `dir="auto"`, which results in content and text not being properly localized.

### Expected behavior

* No displayed user data or content should have `right` or `left` set for the `text-align` property (`center` is OK).
* All displayed user data and content should have `dir="auto"`
* `dir="auto"` should not generally be set on 'wrapper' nodes - it should be as close to the text as possible
* Translated "app text" does not generally need to have `dir="auto"`

| Incorrect | Correct |
|--|--|
| ![image](https://user-images.githubusercontent.com/2367265/41183996-500f2f12-6b32-11e8-99ce-1f6ea6df4912.png) | ![image](https://user-images.githubusercontent.com/2367265/41183934-09875b0a-6b32-11e8-80a6-08ae6b2386de.png) |

### User-facing consequences

Mixed LTR/RTL text is harder to read

### Context

0.10.0

Priyaraj17:
Is this issue open? May I work on it?

nucleogenesis:
@Priyaraj17 sure that'd be much appreciated. If you have any questions please let us know.

Priyaraj17:
Sure, I will get started. Can you tell me how to reproduce this issue?

nucleogenesis:
Yes - @Priyaraj17 this issue is a matter of ensuring that we don't use the `text-align` property in our CSS with the `right` and `left` values.

Resolving the issue would involve:

- [ ] Finding all instances of `text-align` in our CSS (both inline CSS and the stuff between `<style>` tags) and making sure that they do not assign the value of `right` or `left`.
- [ ] If you make a change, ensure that the content affected by that CSS change displays correctly in RTL and LTR. If it doesn't, you may need to add `dir="auto"` attribute to the HTML.

When you create a PR - and as you go - it would be helpful during review if you provide screenshots.

Priyaraj17:
Thank you. I will start ASAP.

indirectlylit:
Just to clarify a subtle but important distinction:

Using `text-align` with `right` and `left` is entirely acceptable for page layout. These values should be written assuming an LTR layout, and they will automatically be swapped when viewed in an RTL language using [RTLCSS](https://rtlcss.com/).

[...]

PR #8466, opened 2021-09-24
Side Panels Implementation

nucleogenesis:
## Summary

Follows up on previous work in #8160 which was merged for String Freeze.

`FullScreenSidePanel` is a component in `kolibri.coreVue.components` that provides responsive functionality and a default slot into which content must be added. This allows this backdrop-masked side bar to be used across Kolibri without coupling it to any particular data.

This PR covers two kinds of side panel content, accessible in three places by the user.

[Scenario 1 - A User is Viewing Content](https://www.figma.com/file/uiZuCIqh2KYvBUfbCiJyyA/Hybrid-Learning?node-id=328%3A25722)

| Actual Content | With Fake Data |
|-|-|
|actual-irl-video-metadata|metadata-with-all-the-fixins|

[Scenario 2 - A User is Browsing Content](https://www.figma.com/file/uiZuCIqh2KYvBUfbCiJyyA/Hybrid-Learning?node-id=55%3A8)

`SidePanelResourceMetadata` was initially the implementation covering one of the display styles. I opted to strip this component and base two others off of it. This resulted in my using `kolibri.utils.i18n.crossComponentTranslator` to get these strings into the two new components I've added covering the two scenarios:

1) `CurrentlyViewedResourceMetadata` covers the first scenario and is rendered in `LearnImmersiveLayout`
2) `BrowseResourceMetadata` covers the second scenario and is rendered in `LibraryPage` and `TopicsPage`

**Important notes**

- This focuses firstly on providing the styles and layout so that once we decide on how the metadata is structured, we can wire it all up to working layout.
- Ideally, the data passed to the `*ResourceMetadata` components will readily and easily include everything needed - so any state mapping can be done higher in the hierarchy. It would be super nice to have a bespoke data type - perhaps returned from a getter - that can be made readily available to components rendering the side panels.
- There is a lot of duplication here but that feels better to me than trying to work out some convoluted system for displaying the metadata in a `v-for` of some kind. I'm definitely open to suggestions on this approach or ways we might be able to improve.

## References

#8107

## Reviewer guidance

The best way to test this is locally in the code and by putting fake data in for each metadata slot if you want to see it.

I don't really think there is much to test aside from making sure the side bars are toggling correctly, have the correct metadata in place per the specs and are on the right track to being ready once we know the metadata in a concrete way.

Note, the "related resources" is missing. We can work through that after the fact, this is an aging PR and is stable enough to get in and we can build up on it in coming weeks.

----

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated


MisRob:
I'll skim through the code soon. Thank you for the summary, this approach sounds good to me.

Regarding TODOs, I am not sure if the following will be helpful in all cases, but if you'd like to you can check it out and see, or adjust as needed:

> Derive and display thumbnail (get the styles correct)

`ContentNodeThumbnail` and `QuizThumbnail` made for hybrid learning features [here](https://github.com/learningequality/kolibri/tree/develop/kolibri/plugins/learn/assets/src/views/thumbnails) might be helpful

> Figure out how / where to get Related resources & "location" in channel, where they link to, etc for BrowseResourceMetadata

Regarding related resources, @rtibbles recommended using our existing endpoint for now so I got them from [recommended getter](https://github.com/MisRob/kolibri/blob/0c3125b5a81c4b28ec3e8fc1e09e6be29690cc6f/kolibri/plugins/learn/assets/src/views/ContentPage.vue#L202) (however, this getter seems to be associated with the current topic tree) and fetched them with [this function](https://github.com/learningequality/kolibri/pull/8267/files#diff-deb71a7e53686b5ffe1c519dae8566920a62f425019dcf4416593c19947af6d1R23) for [the completion modal](https://www.figma.com/file/uiZuCIqh2KYvBUfbCiJyyA/Hybrid-Learning?node-id=2254%3A6) If these related resources are supposed to be the same as those on the modal, some of this logic might help.

MisRob:
I've checked code focusing on the high-level architecture of side panels and having two components, `CurrentlyViewedResourceMetadata` and `BrowseResourceMetadata`, seems to be simpler to me too. Even from designs, there are relatively lots of differences, so I consider this direction to be good.

 Regarding duplicates, I wouldn't be too worried about it. We can always extract and start sharing areas that intersect. Might be rather my opinion - I find it simpler than guessing how to generalize. I see two areas that we could improve, not necessarily in this PR though:

**Metadata**

I agree that using `v-for` would most likely end up very complex since we also need to format various metadata etc. so I appreciate that you go one by one. It brings clarity to it. Maybe at some point, we could move this code to a metadata component that would receive all content node's metadata in `props` and it could also accept a `prop` with a list of metadata keys in the desired order by which to filter metadata to be displayed (e.g. `[..., 'language', 'author', 'licenseHolder', ...]` for `CurrentlyViewedResourceMetadata` and  `..., ['language', 'accessibility, 'whatYouWillNeed', 'author', 'licenseHolder' ...]` for `BrowseResourceMetadata` ). However maybe it'd be better to look into it after we have the final version of the metadata structure?

**Truncation of description text**

I hope this new [(work in progress) truncator](https://github.com/learningequality/kolibri/pull/8464) could help. I think it should be possible to add view more/less functionality.

kolibri/core/assets/src/views/FullScreenSidePanel/SidePanelContentMetadata.vue
```diff
@@ -0,0 +1,17 @@
+<template>
```
MisRob: Is this component something to be implemented in the near future?

kolibri/core/assets/src/views/FullScreenSidePanel/index.vue
```diff
@@ -127,15 +138,17 @@
     position: fixed;
     top: 0;
     right: 0;
-    z-index: 16;
-    width: 100vw;
+    bottom: 0;
+    // Must be <= 12 z-index so that KDropdownMenu shows over
```
MisRob, line 142: Thank you for all explanations for z-index values

kolibri/plugins/learn/assets/test/util/gen-content-link.spec.js
```diff
@@ -0,0 +1,34 @@
...
+    expect(topicLink.params.id).toEqual(19);
+  });
+});
+
+describe('genContentLink - generating for a topic (isLeaf == true)', () => {
+  it('returns a valid link object', () => {
+    expect(validateLinkObject(contentLink)).toBeTruthy();
+  });
+
+  it('returns an object with name pointing to PageName.TOPICS_TOPIC', () => {
```
MisRob: returns an object with name pointing to PageName.TOPICS_**CONTENT**

kolibri/plugins/learn/assets/src/views/LearningActivityChip.vue
```diff
@@ -0,0 +1,110 @@
...
+    [LearningActivities.CREATE]: 'create',
+    [LearningActivities.LISTEN]: 'listen',
+    [LearningActivities.REFLECT]: 'reflect',
+    [LearningActivities.PRACTICE]: 'practice',
+    [LearningActivities.READ]: 'read',
+    [LearningActivities.WATCH]: 'watch',
+    [LearningActivities.EXPLORE]: 'interact',
+  };
+
+  /**
```
MisRob, line 20: This doesn't describe the chip component precisely since it also displays a label in addition to an icon

kolibri/plugins/learn/assets/src/views/LearningActivityChip.vue
```diff
@@ -0,0 +1,110 @@
+<template>
+
+  <div class="activity-chip">
+    <KLabeledIcon :icon="icon" :label="coreString(kind)" class="activity-chip-label" />
```
MisRob: Was there any reason for re-implementing the whole `LearningActivityIcon` logic from scratch in this component instead of using `LearningActivityIcon` in the icon slot of `KLabeledIcon`?

```
<KLabeledIcon :label="coreString(kind)" class="activity-chip-label" />
  <template #icon>
    <LearningActivityIcon :kind="kind" />
  </template>
</KLabeledIcon>
```

MisRob: Regarding `:label="coreString(kind)"`, what is supposed to happen when `kind` is an array?

kolibri/plugins/learn/assets/src/views/LearningActivityChip.vue
```diff
@@ -0,0 +1,110 @@
...
+          } else {
+            return false;
+          }
+        },
+      },
+      /**
+       * Icon is solid by default.
+       * Set to `true` to make it shaded.
+       */
+      shaded: {
```
MisRob: From what I've seen in designs, I'd say that there's no need for the chip component to provide this settings in its API since chips are so small that they'll have `solid` learning activity icons:

Screenshot from 2021-10-12 13-11-10
Screenshot from 2021-10-12 13-11-30

kolibri/plugins/learn/assets/src/views/LearningActivityChip.vue
```diff
@@ -0,0 +1,110 @@
...
+  };
+
+</script>
+
+
+<style lang="scss" scoped>
+
+  .activity-chip {
+    display: inline-block;
+    padding: 12px; // Aligns with close button
```
MisRob: Was this meant to refer to the close button in side panels? If yes, I'd suggest moving this positioning CSS to side panels CSS (maybe `chips` class?) because it's specific to the particular usage of this component in the side panel context and doesn't have much to do with the chip component itself.

kolibri/plugins/learn/assets/src/views/BrowseResourceMetadata.vue
```diff
@@ -0,0 +1,365 @@
...
+          :text="metadataStrings.$tr('viewResource')"
+          appearance="raised-button"
+          :primary="false"
+          :to="genContentLink(content.id, content.is_leaf)"
+        />
+      </div>
+    </div>
+
+    <div class="section">
+      <Thumbnail :thumbnailUrl="content.thumbnail" />
```
MisRob: I made some updates to hybrid learning thumbnails and `Thumbnail` is now only a base component for other components (now I'm thinking it might make more sense to name it `BaseThumbnail`) Although this works, it won't display a learning activity icon when a thumbnail image is not available because this behavior is not part of `Thumbnail` implementation anymore. Could you please update thumbnails from `Thumbnail` to `ContentNodeThumbnail` so it works again as originally intended?

kolibri/plugins/learn/assets/src/views/BrowseResourceMetadata.vue
```diff
@@ -0,0 +1,365 @@
...
+
+    <KButton
+      v-if="descriptionOverflow"
+      :text="showMoreOrLess"
+      appearance="basic-link"
+      class="show-more-button"
+      :primary="true"
+      @click="toggleShowMoreOrLess"
+    />
+    <!-- this v-else ensures spacing remains consistent without show more -->
```
MisRob, line 68: Thank you, these template comments are helpful

kolibri/plugins/learn/assets/src/views/BrowseResourceMetadata.vue
```diff
@@ -0,0 +1,365 @@
...
+        {{ licenseShortName || '' }}
+        <KIconButton
+          :icon="licenseDescriptionIsVisible ? 'chevronUp' : 'chevronDown'"
+          :ariaLabel="metadataStrings.$tr('toggleLicenseDescription')"
+          size="small"
+          type="secondary"
+          class="license-toggle"
+          @click="licenseDescriptionIsVisible = !licenseDescriptionIsVisible"
+        />
+        <div v-if="licenseDescriptionIsVisible" dir="auto" class="license-details">
```
MisRob: Should automatic text directionality be used for all user-generated texts in Kolibri codebase or are there any other recommendations for `dir="auto"`? I can't understand why it's used only for the license description in this component (and often I am not sure when to add it to my code in Kolibri) so I'd be grateful for clarification.

kolibri/plugins/learn/assets/src/views/BrowseResourceMetadata.vue
```diff
@@ -0,0 +1,365 @@
...
+        licenseDescriptionIsVisible: false,
+        showMoreOrLess: 'Show More',
+        truncate: 'truncate-description',
+        descriptionOverflow: false,
+        metadataStrings: { $tr: () => null },
+      };
+    },
+    computed: {
+      licenseShortName() {
+        return licenseShortName(this.content.license_name);
```
MisRob: I'd suggest adding guards to check that `this.content` is truthy before accessing its attributes in this `computed` as well as in the following ones.

As long as `content` is indeed an `Object` as documented in `props`, accessing `this.content.<attribute>` should be fine, but I think that's rather information for developers and won't prevent from actually getting `undefined` or `null` through props so it could still fail here in such cases, I think?

kolibri/plugins/learn/assets/src/views/BrowseResourceMetadata.vue
```diff
@@ -0,0 +1,365 @@
...
+  .download-button {
+    margin-top: 16px;
+  }
+
+  .beginners-chip {
+    display: inline-block;
+    padding: 10px;
+    font-weight: bold;
+    color: white;
+    background: #328168; // brand.secondary.v_600
```
MisRob: Is there any reason for not using our brand color definition directly with computed class or style?

kolibri/plugins/learn/assets/src/views/BrowseResourceMetadata.vue
```diff
@@ -0,0 +1,365 @@
+<template>
```
MisRob, line 1: Most of the review comments for this component also apply to _CurrentlyViewedResourceMetadata.vue_

MisRob (approved):
Thank you @nucleogenesis.

Regarding code, I haven't noticed any blockers. It'd be good to have a look at the chip component and thumbnails.

Since Kolibri is still quite broken on this branch and side panels are work in progress, I didn't spend much time testing and only briefly checked panels toggling which seems to work well.

kolibri/core/assets/src/views/FullScreenSidePanel/SidePanelContentMetadata.vue
nucleogenesis: Oops - no I think that I started an outline component but bailed on it when I realized I wanted to make two separate metadata components - will remove this one :)

kolibri/plugins/learn/assets/test/util/gen-content-link.spec.js
nucleogenesis: Great catch! `jest` caught my copy-pasting error in the code but not in the comments :-p

kolibri/plugins/learn/assets/src/views/LearningActivityChip.vue
nucleogenesis: I have not implemented the list version because - well I don't know what the data structure will look like when we get it.

IMO it should just always be an array and it should be called `activityTypes` but without the metadata in place I didn't see reason to put it until we know the shape (or potential shapes) of the incoming data.

nucleogenesis: As for re-implementing, I think that I just hadn't really fully considered the existence of LearningActivytIcon - will import and use that here

kolibri/plugins/learn/assets/src/views/BrowseResourceMetadata.vue
nucleogenesis: The designs that I worked from didn't show the thumbnail so I just assumed that we wouldn't want the activity icon on the thumbnail in this context because the activity icon(s) are visible at the top of the side panel

kolibri/plugins/learn/assets/src/views/BrowseResourceMetadata.vue
nucleogenesis: Honestly - I think I may have put it there while trying to debug an RTL issue and forgot to remove it - or I copy-pasted it. 

Perhaps @indirectlylit would have a better understanding of when/how we should use that dir attribute. 

In any case, this could go very well in our [KDS Inclusive Design ](https://design-system.learningequality.org/inclusive/) documentation... which is where I looked in case there was guidance there :-)

kolibri/plugins/learn/assets/src/views/BrowseResourceMetadata.vue
nucleogenesis: I was tweaking my CSS in `devserver-hot` and didn't want to stop and do JS while I was in my styling groove :) - I'll fix this in my next pass.

kolibri/plugins/learn/assets/src/views/BrowseResourceMetadata.vue
indirectlylit: Good question!

* I have an open Kolibri issue with some guidance here: https://github.com/learningequality/kolibri/issues/3781
* Opened a new KDS issue here: https://github.com/learningequality/kolibri-design-system/issues/266

kolibri/plugins/learn/assets/src/views/BrowseResourceMetadata.vue
rtibbles: Yes, validation doesn't run in production, only in development.

kolibri/plugins/learn/assets/src/views/LearningActivityChip.vue
MisRob: > IMO it should just always be an array and it should be called activityTypes but without the metadata in place I didn't see reason to put it until we know the shape (or potential shapes) of the incoming data.

Good point. `activityTypes` will make much more sense than `kind` for `LearningActivityIcon` and similar components `props`. I don't recall why I allowed `String` type alongside `Array` but would say it wasn't anything important and might be caused by not knowing that we can have more activities for a content node in the beginning of development and adding support for that later. So I think it should be fine to simplify it to accept only `Array`. No need to do it in this PR for `LearningActivityIcon`

kolibri/plugins/learn/assets/src/views/BrowseResourceMetadata.vue
MisRob: The activity icon on the thumbnail only serves as a generic thumbnail placeholder when an image is not available so I don't consider having the activity icon somewhere close to the thumbnail a problem. There are already other places in the app when this is happening so if we really want to display the thumbnail here I'd suggest keeping it consistent. Here's one example [in designs](https://www.figma.com/file/uiZuCIqh2KYvBUfbCiJyyA/Hybrid-Learning?node-id=1915%3A82994) (note that later we made a decision to make activities icons on thumbnails a bit more colorful):

Screenshot from 2021-10-13 13-11-40

Besides that, if we'd displayed only gray background without anything else, I'd say that it might be a bit confusing because giving a false impression that it's a placeholder for an image that will load soon but that won't actually happen.

kolibri/plugins/learn/assets/src/views/BrowseResourceMetadata.vue
MisRob: Thank you

kolibri/plugins/learn/assets/src/views/BrowseResourceMetadata.vue
MisRob: Thank you both

kolibri/plugins/learn/assets/src/views/BrowseResourceMetadata.vue
nucleogenesis: Ahh this was my mistake - I thought that it was a thumbnail that would overlay the text of the activity like "Read" overlaid on the thumbnail like is shown in the content browsing experience. This makes much more sense - will have a fix shortly.

nucleogenesis:
@MisRob if all checks pass, this should be ready for a final review

pcenov:
Hi @nucleogenesis while testing the [buildkite build](https://buildkite.com/learningequality/kolibri-python-package/builds/3890) I noticed that I cannot open the side panel from the Learn>Classes Tab and also that the Close icon can overlap other controls when scrolling down:


Other than that all looks fine to me. I'll report the above mentioned issues separately if you say that they are in scope/not already known to you.

nucleogenesis:
@pcenov thank you! I'm following up on related work and will be sure to hook that up

PR #8351, opened 2021-08-26
Preseed the sqlite database

rtibbles:
## Summary
* Mainlines logic from the android installer to preseed a SQLite database for faster first time startup
* Skips update during our makemessages task to avoid unnecessary migrations
* Only runs upgrade logic when version upgrades are detected, not from scratch

## Reviewer guidance
Mainlined from https://github.com/learningequality/kolibri-installer-android/blob/develop/Makefile#L55

----

## Testing checklist

- [x] Contributor has fully tested the PR manually
- [ ] If there are any front-end changes, before/after screenshots are included
- [x] Critical user journeys are covered by Gherkin stories
- [x] Critical and brittle code paths are covered by unit tests

## PR process

- [x] PR has the correct target branch and milestone
- [x] PR has 'needs review' or 'work-in-progress' label
- [x] If PR is ready for review, a reviewer has been added. (Don't use 'Assignees')
- [ ] If this is an important user-facing change, PR or related issue has a 'changelog' label
- [ ] If this includes an internal dependency change, a link to the diff is provided

## Reviewer checklist

- Automated test coverage is satisfactory
- PR is fully functional
- PR has been tested for [accessibility regressions](http://kolibri-dev.readthedocs.io/en/develop/manual_testing.html#accessibility-a11y-testing)
- External dependency files were updated if necessary (`yarn` and `pip`)
- Documentation is updated
- Contributor is in AUTHORS.md

PR #8468, opened 2021-09-24
Preseed all the databases

rtibbles:
## Summary
* We started preseeding SQLite db files to improve first time startup
* We then added a bunch of new DB files to improve write concurrency
* This PR generalizes our preseeding logic to handle all defined additional SQLite databases
* It also allows preseeding of 'new' databases that are added at version upgrades, rather than only on fresh installations

## References
Follow up from https://github.com/learningequality/kolibri/pull/8442 and https://github.com/learningequality/kolibri/pull/8351

## Reviewer guidance
Does everything run fine? Do new databases get copied in and migrations avoided?

----

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/build_tools/preseed_home.py
+++ b/build_tools/preseed_home.py
@@ -1,0 +1,30 @@
+import os
+import shutil
+import tempfile
+
+temphome = tempfile.mkdtemp()
+os.environ["KOLIBRI_HOME"] = temphome
+
+from kolibri.main import initialize  # noqa E402
+from kolibri.deployment.default.sqlite_db_names import (  # noqa E402
+    ADDITIONAL_SQLITE_DATABASES,
+)
+from django.core.management import call_command  # noqa E402
+
+move_to = os.path.join(os.path.dirname(__file__), "..", "kolibri", "dist", "home")
+shutil.rmtree(move_to)
+os.mkdir(move_to)
+
+print("Generating preseeded home data in {}".format(temphome))
+
+initialize()
+call_command("deprovision", "--no-input")
+
+shutil.move(os.path.join(temphome, "db.sqlite3"), move_to)
+
+for db_name in ADDITIONAL_SQLITE_DATABASES:
+    shutil.move(os.path.join(temphome, "{}.sqlite3".format(db_name)), move_to)
+
+print("Moved all preseeded home data to {}".format(move_to))
+
+shutil.rmtree(temphome)
--- a/build_tools/preseed_home.py
+++ b/build_tools/preseed_home.py
@@ -18,7 +18,9 @@ os.mkdir(move_to)
 print("Generating preseeded home data in {}".format(temphome))
 
 initialize()
-call_command("deprovision", "--no-input")
+call_command(
+    "deprovision", "--destroy-all-user-data", "--permanent-irrevocable-data-loss"
+)
 
 shutil.move(os.path.join(temphome, "db.sqlite3"), move_to)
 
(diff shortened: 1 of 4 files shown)


kolibri/core/auth/management/commands/deprovision.py
```diff
@@ -60,13 +70,14 @@ def handle_async(self, *args, **options):
...
-            "ARE YOU SURE? If you do this, there is no way to recover the user data on this device."
-        )
+        if options["interactive"]:
+            # ensure the user REALLY wants to do this!
+            confirm_or_exit(
+                "Are you sure you wish to deprovision your database? This will DELETE ALL USER DATA!"
+            )
+            confirm_or_exit(
+                "ARE YOU SURE? If you do this, there is no way to recover the user data on this device."
+            )
```
jamalex, line 87: We had previously made the explicit choice not to make it that easy to deprovision without warnings, given how destructive it is.

If a one-liner is really needed, we've suggested `yes yes|kolibri manage deprovision` in the past. I guess that doesn't work on Windows, though.

jamalex: Ah, I see it's for the purposes of `call_command`, though -- I wonder about using a nonstandard name like `--accept-data-loss` or something to make it a bit harder to do accidentally.

I'm probably being paranoid. Except if there's anything we've learned from what users do.....

rtibbles: Yeah, I only added it because of the python invocation, happy to make it something ridiculous!

jamalex: Throw some unicode for a few different languages in there and we should be good, then!

build_tools/preseed_home.py
```diff
@@ -0,0 +1,32 @@
...
+os.environ["KOLIBRI_HOME"] = temphome
+
+from kolibri.main import initialize  # noqa E402
+from kolibri.deployment.default.sqlite_db_names import (  # noqa E402
+    ADDITIONAL_SQLITE_DATABASES,
+)
+from django.core.management import call_command  # noqa E402
+
+move_to = os.path.join(os.path.dirname(__file__), "..", "kolibri", "dist", "home")
+shutil.rmtree(move_to)
```
jredrejo: First time this is executed this line will trigger a `FileNotFoundError` error . It needs to be in a` try..except `

kolibri/utils/main.py
```diff
@@ -160,6 +161,28 @@ def _setup_django():
...
 
+def _copy_preseeded_db(db_name, target=None):
+    target = target or "{}.sqlite3".format(db_name)
+    target = os.path.join(KOLIBRI_HOME, target)
+    if not os.path.exists(target):
+        try:
+            import kolibri.dist
+
+            db_path = os.path.join(
+                kolibri.dist.__file__, "home/{}.sqlite3".format(db_name)
```
jredrejo: Using `__file__ `here is adding `__init__.py` to the path creating something like `.. kolibri/kolibri/dist/__init__.py/home/db.sqlite3 ` .
To make it work I had to change it to 
`os.path.join(kolibri.dist.__path__[0], "home/{}.sqlite3".format(db_name))`

jredrejo (changes requested):
I've found a couple of problems with the current code.
Also, I haven't found any where where 
`make preseeddb` is called, so I wonder what's the idea of this when used with the different installers (Debian, pex, Windows, etc.)

build_tools/preseed_home.py
rtibbles: Added `ignore_errors=True`

kolibri/utils/main.py
rtibbles: Wrapped in `os.path.dirname`

jredrejo (approved):
Everything looks fine, made some tests and it all worked without any problem

Issue #527, opened 2016-09-16
Versioning plugins and dependencies between plugins and kolibri


rtibbles:
Help ensure that standalone plugins are built/used with the correct version of Kolibri.

Static JS assets are now built into separate bundles such that plugins can use different versions of NPM dependencies (right @rtibbles ?)

So, the remaining piece of the puzzle is to be able to specify the versions of the Kolibri core which a plugin is dependent on. This should be enforced at build time but possibly also feed into the `setup.py` metadata such that the plugin isn't installed with the wrong version of Kolibri.

Finally, at Kolibri load-time, the plugin should fail to load if the wrong version of Kolibri is calling it.

We can find inspiration possibly by how Firefox and Thunderbird extensions depend on their host systems.

We should also ensure that all aspects of the core API are properly documented for each version, so it's clear when the API version needs to be updated. In particular, we currently use `apiSpec` to enumerate globally shared python modules. However this might not fully capture the API, including:

* core component props, events, and even potentially CSS- level behaviors (e.g. `display-inline` vs `display-block`)
* core CSS variables and functions
* python hooks
* probably others - anything that could break a plugin if we update something in Core

benjaoming:
Updated the description!

indirectlylit:
This is a high-level task that needs some planning and design before it can be addressed with a PR. Migrating to: https://trello.com/c/iRzHX80d/690-core-api-versioning

indirectlylit:
re-opening - back from Trello

PR #8470, opened 2021-09-27
Do not show password if not needed

### LINKED ISSUE #8390: password field is always shown during user import
## Observed behavior

The password field is always shown:

image

## Expected behavior

User should only be prompted to enter a password when necessary

## User-facing consequences

Unclear how to import learners from a device that has password-free login enabled. Leaving it blank seems to work, but this is a bit confusing.

## Context

latest develop

jredrejo:
## Summary
When syncing from a facility that has password-free login enabled, the setup wizard does not show the password input in order to sync an user:

image

## References
Closes #8390 

## Reviewer guidance
Start provisioning a new kolibri installation syncing users from another facility. Check that if the facility has the password-free login option enabled, only username is required to sync (and syncing keeps working).

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/plugins/setup_wizard/assets/src/api.js
+++ b/kolibri/plugins/setup_wizard/assets/src/api.js
@@ -2,7 +2,7 @@ import { Resource } from 'kolibri.lib.apiResource';
 import pickBy from 'lodash/pickBy';
 import urls from 'kolibri.urls';
 import redirectBrowser from 'kolibri.utils.redirectBrowser';
-import clientFactory from '../../../../core/assets/src/core-app/baseClient';
+import clientFactory from 'kolibri.utils.clientFactory';
 
 export const FacilityImportResource = new Resource({
   name: 'facilityimport',
--- a/kolibri/plugins/setup_wizard/assets/src/api.js
+++ b/kolibri/plugins/setup_wizard/assets/src/api.js
@@ -2,6 +2,7 @@ import { Resource } from 'kolibri.lib.apiResource';
 import pickBy from 'lodash/pickBy';
 import urls from 'kolibri.urls';
 import redirectBrowser from 'kolibri.utils.redirectBrowser';
+import clientFactory from '../../../../core/assets/src/core-app/baseClient';
 
 export const FacilityImportResource = new Resource({
   name: 'facilityimport',
@@ -55,6 +56,8 @@ export const SetupSoUDTasksResource = new Resource({
    */
   createTask(task, params) {
     const args = { task: task, ...pickBy(params) };
+    // use baseClient to  avoid interceptors from client:
+    this.client = clientFactory();
     return this.postListEndpoint('list', args).then(response => {
       return response.data;
     });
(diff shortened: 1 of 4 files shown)


radinamatic:
@pcenov Could you test this, please? Thank you!

pcenov:
Hi @jredrejo, I had a chance to test this PR today and the main scenario works correctly as described.
At the same time I noticed the following issues:

1 | 2
---- | ---
When I'm importing several users there doesn't seem to be a check whether the user has already been added so that the same user is listed several times (see below) | If I accidentally input an incorrect or not existing username I briefly see an "Incorrect username or password" validation message and then I'm automatically signed in with the last valid user. In this case the validation message is incorrect because I'm not inputting a password and also I should be given a chance to correct the mistake and add the correct user (see below)
Screenshot_20210930-190027_Kolibri | Screenshot_20210930-190048_Kolibri

jredrejo:
Hello @pcenov  are you by chance using a virtual machine to test these things? I am not able to reproduce the problems you're having, so if you're using a VM for the server maybe you could pass it to me

pcenov:
Hi @jredrejo I'm always testing with both real and virtual devices. In this case I'm experiencing the reported issues both on real and virtual devices. I'm on a sick leave today, will reach out to you next week to see why you are not able to replicate them.

jredrejo:
hello @pcenov  I finally could reproduce the two issues you had. It happened they existed since a long time and were not due to this PR and they were quite tricky to fix. Please, check it when you can. They should be fixed.

kolibri/plugins/setup_wizard/assets/src/api.js
```diff
@@ -2,6 +2,7 @@ import { Resource } from 'kolibri.lib.apiResource';
 import pickBy from 'lodash/pickBy';
 import urls from 'kolibri.urls';
 import redirectBrowser from 'kolibri.utils.redirectBrowser';
+import clientFactory from '../../../../core/assets/src/core-app/baseClient';
```
indirectlylit: if we cannot use `import client from 'kolibri.client'` then `clientFactory` should probably be added to `apiSpec.js` to avoid relative imports across the plugin/core boundary

indirectlylit (commented):
Always great to fix a long-standing issue!

@jredrejo left a note on the new relative import

> main scenario works correctly as described. At the same time I noticed the following issues

@pcenov in the future for this kind of situation it can be useful to file a follow-up issue. this can help the product team assign a priority to the work independently of the current PR

pcenov:
Perfect @jredrejo I've tested this again on Windows, Ubuntu and Android and everything works correctly as described. :)

kolibri/plugins/setup_wizard/assets/src/api.js
jredrejo: @indirectlylit  Thank you! I didn't want to alter the kolibri JS api but it's clear this can't cause any damage, so done!

indirectlylit: Yeah currently `apiSpec` contains both "external API" that external plugins reference as well as "internal API" that is only used in the kolibri repo and is free to change at any time. @rtibbles and I have discussed making this distinction more explicit and also starting to version the external parts (https://github.com/learningequality/kolibri/issues/527)

indirectlylit (approved):
thank you!

PR #8476, opened 2021-10-01
API changes to extend custom presentation

danigm:
## Summary

Extend the custom content renderer javascript API to be able to do advanced content query. This is the list of changes made to the API:

 * Add pagination parameters to `getContentByFilter` and `searchContent` queries and responses.
 * Filter all queries in the API by the channel so there's no nodes outside of the channel scope.
 * Add more options to `getContentByFilter`: onlyTopics, onlyContent, kinds, authors, tags and random
 * Add node thumbnail url to the node serialization
 * Add a new `getChannelMetadata` API method
 * Add a new  `getChannelFilterOptions` API method to get available authors, content kinds and tags in a channel
 * New random filter for content nodes that sorts the nodes randomly
 * New filter by tags for content nodes

----

## Testing checklist

- [x] Contributor has fully tested the PR manually. We are updating our custom presentation for the Endless Key using this new API and we're testing all the new methods and calls.

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated


kolibri/core/content/api.py
```diff
@@ -455,6 +455,11 @@ def process_items(self, items, queryset, ancestor_lookup_method=None):
             if ancestor_lookup_method:
                 item["ancestors"] = ancestor_lookup_method(item)
             item["is_leaf"] = item.get("kind") != content_kinds.TOPIC
+            children = models.ContentNode.objects.filter(parent=item["id"])
```
rtibbles: This will cause a massive performance regression, btw - so we cannot merge this as is.

rtibbles:
See here for the latest updates to the contentnode backend re pagination and metadata: https://github.com/learningequality/kolibri/pull/8488

kolibri/core/content/api.py
danigm: I've just removed this commit for now. We want to know the number of children and topic children to be able to detect the last level in the tree (if it has no topic children) and also to show some information in topic nodes, the number of children.

But we can rethink all of this frontend functionality if we don't find an efficient way to add this information to nodes.

manuq:
> See here for the latest updates to the contentnode backend re pagination and metadata: #8488

OK, that one got merged so we should rebase then.

danigm:
I've detected that the random filter in this PR is not random anymore with the new pagination system. I'll take a look to this to try to solve.

danigm:
> I've detected that the random filter in this PR is not random anymore with the new pagination system. I'll take a look to this to try to solve.

After some testing, I can confirm that the random filter in combination with the pagination is not a solution anymore, so I've implemented the random query as a new API method: https://github.com/endlessm/kolibri/commit/5c94506f906638cea8a4fb3f0ed6d0533515cac6

kolibri/plugins/learn/assets/src/views/ChannelRenderer/CustomContentRenderer.vue
```diff
@@ -114,19 +132,33 @@
...
+          return createReturnMsg({ message, err });
+        }
+
         return ContentNodeResource.fetchCollection({
           getParams: {
             ids: options.ids,
+            authors: options.authors,
+            tags: options.tags,
             parent: options.parent === 'self' ? this.topic.id : options.parent,
+            channel_id: this.topic.channel_id,
```
rtibbles: The parent option should be used for filtering here - this will prevent custom renderers from displaying any content from outside their channels.

Alternatively, can add a `sameChannel` option to toggle this.

kolibri/plugins/learn/assets/src/views/ChannelRenderer/CustomContentRenderer.vue
```diff
@@ -114,19 +132,33 @@
...
+            max_results: options.maxResults ? options.maxResults : 50,
+            cursor: options.cursor,
+            kind: onlyTopics ? ContentNodeKinds.TOPIC : undefined,
+            kind_in: onlyContent ? allButTopicTypes : kinds,
```
rtibbles: I think this would be better done as an exclude filter on topic.

kolibri/plugins/learn/assets/src/views/ChannelRenderer/CustomContentRenderer.vue
```diff
@@ -114,19 +132,33 @@
...
+
         return ContentNodeResource.fetchCollection({
           getParams: {
             ids: options.ids,
+            authors: options.authors,
+            tags: options.tags,
             parent: options.parent === 'self' ? this.topic.id : options.parent,
+            channel_id: this.topic.channel_id,
+            max_results: options.maxResults ? options.maxResults : 50,
+            cursor: options.cursor,
```
rtibbles, line 187: I think rather than passing `cursor` here - it would be better to define a separate `fetchMore` pathway, and the `more` object that was returned can just be returned back. Otherwise, mixing the cursor with potential other changes to parameters could lead to unexpected results.

kolibri/plugins/learn/assets/src/views/ChannelRenderer/CustomContentRenderer.vue
```diff
@@ -245,6 +281,75 @@
         const newMsg = createReturnMsg({ message, data: __version });
         return this.hashi.mediator.sendMessage(newMsg);
       },
+      sendChannelMetadata(message) {
+        return ChannelResource.fetchModel({ id: this.topic.channel_id })
```
rtibbles: We can retrieve this from bootstrapped metadata now, see this composable: https://github.com/learningequality/kolibri/blob/release-v0.15.x/kolibri/plugins/learn/assets/src/composables/useChannels.js

This will avoid a request.

kolibri/plugins/learn/assets/src/views/ChannelRenderer/CustomContentRenderer.vue
```diff
@@ -245,6 +281,75 @@
...
+      },
+      sendRandomCollection(message) {
+        const { options } = message;
+        const { kinds, onlyContent } = options;
+
+        return ContentNodeResource.fetchRandomCollection({
+          getParams: {
+            parent: options.parent === 'self' ? this.topic.id : options.parent,
+            channel_id: this.topic.channel_id,
+            max_results: options.maxResults ? options.maxResults : 10,
```
rtibbles, line 356: Pagination forces an ordering, so I am guessing this is not working.

rtibbles (changes requested):
Mostly seems mergeable, but some tweaks that I think would be needed.

kolibri/plugins/learn/assets/src/views/ChannelRenderer/CustomContentRenderer.vue
danigm: I've added a new option `limitToChannel` to toggle this. With the parent we limit to just direct children, but with this option we can get more than one level at once, so this is more flexible.

kolibri/plugins/learn/assets/src/views/ChannelRenderer/CustomContentRenderer.vue
danigm: I've discovered that this filter is already implemented in the `kind` filter, we just need to pass the value `content`. So I've updated the logic to use that filter instead of using the `kind_in` or create a new one.

kolibri/plugins/learn/assets/src/views/ChannelRenderer/CustomContentRenderer.vue
danigm: Okay, I didn't thought about that problem. I've created a new method in the API, `getContentPage`, to request new pages with the cursor and just removed the `cursor` from the `getContentByFilter`.

kolibri/plugins/learn/assets/src/views/ChannelRenderer/CustomContentRenderer.vue
danigm: I've just updated this API call to use the `useChannels` function instead of making a request to the backend.

kolibri/plugins/learn/assets/src/views/ChannelRenderer/CustomContentRenderer.vue
danigm: This new API method doesn't use the cursor pagination, it doesn't have any pagination, the `max_results` is only used to limit the number of nodes, but it's not possible to get more with the same "random" order, each call should provide a new set of randoms.

https://github.com/learningequality/kolibri/pull/8476/commits/6225201694aa59aa129bd9f97f4acb06656dfcc7#diff-f2b97d70d258e7c248156962c7e63fafe868abced1c0644b8b25a0aa1d0addd2

rtibbles:
Some linting issues, can run:
`pre-commit run --all-files` and commit the results to fix.

rtibbles (approved):
This all seems reasonable to me - doesn't seem like a huge maintenance burden by continuing to support all of these.

rtibbles:
Linting failure has been fixed on release branch, merging.

PR #8488, opened 2021-10-08
Content metadata fields, API, and Search

rtibbles:
## Summary
* Adds backend fields for new metadata fields
* Updates content schema and tests for content import
* Adds upgrade migration for content kinds to learning activities for those that can be automatically mapped
* Adds import migrations for content kinds to learning activities when importing from older database schemas
* Updates ContentNode API endpoints to return the new metadata as arrays of ids
* Adds ValuesViewset compatible cursor pagination for use with the ContentNode endpoint
* Updates LimitOffset pagination to fix some errors in previous implementation
* Adds filtering of the ContentNode endpoint by the new metadata fields, using `contains` queries
* Adds a new 'keyword' filter that implements a simpler version of the behaviour of the ContentNodeSearchViewset
* Integrates the Library Search interface with backend filtering
* Cleans up and deletes a lot of previous frontend Search code and defers all searching to the Library page
* Adds annotation and caching of labels that are available for queries
* Bootstraps available labels for all available nodes into the learn page
* Uses bootstrapped available labels to restrict the filters that are shown to users in search (off by default in development for now)
* Returns available labels for the query when paginated requests are made to the ContentNode endpoint
* Uses available labels from a search to disable filter options

## References
Fixes #8311 
Fixes #8312 
Fixes #8313
Fixes #8493

## Reviewer guidance
Caveat - the introduction of pure ids into this PR means that areas of the frontend code that were relying on english string ids coming from the backend are now broken.

Test the Library page and confirm that searching is working as might be expected.

1. Filtering by Learning Activity - ensure that filtering by Learning Activity does indeed produce a list of content for just that learning activity (although rendering of both the content and the icons for the learning activities will currently be broken).
2. Filtering by channel - ensure that filtering by channel produces results from within that channel only
3. Filtering by language - ensure that filtering by languages produces search results only for that language
4. Filter by keyword search - ensure that filtering by keyword searches produces somewhat sensible results (plenty of room for interpretation here, but compare against previous keyword search behaviour!)

In all of the above cases, ensuring that the offered filter options match the content that has been loaded onto the device.

Running from production assets specific:
* Ensure that only filters that are present on the local device metadata (i.e. learning activities, language, channels) appear to the end user.

Things that are definitely broken:
* Rendering content
* Most card metadata display

search

----

## Testing checklist

- [x] Contributor has fully tested the PR manually
- [x] If there are any front-end changes, before/after screenshots are included
- [ ] Critical user journeys are covered by Gherkin stories
[...]

PR #8492, opened 2021-10-11
Content metadata optimizations

rtibbles:
## Summary
* Adds BigIntegerField bitmask columns for every 64 allowed labels in each new categorical metadata column
* Partitions the labels between the columns and assigns a bit value to each based on position in the list of labels represented in that column
* Adds a `has_labels` method to the ContentNode queryset that does bitwise matching of one or more labels for a specific categorical metadata field
* Updates API endpoints to use this - seems to give a roughly 50x speedup over LIKE lookups
* Adds upgrade and annotation logic for setting these bitmasks on upgrade and at channel import time
* Uses these bitmasks for the 'available label' queries also
* Adds a new 'ancestors' JSON field onto ContentNodes (but not on the importable schema)
* Adds function to annotate the ancestors fields for ContentNodes in a channel
* Adds an upgrade step to apply to all existing channels
* Adds to channel import to run this step after channel import
* Removes ancestor querying logic from API endpoints and just loads this value directly instead, this saves a query that was particularly poorly performing for search results, due to the need to lookup ancestors across multiple tree locations - eliminating this allows search results to return in ~6ms on my dev machine, if the cache for available labels is hot.

## References
Follow up to #8488

## Reviewer guidance
Any areas of the code where more comments might be useful?
Any way to optimize the available labels code further?
Still needs tests.

----

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated


kolibri/core/content/migrations/0029_metadata_bitmasks.py
```diff
@@ -0,0 +1,41 @@
...
+        ),
+        migrations.AddField(
+            model_name="contentnode",
+            name="categories_bitmask_0",
+            field=models.BigIntegerField(blank=True, default=0, null=True),
+        ),
+        migrations.AddField(
+            model_name="contentnode",
+            name="grade_levels_bitmask_0",
+            field=models.BigIntegerField(blank=True, default=0, null=True),
```
jredrejo, line 29: do we really  need BigIntegerField (64 bits) for these fields? or it's just to make it equal thus simpler? Because I understand it for categories (it even could not be enough for that case) but I can't imagine 64 levels or 64 learning activities

kolibri/core/content/models.py
```diff
@@ -162,6 +185,10 @@ class ContentNode(base_models.ContentNode):
     # then it is 1 or 0 depending on availability
     on_device_resources = models.IntegerField(default=0, null=True, blank=True)
 
+    # Use this to annotate ancestor information directly onto the ContentNode, as it can be a
+    # costly lookup
+    ancestors = JSONField(default=[], null=True, blank=True)
```
jredrejo, line 190: could you comment why do we need the ancestors? I can barely understand how they are calculated but don't see why they are needed when searching.

jredrejo (commented):
It's impossible to test this PR using the frontend as https://github.com/learningequality/kolibri/blob/release-v0.15.x/kolibri/plugins/learn/assets/src/views/EmbeddedSidePanel/ActivityButtonsGroup.vue#L56 is broken (`plugin_data.learningActivities` is undefined)
is there a known workaround (appart from trying to debug and solve it) ?

kolibri/core/content/migrations/0029_metadata_bitmasks.py
rtibbles: I did originally implement it to be smart and choose BigIntegerField or just IntegerField depending on the length. I am happy to reinstate that cleverness, but I thought just doing the simple thing and adding smarter choices by request was better :)

kolibri/core/content/models.py
rtibbles: They're not needed at all by searching - the reason I added this here is that we annotate these onto the search results in the API for consistent return data across all ContentNode return - and it was the main thing causing slow down once the actual search query was setup.

This is separable from the search work, but does provide a very good speedup, so I think this is worth including (also, the tree structure only changes at very predictable times, so this seems like a low risk thing to do).

rtibbles:
Sorry @jredrejo - looks like I need to map from snake case to camelCase after my last commit, will update.

jredrejo:
> Sorry @jredrejo - looks like I need to map from snake case to camelCase after my last commit, will update.

thanks, let me know when I can test it

rtibbles:
Have updated so that it can be tested - have not made any updates as yet changing the integer fields.

jredrejo (commented):
Code looks good to me (there's an odd linting error that has not been introduced in this commit and it's already resolved in the v0.15.x  branch) so rebasing should solve it.
I have tested it as much as I have been able and searching,  upgrades & migrations work. Performance is hard to compare without a previous reference but the app works fine on my laptop.

There are a good amount of frontend errors that seem to be related to things this PR is not affecting, but just in case, this is the list I have seen:

- Continue missing icon errors at Sub.svgIconComponent
- http://localhost:8080/es-es/learn/ page keeps in blank with the loading animation forever until one tab is clicked
- No way to clean the search filters 
- `vue-router.esm.js:2062 Uncaught (in promise) Error: Avoided redundant navigation to current location: "/library?languages=es".` error after searching filtering by language or channel

I think all these problems are related to different PR or issues, but I have doubts with the latest one. If this PR is not causing this issue, we can approve it

rtibbles:
Yeah, I don't think any of these are caused by this PR.

jredrejo (approved):

rtibbles:
Reverting to draft here until I can update the bit generation to come stably and consistently from le-utils.

Issue #7960, opened 2021-04-06
Collect enhanced telemetry data


rtibbles:
Get more accurate and better reporting of user devices, including:
* Browser information of client devices
* Accurate installer information
* Information about apps and other 'launch' modes (e.g. cherrypy, uwsgi, gunicorn, apache)
* Project specific tagging so we can trace more information about projects we are working directly with

Installer information spec:
https://docs.google.com/spreadsheets/d/1EpN4yIO9H0qkP0pY6jiEtfQIdyrEPChs3aDZwOMupcI/edit#gid=0

*If the task runner has been invoked either with `CHERRYPY_START` set to `False` and/or has been invoked using `kolibri services`. Otherwise this will be `False`.

jamalex:
@rtibbles Another piece of this is presumably including some aggregate stats about browsers, which we now track via the session endpoint? I'm not sure where we're tracking that.

jamalex:
>Accurate installer information

I have copied and updated the table in the original issue description into this spreadsheet, and bolded the places where I am proposing something different than what was laid out there:
https://docs.google.com/spreadsheets/d/1EpN4yIO9H0qkP0pY6jiEtfQIdyrEPChs3aDZwOMupcI/edit#gid=0
(important notes have been added to the last column)

>Information about apps and other 'launch' modes (e.g. cherrypy, uwsgi, gunicorn, apache)

We're folding the "app" info into the installer field, as they're tightly coupled. Determining that gunicorn, apache, etc are being used doesn't seem feasible, given that in those scenarios the task runner would need to be run separately anyway, so it would be hard to pick it up from the context of the ping.

>Project specific tagging so we can trace more information about projects we are working directly with

It looks as if this is already in place on the Kolibri side, with:
`"project": conf.OPTIONS["Deployment"]["PROJECT"]`

>Browser information of client devices

It looks as if this is in place as well:
`"dis": usersess_devinf,`

jredrejo:
I miss here installations done with `pex` and with `pip`.

jamalex:
The table that was in the initial issue description was not the latest -- I updated the table (including adding pex/pip) in the spreadsheet linked in my last comment:
https://docs.google.com/spreadsheets/d/1EpN4yIO9H0qkP0pY6jiEtfQIdyrEPChs3aDZwOMupcI/edit#gid=0
(so that should be considered the canonical values)

rtibbles:
This Epic was resolved with the release of 0.15.0.

PR #8500, opened 2021-10-14
Use, when possible, an ENV var to get the installation type

### LINKED ISSUE #7957: Update installation_type function to read from an environment variable
## Observed behavior
Currently the `installation_type` function: https://github.com/learningequality/kolibri/blob/develop/kolibri/utils/server.py#L564 uses a series of sniffing steps to try to infer how the package in question was installed.

The reliance on the `sys.argv` sniffing in particular causes issues in disambiguating the `kolibri` debian package from invocations of Kolibri from within the `kolibri-server` debian package. In the device info page, this works, as the requests are being handled by a process launched by uwsgi - however, for analytics pingbacks, the worker process is not launched by uwsgi, so the wrong info is inferred.

## Expected behavior
We should instead set an environment variable `KOLIBRI_INSTALLATION_TYPE` inside the invocation scripts of each of our installers, and use that to infer the proper type.

jredrejo:
## Summary

Whenever it's available, use the `KOLIBRI_INSTALLATION_TYPE` env var to show the installation type instead of trying to guess it 

In any case, with or without the env var, the output of the installation type follow the specs at https://docs.google.com/spreadsheets/d/1EpN4yIO9H0qkP0pY6jiEtfQIdyrEPChs3aDZwOMupcI/edit#gid=0

Example:
image

## References
Closes: #7957 

## Reviewer guidance
Do the new constants names and descriptions make sense?

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/utils/constants/installation_types.py
+++ b/kolibri/utils/constants/installation_types.py
@@ -1,0 +1,22 @@
+"""
+This module contains constants representing the type of "installers" used to install Kolibri.
+"""
+from __future__ import unicode_literals
+
+APK = "apk"
+APT = "apt"
+KOLIBRI_SERVER = "kolibri-server"
+MACOS = "Mac"
+PEX = "pex"
+WHL = "whl"
+WINDOWS = "Windows"
+
+install_type_map = {
+    APK: "APK",
+    APT: "APT - Debian package",
+    KOLIBRI_SERVER: "kolibri(apt) with kolibri_server",
+    MACOS: "MacOS Installer",
+    PEX: "PEX file",
+    WHL: "WHL Python package",
+    WINDOWS: "Windows Installer",
+}
--- a/kolibri/utils/constants/installation_types.py
+++ b/kolibri/utils/constants/installation_types.py
@@ -8,6 +8,7 @@ APT = "apt"
 KOLIBRI_SERVER = "kolibri-server"
 MACOS = "Mac"
 PEX = "pex"
+UWSGI = "uwsgi"
 WHL = "whl"
 WINDOWS = "Windows"
 
@@ -17,6 +18,7 @@ install_type_map = {
     KOLIBRI_SERVER: "kolibri(apt) with kolibri_server",
     MACOS: "MacOS Installer",
     PEX: "PEX file",
+    UWSGI: "UWSGI process",
     WHL: "WHL Python package",
     WINDOWS: "Windows Installer",
 }
--- a/kolibri/utils/constants/installation_types.py
+++ b/kolibri/utils/constants/installation_types.py
@@ -4,7 +4,7 @@ This module contains constants representing the type of "installers" used to ins
 from __future__ import unicode_literals
 
 APK = "apk"
-APT = "apt"
+DEB = "deb"
 KOLIBRI_SERVER = "kolibri-server"
 MACOS = "Mac"
 PEX = "pex"
@@ -14,10 +14,10 @@ WINDOWS = "Windows"
 
 install_type_map = {
     APK: "APK",
-    APT: "APT - Debian package",
-    KOLIBRI_SERVER: "kolibri(apt) with kolibri_server",
-    MACOS: "MacOS Installer",
-    PEX: "PEX file",
+    DEB: "Debian package",
+    KOLIBRI_SERVER: "kolibri Debian package with kolibri_server",
+    MACOS: "macOS desktop app",
+    PEX: "PEX executable",
     UWSGI: "UWSGI process",
     WHL: "WHL Python package",
     WINDOWS: "Windows Installer",
--- a/kolibri/utils/constants/installation_types.py
+++ b/kolibri/utils/constants/installation_types.py
@@ -5,20 +5,24 @@ from __future__ import unicode_literals
 
 APK = "apk"
 DEB = "deb"
-KOLIBRI_SERVER = "kolibri-server"
-MACOS = "Mac"
+FLATPAK = "flatpak"
+GNOME = "gnome"
+KOLIBRI_SERVER = "kolibriserver"
+MACOS = "mac"
 PEX = "pex"
-UWSGI = "uwsgi"
 WHL = "whl"
-WINDOWS = "Windows"
+WINDOWS = "windows"
+WINDOWS_APP = "windowsapp"
 
 install_type_map = {
-    APK: "APK",
-    DEB: "Debian package",
-    KOLIBRI_SERVER: "kolibri Debian package with kolibri_server",
-    MACOS: "macOS desktop app",
-    PEX: "PEX executable",
-    UWSGI: "UWSGI process",
-    WHL: "WHL Python package",
-    WINDOWS: "Windows Installer",
+    APK: "apk - {}",
+    DEB: "deb kolibri - {}",
+    FLATPAK: "Flatpak - {}",
+    GNOME: "GNOME - {}",
+    KOLIBRI_SERVER: "deb kolibri-server - {}",
+    MACOS: "Mac - {}",
+    PEX: "pex",
+    WHL: "whl",
+    WINDOWS: "Windows - {}",
+    WINDOWS_APP: "Windows App - {}",
 }
--- a/kolibri/utils/server.py
+++ b/kolibri/utils/server.py
@@ -914,7 +914,7 @@ def get_urls(listen_port=None):
         return e.status_code, []
 
 
-def get_installer_version(installer_type):
+def get_installer_version(installer_type):  # noqa: C901
     def get_debian_pkg_version(package):
         """
         In case we want to distinguish between dpkg and apt installations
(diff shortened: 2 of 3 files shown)


kolibri/utils/constants/installation_types.py
```diff
@@ -0,0 +1,24 @@
...
+APT = "apt"
+KOLIBRI_SERVER = "kolibri-server"
+MACOS = "Mac"
+PEX = "pex"
+UWSGI = "uwsgi"
+WHL = "whl"
+WINDOWS = "Windows"
+
+install_type_map = {
+    APK: "APK",
```
rtibbles: This seems like a good way of constraining the values that end up in our analytics database - @jamalex any concerns with centralizing the possible values here?

kolibri/utils/constants/installation_types.py
```diff
@@ -0,0 +1,24 @@
...
+    APT: "APT - Debian package",
+    KOLIBRI_SERVER: "kolibri(apt) with kolibri_server",
+    MACOS: "MacOS Installer",
+    PEX: "PEX file",
+    UWSGI: "UWSGI process",
+    WHL: "WHL Python package",
+    WINDOWS: "Windows Installer",
+}
```
indirectlylit: I'm a bit confused why there are two inconsistent variations of these strings.

Also, if these are permanent strings that cannot be changed we might want to tighten them a bit and make them a bit more consistent and correct:

* package manager, file format, or operating system is the primary piece of info?
* "apt" is spelled two different ways above
* "MacOS" is actually spelled "macOS", and it's technically not an "installer", just an "app"
* For "Windows" we might want to distinguish the current "server" version from a potential future "app" version more similar to mac



```suggestion
ANDROID = "Android APK"
DEBIAN = "Debian package"
DEBIAN_APT = "Debian APT-installed package"
DEBIAN_APT_SERVER = "Debian APT package with kolibri-server"
MACOS_APP = "macOS desktop app"
PEX = "PEX executable"
UWSGI = "UWSGI process"
WHL = "WHL Python package"
WINDOWS_BG = "Windows background process"

runtime_types = (
    ANDROID,
    DEBIAN,
    DEBIAN_APT,
    DEBIAN_APT_SERVER,
    MACOS_APP,
    PEX,
    UWSGI,
    WHL,
    WINDOWS_BG,
)
```

indirectlylit (commented):
Slightly pedantic but possibly important: these are generally "runtime" types, not "installation" types.

kolibri/utils/constants/installation_types.py
rtibbles: The spec for this is here: https://github.com/learningequality/kolibri/issues/7960 this does need some updates to fit that spec (such as being able to insert the version number), but better to iterate there. Also, as there is already historic data in this format for some installers, updates may not be preferred for consistent analytics reporting.

indirectlylit: Ah, didn't realize there there was already data. At the least then there should only be one set of strings, not two? And we can still choose more consistent constant names even if the actual strings are a bit noisy.

jredrejo: @rtibbles great, I was not aware of #7960 happy to fix these inconsistences after knowing the details. 

@indirectlylit I agree with some of your changes (I didn't know we  have an app for MacOs instead of an installer) but not with others, For Windows, for example installation type is clearly an installer, not a background process.
Also , there's not practical way to know if a Debian package has been installed using `dpkg` or `apt`, the current logic tries to find it out, but it can be easily cheated. In my opinion, we should just say "Deb package" removing the `apt` reference.

indirectlylit: for what it's worth these comments are not critical or intended to be blocking.

> For Windows, for example installation type is clearly an installer, not a background process.

Is it the "installation type" we care about or the runtime app? Sometimes it's both but often they are independent. This is like the Debian package which can be installed either with `dpkg` or `apt`.

The APK can be installed by sideloading or via the app store.

The [current Windows repo](https://github.com/learningequality/kolibri-installer-windows) contains three things:
* an installer script which runs once
* a background process (`kolibri.exe`) which runs Kolibri in the background, potentially on start-up
* a small GUI which allows some basic config and control

Additionally at some point we'll also have a Windows "foreground app" similar to the Endless, Android, and macOS apps [based on this repo](https://github.com/learningequality/kolibri-app). When this happens, it may also have its own installer.

> In my opinion, we should just say "Deb package" removing the apt reference.

agreed!

jredrejo: > Is it the "installation type" we care about or the runtime app? 

well, this is a part of a function called `installation_type` :smile: , probably knowing the runtime app is useful too, but when trying to get information from the user to solve issues, knowing how Kolibri was installed is important.

kolibri/utils/constants/installation_types.py
jamalex: It helps with having this centrally documented/enforced, so the various installers don't go and do their own things, yeah.

Main thing is that it'd need to be considered a template/prefix, rather than a simple equivalence match.

jamalex:
Main thing here is updating to match the spec:
https://github.com/learningequality/kolibri/issues/7960#issuecomment-955072973

(and, as noted in the comment, ensuring that the set of constants are treated as prefixes, rather than exact matches)

jredrejo:
@jamalex  done according to the specs (I have updated the description of this PR accordingly).
Now it's on the respective installers to set the env var if we want to be exact, if not , the  `installation_type` function tries to guess it. Issues for this are linked in #7957 
For the version, I know how to get the version of the `apk`, `deb` and `kolibri-server` installers but not for the rest, so by now, the version is not returned.

jamalex (approved):
This looks good to me overall, functionally!

I note that it seems the `install_type` variable is used to hold the keys from the constants, rather than the values, but we also store the value from the envvar in there. I had been imagining that the envvar passed in from the installer would be the full value (including the version, etc). Do you see any issues with that, either on the installer side (detecting version etc there, to put in the envvar) or here, with the codepaths (it seems like it would probably just skip over the other code and directly return the value from the envvar, so we'd be fine here)? It's nice having the fallback logic here just in case, too.

jredrejo:
> This looks good to me overall, functionally!
> 
> I note that it seems the `install_type` variable is used to hold the keys from the constants, rather than the values, but we also store the value from the envvar in there. I had been imagining that the envvar passed in from the installer would be the full value (including the version, etc). Do you see any issues with that, either on the installer side (detecting version etc there, to put in the envvar) or here, with the codepaths (it seems like it would probably just skip over the other code and directly return the value from the envvar, so we'd be fine here)? It's nice having the fallback logic here just in case, too.

My  main concern is having envvars with spaces or other chars to separate version and installer type, I am not sure that will work correctly in the different combinations of OS/installers we have. If you prefer the installer to provide the version I'd go for one envvar for the installer type and another for the version.
I have added a `KOLIBRI_INSTALLER_VERSION` envvar that, if set by the installer, will be used by kolibri instead of guessing it out.

kolibri/utils/server.py
```diff
@@ -913,77 +914,116 @@ def get_urls(listen_port=None):
...
+        In case we want to distinguish between dpkg and apt installations
+        we can use apt-cache show madison and compare versions with dpkg
+        if dpkg > madison, it's dpkg otherwise it's apt
+        """
+        try:
+            output = check_output(["dpkg", "-s", package])
+            if hasattr(output, "decode"):  # needed in python 2.x
+                output = output.decode("utf-8")
+            package_info = output.split("\n")
+            version_info = [output for output in package_info if "Version" in output]
```
indirectlylit, line 929: Not sure why this line didn't raise an issue before, but it's now causing a python 2 syntax check flake8 failure:

```
kolibri/utils/server.py:929:40: F812 list comprehension redefines 'output' from line 927
```
https://github.com/indirectlylit/kolibri/runs/4301449205?check_suite_focus=true

Perhaps this should be:

```python
             version_info = [info for info in package_info if "Version" in info]
```

Weirdly, I see the linting error in my fork but not in the LE org repo.

cc @jredrejo @rtibbles

indirectlylit: interesting, the check is skipped in LE but not in my repo:

https://github.com/learningequality/kolibri/runs/4301448839?check_suite_focus=true

image

rtibbles: Oh - probably because the path match check looks back through the history of runs (which are constrained to a specific fork) so because the learning equality repo had a test run that had previously passed for this set of files, it doesn't run again, whereas your fork didn't.

indirectlylit: filed a follow-up here: https://github.com/learningequality/kolibri/issues/8772

jredrejo: weird, that's not new code, it has been there since kolibri 0.11

rtibbles: I think it is more likely a floating linter version issue than an issue with the CI setup.

PR #8501, opened 2021-10-15
Update Content Renderer to use Props

marcellamaki:
- Switches from using mapState to using props passed by `LearnImmersiveLayout`. This is to fix the issues with the content renderer, and now it seems to be working correctly, no matter which page it is accessed from. 

Reviewer guidance: 
Any edge cases I am missing?

### DIFF
--- a/kolibri/plugins/learn/assets/src/views/ContentPage.vue
+++ b/kolibri/plugins/learn/assets/src/views/ContentPage.vue
@@ -96,6 +96,28 @@
       AssessmentWrapper,
     },
     mixins: [commonLearnStrings],
+    props: {
+      content: {
+        type: Object,
+        required: true,
+        default: null,
+      },
+      contentId: {
+        type: String,
+        required: true,
+        default: null,
+      },
+      channelId: {
+        type: String,
+        required: true,
+        default: null,
+      },
+      contentKind: {
+        type: String,
+        required: true,
+        default: null,
+      },
+    },
     data() {
       return {
         wasIncomplete: false,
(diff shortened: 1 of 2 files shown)


kolibri/plugins/learn/assets/src/views/ContentPage.vue
```diff
@@ -96,6 +96,28 @@
       AssessmentWrapper,
     },
     mixins: [commonLearnStrings],
+    props: {
+      content: {
+        type: Object,
+        required: true,
+        default: null,
+      },
+      contentId: {
```
rtibbles: Can replace this with a computed prop:
```
this.content && this.content.content_id || null
```

kolibri/plugins/learn/assets/src/views/ContentPage.vue
```diff
@@ -96,6 +96,28 @@
...
+        type: String,
+        required: true,
+        default: null,
+      },
+      channelId: {
+        type: String,
+        required: true,
+        default: null,
+      },
+      contentKind: {
```
rtibbles: Can replace this with a computed prop:
```
this.content && this.content.kind || null
```

rtibbles (commented):
One suggested tweak, but I think this makes perfect sense.

kolibri/plugins/learn/assets/src/views/ContentPage.vue
```diff
@@ -96,6 +96,28 @@
       AssessmentWrapper,
     },
     mixins: [commonLearnStrings],
+    props: {
+      content: {
+        type: Object,
+        required: true,
+        default: null,
+      },
```
indirectlylit: Props also have the advantage of natively supporting validator functions which can help encode some expectations about the shape and values of the `Object` being passed in. (Vuex does not have this natively though we experimented with schemas in #5332).

A very simple, better-than-nothing validator might simply check that `content` has all required attributes, for example like this one:

https://github.com/learningequality/kolibri/blob/1658e8407839cc9f20bfd14ecfa10b2996b2847d/kolibri/core/assets/src/views/sync/FacilityAdminCredentialsForm.vue#L53-L60

Validators can also be more sophisticated and verify the actual values, like this mastery model object validator:

https://github.com/learningequality/kolibri/blob/1658e8407839cc9f20bfd14ecfa10b2996b2847d/kolibri/plugins/coach/assets/src/views/common/MasteryModel.vue#L12-L38

Note of this is required but at least checking the attributes is certainly a nice-to-have.

rtibbles (approved):

MisRob:
@marcellamaki Thank you for your help

PR #8506, opened 2021-10-20
Fix bug with import resource when there are duplicates.

### LINKED ISSUE #6351: "resources to be updated" diff count does not match reported import behavior and final count
### Observed behavior

observed by @radinamatic in https://github.com/learningequality/kolibri/issues/6280

* 23 resources initially imported
* diff reports 6 resources to be updated and 17 deleted
* task manager reports 10 resources updated
* final count reported on device is 12

|initial import| diff page | task page | on device summary |
|--|--|--|--|
| image| image |  image |image|

### Expected behavior

counts should add up

### User-facing consequences

confusion about actual behavior

### Steps to reproduce

update a channel with some content already imported

### Context

* Kolibri version: 0.13.0b5
* Operating system: Ubuntu 16.04
* Browser: Firefox & Chrome

### LINKED ISSUE #6509: incorrect counting of resources when importing duplicates
### Observed behavior

reported by @jamalex 

> This happens when some of the resources selected already existed from being imported via another place in the tree. Have we considered displaying this differently? My first assumption on seeing it was that some of the files failed to import.
> 
> image
> 
> (and the 22 of 46 / 38MB of 38MB feels a bit inconsistent too)

### Expected behavior

The intended behavior here is that it would show 46 of 46 because we are now treating duplicates as distinct resources from the end-user’s perspective.

However, the total file size should represent only the newly transferred stuff. For example if you imported 5 resources, all dupes, it should say “5 of 5 resources (0 MB of 0 MB) imported”

### User-facing consequences

confusing/inconsistent counting

### Steps to reproduce

* Create or find a channel that has some resources duplicated across multiple topics
* Import one copy
* Import another copy

### Context

0.13.0

AtKristijan:
## Summary

Remove distinct calculation for duplicate resource. 
…

## References
Fixes #6509
Fixes #6351 
…

## Reviewer guidance

…

----

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated


kolibri/core/content/utils/import_export_content.py
```diff
@@ -170,17 +170,16 @@ def get_import_export_data(  # noqa: C901
...
 
         # Only bother with this query if there were any resources returned above.
         if included_content_ids:
             content_ids.update(included_content_ids)
             file_objects = LocalFile.objects.filter(
                 files__contentnode__in=nodes_segment
-            ).values("id", "file_size", "extension")
-            if available is not None:
+            )
+            if available:
```
rtibbles: This change needs to be reverted, as this will exclude the case when `available` is `False`.

kolibri/core/content/utils/import_export_content.py
```diff
@@ -139,7 +139,7 @@ def get_import_export_data(  # noqa: C901
 
     queried_file_objects = {}
 
-    content_ids = set()
+    content_ids = list()
```
rtibbles: The only purpose of `content_ids` here is to deduplicate the number of resources - so we can remove most of the logic related to `content_ids` and just keep a count of total number of resources instead.

rtibbles (changes requested):
Some changes needed.

kolibri/core/content/utils/upgrade.py
```diff
@@ -566,7 +566,7 @@ def get_import_data_for_update(
     i = 0
 
     updated_ids_slice = updated_resource_ids[i : i + batch_size]
-
+    print("updated_ids_slice", updated_ids_slice)
```
rtibbles: Should remove the print statement here.

kolibri/core/content/utils/import_export_content.py
rtibbles: This should still be addressed - instead of keeping a list of `content_ids` we can just start a counter:
```
number_of_resources = 0
```
and increment it by the number of resources within each iteration.

kolibri/core/content/utils/import_export_content.py
```diff
@@ -207,7 +207,6 @@ def get_import_export_data(  # noqa: C901
     files_to_download = list(queried_file_objects.values())
 
     total_bytes_to_transfer = sum(map(lambda x: x["file_size"] or 0, files_to_download))
-
     return len(content_ids), files_to_download, total_bytes_to_transfer
```
rtibbles: Would then return `number_of_resources` here instead of `len(content_ids)`

rtibbles (changes requested):
Some clean up still to do here, plus linting.

rtibbles (approved):

PR #8513, opened 2021-10-21
Handle mismatching datasets for exams and lessons to prevent syncing issues

### LINKED ISSUE #7453: Don't set user as created_by for lesson when superadmin from Facility B creates a lesson in Facility A
### Observed behavior

This is an edge case scenario for a multi-facility setup i.e. `Facility A` and `Facility B` available on the same device. In the current implementation of 0.14 it is possible that a `Super admin B` of the device and associated with `Facility B` creates a lesson `Lesson AB` for a Class of `Facility A`.

If this happens then upon syncing `Facility A` to another device, it does not sync the Lesson created by the `Super admin B`. 

### Expected behavior

1. Do not allow even a super admin not associated with a facility to modify any lessons in the unassociated facility.
**or**
2. Allow sync of lessons if created by a device super admin irrespective of associated facility. This will be consistent with the existing behaviour of syncing users and other artefacts created by such a super admin. *Am not aware of any demerits of allowing the same. Rather it shall be convenient for a device super user to manage.*

### User-facing consequences

Confusion about missing data.

### Errors and logs

NA

### Steps to reproduce

1. Create two facilities - `Facility A` and `Facility B` on a device with superusers `Super Admin A` and `Super Admin B` respectively.
2. Login as `Super Admin B` and create a class and `Lesson AB` in `Facility A`
3. Sync `Facility A` to another device.
4. Note that `Lesson AB` will not sync to the destination device.

### Context

Kolibri 0.14.0

bjester:
## Background
Lesson and exam models have foreign references to the user model, and if those models are created by a user who was in a different facility (like a superuser), then the models would fail to deserialize when synced

## Summary

- Makes `creator`, `created_by`, and `assigned_by` nullable
- Adds enforcement to still require users for the above fields, but allows it to be null for the cross-facility superuser edge case
- Adds additional enforcement to ensure assignment datasets match datasets of their referenced exams/lessons
- Adds migrations to fix un-syncable lessons and exams caused by the issue #7453 

## References

Resolves: https://github.com/learningequality/kolibri/issues/7453
Builds off: https://github.com/learningequality/kolibri/pull/7887
Ref: https://github.com/learningequality/kolibri/issues/7859

## Reviewer guidance

…

----

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated


kolibri/core/exams/upgrade.py
```diff
@@ -0,0 +1,43 @@
...
+    reside in, which caused a mismatch in the exam dataset and prevented syncing
+    """
+    # un-set creator for all exams created by user in a different dataset
+    exam_sub_query = Subquery(
+        Collection.objects.filter(pk=OuterRef("collection_id")).values("dataset_id")[
+            :1
+        ],
+        output_field=UUIDField(),
+    )
+    Exam.objects.exclude(collection__dataset_id=F("dataset_id")).update(
```
rtibbles, line 32: This took me a couple of reads to get my head around. I think I understand what is going on though.

The dataset_id of the Exam has previously been set from the `creator`, and so if there is a mismatch between the dataset_id of the Exam and the dataset_id of the collection it belongs to, then we have discovered our problem exams.

kolibri/core/exams/upgrade.py
```diff
@@ -0,0 +1,43 @@
...
+        creator=None, dataset_id=exam_sub_query
+    )
+
+    # un-set assigned_by for all exam assignments assigned by a user in a different dataset
+    assignment_sub_query = Subquery(
```
rtibbles, line 37: Presumably this is now relying on the migration above having happened?

rtibbles (approved):
I had to read a bit harder to parse what was happening in the upgrade steps - but nothing blocking. Code makes sense and tests cover it!

kolibri/core/exams/upgrade.py
bjester: Yeah the dataset inference was changed in https://github.com/learningequality/kolibri/pull/7887 so this migrates that change, like you said where the collection dataset doesn't match the exam's. This wasn't as simple as I wanted, and as simple as the query could have been, but Django constrained me a little bit.

kolibri/core/exams/upgrade.py
bjester: Yeah that's a good note, it does depend on that to achieve the end result we want

PR #8502, opened 2021-10-15
Add utility to migrate from ExamLog and ExamAttemptLog to consolidated logging.

rtibbles:
## Summary
* Implements a utility to map ExamLog and ExamAttemptLog to our more general Content logging models
* Removes restrictive validation on MasteryLog mastery_level to allow mastery levels that will prevent syncing to historic versions of Kolibri
* Allows channel_id to be nullable on ContentSession and ContentSummary log models
* Adds tests for the function

## References
Fixes #8477

## Reviewer guidance
Can you think of any edge cases not covered by the tests?

----

## Testing checklist

- [x] Contributor has fully tested the PR manually
- [ ] If there are any front-end changes, before/after screenshots are included
- [ ] Critical user journeys are covered by Gherkin stories
- [x] Critical and brittle code paths are covered by unit tests

## PR process

- [x] PR has the correct target branch and milestone
- [x] PR has 'needs review' or 'work-in-progress' label
- [x] If PR is ready for review, a reviewer has been added. (Don't use 'Assignees')
- [ ] If this is an important user-facing change, PR or related issue has a 'changelog' label
- [ ] If this includes an internal dependency change, a link to the diff is provided

## Reviewer checklist

- Automated test coverage is satisfactory
- PR is fully functional
- PR has been tested for [accessibility regressions](http://kolibri-dev.readthedocs.io/en/develop/manual_testing.html#accessibility-a11y-testing)
- External dependency files were updated if necessary (`yarn` and `pip`)
- Documentation is updated
- Contributor is in AUTHORS.md

PR #8525, opened 2021-10-22
Integrated progress tracking API

### LINKED ISSUE #8257: Timer that tracks elapsed time for content does not actually pause when a user is idle
## Observed behavior
The elapsed time timer does not pause when a user is idle, thus causing discrepancies in the logged time elapsed for content if someone is idle but does not get logged out.

## Expected behavior
If a user is flagged as idle, the timer should no longer be accumulating time spent.

## User-facing consequences
Unexpectedly long time spent from backgrounded tabs with content open in them.

## Steps to reproduce
Open content in a tab
Open another tab and use kolibri actively for X minutes
Observe that the time spent for the content in the backgrounded tab gets saved as X minutes when you return to the backgrounded tab.

## Context
 * Kolibri version: 0.4.x and onwards

### LINKED ISSUE #8480: Create an upgrade step when upgrading from versions older than 0.15 to run the utility function to convert ExamLog and ExamAttemptLog to MasteryLog and AttemptLog

### LINKED ISSUE #8483: Update Kolibri to use MasteryLog and AttemptLog for reading quiz progress in coach reports

### LINKED ISSUE #8484: Update the learner quiz engagement to log attempts and progress via MasteryLog and AttemptLog

### LINKED ISSUE #8485: Update stat pingbacks to use AttemptLogs and MasteryLogs filtered by 'quiz' or not
https://github.com/learningequality/kolibri/blob/b87f9da274a6a9218419cf9f858be695aa04304a/kolibri/core/analytics/utils.py#L282

These stats pingbacks need to be filtered to either exclude quizzes logged as content, or include only quizzes logged as content, depending on whether they are currently using 'ExamLog' and 'ExamAttemptLog'.

### LINKED ISSUE #8486: Update data exports for content session and summary logs to exclude the 'quiz' kind

rtibbles:
## Summary
* Creates an integrated progress tracking API
* Migrates content resource progress tracking to use this API
* Migrates quiz progress tracking to use this API
* Updates Kolibri to use MasteryLog and AttemptLog for reading quiz progress in coach reports
* Updates stat pingbacks to use AttemptLogs and MasteryLogs filtered by 'quiz' or not
* Creates an upgrade step when upgrading from versions older than 0.15 to run the utility function to convert ExamLog and ExamAttemptLog to MasteryLog and AttemptLog
* Updates data exports for content session and summary logs to exclude the 'quiz' kind

## References
Depends on #8502

Fixes #8484
Fixes #8486
Fixes #8480
Fixes #8483
Fixes #8485
Does the first stage to fix 8482 - but is waiting on #8527 to do the final connection to avoid merge conflicts and to ensure the order of operations!
Fixes #8257 

## Reviewer guidance
For manual QA:
Run through gherkin scenarios for:
* content resource engagement
* lesson resource engagement
* quiz engagement
* viewing learner quiz reports
* viewing coach reports for all of the above activity

For code review:
The API is a bit of a beast - what else can I do to make it easy to read and inspect?
The frontend code is mostly simplifying - but it is possible I've missed something - not changed the value of something in the frontend from `closed` to `complete` is the most likely, as that has been a switchover in some places for ExamLog to MasteryLog.

----

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated


kolibri/core/assets/src/state/modules/logging.js
```diff
@@ -1,121 +1,82 @@
...
+    ADD_OR_UPDATE_ATTEMPT(state, attempt) {
+      if (attempt.id) {
+        if (!state.pastattemptMap[attempt.id]) {
+          state.pastattempts.unshift(attempt);
+          Vue.set(state.pastattemptMap, attempt.id, attempt);
+          state.totalattempts += 1;
+        } else {
+          Object.assign(state.pastattemptMap[attempt.id], attempt);
+        }
+      }
```
indirectlylit: what happens if if `attempt.id` is not defined? Is this an error condition that should be logged?

kolibri/core/analytics/utils.py
```diff
@@ -281,11 +281,11 @@ def extract_facility_statistics(facility):
         # exam_count
         "ec": Exam.objects.filter(dataset_id=dataset_id).count(),
         # exam_log_count
-        "elc": ExamLog.objects.filter(dataset_id=dataset_id).count(),
+        "elc": MasteryLog.objects.filter(dataset_id=dataset_id, summarylog__kind=content_kinds.QUIZ).count(),
```
indirectlylit, line 284: are 'mastery logs' used for both exercises and quizzes?

kolibri/core/logger/api.py
```diff
@@ -1,192 +1,774 @@
...
+        item_content_id, item = log["item"].split(QUIZ_ITEM_DELIMETER)
+        log["item"] = item
+    except ValueError:
+        item_content_id = ""
+    log["content_id"] = item_content_id
+    return log
+
+
+class ProgressTrackingViewSet(viewsets.GenericViewSet):
+    def _validate_quiz(self, user, quiz_id):
```
indirectlylit: seems this function does not actually validate the quiz

```suggestion
    def _check_permission(self, user, quiz_id):
```

kolibri/core/logger/api.py
```diff
@@ -1,192 +1,774 @@
...
+        if not Exam.objects.filter(
+            active=True,
+            assignments__collection_id__in=user.memberships.all().values(
+                "collection_id"
+            ),
+            id=quiz_id,
+        ).exists():
+            raise PermissionDenied("User does not have access to this quiz_id")
+
+    def _validate_context(self, user, validated_data):
```
indirectlylit: since this returns values, it does more than just validation

```suggestion
    def _extract_context(self, user, validated_data):
```

kolibri/core/logger/api.py
```diff
@@ -1,192 +1,774 @@
...
+        node_id = validated_data.get("node_id")
+        quiz_id = validated_data.get("quiz_id")
+        lesson_id = validated_data.get("lesson_id")
+
+        context = {}
```
indirectlylit: maybe this could be a simple `LogContext` class so the possible members are well-defined and documented in one place?

kolibri/core/logger/api.py
```diff
@@ -1,192 +1,774 @@
...
+        progress - any previous progress on this content resource
+        time_spent - any previous time spent on this content resource
+        extra_fields - any previously recorded additional data stored for this resource
+
+        if this is an assessment, will also include
+        mastery_criterion - the mastery criterion that should be applied to determine completion
+        pastattempts - a serialized subset of previous responses within this run, that can be used
+                       to determine completion
+        totalattempts - the total number of previous responses within this run of the assessment resource.
+        """
```
indirectlylit: possible readability updates


```suggestion
        """
        Make a POST request to start a content session.

        Requires one of either:
        - node_id: the pk of the resource
        - quiz_id: the pk of the quiz (Exam) object

        Optional parameters:
        - repeat: whether to reset previous progress on this content to zero and start fresh
        - lesson_id: if this is being engaged within a lesson

        Returns object with properties:
        - session_id: id of the session object that was created by this call
        - context: contains node_id, quiz_id, lesson_id, and mastery_level as appropriate

        If user is logged in, return object will also include:
        - progress: any previous progress on this content resource
        - time_spent: any previous time spent on this content resource
        - extra_fields: any previously recorded additional data stored for this resource

        If this is an assessment, return object will also include:
        - mastery_criterion: mastery criterion that should be applied to determine completion
        - pastattempts: serialized subset of recent responses, used to determine completion
        - totalattempts: total number of previous responses within this run of the assessment resource
        """
```

kolibri/core/logger/api.py
```diff
@@ -1,192 +1,774 @@
...
+            if mastery_model.get("coach_assigned"):
+                # To prevent coach assigned quiz mastery logs from propagating to older
+                # Kolibri versions, we use negative mastery levels for these.
+                mastery_level = (
+                    masterylog.mastery_level - 1 if masterylog is not None else -1
+                )
+            else:
+                mastery_level = (
+                    masterylog.mastery_level + 1 if masterylog is not None else 1
+                )
```
indirectlylit: Without more context on what a "mastery level" is and what kinds of values it can contain, the intent of this code is a bit confusing.

It's being incremented and decremented based on whether or not something is coach-assigned? This is also complicated by the use of `-1` as a special flag because it looks like it's possible for `mastery_level` to be decremented below zero?

If `mastery_level` can only take on a few values (e.g. `-1`, `0`, or `1`) I'd recommend defining these as named constants and just setting them directly rather than doing arithmetic on them. If the valid range is `>= 0` and `-1` is the only special case, then maybe just a single named constant for `-1`?

kolibri/core/logger/api.py
```diff
@@ -1,192 +1,774 @@
...
+                summarylog,
+                context["node_id"],
+                context["lesson_id"],
+            )
+
+    def _process_masterylog_created_notification(self, masterylog, context):
+        if "quiz_id" in context:
+            wrap_to_save_queue(
+                quiz_started_notification, masterylog, context["quiz_id"]
+            )
```
indirectlylit: Non-blocking comment, similar feedback I gave recently on another PR:

I think the [return early strategy](https://szymonkrajewski.pl/why-should-you-return-early/) can really help readability, especially in cases like these "process * notification" functions. For example, it can help break up complex single conditionals into multiple simpler conditionals.

Copying the illustrations from that blog post:

image


image


```suggestion
    def _process_created_notification(self, summarylog, context):
        if "lesson_id" not in context:
            return
        if "node_id" not in context:
            return
        if summarylog.kind == content_kinds.EXERCISE:
            return

        wrap_to_save_queue(
            start_lesson_resource,
            summarylog,
            context["node_id"],
            context["lesson_id"],
        )

    def _process_masterylog_created_notification(self, masterylog, context):
        if "quiz_id" not in context:
            return

        wrap_to_save_queue(quiz_started_notification, masterylog, context["quiz_id"])
```

kolibri/core/logger/api.py
```diff
@@ -1,192 +1,774 @@
...
+                attemptlog.correct = correct
+                update_fields += ("correct",)
+
+                if answer:
+                    attemptlog.answer = answer
+                    update_fields += ("answer",)
+
+                if simple_answer:
+                    attemptlog.simple_answer = simple_answer
+                    update_fields += ("simple_answer",)
```
indirectlylit: Maybe a small helper class for processing attempts and update fields could help create a simpler abstraction and hide the complexity and hard-coded strings?

I also see some of these strings might be defined in `constants/interaction_types.py`?

kolibri/core/logger/api.py
```diff
@@ -1,192 +1,774 @@
...
+        attempt - a nested object, if creating an attempt, must include item
+                  if updating an existing attempt, must include attempt_id
+
+        returns an object with the following properties
+        complete - boolean indicating if the resource is completed
+
+        if an attempt at an assessment was included, then the following parameter will be included
+        attempt - a serialized form of the attempt, equivalent to that returned in pastattempts from
+                  session initialization.
+        """
```
indirectlylit:
```suggestion
        """
        Make a PUT request to update the current session

        Requires one of either:
        - progress_delta: increase the progress by this amount
        - progress: set the progress to this amount

        Can also update time spent recorded with a delta:
        - time_spent_delta: number of seconds to increase time_spent by

        And update the extra_fields value stored:
        - extra_fields: the complete representation to set extra_fields to

        If creating or updating an attempt for an assessment must include:
        - attempt: a nested object, if creating an attempt, must include item
                  if updating an existing attempt, must include attempt_id

        Returns an object with the properties:
        - complete: boolean indicating if the resource is completed

        If an attempt at an assessment was included, then this parameter will be included:
        - attempt: serialized form of the attempt, equivalent to that returned in pastattempts from
                  session initialization
        """
```

kolibri/core/logger/utils/exam_log_migration.py
```diff
@@ -0,0 +1,275 @@
...
+    This serves two purposes:
+    1. It allows user progress data about quizzes to be synced via single user syncing
+    mechanisms. This was not possible for the ExamLog and ExamAttemptLog, due to the
+    choice made about the partition used for that data, when the original intent for the
+    'Exam' was made to be more summative and less formative, with a focus on timed classroom
+    testing.
+    2. It allows quizzes to be treated as a content type, which means that a quiz like experience
+    can be exposed to a learner without coach intervention, and the same user progress tracking
+    is used irrespective of whether it is a coach assigned quiz, or one discovered by the learner
+    independently.
```
indirectlylit: would be excellent to have some gherkin stories to exercise these migration paths

indirectlylit (commented):
exciting stuff! adding some quick feedback from an initial skim below

pcenov:
Hi @rtibbles, I was able to find the following issues while testing the [Windows and Linux builds](https://buildkite.com/learningequality/kolibri-python-package/builds/3915):

1. Incorrect points value after completing a single lesson on a learner-only device: 

2. 400 error for /api/logger/trackprogress while going through an exercise: 

3. 400 error for /api/logger/masterylog after completing an exercise: 

4. 500 error for /api/logger/trackprogress when viewing a resource as a guest user: 

Attached logs:
[WindowsLogs.zip](https://github.com/learningequality/kolibri/files/7398478/WindowsLogs.zip)
[UbuntuLogs.zip](https://github.com/learningequality/kolibri/files/7398549/UbuntuLogs.zip)

kolibri/core/assets/src/state/modules/logging.js
rtibbles: No - but I can add a clarifying comment here. It gets added in after the attempt has been created and the id assigned (when this mutation is called again from the action).

kolibri/core/analytics/utils.py
rtibbles: Now they are, yes.

kolibri/core/logger/api.py
rtibbles: Fair.

kolibri/core/logger/api.py
rtibbles: True.

kolibri/core/logger/api.py
rtibbles: Yeah, can set some slots to document the attributes.

kolibri/core/logger/api.py
rtibbles: Yeah - I am never quite sure how best to document functions that handle body or query params - they're kind of arguments, but not a traditional doc string.

kolibri/core/logger/api.py
rtibbles: I was attempting to explain this here:
```
                # To prevent coach assigned quiz mastery logs from propagating to older
                # Kolibri versions, we use negative mastery levels for these.
```

The MasteryLevel can be any integer - for exercises it starts at 1 and increments from there (although we've not actually implemented the 'practice again' behaviour in the frontend).

This may need to be iterated on further also, as one key element of the `mastery_level` is that it helps to determine the unique id for the MasteryLog.

In the case of coach assigned quizzes we may be better off assigning a random negative integer for the mastery level so that independent attempts at the same quiz started on different devices do not collide.

kolibri/core/logger/api.py
rtibbles: I've referenced the constants where relevant - the pure strings here are almost all model field names on the attemptlog.

And yes, I was struggling with how to handle the combination of the nesting and not nesting here.

I think at the very least breaking apart the create and update into separate functions would help.

rtibbles:
Thanks @pcenov - I'll track these down.

kolibri/core/logger/api.py
rtibbles: I think in this case, because of the single indent for the conditional, and because there's not any intervening processing between each of the conditions, it feels less of a readability concern.

rtibbles:
@pcenov one thing I am confused by is

> 400 error for /api/logger/masterylog after completing an exercise

I have removed all calls to this endpoint (and to the attemptlog endpoint I also see in the network tab history) is there any possibility this error arose on an asset not from this PR?

kolibri/core/logger/api.py
indirectlylit: yeah it's not a particular concern – just a general style I like and using this as an example

pcenov:
Hi @rtibbles, yes the '400 error for /api/logger/masterylog after completing an exercise' was most probably observed on an asset not from this PR as I'm not able to replicate it today using the latest build asset.

pcenov:
Hi @rtibbles, happy to report that I didn't see any api/logger errors in the latest Buildkite build. At the same time:
1. The issue with the incorrectly assigned points is still valid - for example after completing a lesson with a single video resource I get assigned 1500 points and after refreshing the page it goes back to the correct 500 points.
2. I'm not able to get a single lesson with exercises marked as completed in the Reports or the learner's home page:


3. In the Reports for a lesson there is no value in the 'Average time spent' column (could not be caused by changes made in this PR if so, I'll report it separately):


Logs: [logs.zip](https://github.com/learningequality/kolibri/files/7418598/logs.zip)
Database: https://drive.google.com/file/d/1dMy0s3nI4Nac6KBp5bn4uxxtH3Mx_KZ1/view?usp=sharing

kolibri/plugins/learn/assets/src/views/ExamPage/index.vue
```diff
@@ -193,17 +192,14 @@
     data() {
       return {
         submitModalOpen: false,
+        startTime: Date.now(),
```
indirectlylit, line 197: This is just used to compute client-side time deltas in `time_spent`, correct? (Probably don't want to pass it to the server as a timestamp in case of clock discrepancies.)

kolibri/plugins/learn/assets/src/views/AssessmentWrapper/index.vue
```diff
@@ -393,58 +388,34 @@ oriented data synchronization.
...
       hintTaken({ answerState }) {
-        this.updateAttemptLogInteractionHistory({
-          type: InteractionTypes.hint,
-          answer: answerState,
-        });
+        this.hintWasTaken = true;
+        this.updateAttempt({ answerState });
+      },
+      updateAttempt({ answerState, simpleAnswer } = {}) {
+        const response = {
```
indirectlylit: should this be `request`?

kolibri/plugins/learn/assets/src/views/AssessmentWrapper/index.vue
```diff
@@ -301,74 +343,27 @@ oriented data synchronization.
         }
         return null;
       },
-    },
-    watch: {
-      exerciseProgress() {
-        this.updateExerciseProgressMethod();
+      renderer() {
+        // TODO: rtibbles - update the KContentRenderer API to expose hint info
+        // and add takeHint public method
```
indirectlylit, line 324: Should this be an issue in KDS?

(If so, let's do this at the same time - https://github.com/learningequality/kolibri-design-system/issues/234)

kolibri/plugins/coach/assets/src/views/reports/ReportsQuizQuestionPage.vue
```diff
@@ -5,7 +5,7 @@
     :immersivePagePrimary="true"
     immersivePageIcon="back"
     :immersivePageRoute="toolbarRoute"
-    :appBarTitle="exam.title"
+    :appBarTitle="exam && exam.title"
```
indirectlylit: minor passing thought:

would be cool if we could use [optional chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining)

This might be more explicit than the boolean expression?

```suggestion
    :appBarTitle="exam ? exam.title : null"
```

kolibri/core/logger/utils/user_data.py
```diff
@@ -540,17 +540,44 @@ def create_exams_for_classrooms(**options):
...
         # everyone in the class has to take the exam
         for user in classroom.get_members():
-            # create exam log per user
-            examlog = ExamLog.objects.create(
-                exam=exam, user=user, completion_timestamp=now
+            random_seconds = random.randint(1, 500)
+            seconds = timezone.timedelta(seconds=random_seconds)
+            then = now - seconds
+            # create mastery log per user
+            sl = ContentSessionLog.objects.create(
```
indirectlylit: would help to distinguish from `summarylog`

```suggestion
            sessionlog = ContentSessionLog.objects.create(
```

kolibri/core/logger/api.py
```diff
@@ -1,192 +1,841 @@
...
+            )
+            .order_by("-complete", "-end_timestamp")
+            .first()
         )
 
-    def filter_learner_group(self, queryset, name, value):
-        return queryset.filter(user__memberships__collection_id=value)
+        if masterylog is None or (masterylog.complete and repeat):
+            # There is no previous masterylog, or there is no incomplete
+            # previous masterylog, and the request is requesting a new attempt.
```
indirectlylit: confusing phrasing

```suggestion
            # There is no previous masterylog, or the previous masterylog
            # is complete, and the request is requesting a new attempt.
```

kolibri/core/logger/api.py
```diff
@@ -1,192 +1,841 @@
...
+            # unified under a single retry.
+            if mastery_model.get("coach_assigned"):
+                # To prevent coach assigned quiz mastery logs from propagating to older
+                # Kolibri versions, we use negative mastery levels for these.
+                # In older versions of Kolibri the mastery_level is validated to be
+                # between 1 and 10 - so these values will fail validation and hence will
+                # not be deserialized from the morango store.
+                # TODO: rtibbles - determine if we should instead randomize these values
+                # to ensure that coach assigned quiz attempts on different devices are not
+                # mixed together.
```
indirectlylit: I don't really understand this. Shouldn't `mastery_level` be equal to `len(MasteryLog.objects.filter(summarylog, user))` and if so why do we store it separately?

Also, don't `MasteryLog` objects have random UUIDs that can serve to disambiguate?

kolibri/core/logger/api.py
```diff
@@ -1,192 +1,841 @@
...
+            # Having the same mastery_level/summarylog (and hence user) pair will result in the same
+            # identifier being created. So if the same user engages with the same assessment on different
+            # devices, when the data synchronizes, if the mastery_level is the same, this data will be
+            # unified under a single retry.
+            if mastery_model.get("coach_assigned"):
+                # To prevent coach assigned quiz mastery logs from propagating to older
+                # Kolibri versions, we use negative mastery levels for these.
+                # In older versions of Kolibri the mastery_level is validated to be
+                # between 1 and 10 - so these values will fail validation and hence will
+                # not be deserialized from the morango store.
```
indirectlylit, line 455: Seems we might need a more systematic pattern for preventing newer schemas from syncing to older devices and specifying compatibility as the schema evolves. Maybe introducing `schema_version_major`/`schema_version_minor` fields to all syncable models at the Morango level?

If we do that in 0.15 we'll have the tool available for the next time something like this happens.

indirectlylit (commented):
gave the code another look - looking good

kolibri/plugins/learn/assets/src/views/ExamPage/index.vue
rtibbles: Yes, this is just used to generate time spent deltas - nothing else. I've also deferred almost all time stamping to the backend in this PR (the only thing that isn't deferred is the generation of attempt `end_timestamp`).

kolibri/plugins/learn/assets/src/views/AssessmentWrapper/index.vue
rtibbles: No - it's a 'response to an attempt', I had not considered that this might be confusing.

There's a weird hierarchy here of 'sequence of attempts on an exercise -> attempt (mapping to attemptlog) -> sequence of responses/interactions in an attempt'.

Maybe I should call it an 'interaction' instead?

kolibri/plugins/learn/assets/src/views/AssessmentWrapper/index.vue
rtibbles: Yes, yes it should..

kolibri/core/logger/utils/user_data.py
rtibbles: I did this so as not to rename a previous `sessionlog` above.

kolibri/core/logger/api.py
rtibbles: > Also, don't MasteryLog objects have random UUIDs that can serve to disambiguate?

No - their UUID is deterministic based on the summarylog_id and the mastery_level.

kolibri/core/logger/api.py
rtibbles: I'm not sure what the best way to handle that would be in morango - checking `min_kolibri_version` during validation might be the simplest.

I don't want to open that rabbit hole at this point as I think doing this systematically would take us well beyond the already fairly heavy refactor. But maybe we could look at this for Morango 0.7?

kolibri/plugins/learn/assets/src/views/AssessmentWrapper/index.vue
indirectlylit: that would avoid confusing with HTTP concepts yes

indirectlylit: if it's a widely-used term might as well make it even more explicit e.g. `attemptInteraction`

kolibri/core/logger/api.py
indirectlylit: Yeah, definitely not advocating adding it to this PR - just doing something before 0.15 release.

We could initialize all Morango models to 1.0. Then in the future we can bump minor for non-breaking and major for breaking changes which would be more granular than blanket Kolibri version restrictions.

kolibri/plugins/learn/assets/src/views/AssessmentWrapper/index.vue
rtibbles: https://github.com/learningequality/kolibri-design-system/issues/271

kolibri/core/logger/utils/user_data.py
rtibbles: I tell a lie - this was a hot fresh bowl of copy pasta.

kolibri/core/logger/api.py
rtibbles: I would be concerned about doing this within the scope of 0.15 as a whole, to be honest.

rtibbles:
@pcenov I should now have fixed the remaining three issues you identified.

@indirectlylit I have addressed all of your review comments to the best of my ability!

radinamatic:
Confirming all issues previously reported by @pcenov as resolved on the most recent DEB asset installed on Ubuntu 18.04: 
* no errors in the Network tab
* resource, lesson, and quiz engagement by learners  ✔️ 
* lessons and quizzes completion reported correctly to both learners and coaches  ✔️ 
* quiz report viewing by both learners and coaches  ✔️ 

Finicky behavior of the **Completed** icon on video/audio resource viewer: **In progress** icon disappears when the resource is completed, but the **Completed** icon does not appear unless the page is reloaded. 

completed-icon2

Did you encounter ⬆️ in your testing, @pcenov?

I'll report separately if not related to this PR.

pcenov:
I also confirm that the issues reported so far are resolved.
@radinamatic yes, I also noticed the issue with the disappearing progress icon while testing the Windows build plus I was able to spot another similar one where the icon for a correctly answered question is not displayed and in fact if the user decides to proceed by clicking the Next button the question doesn't get changed to a new one. Refreshing the page seems to fix that issue as well.
Screenshots:



Logs and DB: https://drive.google.com/file/d/1ZThxD24i7ji2N-0ocvpOQLvDYwnplvWA/view?usp=sharing

rtibbles:
Thanks for the careful testing here. I had seen the 'not moving onto the next question' issue at one point during testing, but then it had failed to replicate. Will push on this more, as I suspect I know where the issue is arising.

Thank you for the screenshots of the console @pcenov. I am going to write some automated tests for this logic to put this on a surer footing as well.

kolibri/core/device/middleware.py
```diff
@@ -99,3 +101,38 @@ def process_exception(self, request, exception):
 
     def __call__(self, request):
         return self.get_response(request)
+
+
+class DatabaseBusyErrorHandler(object):
+    """
+    A middleware class to raise a 503 when the database is under heavy load
+    For SQLite this will trigger for database locked errors.
+    For Postgres this will trigger for
```
jredrejo: for... what?

kolibri/core/exams/permissions.py
```diff
@@ -30,23 +38,31 @@ def user_can_read_object(self, user, obj):
...
+        from kolibri.core.logger.models import MasteryLog
 
         # If they are not a member of the assignment's collection, don't bother with any other checks
         return obj.assignments.objects.filter(
             collection_id__in=user.memberships.all().values("collection_id")
         ).exists() and (
-            obj.active or ExamLog.objects.filter(exam=obj, user=user).exists()
+            obj.active
+            or MasteryLog.objects.filter(
+                summarylog__content_id=obj.id, user=user
```
jredrejo, line 49: I'm concerned for the performance of queries like this in DB with a long story, because the logs will be huge. I have checked summarylog does not have an index for content_id. I think it would be good to add it. (It has it for the user though)

kolibri/core/exams/permissions.py
```diff
@@ -30,23 +38,31 @@ def user_can_read_object(self, user, obj):
...
+        user_masterylog_content_ids = MasteryLog.objects.filter(user=user).values(
+            "summarylog__content_id"
+        )
 
         assignments = ExamAssignment.objects.filter(
             collection_id__in=user.memberships.all().values("collection_id")
         )
         return Q(assignments__in=assignments) & Q(
-            Q(active=True) | Q(examlogs__user=user)
+            Q(active=True) | Q(id__in=user_masterylog_content_ids)
```
jredrejo, line 67: I'd also recommend an index on `active` in the `ExamAssignment` model as it's used in many places for filtering

kolibri/core/logger/api.py
```diff
@@ -1,192 +1,855 @@
...
+    "error",
+    "item",
+    "answer",
+    "time_spent",
+]
+
+
+class LogContext(object):
+
+    __slots__ = "node_id", "quiz_id", "lesson_id", "mastery_level"
```
jredrejo, line 151: I love the use of slots for the class, but I'd also love some doc on the purpose of this LogContext object as it's hard to track how this is used to pass context to the serializers

jredrejo (commented):
I have a few minor doubts, but it's a huge PR so my eyes could have missed something. I have tested it and haven't found any problem after the latest fixes.

Appart from the code comments, I wonder if we could get totally rid of the `ExamLogs` model. I see it's still used for some morango magic in `batch_process_examlogs` in the notifications api but I'm not sure if this is still needed or something this PR should change too.

kolibri/core/device/middleware.py
rtibbles: I left this as an exercise for the reader, obviously! It's currently triggering for deadlocks, will update.

kolibri/core/exams/permissions.py
rtibbles: I think would be good to follow up and do some optimization checks on here, but I don't want it to get any bigger.

I'll open a follow up issue for this.

kolibri/core/exams/permissions.py
rtibbles: `active` is on the Exam model here, I think - I'll add it to the issue I'll file for this.

I'll do a general look through and try to log all the possible places to check for slowdowns as a result of this work!

kolibri/core/logger/api.py
rtibbles: Can do - I added this class in the first place in response to @indirectlylit's suggestion (previously it was just using a dict) will document more!

rtibbles:
@jredrejo I filed a follow up issue to look at optimizations here: https://github.com/learningequality/kolibri/issues/8549

We can't remove the ExamLog and ExamAttemptLog models yet (and probably can't for a while) as we still need to be able to upgrade and sync in these models from older Kolibris.

rtibbles:
@pcenov @radinamatic I believe I have now fixed the two issues you discovered, I was able to reliably reproduce the completion mark disappearing, and that is now definitely fixed. The moving to the next item should also be fixed (I have not been able to reproduce since adding my fix).

I also added one more fix, where I was seeing videos not getting marked as completed even though they had been played all the way through (this was due to some floating point arithmetic issues, where the total progress was adding up to 0.996 instead of 1). I have now added some rounding logic to ensure that we always hit 1 - but it may cause mildly premature completion, but as it's rounding to thousandths, I don't see that it will be practically perceptible by a user.

pcenov:
@rtibbles I've verified the resolved issues as fixed and didn't identify any new issues in my regression testing. Looking good!

kolibri/core/logger/api.py
jredrejo: > @jredrejo I filed a follow up issue to look at optimizations here: #8549
> 
> We can't remove the ExamLog and ExamAttemptLog models yet (and probably can't for a while) as we still need to be able to upgrade and sync in these models from older Kolibris.

Ok, and the use of `ExamLog` in `batch_process_examlogs` (in kolibri/core/notifications/api.py ) is still needed or you forgot to change it?

indirectlylit (approved):
> @indirectlylit I have addressed all of your review comments to the best of my ability!

thank you!

I haven't tested but once you're comfortable with the rest of the feedback it's good from my end

kolibri/core/logger/api.py
rtibbles: We can probably delete those, but I wanted to integrate with @bjester's sync operations work before making any changes to how we handle the notification generation.

PR #8531, opened 2021-10-25
Build Parallelization Mods + BCK Build Requirements

DXCanas:
## Summary

Hello! Guest PR from infra.

This should be a fairly simple modification to the way we extract build assets from the docker container that builds them.

I'll add more comments in the PR, but in a nutshell: we copy the assets out of the container post-build. As opposed to mounting a folder into the container prior to build, only to copy assets into it afterwards (within the same build context). Happy side effect (🌈) is that the asset is no longer owned by root, which has caused weird issues in the past (see the awkward/seemingly unnecessary `mkdir -p` in the buildkite YAML of this repo, as well as in the BCK pipeline.

The reason I'm doing this _now_ is that it became important to be able to reference _the exact build container_ for BCK; we're extracting assets other than the WHL for that pipeline. 
> Note: I could have added more commands to Kolibri's build steps to do the same, or at least to trigger the other pipeline without rebuilding the entire WHL. But I thought it'd be best to avoid adding complexity to this pipeline, given the recent discussions around GHActions with @rtibbles. For the same reason, I've avoided consolidating more of the build logic to runtime and slimming down the build image. 

*_HOWEVER_*, I'd like to point out that all (most?) of the other installers have been given this treatment (where we forego image/container names in favor of IDs). In these cases, it was done to allow for build parallelization, because reusing image/container between builds means that there could be collisions in a parallel-build scenario. Bad times.

Hence the name of feature branch. Another happy side effect (🌈 🌈 ) to getting this merged would be that we could (in theory) add more build agent processes to our builders, which would parallelize builds and reduce wait times. But I'd hold off on those experiments until after release :)

## Reviewer guidance

Keep an eye on the build associated with this PR and ensure that the right assets are uploaded to buildkite when it ends.

During testing, I noticed some odd GHAction check errors around documentation. Please advise @rtibbles 

----

## Testing checklist

## PR process

    It wouldn't HURT to merge this back to older Kolibri versions. And if we end up moving forward with parallelization, it might be necessary. Looking for guidance here @rtibbles 

## Reviewer checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -252,10 +252,10 @@ docker-whl: docker-envlist docker-clean
 	docker run \
 		--env-file ./docker/env.list \
 		--cidfile docker-whl.cid \
-		-v $$PWD/dist:/kolibridist \
 		-v yarn_cache:/yarn_cache \
 		-v cext_cache:/cext_cache \
 		`cat docker-whl.iid`
+	docker cp `cat docker-whl.cid`:/kolibridist/* dist
 	git checkout -- ./docker/env.list  # restore env.list file
 
 docker-build-base: writeversion
--- a/Makefile
+++ b/Makefile
@@ -242,7 +242,7 @@ i18n-download-glossary:
 i18n-upload-glossary:
 	python packages/kolibri-tools/lib/i18n/crowdin.py upload-glossary
 
-docker-whl: writeversion docker-envlist
+docker-whl: docker-envlist
 	docker image build -t "learningequality/kolibri-whl" -f docker/build_whl.dockerfile .
 	docker run \
 		--env-file ./docker/env.list \
--- a/Makefile
+++ b/Makefile
@@ -255,7 +255,7 @@ docker-whl: docker-envlist docker-clean
 		-v yarn_cache:/yarn_cache \
 		-v cext_cache:/cext_cache \
 		`cat docker-whl.iid`
-	docker cp `cat docker-whl.cid`:/kolibridist/* dist
+	docker cp `cat docker-whl.cid`:/kolibri/dist dist
 	git checkout -- ./docker/env.list  # restore env.list file
 
 docker-build-base: writeversion
--- a/Makefile
+++ b/Makefile
@@ -242,14 +242,20 @@ i18n-download-glossary:
 i18n-upload-glossary:
 	python packages/kolibri-tools/lib/i18n/crowdin.py upload-glossary
 
-docker-whl: docker-envlist
-	docker image build -t "learningequality/kolibri-whl" -f docker/build_whl.dockerfile .
+docker-clean:
+	rm -f *.iid *.cid
+
+docker-whl: docker-envlist docker-clean
+	docker build \
+		--iidfile docker-whl.iid \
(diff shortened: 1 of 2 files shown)


DXCanas:
@marcellamaki Tagging because you've expressed interest. Feel no obligation to review or even understand. Will happily take questions either way.

Makefile
```diff
@@ -242,14 +242,20 @@ i18n-download-glossary:
 i18n-upload-glossary:
 	python packages/kolibri-tools/lib/i18n/crowdin.py upload-glossary
 
-docker-whl: writeversion docker-envlist
-	docker image build -t "learningequality/kolibri-whl" -f docker/build_whl.dockerfile .
```
DXCanas, line 246: Note removal of `-t` (for "tag", as in Docker tag) flag.

Makefile
```diff
@@ -242,14 +242,20 @@ i18n-download-glossary:
...
+	docker build \
+		--iidfile docker-whl.iid \
+		-f docker/build_whl.dockerfile .
 	docker run \
 		--env-file ./docker/env.list \
-		-v $$PWD/dist:/kolibridist \
+		--cidfile docker-whl.cid \
 		-v yarn_cache:/yarn_cache \
 		-v cext_cache:/cext_cache \
-		"learningequality/kolibri-whl"
```
DXCanas, line 252: Note removal of `-t` (for "tag", as in Docker tag) flag.

Makefile
```diff
@@ -242,14 +242,20 @@ i18n-download-glossary:
...
+docker-clean:
+	rm -f *.iid *.cid
+
+docker-whl: docker-envlist docker-clean
+	docker build \
+		--iidfile docker-whl.iid \
+		-f docker/build_whl.dockerfile .
 	docker run \
 		--env-file ./docker/env.list \
-		-v $$PWD/dist:/kolibridist \
```
DXCanas, line 249: Note removal of local dir mounting. Avoids permissions mixup that used to happen as a result of [copying files to this mounted directory from inside the container's context](https://github.com/learningequality/kolibri/pull/8531/files#diff-762d9324eb4f6cbd300524e73492fc1b699f454901c397ea774dda65a943c048L56).

DXCanas (commented):
Some notes to describe changes for the curious.

Makefile
```diff
@@ -242,14 +242,20 @@ i18n-download-glossary:
...
+		--cidfile docker-whl.cid \
 		-v yarn_cache:/yarn_cache \
 		-v cext_cache:/cext_cache \
-		"learningequality/kolibri-whl"
+		`cat docker-whl.iid`
```
DXCanas, line 257: I'm aware that some folks have opinions about the backtick syntax in `sh`/`bash`.

To those folks: I hear you. It was either this or arguing with GNU Make's `${VARIABLE_REFERENCE_SYNTAX}`.

I'm sure there are other ways. But this one was POSIX standard and did everything needed here. I also used it in another, similar, script within the infra universe. So precedence.

rtibbles (approved):

Issue #8307, opened 2021-08-17
Implement Side Informational Panel throughout new Learn experience


marcellamaki:
**Overview:**
While the container component is handled in #8107, the implementation of the component, and connection to relevant cards, toolbars, etc. still needs to be completed. 

The sidepanel can be accessed 3 ways:
- from the content renderer (as a resource informational panel) 
- from the content renderer (as a topic/lesson informational panel) 
- from a single card by the (i) icon (as a resource informational panel) 

**Done when:**
- The three interactions above all work and the correct metadata is displayed for each

<img width="443" alt="Screen Shot 2021-08-17 at 10 46 41 AM" src="https://user-images.githubusercontent.com/17235236/129748169-0dfa7378-e38a-45b1-8e96-7728e5311e6d.png">
<img width="466" alt="Screen Shot 2021-08-17 at 10 46 56 AM" src="https://user-images.githubusercontent.com/17235236/129748171-0eba9805-ce37-4b12-b3bf-0ac314aff1c7.png">
<img width="682" alt="Screen Shot 2021-08-17 at 10 45 15 AM" src="https://user-images.githubusercontent.com/17235236/129748166-0ff54a33-2ff2-459a-8666-965a97fc5284.png">

PR #8545, opened 2021-10-27
Side Panel Metadata

nucleogenesis:
## Summary

Implements 2 / 3 major Side Panel displays **with proper metadata** and tests and such. 

The two of the three which are done here are:

- When you are browsing the library/channels, there is an (i) icon which will open extended metadata in  the right sidepanel.

image

- When you are viewing content, you can click the top-right context menu, then "view more information" to see a smaller subset of metadata.

image

The one which is not completed (maybe even two of its own, but Figma seems like they'll look the same anway):
- When viewing a piece of content, you can click "View Folder Resources" or "View Lesson Resources" which shows the currently viewed piece of content's siblings.

This will be done in a follow-up PR because I have a bad habit of making monolithic PRs and caught myself before I just plowed on to the next bit. 

## References

Continues to work toward #8107 #8307 - soon to be closed!

## Reviewer guidance

Please check to see if anything breaks, that the panels appear with the content they should have. You'll need to get like KA English and import some folders that have the same content to test that the Locations section in the browse resource metadata shows "other locations in this channel" properly. I used KA-en early math stuff.

----

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/core/assets/src/mixins/commonCoreStrings.js
+++ b/kolibri/core/assets/src/mixins/commonCoreStrings.js
@@ -1075,7 +1075,7 @@ export default {
      * if available.
      *
      * @param {string} key - A key as defined in the coreStrings translator; also accepts keys
-     * for the object kolibri.coreVue.vuex.constants.MetadataLookup.
+     * for the object MetadataLookup.
      * @param {object} args - An object with keys matching ICU syntax arguments for the translation
      * string mapping to the values to be passed for those arguments.
      */
--- a/kolibri/core/assets/src/mixins/commonCoreStrings.js
+++ b/kolibri/core/assets/src/mixins/commonCoreStrings.js
@@ -1091,7 +1091,7 @@ export default {
         return coreStrings.$tr(nonconformingKeys[key], args);
       }
 
-      return coreStrings.$tr(camelCase(key), args);
+      return coreStrings.$tr(key, args);
     },
     /**
      * Shows a specific snackbar notification from our notificationStrings translator.
--- a/kolibri/core/assets/src/mixins/commonCoreStrings.js
+++ b/kolibri/core/assets/src/mixins/commonCoreStrings.js
@@ -1,4 +1,8 @@
 import { createTranslator } from 'kolibri.utils.i18n';
+import camelCase from 'lodash/camelCase';
+import get from 'lodash/get';
+import invert from 'lodash/invert';
+import * as METADATA from 'kolibri.coreVue.vuex.constants';
 import notificationStrings from './notificationStrings';
 
 export const coreStrings = createTranslator('CommonCoreStrings', {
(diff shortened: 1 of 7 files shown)


kolibri/plugins/learn/assets/src/views/BrowseResourceMetadata.vue
```diff
@@ -151,48 +143,55 @@
...
       <div class="label">
-        {{ metadataStrings.$tr('locationsInChannel', { 'channelName': content.parent.title }) }}:
+        {{
+          metadataStrings.$tr('locationsInChannel', { 'channelname': content.ancestors[0].title })
+        }}:
+      </div>
+      <div v-for="location in locationsInChannel" :key="location.id">
+        <div>
+          <KRouterLink
+            :to="genContentLink(location.ancestors[location.ancestors.length - 1].id, false)"
```
rtibbles: Could also do:
```
location.ancestors.slice(-1)[0].id
```
here for a little more concision.

kolibri/plugins/learn/assets/src/views/BrowseResourceMetadata.vue
```diff
@@ -254,6 +295,18 @@
...
+      ContentNodeResource.fetchRecommendationsFor(this.content.id).then(recommendations => {
+        const threeRecs = recommendations.splice(0, 3);
+        this.recommendations = threeRecs.length ? threeRecs : null;
+      });
+
+      if (this.showLocationsInChannel) {
+        // Retreives any topics in this same channel
+        ContentNodeResource.fetchCollection({
+          getParams: { content_id: this.content.content_id, channel_id: this.content.channel_id },
+        }).then(parents => (this.locationsInChannel = parents && parents.length ? parents : null));
```
rtibbles: Might be good to not call this temporary variable `parents` as it's the other locations, not the parents of those other locations.

kolibri/core/assets/src/mixins/commonCoreStrings.js
```diff
@@ -1028,17 +1032,76 @@ const noneOfTheAboveTranslator = createTranslator('PerseusInternalMessages', {
...
+     * Return translated string for key defined in the coreStrings translator. Will map
+     * ID keys generated in the kolibri-constants library to their appropriate translations
+     * if available.
+     *
+     * @param {string} key - A key as defined in the coreStrings translator; also accepts keys
+     * for the object MetadataLookup.
+     * @param {object} args - An object with keys matching ICU syntax arguments for the translation
+     * string mapping to the values to be passed for those arguments.
+     */
     coreString(key, args) {
```
rtibbles, line 1082: I think I'd rather still do this as a separate method that accepts `labelType, key, args` as arguments. At the moment this works because (by some miracle) the keys for all of the different METADATA categories do not collide (in spite of my fears that they do).

There's nothing in le-utils that guarantees this, so having the separate method would help to avoid very subtle and confusing bugs for our future selves in the case where a key just happens to collide.

rtibbles (commented):
Looking good from a quick read through - a couple of comments/caveats.

kolibri/core/assets/src/mixins/commonCoreStrings.js
nucleogenesis: That approach seems like it'd necessarily result in an api like:

`coreMetadataStrings(MetadataKinds.ACCESSIBILITY_CATEGORIES, content.accessibility_categories)` along with an `import MetadataKinds from 'kolibri.coreVue.vuex.constants'` in every file where it's used.

Avoiding that was my main motivation for the simpler approach of a single method & an escape hatch check against the `nonconformingKeys` where we could put rare one-off oddities like misspellings and such until the next strings pass.

> There's nothing in le-utils that guarantees this

Can we write a test that just checks all of those files and that they have no key collisions in LE Utils? So whenever they're generated, we can know for sure that there are no collisions.

Additionally, since these strings are so tightly coupled to LE Utils now, it's maybe worth writing tests in Kolibri to ensure conformity to our expected relationship between the two. Perhaps we could guarantee in test that all metadata strings defined are camelCase and successfully map to some equivalent SNAKE_CASE entry in the `kolibri-constant` lib?

rtibbles: We can definitely go the route of ensuring that this never happens in the future.

Have opened an issue here in le-utils for that: https://github.com/learningequality/le-utils/issues/96

rtibbles:
Only thing I noticed in manual review was this floating close button that perhaps needed some contrast to the background to make it more like a material FAB?

Screenshot from 2021-10-28 13-42-35

Going to merge this, and we'll open a follow up issue if needed.

rtibbles (approved):
Code looks good, manual testing checks out. Anything else can be done as follow up.

Issue #8533, opened 2021-10-25
Audit and cleanup components introduced for hybrid learning features


MisRob:
To avoid regressions, we often introduced new components for hybrid learning features (e.g. thumbnails and cards used on the home page). This issue is to check the original before hybrid-learning components and merge them with the new components whenever appropriate. Also, it'd be good to check if we have some duplicate hybrid learning components since we were often developing features with similar elements in parallel.

## User-facing consequences

Improved UX/UI consistency

MisRob:
For cards, we may even consider having a base KDS component, see https://github.com/learningequality/kolibri-design-system/issues/263

marcellamaki:
Part of the card work should include considering the `HybridLearningCardGrid` and the various card designs within it see, [HybridLearningCardGrid.vue](https://github.com/learningequality/kolibri/blob/181a2e7d19b5da4a50523d16a5fe3ebffdac9eda/kolibri/plugins/learn/assets/src/views/HybridLearningCardGrid.vue) 

Points to consider should include: simplifying conditional logic (such as more sustainable/straightforward ways of determining the relationship between card display, current page, and current screen size)  and potential, as part of that, reducing the number of card components used

MisRob:
Related: #8564, #8566

PR #8504, opened 2021-10-18
The hybrid learning home page

MisRob:
## Summary

Add a new home page. It doesn't follow the full spec in Figma but is rather an MVP based on #8137. See test scenarios below for details. This PR is meant to be reviewed together with #8430, both code and testing, which contains the majority of work and was merged prematurely due to the string freeze. You're welcome to add review comments to already merged #8430.

## References

- closes #8137
- [full feature spec in Figma](https://www.figma.com/file/uiZuCIqh2KYvBUfbCiJyyA/Hybrid-Learning?node-id=1788%3A6234)
- [home page cards designs in Figma](https://www.figma.com/file/uiZuCIqh2KYvBUfbCiJyyA/Hybrid-Learning?node-id=114%3A8916)

## Reviewer guidance

See [the following test scenarios in Notion](https://learningequality.notion.site/Home-Page-QA-8ffb1011fa034e21bc1f52a2aad585ac)

----

## Testing checklist

- [x] Contributor has fully tested the PR manually
- [x] If there are any front-end changes, before/after screenshots are included
- [ ] Critical user journeys are covered by Gherkin stories
- [x] Critical and brittle code paths are covered by unit tests

## PR process

- [x] PR has the correct target branch and milestone
- [x] PR has 'needs review' or 'work-in-progress' label
- [x] If PR is ready for review, a reviewer has been added. (Don't use 'Assignees')
- [x] If this is an important user-facing change, PR or related issue has a 'changelog' label
- [ ] If this includes an internal dependency change, a link to the diff is provided

## Reviewer checklist

- Automated test coverage is satisfactory
- PR is fully functional
- PR has been tested for [accessibility regressions](http://kolibri-dev.readthedocs.io/en/develop/manual_testing.html#accessibility-a11y-testing)
- External dependency files were updated if necessary (`yarn` and `pip`)
- Documentation is updated
- Contributor is in AUTHORS.md

PR #8548, opened 2021-10-27
Many hybrid learning things

marcellamaki:
## Summary

Due to having many open PRs that have been built on top of each other, I have created a new branch that has merged or rebased that work, and then done a polish. This PR covers quite a few front end issues for the hybrid learning work, but hopefully it is in a state where reviewing it together, while extensive, does make testing and QA a bit easier. 

There are still a few bugs that are not fixed, but for the most part, the core functionality and mobile responsiveness should be in place. 🤞 

### Updated Lessons Cards on Class Page 
|  Description  |  Screenshot (minimal interactions to demo in this section, so just the screenshot |
|---|---|
| Desktop view  |  Screen Shot 2021-10-27 at 6 45 40 PM|
|  Mobile view |  Screen Shot 2021-10-27 at 6 45 50 PM |

### Library Page
| Description  |  Gif or Screenshot  |
|---|---|
|  "Default"/ Library landing view |  library-landing-page |
| Filtering options  |  library-filtering |
|  Category Search Modal  | category-search-modal |
|  Mobile Library view and search interactions  |  mobile-library-search |

### Channel Topic Browsing
| Description  |  Gif or Screenshot  |
|---|---|
|  Desktop channel browse  |  channel-topics-browsing |
|  Medium screen   | medium-channel-browse  |
| Mobile  | mobile-channel-browse  |

### Bookmarks Page

|   |   |
|---|---|
|  Bookmarks desktop |  Screen Shot 2021-10-27 at 6 48 24 PM |
| Mobile bookmarks, and remove bookmark interaction  |  mobile-bookmarks|

## Reviewer guidance

There is quite a lot of code in this PR, and I think _all_ of it deserves extra testing. Key points to test for (drawn from other PRs).

**Across ALL pages**
1. Accurate reflection of Figma specs (cc @jtamiace) - although not all metadata will be displayed
 - Updated card designs
 - Responsive layout working as expected across browsers and screens
 - Have I used the correct cards in all places, and have I interpreted the mobile responsive differences correctly? (I have done by best, but fresh eyes would certainly help because I have been staring at this for a very long time)
- Overall layout
- Buttons and tooltips (note, persistent "active" state not implemented yet)

**On Library, and Channel and topics browsing**
Filtering and keyword search as implemented by @rtibbles still working 😄. On Library page, filter for all channels that exist on device. On channel page, search results should be limited to the current channel:
 - Can search by keyword
 - Can search by single filter
 - Can search by multiple filters
 - Can view more
 - Can clear all
 - Can clear filters individually
 - Side panel accurately disables and/or hides filters and filter groups that are not available on search results

**On Bookmarks page**
- Complete list of bookmarks displayed 
- "Bookmarked time" accurately displayed on card 
- Removing a bookmark removes the _correct_ bookmark

#### General notes and comments

@pcenov has already pointed out that there need to be some improvements made on IE11, and I haven't forgotten this, but I haven't implemented it just yet. Would welcome any feedback (either in the PR, or on Slack or in a call) from @MisRob @nucleogenesis or @indirectlylit about how I can make that experience better.

Another particular area that needs some "look and feel" improvements is the Topics browsing generally... jankiness from the sticky sidebar which I am trying to calculate with computed props. Would _love_ an idea about how to do that differently/correctly/better. 

The pages themselves are a bit clunky - I had great aspirations to refactor into some smaller sub components, but for time reasons, I've just tried to add a few comments to layout what is happening overall, and which sections of the file are responsible for what on the Library and Topics pages, which are quite long. If there are any other areas that you think comments would be a helpful addition, please let me know and I will be happy to add before merging - it would be a low lift, and I think quite worth it. 

----

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated


kolibri/plugins/learn/assets/src/views/TopicsPage.vue
```diff
@@ -238,29 +434,213 @@
...
+        if (header) {
+          let position = header.getBoundingClientRect();
+          if (position.bottom >= 64) {
+            this.stickyTop = `${position.bottom}px`;
+          } else {
+            this.stickyTop = '64px';
+          }
+        } else {
+          null;
+        }
```
bjester, line 662: Since this is being called on scroll, you might think about optimizing this code. Having this logic executed within an [animation frame](https://developer.mozilla.org/en-US/docs/Web/API/window/requestAnimationFrame) seems acceptable too. For that, I'm not sure if we have any utils for that in Kolibri but I made a [small one in studio](https://github.com/learningequality/studio/blob/unstable/contentcuration/contentcuration/frontend/shared/utils/helpers.js#L308).

kolibri/core/assets/src/views/KeenUiToolbar.vue
```diff
@@ -34,8 +34,8 @@
     }
   }
 
-  .ui-toolbar {
```
MisRob: What was the reason for commenting this style out? It'll affect all toolbars in the app built around `KeenUiToolbar`/`UiToolbar`. For example, seems to be the cause of the following regression. See space on the right side of the dropdown:

| Before | After |
| --- | --- |
| toolbar-before | toolbar-after |

kolibri/plugins/learn/assets/src/views/HybridLearningCardGrid.vue
```diff
@@ -0,0 +1,200 @@
...
+      },
+      cardViewStyle: {
+        type: String,
+        required: true,
+        default: 'card',
+        validator(value) {
+          return ['card', 'list'].includes(value);
+        },
+      },
+      currentPage: {
```
MisRob, line 138: Using `currentPage` in `HybridLearningCardGrid` to decide if lessons cards or content cards should be displayed creates coupling that will presumably be hard to work with in the future.

I assume this whole component is a part of larger tech debt related to hybrid learning - could you please document it together with other similar components from this PR [in this issue](https://github.com/learningequality/kolibri/issues/8533) so that we can wrap our heads around more easily around what needs to be done in KDS or Kolibri?

MisRob (commented):
Thank you, @marcellamaki

I haven't managed to go through all files yet but am posting what I've noticed so far. I'll continue with the next review batch later. If @nucleogenesis gets to finishing the review sooner than me, feel free to proceed with merging after addressing important feedback.

MisRob:
> @pcenov has already pointed out that there need to be some improvements made on IE11, and I haven't forgotten this, but I haven't implemented it just yet. Would welcome any feedback (either in the PR, or on Slack or in a call) from @MisRob @nucleogenesis or @indirectlylit about how I can make that experience better.

Firstly, can you run IE11 on your machine, in VM for example? Maybe @jacob and @devon will have some general advice, I don't have any. I usually just try to at least briefly preview new components there and if I need to fix something, it usually requires lots of googling for known IE problems. I am glad to help with that, however would need to hear in more detail about particular issues that you have problems with. If it's nothing blocking, we can also open follow-up issues for this.

kolibri/plugins/learn/assets/src/views/TopicsPage.vue
marcellamaki: Thanks, @bjester - having the code reference is really helpful!

kolibri/core/assets/src/views/KeenUiToolbar.vue
marcellamaki: This was accidental! Thanks for catching this, Misha. I'll revert.

pcenov:
Hi @marcellamaki, today I had some time to test mainly the Browse channel pages and here are the findings as compared to Figma:
1. The channel name is not positioned correctly, there is too much white space, the Folders option is not underlined and selected by default, and there is displayed breadcrumb navigation while in Figma it isn't. Also clicking the Channels link doesn't result in anything:



2. Missing 'Show more' option:



3. Resources view - several issues, see the screenshot:



4. End folder view - the Folders section is missing, missing folder title and options



5. Search results: the search tab is not selected and underlined, the cards are squished, the x icon is incorrect, no breadcrumb navigation should be displayed and missing options for grid/list view 



Other issues:
7. Both me and Radina saw an issue where the breadcrumb navigation gets mixed up but I didn't have time to find a way to reliably replicate it:

indirectlylit:
Related to IE11, [cross-posting from Slack](https://learningequality.slack.com/archives/C023DNC5GCX/p1635438508097100):

> One general rule we used try and enforce in Kolibri was "no flexbox allowed" because IE and early Android implemented different versions of it with slight incompatibilities and gotchas. For example, our KGrid is implemented using inline blocks specifically to avoid flexbox.
> 
> Over time my sense is that engineers really wanted to have that tool in our CSS toolbox and we've started using it more liberally. If many of the IE issues we're encountering turn out to be flexbox related, perhaps it's time to revisit this rule and "ban" flexbox again
> 
> Here's a [recent example](https://github.com/learningequality/kolibri/pull/8504/files#diff-00d7eb78df39a30f6e86e658b461bce7699a9ff9bd555cf37adce16d61d3a292R45) from @MisRob's PR
>
> Most of these guidelines aren't relevant to the IE issue, but [here's where that "rule" was documented](https://kolibri-dev.readthedocs.io/en/develop/frontend_architecture/conventions.html#styling-anti-patterns) (without justification or explanation unfortunately)

kolibri/plugins/learn/assets/src/views/CategorySearchModal/CategorySearchModalGrid.vue
```diff
@@ -0,0 +1,131 @@
...
+        return Object.values(this.displaySelectedCategories).some(
+          obj => Object.keys(obj.nested).length
+        );
+      },
+      displaySelectedCategories() {
+        return LibraryCategories[this.selectedCategory].nested;
+      },
+      isLevel0() {
+        return this.windowWidth <= 330;
+      },
```
nucleogenesis: What caused the need to check for something smaller than what `this.windowIsSmall` uses (<480px)? If it's supposed to handle even smaller-than-small screens it'd help to have a comment block above this explaining what it does and why.

I also don't see how I can even open this window on the smaller screen right now. When I make my screen that thin I just see the cards and no search / filtering links.

kolibri/plugins/learn/assets/src/views/CategorySearchModal/CategorySearchModalGrid.vue
```diff
@@ -0,0 +1,131 @@
...
+
+<style lang="scss" scoped>
+
+  .level-0 {
+    position: fixed;
+    top: 0;
+    right: 0;
+    bottom: 0;
+    left: 0;
+    z-index: 35;
```
nucleogenesis: We want to try to keep the z-index between 0-24 per Material specifications wherever we can - if this is not visible at 24 then you may be able to adjust. If you find that you really need to go higher than 24 for some reason, then a comment documenting that would be very helpful in case someone else can't figure out why they cannot put the snackbar on top of your modal or something.

This isn't a context where we'll suddenly decide we need _another_ thing, even a snackbar, on top of the modal so I wouldn't spend too much time hammering it out if you run into trouble with regressions tweaking other z-indexes

kolibri/plugins/learn/assets/src/views/CategorySearchModal/CategorySearchModalGrid.vue
```diff
@@ -0,0 +1,131 @@
...
+        );
+      },
+      displaySelectedCategories() {
+        return LibraryCategories[this.selectedCategory].nested;
+      },
+      isLevel0() {
+        return this.windowWidth <= 330;
+      },
+      responsiveSpan() {
+        if (this.windowWidth < 480) {
```
nucleogenesis: This is redundant with `this.windowIsSmall` - also there is a dire looking warning about using `this.windowWidth` in computed props in KResponsiveWindowMixin. The helpers provided here should help avoid any issues

https://github.com/learningequality/kolibri-design-system/blob/04d71dabf0cb4030da31b7a33a2a8f754df1c6c9/lib/KResponsiveWindowMixin.js#L135-L147

kolibri/plugins/learn/assets/src/views/CategorySearchModal/CategorySearchModalGrid.vue
```diff
@@ -0,0 +1,131 @@
...
+          return 6;
+        } else {
+          return 4;
+        }
+      },
+    },
+    methods: {
+      camelCase(val) {
+        return camelCase(val);
+      },
```
nucleogenesis, line 93: You can also just put `methods: { camelCase },` to the same effect - assigning the value of the function in the `camelCase` variable to the key of the same name in the methods object.

kolibri/plugins/learn/assets/src/views/ContentCard/CardThumbnail.vue
```diff
@@ -4,32 +4,148 @@
...
+            return { fill: this.$themeTokens.audio };
+          case ContentNodeKinds.DOCUMENT:
+            return { fill: this.$themeTokens.document };
+          case ContentNodeKinds.HTML5:
+            return { fill: this.$themeTokens.html5 };
+          case ContentNodeKinds.SLIDESHOW:
+            return { fill: this.$themeTokens.slideshow };
+          default:
+            return { fill: this.$themeTokens.topic };
+        }
```
nucleogenesis, line 147: A slightly more concise and clear alternative to a big switch like this is to return an object where all of the potential keys are mapped, then pass `this.kind` to access the correct key's value:

```
return  {
    [ContentNodeKinds.EXERCISE]: { fill: this.$themeTokens.exercise },
    [ContentNodeKinds.VIDEO]: { fill: this.$themeTokens.video },
    [ContentNodeKinds.AUDIO]: { fill: this.$themeTokens.audio },
    [ContentNodeKinds.DOCUMENT]: { fill: this.$themeTokens.document },
    [ContentNodeKinds.HTML5]: { fill: this.$themeTokens.html5 },
    [ContentNodeKinds.SLIDESHOW]: { fill: this.$themeTokens.slideshow },
    [ContentNodeKinds.TOPIC]: { fill: this.$themeTokens.topic },
}[this.kind];
```

kolibri/plugins/learn/assets/src/views/HybridLearningCardGrid.vue
```diff
@@ -0,0 +1,200 @@
...
+      :isLeaf="content.is_leaf"
+      :progress="content.progress_fraction || 0"
+      :numCoachContents="content.num_coach_contents"
+      :link="genContentLink(content.id, content.is_leaf)"
+      :contentId="content.content_id"
+      :copiesCount="content.copies_count"
+      :footerIcons="footerIcons"
+      :createdDate="content.bookmark ? content.bookmark.created : null"
+      @openCopiesModal="openCopiesModal"
+      @removeFromBookmarks="removeFromBookmarks(content, contents)"
```
nucleogenesis: There are a ton of props here and a lot of them are just extractions of values from `content`. If you instead just pass one prop `:content="content"` to the `HybridLearningContentCardListView` component, then you'll save a lot of code in that component by removing all of the prop definitions and in every component using it as well. The goes for the `HybridLearning(Lesson|Content)Card` components as well. In those components you can then just reference `content` wherever you need it's properties.

I don't think that this is blocking as it's mostly stylistic for now - and I'd be happy to write a follow-up issue for this as well.

kolibri/plugins/learn/assets/src/views/HybridLearningCardGrid.vue
```diff
@@ -0,0 +1,200 @@
...
+      currentPage: {
+        type: String,
+        default: null,
+        required: false,
+      },
+      numCols: {
+        type: Number,
+        required: true,
+      },
+      genContentLink: {
```
nucleogenesis: This is available at `learn/assets/src/utils/genContentLink` to import so it doesn't have to be passed as a prop

kolibri/plugins/learn/assets/src/views/LibraryPage.vue
```diff
@@ -370,4 +503,54 @@
...
+    margin: 4px;
+    margin-left: 8px;
+  }
+
+  .filter-chip {
+    display: inline-block;
+    margin: 2px;
+    font-size: 14px;
+    vertical-align: top;
+    background-color: #dedede;
```
nucleogenesis, line 545: This is our `$themeTokens.fineLine` color but you could also use the `$themePalette.grey.v_###` tokens to get on-theme colors for this. Unfortunately you'd have to do this in the JS above and apply partial styles there. 

Not blocking or major but just a note.

nucleogenesis (commented):
I think that most of my notes here shouldn't be blocking as long as their tracked in a follow-up issue and handled soon.

The only thing important before release I think is the places where you used `this.windowWidth` from `KResponsiveWindowMixin` - there is a stern warning in the mixin not to do that.

The rest are notes about code style/conventions and such for the most part and can be tackled later on in a follow-up issue.

kolibri/plugins/learn/assets/src/views/HybridLearningCardGrid.vue
nucleogenesis: I saw a few other places where there was a prop being passed with this value when it could instead be imported so we should look out for other areas when writing this into a follow-up

kolibri/core/assets/src/views/TextTruncator.vue
```diff
@@ -135,6 +135,9 @@
   .truncated {
     overflow: hidden;
     text-overflow: ellipsis;
+    a {
+      text-decoration: none !important;
+    }
```
indirectlylit, line 142: It looks like this is overriding styling coming from inside `shave`? It could be useful to add a comment describing what the intended outcome is, e.g.:

```suggestion
    // this style ensures that ____
    a {
      text-decoration: none !important;
    }
```

kolibri/plugins/learn/assets/src/modules/coreLearn/utils.js
```diff
@@ -3,16 +3,25 @@ import { ContentNodeKinds } from 'kolibri.coreVue.vuex.constants';
 import { assessmentMetaDataState } from 'kolibri.coreVue.vuex.mappers';
 import { getContentNodeThumbnail } from 'kolibri.utils.contentNode';
 import tail from 'lodash/tail';
+import plugin_data from 'plugin_data';
+
+const channels = {};
+for (let channel of plugin_data.channels) {
```
rtibbles, line 9: Once this is merged - we can fully bootstrap available channels into the home page and use plugin_data for @MisRob's getChannels composition helper.

kolibri/plugins/learn/assets/src/views/CategorySearchModal/CategorySearchModalOptions.vue
```diff
@@ -1,61 +1,73 @@
...
         <KIcon
           icon="info"
           size="large"
         />
-        <h2>{{ coreString(camelCase(key)) }}</h2>
-        <p
-          v-for="(item, nestedKey) in nestedObject.nested"
-          :key="item.value"
+        <h2 @click="$emit('input', nestedObject.value)">
+          {{ coreString(camelCase(key)) }}
```
rtibbles, line 23: Once this is merged, we can clean this up to use the consolidated translation logic that @nucleogenesis implemented.

I think we need a general issue for string follow up. I can file, as I think I know most of the extant issues.

kolibri/plugins/learn/assets/src/views/CategorySearchModal/CategorySearchModalOptions.vue
```diff
@@ -154,12 +176,16 @@
         return camelCase(val);
       },
     },
-    $trs: {
-      title: {
-        message: 'Choose a category',
```
rtibbles, line 159: Flagging for any other concerned reviewers that this string has stayed in the same namespace by virtue of persisting in the main `CategorySearchModal` component.

kolibri/plugins/learn/assets/src/views/CategorySearchModal/index.vue
```diff
@@ -0,0 +1,118 @@
...
+      <CategorySearchModalOptions
+        :selectedCategory="selectedCategory"
+        :availableLabels="availableLabels"
+        :span="windowIsMedium ? 6 : 4"
+        numCols="12"
+        v-on="$listeners"
+      />
+    </KModal>
+
+    <h2>{{ $tr('title') }}</h2>
```
rtibbles: Shouldn't this block be wrapped in a div with a `v-else` or `v-else-if='fullscreen'`?

kolibri/plugins/learn/assets/src/views/ChannelCardGroupGrid.vue
```diff
@@ -66,7 +66,12 @@
       },
     },
     methods: {
-      genContentLink,
+      genChannelLink(channel_id) {
```
rtibbles, line 69: If there's a specific reason not to use the util here, put a code comment. If there's not, let's revert this change!

kolibri/plugins/learn/assets/src/views/ContentCard/CardThumbnail.vue
```diff
@@ -4,32 +4,148 @@
     class="card-thumbnail-wrapper"
     :class="{ 'mobile-thumbnail': isMobile }"
     :style="thumbnailBackground"
-  ></div>
+  >
+
+    <ContentIcon
```
rtibbles, line 9: I'm still a little confused why we are still showing kind icons anywhere in the user facing UI. My preference would be to remove this completely and use learning activity iconography instead. @jtamiace thoughts? I believe this component is being used in the lesson display?

kolibri/plugins/learn/assets/src/views/EmbeddedSidePanel/SelectGroup.vue
```diff
@@ -157,23 +157,41 @@
 
 <style lang="scss" scoped>
 
+  /deep/ .ui-select-label-text.is-inline {
+    position: absolute;
+    bottom: 45px;
+    left: 10px;
+    font-size: 12px;
+    color: black;
```
rtibbles: Should this be black or some theming derived colour?

kolibri/plugins/learn/assets/src/views/CategorySearchModal/CategorySearchModalGrid.vue
indirectlylit: Agreed that it's best to use the built-in small/medium/large shortcuts whenever possible.

In this case though, it seems like maybe `KGridItem` and `KGrid` could provide the responsiveness? For example:

```vue
     <KGrid :style="{ margin: '24px' }">
        <KGridItem
          v-for="(value, key) in displaySelectedCategories"
          :key="value.value"
          :layout="{ span: 4 }"
          :layout12="{ span: 3 }"
        >
          <KIcon icon="info" size="large" />
          <h2>{{ coreString(camelCase(key)) }}</h2>
        </KGridItem>
      </KGrid>
```

https://design-system.learningequality.org/kgriditem/

> also there is a dire looking warning about using this.windowWidth in computed props

The comment could have used more explanation, but the reason was that when resizing a window the computed prop will be recomputed for every pixel. On the other hand, the small/medium/large shortcuts only trigger recomputes on the thresholds.

kolibri/plugins/learn/assets/src/views/HybridLearningCardGrid.vue
```diff
@@ -0,0 +1,200 @@
...
+          :copiesCount="content.copies_count"
+          :description="content.description"
+          :channelThumbnail="content.channel_thumbnail"
+          :channelTitle="content.channel_title"
+          @openCopiesModal="openCopiesModal"
+        />
+      </KFixedGridItem>
+    </KFixedGrid>
+    <HybridLearningLessonCard
+      v-for="content in contents"
```
rtibbles: Note that using v-for and v-if on the same element is not recommended: https://vuejs.org/v2/guide/conditional.html#v-if-with-v-for

If you use v-if/v-else-if/v-else on a wrapper element and then do the v-for within that wrapped element, then you will avoid unnecessary looping as otherwise the v-for takes precedence.

kolibri/plugins/learn/assets/src/views/CategorySearchModal/CategorySearchModalGrid.vue
indirectlylit: mentioned in this KDS issue: 

https://github.com/learningequality/kolibri-design-system/issues/265#issuecomment-924440088

kolibri/plugins/learn/assets/src/views/HybridLearningCardGrid.vue
```diff
@@ -0,0 +1,200 @@
...
+
+    <HybridLearningContentCardListView
+      v-for="content in contents"
+      v-else
+      :key="content.id"
+      :contentNode="content"
+      :channelThumbnail="content.channel_thumbnail"
+      :channelTitle="content.channel_title"
+      :description="content.description"
+      :activityLength="content.activity_length"
```
rtibbles: Also worth noting that `activity_length` is not a valid attribute. `duration` is what you are looking for.

Seeing how it is used below, also worth noting that duration will be a numeric value, this will need to be mapped to the 'short' 'long' values for display.

kolibri/plugins/learn/assets/src/views/HybridLearningCardGrid.vue
```diff
@@ -0,0 +1,200 @@
...
+      :thumbnail="content.thumbnail"
+      :kind="content.kind"
+      :isLeaf="content.is_leaf"
+      :progress="content.progress_fraction || 0"
+      :numCoachContents="content.num_coach_contents"
+      :link="genContentLink(content.id, content.is_leaf)"
+      :contentId="content.content_id"
+      :copiesCount="content.copies_count"
+      :footerIcons="footerIcons"
+      :createdDate="content.bookmark ? content.bookmark.created : null"
```
rtibbles, line 86: I introduced this slightly brittle way of annotating the bookmark onto the content node - we should revisit this to think about how to handle this best.

kolibri/plugins/learn/assets/src/views/CategorySearchModal/index.vue
```diff
@@ -0,0 +1,118 @@
...
+            nested: {},
+          };
+        }
+        nested = nested[nestedKey].nested;
+        path += '.';
+      } else {
+        break;
+      }
+    }
+  }
```
indirectlylit, line 85: this code block could use some comments – the intent isn't totally obvious from a quick skim.

Maybe even a unit test as a follow-up?

kolibri/plugins/learn/assets/src/views/LearnIndex.vue
```diff
@@ -251,6 +251,12 @@
             hasSidebar: true,
           };
         }
+        // if (this.pageName === ClassesPageNames.LESSON_PLAYLIST) {
```
rtibbles: Reason for this commenting out?

kolibri/plugins/learn/assets/src/views/LibraryPage.vue
```diff
@@ -1,103 +1,208 @@
...
-      >
-        <div v-if="!displayingSearchResults">
-          <h2>{{ coreString('channelsLabel') }}</h2>
-          <ChannelCardGroupGrid
-            v-if="channels.length"
-            class="grid"
-            :contents="channels"
-            @toggleInfoPanel="toggleInfoPanel"
+      <div v-if="!windowIsLarge">
+        <!-- TO DO Marcella swap out new icon after KDS update -->
```
rtibbles, line 9: Let's do this as a follow up issue?

kolibri/plugins/learn/assets/src/views/HybridLearningContentCardListView.vue
```diff
@@ -0,0 +1,373 @@
...
+    right: 16px;
+    bottom: 16px;
+    display: inline;
+    .button {
+      width: 32px !important;
+      height: 32px !important;
+      /deep/ svg {
+        top: 4px !important;
+      }
+    }
```
indirectlylit: Using `!important` can sometimes cause unexpected bugs and behaviors in the future, e.g. if components are refactored. When it's used, we should generally add a comment explaining why it was necessary and look for opportunities to remove it eventually.

kolibri/plugins/learn/assets/src/views/LearnIndex.vue
```diff
@@ -251,6 +251,12 @@
...
+        //   return {
+        //     immersivePage: false,
+        //     hasSidebar: true,
+        //   };
+        // }
```
indirectlylit: maybe a mistaken addition?

kolibri/plugins/learn/assets/src/views/LibraryPage.vue
```diff
@@ -108,18 +213,19 @@
   import { mapState } from 'vuex';
   import uniq from 'lodash/uniq';
 
-  import FullScreenSidePanel from 'kolibri.coreVue.components.FullScreenSidePanel';
   import responsiveWindowMixin from 'kolibri.coreVue.mixins.responsiveWindowMixin';
   import { ContentNodeProgressResource, ContentNodeResource } from 'kolibri.resources';
   import commonCoreStrings from 'kolibri.coreVue.mixins.commonCoreStrings';
   import { AllCategories, NoCategories } from 'kolibri.coreVue.vuex.constants';
+  import genContentLink from '../utils/genContentLink';
+  import FullScreenSidePanel from '../../../../../core/assets/src/views/FullScreenSidePanel';
```
indirectlylit: would be better to add this to the core API via `apiSpec`

https://github.com/learningequality/kolibri/blob/release-v0.15.x/kolibri/core/assets/src/core-app/apiSpec.js

kolibri/plugins/learn/assets/src/views/LibraryPage.vue
```diff
@@ -296,26 +419,39 @@
...
             this.more = data.more;
             this.labels = data.labels;
             this.searchLoading = false;
           });
         }
+        // else {
+        //   ContentNodeResource.fetchCollection({ getParams }).then(data => {
+        //     this.labels = data.labels;
+        //   });
+        // }
```
indirectlylit: more dead code

kolibri/plugins/learn/assets/src/views/LibraryPage.vue
```diff
@@ -1,103 +1,208 @@
...
-              v-if="more"
-              :text="coreString('viewMoreAction')"
-              :primary="false"
-              :disabled="moreLoading"
-              @click="searchMore"
-            />
+        </div>
+        <div class="results-header-group">
+          <div
+            v-for="item in Object.values(searchTermChipList)"
```
rtibbles: To avoid having to do the search in `removeFilterTag` you can preserve both the key and the value here.
```suggestion
            v-for="(item, key) in searchTermChipList"
```

kolibri/plugins/learn/assets/src/views/LibraryPage.vue
indirectlylit: yes, at least before 0.15 is released it would be best to remove hard-coded colors. This probably won't cause an immediate issue because it's just a background color but if we were ever to introduce a 'dark' theme this will end up being an issue

kolibri/plugins/learn/assets/src/views/LibraryPage.vue
```diff
@@ -1,103 +1,208 @@
...
+            class="filter-chip"
+          >
+            <!-- TODO Marcella convert to strings, and add relevant aria label -->
+            <span>
+              <p class="filter-chip-text">{{ item }}</p>
+              <KIconButton
+                icon="close"
+                size="mini"
+                class="filter-chip-button"
+                @click="removeFilterTag(item)"
```
rtibbles: Can then invoke this function with both the item and the key:
```suggestion
                @click="removeFilterTag(item, key)"
```

kolibri/plugins/learn/assets/src/views/LibraryPage.vue
```diff
@@ -370,4 +503,54 @@
...
+  .filter-chip-text {
+    display: inline-block;
+    margin: 4px 0 4px 8px;
+    font-size: 14px;
+  }
+
+  .filter-chip-button {
+    padding-top: 4px;
+    margin: 2px;
+    color: #dadada;
```
indirectlylit, line 558: if you need a light text color there are:

* https://design-system.learningequality.org/colors/#tokens-annotation
* https://design-system.learningequality.org/colors/#tokens-textDisabled
* https://design-system.learningequality.org/colors/#tokens-textInverted

kolibri/plugins/learn/assets/src/views/LibraryPage.vue
```diff
@@ -230,6 +326,30 @@
       displayingSearchResults() {
         return Object.values(this.searchTerms).some(v => Object.keys(v).length);
       },
+      searchTermChipList() {
```
rtibbles, line 341: I would suggest using the `searchTerms` computed prop for this, rather than the query parameters - the searchTerms prop is formatted for access by the frontend code, the query params are formatted to make a valid URL.

kolibri/plugins/learn/assets/src/views/TopicsPage.vue
indirectlylit: FYI in Kolibri we have the [frame-throttle](https://www.npmjs.com/package/frame-throttle) library which is similar to @bjester's and also adds a fallback for older browsers that don't support `requestAnimationFrame`.

See for example:

https://github.com/learningequality/kolibri/blob/48075ed052ee4e61b6865e9e212880d527ae3a9e/kolibri/core/assets/src/views/CoreBase/index.vue#L382-L385

kolibri/plugins/learn/assets/src/views/LibraryPage.vue
```diff
@@ -296,26 +419,39 @@
...
             this.more = data.more;
             this.labels = data.labels;
             this.moreLoading = false;
           });
         }
       },
-      toggleInfoPanel(content) {
-        this.sidePanelContent = content;
+      removeFilterTag(value) {
+        let query = JSON.parse(JSON.stringify(this.$route.query));
```
rtibbles: This works, but seems unnecessary.

Also, you can use `const` here as you are never reassigning `query` to a different value (even though you are modifying the contents of the object that `query` is pointing to).
```suggestion
    removeFilterTag(value, key) {
        const keyObject = this.searchTerms[key];
        delete keyObject[value];
        this.searchTerms = {
            ...this.searchTerms,
            [key]: keyObject,
        };        
```

This way you lean on exactly the same setter/getter code that all the other selection components are using.

kolibri/plugins/learn/assets/src/views/LibraryPage.vue
```diff
@@ -296,26 +419,39 @@
...
+        const key = Object.keys(query).filter(function(key) {
+          return query[key] === value;
+        })[0];
+        delete query[key];
+        this.$router.replace({ query: query });
+      },
+      clearSearch() {
+        this.$router.push(this.$router.getRoute(PageNames.LIBRARY));
```
rtibbles:
```suggestion
        this.searchTerms = {};
```

Would also work here.

indirectlylit (commented):
looking great!

kolibri/plugins/learn/assets/src/views/TopicsPage.vue
```diff
@@ -238,29 +434,213 @@
...
         };
       },
-    },
-    methods: {
-      ...mapMutations('topicsTree', ['ADD_MORE_CONTENTS', 'ADD_MORE_CHILD_CONTENTS']),
-      loadMore() {
-        return ContentNodeResource.fetchTree(this.topic.children.more).then(data => {
-          this.ADD_MORE_CONTENTS(data.children);
-        });
+      searchTerms: {
```
rtibbles, line 450: Non-blocking, but I think it would be very helpful to make a composition API helper for the search logic here, and to do it before we release 0.15, to ensure we have consistent behaviour.

rtibbles (commented):
Nothing that I see as blocking - but definitely some things we can follow up on.

kolibri/plugins/learn/assets/src/views/CategorySearchModal/index.vue
rtibbles: I think I can take the blame for this code block...

kolibri/plugins/learn/assets/src/views/ContentCard/CardThumbnail.vue
jtamiace: It's hard for me to tell exactly what this component looks like, but yes, eventually the learning activity should eventually replace the old kind icons everywhere.

kolibri/plugins/learn/assets/src/views/HybridLearningCardGrid.vue
marcellamaki: I actually was wondering about this as I was doing it -- if there were any downsides to just passing the content object and then using computed props (i.e. performance considerations - either good or bad) so thanks for answering that this is a better approach, without me even asking! :)

rtibbles: We should definitely do the object based approach - the main caveat, thinking more generally, of passing objects is the lack of specification of what attributes that object should have. Ideally, we should have a companion function to `normalizeContentNode` - `validateContentNode` that would assert the presence of all relevant attributes on the content node object, that can then be used for the validator on the prop.

kolibri/plugins/learn/assets/src/views/CategorySearchModal/CategorySearchModalOptions.vue
rtibbles: Follow up issue filed here: https://github.com/learningequality/kolibri/issues/8555

kolibri/plugins/learn/assets/src/views/CategorySearchModal/CategorySearchModalGrid.vue
marcellamaki: So, there are "Level 0" responsive designs throughout the specs on many different pages, where the layout is different on xsmall screens (320px). I think in this case, I can remove this degree of granularity and with @indirectlylit 's suggestions below, some of this responsiveness could be achieved other ways, so the design implications would be fairly minimal. Seems like there's a follow up question though about having designs for xsmall screens, and as a product team, it is worth making sure we are on the same page for the future -- and maybe we can make some follow up issues in KDS to the responsiveWindowMixin as needed to account for this addition, if @jtamiace will be including xs screen designs in specs for the most or all future product work?

kolibri/plugins/learn/assets/src/modules/coreLearn/utils.js
rtibbles: Follow up issue filed here: https://github.com/learningequality/kolibri/issues/8556

kolibri/plugins/learn/assets/src/views/TopicsPage.vue
rtibbles: Follow up issue filed here: https://github.com/learningequality/kolibri/issues/8557

pcenov:
Hi @marcellamaki these are the additional issues I was able to identify after today's round of testing on all the major workflows and both me and Radina don't see as blockers:
1. A long description issue in Browse channel:


2. A similar issue in the Bookmarks:


3. Search results in mobile view when compared to Figma:


4. Sometimes the list/grid toggles appear in mobile view when there is a single column and they shouldn't. I tested this in real mobile phones as well:

kolibri/plugins/learn/assets/src/views/LibraryPage.vue
```diff
@@ -296,26 +419,38 @@
...
           ContentNodeResource.fetchCollection({ getParams }).then(data => {
-            this.results = data.results;
+            this.results = data.results.map(normalizeContentNode);
             this.more = data.more;
             this.labels = data.labels;
             this.searchLoading = false;
           });
+        } else {
+          ContentNodeResource.fetchCollection({ getParams }).then(data => {
+            this.labels = data.labels;
```
rtibbles: The available labels for the unfiltered case are bootstrapped into the page with plugin_data - so we should avoid doing this extra request when it isn't needed.

If you are trying to get the display to match what is currently available, then you should remove the thing I added for developer ease, where this available label filtering is not active in development mode.

kolibri/plugins/learn/assets/src/views/LibraryPage.vue
```diff
@@ -1,103 +1,208 @@
...
+            icon="channel"
+            :ariaLabel="$tr('viewAsGrid')"
+            :color="$themeTokens.text"
+            :tooltip="$tr('viewAsGrid')"
+            @click="toggleCardView('card')"
+          />
+        </div>
+        <div class="results-header-group">
+          <div
+            v-for="(item, key) in searchTermChipList"
```
rtibbles, line 87: searchTermChipList is still referencing the query parameters - this will break in an instance where there is a comma separated value in the query parameters.

I'd recommend just directly referencing `searchTerms` here, with the caveat that the 'item' is actually an object acting as a map with `[item]: true` - so you'll need to iterate over `Object.keys` for each of the spans.

Lastly, `keyword` is a special case which is just set as the text, so you would need to handle that as an exception.

kolibri/plugins/learn/assets/src/views/HybridLearningCardGrid.vue
marcellamaki: added to the issue!

kolibri/core/assets/src/views/TextTruncator.vue
marcellamaki: Added comment!

kolibri/plugins/learn/assets/src/views/HybridLearningContentCardListView.vue
marcellamaki: added explanatory comment

kolibri/plugins/learn/assets/src/views/LearnIndex.vue
marcellamaki: yes - removed

kolibri/plugins/learn/assets/src/views/ChannelCardGroupGrid.vue
marcellamaki: After giving this a look over, I think genContentLink was added by accident when I was rebasing, but `genChannelLink` is actually what's needed - so this is swapping back that mistake

kolibri/plugins/learn/assets/src/views/LibraryPage.vue
marcellamaki: yep, added a follow up https://github.com/learningequality/kolibri/issues/8560

kolibri/plugins/learn/assets/src/views/LibraryPage.vue
marcellamaki: Ah, I think I knew this from when we first talked about it, but I forgot. Thanks for the reminder - removed!

kolibri/plugins/learn/assets/src/views/HybridLearningCardGrid.vue
marcellamaki: Follow up here https://github.com/learningequality/kolibri/issues/8564

kolibri/plugins/learn/assets/src/views/LibraryPage.vue
marcellamaki: Follw up here https://github.com/learningequality/kolibri/issues/8565

marcellamaki:
Follow up for @pcenov's very diligent design alignment, as well as some other styling issues (covering tokens instead of hex, icon usage) here: https://github.com/learningequality/kolibri/issues/8563

kolibri/plugins/learn/assets/src/views/ContentCard/CardThumbnail.vue
marcellamaki: While I was starting this work, I had refactored the ContentCard component, but I realized it would cause either regressions or breaking changes in other places (namely, Coach). That work got merged as part of string freeze, and now after some refactoring to fix it (creating new components) my PR has some of this original code in it that's showing up as an "addition" in the diff. After reviewing, I am reasonably certain that all `ContentCard` references exist in places that have not been refactored as part of hybrid learning - either in the coach plugin, or in components that can probably be removed all together.  I can create a follow up issue for removing the unused components, but it seems like having this still in might be important for not having regression in coach? I'm not familiar enough with the places where swaps would happen to know if that could be done easily as part of this...

rtibbles: I don't see anywhere in the coach plugin that is importing any ContentCard components from the learn plugin, so not sure that this caution is needed, unless I'm missing something.

marcellamaki: You are right, I was not thinking about the plugins :D

marcellamaki: Actually, I remembered that the current `ChannelCard` builds on the `ContentCard` -- I'll create a follow up issue to pull the dependent styles into the `ChannelCard` and then remove various components and files that aren't being used

kolibri/plugins/learn/assets/src/views/CategorySearchModal/CategorySearchModalGrid.vue
rtibbles: Will resolve in translations follow up.

kolibri/plugins/learn/assets/src/views/ContentCard/CardThumbnail.vue
marcellamaki: Follow up https://github.com/learningequality/kolibri/issues/8566

kolibri/plugins/learn/assets/src/views/CategorySearchModal/index.vue
rtibbles: Will address here: https://github.com/learningequality/kolibri/issues/8565

kolibri/plugins/learn/assets/src/views/ContentCard/CardThumbnail.vue
rtibbles: Follow up issue here: https://github.com/learningequality/kolibri/issues/8566

kolibri/plugins/learn/assets/src/views/ContentCard/CardThumbnail.vue
rtibbles: Non-blocking.

kolibri/plugins/learn/assets/src/views/LibraryPage.vue
rtibbles: Follow up issue here: https://github.com/learningequality/kolibri/issues/8565

kolibri/plugins/learn/assets/src/views/ContentCard/CardThumbnail.vue
marcellamaki: Can follow up with this in https://github.com/learningequality/kolibri/issues/8566

kolibri/plugins/learn/assets/src/views/LibraryPage.vue
marcellamaki: Theme tokens address as task in https://github.com/learningequality/kolibri/issues/8563

kolibri/plugins/learn/assets/src/views/LibraryPage.vue
marcellamaki: Theme tokens address as task in https://github.com/learningequality/kolibri/issues/8563

rtibbles (approved):
All comments resolved - manual QA passed. Let's merge this and cross our fingers!

PR #8527, opened 2021-10-22
Add sync operation support to Kolibri plugins with migration for exam logs

bjester:
## Summary

- Adds support for defining Morango sync operations on a per plugin basis
- Adds Kolibri version information to Morango instance data shared during sync
- Establishes guidelines for deciding whether to use a sync hook or a sync operation
- Reorganizes existing sync hooks/operations to align with new guidelines
- Re-enables syncing of exams for single-user devices
- Adds new sync operation to migrate exam logs when syncing full facilities from older Kolibris
- Adds and updates Morango integration tests

## References

Resolves #8481 
Resolves #8487 

## Reviewer guidance

Note: notification generation post-sync hasn't yet been updated for new log structure
- Test full facility sync with Kolibri version < 0.15 and verify exam logs
- Single user syncing of exams and related logs

----

## Testing checklist

- [x] Contributor has fully tested the PR manually
- [ ] If there are any front-end changes, before/after screenshots are included
- [ ] Critical user journeys are covered by Gherkin stories
- [x] Critical and brittle code paths are covered by unit tests

## PR process

- [ ] PR has the correct target branch and milestone
- [x] PR has 'needs review' or 'work-in-progress' label
- [x] If PR is ready for review, a reviewer has been added. (Don't use 'Assignees')
- [ ] If this is an important user-facing change, PR or related issue has a 'changelog' label
- [ ] If this includes an internal dependency change, a link to the diff is provided

## Reviewer checklist

- Automated test coverage is satisfactory
- PR is fully functional
- PR has been tested for [accessibility regressions](http://kolibri-dev.readthedocs.io/en/develop/manual_testing.html#accessibility-a11y-testing)
- External dependency files were updated if necessary (`yarn` and `pip`)
- Documentation is updated
- Contributor is in AUTHORS.md

PR #8567, opened 2021-10-29
Examattemptlog migration in case where no examlog has been changed

rtibbles:
## Summary
* The examlog migration code assumed that an examlog would always have been when needed
* However, it is possible for examattemptlogs to have been updated without an associated examlog being updated
* This adds a test for this scenario, and adds the ability to pass an iterable of modified ids for exam attempt logs
* It then narrows down that list to those that have not otherwise been updated and then updates them
* Further, it reads touched examattemptlog ids in the morango sync hook and passes them to the utility

## References
Follow up to #8527 and #8502

## Reviewer guidance
Does the test cover the relevant cases? Is there anything we should add on the sync hook side?

----

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/core/logger/utils/exam_log_migration.py
+++ b/kolibri/core/logger/utils/exam_log_migration.py
@@ -69,6 +69,13 @@ attempt_log_fields_for_update = [
 ]
 
 
+BATCH_READ_SIZE = 750
+
+# Things that we will set as constant for every examlog we migrate
+kind = content_kinds.QUIZ
+mastery_criterion = {"type": content_kinds.QUIZ, "coach_assigned": True}
+
+
 def _update_attempt_logs(masterylog_id, logs):
     existing_logs = AttemptLog.objects.filter(masterylog_id=masterylog_id)
     existing_log_items = {log.item: log for log in existing_logs}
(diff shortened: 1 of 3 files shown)


kolibri/core/logger/utils/exam_log_migration.py
```diff
@@ -285,3 +302,31 @@ def migrate_from_exam_logs(source_logs):  # noqa C901
...
+                    .values_list("id", flat=True)
+                    .first()
+                )
+            masterylog_id = examlog_id_to_masterylog[examattemptlog.examlog_id]
+            if masterylog_id:
+                attemptlog = _create_attemptlog(examattemptlog, None, masterylog_id)
+                attempt_logs.append(attemptlog)
+
+    for masterylog_id, logs in groupby(
+        sorted(attempt_logs, key=lambda x: x.masterylog_id), lambda x: x.masterylog_id
```
bjester: This can achieve a `referenced before assignment` error for `attempt_logs` here if there are no are logs of any kind. It's defined in the batch loop but then redefined later with these new updates. It's not clear which variable this for-loop is referencing. Perhaps separate this into a separate function?

bjester (changes requested):
Function might be too long!

kolibri/core/logger/utils/exam_log_migration.py
rtibbles: Fair - can separate it out.

The function definitely is too long.

kolibri/core/logger/utils/exam_log_migration.py
```diff
@@ -108,7 +117,126 @@ def _update_attempt_logs(masterylog_id, logs):
...
+
+
+def _handle_unprocessed_attemptlog_ids(unprocessed_attempt_log_ids):
+    if unprocessed_attempt_log_ids:
+        examlog_id_to_masterylog = {}
+        unprocessed_attempt_log_ids = list(unprocessed_attempt_log_ids)
+        attempt_logs = []
+        i = 0
+        source_logs = ExamAttemptLog.objects.filter(
+            id__in=unprocessed_attempt_log_ids[i : i + BATCH_READ_SIZE]
```
bjester, line 215: It looks like this is doing batching but I think it only does one batch?

rtibbles: Yes, I apparently forgot to while here.

Review saves lives again.

rtibbles: Updated to actually loop.

bjester (approved):
:+1:

Issue #8563, opened 2021-10-29
Hybrid Learning Library, Topics pages -- design alignment follow up to-dos 


marcellamaki:
Topics Page 
- [x] Header - title alignment, better truncation, and overall spacing. 
- [x] Header - add text truncation for description 
- [x] Remove "channels" breadcrumb
- [x] Add "active" styles for folders and search tab 
- [x] Add "show more" option below trimmed channel resources 
- [x] Add "end folder view" 
- [x] Add header for folder topic above topic links
- [x] update footer: add progress, remove "more options" footer icon 
- [x] Remove breadcrumbs for search 
- [x] investigate and potentially fix breadcrumbs - some of the ordering seems incorrect
- [x] Fix "squished" view on limited search results
- [x] Remove thumbnail outline on topic thumbnail in header

Library page 
- [ ] change colors, and background color to tokens
- [x] Fix mobile card view that sometimes shows up as a grid with toggle, when should just be single card
- [x] Fix "squished" view on limited search results
- [x] Fix text overflow caused by shrinking page and long title
- [x] search: card spacing, and ensure toggle list/grid is visible 
- [ ] remove list/grid toggle if no content cards displayed

Classes page 
- [x] remove 'Library' breadcrumbs

Bookmarks page 
- [x] Card: Add text truncation to manage long description overflow and spacing issues 

Search panel
- [ ] Clear an activity or category by clicking on an active button
- [x] Fix spacing on side bar to ensure scroll to the bottom of the panel works, even on a limited folder list in the main content [ETA: this is actually because Marcella is so used to seeing the dev view, and it was not actually broken]
- [x] Activity buttons should appear on mobile

Cards 
- [ ] double-digit coach count overflows side of card
- [x] [Fix label ordering and alignment](https://github.com/learningequality/kolibri/pull/8708#pullrequestreview-808114144)

pcenov:
Hi @marcellamaki, posting here some additional issues spotted after testing [this 0.15 build](https://buildkite.com/learningequality/kolibri-python-package/builds/4003) so that you can sift through them and maybe disregard or create follow up to-dos:

Major issues:
1. Library - Clear all is not functioning now - clicking on it results in a TypeError: Cannot convert undefined or null to object

![2021-11-02_14-31-33](https://user-images.githubusercontent.com/79847249/139894499-0e069e9d-9cb4-4e39-9a60-0ab5a82c2642.png)

[...]

PR #8570, opened 2021-11-01
Search updates

### LINKED ISSUE #8555: Finalize metadata label string translations
https://github.com/learningequality/kolibri/pull/8545 consolidated logic for translating metadata labels.

To complete this we need to:
- [ ] Update kolibri-constants JS library to the latest version
- [ ] Tweak the translation for the BASIC_SKILLS subject category as it has been renamed to FOUNDATIONS to avoid name collision
- [ ] Tweak the translation for FOUNDATIONS_LOGIC_AND_CRITICAL_THINKING which was renamed from LOGIC_AND_CRITICAL_THINKING to avoid a name collision
- [ ] Update all translation of metadata labels to use simply `coreString(labelId)`

### LINKED ISSUE #8556: Bootstrap the complete representation of all available channels into the learn page using plugin data
https://github.com/learningequality/kolibri/pull/8548 bootstraps the channel thumbnail into the page to make it available for use by any contentnode within the channel.

To avoid an additional request, we should bootstrap all available channels into learn's plugin_data and use that instead of an API request.

These can then be bootstrapped into this composable: https://github.com/learningequality/kolibri/blob/release-v0.15.x/kolibri/plugins/learn/assets/src/composables/useChannels.js

### LINKED ISSUE #8557: Turn search functionality used in LibraryPage and TopicsPage into a composition helper
https://github.com/learningequality/kolibri/pull/8548 duplicates the search code used in LibraryPage for TopicsPage to provide channel scoped search.

We can abstract this into a composition helper in order to ensure consistent behaviour (as the underlying code has some complexity).

### LINKED ISSUE #8565: Fix searchTermChipList reference to query params
Reference `searchTerms` instead 
Account for 'keyword' special case

This would need to be updated on both the Library page and Topic page

- [x] Add comments to convoluted code in CategorySearchModal https://github.com/learningequality/kolibri/pull/8548/files#r738834556

rtibbles:
## Summary
* Bootstraps all available channel data into the learn page
* Uses this throughout the learn plugin to avoid API calls
* Creates a composable helper for search functionality
* Creates a SearchChips component to display and translate currently selected search terms
* Adds a route for searching within topics and uses route state for most behaviour on topics page
* Gets rid of TOPICS_CHANNEL route in favour of just using TOPICS_TOPIC
* Fixes some issues with breadcrumbs
* Makes search in topics constrained to descendants of the currently navigated to topic
* Deletes unused code that referenced the now non-existent TOPICS_ROOT and SEARCH routes
* Upgrades kolibri-constants and adds additional unhandled translations to core translator

## References
Fixes #8565
Fixes #8555
Fixes #8556
Fixes #8557
Also resolves some TODOs from #8563:
* Remove breadcrumbs for search
* Remove "channels" breadcrumb (although changes to Library - maybe should remove completely?)

## Reviewer guidance
search

----

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated


kolibri/plugins/learn/assets/src/composables/useSearch.js
```diff
@@ -0,0 +1,173 @@
...
+    set(labels, null);
+  }
+
+  function setCategory(category) {
+    set(searchTerms, { ...this.searchTerms, categories: { [category]: true } });
+  }
+
+  watch(searchTerms, search);
+
+  return {
```
MisRob, line 174: Since it's quite critical logic used from more places, would be good to have this composable's computed and methods that are publicly exposed here covered by tests, at least those that are more complex.  I realize that even without that this PR is a step towards reducing some tech debt, so it's fine to open a follow-up issue.

kolibri/plugins/learn/assets/src/views/TopicsPage.vue
```diff
@@ -519,31 +508,18 @@
...
+        this.setSearchWithinDescendant(this.topic);
       },
     },
     beforeDestroy() {
       window.removeEventListener('scroll', this.throttledHandleScroll);
     },
     created() {
+      this.translator = crossComponentTranslator(LibraryPage);
       window.addEventListener('scroll', this.throttledHandleScroll);
+      this.setSearchWithinDescendant(this.topic);
```
MisRob, line 521: Since compostables can be used from anywhere, I'd suggest also having `resetSearchWithinDescendant` function available in the `useSearch` composable that would reset [`descendant`'s value here](https://github.com/learningequality/kolibri/pull/8570/files#diff-da88993a9f71c968b198aaca81c2fe54512de05371c08f6620e1a5dea267502dR28) to `undefined` and calling the reset somewhere in the `TopicsPage` after finishing all search, maybe in the `beforeDestroy` hook? Alternatively we could just call `setSearchWithinDescendant(undefined)`.

kolibri/plugins/learn/assets/src/views/TopicsPage.vue
```diff
@@ -519,31 +508,18 @@
...
       this.search();
-      if (this.$store.getters.isUserLoggedIn) {
-        const contentNodeIds = uniq(
-          [...this.trimmedNextSteps, ...this.trimmedPopular, ...this.trimmedResume].map(
-            ({ id }) => id
-          )
-        );
-
-        if (contentNodeIds.length > 0) {
-          ContentNodeProgressResource.fetchCollection({ getParams: { ids: contentNodeIds } }).then(
```
MisRob, line 540: What was the reason for removing fetching content nodes' progress data? It seems that `HybridLearningContentCard` on `TopicsPage` is prepared to display progress and I haven't noticed an alternative API call in this PR diff.

MisRob (dismissed):
Thank you, @rtibbles. I haven't managed to understand and review all parts of this PR yet. I added some comments to parts that I went through in more detail and I also see the following issues:

**1. I can't remove search by clicking on the delete button in a search chip**

bug-chip

**2. Navigation from a channel card to a channel page on the home page stopped working**
bug-navigation

**3. Home page and classes pages content area is wider, is that change intentional?**

| Before | After |
| -------- | ------- |
| classes-page-before | classes-page-after |
| class-page-before | class-page-after |
| home-page-before | home-page-after |

kolibri/plugins/learn/assets/src/views/TopicsPage.vue
rtibbles: All the data is local to the invocation of the composable though, so this won't get reused elsewhere.

kolibri/plugins/learn/assets/src/composables/useSearch.js
rtibbles: Yes, some tests would be good, I think not unreasonable to do them within this PR.

kolibri/plugins/learn/assets/src/views/TopicsPage.vue
rtibbles: I think this may have been a half finished thought - will follow up.

rtibbles:
Thanks @MisRob:

> 1. I can't remove search by clicking on the delete button in a search chip

Looks like this was specific to key word searches only. Was unable to replicate with a non-keyword search item.

> 2. Navigation from a channel card to a channel page on the home page stopped working

This is probably related to the route simplification, will fix.

> 3. Home page and classes pages content area is wider, is that change intentional?

This was not intentional - it's the result of slightly overzealous LearnIndex cleanup. Will revert.

marcellamaki (changes requested):
I am going to do a closer code read-through (I've done a quick pass to understand the general changes). Overall the changes look good, although I appreciate that @MisRob is able to weigh in on the composables because I'm not sure I would things there. 

The one bug I've noticed in manual QA is that the options for the filter options persist the previous state after clearing the search:
filters-persist

Will update with additional questions/comments this afternoon after a closer code review.

rtibbles:
Ah yes - good catch, can clear labels too on search change.

kolibri/plugins/learn/assets/src/views/TopicsPage.vue
MisRob: Ah, that's right, it's in the function, not outside of it. I didn't realize that when writing the comment, you can ignore it then.

kolibri/plugins/learn/assets/src/views/TopicsPage.vue
rtibbles: Ah yes, this was intentional. The content nodes for which progress is being fetched here are not rendered anywhere on this page, this seemed to have been copy pasted from the `created` implementation in the `LibraryPage.vue` so was a superfluous request.

It does mean we still need to load progress for the items on the topics page, but this change is not a regression.

kolibri/plugins/learn/assets/src/modules/topicsTree/handlers.js
```diff
@@ -1,37 +1,27 @@
...
 
-  const promises = [
-    ContentNodeResource.fetchModel({ id }),
-    ContentNodeResource.fetchNextContent(id),
-    store.dispatch('setChannelInfo'),
-  ];
-  ConditionalPromise.all(promises).only(
+  ContentNodeResource.fetchModel({ id }).only(
     samePageCheckGenerator(store),
-    ([content, nextContent]) => {
```
marcellamaki, line 27: I can't totally tell if this is a refactor, or a removal of "next content". Does any of @nucleogenesis 's side panel work depend on this for the topics/lesson overview side panel?

rtibbles: My recollection was that this was being independently fetched in the side panel, and so it was now superfluous here.

I think that is still correct, as I see the recommendations being loaded here: https://github.com/learningequality/kolibri/blob/release-v0.15.x/kolibri/plugins/learn/assets/src/views/BrowseResourceMetadata.vue#L298

kolibri/plugins/learn/assets/src/views/SearchChips.vue
```diff
@@ -0,0 +1,132 @@
...
+              key,
+              text: this.translate(key, val),
+            };
+          });
+        });
+      },
+    },
+    created() {
+      const LibraryPageComponent = require('./LibraryPage').default;
+      this.translator = crossComponentTranslator(LibraryPageComponent);
```
marcellamaki, line 72: This is new!

rtibbles: Yes, there were strings required to display the search chips that did not exist in the TopicsPage component, only in the LibraryPage component - but because we import the SearchChips into the LibraryPage, I didn't want to setup a circular import, so did the import here inside the closure.

rtibbles: And also, I hate myself for using crossComponentTranslator.

kolibri/plugins/learn/assets/src/routes/index.js
```diff
@@ -99,23 +89,23 @@ export default [
     },
   },
   {
-    name: PageNames.TOPICS_CHANNEL,
-    path: '/topics/:channel_id',
```
marcellamaki, line 103: This feels like a notable change. Not blocking, and not a technical comment, but I'm wondering if this might require some toolkit or other updates for public channels to places like on the demo server, etc? I'm not sure how many places we have links to specific channel resources but thought that might be worth flagging

rtibbles: I think it would probably be a good idea to retain the route, and instead do a redirect to the new page, in the interests of the above.

kolibri/plugins/learn/assets/src/views/Breadcrumbs.vue
```diff
@@ -48,67 +58,25 @@
...
-        return {};
-      },
       topicsBreadcrumbs() {
         return [
           // All Channels Link
           {
-            text: this.coreString('channelsLabel'),
-            link: { name: PageNames.TOPICS_ROOT },
+            text: this.learnString('libraryLabel'),
+            link: { name: PageNames.LIBRARY },
```
marcellamaki: Breadcrumbs in the topics page don't need to start until the user is down at least one folder level. No reference to Library page is needed, just starting with the return to the highest topic (channel) folder. Which I only know because Peter already provided me this feedback 😄 
Screen Shot 2021-11-01 at 3 38 55 PM

kolibri/plugins/learn/assets/src/routes/index.js
rtibbles: Have added the redirect.

kolibri/plugins/learn/assets/src/views/Breadcrumbs.vue
rtibbles: Have updated to remove the Library from the breadcrumbs.

rtibbles:
Resolved review feedback, except for adding tests for the search composable.

kolibri/plugins/learn/assets/src/composables/useSearch.js
rtibbles: Have now added tests for every branch in here, I think.

rtibbles:
Tests for search composable have now been added.

kolibri/plugins/learn/assets/src/composables/__tests__/useSearch.spec.js
```diff
@@ -0,0 +1,460 @@
...
+import { AllCategories, NoCategories } from 'kolibri.coreVue.vuex.constants';
+import useSearch from '../useSearch';
+import { normalizeContentNode } from '../../modules/coreLearn/utils';
+
+Vue.use(VueRouter);
+
+const name = 'not important';
+
+function prep(query = {}) {
+  const store = coreStoreFactory({
```
MisRob, line 15: No action is needed. Just wanted to note that importing the whole store machinery can often bring quite a lot of complexity behind scenes and am not sure if it's worth it only to mock `state.route`, especially in a composable that is otherwise quite independent of the rest of the store. Was there any particular reason for bringing the whole core module into tests when compared to using an empty mock store or router and defining only values and methods that are needed?

rtibbles: I did this because it's more representative of what is actually happening in production, and I didn't want any weird surprises to happen there compared to the tests.

Ideally, I would have just used the [VueRouter composable functions](https://next.router.vuejs.org/guide/advanced/composition-api.html) in here to access the route and the router, and not had the complexity inside or outside the tests, but that requires upgrading to VueRouter 4.x, and that seemed like a step too far for now!

MisRob: > I did this because it's more representative of what is actually happening in production, and I didn't want any weird surprises to happen there compared to the tests

Thank you for the explanation, makes sense. I usually find balancing these things quite tricky. It sounds that this approach has the advantage of being more representative as you say.

MisRob: No need to upgrade Vue Router for this :smile:

marcellamaki:
Seems like feedback has been addressed, although I know @jtamiace sent some feedback about the homepage -- not sure if that will be included in this PR or not. Otherwise, the only thing I see is this "extra chip" popping up. Seems like one empty one is appended to the end of the list of filters. 

Screen Shot 2021-11-02 at 10 40 38 AM

rtibbles:
> not sure if that will be included in this PR or not

I think best to do as a follow up.

The extra chip is probably coming from an empty `keywords` parameter, will update.

rtibbles:
> The extra chip is probably coming from an empty keywords parameter, will update.

Replicated and updated.

nucleogenesis (commented):
Code look good. I really love seeing the composition API stuff in action and seeing it mixed with Vuex stuff - love it. I didn't see anything that jumped out at me. 

I found a bug in the keywords processing in Python which is that it fails with the following queries, separated by line:

```
10
100
hello 10
hello 100
10 hello
10 x 10
10 10 10
100 10 100
or
and or
```

These are all unlikely queries, of course, but they may signal that there are more that I'm not finding. I wouldn't block merging on this but it's worth following up on, perhaps after bug bash to see if people run into any more normal queries that return 500.

The server gives a message from Django that it [needs to be valid JSON (traceback)](https://gist.github.com/nucleogenesis/d0a8ac74658ee531df8312170c4ae8a3).

In any case - the fail state there when it gets the 500 is to show a spinner forever so that could use some better error handling for the UX - maybe just showing 0 results along with some existing "we encountered an error" strings. The user can return to normal by clearing the filters that appear once you click something in the side bar.

nucleogenesis (commented):
Oh just noticed this as I was wrapping up but the X shows in the search bar when you haven't searched any kind of text query.

So - I click "Read" and the X shows in both searches (top bar and left nav). 

When I click it, nothing happens. This is actually a bug altogether - even when I type text in and click the X to clear the search, it doesn't have any effect on the chips or query results.

If this isn't in scope I can make sure to create a follow up issue.

rtibbles:
Fixed the 500 in #8579

Have fixed the search box `X` issue:

Screenshot from 2021-11-02 14-39-45

Screenshot from 2021-11-02 14-39-50

marcellamaki (approved):

Issue #8577, opened 2021-11-02
Add lesson_id context when initializing logging session in ContentPage


rtibbles:
## Expected behavior
* The progress tracking API allows setting a `lesson_id` in addition to the `node_id` in order to set a more detailed context for the session
* This should be set within the ContentPage, possibly in conjunction with setting the appropriate `back` arrow behaviour as described in https://github.com/learningequality/kolibri/issues/8512
* If the lesson_id is encoded in a query parameter, then it can be used to set the lessonId computed prop https://github.com/learningequality/kolibri/blob/release-v0.15.x/kolibri/plugins/learn/assets/src/views/ContentPage.vue#L141 (instead of the current URL params that will never be defined).

## User-facing consequences
None at the moment, but this would give us better insight into the use of the lesson functionality in synced data.

rtibbles:
Fixed in #8588

PR #8588, opened 2021-11-03
Fix back button in LearnImmersiveLayout

### LINKED ISSUE #8512: Back button in LearnImmersiveLayout is Broken
The back button in the learn immersive layout page is not working as expected 

When the previous page is the library page, the back button returns to the topics browser page

From other pages, the back button either does nothing or results in an `Avoided redundant navigation to current location` error

marcellamaki:
## Summary
When leaving content renderer, back button should correctly revert to home, lesson, library, topic, or bookmarks page. 

## References
Fixes #8512

## Reviewer guidance
I tried to use `last` and `prevContext` to account for both the previous page name and the previous page params, rather than having many conditions trying to match up with page names, but I'm not sure if this is more confusing, since it's less explicit.

Any scenarios this doesn't account for?

----

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/plugins/learn/assets/src/utils/genContentLink.js
+++ b/kolibri/plugins/learn/assets/src/utils/genContentLink.js
@@ -1,8 +1,10 @@
 import { PageNames } from '../constants';
 
-export default function genContentLink(id, isLeaf) {
+// previous context allows to pass in ids or params from the last route
+export default function genContentLink(id, isLeaf, last, prevContext) {
   return {
     name: isLeaf ? PageNames.TOPICS_CONTENT : PageNames.TOPICS_TOPIC,
     params: { id },
+    query: { last: last, prevContext },
   };
 }
--- a/kolibri/plugins/learn/assets/src/views/HybridLearningCardGrid.vue
+++ b/kolibri/plugins/learn/assets/src/views/HybridLearningCardGrid.vue
@@ -18,7 +18,7 @@
           :isLeaf="content.is_leaf"
           :progress="content.progress_fraction || 0"
           :numCoachContents="content.num_coach_contents"
-          :link="genContentLink(content.id, content.is_leaf, backRoute)"
+          :link="genContentLink(content.id, content.is_leaf, backRoute, context)"
           :contentId="content.content_id"
           :copiesCount="content.copies_count"
           :description="content.description"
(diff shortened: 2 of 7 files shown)


kolibri/plugins/learn/assets/src/views/HybridLearningCardGrid.vue
```diff
@@ -79,7 +79,7 @@
       :isLeaf="content.is_leaf"
       :progress="content.progress_fraction || 0"
       :numCoachContents="content.num_coach_contents"
-      :link="genContentLink(content.id, content.is_leaf)"
+      :link="genContentLink(content.id, content.is_leaf, backRoute)"
```
rtibbles: Why is the `genContentLink` above passed the context, but this one isn't?

kolibri/plugins/learn/assets/src/utils/genContentLink.js
```diff
@@ -1,8 +1,10 @@
 import { PageNames } from '../constants';
 
-export default function genContentLink(id, isLeaf) {
+// previous context allows to pass in ids or params from the last route
+export default function genContentLink(id, isLeaf, last, prevContext) {
   return {
     name: isLeaf ? PageNames.TOPICS_CONTENT : PageNames.TOPICS_TOPIC,
     params: { id },
+    query: { last: last, prevContext },
```
rtibbles, line 8: How does the `prevContext` object get encoded into the URL? I assume it's fine?

rtibbles (commented):
Haven't tested, and while I share your aesthetic disregard for the lack of specificity, this seems like the simplest way for now.

One argument for making it more explicit would be for resolving this: https://github.com/learningequality/kolibri/issues/8577 so we have a canonical source of truth on the content page about the context it was loaded from.

kolibri/plugins/learn/assets/src/utils/genContentLink.js
marcellamaki: That's an oversight -- I've now added encoding

kolibri/plugins/learn/assets/src/views/HybridLearningCardGrid.vue
marcellamaki: It's an optional param. i.e. library page does not need it, but since the lessons cards _do_ since their route requires a specific classid and lessonid, it is passed there. Maybe I can name it something clearer...

rtibbles: I think would be good to just always pass it for consistency - means that we won't get bitten in the behind in future when we decide we need to pass context from the library page (like the search term that someone arrived at the content from) and we can't understand why the context is missing.

rtibbles (approved):

PR #8592, opened 2021-11-04
Resolve missing kind preventing successful sync of quiz logs and quieter sync logging

### LINKED ISSUE #8394: noisy syncing logs
## Observed behavior

when devices are auto-syncing, both the source and target device get a lot of repetitive information written to the logs:

image

## Expected behavior

If possible, these should be made quieter - maybe one line for each attempted sync?

## User-facing consequences

This could make it more difficult for users (and us) to find important information like errors in the logs, and could also cause log file sizes to grow very large

## Context

latest `develop`

bjester:
## Summary

- Updates logging API to set `kind` for `ContentSessionLog`
- Adds minimum length validator to some non-blank char fields to ensure they're not blank
- Reduces log messaging originating from sync

## References

Addresses @pcenov's [comment](https://github.com/learningequality/kolibri/pull/8527#issuecomment-956380573)
Resolves https://github.com/learningequality/kolibri/issues/8394

## Reviewer guidance

…

----

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated


rtibbles (approved):
Code changes look good - seems like Django is not entirely convinced that the migration for the logger app exists? Possible it's some weird Py2/3 discrepancy, so maybe try rerunning makemigrations with Python 2 and see if anything different pops up?

bjester:
@rtibbles I didn't think it needed a migration, and thought I would see a warning running tests if it did, so I didn't initially commit a migration :+1:

rtibbles:
Ah - gotcha! Yeah, Django likes to add a migration for any model change, regardless of whether it actually affects the DB or not.

pcenov:
Hi @bjester, while testing this I noticed that only the first part of my comment is fixed with this PR namely a Coach is now able to see the scores of a learner and any questions answered.
At the same time my second point is still valid:  I imported a full 0.14 facility to a 0.15 one and synced it. When I go to check the scores of a learner as a Coach the page with the quiz scores is not displayed and there's an error in the console: `Uncaught (in promise) TypeError: currentAttempt.interaction_history.filter is not a function`:

rtibbles:
@bjester - weird that interaction history is not an array, but perhaps we can coerce that in the API request, in the case that it is falsy for some reason?

bjester:
Thanks for testing @pcenov! What instance was the quiz completed on? And what instance did you check the result as the Coach? I should have noted in the prior PR that we didn't make the quiz results compatible with older versions, so if the quiz was completed on 0.15 then I wouldn't expect it to show up correctly in 0.14, but the other way around should work appropriately

pcenov:
The quiz was completed on the 14.7 instance which was then imported in 0.15 where I was subsequently checking the results. So yeah, it seems that there is an issue indeed.

bjester:
Okay yes that sounds like a bug. I will investigate.

Issue #8554, opened 2021-10-29
Update API for difficult questions in practice quizzes


sairina:
## Observed behavior
Currently, to access a difficult questions report, there are only two options for viewsets - one for an exercise and one for a quiz. However, for practice quizzes, we need some new functionality that is not present in either viewset.

|Frame|Notes|
|--|--|
|![Screen Shot 2021-10-28 at 9 29 13 PM](https://user-images.githubusercontent.com/13563002/139375806-e9814259-7fea-4650-b4f6-4733bb6723d4.png)|the practice quiz difficult questions report|
|![Screen Shot 2021-10-28 at 9 32 52 PM](https://user-images.githubusercontent.com/13563002/139376120-33d376da-c9ba-41a2-8046-afb07d01a3d3.png)|the possible responses that show up as a result of comparing the question with the learner's previous attempt|

## Expected behavior
We need additional functionality in our viewsets to address the ability to add:
- [x] restrict the returned attempts to only attempts for the most recently completed try for a user
- [x] comparatively, how the learner did on the question this time compared to their previous attempt - this could either be achieved by annotation on the attempt, or by fetching the other attemptlog from the penultimate attempt

## User-facing consequences
Without this, coaches will not be able to utilize the full functionality and purpose for having a student take a practice quiz several times.

## Steps to reproduce
1. Sign in as a coach
2. Click on the reports tab
3. Click on the lessons subtab. It should open lesson list page
4. Making sure you are in the report subtab here, click on the title for a resource that is an independent practice quiz
5. Click on the difficult questions subtab
6. Click on the question - this should take you to the difficult question report view.

bjester:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Kolibri-Ecosystem-6047e525b8b9d900182e2d6f/issues/learningequality/kolibri/8554?planning-poker) with ZenHub @marcellamaki @sairina @jredrejo @rtibbles

bjester:
Intersects a bit with API for reporting on practice quizzes which also has question diff

rtibbles:
Both requirements here were fulfilled by a combination of #8740 and #8645

PR #8634, opened 2021-11-08
Listing of practice quiz difficult questions

rtibbles:
## Summary
* Adds an endpoint for providing listings of question correctness for practice quizzes
* Adds testing to ensure only the most recent completed tries of a quiz are used to calculate results
* Adds initial integration inside the `questionList` module with a 'practiceQuiz' boolean flag to disambiguate an exercise from a practice quiz.

## References
Necessary precursor to #8554 but does not implement the fix detailed therein.

## Reviewer guidance
Does the API endpoint provide the required unique functionality for returning a list of all the questions that have been answered in completed tries, and return a summary of their correct responses and total responses?

Does the integration into the existing questionList module make sense? Should we use something other than a boolean flag to indicate the practiceQuiz?

----

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/plugins/coach/api.py
+++ b/kolibri/plugins/coach/api.py
@@ -2,6 +2,7 @@ import datetime
 
 from django.db import connections
 from django.db.models import Count
+from django.db.models import F
 from django.db.models import OuterRef
 from django.db.models import Subquery
 from django.db.models import Sum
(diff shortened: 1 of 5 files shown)


rtibbles:
Seems like my code does not work correctly on postgres, will update.

kolibri/plugins/coach/api.py
```diff
@@ -408,3 +410,69 @@ def retrieve(self, request, pk):
...
+            collection_id = group_id or classroom_id
+            masterylog_queryset = masterylog_queryset.filter(
+                user__memberships__collection_id=collection_id
+            )
+
+        masterylog_queryset = masterylog_queryset.filter(
+            id__in=Subquery(
+                MasteryLog.objects.all()
+                .order_by("-completion_timestamp")
+                .filter(user_id=OuterRef("user_id"), summarylog__content_id=pk)
```
bjester: Needs `masterylog__complete=True`? I forgot that on a similar query in my API work

rtibbles: Yes, I remembered it on the outer queryset, but it's needed here to.

Will also be updating the ordering for consistent null sorting, as I think that's the cause of the postgres failure here.

rtibbles: Added a regression test for this and updated.

bjester (approved):
Looks good to me

PR #8650, opened 2021-11-10
Consolidate Learn data loading

### LINKED ISSUE #8068: Create a prototype hydration viewset base class that allows multiple viewsets to be retrieved simultaneously
An extension of the class_summary_api used in the Coach plugin, we will make a viewset that allows amalgamating the output of multiple viewsets in order to do a smaller number of fetches for page hydration.

Open questions:
* Will it be helpful to be able to requery the endpoint specifying which fields/sub paths to return, or will just querying the separate endpoints be sufficient?
* How should we specify dependencies and handle efficient caching?

### LINKED ISSUE #8604: Make a dedicated endpoint for hydrating the learn page, that will include all recent items in lessons
## Observed behavior
Currently we load the resume endpoint and cross reference with the loaded lessons to find recent content. The resume endpoint only loads 10 nodes, so it is easy for content to get missed.

## Expected behavior
All resumable content in lessons should display.

We should also refresh this whenever a use navigates back to the home page or classes page to ensure we have up to date data.

Caveat: Pagination may be required.

### LINKED ISSUE #8605: Ensure all recent nodes can be displayed on the LibraryPage
## Observed behavior
Currently nodes that have been recently engaged with may not show up on the Recent section of the library page
This may either be because we are doing a non-forced fetch to the backend, or the new node is not in the ten that get returned, or both!

## Expected behavior
Resume should return a paginated list and the frontend should force the fetch.
We should also allow the user to View More when there are more available.

### LINKED ISSUE #8607: Loading ContentNodeProgress data when loading more topics causes errors
## Observed behavior
When clicking 'view more' for loading additional topics, a TypeError occurs from the call to fetch ContentNodeProgress.

In addition, this call seems to happen even if not logged in.

## Expected behavior
No errors. Don't make the call if not logged in.

### LINKED ISSUE #8609: Conditionalize "Classes"
Classes tab and "your classes" section on home page should only appear when learner is enrolled in classes

### LINKED ISSUE #8626: Learn breadcrumbs seem to show up on wrong page
## Observed behavior

(this may be tracked in https://github.com/learningequality/kolibri/issues/8563 - if so feel free to close as a duplicate)

It appears that the breadcrumb is showing up on the Home page (where it probably shouldn't) and not on the Library page (where it probably should):

| Home |  Library |
|--|--|
|image |image |

Further, on the home page the `KBreadcrumbs` component is getting invalid data passed to its `items` prop:

```
Invalid prop: custom validator check failed for prop "items".
```

The `items` array triggering this message seems to have two objects as elements, the second of which is missing a required `text` attribute. This is probably why it's showing "one and a half" breadcrumbs:

> image

## Expected behavior

* breadcrumbs should match the current page
* there should never be a trailing `>` on breadcrumbs without a final 'current page' indicator

## User-facing consequences

confusing navigation

## Steps to reproduce

view 'home' and 'library' pages

## Context

0.15 beta 1

rtibbles:
## Summary
* Consolidates and simplifies data loaded at Learn app initialization
* Creates a new utility to load data from viewsets to allow for multiple data sources to be loaded from one API endpoint
* Updates state management and loading for the home page
* Inlines content node loading for the learner classroom viewset
* Fixes display of resumable content, and adds pagination in all displays
* Hides classes section when not available
* Hides unnecessary breadcrumbs

## References
Fixes #8605
Fixes #8626
Fixes #8609 
Fixes #8607 
Fixes #8604 
Fixes #8068

## Reviewer guidance
Check fixes for all of the above.
Are the code changes comprehensible? Have I put the state in the right place? Should I be putting the data and resource layer manipulation elsewhere?

I have not updated tests yet.

----

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated


marcellamaki:
Overall, when reviewing, I found it fairly readable, although I'm not sure I am the best person to comment on either the python or the use of composables. On the QA side, it also seems to be working. 

My main point of confusion was around this -- after reading through the code, my initial understanding was that the `resumable_resources` would be used to set _all_ of the "continue learning" section. However, when I load the homepage and the see the network request, that is an empty array, with the data coming from classes. My guess is that this is actually the expected behavior, and that there _is_ a distinction between continuing from your classes vs. on your own (`uniqueResumableClassesResources` vs. `resumableContentNodes` ?) but just want to confirm that I am understanding this correctly... 

Screen Shot 2021-11-10 at 11 01 12 AM

rtibbles:
Yes - the logic in the `ContinueLearning` component is only to show non-classes resumable content nodes when there is nothing to show for classes resumable content - so in the backend we do this check and only return the 'resumable content nodes' data in the case that it is needed.

rtibbles:
Anything resumable from within classes, we already know about because the progress is annotated on the resources within lessons (I added this, but we were already fetching this progress data previously, but were only annotating it at the lesson level).

marcellamaki (approved):
With clarifying questions answered, looks good to me.

PR #8656, opened 2021-11-10
Update `pre-commit` config python version to 3.6

sairina:
## Summary
Change the python `lang_version` from 2.7 to 3.6 for the pre-commit file. We already have python 2.7-specific tests that should catch any issues that arise from this, so hopefully reordering imports won't cause errors?

## Testing checklist
- [x] Contributor has fully tested the PR manually (as far as I know, I am now able to use the `pre-commit` hook)

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -14,7 +14,7 @@ repos:
     rev: v1.3.3
     hooks:
     -   id: reorder-python-imports
-        language_version: python2.7
+        language_version: python3.6
 -   repo: local
     hooks:
     -   id: lint-frontend
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -14,7 +14,7 @@ repos:
     rev: v1.3.3
     hooks:
     -   id: reorder-python-imports
-        language_version: python3.6
+        language_version: python3
 -   repo: local
     hooks:
     -   id: lint-frontend
--- a/kolibri/core/content/management/commands/importcontent.py
+++ b/kolibri/core/content/management/commands/importcontent.py
@@ -1,7 +1,7 @@
+import concurrent.futures
 import logging
 import os
 
-import concurrent.futures
 import requests
 from django.core.management.base import CommandError
 from le_utils.constants import content_kinds
--- a/kolibri/core/tasks/compat.py
+++ b/kolibri/core/tasks/compat.py
@@ -1,6 +1,5 @@
 import multiprocessing
 import threading
-
 from concurrent import futures
 
 from kolibri.utils.conf import OPTIONS
--- a/kolibri/core/tasks/worker.py
+++ b/kolibri/core/tasks/worker.py
@@ -1,6 +1,5 @@
 import logging
 import traceback
-
 from concurrent.futures import CancelledError
 
 from kolibri.core.tasks.compat import PoolExecutor
--- a/kolibri/utils/debian_check.py
+++ b/kolibri/utils/debian_check.py
@@ -1,9 +1,8 @@
 import getpass
 import os
 import sys
-from shutil import rmtree
-
 from builtins import input
+from shutil import rmtree
 
 from .conf import KOLIBRI_HOME
 from .server import installation_type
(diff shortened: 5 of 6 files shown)


.pre-commit-config.yaml
```diff
@@ -14,7 +14,7 @@ repos:
     rev: v1.3.3
     hooks:
     -   id: reorder-python-imports
-        language_version: python2.7
+        language_version: python3.6
```
indirectlylit: After trying this, I now get an error:

```bash
.venv ❯ pre-commit run --all-files
[INFO] Installing environment for https://github.com/asottile/reorder_python_imports.
[INFO] Once installed this environment will be reused.
[INFO] This may take a few minutes...
An unexpected error has occurred: CalledProcessError: Command: ('/Users/d/Projects/le/kolibri/.venv/bin/python', '-mvirtualenv', '/Users/d/.cache/pre-commit/repojskl9o3e/py_env-python3.6', '-p', 'python3.6')
Return code: 1
Expected return code: 0
Output: 
    RuntimeError: failed to find interpreter for Builtin discover of python_spec='python3.6'
    
Errors: (none)

Check the log at /Users/d/.cache/pre-commit/pre-commit.log

~/Projects/le/kolibri update-precommit 21s
```

The same error is appearing in Github actions:

https://github.com/learningequality/kolibri/runs/4172291713?check_suite_focus=true

My virtual environment uses Python 3.8 which could be the reason:

```bash
.venv ❯ python --version          
Python 3.8.2
```

I think this line can actually be deleted entirely, and it should just use whatever environment is available.

indirectlylit (changes requested):
We might be able to avoid specifying any version of Python for the import reorder hook to run in.

Also, we may want to target this to 0.15 to help avoid merge conflicts

.pre-commit-config.yaml
rtibbles: The only concern with that is if different language versions produce different import orderings (e.g. changing to Python 3.6 caused some imports to be reordered).

If you use Python 3.8 does it reorder any imports?

indirectlylit: Switching to 3.8 retains the updated ordering in this PR, so I imagine it's a 2/3 inconsistency.

Assuming that's the case, another option is just the `python3` alias:

```suggestion
        language_version: python3
```

rtibbles: That should work - I think that _should_ also fix the Github action error.

indirectlylit (approved):
nice, tests are passing.

Not required, but targeting to 0.15 would keep the CI environments consistent and may reduce the chance of future merge conflicts. (Also it would allow you to run `pre-commit` in both branches!)

sairina:
> it would allow you to run pre-commit in both branches!

Explain that wizardry. I currently am working off develop and need to start commiting things (I have 2 new branches from develop just hanging out locally), so how would targeting to 0.15 allow me to do that right now? Does that mean I would have to merge into 0.15 and then merge again into develop?

indirectlylit:
Yeah, 0.15 can and should get merged to develop on a regular basis. In general, any change that we want in multiple branches should be committed to the "oldest" applicable release branch so that merges always go from older to newer code and never back the other direction.

In this case I would personally:

* target this change to 0.15
* we'll merge it and then immediately merge 0.15 -> develop
* then you can rebase your develop work on the latest

indirectlylit:
> merges always go from older to newer code and never back the other direction

For additional context, the reason for this is that going the other direction usually triggers a very annoying merge conflict on subsequent 'waterfall' merges. For example:

* make a change in develop
* decide we want that change in 0.15 also and commit it there
* then the next time we merge 0.15 -> develop, the two differently hashed but otherwise identical changes conflict with each other

rtibbles:
I'd be happy to have this rebased to release-v0.15.x and merged there, don't want @sairina to be permanently barred from making bugfixes for 0.15.x!

indirectlylit:
thank you! opened a PR to merge 0.15 -> develop here: https://github.com/learningequality/kolibri/pull/8681

PR #8659, opened 2021-11-11
Store all currently displayed locations, not just the start location for EPUB progress

### LINKED ISSUE #8309: Investigate potential bug with progress/completion

rtibbles:
## Summary
* Stores all the locations currently being displayed on the screen, rather than just the starting location
* This means that all displayed locations count as visited
* Allows EPUBs to be completed
* Fixes display of Next and Previous buttons for RTL books

## References
Fixes #8309

## Reviewer guidance
If you read every page of an EPUB does it get marked as complete?

RTL reading still seems not to work, but I can't narrow down if it is due to our implementation of EPUB.js or the library itself.

----

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/plugins/epub_viewer/assets/src/views/EpubRendererIndex.vue
+++ b/kolibri/plugins/epub_viewer/assets/src/views/EpubRendererIndex.vue
@@ -101,7 +101,16 @@
         <div
           class="column epubjs-navigation"
         >
+          <NextButton
+            v-if="contentIsRtl"
+            v-show="!isAtEnd"
+            :color="navigationButtonColor"
+            :isRtl="contentIsRtl"
+            :style="{ backgroundColor }"
+            @goToNextPage="goToNextPage"
+          />
           <PreviousButton
+            v-else
             v-show="!isAtStart"
             :color="navigationButtonColor"
             :isRtl="contentIsRtl"
@@ -118,7 +127,16 @@
         <div
           class="column epubjs-navigation"
         >
+          <PreviousButton
+            v-if="contentIsRtl"
+            v-show="!isAtStart"
+            :color="navigationButtonColor"
+            :isRtl="contentIsRtl"
+            :style="{ backgroundColor }"
+            @goToPreviousPage="goToPreviousPage"
+          />
           <NextButton
+            v-else
             v-show="!isAtEnd"
             :color="navigationButtonColor"
             :isRtl="contentIsRtl"
--- a/kolibri/plugins/epub_viewer/assets/src/views/EpubRendererIndex.vue
+++ b/kolibri/plugins/epub_viewer/assets/src/views/EpubRendererIndex.vue
@@ -493,9 +493,11 @@
         }
       },
       storeVisitedPage(currentLocation) {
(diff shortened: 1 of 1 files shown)


sairina:
With an RTL epub, the short answer is, "no" - it does not get marked complete. However:
- The slider only goes up to 84%
- Any manual movement of the slider makes it so that the bar is fully done, but the number is still at "84%"
- The buttons (left and right) on the epub don't actually work once the slider is moved beyond the "84%

Screen Shot 2021-11-11 at 1 15 43 PM
- Here, on the Home page, the progress is actually (even though it was completely read and is at 84% in the renderer itself) quite low.
Screen Shot 2021-11-11 at 1 17 03 PM

marcellamaki (approved):
This looks good and I was able to see the completion modal pop up at the end of an EPUB.

kolibri/plugins/epub_viewer/assets/src/views/EpubRendererIndex.vue
```diff
@@ -118,7 +127,16 @@
...
+          <PreviousButton
+            v-if="contentIsRtl"
+            v-show="!isAtStart"
+            :color="navigationButtonColor"
+            :isRtl="contentIsRtl"
+            :style="{ backgroundColor }"
+            @goToPreviousPage="goToPreviousPage"
+          />
           <NextButton
+            v-else
```
indirectlylit, line 139: Non-blocking comment:

One subtle issue with this strategy is that it reverses tab order between RTL and LTR, which is generally not supposed to happen because tab order should be independent of text direction.

In the quiz page we work around this for the 'Next/Previous' buttons by dynamically changing the text direction `dir`:

https://github.com/learningequality/kolibri/blob/f6318032221d72c9e9e54d4f41285b357299acff/kolibri/plugins/learn/assets/src/views/ExamPage/index.vue#L46

https://github.com/learningequality/kolibri/blob/f6318032221d72c9e9e54d4f41285b357299acff/kolibri/plugins/learn/assets/src/views/ExamPage/index.vue#L271-L279

Another nice effect here is that the more important 'next' button gets focus _before_ the 'previous' button even though it visually appears "second".

| Visual / reading order | Tab focus order |
|--|--|
| Quiz questions, Previous, Next | Quiz questions, Next, Previous |

indirectlylit (commented):
comment on subtle text direction + tab order issue

kolibri/plugins/epub_viewer/assets/src/views/EpubRendererIndex.vue
rtibbles: True - would probably be a better fix. There is still some overall breakage of RTL epubs at the moment, so can fix this during that follow up.

PR #8660, opened 2021-11-11
Back navigation

### LINKED ISSUE #8614: When referred from the topics page, make sure to generate the back link using the parent of the content item
## Observed behavior
When navigating to the content page from the topics page, the generated backlink to the topics page uses the id of the node being displayed on the content page, and not the id of its parent.

Because it is not a topic, this causes navigation to appear to fail.

## Expected behavior
Back navigation should work.

marcellamaki:
## Summary
Updates genContentLink to take a topicId, rather than conditionally referring to the parent in Topic/TopicSearch page 
Fixes the issue with topic page back button appearing to break, and also ensures that the back button takes the user to their previous view, not just the parent (which can be disorienting, due to nested folder display)

## References
Fixes #8614

topics-back-button

…

## Reviewer guidance
Do all of the back links still work? 
Any edge cases not considered or potentially breaking?

----

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/plugins/learn/assets/src/views/HybridLearningCardGrid.vue
+++ b/kolibri/plugins/learn/assets/src/views/HybridLearningCardGrid.vue
@@ -12,7 +12,7 @@
           :isMobile="windowIsSmall"
           :content="content"
           :thumbnail="content.thumbnail"
-          :link="genContentLink(content.id, content.is_leaf, backRoute, context)"
+          :link="genContentLink(content.id, topicId, content.is_leaf, backRoute, context)"
           @openCopiesModal="openCopiesModal"
           @toggleInfoPanel="$emit('toggleInfoPanel', content)"
         />
@@ -26,7 +26,7 @@
         :thumbnail="content.thumbnail || getContentNodeThumbnail(content)"
         class="card-grid-item"
         :isMobile="windowIsSmall"
-        :link="genContentLink(content.id, content.is_leaf, backRoute, context)"
+        :link="genContentLink(content.id, topicId, content.is_leaf, backRoute, context)"
       />
     </div>
     <CardGrid
@@ -37,7 +37,7 @@
 
         :key="`resource-${idx}`"
         :contentNode="content"
-        :to="genContentLink(content.id, content.is_leaf, backRoute, context)"
+        :to="genContentLink(content.id, topicId, content.is_leaf, backRoute, context)"
       />
     </CardGrid>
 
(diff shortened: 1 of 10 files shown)


kolibri/plugins/learn/assets/src/views/HybridLearningCardGrid.vue
```diff
@@ -12,7 +12,7 @@
           :isMobile="windowIsSmall"
           :content="content"
           :thumbnail="content.thumbnail"
-          :link="genContentLink(content.id, content.is_leaf, backRoute, context)"
+          :link="genContentLink(content.id, topicId, content.is_leaf, backRoute, context)"
```
indirectlylit, line 15: interesting – did the API change? I found a couple other instances in the codebase that seem to lack a `topicId`:


https://github.com/learningequality/kolibri/blob/802f02eabecd24f99085add228b15dfd06835eac/kolibri/plugins/learn/assets/src/views/CompletionModal/index.vue#L115-L120

https://github.com/learningequality/kolibri/blob/802f02eabecd24f99085add228b15dfd06835eac/kolibri/plugins/learn/assets/src/views/CompletionModal/index.vue#L262-L267

Any chance there are others?

One strategy to avoid regressions would be to give default values to parameters in `genContentLink`, e.g.

```js
export default function genContentLink(
  id,
  topicId = null,
  isLeaf = false,
  last = null,
  prevContext = {}
) {
  const query = { ...prevContext };
  if (last) {
    query.last = last;
  }
  if (topicId) {
    query.topicId = topicId;
  }
  return {
    name: isLeaf ? PageNames.TOPICS_CONTENT : PageNames.TOPICS_TOPIC,
    params: { id },
    query,
  };
}
```


Another option is to have `genContentLink` take an options object with all optional parameters packed in there:

```js
export default function genContentLink(id, options = {}) {
  // ...
```

indirectlylit (commented):
confirmed that the 'back' link works in the case reported by the issue

kolibri/plugins/learn/assets/src/views/HybridLearningCardGrid.vue
marcellamaki: Good catch - I've update with the defaults and I've also gone through and added params explicitly. Yes - it did change. I added the topic ID to accommodate some nested folder displays on the topics page, which means you can't just reference `parent` anymore

indirectlylit (approved):
cool, thank you. 

flagging that there is now a conflict in `BrowseResourceMetadata.vue`

PR #8651, opened 2021-11-10
Also in this sidepanel

nucleogenesis:
## Summary

Provides context-informed side panel for content viewing per [Figma specs](https://www.figma.com/file/uiZuCIqh2KYvBUfbCiJyyA/Hybrid-Learning?node-id=212%3A0).

Video shows, roughly in order:
- Guest user (no progresses shown - no lesson available)
- Logged in user in library context (progress shown)
- "Next folder navigation"
- Back route navigation
- Lesson content viewing
- Lesson content navigation (within the lesson)

https://user-images.githubusercontent.com/6356129/141061287-40fe4d11-7154-41dc-ada9-7b89a97ae69a.mp4

Overall, this was more complicated than I anticipated. Some notes:

- This is not particularly pretty. I added to the bloat in `LearnImmersiveLayout` with some methods that gather and map progresses to (if user is logged in) the returned nodes. It gets the Lesson model when in lesson context as well - this is all to pass it on to the `AlsoInThis` component (which is the body of the Side Panel). I hope the comments help clarify this.
- The CSS uses a lot of absolute positioning - I don't suspect it would break if the styles of the side panel width was changed but would be simple enough to fix if it did. It's not often you are given a finite space to work in the web as the side panel fits comfortably on nearly all screen sizes.
- The `detail_route` `next_content` now looks for a query param `topicOnly` - if true, then it will return the next topic. This was in order to easily get the "Next folder" content node
- ProgressBar had a `contentNode` prop, but didn't check if there was a `progress` property set - instead opting to look into a progress mapping. Perhaps I've missed something w/ that `useContentNodeProgress()` composable @rtibbles .
- I tried my hand at a quick Gherkin - definitely would love some feedback @radinamatic - I hope this covers enough and gets the idea across.
- The routing currently will go 2 resources deep - then if you back out twice it'll pop you back into the Library or Lesson you were in. This isn't intentional but I think it's a good UX overall. Interested in thoughts @jtamiace 

General thoughts on future improvements:
- `fetchSiblings` `detail_route` might be helpful and simplify this a bit? 
- Move fetching business into `AlsoInThis` to pare down `LearnImmersiveLayout` 
- `AlsoInThis` should probably be two separate components `AlsoInThisLesson` and `AlsoInThisFolder` I just wish I'd seen it earlier on :-/ 

## References

Fixes #8107 
Fixes #8307 

## Reviewer guidance

It ended up being more complicated than I had anticipated but it should be good to go :) 

Mainly be sure that you can get the 

[...]

PR #8674, opened 2021-11-12
AlsoInThis side panel follow-up

nucleogenesis:
## Summary

https://user-images.githubusercontent.com/6356129/141417838-65d424dd-5263-43b6-9848-07d0c8e73450.mp4

- Title can now be 2 lines @jtamiace 
- Less `position:absolute`, refactored layout a bit for simpler approach to accommodate the 2-line request
- CSS cleaned up quite a bit
- Use `useContentNodeProgress` various tools to build and use a shared store (:heart_eyes:) 
- General clean up, improved comments, etc

## References

I don't see an issue to fix, but this is follow up on #8651 

## Reviewer guidance

The above is what I intended to do here. There was discussion around routing that should be handled in a separate follow-up PR. 

Copy-pasted from #8651 - all should still work:

It ended up being more complicated than I had anticipated but it should be good to go :)

Mainly be sure that you can get the

General

Progress indicator on content where there is progress. Just a star if completed
Scrollbar appears in largely populated folders and all items are visible when scrolled down
The content you're viewing doesn't get listed in the side bar
Content which has a duration will show the time under the title
A channel with lots of folders, if you view a content a level or two deep, then you should be guaranteed to see the link on the bottom of the sidebar linking to the folder.
Lesson context

The contents listed should be from within the same lesson, not from siblings of that content in it's other locations (in channels in the library).
Links should remain in Lesson context - so you should always see the "view lesson resources" icon as you click through side panel "next in lesson" resources
Mobile

The progress / completed icon appears underneath the title (and optionally the duration if present)

----

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/plugins/learn/assets/src/views/AlsoInThis.vue
+++ b/kolibri/plugins/learn/assets/src/views/AlsoInThis.vue
@@ -177,7 +177,6 @@
         const progress = {
           progress: this.contentNodeProgressMap[node.content_id] || 0,
         };
-        console.log({ ...node, ...progress });
         return { ...node, ...progress };
       },
     },
--- a/kolibri/plugins/learn/assets/src/views/AlsoInThis.vue
+++ b/kolibri/plugins/learn/assets/src/views/AlsoInThis.vue
@@ -43,12 +43,12 @@
           </div>
           <div class="progress">
             <KIcon
-              v-if="content.progress === 1"
+              v-if="withProgress(content).progress === 1"
               icon="star"
               class="mastered-icon"
               :style="{ fill: $themeTokens.mastered }"
             />
-            <ProgressBar v-else :contentNode="content" class="bar" />
+            <ProgressBar v-else :contentNode="withProgress(content)" class="bar" />
           </div>
         </div>
 
@@ -87,6 +87,7 @@
   import TimeDuration from 'kolibri.coreVue.components.TimeDuration';
   import KResponsiveWindowMixin from 'kolibri-design-system/lib/KResponsiveWindowMixin';
   import SidePanelResourcesList from '../../../../../../kolibri/core/assets/src/views/FullScreenSidePanel/SidePanelResourcesList';
+  import useContentNodeProgress from '../composables/useContentNodeProgress';
   import genContentLink from '../utils/genContentLink';
   import LearningActivityIcon from './LearningActivityIcon.vue';
   import ProgressBar from './ProgressBar';
@@ -102,6 +103,10 @@
       TimeDuration,
     },
     mixins: [KResponsiveWindowMixin],
+    setup() {
+      const { contentNodeProgressMap } = useContentNodeProgress();
+      return { contentNodeProgressMap };
+    },
     props: {
       /**
        * contentNodes - The contentNode objects to be displayed. Each
@@ -166,7 +171,16 @@
         return context;
       },
     },
-    methods: { genContentLink },
+    methods: {
+      genContentLink,
+      withProgress(node) {
+        const progress = {
+          progress: this.contentNodeProgressMap[node.content_id] || 0,
+        };
+        console.log({ ...node, ...progress });
+        return { ...node, ...progress };
+      },
+    },
   };
 
 </script>
--- a/kolibri/plugins/learn/assets/src/views/AlsoInThis.vue
+++ b/kolibri/plugins/learn/assets/src/views/AlsoInThis.vue
@@ -24,31 +24,34 @@
           :kind="content.learning_activities"
         />
         <KIcon v-else class="topic-icon" icon="topic" />
-        <div class="content-meta" :style="{ top: (content.duration ? '-8px' : '8px') }">
-          <TextTruncator
-            :text="content.title"
-            :maxHeight="24"
-            :style="{ marginTop: (content.duration ? '8px' : '0px') }"
-          />
-          <TimeDuration
-            v-if="content.duration"
-            :style="{
-              color: $themeTokens.annotation,
-              top: '16px',
-              position: 'relative'
-            }"
-            :seconds="content.duration"
-          />
+
+        <div class="content-meta">
+          <div class="text-and-time">
+            <TextTruncator
+              class="content-title"
+              :text="content.title"
+              :maxHeight="72"
+            />
+            <TimeDuration
+              v-if="content.duration"
+              class="time-duration"
+              :style="{
+                color: $themeTokens.annotation,
+              }"
+              :seconds="content.duration"
+            />
+          </div>
+          <div class="progress">
+            <KIcon
+              v-if="content.progress === 1"
+              icon="star"
+              class="mastered-icon"
+              :style="{ fill: $themeTokens.mastered }"
+            />
+            <ProgressBar v-else :contentNode="content" class="bar" />
+          </div>
         </div>
-        <div class="progress">
-          <KIcon
-            v-if="content.progress === 1"
-            icon="star"
-            class="mastered-icon"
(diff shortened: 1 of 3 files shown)


kolibri/plugins/learn/assets/src/views/AlsoInThis.vue
```diff
@@ -24,31 +24,34 @@
...
+              class="time-duration"
+              :style="{
+                color: $themeTokens.annotation,
+              }"
+              :seconds="content.duration"
+            />
+          </div>
+          <div class="progress">
+            <KIcon
+              v-if="withProgress(content).progress === 1"
```
rtibbles: The progress bar already contains this conditional logic: https://github.com/learningequality/kolibri/blob/release-v0.15.x/kolibri/plugins/learn/assets/src/views/ProgressBar.vue#L11

And accesses progress via the contentNodeMap from the useContentNodeProgress composable - so no need to mix it into the content node first.

Can also remove this addition from the previous PR: https://github.com/learningequality/kolibri/blob/release-v0.15.x/kolibri/plugins/learn/assets/src/views/ProgressBar.vue#L60

kolibri/plugins/learn/assets/src/views/LearnImmersiveLayout.vue
```diff
@@ -275,38 +273,22 @@
...
-       * Prepares a list of content nodes which are also in the same lesson as this.lessonId
-       * @modifies this.viewResourcesContents - Sets it to the progress-mapped nodes
+       * Retrieves the lesson, then populates contentNodeProgressMap for the resulting
+       * nodes, then sets those nodes (sans this.content) to viewResourcesContents
+       * @modifies this.viewResourcesContents - Assigned the content nodes retrieved
+       * @modifies useContentNodeProgress.contentNodeProgressMap (indirectly) if the user
+       * is logged in
        */
       fetchLessonSiblings() {
         LessonResource.fetchModel({ id: this.lessonId }).then(lesson => {
```
rtibbles, line 290: Can just call `this.fetchContentNodeProgress({ lesson_id: this.lessonId });` here in parallel to the lesson fetch to avoid chaining (with the isUserLoggedIn check).

kolibri/plugins/learn/assets/src/views/LearnImmersiveLayout.vue
```diff
@@ -315,27 +297,30 @@
               ids: lesson.resources.map(resource => resource.contentnode_id),
             },
           }).then(contentNodes => {
-            return this.progressPromise(contentNodes);
+            const removeSelfFilter = node => node.id !== this.content.id;
+            if (this.$store.getters.isUserLoggedIn) {
+              this.mapSidePanelContentProgress(contentNodes).then(
+                () => (this.viewResourcesContents = contentNodes.filter(removeSelfFilter))
```
rtibbles: I would suggest treating the contentNodeProgress load as a lazy load (this is how we treat it in other places in the codebase), and not block loading of the main content for it.

If you just rely on the built in ProgressBar functionality as I suggest above, it will reactively update when the progress data has been loaded.

kolibri/plugins/learn/assets/src/views/LearnImmersiveLayout.vue
```diff
@@ -154,7 +151,8 @@
     mixins: [responsiveWindowMixin, commonCoreStrings],
     setup() {
       const { canDownload } = useCoreLearn();
-      return { canDownload };
+      const { contentNodeProgressMap, fetchContentNodeProgress } = useContentNodeProgress();
+      return { canDownload, contentNodeProgressMap, fetchContentNodeProgress };
```
rtibbles: Not seeing where `contentNodeProgressMap` is used in the component.

kolibri/plugins/learn/assets/src/views/LearnImmersiveLayout.vue
```diff
@@ -315,27 +297,30 @@
...
        */
       fetchSiblings() {
-        // Guard against this in lesson context
-        if (this.lessonId) {
-          return;
-        }
-
         ContentNodeResource.fetchTree({
           id: this.content.parent,
           params: {
```
rtibbles, line 317: If you're only interested in the immediate children, can also pass `depth: 1` in the params, to avoid recursing to the grand children of the parent.

rtibbles (dismissed):
I think the progress fetching and tracking can be made lazy and some of the previously added complexity removed.

I think for the sake of simplicity and not waiting on chaining, just using direct params filtering for the content node progress fetch (rather than waiting for nodes to return then fetching ids) will make for a snappier user experience.

kolibri/plugins/learn/assets/src/views/AlsoInThis.vue
nucleogenesis: Thanks for the reminder - once I got the bits hooked up elsewhere and they worked I was too pleased to remember I had to make proper use of ProgressBar

kolibri/plugins/learn/assets/src/views/AlsoInThis.vue
```diff
@@ -24,31 +24,34 @@
...
+              v-if="content.duration"
+              class="time-duration"
+              :style="{
+                color: $themeTokens.annotation,
+              }"
+              :seconds="content.duration"
+            />
+          </div>
+          <div class="progress">
+            <KIcon
```
rtibbles, line 45: This `KIcon` is also inside the `ProgressBar` component - is there a styling or layout reason to not to defer to it here? If so, would be good to have a code comment to explain.

kolibri/plugins/learn/assets/src/views/LearnImmersiveLayout.vue
```diff
@@ -275,81 +273,70 @@
...
+          );
 
-          /** A promise to get the nextFolder content - done always */
-          const nextFolderPromise = ContentNodeResource.fetchNextContent(parent.id, {
+          // Fetch and map the progress for the nodes if logged in
+          if (this.$store.getters.isUserLoggedIn) {
+            const getParams = {
+              ids: this.viewResourcesContents.map(node => node.id),
+            };
+            this.fetchContentNodeProgress(getParams);
```
rtibbles, line 335: Can avoid the delayed load here still by using `fetchContentNodeTreeProgress`:
```
this.fetchContentNodeTreeProgress({
          id: this.content.parent,
          params: {
            include_coach_content:
              this.$store.getters.isAdmin ||
              this.$store.getters.isCoach ||
              this.$store.getters.isSuperuser,
            depth: 1,
          },
        })
```

Can then be invoked in parallel to the node fetching.

rtibbles (commented):
A couple of comments but nothing blocking code wise. Have not reviewed functionality!

jtamiace (approved):
Was able to get all the scenarios listed working. Thank you for doing the follow up!

jtamiace:
Actually... I think I found a bug. If you click on a resource from the Bookmarks tab, and that resource was bookmarked from a Channel, the 'Also in this' panel shows a random lesson that the bookmarked resource is not a part of.

I realize this happens in a specific sequence:
1. Go to a resource in a lesson
2. Exit the lesson
3. Go to the bookmarks tab
4. Go to a bookmarked resource
5. Open in the 'Also in this' panel

It should be the lesson list. But if you: (1) Exit the bookmarked resouce renderer, (2) refresh the page, (3) click the bookmarked resource again, (4) open the 'Also in this' panel

It should return to normal

rtibbles:
I am going to file this as a follow up issue, as this was pre-existing to this PR, I think.

PR #7431, opened 2020-08-04
Preliminary handling for missing content in editing lessons and quizzes

rtibbles:
### Summary
* Stops frontend from auto-removing missing content
* Removes separate ancestor fetching to avoid errors on content management page for lessons
* Adds missing resource placeholder for lesson resource details
* Allows saving resources that are missing
* Adds a message in quiz preview when trying to render a question that is missing
* Adds an icon and tooltip to each question that is missing its exercise

### Reviewer guidance
Quiz preview
Screenshot from 2020-08-03 17-51-05

Lesson resources
Screenshot from 2020-08-03 17-52-18

### References
Fixes #7327

…

----

### Contributor Checklist

PR process:

- [x] PR has the correct target branch and milestone
- [x] PR has 'needs review' or 'work-in-progress' label
- [x] If PR is ready for review, a reviewer has been added. (Don't use 'Assignees')
- [ ] If this is an important user-facing change, PR or related issue has a 'changelog' label
- [ ] If this includes an internal dependency change, a link to the diff is provided

Testing:

- [x] Contributor has fully tested the PR manually
- [x] If there are any front-end changes, before/after screenshots are included
- [ ] Critical user journeys are covered by Gherkin stories
- [ ] Critical and brittle code paths are covered by unit tests

### Reviewer Checklist

- Automated test coverage is satisfactory
- PR is fully functional
- PR has been tested for [accessibility regressions](http://kolibri-dev.readthedocs.io/en/develop/manual_testing.html#accessibility-a11y-testing)
- External dependency files were updated if necessary (`yarn` and `pip`)
- Documentation is updated
- Contributor is in AUTHORS.md

PR #8677, opened 2021-11-12
Adds guards against non-locally available content.

### LINKED ISSUE #6563: Make accessing classSummary/contentMap (and other maps) safer in Coach page
### Observed behavior

#6552 was caused by a runtime error where we tried to access a property from an undefined object since the `classSummary/contentMap` getter did not have an associated value for a content ID, because the content item was possibly deleted and the lessons using that content were not properly cleaned up.

We should make sure that we don't get runtime errors like this that make pages inaccessible.

Should also look into other maps from the `classSummary` module that might pose a similar risk.

### Expected behavior

The possible mismatches between the Kolibri content library + the logs data + the lessons + quizzes, etc. don't cause crashes anywhere in Kolibri, and esp. the coach pages.

### User-facing consequences

…

### LINKED ISSUE #7325: 👻 Unavailable resources on Coach Page are properly displayed
### Observed behavior

Some work related to missing content didn't get implemented in 0.14, but the strings for the functionality were added and have been translated.

Refs:

* https://www.figma.com/file/LM0YUkbWAOIerZ3EEg0FN6bS/0.14-P2P-Sync?node-id=547%3A11
* https://www.notion.so/learningequality/Data-syncing-and-facility-management-0-14-0-44a8d0e4a5674d35b2a18a440e967cea

image

### User-facing consequences

Users of data syncing could end up with very confusing user experiences

### Context

0.14.0

rtibbles:
## Summary
* Consolidates report components with repetitive boilerplate across group/non-group to prevent deviations
* Adds MissingResourceAlert component to show consistent messages to end users about missing resources
* Updates all reports to prevent missing resources from disrupting display, while ensuring the maximal available information is displayed to the end user where possible.
* Adds fetching of metadata that is available, but where the resource is unavailable to maximize the contextual information provided to coaches.
* Updates learn lesson playlist and quiz interfaces to display warnings
* Updates learn lesson also in this to display warnings

## References
Fixes #6563
Fixes #7325

## Reviewer guidance
Create quizzes and lessons assigned to both classes and groups.
Interact with them with learner accounts, but do not complete them.
Delete 1 or more of the resources required by the quizzes and lessons.
Review all reports and revisit the plan page for each quiz and lesson. View the reports for both non-group and group reports.
Return to the lessons and quizzes as a learner.

Learner quiz
Screenshot from 2023-02-13 17-34-53

Learner Lesson playlist
Screenshot from 2023-02-13 17-40-37

Learner lesson 'also in this'
file:///home/richard/Pictures/Screenshots/Screenshot%20from%202023-02-13%2017-52-08.png

Coach notifications
Screenshot from 2023-02-13 17-35-22

Coach lesson report
Screenshot from 2023-02-13 17-35-37

Coach lesson resource detail report
Screenshot from 2023-02-13 17-35-43

Coach lesson report preview
Screenshot from 2023-02-13 17-35-47

Coach quiz report
Screenshot from 2023-02-13 17-36-05

Coach quiz difficult questions report
Screenshot from 2023-02-13 17-36-11

Coach lesson plan (at the moment this is mostly unchanged, showing resources not available across the board - should I update to showing available metadata parallel to the reports?)
Screenshot from 2023-02-13 17-39-34

Coach quiz plan preview
Screenshot from 2023-02-13 17-40-03

----

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/core/assets/src/exams/utils.js
+++ b/kolibri/core/assets/src/exams/utils.js
@@ -119,6 +119,7 @@ export function fetchNodeDataAndConvertExam(exam) {
   return ContentNodeResource.fetchCollection({
     getParams: {
       ids: uniq(exam.question_sources.map(item => item.exercise_id)),
+      no_available_filtering: true,
     },
   }).then(contentNodes => {
     return {
@@ -160,6 +161,7 @@ export function getExamReport(examId, tryIndex = 0, questionNumber = 0, interact
           contentPromise = ContentNodeResource.fetchCollection({
             getParams: {
               ids: uniq(questionSources.map(item => item.exercise_id)),
+              no_available_filtering: true,
             },
           });
         } else {
--- a/kolibri/plugins/coach/assets/src/views/plan/LessonEditDetailsPage/EditDetailsResourceListTable.vue
+++ b/kolibri/plugins/coach/assets/src/views/plan/LessonEditDetailsPage/EditDetailsResourceListTable.vue
@@ -89,7 +89,6 @@
   import ContentIcon from 'kolibri.coreVue.components.ContentIcon';
   import commonCoreStrings from 'kolibri.coreVue.mixins.commonCoreStrings';
   import CoachContentLabel from 'kolibri.coreVue.components.CoachContentLabel';
-  import { coachStringsMixin } from '../../common/commonCoachStrings';
 
   // This is a simplified version of ResourceListTable that is supposed to work
   // outside of the LessonSummaryPage workflow.
@@ -103,7 +102,7 @@
       ContentIcon,
       CoachContentLabel,
     },
-    mixins: [coachStringsMixin, commonCoreStrings],
+    mixins: [commonCoreStrings],
     props: {
       // Array<{ contentnode_id, content_id, channel_id }>
       resources: {
@@ -155,7 +154,7 @@
         });
       },
       resourceMissingText() {
-        return this.getMissingContentString('resourceNotFoundOnDevice');
+        return this.coreString('resourceNotFoundOnDevice');
       },
     },
     methods: {
--- a/kolibri/plugins/coach/assets/src/views/plan/LessonEditDetailsPage/EditDetailsResourceListTable.vue
+++ b/kolibri/plugins/coach/assets/src/views/plan/LessonEditDetailsPage/EditDetailsResourceListTable.vue
@@ -146,11 +146,10 @@
               num_coach_contents: match.num_coach_contents,
               channelTitle: this.resourceChannelTitle(contentnode_id),
             };
-          } else {
-            // Need to filter out objects not in the contentNodeMap.
-            // Hopefully the contentNodeMap is always updated.
-            return resource;
           }
+          // Need to filter out objects not in the contentNodeMap.
+          // Hopefully the contentNodeMap is always updated.
+          return resource;
         });
       },
       resourceMissingText() {
@@ -163,7 +162,11 @@
         this.$emit('update:resources', resources);
       },
       resourceChannelTitle(id) {
-        const match = this.$store.getters['getChannelObject'](this.contentNodeMap[id].channel_id);
+        const contentNode = this.contentNodeMap[id];
+        if (!contentNode) {
+          return '';
+        }
+        const match = this.$store.getters['getChannelObject'](contentNode.channel_id);
         return match ? match.title : '';
       },
       removeResource(resource) {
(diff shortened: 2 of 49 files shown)


kolibri/plugins/coach/assets/src/views/plan/LessonEditDetailsPage/EditDetailsResourceListTable.vue
```diff
@@ -164,7 +164,9 @@
         this.$emit('update:resources', resources);
       },
       resourceChannelTitle(id) {
-        const match = this.$store.getters['getChannelObject'](this.contentNodeMap[id].channel_id);
+        const match = this.$store.getters['getChannelObject'](
+          (this.contentNodeMap[id] || {}).channel_id
+        );
```
indirectlylit: This looks like a probable bug:

```
(this.contentNodeMap[id] || {}).channel_id
```

because it will sometimes evaluate to `undefined`.

If that is in fact intended behavior, there might be a more clear way to handle this.

kolibri/plugins/coach/assets/src/views/reports/ReportsGroupReportLessonPage.vue
```diff
@@ -117,7 +117,9 @@
         return this.getLearnersForGroups([this.$route.params.groupId]);
       },
       table() {
-        const contentArray = this.lesson.node_ids.map(node_id => this.contentNodeMap[node_id]);
+        const contentArray = this.lesson.node_ids
+          .map(node_id => this.contentNodeMap[node_id])
+          .filter(Boolean);
```
indirectlylit: These changes seem fine to me

indirectlylit (commented):
I can't think of any particular issue with this.

In the longer run though we may need to revert in order to support https://github.com/learningequality/kolibri/issues/7325

indirectlylit:
actually, does this break the 'Resource not found on device' behavior?


https://github.com/learningequality/kolibri/pull/7431

image

rtibbles:
It may - I reverted this to a draft, to properly follow up on #7325 here instead.

pcenov:
@radinamatic Tested in both Windows and Ubuntu - no issues observed, implemented as specified!

radinamatic (approved):
Manual QA revealed no issues or regressions. good to go! 💯 :shipit:

rtibbles:
Suggested updates to the learner experience from @jtamiace:
image

Details here:
> the “missing resources” alert only shows up for the question that’s missing, not for all questions. i was hoping that replacing the gray question status bubbles on the left with the warning icon can play a similar role to the alert while giving an overview of which questions are problematic
removed the red error message. the new yellow one functionally does the same thing and is more specific to the problem happening
remove the submit quiz button and replace with an explanation - doesn’t seem fair to record scores for questions they can’t do
potentially adding the missing resource indicator on quiz cards on the home & class pages. i forget if we touched on this or whether it wasn’t desirable, but it could help show extent of the missing resource problem?

pcenov:
@rtibbles and @radinamatic Retested on both Windows and Ubuntu and the latest changes are working correctly as specified above.
I am only going to mention that if a learner has already submitted a quiz prior to the removal of some of the questions in that particular quiz, then the result is not being recalculated to exclude the missing questions which might or might not be problematic and should be further discussed.

radinamatic (approved):
Re-tested and ready to go! 💯 :shipit:

kolibri/core/assets/src/exams/utils.js
```diff
@@ -119,6 +119,7 @@ export function fetchNodeDataAndConvertExam(exam) {
   return ContentNodeResource.fetchCollection({
     getParams: {
       ids: uniq(exam.question_sources.map(item => item.exercise_id)),
+      no_available_filtering: true,
```
marcellamaki, line 122: It seems like this param is newly introduced by this PR, and this is a bit of a (definitely non-blocking) nitpick but it is pretty hard for me to deduce from the name inline in some of the places it's called in the frontend what exactly this means. Does it mean the param is filtering for "none available"? Does it mean there are no available filters? I think based on the context of the PR it is filtering out unavailable content nodes, but seeing this outside of the PR, I'm not sure I would deduce that

marcellamaki (approved):
Code read through is long, but looks good to me. I like the page refactoring for some of the base pages.

I added one very nit-picky, non-blocking comment about names, but, good to go

kolibri/core/assets/src/exams/utils.js
rtibbles: By default the ContentNode endpoints filter by available=True, this flag tells the backend to do no available filtering, and return all nodes. I am happy to change the name but couldn't think of a more descriptive one without being incredibly verbose.

marcellamaki: There probably isn't a better option! thanks for the explanation.

PR #8680, opened 2021-11-12
Removes hard coded Studio URL from frontend.

### LINKED ISSUE #8139: Hard coded Kolibri Studio URL in frontend breaks upgrades when option is set
## Observed behavior
When setting this option https://github.com/learningequality/kolibri/blob/develop/kolibri/utils/options.py#L366 Kolibri is able to import solely from another content source.
However, when this is set, the hard coded URL here: https://github.com/learningequality/kolibri/blob/develop/kolibri/plugins/device/assets/src/views/ManageContentPage/api.js#L6 causes errors during the upgrade work flow, as it attempts to generate the diff stats from Studio and not the substituted server.

## Expected behavior
The Studio URL should not be hard coded in the frontend. The backend handles a missing `baseurl` parameter and in fills with the option value appropriately.

If having the URL is still needed, we should use the `plugin_data` mechanisms to pass the option value to the frontend.

## User-facing consequences
Breakage if pointing to another source URL for import.

rtibbles:
## Summary
* The default Studio URL was hardcoded unnecessarily in the frontend
* This caused upgrades for devices that use a non-Studio URL as their base to break
* Removes this dependency and code that assumed this

## References
Fixes #8139

## Reviewer guidance
Run Kolibri with the default content download source set to something else - e.g.
```
KOLIBRI_CENTRAL_CONTENT_BASE_URL=https://hotfixes.studio.learningequality.org
```

 as an env var.

Try to upgrade a channel that is on that download source. See that it successfully complete.

----

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/plugins/device/assets/src/views/ManageContentPage/api.js
+++ b/kolibri/plugins/device/assets/src/views/ManageContentPage/api.js
@@ -3,8 +3,6 @@ import { TaskResource, ChannelResource, RemoteChannelResource } from 'kolibri.re
 import { TaskTypes } from '../../constants';
 import { NetworkLocationResource } from '../../apiResources';
 
-const kolibriStudioUrl = 'https://studio.learningequality.org';
-
 function getChannelOnDrive(driveId, channelId) {
   const reject = () => Promise.reject('CHANNEL_NOT_ON_DRIVE');
   return TaskResource.localDrives().then(response => {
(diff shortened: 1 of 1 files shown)


jredrejo (approved):
Reviewed and tested, everything continue working fine (theorically better removing the hardcoded url and allowing the task to provide the baseurl)

PR #8709, opened 2021-11-16
Export size bug

### LINKED ISSUE #7625: Export to attached drive > Cancel task - on-click the Cancel button the *N of M resources (size)* exported text is missing under the task
### Observed behavior

Export to attached drive > Cancel task - on-click the Cancel button the *N of M resources (size)* exported text is missing under the task.

### Expected behavior

After clicking the Cancel button of an Export task the *N of M resources (size)* exported text must be visible under the task as it is with all the other tasks (Import, Delete, etc)

### User-facing consequences

…

### Steps to reproduce

Preconditions: Resources imported, attached drive to the device
1. Login as a super admin user
2. Go to Device > Channels
3. Click the Manage button of a channel
4. Select some resources
5. Click the Export button
6. Select the attached drive and click Export
7. After some resources are exported click the Cancel button
8. See the text below the task

### Context

Kolibri 0.14.3
Raspberry Pi 4
OS: Raspbian GNU, Linux 10
Chrome (latest) + Firefox (latest)

AtKristijan:
## Summary

Now when we are exporting first we get a bar with progress and total amount of bytes and total number of resources. When we press cancel now at the end it will be show the total amount of bytes. 

## References
Fixes https://github.com/learningequality/kolibri/issues/7625

## Reviewer guidance

…

----

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/core/content/management/commands/exportchannel.py
+++ b/kolibri/core/content/management/commands/exportchannel.py
@@ -39,8 +39,4 @@ class Command(AsyncCommand):
                     pass
 
                 if self.is_cancelled():
-                    try:
-                        os.remove(dest)
-                    except IOError:
-                        pass
                     self.cancel()
--- a/kolibri/core/content/management/commands/exportcontent.py
+++ b/kolibri/core/content/management/commands/exportcontent.py
@@ -15,6 +15,9 @@ logger = logging.getLogger(__name__)
 
 
 class Command(AsyncCommand):
+    exported_size = 0
+    total_resources = 0
+
     def add_arguments(self, parser):
         node_ids_help_text = """
         Specify one or more node IDs to import. Only the files associated to those node IDs will be imported.
@@ -86,7 +89,6 @@ class Command(AsyncCommand):
         with self.start_progress(
             total=total_bytes_to_transfer
         ) as overall_progress_update:
-
             for f in files:
 
                 if self.is_cancelled():
@@ -97,14 +99,10 @@ class Command(AsyncCommand):
                     exported_files.append(dest)
 
             if self.is_cancelled():
-                # Cancelled, clean up any already downloading files.
-                for dest in exported_files:
-                    os.remove(dest)
                 self.cancel()
 
     def export_file(self, f, data_dir, overall_progress_update):
         filename = get_content_file_name(f)
-
         try:
             srcpath = paths.get_content_storage_file_path(filename)
             dest = paths.get_content_storage_file_path(filename, datafolder=data_dir)
@@ -117,17 +115,21 @@ class Command(AsyncCommand):
         if os.path.isfile(dest) and os.path.getsize(dest) == f["file_size"]:
             overall_progress_update(f["file_size"])
             return
-
         copy = transfer.FileCopy(srcpath, dest, cancel_check=self.is_cancelled)
-
         with copy, self.start_progress(
             total=copy.total_size
         ) as file_cp_progress_update:
             try:
                 for chunk in copy:
                     length = len(chunk)
+                    self.exported_size = self.exported_size + length
                     overall_progress_update(length)
                     file_cp_progress_update(length)
             except transfer.TransferCanceled:
+                job = get_current_job()
+                if job:
+                    job.extra_metadata["file_size"] = self.exported_size
+                    job.extra_metadata["total_resources"] = 0
+                    job.save_meta()
                 return
         return dest
--- a/kolibri/plugins/device/assets/src/views/ManageTasksPage/TaskPanel.vue
+++ b/kolibri/plugins/device/assets/src/views/ManageTasksPage/TaskPanel.vue
@@ -55,7 +55,7 @@
           />
         </template>
       </div>
-      <template v-if="!taskIsCompleted && !taskIsCanceled">
+      <template v-if="!taskIsCompleted">
         <p v-if="sizeText" class="details-size">
           {{ sizeText }}
         </p>
@@ -192,6 +192,10 @@
             numResources: total_resources,
             bytesText: bytesForHumans(file_size),
           });
+        } else if (file_size) {
+          return this.$tr('cancelSize', {
+            bytesText: bytesForHumans(file_size),
+          });
         }
         return '';
       },
@@ -261,6 +265,11 @@
           '{numResources} {numResources, plural, one {resource} other {resources}} ({bytesText})',
         context: 'Indicates the number of resources and their size.',
       },
+      cancelSize: {
+        message: 'Exported size: ({bytesText})',
+        context: 'Indicates the number of resources and their size.',
+      },
+
       statusInProgress: {
         message: 'In-progress',
         context: 'Label indicating that a task is in progress.',
--- a/kolibri/core/content/test/test_import_export.py
+++ b/kolibri/core/content/test/test_import_export.py
@@ -1220,7 +1220,7 @@ class ExportChannelTestCase(TestCase):
             local_src_path, local_dest_path, cancel_check=is_cancelled_mock
         )
         cancel_mock.assert_called_with()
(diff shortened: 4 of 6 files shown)


rtibbles (commented):
Code changes look right to me.

radinamatic:
Tested the EXE asset on Windows 7, and I'm not seeing the full wording _N of M resources (X MB) successfully exported_, although I'm not sure if that was included in the fix. Findings for each of 4 export tests:

1. Export of one folder from PhET channel (46 resources, 84 MB) was cancelled right after the DB was exported, and none of the resources were saved on the external drive: wording states the full amount of resources that were initially set to be exported.
2. Successful export, size was too small for me to cancel in time: wording as expected.
3. Export of 191 resources (260 MB) from KA BG channel was at 50% when I cancelled: wording is missing completely.
4. Export of the full PhET channel was at 10% when I cancelled: wording states the full size of the channel, but not how much was exported. Seems that the partial export is somehow cancelled completely and no resources register as exported at all...? I checked and even re-tried, waiting to export 90% before I cancelled: cancelled or failed exports only report the full amount and size of resources, not that of the actual export task. 

    If I delete the channel on the device, and try to import from the drive where I previously exported 90%, the channel seems not available for import. Cancelling itself also took a long time, and when I checked the drive, the ``/KOLIBRI_DATA/content/databases`` folder in the drive was empty (no wonder nothing could be imported). Why is that? Makes no sense to invalidate 90% of the exported resources from the channel... 😕 

    **Does Kolibri delete the DB of the channel if the export is cancelled?** 

Win7 (start)  Running  - Oracle VM VirtualBox_023

[logs.zip](https://github.com/learningequality/kolibri/files/7566720/logs.zip)


Does this need to be repeated on other platforms? We don't have a working RPi image where the issue was originally reported.

rtibbles:
If an export is cancelled, it cleans up and deletes anything that was copied, we can certainly change this behaviour, but that is the current behaviour.

This is why for cancelled tasks it is consistent that no exported statistics are reported.

radinamatic:
> If an export is cancelled, it cleans up and deletes anything that was copied

I can certainly imagine the frustration of the user who patiently waited for hours to have the large channel of several GBs exported to external drive (we know that process can be slow), and seeing all that vanish if they cancel by mistake. Even more so since there is no confirmation modal alerting them that everything that has been copied up to that point will be deleted. What was the reason for this approach? I could understand that deleting specific files that were half-way and partially copied might be prudent, as the resources could not be correctly rendered, but deleting ALL...? cc @jtamiace 

Finally, now I'm not even sure what was the exact point of this PR. From the linked issue I assumed that the initially reported problem of not having the wording _N of M resources (X MB) successfully exported_, is what was being fixed here, and from what @rtibbles is saying above that seems not to have been the goal, so I'm confused. @AtKristijan, it could help if you clearly state under **Summary** and **Reviewer guidance**, what exactly needed to be tested.

rtibbles:
@radinamatic I've discussed this with @AtKristijan on slack, and he is now following up to bring the export cancellation behaviour in line with what you have outlined (which also makes it consistent with the cancellation behaviour of content import).

radinamatic:
Excellent, I'll be happy to test again once that is implemented! 🙂

AtKristijan:
@radinamatic  
Only N x M here was made, but I will implement and I will change cancel behavior. 

So to resume: 
1) We need to keep the data even if we press cancel.
2) Bytes and total resource need to be show only on exported data

AtKristijan:
only size

kolibri/core/content/management/commands/exportchannel.py
```diff
@@ -39,8 +39,8 @@ def handle_async(self, *args, **options):
                     pass
 
                 if self.is_cancelled():
-                    try:
```
rtibbles, line 42: Delete this code block rather than commenting out.

kolibri/core/content/management/commands/exportcontent.py
```diff
@@ -58,6 +61,8 @@ def add_arguments(self, parser):
     def update_job_metadata(self, total_bytes_to_transfer, total_resource_count):
         job = get_current_job()
         if job:
+            print("file_size",total_resource_count)
```
rtibbles: Remove print debugging statements here.

kolibri/core/content/management/commands/exportcontent.py
```diff
@@ -98,13 +102,12 @@ def handle_async(self, *args, **options):
 
             if self.is_cancelled():
                 # Cancelled, clean up any already downloading files.
-                for dest in exported_files:
```
rtibbles, line 101: Delete this code block rather than commenting out.

kolibri/core/content/management/commands/exportcontent.py
```diff
@@ -117,17 +120,23 @@ def export_file(self, f, data_dir, overall_progress_update):
...
         copy = transfer.FileCopy(srcpath, dest, cancel_check=self.is_cancelled)
-
         with copy, self.start_progress(
             total=copy.total_size
         ) as file_cp_progress_update:
             try:
                 for chunk in copy:
                     length = len(chunk)
+                    self.exported_size = self.exported_size + length
+                    print(self.exported_size)
```
rtibbles: Remove print statement.

kolibri/core/content/management/commands/exportcontent.py
```diff
@@ -117,17 +120,23 @@ def export_file(self, f, data_dir, overall_progress_update):
...
                     overall_progress_update(length)
                     file_cp_progress_update(length)
             except transfer.TransferCanceled:
+                job = get_current_job()
+                if job:
+                    job.extra_metadata["file_size"] = self.exported_size
+                    job.extra_metadata['total_resources'] = 0
+                    job.save_meta()
+                    print(job.extra_metadata)
```
rtibbles: Remove print statement.

kolibri/plugins/device/assets/src/views/ManageTasksPage/TaskPanel.vue
```diff
@@ -261,6 +265,11 @@
           '{numResources} {numResources, plural, one {resource} other {resources}} ({bytesText})',
         context: 'Indicates the number of resources and their size.',
       },
+      cancelSize: {
```
rtibbles, line 268: We can't add new strings in a patch version release - and I almost think it might be better to see if we can find a way to emulate the resource import flow and be able to count the number of resources that we have successfully exported in order to be able to use the `exportSuccess` message below instead.

kolibri/core/content/management/commands/exportcontent.py
```diff
@@ -117,17 +120,23 @@ def export_file(self, f, data_dir, overall_progress_update):
...
                     length = len(chunk)
+                    self.exported_size = self.exported_size + length
+                    print(self.exported_size)
                     overall_progress_update(length)
                     file_cp_progress_update(length)
             except transfer.TransferCanceled:
+                job = get_current_job()
+                if job:
+                    job.extra_metadata["file_size"] = self.exported_size
+                    job.extra_metadata['total_resources'] = 0
```
rtibbles: I think being able to give a meaningful number here would be the best result - I can give some guidance on how to estimate these using some temporary annotation techniques, similar to what we use for importability.

rtibbles (commented):
Some cleanup - the path I am suggesting adds a little more complexity, but we also can't add a new string for a 0.15.x patch.

rtibbles (commented):
Still some commented out code that could be deleted.

We also can't have a new string within the patch release. However, it would be fairly complex to resolve this without a new string, so maybe we should retarget this to develop and add the string @marcellamaki?

marcellamaki:
> should retarget this to develop and add the string

Yep, that makes sense to me, @rtibbles !

rtibbles (approved):

PR #8740, opened 2021-11-18
Evaluation api for mastery/attempt logs

### LINKED ISSUE #8543: Provide API for practice quiz report
Due to the updates in the examlogging API, we will need new functionality to take advantage of the exam attempts to return the following data for use in the components below:

### Component: `PracticeQuizReport`
**Summary information in header:**
- Status (in progress or completed, though, tbh, it would never _not_ be completed, right?)
- Best score out of all attempts
- For the best attempt (the one with the highest score), how many answers were answered correctly
- For the best attempt (the one with the highest score), what was the time?

**Information in the main content area (per attempt from the dropdown - the 5 most recent attempts are shown in the dropdown):**
- Score for that attempt
- Total questions answered correctly
- Total questions answered correctly from previous attempt
- Individual question answered this attempt and the previous attempt
- Time spent for that attempt
- Time spent for the previous attempt
- Last attempted

See Figma below for details.
Screen Shot 2021-10-28 at 9 53 29 PM

bjester:
## Summary

- Adds new endpoints for reporting practice quiz improvements:
  - `masterylog/summary?content={}&user={}`: 
  Returns a list of masterylogs annotated with total `correct`
  - `masterylog/{pk}/diff`: 
  Returns a materylog annotated with `diff` of `null` or `{correct: signed int, time_spent: signed int (seconds)}` as well as `attemptlogs` annotated with `diff` compared with previous try
  - `attemptlog/diff?content={}&user={}&item={}&masterylog={}`: 
  Returns a list of attemptlogs annotated with `diff` of `{correct: (null,-1,0,1)}`
- Adds frontend resource methods for new endpoints

| `attemptlog.correct` | `attemptlog.diff.correct` | Diff text |
| --- | --- | --- |
| 0 | null | N/A |
| 1 | null | N/A |
| 0 | 0 | "You also answered this question incorrectly on the previous attempt" |
| 1 | 0 | N/A |
| 0 | 1 | Not possible |
| 1 | 1 | "You improved the incorrect answer from the previous attempt" |
| 0 | -1 | "You answered this question correctly on the previous attempt" |
| 1 | -1 | Not possible |

## References

Resolves https://github.com/learningequality/kolibri/issues/8543

## Reviewer guidance

…

----

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/core/auth/api.py
+++ b/kolibri/core/auth/api.py
@@ -85,9 +85,13 @@ class KolibriAuthPermissionsFilter(filters.BaseFilterBackend):
     to the ``filter_readable`` method on ``KolibriAbstractBaseUser``.
     """
 
+    def __init__(self, endpoints=None):
+        self.endpoints = endpoints or ["-list"]
+
     def filter_queryset(self, request, queryset, view):
-        if request.method == "GET" and request.resolver_match.url_name.endswith(
-            "-list"
+        if request.method == "GET" and any(
+            request.resolver_match.url_name.endswith(endpoint)
+            for endpoint in self.endpoints
         ):
             # only filter down the queryset in the case of the list view being requested
             return request.user.filter_readable(queryset)
--- a/kolibri/core/auth/api.py
+++ b/kolibri/core/auth/api.py
@@ -85,14 +85,13 @@ class KolibriAuthPermissionsFilter(filters.BaseFilterBackend):
     to the ``filter_readable`` method on ``KolibriAbstractBaseUser``.
     """
 
-    def __init__(self, endpoints=None):
-        self.endpoints = endpoints or ["-list"]
-
     def filter_queryset(self, request, queryset, view):
-        if request.method == "GET" and any(
-            request.resolver_match.url_name.endswith(endpoint)
-            for endpoint in self.endpoints
-        ):
+        # if the url name ends with "-list" or the endpoint is explicitly declared a list endpoint
+        # with .detail=False
+        is_list = request.resolver_match.url_name.endswith("-list") or not getattr(
+            request.resolver_match.func, "detail", True
+        )
+        if request.method == "GET" and is_list:
             # only filter down the queryset in the case of the list view being requested
             return request.user.filter_readable(queryset)
         # otherwise, return the full queryset, as permission checks will happen object-by-object
--- a/kolibri/core/logger/api.py
+++ b/kolibri/core/logger/api.py
@@ -5,9 +5,11 @@ from random import randint
 
 from django.core.exceptions import PermissionDenied
 from django.db import transaction
+from django.db.models import IntegerField
 from django.db.models import OuterRef
 from django.db.models import Subquery
 from django.db.models import Sum
+from django.db.models import Value
 from django.http import Http404
 from django_filters.rest_framework import CharFilter
 from django_filters.rest_framework import DjangoFilterBackend
@@ -18,6 +20,7 @@ from le_utils.constants import exercises
 from rest_framework import filters
 from rest_framework import serializers
 from rest_framework import viewsets
+from rest_framework.decorators import action
 from rest_framework.exceptions import ValidationError
 from rest_framework.response import Response
 
@@ -36,6 +39,11 @@ from kolibri.core.exams.models import Exam
 from kolibri.core.lessons.models import Lesson
 from kolibri.core.logger.constants import interaction_types
 from kolibri.core.logger.constants.exercise_attempts import MAPPING
+from kolibri.core.logger.evaluation import attempts_diff
+from kolibri.core.logger.evaluation import find_previous_tries_attempts
+from kolibri.core.logger.evaluation import find_tries
+from kolibri.core.logger.evaluation import get_previous_try
+from kolibri.core.logger.evaluation import try_diff
 from kolibri.core.notifications.api import create_summarylog
 from kolibri.core.notifications.api import parse_attemptslog
 from kolibri.core.notifications.api import parse_summarylog
@@ -852,7 +860,7 @@ class MasteryFilter(FilterSet):
 
     class Meta:
         model = MasteryLog
-        fields = ["content"]
+        fields = ["content", "user"]
 
 
 class MasteryLogViewSet(ReadOnlyValuesViewset):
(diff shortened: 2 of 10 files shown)


kolibri/core/logger/test/test_api.py
```diff
@@ -184,3 +186,121 @@ def test_csv_download_unicode_username(self):
...
+                self.assertEqual(tries_count, len(response.data))
+                for try_index, mastery_log in enumerate(response.data):
+                    self.assertEqual(
+                        self.user_tries[user_index][try_index].id, mastery_log["id"]
+                    )
+
+    def test_diff(self):
+        for user_index, user_tries in enumerate(self.user_tries):
+            for try_index, user_try in enumerate(user_tries):
+                self.client.force_login(self.users[user_index])
```
bjester, line 217: @rtibbles I needed this as it failed checking permissions, but I didn't run into this with any of the other tests, which makes me think the `KolibriAuthPermissionsFilter` isn't working correctly (or coder error) for the others

bjester: Fixed it

kolibri/core/logger/api.py
```diff
@@ -871,6 +883,51 @@ class MasteryLogViewSet(ReadOnlyValuesViewset):
...
+        self.attempt_log_view_set = AttemptLogViewSet()
+
+    @action(detail=False)
+    def summary(self, request):
+        user_id = request.GET.get("user")
+        content_id = request.GET.get("content")
+
+        # required parameters
+        if not user_id:
+            raise Http404("Parameter `user` is required")
```
rtibbles: Not blocking - I've used 412 previously for missing parameters (e.g. in the required parameters decorator https://github.com/learningequality/kolibri/blob/release-v0.15.x/kolibri/core/decorators.py#L35 - which I realize I have failed to make usable just specific endpoint methods).

kolibri/core/auth/api.py
```diff
@@ -85,9 +85,13 @@ class KolibriAuthPermissionsFilter(filters.BaseFilterBackend):
     to the ``filter_readable`` method on ``KolibriAbstractBaseUser``.
     """
 
+    def __init__(self, endpoints=None):
+        self.endpoints = endpoints or ["-list"]
```
rtibbles: I am sure there is a clever DRF way to detect whether something is a list endpoint or not - but it's not coming to mind...

kolibri/core/logger/api.py
```diff
@@ -913,4 +974,50 @@ class AttemptLogViewSet(ReadOnlyValuesViewset):
...
+
+    @action(detail=False)
+    def diff(self, request):
+        # protect against expensive queries from loose unsupported filtering
+        min_required_filter_sets = (
+            {"masterylog"},
+            {"content", "item"},
+            {"content", "user"},
+        )
+        for filter_set in min_required_filter_sets:
```
rtibbles, line 996: Nice - the required parameters stuff I'd done previously that I linked above didn't allow for multiple disjoint sets - and then the required parameters I did in [Studio](https://github.com/learningequality/studio/blob/unstable/contentcuration/contentcuration/viewsets/base.py#L838) was just preventing no queries.

The other alternative might be to enforce cursor pagination on endpoints.

kolibri/core/logger/api.py
```diff
@@ -913,4 +974,50 @@ class AttemptLogViewSet(ReadOnlyValuesViewset):
...
+            return Response("Minimum filter condition not met", status=412)
+
+        # apply filters, narrow down to most recent try
+        target_attempt_logs = self.get_queryset()
+        # masterylog filter would filter it to a specific try
+        if "masterylog" not in request.GET:
+            # without masterylog filter, narrow down to most recent attempt for user+item
+            target_attempt_logs = AttemptLog.objects.filter(
+                id__in=Subquery(
+                    target_attempt_logs.order_by("-completion_timestamp")
```
rtibbles: I think it would be safer to order by `end_timestamp` here - as we have no real guarantee that `completion_timestamp` is set for quizzes.

The other thing that might be good is to add test coverage for the case of multiple attemptlogs per user - I think my similar code here https://github.com/learningequality/kolibri/pull/8634/files# failed because of the difference in sort order in postgres.

Or... as I have just discovered, we can use this feature to be explicit about where we want nulls to be, TIL! https://docs.djangoproject.com/en/dev/ref/models/expressions/#using-f-to-sort-null-values

Might be good to use this in the order_by so that we know where we're putting the nulls (although should be less of an issue if using `end_timestamp` as it should never be null in practice).

kolibri/core/logger/evaluation.py
```diff
@@ -0,0 +1,145 @@
...
+    :param user_id: A str referencing user.id
+    :type user_id: str|django.db.models.OuterRef
+    :return: A queryset representing the found tries
+    :rtype: django.db.models.QuerySet
+    """
+    return MasteryLog.objects.filter(
+        summarylog__content_id=content_id,
+        user_id=user_id,
+        complete=True,
+    ).order_by("-completion_timestamp")
```
rtibbles: Shouldn't ever be null in practice if `complete=True`, so this should be safe.

kolibri/core/logger/evaluation.py
```diff
@@ -0,0 +1,145 @@
...
+    :type target_try_attempts: django.db.models.QuerySet
+    :return: A queryset matching previous tries' attempts for each try in `target_try_attempts`
+    :rtype: django.db.models.QuerySet
+    """
+    target_tries = MasteryLog.objects.filter(
+        id__in=target_try_attempts.values_list("masterylog_id", flat=True).distinct(),
+    )
+    return AttemptLog.objects.filter(
+        masterylog_id__in=find_previous_tries(target_tries).values_list("id"),
+    ).order_by("-completion_timestamp")
```
rtibbles: Again, I think `end_timestamp` would be safer here.

rtibbles (commented):
I think the main thing here will be just switching AttemptLog ordering from completion_timestamp to end_timestamp - everything else should just drop out of that.

kolibri/core/auth/api.py
bjester: Found it

kolibri/core/logger/api.py
bjester: It's already testing with multiple attemptlogs per user, just nothing with null timestamps.

rtibbles (approved):
All makes sense to me!

Issue #8091, opened 2021-05-18
Allow a facility to be dissolved and reconstituted


jamalex:
We've seen many cases of devices being cloned without first running the `deprovision` management command, hence duplicating the facility to each device, which leads to issues when it comes time to sync. Currently there is no way to recover from this situation and split the data out into separate facilities, so the only approach currently is to wipe one of the devices and start fresh, or never sync.

A possible flow for addressing this would be creating a management command, perhaps called `fork_facility`, that does the following:
- Creates a new facility B on the device, with the same metadata and settings as facility A
- One model at a time, starting from the root of the foreign-key chain, take chunks of records, update their `_morango_partition` and possibly also `_morango_source_id` (for those that include IDs of other records in them) and then recalculating and setting their `id`, then doing a bulk creation across the record batch.
- Remove the old facility once all that is done, including from the morango Store

Two specific places this could be leveraged, in addition to manually running the management command:
- When a user is renaming a facility, ask them if their intention is actually to make this into a new facility, or just correct the name. If they want it to be a new facility, trigger the forking.
- Automatically run this when a device is provisioned and then cloned! When a device starts up, if a) the instance_id has changed, and b) the device has never been synced (no records in the Store), and c) the facility doesn't have any/much user data.

rtibbles:
Closed by #8742 - follow up filed here: https://github.com/learningequality/kolibri/issues/8824

PR #8742, opened 2021-11-18
Allow facility forking and recreation

### LINKED ISSUE #8435: If the device settings cache gets stale with a non-existent facility, Kolibri does not recover
## Observed behavior
If the Facility that is stored in the device settings cache as the default facility has been deleted, multiple places that attempt to access this information can fall into unrecoverable error states.

## Errors and logs

```
ERROR 2021-09-14 15:42:23,096 django.request Internal Server Error: /api/auth/session/current/
  File "/home/kivy/build/android/kolibri/dist/django/db/models/fields/related_descriptors.py", line 178, in __get__
AttributeError: 'DeviceSettings' object has no attribute '_default_facility_cache'

During handling of the above exception, another exception occurred:

  File "/home/kivy/build/android/kolibri/dist/django/core/handlers/exception.py", line 41, in inner
  File "/home/kivy/build/android/kolibri/dist/django/core/handlers/base.py", line 187, in _get_response
  File "/home/kivy/build/android/kolibri/dist/django/core/handlers/base.py", line 185, in _get_response
  File "/home/kivy/build/android/kolibri/dist/django/views/decorators/csrf.py", line 58, in wrapped_view
  File "/home/kivy/build/android/kolibri/dist/rest_framework/viewsets.py", line 116, in view
  File "/home/kivy/build/android/kolibri/dist/django/utils/decorators.py", line 67, in _wrapper
  File "/home/kivy/build/android/kolibri/dist/django/utils/decorators.py", line 149, in _wrapped_view
  File "/home/kivy/build/android/kolibri/dist/django/utils/decorators.py", line 63, in bound_func
  File "/home/kivy/build/android/kolibri/dist/rest_framework/views.py", line 495, in dispatch
  File "/home/kivy/build/android/kolibri/dist/rest_framework/views.py", line 455, in handle_exception
  File "/home/kivy/build/android/kolibri/dist/rest_framework/views.py", line 492, in dispatch
  File "/home/kivy/build/android/kolibri/core/auth/api.py", line 759, in update
  File "/home/kivy/build/android/kolibri/core/auth/api.py", line 765, in get_session_response
  File "/home/kivy/build/android/kolibri/core/auth/models.py", line 522, in session_data
  File "/home/kivy/build/android/kolibri/core/auth/models.py", line 1311, in get_default_facility
  File "/home/kivy/build/android/kolibri/core/device/utils.py", line 29, in get_device_setting
  File "/home/kivy/build/android/kolibri/dist/django/db/models/fields/related_descriptors.py", line 184, in __get__
  File "/home/kivy/build/android/kolibri/dist/django/db/models/fields/related_descriptors.py", line 159, in get_object
  File "/home/kivy/build/android/kolibri/dist/django/db/models/query.py", line 378, in get
kolibri.core.auth.models.DoesNotExist: Facility matching query does not exist.
```

## Expected behavior

Kolibri should catch this error, clear the stale cache, and recover.

## User-facing consequences
Seemingly permanent breakage of Kolibri.

## Steps to reproduce
Delete a facility, having already accessed the default facility from the Device Settings.

rtibbles:
## Summary
* Fixes a possible bug in user merging, by ensuring that serialized data is passed through db_to_value
* Adds a utility for cloning a facility and all its data
* Refactors some auth app utils into separate modules by function
* Updates the deletefacility command to delete the morango `DeletedModels` model and the new `Bookmark` model
* Uses the cloning facility and the updated delete facility functions to add a migrate_facility function and recreatefacility management command to allow the migration of a facility to a fresh start

## References
Fixes all but the automated parts of: #8091
Fixes #8435

## Reviewer guidance
Anything I have failed to copy?
Any morango subtleties I might have missed?

----

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/core/auth/management/commands/recreatefacility.py
+++ b/kolibri/core/auth/management/commands/recreatefacility.py
@@ -1,0 +1,55 @@
+import logging
+
+from django.db import transaction
+
+from kolibri.core.auth.management.utils import confirm_or_exit
+from kolibri.core.auth.management.utils import get_facility
+from kolibri.core.auth.utils.migrate import migrate_facility
+from kolibri.core.tasks.management.commands.base import AsyncCommand
+
+
+logger = logging.getLogger(__name__)
+
+
+class Command(AsyncCommand):
+    help = "This command initiates the recreation process for a facility and all of its related data."
+
+    def add_arguments(self, parser):
+        parser.add_argument(
+            "--facility",
+            action="store",
+            type=str,
+            help="The ID of the facility to recreate",
+        )
+        parser.add_argument("--noninteractive", action="store_true")
+
+    def handle_async(self, *args, **options):
+        noninteractive = options["noninteractive"]
+        facility = get_facility(
+            facility_id=options["facility"], noninteractive=noninteractive
+        )
+        dataset_id = facility.dataset_id
+
+        logger.info(
+            "Found facility {} <{}> for recreation".format(facility.id, dataset_id)
+        )
+
+        if not noninteractive:
+            # ensure the user REALLY wants to do this!
+            confirm_or_exit(
+                "Are you sure you wish to permanently recreate this facility? This will TRANSFER ALL FACILITY DATA TO A NEW FACILITY."
+            )
+            confirm_or_exit(
+                "ARE YOU SURE? If you do this, there is no way to recover the original facility data."
+            )
+
+        logger.info(
+            "Proceeding with facility recreation. Recreating all data for facility <{}>".format(
+                dataset_id
+            )
+        )
+
+        with transaction.atomic():
+            migrate_facility(facility)
+
+        logger.info("Recreation complete.")
--- a/kolibri/core/auth/test/test_utils.py
+++ b/kolibri/core/auth/test/test_utils.py
@@ -21,8 +21,8 @@ from kolibri.core.auth.test.test_api import ClassroomFactory
 from kolibri.core.auth.test.test_api import FacilityFactory
 from kolibri.core.auth.test.test_api import FacilityUserFactory
 from kolibri.core.auth.test.test_api import LearnerGroupFactory
-from kolibri.core.auth.utils import fork_facility
-from kolibri.core.auth.utils import merge_users
+from kolibri.core.auth.utils.migrate import fork_facility
+from kolibri.core.auth.utils.migrate import merge_users
 from kolibri.core.logger import models as log_models
 
 
--- a/kolibri/core/auth/test/test_utils.py
+++ b/kolibri/core/auth/test/test_utils.py
@@ -13,8 +13,15 @@ from django.test import TestCase
 
 from ..models import Facility
 from kolibri.core.auth.management import utils
+from kolibri.core.auth.models import AdHocGroup
+from kolibri.core.auth.models import Classroom
+from kolibri.core.auth.models import FacilityUser
+from kolibri.core.auth.models import LearnerGroup
+from kolibri.core.auth.test.test_api import ClassroomFactory
 from kolibri.core.auth.test.test_api import FacilityFactory
 from kolibri.core.auth.test.test_api import FacilityUserFactory
+from kolibri.core.auth.test.test_api import LearnerGroupFactory
+from kolibri.core.auth.utils import fork_facility
 from kolibri.core.auth.utils import merge_users
 from kolibri.core.logger import models as log_models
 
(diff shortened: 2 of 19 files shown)


kolibri/core/auth/management/commands/recreatefacility.py
```diff
@@ -0,0 +1,55 @@
...
+            "Found facility {} <{}> for recreation".format(facility.id, dataset_id)
+        )
+
+        if not noninteractive:
+            # ensure the user REALLY wants to do this!
+            confirm_or_exit(
+                "Are you sure you wish to permanently recreate this facility? This will TRANSFER ALL FACILITY DATA TO A NEW FACILITY."
+            )
+            confirm_or_exit(
+                "ARE YOU SURE? If you do this, there is no way to recover the original facility data."
```
jamalex, line 43: I wonder if "the original facility" might be clearer. It kind of depends on how we... define identity of data, I guess.

kolibri/core/auth/management/commands/recreatefacility.py
```diff
@@ -0,0 +1,55 @@
...
+        logger.info(
+            "Proceeding with facility recreation. Recreating all data for facility <{}>".format(
+                dataset_id
+            )
+        )
+
+        with transaction.atomic():
+            migrate_facility(facility)
+
+        logger.info("Recreation complete.")
```
jamalex, line 55: ⚽ 

Now, we must return to work!

kolibri/core/auth/test/test_utils.py
```diff
@@ -481,3 +491,174 @@ def test_contentsummarylogs(self):
...
+
+
+class ForkFacilityTestCase(TestCase):
+    @classmethod
+    def setUpTestData(cls):
+        cls.facility = FacilityFactory.create()
+        content_identifiers = [(uuid.uuid4().hex, uuid.uuid4().hex) for i in range(3)]
+        ex_identifiers = (uuid.uuid4().hex, uuid.uuid4().hex)
+        cls.count = 2
+        for i in range(1, 1 + cls.count):
```
jamalex, line 512: Optional: would read more clearly to me as `cls.count + 1`.

kolibri/core/auth/utils/delete.py
```diff
@@ -0,0 +1,305 @@
...
+def _get_morango_models(dataset_id):
+    querysets = [DatabaseMaxCounter.objects.filter(partition__startswith=dataset_id)]
+
+    stores = Store.objects.filter(partition__startswith=dataset_id)
+    store_ids = stores.values_list("pk", flat=True)
+
+    for store_ids_chunk in chunk(list(store_ids), 300):
+        querysets.append(
+            RecordMaxCounter.objects.filter(store_model_id__in=store_ids_chunk)
+        )
```
jamalex, line 249: Note that these would get cascade deleted from deleting the Store records.

kolibri/core/auth/utils/delete.py
```diff
@@ -0,0 +1,305 @@
...
+            ExamAttemptLog.objects.filter(dataset_id_filter),
+            ExamLog.objects.filter(dataset_id_filter),
+            MasteryLog.objects.filter(dataset_id_filter),
+            UserSessionLog.objects.filter(dataset_id_filter),
+        ],
+    )
+
+
+def _get_morango_models(dataset_id):
+    querysets = [DatabaseMaxCounter.objects.filter(partition__startswith=dataset_id)]
```
jamalex, line 241: There's a case we should double-check here (I know @bjester and I dealt with it somewhere) around the records created on this device itself (which will be a common case). I believe it saves a DatabaseMaxCounter with empty partition for its own instance_id, given that the local device, by definition (except with this sort of removal), has all of the data that it produced itself. I believe we dealt with that by converting the ones with empty partitions into ones with the dataset prefixes for other facilities on the device, or similar. Will come back to this.

kolibri/core/auth/utils/migrate.py
```diff
@@ -0,0 +1,244 @@
...
+                    field_obj = LogModel._meta.get_field(field)
+                    if hasattr(field_obj, "from_db_value"):
+                        value = field_obj.from_db_value(value, None, None, None)
+                except FieldDoesNotExist:
+                    pass
+                setattr(new_log, field, value)
+            if not new_log._morango_source_id:
+                new_log.id = new_log.calculate_uuid()
+            else:
+                # Have to do this, otherwise morango will overwrite the current source id on the model
```
jamalex, line 111: Until https://github.com/learningequality/morango/issues/141 is resolved, at least!

jamalex (changes requested):
Looks great! Super thorough testing, too. :O

Most comments are small, but one thing we need to look into a bit (regarding DatabaseMaxCounters).

One thing I wonder about, as we add things like this that are tightly coupled to the specific set of models we currently have, is how we could add a test that specifically catches cases where new models have been added, so we can make sure we update this code to handle them. One option I was thinking would be to just enumerate the set of syncable models and assert that they are an exact set. And then in that test's docstring, we provide pointers to the various places around the code that would need updating if a new model is added.

This snippet can be used to enumerate the current set of syncable models:
```
from morango.registry import syncable_models
all_models = syncable_models.get_models("facilitydata")
```

kolibri/core/auth/utils/delete.py
rtibbles: This code isn't actually new in this PR, I just moved it over from the original deletefacility command code.

The line below for `DeletedModels` is the new one - but it seems reasonable to do this out of an abundance of caution!

These querysets are also used to count the total number of records to be deleted.

kolibri/core/auth/utils/delete.py
bjester: I don't recall that

kolibri/core/auth/utils/migrate.py
bjester: https://github.com/learningequality/morango/pull/142

rtibbles:
This should be ready for re-review!

kolibri/core/auth/utils/delete.py
jamalex: I looked into this a bit further. The scenario of concern relates to when we serialize from the app models into the Store, and then update the DatabaseMaxCounter values for our own instance ([here](https://github.com/learningequality/morango/blob/master/morango/sync/operations.py#L258)). We used to do unfiltered serialization (serialize everything, even if it doesn't match the filter), which means that the code linked above would set the DatabaseMaxCounter for its own instance_id to `""`, basically saying (correctly) "I have all records, across all partitions, for my own instance" (because, without removal, that's intrinsically true -- if data originated on this instance, then this instance automatically already has it all). If we then removed data that had originated on that instance, but left the `""` counter, it would still think it had everything for that instance, so if we tried to sync data for the removed facility back onto the device, it wouldn't get pulled in.

In practice:
- We shifted in Morango (at some point) to having serialization be filter-specific, rather than serializing everything unconditionally prior to each sync. So the line of code linked above will basically never be hit anymore, because `filter` will always have a value set.
- However, devices that had synced on previous versions may still have those old DatabaseMaxCounter records around. For instance, there are 8 of them on Kolibri Data Portal. So those devices could be impacted if we don't do something here.

What I would propose is that as a separate operation during facility removal, we do the following:
```
# remove any legacy counters with empty partition, and add corresponding counters for remaining facility datasets
for dmc in DatabaseMaxCounter.objects.filter(partition=""):
    for dataset in FacilityDataset.objects.all():
        newdmc, _ = DatabaseMaxCounter.objects.get_or_create(instance_id=dmc.instance_id, partition=dataset.id)
        if newdmc.counter != dmc.counter:
            newdmc.counter = max(newdmc.counter, dmc.counter)
            newdmc.save()
    dmc.delete()
```

rtibbles: This is now implemented.

jamalex (approved):
Looks good to me, thanks!

PR #8746, opened 2021-11-19
Fix `KTextbox` in form vertical shift

### LINKED ISSUE #8675: shifting UI elements cause a 'missed click'
## Observed behavior

When the username field loses keyboard focus it causes a shift in UI elements.

If the user is trying to click a UI element such as the 'guest' link, it can causes an annoying 'missed click' where the UI element moves at just the wrong time:

 

## Expected behavior

In general, UI elements should not shift on the screen unless absolutely necessary.

## User-facing consequences

Frustration

## Steps to reproduce

* refresh login screen
* click somewhere near the top of the 'guest' link

## Context

* 0.15.0-beta1
* Chrome and Firefox
* MacOS

sairina:
## Summary
Fixes the shifting HTML elements in the following views:
- Setup wizard
- Creating a quiz page
- And the two below:

|View|Initial|Change|
|--|--|--|
|User sign-in page|||
|Password box in the user creation page|||

## References
#8675

**Important!** To fix the issue, KDS also needs to be updated in Kolibri. See https://github.com/learningequality/kolibri-design-system/pull/279

## Reviewer guidance
Please review manually in the following components:
- User accounts
- `CreateExamPage`
- `ImportIndividualUserForm` in setup wizard
- `SignInPage`

----

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/core/assets/src/views/userAccounts/PasswordTextbox.vue
+++ b/kolibri/core/assets/src/views/userAccounts/PasswordTextbox.vue
@@ -146,4 +146,10 @@
 </script>
 
 
-<style lang="scss" scoped></style>
+<style lang="scss" scoped>
+
+  /deep/ .ui-textbox-feedback {
+    display: block !important;
+  }
+
+</style>
--- a/kolibri/core/assets/src/views/userAccounts/PasswordTextbox.vue
+++ b/kolibri/core/assets/src/views/userAccounts/PasswordTextbox.vue
@@ -149,6 +149,7 @@
 <style lang="scss" scoped>
 
   /deep/ .ui-textbox-feedback {
+    // Add to fix vertical shifting of textboxes
     display: block !important;
   }
 
--- a/kolibri/plugins/coach/assets/src/views/plan/CreateExamPage/index.vue
+++ b/kolibri/plugins/coach/assets/src/views/plan/CreateExamPage/index.vue
@@ -693,4 +693,8 @@
     margin-left: 15rem;
   }
 
+  /deep/ .ui-textbox-feedback {
+    display: block !important;
+  }
+
 </style>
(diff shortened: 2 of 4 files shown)


indirectlylit:
I'll take a closer look at this today, but an initial question:

Might there be a way to fix this without using `/deep/` or `!imortant`? Both of these tools are risky in the sense that they break the encapsulation pattern of scoped styles. The danger is that any future refactors or updates to the text box CSS in KDS are liable to break parent components that rely on private implementation details.

(See also [these notes on !important](https://developer.mozilla.org/en-US/docs/Web/CSS/Specificity#the_!important_exception))

indirectlylit:
see https://github.com/learningequality/kolibri-design-system/pull/279#discussion_r755339986

I think that preferably the textbox will be a fixed height that always reserves space below for validation. The only potential issue I can think of here is whether there are some designs that require "tighter" spacing. If that's the case, this should probably be a design conversation and a new `KTextbox` parameter like `noSpaceReserved`

sairina:
> The only potential issue I can think of here is whether there are some designs that require "tighter" spacing. If that's the case, this should probably be a design conversation and a new `KTextbox` parameter like `noSpaceReserved`

@indirectlylit Should this be opened as a follow-up issue in KDS?

indirectlylit:
closing in favor of KDS-only changes in https://github.com/learningequality/kolibri-design-system/pull/279


>>   The only potential issue I can think of here is whether there are some designs that require "tighter" spacing. If that's the case, this should probably be a design conversation and a new KTextbox parameter like noSpaceReserved
>
> @indirectlylit Should this be opened as a follow-up issue in KDS?

I'm just speculating here - I don't actually know that there _is_ an issue :) My sense is it's probably not but something to keep an eye out for.

indirectlylit:
closing in favor of KDS-only changes in https://github.com/learningequality/kolibri-design-system/pull/279


>>   The only potential issue I can think of here is whether there are some designs that require "tighter" spacing. If that's the case, this should probably be a design conversation and a new KTextbox parameter like noSpaceReserved
>
> @indirectlylit Should this be opened as a follow-up issue in KDS?

I'm just speculating here - I don't actually know that there _is_ an issue :) My sense is it's probably not but something to keep an eye out for.

PR #8755, opened 2021-11-19
Update KDS reference to point at newest commit

marcellamaki:
To include timer icon for header bar for practice quizzes


sairina (approved):
lgtm!

PR #8765, opened 2021-11-22
Fixes and tests for metadata labels

### LINKED ISSUE #8737: Search - Getting a 500 error after having applied too many filters
## Observed behavior
When searching I'm getting a 500 error after having applied too many filters. The issue could be caused by the fake data used but should be further investigated.

## Expected behavior
Fully functional search functionality. 

1. Install the following [build.](https://github.com/learningequality/kolibri/releases/tag/v0.15.0-beta2)
2. Import some channels and run the following management command: `kolibri manage labeltestdata `
3. Filter by as many criteria as possible

## Logs

[logs.zip](https://github.com/learningequality/kolibri/files/7564984/logs.zip)

## Usage Details
 - OS: Windows 10
 - Browser: Chrome

rtibbles:
## Summary
* Adds extensive testing for searching by metadata labels and the available labels generation code
* This showed that the available metadata labels code was not quite doing what it ought
* Changes how the available metadata labels code produces the available metadata labels to make the tests pass and also improve performance
* Uses a Bit wise OR in Postgres to determine which labels are present
* Implements a Bit wise OR shim for SQLite that functions identically to the postgres Bit wise OR, except that the maximum size of bits must be specified in advance
* Uses a bit wise OR to do the label aggregation in a single query without subqueries, which significantly improves the performance, and gives full test compliance
* Limits the data returned for Channels and Languages in the available label metadata to just the ids (reduces query times for these two specific items by roughly 100x for each)
* Bootstraps full language data into the learn page explicitly to compensate for only the ids (to compensate for the full data no longer coming from the available labels)

Fixes #8737

## Reviewer guidance
Does filtering seem to produce the correct remaining labels?
When testing I changed this line to `if True:` https://github.com/learningequality/kolibri/blob/release-v0.15.x/kolibri/core/content/utils/search.py#L91 to repeatably execute the queries that are being optimized here.

----

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/core/content/utils/search.py
+++ b/kolibri/core/content/utils/search.py
@@ -60,30 +60,14 @@ for key, labels in metadata_lookup.items():
 
 
 def _get_available_languages(base_queryset):
-    from kolibri.core.content.models import Language
-
     lang_ids = (
-        base_queryset.exclude(lang=None)
-        .order_by("lang_id")
-        .values_list("lang_id", flat=True)
+        base_queryset.exclude(lang=None).values_list("lang_id", flat=True).distinct()
     )
-    return list(
-        Language.objects.filter(id__in=lang_ids)
-        .order_by("id")
-        .values("id", "lang_name")
-    )
+    return list(lang_ids)
 
 
 def _get_available_channels(base_queryset):
-    from kolibri.core.content.models import ChannelMetadata
-
-    return list(
-        ChannelMetadata.objects.filter(
-            id__in=base_queryset.values_list("channel_id", flat=True)
-        )
-        .order_by()
-        .values("id", "name")
-    )
+    return list(base_queryset.values_list("channel_id", flat=True).distinct())
 
 
 class SQLiteBitwiseORAggregate(Aggregate):
--- a/kolibri/core/content/utils/search.py
+++ b/kolibri/core/content/utils/search.py
@@ -101,7 +101,7 @@ def get_available_metadata_labels(base_queryset):
         aggregates = {}
         for field in bitmask_fieldnames:
             field_agg = field + "_agg"
-            if connections[base_queryset.db].vendor == "sqlite":
+            if connections[base_queryset.db].vendor == "sqlite" or BitOr is None:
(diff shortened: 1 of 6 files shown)


pcenov:
Tested in this [build ](https://buildkite.com/learningequality/kolibri-python-package/builds/4346) and the 500 error is fixed now plus filtering is fully functional.

kolibri/core/content/utils/search.py
```diff
@@ -4,10 +4,17 @@
 """
 import hashlib
 
+try:
+    from django.contrib.postgres.aggregates import BitOr
+except ImportError:
+    BitOr = None
```
bjester, line 10: Would this ever actually fail to import?

kolibri/core/content/utils/search.py
```diff
@@ -89,27 +97,25 @@ def get_available_metadata_labels(base_queryset, limit_to_known_fields=True):
...
-                    for value in values
-                }
-            )
-            all_values += values
-        result = queryset.values(*all_values).first()
+        base_queryset = base_queryset.order_by()
+        aggregates = {}
+        for field in bitmask_fieldnames:
+            field_agg = field + "_agg"
+            if connections[base_queryset.db].vendor == "sqlite":
```
bjester: With the above `BitOr` potentially being `None`, perhaps this should also include `or BitOr is None`?

bjester (approved):
Couple comments, but looks good!

kolibri/core/content/utils/search.py
jredrejo: yes
`ImportError: No module named psycopg2.extras`
in python 2.7

jredrejo (approved):
I have tested it as much as I have been able (in sqlite only) and it has worked perfectly.
Also, seeing how `pytest kolibri/core/content/test/test_search.py` results in 
`371 passed in 63.65 seconds` it's quite impressive.
Great work!

PR #8795, opened 2021-12-02
Fix learning activity label style on bookmarks cards

### LINKED ISSUE #8725: Bookmark Card - Metadata misalignments
## Observed behavior

These likely have some padding/margin applied that makes sense at larger sizes but not when reflowed for a smaller screen.

Follows up from #8708 

image

## Expected behavior

Text aligns on the left.

## User-facing consequences

It looks off.

MisRob:
## Summary

Fixes learning activity labels on cards on the bookmarks page so that learning activity label, icon, and duration show on one line as specced [in Figma designs](https://www.figma.com/file/uiZuCIqh2KYvBUfbCiJyyA/Hybrid-Learning?node-id=130%3A11706). For that purpose, `condensed` prop has been added to `LearningActivityLabel` component.

| Before | After |
| -------- | ------- |
| Screenshot from 2021-12-02 11-05-22 | Screenshot from 2021-12-02 11-04-57 |

## References

Closes #8725

## Reviewer guidance

Check cards on the bookmarks page for various resolutions, learning activities, content kinds, and their durations.

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/plugins/learn/assets/src/views/HybridLearningContentCardListView.vue
+++ b/kolibri/plugins/learn/assets/src/views/HybridLearningContentCardListView.vue
@@ -32,7 +32,11 @@
           class="metadata-info"
           :style="{ color: $themePalette.grey.v_700 }"
         >
-          <LearningActivityLabel :contentNode="content" class="learning-activity-label" />
+          <LearningActivityLabel
+            :contentNode="content"
+            labelAfter
+            condensed
+          />
         </div>
         <h3 class="title">
           <TextTruncator
@@ -265,13 +269,6 @@
     bottom: 0;
     width: 100%;
     padding: $margin;
-  }
-
-  .learning-activity-label {
-    width: 100px;
-    /deep/ .learning-activity {
-      justify-content: flex-start;
-    }
   }
 
   .metadata-info-footer {
--- a/kolibri/plugins/learn/assets/src/views/cards/ResourceCard/LearningActivityLabel.vue
+++ b/kolibri/plugins/learn/assets/src/views/cards/ResourceCard/LearningActivityLabel.vue
@@ -1,10 +1,10 @@
 <template>
 
-  <div v-if="contentNode">
+  <div v-if="contentNode" :class="[ condensed ? 'condensed' : '']">
     <div class="learning-activity">
       <span
         v-if="!labelAfter"
-        class="label"
+        class="label-before"
         data-test="label"
       >
         {{ label }}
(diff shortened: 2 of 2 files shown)


kolibri/plugins/learn/assets/src/views/HybridLearningContentCardListView.vue
```diff
@@ -32,7 +32,11 @@
           class="metadata-info"
           :style="{ color: $themePalette.grey.v_700 }"
         >
-          <LearningActivityLabel :contentNode="content" class="learning-activity-label" />
+          <LearningActivityLabel
+            :contentNode="content"
+            labelAfter
```
jredrejo, line 37: for consistency in the code, would you  mind to change https://github.com/learningequality/kolibri/blob/release-v0.15.x/kolibri/plugins/learn/assets/src/views/HybridLearningContentCard/index.vue#L22 to use the same syntax?

jredrejo (approved):
Code looks good and testing it, it really looks good to me!

nucleogenesis (approved):
LGTM!

kolibri/plugins/learn/assets/src/views/HybridLearningContentCardListView.vue
MisRob: Thanks, @jredrejo. I'm sorry for the late reply. I checked how we use Boolean props elsewhere, and it seems that across the whole codebase, we use explicit `true` value for many other boolean props, so even if I removed Boolean value from your reference, we'd still have many other places (just search for  `="true"` ) so it wouldn't probably help much overall with consistency. I don't feel strongly opinionated about these two syntaxes. Did you consider syntax without the Boolean value confusing? I could also use explicit `:labelAfter="true"` instead.

jredrejo: Specs don't have a favourite setting, https://html.spec.whatwg.org/multipage/common-microsyntaxes.html#boolean-attribute. From a personal point, I prefer the presence/absence of the attribute to reflect its value. It looks more ellegant and readable for me, but it's a completely personal preference.
In any case, I think it's important to have consistency in the code. For this attribute we're using it only in one more place, so whatever is decided is good for me is it's the same in both parts.

MisRob: Thanks. I appreciate its simplicity too. On the other hand, the advantage of having it explicitly defined might be that it's clear that it's one of Vue's `prop`s.

In any case, I'd say it's a matter of all Boolean attributes rather than this one specific attribute, and since we mostly use explicit syntax across the whole codebase, I think I'll use that.  I'm going to sneak the update in one of my next PRs.

MisRob: Addressed here https://github.com/learningequality/kolibri/pull/8891

PR #8796, opened 2021-12-02
CustomContent: Add withinDescendant parameter to API

danigm:
## Summary

This patch adds a new optional parameter to the getContentByFilter API call to be able to retrieve all descendant nodes using the MPTT model fields.

With this new parameter it's possible to filter all the nodes below a specific topic node, useful when combining with other filters like kind or tags.

### DIFF
--- a/packages/hashi/src/kolibri.js
+++ b/packages/hashi/src/kolibri.js
@@ -114,6 +114,8 @@ export default class Kolibri extends BaseShim {
        * an array of matching metadata
        * @param {Object} options - The different options to filter by
        * @param {string=} options.parent - id of the parent node to filter by, or 'self'
+       * @param {string=} options.descendantOf - id of the root node to filter
+       * by, to show all descendants, not just direct children
        * @param {string[]} options.ids - an array of ids to filter by
        * @param {number} [options.maxResults=50] - the maximum number of nodes per request
        * @param {string[]} options.kinds - an array of kinds to filer by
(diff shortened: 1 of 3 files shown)


packages/hashi/src/kolibri.js
```diff
@@ -123,6 +130,7 @@ export default class Kolibri extends BaseShim {
        * @param {boolean} options.onlyContent - set to true to query only content nodes
        * @param {boolean} [options.limitToChannel=true] - true to limit the
        * results to the topic channel
+       * @param {MPTTNode} options.withinDescendant - Object to filter nodes descendant of this one
```
rtibbles: This requires that the HTML5 apps are aware of the implementation details of MPTT - perhaps it would be better for this to just be the id of the node to fetch descendants of?

We can add a `descendant_of` filter to the backend: https://github.com/learningequality/kolibri/blob/release-v0.15.x/kolibri/core/content/api.py#L250 that does the lookup for the node - something like:
```
def filter_descendant_of(self, queryset, name, value):
    try:
        node = ContentNode.objects.values("lft", "rght", "tree_id").get(pk=value)
    except (ContentNode.DoesNotExist, ValueError):
        return queryset.none()
    return queryset.filter(lft__gt=node["lft"], rght__lt=node["rght"], tree_id=node["tree_id"])
```

danigm: Okay, that's a better API and moving the logic to the backend simplifies the frontend logic. I'll update the PR to follow that implementation

danigm: I've just updated the PR with the new filter and the parameter `descendantOf` that should be a node id.

rtibbles (approved):

Issue #8809, opened 2021-12-06
Content cards with folders shorter than other content cards


sairina:
## Observed behavior
In the learner's library, content cards with folders are slightly shorter in height than the rest.
|without hover shadow|with hover shadow|
|--|--|
|<img width="674" alt="Screen Shot 2021-12-06 at 1 33 10 PM" src="https://user-images.githubusercontent.com/13563002/144926066-9b3bcfd3-5e58-4cce-a621-2c73b5131e00.png">|<img width="674" alt="Screen Shot 2021-12-06 at 1 29 59 PM" src="https://user-images.githubusercontent.com/13563002/144925828-653c2f38-0d26-4323-a328-00fad2fd54a8.png">|

## Errors and logs
None

## Expected behavior
The cards should be the same height.

## User-facing consequences
Just a bit off.

## Steps to reproduce
1. Log in as a learner
2. Click on "Library" tab
3. On left sidebar, go to Category > "None of the above" to view everything (including folders)
4. See that cards with folders are slightly shorter

## Context
- Kolibri v0.15
- Chrome

rtibbles:
Fixed in #8856

PR #8856, opened 2021-12-07
Fix height on content cards with folders

sairina:
## Summary
Update CSS to fix the height and text size on content cards with folders so that it matches the text size and height of the other content cards.

|Before|After|
|--|--|
|Screen Shot 2021-12-06 at 1 33 10 PM|Screen Shot 2021-12-07 at 1 48 42 PM|
## References
Addresses #8809

## Reviewer guidance
Double check that the heights now match that of the content cards with activities:
1. Log in as a learner
2. Click on "Library" tab
3. On left sidebar, go to Category > "None of the above" to view everything (including folders)
4. See that cards with folders are the same height as other cards.

----

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/plugins/learn/assets/src/views/HybridLearningContentCard/index.vue
+++ b/kolibri/plugins/learn/assets/src/views/HybridLearningContentCard/index.vue
@@ -135,7 +135,6 @@
       },
       footerLength() {
         return (
-          1 +
           this.content.is_leaf +
           (this.isUserLoggedIn && !this.isLearner && this.content.num_coach_contents) +
           (this.content.num_coach_contents > 0) +
@@ -211,8 +210,8 @@
   .folder-header-text {
     display: inline-block;
     padding: 0;
-    margin: 0;
-    font-size: 16px;
+    margin: 7px 0;
+    font-size: 13px;
   }
 
   .k-labeled-icon {
--- a/kolibri/plugins/learn/assets/src/views/HybridLearningContentCard/index.vue
+++ b/kolibri/plugins/learn/assets/src/views/HybridLearningContentCard/index.vue
@@ -192,6 +192,7 @@
   .header-bar {
     display: flex;
     justify-content: space-between;
+    height: 48px;
     padding: 8px 16px;
     font-size: 13px;
     .channel-logo {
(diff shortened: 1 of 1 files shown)


kolibri/plugins/learn/assets/src/views/HybridLearningContentCard/index.vue
```diff
@@ -211,8 +210,8 @@
   .folder-header-text {
     display: inline-block;
     padding: 0;
-    margin: 0;
-    font-size: 16px;
+    margin: 7px 0;
```
MisRob: ~~Not familiar with designs but this looks like something where we could follow 8px increments guidelines?~~

^ Can be ignored. I didn't realize that it's the precise value needed to keep headers the same height.

kolibri/plugins/learn/assets/src/views/HybridLearningContentCard/index.vue
```diff
@@ -135,7 +135,6 @@
       },
       footerLength() {
         return (
-          1 +
```
MisRob, line 138: Why was this update of the progress bar max width calculation needed to fix cards heights?

kolibri/plugins/learn/assets/src/views/HybridLearningContentCard/index.vue
indirectlylit: This is a somewhat brittle way to make the header bar heights consistent because it is tightly coupled to the font size, so any change there will break the height again.

A more robust fix would be to set a fixed height (e.g. `48px`) on `.header-bar`. In this way we decouple container height from contents to a degree.

kolibri/plugins/learn/assets/src/views/HybridLearningContentCard/index.vue
sairina: From my understanding of what @rtibbles told me, there may have been a three-dots button for a menu in the footer that wasn't conditional, so the `1` was there to provide some space for that. However, that element was removed, but the `1` did not get removed.

MisRob: Ah okay, sounds like an additional cleanup not related to the height fix. Thank you.

MisRob (approved):
Thank you, @sairina

PR #8859, opened 2021-12-07
Remove unnecessary cruft from Bookmark.infer_dataset

jamalex:
Quick flyby cleanup: it looks as if the code here was copied from the logger models, but we can keep it simpler here. The logs can be anonymous (no user), so we need to jump through more hoops to decide on a facility to associate them with, whereas bookmarks require that the user be specified.


rtibbles (approved):

rtibbles:
Test failure is due to still stubbornly flaky test :/

Issue #8838, opened 2021-12-07
Remove top bar search on library page


marcellamaki:
To avoid confusion/conflict between search bars

https://www.notion.so/learningequality/Feedback-Learn-Library-Search-dfe09aeae356463bbaf753692de2e480

rtibbles:
Fixed in #8862

PR #8862, opened 2021-12-07
Removes search box from top navbar when user is in library

sairina:
## References
Addresses #8838 

## Reviewer guidance
Check to see that the top search box is removed when a user is in the library, and that it still works as intended on other pages.

----

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/plugins/learn/assets/src/views/LearnIndex.vue
+++ b/kolibri/plugins/learn/assets/src/views/LearnIndex.vue
@@ -132,6 +132,9 @@
       currentPage() {
         return pageNameToComponentMap[this.pageName] || null;
       },
+      pageName() {
+        return this.$route.name;
+      },
       currentPageIsTopic() {
         return [
           pageNameToComponentMap[PageNames.TOPICS_TOPIC],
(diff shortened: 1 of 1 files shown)


kolibri/plugins/learn/assets/src/views/LearnIndex.vue
```diff
@@ -132,6 +132,9 @@
       currentPage() {
         return pageNameToComponentMap[this.pageName] || null;
       },
+      pageName() {
```
rtibbles: This computed prop is colliding with `...mapState(['pageName']),` above - personally, I think your computed prop is far more reliable here, as I have seen some funky behaviour with the vuex store page name not being what it ought to be.

Should remove the mapState reference though, and will also need to update the tests that rely on mocking `pageName` in vuex state in order to test functionality.

rtibbles (approved):
This works as intended, just needs the tests to pass and we're good to go!

kolibri/plugins/learn/assets/src/views/LearnIndex.vue
sairina: I definitely missed that! Based on the conversations on Slack, it seems like we can just remove that `<ActionBarSearchBox />` component completely from the Learn index, so this should be okay.

sairina:
@jtamiace 
This is how it currently looks - does it work?

rtibbles:
If @jtamiace is good with this, we can go ahead and merge!

jtamiace:
Looks great! Thank you @sairina

PR #8878, opened 2021-12-08
add core validator generator for Object-type props

### LINKED ISSUE #8640: add core validator generator for Object-type props
## Observed behavior

There are a large number of `Object`-type props with no validator function. (Search the codebase using a multi-line regex to help find them: `{[^}]*type: Object[^}]*}`)

Normal component props are well-documented using `type`, `required`, and sometimes `validator`. The keys of `Object` props are as much a part of the component API as the other ones, but they are often under-documented.

Even when validation of `Object` props is done in Kolibri, it is performed in an ad-hoc manner with little consistency and the logging is less useful because usually Vue will simply say the validator failed without specifying why. Some examples:

https://github.com/learningequality/kolibri/blob/c3238a85915a75386150c5f1191de6735e495d1c/kolibri/core/assets/src/views/sync/FacilityAdminCredentialsForm.vue#L54-L60

https://github.com/learningequality/kolibri/blob/c3238a85915a75386150c5f1191de6735e495d1c/kolibri/plugins/setup_wizard/assets/src/views/importFacility/SelectFacilityForm.vue#L67-L73

https://github.com/learningequality/kolibri/blob/c3238a85915a75386150c5f1191de6735e495d1c/kolibri/plugins/learn/assets/src/views/EmbeddedSidePanel/SelectGroup.vue#L72-L79

## Expected behavior

* The expected schema of `Object` props should be documented and enforced similar to other props. This includes the ability to specify `type`, `required`, and `validator` for object properties
* Logging from failed validators should produce helpful results for developers
* Object validators should be easy to construct using syntax similar to normal prop definitions
* Nice-to-have: add a new linting rule which requires that all `Object`-type props have a validator function

## User-facing consequences

* Developers who want to leverage `Object`-based props in new components do not have sufficient support from our built-in tooling for quickly writing effective and useful validators
* Developers who are working on existing components that use `Object`-based props do not have as much insight and assurance into what data might exist in the object without looking at parent components that pass the data in

## Proposal

Add a new core function `objectValidator` to `kolibri.utils.validators`.

* **Input:** takes an object similar to the Vue `props` object, where each key is the name of an expected property, and each value may contain information about the `type`, `required`, and `validator` for that property.
* **Output:** returns a new validator function which itself takes an object as an input and returns `true` or `false` depending on whether the object conforms to the schema. Additionally, when the value is `false` the function should log a helpful error message specific to the mode of failure.

Then this can be used to define validators on `Object`-type properties. For example:

```js
    props: {
      user: {
        type: Object,
        required: true,
        validator: objectValidator({
          name: {
            type: String,
            required: true,
          },
          age: {
            type: Number,
            required: true,
            validator(value) {
              return value > 0;
            },
          },
        }),
      },
    },
```

## Context

0.15

hamzamunaf:
## Summary
Adding Core Validator Generator for Object Type Props

…

## References
Closes #8640 

…

## Reviewer guidance
Use the ObjectPropValidator function like this: 

			var users = {

				"props": {

					"user": {
						"name": "Hamza Munaf",
						"age": 24
					}

				}

			};

			//
			objectPropValidator( { props: {
					user: {
						type: Object,
						required: true,
						validator: objectPropValidator({
							name: {
								type: String,
								required: true,
							},
							age: {
								type: Number,
								required: true,
								validator(value) {
									return value > 0;
								},
							},
						}),
					},
				}
			} )(users);

…

----

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/core/assets/src/validators.js
+++ b/kolibri/core/assets/src/validators.js
@@ -25,3 +25,195 @@ export function validateLearningActivity(arr) {
   const isValidLearningActivity = v => Object.values(LearningActivities).includes(v);
   return arr.length > 0 && arr.every(isValidLearningActivity);
 }
+export function objectPropValidator( options ){
+	var args = arguments || [];
+    var _getType = function(variable) {
+        var match = false;
+        var rawType = "";
+        try {
+            match = variable && variable.toString().match(/^\s*function (\w+)/);
+        } catch {}
+        try {
+            rawType = Object.prototype.toString.call(variable).slice(8, -1);
+        } catch {}
+        return match ? match[1] : (rawType.length > 0 ? rawType : ((typeof variable == "function" || typeof variable.name != "undefined") ? variable.name : ''));
+    };
+    var _isUndefined = function(variable) {
+        return (typeof variable === "undefined");
+    };
+    var _isSet = function(variable) {
+        return (typeof variable !== "undefined");
+    };
+    var _isString = function(variable) {
+        return (typeof variable === "string");
+    };
+    var _isNumber = function(variable) {
+        return (typeof variable === "number");
+    };
+    var _isFinite = function(variable) {
+        return isFinite(variable);
+    };
+    var _isInfinity = function(variable) {
+        return (typeof variable === "number" && !isFinite(variable));
+    };
+    var _isBoolean = function(variable) {
+        return (typeof variable === "boolean");
+    };
+    var _isBigInt = function(variable) {
+        return (typeof variable === "bigint");
+    };
+    var _isFlatten = function(variable) {
+        return (typeof variable === "string" || typeof variable === "number" || typeof variable === "boolean");
+    };
+    var _isEmpty = function(variable) {
+        if (typeof variable === 'undefined' || variable === "" || variable === 0 || variable === [] || variable === {} || variable === false || variable === null || isNaN(variable) || variable === function() {}) {
+            return true;
+        }
+        return false;
+    };
+    var _isArray = function(variable) {
+        if (typeof Array != "undefined" && typeof Array.isArray != "undefined") {
+            return Array.isArray(variable);
+        }
+        if (variable.constructor && variable.constructor.name && variable.constructor.name == "Array") {
+            return true;
+        }
+        if (typeof Object != "undefined" && Object.prototype && Object.prototype.toString && Object.prototype.toString.call && Object.prototype.toString.call(variable) == "[object Array]"); {
+            return true;
+        }
+        return false;
+    };
+    var _inArray = function(needle, haystack) {
+        if (typeof haystack.indexOf != "undefined") {
+            return haystack.indexOf(needle) > -1;
+        }
+        return false;
+    };
+    var _isFunction = function(variable) {
+        return (typeof variable === "function");
+    };
+    var _isObject = function(variable) {
+        return (typeof variable === "object");
+    };
+    var _isPureObject = function(variable) {
+        return (typeof variable === "object" && typeof Object != "undefined" && variable === Object(variable));
+    };
+    var _isSymbol = function(variable) {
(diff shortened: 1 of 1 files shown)


kolibri/core/assets/src/validators.js
```diff
@@ -25,3 +25,195 @@ export function validateLearningActivity(arr) {
   const isValidLearningActivity = v => Object.values(LearningActivities).includes(v);
   return arr.length > 0 && arr.every(isValidLearningActivity);
 }
+export function objectPropValidator( options ){
```
rtibbles, line 28: Would be good to have a test suite to demonstrate the validation happening here.

kolibri/core/assets/src/validators.js
```diff
@@ -25,3 +25,195 @@ export function validateLearningActivity(arr) {
...
+        var rawType = "";
+        try {
+            match = variable && variable.toString().match(/^\s*function (\w+)/);
+        } catch {}
+        try {
+            rawType = Object.prototype.toString.call(variable).slice(8, -1);
+        } catch {}
+        return match ? match[1] : (rawType.length > 0 ? rawType : ((typeof variable == "function" || typeof variable.name != "undefined") ? variable.name : ''));
+    };
+    var _isUndefined = function(variable) {
```
rtibbles, line 41: I would suggest using lodash helper functions here e.g. https://lodash.com/docs/4.17.15#isBoolean instead of defining your own functions here.

kolibri/core/assets/src/validators.js
```diff
@@ -25,3 +25,195 @@ export function validateLearningActivity(arr) {
...
+        return (typeof variable === "undefined");
+    };
+    var _isSet = function(variable) {
```
rtibbles, line 44: This function name is confusing, as `Set` is an ES6 native type.

kolibri/core/assets/src/validators.js
```diff
@@ -25,3 +25,195 @@ export function validateLearningActivity(arr) {
   const isValidLearningActivity = v => Object.values(LearningActivities).includes(v);
   return arr.length > 0 && arr.every(isValidLearningActivity);
 }
+export function objectPropValidator( options ){
+	var args = arguments || [];
```
rtibbles, line 29: Preferable to use ES6 `const` and `let` declarations.

kolibri/core/assets/src/validators.js
rtibbles, line 28: I think overall this could be simplified and cleaned up with a type map:
```
import isNumber from 'lodash/isNumber';
import isBoolean from 'lodash/isBoolean';

const typeValidatorMap = {
  Number: isNumber,
  Boolean: isBoolean,
}
```
This is an example, but could be extended to all supported types.

Then when going to check can just do:
```
typeValidatorMap[option.type](data);
```

rtibbles (commented):
A good start, a test suite would help to demonstrate that the validator is working as intended. Some cleanup and simplification can be done too.

indirectlylit:
Minor request: could we please rename `objectPropValidator` to just `objectValidator`? I have a use-case for this which is not a prop.

indirectlylit:
also, please check the linting and formatting:

https://kolibri-dev.readthedocs.io/en/develop/getting_started.html#linting-and-auto-formatting

indirectlylit:
Hi @hamzamunaf – thank you very much for your work!

I wanted to move forward with using the results in some ongoing refactoring I'm doing, so I took your commit and integrated it into a new PR here along with the above feedback from @rtibbles: https://github.com/learningequality/kolibri/pull/8902, so I'll close this as superseded. Your review of the new PR would be appreciated.

I've also opened a new follow-up issue here: https://github.com/learningequality/kolibri/issues/8903 – if you're interested in continuing, your help there would be appreciated!

PR #8879, opened 2021-12-08
Bookmark font size

### LINKED ISSUE #8844: Metadata font sizes on bookmarks page cards (below thumbnail) are incorrect
Update to ensure fonts match figma spec 
https://www.notion.so/learningequality/FEEDBACK-Learn-Bookmarks-List-80103926efec4dcb8d6e1a1467234278

Untitled (1)

sairina:
## Summary
Some UI updates to make the bookmark cards more closely aligned to [spec in Figma](https://www.figma.com/file/uiZuCIqh2KYvBUfbCiJyyA/Hybrid-Learning?node-id=76%3A6747).
- Moved the right icons into their own separate row, per spec
- Updated size of thumbnail
- Updated margins for card

Screen Shot 2021-12-08 at 12 05 04 PM

## References
Fixes #8844 

## Reviewer guidance

- [ ] Check to see if bookmarked cards now more closely match the specs
- [ ] Check to see if there is brittleness in CSS that should be fixed
- [ ] Check for regressions

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/plugins/learn/assets/src/views/HybridLearningContentCardListView.vue
+++ b/kolibri/plugins/learn/assets/src/views/HybridLearningContentCardListView.vue
@@ -268,7 +268,7 @@
     position: absolute;
     bottom: 0;
     width: 100%;
-    padding: $margin 24px;
+    padding: 16px 24px;
   }
 
   .metadata-info-footer {
--- a/kolibri/plugins/learn/assets/src/views/HybridLearningContentCardListView.vue
+++ b/kolibri/plugins/learn/assets/src/views/HybridLearningContentCardListView.vue
@@ -275,7 +275,7 @@
     flex: auto;
     align-self: center;
     margin: 0;
-    font-size: 14px;
+    font-size: 13px;
   }
 
   .footer-elements {
@@ -307,7 +307,6 @@
   .thumbnail {
     display: inline-block;
     width: 240px;
-    height: 100px;
     margin-top: 8px;
     margin-right: 24px;
     margin-left: 24px;
--- a/kolibri/plugins/learn/assets/src/views/HybridLearningContentCardListView.vue
+++ b/kolibri/plugins/learn/assets/src/views/HybridLearningContentCardListView.vue
@@ -69,18 +69,20 @@
         />
       </span>
     </router-link>
-    <KFixedGrid :numCols="10" class="footer">
-      <KFixedGridItem span="3" class="footer-elements">
-        <p
-          v-if="isBookmarksPage"
-          class="metadata-info-footer"
-          :style="{ color: $themePalette.grey.v_700 }"
-        >
-          {{ bookmarkCreated }}
-        </p>
-        <ProgressBar :contentNode="content" />
-      </KFixedGridItem>
-      <KFixedGridItem span="10" alignment="right" class="footer-elements footer-icons">
+    <div class="footer">
+      <div class="footer-elements">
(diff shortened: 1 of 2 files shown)


kolibri/plugins/learn/assets/src/views/HybridLearningContentCardListView.vue
```diff
@@ -268,24 +268,20 @@
     position: absolute;
     bottom: 0;
     width: 100%;
-    padding: $margin;
+    padding: $margin 24px;
```
indirectlylit: Is there a reason to keep `$margin` around? It doesn't appear to be used except here.

We can delete it and just do:


```suggestion
    padding: 16px 24px;
```

kolibri/plugins/learn/assets/src/views/HybridLearningContentCardListView.vue
```diff
@@ -70,7 +70,7 @@
       </span>
     </router-link>
     <KFixedGrid :numCols="10" class="footer">
-      <KFixedGridItem span="6" class="footer-elements">
+      <KFixedGridItem span="3" class="footer-elements">
```
indirectlylit: This is a bit confusing because `numCols` is 10 but the grid item spans are 3 and 10, which don't add up.

If these items are always on separate rows, we don't need the grid at all and can just use two divs.

indirectlylit (approved):
no blocking changes, but some cleanup would be nice for future editors

kolibri/plugins/learn/assets/src/views/HybridLearningContentCardListView.vue
```diff
@@ -278,13 +278,18 @@
...
+
+  .footer-progress {
+    width: 240px;
   }
 
   .footer-icons {
     // this override fixes an existing KDS bug with
     // the hover state circle being squished
     // and can be removed upon that hover state fix
+    float: right;
```
indirectlylit, line 292: two issues here:

1. the new line is inserted below a comment that describes something completely unrelated! that will be very confusing for future editors.
2. `float: right` is a very brittle mechanism for doing layout. Instead, it's usually better to use one or more of `position: absolute; right: 0`, `text-align: right`, `KGrid`, or flexbox

indirectlylit (changes requested):
Today I looked closer at both the changes in this PR and also the preexisting code and a few things raised some flags.

I went back to check [the Figma spec](https://www.figma.com/file/uiZuCIqh2KYvBUfbCiJyyA/Hybrid-Learning?node-id=76%3A6747), and unfortunately it isn't explicit about how the grid layout within the card is supposed to work. This is something that future KDS reviews should hopefully help address.

Looking at it though, my impression was that the intent is something like this:

image

However the actual DOM structure looks like this:



One serious user-facing consequence of this is that the whole bottom-half of the card looks like it should be clickable but is not. This will cause especially tablet users a ton of frustration:



I think the overall layout of this component needs to be cleaned up. If it would be helpful, I'd be happy to take this on in order to demonstrate how `KGrid` can be useful here.

kolibri/plugins/learn/assets/src/views/HybridLearningContentCardListView.vue
indirectlylit: now that `$margin` is completely unused, the definition should be deleted

PR #8892, opened 2021-12-10
html5_viewer: Allow fullscreen in iframe

danigm:
## Summary
The T2 games (unity games) that we've on the Endless Key have a fullscreen button that don't work without the `allow="fullscreen"`. Adding this option to the iframe fixes the issue.


rtibbles (approved):

rtibbles:
This has no implications for the sandbox and shouldn't interfere with our own full screen button.

PR #8898, opened 2021-12-11
Encapsulate progress tracking logic in a composable to prevent content pages crossing their streams

### LINKED ISSUE #8776: Rapidly navigating back and forth between two content pages causes all currently pending requests to be cancelled
## Observed behavior
Rapidly pressing the forward and back button between two content pages seems to cause Kolibri to think it has very briefly been disconnected and for all requests to be cancelled. Not entirely clear why this happens, but may be an axios or xhr behaviour surfacing oddly?

kolibribackandforth

## Errors and logs
The only errors are the cascade of errors caused by the cavalcade of cancelled requests.

## Expected behavior
Should be able to navigate back and forth as you please.

Ideally we should delay navigation until all pending requests on the ContentPage have resolved.

## User-facing consequences
This seems like an edge case, but seems like an extension of the concerns raised in #2355.

## Steps to reproduce
View a resource on a content page, use the also in this panel to navigate directly to another resource. Then rapidly press the back and forwards buttons in the browser.

## Context

 * Kolibri version: 0.15
 * Operating system: Ubuntu
 * Browser: Chrome

### LINKED ISSUE #8848: Potential completion modal error when clicking "Move On"
From @jtamiace: Clicking 'Move on' from the completion modal gave me an error but I'm having trouble reproducing this

https://www.notion.so/learningequality/REGRESSION-Learn-Content-renderer-27-BUG-3d5728de3b8c430e848bf654bf27bf77
Screen Shot 2021-12-07 at 10 15 58 AM

rtibbles:
## Summary
* Migrates core logging state and actions into a composable in learn
* Migrates and adds to tests
* Adds retry behaviour when encountering a 503 for initializing a content session, parallel to the 503 retry handling for updating a content session
* Fixes previously undocumented issue unique to quizzes where multiple interactions with the same attempt could result in the attempt object not being properly updated
* Gets rid of use of patching the ContentNodeProgressResource cache to maintain client side representation, as this is now always bypassed
* Instead updates contentNodeProgressMap from the useContentNodeProgress composable
* Uses this source of truth for displaying the current content progress in the top bar of the content page

## References
Fixes #8776
Fixes #8848 

## Reviewer guidance
Take a coach assigned quiz
Engage with an exercise
Watch a video

Does progress all get saved and tracked properly?

This will necessitate small tweaks to the QuizRenderer in develop - but they will be quick to implement.

----

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated


sairina:
When doing manual QA based on the Reviewer Guidance above, the tracking in the UI was fine, but in the console, there were some errors:

### Take a coach assigned quiz
1. Assigned a quiz to a class
2. As learner, took quiz, but before finishing quiz, returned to `ClassAssignmentsPage`
3. Returned to finish quiz, and saw the screenshot below. Icons had disappeared (not sure if known issue)
Screen Shot 2021-12-12 at 3 55 25 PM

4. As learner, when returned back to `ClassAssignmentsPage`, received redundant navigation error and 403 errors (one in `useProgressTracking`):
Screen Shot 2021-12-12 at 3 55 44 PM

### When testing by doing an exercise and then a video, also received 403 errors
Screen Shot 2021-12-12 at 3 56 12 PM

pcenov:
Tested with the latest [Buildkite deb build](https://buildkite.com/learningequality/kolibri-debian/builds/3400#f1679802-429c-43e4-a7be-379ed6c6d2c0):
Observations:
1. Assigned lesson: when I reopen a resource which is already marked as 'in progress' I'm not starting from the point where I left off - start a video and go to 20s then go back and then click again the video resource - it starts from the beginning. No errors in the console in that case.
2. Assigned Lesson: noticed 503 errors when navigating back and forth after completion. The error is displayed in the console, while what the user observes is some slowness while loading the completed resource. I compared with a previous build and I was getting an error screen with multiple errors in the console in that scenario.
3. Assigned Quiz: Getting a 'TypeError: Cannot read properties of null (reading 'hintsRenderer')' after exiting and going back to the quiz but the progress is kept. Also getting File not found errors for some exercises as reported by Sairina.


[logs.zip](https://github.com/learningequality/kolibri/files/7704451/logs.zip)

rtibbles:
> Icons had disappeared (not sure if known issue)

I can't imagine how my progress tracking PR might have affected this, but I'll take a closer look as we want this to be fixed regardless.

> redundant navigation error

This is pre-existing to this PR.

> 403 errors (one in useProgressTracking)

I think this may be new, because I think the quiz renderer was failing to call stop tracking after submission. I can fix this.

> When testing by doing an exercise and then a video, also received 403 errors

Looking at the screenshot, this is also a consequence of the failure to call stop tracking - it is repeatedly trying to make updates to the progress of the closed quiz, even though it is closed! So, I will make sure we call stop tracking before tear down, and also add some guard logic to the composable to prevent us trying to continue tracking user progress when they are no longer on the page.

> start a video and go to 20s then go back and then click again the video resource - it starts from the beginning.

Hrm, this sounds like a timing issue - where we're not ensuring that progress data is saved quickly enough. I think adding some guard logic to ensure that the progress tracking data is finalized would help here.

> I compared with a previous build and I was getting an error screen with multiple errors in the console in that scenario.

Excellent - this is one of the things this PR should be addressing!

> Getting a 'TypeError: Cannot read properties of null (reading 'hintsRenderer')' after exiting and going back to the quiz

I think this is a regression in the perseus viewer plugin after recent updates, can fix.

jredrejo:
@rtibbles is this PR fixing #8848 too?

kolibri/plugins/learn/assets/src/composables/useProgressTracking.js
```diff
@@ -0,0 +1,470 @@
...
+    if (!lastElapsedTimeCheck) {
+      return 0;
+    }
+    const currentTime = new Date();
+    const timeElapsed = currentTime - lastElapsedTimeCheck;
+    lastElapsedTimeCheck = currentTime;
+    // Some browsers entirely suspend code execution in background tabs,
+    // which can lead to unreliable timing if a tab has been in the background
+    // if the time elasped is significantly longer than the interval that we are
+    // checking this at, we should discard the result.
```
marcellamaki: what is "the result" referring to? the measure of elapsed time?

kolibri/plugins/learn/assets/src/views/ChannelRenderer/ContentItem.vue
```diff
@@ -159,21 +161,20 @@
...
           interaction,
           progress,
-        }).then(() => updateContentNodeProgress(this.contentNodeId, this.progress));
+        }).then(this.cacheProgress);
       },
       updateProgress(progress) {
-        this.updateContentSession({ progress }).then(() =>
-          updateContentNodeProgress(this.contentNodeId, this.progress)
-        );
+        this.updateContentSession({ progress }).then(this.cacheProgress);
```
marcellamaki, line 176: I don't understand why the content session is being updated and _then_ the progress is cached?

rtibbles:
@jredrejo I think it may well do that, looking at the errors in there! Thanks, I knew there were some other issues out there like this.

marcellamaki (commented):
It's a lot of changes, but the tests are very helpful both for catching things and for understanding what the intended behaviors are. I'm going to do some manual testing and another read through after the updates for the issues mentioned previously during QA.

kolibri/plugins/learn/assets/src/views/ChannelRenderer/ContentItem.vue
marcellamaki: (this pattern is used in a number of places, which is why I'm asking about it)

kolibri/plugins/learn/assets/src/composables/useProgressTracking.js
rtibbles: Yes, can update the comment accordingly.

kolibri/plugins/learn/assets/src/views/ChannelRenderer/ContentItem.vue
rtibbles: The previous utility function was slightly erroneously named, but it was updating the client side cache of content node progress. This is doing the same just with a different mechanism.

rtibbles:
> start a video and go to 20s then go back and then click again the video resource - it starts from the beginning.

Note: this was actually an issue with a typo - I was referring to `extraFields` in camelCase rather than `extra_fields` in snake_case. A similar issue was affecting `time_spent`.

kolibri/plugins/learn/assets/src/views/ChannelRenderer/ContentItem.vue
rtibbles: Added a code comment to the `this.cacheProgress` method to try to explain this better.

marcellamaki:
Manual QA after the latest changes: 
- I don't see any 403 errors in the console when viewing PDFs ✔️ 
- I don't see any 403 errors when watching video ✔️ 
- When I exist and resume a video, it loads to the correct spot where I previous exited ✔️ 

One error that I still see that @pcenov maybe was referencing is after completing an exercise in a lesson, when clicking "move on" I get a 503 error. First, this seems to not do anything (i.e. the page does not "move on" but continues to show the completed resource after the modal has closed), then there is a long loading spinner, then the content loads, after ~10 seconds...? (trying to replicate for more details) 

Error is `POST http://127.0.0.1:8000/api/logger/trackprogress/ 503 (Service Unavailable)`
The response is `Database is not available for write operations`

rtibbles:
> I get a 503 error

This is one of those cases where a properly handled API error is part of expected behaviour. This happens when the request receives a database locked error - it then waits 10 seconds before retrying. The user experience of this _could_ be better, but would probably require strings.

The alternative, as @pcenov noted is a big fat error page and not actually saving the data.

marcellamaki:
Got it - thanks for clarifying. Code read through makes sense with that, and I don't see any blockers with the new changes

marcellamaki (approved):
With updates, looks good to me

PR #8900, opened 2021-12-12
theme code simplification

indirectlylit:
## Summary

Follow-up from https://github.com/learningequality/kolibri/pull/8805 with more to come after.

In the theme code I originally tried defining constants in the Python `theme_hook` module which are simply references to strings used elsewhere in the code. These constants turn out to be more trouble than they're worth, mainly because they add a layer of indirection when someone is trying to figure out how all the theme logic fits together. They [fail the grep test](http://jamie-wong.com/2013/07/12/grep-test/).

Additional changes:

- remove unnecessary checks
- simplify default theme

## References

Will link to this PR from other relevant repos and PRs. See back-references below description.

## Reviewer guidance

Seem reasonable?

----

## Testing checklist

- [x] Contributor has fully tested the PR manually
- [ ] If there are any front-end changes, before/after screenshots are included
- [ ] Critical user journeys are covered by Gherkin stories
- [ ] Critical and brittle code paths are covered by unit tests

## PR process

- [x] PR has the correct target branch and milestone
- [x] PR has 'needs review' or 'work-in-progress' label
- [x] If PR is ready for review, a reviewer has been added. (Don't use 'Assignees')
- [ ] If this is an important user-facing change, PR or related issue has a 'changelog' label
- [ ] If this includes an internal dependency change, a link to the diff is provided

## Reviewer checklist

- Automated test coverage is satisfactory
- PR is fully functional
- PR has been tested for [accessibility regressions](http://kolibri-dev.readthedocs.io/en/develop/manual_testing.html#accessibility-a11y-testing)
- External dependency files were updated if necessary (`yarn` and `pip`)
- Documentation is updated
- Contributor is in AUTHORS.md

PR #8902, opened 2021-12-13
add object validation and apply it to themes

### LINKED ISSUE #8640: add core validator generator for Object-type props

indirectlylit:
## Summary

Note: please review & merge  #8900 first.

* Introduces two new functions, `validateObject` and `objectWithDefaults`, which bring Vue prop-like validation and default behavior to nested javascript objects. This can be used with component props, vuex  and API validation, and other scenarios
* Applies the new functions to theme specification

## References

* Builds off of  #8900 and  #8878
* Fixes #8640 and https://github.com/learningequality/kolibri-design-system/issues/151
* Created a follow-up issue here: https://github.com/learningequality/kolibri/issues/8903

## Reviewer guidance

* does the API make sense?
* are the tests sufficient?

----

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated


indirectlylit:
Switching to draft until we decide what branch this should go into.

kolibri/core/assets/src/objectSpecs.js
```diff
@@ -0,0 +1,193 @@
...
+
+/*
+ * Given a JS object and a spec object as defined above, return `true` if the object
+ * conforms to the spec and `false` otherwise
+ */
+export function validateObject(object, spec) {
+  // skip validation in production
+  if (process.env.NODE_ENV === 'production') {
+    return true;
+  }
```
nucleogenesis, line 139: Might be good to give a warning in this case otherwise you'll never know your object is failing the spec outside of production.

indirectlylit:
merging into `theme-updates` feature branch

ref: https://github.com/learningequality/kolibri/pull/8918

Issue #8871, opened 2021-12-08
remove 'all' options from filters


indirectlylit:
## Observed behavior

| observation | illustration |
|--|--|
| The default unfiltered state really corresponds to these "all" filters being selected, but that is not currently the case  |    ![image](https://user-images.githubusercontent.com/2367265/145222814-e3c75429-b570-40e6-8b50-5d69bba0ed52.png) |
| This leads to contradictory user-facing text  |     ![image](https://user-images.githubusercontent.com/2367265/145222362-8b27ff1a-ff86-4ed2-ade2-2c117dd12534.png) |
| Separately, the filter "None of the above" (a.k.a. 'uncategorized' - see https://github.com/learningequality/kolibri/issues/8872) doesn't make any sense when one thing above is "All categories" |     ![image](https://user-images.githubusercontent.com/2367265/145222470-30fadd8f-435b-43c0-9773-75e2c4ca226f.png) |

From @rtibbles:

> filtering by `all <X>/no <X>` probably isn’t terribly useful, and much more useful to just be able to turn off the filter

From @jtamiace:

> also becomes ambiguous what should happen when multiple filters are selected - what’s the difference between an empty filter and when it’s set to ‘All’?

## Expected behavior

Remove 'all' options from filters, where empty is implicitly ‘All’. this is in line with our decision to turn the dropdowns for Channel, language, etc. into 'clearable' dropdowns.

> ![image](https://user-images.githubusercontent.com/2367265/145222213-707dfe44-b743-474b-bc77-0ff2acb3c489.png)

## User-facing consequences

Confusing filter options

## Context

0.15.0-beta3

akash5100:
Can you help me, which files are needed to modify? 
------------------------------------------
In the case of the `all` filter, selecting the `all` filter will set the filters to default and the `clear all` button is doing the same thing, right? 
So we are going to remove the `all`, `All categories` option completely and rename `None of the above` to something like `clear filter`/`clear` etc.

indirectlylit:
> In the case of the `all` filter, selecting the `all` filter will set the filters to default and the `clear all` button is doing the same thing, right? 

That's correct.  the 'clear all' button has the same effect as setting the filters to default, which is "unfiltered"

Alternatively, the user can achieve the same result by deselecting any active filters one at a time.

> So we are going to remove the `all`, `All categories` option completely and rename `None of the above` to something like `clear filter`/`clear` etc

[...]

PR #8904, opened 2021-12-13
remove all option from filters

akash5100:
## Summary

This PR is a draft,

TODO's

- [x] removing `All categories` from `filter tab` 
- [x] removing `All` button from `Activity tab`

Thankyou @indirectlylit for helping.

 Before | After 
----------|----------
Screenshot from 2021-12-13 11-41-06 | Screenshot from 2021-12-13 11-26-00
Screenshot from 2021-12-15 13-18-54|Screenshot from 2021-12-13 11-26-11

* No changes in `clear all` filter

https://user-images.githubusercontent.com/53405133/145761580-b66dd6a1-e963-4033-aed2-75f4c36b185f.mp4

https://user-images.githubusercontent.com/53405133/146146663-7e966f28-73b1-4fc2-b4f5-d45143882e32.mp4

Please Review the file changes

## References

Completing this PR will Close issue #8871

----

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated


akash5100:
I have a doubt, do we have to clean code from where we import `All categories`.
This [line](https://github.com/learningequality/kolibri/blob/release-v0.15.x/kolibri/plugins/learn/assets/src/views/EmbeddedSidePanel/index.vue#L139)

indirectlylit:
Hi @akash5100 – 

Searching for `AllCategories`, we see that it is used in a couple important places:

kolibri/kolibri/plugins/learn/assets/src/composables/useSearch.js:

```
  102        for (let key of searchKeys) {
  103          if (key === 'categories') {
  104:           if (terms[key][AllCategories]) {
  105              getParams['categories__isnull'] = false;
  106              continue;
```

kolibri/kolibri/plugins/learn/assets/src/views/SearchChips.vue:

```
   86            return this.channelsMap[value].name;
   87          }
   88:         if (key === 'categories' && value === AllCategories) {
   89            return this.allCategoriesTranslator.$tr('allCategories'); // eslint-disable-line kolibri/vue-no-undefined-string-uses
   90          }
```

The change you're making should only affect the user-facing components in the side panel, but the underlying data handling behavior and other components should be left intact. Therefore you'll probably need to update the logic to explicitly set `AllCategories` if and only if no other category filters are selected.

kolibri/plugins/learn/assets/src/views/EmbeddedSidePanel/index.vue
```diff
@@ -307,9 +296,6 @@
     },
     methods: {
       genContentLink,
-      allCategories() {
-        this.$emit('input', { ...this.value, categories: { [AllCategories]: true } });
```
indirectlylit: This event will still need to be emitted in order for other parts of the app to continue working correctly. Instead of being emitted when the button is clicked, instead it should be emitted when the number of selected categories goes to zero.

akash5100:
Hi @indirectlylit,
can you help me, what should we do for:-
> you'll probably need to update the logic to explicitly set AllCategories if and only if no other category filters are selected.

Sorry for being late due to my exams.

indirectlylit:
Hi @akash5100 – hope your exams went well!

I'm not the best person to advise here so I'll defer to others to chime in. I may have been mistaken in my guidance above and don't want to lead you astray. (Note however that some folks are on vacation, so there might be a bit of a delay in response.)

That said, here are some brief notes:

At a high level, you'll need to make sure that showing "all categories" and "all activities" still works correctly after these buttons are removed and that any old code is cleaned up. For example, you might do a global search for the `AllCategories` constant and figure out what behavior that is driving:

https://github.com/learningequality/kolibri/blob/1dc6ba5c05469399e2d9d2da6b3481af4f95a0f0/kolibri/plugins/learn/assets/src/views/EmbeddedSidePanel/index.vue#L134-L139

Then you'll need to decide whether it's more correct to delete the constant and all code that references it, or to keep it and update the embedded side panel to emit appropriate events and update global vuex state using the new UI.

Also note that these files changed recently in #8959, and this introduced a conflict with your code that will need to be resolved. It's also worth looking at [the changes](https://github.com/learningequality/kolibri/pull/8959/files) there to better understand how the logic works.

Good luck!

marcellamaki (changes requested):
Hi @akash5100 -- as was mentioned above, there have been some changes to these files recently. I'd suggest rebasing from `release-v0.15.x`, resolving any conflicts (ensuring that all changes in 0.15.x branch are included), and then continuing. 

Overall, I think your approach should work, and that you are on the right track. It should be okay to remove the extra references to `AllCategories` that you mention. The clearing of search terms and returning to the default view is handled separately, so you won't inadvertently break that. You can see two places where you will see linting errors also where there are extra definitions that can be removed, since the only places those are being referenced (a method for `click` and a translation referenced`$tr`) are in buttons that you are removing. 

If you'd like to go ahead with those updates, I'll do a final review. Thanks for your work on this.

akash5100:
Thank you @marcellamaki for the help!

Issue #8767, opened 2021-11-22
Update learning activity label metadata to correctly display reference and duration


marcellamaki:
Now that we have the backend set up, we will need to make some adjustments to duration/reference displays that are used on the Learning Activity Label (which is on some of the cards). 

Specifically my guess is that '`durationEstimate()` in `LearningActivityLabel.vue` in the place that needs to be updated to fix the data mismatch.

rtibbles:
Full spec for how this will be encoded in content node options is here: https://github.com/learningequality/studio/issues/3281

Issue #6003, opened 2019-10-29
Add warnings to catch invocations of translation functions before polyfill loading occurs


rtibbles:
### Observed behavior
On two occasions now we have fixed Kolibri due to invocation of translation machinery in module scope breaking on Intl-polyfilled browsers. (#5254, #6002)

### Expected behavior
The linter should catch these invocations in module scope and error.

Alternatively, we can add a flag like `i18nReady = false` that gets set to `true` after `setupi18n` has been invoked. Then the translation machinery should complain if run too early

### User-facing consequences
Total breakage in old browsers.

jredrejo:
Would not it better to set 0.12.x as the milestone?

indirectlylit:
Updated description and title.

I don't believe this should be a blocker for 0.12.9

PR #8920, opened 2021-12-14
Learning activities labels and duration improvements

### LINKED ISSUE #8829: Broken string for duration overlay tag
The overlay on the bottom of a thumbnail image, which displays duration, seems to have a broken string. Potentially also in the chips on the library page
https://www.notion.so/learningequality/REGRESSION-Anywhere-potentially-Broken-52ca67f257f84c1d966e941092132618

MisRob:
## Summary

- Move shared logic related to learning activities into a composable function
- Extract logic from `LearningActivityLabel` related to duration to new `LearningActivityDuration` component
- Use  `LearningActivityDuration` on `HybridLearningContentCard` for displaying duration chips in the thumbnail area for larger resolutions - fixes  #8829. Adding mobile screenshot too here just FYI:

| Mobile | Desktop |
| -------- | ----------- |
| mobile | desktop |

- Simplify `LearningActivityLabel` styles, decouple them from `ResourceCard`, and make them safer so that the component can be used easily from other cards
- Move `LearningActivityLabel` and `LearningActivityDuration` components outside of `ResourceCard` directory since `ResourceCard` is not the only component using them

Note that no updates to logic around what should be source metadata for deciding how to display duration were made. I only extracted existing logic from `LearningActivityLabel`. This might need some more updates after we know more about metadata related to this (#8767).

## Reviewer guidance

For resources with various learning activities and durations, can you see any regressions of how learning activities labels, icons, and durations are displayed on cards on Home, Library, and Bookmarks pages?

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/plugins/learn/assets/src/composables/useLearningActivities.js
+++ b/kolibri/plugins/learn/assets/src/composables/useLearningActivities.js
@@ -6,7 +6,7 @@
 import { computed } from 'kolibri.lib.vueCompositionApi';
 import { get } from '@vueuse/core';
 import { LearningActivities } from 'kolibri.coreVue.vuex.constants';
-import { coreStrings } from '../../../../../core/assets/src/mixins/commonCoreStrings';
+import coreStrings from 'kolibri.utils.coreStrings';
 
 export default function useLearningActivities(contentNode) {
   const ReferenceLabel = coreStrings.$tr('readReference');
--- a/kolibri/plugins/learn/assets/src/composables/useLearningActivities.js
+++ b/kolibri/plugins/learn/assets/src/composables/useLearningActivities.js
@@ -1,0 +1,118 @@
+/**
+ * Everything related to learning activities of a content node
+ * and their duration.
+ */
+
+import { computed } from 'kolibri.lib.vueCompositionApi';
+import { get } from '@vueuse/core';
+import { LearningActivities } from 'kolibri.coreVue.vuex.constants';
+import { coreStrings } from '../../../../../core/assets/src/mixins/commonCoreStrings';
+
+const ReferenceLabel = coreStrings.$tr('readReference');
+const _LearningActivityToLabelMap = {
+  [LearningActivities.CREATE]: coreStrings.$tr('create'),
+  [LearningActivities.LISTEN]: coreStrings.$tr('listen'),
+  [LearningActivities.REFLECT]: coreStrings.$tr('reflect'),
+  [LearningActivities.PRACTICE]: coreStrings.$tr('practice'),
+  [LearningActivities.READ]: coreStrings.$tr('read'),
+  [LearningActivities.WATCH]: coreStrings.$tr('watch'),
+  [LearningActivities.EXPLORE]: coreStrings.$tr('explore'),
+};
+
+export default function useLearningActivities(contentNode) {
+  /**
+   * @returns {Boolean} Does the content node have exactly
+   *                    one learning activity associated with it?
+   */
+  const hasSingleActivity = computed(() => {
+    return (
+      contentNode && contentNode.learning_activities && contentNode.learning_activities.length === 1
+    );
+  });
+
+  /**
+   * @returns {String,null} A constant of the first learning activity
+   *                        associated with the content node.
+   *
+   */
+  const firstActivity = computed(() => {
+    if (
+      !contentNode ||
+      !contentNode.learning_activities ||
+      contentNode.learning_activities.length === 0
+    ) {
+      return null;
+    }
+    return contentNode.learning_activities[0];
+  });
+
+  /**
+   * @returns {Boolean} `true` if the content node has exactly one learning
+   *                    activity associated and that activity is reading
+   */
+  const isReference = computed(() => {
+    return get(hasSingleActivity) && get(firstActivity) === LearningActivities.READ;
+  });
+
+  /**
+   * @returns {Boolean} Does the content node have truthy duration?
+   */
+  const hasDuration = computed(() => {
+    return contentNode && contentNode.duration;
+  });
+
+  /**
+   * Should we display precise time duration for the content node?
+   *
+   * @returns {Boolean} `true` when the content node has exactly one learning activity
+   *                    associated and that activity is watching or listening
+   */
+  const displayPreciseDuration = computed(() => {
+    return (
+      contentNode &&
+      contentNode.duration &&
+      get(hasSingleActivity) &&
+      [LearningActivities.WATCH, LearningActivities.LISTEN].includes(get(firstActivity))
+    );
+  });
+
+  /**
+   * @returns {Number}
+   */
+  const durationInSeconds = computed(() => {
+    return contentNode && contentNode.duration ? contentNode.duration : 0;
+  });
+
+  /**
+   * @returns {String} Returns the translated 'Short activity' label when duration is less
+   *                   than 30 minutes. Otherwise returns the translated 'Long activity' label.
+   */
+  const durationEstimation = computed(() => {
+    if (!get(hasDuration)) {
+      return '';
+    }
+    return contentNode.duration < 1800
+      ? coreStrings.$tr('shortActivity')
+      : coreStrings.$tr('longActivity');
+  });
+
+  /**
+   * @param {String} learningActivity A learning activity constant
+   * @returns {String} A translated label for the learning activity
+   */
+  function getLearningActivityLabel(learningActivity) {
+    return _LearningActivityToLabelMap[learningActivity];
+  }
+
+  return {
+    ReferenceLabel,
+    hasSingleActivity,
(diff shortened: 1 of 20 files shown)


kolibri/plugins/learn/assets/src/views/HybridLearningContentCard/index.vue
```diff
@@ -84,7 +84,7 @@
   import TextTruncatorCss from 'kolibri.coreVue.components.TextTruncatorCss';
   import commonCoreStrings from 'kolibri.coreVue.mixins.commonCoreStrings';
   import ProgressBar from '../ProgressBar';
-  import LearningActivityLabel from '../cards/ResourceCard/LearningActivityLabel';
+  import LearningActivityLabel from '../LearningActivityLabel';
```
rtibbles, line 87: Need to update v-bind for `CardThumbnail` and `LearningActivityLabel` in here to correlate with the change to just the `contentNode` prop.

Getting vue warnings from use of this component as a result.

kolibri/plugins/learn/assets/src/composables/useLearningActivities.js
```diff
@@ -0,0 +1,118 @@
...
+ */
+
+import { computed } from 'kolibri.lib.vueCompositionApi';
+import { get } from '@vueuse/core';
+import { LearningActivities } from 'kolibri.coreVue.vuex.constants';
+import { coreStrings } from '../../../../../core/assets/src/mixins/commonCoreStrings';
+
+const ReferenceLabel = coreStrings.$tr('readReference');
+const _LearningActivityToLabelMap = {
+  [LearningActivities.CREATE]: coreStrings.$tr('create'),
```
rtibbles: Out of an abundance of caution, if we could move this logic inside the function scope that would be preferable, as we've previously had issues caused by accessing translations in module scope: https://github.com/learningequality/kolibri/issues/6003

(note - these only show up when the Intl polyfill is used, so it's hard to spot, and would only really occur in Safari).

rtibbles (changes requested):
Code read through makes sense, and manual testing checks out except for the HybridLearningContentCard. Two changes to make here before merging.

MisRob:
Thank you, @rtibbles. Resolved.

kolibri/plugins/learn/assets/src/composables/useLearningActivities.js
```diff
@@ -8,18 +8,18 @@ import { get } from '@vueuse/core';
 import { LearningActivities } from 'kolibri.coreVue.vuex.constants';
 import { coreStrings } from '../../../../../core/assets/src/mixins/commonCoreStrings';
```
rtibbles: Sorry, I missed this in the first pass - we shouldn't import core API objects in this way, as otherwise this will replicate this code into the learn plugin (using the import below, it accesses it at runtime via the default frontend object).
```suggestion
import coreStrings from 'kolibri.coreVue.mixins.commonCoreStrings';
```

rtibbles:
All issues from previous pass resolved - just one more thing that I missed first time.

kolibri/plugins/learn/assets/src/composables/useLearningActivities.js
MisRob: This ends up with the following Webpack error

```
[Vue warn]: Error in data(): "TypeError: kolibri_coreVue_mixins_commonCoreStrings__WEBPACK_IMPORTED_MODULE_3___default.a.$tr is not a function"
```

I think this may be happening because `commonCoreStrings` [exports Vue mixin by default](https://github.com/MisRob/kolibri/blob/f9fcd0a6f4e6315ffa6e0af64958e7d020cbf495/kolibri/core/assets/src/mixins/commonCoreStrings.js#L1089) that is then registered [in the core API](https://github.com/MisRob/kolibri/blob/f9fcd0a6f4e6315ffa6e0af64958e7d020cbf495/kolibri/core/assets/src/core-app/apiSpec.js#L48). I'd like to only use [this translator](https://github.com/MisRob/kolibri/blob/f9fcd0a6f4e6315ffa6e0af64958e7d020cbf495/kolibri/core/assets/src/mixins/commonCoreStrings.js#L8) though. Is that a problem too? I think I already saw translator functions being imported and used in javascript files.

rtibbles: It's fine to import translators across files, it's just a special case for the core API. The only issue is that it will bundle the code for the core translator into the learn plugin, and then the string extraction machinery will duplicate the core strings into the learn plugin message bundles too, so it will cause a fair amount of bloat.

I think maybe better to add just the strings translator to the core API?

Can update this import https://github.com/learningequality/kolibri/blob/release-v0.15.x/kolibri/core/assets/src/core-app/apiSpec.js#L48 to import `commonCoreStrings, { coreStrings }` and then add `coreStrings` to the core API here under utils? https://github.com/learningequality/kolibri/blob/release-v0.15.x/kolibri/core/assets/src/core-app/apiSpec.js#L218

MisRob: I see now, thank you. Yes, I can add it to the core utils.

rtibbles: Only thing to be wary about here is that in tests you may need to explicitly mock the `coreString` import in this case, as our Jest auto-import patching for coreAPI imports only works for default exports.

MisRob: `import commonCoreStrings, { coreStrings } from '../mixins/commonCoreStrings';` is breaking the build process:

```
Error: Cannot find module '/home/michaela/dev/kolibri/packages/kolibri-tools/dist/apiSpec.json'
    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:636:15)
    at Function.Module._load (internal/modules/cjs/loader.js:562:25)
    at Module.require (internal/modules/cjs/loader.js:692:17)
    at require (internal/modules/cjs/helpers.js:25:18)
    at Object.<anonymous> (/home/michaela/dev/kolibri/packages/kolibri-tools/lib/apiSpecExportTools.js:106:13)
    at Module._compile (internal/modules/cjs/loader.js:778:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:789:10)
    at Module.load (internal/modules/cjs/loader.js:653:32)
    at tryModuleLoad (internal/modules/cjs/loader.js:593:12)
    at Function.Module._load (internal/modules/cjs/loader.js:585:3)
    at Module.require (internal/modules/cjs/loader.js:692:17)
    at require (internal/modules/cjs/helpers.js:25:18)
    at Object.<anonymous> (/home/michaela/dev/kolibri/packages/kolibri-tools/lib/webpack.config.base.js:20:23)
    at Module._compile (internal/modules/cjs/loader.js:778:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:789:10)
    at Module.load (internal/modules/cjs/loader.js:653:32)
internal/modules/cjs/loader.js:638
    throw err;
    ^
```

and the linter as well:

```
kolibri/plugins/user_auth/assets/test/views/auth-select-page.spec.js
  1:1   error    Resolve error: Cannot find module '/home/michaela/dev/kolibri/packages/kolibri-tools/dist/apiSpec.json'
    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:636:15)
    at Function.Module._load (internal/modules/cjs/loader.js:562:25)
    at Module.require (internal/modules/cjs/loader.js:692:17)
    at require (internal/modules/cjs/helpers.js:25:18)
    at Object.<anonymous> (/home/michaela/dev/kolibri/packages/kolibri-tools/lib/apiSpecExportTools.js:106:13)
    at Module._compile (internal/modules/cjs/loader.js:778:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:789:10)
    at Module.load (internal/modules/cjs/loader.js:653:32)
    at tryModuleLoad (internal/modules/cjs/loader.js:593:12)
    at Function.Module._load (internal/modules/cjs/loader.js:585:3)  import/namespace
```

I wasn't able to find out what's the problem in a reasonable timeframe so I just left those two imports on different lines.

MisRob:
I forgot to add some related changes to the last commit, so I forced pushed it now. All previous commits should stay the same as before.

kolibri/plugins/learn/assets/src/composables/useLearningActivities.js
MisRob: Should be resolved and will follow-up on Slack with modules resolution errors

rtibbles: Yeah that seems fine, probable that some AST parsing code is not setup to handle the same line import.

rtibbles (approved):

PR #8926, opened 2021-12-14
Filter bookmarks by available nodes.

### LINKED ISSUE #8823: Use the available = true filter on the frontend for bookmarks
So that the availability filter on the backend is useful and so that learners do not see bookmarks for content that is not available.

rtibbles:
## Summary
* Only return available bookmarks to the user to avoid confusion

## References
Fixes #8823

## Reviewer guidance
Bookmark a content node, then delete that content node from your device. Check that it no longer shows up in your bookmarks.

Before deleting node:
Screenshot from 2021-12-14 12-26-00

After deleting the top node from the device (marked as unavailable):
Screenshot from 2021-12-14 12-27-01

----

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/core/content/api.py
+++ b/kolibri/core/content/api.py
@@ -1246,11 +1246,12 @@ class BookmarkFilter(FilterSet):
         queryset = queryset.annotate(
             available=Subquery(
                 models.ContentNode.objects.filter(
-                    available=True,
                     id=OuterRef("contentnode_id"),
                 ).values_list("available", flat=True)[:1]
             )
         )
+
+        return queryset.filter(available=value)
 
 
 class ContentNodeBookmarksViewset(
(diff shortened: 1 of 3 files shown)


kolibri/core/content/api.py
```diff
@@ -1246,12 +1246,13 @@ def filter_available(self, queryset, name, value):
...
             available=Subquery(
                 models.ContentNode.objects.filter(
-                    available=True,
                     id=OuterRef("contentnode_id"),
                 ).values_list("available", flat=True)[:1]
             )
         )
 
+        return queryset.filter(available=value)
+
```
sairina, line 1255: What's the difference between returning `available=value` vs. filtering by `available=True`?

sairina (approved):
Manual QA works. lgtm!

kolibri/core/content/api.py
rtibbles: If `value` is `False` then it would only return not available ones (although, technically it would need to be `False` or `None`) so it wouldn't work quite as expected for the negative case.

PR #8928, opened 2021-12-14
Topics page fixes

### LINKED ISSUE #8847: ux issues with 'search' tab in Library
## Observed behavior

| observation | illustration |
|--|--|
| At smaller sizes the 'Search' tab launches the side panel which seems incorrect |   |
| The tabs also initially start out with neither selected | | 
| Additionally, on mid-size responsive views when there are no folders there is a only single tab. At this point, it should probably be replaced with just a header or a label rather than being clickable. | image |

## Expected behavior

* Make sure there is always one tab selected
* Do not overload the 'search' tab to launch filter side view
* Display as header instead of tab when there is only one tab

## User-facing consequences

confusing navigation and interaction

marcellamaki:
## Summary
- Updates 'search' button to be 'filter' button per spec (and to reduce confusion about multiple references to 'search'ing)
- Fixes tab behavior - tabs update main view and routes, whereas filter button opens filtering side panel
- Folder tab doesn't display when we are at the last child folder
- replaces breadcrumbs that seem to have disappeared on mobile

## References
Fixes #8847

**Figma spec for medium screens (where the toggling changes were most relevant)**
Screen Shot 2021-12-14 at 3 48 00 PM

**Implementation**
updated-topics-page

updated-topics-mobile

## Reviewer guidance
- Any oversights or regressions that I missed?

----

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/plugins/learn/assets/src/views/TopicsPage.vue
+++ b/kolibri/plugins/learn/assets/src/views/TopicsPage.vue
@@ -617,7 +617,9 @@
       },
       searchTerms(newVal, oldVal) {
         if (!isEqual(newVal, oldVal)) {
-          this.$router.push({ ...this.searchTabLink });
+          if (!isEqual(this.searchTabLink, this.$route)) {
+            this.$router.push({ ...this.searchTabLink }).catch(() => {});
+          }
           this.sidePanelIsOpen = false;
         }
       },
@@ -727,8 +729,9 @@
   .header {
     position: relative;
     width: 100%;
+    height: 324px;
     padding-top: 32px;
-    padding-bottom: 32px;
+    padding-bottom: 48px;
     padding-left: 32px;
     background-color: white;
     border: 1px solid #dedede;
--- a/kolibri/plugins/learn/assets/src/views/TopicsPage.vue
+++ b/kolibri/plugins/learn/assets/src/views/TopicsPage.vue
@@ -72,7 +72,7 @@
           />
           <HeaderTab
             :text="coreString('searchLabel')"
-            :to="topics.length ? searchLink : {} "
+            :to="topics.length ? searchTabLink : {} "
           />
         </HeaderTabs>
       </div>
@@ -81,6 +81,7 @@
         <div class="mobile-header-contents">
           <HeaderTabs>
             <HeaderTab
+              v-if="topics.length"
               :text="coreString('folders')"
               :to="foldersLink"
             />
@@ -100,7 +101,7 @@
         class="main-content-grid"
         :style="gridOffset"
       >
-        <slot v-if="windowIsSmall" name="breadcrumbs" class="breadcrumbs"></slot>
+        <KBreadcrumbs v-if="breadcrumbs.length && windowIsSmall" :items="breadcrumbs" />
         <div
           class="card-grid"
         >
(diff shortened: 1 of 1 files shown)


kolibri/plugins/learn/assets/src/views/TopicsPage.vue
```diff
@@ -99,15 +101,15 @@
         class="main-content-grid"
         :style="gridOffset"
       >
-        <slot v-if="windowIsSmall" name="breadcrumbs" class="breadcrumbs"></slot>
+        <KBreadcrumbs v-if="breadcrumbs.length && windowIsSmall" :items="breadcrumbs" />
```
rtibbles, line 104: Ah, this was my fault, I think - I didn't replace both the breadcrumbs slots.

kolibri/plugins/learn/assets/src/views/TopicsPage.vue
```diff
@@ -80,12 +81,13 @@
...
           <HeaderTabs>
             <HeaderTab
+              v-if="topics.length"
               :text="coreString('folders')"
               :to="foldersLink"
             />
             <HeaderTab
               :text="coreString('searchLabel')"
-              :to="searchLink"
+              :to="searchLink "
```
rtibbles: Weird additional whitespace here?

rtibbles (changes requested):
There are still some issues here from testing.

rtibbles:
When first navigating to the topics page, the folder header is not highlighted, because it requires the query parameter to be set.

Screenshot from 2021-12-15 05-50-33

Overall, I think the interaction on mobile is still confusing - the header links being used as buttons to toggle the side panels feels counterintuitive. As such, the fact that the button for filter is only available in tablet layout but not smaller layouts seems like an issue.

rtibbles:
Also, when first navigating to the search tab, I think we should show the search side panel - otherwise we are requiring users to click the search header and _then_ open the search panel separately, making two clicks to express what they want to do.

marcellamaki:
Updated interactions -- this briefly broke my brain, but I think it covers the scenarios discussed earlier...
updated-topics-desktop

updated-topics-tablet

updated-topics-mobile

kolibri/plugins/learn/assets/src/views/TopicsPage.vue
```diff
@@ -619,8 +616,9 @@
         }
       },
       searchTerms(newVal, oldVal) {
-        if (!isEqual(newVal, oldVal) && this.displayingSearchResults) {
-          this.$router.push({ ...this.searchLink, sidePanel: false });
+        if (!isEqual(newVal, oldVal)) {
+          this.$router.push({ ...this.searchTabLink });
```
rtibbles: Getting redundant navigation errors here now - simplest fix might be to just to update it to:
```suggestion
          if (!isEqual(this.searchTabLink, this.$route)) {
            this.$router.push({ ...this.searchTabLink });
          }
```

Don't know if this will actually work, alternatively can just wrap it in a try/catch.

rtibbles (changes requested):
Manual testing checks out - just need to prevent the redundant navigation error.

rtibbles (approved):

PR #8931, opened 2021-12-15
Finalize H5P Integration

### LINKED ISSUE #8927: Finalize H5P integration
- [ ] Disable time based progress tracking for H5P files in the HTML5 viewer
- [ ] Log all logged xAPI statements even when they fail validation, but mark them as such
- [ ] Add logic in the score calculation logic to handle invalid statements

rtibbles:
## Summary
-  Disable time based progress tracking for H5P files in the HTML5 viewer
-  Log all logged xAPI statements even when they fail validation, but mark them as such
-  Add logic in the score calculation logic to handle invalid statements and other statement handling logic
- Prevent H5P from returning statements that imply completion when they are only about subContentIds within the main content (seemingly contrary to the CMI5 spec https://github.com/AICC/CMI-5_Spec_Current/blob/quartz/cmi5_spec.md#verbs_completed)
- Clean up old mutation observer/resizing code that is no longer used but was causing an unreported error that was preventing data from being saved

## References
Fixes #8927

## Reviewer guidance
Engage with H5P content in the Kolibri QA channel.
Engage with the H5P node here: `birid-riziv`

Ensure in both instances that the content is not completed too soon at the very least (possible that it will never be marked as complete, but there's only so much we can do).

----

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/packages/hashi/src/iframeClient.js
+++ b/packages/hashi/src/iframeClient.js
@@ -82,6 +82,7 @@ export default class SandboxEnvironment {
         this.xAPI.iframeInitialize(this.iframe.contentWindow);
         patchIndexedDB(this.contentNamespace, this.iframe.contentWindow);
       } catch (e) {
+        console.debug(e);
         console.log('Shimming storage APIs failed, data will not persist'); // eslint-disable-line no-console
       }
     }
--- a/packages/hashi/src/iframeClient.js
+++ b/packages/hashi/src/iframeClient.js
@@ -81,11 +81,6 @@ export default class SandboxEnvironment {
         this.H5P.iframeInitialize(this.iframe.contentWindow);
         this.xAPI.iframeInitialize(this.iframe.contentWindow);
         patchIndexedDB(this.contentNamespace, this.iframe.contentWindow);
-        this.mutationObserver.observe(this.iframe.contentDocument, {
-          attributes: true,
-          childList: true,
-          subtree: true,
-        });
       } catch (e) {
         console.log('Shimming storage APIs failed, data will not persist'); // eslint-disable-line no-console
       }
@@ -93,9 +88,6 @@ export default class SandboxEnvironment {
   }
 
   clearIframe() {
-    try {
-      this.mutationObserver.disconnect();
-    } catch (e) {} // eslint-disable-line no-empty
     try {
       document.body.removeChild(this.iframe);
     } catch (e) {} // eslint-disable-line no-empty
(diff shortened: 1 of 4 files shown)


packages/hashi/src/iframeClient.js
```diff
@@ -81,21 +81,13 @@ export default class SandboxEnvironment {
         this.H5P.iframeInitialize(this.iframe.contentWindow);
         this.xAPI.iframeInitialize(this.iframe.contentWindow);
         patchIndexedDB(this.contentNamespace, this.iframe.contentWindow);
-        this.mutationObserver.observe(this.iframe.contentDocument, {
```
marcellamaki, line 84: Is this functionality replaced with other code somewhere, or is it just obsolete?

marcellamaki (commented):
So far the code read through looks good to me -- easy to follow even with my only rudimentary understanding of H5P. I'm not sure how to best do some manual QA on this so open to recommendations @radinamatic or @rtibbles if you have any. Once that checks out I will approve

packages/hashi/src/iframeClient.js
rtibbles: It was previously replaced in https://github.com/learningequality/kolibri/pull/8458, but these references to `mutationObserver` were missed.

radinamatic:
Tested with Kolibri QA channel, `birid-riziv` and the main `piman-fapum`, no _premature_ completion modal appeared. 
As you mentioned, in many cases there was no way to achieve completion (which is a bit anticlimactic), but some do not even register progress, even when one goes through all the _motions_ (`Dialog cards`, for example). Not sure if that is due to a resource or Kolibri...? 🤔 

1 | 2
--- | ---
LEDev2111 (take)  Running  - Oracle VM VirtualBox_016 | LEDev2111 (take)  Running  - Oracle VM VirtualBox_015

radinamatic (approved):
For the stated PR scope, manual QA is passed! :shipit: 🙂

packages/hashi/src/iframeClient.js
```diff
@@ -81,21 +81,14 @@ export default class SandboxEnvironment {
...
-          attributes: true,
-          childList: true,
-          subtree: true,
-        });
       } catch (e) {
+        console.debug(e);
```
nucleogenesis, line 85: Is this committed intentionally?

rtibbles:
Have confirmed with @radinamatic that this seems to be because the H5P content is not emitting any events at all on which we would judge progress.

packages/hashi/src/iframeClient.js
rtibbles: Yes - the reason one of the bugs I fixed in here went unnoticed for so long was because the error was silently swallowed.

PR #8933, opened 2021-12-15
API: Add query context to the pagination object

danigm:
## Summary
This patch adds the getParams to the pagination object so any call to the getContentPage has the same filter parameters than the initial query call, making the pagination consistent and simpler.

### DIFF
--- a/kolibri/plugins/learn/assets/src/views/ChannelRenderer/CustomContentRenderer.vue
+++ b/kolibri/plugins/learn/assets/src/views/ChannelRenderer/CustomContentRenderer.vue
@@ -163,12 +163,14 @@
           },
         })
           .then(contentNodes => {
+            const { more, results } = contentNodes;
+
             return createReturnMsg({
               message,
               data: {
                 maxResults: options.maxResults ? options.maxResults : 50,
-                more: contentNodes.more,
-                results: contentNodes.results,
+                more,
+                results,
               },
             });
           })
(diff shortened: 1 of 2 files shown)


kolibri/plugins/learn/assets/src/views/ChannelRenderer/CustomContentRenderer.vue
```diff
@@ -149,26 +149,35 @@
...
-            descendant_of: options.descendantOf,
-          },
+          getParams,
         })
           .then(contentNodes => {
+            const { more, results } = contentNodes;
+
+            if (more) {
+              // Add query parameters to the more object
+              more.query = getParams;
```
rtibbles: This is the format of an example `more` object returned by the API:
```
{"include_coach_content":"true","max_results":"25","learning_activities":"wA01urpi","contentCacheKey":"1637348017","format":"json","cursor":"cD0yZTA3YTllYWU3NDc1YWQ0YjA1Y2VhODQ0Y2RlODcwNg=="}
```
It includes all the parameters needed to make the subsequent query, so this additional encoding is unnecessary.

All that is needed is for it to be passed back as the parameters for the next request.

I'd suggest instead just having an `options.more` on the options object to this parameter and then just having `fetchMore` take one option `options.more` that is just this entire object.

kolibri/plugins/learn/assets/src/views/ChannelRenderer/CustomContentRenderer.vue
```diff
@@ -185,17 +194,24 @@
 
         return ContentNodeResource.fetchCollection({
           getParams: {
+            ...options.query,
```
rtibbles: See above - can just set getParams to `options.more` and not explicitly encode anything here.

This also has the advantage of not being brittle to changes in the naming of the cursor, for example.

rtibbles (changes requested):
Can simplify further to just pass back the `more` object rather than attempting to deconstruct and reconstruct it.

kolibri/plugins/learn/assets/src/views/ChannelRenderer/CustomContentRenderer.vue
danigm: Ups, I misunderstood the more object, it's simpler than what I was thinking, I've updated the commit to remove the `query` attribute, just pass the full `options` as `getParams` in the `fetchMore` method.

rtibbles (approved):

PR #8937, opened 2021-12-16
PDF renderer completion fixes

### LINKED ISSUE #8868: PDF resource completion
## Observed behavior
It does not seem possible to achieve completion on any PDF file, no matter how short they are, or time learner spends with the resource. I added 3 more short PDF files (3-9 pages) to the QA channel, but it seems impossible to complete them.

## Errors and logs
…

## Expected behavior
Ability to complete PDF resources.

## User-facing consequences
... 

## Steps to reproduce
...

## Context

 * Kolibri version: 0.15.0b3
 * Operating system: Android 8.0 and 9.0
 * Browser: -

cc @marcellamaki @rtibbles

sairina:
## Summary
- Fix edge case of 2-page pdfs
- Add visit to 1st pdf page for all pdfs
- Fix progress tracking from PDF to content card

## References
Fixes #8868 

## Reviewer guidance
1. Sign in as learner
2. Go to the library and choose the 2-pg PDF in the QA channel
3. Read through the PDF; check to see that it gets completed and that the completed icon shows up in the app bar
4. Go back to the library and check to see that the progress on that content card is now complete
5. Repeat for PDFs of other lengths, but also periodically check to see that the progress on the content card shows accurate progress while you're reading through the PDF

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/plugins/pdf_viewer/assets/src/views/PdfRendererIndex.vue
+++ b/kolibri/plugins/pdf_viewer/assets/src/views/PdfRendererIndex.vue
@@ -325,8 +325,9 @@
           // TODO: there is a miscalculation that causes a wrong position change on scale
           this.savePosition(this.calculatePosition());
 
-          // determine how many pages user has viewed/visited
-          let currentPage = parseInt(this.currentLocation * this.totalPages) + 1;
+          // determine how many pages user has viewed/visited; fix edge case of 2 pages
+          let currentPage =
+            this.totalPages === 2 ? 2 : parseInt(this.currentLocation * this.totalPages) + 1;
           this.storeVisitedPage(currentPage);
           this.updateProgress();
           this.updateContentState();
--- a/kolibri/plugins/pdf_viewer/assets/src/views/PdfRendererIndex.vue
+++ b/kolibri/plugins/pdf_viewer/assets/src/views/PdfRendererIndex.vue
@@ -260,6 +260,8 @@
         }
         this.$emit('startTracking');
         this.updateContentStateInterval = setInterval(this.updateProgress, 30000);
+        // Even if user does not pause while scrolling on first page, we store that as visited
+        this.storeVisitedPage(1);
       });
     },
     beforeDestroy() {
--- a/kolibri/plugins/learn/assets/src/composables/__tests__/useProgressTracking.spec.js
+++ b/kolibri/plugins/learn/assets/src/composables/__tests__/useProgressTracking.spec.js
@@ -354,6 +354,13 @@ describe('useProgressTracking composable', () => {
       expect(get(progress_delta)).toEqual(0.1);
       expect(client).not.toHaveBeenCalled();
     });
+    it('should increment progress_delta if progress is updated twice', async () => {
+      const { updateContentSession, progress_delta } = await initStore();
+      await updateContentSession({ progress: 0.6 });
+      await updateContentSession({ progress: 0.7 });
+      expect(get(progress_delta)).toEqual(0.2);
+      expect(client).not.toHaveBeenCalled();
+    });
     it('should update progress and store progress_delta if progress is updated over threshold', async () => {
       const { updateContentSession, progress } = await initStore();
       await updateContentSession({ progress: 1 });
--- a/kolibri/plugins/learn/assets/src/composables/useProgressTracking.js
+++ b/kolibri/plugins/learn/assets/src/composables/useProgressTracking.js
@@ -65,7 +65,7 @@ export default function useProgressTracking(store) {
   store = store || getCurrentInstance().proxy.$store;
   const complete = ref(null);
   const progress_state = ref(null);
-  const progress_delta = ref(null);
+  const progress_delta = ref(0);
   const time_spent = ref(null);
   const time_spent_delta = ref(null);
   const session_id = ref(null);
@@ -350,7 +350,9 @@ export default function useProgressTracking(store) {
       progress = _zeroToOne(progress);
       progress = threeDecimalPlaceRoundup(progress);
       if (get(progress_state) < progress) {
-        set(progress_delta, threeDecimalPlaceRoundup(progress - get(progress_state)));
+        const newProgressDelta =
+          get(progress_delta) + threeDecimalPlaceRoundup(progress - get(progress_state));
+        set(progress_delta, newProgressDelta);
         set(progress_state, progress);
       }
     }


kolibri/plugins/pdf_viewer/assets/src/views/PdfRendererIndex.vue
```diff
@@ -323,8 +325,9 @@
           // TODO: there is a miscalculation that causes a wrong position change on scale
           this.savePosition(this.calculatePosition());
 
-          // determine how many pages user has viewed/visited
-          let currentPage = parseInt(this.currentLocation * this.totalPages) + 1;
+          // determine how many pages user has viewed/visited; fix edge case of 2 pages
```
rtibbles, line 328: What was the edge case here for 2 pages? Seems like both 1 and 2 are edge cases, which seems odd.

rtibbles: My thought here is that the subsequent progress_delta fix might also fix this issue

sairina: Ah, yes, it would make sense that it was both 1 and 2, but I didn't have a PDF that was only 1 page.

sairina: Hmm, no, that doesn't seem to fix it. When there are two pages in the pdf, `currentPage` seems to be constantly set to 1 because `parseInt(this.currentLocation * this.totalPages)` always ends up at `0` because `this.currentLocation` doesn't get high enough.

rtibbles: Hrm - my I am curious why that is specific to 2 page PDFs, and why on a 3 page PDF '2' does become a currentPage at some point, makes me think there might be something slightly off about our currentLocation calculation.

rtibbles: (this is just curiosity, not a blocker)

rtibbles (approved):

kolibri/plugins/pdf_viewer/assets/src/views/PdfRendererIndex.vue
akash5100: Hello sir @rtibbles, I spent my time learning about the rendering of PDFs in Kolibri. I tried to understand the files `LearnImmersivelayout`, `LearningActivityBar`, `ProgressIcon` etc. If I am not wrong the PDFs are rendered [here](https://github.com/learningequality/kolibri/blob/develop/kolibri/plugins/pdf_viewer/assets/src/views/PdfPage.vue#L12) on a canvas. I have a doubt, about how can we track the current page from the rendered canvas? , I think if we get to know which page the user is at then we can fix the `ProgressIcon` not updating bug.  
(I will try to research more about PDF.js to understand this problem more)

rtibbles: We don't render every page at once (as this causes performance issues for larger PDFs) so we know which pages are currently rendered, and control that based on where the user is.

The component you linked to controls the rendering of a single page - it is then used inside a `RecycleList` component that handles only actually rendering the pages that are visible by the user at the current time: https://github.com/learningequality/kolibri/blob/develop/kolibri/plugins/pdf_viewer/assets/src/views/PdfRendererIndex.vue#L47

It's in ^ this component that the progress tracking logic is happening, this is where the fix would almost certainly need to be.

akash5100: Hi @rtibbles and @sairina, thanks for all your help I found a solution but I am not sure it is correct.
Let me explain what I came up with...
In the `<progressIcon>` we have a prop called "progress".
The value of "progress" determines the state of `<progressIcon>`
Screenshot from 2022-03-13 12-47-40
Screenshot from 2022-03-13 13-02-54
In React we use a hook `useEffect` which lets you perform side effects in function components (I am new to vue.js so I don't know what features we have in vue.js, we can use something like that). 

My Idea is when we scroll we can keep on updating the "progress" value maybe from the `PdfRendererIndex.vue` (not sure from where) something like:
```
console output:
> 0.638
> 0.639
> 0.640
> 0.641
> 0.642
(this output is a simple console log of an onScroll event, all I wanted to show is my idea)
```
 but I don't know how we can access the value of "this.progress" when we scroll

In the end when the "progress" value is equal to `1`. we can change the `isComplete` value to `True` and `isInProgress` value to `False` (I think, this part already exists in the code). 
Thank you for reading this, please tell me if wrong, I will search for another way to fix it. Or maybe it's just a small bug that I need to find and don't need all the stuff I mentioned above (sorry 😛)

sairina: Ooh, I never even considered looking at `ProgressIcon`! @rtibbles would probably know more about whether this is the right way to go in terms of efficiency, but if you want to explore it a bit more (if you haven't already):
- `progress` is a prop, and its value is coming from the [`contentProgress` prop in the `LearningActivityBar` component ](https://github.com/learningequality/kolibri/blob/release-v0.15.x/kolibri/plugins/learn/assets/src/views/LearningActivityBar.vue#L181)
- `contentProgress` is a computed prop coming from [`contentNodeProgress` in the `LearnImmersiveLayout` component](https://github.com/learningequality/kolibri/blob/release-v0.15.x/kolibri/plugins/learn/assets/src/views/LearnImmersiveLayout.vue#L238)
- [`useContentNodeProgress.js`](https://github.com/learningequality/kolibri/blob/release-v0.15.x/kolibri/plugins/learn/assets/src/composables/useContentNodeProgress.js) is where this is housed. It's called a composable, which is pretty similar to a React hook, you can see it in the code through `setup()` and one of the difference between this and React hooks is that it is only run once (I believe).
   - you can see it in use in the [`ProgressBar`](https://github.com/learningequality/kolibri/blob/release-v0.15.x/kolibri/plugins/learn/assets/src/views/ProgressBar.vue#L46) and also in the [`LearnImmersiveLayout`](https://github.com/learningequality/kolibri/blob/release-v0.15.x/kolibri/plugins/learn/assets/src/views/LearnImmersiveLayout.vue#L183)

PR #8946, opened 2021-12-17
Schnittstellenfehler - Visual bugs in German

### LINKED ISSUE #7337: Possible responsivenes and/or design issues appearing on small screens (Android app)
### Observed behavior

To better explore potential issues with long strings/labels in buttons, I used the German locale (will provide equivalent in English for easier debugging).

@jonboiser suggested I use one long issue, not sure if separate ones would be better. 
cc @nucleogenesis @kollivier 

1. Padding/margins

   - Missing padding on the button
     Android (Import content after Advanced Setup Wizard; German) | English
     ------------ | -------------
     DevTools - localhost:8080-de-device-_839 | (will add later)

     Second example:
     Android (Device > Facilities; German) | Comments
     ------------ | -------------
     DevTools - localhost:8080-de-device-_841 | Technically it was possible to scroll the Facility (**Enrichtung**) block horizontally by touch, but it was difficult to control, and more importantly *there was no indication that user could scroll that block*. Is there a possibility to display a scrollbar at the bottom whenever the content is wider than the available width?

     Third example:
     Android (Device > Channels; German) | Comments
     ------------ | -------------
     DevTools - localhost:8080-de-device-_847 | - 

     Fourth example:
     Android (Device > Channels > Import; German) | Comments
     ------------ | -------------
     DevTools - localhost:8080-de-device-_870 | 2. is the problem here, 1. is part of the strings too long and cut off that we may not decide to resolve in this round.

   -  Android (Advanced Setup Wizard; German) | Comments
      ------------ | -------------
      DevTools - localhost:8080-de-setup_wizard---_836 | This is probably according to specs (minus the missing margin), but still weird looking with the lower button not looking like a button but having this positioning.

2. Strings too long and cut out

   - Android (Device > Permissions; German) | Dev server with responsive view (English) 
     ------------ | -------------
     DevTools - localhost:8080-de-device-_840 | LEDev20 04 (start)  Running  - Oracle VM VirtualBox_872

     I'm guessing 1. is according to specs, not sure what we can do with 2., but there should be a way to deal with 3.

   - Android (Device > Info; German) | Dev server with responsive view (English) 
     ------------ | -------------
     DevTools - localhost:8080-de-device-_844 | LEDev20 04 (start)  Running  - Oracle VM VirtualBox_873

   - Android (Facility > Settings; German) | Dev server with responsive view (English) 
     ------------ | -------------
     DevTools - localhost:8080-de-facility---users_852 | LEDev20 04 (start)  Running  - Oracle VM VirtualBox_853

2. String overflowing the container element

   - Android (Device > Settings; German) | Dev server with responsive view (English) 
     ------------ | -------------
     DevTools - localhost:8080-de-device-_845 | LEDev20 04 (start)  Running  - Oracle VM VirtualBox_874

### Expected behavior
App fully responsive and usable for the super admin user even when using the small screen.

### User-facing consequences
Unable to use the app on small screen devices.

### Errors and logs
...

### Steps to reproduce
Provided with screenshots.

### Context
 * Kolibri version: Beta 12
 * Device: Moto G Play 6 
 * Operating system: Android 9

nucleogenesis:
## Summary

Addressing some visual issues in #7337 and finding that some no longer exist:

In order as they appear in #7337 

1) "N Channels selected [Import]" where text flows atop the button *not fixed here*

2) Facilites list table - buttons being cut off. (Also adding space between the KButtonGroup buttons when stacked)

https://user-images.githubusercontent.com/6356129/146487986-c0fd3a8f-4969-451a-b4a8-669b7169859c.mp4

3) Options and Import button hugging when stacked - added some space there:

image

4) Device > Channels > Import 
4a) Table items misaligned - already improved by other work it seems. 
4b) Bottom app bar button & text like (1) above *not fixed here*

5) Stacked flat and primary buttons weird left-alignment - *not an issue the designs have changed* (although KButtonGroup should be refactored to accommodate more flexible designs)

6) Device permissions title overflow - I added `word-break: break-all` to that title. Not sure if this is the best way to handle this for all languages (or any languges for that matter @radinamatic) - now it just does this:

image
 
7) Device > Info table URL cut off - *not fixed here* 

8) Probably not fixable except by wrapping the text within the button (gross and more work than it's worth to avoid regressions probably)

9) Device > Settings title overflowing - *not fixed here* but with feedback on (6) can apply it here if we don't undo the other similar word-break fix.

## References

Partially fixes #7337 

## Reviewer guidance

Check it out - did I miss anything? Does it look okay in other non-German and non-English languages? How about RTL languages?

KButtonGroup could use some tweaks I think if we want to have it work so that it stacks. I think that IE11-safe flexbox might suit it well because we don't need anything from flex aside from basic flow-direction and content justification. 

Should the CoreTable in FacilitesPage just be a KGrid? Also I just thought maybe I could have used a colspan attr or something but if this works then it works :) 

----

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/plugins/device/assets/src/views/FacilitiesPage/index.vue
+++ b/kolibri/plugins/device/assets/src/views/FacilitiesPage/index.vue
@@ -3,14 +3,17 @@
   <div>
     <HeaderWithOptions :headerText="coreString('facilitiesLabel')">
       <template #options>
+        <!-- Margins to and bottom adds space when buttons are vertically stacked -->
         <KButtonGroup>
           <KButton
             :text="$tr('syncAllAction')"
+            style="margin-top: 16px; margin-bottom: -16px;"
             @click="showSyncAllModal = true"
           />
           <KButton
             :text="$tr('importFacilityAction')"
             primary
+            style="margin-top: 16px; margin-bottom: -16px;"
             @click="showImportModal = true"
           />
         </KButtonGroup>
@@ -29,7 +32,43 @@
         <th>{{ coreString('facilityLabel') }}</th>
       </template>
       <template #tbody>
-        <tbody>
+        <!-- On mobile, put buttons on a row of their own -->
+        <tbody v-if="windowIsSmall">
+          <template v-for="(facility, idx) in facilities">
+            <tr :key="idx" style="border: none!important">
+              <td>
+                <FacilityNameAndSyncStatus
+                  :facility="facility"
+                  :isSyncing="facilityIsSyncing(facility)"
+                  :isDeleting="facilityIsDeleting(facility)"
+                  :syncHasFailed="facility.syncHasFailed"
+                />
+              </td>
+            </tr>
+            <!-- May cause error on device with > 10000 facilities... -->
+            <tr :key="idx + 10000">
+              <td style="padding: 0 0 16px 0;">
+                <!-- Gives most space possible to buttons and aligns them with text -->
+                <KButtonGroup style="margin-left: -16px; margin-right: -16px; max-width: 100%">
+                  <KButton
+                    :text="coreString('syncAction')"
+                    appearance="flat-button"
+                    @click="facilityForSync = facility"
+                  />
+                  <KDropdownMenu
+                    :text="coreString('optionsLabel')"
+                    :options="facilityOptions(facility)"
+                    appearance="flat-button"
+                    @select="handleOptionSelect($event.value, facility)"
+                  />
+                </KButtonGroup>
+              </td>
+            </tr>
+          </template>
+        </tbody>
+
+        <!-- Non-mobile -->
+        <tbody v-else>
           <tr v-for="(facility, idx) in facilities" :key="idx">
             <td>
               <FacilityNameAndSyncStatus
@@ -112,6 +151,7 @@
 <script>
 
   import commonCoreStrings from 'kolibri.coreVue.mixins.commonCoreStrings';
+  import responsiveWindowMixin from 'kolibri.coreVue.mixins.responsiveWindowMixin';
   import commonSyncElements from 'kolibri.coreVue.mixins.commonSyncElements';
   import CoreTable from 'kolibri.coreVue.components.CoreTable';
   import { FacilityResource } from 'kolibri.resources';
@@ -153,7 +193,7 @@
       SyncAllFacilitiesModal,
       TasksBar,
     },
-    mixins: [commonCoreStrings, commonSyncElements, facilityTaskQueue],
+    mixins: [commonCoreStrings, commonSyncElements, facilityTaskQueue, responsiveWindowMixin],
     data() {
       return {
         showSyncAllModal: false,
--- a/kolibri/plugins/setup_wizard/assets/src/views/onboarding-forms/OnboardingForm.vue
+++ b/kolibri/plugins/setup_wizard/assets/src/views/onboarding-forms/OnboardingForm.vue
@@ -109,4 +109,9 @@
     margin-top: 24px;
   }
 
+  /deep/ .truncate-text {
+    // Ensure long text (like German) wrap and stay on screen
+    white-space: normal;
+  }
+
 </style>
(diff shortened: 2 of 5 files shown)


kolibri/plugins/device/assets/src/views/FacilitiesPage/index.vue
```diff
@@ -29,7 +32,43 @@
...
+                <FacilityNameAndSyncStatus
+                  :facility="facility"
+                  :isSyncing="facilityIsSyncing(facility)"
+                  :isDeleting="facilityIsDeleting(facility)"
+                  :syncHasFailed="facility.syncHasFailed"
+                />
+              </td>
+            </tr>
+            <!-- May cause error on device with > 10000 facilities... -->
+            <tr :key="idx + 10000">
```
rtibbles, line 49: I don't think this is a genuine worry, but you can always make a string key with a prefix instead `buttongroup-${idx}`

kolibri/plugins/setup_wizard/assets/src/views/onboarding-forms/OnboardingForm.vue
```diff
@@ -109,4 +109,9 @@
     margin-top: 24px;
   }
 
+  /deep/ .truncate-text {
```
rtibbles, line 112: I am guessing that a text truncator is being inserted into one of the slots here, and that's what this `/deep/` selector is targeting? Might we get better results using the new `TextTruncatorCSS` instead?

kolibri/plugins/device/assets/src/views/FacilitiesPage/index.vue
```diff
@@ -3,14 +3,17 @@
   <div>
     <HeaderWithOptions :headerText="coreString('facilitiesLabel')">
       <template #options>
+        <!-- Margins to and bottom adds space when buttons are vertically stacked -->
         <KButtonGroup>
           <KButton
             :text="$tr('syncAllAction')"
+            style="margin-top: 16px; margin-bottom: -16px;"
```
indirectlylit, line 10: added example of how this could be addressed globally here

https://github.com/learningequality/kolibri-design-system/pull/290

kolibri/plugins/setup_wizard/assets/src/views/onboarding-forms/OnboardingForm.vue
nucleogenesis: It's in the KRadioButton component - I'm not sure why as it seems a radio button label is better off wrapping but I'm sure there's something I'm missing. I think a follow-up issue in KDS can help address better

pcenov:
Tested the following points on this [build](https://buildkite.com/learningequality/kolibri-python-package/builds/4677) and everything is looking correctly and as described:
- Facilites list table - buttons being cut off. (Also adding space between the KButtonGroup buttons when stacked)
- Options and Import button hugging when stacked

About the following point: 6. 'Device permissions title overflow' - it seems to be implemented only at http://127.0.0.1:8080/de/device/#/settings But if I go to: http://127.0.0.1:8080/de/device/#/permissions I can still see the text not overflowing:


Tested both in Chrome's device emulator and on a real device (Samsung S7) with installed Kolibri app to the same result.

nucleogenesis:
Unfortunately, the result of the fix is that this is how multiple-word text will break:

image
 
So on permissions, the English being "Device Permissions" it is long and wraps, but "Device Settings" does not.

However, in Swahili, the Device Settings string breaks unnaturally as well with this fix.

I think that letting single-worded phrases that are too long run off for now is the ideal solution rather than making it so that the text is made a bit more illegible in other cases.

image

I'll be undoing the change to Device Settings and pushing, then merging for now.

marcellamaki:
> I think that letting single-worded phrases that are too long run off for now is the ideal solution rather than making it so that the text is made a bit more illegible in other cases.

This makes sense to me @nucleogenesis

indirectlylit:
> I think that letting single-worded phrases that are too long run off for now is the ideal solution rather than making it so that the text is made a bit more illegible in other cases.


Agreed. Wrapping behavior is also notoriously inconsistent across browsers (we've seen in cards for example) and it seems risky to try to change that at this point

Issue #8627, opened 2021-11-07
channel navigation header tears away from side nav


indirectlylit:
## Observed behavior

at some browser sizes the header tears away from the side-nav:

![2021-11-07 15 29 26](https://user-images.githubusercontent.com/2367265/140663929-35b051cc-81b4-498d-92ce-f6d5b95a58d0.gif)

## Expected behavior

I would expect the side nav to continue scrolling up to the top when the header disappears

## User-facing consequences

slightly broken-looking layout

## Steps to reproduce

set browser width to ~750 px and view resources in library browser

## Context

v0.15.0-beta1

indirectlylit:
@marcellamaki FYI this behavior is improved in your PR https://github.com/learningequality/kolibri/pull/8637

![2021-11-11 18 05 54](https://user-images.githubusercontent.com/2367265/141381658-56005516-3ba3-4088-b106-6fc171fc2bc1.gif)

MisRob:
This was related to issues caused by certain layout of elements within one breakpoint. The layout seems to haven been updated recently and I can't find conditions to reproduce this particular issue anymore. I made some improvements in #8947 that were meant as a fix for this issue originally, but even without them things seems to work pretty well now to me, so closing.

PR #8947, opened 2021-12-17
Sticky side navigation improvements

MisRob:
## Summary

This was originally fix for #8627. However, it seem that due to recent changes, behavior of the whole page is now different in regards to what is displayed at certain breakpoints and the problematic constellation of elements might be gone for now. or I wasn't able to find new conditions to reproduce it. However, I thought it might be good to submit the work, since it accounts for more scenarios in regards to hidden/visible top bar and also helped with jumping side navigation due to previously frequent styles updates.

## Reviewer guidance

Go to Library, select a channel, and scroll the page - it should scroll at first but as soon as the header is not visible, the side navigation should stick to the top of the page so that it's whole area is visible. It shouldn't get stuck or jump. Please test for more screen resolutions.

----

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/plugins/learn/assets/src/views/TopicsPage.vue
+++ b/kolibri/plugins/learn/assets/src/views/TopicsPage.vue
@@ -7,7 +7,7 @@
 
     <div v-else class="page">
       <!-- Header with thumbail and tagline -->
-      <div v-if="!windowIsSmall" class="header">
+      <div v-if="!windowIsSmall" ref="header" class="header">
         <KGrid>
           <KGridItem
             class="breadcrumbs"
@@ -237,6 +237,7 @@
         v-if="!!windowIsLarge"
         v-model="searchTerms"
         :topicsListDisplayed="!desktopSearchActive"
+        class="side-panel"
         topicPage="True"
         :topics="topics"
         :activeActivityButtons="activeActivityButtons"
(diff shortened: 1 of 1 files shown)


kolibri/plugins/learn/assets/src/views/TopicsPage.vue
```diff
@@ -659,21 +657,25 @@
...
-          let position = header.getBoundingClientRect();
-          let topbarPosition = topbar.getBoundingClientRect();
-          if (position.bottom >= 64) {
-            this.stickyTop = `${position.bottom}px`;
-          } else if (position.bottom < 0 && topbarPosition.bottom < 0) {
-            this.stickyTop = '0px';
-          } else {
-            this.stickyTop = '64px';
-          }
+        const header = this.$refs.header;
```
sairina, line 666: What is the difference between using the `header` ref above as compared to the original: `let header = document.getElementsByClassName('header')[0]`?

sairina (approved):
Manual testing shows the appropriate behavior for the side nav

kolibri/plugins/learn/assets/src/views/TopicsPage.vue
rtibbles: `let header = document.getElementsByClassName('header')[0]` seems like a riskier reference, as it could pick up a `header` class from anywhere in the DOM, whereas this will get the header specifically in this component.

rtibbles: (also `let` is unnecessary unless we are intending to reassign the `header` variable later)

MisRob: Yes, I made those changes to make things a bit safer in that regards.

PR #7797, opened 2021-02-17
Use `whitenoise` to pre-compress static files

rtibbles:
### Summary
* Uses whitenoise to serve precompressed static files when available
* Adds a kolibri-tools utility to precompress static files
* Adds a build step to compress all files in static folders

### Reviewer guidance
Does Kolibri still run properly?
Do static assets serve properly and with gzip encoding?

### References
Fixes #7791 

----

### Contributor Checklist

PR process:

- [ ] PR has the correct target branch and milestone
- [x] PR has 'needs review' or 'work-in-progress' label
- [x] If PR is ready for review, a reviewer has been added. (Don't use 'Assignees')
- [ ] If this is an important user-facing change, PR or related issue has a 'changelog' label
- [ ] If this includes an internal dependency change, a link to the diff is provided

Testing:

- [x] Contributor has fully tested the PR manually
- [ ] If there are any front-end changes, before/after screenshots are included
- [ ] Critical user journeys are covered by Gherkin stories
- [ ] Critical and brittle code paths are covered by unit tests

### Reviewer Checklist

- Automated test coverage is satisfactory
- PR is fully functional
- PR has been tested for [accessibility regressions](http://kolibri-dev.readthedocs.io/en/develop/manual_testing.html#accessibility-a11y-testing)
- External dependency files were updated if necessary (`yarn` and `pip`)
- Documentation is updated
- Contributor is in AUTHORS.md

PR #8958, opened 2021-12-19
Truncate uncompressed files when packaging

rtibbles:
## Summary
* Truncates compressed files after successful compression so that we save space in our whl file.
* Removes capability to inline JS and CSS assets - this seems OK, as it shifts a lot of assets that are shared across multiple SPA pages out of HTML, and should on the whole improve the load time. It also interacted badly with the truncated assets, as it would just inline those!

## References
Follow up to https://github.com/learningequality/kolibri/pull/7797 which introduced the pre-compression of static files, but did not remove uncompressed files.

## Reviewer guidance
This is a higher risk change than the other packaging PRs but seems to produce the most dramatic results on the final whl size.

In terms of impact on users, if the files are served as gzipped only, this should not impact users as this is supported by nearly all browsers: https://caniuse.com/sr_content-encoding-gzip

However, it is possible there may be quirks in the static file serving that cause it to not work as expected when the original files are missing.

As long as static assets load as expected from the built asset, this should be safe.

New WHL size: 87MB

**The main consideration here is that if a device explicitly requests the `identity` `Accept-Encoding` header it will return an empty file instead of the actual file. The other alternative is to return a 406 status code if the identity header is passed for one of these files, but this seems to be [explicitly disallowed](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept-Encoding).**

----

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -56,6 +56,7 @@ clean: clean-build clean-pyc clean-assets clean-staticdeps
 
 clean-assets:
 	yarn run clean
+	rm -fr kolibri/core/content/static/hashi/
 
 clean-build:
 	rm -f kolibri/VERSION
--- a/kolibri/core/hooks.py
+++ b/kolibri/core/hooks.py
@@ -26,9 +26,7 @@ from kolibri.plugins.utils import plugin_url
 
 @define_hook
 class NavigationHook(WebpackBundleHook):
-
-    # Set this to True so that the resulting frontend code will be rendered inline.
-    inline = True
+    pass
 
 
 @define_hook
--- a/kolibri/core/kolibri_plugin.py
+++ b/kolibri/core/kolibri_plugin.py
@@ -164,7 +164,6 @@ class FrontendHeadAssetsHook(WebpackBundleHook):
     """
 
     bundle_id = "frontend_head_assets"
-    inline = True
 
     def render_to_page_load_sync_html(self):
         """
--- a/kolibri/core/webpack/hooks.py
+++ b/kolibri/core/webpack/hooks.py
@@ -69,10 +69,6 @@ class WebpackBundleHook(hooks.KolibriHook):
     def bundle_id(self):
         pass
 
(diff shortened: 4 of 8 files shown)


nucleogenesis (commented):
> However, it is possible there may be quirks in the static file serving that cause it to not work as expected when the original files are missing.

I'm not sure how best to test this aside from smoke testing. I think that we'll probably want to get this in right away so that we have as many chances as possible during integration testing to find a potential situation where a quirk might show itself. Seems low risk enough though.

jredrejo (dismissed):
LGTM!

radinamatic:
I do like pink, but this amount makes me a bit nauseous 😅 

Firefox on Windows 7 

[logs-db.zip](https://github.com/learningequality/kolibri/files/7924644/logs-db.zip)


Win7 (start)  Running  - Oracle VM VirtualBox_004

rtibbles:
This does look rather like exactly the failure case I was imagining. This should not be merged.

rtibbles:
I think this is still recoverable, but will require a little bit of additional work.

rtibbles:
I have a strategy for achieving this, but it is probably best rebased onto https://github.com/learningequality/kolibri/pull/9047 first.

rtibbles:
Have uncovered another problematic interaction here with the inlining of JS and CSS assets.

rtibbles:
> Have uncovered another problematic interaction here with the inlining of JS and CSS assets.

Got around this by removing the inlining of assets - which I think might have been a bad idea in the first place.

rtibbles:
While I appreciate the cunning and inventiveness here @rtibbles - I think we'll close this as too risky as it requires circumventing HTTP standards.

I think it would be worthwhile to include the compression of html files that this adds in release-v0.15.x, and the removal of asset inlining in develop.

PR #8966, opened 2021-12-22
Add 0.15.0 changelog notes

marcellamaki:
Most is pulled from previous release notes. Since there was so much, I have a few levels of details. Am I missing anything critical that we want to highlight? Especially thinking about things that might have started before the summer that I'd be less aware of...

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,108 @@
 
 List of the most important changes for each release.
 
+
+## 0.15.0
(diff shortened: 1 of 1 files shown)


radinamatic (approved):
LGTM! :shipit: 👏🏽

jredrejo (approved):
LGTM, let's fix this asap. If any bullet point is missing it can be added later without any effort

PR #8976, opened 2022-01-06
frontend architecture docs correction re: default sass import

nucleogenesis:
This is a super simple update. I came across this error while grepping for the proper path for some other use and saw the docs were wrong.

### DIFF
--- a/docs/frontend_architecture/core.rst
+++ b/docs/frontend_architecture/core.rst
@@ -48,7 +48,7 @@ These modules would now be available for import anywhere with the following stat
 Styling
 -------
 
-To help enforce style guide specs, we provide global variables that can be used throughout the codebase. This requires including  ``@import '~kolibri.styles.definitions';`` within a SCSS file or a component's ``<style>`` block. This exposes all variables in ``definitions.scss``.
+To help enforce style guide specs, we provide global variables that can be used throughout the codebase. This requires including  ``@import '~kolibri-design-system/lib/styles/definitions';`` within a SCSS file or a component's ``<style>`` block. This exposes all variables in ``definitions.scss``.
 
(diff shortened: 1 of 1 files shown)


nucleogenesis:
@radinamatic I've not done much with docs lately - have you seen failures like this from simple changes before? (cc @rtibbles)

radinamatic:
I see nothing in your edits that would warrant this error, but during the release last month there must have been some rushed merge of the changes in the `CHANGELOG.md`, that introduced non-kosher formatting that is provoking this error. I suggest you delete the [repeated `###` in line 26](https://github.com/learningequality/kolibri/blob/release-v0.15.x/CHANGELOG.md?plain=1#L26), and properly format the lists of PRs below, to see if that resolves the error.

This also causes those PR titles to appear in the docs TOC:

Selection_002

rtibbles (approved):

rtibbles:
Docs failure is now fixed on release-v0.15.x merging this.

radinamatic:
The proper format of the lists of PRs in the 0.15 section is still pending, and therefore PR titles continue to appear in the ToC (built the project 2 times on RTD to be sure it's not cached).

I'll fix that in a separate PR.

rtibbles:
Is the bulleting here not sufficient to fix that? https://github.com/learningequality/kolibri/blob/release-v0.15.x/CHANGELOG.md#fixed

radinamatic:
You're correct, it should have been...
At this point the only difference I see from the 0.14 section is the indentation 🤔

radinamatic:
Or rather it most probably would be sufficient, if it was on the `develop` branch, from where the dev docs are built on RTD... 😅 

Selection_130

PR #3798, opened 2018-06-13
Move all core apps into core

rtibbles:
### Summary
Moves `auth`, `content`, `logger`, and `tasks` apps into `core`.

Updates all references to these apps to reflect the change in path.

### Reviewer guidance
Do tests pass?
Does Kolibri run?

Before completing each commit, I ran all the Python tests, and ran the devserver.

### References
Fixes #31 

----

### Contributor Checklist

- [x] Contributor has fully tested the PR manually
- [x] PR has the correct target branch and milestone
- [x] PR has 'needs review' or 'work-in-progress' label
- [x] If PR is ready for review, a reviewer has been added. (Don't use 'Assignees')
- [ ] If there are any front-end changes, before/after screenshots are included
- [ ] If this is an important user-facing change, PR or related issue has a 'changelog' label

### Reviewer Checklist

- [ ] Automated test coverage is satisfactory
- [ ] Reviewer has fully tested the PR manually
- [ ] PR has been tested for [accessibility regressions](http://kolibri-dev.readthedocs.io/en/develop/manual_testing.html#accessibility-a11y-testing)
- [ ] External dependencies files were updated (`yarn` and `pip`)
- [ ] Documentation is updated
- [ ] Link to diff of internal dependency change is included
- [ ] CHANGELOG.rst is updated for high-level changes
- [ ] Contributor is in AUTHORS.rst

PR #8999, opened 2022-01-14
reduce and fix migrations

jredrejo:
## Summary

- Strip initial migrations (touching the non-existing DeviceOwner model/tables) that are causing random problems when running tests with Postgresql on Docker.
- Adds one more error exception to migration helper

## References

These migrations deleted an old model that has not been used since, at least, #3798 , i.e. Jun 2018 (likely much before as they are initial migrations, but the history has been lost when  they were moved from auth to auth.core) , so it should not affect current users.

This PR is targed to 0.14.7 so we could use it in current stable KDP installations in case a new 0.14.x release is done.

## Reviewer guidance
* do tests pass?
* might any possible old user be affected by this change?

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated


kolibri/core/auth/migrations/0004_auto_20170816_1607.py
```diff
@@ -7,69 +7,6 @@
 from kolibri.core.auth.constants.role_kinds import ADMIN
 
 
-def device_owner_to_super_user(apps, schema_editor):
```
rtibbles, line 10: Deleting this migration would drop support for upgrading from 0.6.0 and below. I think this is OK, but I think it's useful for us to be explicit that this is happening.

rtibbles: The other consideration here is that if we are preventing this upgrade path, it actually gives us license to make more aggressive edits to the migrations - we could prevent the DeviceOwner table from ever being created in the first place.

jredrejo: I just wanted to remove the code that was causing problems in the tests, to be as less intrusive as possible. Of course, having the DeviceOwner table in the previous migrations does not make sense anymore from a code point of view, so I'm happy to delete it :)

rtibbles: Looks like we need to remove references to it from the tests for it too: `kolibri/core/auth/test/test_data_migrations.py`

jredrejo: Any problem in removing totally this test?
I don't see the point of it if the migration between models does not exist any more and it's gonna be really hard to make it pass because using `DevicePermissions.objects.create(user=deviceowner, is_superuser=True)` requires the user to be saved, but the user can't be saved at that moment of the migrations because `FacilityUser.objects.create` adds `gender` or `birth_year` fields that can't be saved as they didn't exist in the db yet.

rtibbles (approved):
Failures are unrelated to this PR, merging.

PR #9000, opened 2022-01-14
Survey modality

nucleogenesis:
## Summary

Introduces Survey modality support by adding some display logic based on whether the content is a Survey or not. 

- No "correct" / "incorrect" iconography on the answering or reporting
- Simplified CurrentTryOverview - only use that for surveys because TriesOverview summarizes multiple attempts which we don't need here.

It's not super complex logic there in CurrentTryOverview, but I wanted to be sure it was covered and started writing tests for the bits I was going to change and I figured that since I was there I should cover all of the props and display logic. 

https://user-images.githubusercontent.com/6356129/149592925-83f1b683-b021-4d55-b750-6b85c47015f1.mp4

## References

[Figma](https://www.figma.com/file/PxwIrfaqoVeigF19AvQsDj/Surveys?node-id=2%3A1013)

## Reviewer guidance

I tried to make the video cover everything so you don't have to do all this but I think it'd be good if someone spun this up.

### Setup

- Be sure to yarn install and pip install requirements.txt so you have the latest LE Utils & Kolibri constants.
- Make sure you have the latest Kolibri QA Channel
- Go to Exercises -> Channel based quizzes 
- Pick a piece of content from there that you want to update and update it's `options: { modality: SURVEY }` value. I do this in SQLite Browser, but you can probably do it in the Kolibri Shell getting the content node by id, updating it's `extra_fields` JSON accordingly.

### Actually testing
 
- Review the Figma
- Check mobile
- Take multiple tries
- Test while not logged in

----

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated


kolibri/core/assets/src/views/ExamReport/CurrentTryOverview.vue
```diff
@@ -128,9 +129,15 @@
...
       },
+      // Whether the modality is a Survey - conditionalizes styles
+      isSurvey: {
+        type: Boolean,
+        default: false,
+      },
     },
     computed: {
       ...mapGetters(['currentUserId']),
+      // QUESTION: Why? Wouldn't anything else fail validation? Why not just call tryValidator here?
```
rtibbles: Validators don't run in production, only in development. This value can still be undefined, so we need to handle that case. Could swap it out for `tryValidator` instead, which might be more robust.

kolibri/core/assets/src/views/ExamReport/CurrentTryOverview.vue
```diff
@@ -171,7 +175,9 @@
         return this.currentTryDefined && !isUndefined(this.currentTry.correct);
       },
       score() {
-        return this.currentTryDefined ? this.currentTry.correct / this.totalQuestions : 0;
```
rtibbles, line 174: Have lost the division here in the small refactor.

kolibri/core/assets/src/views/ExamReport/index.vue
```diff
@@ -403,6 +421,7 @@
           return;
         }
         this.loading = true;
+        console.log(this.getParams());
```
rtibbles: Spare `console.log`

rtibbles (dismissed):
Spotted 1 potential regression in code review. Have not manually tested yet.

kolibri/core/assets/src/views/ExamReport/CurrentTryOverview.vue
nucleogenesis: If the component defines a prop as required and validates it then I think it should be able to operate under the assumption that it won't be used incorrectly. The parent component should (and I think in all/most cases do - can confirm in a bit) guard rendering CurrentTryOverview w/ v-if

kolibri/core/assets/src/views/ExamReport/CurrentTryOverview.vue
nucleogenesis: Will update the test scenario to be more robust thank you!

kolibri/core/assets/src/views/ExamReport/CurrentTryOverview.vue
rtibbles: I don't disagree - but it's not how Vue handles it, so better to be defensive, I think.

pcenov:
Hi @radinamatic and @nucleogenesis, I tried testing this PR with all of the[ Buildkite builds](https://buildkite.com/learningequality/kolibri-python-package/builds/4807) plus directly from the nucleogenesis:survey-modality branch but I am not able to see any of the changes.
I tested with all the available survey exercises and also updated the `options: { modality: SURVEY } `value of an exercise using the SQLite Browser and still I'm not seeing any of the new functionality. Wondering am I missing something? :)

marcellamaki:
@pcenov and @nucleogenesis -- I'm running into a similar problem after having updated an exercise to the SURVEY modality while trying to do some manual QA after a code read through this morning (which looks good!). One thing I did notice as I was troubleshooting is that Modalities.SURVEY coming from `le-utils` seems to be undefined when console logging it. I can't quite sort that out yet, since I've double checked and I have the new `le-utils` version and I checked the PR for the `le-utils` version and it all makes sense reading through. I will do a little more investigating and see if I can find anything else!

rtibbles:
@nucleogenesis looks like you updated the version of `le-utils` but not its corollary JS package `kolibri-constants`. Need to update both! (because of the jank in kolibri-core, will need to update the version in both places it appears in two separate package.jsons)

kolibri/plugins/learn/assets/src/views/CompletionModal/index.vue
```diff
@@ -447,7 +478,7 @@
         message: 'Open the quiz report to review your answers',
         context: 'After learner submitted a practice quiz, they can view the report page.',
       },
-      reviewQuizButtonLabel: {
+      reviewReportButtonLabel: {
```
rtibbles: To avoid faff, probably best not to change message ids right now?

kolibri/plugins/learn/assets/src/views/QuizRenderer/QuizReport.vue
```diff
@@ -84,6 +91,10 @@
         message: 'Try again',
         context: 'Label for a button used to retake the quiz',
       },
+      submitAgainButton: {
+        message: 'Submit again',
+        context: 'Label for a button used to retake the survey',
```
rtibbles:
```suggestion
        context: 'Label for a button used to resubmit the survey',
```

kolibri/plugins/learn/assets/src/views/QuizRenderer/index.vue
```diff
@@ -462,6 +466,11 @@
         context:
           'Action that learner takes to submit their quiz answers so that the coach can review them.',
       },
+      submitSurvey: {
+        message: 'Submit survey',
+        context:
+          'Action that learner takes to submit their exam answers so that the coach can review them.',
```
rtibbles:
```suggestion
          'Action that learner takes to submit their survey answers so that they can be reviewed.',
```

kolibri/plugins/learn/assets/src/views/CompletionModal/index.vue
```diff
@@ -439,6 +462,14 @@
         message: 'Stay here',
         context: 'Label for a button used if learner decides to repeat the completed resource.',
       },
+      reviewSurveyTitle: {
+        message: 'Review survey',
+        context: 'Message to the user to review a survey after they completed it.',
+      },
+      reviewSurveyDescription: {
+        message: 'Open the survey report to review your answers',
+        context: 'After learner submitted a practice survey, they can view the report page.',
```
rtibbles:
```suggestion
        context: 'After learner submitted a survey, they can view the report page.',
```

radinamatic:
Tested running from source (@pcenov please re-test with DEB and EXE assets), and seeing various issues in Firefox on Linux.

> when looking at reports, (surveys) should not display any correct/incorrect information, any scores, or completion criteria.

1. Not seeing any completion criteria, but scores are visible in the `Attempts` dropdown when surveys are submitted multiple times. 
2. When survey resources are browsed in their original channel (`gutuf-potuv` token), all questions display correctly. However, when I try to browse those same resources that are imported in the Kolibri QA channel, the first question (Teacher ID?) never loads properly, but the rest of them do. At one time I caught a glimpse of the `Firefox can’t establish a connection to the server at ws://127.0.0.1:3016/sockjs-node/889/0vwz1pqo/websocket` error that I never saw before 😕 

1 | 2
--- | ---
survey4 | LEDev2111 (start)  Running  - Oracle VM VirtualBox_005

There is also a bit of weirdness in displaying the question area when changing browsing width, and hint appearance seems to be affecting the width of the paragraph (shifting may be affecting readability), but those could be issues with the channel/ricecooker, and not this PR.

1 | 2
--- | ---
survey2 | survey3

pcenov:
In addition to the issues observed by @radinamatic I've reported separately the following issues: https://github.com/learningequality/kolibri/issues/9020, https://github.com/learningequality/kolibri/issues/9021, https://github.com/learningequality/kolibri/issues/9022, https://github.com/learningequality/kolibri/issues/9023
I didn't find any regressions for the learner facing practice quiz scenarios.

rtibbles:
I think we need to fix #9020 and #9023 - before merging the PR. The other two are intended behaviour at the moment as far as I can tell.

nucleogenesis:
@pcenov @radinamatic 

The following fixes have been pushed here:

- Questions ordering in report is now correct
- Infinite loader on submission without answers - now the user sees the dropdown to look through past submissions only
- (Unreported) the "submit" button was not visible so I added some padding on the bottom for small screens. It was getting hidden under the arrow icons at the bottom of the screen
- Squished renderer of the content
- % removed from list of attempts

I could not replicate the issue where the first question stopped working when accessed through the QA channel because I cannot upgrade my channel. I'll see if I can get it working later on.

radinamatic:
1.  
   - Questions ordering in report is now correct ✔️ 
   - Infinite loader on submission without answers - now the user sees the dropdown to look through past submissions only ✔️ 
   - (Unreported) the "submit" button was not visible so I added some padding on the bottom for small screens. It was getting hidden under the arrow icons at the bottom of the screen ✔️ 
   - % removed from list of attempts ✔️ 

2. Squished renderer of the content 🔴 

   Mobile view looks OK, but the issue is still present while resizing on wider screens, or on tablet sizes in landscape mode:

   Firefox | Chrome
   --- | ---
   Selection_112 | Selection_113

   https://user-images.githubusercontent.com/1457929/150770035-9e440611-cc65-4fe4-a974-417ac41caf5e.mp4

3. Hints

   Are still appearing, are those going away? And I did notice that the ID fields are not counting answers that contain letters as _answered_. Did we agree with partners that the student and teachers IDs can contain only numbers @rtibbles?

   https://user-images.githubusercontent.com/1457929/150771675-ec5cada7-b763-4916-8b25-401a43a1b79a.mp4

4. Last thing: seems possible to open the survey even if the user is not signed in (c/p URL), oversight or meant to be...? 🤔

kolibri/plugins/perseus_viewer/assets/src/views/PerseusRendererIndex.vue
```diff
@@ -18,9 +18,10 @@
           <!-- Layout notes
             - Layout12 span8 -> ~66% width on windowIsLarge
             - No other layout definitions means span will be 100%
-            - When not interactive, this should be full width
+            - If we're not allowingHints, then it should be 100% because
+              they're the reason we'd go smaller at all
           -->
-          <KGridItem :layout12="{ span: interactive ? 6 : 12 }">
+          <KGridItem :layout12="{ span: allowHints || interactive ? 6 : 12 }">
```
rtibbles: `allowHints` defaults to `true` so this means that `interactive` will still not have any effect. Needs to be an `&&`.

https://github.com/learningequality/kolibri-design-system/blob/main/lib/content/mixin.js#L103

rtibbles:
> Are still appearing, are those going away? And I did notice that the ID fields are not counting answers that contain letters as answered. Did we agree with partners that the student and teachers IDs can contain only numbers @rtibbles?

Tooltips are not hints. There's nothing that should be stopping the tooltips from appearing, so this is expected behaviour.
It is a numeric input box, so it can only contain numbers - we have communicated this.

> Last thing: seems possible to open the survey even if the user is not signed in (c/p URL), oversight or meant to be...? thinking

Users can take surveys anonymously - I'd say this is intended behaviour for now (even if potentially undesirable in some contexts).

radinamatic:
Installed the APK asset on an Android tablet very similar to the one partner will be using, and content is rendering fine (no squishing), as are apparently all the other survey features 💯 

The only doubt remains why the resources in this channel are not displaying correctly when imported in other channels...? 🤔 
To recap:
> When survey resources are browsed in their original channel (`gutuf-potuv` token), all questions display correctly. However, when I try to browse those same resources that are imported in the Kolibri QA channel, most of the time I get just the loader spinning

To exclude this being a QA channel quirk, I created a brand new channel (`jotuh-mojop` token) and imported only the 4 survey resources from the source channel, but no change. I even thought this might be because the `Allow as channel quiz` option (which is present in the source channel) is not automatically set up when the resources are imported in another channel, and went to check and apply them manually on my new channel, but in Kolibri the only result was to have the counter appear (as it should with channel quizzes), and the resources still do not display correctly. 

The frustrating part is that hardly anything appears as error in the console... 😒 

Source channel | New channel
--- | ---
greece-full-cok | greece-cok

If we are sure that the partner will only be using the `gutuf-potuv` token to view these resources on their tablets, this might not be a blocker, but if they are creating derivative channels, trouble... 😕

rtibbles:
The error looks like an issue with loading the next question inside Perseus - so it is probably being passed a question identifier that does not exist inside the perseus file.

radinamatic:
Ok, since we decided this is not a blocker, let's merge this in! 👍🏽

radinamatic (approved):
Manual QA cleared up, good to go! 🎉

PR #9002, opened 2022-01-18
Content API: Change default ordering to "lft"

manuq:
The API should return the nodes ordered in the same way as they where
authored in Kolibri Studio.

## Reviewer guidance
Create a custom presentation for a channel. Then check that the ordering is the same as the default Kolibri presentation. Which at the same time should be the same ordering as in Kolibri Studio. Here are before/after screenshots:

Before:
regression-before

After:
regression-after

And default (matches after):
regression-default

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/core/content/api.py
+++ b/kolibri/core/content/api.py
@@ -555,14 +555,11 @@ class BaseContentNodeMixin(object):
 
 
 class OptionalPagination(ValuesViewsetCursorPagination):
-    ordering = "id"
+    ordering = ("lft", "id")
     page_size_query_param = "max_results"
 
 
 class OptionalContentNodePagination(OptionalPagination):
-    ordering = "id"
-    page_size_query_param = "max_results"
-
     def paginate_queryset(self, queryset, request, view=None):
         # Record the queryset for use in returning available filters
         self.queryset = queryset


kolibri/core/content/api.py
```diff
@@ -555,12 +555,12 @@ def consolidate(self, items, queryset):
 
 
 class OptionalPagination(ValuesViewsetCursorPagination):
-    ordering = "id"
+    ordering = "lft"
```
rtibbles:
```suggestion
    ordering = ("lft", "id")
```
This seems to have minimal impact on performance and will give a unique ordering every time.

kolibri/core/content/api.py
```diff
@@ -555,12 +555,12 @@ def consolidate(self, items, queryset):
...
     page_size_query_param = "max_results"
 
 
 class OptionalContentNodePagination(OptionalPagination):
-    ordering = "id"
```
rtibbles, line 563: These two lines don't actually need to be here, can just delete them both, as they are inheriting from above.

rtibbles (dismissed):
Doing the joint ordering seems to be the best solution here.

manuq:
Thanks! Suggestion applied and comment addressed.

rtibbles (approved):
QA found no regressions, this is good to go.

rtibbles:
Unrelated build failure in the Debian build. Merging.

PR #9029, opened 2022-01-21
Fix focus for mobile full screen side panel; update `CoreMenu`, `CompletionModal`, and `FocusTrap` API

### LINKED ISSUE #8992: Side panel should gain focus when opened w/keyboard
The side panel does not have focus when a user using a keyboard as input modality opens it via the "Filter" button on the Library page (or the "Folders" button on the Topics page). The keyboard focus is still on the main content underneath the side panel.

Impact:
User is unable to use the filters in the side panel in "Filter"
User is unable to access folders in the side panel in "Folders" (on the Topics page)

keyboard-navigation-sidebar

_Originally posted by @radinamatic in https://github.com/learningequality/kolibri/issues/8986#issuecomment-1009674923_

sairina:
## Summary
- Fixes missing focus on side panels that are opened on mobile.
- Add `FocusTrap` component in `FullScreenSidePanel`
- Updates `FocusTrap` API to emit events rather than pass props
- Updates existing implementation of components that use `FocusTrap` to emit events
  - `CoreMenu` in `AppBar`
  - `CoreMenu` in `SideNav`
  - `CoreMenu` in `LearningActivityBar`
  - `CoreMenu` in `SyncInterface` (gif not provided)
  - `CompletionModal`

|Library: filter/search|Library: info panel|
|--|--|
|||

|Topics: folders|Topics: info panel|
|--|--|
|||

|Completion modal|
|--|
||

|Core menu in side nav and app bar|
|--|
||

|Learning activity bar - core menu in mobile|
|--|
||

## References
Fixes #8992

## Reviewer guidance
Please check any uses of the `FullScreenSidePanel`.
1. Log in as learner and make sure you are on a responsive/mobile screen
2. Click in the library tab

### Library Page
**Filter button**
1. Using the keyboard, navigate to the "filter" button
2. Once it opens, use tab to navigate through all the clickable elements
3. Make sure that the focus is "trapped" in the side panel and loops back to the first element after getting to the last element (the "x"/close icon)
4. Make sure that the "x"/close icon works via keyboard. 

### Topics Page
1. Click into a channel via the content card

**Filter button** (see above)
**Folders button**
1. Using the keyboard, navigate to the "folders" button
2. Once it opens, use tab to navigate through all the folders links
3. Make sure that the focus is "trapped" in the side panel and loops back to the first element after getting to the last element (the "x"/close icon)
4. Make sure that the "x"/close icon works via keyboard

**Info button on content card**
1. Using the keyboard, navigate to a content card, and then to the "i"/info icon
2. Once it opens, use tab to navigate through the clickable elements
3. Make sure that the focus is "trapped" in the side panel and loops back to the first element after getting to the last element (the "x"/close icon)
4. Make sure that the "x"/close icon works via keyboard

### Completion modal
1. As a learner, complete any activity
2. When the completion modal appears, make sure that a user can navigate through it using the keyboard

### Core menu in side nav and app bar
1. Sign in as a learner
2. Navigate with the keyboard to the user menu button on the righthand side of the app bar
3. Enter and make sure you can navigate through all options with the keyboard
4. ESC to exit and navigate to the hamburger menu on the left to open the side nav
5. Enter and make sure you can navigate through all options with the keyboard

### Learning activity bar - core menu in mobile
1. In mobile view, open an activity
2. In the Learning Activity Bar at the top of the activity, use the keyboard to navigate to the three dots horizontal icon
3. When you press enter, make sure that the focus can navigate through all options

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/core/assets/src/views/CoreMenu/index.vue
+++ b/kolibri/core/assets/src/views/CoreMenu/index.vue
@@ -20,8 +20,8 @@
         ref="focusTrap"
         class="ui-menu-options"
         :disabled="!containFocus"
-        :firstEl="firstFocusableEl"
-        :lastEl="lastFocusableEl"
+        @shouldFocusFirstEl="$emit('shouldFocusFirstEl')"
+        @shouldFocusLastEl="focusLastEl"
       >
         <slot name="options"></slot>
       </FocusTrap>
@@ -77,12 +77,6 @@
         showActive: this.showActive,
       };
     },
-    data() {
-      return {
-        firstFocusableEl: null,
-        lastFocusableEl: null,
-      };
-    },
     computed: {
       classes() {
         return {
@@ -102,9 +96,25 @@
       // make sure that all child components have been mounted
       // before attempting to access their elements
       this.$nextTick(() => {
-        this.firstFocusableEl = this.$el.querySelector('.core-menu-option');
-        this.lastFocusableEl = last(this.$el.querySelectorAll('.core-menu-option'));
+        // this.$nextTick(this.$refs.modal.focus());
+        this.focusFirstEl();
       });
+    },
+    destroyed() {
+      window.setTimeout(() => this.lastFocus.focus());
+    },
+    methods: {
+      focusLastEl() {
+        last(this.$el.querySelectorAll('.core-menu-option')).focus();
+      },
+      /**
+       * @public
+       * Focuses on correct first element for FocusTrap depending on content
+       * rendered in CoreMenu.
+       */
+      focusFirstEl() {
+        this.$el.querySelector('.core-menu-option').focus();
+      },
     },
   };
 
--- a/kolibri/core/assets/src/views/CoreMenu/index.vue
+++ b/kolibri/core/assets/src/views/CoreMenu/index.vue
@@ -92,11 +92,13 @@
         }
       },
     },
+    beforeMount() {
+      this.lastFocus = document.activeElement;
+    },
     mounted() {
       // make sure that all child components have been mounted
       // before attempting to access their elements
       this.$nextTick(() => {
-        // this.$nextTick(this.$refs.modal.focus());
         this.focusFirstEl();
       });
     },
@@ -104,7 +106,13 @@
       window.setTimeout(() => this.lastFocus.focus());
     },
     methods: {
+      /**
+       * @public
+       * Focuses on correct last element for FocusTrap depending on content
+       * rendered in CoreMenu.
+       */
       focusLastEl() {
+        this.$emit('shouldFocusLastEl');
         last(this.$el.querySelectorAll('.core-menu-option')).focus();
       },
       /**
@@ -113,7 +121,9 @@
        * rendered in CoreMenu.
        */
       focusFirstEl() {
-        this.$el.querySelector('.core-menu-option').focus();
+        if (this.$el.querySelector('.core-menu-option')) {
+          this.$el.querySelector('.core-menu-option').focus();
+        }
       },
     },
   };
--- a/kolibri/core/assets/src/views/CoreMenu/index.vue
+++ b/kolibri/core/assets/src/views/CoreMenu/index.vue
@@ -112,7 +112,6 @@
        * rendered in CoreMenu.
        */
       focusLastEl() {
-        this.$emit('shouldFocusLastEl');
         last(this.$el.querySelectorAll('.core-menu-option')).focus();
       },
       /**
--- a/kolibri/core/assets/src/views/FocusTrap.vue
+++ b/kolibri/core/assets/src/views/FocusTrap.vue
@@ -59,15 +59,27 @@
         e.stopPropagation();
         if (!this.isTrapActive) {
           // On first focus, redirect to first option, then activate trap
-          this.firstEl.focus();
+          this.focusFirstEl();
           this.isTrapActive = true;
         } else {
-          this.lastEl.focus();
+          this.focusLastEl();
         }
       },
       handleLastTrapFocus(e) {
         e.stopPropagation();
-        this.firstEl.focus();
+        this.focusFirstEl();
+      },
+      focusFirstEl() {
+        this.$emit('shouldFocusFirstEl');
+        if (this.firstEl) {
+          this.firstEl.focus();
+        }
+      },
+      focusLastEl() {
+        this.$emit('shouldFocusLastEl');
+        if (this.lastEl) {
+          this.lastEl.focus();
+        }
       },
       /**
        * @public
--- a/kolibri/core/assets/src/views/FocusTrap.vue
+++ b/kolibri/core/assets/src/views/FocusTrap.vue
@@ -33,16 +33,6 @@
   export default {
     name: 'FocusTrap',
     props: {
-      firstEl: {
-        type: HTMLElement,
-        required: false,
-        default: null,
-      },
-      lastEl: {
-        type: HTMLElement,
-        required: false,
-        default: null,
-      },
       disabled: {
         type: Boolean,
         required: false,
@@ -71,15 +61,9 @@
       },
       focusFirstEl() {
         this.$emit('shouldFocusFirstEl');
-        if (this.firstEl) {
-          this.firstEl.focus();
-        }
       },
       focusLastEl() {
         this.$emit('shouldFocusLastEl');
-        if (this.lastEl) {
-          this.lastEl.focus();
-        }
       },
       /**
        * @public
--- a/kolibri/core/assets/src/views/FullScreenSidePanel/index.vue
+++ b/kolibri/core/assets/src/views/FullScreenSidePanel/index.vue
@@ -2,53 +2,61 @@
 
   <div
     ref="sidePanel"
+    :tabindex="0"
     :class="{ 'is-rtl': isRtl, 'is-mobile': isMobile }"
     @keyup.esc="closePanel"
   >
     <transition name="side-panel">
-      <div
-        class="side-panel"
-        :style="sidePanelStyles"
+      <FocusTrap
+        :firstEl="firstFocusableEl"
+        :lastEl="lastFocusableEl"
       >
-
-        <!-- Fixed header with optional close button -->
         <div
-          v-show="$slots.header"
-          ref="fixedHeader"
-          class="fixed-header"
-          :style="fixedHeaderStyles"
+          class="side-panel"
+          :style="sidePanelStyles"
         >
-          <div class="header-content" tabindex="0">
-            <slot name="header">
-            </slot>
+
+          <!-- Fixed header with optional close button -->
+          <div
+            v-show="$slots.header"
+            ref="fixedHeader"
+            class="fixed-header"
+            :style="fixedHeaderStyles"
+          >
+            <div class="header-content">
+              <slot name="header">
+              </slot>
+            </div>
           </div>
+
+          <!-- Default slot for inserting content which will scroll on overflow -->
+          <div class="side-panel-content" :style="contentStyles">
+            <slot></slot>
+          </div>
+
+          <KIconButton
+            v-if="fullScreenSidePanelCloseButton"
+            ref="closeButton"
+            icon="close"
+            class="close-button"
+            :style="closeButtonFullScreenSidePanelStyles"
+            :ariaLabel="coreString('closeAction')"
+            :tooltip="coreString('closeAction')"
+            @click="closePanel"
+          />
+          <KIconButton
+            v-else
+            ref="closeButton"
+            icon="close"
+            class="close-button"
+            :style="closeButtonStyles"
+            :ariaLabel="coreString('closeAction')"
+            :tooltip="coreString('closeAction')"
+            @click="closePanel"
+          />
+
         </div>
-
-        <KIconButton
-          v-if="fullScreenSidePanelCloseButton"
-          icon="close"
-          class="close-button"
-          :style="closeButtonFullScreenSidePanelStyles"
-          :ariaLabel="coreString('closeAction')"
-          :tooltip="coreString('closeAction')"
-          @click="closePanel"
-        />
-        <KIconButton
-          v-else
-          icon="close"
-          class="close-button"
-          :style="closeButtonStyles"
-          :ariaLabel="coreString('closeAction')"
-          :tooltip="coreString('closeAction')"
-          @click="closePanel"
-        />
-
-        <!-- Default slot for inserting content which will scroll on overflow -->
-        <div class="side-panel-content" :style="contentStyles">
-          <slot></slot>
-        </div>
-
-      </div>
+      </FocusTrap>
     </transition>
 
     <Backdrop
@@ -66,11 +74,13 @@
   import Backdrop from 'kolibri.coreVue.components.Backdrop';
   import commonCoreStrings from 'kolibri.coreVue.mixins.commonCoreStrings';
   import responsiveWindowMixin from 'kolibri.coreVue.mixins.responsiveWindowMixin';
+  import FocusTrap from 'kolibri.coreVue.components.FocusTrap';
 
   export default {
     name: 'FullScreenSidePanel',
     components: {
       Backdrop,
+      FocusTrap,
     },
     mixins: [responsiveWindowMixin, commonCoreStrings],
     props: {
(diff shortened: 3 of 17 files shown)


marcellamaki (commented):
This looks great overall. The focus trap implementation seems to be working in each of the scenarios, and this is a huge improvement to the user experience

I _think_ this is a **non-blocking piece of feedback**, but one note: on opening the filter panel, the user "loses focus" and has to tab twice to get into the search bar. It's hard to capture in a gif, because you can't actually see me tabbing but it's here: 
tab-filter-panel

@radinamatic maybe you can confirm if updating this ^^ to just one tab needed is "essential" or a "nice to have".

My initial thought of what might be causing this is the tabindex="0" on the whole div at the very top of the file, which is adding the div itself to the tab order.

```
<div
    ref="sidePanel"
    :tabindex="0"
    :class="{ 'is-rtl': isRtl, 'is-mobile': isMobile }"
    @keyup.esc="closePanel"
  >
  ``` 
  
If you skip adding the tabindex="0" here, the focus won't shift to the side panel, so simply removing it wouldn't work. I've done a little bit of googling of how to shift the focus directly to a child element, but I haven't found anything just yet. Not sure if @MisRob might have a suggestion/experience with this. I can investigate a little bit more, or we can merge this (pending ✅ from Radina), and add that as a follow up.

Great work 👏

**Comment by MisRob** [comment]
I don't think we can get rid of `tabindex=0` on the whole `div` as we need to make that `div` focusable. `tabindex=0` is used in the user menu

Screenshot from 2022-01-26 13-00-24

as well where only one TAB keypress is needed to focus on the first option so I don't think it should be, on its own, causing the problem with two TABs needed at this point.

**Comment by MisRob** [comment]
Here's the sequence of my actions and elements that have focus as I'm navigating with the keyboard:

(1) I press TAB and see the focus ring around the "FILTER" button

`document.activeElement=button.button.KButton-noKey-0_1tum2rb.raised`

(2) I press ENTER on the "FILTER" button

`document.activeElement=div.full-screen-side-panel`

(3) I press TAB

`document.activeElement=form.search-box`

(4) I press TAB again

`document.activeElement=input#searchfield.search-input.SearchBox-noKey-0_1jjfqpe`

From above, I'd say that the extra TAB press is needed because of the search form (3) - the focus goes first into the element with the form and I need to press TAB again to go to the input in the form.

File: kolibri/plugins/learn/assets/src/views/EmbeddedSidePanel/index.vue [hunk 1]
```diff
@@ -34,6 +34,7 @@
       <SearchBox
         key="channel-search"
         placeholder="findSomethingToLearn"
+        :tabindex="0"
```
MisRob: What was the reason for adding `tabindex="0"` on the search box `form` element? Wondering if this may be related to problems with two TAB key presses needed.

kolibri/core/assets/src/views/FullScreenSidePanel/index.vue
```diff
@@ -171,24 +184,44 @@
...
     methods: {
       closePanel() {
         this.$emit('closePanel');
       },
+      setFocusTrap() {
+        if (this.$refs.sidePanel && !this.$refs.sidePanel.contains(document.activeElement)) {
+          this.focusSidePanel();
+        }
+        this.firstFocusableEl =
+          this.$el.querySelector('.search-box') || // if `EmbeddedSidePanel` is for search/filter btn
```
MisRob: I'd recommend moving focus logic specific to the embedded side panel to the `EmbeddedSidePanel` component as we chatted briefly earlier. These dependencies seem to be brittle and I can imagine that focus logic could stop working silently - for example, if we updated some classes in the `EmbeddedSidePanel`, we may easily not notice that some updates are needed in the `FullScreenSidePanel` as well.

I recall that you mentioned that we don't want to use focus trapping when `EmbeddedSidePanel` is used on its own, outside of `FullScreenSidePanel`. I'm not sure which way would be better feasible here off the top of my head, but can imagine two directions (their combination might be needed) to play with that could help us with moving the `FocusTrap` component and related to the `EmbeddedSidePanel`:

- Providing `setFocusTrap` as a public method from the `EmbeddedSidePanel` to allow other components to call it (you can search for JSDoc `@public` tag in our codebase to see some examples of how we use public methods)
- Adding `Boolean` `enableFocusTrap` property to the `EmbeddedSidePanel` while setting it to `true` only when used inside of the `FullScreenSidePanel`. Focus-related logic would then be only run in the `EmbeddedSidePanel` when `enableFocusTrap` would be `true` and skipped otherwise. It should be possible to propagate this information into the `FocusTrap` component since it provides `disabled` property.

MisRob (changes requested):
Thank you @sairina, this is a great start on focus issues that are always tricky to debug, as we found out together as well :)

I'm requesting changes for issues I mentioned above since I'd say that it'd be good to deal with them rather sooner than later so that we're not adding tech debt in times of trying to reduce it. If you encounter some blockers or if it shows important to have this merged for the next patch release since it's still a great improvement, I'm fine with merging it as long as there's a follow-up issue open.

kolibri/core/assets/src/views/FullScreenSidePanel/index.vue
bjester: Going off what @MisRob said, a public method on `EmbeddedSidePanel` to focus it on the appropriate internal element makes sense to me. 

There's a pattern we can follow in plain HTML-- you can think of `<input>` as just another Vue component, so when you want to focus on it, you call that method `.focus()`. That allows the component to decide how to handle that appropriately.

The other case is when the element or Vue component wraps other elements or components, like the `FullScreenSidePanel`. In this instance, you want it to trap the focus, so implementing `FocusTrap` would either be inside this component or in whatever component defines the wrapped content, since that component has access to its children. In this case, I think the pattern is to use events. So `FullScreenSidePanel` could emit and event like `@shouldFocus` to trigger focusing on the appropriate element.

Although, the current `FocusTrap` implementation relies on access to the component children, which conflicts with the use case of a wrapper component. To make it work with this approach, it could also emit events when the first and last elements should be focused and avoid requiring the elements as props.

So as an example, it could look like this:
```html
<FullScreenSidePanel @shouldFocus="$refs.embedded.focus()">
  <EmbeddedSidePanel ref="embedded" />
</FullScreenSidePanel>
```
where `$emit("shouldFocus")` occurs in the `$nextTick` on `mount()` and on `<FocusTrap @shouldFocusFirstEl="$emit('shouldFocus')">` if it was modified to emit events.

kolibri/plugins/learn/assets/src/views/EmbeddedSidePanel/index.vue
sairina: Yes - this is definitely the reason. I originally put this here because I was trying to get the focus _into_ the search field, but I realize now that I could just find that via the input attribute of the form. I'll fix that. Thank you!

kolibri/core/assets/src/views/FullScreenSidePanel/index.vue
MisRob: Thank you @bjester, I appreciate that with the solution utilizing `events`, it sounds like we wouldn't need `enableFocusTrap` prop on the `EmbeddedSidePanel` and it'd still be `FullScreenSidePanel` that's responsible for controlling most of the focus logic which makes sense to me since we need to handle focus only when we use `FullScreenSidePanel`, and having it all in the `EmbeddedSidePanel` might not be relevant in cases we use it outside of the `FullScreenSidePanel` context.

sairina:
Made some changes (thank you, all for the feedback and guidance). In general:
- Refactored some components to use event listeners instead of passing around the elements to be focused by `FocusTrap`
- Moved the logic for determining first element for `FocusTrap` into `EmbeddedSidePanel` and `BrowseResourceMetadata` instead of sitting in `FullScreenSidePanel`
- Fixed background scrolling after looping through modal that has focus on it

More feedback and comments appreciated.

kolibri/core/assets/src/views/FocusTrap.vue
```diff
@@ -59,15 +59,27 @@
...
       },
       handleLastTrapFocus(e) {
         e.stopPropagation();
-        this.firstEl.focus();
+        this.focusFirstEl();
+      },
+      focusFirstEl() {
+        this.$emit('shouldFocusFirstEl');
+        if (this.firstEl) {
+          this.firstEl.focus();
```
MisRob: Emitting `shouldFocusFirstEl`/`shouldFocusLastEl` events and shortly after focusing on the first/last element here in the `FocusTrap` seems to be confusing, as well as looking at `FocusTrap` API from outside:

```
<FocusTrap
  :firstEl="firstEl"
  :lastEl="lastEl"
  @shouldFocusFirstEl="firstEl.focus()"
  @shouldFocusLastEl="lastEl.focus()"
>
```
where it's not clear whether it's `FocusTrap` or its parent component that is responsible for handling focus.

One way to solve this would be to document in comments that when `firstEl`/`lastEl`  props are provided, `FocusTrap` will handle focus and `shouldFocusFirstEl`/`shouldFocusLastEl` events should be ignored.

However, I think it may be much simpler to remove this logic and `firstEl`/`lastEl` prop from `FocusTrap` API completely now that we can use events. This would also help us to have a consistent way of handling focuses everywhere. I can't imagine a scenario when we couldn't use events since we can pretty easily migrate from

```
<FocusTrap
  :firstEl="firstEl"
  :lastEl="lastEl"
>
```

to

```
<FocusTrap
  @shouldFocusFirstEl="firstEl.focus()"
  @shouldFocusLastEl="lastEl.focus()"
>
```


@bjester Do you have any thoughts?

kolibri/core/assets/src/views/FullScreenSidePanel/index.vue
```diff
@@ -2,53 +2,63 @@
...
+    :tabindex="0"
     :class="{ 'is-rtl': isRtl, 'is-mobile': isMobile }"
     @keyup.esc="closePanel"
   >
     <transition name="side-panel">
-      <div
-        class="side-panel"
-        :style="sidePanelStyles"
+      <FocusTrap
+        :firstEl="firstFocusableEl"
```
MisRob: `firstFocusableEl` and `lastFocusableEl` elements are always `null` so lines 11-12 can be removed

kolibri/core/assets/src/views/FullScreenSidePanel/index.vue
```diff
@@ -97,14 +109,17 @@
       return {
         /* Will be calculated in mounted() as it will get the height of the fixedHeader then */
         fixedHeaderHeight: 0,
+        firstFocusableEl: null,
+        lastFocusableEl: null,
```
MisRob: `firstFocusableEl` and `lastFocusableEl` data can be removed

MisRob:
@sairina This is looking great, thank you. Also, I again appreciate @bjester's idea of using events for handling focuses - now after seeing code for it, it seems to be particularly useful for bringing clarity into this quite complex situation when we have elements that need to get focus scattered among more components.

kolibri/core/assets/src/views/FocusTrap.vue
marcellamaki: I am not @bjester, but that makes sense to me, @MisRob :)

marcellamaki:
This looks good @sairina -- I agree with @MisRob about the small adjustments to just use events. 

The only other thing I'm noticing is that the focus doesn't shift when opening the info panel after these new changes. Is that something you wanted to include in this PR, or will that be follow up? You can see the focus moving behind the backdrop, rather than shifting, here:

resource-panel-toggle

sairina:
> The only other thing I'm noticing is that the focus doesn't shift when opening the info panel after these new changes. Is that something you wanted to include in this PR, or will that be follow up? You can see the focus moving behind the backdrop, rather than shifting, here:
> 
> resource-panel-toggle

I just checked out the `release-v0.15.x` branch, and this seems to be a pre-existing issue for when things are not in mobile view. I can create a new issue for follow-up after we merge this one. -- nevermind. I fixed it.

marcellamaki:
Sounds good to me! 👍

kolibri/core/assets/src/views/FocusTrap.vue
sairina: That works. Does it mean that we should change the `FocusTrap` API in this PR and update the other uses of `FocusTrap`, or create a new PR that does that and updates any additional uses of `FocusTrap`?

MisRob: I assume it'd make sense to have these changes logically together as sometimes happens in PRs when updating components APIs and wouldn't expect complications with updating old places, but maybe you can decide according to how it goes. I feel fine about both ways as long as it's addressed.

bjester: Since this is a bug fix, and because the issue resided outside the `<FocusTrap>` implementation, I think it's @sairina's call as to whether she wants to close this out quickly or continue with additional refactoring. The situation definitely felt like it necessitated a new API for `<FocusTrap>`, so I think as a team, in this type of situation, we should strive for iterative development to avoid getting ourselves into larger tangential refactors. Long term, we should definitely replace the props with events as @MisRob said!

bjester (approved):
Checked it out, works for me!

With locking the background page from scrolling, a future enhancement could be to allow the viewport to scroll the side panel content. I guess this is sort of analogous to trapping focus-- scroll trapping!?

sairina:
Updated the summary, additional manual testing, and the PR title to better reflect what is included. There are still a couple of extant bugs about focus that are not addressed here, but I think this is important to get into `0.15.1` if possible. The additional bugs are recorded in the following issues:
- https://github.com/learningequality/kolibri/issues/9095
- https://github.com/learningequality/kolibri/issues/9096

kolibri/core/assets/src/views/CoreMenu/index.vue
```diff
@@ -98,14 +92,40 @@
...
+      window.setTimeout(() => this.lastFocus.focus());
+    },
+    methods: {
+      /**
+       * @public
+       * Focuses on correct last element for FocusTrap depending on content
+       * rendered in CoreMenu.
+       */
+      focusLastEl() {
+        this.$emit('shouldFocusLastEl');
```
MisRob: Was there any reason for emitting the event here?

kolibri/plugins/learn/assets/src/views/EmbeddedSidePanel/index.vue
```diff
@@ -345,6 +347,18 @@
...
       },
+      /**
+       * @public
+       * Focuses on correct first element for FocusTrap depending on content
+       * rendered in EmbeddedSidePanel.
+       */
+      focusFirstEl() {
+        if (this.$refs.searchBox) {
+          this.$refs.searchBox.focusSearchBox();
+        } else if (this.$refs.folders) {
```
MisRob: Since we access the first item of the `folders` array one line below, it'd be good to also check that there is at least one item in the array `(this.$refs.folders && this.$refs.folders.length > 0)`

kolibri/core/assets/src/views/FocusTrap.vue
```diff
@@ -59,15 +59,27 @@
...
+        }
+      },
+      focusLastEl() {
+        this.$emit('shouldFocusLastEl');
+        if (this.lastEl) {
```
MisRob: `firstEl` and `lastEl` props can now be removed from `FocusTrap` since they're not used from anywhere anymore (lines 36-45), including this code on lines 80-82 and above on 74-76.

MisRob (approved):
Thank you @sairina, great work. I'm leaving some last notes but already approving since overall it's looking very well. I tested most of the places affected locally, if not all, and everything related to this PR works smoothly. Also thanks for paying so much attention to keyboard users by reporting all issues you discovered during this work.

kolibri/core/assets/src/views/FocusTrap.vue
MisRob: Makes sense to me. Initially, I assumed that it wouldn't be a larger diff, but it is as I can see now. In any case, thanks @sairina for iterating on this all.

kolibri/core/assets/src/views/CoreMenu/index.vue
sairina: Originally, I saw that `SideNav` needed to control the close icon from the component, and I think my original thinking was to emit that so that `SideNav` would control it there, but I think I must have missed taking it out since I figured out the right information being passed along. Thanks for the catch!

PR #9073, opened 2022-02-01
Fix morango 144: integrate new version of morango to address single-user syncing issue

### LINKED ISSUE #144: Current FSIC format is not expressive enough to support full needs of single-user syncing in Kolibri
## Background

We discovered an issue in Kolibri with syncing from a single-user device A to a full-facility device B, and then from there to another full-facility device C, in which the data from A does not make it onto C.

The issue turned out to be with how we compute, communicate, and use the "FSIC" (Filter-Specific Instance Counter) data structure in Morango (also referred to as ["Filter Max Counters" in the docs](https://morango.readthedocs.io/en/latest/counters/index.html#counters)).

The current format for the FSIC is a dict of the form `{instance_id: counter}`, for the specific filter (set of partition prefixes) being used in the sync, mapping each `instance_id` to a `counter` level. Each side of the sync generates its own FSIC, which basically says "For data matching this filter, I have everything from these devices (instances) up through these counter levels."

## Problem

Here is a high-level cartoonish summary of how this format prevented the syncing of A's data from B to C. (Note that to simplify I'll ignore counters and just say "has the latest data created on", since it doesn't materially impact the current problem. And when I say "sync data for the full facility" or "sync data for user X", those are referring to the filter being used in that sync.)
- To start with, each of the devices has all the data that it created itself ("A has all the latest data created on A", etc).
- When A syncs data for user X to B, A says to B "I have all the data for user X created on A" and B says "I have all the data for user X created on B, but not that was created on A, so give me all the data for user X created on A". After that sync, B knows that it has all the data for user X created on A, and still has all data for the full facility created on B (itself).
- When B then syncs data for the full facility to C, both the data created on B and the additional data for user X created on A _should_ be included within that (and B knows that it _has_ it), but when B tells C what data it has for the full-facility filter, it's only currently able to say "I have the latest data created on B for the full facility", because it can't say with confidence that it also has ALL the data created on A _for the full facility_, because there might have been additional data created on A (e.g. for a user Y) that B didn't receive during the single-user sync for user X.

## Solution (in-progress)

The new FSIC format basically allows B to tell C "I have the latest data created on B for the full facility, and in addition I also have the latest data for user X created on A", which then allows them to negotiate what pieces need to be sent along, and have the data from A be included as part of that.

At a more technical level, the new FSIC format is of the form `{"super": {partition: {instance_id: counter}}, "sub": {partition: {instance_id: counter}}}`. The `partition` roughly corresponds to the "for user X" or "for the full facility" from the example above -- so, it's now able to say that it has everything under a specific partition from a specific device, without saying that it has _all_ the data from that device matching the current sync filter.

The `"sub"` key contains the partitions that are under the current sync filter, whereas the `"super"` key contains partitions that are prefixes of the filter. This allows us to use space as efficiently as possible, as we mark the counters for an instance at the highest-level partition, removing redundant counters on partitions that are contained within that partition (i.e. for which it is a prefix). The other side is then able to "reinflate" this structure into a format usable for filtering Store records for queuing, by propagating the counters down from superpartitions onto the subpartitions.

Things are looking good with the first-pass implementation of this new format (which will be enabled through a "capabilities" flag, so that new versions can still talk to old versions, though with the single-user syncing constraints noted above).

jamalex:
## Summary

This fixes a syncing issue around a SoUD syncing to a Full Facility device and from there to another Full Facility device. More details of the issue can be found at https://github.com/learningequality/morango/issues/144

## References

Fixes https://github.com/learningequality/morango/issues/144 through updating to Morango v0.6.8a0, which is based on the fixes in https://github.com/learningequality/morango/pull/145 (note: not yet merged/released, so this PR won't yet build).

## Reviewer guidance

We'll want to test a diverse range of syncing scenarios, to ensure everything is solid, with the specific one that was fixed here (and should probably be part of our standard manual testing workflows now too) being:
1. Setup device A with a new facility. Create user X.
2. Single-user sync user X to another device B.
3. Generate data for user X on device B.
4. Single-user sync B with A again, and ensure the data created on B has come across.
5. Full-facility sync A with another device C, and ensure that the data from B has made it onto device C as well.

----

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/core/auth/test/test_morango_integration.py
+++ b/kolibri/core/auth/test/test_morango_integration.py
@@ -1193,3 +1193,52 @@ class SingleUserSyncRegressionsTestCase(TestCase):
             .filter(**base_log_params)
             .exists()
         )
+
+    @multiple_kolibri_servers(3)
+    def test_morango_issue_144(self, servers):
+        """
+        This is to test the syncing issue identified in https://github.com/learningequality/morango/issues/144
+        by ensuring that the data synced from a SoUD to a full-facility device is then synced onwards from there
+        through a full-facility sync to another full-facility device.
+        """
+
+        server1, soud, server2 = servers
+
+        facility, learner, _ = server1.generate_base_data()
+
+        # set up soud as a single-user device
+        soud.sync(server1, facility, user=learner)
+
+        # create a log on soud
+        base_log_params = {
+            "channel_id": "725257a0570044acbd59f8cf6a68b2be",
+            "content_id": "9f9438fe6b0d42dd8e913d7d04cfb277",
+            "user_id": learner.id,
+        }
+
+        soud.create_model(
+            ContentSummaryLog,
+            start_timestamp=timezone.now(),
+            kind="audio",
+            **base_log_params
+        )
+
+        # sync the log from soud back to the first server
+        soud.sync(server1, facility, user=learner)
+
+        # verify that it was synced correctly
+        assert (
+            ContentSummaryLog.objects.using(server1.db_alias)
+            .filter(**base_log_params)
+            .exists()
+        )
+
+        # sync the full facility onto a new server
+        server2.sync(server1, facility)
+
+        # verify that the log has made it over to the new server as well
+        assert (
+            ContentSummaryLog.objects.using(server2.db_alias)
+            .filter(**base_log_params)
+            .exists()
+        )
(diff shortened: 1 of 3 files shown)


jamalex:
(Noting again here that this won't pass/build until https://github.com/learningequality/morango/pull/145 is approved, merged, and released)

bjester:
Looks like some of the test errors are related to my assertion changes in Morango. I will push fixes

bjester (approved):

jamalex:
Oh, sorry, didn't see your message here -- I pushed some updates in parallel, looks to be doubly passing now! 🥳

PR #9086, opened 2022-02-08
Update content import flow for channel upgrades

### LINKED ISSUE #8826: 'import more' flow is interrupted by channel upgrade
## Observed behavior

After clicking 'import more' on a channel, if the channel has updates available, the flow is interrupted:

1. the bottom section showing known resources is completely hidden
2. after upgrading the channel, the user is kicked out of the import flow

## Expected behavior

The user should remain in the 'import more' flow regardless of whether upgrades are available or if they choose to complete an available upgrade

## User-facing consequences

* The user is forced to upgrade even if they don't want to in order to import more
* If the user does want to upgrade, they lose their place in the task and need to start over

## Steps to reproduce

* Import items from a channel
* Upgrade the channel
* Go to 'import more' from the channel

## Context

0.15.0 beta 3

marcellamaki:
## Summary

This is a PR with very few code changes but some significant differences to the UX. It adjusts the channel upgrade flow when importing additional resources. We had tagged this on `0.15.1` as Devon thought it may have been a regression (thus, more important to fix promptly) but after some [investigating](https://learningequality.slack.com/archives/CB37UM23A/p1643988582256619), it seems as though this is just an improvement, rather than a bug fix. So, I am open to suggestions about if this should go into 15 or into develop.

1. The content tree view is now always visible, even if there is an upgrade available
2. After upgrade, the user is returned to the page they were on ("Import More") rather than being returned to the main channel page

## References

Fixes #8826 

| Description | Before  |  After |
|---|---|---|
| Content tree view  | old-version-import (1)  | new-content-tree-view |
| Redirecting after upgrade |  old-redirect-import-flow  | new-upgrade-channel-redirect |

## Reviewer guidance
- Does this sufficiently address the issue? 
- Would adding some kind of "feedback" or loading state between the "upgrade channel" and being returned back to the original page be helpful (cc @jtamiace) 

----

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/plugins/device/assets/src/views/ManageContentPage/NewChannelVersionPage.vue
+++ b/kolibri/plugins/device/assets/src/views/ManageContentPage/NewChannelVersionPage.vue
@@ -2,7 +2,7 @@
 
   <div v-if="!loadingChannel">
 
-    <section v-if="isUpdatingChannel">
+    <section v-if="!isUpdatingChannel">
       <h1>
         {{ $tr('updateChannelAction') }}
       </h1>
@@ -109,7 +109,7 @@
 
     <BottomAppBar>
       <KButton
-        v-if="!channelIsUpdated"
+        v-if="channelIsUpdated"
         :text="$tr('updateChannelAction')"
         appearance="raised-button"
         :primary="true"
@@ -246,7 +246,7 @@
     },
     methods: {
       handleSubmit() {
-        this.disableModal = true;
+        this.showModal = false;
         const updateParams = {
           sourcetype: 'remote',
           channel_id: this.params.channelId,
--- a/kolibri/plugins/device/assets/src/views/ManageContentPage/NewChannelVersionPage.vue
+++ b/kolibri/plugins/device/assets/src/views/ManageContentPage/NewChannelVersionPage.vue
@@ -244,7 +244,9 @@
                 }
               });
             } else {
-              this.$router.push(this.$router.getRoute(PageNames.MANAGE_TASKS));
+              this.$router.push({
+                ...this.$router.getRoute(PageNames.SELECT_CONTENT),
+              });
             }
           })
           .catch(error => {
--- a/kolibri/plugins/device/assets/src/views/ManageContentPage/NewChannelVersionPage.vue
+++ b/kolibri/plugins/device/assets/src/views/ManageContentPage/NewChannelVersionPage.vue
@@ -2,83 +2,99 @@
 
   <div v-if="!loadingChannel">
 
-    <section>
+    <section v-if="isUpdatingChannel">
       <h1>
-        {{ versionAvailableText }}
+        {{ $tr('updateChannelAction') }}
       </h1>
-      <p> {{ $tr('youAreCurrentlyOnVersion', { currentVersion }) }}</p>
-      <p v-if="channelIsIncomplete">
-        {{ $tr('channelIsIncomplete', { available, total }) }}
-      </p>
-    </section>
-
-    <section>
-      <p>
-        <strong>
-          {{ $tr('versionChangesHeader', {
-            oldVersion: currentVersion,
-            newVersion: nextVersion
-          }) }}
-        </strong>
-      </p>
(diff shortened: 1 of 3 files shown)


jtamiace:
Yeah, I think this could do with a loading state. Option 1 here could be a smooth experience since it keeps them on the content selection page. 

Others: option 2 is simpler, used indeterminate state assuming that updates seem to be smaller and quicker to load? Although I'm not sure this would always be true. Option 3 could help to avoid any jarring experience of seeing the channel structure suddenly change, but also creates an extra step for the user. 

update

nucleogenesis (approved):
Code LGTM - I cannot test manually due to issues I'm having with connecting Kolibri to Studio :-/

rtibbles (changes requested):
I think the main issue here is that the redirect to the select content page happens immediately, while the upgrade might still be happening in the background. We need to have some sort of interstitial state that waits for the upgrade task to complete before we do the redirect.

marcellamaki:
I've made the following changed based on Jessica's options above (thank you for the help @jtamiace !!), since it was easily available and doesn't require any new strings. Let me know what you think. @rtibbles, it is not an intermediary state before the redirect -- maybe it would be better to remain on that page (i.e. Option 3) ....

task-status-bar

rtibbles:
Yes, I think best to wait for the task to complete before doing the redirect (the issue is that the loaded channel tree on that page could now be out of date after the upgrade has happened).

radinamatic:
Running this from source on Firefox/Ubuntu, and I see 2 issues:

1. Once the channel is updated, all the recent update tasks on that channel appear listed (instead of just the most recent one), which puts the topic tree completely below the fold (~800px height at 80% zoom), which leaves the user wondering _Wth now...?_ 😕 

   https://user-images.githubusercontent.com/1457929/153808407-8fa83aaf-d733-432e-871e-9ec19c9135f5.mp4

2. On the second test I could still see the _Version 102 of the channel available_ in between all the tasks and the topic tree, which was also confusing (_Didn't I just update the channel...?_)

   https://user-images.githubusercontent.com/1457929/153808404-6eadef46-e3c6-4d7d-8a42-314030fbed49.mp4 

Separate question is that there was nothing actionable in the topic tree after updating the channel... Shouldn't there be an action for the user to confirm that one resource that needed to be deleted? 🤔

rtibbles:
I think this still needs updates, as the current flow will not be loading the upgraded channel data on the topic tree. @marcellamaki and I have discussed this for a next step forward, but it hasn't been implemented yet.

marcellamaki:
@rtibbles @radinamatic this is ready for re-review

pcenov:
@marcellamaki this seemed to work fine for a while but after updating several of my channels it became impossible to update a channel.
Here's a video with the correct flow captured for a channel called MTest:
https://user-images.githubusercontent.com/79847249/160603359-b729dae8-427e-4918-9806-f8f4096cde42.mp4

Here's a video with an incorrect flow where after clicking "View changes" I don't see the changes for the selected MTest channel but instead I see a notification for another of  my channels which I had updated previously. Clicking Continue results in an attempt to list the changes for version 12 of the  MTest channel which never finishes. If I click the Back arrow I can no longer get into the content import flow:
https://user-images.githubusercontent.com/79847249/160604409-3d8e9141-6210-4499-b4e5-f60a59a43093.mp4

Tested with the following build: https://buildkite.com/learningequality/kolibri-debian/builds/3988

marcellamaki:
Thank you for the notes @pcenov - very helpful! I will look into this

marcellamaki:
@pcenov -- I've created a new PR #9252 (using a very different coding approach) for this issue. I hope the changes I have made will address the issues you ran into. If you could take a look at that and let me know your feedback, I would appreciate it! I have added a (not very official 😄 ) gherkin scenario to try to explain the workflow also, which hopefully will be helpful. Thank you! Let me know if you have questions

marcellamaki:
Superseded by #9252

Issue #9100, opened 2022-02-11
Internet Explorer - The info side panel is not displayed


pcenov:
## Observed behavior
The info side panel of a resource is not displayed when I click the info icon of a card at the Library page or at the Bookmarks page while using IE browser.
This issue was initially reported here: https://github.com/learningequality/kolibri/pull/9063

## Expected behavior
Fully functional side panel in IE browser.

## Steps to reproduce the issue
1. Install the following build: https://buildkite.com/learningequality/kolibri-windows/builds/3154
2. Go to the Library/Bookmarks page and click the info icon of a card.

## Screenshots
![2022-02-11_15-25-53](https://user-images.githubusercontent.com/79847249/153612389-bcce6a43-d08c-44f3-8a00-582d0caffb56.png)
![2022-02-11_16-20-04](https://user-images.githubusercontent.com/79847249/153612392-0b6b09ac-4cf8-4db9-99da-43cc4c17e7a7.png)

## Usage Details
Windows 7 and 10, Internet Explorer

pcenov:
@radinamatic

radinamatic:
I'm able to replicate this in IE11, and apparently the same error happens when user clicks on the *View folder resources* or *View information* buttons. 

On top of it, after that error occurs, which has no user-facing notification (just the console), the _Go back_ button (top left arrow) does not work, and the user is effectively stuck there with no way to go anywhere, a part from manually editing the URL (if they know how) 😞 

*View folder resources* | *View information* | _Go back_ (top left arrow)
--- | --- | ---
![Win10 (start)  Running  - Oracle VM VirtualBox_001](https://user-images.githubusercontent.com/1457929/153804625-f029d6b0-1e49-4393-8a29-26fd0c825be1.png) | ![Win10 (start)  Running  - Oracle VM VirtualBox_002](https://user-images.githubusercontent.com/1457929/153804624-f39ae7e9-8dfe-4713-9b40-96363e16f3b2.png) | ![Win10 (start)  Running  - Oracle VM VirtualBox_003](https://user-images.githubusercontent.com/1457929/153804621-fefaf725-d2d0-4d7d-8371-d02ca4d3d644.png)

radinamatic:
@marcellamaki This looks like P0 (blocker) to me.

MisRob:
Seems to be more important than what I've been working on so I'll check if I can find out something before our US friends get up :)

MisRob:
[...]

PR #9080, opened 2022-02-03
Remove kfixedgrid from category search modal

marcellamaki:
## Summary

Fixes #8988 

This PR cleans up the `CategorySearchModal` implementation in several ways. 
1) Rather than using `KFixedGrid` (which does not allow for numCols === 1), it uses `KGrid`
2) This eliminates the need for some props (that I probably never should have added) also
3) Removes a file that wasn't being used at all! 🆘 
4) Changes the display of the `CategorySearchModal` on the Library page to use `FullScreenSidePanel` on medium screens, per [slack discussion](https://learningequality.slack.com/archives/CB37UM23A/p1642790416108000).

## References
Most of this refactor should result in no changes to the UI, with the exception of the "medium screen" change (#4): 

medium-screen-category-search

## Reviewer guidance

- Does the category search work at each screen resolution? 
- Are there any bugs in the layout (overlapping/squished words, etc.)? 

----

## Testing checklist

- [x] Contributor has fully tested the PR manually
- [x] If there are any front-end changes, before/after screenshots are included
- [ ] Critical user journeys are covered by Gherkin stories
- [ ] Critical and brittle code paths are covered by unit tests

## PR process

- [x] PR has the correct target branch and milestone
- [x] PR has 'needs review' or 'work-in-progress' label
- [x] If PR is ready for review, a reviewer has been added. (Don't use 'Assignees')
- [ ] If this is an important user-facing change, PR or related issue has a 'changelog' label
- [ ] If this includes an internal dependency change, a link to the diff is provided

## Reviewer checklist

- Automated test coverage is satisfactory
- PR is fully functional
- PR has been tested for [accessibility regressions](http://kolibri-dev.readthedocs.io/en/develop/manual_testing.html#accessibility-a11y-testing)
- External dependency files were updated if necessary (`yarn` and `pip`)
- Documentation is updated
- Contributor is in AUTHORS.md

PR #9103, opened 2022-02-14
First part of the fix for the info side panel not displayed in Internet Explorer

MisRob:
Fixes the info side panel not displayed in IE after clicking the info button (https://github.com/learningequality/kolibri/issues/9100). See the commit message for an explanation.

However, I tested this fix on the bookmarks page only, and even though the side panel appears now in Internet Explorer as expected for me, there still needs to be some work done. It seems that over time we somehow updated styles related to positioning the close button in the full-screen side panel and after fixing the calculation in this PR, the button seems to be off now in ALL browsers:

**(1) When there are some learning activities**
Screenshot from 2022-02-14 14-00-26

**(2) No learning activities or when the header slot is empty**
Not sure if it happens with production data but in any case `FullScreenSidePanel` should be ready for this situation since the header slot is optional or when learning activities would be missing.

Screenshot from 2022-02-14 14-01-35

I might have also overlooked something and so far was debugging on Browserstack where things were getting quite stuck. Since #9100 seems to be a blocker to the release, I'm opening a draft PR if someone would like to finish this sooner, before I can get to it again.

### DIFF
--- a/kolibri/core/assets/src/views/FullScreenSidePanel/index.vue
+++ b/kolibri/core/assets/src/views/FullScreenSidePanel/index.vue
@@ -171,7 +171,7 @@
       },
       closeButtonStyles() {
         return {
-          top: `calc((${this.fixedHeaderHeight} - 40px) / 2)`,
+          top: `${(this.fixedHeaderHeight - 40) / 2}px`,
         };
       },
       closeButtonFullScreenSidePanelStyles() {


MisRob:
Closing in favor of https://github.com/learningequality/kolibri/pull/9080, where the problematic calculation has been removed completely as part of a larger refactor and where @marcellamaki also added fixes for buttons positions. Thank you, @marcellamaki, @radinamatic, and @pcenov.

PR #9128, opened 2022-02-19
New `NotificationsRoot` component

### LINKED ISSUE #9102: New CoreBase
# Context

EDIT: No more borrowing PageRoot - let's make a new component altogether.

In order to begin updating Kolibri's many plugins and uses of CoreBase, we must establish a new component built to our new ideals that we can start implementing without regressions. 

My suggestion is to take create a new component for this purpose. This will give us a component that accommodates the following suggestions:

- The GlobalSnackbar component should be loaded into every page. This means that the GlobalSnackbar can show itself anytime it sees the correct Vuex state.
- The Core Vuex store's `error` state property is universally how we will decide if the user should see one of either an AuthMessage error message or a general AppError message.
- The UpdateNotification component should be available on every page (and, naturally, rendered under the correct circumstances)
- Renders, otherwise, only a default slot, into which any implementing page in the app can display what they need to.

# Acceptance Criteria

- [x] Create a new root component, include it in the apiSpec
- [x] Move the four aforementioned components to the new component. The error components should follow a similar logical flow as in CoreBase. The other two components need be loaded into the app.
- [x] The component should render the default slot when there are no errors.
- [x] Unit tests covering conditions for showing error messages vs router view

sairina:
## Summary
- Removes 4 components from `CoreBase` into new `NotificationsRoot` component
  - `AuthMessage`
  - `AppError`
  - `GlobalSnackbar`
  - `UpdateNotifications`
- Adds tests
- Registers `NotificationsRoot` in `apiSpec`

## References
Fixes #9102 

## Reviewer guidance
Check to see that the component is doing what is described in the issue (#9102) and check that the tests pass with the `yarn run test-jest` command.

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated


kolibri/core/assets/src/views/NotificationsRoot.vue
```diff
@@ -0,0 +1,254 @@
+<template>
+
+  <div
+    ref="mainWrapper"
```
MisRob: I don't see this reference used from anywhere, so I think it could be removed

MisRob (changes requested):
Thank you, @sairina . The overall direction is looking great. I have two high-level suggestions for now that I think would be good to address.

For later reference, the  `NotificationsRoot` is visually responsible for the following four components which I'll call the messaging components

| `AuthMessage` | `AppError` | `GlobalSnackbar` | `UpdateNotification` |
| -------------------- | -------------- | ----------------------- | --------------------------- |
| AuthMessage | AppError  | GlobalSnackbar | UpdateNotification |

### (1) Decoupling of unnecessary style and layout information (including deletion of related props)

 I’d suggest removing all styles that are not directly related to the messaging components. That’s the majority of computed styles (e.g. `mainWrapperStyles` , `contentStyles`) as well as styles defined in `<style>` block. It’d be fine to keep styles that are unnecessary to position these components on a page if there are some. This also applies to related props: `fullScreen` , `hasSidebar` ,`maxMainWidth` , and `marginBottom`.

**Reason**

`NotificationsRoot` is planned to be used in the following way from plugin index pages:

```javascript
// LearnIndex

<template>
  <NotificationsRoot>
    <router-view />
  </NotificationsRoot>
</template>
```

One of the points of `CoreBase` refactor is to allow each page of the plugin that’s rendered in the `<router-view />` to have full control over its layout, including margins, padding, etc. to gain more flexibility. That wouldn’t be fully possible if we kept unnecessary layout and style dependencies in the `NotificationsRoot`.

### (2) Decoupling loading state from the default slot

I’d suggest removing  `v-if="!loading"` from [the whole wrapper div here](https://github.com/learningequality/kolibri/pull/9128/files#diff-4a205996a86d148f2aeb0e7a84a2f6ea6325fe25f0f548eae3230a4390abdce4R9) and only using it directly on components that need it, that’s `AuthMessage` and `AppError` . For example

```javascript
<KPageContainer v-if="!loading && notAuthorized">
  <AuthMessage
    :authorizedRole="authorizedRole"
    :header="authorizationErrorHeader"
    :details="authorizationErrorDetails"
  />
</KPageContainer>
```

**Reason**

Keeping `v-if="!loading"` on the div that’s also wrapping [the default slot](https://github.com/learningequality/kolibri/pull/9128/files#diff-4a205996a86d148f2aeb0e7a84a2f6ea6325fe25f0f548eae3230a4390abdce4R33) has unhappy consequences. Consider the situation demonstrated in the following example when we’d try to set loading state from one of the pages of the Learn plugin:

```javascript
// LearnIndex

<template>
  <NotificationsRoot>
    <router-view />
  </NotificationsRoot>
</template>
```

```javascript
// Library (rendered within the <router-view /> of LearnIndex)

<script>

export default {
  created() {
    state.loading = true
    fetchData().then(() => state.loading = false)
  }
}

</script>
```

At the moment we set `state.loading = true` in the `created` hook, the Library component would get destroyed (therefore `state.loading = false` would never get called) if we kept `v-if="!loading"` as is in the `NotificationsRoot` . This would make all logic around the loading state rigid and difficult to refactor later (and debug - true story :) - for example, we wouldn't be able to use the global loading state at all in a way shown in the example above (and we actually already need to do this in some cases and use hacks to overcome the current limitation).

---

To sum it up, this means we’d only keep things that are absolutely unnecessary for notifications and messaging components in the `NotificationsRoot`  (and if we find out we need something more from it in the future, we can do that when a need arises while considering architecture and alternative options to solving that particular need).

MisRob:
I realize that there might be a need to re-use some of the logic that I’m suggesting for deletion from this particular component so here are some possible directions (posting here since it's closely related to what I'm suggesting, but final decisions and implementation are out of the scope of this PR):

- If we need to re-use some common layout styles in the future, we can create a default layout component that will encapsulate them and then import this component from each page. Pages will still have freedom to not use a layout and define their own layout (e.g. Library is one example of a page with an atypical layout). We could even have more layout components if we need to re-use different layouts.
- Regarding the default gray background color of a page (currently defined in `mainWrapperStyles`), even though I don’t consider it such a problem, it seems to be rather a relic from the old `CoreBase` that contains much more layout and styles. Keeping it in the `NotificationsRoot` that’s responsible primarily for messaging would feel a bit confusing in my view, so maybe we could rather apply it to `<body>` element somewhere [from the `KolibriApp`](https://github.com/learningequality/kolibri/blob/develop/kolibri/core/assets/src/kolibri_app.js#L16)?

MisRob:
Also, I know it’s quite an abstract task where we’re trying to switch to a different architecture mindset that's quite the opposite of what we’ve relied on so far in the old `Corebase` while not having places to introduce the new component yet, and therefore it might be tricky to determine what should be preserved in the new one. If that’d be helpful, you may create some temporary test pages where this root component would be used similarly to how we plan to use it from plugin index pages. That would hopefully make development simpler since we could test what we need in the `NotificationsRoot` for those four messaging components.

nucleogenesis (commented):
I agree with @MisRob  - I think perhaps that we'll end up seeing how removing the styles plays out in the next sprint as someone takes on #9134 I think that referring to the commit(s) where such styling is removed may serve the person working on that well

sairina:
> If that’d be helpful, you may create some temporary test pages where this root component would be used similarly to how we plan to use it from plugin index pages.

What I did prior to the PR was remove `CoreBase` from a place and use `NotificationsRoot` to see if I could get the error-handling and authorization to work 😄 , so I definitely did a hacky version of that. But, given that there's no styling, I'm wondering if the tests for this component (which were passing and still are all currently passing) are sufficient for this `NotificationsRoot`?

kolibri/core/assets/src/views/NotificationsRoot.vue
```diff
@@ -0,0 +1,140 @@
...
+
+  <div>
+    <KPageContainer v-if="!loading && notAuthorized">
+      <AuthMessage
+        :authorizedRole="authorizedRole"
+        :header="authorizationErrorHeader"
+        :details="authorizationErrorDetails"
+      />
+    </KPageContainer>
+    <KPageContainer v-else-if="error">
```
MisRob: I don't have an example of any specific problematic scenario, but I think we may need the loading condition here too so that we can be sure that the `AppError` doesn't get displayed prematurely? (in the original `CoreBase`, the `AppError` too didn't get displayed until loading finished)

`<KPageContainer v-else-if="!loading && error">`

kolibri/core/assets/test/views/NotificationsRoot.spec.js
```diff
@@ -0,0 +1,87 @@
...
+      store.state.core.loading = false;
+      store.state.core.error = 'some error here';
+      await wrapper.vm.$nextTick();
+
+      expect(wrapper.findComponent({ name: 'AppError' }).exists()).toBeTruthy();
+      expect(wrapper.findComponent({ name: 'AuthMessage' }).exists()).toBeFalsy();
+      expect(wrapper.find('[data-test="main"]').exists()).toBeFalsy();
+    });
+
+    it('UpdateNotification should be rendered ', async () => {
```
MisRob: I'd recommend avoiding variables and component names in tests descriptions, and rather describing things from a user's perspective, at least whenever possible. One of the main reasons is the readability of the test suite (e.g. when test descriptions are displayed in a console) for people who are not familiar with internal implementation. This applies to some other test descriptions in this suite too.

kolibri/core/assets/test/views/NotificationsRoot.spec.js
```diff
@@ -0,0 +1,87 @@
...
+      const { wrapper, store } = makeWrapper({ computed: { showNotification: () => true } });
+      store.state.core.loading = false;
+      await wrapper.vm.$nextTick();
+
+      expect(wrapper.findComponent({ name: 'UpdateNotification' }).exists()).toBeTruthy();
+    });
+
+    it('UpdateNotification should not be rendered ', async () => {
```
MisRob: Could you please add information about when update notification should/should not be rendered to this test and the previous test descriptions too? Similarly to how you did it for the authentication message test _"if user is not authorized, AuthMessage component should be rendered"_

kolibri/core/assets/test/views/NotificationsRoot.spec.js
```diff
@@ -0,0 +1,87 @@
...
+      store.state.core.loading = false;
+      await wrapper.vm.$nextTick();
+
+      expect(wrapper.find('[data-test="main"]').exists()).toBeTruthy();
+      expect(wrapper.findComponent({ name: 'AuthMessage' }).exists()).toBeFalsy();
+      expect(wrapper.findComponent({ name: 'AppError' }).exists()).toBeFalsy();
+    });
+
+    it('if user is not authorized, AuthMessage component should be rendered', async () => {
+      const { wrapper, store } = makeWrapper({ computed: { notAuthorized: () => true } });
```
MisRob: This test, as implemented right now, doesn't really test this scenario. Assume that over time we'd make a mistake in code of [the `notAuthorized` computed property ](https://github.com/learningequality/kolibri/pull/9128/files#diff-4a205996a86d148f2aeb0e7a84a2f6ea6325fe25f0f548eae3230a4390abdce4R83), for example:

```javascript
notAuthorized() {
  if (
    this.error &&
    this.error.response &&
    this.error.response.status &&
    this.error.response.status == 403
  ) {
    return false; // should be `return true`
  }
  return !this.authorized;
},
```

Even though `AuthMessage` wouldn't get displayed after we got 403 response, this test would still pass because `notAuthorized` is set to `true` no matter of what due to how the test wrapper is created `makeWrapper({ computed: { notAuthorized: () => true } })`, giving us the false impression that things work as expected.

I see this pattern in couple of other tests of this suite too. I'd suggest removing anything related to `computed` from the whole test suite and mocking rather things that are coming from outside of the `NotificationsRoot` (store state etc.) instead of mocking computed properties, similarly to how you did it [on lines 57-58](https://github.com/learningequality/kolibri/pull/9128/files#diff-0ad00cb2ee84c775bb92409d2ea55408ffa548d7bb3546116225a01dce333913R58).

I wrote more generally about problems of reaching into internal implementation in tests in https://github.com/learningequality/kolibri/pull/9065#pullrequestreview-870732578 if some more information would help.

MisRob (approved):
Thank you for the updates and your work, @sairina, the component is looking good to me.

> But, given that there's no styling, I'm wondering if the tests for this component (which were passing and still are all currently passing) are sufficient for this NotificationsRoot

I think that these main scenarios are sufficient. In my opinion, it is better to focus on fewer reliable tests that can be easily maintained rather than plenty of tests just to have everything covered.

I left some comments regarding the implementation of some tests that I think would be good to address but don't see that as blocking.

nucleogenesis (approved):
Everything looks great to me! If there are any issues they'll come out in the wash as we implement it elsewhere. Great work!

nucleogenesis:
@sairina I think it's good to merge this in spite of the DMG build failing

rtibbles:
Yes, the DMG build failure was definitely not caused by any code in this PR!

PR #9141, opened 2022-02-25
Delete any annotated channelmetadata many to many fields to avoid integrity errors

rtibbles:
## Summary
* Currently `included_languages` is a many to many field from ChannelMetadata to Languages
* Django implements this via a through table
* When we delete the ChannelMetadata object via SQLAlchemy, this does not get properly deleted, because the schema we are using does not know about the included languages table
* This is because using a schema that did know about it would cause other errors, so we let sleeping dogs lie
* To workaround this, we explicitly delete any many to many through tables that are pointing at ChannelMetadata in order to prevent Integrity errors caused by trying to delete a channel metadata row when a many to many through row still points at it
* Adds a regression test that does not replicate the integrity errors, but does at least assert that the deletion of the included languages has occurred.

## Reviewer guidance
Does the test make sense?
Can you think of another way to try to actually get the reported integrity error to trigger in the test?

----

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/core/content/test/test_channel_import.py
+++ b/kolibri/core/content/test/test_channel_import.py
@@ -29,6 +29,7 @@ from kolibri.core.content.models import AssessmentMetaData
 from kolibri.core.content.models import ChannelMetadata
 from kolibri.core.content.models import ContentNode
 from kolibri.core.content.models import File
+from kolibri.core.content.models import Language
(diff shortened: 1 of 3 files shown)


bjester (approved):
Code changes make sense, but I haven't tested

PR #9142, opened 2022-02-26
Redirect for Bookmarks when user is not logged in

### LINKED ISSUE #9057: Signed-out user can access the bookmarks page when users are allowed to explore resources without signing in
## Observed behavior

With this device settings:

device

I can access the bookmarks page as a signed-out user via URL `/en/learn/#/bookmarks`

bookmarks-page

## Expected behavior

Even though users are allowed to explore resources without signing in, giving access to the bookmarks page still doesn't make sense for anonymous users. For signed-out users, it should rather show "You must be signed in" message:

message

## User-facing consequences

Anonymous users can access pages that shouldn't be visible to them

## Steps to reproduce

1. Select "Allow users to explore resources without signing in" in device settings as an admin
2. Sign out
3. Type `/en/learn/#/bookmarks` URL to the browser address bar

## Context

- Kolibri version: Kolibri 0.15.1.dev0+git.20220127215827
- Operating system: Ubuntu 20.04.3 LTS
- Browser: Chrome Version 97.0.4692.99 (Official Build) (64-bit)

sairina:
## Summary
Added a check to see whether or not a user is logged in. If a user is not logged in, they are redirected to the "content unavailable" view.

## References
Fixes #9057

## Reviewer guidance
Follow the guidance from the original issue:
1. Select "Allow users to explore resources without signing in" in device settings as an admin
2. Sign out
3. Type /en/learn/#/bookmarks URL to the browser address bar
4. Make sure that you see the image below

message

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/plugins/learn/assets/src/views/LearnIndex.vue
+++ b/kolibri/plugins/learn/assets/src/views/LearnIndex.vue
@@ -95,6 +95,9 @@
       ...mapState('examReportViewer', ['exam']),
       ...mapState(['pageName']),
       userIsAuthorized() {
+        if (this.pageName === PageNames.BOOKMARKS) {
+          return this.isUserLoggedIn;
+        }
         return (
           (plugin_data.allowGuestAccess && this.$store.getters.allowAccess) || this.isUserLoggedIn
         );


pcenov:
Seems that this is partially fixed because navigating to /en/learn/#/bookmarks as a signed out user results in the user seeing the /en/learn/#/resources-unavailable page while as pointed out above the expected result is for the user to see /en/device/#/permissions.

Observed:


Expected:

MisRob:
I can confirm, I was just in the middle of the review when I saw this too.

Also, after this is resolved, we also need to make sure that for "Allow users to explore resource without signing in" device settings, users will get redirected to the bookmarks page after they sign in. This can be achieved by using the `next` query parameter similarly it's currently done for "Learners must sign in to explore resources" device settings (see `next=http%253A%252F%252F127.0.0.1%253A8000%252Fen%252Flearn%252F%2523%252Fbookmarks` in the URL)

MisRob (dismissed):
Thank you, @sairina, it's definitely better than it was. It would be good to resolve the issues mentioned above before merging.

kolibri/plugins/learn/assets/src/routes/index.js
```diff
@@ -179,6 +179,16 @@ export default [
   {
     name: PageNames.BOOKMARKS,
     path: '/bookmarks',
+    beforeEnter: (to, from, next) => {
```
rtibbles: I think we can leverage the authorized user messaging in a different way:

I know this should and will later be refactored, but while we have it, just updating the `userIsAuthorized` computed prop in `LearnIndex.vue` so that if the page is `PageName.BOOKMARKS`, then the user gets flagged as not authorized.

https://github.com/learningequality/kolibri/blob/develop/kolibri/plugins/learn/assets/src/views/LearnIndex.vue#L97

This feels a little bit more explicit than setting an error to 403 in the core errors, which might be interpreted by other parts of the app in another way.

kolibri/plugins/learn/assets/src/app.js
```diff
@@ -25,6 +25,7 @@ class LearnModule extends KolibriApp {
     router.beforeEach((to, from, next) => {
       if (
         to.name !== PageNames.CONTENT_UNAVAILABLE &&
+        to.name !== PageNames.BOOKMARKS && // We want to handle bookmarking behavior separately
```
rtibbles: I think this would be unneeded with the change I suggest below.

rtibbles (changes requested):
I think this is best targeted in LearnIndex at the moment, not in the router - router guards are best used to redirect to other pages, not conditionalize behaviour within a page.

kolibri/plugins/learn/assets/src/routes/index.js
sairina: That is infinitely easier and clearer. Thanks!

sairina:
With the changes, this is the behavior:
|User (not logged in) goes to `/learn/#/bookmarks` and clicks on "sign in" link|User is able to sign in (note the update in the URL)|
|--|--|
Screen Shot 2022-03-04 at 9 58 26 AM|Screen Shot 2022-03-04 at 9 58 39 AM|

kolibri/plugins/learn/assets/src/views/LearnIndex.vue
```diff
@@ -95,6 +95,9 @@
       ...mapState('examReportViewer', ['exam']),
       ...mapState(['pageName']),
       userIsAuthorized() {
+        if (this.pageName === PageNames.BOOKMARKS) {
```
rtibbles, line 98: Delightfully simple!

rtibbles (approved):

PR #9171, opened 2022-03-09
Use scroll position

### LINKED ISSUE #9099: Create useScrollPosition module
# Context

Currently, there is some code in CoreBase that handles mapping the user's scrollPosition to unique router history state hashes. This provides us with everything we need to ensure the user's scrollPosition remains consistent while navigating between pages and views within the app. 

For example, a Learner may scroll down a dozen or more resources, select and view a resource, then hit the back button on their browser _or_ the back arrow within the heading app bar. That Learner would not want to start the the top of the previous page, they'd expect to be where they left off from. (If this doesn't happen right now, making it happen isn't in the scope of this PR).

# Acceptance Criteria

- [ ] A new module is created using the `kolibri.lib.vueCompositionApi` to encapsulate the logic previously [defined here](https://github.com/learningequality/kolibri/blob/e0d85dac127107b9522467ada0eec79b97e05c68/kolibri/core/assets/src/views/CoreBase/index.vue#L148-L167).

- [ ] The module exposes a public API accommodating the following use cases

- Get the scrollPosition (`{x: Number, y: Number}`) for the current history state without having to pass the history state.
- Set the scrollPosition for the current history state after passing in a set of (x,y) coordinates OR a string (which could be a CSS selector pointing to an element).
- scrollToSavedPosition() - a function that will set the scrollPosition of the window to the saved position state for the current history state OR to the default scrollPosition `{ x: 0, y: 0}`
- Clear the saved scrollPosition for the current history state (ie, `doNotSaveScrollPosition()` or something similar might convey the purpose better - which is to give any page/route the ability to say "don't save the position here")

- [ ] The state is stored in the module scope - making it a global store that any component can interface with.

- [ ] Unit tests covering most or all functions, especially those with side effects (those that touch the state itself or call window.scrollTo)

# Not to do 

- Don't remove this from CoreBase, but could put a comment noting that it will be deprecated perhaps next to it there?
- This module will provide functionality that is not coupled to Vue Router's [ScrollBehavior](https://v3.router.vuejs.org/guide/advanced/scroll-behavior.html#scroll-behavior) - however it will work closely with it.

marcellamaki:
## Summary

Moves the existing features of scrollPosition in CoreBase into a composable in `kolibri.core`

Does _not_ remove the function from CoreBase, but adds a note to remove it when the file is refactored, and use the composable instead. 

Adds basic tests.

## References

Fixes #9099

## Reviewer guidance

I have not added a test for the 'css selector' condition, because I couldn't figure out how to test it, but if anyone has a suggestion, I would be happy to add it.

----

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/core/assets/src/composables/useScrollPosition.js
+++ b/kolibri/core/assets/src/composables/useScrollPosition.js
@@ -1,0 +1,72 @@
+/* eslint-disable */
+
+export default function useScrollPosition() {
+  let _scrollPositions = {};
+
+  const scrollPosition = computed({
+    get() {
+      // Use key set by Vue Router on the history state.
+      const key = (window.history.state || {}).key;
+      const defaultPos = { x: 0, y: 0 };
+      if (key && this._scrollPositions[key]) {
+        return this._scrollPositions[key];
+      }
+      return defaultPos;
+    },
+    set(oldLocation, newLocation) {
+      if (newLocation !== oldLocation) {
+        // reference location can be { x: 0, y: 0 } coordinates using a Object,
+        // or a string i.e. a css selector
+        // for any instance
+        let x, y;
+        if (newLocation instanceof String) {
+          if (document.querySelector(newLocation)) {
+            const position = document.querySelector(newLocation).getBoundingClientRect();
+            x = position.left;
+            y = position.top;
+          } else {
+            x = 0;
+            y = 0;
+          }
+        } else if (newLocation instanceof Object) {
+          if (newLocation.x instanceof Number && newLocation.x >= 0) {
+            x = newLocation.x;
+          } else {
+            x = 0;
+          }
+          if (newLocation.y instanceof Number && newLocation.y >= 0) {
+            y = newLocation.y;
+          } else {
+            y = 0;
+          }
+        } else {
+          x = 0;
+          y = 0;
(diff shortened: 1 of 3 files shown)


kolibri/core/assets/src/composables/useScrollPosition.js
```diff
@@ -0,0 +1,77 @@
...
+        }
+      } else {
+        x = 0;
+        y = 0;
+      }
+      const key = (window.history.state || {}).key;
+      // Only set if we have a vue router key on the state,
+      // otherwise we don't do anything.
+      if (key) {
+        _scrollPositions[window.history.state.key] = { x: x, y: y };
```
nucleogenesis, line 50: Looks like `window.history.state.key` could just be `key` in this line (not a blocker)

nucleogenesis (approved):
This all looks good to me! I'm not sure how we would test the CSS. Maybe could render a component, then dig into the wrapper and get one of the nodes and get it's bounding client rect - then should be able to expect it to be the value saved for that history state I think?

I wouldn't block this on that though. I'm not entirely sure it ever comes up - but we'll see as we continue to remove CoreBase.

PR #9176, opened 2022-03-14
Update the kolibri-tools node version to 16.x

danigm:
Update the node-version in kolibri-tools to match the version in kolibri.


rtibbles:
This update has been done specifically in the develop branch - this will break future builds of release-v0.15.x.

jredrejo:
so I must assume it would also break 0.14.x builds, right? I was thinking of moving KDP to node 16.x too to make my dev life easier.

rtibbles:
Yes.

PR #9196, opened 2022-03-15
Remove corebase from facility index

### LINKED ISSUE #9134: Remove CoreBase from FacilityIndex
# Context

We can begin removing CoreBase from what appears to be a _relatively_ simple component before diving into the more complex and OKR-specific Learn plugin.

_Relatively_ is a keyword there, however. This process will require sweeping changes across the Facility plugin and many of its components.

I will outline a plan of action and some expectations, but there may be some complications that are unforeseeable before getting into the work itself.

CoreBase also handled some styling which will likely need to be applied moving forward. Ultimately, this may result in our building new base styling components, but for now we can apply these styles directly.

# A new architecture

In #9128 a new component is being introduced to handle much of CoreBase's responsibility we've agreed should remain housed in a single component. This component is designed to bootstrap once into each plugin the functionality around:

- Authentication error messaging
- General application error messaging
- Snackbar messaging
- Update notification messaging

There are a few things **no longer included in that new #9128 component** but **that need to be maintained through the refactor**:

- Scroll position mapping and restoration between route navigations. #9099 and #9101 are issues dedicated to encapsulating that logic. _Note that completion of this issue may result in Facility not having this feature until #9101 is resolved and that is okay_
- Implementing the correct AppBar / ImmersiveToolbar for the page (more on this in the following section) along with the wrapping ScrollingHeader component in addition to subnavs being passed to the proper AppBar slot when necessary.
- Implement the SideNav whenever it is needed
- Styling
- Managing the loading state hiding content when loading

# Acceptance Criteria

- Replace CoreBase in FacilityIndex with the new #9128 component
- In each Page that is addressed within FacilityIndex's immersivePageProps, implement directly the AppBar/ImmersiveToolbar with the proper configuration. Rather than having FacilityIndex owning this, each *Page component should own this itself.

nucleogenesis:
## Summary

Removes CoreBase from Facility plugin. The first and last commits' messages explain roughly the approach that has come out of this, but I will summarize here.

### Introducing `AppBarCorePage`

This component bootstraps the `ScrollingHeader, SideNav, and LanguageSwitcherModal` components into a simple component that exposes this API

- default slot - where the body of the page content will go
- subNav slot - where an optional subNav (the bar under the main AppBar) can go
- title prop - where the page's title will go (passed to AppBar)

I initially did not make this bit a core component but instead did something like this specific to the Facility plugin - but that did not cover the case of the error pages bootstrapped in NotificationsRoot - so you'll see that those parts of NotificationsRoot are now wrapped with `<AppBarCorePage>` (with no title) - so auth errors or other errors will still give people the nav items to go to Learn, change languages, or use the user dropdown to get to the sign in (which is redundant with the link on the auth error page but that's neither here nor there).

### The `FacilityAppBarPage`

The `FacilityAppBarPage` simply implements the `FacilityTopNav` that exists on every AppBar-including page (except one, which is handled using `AppBarCorePage` directly itself).

It also calculates the proper title string for the `AppBar` depending on whether the user is an admin of a multi-facility device or if the `appBarTitle` prop has a value.

#### Default titles

Facility has a couple of default titles used throughout - either it just says "Facility" or it says "Facility - Name of the facility". `FacilityAppBar` exposes the `title` prop in case it needs to be overridden, but the aforementioned logic is used I think in every case.

### ImmersivePage (this is only in Facility right now, but should it be moved to a Core page?)

This is simlar to the `AppBarCorePage` except I wasn't compelled to move it to the Core plugin (perhaps this can be a task for the Learn plugin if it is ever necessary there).

In any case, @jtamiace this component defaults to having a black immersive toolbar, but when I look at Kolibri Beta, it is primary purple - is this unique to Facility or the expectation of the ImmersiveToolbar altogether?

### Then...

All of the various *Page components were wrapped in these new wrapper components.

**Note:** I'll add tests later today for these new components

## References

Fixes #9134 

## Reviewer guidance

Compare all of Facility to Kolibri beta.

When on a single-facility device, the title should be "Facility" and not "Facility - Name of selected facility".

----

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated


kolibri/core/assets/src/mixins/nav-components.js
```diff
@@ -1,7 +1,9 @@
+import Vue from 'vue';
 import { mapGetters } from 'vuex';
 import { UserKinds } from 'kolibri.coreVue.vuex.constants';
+import responsiveWindowMixin from 'kolibri.coreVue.mixins.responsiveWindowMixin';
 
-export default {
+const mixin = Vue.util.mergeOptions(responsiveWindowMixin, {
```
marcellamaki, line 6: 🤯

kolibri/core/assets/src/views/CoreBase/index.vue
```diff
@@ -313,18 +312,12 @@
...
         if (this.showSubNav) {
           // Adds the height of KNavBar
-          return this.headerHeight + 48;
+          return this.topBarHeight + 48;
         }
-        return this.headerHeight;
-      },
-      navWidth() {
-        return this.headerHeight * 4;
+        return this.topBarHeight;
```
marcellamaki, line 320: Seems like this is missing the factor of 4 (or different/updated factor)?

pcenov:
@nucleogenesis I attempted to test this by using the following build: https://buildkite.com/learningequality/kolibri-debian/builds/3923
Unfortunately none of the Facility pages can be accessed and I can see the following error in the browser console:
```
vue.runtime.esm.js:1897 TypeError: this.coreString is not a function
    at Sub.title (FacilityAppBarPage.vue:41:1)
    at Watcher.get (vue.runtime.esm.js:4495:25)
    at Watcher.evaluate (vue.runtime.esm.js:4597:21)
    at Sub.title (vue.runtime.esm.js:4851:17)
    at Sub.FacilityAppBarPagevue_type_template_id_2b498338_render (FacilityAppBarPage.vue?6cd9:8:27)
    at vue-composition-api.esm.js:1585:80
    at activateCurrentInstance (vue-composition-api.esm.js:1545:16)
    at Sub.$options.render (vue-composition-api.esm.js:1585:24)
    at Sub.Vue._render (vue.runtime.esm.js:3569:22)
    at Sub.updateComponent (vue.runtime.esm.js:4081:21)
```
Here's a screenshot:


Zipped .kolibri folder for additional logs and DB:
[kolibri.zip](https://github.com/learningequality/kolibri/files/8262387/kolibri.zip)

@radinamatic

nucleogenesis:
@pcenov thanks for the details. What is strange is that I tested with the PEX from that same build and did not see the issue, so I'm wondering if maybe there is an issue with the Debian builder specifically in this case? ( cc @DXCanas )

DXCanas:
The Debian build essentially just wraps the WHL as its source iirc - can’t think of any way it’d made this sort of change. @jredrejo can confirm there. 

I think @rtibbles might have a bit more insight — he’s the most up to date and has made some changes on the GHActions side.

jredrejo:
> The Debian build essentially just wraps the WHL as its source iirc - can’t think of any way it’d made this sort of change. 

Yep, it takes it from the `tar.gz`, adds the `debian` directory and makes the packaging. The only place where the Debian package removes files is in the `rules` Makefile, https://github.com/learningequality/kolibri-installer-debian/blob/master/debian/rules and I don't see there anything related to this problem.

rtibbles:
Given the error, it seems much more likely that this reference to `this.coreString` is to blame: https://github.com/learningequality/kolibri/pull/9196/files#diff-fb959b4b27924b9c405741ff3b9c15260ff7a5c4cff3728e82064bd64f01fb82R41

This component does not import the core string mixin anywhere, so it seems likely that this is undefined.

nucleogenesis:
@pcenov - thanks to @rtibbles note I got it worked out and the tests all fixed - I had expected incorrectly that the linter would have caught that particular issue. 

In any case, should be ready to review now.

pcenov:
@nucleogenesis it's all looking good now with the exception of enrolling a Learner or assigning a Coach to a class. In that case I'm seeing a blank page instead of the class page with the Coaches and Learners tables.
Video:
https://user-images.githubusercontent.com/79847249/159007807-81f9625d-1cc0-4ce2-a28c-49379e321b7e.mp4

Logs:
[logs.zip](https://github.com/learningequality/kolibri/files/8304639/logs.zip)

kolibri/core/assets/src/views/AppBarCorePage.vue
```diff
@@ -0,0 +1,78 @@
+<template>
+
+  <!-- TODO useScrollPosition to set scrollPosition...
+    here or in router, but somewhere -->
+  <div>
+    <ScrollingHeader :scrollPosition="0">
```
marcellamaki, line 6: This isn't imported -- not sure if it's this that should be removed, or if the import is missing

kolibri/core/assets/src/views/AppBarCorePage.vue
```diff
@@ -0,0 +1,78 @@
...
+      <AppBar
+        ref="appBar"
+        :title="title"
+        :hideNavBar="hideNavBar"
```
marcellamaki: getting a console error undefined for this

marcellamaki (changes requested):
Looking good, @nucleogenesis! I think updating two routes with the `classId` params will help solve the problem that @pcenov mentioned -- one in CoachClassAssignmentPage > assignCoaches() > .push(this.$store.getters.facilityPageLinks.ClassEditPage(**this.class.id**)

and the same in 

LearnerClassEnrollementPage > enrollLearners() > this.$store.getters.facilityPageLinks.ClassEditPage(**this.class.id**)

There are a couple of console errors that I note also -- they don't seem to be causing anything to _break_, but I couldn't quite sort out if it was old code that needed to be removed, or if there were a few imports missing. 

Other than that, looks great! Excited to see this moving forward! 🎉

kolibri/core/assets/src/views/CoreBase/index.vue
nucleogenesis: That factor of 4 was used in the `navWidth` computed function which is removed and no longer used here.

kolibri/core/assets/src/views/AppBarCorePage.vue
nucleogenesis: Fix pushed for this

kolibri/core/assets/src/views/AppBarCorePage.vue
nucleogenesis: Fix pushed for this

nucleogenesis:
Thanks @pcenov - I've pushed a fix for the redirect issue and some other visual fixes and cleaned up some console errors that I was seeing.

marcellamaki (approved):

marcellamaki:
Changes look good to me @nucleogenesis!

pcenov:
@nucleogenesis I can also confirm that the redirect issue is fixed now and there are no other issues spotted by manual QA.

PR #9222, opened 2022-03-25
Remove corebase from learn index

### LINKED ISSUE #9137: Remove CoreBase from Learn
# Context

I think that we will look to our experience with #9134 before I flesh this out, however, the context there is largely the same. We will remove immersivePageProps and apply the new architecture described there in Learn.

This issue will be blocked by #9134 until we've had more experience in the process so this issue can be as insightful as possible to whoever takes it on.

marcellamaki:
### Overview 

Fixes #9137

This PR removes CoreBase from `LearnIndex.vue` and adopts the `<router-view/>` pattern created by @nucleogenesis in his `FacilityIndex` refactor. (For more background on the logic of the architecture, see #9196)

There are two wrapper pages, `LearnAppBarPage` and `ImmersivePageRoot` that manage the style of the page. 

In the Learn plugin versions, I've added a new prop `applyStandardLayout` which defaults to true and applies the default  styles of each type of page, but can also be overridden, such as in the Library and Topics pages, where the use the full screen, rather than a centered block of main content.

This PR also removes `LearnImmersiveLayout` from `LearnIndex` and puts it in a router-view (although this is probably out of scope) for the sake of consistency. I have not addressed the rather chaotic `backRoute` situation in `LearnImmersiveLayout`, but the existing logic has been moved out of `LearnIndex` and is working.

#### Places of consideration 

Basically all of Learn has been refactored, so this will require a thorough going-over with manual QA. I have tried to be very thorough with my testing on multiple browsers and screen sizes in advance of this PR, but... definitely probably that I have missed something. On the flip side, very few of the code changes are different from the pattern @nucleogenesis set up, and they actually are mostly straightforward (I hope). 

Tricky spots were mostly around `content` since it is no longer passed from LearnIndex. So, things to think about here: 
1. Is actually working everywhere? 
2. Are there any places where I dropped the loading state handling and things look janky while content is being fetched? 

CSS changes to note: 
1. (especially @MisRob) I want to be sure the improvements made for 0.15.1 fixes, especially around the scrolling header jumpiness and the height of various elements of the page layout has not been lost. It looks okay to me, and I have done QA on multiple browsers, but your attention and review there would be helpful in case I accidentally removed something that was needed. (cc @pcenov and @radinamatic for testing on IE)
2. I have had to add a z-index to the `ImmersiveToolbar` that is in `LearnIndex` due to the cards (without it, they scroll over it). Is this okay? Is there something I should do instead that would make adding an explicit z-index unnecessary?
3. I updated the `card` width property which resolved the problems with the grid. I do not know why this is necessary in this version 🙃 but it seems to be working now. Are there any places on Library or Topics page where the grid layout is wrong?

### DIFF
--- a/kolibri/core/assets/src/views/AppBarCorePage.vue
+++ b/kolibri/core/assets/src/views/AppBarCorePage.vue
@@ -59,9 +59,10 @@
         type: String,
         default: null,
       },
-      applyStandardLayout: {
-        type: Boolean,
-        default: true,
+      appearanceOverrides: {
+        type: Object,
+        required: false,
+        default: null,
       },
     },
     data() {
@@ -76,8 +77,9 @@
         loading: state => state.core.loading,
       }),
       wrapperStyles() {
-        return this.applyStandardLayout
-          ? {
+        return this.appearanceOverrides
+          ? this.appearanceOverrides
+          : {
               width: '100%',
               display: 'inline-block',
               backgroundColor: this.$themePalette.grey.v_100,
@@ -86,8 +88,7 @@
               paddingTop: this.appBarHeight + 32 + 'px',
               paddingBottom: '72px',
               marginTop: 0,
-            }
-          : '';
+            };
       },
     },
     mounted() {
--- a/kolibri/core/assets/src/views/AppBarCorePage.vue
+++ b/kolibri/core/assets/src/views/AppBarCorePage.vue
@@ -49,7 +49,6 @@
   import AppBar from 'kolibri.coreVue.components.AppBar';
   import ScrollingHeader from 'kolibri.coreVue.components.ScrollingHeader';
   import SideNav from 'kolibri.coreVue.components.SideNav';
-  import { mapState } from 'vuex';
 
   export default {
     name: 'AppBarCorePage',
@@ -64,6 +63,10 @@
         required: false,
         default: null,
       },
+      loading: {
+        type: Boolean,
+        default: null,
+      },
     },
     data() {
       return {
@@ -73,9 +76,6 @@
       };
     },
     computed: {
-      ...mapState({
-        loading: state => state.core.loading,
-      }),
       wrapperStyles() {
         return this.appearanceOverrides
           ? this.appearanceOverrides
--- a/kolibri/core/assets/src/views/AppBarCorePage.vue
+++ b/kolibri/core/assets/src/views/AppBarCorePage.vue
@@ -6,6 +6,7 @@
     <ScrollingHeader :scrollPosition="0">
       <AppBar
         ref="appBar"
+        class="app-bar"
         :title="title"
         @toggleSideNav="navShown = !navShown"
         @showLanguageModal="languageModalShown = true"
@@ -14,6 +15,11 @@
           <slot name="subNav"></slot>
         </template>
       </AppBar>
+      <KLinearLoader
+        v-if="loading"
+        type="indeterminate"
+        :delay="false"
+      />
     </ScrollingHeader>
 
     <div class="main-wrapper" :style="wrapperStyles">
@@ -43,6 +49,7 @@
   import AppBar from 'kolibri.coreVue.components.AppBar';
   import ScrollingHeader from 'kolibri.coreVue.components.ScrollingHeader';
   import SideNav from 'kolibri.coreVue.components.SideNav';
+  import { mapState } from 'vuex';
 
   export default {
     name: 'AppBarCorePage',
@@ -51,6 +58,10 @@
       title: {
         type: String,
         default: null,
+      },
+      applyStandardLayout: {
+        type: Boolean,
+        default: true,
       },
     },
     data() {
@@ -61,17 +72,22 @@
       };
     },
     computed: {
+      ...mapState({
+        loading: state => state.core.loading,
+      }),
       wrapperStyles() {
-        return {
-          width: '100%',
-          display: 'inline-block',
-          backgroundColor: this.$themePalette.grey.v_100,
-          paddingLeft: '32px',
-          paddingRight: '32px',
-          paddingTop: this.appBarHeight + 32 + 'px',
-          paddingBottom: '72px',
-          marginTop: 0,
-        };
+        return this.applyStandardLayout
+          ? {
+              width: '100%',
+              display: 'inline-block',
+              backgroundColor: this.$themePalette.grey.v_100,
+              paddingLeft: '32px',
+              paddingRight: '32px',
+              paddingTop: this.appBarHeight + 32 + 'px',
+              paddingBottom: '72px',
+              marginTop: 0,
+            }
+          : '';
       },
     },
     mounted() {
@@ -82,3 +98,16 @@
   };
 
 </script>
+
+
+<style lang="scss" scoped>
+
+  @import '~kolibri-design-system/lib/styles/definitions';
+
+  .app-bar {
+    @extend %dropshadow-4dp;
+
+    width: 100%;
+  }
+
+</style>
--- a/kolibri/core/assets/src/views/AppBarCorePage.vue
+++ b/kolibri/core/assets/src/views/AppBarCorePage.vue
@@ -2,7 +2,7 @@
 
   <!-- TODO useScrollPosition to set scrollPosition...
     here or in router, but somewhere -->
-  <div>
+  <div class="main">
     <ScrollingHeader :scrollPosition="0">
       <AppBar
         ref="appBar"
@@ -81,7 +81,8 @@
           ? this.appearanceOverrides
           : {
               width: '100%',
-              display: 'inline-block',
+              maxWidth: '1064px',
+              margin: 'auto',
               backgroundColor: this.$themePalette.grey.v_100,
               paddingLeft: '32px',
               paddingRight: '32px',
--- a/kolibri/plugins/learn/assets/src/modules/pluginModule.js
+++ b/kolibri/plugins/learn/assets/src/modules/pluginModule.js
@@ -1,3 +1,4 @@
+// import { LastPages } from './../constants/lastPagesConstants';
 import { PageNames, ClassesPageNames } from './../constants';
 import mutations from './coreLearn/mutations';
 import * as getters from './coreLearn/getters';
@@ -24,53 +25,42 @@ export default {
   getters: {
     ...getters,
     learnPageLinks() {
-      const params = {};
       return {
         HomePage: {
           name: PageNames.HOME,
-          params,
         },
         LibraryPage: {
           name: PageNames.LIBRARY,
-          params,
         },
         TopicsPage: {
           name: PageNames.TOPICS_TOPIC,
-          params,
         },
         TopicsSearchPage: {
           name: PageNames.TOPICS_TOPIC_SEARCH,
-          params,
         },
         ContentUnavailablePage: {
           name: PageNames.CONTENT_UNAVAILABLE,
-          params,
         },
         BookmarksPage: {
           name: PageNames.BOOKMARKS,
-          params,
         },
         ExamPage: id => {
           return {
             name: ClassesPageNames.EXAM_VIWER,
-            params: { params, quizId: id },
+            params: { quizId: id },
           };
         },
         ExamReportViewer: {
           name: ClassesPageNames.EXAM_REPORT_VIEWER,
-          params,
         },
         AllClassesPage: {
           name: ClassesPageNames.ALL_CLASSES,
-          params,
         },
         ClassAssignmentsPage: {
           name: ClassesPageNames.CLASS_ASSIGNMENTS,
-          params,
         },
         LessonPlaylistPage: {
           name: ClassesPageNames.LESSON_PLAYLIST,
-          params,
         },
       };
     },
--- a/kolibri/plugins/learn/assets/src/modules/pluginModule.js
+++ b/kolibri/plugins/learn/assets/src/modules/pluginModule.js
@@ -1,3 +1,4 @@
+import { PageNames, ClassesPageNames } from './../constants';
 import mutations from './coreLearn/mutations';
 import * as getters from './coreLearn/getters';
 import * as actions from './coreLearn/actions';
@@ -20,7 +21,60 @@ export default {
     };
   },
   actions,
-  getters,
+  getters: {
+    ...getters,
+    learnPageLinks() {
+      const params = {};
+      return {
+        HomePage: {
+          name: PageNames.HOME,
+          params,
+        },
+        LibraryPage: {
+          name: PageNames.LIBRARY,
+          params,
+        },
+        TopicsPage: {
+          name: PageNames.TOPICS_TOPIC,
+          params,
+        },
+        TopicsSearchPage: {
+          name: PageNames.TOPICS_TOPIC_SEARCH,
+          params,
+        },
+        ContentUnavailablePage: {
+          name: PageNames.CONTENT_UNAVAILABLE,
+          params,
+        },
+        BookmarksPage: {
+          name: PageNames.BOOKMARKS,
+          params,
+        },
+        ExamPage: id => {
+          return {
+            name: ClassesPageNames.EXAM_VIWER,
+            params: { params, quizId: id },
+          };
+        },
+        ExamReportViewer: {
+          name: ClassesPageNames.EXAM_REPORT_VIEWER,
+          params,
+        },
+        AllClassesPage: {
+          name: ClassesPageNames.ALL_CLASSES,
+          params,
+        },
+        ClassAssignmentsPage: {
+          name: ClassesPageNames.CLASS_ASSIGNMENTS,
+          params,
+        },
+        LessonPlaylistPage: {
+          name: ClassesPageNames.LESSON_PLAYLIST,
+          params,
+        },
+      };
+    },
+  },
   mutations,
   modules: {
     classAssignments,
--- a/kolibri/plugins/learn/assets/src/modules/pluginModule.js
+++ b/kolibri/plugins/learn/assets/src/modules/pluginModule.js
@@ -1,5 +1,4 @@
 // import { LastPages } from './../constants/lastPagesConstants';
-import { PageNames, ClassesPageNames } from './../constants';
 import mutations from './coreLearn/mutations';
 import * as getters from './coreLearn/getters';
 import * as actions from './coreLearn/actions';
@@ -22,49 +21,7 @@ export default {
     };
   },
   actions,
-  getters: {
-    ...getters,
-    learnPageLinks() {
-      return {
-        HomePage: {
-          name: PageNames.HOME,
-        },
-        LibraryPage: {
-          name: PageNames.LIBRARY,
-        },
-        TopicsPage: {
-          name: PageNames.TOPICS_TOPIC,
-        },
-        TopicsSearchPage: {
-          name: PageNames.TOPICS_TOPIC_SEARCH,
-        },
-        ContentUnavailablePage: {
-          name: PageNames.CONTENT_UNAVAILABLE,
-        },
-        BookmarksPage: {
-          name: PageNames.BOOKMARKS,
-        },
-        ExamPage: id => {
-          return {
-            name: ClassesPageNames.EXAM_VIWER,
-            params: { quizId: id },
-          };
-        },
-        ExamReportViewer: {
-          name: ClassesPageNames.EXAM_REPORT_VIEWER,
-        },
-        AllClassesPage: {
-          name: ClassesPageNames.ALL_CLASSES,
-        },
-        ClassAssignmentsPage: {
-          name: ClassesPageNames.CLASS_ASSIGNMENTS,
-        },
-        LessonPlaylistPage: {
-          name: ClassesPageNames.LESSON_PLAYLIST,
-        },
-      };
-    },
-  },
+  getters,
   mutations,
   modules: {
     classAssignments,
--- a/kolibri/plugins/learn/assets/src/routes/index.js
+++ b/kolibri/plugins/learn/assets/src/routes/index.js
@@ -12,6 +12,9 @@ import { showLibrary } from '../modules/recommended/handlers';
 import { PageNames, ClassesPageNames } from '../constants';
 import LibraryPage from '../views/LibraryPage';
 import HomePage from '../views/HomePage';
+import TopicsPage from '../views/TopicsPage';
+import ContentUnavailablePage from '../views/ContentUnavailablePage';
+import BookmarkPage from '../views/BookmarkPage.vue';
 import classesRoutes from './classesRoutes';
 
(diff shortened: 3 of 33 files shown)


kolibri/core/assets/src/views/AppBarCorePage.vue
```diff
@@ -52,6 +59,10 @@
         type: String,
         default: null,
       },
+      applyStandardLayout: {
+        type: Boolean,
+        default: true,
+      },
```
nucleogenesis: I wonder if using something like how [KDS buttonMixin's appearanceOverrides](https://github.com/learningequality/kolibri-design-system/blob/28111dfaee487cbcaed29bbda80289e343004ddf/lib/buttons-and-links/buttonMixin.js#L31-L37) works might be a more flexible / empowering solution to this issue because it would allow implementing components to directly change the styles of the content wrapper, but keep some of the fairly consistent styling such as that around the calculated `paddingTop` that ensures the content renders below the AppBar, however tall it is.

Then, similar to [buttonMixin](https://github.com/learningequality/kolibri-design-system/blob/28111dfaee487cbcaed29bbda80289e343004ddf/lib/buttons-and-links/buttonMixin.js#L135), AppBarCorePage could merge / Object.assign their overrides over the default styles.

kolibri/core/assets/src/views/AppBarCorePage.vue
```diff
@@ -61,17 +72,22 @@
       };
     },
     computed: {
+      ...mapState({
+        loading: state => state.core.loading,
+      }),
```
nucleogenesis: :cry: I really thought I got away with there being no Vuex in this component :stuck_out_tongue:

kolibri/plugins/learn/assets/src/views/ImmersivePageRoot.vue
```diff
@@ -0,0 +1,85 @@
...
+</template>
+
+
+<script>
+
+  import ImmersiveToolbar from 'kolibri.coreVue.components.ImmersiveToolbar';
+  import { mapState } from 'vuex';
+
+  export default {
+    name: 'ImmersivePageRoot',
```
nucleogenesis, line 26: I think this might be worth having as a Core component - I think I probably should have done this in the Facility PR as I had no good reason not to - also your version addresses the lack of KLinearLoader in the [Facility ImmersiveToolbar](https://github.com/learningequality/kolibri/blob/develop/kolibri/plugins/facility/assets/src/views/ImmersivePageRoot.vue) which we'll need in that plugin too.

If that's not too much to add into this PR, then you could delete the one in Facility and update the imports (and maybe add it to the API Spec).

nucleogenesis: Otherwise we can make a follow-up issue to do that before moving on into the other components for CoreBase extraction

kolibri/plugins/learn/assets/src/views/ImmersivePageRoot.vue
```diff
@@ -0,0 +1,85 @@
...
+        default: '',
+      },
+      route: {
+        type: Object,
+        default: null,
+      },
+      applyStandardLayout: {
+        type: Boolean,
+        default: true,
+      },
```
nucleogenesis: Ditto my comment on the same change in AppBarCorePage

kolibri/plugins/learn/assets/src/views/ExamPage/index.vue
```diff
@@ -173,22 +181,25 @@
...
-    metaInfo() {
-      return {
-        title: this.exam.title,
-      };
-    },
+    // metaInfo() {
+    //   return {
+    //     title: this.exam.title,
+    //   };
+    // },
```
nucleogenesis: Is this no longer needed?

kolibri/plugins/learn/assets/src/views/LearnImmersiveLayout.vue
```diff
@@ -284,6 +275,40 @@
...
+          const lastPage = this.$route.query.last;
+          route = this.$router.getRoute(lastPage, {}, query);
+        } else if (this.$route.query && this.$route.query.last) {
+          const last = this.$route.query.last;
+          route = this.$router.getRoute(last, query);
+        } else {
+          route = this.$router.getRoute(PageNames.HOME);
+        }
+        return route;
+      },
```
nucleogenesis: Just a thought for making this function a bit easier to grok.

If you declare something like `const lastPage = (this.$route.query || {}).last;` (or use `lodash.get(this, '$route.query.last', null)`) before entering the conditional block, you can clean up all of the conditionals using and assignments of `$this.route.query.last` throughout the function making it a bit easier to scan and read.

kolibri/plugins/learn/assets/src/views/classes/LessonPlaylistPage.vue
```diff
@@ -89,26 +94,29 @@
...
-            link: {
-              name: ClassesPageNames.CLASS_ASSIGNMENTS,
-              params: { classId: this.currentLesson.classroom.id },
-            },
-          },
-          {
-            text: this.currentLesson.title,
-          },
-        ];
+        console.log(this.currentLesson);
```
nucleogenesis: Console log snuck its way in

nucleogenesis (changes requested):
Overall, this looks great! I'm really happy that you found it relatively easy to take the patterns over. I've left a few comments requesting changes:

- One cleaning up a function a bit
- Open for discussion re: the styles overriding, just want to hear your and @MisRob 's thoughts about my suggestion vs keeping as-is
- Optional and non-blocking suggestion that maybe we move ImmersivePageRoot to core (and clean up Facility's version and imports in the process - but can do that in follow-up too)

kolibri/plugins/learn/assets/src/views/LearnIndex.vue
```diff
@@ -1,269 +1,36 @@
...
-
-    <template #sub-nav>
-      <LearnTopNav />
-    </template>
-
-    <template #totalPointsMenuItem>
-      <TotalPoints />
-    </template>
-
     <div>
```
MisRob: Was there a reason for keeping this `div`?

kolibri/plugins/learn/assets/src/views/LearnIndex.vue
```diff
@@ -275,6 +42,7 @@
   @import './learn';
 
   .content {
```
MisRob: I don't see this class being used in the template, so I think it could be removed

kolibri/plugins/learn/assets/src/views/LearnAppBarPage.vue
```diff
@@ -0,0 +1,67 @@
+<template>
+
+  <AppBarCorePage :title="appBarTitle" :applyStandardLayout="applyStandardLayout">
+
+    <template #subNav>
+      <LearnTopNav ref="topNav" />
+    </template>
+
+    <div v-if="!loading" :style="styles">
```
MisRob: Let's remove `v-if="!loading"` condition from here and instead let pages rendered within the default slot decide what should/shouldn't be displayed when `state.core.loading` is truthy. Leaving it here would have unhappy consequences. Imagine that we would set `state.core.loading` to `true` from a page rendered within this slot, typically done when fetching data in the `created` hook. At that very moment, the page would get destroyed because of this condition and before we would get a chance to set `state.core.loading` to `false` after data is fetched. This pattern troubled us in the past already with our previous router hooks architecture. I think it might the reason why we couldn't use `state.core.loading` state on the `Bookmarks` page and instead, internal component data `loading` was used

https://github.com/learningequality/kolibri/blob/e207ffbd006e61562a2a915af3a0df51a39d9613/kolibri/plugins/learn/assets/src/views/BookmarkPage.vue#L109-L111

https://github.com/learningequality/kolibri/blob/e207ffbd006e61562a2a915af3a0df51a39d9613/kolibri/plugins/learn/assets/src/views/BookmarkPage.vue#L125-L129


 I tried to illustrate this problem in more detail in [section (2) of this comment](https://github.com/learningequality/kolibri/pull/9128#pullrequestreview-893668096) if that'd be helpful.

kolibri/plugins/learn/assets/src/views/LearnAppBarPage.vue
```diff
@@ -0,0 +1,67 @@
...
+      <slot></slot>
+    </div>
+    <KLinearLoader
```
MisRob: I can see that `KLinearLoader` has been added to `AppBarCorePage` as well. What was the reason for having them in both these components?

kolibri/plugins/learn/assets/src/views/ImmersivePageRoot.vue
```diff
@@ -0,0 +1,85 @@
+<template>
+
+  <div class="main-wrapper" :style="wrapperStyles">
+    <ImmersiveToolbar
+      ref="appBar"
+      :appBarTitle="appBarTitle"
+      :route="route"
+    />
+    <slot v-if="!loading"></slot>
```
MisRob: Let's remove `v-if="!loading"` from here. Explained in a similar comment for `LearnAppBarPage`.

kolibri/plugins/learn/assets/src/modules/pluginModule.js
```diff
@@ -20,7 +21,60 @@ export default {
...
+          params,
+        },
+        BookmarksPage: {
+          name: PageNames.BOOKMARKS,
+          params,
+        },
+        ExamPage: id => {
+          return {
+            name: ClassesPageNames.EXAM_VIWER,
+            params: { params, quizId: id },
```
MisRob: I think this was supposed to be `params: { quizId: id }`?

kolibri/plugins/learn/assets/src/modules/pluginModule.js
```diff
@@ -20,7 +21,60 @@ export default {
     };
   },
   actions,
-  getters,
+  getters: {
+    ...getters,
+    learnPageLinks() {
+      const params = {};
```
MisRob: Just a stylistic note, sharing the same and empty `params` object across unrelated routes feels a bit confusing to me.

kolibri/plugins/learn/assets/src/routes/index.js
```diff
@@ -168,13 +176,15 @@ export default [
...
     },
+    component: TopicsPage,
   },
   {
     name: PageNames.TOPICS_CONTENT,
     path: '/topics/c/:id',
     handler: toRoute => {
       showTopicsContent(store, toRoute.params.id);
     },
+    component: LearnImmersiveLayout,
```
MisRob: This component seems to be used only on this route now and its purpose isn't to provide layout anymore so I'd suggest renaming it to something like `TopicsContentPage`

kolibri/plugins/learn/assets/src/views/LearnAppBarPage.vue
```diff
@@ -0,0 +1,67 @@
...
+
+<script>
+
+  import { mapState } from 'vuex';
+  import AppBarCorePage from 'kolibri.coreVue.components.AppBarCorePage';
+  import commonCoreStrings from 'kolibri.coreVue.mixins.commonCoreStrings';
+  import LearnTopNav from './LearnTopNav';
+
+  export default {
+    name: 'LearnAppBarPage',
```
MisRob, line 27: I'd suggest not using `_Page` in layout components names to make their purpose clear and to allow ourselves easily differentiate between layout components and real pages connected to routes. Maybe `LearnAppBarLayout`?

MisRob (dismissed):
Good work, thank you @marcellamaki. Overall, it's looking really great.

Regarding code, I left several comments, most of them non-blocking. I think it's important to resolve `loading` `v-if`s in `slot`s and also need some clarification in regard to linear loaders.

I noticed some issues when testing sidebar scrolling locally and maybe one or two other problems. I want to check those places on `develop` at first to be sure it's not broken there. I'll report it all in another comment later.

kolibri/core/assets/src/views/AppBarCorePage.vue
MisRob: I think it'd be fine to use the loading state here

kolibri/core/assets/src/views/AppBarCorePage.vue
MisRob: That would make sense to me, too

MisRob:
@marcellamaki I tested only briefly and in Chrome. So far I've noticed the following problems that don't seem to be present on `develop`:

#### (1) Blank page after navigating to a channel for the first time from the library or after the channel page reload

1. Go to the "Library" and open a channel folder in the "Channels" section (make sure that you haven't visited the channel folder before, for example by reloading the "Library" page)
2. See the blank page and `"TypeError: Cannot read properties of undefined (reading 'modality')"` error in the console

Interestingly, when I navigate back to the library from the blank page and click again the channel folder, I can then see the channel as expected. However, reloading it results in the same blank page and console error.

open-channel-bug

#### (2) Scrolling behavior

- There are two vertical scrollbars when browsing a channel in "Library"
- There seems to be a regression with the stickiness of the top bar when I scroll up and down a couple of times:

scroll

Both "China" and "Russia" items in the side panel should be visible. Note that the problem with "China" disappearing is present on `develop` (might be a regression with calculating the height of the top bar I'd guess), however, I don't see a problem with stickiness not working at all after a while on `develop`.

marcellamaki:
Thanks @MisRob and @nucleogenesis for your comments here! Sorry it has taken me so long to get back to them. I hope to have the changes ready for review by end of day tomorrow.

kolibri/plugins/learn/assets/src/views/ExamPage/index.vue
marcellamaki: accidentally commented out! Re-added

kolibri/plugins/learn/assets/src/views/LearnImmersiveLayout.vue
marcellamaki: Great suggestion -- thanks!

kolibri/plugins/learn/assets/src/views/LearnIndex.vue
marcellamaki: no, good catch. I've removed it

kolibri/plugins/learn/assets/src/modules/pluginModule.js
marcellamaki: That's a good point @MisRob -- I modeled this after [the pattern shown in the facility plugin](https://github.com/learningequality/kolibri/blob/develop/kolibri/plugins/facility/assets/src/modules/pluginModule.js#L46). 

Do you have a recommendation of how to do it differently? 
either 
`params : {}`
or...
not defining it at all unless needed?

MisRob: Ah okay, in the facility plugin it seems that it was needed because there is a shared `facility_id` field in `params` for all routes. In such a case, it'd make sense to me. I didn't see any shared data here and that's why I was wondering why it's been done this way, but it's not a big deal. Up to you :)

kolibri/core/assets/src/views/AppBarCorePage.vue
rtibbles: Could we turn this into a prop instead? Then we could get one step closer to doing away with the global `loading` Vuex state.

kolibri/plugins/learn/assets/src/modules/pluginModule.js
```diff
@@ -20,7 +22,49 @@ export default {
     };
   },
   actions,
-  getters,
+  getters: {
+    ...getters,
+    learnPageLinks() {
+      return {
```
rtibbles: Why does this need to be in vuex state? There's nothing reactive or dynamic here, these are all constants (even the ExamPage function).

kolibri/plugins/learn/assets/src/modules/pluginModule.js
```diff
@@ -20,7 +22,49 @@ export default {
...
+        },
+        ContentUnavailablePage: {
+          name: PageNames.CONTENT_UNAVAILABLE,
+        },
+        BookmarksPage: {
+          name: PageNames.BOOKMARKS,
+        },
+        ExamPage: id => {
+          return {
+            name: ClassesPageNames.EXAM_VIWER,
```
rtibbles: typo

kolibri/plugins/learn/assets/src/views/ExamPage/index.vue
```diff
@@ -324,15 +337,15 @@
       attemptLogItemValue(newVal, oldVal) {
         // HACK: manually dismiss the perseus renderer message when moving
         // to a different item (fixes #3853)
-        if (newVal !== oldVal) {
+        if (newVal !== oldVal && this.$refs.contentRenderer) {
```
rtibbles, line 342: Could you file a follow up issue here to properly fix this in the perseus renderer?

kolibri/plugins/learn/assets/src/views/ImmersivePageRoot.vue
```diff
@@ -0,0 +1,86 @@
...
+      },
+    },
+    data() {
+      return {
+        appBarHeight: 0,
+      };
+    },
+    computed: {
+      ...mapState({
+        loading: state => state.core.loading,
```
rtibbles: Let's make this a prop here too.

kolibri/plugins/learn/assets/src/views/TopicsContentPage.vue
```diff
@@ -286,6 +277,36 @@
       timeSpent() {
         return this.contentPageMounted ? this.$refs.contentPage.time_spent : 0;
       },
+      back() {
+        let route = {};
```
rtibbles: No need to assign a let here, just return instead of assigning to `route`.

Also, what happens if `route` is an empty object? Maybe it's more semantic for it to be null, then we can do falsy checking to see if we should display a link at all?

kolibri/plugins/learn/assets/src/views/TopicsContentPage.vue
```diff
@@ -286,6 +277,36 @@
...
+          const lastPage = (this.$route.query || {}).last;
+          delete query.last;
+          delete query.topicId;
+          // returning to a topic page requires an id
+          if (lastPage === PageNames.TOPICS_TOPIC_SEARCH || lastPage === PageNames.TOPICS_TOPIC) {
+            const lastId = this.$route.query.topicId
+              ? this.$route.query.topicId
+              : this.content.parent;
+            // Need to guard for parent being non-empty to avoid console errors
+            route = this.$router.getRoute(
```
rtibbles:
```suggestion
            return this.$router.getRoute(
```

kolibri/plugins/learn/assets/src/views/TopicsContentPage.vue
```diff
@@ -286,6 +277,36 @@
...
+              lastPage,
+              {
+                id: lastId,
+              },
+              query
+            );
+          } else if (lastPage === PageNames.LIBRARY) {
+            route = this.$router.getRoute(lastPage, {}, query);
```
rtibbles:
```suggestion
            return this.$router.getRoute(lastPage, {}, query);
```

kolibri/plugins/learn/assets/src/views/TopicsContentPage.vue
```diff
@@ -286,6 +277,36 @@
...
+          } else if (lastPage) {
+            route = this.$router.getRoute(lastPage, query);
```
rtibbles:
```suggestion
            return this.$router.getRoute(lastPage, query);
```

kolibri/plugins/learn/assets/src/views/TopicsContentPage.vue
```diff
@@ -286,6 +277,36 @@
...
+          } else {
+            route = this.$router.getRoute(PageNames.HOME);
```
rtibbles:
```suggestion
            return this.$router.getRoute(PageNames.HOME);
```

kolibri/plugins/learn/assets/src/views/TopicsContentPage.vue
```diff
@@ -286,6 +277,36 @@
       timeSpent() {
         return this.contentPageMounted ? this.$refs.contentPage.time_spent : 0;
       },
+      back() {
+        let route = {};
+        if (this.$route) {
```
rtibbles: Could instead use a [return early strategy](https://szymonkrajewski.pl/why-should-you-return-early/) here, which will let you unnest the rest of the function:

```
if (!this.$route) {
  return null;
}
```
then the rest of the function can just continue without being inside an if statement.

kolibri/plugins/learn/assets/src/views/TopicsPage.vue
```diff
@@ -4,11 +4,18 @@
     <div v-if="currentChannelIsCustom">
       <CustomContentRenderer :topic="topic" />
     </div>
-
-    <div v-else class="page">
+    <!-- appearanceOverrides overrides the default page styling -->
```
rtibbles: Presumably this is going to need a rebase on top of @nucleogenesis's topics refactor?

rtibbles (dismissed):
As we're doing a refactor, I'm going to be slightly picky and request  changes rather than just comment for these things.

kolibri/plugins/learn/assets/src/views/TopicsPage.vue
marcellamaki: yes

kolibri/plugins/learn/assets/src/views/ExamPage/index.vue
marcellamaki: [Follow up issue here](https://github.com/learningequality/kolibri/issues/9402)

MisRob:
@marcellamaki Thank you for all the new work on this. Did you have a chance to look into the problem with the blank page and scrolling behavior, too (https://github.com/learningequality/kolibri/pull/9222#issuecomment-1091706443)? Should I re-review something or is it fine to just dismiss my review? Wouldn't like to block this and can see that other folks had a chance to review while I was gone. Please let me know if I can do something to support. Thank you.

marcellamaki:
Hi @MisRob -- I remember looking into them, but now I cannot remember if I made and committed the changes. I will review and follow up with you if a re-review will be helpful! Thank you

nucleogenesis:
@marcellamaki - #9335 is merged so this should be hopefully unblocked except for by the conflicts that I've created :sweat_smile:

marcellamaki:
I believe all feedback has now been addressed, and this is ready for a final review (and hopefully approval, or close to it! 🤞)

radinamatic:
@pcenov Let's test this tomorrow, thank you!

nucleogenesis (approved):
If @pcenov 's recently created issues cover all outstanding bugs/problems with this PR, then I think it should be good to merge once the conflicts are resolved.

Issue #9224, opened 2022-03-27
ImportError on running migrate while using Python 3.10.0


nitheezkant:
## Observed behavior

On running  `kolibri manage migrate` command while using Python version 3.10.0, ImportError is observed.
Complete screenshot of Terminal is attached:
<img width="1343" alt="image" src="https://user-images.githubusercontent.com/85250569/160286893-58f1e248-a0e8-4d07-b60e-d5ee0db87edb.png">

## Errors and logs

```
ImportError: cannot import name 'Iterator' from 'collections' (/Library/Frameworks/Python.framework/Versions/3.10/lib/python3.10/collections/__init__.py)
```

## Expected behavior

Migration is successful when lower versions of python is used.

## Steps to reproduce

Make sure you are using Python 3.10.0 inside the virtual environment and then run `kolibri manage migrate`.

## Context

OS:macOS, M1

rtibbles:
Kolibri does not support Python 3.10.0 or later, and cannot until we drop Python 2.7 support and upgrade our Django version.

However, this could be better specified in our setup.py with a `python_requires` kwarg.

PR #9225, opened 2022-03-27
Update setup.py

nitheezkant:
## Summary

Following the issue https://github.com/learningequality/kolibri/issues/9224, since Kolibri doesn't support python 3.10.0 and above, it has been specified in python_requires. 

## References

This pull request is made following the issue that has been closed recently. The link to the same is here:https://github.com/learningequality/kolibri/issues/9224

## Reviewer guidance

Test it by trying to setup using python version that is not supported.

### DIFF
--- a/docs/getting_started.rst
+++ b/docs/getting_started.rst
@@ -63,7 +63,7 @@ Python and Pip
 
 To develop on Kolibri, you'll need:
 
-* Python 3.4+ or Python 2.7+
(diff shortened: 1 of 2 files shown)


rtibbles (approved):
Thanks!

nitheezkant:
I have further specified in the documentation that Python 3.10.0 and higher is not supported. It has been written in parenthesis under Python and Pip section under Getting Started.

PR #9230, opened 2022-03-28
Library page refactor

### LINKED ISSUE #8978: Refactor library page using a reusable/props-driven component for main content/content grid(s)
These sections can be turned into a "main view" component. This component could either be used twice within the page (one for the default view and once for results) or could be used once and have the contents changed based on whether or not search results are displayed. As a follow up issue, the same changes can be applied in the Topics Page.

**Writing unit tests for this new component is part of this issue being done**

nucleogenesis:
## Summary

LibraryPage now has become a few other pieces:

**SidePanel**

Contains logic around mobile and desktop display of the EmbeddedSidePanel (the filters)

**ResumableContentGrid**

Shows, conditionally, the resumable content for the user. Tests were basically moved over to the new component 1-to-1.

**SearchResultsGrid**

Much like ResumableContentGrid, this was migrated over with most test 1-to-1.

**useCopies** was created to allow shared store and simple API for managing the displayedCopies - was applied everywhere CopiesModal is used (and CopiesModal was updated too)

**useSearch** now exposes a `currentRoute` function because it already did the business of extracting the route and making a computed for it so it seemed convenient to do it this way.

**BUGFIX** - Embedded side panel didn't disable the buttons it ought to have, this is fixed in that file's only changes

## References

Fixes #8978 

## Reviewer guidance

Does all functionality track over?

- Mobile filters sidepanel
- Content info side panels
- Search results, clearing them
- Copies in all places it is relevant (see code)

Does the code, particularly the useCopies business and the way LibraryPage is broken down make sense? Is it cleaner and easier to reason about? Any comments that would help clarify how things are working?

----

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated


kolibri/plugins/learn/assets/src/composables/__tests__/useCopies.spec.js
```diff
@@ -0,0 +1,22 @@
...
+  it('should default to an empty array', () => {
+    const { displayedCopies } = useCopies();
+    expect(displayedCopies.copies).toEqual([]);
+  });
+  it('should set displayed copies to value passed to setCopies function', () => {
+    const { displayedCopies, setCopies } = useCopies();
+    setCopies([{ id: '1' }, { id: '2' }]);
+    expect(displayedCopies.copies).toEqual([{ id: '1' }, { id: '2' }]);
+  });
+  it('displayedCopies should be a module-level store, so the previous test should still pass', () => {
```
marcellamaki: (Non-blocking, just a learning moment question for me) Is this a good way to test persistence of data? I am still wrapping my mind around test behavior and what is unique about writing jest tests that is different from other code. Given the troubles I've run into recently around order of execution in tests, and what gets persisted where and why, I am immediately a little bit 🤔 wondering how to tell when things are testing something like the module, vs. testing something that is perhaps more related to jest itself... Not sure I am explaining this well, but if you have any thoughts here, would be happy to hear them.

marcellamaki (changes requested):
@nucleogenesis -- looking good! 

- [comment/feedback] For code read-through, on a high level this is a great improvement! It makes the LibraryPage itself so much easier to read, and I think the page is broken down into sub components in a really logical way that very much aligns with what I was thinking about when I was thinking about the refactoring as a next step. So, thanks for the work on this!

- [comment] I would love to also hear @MisRob's perspective on the code changes, since I am much more familiar with this page and so even though you've done a lot of work with refactoring, a lot of it I have context for. Having her insight on readability, etc. would be very valuable, so looking forward to her review! 

- [comment/non-blocking question] The `useCopies()` bit seems fine. I don't totally understand the _why_ of that refactoring -- is it for more consistency with using composables in more places? Or is there something it wasn't handling very well that this is specifically trying to address? I think now that it's in, totally fine to keep, but just for my own learning would love to hear more about your thinking 

- [QA feedback: **blocking**] Something is missing with the content card grids. Channels are showing up for me, but when not signed in, I don't see any search results (although, I also am not getting any errors) and when signed in, I don't see any search results or and resumable content. Search results -- no errors; resumableContent is a `.length` of `undefined` type error. Once that is sorted, I will continue with some manual QA. 

- [QA feedback: ✅] For QA, most of the page layout seems fine translated across screen sizes, and functionality of the filter side panel interactions as expected

Thanks again for your work on this. So exciting (and a relief for me personally due to my lingering shame over the state of the mess I made during 0.15 😔)

nucleogenesis:
Thanks @marcellamaki - to this question, which I'm very interested in your and @MisRob 's opinions on:

> The useCopies() bit seems fine. I don't totally understand the why of that refactoring -- is it for more consistency with using composables in more places? Or is there something it wasn't handling very well that this is specifically trying to address? I think now that it's in, totally fine to keep, but just for my own learning would love to hear more about your thinking

I was motivated to do this because it felt nice to me to have a consistent API through which the `CopiesModal` could be interacted with all without having to bootstrap into each component that uses `CopiesModal`: 

- Owning a data property `displayedCopies` of its own
- Writing the function to set its value and passing it to an event listener
- Listening to the submit event on `CopiesModal` and calling a function to clear your local `displayedCopies`

Not that this is the worst thing ever, and now that I put it that way it seems like a good candidate for a mixin, which may be simpler, but I guess as a matter of personal preference the `useCopies` approach feels like an encapsulated bit of logic that exposes an API for the domain of storing when and what copies should be shown. I also like how it simplifies the API of the `CopiesModal` component itself, no longer needing props or to emit events because it just has in hand everything it needs to manage itself from `useCopies`.

I'm not sure if this is really the best use of this pattern or if this adds some complexity that I'm not seeing at this time, but I _think_ this should be a relatively easy-to-change system. 

One thing I'd note, though, is that I was thinking maybe that our current directory structure where we dump all composables into one place might not be the best as we continue adding them (particularly if we start making and using them like this) as they might be better placed alongside their domain. For example, you'll never import `useCopies` unless you're importing `CopiesModal` - they're basically coupled together in this setup - so I wonder if it would be better to lay them out next to each other.

kolibri/plugins/learn/assets/src/composables/__tests__/useCopies.spec.js
nucleogenesis: I'm not entirely sure off hand - but I believe that once you import something like `useCopies` into your file, its module-level scope is persisted for the duration of the running of that file, in this case the entire test suite.

So, since `useCopies` is a function that returns a module-level `reactive()` variable like `displayedCopies`, I expect that that value should persist across all tests, so any time you need it to be a value, you need to set it anew.

If that is correct, then if I were testing a component using `useCopies` and wanted to mock it, then I'd want to `jest.resetAllMocks` in a `beforeEach` if I needed each test to presume the original state or if I wanted to ensure a specific state for each pass like:

```
jest.mock('./useFns');
// ... //
// later on in a describe() block
var { setFn, reactiveValue } = useFns();

beforeEach(() => {
    jest.resetAllMocks();
    setFn({ value: 'that I want to initialize in each test in this describe block' });
}

// then my tests that depend on that initial value in this test case
```

(cc @rtibbles does this all sound right?)

MisRob: (1) Persisting state across more tests
(2) Order of execution of test

Regarding (1), I'm not sure what would be best, it depends. I tend to set test conditions anew each time whenever possible, for example in `beforeEach` function that is wrapped in `describe`/`context` block together with relevant tests. But sometimes resetting mocks or sharing some test conditions is needed too on more complex pages. I'm not very opinionated, just try to make tests clear enough for others in whatever way appropriate.

Regarding (2), I think @marcellamaki made a good point. I wouldn't expect any test suite to break if I for example added some tests in between other tests that have some dependencies on order. This could happen in this test suite. For that reason, in all tests that this note is relevant for, I'd rather suggest making sure that there are no such dependencies.

For example for the last test we could do:

```
`it('should reset displayedCopies to an empty array when clearCopies is called', () => {
    const { displayedCopies, setCopies } = useCopies();
    setCopies([{ id: '1' }, { id: '2' }]);
    clearCopies();
    expect(displayedCopies.copies).toEqual([]);
`  });
```
or use `beforeEach` or `beforeAll` to `setCopies([{ id: '1' }, { id: '2' }]);` for all tests that need it, including this one.

kolibri/plugins/learn/assets/src/views/LibraryPage/index.vue
```diff
@@ -0,0 +1,249 @@
...
+        message: 'Clear all',
+        context: 'Clickable link which removes all currently applied search filters.',
+      },
+    },
+  };
+
+</script>
+
+
+<style lang="scss" scoped>
```
MisRob, line 273: In the template, I can only see these three classes: `main-grid`, `loader`, `grid`, and I haven't noticed any other computed styles or classes in this component. So I think most of the styles here are obsolete and could be removed?

kolibri/plugins/learn/assets/test/views/library-page-side-panel.spec.js
```diff
@@ -0,0 +1,23 @@
...
+  describe('handleCategory method', () => {
+    it('passes the first param to useSearch#setCategory', () => {
+      const setCategorySpy = jest.spyOn(useSearch(), 'setCategory');
+      const wrapper = shallowMount(SidePanel);
+      wrapper.vm.handleCategory('category');
+      expect(setCategorySpy).toBeCalledWith('category');
+    });
+  });
+  describe('closeCategoryModal', () => {
+    it('sets this.currentCategory to null', () => {
```
MisRob: What is this supposed to test?

kolibri/plugins/learn/assets/test/views/resumable-content-grid.spec.js
```diff
@@ -0,0 +1,73 @@
...
+
+  it('does not show the grid / list toggle buttons when on extra small screens', async () => {
+    const wrapper = shallowMount(ResumableContentGrid, {
+      computed: { windowIsSmall: () => true },
+    });
+    expect(wrapper.find('[data-test="toggle-view-buttons"]').element).toBeFalsy();
+  });
+
+  it('displays HybridLearningCardGrid', () => {
+    const wrapper = shallowMount(ResumableContentGrid, {});
```
MisRob, line 46: What's the purpose of this test?

kolibri/plugins/learn/assets/test/views/resumable-content-grid.spec.js
```diff
@@ -0,0 +1,73 @@
...
+      useLearnerResourcesMock({
+        resumableContentNodes: [{ node: 1 }],
+        moreResumableContentNodes: [],
+      })
+    );
+    const wrapper = shallowMount(ResumableContentGrid, {});
+    expect(wrapper.find('[data-test="more-resumable-nodes-button"').element).toBeFalsy();
+  });
+  describe('showing the side panel when there is `sidePanelContent`', async () => {
+    it('shows BrowseResourceMetadata', async () => {
```
MisRob: It's hard to understand what scenario is being tested here without knowing the internals of the component in detail and this test could pass even when the side panel wouldn't get displayed properly because the `sidePanelContent` data being truthy is just a piece of all related logic.

I'd suggest testing something like "resource metadata will be displayed in a side panel after clicking info toggle button", in the test implementation as well as using it as the test's description.

kolibri/plugins/learn/assets/test/views/search-results-grid.spec.js
```diff
@@ -0,0 +1,126 @@
...
+/* eslint-enable import/named */
+
+const SearchStrings = createTranslator('SearchResultsGrid', SearchResultsGrid.$trs);
+const coreStrings = commonCoreStrings.methods.coreString;
+
+jest.mock('../../src/composables/useSearch');
+
+describe('when search results are loaded', () => {
+  /* useSearch#displayingSearchResults is truthy and isLoading is false */
+  it('does not display a ChannelCardGroupGrid', () => {
```
MisRob: I'd suggest renaming this in a way that you can understand what's not being displayed even without being familiar with `ChannelCardGroupGrid`, e.g. "does not display a list of channels".

kolibri/plugins/learn/assets/src/composables/useCopies.js
```diff
@@ -0,0 +1,21 @@
+import { set } from '@vueuse/core';
+import { reactive } from 'kolibri.lib.vueCompositionApi';
+
+// Shared store - an object because it uses Vue.observable under the hood
+var displayedCopies = reactive({ copies: [] });
```
MisRob: As I mentioned on Slack when we chatted about other composable, I think it'd be fine to use a shared global state in some composables when there's a good reason to do so. However, I don't understand why it's needed for this data? From code and clicking through the library, it seems to me copies show only in the copies modal that is open for one resource at a time. Was there any reason why having it defined in the function scope of `useCopies` wouldn't be sufficient?

MisRob (commented):
Good work, thank you @nucleogenesis. I added some non-blocking comments.

Like @marcellamaki, I wasn't able to see the "Recent" section on the "Library" page.

MisRob:
~Regarding `useCopies` and `CopiesModal` I think we need to get some experience but haven't noticed any problems. In this particular scenario, having the logic gathered this way makes sense to me and personally, I consider this approach more readable than using a mixin.~

~I only left one note about the internal implementation of `useCopies` in regard to global data.~

EDIT: I misunderstood the purpose of the composable previously, assuming that it was implemented to replace the mixin alternative which doesn't seem to be the case because mixins are not built around a global state. I agree that it'd be better not to move towards sharing state globally unless necessary.

kolibri/plugins/learn/assets/src/composables/__tests__/useCopies.spec.js
rtibbles: I am inclined to agree, sharing state across tests is a code smell to me. The only time sharing state across tests is a good idea is when the state takes a long time to setup (like in Django when we setup database state that is shared across multiple tests - but due to Django transactional encapsulation of tests, there is no actual sharing of state in that case).

Given that all state is being setup in memory here, there are no performance benefits, so I think we should use a `beforeEach` to set this up for each test.

kolibri/plugins/learn/assets/src/composables/useCopies.js
rtibbles: It looks like the global state is needed because this is referenced in the parent component that has the `CopiesModal` in it, and the child `CopiesModal` component.

I think using the composable in this case is adding an additional layer of indirection here - rather than simply passing in the copies via a prop, we are setting it in some global state, then reading from that global state inside another component.

rtibbles (changes requested):
I think the changes to CopiesModal here should either be reverted, or at the very least just changed back to props, the use of the composable to create a mini-global store adds indirection that I think is in the wrong direction of travel for explicit, primarily props driven data passing.

I don't think this is a case where we should use a composable plus provide/inject, as it's literally just passing data from a parent to a child - a clear use case for props.

kolibri/plugins/learn/assets/src/composables/useCopies.js
MisRob: Ah okay, I haven't noticed this. In that case, it'd make sense to stay with props to me, too.

kolibri/plugins/learn/assets/src/composables/__tests__/useCopies.spec.js
nucleogenesis: This is all great feedback, thank you all.

I did it how I did to test the fact that the state was indeed outside of the function scope, but I could have tested that within the same `it` block.

I think that with these composables, using the mocks and the `resetAllMocks` function makes it clear when/how the state is being reset - but in the case above this isn't immediately clear.

In any case, I'll be refactoring this out but will definitely take this with me as I continue to develop test suites.

kolibri/plugins/learn/assets/test/views/library-page-side-panel.spec.js
nucleogenesis: Ah it is supposed to test that the method changes the state - but seems like it never sets it to open in the first place to be sure that the data has changed - I believe it defaults to null in this case.

I'll revisit this test

nucleogenesis (commented):
I think I got ahead of myself trying to DRY things up with the `useCopies` business. FWIW - the thing I like about it is that it is an encapsulation of a piece of logic with a simple API to do the things that this specific instance would need. It's like a microservice for storing a piece of data that one component uses without having to wire up props and events and their accompanying methods and local state. CopiesModal reads the state and acts based on it, other components can write to the state when they want the CopiesModal to be visible. In any case - I definitely see how the indirection here is overkill for the problem at hand.

The thought re: a mixin is less about "global state" and more about just making all of that stuff, the props, events, handlers, local state live in one place that can be mixed in wherever the `CopieModal` component would be used - but even here this is only used in a couple places and I think I'd rather have duplicated logic between components than a mixin in this case, particularly because I doubt the solution is inclined to be changed much.

Thank you all for the feedback - I am glad I tried something that is relatively new to me and for your critical eyes on it :) 

---

I'll be following up on this PR tomorrow and pushing changes hopefully midday tomorrow as I don't think it'll take too much time to clean up the tests per Misha's comments and to move the CopiesModal business into the components.

kolibri/plugins/learn/assets/test/views/library-page-side-panel.spec.js
MisRob: I think that even if this method changes the state as expected, it doesn't necessarily mean that closing the category modal works? If that was the scenario that you wanted to test (that's why I asked first)

kolibri/plugins/learn/assets/test/views/resumable-content-grid.spec.js
nucleogenesis: Just to test that a component is loaded - I think it was already in the LibararyPage tests that I broke up into multiple components - not particularly useful so I've removed it

nucleogenesis:
@MisRob @rtibbles @marcellamaki 

I've updated the PR and fixed the issues with search not showing up (which was because I was treating `useSearch` as if it had module-level shared state) - now all search business is instantiated still in LibraryPage, then passed to the new children via props.

useCopies is gone and the feature uses props now

MisRob (dismissed):
Thank you, @nucleogenesis. Code changes look good to me. I can still see the problem with the "Recent" section not showing up (and most likely related `TypeError` in the console):

| develop (`7cfab50f9b3fbbe684775f8f01c2e09507d75e7f` commit) | PR (`c4e74b404867365f89f7b8ce8d64e1d5349a5977` commit) |
| --- | --- |
| 7cfab50f9b3fbbe684775f8f01c2e09507d75e7f | c4e74b404867365f89f7b8ce8d64e1d5349a5977 |

I won't be able to re-review again since this is the last day before my time off in the upcoming weeks and I haven't noticed any other problems, so after this issue is fixed, I think it'd be good to merge, unless @marcellamaki has some more notes.

nucleogenesis:
Thanks @MisRob - the issue there is resolved

marcellamaki (approved):
`resumableContentGrid` now appearing! ✅

rtibbles (approved):

PR #9241, opened 2022-04-01
ref(FSSP): Implements suggestions in #8989

### LINKED ISSUE #8989: Refactor: adding optional/required props in `FullScreenSidePanel`
## Observed behavior
Currently, there are two places where the `close` button that appears in mobile can be rendered from: `FullScreenSidePanel` and `LibraryPage`. As it currently stands in the code, the `close` button in `FullScreenSidePanel` is always hidden when a user is in the Library page and in a Topic page, and that button is rendered instead by the `LibraryPage` and `TopicsPage` parent components. 

Two decisions to be made:
- [ ] Which component should be rendering the `close` button - the `FullScreenSidePanel` or the parents (`LibraryPage`/`TopicsPage`)?
- [ ] Should the `closeButtonHidden` prop be required, and if so, should we refactor all instances of `FullScreenSidePanel` with that required prop?

Screen Shot 2022-01-12 at 2 13 50 PM

## Errors and logs
No errors.

## Expected behavior
See decisions above.

## User-facing consequences
No user-facing consequences, but for good for cleaning up our code.

## Steps to reproduce
1. Log in as Learner and on mobile
2. Go to library page and click on the "Filter" button
3. This is in the `LibraryPage` and `FullScreenSidePanel` components

d0sadata:
## Summary:
Changed location of closeButton,
Changed handling in Library and Topics,
Added props in all 4 pages that implement,
Added RTL and LTR handling for style of button,
Added icon handler for button.

In the following screenshots you can see the future implementation:
fssp_LTR
fssp_LTR_arrow
fssp_RTL
fssp_RTL_arrow

Moved the logic of the button to FSSP, 
Added event handlers that mimic the previous ones in Library and Topic, 
Removed the check if the FSSP close button needs to be shown since we always use it now,
Added prop to determine which Icon it is.
Added computed properties that change icon according to FSSP width in Topic and Library.

## References
Resolves #8989

## Reviewer guidance
Open Library Page, Any Folders Info or filter, 
Infor or Bookmark, 
And in Learn > Any Resource > The book icon or Info icon. 
These are all the uses of FSSP

## Tech Debt
There is a props error in console, Width prop on the child  EmbeddedSidePanel is missing when it's used in FSSP, yet it uses a computed property otherwise. 
I need consultation on how to make the Top and Right/left positioning on the sidepanel more scalable and adjustable to different components it appears in.

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/core/assets/src/views/FullScreenSidePanel/index.vue
+++ b/kolibri/core/assets/src/views/FullScreenSidePanel/index.vue
@@ -77,6 +77,9 @@
         type: String,
         required: true,
         default: 'close',
+        validator: value => {
+          return ['close', 'back'].includes(value);
+        },
       },
       /* Optionally override the default width of the side panel with valid CSS value */
       sidePanelWidth: {
@@ -150,7 +153,7 @@
           'z-index': 12,
         };
       },
-      /* Change of position with change of close button type, default is close, */
+      /* Change of position with change of close button type, default is close */
       closeButtonStyle() {
         if (this.isRtl) {
           if (this.closeButtonIconType === 'close') {
--- a/kolibri/core/assets/src/views/FullScreenSidePanel/index.vue
+++ b/kolibri/core/assets/src/views/FullScreenSidePanel/index.vue
@@ -30,9 +30,9 @@
           </div>
 
           <KIconButton
-            v-if="fullScreenSidePanelCloseButton"
-            icon="close"
+            :icon="closeButtonIconType"
             class="close-button"
+            :style="closeButtonStyle"
             :ariaLabel="coreString('closeAction')"
             :tooltip="coreString('closeAction')"
             @click="closePanel"
@@ -72,9 +72,11 @@
     },
     mixins: [responsiveWindowMixin, commonCoreStrings],
     props: {
-      fullScreenSidePanelCloseButton: {
-        type: Boolean,
-        default: true,
+      /* CloseButtonIconType icon from parent component */
+      closeButtonIconType: {
+        type: String,
+        required: true,
+        default: 'close',
       },
       /* Optionally override the default width of the side panel with valid CSS value */
       sidePanelWidth: {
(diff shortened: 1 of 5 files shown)


marcellamaki:
~~@d0sadata -- I've started a code read through of this, but I'm having a syntax error when I try to pull your changes for QA. Can you update the name of your branch to something without parentheses? Thanks!~~

False alarm. @sairina putting in '' will resolve if you run into the same issue 
`git fetch upstream pull/9241/head:'fix(FSSP-close-functionallity)'`

kolibri/core/assets/src/views/FullScreenSidePanel/index.vue
```diff
@@ -72,9 +72,11 @@
...
     mixins: [responsiveWindowMixin, commonCoreStrings],
     props: {
-      fullScreenSidePanelCloseButton: {
-        type: Boolean,
-        default: true,
+      /* CloseButtonIconType icon from parent component */
+      closeButtonIconType: {
+        type: String,
+        required: true,
+        default: 'close',
```
marcellamaki, line 79: You might add a validator here, to ensure that the strings that are passed as props are ones that are `===` to one of the options that is needed. 

Here is [an example](https://github.com/learningequality/kolibri/blob/develop/kolibri/plugins/learn/assets/src/views/LearningActivityDuration/index.vue#L76)

kolibri/plugins/learn/assets/src/views/LibraryPage.vue
```diff
@@ -156,19 +156,19 @@
...
       alignment="left"
-      :fullScreenSidePanelCloseButton="displayCloseButton"
+      :closeButtonIconType="closeButtonIcon"
       :sidePanelOverrideWidth="`${sidePanelOverlayWidth}px`"
-      @closePanel="toggleSidePanelVisibility"
+      @closePanel="closeEventHandler()"
       @shouldFocusFirstEl="findFirstEl()"
     >
-      <KIconButton
+      <!-- <KIconButton
```
marcellamaki: I think this can be removed, not just commented out?

marcellamaki (commented):
Hi @d0sadata -- overall this is looking great! I've done some manual QA and so far I haven't seen any issues. I'm going to give it another look over tomorrow morning, but I've added two (very small) comments that you can take a look at in the meantime.

I also was wondering what you meant by making your changes "more scalable" for all uses. Are there specific uses you are thinking about? I'm not totally sure what the tech debt is here.

d0sadata:
In regards to the close button positioning, It's currently rendered by fixed values, which i don't think will work properly for all type of displays. since the button is a fixed top: and left or right(depending on type of button and RTL check) position away from the element.
image

To solve this I needed to added a height adjustment to the base element that adjusts it down by a certain pixel amount if the adjustment is not done allready.
image


and in regards to the other thing I mentiont it's in reference to this prop. 
image

In the image above, you can see the difference when EmbeddedSidePanel is implemented inside and outside FSSP
When it's a standalone side panel the **width** prop, whis is a required prop is passed. 
When it's inside it has no prop and throws a prop error. My question was, should we add width there or make it an optional prop in EmbeddedSidePanel so it has no error. 
The width is handled by FSSP when it's the child, and by itself when it's not FSSP's child.

marcellamaki (approved):
Hi @d0sadata 

This looks good. I think in regards to the width, making it an option prop makes sense. If you would like to add that to this PR, go ahead. Otherwise, you can make a follow up issue and this is good to go.

rtibbles:
Follow up issue here: https://github.com/learningequality/kolibri/issues/9349

PR #9246, opened 2022-04-01
Remote_Content frontend implementation

### LINKED ISSUE #8994: REMOTE_CONTENT option does not affect required storage at content import
## Observed behavior
When using the `REMOTE_CONTENT` option, the device plugin is not aware that this means that no local storage space is required in order to import content.

This means that on devices using this option, the amount of content that can be imported at one time will be artificially restricted.

## Expected behavior
Should be able to import any amount of content when `REMOTE_CONTENT` is set.

## User-facing consequences
Not able to provision servers effectively.

d0sadata:
## Summary

added a path to check if remote content is enabled, Thanks @rtibbles 
added checks for it to stop the disable of import buttons,
added check to hide free space available

## Screenshots
When Remote_Content is off, And file storage space is enough and not enough:
remote_off
remote_off_file_low

When Remote_Content is on, and file storage space is enough and not enough:
remote_on
remote_on_file_low

## References
@rtibbles Helped me with the python part, and linking it to the JS data ingestion point.
Resolves #8994

## Reviewer guidance
by using the Options.ini file to change REMOTE_CONTENT=TRUE or optimally by loading in remote content

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/plugins/device/assets/src/views/AvailableChannelsPage/index.vue
+++ b/kolibri/plugins/device/assets/src/views/AvailableChannelsPage/index.vue
@@ -126,6 +126,7 @@
   import ChannelTokenModal from './ChannelTokenModal';
   import ChannelUpdateModal from './ChannelUpdateModal';
   import { getFreeSpaceOnServer } from './api';
+  import plugin_data from 'plugin_data';
 
   export default {
     name: 'AvailableChannelsPage',
@@ -154,6 +155,7 @@
         freeSpace: null,
         disableBottomBar: false,
         disableModal: false,
+        remoteContentEnabled: plugin_data.isRemoteContent,
       };
     },
     computed: {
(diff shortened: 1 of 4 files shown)


rtibbles (commented):
Code looks good to me.

pcenov:
Hi @d0sadata I tested this with both the Ubuntu and Windows builds and things are looking as described by you with the exception of a message about not enough space which is still being displayed after I've set `REMOTE_CONTENT = True` which I guess should also be removed:


A question for @rtibbles would be that since the imported content looks imported but it's actually not how can we test that we can interact with such remote content? Is this part of a separate PR?

rtibbles:
> a message about not enough space which is still being displayed after I've set REMOTE_CONTENT = True which I guess should also be removed:

Looks like I missed this when going through this with you @d0sadata - I think this would be good to hide in this case too. I think this computed prop is what needs to be updated: https://github.com/learningequality/kolibri/blob/develop/kolibri/plugins/device/assets/src/views/AvailableChannelsPage/index.vue#L237


@pcenov the ability to serve content from remote source is not built into Kolibri - it requires additional server configuration external to Kolibri at this point, so there's no way to test the content once imported.

rtibbles (approved):
This is the correct additional change to handle the other messaging!

pcenov:
Retested and verified as fixed in the latest [build](https://buildkite.com/learningequality/kolibri-python-package/builds/5222).

PR #9254, opened 2022-04-08
Fix Kolibri blocking after starting with port 0

dylanmccall:
## Summary

After the server starts, `BaseServerPlugin` waits for it to respond on the
expected port number. For this feature to work, we need it to see the
bind address from the associated `httpserver,` which is different from the
original bind address when binding to port 0.

For clarity, we will not initialize the `cheroot.wsgi` Server with a
`bind_addr`. Instead, it will be controlled by `ServerPlugin` itself.

With this change in place, we can remove `PortCache` because we no longer
need to reuse port numbers.

## References

The port cache was originally implemented in #8026.

This is dealing with a peculiarity of magicbus, where it waits for the port number described in `ServerPlugin.bind_addr` to be used (and, later, to be freed): https://github.com/cherrypy/magicbus/blob/main/magicbus/plugins/servers.py#L194-L205.

## Reviewer guidance

We need to check that Kolibri starts and stops as expected with this change, both with `KOLIBRI_HTTP_PORT=0` and with some other valid port number such as `KOLIBRI_HTTP_PORT=8080`.

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/utils/server.py
+++ b/kolibri/utils/server.py
@@ -61,9 +61,6 @@ PID_FILE = os.path.join(conf.KOLIBRI_HOME, "server.pid")
 # File used to activate profiling middleware and get profiler PID
 PROFILE_LOCK = os.path.join(conf.KOLIBRI_HOME, "server_profile.lock")
 
-# File used to store previously available ports
-PORT_CACHE = os.path.join(conf.KOLIBRI_HOME, "port_cache")
-
 # File used to send a state transition command to the server process
 PROCESS_CONTROL_FLAG = os.path.join(conf.KOLIBRI_HOME, "process_control.flag")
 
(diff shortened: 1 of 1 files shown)


kolibri/utils/server.py
```diff
@@ -129,73 +126,31 @@ def port_is_available_on_host(host, port):
...
-
 class ServerPlugin(BaseServerPlugin):
-    def subscribe(self):
-        super(ServerPlugin, self).subscribe()
-        self.bus.subscribe("ENTER", self.ENTER)
+    def __init__(self, *args, **kwargs):
+        httpserver = kwargs.get("httpserver")
+        if not isinstance(httpserver, BaseServer):
+            raise TypeError("httpserver must be a cheroot.wsgi.BaseServer")
+        super(ServerPlugin, self).__init__(*args, **kwargs)
```
rtibbles, line 137:
```suggestion
        # The server is initialized without a bind address before it is passed into the plugin.
        # Because of the property setter below for `bind_addr` the setting of `self.bind_addr`
        # in the super invocation here, results in the httpserver's `bind_addr` property being set.
        super(ServerPlugin, self).__init__(*args, **kwargs)
```

rtibbles (approved):
Manual testing checks out, just added one code comment suggestion to encode what I had to read very carefully to work out/understand.

dylanmccall:
:+1: good extra comment. Thank you, @rtibbles! :)

PR #9259, opened 2022-04-08
Add importcontent --fail-on-error option

### LINKED ISSUE #9258: importcontent command should allow failure on import errors
## Observed behavior

When running the `importcontent` management command, it succeeds with a warning when files failed to be imported.

## Errors and logs

```
WARNING  32 files are skipped, because errors occurred during the import.
```

## Expected behavior

`importcontent` fails when it can't import all files.

## User-facing consequences

If you're trying to provision a fully populated device, then you'll have to dig through the logs to find out if it really succeeded or not.

## Steps to reproduce

```
python -m kolibri manage importchannel network 2091ca47ff544c96b4ae02b3a92346e1
python -m kolibri manage importcontent network --baseurl https://example.com 2091ca47ff544c96b4ae02b3a92346e1
echo $?
```

The last step would print `0` while I'd expect it to have some non-`0` failure code.

## Context

This is the develop branch as of 6f7f0b866e71fd4e89ff9bf9a3a686738bee29af on Linux (EndlessOS). Where this really affects us is in our image builder. There we're installing Kolibri via the `org.learningequality.Kolibri` flatpak app in flathub. Currently this uses kolibri-installer-gnome 2.0, which in turn includes Kolibri 0.14.7 (AFAICT).

We want to pre-populate our OS images with Kolibri content. If the content isn't reliably downloaded, then we might ship images that are missing channel objects.

dbnicholson:
Normally importcontent carries on when it can't import a file and prints
a warning on the number of skipped files. That makes it unreliable when
trying to ensure a fully provisioned device. Add the --fail-on-error
option to make import errors fatal.

Fixes: #9258

## Summary

If you're trying to provision a device and want to ensure fully populated channels, then you want `importcontent` to fail instead of warn when it can't import files. I've run the added test locally as well as manually run `importcontent --fail-on-error` with a fake `--baseurl`.

## References

#9258

## Reviewer guidance

This seems safe to add since it's narrowly scoped and the default is unchanged. The tests already cover a wide range of importcontent cases, so it should be noticeable if there are any regressions.

I'd appreciate some guidance on the test I added. I basically copied the 404 test, but I admit to not understanding the content API well enough to really understand the best strategy for testing this feature.

----

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/core/content/management/commands/importcontent.py
+++ b/kolibri/core/content/management/commands/importcontent.py
@@ -112,6 +112,14 @@ class Command(AsyncCommand):
             help="Import all updated content after a channel version upgrade",
         )
 
+        parser.add_argument(
+            "--fail-on-error",
+            action="store_true",
+            default=False,
+            dest="fail_on_error",
+            help="Raise an error when a file has failed to be imported",
+        )
+
         # to implement these two groups of commands and their corresponding
         # arguments, we'll need argparse.subparsers.
         subparsers = parser.add_subparsers(
@@ -160,6 +168,7 @@ class Command(AsyncCommand):
         peer_id=None,
         renderable_only=True,
         import_updates=False,
+        fail_on_error=False,
     ):
         self._transfer(
             DOWNLOAD_METHOD,
@@ -170,6 +179,7 @@ class Command(AsyncCommand):
             peer_id=peer_id,
             renderable_only=renderable_only,
             import_updates=import_updates,
+            fail_on_error=fail_on_error,
         )
 
     def copy_content(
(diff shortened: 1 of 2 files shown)


dbnicholson:
After writing the description and commit message, I think the option should be called `--fail-on-error`. The command sees errors and what this does is make it fail instead of warn. What do you think?

rtibbles:
Yep, that works!

dbnicholson:
Done. I didn't change the branch name since it would require opening a new PR.

rtibbles:
I think we can live with that! :)

rtibbles (approved):
Test case covers the issue, and all appropriate code paths updated.

PR #9291, opened 2022-04-13
Update process bus for easier reuse

### LINKED ISSUE #8273: Feature request: make LISTEN_ADDRESS an option
At the moment, Kolibri's server always listens on all network interfaces, with the following code in [utils/server.py](https://github.com/learningequality/kolibri/blob/release-v0.14.x/kolibri/utils/server.py#L64-L65):

```
# Currently non-configurable until we know how to properly handle this
LISTEN_ADDRESS = "0.0.0.0"
```

On Windows, listening on all interfaces requires a firewall exception, meaning a user is greeted with a scary firewall dialog when they first run Kolibri. This makes sense usually (we want Kolibri instances to talk to each other and firewall dialogs are not exactly uncommon in Windows), but it is problematic in an application where we are providing pre-installed Kolibri instances for _offline_ users who don't need that functionality. So, it would be nice if the listen address was configurable :)

### LINKED ISSUE #9260: Refactor the KolibriProcessBus to allow for easier subclassing and selective inclusion of different architectural features
Currently the KolibriProcessBus is a monolith that includes a range of different functionality with conditional flags for enabling and disabling.

This means that running a slightly different architecture than the one envisaged by the core repo involves duplicating a lot of Kolibri code, rather than just piecing together the different parts that are needed.

The bus should be broken down into three classes that inherit from each other:

BaseKolibriProcessBus - this includes all core magicbus plugins that should always be run:
* ThreadWait plugin
* PIDPlugin
* LogPlugin
* SignalHandler plugin
* Autoreloader for development mode
* ProcessControlPlugin

KolibriServicesProcessBus - which inherits from the above and includes the following additional plugins:
* ServicesPlugin
* ZeroConfPlugin

KolibriProcessBus - which inherits from the above and includes the following additional plugins:
* KolibriServerPlugin
* ZipContentServerPlugin

The DaemonizePlugin should be added in the `start` function, as most external architectures do not need to leverage it.

rtibbles:
## Summary
* Makes listen address configurable.
* Refactors the process bus to have multiple inheriting classes that each attach different plugins to the process bus
* Moves all plugin specific logic into plugins themselves, rather than having it at the bus level
* Does this to allow wrapper apps (like the Android App) to implement their own architectures for running Kolibri more easily.

## References
Fixes #9260
Fixes #8273

## Reviewer guidance
Has anything been lost in the reshuffle? Do `kolibri start` and `kolibri services` and the `--foreground` flag all work as expected?

----

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/utils/server.py
+++ b/kolibri/utils/server.py
@@ -99,6 +99,18 @@ class NotRunning(Exception):
         super(NotRunning, self).__init__()
 
 
+class PortOccupied(OSError):
+    pass
+
+
+class RunningException(PortOccupied):
+    """
+    Raised when server already appears to be running
+    """
+
+    pass
+
+
 class Server(BaseServer):
     def error_log(self, msg="", level=20, traceback=False):
         if traceback:
@@ -195,7 +207,7 @@ class KolibriServerPlugin(ServerPlugin):
                 "There is another Kolibri server running. "
                 "Please use `kolibri stop` and try again."
             )
-            sys.exit(1)
+            raise RunningException("There is another Kolibri server running.")
         super(KolibriServerPlugin, self).__init__(bus, port)
 
     @property
@@ -399,7 +411,7 @@ def _port_check(port):
             "Please check that you do not have other processes "
             "running on this port and try again.\n".format(port)
         )
-        sys.exit(1)
+        raise PortOccupied("Port {} is occupied.".format(port))
 
 
 class DaemonizePlugin(SimplePlugin):
--- a/kolibri/utils/server.py
+++ b/kolibri/utils/server.py
@@ -281,10 +281,21 @@ class ServicesPlugin(SimplePlugin):
 class ZeroConfPlugin(Monitor):
     def __init__(self, bus, port):
         self.port = port
-        Monitor.__init__(self, bus, self.run, frequency=5)
+        if conf.OPTIONS["Deployment"]["LISTEN_ADDRESS"] == "0.0.0.0":
+            # Only bother doing dynamic updates of the zeroconf service if we're bound
+            # to all available IP addresses.
+            Monitor.__init__(self, bus, self.run, frequency=5)
         self.bus.subscribe("SERVING", self.SERVING)
         self.bus.subscribe("UPDATE_ZEROCONF", self.UPDATE_ZEROCONF)
         self.broadcast = None
+
+    @property
+    def interfaces(self):
+        return (
+            InterfaceChoice.All
+            if conf.OPTIONS["Deployment"]["LISTEN_ADDRESS"] == "0.0.0.0"
+            else [conf.OPTIONS["Deployment"]["LISTEN_ADDRESS"]]
+        )
 
     def SERVING(self, port):
         # Register the Kolibri zeroconf service so it will be discoverable on the network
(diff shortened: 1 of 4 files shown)


kolibri/utils/server.py
```diff
@@ -343,9 +388,46 @@ def EXIT(self):
...
+            conf.OPTIONS["Deployment"]["LISTEN_ADDRESS"], int(port)
+        )
+    ):
+        # Port is occupied
+        logger.error(
+            "Port {} is occupied.\n"
+            "Please check that you do not have other processes "
+            "running on this port and try again.\n".format(port)
+        )
+        sys.exit(1)
```
dylanmccall: Is it reasonable for this to raise a (regular) exception, instead? It seems a bit rude for a function in utils to call `sys.exit()` (unless the calling code has clearly specified that that is okay).

kolibri/utils/server.py
```diff
@@ -68,9 +68,6 @@
 # might not have made it to the log file!
 DAEMON_LOG = os.path.join(conf.LOG_ROOT, "daemon.txt")
 
-# Currently non-configurable until we know how to properly handle this
-LISTEN_ADDRESS = "0.0.0.0"
```
dylanmccall, line 72: 🥳

kolibri/utils/server.py
rtibbles: Yeah that does seem reasonable, we should move the exits to the cli only.

jredrejo (changes requested):
There are a couple of changes needed, all in the `kolibri/utils/server.py` file, due to the use of the ZeroConf `get_all_addresses` function. This function now only can be called if `LISTEN_ADDRESS == '0.0.0.0'`
1. `ZeroConfPlugin::run `
2. `get_urls` function

Other than that, it works perfectly, and code looks fine.
is this a feature planned for 0.15.x or 0.16.x? It would be good to know it to schedule the work on https://github.com/learningequality/kolibri-server/issues/95

rtibbles:
This is targeted at develop, so is 0.16 only.

rtibbles:
Added a couple of extra commits to address the two areas you identified, @jredrejo, thanks!

kolibri/utils/server.py
```diff
@@ -289,11 +289,22 @@ def STOP(self):
 class ZeroConfPlugin(Monitor):
     def __init__(self, bus, port):
         self.port = port
-        Monitor.__init__(self, bus, self.run, frequency=5)
+        if conf.OPTIONS["Deployment"]["LISTEN_ADDRESS"] == "0.0.0.0":
+            # Only bother doing dynamic updates of the zeroconf service if we're bound
```
jredrejo, line 285: why does listening on a single one of the device interfaces block dynamic updates?

jredrejo (commented):
I still have one doubt and, after the latest chagnes in develop, now there are merge conflicts that need to be solved.

kolibri/utils/server.py
rtibbles: The dynamic updates are only required when new IP addresses are available to broadcast/listen on in zeroconf. If we have bound to a specific listen address, then this is not necessary, as we are always listening/broadcasting only on that specific listen address.

If we've bound to 0.0.0.0 then if a new IP address is now bound to a local interface, we should start listening and broadcasting on it.

jredrejo: thanks for the explanation!

jredrejo (approved):

PR #9292, opened 2022-04-13
Simplify task execution

### LINKED ISSUE #9261: Refactor the task API to collapse task scheduling and task enqueueing and reduce worker responsibility
Currently, scheduled tasks are treated entirely separately from enqueued tasks, and to be executed scheduled tasks must be picked up by our task scheduler and then put into the job queue.

This design was inspired by [rq-scheduler](https://github.com/rq/rq-scheduler) which is an addon for [rq](https://python-rq.org/) and stores its own schedule of tasks and then puts them into the queue at the appropriate time.

While this approach works, this separation is not necessary in our context, where we have complete control over the entire job queue and can embellish our data models to store all the relevant information in the same store.

Instead we can delegate all of the responsibilities of the task scheduler to a combination of the `Storage` class, and the `Worker` class.

Firstly, the storage class can house all the methods of the current scheduler, with an enhancement of the ORMJob class to add the additional metadata. In this way, an enqueued job simply becomes a special case of a scheduled job with a `repeat` value of `0` - it is executed once and that is it.

Secondly the worker class, when checking for the next job to execute can look at enqueued and scheduled tasks at the same time, and choose the highest priority from each. To simplify the search, a task that is simply enqueued can be enqueued with a `scheduled_time` of `now`, so it is immediately eligible for execution.

Lastly, the job status reporting functions (which have already been partially delegated to the [job execution function](https://github.com/learningequality/kolibri/blob/develop/kolibri/core/tasks/job.py#L96) by allowing tasks to directly report to the storage class via the job, progress updates and changes to additional metadata) should be entirely delegated to the job execution function. This means that we can swap in alternative task runners, while maintaining our task status reporting. This is especially important for the Android use case, where tasks would be executed by Android APIs and not via an ExecutorPool initiated by our Worker class.

rtibbles:
## Summary
* Consolidates all task reporting functionality into the task execution function - meaning that the worker no longer has to collect and report back task status.
* This is important as it allows us to continue to do all of our status tracking, even if we slot in another worker implementation as long as it still executes the `execute_job` function
* Consolidates the scheduler with the storage and worker classes
* This means that we have a single table for all jobs, allowing for task enqueuing and scheduling to trigger external task handlers (such as the Android WorkManager API).
* Also means we can avoid running an additional process for checking for scheduled tasks to execute.
* The worker now looks for tasks that are ready for execution
* Enqueued tasks are now just a special case of scheduled tasks, that always repeat 0 times.

## References
Fixes #9261

## Reviewer guidance
Confirm all the test cases have been properly updated.

----

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated


vkWeb (approved):
The code looks good. 

Also, I've manually tested running some tasks and made sure that our fetching of next queued job is correct. So, LGTM! :tada:

@rtibbles we just need to update our tasks documentation about `Priorirty` changes: https://github.com/learningequality/kolibri/blob/develop/docs/backend_architecture/tasks.rst. Can be done in a future PR or on this one itself.

jredrejo (commented):
Code looks good, and tested with some tasks it work fine.
However I haven't been able to test it with the setup wizard because after merging #8303 into develop, the wizard setup has stopped working when importing users to create a SoUD. I'll create an issue to fix it.

rtibbles:
Have rebased to latest develop and added documentation updates to reflect the change in our Priority mapping.

PR #9294, opened 2022-04-14
Release process automation

rtibbles:
## Summary
* Modularizes some previous asset handling Javascript
* Implements release process automation when a new release is published on the Kolibri repo
* Builds all assets
* Builds the DMG with the release flag set to `true` so that it builds a codesigned DMG
* Does an upload to testpypi of the whl and tar.gz files
* Specifies a `release` environment that requires manual approval before proceeding to later jobs
* Uploads to pypi, dependent on the manual verification step
* Uploads to the release assets page, dependent on the manual verification step
* Uploads to Google Cloud Storage, dependent on the manual verification step

## Reviewer guidance
This is a work in progress and will not currently function.

Things currently missing:
- [x] Setting up all needed secrets on the Kolibri repo, including:
  - [x] PyPi token
  - [x] Test PyPi token
  - [x] a GPG signing private key and passphrase
  - [x] Secrets for GCS Bucket and credentials
- [x] It's possible that the secrets will need to be passed into the PyPi upload workflow, as it is being invoked using workflow call
- [x] The windows installer is not currently being signed, that pipeline needs a release flag added and code signing functionality included

Is there anything else I've missed?

----

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated


.github/workflows/pypi_upload.yml
```diff
@@ -0,0 +1,58 @@
...
+
+on:
+  workflow_call:
+    inputs:
+      whl-file-name:
+        required: true
+        type: string
+      tar-file-name:
+        required: true
+        type: string
```
nucleogenesis, line 11: Is this something you manually enter when triggering the build?

nucleogenesis (commented):
Nice to have the JS in its own file :+1:  

You mention needing to put secrets on the repo - curious how that'll work. Is that a Github feature in the repo settings or something that'd need to be committed?

.github/workflows/pypi_upload.yml
rtibbles: The `workflow_call` event allows workflows to be called from other workflows, so these inputs are specified there.

If this had a `workflow_dispatch` event, that would allow manually triggering the workflow, in which case inputs could be manually entered - although in the places where I have done that, it has been to specify a URL where the input asset can be downloaded from.

.github/workflows/release_kolibri.yml
```diff
@@ -0,0 +1,112 @@
...
+      ref: master
+  test_pypi_upload:
+    name: Upload to TestPyPi
+    needs: whl
+    uses: ./.github/workflows/pypi_upload.yml
+    with:
+      whl-file-name: ${{ needs.whl.outputs.whl-file-name }}
+      tar-file-name: ${{ needs.whl.outputs.tar-file-name }}
+      test: true
+  deployment_block:
```
marcellamaki: A bit nit picky, but maybe this should be renamed? At first I read this as "the block of code that will deploy". Then when I read the description, it seems like it is _blocking_ deployment. But then the steps is "Release now publishing!"... so, I'm not 100% sure what exactly this should be handling and if the feedback in the logs will be clear.

rtibbles: Ah, yes - this is meant to be `the job that blocks deployment of the release` - thus requiring further manual intervention to make the release go.

The intention here is that we can test all the assets for the release, and then manually say "Good to go", and let the rest happen.

The reason that the only step is `Release now publishing!` is that this step will not run until it has been approved, so that message would only appear in the logs once we were going forward with the release.

rtibbles:
> Is that a Github feature in the repo settings or something that'd need to be committed?

It's a Github feature - we have organization level secrets that we can let certain repositories access. This means that we don't have to commit them to the repo (++good), and that their values are explicitly hidden in the Github Actions logs.

DXCanas:
> -  PyPi token
> -  Test PyPi token

Given that we're doing something similar for other packages, seems like we should bubble these up to the org and use them here too?

> a GPG signing private key and passphrase

Jose went through the trouble of demoing how to sign debs with me and I never used that training after the first and only time I made a release over a year ago. He would probably be best suited to consult here?

> Secrets for GCS Bucket and credentials

Is this for `le-downloads`/the download page on the site? Or for the BCK uploads? Either way, I recently went through the automation process here and can definitely assist here; would most likely be a service acct JSON blob. Would be hard to imagine there's not action for this already.

> The windows installer is not currently being signed, that pipeline needs a release flag added and code signing functionality included

After getting a little more familiar with this recently, I don't think it'd be a huge challenge. All the docs online insist on adding the creds to the sort of "keychain" concept that Windows encrypts and stores away in fancy hardware. But the `signtool.exe` CLI takes creds as an argument, making it a lot easier.

rtibbles:
> Jose went through the trouble of demoing how to sign debs with me and I never used that training after the first and only time I made a release over a year ago

I know how to create the keys, I just didn't think creating them on my personal laptop was the best course? Assumed it might be better to have the private key and passphrase created in some institutionally secure way and stored in our internal secrets.

> Is this for le-downloads/the download page on the site?

Yes, this is for uploading to the `le-downloads` bucket. I am using a Google Cloud Platform provided action to setup the credentials, so I just need the appropriate credentials for that!

marcellamaki (approved):
Looks good - I don't think we are missing any steps. Let's merge and try it 🤞 for our upcoming alpha

PR #9335, opened 2022-04-19
Refactor topics page

### LINKED ISSUE #8979: Refactor topic page using a reusable/props-driven component for main card/content grid(s)
Apply refactor from Library page to topics page. **Updating the new component unit tests if needed is part of this issue being done**

nucleogenesis:
## Summary

**This is based on #9230 - this is why there is a long commit history. This should be merged after that PR is.**

- Move header content into HeaderTab and HeaderTabs and Header components (similar to how it is done in Coach)
- Create a MobileHeader component for the mobile use case
- Move the Search/Folders side panel logic into components - one for the mobile use case and one for larger screens
- Create TopicRow component to handle rendering each row of topic headings + cards + "show more" links to simplify the TopicsPage component
- Move the SearchResultsGrid out of the views/LibraryPage dir into the views dir & update import references to it in the LibraryPage components and tests
- Use the SearchResultsGrid to render search results in TopicsPage

## References

Fixes #8979 

## Reviewer guidance

Do all the things  on the TopicsPage (viewing a channel)

- Search at the top level
- Search from within a subsection of the topic
- Folders navigation
- Show more
- Informational side panel
- Search / categories on mobile and desktop sizes


pcenov:
Hi @nucleogenesis I was asked by @radinamatic to test this PR which I did using the following build: https://buildkite.com/learningequality/kolibri-debian/builds/4112#e9493169-aedb-4897-a2d6-e232233f2fac
Tested in all supported browsers and in mobile view.

Here are the issues I was able to identify and I'm listing everything just to be on the safe side:

| Issue| Screenshot |
| Library Page - Missing 'Recent' section and errors in the console |  |
| Topics Page - The options 'View as list' and 'View as grid' are not working |  |
| Topics Page - Opening the info panel squishes the card (especially noticeable when there is only one) |  |
| Topics Page - Clicking the View resource button from the Info panel results in showing the resource in small size portion of the screen | https://user-images.githubusercontent.com/79847249/164241012-b161bad7-8a46-4221-a391-a375899ed9a9.mp4 |

nucleogenesis:
Thank you @pcenov - I've addressed all noted issues.

One thing I noticed is that the Card / List selection isn't currently implemented in the search results before these changes - but rather this seems to be a consequence of my work in #9230 - is there any reason why this option to switch views should be hidden in the search results of the topic page @jtamiace or is keeping this okay?

pcenov:
Hi @nucleogenesis, I've verified that the previously reported issues are now fixed and in the process of doing so I was able to identify some new issues:

| Issue| Screenshot |
| Topics Page - Clicking the folder breadcrumb results in a "Resource not found error" |  |
| Topics Page - Clicking 'View all' results in console errors and clicking it again does nothing at all | https://user-images.githubusercontent.com/79847249/165321733-671834c9-5923-4c34-9b4c-b3bbdfc6b953.mp4 |
| Library/Topics Page - A bottom scroll is shown persistently |  |

jtamiace:
@nucleogenesis can't think of any downsides to hide the list/card toggle 👍🏼

kolibri/plugins/learn/assets/src/views/CopiesModal.vue
```diff
@@ -53,6 +53,7 @@
         required: true,
       },
     },
+    computed: {},
```
marcellamaki: I am not sure what this is for?

kolibri/plugins/learn/assets/src/views/TopicsPage/FloatingSidePanel.vue
```diff
@@ -0,0 +1,120 @@
...
+
+<script>
+
+  import FullScreenSidePanel from 'kolibri.coreVue.components.FullScreenSidePanel';
+  import genContentLink from '../../utils/genContentLink';
+  import EmbeddedSidePanel from '../EmbeddedSidePanel';
+
+  export default {
+    name: 'FloatingSidePanel',
+    components: { EmbeddedSidePanel, FullScreenSidePanel },
```
marcellamaki: I say this with **full acknowledgement** I have been a large contributor to the number of SidePanels 😄 but ... I'm having a bit of trouble wrapping my mind around what exactly makes this component different/necessary? Is it just to break things out of the main page into more modular pieces? I feel very comfortable with this part of the codebase in general but having what is essentially 3 layers of side panels feels a bit confusing.

 Do you think it would be worth it to have a follow up issue for refactoring some of this, similar to the issue around the card grid to simplify it? I'm not sure if you think that's a good use of time or if there are even improvements to be made, given you've just spend quite a bit of time here.

kolibri/plugins/learn/assets/src/views/TopicsPage/index.vue
```diff
@@ -7,109 +7,15 @@
...
-        <KGridItem
-          :layout4="{ span: 1 }"
-        >
-          <img
-            :src="topic.thumbnail"
-            class="channel-logo"
-          >
-        </KGridItem>
-      </KGrid>
+      <MobileHeader v-else :topic="topic" />
```
marcellamaki: I like the split into a MobileHeader here -- great idea!

kolibri/plugins/learn/assets/src/views/TopicsPage/index.vue
```diff
@@ -141,58 +47,23 @@
...
-              <!-- card grid of items in folder -->
-              <LibraryAndChannelBrowserMainContent
-                v-if="t.children && t.children.length"
-                data-test="children-cards-grid"
-                :contents="t.children"
-                currentCardViewStyle="card"
-                :gridType="2"
+            <!-- Rows of cards and links / show more for each Topic -->
+            <template v-for="t in topicsForDisplay">
+              <TopicRow
```
marcellamaki: I like the TopicRow too!

kolibri/plugins/learn/assets/src/views/TopicsPage/index.vue
```diff
@@ -325,11 +190,11 @@
         />
       </FullScreenSidePanel>
       <CategorySearchModal
-        v-if="(windowIsMedium || windowIsLarge) && currentCategory"
+        v-if="currentCategory"
         :selectedCategory="currentCategory"
         :numCols="numCols"
         :availableLabels="labels"
-        position="modal"
+        position="windowIsSmall ? 'fullscreen' : 'modal'"
```
marcellamaki: This seems to have originated from code that I added (and for the life of me I cannot remember what I was intending....), but now I don't think we actual reference the position within this component? Or use this value anywhere? It seems like the layout of the modal is handled within `CategorySearchModal` by using `windowIsLarge`. 

`numCols` also seems obsolete (although I don't think that's your problem, I think that's a me problem 😂)

marcellamaki (commented):
Overall, the code read through looks pretty good to me! I like how you have split things up into more manageable components, and I found most of it quite easy to read. I'm going to do some additional manual QA to ensure I fully understand how the pieces fit together, once I can sort out the dependencies issues I'm having and will add more comments then.

kolibri/plugins/learn/assets/src/composables/useCopies.js
```diff
@@ -0,0 +1,21 @@
+import { set } from '@vueuse/core';
+import { reactive } from 'kolibri.lib.vueCompositionApi';
+
+// Shared store
+var displayedCopies = reactive([]);
+
+export default function useCopies() {
```
rtibbles: The copies composable that we nixed in the previous PR seems to have crept back in here?

kolibri/plugins/learn/assets/src/views/CopiesModal.vue
```diff
@@ -39,21 +40,19 @@
 <script>
 
   import commonCoreStrings from 'kolibri.coreVue.mixins.commonCoreStrings';
+  import genContentLink from '../utils/genContentLink';
+  import useCopies from '../composables/useCopies';
 
   export default {
     name: 'CopiesModal',
     mixins: [commonCoreStrings],
-    props: {
```
rtibbles, line 46: Let's revert this back to whatever was merged with the Library page refactor PR. Might need to drop a commit or two in the rebase?

kolibri/plugins/learn/assets/src/views/TopicsPage/FloatingSidePanel.vue
rtibbles: Also, the names are bloody confusing. 'FullScreenSidePanel' feels like it is in some way taking up the whole screen, when it isn't - embedded side panel is embedded inside the full screen side panel, and now we are nesting both inside a floating side panel?

image

kolibri/plugins/learn/assets/src/views/TopicsPage/HeaderTab.vue
```diff
@@ -0,0 +1,118 @@
...
+  };
+
+</script>
+
+
+<style lang="scss" scoped>
+
+  @import '~kolibri-design-system/lib/styles/definitions';
+
+  // a lot copied from KButton
```
rtibbles: I was wondering where all the extra styles were coming from here.

At the risk of dog piling - is it worth checking with @jtamiace who I know had a couple of regrets about how these header links were done, and I think slightly redesigned them for 0.16. Maybe we could just do that directly here instead?

kolibri/plugins/learn/assets/src/views/TopicsPage/HeaderTabs.vue
```diff
@@ -0,0 +1,42 @@
+<template>
+
+  <!-- Stolen from Coach -->
+  <div
```
rtibbles: Does this component get reused at all? It feels like too much breaking out if not.

kolibri/plugins/learn/assets/src/views/TopicsPage/TopicRow.vue
```diff
@@ -0,0 +1,115 @@
+<template>
```
rtibbles: I can't comment on it because it's empty ^^ but why the empty file?

kolibri/plugins/learn/assets/src/views/TopicsPage/TopicRow.vue
```diff
@@ -0,0 +1,115 @@
...
+
+
+<script>
+
+  import commonCoreStrings from 'kolibri.coreVue.mixins.commonCoreStrings';
+  import genContentLink from '../../utils/genContentLink';
+  import LibraryAndChannelBrowserMainContent from '../LibraryAndChannelBrowserMainContent';
+
+  export default {
+    name: 'TopicRow',
```
rtibbles: Is this really just a row, or a topic subsection?

kolibri/plugins/learn/assets/src/views/TopicsPage/index.vue
rtibbles: As I noted above, I think it's the wrong name though.

rtibbles (commented):
A couple of comments, might need another look.

kolibri/plugins/learn/assets/src/composables/useCopies.js
nucleogenesis: Bah - rebasing is a pain I'm sure that's where it came from.

kolibri/plugins/learn/assets/src/views/TopicsPage/FloatingSidePanel.vue
marcellamaki: I am the one responsible for the fullscreen vs embedded situation... 🙃  I'd thought previously about renaming fullscreen as 'overlaid' or something similar and then never took any action because I wasn't sure how much better it was

nucleogenesis: I'm definitely up for a group chat to figure out what to call these things :smile:

kolibri/plugins/learn/assets/src/views/TopicsPage/HeaderTabs.vue
nucleogenesis: Nah it's just the structure that came with it when I stole it from Coach - seems more reasonable to make something specific to the TopicsPage anyway here and would go hand-in-hand with the work to move them per @jtamiace 's latest TopicsPage updated designs

kolibri/plugins/learn/assets/src/views/TopicsPage/HeaderTab.vue
nucleogenesis: If @jtamiace signs off on implementing the changes she shared during the content sync review (for context, these are the Folders & Search tabs) - then it wouldn't be too much work to add that change here while cleaning up the header + tabs code a bit more

kolibri/plugins/learn/assets/src/views/TopicsPage/TopicRow.vue
nucleogenesis: :man_shrugging:

kolibri/plugins/learn/assets/src/views/TopicsPage/TopicRow.vue
nucleogenesis: It's a complicated row :smile: - subsection makes sense to me

kolibri/plugins/learn/assets/src/views/TopicsPage/TopicRow.vue
nucleogenesis: You commented later on in this file so I'm wondering if you hit a bug in Github / VS Code or something - or maybe I'm missing something

rtibbles: My comment was targeted at the `kolibri/plugins/learn/assets/src/views/TopicsPage/MobileSidePanel.vue` which still appears to be empty?

kolibri/plugins/learn/assets/src/views/TopicsPage/FloatingSidePanel.vue
nucleogenesis: I could rename them here to something like (open to suggestions)

**FullScreenSidePanel -> SidePanelModal | DrawerModal | PopOutModal**

- Modal seems to imply the "it pops up, steals focus, lets you do a thing, and you can dismiss it"
- PopOut and Drawer to me both say "I come out from the side and go away when you're done with me".

**EmbeddedSidePanel -> SearchForm | SearchControlPanel**

- The EmbeddedSidePanel is basically the helm for search controls so something around that functionality seems to be a good way forward.

**FloatingSidePanel -> SearchControlPanelModal | SearchPopOutModal**

- The FloatingSidePanel is basically just wrapping EmbeddedSidePanel with FullScreenSidePanel... so its name might be best decided after the other two are :laughing:

kolibri/plugins/learn/assets/src/views/TopicsPage/TopicRow.vue
nucleogenesis: Ah - weird - GH says your comment was on TopicsPage/TopicRow.vue - thanks for clarifying, will remove the empty.###

kolibri/plugins/learn/assets/src/views/TopicsPage/HeaderTab.vue
nucleogenesis: I think I'll hold off for now and wait for the redesigns to come through as the header tab designs may be further impacted by design decisions and they're not finalized yet.

nucleogenesis:
@rtibbles @marcellamaki 

I've made some changes to address feedback. 

- I went ahead and renamed the SidePanels as follows - happy to apply any suggested changes pre-merge:

FullScreenSidePanel -> SidePanelModal
EmbeddedSidePanel -> SearchFiltersPanel
FloatingSidePanel -> SearchPanelModal

- I removed the copies business I managed to reintroduce during rebasing
- Changed TopicRow -> TopicSubsection
- Removed empty file

kolibri/plugins/learn/assets/src/views/TopicsPage/index.vue
```diff
@@ -141,58 +48,23 @@
...
-                data-test="children-cards-grid"
-                :contents="t.children"
-                currentCardViewStyle="card"
-                :gridType="2"
+            <!-- Rows of cards and links / show more for each Topic -->
+            <template v-for="t in topicsForDisplay">
+              <TopicSubsection
+                :key="t.id"
+                :topic="t"
+                :subTopicLoading="subTopicLoading"
```
rtibbles: Maybe we could change `subTopicLoading` prop on `TopicSubsection` to be a boolean, and then do the equality check here?
```suggestion
                :subTopicLoading="t.id === subTopicLoading"
```

That way the TopicSubsection component doesn't need to worry about things that are outside its purview.

kolibri/plugins/learn/assets/src/views/LibraryAndChannelBrowserMainContent.vue
```diff
@@ -32,7 +32,7 @@
     </CardGrid>
     <!-- large displays, list view -->
     <HybridLearningContentCardListView
-      v-for="content in contents"
+      v-for="(content, idx) in contents"
```
rtibbles, line 35: This file does rather seem like it's acting as a responsive Card grid with a loading state? Could we do a rename to something that more resembles that?

rtibbles (commented):
I might have some more nitpick comments in me, but I'm not finding them just yet. Code wise this is looking solid, but I have not manually tested.

radinamatic:
Just started the testing and noticed 2 issues on browsing the Library (replicable in both Firefox and Chrome on Windows 10):

1. On first load of the Library home page there is an unnecessary horizontal scrollbar in the browser window.
2. Does not happen when opening the 'View information' side panel from a resource on the Library home page, but when the user is browsing a channel, and opens the side panel for a resource, it seems impossible to close it: neither clicking outside nor on the ``X`` close button seem to have effect. The only remaining thing seems to be to hit the browser back button, and then one lands on the Library home page (or the parent topic) which cannot be scrolled vertically until the user reloads the page.


https://user-images.githubusercontent.com/1457929/167141154-408e9a76-37fe-4303-aaf0-a78861355bf3.mp4

pcenov:
@radinamatic - I'm also able to replicate the above mentioned 2 issues plus found the following additional regressions:

| Issue| Screenshot |
| Topics Page - "Folders" and "Search" tabs get disabled after interacting with a resource and clicking the Back arrow | https://user-images.githubusercontent.com/79847249/167422815-16977f61-4fda-4df2-97ff-ab60954536ab.mp4|
| Topics Page - incorrectly displayed video resource in ultra wide resolution - the header is hidden and the video is stretched too much  | https://user-images.githubusercontent.com/79847249/167425327-46818371-e2f4-4ad1-b32a-99504aa6d0dc.mp4 |
| Mobile View - Topics Page - squished cards, incorrectly displayed 3 instead of max 2 cards | https://user-images.githubusercontent.com/79847249/167427143-77687722-36a7-412f-814c-31d60a1f5502.mp4 |

nucleogenesis:
@radinamatic @pcenov 

Thank you for the reviews here.

I think that some of these issues will be fixed - or at least _should be addressed_ after merging this and #9137 - I think that from there, some of the above mentioned issues should be addressed anew as we risk complicating this code altogether quite a bit more than if we just fix any issues that come out in the wash after the fact.

If the tests and checks pass, then @marcellamaki can merge her PR and we can see where things stand w/ the Library & Topics pages and hammer out anything that remains.

rtibbles:
I think this seems sensible, we'll see how things stand after both the PRs are merged.

radinamatic:
@nucleogenesis Given the conversation today around the assets for ``develop`` not being available, after the PR is merged we'll have to test this from the source, correct?

kolibri/core/assets/src/core-app/apiSpec.js
```diff
@@ -157,7 +157,7 @@ export default {
       PermissionsIcon,
       AppBarCorePage,
       CoreBase,
-      FullScreenSidePanel,
+      SidePanelModal,
```
rtibbles, line 160: I probably asked this at the time, but why do we have this in the core API? I thought it was only used in learn?

rtibbles (approved):

nucleogenesis:
@radinamatic I think maybe there was some confusion - where I understood that conversation landing was that we're basically accepting that some parts of `develop` may be a bit busted for a couple weeks at a time - particularly Tasks and Onboarding at this point. 

You'll be able to test this and any follow-up PRs with the assets as normal.

The case in which assets would not be built is if I had done this work in a feature branch (ie, not merging this or the onboarding work into `develop` but instead into its own branch).

So in the end `develop` is always liable to be a bit borked for the next few weeks. In this PRs particular case, we will follow up quickly to hammer out any issues that make it through the merges of my and @marcellamaki 's PR

rtibbles:
@nucleogenesis I think the issue is that without the setup wizard working, and tasks working, testing from assets in a fresh VM doesn't work, because you can't initialize Kolibri to test it.

So, there's a need to test from source, because that way it can be tested with an existing KOLIBRI_HOME dir. @radinamatic feel free to correct me if I am mistaken.

radinamatic:
Apologies, I was not clear enough: assets are available, but the onboarding does not work - so no setup wizard... 🤔

radinamatic:
What Richard said, right!

nucleogenesis:
Ah - I was unaware that each QA session works in a freshly generated VM. So in that case, you're correct Radina. Sorry for the misunderstanding @radinamatic

kolibri/core/assets/src/core-app/apiSpec.js
nucleogenesis: What a great question... will create an issue for this to follow-up on

radinamatic:
We could copy an up-to-date ``./kolibri`` folder from the development setup and install the 0.16 asset on top, that might work 🤔

PR #9339, opened 2022-04-19
Fixes incorrect run method for full Kolibri Process Bus.

rtibbles:
## Summary
Fixes regression from #9291
Accidentally inherited the run method from the services bus, which includes an additional publish of the SERVING event - which will instead happen from the Server plugin.

## Reviewer guidance
When the main port was set to 0, this might have caused issues.
Compare with previous code: https://github.com/learningequality/kolibri/blob/release-v0.15.x/kolibri/utils/server.py#L701 where the publish was only done when HTTP is not being served.

----

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/utils/server.py
+++ b/kolibri/utils/server.py
@@ -720,6 +720,10 @@ class KolibriProcessBus(KolibriServicesProcessBus):
         kolibri_server.subscribe()
         alt_port_server.subscribe()
 
+    def run(self):
+        self.graceful()
+        self.block()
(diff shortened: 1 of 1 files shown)


jredrejo (approved):
cloning what was working can't be harmful. Honestly I don't know Magicbus enough to know why it needs this two steps, but having them back makes sense if they were there before

PR #9341, opened 2022-04-20
Add API endpoint for UI initiated device restart

### LINKED ISSUE #9293: Create API endpoint to initiate a restart of the Kolibri server
This endpoint should only handle two specific HTTP verbs:
`GET`:
* In this case, the API endpoint should return either 'restarting' or 'running' - this will be based on whether a restart command has previously been issued, and whether the server has since restarted

`POST`:
* In this case, the API request requires no body, and will set the server as 'restarting' - as it is tricky to cancel a restart, this will not be supported.

The API endpoint should require SuperAdmin permissions.

The API endpoint will call the `restart` function in `server.py` to initiate the restart. A new function will need to be created in there that will return the status for running or restarting - but this should be read from the PID file.

Caveat - these methods may be considerably less reliable or not work at all in a kolibri-server or BCK implementation.

rtibbles:
## Summary
* Adds an API endpoint to call the 'restart' function from `server.py`
* Modifies the `restart` function to only emit the signal, and adds a `restart_and_wait` function that is used by the CLI command
* Adds a frontend composable to handle the restart process, exposing a `restart` method and a `restarting` Boolean state
* Caveat: This will not work for kolibri-server or BCK. We need to define some way of disabling this behaviour or making it work for uwsgi (there is a uwsgi.restart() method that we could use).
* Adds a `RESTART_HOOKS` option to allow completely disabling restart behaviour, or adding additional restart behaviour for, say, uwsgi deployment scenarios.

## References
Fixes #9293

## Reviewer guidance
To test this I added this diff to `kolibri/plugins/device/assets/src/views/DeviceSettingsPage/index.vue`
```
@@ -3,10 +3,13 @@
   <div>
     <section>
       <h1>
         {{ $tr('pageHeader') }}
       </h1>
+      <KButton :disabled="restarting" @click="restart">
+        RESTART
+      </KButton>
       <p>
         {{ $tr('pageDescription') }}
         <KExternalLink
           v-if="!isMultiFacilitySuperuser && getFacilitySettingsPath()"
           :text="$tr('facilitySettings')"
@@ -127,10 +130,11 @@
   import sortLanguages from 'kolibri.utils.sortLanguages';
   import { LandingPageChoices } from '../../constants';
+  import useDeviceRestart from '../../composables/useDeviceRestart';
   import { getDeviceSettings, saveDeviceSettings } from './api';
 
   const SignInPageOptions = Object.freeze({
     LOCKED_CONTENT: 'LOCKED_CONTENT',
     DISALLOW_GUEST_ACCESS: 'DISALLOW_GUEST_ACCESS',
@@ -143,10 +147,17 @@
       return {
         title: this.$tr('pageHeader'),
       };
     },
     mixins: [commonCoreStrings],
+    setup() {
+      const { restarting, restart } = useDeviceRestart();
+      return {
+        restarting,
+        restart,
+      };
+    },
     data() {
       return {
         language: {},
         landingPage: '',
         allowPeerUnlistedChannelImport: null,
```

This produced this result:

https://user-images.githubusercontent.com/1680573/164297491-e36801df-f435-4bb3-8d18-bfdeb8d25582.mp4

----

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/core/assets/src/core-app/baseClient.js
+++ b/kolibri/core/assets/src/core-app/baseClient.js
@@ -19,6 +19,9 @@ export default function clientFactory(options) {
   client.interceptors.response.use(
     response => response,
     function(error) {
+      if (!error) {
+        error = {};
+      }
       if (!error.response) {
         error.response = {
           status: 0,
--- a/kolibri/utils/cli.py
+++ b/kolibri/utils/cli.py
@@ -354,7 +354,7 @@ def restart():
     """
     Restarts the server if it is running
     """
-    if server.restart():
+    if server.restart_and_wait():
         logger.info("Kolibri has successfully restarted")
         sys.exit(0)
     logger.info("Kolibri has failed to restart - confirm that the server is running")
--- a/kolibri/utils/options.py
+++ b/kolibri/utils/options.py
@@ -694,7 +694,7 @@ def _set_from_envvars(conf):
     for section, opts in option_spec.items():
         for optname, attrs in opts.items():
             for envvar in attrs.get("envvars", []):
-                if os.environ.get(envvar):
+                if envvar in os.environ:
                     deprecated_envvars = attrs.get("deprecated_envvars", ())
                     if envvar in deprecated_envvars:
                         logger.warn(
--- a/kolibri/utils/options.py
+++ b/kolibri/utils/options.py
@@ -112,6 +112,27 @@ def _process_language_string(value):
     raise ValueError
 
 
+def _process_list(value, separator=","):
+    """
+    Used to validate list values.
+    The only valid argument in this case is that it is a list
+    so we first try to coerce it to a list, then do some checks
+    to see if it is any of our special values. Then if it is an
+    appropriate list value.
+    If no value is appropriate, raise a ValueError.
+    """
+
+    # Check the supplied value is a list
+    if not isinstance(value, list):
+        if not value:
+            value = []
+        elif isinstance(value, string_types):
+            value = value.split(separator)
+        else:
+            value = [value]
+    return value
+
+
 def language_list(value):
     """
     Check that the supplied value is a list of languages,
@@ -127,9 +148,7 @@ def language_list(value):
     or one of the special strings represented by ALL_LANGUAGES or SUPPORTED_LANGUAGES
     A list must be a list of these strings.
     """
-    # Check the supplied value is a list
-    if not isinstance(value, list):
-        value = [value]
+    value = _process_list(value)
 
     out = set()
     errors = []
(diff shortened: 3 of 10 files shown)


kolibri/utils/cli.py
```diff
@@ -354,7 +354,7 @@ def restart():
     """
     Restarts the server if it is running
     """
-    if server.restart():
+    if server.restart_and_wait():
```
marcellamaki, line 357: What is it that has changed that makes the "`_and_wait`" and important addition? (This is just for my own understanding, not a rhetorical to suggest it is unnecessary)

marcellamaki (commented):
QA of this checks out, both in terms of the UI interaction, and what is happening in the terminal/network tab checks out. I think @jredrejo might have better insight into any potential problems than I do, but I found the code relatively readable, even with my limited python knowledge.

kolibri/utils/cli.py
rtibbles: The `restart` function now just signals that the server should restart - previously it was doing `restart_and_wait` but it was useful to be able to _just_ signal that the server should restart (via the API call) rather than wait for the restart to happen - which always meant that the request failed to respond, because the request thread gets killed while the request is being served.

kolibri/utils/options.py
```diff
@@ -246,6 +267,71 @@ def multiprocess_bool(value):
         return False
 
 
+class LazyImportFunction(object):
```
jredrejo, line 271: isn't this the same purpose as the builtin `importlib.util.LazyLoader` class? I know this is not available in 2.7 but just wondering if we could hijack its code for 2.7 and use the original one  for > 3.5 https://github.com/python/cpython/blob/main/Lib/importlib/util.py#L268

If it's not adecuate, I think we should have at least some test case for this as it's a quite sofisticated piece of code (first time I see a real case use for AST out of parsers world)

kolibri/core/assets/src/core-app/baseClient.js
```diff
@@ -19,6 +19,9 @@ export default function clientFactory(options) {
   client.interceptors.response.use(
     response => response,
     function(error) {
+      if (!error) {
+        error = {};
```
jredrejo, line 23: don't we have any case in our code using `error === undefined`? if we have it this can cause new problems

jredrejo (approved):
I have a couple of questions to solve some doubts, but overall the code is good and the logic seems correct to me (not to mention it works when testing it ;) )

kolibri/utils/options.py
rtibbles: Yes - it does seem better to vendor this, I was just totally unaware of it.

kolibri/core/assets/src/core-app/baseClient.js
rtibbles: If we do, it's currently dead code, because this interceptor will error out before it even gets there.

PR #9343, opened 2022-04-21
Add REST API for enabling and disabling plugins

### LINKED ISSUE #9290: Create API endpoint for enabling and disabling plugins
In order to allow for plugins to be fully disabled, we need to allow the frontend to be able to initiate requests that disable plugins.

The first use case for this is to disable the Facility and Coach plugins during the Quick Start onboarding flow - as this targets a single, standalone user, the need for the Facility and Coach plugins is limited.

This same endpoint could then be reused for later enabling the plugins.

In considering the technical implementation for this API endpoint, there are some specific technical considerations that need to be handled:

1. This functionality requires restarting/reloading the Kolibri stack, as the Kolibri Django app, Options processing etc. are dependent on the enabled plugins. This means that it is not as simple as sending RESTful requests to an endpoint, there must also be an interplay between the client and the server to indicate when the restart should be initiated, and when it has completed.
2. As this requires restarting/reloading Kolibri, specific allowances will have to be made for Kolibri running in alternative architectures (e.g. the kolibri-server package which runs Kolibri with uwsgi, or BCK that runs it on a cloud based architecture).

The simplest mechanism therefore, might be an API endpoint for editing the plugins enabled and disabled - this could be a simple RESTful endpoint that will return a list of plugins and their enabled/disabled states, and then a detail endpoint that accepts PATCHes that update whether that plugin is enabled or disabled.

The list endpoint should return objects of the following type:

```
/**
 * Type definition for Plugin metadata
 * @typedef {Object} Plugin
 * @property {string} id - a module path specific to the plugin
 * @property {boolean} enabled - whether the plugin is enabled
 */
```

This API endpoint should be implemented in the `kolibri.core.device` app.

Another API endpoint for handling the restart would be implemented separately.

rtibbles:
## Summary
* Cleans up cruft and adds capability to the plugin object
* Adds name method to allow internationalization of plugin names
* Uses this in the CLI
* Adds language prefixed REST API endpoint for plugins
* Adds composable for managing plugin state

## References
Fixes #9290 

## Reviewer guidance
I tested this by adding this diff to `kolibri/plugins/device/assets/src/views/DeviceSettingsPage/index.vue`:
```
@@ -14,10 +14,19 @@
         />
       </p>
     </section>
 
     <section>
+      <ul v-if="plugins">
+        <li v-for="plugin in plugins" :key="plugin.id">
+          <span>{{ plugin.name }}</span>
+          <KButton
+            :text="plugin.enabled ? 'Disable' : 'Enable'"
+            @click="plugin.enabled ? disablePlugin(plugin.id) : enablePlugin(plugin.id)"
+          />
+        </li>
+      </ul>
       <div class="fieldset">
         <KSelect
           v-model="language"
           :label="$tr('selectedLanguageLabel')"
           :options="languageOptions"
@@ -127,10 +136,11 @@
   import sortLanguages from 'kolibri.utils.sortLanguages';
   import { LandingPageChoices } from '../../constants';
+  import usePlugins from '../../composables/usePlugins';
   import { getDeviceSettings, saveDeviceSettings } from './api';
 
   const SignInPageOptions = Object.freeze({
     LOCKED_CONTENT: 'LOCKED_CONTENT',
     DISALLOW_GUEST_ACCESS: 'DISALLOW_GUEST_ACCESS',
@@ -143,10 +153,18 @@
       return {
         title: this.$tr('pageHeader'),
       };
     },
     mixins: [commonCoreStrings],
+    setup() {
+      const { plugins, enablePlugin, disablePlugin } = usePlugins();
+      return {
+        plugins,
+        enablePlugin,
+        disablePlugin,
+      };
+    },
     data() {
       return {
         language: {},
         landingPage: '',
         allowPeerUnlistedChannelImport: null,
```

https://user-images.githubusercontent.com/1680573/164490139-cbaca380-ea2e-401a-8bdb-37e626bc12c1.mp4

----

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/plugins/__init__.py
+++ b/kolibri/plugins/__init__.py
@@ -201,6 +201,14 @@ class KolibriPluginBase(with_metaclass(SingletonMeta)):
     # }
     kolibri_option_defaults = None
 
+    #: Comment
+    # Property that indicates whether it is safe to allow this plugin to be disabled via the GUI
+    # This is used to prevent plugins that are required by the system from being disabled except
+    # through the command line. This property is purely advisory, and does not affect the
+    # functionality of the plugin, or the ability to disable it via the command line or other
+    # configuration mechanisms.
+    can_manage_while_running = False
+
     def __init__(self):
         self.INSTALLED_APPS = []
 
--- a/kolibri/plugins/__init__.py
+++ b/kolibri/plugins/__init__.py
@@ -201,15 +201,6 @@ class KolibriPluginBase(with_metaclass(SingletonMeta)):
     # }
     kolibri_option_defaults = None
 
-    # : Suggested property, not yet in use
-    migrate_on_enable = False
-
-    # : Suggested property, not yet in use
-    collect_static_on_enable = False
-
-    # : Suggested property, not yet in use
-    collect_static_on_enable = False
-
     def __init__(self):
         self.INSTALLED_APPS = []
 
@@ -220,6 +211,13 @@ class KolibriPluginBase(with_metaclass(SingletonMeta)):
     @property
     def module_path(self):
         return self.class_module_path()
+
+    def name(self, lang):
+        """
+        A function to return a name for the plugin.
+        Accepts a lang argument to allow for internationalization of the name.
+        """
+        return self.__class__.__name__
 
     def _installed_apps_add(self):
         """Call this from your enable() method to have the plugin automatically
@@ -234,6 +232,10 @@ class KolibriPluginBase(with_metaclass(SingletonMeta)):
     def enable(self):
         """Modify the kolibri config dict to your plugin's needs"""
         self._installed_apps_add()
+
+    @property
+    def enabled(self):
+        return self.module_path in config.ACTIVE_PLUGINS
 
     def disable(self):
         """Modify the kolibri config dict to your plugin's needs"""
--- a/kolibri/plugins/device/assets/src/composables/usePlugins.js
+++ b/kolibri/plugins/device/assets/src/composables/usePlugins.js
@@ -1,0 +1,49 @@
+/**
+ * A composable function containing logic related to channels
+ */
+
+import { ref } from 'kolibri.lib.vueCompositionApi';
+import client from 'kolibri.client';
+import urls from 'kolibri.urls';
+
+export default function usePlugins() {
+  const plugins = ref(null);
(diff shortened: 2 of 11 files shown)


kolibri/plugins/__init__.py
```diff
@@ -221,6 +220,13 @@ def class_module_path(self):
     def module_path(self):
         return self.class_module_path()
 
+    def name(self, lang):
+        """
+        A function to return a name for the plugin.
+        Accepts a lang argument to allow for internationalization of the name.
```
nucleogenesis, line 226: Is the i18n baked in via something not clear here - or just not yet implemented?

kolibri/plugins/device/assets/src/composables/usePlugins.js
```diff
@@ -0,0 +1,50 @@
...
+    }).then(response => {
+      plugins.value = response.data;
+    });
+  }
+  fetchPlugins();
+  function togglePlugin(pluginId, value) {
+    const pluginIndex = plugins.value.findIndex(plugin => plugin.id === pluginId);
+    if (pluginIndex !== -1) {
+      const plugin = plugins.value[pluginIndex];
+      console.log(urls['kolibri:core:plugins-detail'](pluginId));
```
nucleogenesis: Was this intentionally committed?

nucleogenesis (approved):
Code LGTM overall.

kolibri/plugins/__init__.py
rtibbles: In specific plugins overriding this method, they would use the lang parameter to activate translation then use regular Django mechanisms to translate.

The reason for not enforcing anything more specific is that in the case of plugins external to Kolibri, they won't be able to use Django translation mechanisms, as the locale files will not be discoverable by Django until the plugin has been enabled, so they would need to use basic Python i18n machinery to offer a translated name.

kolibri/plugins/device/assets/src/composables/usePlugins.js
rtibbles: Nope...

kolibri/plugins/__init__.py
rtibbles: To make this clearer, I added in placeholder functions in all the relevant plugins.

Issue #9280, opened 2022-04-11
Concurrent import of several channels fails after restart of the server


pcenov:
## Observed behavior
This issue was first spotted by @radinamatic at the Bugbash when the concurrent import of 4 channels seemed to go on without issues, but then the VM restarted, and the import of those channels was supposed to restart automatically but the import of the original 4 channels remained stalled and never progressed further than where is was before the restart. 

I was able to replicate this issue on both Windows and Ubuntu after restarting the server while the import is still in progress.

## Expected behavior
The import of the channels should proceed after a restart of the server.

## Steps to reproduce the issue
1. Install the following [0.15.2a1](https://buildkite.com/learningequality/kolibri-python-package/builds/5227#job-c03da187-c9a1-483a-9d5f-e75b5c36f97d) build
2. Attempt to import several channels simultaneously
3. Stop the server while the import of some or all of the channels is still and progress
4. Start again the server and observe the progress of the import

## Screenshots
![2022-04-11_17-12-23](https://user-images.githubusercontent.com/79847249/162772664-c76f7e65-6406-4ab7-8a0e-22a05b7ebd7c.png)

## Ubuntu logs and DB
[Logs.zip](https://github.com/learningequality/kolibri/files/8465509/Logs.zip)

## Usage Details
Windows 10, Ubuntu - Chrome

rtibbles:
Note that this is not currently expected behaviour (in terms of the implementation that is currently in place), but I certainly agree that this behaviour is desirable.

PR #9352, opened 2022-04-25
Mark RUNNING jobs as failed on startup

### LINKED ISSUE #9351: Background running tasks never ends if the server is killed
## Observed behavior

When there's a running background task and the Kolibri server is killed by the operating system for some reason, the tasks will be in the RUNNING state forever.

imagen

## Expected behavior

The RUNNING tasks are marked as "FAILED" when the Kolibri server detects that there are "RUNNING" tasks from a previous Kolibri process.

## User-facing consequences

The user can cancel long running tasks from the interface, so this is not a severe issue. But there's no information in the interface for the user when this happens so it's possible to think that the task is really running in the background and it doesn't.

## Steps to reproduce

 * Start a dev kolibri process
 * Start a background task, for example download some "big" channel
 * Kill the dev kolibri process, make sure to kill completely, all threads
 * Relaunch again the kolibri process
 * You will find the task running and the progress is not updated

## Context

 * Kolibri 0.15.1
 * Detected in the Android version, also reproduced on Linux

danigm:
Fix https://github.com/learningequality/kolibri/issues/9351

### DIFF
--- a/kolibri/core/tasks/storage.py
+++ b/kolibri/core/tasks/storage.py
@@ -191,6 +191,17 @@ class Storage(StorageMixin):
 
             return [self._orm_to_job(job) for job in jobs]
 
+    def get_running_jobs(self, queues=None):
+        with self.session_scope() as s:
+            q = s.query(ORMJob).filter(ORMJob.state == State.RUNNING)
+
+            if queues:
+                q = q.filter(ORMJob.queue.in_(queues))
+
+            jobs = q.order_by(ORMJob.time_created).all()
+
+            return [self._orm_to_job(job) for job in jobs]
+
     def get_all_jobs(self, queue=None):
         with self.session_scope() as s:
             q = s.query(ORMJob)
@@ -279,6 +290,9 @@ class Storage(StorageMixin):
 
     def mark_job_as_running(self, job_id):
         self._update_job(job_id, State.RUNNING)
+
+    def mark_job_as_queued(self, job_id):
+        self._update_job(job_id, State.QUEUED)
 
     def complete_job(self, job_id, result=None):
         self._update_job(job_id, State.COMPLETED, result=result)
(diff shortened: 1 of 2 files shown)


rtibbles (changes requested):
Better if instead we do the behaviour described in: https://github.com/learningequality/kolibri/issues/9280

For best results, I think we should be cancelling tasks during shutdown, waiting for them to finish cancelling, and then setting them to `QUEUED` again, so that upon restart they automatically resume.

Also, I think this change is small enough that we can target it to 0.15.x.

danigm:
> Better if instead we do the behaviour described in: #9280
> 
> For best results, I think we should be cancelling tasks during shutdown, waiting for them to finish cancelling, and then setting them to `QUEUED` again, so that upon restart they automatically resume.
> 
> Also, I think this change is small enough that we can target it to 0.15.x.

As far as I know, right now there's code to "cancel" running jobs [during normal shutdown](https://github.com/learningequality/kolibri/blob/develop/kolibri/core/tasks/worker.py#L36). The problem that I can see is because the OS is killing other threads, or the main server process, and there's no a controlled shutdown, so jobs are not updated and they are forever `RUNNING`.

That's why I'm trying to detect that stalled jobs at startup time. I've updated this PR to "requeue" `RUNNING` jobs at startup, so jobs will continue running if something of this happens.

Thinking about possible problems, it's possible that the `RUNNING` jobs are really running in a different thread during startup? But as far as I know if this happens it's not possible to detect at startup.

rtibbles:
> As far as I know, right now there's code to "cancel" running jobs [during normal shutdown](https://github.com/learningequality/kolibri/blob/develop/kolibri/core/tasks/worker.py#L36).

The issue is that we're not waiting for the cancellation to complete before issuing the shutdown to the PoolExecutor class. We should wait for all active tasks to be in a cancelled state before we shut down the workers.

Noting that this strategy will only work when `shutdown_workers` is called in the context of the overall `shutdown` method, as otherwise the job checker thread is still running, so new items will be sent to the workers.

rtibbles (approved):

PR #9363, opened 2022-04-26
Public signup viewset

### LINKED ISSUE #9356: Create public facility signup viewset
This should just be a public alias for the current SignUpViewset in the auth app. Need to add documentation to that viewset to clarify that it is now being used as a public interface, and so if it is updated for internal uses, updates must now be versioned.

We can either version the public viewset with a prefix, an optional version parameter to the signup payload, or with a 'Cinderella' strategy, where we try to use each possible version for signing up until one fits.

The latter seems simplest to implement, and reflects a strategy that we have already used for our content database schema matching.

rtibbles:
## Summary
* Cleans up existing facility user serializer to consolidate logic
* Removes unused public facility user serializer on read only viewset
* Adds public sign up viewset
* Exposes on the public APIs
* Adds capability for keeping old serializers around to handle historic data formats for future back compatibility

## References
Fixes #9356

## Reviewer guidance
Anything I missed? Does it make sense to even bother with the legacy serializers at this stage?

----

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/core/auth/api.py
+++ b/kolibri/core/auth/api.py
@@ -25,6 +25,7 @@ from django.db.models.functions import Cast
 from django.http import Http404
 from django.utils.decorators import method_decorator
 from django.utils.timezone import now
+from django.views.decorators.csrf import csrf_exempt
 from django.views.decorators.csrf import ensure_csrf_cookie
 from django_filters.rest_framework import CharFilter
 from django_filters.rest_framework import ChoiceFilter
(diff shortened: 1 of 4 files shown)


jredrejo (changes requested):
The implementation has some problems:

- If the user has previously signed in into the facility, the api returns a 404 , PERMISSION_DENIED
- If the facility has been configured to allow users without password, the api should allow it, and it does not. Password is always required. I've tested using ` {"username": "kkkk5", "password": "NOT_SPECIFIED"} `works correctly and the password is not created in the db, but probably the api should be able to handle it automatically, not depending on this `NOT_SPECIFIED` being sent if the facility has this setup.

Also, I'm worried for the security problems this api can introduce. We should have some option to disable it, or at last, throttle it. I can think of kolibri servers being under attack just creating thousands of users. Is it really needed this feature for anonymous users?

rtibbles:
> If the user has previously signed in into the facility, the api returns a 404 , PERMISSION_DENIED

That seems right, if you're currently signed in, you shouldn't be able to sign up? That also prevents a client from accidentally issuing a repeat request.

> If the facility has been configured to allow users without password, the api should allow it, and it does not. Password is always required. I've tested using {"username": "kkkk5", "password": "NOT_SPECIFIED"}works correctly and the password is not created in the db, but probably the api should be able to handle it automatically, not depending on this NOT_SPECIFIED being sent if the facility has this setup.

This is how the existing signup viewset handles it - I didn't see a strong need to depart from the same behaviour. We can certainly drop the `NOT_SPECIFIED` but as this is intended for programmatic access from other Kolibris it feels like this is OK?

> Also, I'm worried for the security problems this api can introduce. We should have some option to disable it, or at last, throttle it. I can think of kolibri servers being under attack just creating thousands of users. Is it really needed this feature for anonymous users?

I am not sure I understand the concern here over and above the existing signup viewset?

jredrejo:
> That seems right, if you're currently signed in, you shouldn't be able to sign up? That also prevents a client from accidentally issuing a repeat request.

So is this intentional?. On one hand, it's perfect to avoid someone creating multiple users from a browser as after the first is done, being in the session disallows it. On the other hand this will force an external api/application to create multiple sessions if multiple users needs to be created. For example, trying to replicate several users in one school server to another. Technically this is solvable but requires more work.

jredrejo (approved):
As I mentioned in the previous comments this is a feature that's dangerous for servers connected to Internet. 
I think we should make it cristal clear some where in our documentation the need to add some middleware or server throttling to avoid easy DDOS attacks.

Appart from that concern, the code is fine and works perfectly as intended.

rtibbles:
> So is this intentional?. On one hand, it's perfect to avoid someone creating multiple users from a browser as after the first is done, being in the session disallows it. On the other hand this will force an external api/application to create multiple sessions if multiple users needs to be created. For example, trying to replicate several users in one school server to another. Technically this is solvable but requires more work.

Oh - actually, I deliberately did not include the 'sign in' behaviour of the internal Signup viewset in the public viewset, so I am surprised that this is happening. Is the 404 happening when you make the exact same POST sequentially?

jredrejo:
> > So is this intentional?. On one hand, it's perfect to avoid someone creating multiple users from a browser as after the first is done, being in the session disallows it. On the other hand this will force an external api/application to create multiple sessions if multiple users needs to be created. For example, trying to replicate several users in one school server to another. Technically this is solvable but requires more work.
> 
> Oh - actually, I deliberately did not include the 'sign in' behaviour of the internal Signup viewset in the public viewset, so I am surprised that this is happening. Is the 404 happening when you make the exact same POST sequentially?

I haven't tried that. Accidentaly I was testing the api being logged in the server and it always returns a 404. I had to log out to be able to use this new api.

rtibbles:
Hrm, I suppose this is not a hard requirement, so I could remove this.

PR #9364, opened 2022-04-26
Allow apps to use OS users

### LINKED ISSUE #9357: Allow device provision endpoint to create a special user account based on the OS user
Add a new capability to the app interface https://github.com/learningequality/kolibri/blob/develop/kolibri/plugins/app/utils.py#L9 called `GETUSERNAME`. When called, the method registered to this capability should return the currently logged in user running the app.

When app context is active on the request (as verified by this function (https://github.com/learningequality/kolibri/blob/develop/kolibri/core/auth/api.py#L825), and the above capability is registered, the device provision endpoint (https://github.com/learningequality/kolibri/blob/develop/kolibri/core/device/api.py#L65) should allow a `superuser` not to be specified for its serializer, and instead to create a superuser whose username is computable from the username returned by the capability.

The password should be set to an opaque value to prevent unauthorized access.

### LINKED ISSUE #9358: Automatically get or create system user in app context if the device has been provisioned
Using the capabilities described in #9357, when a user accesses the App initialize URL: https://github.com/learningequality/kolibri/blob/develop/kolibri/plugins/app/api.py#L43 and is properly authenticated in the app context, if there is no current logged in user, then the capability should return the username, and then the appropriate user looked up.

If the user does not exist it should be created (to allow for simple reuse in multiuser environments), then the user should be automatically logged in within the request.

rtibbles:
## Summary
* Creates a local model to do a lookup from OS username to FacilityUser
* Uses this to allow the device provisioning serializer to not be passed super user info, and instead to create its own OS User specific super user
* Adds a method to get or create a user for a specific OS Username
* Adds a capability to the app interface to allow apps to inject the OS Username lookup
* Allows the initialize endpoint used in the app to login as the relevant OS User when there is no other user logged in already

## References
Fixes #9357
Fixes #9358

## Reviewer guidance
Does the regular setup wizard flow all work as expected?

Edge case that has not been implemented here - what do we do for existing users on apps? Should we let them 'link' a user to their OSUser?

----

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated


kolibri/core/device/serializers.py
```diff
@@ -56,6 +59,18 @@ class Meta:
...
         )
 
+    def validate(self, data):
+        if (
+            "superuser" not in data
+            and GET_USERNAME in interface
+            and "request" in self.context
+            and valid_app_key_on_request(self.context["request"])
+        ):
+            data["os_user"] = True
```
dylanmccall: One issue I have with this approach is our authentication becomes "knows the app key." The trouble is, every client application needs that same key in order to work, but we don't necessarily want every client application to be able to authenticate as any user.

In the implementation of this we have with Kolibri GNOME + Endless OS (https://github.com/endlessm/kolibri-desktop-auth-plugin/), we ended up using a separate authentication token for user logins. That's because, on Endless OS, we have a system-wide Kolibri service that multiple desktop users may talk to. So, each of those users is allowed to see the app key. Instead, we need a different key to be able to say "I am _this_ user."

kolibri/core/auth/models.py
```diff
@@ -634,6 +636,43 @@ def create_superuser(self, username, password, facility=None, full_name=None):
...
+            return os_user.user
+        except OSUser.DoesNotExist:
+            password = uuid4().hex
+            user = None
+            for i in range(0, 10):
+                try:
+                    with transaction.atomic():
+                        user = self.create_user(
+                            "{}{}".format("_" * i, os_username),
+                            password=password,
```
dylanmccall: Instead of generating a random password, we could do `user.set_unusable_password()`. From there, the only way to authenticate (by default) is through the app's desktop user interface. Kolibri handles users with unusable passwords quite nicely. For instance, if you are logged in a such a user, it allows you to set a password (from the _Change password_ dialog) without asking you for an existing password.

kolibri/core/device/models.py
```diff
@@ -295,3 +295,23 @@ class UserSyncStatus(models.Model):
...
+    user = models.OneToOneField(
+        FacilityUser,
+        on_delete=models.CASCADE,
+        related_name="os_user",
+        blank=False,
+        null=False,
+        primary_key=True,
+    )
+    # Apparently the max length on linux is 8, on Windows it's 30, and not sure on Mac,
+    # but hopefully 64 should be long enough for anybody!
```
dylanmccall, line 314: For desktop auth on Linux we ended up using both the UID (short numerical identifier) and username (slightly longer human-readable label). One advantage here is Kolibri shows the Kolibri user's username in places, so it's better if that is their name (like `bob`) rather than a funny number like `1000`.

dylanmccall (commented):
Just some design-related comments. Incidentally, if you have a Linux desktop handy with the newest version of the [Kolibri GNOME app](https://flathub.org/apps/details/org.learningequality.Kolibri) installed, you can test out the automatic login there using…

```
env KOLIBRI_APP_FORCE_AUTOMATIC_LOGIN=yes flatpak run org.learningequality.Kolibri
```

One area it runs into trouble is if a user logs out, there is no way to log in without closing the app and opening it again because there's nothing we can easily hook to for the login screen itself to say "oh, I can automatically authenticate as this user." Having this kind of stuff in core will help tremendously :)

kolibri/core/device/models.py
rtibbles: Yeah, I thought that a username would be more user readable in general, although, we could just use the uid for connecting the OS user to the FacilityUser, and just use the username for naming the FacilityUser - but it seemed simpler to only have to define one capability for the interface.

kolibri/core/auth/models.py
rtibbles: True - just trying to think if that will have any unintended consequences anywhere along the line.

kolibri/core/device/serializers.py
rtibbles: So, the issue is that the service is running as a global user, so getting the OS Username in the backend will always give that user, and not the user that is actually running the app?

So, we could embellish the OSUser model to also generate a unique auth token here - and that gets returned as part of the `get_initialization_url` call, if the capability for getting the OSUser is registered. That means we would move the dynamic user creation from the initialize API endpoint, to the get_initialize_url method, and then add the auth token as an extra (optional) parameter in the initialization URL.

What we would then need to do for the device provision serializer, is instead provide the OSUsername in the provisioning request, in order to generate the user appropriate, as the backend might be running in a different OSUser context.

Does that all seem correct?

kolibri/core/device/models.py
dylanmccall: Ah yes, I think that makes sense indeed. They're both static, after all :)

kolibri/core/device/serializers.py
dylanmccall: > So, the issue is that the service is running as a global user, so getting the OS Username in the backend will always give that user, and not the user that is actually running the app?

Indeed! And also it's a long-running service that may be used by multiple desktop users in the same session.

> So, we could embellish the OSUser model to also generate a unique auth token here - and that gets returned as part of the `get_initialization_url` call, if the capability for getting the OSUser is registered. That means we would move the dynamic user creation from the initialize API endpoint, to the get_initialize_url method, and then add the auth token as an extra (optional) parameter in the initialization URL.
>
> What we would then need to do for the device provision serializer, is instead provide the OSUsername in the provisioning request, in order to generate the user appropriate, as the backend might be running in a different OSUser context.
> 
> Does that all seem correct?

Okay, so the desktop app would get back its own auth token and hold on to that to authenticate with? I think that makes sense. We'd lose a bit of verifiability. In particular, we wouldn't be able to say whether the user is who they claim to be, or an administrator or a standard user. But we'd still get some pleasant behaviour from the app itself.

The moving part in all this is we need something that can reliably say "this client is this desktop user," which is challenging with HTTP. I think that's the spot where we need the ability to plug in a more fine tuned solution. A good tool for that on Linux is D-Bus. So, a client asks a trusted D-Bus service for a token, and that service generates a token and remembers which user it belongs to. (It can also talk to another service to get some details about that user). Then our Kolibri plugin talks to the same D-Bus service and asks it "who is this token for?". But of course, on other platforms we'd have to do that a different way entirely.

rtibbles: The point of the initialization_url is to have a shared secret between the Kolibri backend and the webview that the app frontend is using, what I am suggesting is that we enhance that to be user specific. Rather than just 'this is an app controlled webview', it's now 'this is a webview controlled by this user'.

I note that you're not using the app interface anywhere in the GNOME app in order to provide this sort of guarantee, but it seems like the modules in there do have programmatic access to Kolibri, so this should be workable?

The code that is responsible for loading the webview would therefore be responsible for getting a URL that encodes both the app-key shibboleth, and the new Kolibri generated auth token I am proposing. This would then allow the backend to login to the appropriate user before redirecting to the page that you actually want to go to.

The capability I have defined here for 'get_user' could in theory call out to the D-Bus to get the correctly authenticated username, and ensure that it never returns the wrong one, so this system is already pluggable to the extent that I think is necessary - moving the tokens external to Kolibri seems like extra complexity.

This does not resolve whether the user is a superadmin or not - we could expand the app interface capabilities to get 'this user is an admin' to determine Kolibri permissions as well.

kolibri/core/auth/models.py
```diff
@@ -586,7 +588,7 @@ def create_user(self, username, email=None, password=None, **extra_fields):
         """
         if not username:
             raise ValueError("The given username must be set")
-        if "facility" not in extra_fields:
+        if "facility" not in extra_fields or extra_fields["facility"] is None:
```
jamalex: Could (probably?) just be `if extra_fields.get("facility") is None`.

kolibri/core/auth/models.py
jamalex: We also already have the explicit "NOT_SPECIFIED" dummy password (which isn't hashed etc, just stored raw, [from what I see](https://github.com/learningequality/kolibri/blob/9e53513c43d1c08d66911204dd4b1e80e0eeb5d4/kolibri/core/auth/api.py#L693), so it likely behaves equivalently to `set_unusable_password`, which stores something like `"!" + random_digits`, which will also never be the output of a hashing alg)

kolibri/core/device/serializers.py
jamalex: Just to check my understanding of what's at stake here, by listing out some of my assumptions/interpretations (to be corrected if off-base, plz!):
- This seems specifically relevant for the case where there's a system-wide installation of the service with a shared cross-user KOLIBRI_HOME, which is currently only the case for the GNOME app.
- The concern @dylanmccall is raising is that if the frontend app (which runs in user-space) is able to choose which user to log in as, without some way to prove to the backend that it should have the rights to that user, then someone logged in as user A could (with some poking around with internals, but within the rights of their OS user account) log in and access the data of another user.
- This concern is specifically relevant if we're *not* making everyone a super admin, as noted, since otherwise they can access any user's data anyway once logged in.
- The concern is also only relevant if the OS user accounts can't directly read from the KOLIBRI_HOME directory, since otherwise they could read from the Kolibri DB and access any user's data as well. I imagine the app could be made to function without access to KOLIBRI_HOME except from the background system service, but I'm not sure of whether it currently does that.
- All that being said, if the prior two bullets do hold up, I do understand the concern in principle, and agree that it would be ideal to restrict access for cases like this (i.e. specifically for the way the GNOME app behaves, at least on Endless OS -- I don't think it would currently apply anywhere else?).
- The reason @dylanmccall proposes using D-Bus as a way to get a user-specific token is that it's one way for the app to demonstrate to the Kolibri server process that the token request is coming from a particular actual user, rather than taking it at its word.

Regarding that last bullet, another alternative that comes to mind would be creating and storing a token (non-world-readable) in the user's own data directory, and then associating that token with the user in the backend -- and then the app could re-use that token for re-logging in.[1] For creating a super-admin account, it could use something like visudo to drop its token into a "known admin tokens" directory in KOLIBRI_HOME, thereby proving that the user of that token should be allowed to have an admin account in Kolibri. Not sure whether that's more, or less, general-purpose than relying on D-Bus, but might achieve similar ends.

[1] Update: after reading back through the thread again, I think this is basically where the conversation was heading -- with "the desktop app would get back its own auth token and hold on to that to authenticate with". While the simple version of that approach covers the "can only log in as that same user when you return" case, the piece needs to be covered is distinguishing whether to make the OSUser an admin or learner -- which it seems could be addressed either by D-Bus or the `visudo` type approach I described above.

kolibri/core/device/serializers.py
```diff
@@ -78,12 +93,20 @@ def create(self, validated_data):
             facility.dataset.save()
 
             # Create superuser
-            superuser = FacilityUser.objects.create_superuser(
-                validated_data["superuser"]["username"],
-                validated_data["superuser"]["password"],
-                facility=facility,
-                full_name=validated_data["superuser"].get("full_name"),
-            )
+            if "os_user" in validated_data and validated_data["os_user"]:
```
jamalex: Could just be `if validated_data.get("os_user")`

jamalex (commented):
This is going to be great!

kolibri/core/device/serializers.py
rtibbles: So, I think the approach that would fix this is this:

* `get_initialize_url` takes an additional `token` parameter.
* This token gets set in the URL
* When visiting the `initialize` endpoint, if that token is present, the backend looks up to see if there is a user that is associated with that token
* If so it logs in the user
* If not, it creates that user account and logs them in
* In order to accommodate the provisioning case, if the device is not provisioned, we would need to set the token as a temporary cookie in the initialization endpoint, so that the device provisioning can read that to know to create a superuser for that token
* The other piece here is now how to handle the username - it seems that we can either keep the current capability and execute that within the `get_initialize_url` to extract the username, or perhaps it is simpler to just require the calling app to pass in the username as well?
* In that case, we can remove the username capability, and just make it another argument for `get_initialize_url` - this would then be stored alongside the token in the temporary cookie in the provisioning case.

dylanmccall: Yeah, I think that would work nicely. It's pretty ordinary for a web service to have a one-time login token system anyway, so plugging in to that sort of thing from `get_initialize_url` seems pretty reasonable. It also helps to simplify the setup flow, which would be great :)

For a system where we have less trust between the server and the app (such as when the server is a d-bus daemon running as another user), we can have another API that calls `get_initialize_url` with stuff like username filled in automatically.

jamalex: >In order to accommodate the provisioning case, if the device is not provisioned, we would need to set the token as a temporary cookie in the initialization endpoint, so that the device provisioning can read that to know to create a superuser for that token

If I'm understanding correctly: If the desired behavior is that the first user to interact with Kolibri becomes the sole superuser, then this makes sense. The other direction being discussed above was having a user's superuser status be determined by whether they OS-level admin access (e.g. being root, or a sudoer).

kolibri/core/auth/models.py
rtibbles: Closing the loop here - I think using `NOT_SPECIFIED` has a distinct advantage that it would prompt a user to set a password if they merged their OS user into a facility, but then they would still be able to do passwordless login within their own app.

The main thing is it means we have to set the quick start facility type to passwordless so that we don't prompt users to set their password.

rtibbles: Hrm, now I think about it - this might be a reason not to do this - and we should add an alternative dummy password.

rtibbles: And.... changing my mind once again! We only do this prompting if you have logged in via the SessionViewset create method, so this would not be needed. `NOT_SPECIFIED` it is!

kolibri/core/device/serializers.py
rtibbles: I think I've addressed all the above concerns - to handle the provisioning case, I instead avoid creating a super user during provisioning, and leave it up to the app to redirect the user to the initialize URL after provisioning to create a user or superuser, depending on the returned value from the `get_os_user` capability, which is now expected to return a tuple of `username, is_superuser`.

Let me know what you think!

kolibri/core/auth/serializers.py
```diff
@@ -16,6 +16,7 @@
 from .models import Membership
 from .models import Role
 from kolibri.core import error_constants
+from kolibri.core.auth.constants.demographics import NOT_SPECIFIED
```
jamalex, line 19: Don't think this was introduced here, but feels a bit funny to be using a demographics constant for password stuff.

kolibri/core/device/models.py
jamalex: Usernames aren't actually static, are they, on UNIX systems? Since you can do:
`sudo usermod -l newUsername oldUsername`
(in practice, I don't know that it's a big deal, as it doesn't happen often, but worth noting)

kolibri/plugins/app/utils.py
```diff
@@ -16,25 +22,32 @@ def __init__(self):
...
+    def get_initialize_url(self, next_url=None, auth_token=None):
         if not self.enabled:
             raise RuntimeError("App plugin is not enabled")
         # Import here to prevent a circular import
         from kolibri.core.device.models import DeviceAppKey
 
         url = reverse(
-            "kolibri:kolibri.plugins.app:initialize", args=(DeviceAppKey.get_app_key(),)
+            "kolibri:kolibri.plugins.app:initialize",
+            args=(DeviceAppKey.get_app_key(),),
```
jamalex, line 40: So... is it always still using this centralized token, vs some user-specific token that the app frontend app itself gets? I'm thinking about the EOS case of the backend service running as a different user than the app frontend (of which there may be multiple, under different users).

jamalex: Oh, wait, so there are actually two tokens? The app key and the auth token? Where are the auth tokens generated/stored, and how does the app logic validate them?

jamalex (commented):
Looking good overall! I'm still not 100% clear on the auth token stuff, which is pretty central -- may become clearer with example app code, as well. Happy to chat through.

kolibri/core/auth/serializers.py
rtibbles: No, we should probably add a separate (albeit identical) constant for the password case.

kolibri/plugins/app/utils.py
rtibbles: The app wrapper is responsible for ensuring the auth token is generated/stored. In the case of the Android app, I think we can just handle this on disk in a JSON file or somesuch.

For the Endless app - there will be the separate system service that @dylanmccall suggested for dishing out the tokens to the currently active user.

This seemed to be what was needed in order to be able to distinguish the currently running app user vs the user for the server (which might not be the app user in the Endless case).

jamalex (approved):
I thought I had approved from my side already, but this all looks good to me! The rest of the pudding proofs are in the app code.

PR #9404, opened 2022-05-04
Remove profile dropdown

### LINKED ISSUE #9275: Remove profile dropdown menu and add details to Side Nav
This issue is part of the app bar refresh work. The scope of this issue is removing the current dropdown menu (which will reduce the number of menus and places a user might look for information) and place the relevant profile information into the existing side nav.

### Scope
- [ ] Turn the current user profile dropdown menu should become just a text display of the username. The conditions for truncating the user name should be persisted as they are
- [ ] Move profile information into the side nav
- [ ] Use existing `<SyncStatusDisplay>` component for display on learn only devices

### Acceptance Criteria
- [ ] Layout should be pixel-perfect match to Figma specifications 
- [ ] Side navigation menu should continue to be tab-navigable and tab order should be persisted. The new informational display including learner name, points, and device status do not need keyboard focus, as they are not interactive elements
- [ ] Sync information on Learn-only devices should display accurately based on the current users syncing status or connection
- [ ] Update any tests in `app-bar.spec.js` as needed based on your changes

marcellamaki:
## Summary
This PR moves the data previously in the user profile dropdown menu into the side navigation as part of the 0.16 update. 

Fixes #9275 

## References

Relevant [Figma file](https://www.figma.com/file/MpCG7r5PULOCJsEvnTSti8/Android-Designs-2022?node-id=868%3A14889)

## Reviewer guidance
Most of these changes can be reviewed visually in the front end for proper layout. 

Accessibility has also been tested for the addition of the "change language" option (which is not actually a link, but rather a button that opens the language switcher modal), but should be re-checked.

The section I see as potentially where things went awry would be with the SyncStatus display and the transfer of information from where it was previously to where it is now... while it was mostly a copy and paste, something might have gotten lost in the shuffle. This should probably get properly mocked with QA. 

Screenshots 

| Description  |  Screenshot |
|---|---|
| Main browser view with dropdown menu now removed (additional changes to the points display out of scope)  | Screen Shot 2022-05-04 at 4 38 50 PM |
| Default Learner Display  | Screen Shot 2022-05-04 at 4 52 48 PM |
| Learn Only Device display | Screen Shot 2022-05-04 at 4 40 40 PM |
| Overflow points/username display | Screen Shot 2022-05-04 at 4 41 44 PM |

 
----

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated


kolibri/core/assets/src/views/SideNav.vue
```diff
@@ -33,14 +33,49 @@
...
               :style="{ backgroundColor: $themeTokens.surface }"
               :aria-label="$tr('navigationLabel')"
             >
               <template #options>
-                <component :is="component" v-for="component in menuOptions" :key="component.name" />
+                <component
+                  :is="component"
+                  v-for="component in menuOptions"
+                  :key="component.name"
+                  @openLanguageModal="handleShowLanguageModal()"
```
rtibbles: Rather than adding this listener to every nav component, maybe we should just move the handling of the LanguageSwitcherModal into the ChangeLanguageSideNavEntry instead? Can instantiate it in there just as easily, as it's a modal, so it should bust out of its container.

Alternatively, can do what was done in the original implementation, and just add an extra component after the `v-for` components. If you do that, you can get away without defining a new component at all and just copy the relevant string into here.

kolibri/core/assets/src/views/SideNav.vue
```diff
@@ -243,10 +314,42 @@
...
+      this.isPolling = false;
+    },
     methods: {
       toggleNav() {
         this.$emit('toggleSideNav');
       },
+      handleShowLanguageModal() {
+        this.languageModalShown = true;
+      },
+      pollUserSyncStatusTask() {
```
rtibbles, line 338: In the original impementation, the polling behaviour was conditional on the dropdown being shown - this seems to have been removed in this refactor, even though we have the `navShown` prop that we could conditionalize on in the same way.

kolibri/core/assets/test/views/app-bar.spec.js
```diff
@@ -1,173 +1,17 @@
...
-    });
-    it('should show logout if no components are added and user is logged in', async () => {
-      expect(navComponents).toHaveLength(0);
-      const wrapper = createWrapper(undefined, {});
-      setUserKind(wrapper.vm.$store, UserKinds.LEARNER);
-      await wrapper.vm.$nextTick();
-      expect(wrapper.findComponent(logoutSideNavEntry).element).toBeTruthy();
-    });
-
-    describe('SoUD sync status and learn-only device indicators', () => {
```
rtibbles, line 79: I feel like a couple more of these tests might withstand the great migration.

rtibbles (changes requested):
I think there are a couple of places where doing a more direct port of the original implementation might be beneficial.

kolibri/core/assets/test/views/app-bar.spec.js
marcellamaki: I think aside from the user menu dropdown, many of these were redundant to existing code in `side-nav.spec.js` (and, I moved a few there). Do you have a suggestion of specific tests that should be added back (and/or moved?)

rtibbles: Yeah, expanding the existing tests out they do seem to be redundant. Disgregard!

kolibri/core/assets/src/views/AppBar.vue
```diff
@@ -39,79 +39,7 @@
...
-            </template>
-
-            <template #footer>
-              <!-- Only show this when on a SoUD -->
-              <div v-if="showSoudNotice" class="role" data-test="learnOnlyNotice">
-                <LearnOnlyDeviceNotice />
-              </div>
-            </template>
-          </CoreMenu>
+          <span v-if="isUserLoggedIn" class="username" tabindex="-1">{{ dropdownName }}</span>
```
nucleogenesis: Since it isn't a dropdown anymore, maybe `dropdown` name should be renamed to avoid future confusion?

kolibri/core/assets/src/views/SideNav.vue
```diff
@@ -243,10 +322,46 @@
...
     methods: {
+      ...mapActions(['fetchUserSyncStatus']),
       toggleNav() {
         this.$emit('toggleSideNav');
       },
+      handleShowLanguageModal() {
+        this.languageModalShown = true;
+      },
+      pollUserSyncStatusTask() {
+        console.log('is polling');
```
nucleogenesis: Lingering console.log

nucleogenesis (commented):
This adds a lot of new code to the SideNav component to add a new self-contained section. I'm wondering if it might be worth breaking that stuff into its own separate component to separate the concerns around sync status, user kind, username, fullname, etc. 

This isn't a blocking change request, but this feels like it makes SideNav do and care about more than it probably ought to. Curious about your and others' thoughts in any case.

Overall, the PR looks good. I've not browser tested it, but if QA passes on it should be good to go.

rtibbles (approved):

radinamatic:
This looks much less disruptive to me, thank you! 🙂 

remove-profile

Not a blocker, just noticing that since the string context changed, it's not picking up the translations for the ``Change language`` string anymore...

radinamatic (approved):
LGTM 👍🏽 :shipit:!

PR #9428, opened 2022-05-13
Move drive info endpoint into device app.

rtibbles:
## Summary
[TASK API CLEANUP]
* Moves the drive info endpoint out of the tasks/api.py
* Updates all frontend references to this

----

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/core/device/api.py
+++ b/kolibri/core/device/api.py
@@ -36,6 +36,8 @@ from kolibri.core.auth.api import KolibriAuthPermissions
 from kolibri.core.auth.api import KolibriAuthPermissionsFilter
 from kolibri.core.auth.models import Collection
 from kolibri.core.content.permissions import CanManageContent
+from kolibri.core.content.utils.channels import get_mounted_drive_by_id
+from kolibri.core.content.utils.channels import get_mounted_drives_with_channel_info
 from kolibri.core.device.permissions import IsSuperuser
 from kolibri.core.device.utils import get_device_setting
 from kolibri.core.discovery.models import DynamicNetworkLocation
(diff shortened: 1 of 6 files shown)


kolibri/core/device/api.py
```diff
@@ -336,3 +338,14 @@ def partial_update(self, request, *args, **kwargs):
...
                 plugin.disable()
         return Response(self._serialize(plugin))
+
+
+class DriveInfoViewSet(viewsets.ViewSet):
+    permission_classes = (CanManageContent,)
+
+    def list(self, request):
+        drives = get_mounted_drives_with_channel_info()
+        return Response([mountdata._asdict() for mountdata in drives])
```
marcellamaki, line 368: Why is `_asdict()` added here and what does it change? I know it is literally as a dictonary, but more... why is this necessary and what is it facilitating? And why here? It seems like it's not really used very much at all in other places in kolibri. (ETA: as I kept reading, I realize this was a moving code, not an adding code, but still have the question)

kolibri/plugins/device/assets/src/modules/manageContent/actions/taskActions.js
```diff
@@ -48,7 +50,7 @@ export function refreshTaskList(store) {
 }
 
 export function refreshDriveList(store) {
-  return TaskResource.localDrives().then(({ data }) => {
+  return client({ url: urls['kolibri:core:driveinfo-list']() }).then(({ data }) => {
```
marcellamaki, line 53: Do we have dev docs or a how-to for this pattern somewhere? I know it's used in multiple places, but I get thrown off by the string every time and have to read it a few times. If not, maybe I will task myself with updating our docs to add it...

kolibri/core/device/api.py
rtibbles: The object returned from `get_mounted_drives_with_channel_info` is a list of Python `namedtuple`s - which are not directly JSON serializable, so we use the `namedtuple`'s `_asdict` method in order to turn them into a dict which the backend will serialize to JSON without complaint.

kolibri/plugins/device/assets/src/modules/manageContent/actions/taskActions.js
rtibbles: We appear not to, no. I think we've so far skated by without it, because mostly this URL namespacing is hidden inside the API Resource layer, but some documentation of it would be helpful.

marcellamaki (approved):

PR #9425, opened 2022-05-13
Migrate to tasks API

rtibbles:
## Summary
Draft of changes to migrate completely to the new tasks API.

Includes:
* Refactor to have all args and kwargs that are passed to task functions as explicit arguments, so as to prevent mixing and collision with other function arguments.
* Move to a DRF Serializer based Validator mechanism for job data coming from the frontend
* Cleanup naming conventions - jobs for particular running tasks, tasks for the general functions that can be run.
* Make a new tasks permissions system to be compatible with bulk creation and more aligned with internal Kolibri permissions than DRF.
* Consolidate tasks to share validators through mixins and inheritance.
* Fixes miscellaneous bugs and tests that occurred along the way

## References

…

## Reviewer guidance

…

----

## Testing checklist

- [ ] Contributor has fully tested the PR manually
- [ ] If there are any front-end changes, before/after screenshots are included
- [ ] Critical user journeys are covered by Gherkin stories
- [ ] Critical and brittle code paths are covered by unit tests

## PR process

- [ ] PR has the correct target branch and milestone
- [ ] PR has 'needs review' or 'work-in-progress' label
- [ ] If PR is ready for review, a reviewer has been added. (Don't use 'Assignees')
- [ ] If this is an important user-facing change, PR or related issue has a 'changelog' label
- [ ] If this includes an internal dependency change, a link to the diff is provided

## Reviewer checklist

- Automated test coverage is satisfactory
- PR is fully functional
- PR has been tested for [accessibility regressions](http://kolibri-dev.readthedocs.io/en/develop/manual_testing.html#accessibility-a11y-testing)
- External dependency files were updated if necessary (`yarn` and `pip`)
- Documentation is updated
- Contributor is in AUTHORS.md

PR #9438, opened 2022-05-14
Tasks backend refactor

### LINKED ISSUE #9313: Use args, kwargs explicitly as a standard everywhere for task backend
Currently we pop off `Job` attributes and then assign the remaining attributes as args & kwargs which has potential to be problematic. So to make our tasks backend robust and support future flexibility we should explicitly use args, kwargs for `Job` objects.

rtibbles:
## Summary
[TASK API CLEANUP]

Main features:
* Implementation of the Validation and Permission classes to handle validation and permission for tasks.
* Tweaks to RegisteredJob and JobRegistry (including renames to represent they are about Tasks, not Jobs)
* Refactor of Job signature, to explicitly separate function args and kwargs from Job args and kwargs
* Updates the task API endpoint to use the new validation and permissions
* Updates the logger tasks with the validators and permissions as an example
* Updates and consolidates all auth tasks
* Updates all content tasks

## References
Fixes #9313 

…

## Reviewer guidance
Go commit by commit - can skip the commits:
* [Delete old code from tasks api.py](https://github.com/learningequality/kolibri/commit/9c41c137de97af87552f6f9ab75c7cc202047311)
* [Remove Queue class and update task tests.](https://github.com/learningequality/kolibri/commit/7276cbd047e49a9cf8325b4934d86e2bfd3a1684)

This should give a good overall sense of the changes intended to be introduced by #9425, although by doing a partial implementation here, tests will not pass.

In the interests of easier code review, I am happy to merge with broken tests and then do progressive PRs to fix it up if that's what is needed.

@jredrejo I took your comment on #9425 into account and have changed the `task` field to `type` for consistency with current frontend usage.

----

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated


kolibri/core/logger/tasks.py
```diff
@@ -19,48 +20,39 @@ def get_logs_dir_and_filepath(log_type, facility):
...
-        logs_dir, filepath = get_logs_dir_and_filepath(log_type, facility)
-    else:
-        raise Http404("Impossible to create a csv export file for {}".format(log_type))
-
-    if not os.path.isdir(logs_dir):
-        os.mkdir(logs_dir)
-
-    job_type = "EXPORTSUMMARYLOGCSV" if log_type == "summary" else "EXPORTSESSIONLOGCSV"
+class ExportLogCSVValidator(JobValidator):
+    facility = serializers.PrimaryKeyRelatedField(
```
jredrejo, line 24: thanks, this clarifies a lot how to use the new validation method

kolibri/core/logger/tasks.py
```diff
@@ -19,48 +20,39 @@ def get_logs_dir_and_filepath(log_type, facility):
...
+        if not os.path.isdir(logs_dir):
+            os.mkdir(logs_dir)
 
-    return {
-        "log_type": log_type,
-        "filepath": filepath,
-        "facility": facility,
-        "extra_metadata": job_metadata,
-    }
+        return {
```
jredrejo, line 41: I don't see here some of the data that was in the original implementation. 
The previous code provided to  the frontend  "`started_by`",  "`type`"  and the whole "`facility`" object. Are these values now returned by default by the `JobValidator` always?

kolibri/core/tasks/api.py
```diff
@@ -238,26 +53,34 @@ def validate_create_req_data(self, request):
...
+            "status": job.state,
+            "type": job.func,
+            "exception": job.exception,
+            "traceback": job.traceback,
+            "percentage": job.percentage_progress,
+            "id": job.job_id,
+            "cancellable": job.cancellable,
+            "clearable": job.state in [State.FAILED, State.CANCELED, State.COMPLETED],
+            "facility_id": job.facility_id,
+            "extra_metadata": job.extra_metadata,
```
jredrejo, line 81: `extra_metadata` has been removed in the above `ExportLogCSVValidator`, but here I see it's still available to be used, thus keeping it in the `ExportLogCSVValidator` would save frontend work. Is there any intention to deprecate it or what's the reason to have it available but not using it?

kolibri/core/tasks/api.py
```diff
@@ -307,1225 +130,110 @@ def create(self, request):
...
-                request_data = validator_result
-
-            try:
-                user_facility_id = request.user.facility_id
-            except AttributeError:
-                user_facility_id = None
-
-            job_id = registered_job.enqueue(
-                job_facility_id=user_facility_id, **request_data
+        for registered_task, job in validated_data:
```
jredrejo, line 141: wonderful!

kolibri/core/tasks/api.py
```diff
@@ -307,1225 +130,110 @@ def create(self, request):
...
-        if not task_id:
-            return Response({})
-
-        for _queue in self.queues:
-            _queue.clear_job(task_id)
-
-        return Response({"task_id": task_id})
-
-    @decorators.action(methods=["post"], detail=False)
-    def deletefinishedtasks(self, request):
```
jredrejo, line 467: are we losing this feature?

kolibri/core/tasks/api.py
```diff
@@ -307,1225 +130,110 @@ def create(self, request):
...
-        elif self.action == "startexportlogcsv":
-            return [CanExportLogs]
-        elif self.action in ["importusersfromcsv", "exportuserstocsv"]:
-            return [CanImportUsers]
-
-        # For all other task endpoints.
-        return [CanManageContent]
-
-    @decorators.action(methods=["post"], detail=False)
-    def startchannelupdate(self, request):
```
jredrejo, line 501: all these taks are being implemented in upcoming PRs, aren't they?

jredrejo (commented):
I really like the changes and how they structure the code. I still have some doubts to understand some minor parts of the changes, but looks really good!

kolibri/core/logger/tasks.py
rtibbles: I noticed that `started_by` and `started_by_username` were being added in a lot of tasks, so I moved them into the base `JobValidator` class, so that it always gets added to every task in the extra_metadata.

"type" is now just the import path of the task (so this will require updates to frontend constants) so for this it would be `kolibri.core.logger.tasks.exportuserscsv` - this has the nice side effect that the type of the task is now used to start the task in question and then refer to it later when in progress.

I was confused as to how the `facility` object was being passed around, as it is not JSON serializable. There is now a nullable `facility_id` property on every job object, so it seemed reasonable to use that to refer to which facility the job is being run on.

kolibri/core/tasks/api.py
rtibbles: Still intended to be used, I just migrated all of the previous `extra_metadata` in that specific task to either always be added, or be redundant with other data already on the job object.

kolibri/core/tasks/api.py
rtibbles: I cleaned up the naming of some of these endpoints to be shorter, so this is just `clear` now.

kolibri/core/tasks/api.py
rtibbles: Yes, I have the code and full updates for all of the backend tasks.

jredrejo (approved):
Thanks for the clarifications. Let's roll the snowball!

docs/backend_architecture/tasks.rst
```diff
@@ -38,22 +38,25 @@ We will refer to below sample code in the later sections also.
...
 
-    @register_task(job_id="02", queue="maths", validator=add_validator, priority=Priority.HIGH, cancellable=False, track_progress=True, permission_classes=[IsSuperuser])
+        def validate(self, data):
+            if data['a'] + data['b'] > 100:
+                raise serializers.ValidationError("Sum of a and b should be less than 100")
+            job_data = super(AddValidator, self).validate(data)
+            job_data["extra_metadata"].update({"user": "kolibri"})
+            return job_data
+
+    @register_task(job_id="02", queue="maths", validator=add_validator, priority=Priority.HIGH, cancellable=False, track_progress=True, permission_classes=[IsSuperAdmin])
```
vkWeb: `validator` should be `AddValidator`...?

docs/backend_architecture/tasks.rst
```diff
@@ -94,12 +97,14 @@ A successful response looks like this:
...
 ``permission_classes`` argument of decorator. If the user has permissions then we proceed.
 
-Then, we check whether the ``"task"`` function has a validator associated with it or not. If it has a validator, it
-gets run. The return value of the validator must be a dictionary. The dictionary returned by the validator is passed to the task function as keyword
-arguments.
+Then, we check whether the registered task function has a validator associated with it or not. If it has a validator, it
+gets run. The return value of the validator must be a dictionary that conforms to the function signature of the Job object.
+The dictionary returned by the validator is passed to a Job object to be enqueued. By default, any key value pairs in the
+request object that are registered as input fields on the validator will be passed to the function as kwargs. If no fields
+are defined on the validator, or no validator is registered, then the function will receive no arguments.
```
vkWeb, line 107: So essentially this means every task function needs to decalre a validator for it to receive arguments? :thinking:

docs/backend_architecture/tasks.rst
```diff
@@ -179,3 +186,5 @@ The response will be a list of enqueued jobs like:
         "clearable": False,
       }
     ]
+
+However, if any task fails validation, all tasks in the request will be rejected. Validation happens prior to enqueuing, so tasks will not be partially started in the bulk case.
```
vkWeb, line 190: :rocket:

kolibri/core/auth/tasks.py
```diff
@@ -1,118 +1,168 @@
...
+        if data.get("delete"):
+            args.append("--delete")
+        kwargs = {"locale": data.get("locale"), "facility": data.get("facility_id")}
+        if "user" in self.context:
+            kwargs["userid"] = self.context["user"].id
+        return {
+            "args": args,
+            "kwargs": kwargs,
+            "facility_id": data.get("facility_id"),
+        }
```
vkWeb, line 159: In theory, the validator can return more `args`, `kwargs` than mentioned in the Serializer, right? So how exactly Serialier is serving the purpose of introspecting of expected arguments for a task function?

vkWeb (approved):
The changes are looking good to me! :zap:  

I just have a few questions.

docs/backend_architecture/tasks.rst
rtibbles: Yes - the main motivation here is to ensure an explicit interface, so if you want to receive arguments, you have to define what those arguments will be using a validator, and because the validator uses serializer fields, you will also be specifying the types as well.

kolibri/core/auth/tasks.py
rtibbles: The serializer/validator is an interface between the API request data and the task function, so it allows introspection of what is required to be sent to the API endpoint in order to initialize a task.

As tasks can still be directly enqueued in the backend without using the API, the function signature of the task function itself is still available for that purpose, although it was convenient in at least one point in this refactor to use the validation functionality for a task enqueued entirely from the backend as well.

docs/backend_architecture/tasks.rst
rtibbles:
```suggestion
    @register_task(job_id="02", queue="maths", validator=AddValidator, priority=Priority.HIGH, cancellable=False, track_progress=True, permission_classes=[IsSuperAdmin])
```

kolibri/core/auth/tasks.py
vkWeb: I am skeptical regarding the use of serializers to just introduce introspection abilities. IMO like any other API, documentation should be a single source of truth for conveying what's expected and what will be the response in various scenarios.

docs/backend_architecture/tasks.rst
vkWeb: Agreed! Explicity is always better!

kolibri/core/auth/tasks.py
vkWeb: Moreover the task's function signature is enough to know for our frontend devs about what's required for the task to be enqueued. That's what function signature's purpose is -- to convey required data. And regarding their type, we should use code documentation. 

I feel serializers have introduced more complexity than there should be...

docs/backend_architecture/tasks.rst
vkWeb: Yep that's what I meant there!

kolibri/core/auth/tasks.py
rtibbles: > the task's function signature is enough to know for our frontend devs about what's required for the task to be enqueued

This was never the case though - the validator functions that were already being used did not just validate and pass through frontend arguments to the task function. They validated, added additional data, and did so with no explicit detailing of what arguments they expect.

By turning from validator function to validator serializers, we now have an explicit delineation of what is expected. This can be inspected by looking at the code, or by looking at `/api_explorer` and examining the task endpoint, which now gives a full spec of the accepted task types and their arguments. This also means that frontend devs do not need to delve through Python code nor understand it to have a clear spec of the task arguments.

vkWeb: @rtibbles makes sense!

PR #9450, opened 2022-05-18
Fix pagination user tables

### LINKED ISSUE #9448: Learners not available to enroll in the class
## Observed behavior
Encountered while testing #9446, according to @rtibbles most probably a regression from the recent pagination work.

Facility has 3 new learners, but none seem available to be enrolled in the class:

enroll-issue

## Errors and logs
…

## Expected behavior
Visible list of learners to enroll.

## User-facing consequences
…

## Steps to reproduce
Create learner(s) and try to enroll them in a class.

## Context
 * Kolibri version: 0.15.x branch, EXE asset from #9446  
 * Operating system: Windows 10
 * Browser: all

marcellamaki:
## Summary
Updates `handler.js` for the class enrollment user management, and simplifies the vue components to better match the updates in the FacilityUsers page. Now learners should be accurately displaying in the user table for enrollment in a class.

## References
Fixes #9448
learners

## Reviewer guidance

Is the pagination working as expected with > 30 users? 

Also... am I supposed to be filtering out anyone not exclusively a learner? Truly cannot remember.... I think coaches and admins can still be added to classes

----

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/plugins/facility/assets/src/modules/classAssignMembers/handlers.js
+++ b/kolibri/plugins/facility/assets/src/modules/classAssignMembers/handlers.js
@@ -2,7 +2,6 @@ import ConditionalPromise from 'kolibri.lib.conditionalPromise';
 import pickBy from 'lodash/pickBy';
 import samePageCheckGenerator from 'kolibri.utils.samePageCheckGenerator';
 import { ClassroomResource, FacilityUserResource } from 'kolibri.resources';
-import { UserKinds } from 'kolibri.coreVue.vuex.constants';
 import { _userState } from '../mappers';
 
 export function showLearnerClassEnrollmentPage(store, toRoute) {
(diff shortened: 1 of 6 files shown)


kolibri/plugins/facility/assets/src/modules/classAssignMembers/handlers.js
```diff
@@ -34,7 +33,7 @@ export function showLearnerClassEnrollmentPage(store, toRoute) {
     ([facilityUsers, classroom, classUsers]) => {
       store.commit('classAssignMembers/SET_STATE', {
         facilityUsers: facilityUsers.results.map(_userState),
-        classUsers: classUsers.results.map(_userState),
+        classUsers: classUsers.results,
```
rtibbles: Without pagination parameters passed to the fetchCollection, I think `classUsers` will just be the results directly, rather than having a results attribute?

Also, why are we removing the userState mapping?

rtibbles (commented):
Seems to make sense from an initial read through - one question though.

kolibri/plugins/facility/assets/src/modules/classAssignMembers/handlers.js
marcellamaki: I don't really know why... originally I had both the pagination parameters and the userState, to follow the same pattern, and I kept getting a 404 error that said ID: NOT FOUND. Removing both was the only way that the results were the correct array.

radinamatic:
30< Learners are back! 🥳 

Tested on Chrome & Firefox on Windows 10



And yes, filtering is only in play when assigning coaches, anybody can be a learner 🙂

radinamatic (approved):
LGTM! 👍🏽 :shipit:

kolibri/plugins/facility/assets/src/modules/classAssignMembers/handlers.js
rtibbles: Soo..... the reason this doesn't matter is that we are already excluding non-class members from the original fetch, so there is no reason to even do this fetch any more, it's dead code.

rtibbles (approved):

radinamatic:
Re-tested on Win 10 after the last commits by @rtibbles, 30< learner pagination still looking good, let's merge for the bugbash tomorrow!

Issue #9355, opened 2022-04-25
Create asynchronous task to register a new user on a remote device and then merge current user into that facility user


rtibbles:
The implementation here should leverage the implementation in #9354 with an initial, additional step of creating the new user on the remote device.

The task can then proceed in the same way as #9354.

PR #9454, opened 2022-05-20
Task to merge remote user

### LINKED ISSUE #9354: Create asynchronous task to merge current user into an existing facility user on a remote server
The task should first carry out a single user sync of the user in question - once that has completed, then the following function: https://github.com/learningequality/kolibri/blob/develop/kolibri/core/auth/utils/migrate.py#L55 should be executed on the local user and the newly synced single user.

The task requires all the parameters for a single user sync (https://github.com/learningequality/kolibri/blob/develop/kolibri/plugins/setup_wizard/tasks.py#L139) plus the id of the local facility user that will be merged into the user that is being synced from the remote.

This task should be created in user_profile/tasks.py and registered using the task decorator.

jredrejo:
## Summary

Creation of asyncronous task to merge current user into an existing facility user on a remote server

The steps this task do are:

1. Do a single user sync from the server to the local device (initiated by the local device)
2. Merge the existing user from the local device into the single user synced user
3. Resync with the server to update the merged records
4. Delete the existing user from the local device

## References
Closes: #9354 

## Reviewer guidance
In developer mode, using http://localhost:8000/api_explorer/# a `POST` to the `/api/tasks/tasks` endpoint can be done
This is an example of payload to use it:
```
{
    "type": "kolibri.plugins.user_profile.tasks.mergeuser",
    "baseurl": "http://192.168.0.201:80/",
    "facility": "41d0e8bb1600347f17ab3d9172fff87a",
    "username": "uno",
    "local_user_id": "05685392311d1d259fe01c65c7a6c28e"
}
```
being `facility` and `username` data of the remote server. If the remote server requires password to authenticate user, a "password" parameter must be added, otherwise it's not needed

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated


kolibri/core/auth/tasks.py
```diff
@@ -316,15 +316,15 @@ def validate(self, data):
 
 class PeerFacilitySyncJobValidator(PeerSyncJobValidator):
     username = serializers.CharField()
-    password = serializers.CharField()
+    password = serializers.CharField(required=False)
```
rtibbles: Could set the default here as `"NOT_SPECIFIED"`.

I was also surprised to discover that we don't have a constant in Python code anywhere for the `"NOT_SPECIFIED"` password constant. I guess a consistently named constant vs a consistently named string is about the same?

kolibri/plugins/user_profile/tasks.py
```diff
@@ -0,0 +1,50 @@
...
+        return job_data
+
+
+@register_task(
+    queue="soud",
+    validator=MergeUserValidator,
+    priority=Priority.HIGH,
+    cancellable=False,
+    track_progress=True,
+    permission_classes=[IsAdmin() | NotProvisioned()],
```
rtibbles: If the device isn't provisioned, there won't be any local users to be merging. I think this should be `IsAdmin` or a new `IsSelf` permission, where you can only start this task for your own local user account?

This should be possible by inspecting the `kwargs` property of the job object.

rtibbles (changes requested):
Just one thing that I think needs updating - other comment is just a suggestion.

kolibri/core/auth/tasks.py
jredrejo: Done.
I'm also very surprised for the lack of constant for `NOT_SPECIFIED` We're using it both in JS and Python but it's not defined. I thought of doing it here but I didn't want to taint this PR with that refactoring. I'll do it in a different PR after this work is finished. There are many different parts of the application affected

kolibri/plugins/user_profile/tasks.py
jredrejo: ah, yes, my mistake. I left `Not_Provisioned` there because I was testing the setup wizard too. Of course, it does not make sense. For the `IsSelf`, I agree, that's a good idea. I've implemented it

kolibri/plugins/user_profile/tasks.py
```diff
@@ -0,0 +1,51 @@
...
+from kolibri.core.tasks.decorators import register_task
+from kolibri.core.tasks.job import Priority
+from kolibri.core.tasks.permissions import IsFacilityAdmin
+from kolibri.core.tasks.permissions import IsSelf
+from kolibri.core.tasks.permissions import IsSuperAdmin
+from kolibri.core.tasks.permissions import PermissionsFromAny
+
+
+class MergeUserValidator(PeerImportSingleSyncJobValidator):
+    local_user_id = HexOnlyUUIDField()
```
rtibbles, line 17: This could also be a `PrimaryKeyRelatedField` with the queryset as the `FacilityUser.objects.all()`

rtibbles (approved):

kolibri/plugins/user_profile/tasks.py
jredrejo: good catch, I'm adding this modification in the PR to close #9355

PR #9456, opened 2022-05-23
Creation of remote users if needed

### LINKED ISSUE #9355: Create asynchronous task to register a new user on a remote device and then merge current user into that facility user

jredrejo:
## Summary
Creation of asyncronous task to merge current user into an existing facility user on a remote server. 
If remote server credentials are provided, the user will be created in the remote server.

The steps this task do are:

1. When accesing to the remote server to fetch the user data, if the user does not exist there, it creates it.
2. Do a single user sync from the server to the local device (initiated by the local device)
3. Merge the existing user from the local device into the single user synced user
4. Resync with the server to update the merged records
5. Delete the existing user from the local device

## References
Closes: #9355 

## Reviewer guidance
In developer mode, using http://localhost:8000/api_explorer/# a `POST` to the `/api/tasks/tasks` endpoint can be done
This is an example of payload to use it:
```
{
    "type": "kolibri.plugins.user_profile.tasks.mergeuser",
    "baseurl": "http://192.168.0.201:80/",
    "facility": "41d0e8bb1600347f17ab3d9172fff87a",
    "username": "unonuevo",
    "local_user_id": "05685392311d1d259fe01c65c7a6c28e"
}
```
being `facility` and `username` data of the remote server. If the remote server requires password to authenticate user, a "password" parameter must be added, otherwise it's not needed

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/plugins/user_profile/tasks.py
+++ b/kolibri/plugins/user_profile/tasks.py
@@ -58,6 +58,23 @@ class MergeUserValidator(PeerImportSingleSyncJobValidator):
     ],
 )
 def mergeuser(command, **kwargs):
+    """
+    This is an example of the POST payload to create this task:
+    {
+        "type": "kolibri.plugins.user_profile.tasks.mergeuser",
+        "baseurl": "http://192.168.0.201:80/",
+        "facility": "41d0e8bb1600347f17ab3d9172fff87a",
+        "username": "uno",
+        "local_user_id": "05685392311d1d259fe01c65c7a6c28e"
+    }
+    being baseurl, facility and username all parameters of the remote server.
+    If the remote server requires password to authenticate user,
+    a "password" parameter must be added, otherwise it's not needed.
+
+    If the username/password does not exist in the remote server,
+    this task will try to create the user.
+    """
+
     local_user_id = kwargs.pop("local_user_id")
     local_user = FacilityUser.objects.get(id=local_user_id)
     call_command(command, **kwargs)
--- a/kolibri/plugins/user_profile/tasks.py
+++ b/kolibri/plugins/user_profile/tasks.py
@@ -1,10 +1,15 @@
+from urllib.parse import urljoin
+
+import requests
 from django.core.management import call_command
+from django.core.urlresolvers import reverse
 from rest_framework import serializers
+from rest_framework.exceptions import AuthenticationFailed
+from rest_framework.status import HTTP_201_CREATED
 
 from kolibri.core.auth.models import FacilityUser
 from kolibri.core.auth.tasks import PeerImportSingleSyncJobValidator
 from kolibri.core.auth.utils.migrate import merge_users
-from kolibri.core.serializers import HexOnlyUUIDField
 from kolibri.core.tasks.decorators import register_task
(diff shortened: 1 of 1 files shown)


kolibri/plugins/user_profile/tasks.py
```diff
@@ -14,17 +18,57 @@
...
+        else:
+            baseurl = data["baseurl"]
+
+        facility = data["facility"]
+        user_data = {
+            "username": data["username"],
+            "password": data["password"],
+            "facility": facility,
+        }
+        create_user_url = urljoin(baseurl, reverse("kolibri:core:facilityuser-list"))
```
rtibbles: This should be using the new public signup viewset from this PR: https://github.com/learningequality/kolibri/pull/9363

kolibri/plugins/user_profile/tasks.py
```diff
@@ -14,17 +18,57 @@
...
+        auth_data = {
+            "username": remote_admin_username,
+            "password": remote_admin_password,
+        }
+        auth_url = urljoin(baseurl, reverse("kolibri:core:session-list"))
+        with requests.Session() as req:
+            req.post(auth_url, data=auth_data)
+            req.headers.update(
```
rtibbles: No need to add the CSRF token if using the public endpoint.

kolibri/plugins/user_profile/tasks.py
jredrejo: Ok, I was not aware of this. It's done, now this PR is blocked by #9363 . Also, this change makes this feature incompatible with previous versions of Kolibri on the server side.

rtibbles: Yes - that was intentional, as otherwise future changes in the internal API for the facility user endpoint might cause issues.

kolibri/plugins/user_profile/tasks.py
```diff
@@ -14,17 +18,54 @@
...
+            "facility": facility,
+        }
+        create_user_url = urljoin(baseurl, reverse("kolibri:core:facilityuser-list"))
+        auth_data = {
+            "username": remote_admin_username,
+            "password": remote_admin_password,
+        }
+        auth_url = urljoin(baseurl, reverse("kolibri:core:publicsignup-list"))
+        with requests.Session() as req:
+            req.post(auth_url, data=auth_data)
```
rtibbles: This is now signing up to the remote facility with the admin username and password - that's not what we want to do here.

We should be POSTing the new user data to the public signup URL, and not using the internal facility user API at all.

jredrejo: Ok, I had totally missed the purpose of the new api. 
The change was trivial. I only doubt if having two different endpoints to create/not create users is better. WIth this implementation it's the same endpoint. If the user does not exist in the remote server it tries to create it using the new api.

rtibbles (approved):

PR #9459, opened 2022-05-24
Update KDS to v1.3.1-beta0

marcellamaki:
Updates the reference to KDS to be the latest KDS beta


nucleogenesis (approved):
lgtm when the checks all pass, looks like snapshots need to be regenerated

PR #9460, opened 2022-05-24
Generate a content manifest file with exportcontent

dylanmccall:
## Summary

This change will update Kolibri's `exportcontent` command to generate a manifest file, describing an _intentional_ content selection associated with the exported data files. This is distinct from the user's provided `node_ids` and `exclude_node_ids`, because it needs to be reproducible on another Kolibri instance with a different set of available content. To achieve this, I split `get_import_export_data` into a few different functions, and added another function (`get_content_nodes_selectors`) which finds the optimal list of include nodes to describe the list of content being exported.

_That_ list of nodes is what is written to a `manifest.json` file alongside the exported content. For simplicity, it is possible for the same channel ID to be repeated multiple times, with different sets of include nodes and export nodes, in the same `manifest.json` file. This will only happen if the user runs `exportcontent` multiple times for the same channel.

This pull request does not include the `importcontent` counterpart to this work, which is in #9467.

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/core/content/utils/content_manifest.py
+++ b/kolibri/core/content/utils/content_manifest.py
@@ -1,0 +1,195 @@
+import hashlib
+import itertools
+import json
+import os
+from collections import namedtuple
+
+from kolibri.core.content.models import ChannelMetadata
+from kolibri.core.content.models import ContentNode
+
+try:
+    FileNotFoundError
+except NameError:
+    FileNotFoundError = IOError
+
+
+ContentManifestChannelData = namedtuple(
+    "ContentManifestChannelData", ["channel_id", "channel_version", "include_node_ids"]
+)
+
+
+class ContentManifestParseError(ValueError):
+    pass
+
+
+class ContentManifest(object):
+    """
+    A content manifest describes a selection of Kolibri content across
+    multiple channels. It is usually stored as a JSON file along with exported
+    content. It contains a list of channels, with their versions, and a list
+    of node IDs associated with each channel ID + version.
+
+    When adding content to the manifest, this implementation optimizes the list
+    of node IDs to contain the smallest possible number of topic nodes, instead
+    of every included node. When nodes are added to a channel ID and version
+    that is already in the content manifest, those nodes will be added to the
+    existing list.
+    """
+
+    def __init__(self):
+        self._channels_dict = {}
+
+    def read(self, filenames):
+        if not isinstance(filenames, list):
+            filenames = [filenames]
+
+        files_read = []
+
+        for filename in filenames:
+            try:
+                with open(filename, "r") as fp:
+                    self.read_file(fp)
+            except FileNotFoundError:
+                pass
+            else:
+                files_read.append(filename)
+
+        return files_read
+
+    def read_file(self, fp):
+        json_str = fp.read()
+        if not json_str:
+            return
+        # Raises JSONDecodeError if the file is invalid
+        manifest_data = json.loads(json_str)
+        self.read_dict(manifest_data)
+
+    def read_dict(self, manifest_data):
+        for channel_data in manifest_data.get("channels", []):
+            channel_id = channel_data.get("id", None)
+            channel_version = channel_data.get("version", None)
+            include_node_ids = channel_data.get("include_node_ids", [])
+            if channel_id is None or channel_version is None:
+                raise ContentManifestParseError("id and version are required fields")
+            self._update_channel_data(channel_id, channel_version, include_node_ids)
+
+    def write(self, path):
+        os.makedirs(os.path.dirname(path), exist_ok=True)
+        with open(path, "w") as fp:
+            self.write_file(fp)
+
+    def write_file(self, fp):
+        json.dump(self.to_dict(), fp, indent=4)
+
+    def to_dict(self):
+        channels_list = list(self._iter_channel_dicts())
+        return {
+            "channels": channels_list,
+            "channel_list_hash": hashlib.md5(
+                json.dumps(channels_list, sort_keys=True).encode()
+            ).hexdigest(),
+        }
+
+    def _iter_channel_dicts(self):
+        for channel_data in self._iter_channel_data():
+            yield {
+                "id": channel_data.channel_id,
+                "version": channel_data.channel_version,
+                "include_node_ids": sorted(channel_data.include_node_ids),
+            }
+
+    def _iter_channel_data(self):
+        for channel_id in self.get_channel_ids():
+            for channel_version in self.get_channel_versions(channel_id):
+                yield self.get_channel_data(channel_id, channel_version)
+
+    def get_channel_ids(self):
+        return self._channels_dict.keys()
+
+    def get_channel_versions(self, channel_id):
+        return self._channels_dict.get(channel_id, {}).keys()
+
+    def get_channel_data(self, channel_id, channel_version):
+        channel_data = self._channels_dict.get(channel_id, {}).get(
+            channel_version, None
+        )
+        if channel_data is None:
+            channel_data = ContentManifestChannelData(None, None, set())
+        return channel_data
+
+    def get_include_node_ids(self, channel_id, channel_version):
+        channel_data = self.get_channel_data(channel_id, channel_version)
+        return channel_data.include_node_ids
+
+    def add_content_nodes(self, channel_id, channel_version, nodes_queries_list):
+        include_node_ids = get_content_nodes_selectors(
+            channel_id, channel_version, nodes_queries_list
+        )
+        self._update_channel_data(channel_id, channel_version, include_node_ids)
+
+    def _update_channel_data(self, channel_id, channel_version, include_node_ids):
+        old_include_node_ids = self.get_include_node_ids(channel_id, channel_version)
+        channel_data = ContentManifestChannelData(
+            channel_id,
+            channel_version,
+            set(old_include_node_ids) | set(include_node_ids),
+        )
+        self._set_channel_data(channel_data)
+
+    def _set_channel_data(self, channel_data):
+        assert isinstance(channel_data, ContentManifestChannelData)
+        self._channels_dict.setdefault(channel_data.channel_id, {})[
+            channel_data.channel_version
+        ] = channel_data
+
+
+def get_content_nodes_selectors(channel_id, channel_version, nodes_queries_list):
+    """
+    Returns a set of include_node_ids that can be used with the given
+    channel_id to import only the nodes contained in nodes_queries_list.
+    """
+
+    include_node_ids = set()
+
+    channel_metadata = ChannelMetadata.objects.get(
+        id=channel_id, version=channel_version
+    )
+
+    available_node_ids = set(
+        itertools.chain.from_iterable(
+            nodes_query.values_list("id", flat=True)
+            for nodes_query in nodes_queries_list
+        )
+    )
+
+    available_nodes_queue = [channel_metadata.root]
+
+    while len(available_nodes_queue) > 0:
+        node = available_nodes_queue.pop(0)
+
+        # We could add nodes to exclude_node_ids when less than half of the
+        # sibling nodes are missing. However, it is unclear if this would
+        # be useful.
+
+        if node.kind == "topic":
+            leaf_node_ids = _get_leaf_node_ids(node)
+            matching_leaf_nodes = leaf_node_ids.intersection(available_node_ids)
+            missing_leaf_nodes = leaf_node_ids.difference(available_node_ids)
+            if len(missing_leaf_nodes) == 0:
+                include_node_ids.add(node.id)
+            elif len(matching_leaf_nodes) > 0:
+                available_nodes_queue.extend(node.children.all())
+        elif node.id in available_node_ids:
+            include_node_ids.add(node.id)
+
+    return include_node_ids
+
+
+def _get_leaf_node_ids(node):
+    return set(
+        ContentNode.objects.filter(
+            lft__gte=node.lft, lft__lte=node.rght, channel_id=node.channel_id
+        )
+        .exclude(kind="topic")
+        .values_list("id", flat=True)
+    )
--- a/kolibri/core/content/utils/content_manifest.py
+++ b/kolibri/core/content/utils/content_manifest.py
@@ -47,7 +47,7 @@ class ContentManifest(object):
     def __init__(self):
         self._channels_dict = {}
 
-    def read(self, filenames):
+    def read(self, filenames, validate=False):
         """
         Read content manifest data from a file path, or a list of file paths.
         Returns the number of files which have been read successfully. Note
@@ -64,7 +64,7 @@ class ContentManifest(object):
         for filename in filenames:
             try:
                 with open(filename, "r") as fp:
-                    self.read_file(fp)
+                    self.read_file(fp, validate=validate)
             except FileNotFoundError:
                 pass
             else:
@@ -72,7 +72,7 @@ class ContentManifest(object):
 
         return files_read
 
-    def read_file(self, fp):
+    def read_file(self, fp, validate=False):
         """
         Read content manifest data from a file-like object.
         Raises `ContentManifestParseError` if a file is invalid JSON, or has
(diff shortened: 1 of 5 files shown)


dylanmccall:
Getting this to play nice with `exportcontent` is a bit trickier than it sounds at first because we need to deal with the tool being run multiple times, which means we need to generate a sane set of include_nodes / exclude_nodes, and then _update_ that set as needed. After some poking and prodding today, the solution I have for now is to append a new `{channel id, include nodes, exclude nodes}` to the manifest file for every run. When I'm working on `importcontent` I think I'll be able to add a mechanism that extends the existing selection for a channel as long as it is the same version, but I think it's reasonable to expect the channel to appear multiple times in some cases.

kolibri/core/content/utils/import_export_content.py
```diff
@@ -170,52 +210,160 @@ def get_import_export_data(  # noqa: C901
...
-
-        min_boundary += dynamic_chunksize
-
     files_to_download = list(queried_file_objects.values())
 
     total_bytes_to_transfer = sum(map(lambda x: x["file_size"] or 0, files_to_download))
     return number_of_resources, files_to_download, total_bytes_to_transfer
 
 
+def get_content_nodes_selectors(channel_id, nodes_queries_list):
```
rtibbles: I feel like there's a more efficient way to do this with a hierarchical aggregation, similar to how we do our annotation logic. I will try to sketch out what that might look like and get back to you here.

dylanmccall: Yeah, I think you're right that there's probably a better way. I opted for the simple to read, procedural approach since it's part of a one-off task that users already understand will take a length of time measured in minutes (so a few seconds generating a manifest isn't a big deal as long as we aren't pegging their CPU to do it). I think it's more important here that the code is clearly understandable and somebody can optimize it once it has been proved out.

But fetching all the node IDs into memory is certainly _ugly_. For Khan Academy, for example, that's a really big query and about 800 kB of memory for our set of `available_node_ids`. I'm sure we could do something clever involving the `lft` / `rft` of leaf nodes that are excluded, bubbling up from them to select the correct topics. There's a danger that it could turn into _a lot_ of queries for a problem where we're really just shuffling primary keys around. So while it would involve less duplicate data in memory, I don't think we're guaranteed that it would be more efficient, particularly for the sort of deployment where someone would be using `exportcontent`. Worth exploring in a branch?

rtibbles: We do something very similar in our importability annotation https://github.com/learningequality/kolibri/blob/develop/kolibri/core/content/utils/importability_annotation.py#L36 - the main difference here would be instead of doing annotation based on `some`, we would annotate on `all` children being available. Otherwise the code would be nearly identical. We can then descend down the tree in a read query until we have a set of ids that are marked as 'all' available.

The main trick here is that the write and read all happen within a transaction, and then we roll it back at the end.

kolibri/core/content/utils/import_export_content.py
```diff
@@ -31,6 +34,11 @@
 
     SSLERROR = requests.exceptions.SSLError
 
+try:
+    FileNotFoundError
+except NameError:
+    FileNotFoundError = IOError
```
manuq: :+1: I guess this is to be compliant with older python.

kolibri/core/content/utils/import_export_content.py
```diff
@@ -170,52 +210,160 @@ def get_import_export_data(  # noqa: C901
...
+    number_of_resources = 0
+
+    for nodes_query in nodes_queries_list:
+        number_of_resources = number_of_resources + nodes_query.count()
+
+        file_objects = LocalFile.objects.filter(
+            files__contentnode__in=nodes_query
+        ).values("id", "file_size", "extension")
+        if available is not None:
+            file_objects = file_objects.filter(available=available)
```
manuq, line 222: :+1: although this could be merged with the previous `.filter()`, querysets are lazy so they can be chained like this and it's more readable.

dylanmccall:
Okay, I added a test suite here, which should give us a nice way to measure performance, although the current test fixture isn't quite doing that. (It's a very small tree of content). In general, when I say this is slow I really just mean "it isn't trying very hard to be fast," but (anecdotally) it is acceptable with the channels I've tried it against. In the worst case I've had it spend ~1.5 seconds generating a manifest file for Khan Academy, which is a particularly large channel. It could be faster with some added cleverness, but I don't think we need to block on it here - and the nice thing is the test suite should make it easier to add such cleverness in the future without breaking existing behaviour :)

Here's a handy little script to play with this in isolation:

```python
from kolibri.utils.main import initialize
initialize(skip_update=True)

import json

# Constants to know:
# Khan Academy (c9d7f950ab6b5a1199e3d6c10d7f0103)

from kolibri.core.content.utils.import_export_content import get_import_export_nodes
from kolibri.core.content.utils.import_export_content import get_content_nodes_selectors

KHAN_ACADEMY="c9d7f950ab6b5a1199e3d6c10d7f0103"

nodes_segments = get_import_export_nodes(KHAN_ACADEMY, available=True)
content_selection = get_content_nodes_selectors(KHAN_ACADEMY, nodes_segments)

print("Content selection:", json.dumps(content_selection, indent=4))
```

dylanmccall:
I did a couple tweaks here, moving some stuff over from #9467:

- In `get_import_export_nodes`, differentiate between `node_ids=None` (all node IDs) and `node_ids=[]` (no node IDs). This required changes to a bunch of tests which were using an empty array to mean "all the nodes from this channel", but I think it makes sense. Kolibri's annotation code appears to be making the same distinction: https://github.com/learningequality/kolibri/blob/develop/kolibri/core/content/utils/annotation.py#L207-L228.
- When exporting all nodes in a channel, `include_node_ids` in the manifest file will be set to `[channel_id]`. In a previous version, I had it output an empty list instead.
- With these changes together, we have the ability to clearly express both a channel with no content and a channel with all of the content in our manifest files. In theory, that means we can have `exportchannel` generate a manifest file as well.

rtibbles:
> In get_import_export_nodes, differentiate between node_ids=None (all node IDs) and node_ids=[] (no node IDs). This required changes to a bunch of tests which were using an empty array to mean "all the nodes from this channel", but I think it makes sense. Kolibri's annotation code appears to be making the same distinction: https://github.com/learningequality/kolibri/blob/develop/kolibri/core/content/utils/annotation.py#L207-L228.

I think that's generally helpful, and may be the cause of some behavioural errors I've witnessed elsewhere in Kolibri develop.

> When exporting all nodes in a channel, include_node_ids in the manifest file will be set to [channel_id]. In a previous version, I had it output an empty list instead.

I think for precision this should be `[root_node_id]` - they are frequently identical, but there are some cases where they are not.

> With these changes together, we have the ability to clearly express both a channel with no content and a channel with all of the content in our manifest files. In theory, that means we can have exportchannel generate a manifest file as well.

This would also seem fine.

dylanmccall:
> I think for precision this should be [root_node_id] - they are frequently identical, but there are some cases where they are not.

Ah, good point. That is in fact what it's doing - I didn't realize they could be different, so that's useful to know! In a previous version, it was special-casing this by saying "if the list we end up with is `[channel_id]` then make it `[]`", which was silly, so now it just isn't doing that special case :)

kolibri/core/content/utils/import_export_content.py
```diff
@@ -155,52 +200,157 @@ def get_import_export_data(  # noqa: C901
...
+    except (FileNotFoundError, ValueError):
+        # Use ValueError rather than JSONDecodeError for Py2 compatibility
+        manifest_data = None
+
+    if not isinstance(manifest_data, dict):
+        manifest_data = {}
+
+    channels = manifest_data.setdefault("channels", [])
+
+    # TODO: If the channel is already listed, it would be nice to merge the
```
rtibbles: So, in the case the channel is already in the manifest what happens, do we have two entries for the same channel in the manifest?

rtibbles (commented):
All seems reasonable on a read through, and the tests and generation time for KA gives assurance that this is a reasonable approach.

One question about the merge case.

kolibri/core/content/utils/import_export_content.py
rtibbles: Flagging that we can't merge this with the previous filter, as application of the filter is conditional on `available` being set to a boolean value rather than `None`.

kolibri/core/content/utils/import_export_content.py
dylanmccall: Yeah, exactly. So it could end up with a manifest file like this:

```
{
    "channels": [
        {
            "id": "e409b964366a59219c148f2aaa741f43",
            "version": 10,
            "include_node_ids": [
                "6277aa0c44235435acdc8a9ed98f466b"
            ],
            "exclude_node_ids": []
        },
        {
            "id": "e409b964366a59219c148f2aaa741f43",
            "version": 10,
            "include_node_ids": [
                "e409b964366a59219c148f2aaa741f43"
            ],
            "exclude_node_ids": []
        }
    ],
    "channel_list_hash": "a2c59ee9bab4586a78e56ca8cf13a912"
}
```

Where the first time around, I exported just "6277aa0c44235435acdc8a9ed98f466b", and the second time I exported the entire channel.

Note that identical entries are skipped, but if we have a different selection for the same channel, they each get a new entry. It would be better if it merges them more exhaustively. For instance the second entry is a superset of the first entry. So, I added a TODO about that. Could do that by adding the existing entries to our list of selected nodes.

Come to think of it, that doesn't sound too difficult to implement, so I think I'll give it a shot right now over in #9467 :)

rtibbles: Yes - I was wondering if we could just do the TODO now, and not leave it for later!

kolibri/core/content/utils/import_export_content.py
dylanmccall: Yeah, this should be the same as the code from [`get_import_export_data`](https://github.com/learningequality/kolibri/blob/develop/kolibri/core/content/utils/import_export_content.py#L160-L201) in the develop branch, except the indentation changed slightly :)

kolibri/core/content/utils/import_export_content.py
dylanmccall: I think so, indeed. Ran out of time today, but I'll let you know how it goes tomorrow!

dylanmccall: Okay, I added a commit with a couple things:

- Removed `exclude_node_ids` from manifest.json. It isn't being used right now, and it makes our life easier if it isn't there. If we are only _including_ nodes, we can trivially extend that set of nodes when a new export happens. Otherwise (if there is the possibility of nodes being excluded), we would have to expand the selection into a complete set of nodes and then simplify it again.
- Added a `content_manifest` module to deal with this in one place, moving some file access and json parsing code out of import_export_content.py. It is responsible for reading and writing manifest files, as well as simplifying the provided set of node IDs.

kolibri/core/content/utils/content_manifest.py
```diff
@@ -0,0 +1,195 @@
...
+            else:
+                files_read.append(filename)
+
+        return files_read
+
+    def read_file(self, fp):
+        json_str = fp.read()
+        if not json_str:
+            return
+        # Raises JSONDecodeError if the file is invalid
```
rtibbles: Note in Python 2 this raises a `ValueError`, of which `JSONDecodeError` is a subclass.

kolibri/core/content/utils/content_manifest.py
```diff
@@ -0,0 +1,195 @@
...
+        manifest_data = json.loads(json_str)
```
rtibbles: Any reason not to just do a try...except here, catch any errors and just do `json.load` directly from the `fp` object?

This would also have the benefit of handling non-empty, but invalid JSON.

kolibri/core/content/utils/import_export_content.py
```diff
@@ -97,15 +97,56 @@ def filter_by_file_availability(nodes_to_include, channel_id, drive_id, peer_id)
 
 def get_import_export_data(  # noqa: C901
```
rtibbles: Guessing we can probably remove the `noqa` from here now?

kolibri/core/content/utils/content_manifest.py
```diff
@@ -0,0 +1,195 @@
...
+        # be useful.
+
+        if node.kind == "topic":
+            leaf_node_ids = _get_leaf_node_ids(node)
+            matching_leaf_nodes = leaf_node_ids.intersection(available_node_ids)
+            missing_leaf_nodes = leaf_node_ids.difference(available_node_ids)
+            if len(missing_leaf_nodes) == 0:
+                include_node_ids.add(node.id)
+            elif len(matching_leaf_nodes) > 0:
+                available_nodes_queue.extend(node.children.all())
```
rtibbles: Should we be optionally filtering the node children by an `available` flag here (similarly to how we do in the import export data function).

rtibbles (commented):
A couple of questions - nothing that is completely blocking, but I think it would be good to handle malformed JSON in the manifest reading, and I wonder if optional filtering by available might be useful.

Beyond that - some doc strings on the methods of the ContentManifest class would be helpful - especially to distinguish the role of different but somewhat similarly named methods.

kolibri/core/content/utils/import_export_content.py
dylanmccall: :+1:  Indeed we can!

kolibri/core/content/utils/content_manifest.py
dylanmccall: Oh yeah. I made this clearer along with my fix for the below comment, defining `JSONDecodeError = ValueError` as a fallback.

kolibri/core/content/utils/content_manifest.py
dylanmccall: Yeah, that's funny it ended up looking like this :b Doesn't seem to be any reason, so I switched it to use `json.load(fp)`.

kolibri/core/content/utils/content_manifest.py
dylanmccall: Oh, yeah, I think that's right. Assuming the availability of nodes in the tree is correct, this should be doing `node.children.filter(available=True)`. Unavailable nodes aren't going to resolve to anything useful here so it's just wasting memory. I should make sure the tests actually cover this properly first (might need to adopt that channel builder instead of the questionable database fixture :b), so, I'll do that on Monday.

I see what you mean about passing an `available=` flag to the function to be consistent with the import export data function, although I should tweak some names in that case, and probably wrap my head about the use case there. It does _sound_ like it would be convenient for something.

rtibbles: Yeah, looking at how the function is used, I'm not sure there is a need to have optional available filtering. It's really just that it can be filtered by available=True to limit the search space.

The only question in my mind is.... does the manifest handle the case where a ContentNode has the files it needs in the exported directory, but is not available on the machine doing the exporting?

So, for example, if I exported the basic arithmetic topic from Khan Academy from my machine, and it generated a content manifest - I then pass that USB key to you, and you export more of Khan Academy from your machine, but you don't have basic arithmetic on yours - is that going to now be excluded from the updated manifest?

Would be good to make a test case to cover this scenario, I think.

rtibbles (commented):
I think this is nearly ready to go, I just have one more question that I couldn't be sure of by looking at the code and test cases.

kolibri/core/content/utils/content_manifest.py
dylanmccall: Okay, I added some test cases which explore that. Specifically:

- If we add some new content nodes to a ContentManifest, `include_node_ids` is _extended_ to include those nodes. Any nodes that were listed previously are still there.
- If content nodes are added for one channel version, and then for a different version of the same channel, the two versions are separated; the lists of `include_node_ids` are not merged. (In practice, importcontent merges them in #9467, albeit with a warning message).

This behaviour means a bit of duplication in some cases. That can be solved, but it adds some complexity - especially if we start thinking about channel versions - so I think just something to be aware of for the moment.

dylanmccall:
Funky build failure from the CI system, but the tests passed :)

rtibbles:
Oh yes, the buildkite error was due to a sysops issue with the builder.

rtibbles (approved):

PR #9463, opened 2022-05-27
Content dir argument

danigm:
## Summary

Add `--content_dir` option to the `importchannel` and `importcontent` commands.

This will allow to update content in an external device when running the application with the `CONTENT_FALLBACK_DIRS` configuration.

### DIFF
--- a/kolibri/core/content/management/commands/importcontent.py
+++ b/kolibri/core/content/management/commands/importcontent.py
@@ -159,6 +159,12 @@ class Command(AsyncCommand):
             dest="timeout",
             help="Specify network timeout in seconds (default: %(default)d)",
         )
+        network_subparser.add_argument(
+            "--content_dir",
+            type=str,
+            default=paths.get_content_dir_path(),
+            help="Download the content to the given content dir.",
+        )
 
         disk_subparser = subparsers.add_parser(
             name="disk", cmd=self, help="Copy the content from the given folder."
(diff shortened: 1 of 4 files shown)


kolibri/core/content/management/commands/importcontent.py
```diff
@@ -377,7 +397,9 @@ def _transfer(  # noqa: max-complexity=16
                             f = files_to_download.pop()
                             filename = get_content_file_name(f)
                             try:
-                                dest = paths.get_content_storage_file_path(filename)
+                                dest = paths.get_content_storage_file_path(
+                                    filename, contentfolder=content_dir
```
manuq, line 401: Nit: maybe it's better to use the same variable name everywhere, eg renaming the new CLI parameter to `--content-folder`. Or do you prefer to use "content dir" as the exposed name and "contentfolder" as the internal name?

danigm: `CONTENT_DIR` is the configuration option and it's used in other places, so I think it's better to use `content-dir` for this option.

rtibbles:
1 test needs an update due to the change in function signature breaking an assertion call.

Linting is failing - I'd suggest installing `pre-commit` to prevent linting errors in future.

danigm:
> 1 test needs an update due to the change in function signature breaking an assertion call.
> 
> Linting is failing - I'd suggest installing `pre-commit` to prevent linting errors in future.

I've just fixed the linting issues, sorry for that, I didn't have the `pre-commit` enabled. I've enabled it now for future PRs.

rtibbles (approved):

PR #9466, opened 2022-05-31
Remote content endpoint streaming

### LINKED ISSUE #9386: Create a /content/storage endpoint that will proxy to a remote server
An API endpoint setup under `/remote/content/storage` that will proxy to a specific remote server using the get param `baseurl` specified. If no `baseurl` is specified, it should return a 404.

The API endpoint will then make a further request to `<baseurl>/content/storage/....` and then stream the response from that request as the response to the request being made from the client.

To intelligently cache these resources, this endpoint should first defer to Kolibri's whitenoise implementation to serve up the file - if that returns a 404, then it should attempt to make the request described above.

If the request is successful, and disk space is available and writable, the resource should be cached in the local Kolibri content storage.

This is probably best implemented within the core Kolibri whitenoise implementation, so that it can be used in both the main wsgi and alternate origin wsgi servers.

### LINKED ISSUE #9388: Handle range requests on remote content serving
To better support video streaming from remote servers, we should support proxying range requests, and caching the results of those range requests (where eventually with enough range requests, we'll have a whole file!).

rtibbles:
## Summary
* Moves the file transfer utilities into the `kolibri.utils` namespace so that they can be imported without importing the Django stack
* Makes some updates to the file transfer utilities to allow them to be used solely via `next` and not just via explicit iteration in a loop.
* Catches errors that are thrown when a file transfer attempts to move to its destination, but the source transfer file has disappeared, if the destination file already exists, it silently swallows the error - this allows for multiple simultaneous transfers of the same file to co-exist and not stomp each other.
* Adds additional functionality to the DynamicWhitenoise class, whereby it will accept a `baseurl` GET parameter, which, if the file does not exist on disk, will attempt to fetch the file using the `FileDownload` file transfer class.
* As the file is downloading, the file download is wrapped in a special filelike wrapper object that streams the download into the response.
* This wrapper object supports `seek` in order to support range requests, but still downloads the entire file, as it seems complex to construct the full file out of multiple potentially non-sequential range requests.

Note: Cleaning up the downloaded files is deferred to a follow up issue!

## References
Fixes #9386
Fixes #9388

## Reviewer guidance
Test this out by streaming some files directly from Studio, e.g.:
http://127.0.0.1:8000/content/storage/1/1/11e1c2c83b9dd46e584d75f484386021.mp4?baseurl=https://studio.learningequality.org
http://127.0.0.1:8000/content/storage/9/d/9d523f8a736e372fbe10ac895085fce9.png?baseurl=https://studio.learningequality.org
http://127.0.0.1:8000/content/storage/8/a/8a9c9011e527c0ebf9b09d642eba5f78.pdf?baseurl=https://studio.learningequality.org

Confirm that reloading the file also works.
Delete the file from storage and try the request again, confirm that it works.

Question: Is the range request implementation good enough?

----

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/utils/file_transfer.py
+++ b/kolibri/utils/file_transfer.py
@@ -4,8 +4,21 @@ import shutil
 from time import sleep
 
 import requests
+from requests.exceptions import ChunkedEncodingError
+from requests.exceptions import ConnectionError
+from requests.exceptions import HTTPError
+from requests.exceptions import Timeout
 
-from kolibri.core.content.utils.import_export_content import retry_import
+
+try:
+    import OpenSSL
+
+    SSLERROR = OpenSSL.SSL.Error
+except ImportError:
+    SSLERROR = requests.exceptions.SSLError
+
+RETRY_STATUS_CODE = [502, 503, 504, 521, 522, 523, 524]
+
 
 logger = logging.getLogger(__name__)
 
@@ -26,6 +39,29 @@ class TransferNotYetClosed(Exception):
     pass
 
 
+def retry_import(e):
+    """
+    When an exception occurs during channel/content import, if
+        * there is an Internet connection error or timeout error,
+          or HTTPError where the error code is one of the RETRY_STATUS_CODE,
+          return return True to retry the file transfer
+    return value:
+        * True - needs retry.
+        * False - Does not need retry.
+    """
+
+    if (
+        isinstance(e, ConnectionError)
+        or isinstance(e, Timeout)
+        or isinstance(e, ChunkedEncodingError)
+        or (isinstance(e, HTTPError) and e.response.status_code in RETRY_STATUS_CODE)
+        or (isinstance(e, SSLERROR) and "decryption failed or bad record mac" in str(e))
+    ):
+        return True
+
+    return False
+
+
 class Transfer(object):
     DEFAULT_TIMEOUT = 60
 
@@ -36,7 +72,7 @@ class Transfer(object):
         block_size=2097152,
         remove_existing_temp_file=True,
         timeout=DEFAULT_TIMEOUT,
-        cancel_check=None,
+        cancel_check=lambda: False,
     ):
         self.source = source
         self.dest = dest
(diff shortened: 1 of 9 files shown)


kolibri/utils/file_transfer.py
```diff
@@ -36,7 +72,7 @@ def __init__(
         block_size=2097152,
         remove_existing_temp_file=True,
         timeout=DEFAULT_TIMEOUT,
-        cancel_check=None,
+        cancel_check=lambda: False,
```
jredrejo: what's the advantage of this change in the default value? I only see cancel_check being used in `if` sentences, so None works fine there too

kolibri/utils/kolibri_whitenoise.py
```diff
@@ -182,6 +296,12 @@ def find_and_cache_dynamic_file(self, url):
                     except (IOError, OSError):
                         pass
                 self.add_file_to_dictionary(url, path, stat_cache=stat_cache)
+        elif (
+            remote_baseurl is not None
+            and self.writable_check is not None
+            and self.writable_check.match(url)
+        ):
+            self.files[url] = self.get_streaming_static_file(url, remote_baseurl)
```
jredrejo, line 304: this is really nice

kolibri/utils/file_transfer.py
```diff
@@ -107,7 +157,11 @@ def next(self):
         return chunk
 
     def _move_tmp_to_dest(self):
-        shutil.move(self.dest_tmp, self.dest)
+        try:
+            shutil.move(self.dest_tmp, self.dest)
+        except (IOError, OSError) as e:
```
jredrejo: While we keep using Python 2.7 I like to add at the beginning of the modules having code opening files, this:
```
try:
    FileNotFoundError
except NameError:
    FileNotFoundError = IOError
```
and use `FileNotFound` as a different error catching, what can be more informative in the places where  you're using `IOError` when python 3 is used

jredrejo (approved):
Code looks good to me, and I haven't found any problem while testing it.
I've added a couple of comments, but  none of them are a blocker, they can be used or discarded without any problem.

kolibri/utils/file_transfer.py
rtibbles: I could have changed it to be `if cancel_check and cancel_check()` where it is used - I can update it to that, I was just being slightly lazy.

kolibri/utils/file_transfer.py
rtibbles: Yeah, that's a nice practice, will update.

PR #9467, opened 2022-05-31
Read options from a manifest file in importcontent

### LINKED ISSUE #6067: tracking of content resource visibility in exported channels
# Observed behavior

In the current work on 0.13.0 content management, we're beginning to treat duplicate resources 
individually, so that the user can import or delete them separately. See for example:

* https://github.com/learningequality/kolibri/issues/4082
* https://github.com/learningequality/kolibri/issues/4549

This new 'visibility' tracking information is stored in the Kolibri database, but does not get persisted when a user exports the content to disk.

### Expected behavior

Users may expect content exported to disk to retain the visibility settings, if the export process was intended as a form of curation. While this might not always be true, based on user feedback so far that seems to be the dominant expectation.

Because content databases are immutable, in order to implement this, we would need to add some kind of manifest file that travels along with the content and database file and specifies visibility. This could either be a plain-text file, JSON, or a SQLite DB, for example.

### User-facing consequences

Intentional user decisions made during the "export as curation" process are lost.

### Context

0.13.0 alpha 0

dylanmccall:
## Summary

This adds a `--manifest` option to the `importcontent` management command.

With this change, one can export content from one Kolibri instance…

```
kolibri manage exportcontent c9d7f950ab6b5a1199e3d6c10d7f0103 my_export_dir
```

And then import it in another Kolibri instance…

```
kolbri manage importchannel network c9d7f950ab6b5a1199e3d6c10d7f0103
kolibri manage importcontent --manifest=my_export_dir/content/manifest.json disk c9d7f950ab6b5a1199e3d6c10d7f0103
```

Or…

```
kolbri manage importchannel network c9d7f950ab6b5a1199e3d6c10d7f0103
kolibri manage importcontent disk c9d7f950ab6b5a1199e3d6c10d7f0103 my_export_dir
```

That second option changes some existing behaviour: it automatically detects if there is a file named `manifest.json` in the path we are importing from, and uses _that_ for the default `node_ids` and `exclude_node_ids` (instead of the previous behaviour where it would attempt to import everything). This should be the same in practice, as `manifest.json` should select all of the content which has been exported. I add a `--no_detect_manifest` option to preserve the previous behaviour.

## References

This is a counterpart to my changes in #9460, adding support for _exporting_ a manifest file.

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/core/content/management/commands/importcontent.py
+++ b/kolibri/core/content/management/commands/importcontent.py
@@ -294,7 +294,7 @@ class Command(AsyncCommand):
         if manifest_file:
             content_manifest.read_file(manifest_file)
             use_content_manifest = True
-        elif path and detect_manifest and not (node_ids or exclude_node_ids):
+        elif path and detect_manifest and node_ids is None and exclude_node_ids is None:
             manifest_path = os.path.join(path, "content", "manifest.json")
             if content_manifest.read(manifest_path):
                 use_content_manifest = True
@@ -627,7 +627,9 @@ class Command(AsyncCommand):
         return FILE_TRANSFERRED, data_transferred
 
     def handle_async(self, *args, **options):
-        if options["manifest"] and (options["node_ids"] or options["exclude_node_ids"]):
+        if options["manifest"] and (
+            options["node_ids"] is not None or options["exclude_node_ids"] is not None
+        ):
             raise CommandError(
                 "The --manifest option must not be combined with --node_ids or --exclude_node_ids."
             )
--- a/kolibri/core/content/management/commands/importcontent.py
+++ b/kolibri/core/content/management/commands/importcontent.py
@@ -92,7 +92,7 @@ class Command(AsyncCommand):
             "--node_ids",
             "-n",
             # Split the comma separated string we get, into a list of strings
-            type=lambda x: x.split(","),
+            type=lambda x: x.split(",") if x else [],
             default=None,
             required=False,
             dest="node_ids",
@@ -109,7 +109,7 @@ class Command(AsyncCommand):
         parser.add_argument(
             "--exclude_node_ids",
             # Split the comma separated string we get, into a list of string
-            type=lambda x: x.split(","),
+            type=lambda x: x.split(",") if x else [],
             default=None,
             required=False,
             dest="exclude_node_ids",
--- a/kolibri/core/content/management/commands/importcontent.py
+++ b/kolibri/core/content/management/commands/importcontent.py
@@ -73,7 +73,6 @@ class Command(AsyncCommand):
         """
         parser.add_argument(
             "--manifest",
-            # Split the comma separated string we get, into a list of strings
             type=argparse.FileType("r"),
             default=None,
             required=False,
@@ -108,7 +107,7 @@ class Command(AsyncCommand):
         """
         parser.add_argument(
             "--exclude_node_ids",
-            # Split the comma separated string we get, into a list of string
+            # Split the comma separated string we get, into a list of strings
             type=lambda x: x.split(",") if x else [],
             default=None,
             required=False,
@@ -285,7 +284,8 @@ class Command(AsyncCommand):
             # If manifest_file is stdin, its name will be "<stdin>" and path
             # will become "". This feels clumsy, but the resulting behaviour
             # is reasonable.
-            manifest_dir = os.path.dirname(manifest_file.name)
+            manifest_file_name = getattr(manifest_file, "name", "")
+            manifest_dir = os.path.dirname(manifest_file_name)
             path = os.path.dirname(manifest_dir)
 
         content_manifest = ContentManifest()
--- a/kolibri/core/content/management/commands/importcontent.py
+++ b/kolibri/core/content/management/commands/importcontent.py
@@ -1,3 +1,4 @@
+import argparse
 import concurrent.futures
 import logging
 import os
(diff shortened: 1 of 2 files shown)


kolibri/core/content/utils/import_export_content.py
```diff
@@ -185,7 +185,7 @@ def get_import_export_nodes(  # noqa: C901
             nodes_query = nodes_query.filter(renderable_contentnodes_q_filter)
 
         # filter down the query to remove files associated with nodes we've specifically been asked to exclude
-        if exclude_node_ids:
+        if exclude_node_ids is not None:
```
danigm: The `--node_ids` and `--exclude_node_ids` defaults to `[]` in the `exportcontent.py` and `importcontent.py` commands, so it will never be `None` when calling from there. I don't know if I'm missing something, but when I test the export command with `exclude_node_ids` option, I don't get anything, because the `nodes_query` is empty, because in the previous `if`, the filter is done with an empty `node_ids` list.

kolibri/core/content/management/commands/importcontent.py
```diff
@@ -565,6 +630,16 @@ def _start_file_transfer(self, f, filetransfer):
         return FILE_TRANSFERRED, data_transferred
 
     def handle_async(self, *args, **options):
+        if not options["directory"] and not options["manifest"]:
+            raise CommandError(
+                "Either a directory or a manifest file must be provided."
+            )
```
danigm: This makes impossible to import from network without a manifest. And there's an exception because `options["directory"]` is not an existing option when importing from network

kolibri/core/content/utils/import_export_content.py
dylanmccall: Oh yeah, this commit seems rather suspect. Thanks for pointing it out! If we want to select a whole channel, `exportcontent` should just set `node_ids` to the channel ID (the root node), so this case where node_ids could be None doesn't make sense and it's better if we treat None and `[]` as equivalent :)

I'll revert it, and add some unit tests to make sure things work as needed.

dylanmccall: Never mind, this made sense in the end. I did need to change some tests, but this ended up a useful distinction for the case where a manifest file needs to indicate that there is no content. Just needed a bit more documentation :)

So what's happening here is, if you run `importcontent` without specifying any node_ids, this function gets called with `node_ids=None`. That does the right thing and matches with all nodes in the channel. There was some funny behaviour before, where if you do `importcontent --node_ids=''`, this function ends up being called with `node_ids=['']`, which also does the right thing (it matches with no nodes in the channel), but in the wrong way (it only matches because after an exhaustive search we have concluded there's no node with ID `''`). But that's great, because it means cleaning this up so `None` and `[]` are different doesn't change the behaviour!

So in the end I didn't revert this, but I added an explanatory comment in `get_import_export_nodes`, and I added some unit tests for the specific cases where `node_ids=[]` and `node_ids=None`.

kolibri/core/content/management/commands/importcontent.py
dylanmccall: Good spotting! That's fixed in the newest push.

kolibri/core/content/management/commands/importcontent.py
```diff
@@ -61,6 +63,24 @@ def add_arguments(self, parser):
...
+        manifest_help_text = """
+        Specify a path to a manifest file. Content specified in this manifest file will be imported.
+
+        e.g.
+
+        kolibri manage importcontent --manifest /path/to/KOLIBRI_DATA/content/manifest.json disk
+        """
+        parser.add_argument(
+            "--manifest",
+            # Split the comma separated string we get, into a list of strings
```
rtibbles: This code comment seems like it was a copy paste from below?

rtibbles (changes requested):
I don't see any issues on the face of the code, but it would be good to add some test coverage of this new functionality.

dylanmccall:
Okay, I added some unit tests here, both for `importcontent remote` and `importcontent disk` with manifest files. I think they cover all the cases I've added here, including the `--no-detect-manifest` option and manifest files with mismatched channel versions.

kolibri/core/content/management/commands/importcontent.py
```diff
@@ -71,8 +90,8 @@ def add_arguments(self, parser):
         parser.add_argument(
             "--node_ids",
             "-n",
-            # Split the comma separated string we get, into a list of strings
-            type=lambda x: x.split(","),
+            # Split the comma separated string we get, into a set of strings
+            type=lambda x: set(x.split(",")) if x else set(),
```
rtibbles: Could we do this coersion to a set inside the `handle_async` method instead of here? Currently this introduces a different in behaviour between command line invocations of the management command and calls to `call_command` - as you can see below in the tests you didn't modify, `node_ids` and `exclude_node_ids` are still being set as lists. If the content manifest behaviour in particular is sensitive to this, then this may cause issues.

dylanmccall: Yeah, sure thing! In the newest commit, the argument is parsed into a list (an empty list if it is given an empty string), and we coerce it into a set inside `_transfer()`. I'm not 100% sure about doing it there, but it kind of felt right since we clean up some other inputs there as well.

rtibbles (approved):

PR #9474, opened 2022-06-01
Adds new public content metadata endpoints, allows for remote browsing of metadata

### LINKED ISSUE #9380: Create a public API endpoint for ContentNode data
Create a versioned public API endpoint for returning ContentNode data.

At the moment, it can simply return the same as the current internal ContentNode viewset.

This should be implemented as a v2 endpoint, to match the v2 public channel endpoint in #9381.

As long as fields are only added, and not removed, the viewset can remain as v2.

As soon as there is a more radical overhaul of ContentNode metadata, we will need to be able to offer a backward compatible mapping for v2, and move to a v3API endpoint.

### LINKED ISSUE #9381: Update public Channel API endpoint
Update the public Channel API endpoint to return fields that are a superset of the existing public Channel API endpoint and the internal ChannelMetadata API endpoint.

We should implement this as the v2 spec for this endpoint, as if the v2 spec is unavailable, this will serve as a useful signal that the Kolibri instance cannot be publicly browsed for resources.

### LINKED ISSUE #9382: Allow internal ContentNode endpoint to handle baseurl parameter for browsing non-local libraries
This would be similar in motivation to the RemoteChannelViewset: https://github.com/learningequality/kolibri/blob/develop/kolibri/core/content/api.py#L1447 which is used to provide a local API endpoint for browsing the public channel API of remote Kolibris and Kolibri Studio.

This should then map the data returned from the public endpoint to data that is mapped to the current internal ContentNode configuration.

The implementation should intelligently cache queries, to avoid making more requests than necessary on the local device.

### LINKED ISSUE #9385: Enhance ChannelMetadataViewset to handle baseurl parameter to browse non-local libraries
To allow browsing channels on remotes for learn, we need to copy behaviour from [RemoteChannelViewset](https://github.com/learningequality/kolibri/blob/develop/kolibri/core/content/api.py#L1447).

In addition, full and robust caching of requests should be implemented, to prevent too many requests from being made to remote servers.

It should better handle incomplete requests, or errors during requests, to return meaningful error codes, and not return 500 errors.

It should return fields equivalent to the internal ChannelMetadata viewset.

We should share a request cache between the enhanced ChannelMetadata viewset and the RemoteChannelViewset.

### LINKED ISSUE #9390: Create public ContentNodeTreeViewset or expose publicly in some way on the public content node viewset
To allow efficient, paginated browsing of remote channels, this viewset https://github.com/learningequality/kolibri/blob/develop/kolibri/core/content/api.py#L995 needs to be available in a publicly accessible form. The simplest mechanism would be to make a separate versioned public viewset parallel to the public ContentNode viewset described in #9380.

### LINKED ISSUE #9391: Handle baseurl on ContentNodeTree viewset to allow browsing of non-local topic trees
In parallel to #9382, we should be able to browse non-local topic trees by making a request to the baseurl.

rtibbles:
## Summary
* Cleans up all previously existing 'user data' specific endpoints for content node
* Simplifies and adds etags to all content metadata endpoints
* Adds a new v2 namespace for public content metadata endpoints
* Expands the existing internal ChannelMetadataViewset to include all public data and uses it for the v2 endpoint
* Adds public ContentNodeViewset and ContentNodeTreeViewset to allow remote browsing of public metadata
* Adds `baseurl` query parameter handling to these three internal endpoints
* This means that the internal endpoints can now proxy fetches to other Kolibri's public endpoints

## References
Fixes #9381
Fixes #9380
Fixes #9390
Fixes #9385
Fixes #9382
Fixes #9391

## Reviewer guidance
Test the three endpoints with another instance of Kolibri running (preferably from a different Kolibri home dir, so that different metadata will be returned).

Example URLs:

Channel:
`http://127.0.0.1:8000/api/content/channel/?format=json&baseurl=http://127.0.0.1:8080/`
Public:
`http://127.0.0.1:8000/api/public/v2/channel/?format=json`

ContentNode:
`http://127.0.0.1:8000/api/content/contentnode/?format=json&max_results=25&baseurl=http://127.0.0.1:8080/`
Public:
`http://127.0.0.1:8000/api/public/v2/contentnode/?format=json&max_results=25`

Tree:
`http://127.0.0.1:8000/api/content/contentnode_tree/<topic_id>/?format=json&max_results=2&baseurl=http://127.0.0.1:8080/`
Public:
`http://127.0.0.1:8000/api/public/v2/contentnode_tree/<topic_id>/?format=json&max_results=2`

----

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/core/assets/src/api-resources/contentNode.js
+++ b/kolibri/core/assets/src/api-resources/contentNode.js
@@ -115,11 +115,23 @@ export default new Resource({
     });
     return promise;
   },
-  fetchPopular(getParams) {
-    return this.fetchListCollection('popular', getParams);
+  fetchPopular(params = { popular: true }) {
+    const promise = new ConditionalPromise();
+    const url = urls['kolibri:core:usercontentnode_list']();
+    promise._promise = this.client({ url, params }).then(response => {
+      this.cacheData(response.data);
+      return response.data;
+    });
+    return promise;
   },
-  fetchNextSteps(getParams) {
-    return this.fetchListCollection('next_steps', getParams);
+  fetchNextSteps(params = { next_steps: true }) {
+    const promise = new ConditionalPromise();
+    const url = urls['kolibri:core:usercontentnode_list']();
+    promise._promise = this.client({ url, params }).then(response => {
+      this.cacheData(response.data);
+      return response.data;
+    });
+    return promise;
   },
   cache: {},
   fetchModel({ id }) {
--- a/kolibri/core/public/api_urls.py
+++ b/kolibri/core/public/api_urls.py
@@ -27,9 +27,10 @@ from .api import get_public_channel_lookup
 from .api import get_public_file_checksums
 from .api import InfoViewSet
 from .api import PublicChannelMetadataViewSet
+from .api import PublicContentNodeTreeViewSet
+from .api import PublicContentNodeViewSet
 from .api import SyncQueueViewSet
-from kolibri.core.content.api import ContentNodeTreeViewset
-from kolibri.core.content.api import ContentNodeViewset
+
 
 router = routers.SimpleRouter()
 
@@ -44,10 +45,12 @@ public_content_v2_router.register(
     r"channel", PublicChannelMetadataViewSet, base_name="publicchannel"
 )
 public_content_v2_router.register(
-    r"contentnode", ContentNodeViewset, base_name="publiccontentnode"
+    r"contentnode", PublicContentNodeViewSet, base_name="publiccontentnode"
 )
 public_content_v2_router.register(
-    r"contentnode_tree", ContentNodeTreeViewset, base_name="publiccontentnode_tree"
+    r"contentnode_tree",
+    PublicContentNodeTreeViewSet,
+    base_name="publiccontentnode_tree",
 )
 
 # Add public api endpoints
--- a/kolibri/core/public/api_urls.py
+++ b/kolibri/core/public/api_urls.py
@@ -26,8 +26,8 @@ from .api import get_public_channel_list
 from .api import get_public_channel_lookup
 from .api import get_public_file_checksums
 from .api import InfoViewSet
+from .api import PublicChannelMetadataViewSet
 from .api import SyncQueueViewSet
-from kolibri.core.content.api import ChannelMetadataViewSet
 from kolibri.core.content.api import ContentNodeTreeViewset
 from kolibri.core.content.api import ContentNodeViewset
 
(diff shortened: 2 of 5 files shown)


kolibri/core/assets/src/api-resources/contentNode.js
```diff
@@ -115,11 +115,23 @@ export default new Resource({
     });
     return promise;
   },
-  fetchPopular(getParams) {
-    return this.fetchListCollection('popular', getParams);
+  fetchPopular(params = { popular: true }) {
+    const promise = new ConditionalPromise();
```
jredrejo, line 119: have we solve the problem with `only` in `ConditionalPromise`? this makes me shiver if we need to use all with several promises (I know `fetchListCollection` also uses `ConditionalPromise` but wondering if we could just use `Promise` to avoid this problem)

kolibri/core/public/api_urls.py
```diff
@@ -36,9 +39,21 @@
...
+    r"contentnode", ContentNodeViewset, base_name="publiccontentnode"
+)
+public_content_v2_router.register(
+    r"contentnode_tree", ContentNodeTreeViewset, base_name="publiccontentnode_tree"
+)
+
 # Add public api endpoints
 urlpatterns = [
     url(r"^", include(router.urls)),
+    url(r"v2/", include(public_content_v2_router.urls)),
```
jredrejo, line 59: This does not match with the examples in the review guidance of this PR description. I had to change them to
http://127.0.0.1:8000/api/public/v2/channel/?format=json&baseurl=http://127.0.0.1:8080/
http://127.0.0.1:8000/api/public/v2/contentnode/?format=json&max_results=25&baseurl=http://127.0.0.1:8080/
etc.
Please, change the description for future reviewers

jredrejo (approved):
Code looks good to me. I've tested it and haven't found any issue.

However I'd love to know the purpose of this new _proxy_ api. Obviously it's incompatible with previous versions of kolibri, so I'd like to be sure this is not thought to interact with old servers at any point.

kolibri/core/assets/src/api-resources/contentNode.js
rtibbles: Yeah - this is a precautionary measure to have a consistent API until we can get rid of `ConditionalPromise` - although your comment here has made me realize a potential way to do that right now.

kolibri/core/public/api_urls.py
rtibbles: Well, the primary purpose of this was for the non-public endpoints. I actually think I should remove the proxy behaviour from the public endpoints as it's more of a side effect here than an intended effect. Will do that now.

rtibbles: I have updated the PR description with both the public endpoints and the internal endpoints.

rtibbles:
> However I'd love to know the purpose of this new proxy api

The primary purpose is to allow remote browsing of other Kolibri's libraries, so that the resources can be directly interacted with and previewed, without having to import them first. In an ideal world, we would make this work for any previous version of Kolibri as well, but my sense is that the complexity we would have to introduce to ensure that would be high.

If we do change our content schema in the future, we would have to provide some sort of translation layer (in the same way we do now for metadata import) to map from an older schema to a newer one. I am in two minds about whether we will want to do this, but that's the reason for using the versioned public APIs so we can have a clear signal when we need to do that mapping if needed.

PR #9475, opened 2022-06-01
Remove CoreBase from device

### LINKED ISSUE #9234: Remove CoreBase from the Device plugin
Remove all uses of CoreBase from the Device plugin

See https://github.com/learningequality/kolibri/pull/9196 or https://github.com/learningequality/kolibri/pull/9222 for examples of how this was done in Facility and Learn.

nucleogenesis:
## Summary

Removes CoreBase from Device.

Introduces two components to device which wrap the newly created Core wrapping components. 

`AppBarDevicePage` and `ImmersiveDevicePage` bootstrap their respective properties per specific needs of the Device plugin.  This avoids having to import the `DeviceTopNav` component wherever `AppBar...` is used and to avoid putting the same `KPageContainer` in each of the `Immersive...` pages.

### How page types flow - Immersive icon conventions

You have an AppBar page - which has the sub-menu, is purple, has the sidebar, etc.

This can then transition into an Immersive page. This *first* page will always have an X in the corner (which is the default value for the `icon` prop). This is because the "Immersive" state is a modal state. It's like you've opened up a screen with one purpose and have only one place to return to, which is an entirely different context (ie, you go from a list of users [AppBar] to editing a user [Immersive] - then you go back to the list of users when you're done).

When you are already in an Immersive view and navigate to another view - then you will by convention (not enforcement) see a back arrow. This indicates that you will not leave this Immersive view - but rather will be taken back to where you were previously.

### Facilities Task Page is now immersive

The Task Manager pages are immersive now - the content management one already was Immersive, but the Facility syncing one was not. I updated the string there to better indicate the context as well. 

Now the task manager experience is identical.

## References

Fixes #9234 

## Reviewer guidance

*QA* - Please walk through the Device plugin end-to-end, being sure to navigate to every possible page in the plugin.

*Code* - Thoughts on making device-specific page wrappers & their names? Would love any thoughts on how to make this cleaner. I've tried to make smallish commits that are about one things, so that may be a good way to walk through it. Let me know if you try this - I'm trying to get better about my commit sizes and messages.

----

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated


pcenov:
Observed the following console errors and issues while testing the build assets:

| Issue| Screenshot |
| Select resources from a channel - Erroneously displayed info for a new channel available for each channel. If I click "View changes" I can always see that there are no changes |  |
| Channel import - A 0 channels available screen is displayed briefly prior to showing all the available channels | https://user-images.githubusercontent.com/79847249/171665346-da367ae1-32d4-434f-a684-15ca72084b17.mp4 |
| Info tab - Errors in the console with no other front facing issues |  |
| User permissions modal - Errors in the console with no other front facing issues |  |
| Facilities task manager - Errors in the console with no other front facing issues |  |

kolibri/plugins/device/assets/src/views/DeviceIndex.vue
```diff
@@ -1,6 +1,7 @@
 <template>
 
-  <CoreBase
+  <!-- TODO Remove this stuff before making PR -->
```
marcellamaki: is the TO DO done and the comment not removed, or is there more to do here?

kolibri/plugins/device/assets/src/views/PageWrappers/AppBarDevicePage.vue
```diff
@@ -0,0 +1,68 @@
...
+          // Need to override overflow rule for setting page
+          styles['overflowX'] = 'inherit';
+        }
+
+        return styles;
+      },
+      slotContainerType() {
+        /* Most pages should be rendered in a KPageContainer (maybe all) - this
+           allows a Device page to just be put into a plain div instead depending on the prop */
+        return this.noKPageContainer ? 'div' : 'k-page-container';
```
marcellamaki: So, this is just an escape hatch for some other kind of page? It's not for some known alternative style, but just for the lack of the typical KPageContainer style, right?

kolibri/plugins/device/assets/src/views/FacilitiesPage/FacilitiesTasksPage.vue
```diff
@@ -88,10 +87,9 @@
       },
     },
     $trs: {
-      backToFacilitiesLabel: {
-        message: 'Back to facilities',
-        context:
-          'Link to take user back to the Device > Facilities page where they can see the list of facilities.',
+      facilitiesTaskManagerTitle: {
```
marcellamaki, line 94: I can't quite follow why this string is changing. Is it now the title of the current page, vs. the title of the previous page (that was associated with the back button)?

marcellamaki (commented):
Overall the code read through looks pretty good. I know Peter found a few QA issues and I haven't done any manual testing yet. From the error messages in the console though, it's not super clear to me if those are related to the PR or if those are other things in develop that have slipped through the cracks. (The `KButtonGroup` one in particular seems odd to be related to this, but who knows.)

If you're willing to take a closer look and sort out which ones might be specific to this, and update them, and make follow up issues for any that seem unrelated, I'll do a second review of the changes and do some manual QA as well!

rtibbles:
I think the errors in console with no other issues are due to KButtonGroup. Rebasing to latest develop should resolve these.

kolibri/plugins/device/assets/src/views/DeviceIndex.vue
nucleogenesis: Ah dang - these props are worthless just need to clear them out (and also more thoroughly read my code as I make my commits :grimacing:

kolibri/plugins/device/assets/src/views/PageWrappers/AppBarDevicePage.vue
nucleogenesis: Yep - just allows you to have a blank slate if you want to, but defaults to the predominant (and as it turns out currently only) layout.

kolibri/plugins/device/assets/src/views/FacilitiesPage/FacilitiesTasksPage.vue
nucleogenesis: This is removing the previous string that was a `basic-link` style link with an arrow like `<- Back to Facilities` because this page has changed from being an AppBar page to being an Immersive page -- this is to keep it consistent with the other content task manager page. There is no basic link in that design since the immersive page has the back arrow.

I added a string "Facilities - Task Manager" to give the immersive page a clear name that distinguishes it from the other separate task manager.

nucleogenesis: Sorry - I should have posted screen shots in the PR description

pcenov:
@radinamatic as discussed I had another look at this one today using the latest Linux asset and out of the 5 issues reported by me only the first one is fixed so far: "Select resources from a channel - Erroneously displayed info for a new channel available for each channel. If I click "View changes" I can always see that there are no changes"
I don't see any new issues or regressions.
@nucleogenesis I guess you are aware of that, just ping me or Radina when this one is ready for another round of testing. Thanks!

nucleogenesis:
@radinamatic can you give this another pass at your convenience? I've resolved all of the issues noted above by @pcenov

kolibri/plugins/device/assets/src/views/PageWrappers/AppBarDevicePage.vue
```diff
@@ -0,0 +1,68 @@
+<template>
+
+  <AppBarPage :title="$tr('deviceManagementTitle')">
```
rtibbles: More of a question than a request to change - aside from the title string here, is there anything that would actually need to be consolidated in a component here, or could each top level page in device have its own (slightly boilerplatey) top level component that just directly included these things?

We could move the string out into a JS translator instead to be referenceable in multiple pages.

nucleogenesis: Just to clarify - are you asking if there is any reason for having this component w/ the string here as opposed to having this template applied to every `*Page` component in the plugin directly? The only other thing this does really is include the proper side nav though. 

re: moving the string to a js translator, this and the DeviceManagementSideNavEntry are the only places we use the message "Device" at the moment and we typically won't make a string common unless it's used 3+ times in the same context

Maybe I'm misunderstanding your question though and would be happy to clarify on a call or in Slack

rtibbles: Right, I was just wondering if there was any benefit from having this component instead of just moving the functionality into the top level page components and not having to have these conditionality here.

nucleogenesis: Could do -  then each of them would import DeviceTopNav themselves and they'd just use `AppBarPage` directly and put this somewhere when needed.

```
<template #subNav>
      <DeviceTopNav v-if="!hideSubNav" />
 </template>
```

I can't see any downside to that approach and wish I'd taken it in the first place.

Separately, now looking back I'm wondering why I didn't just have every page put their business in their own KPageContainer rather than adding a prop & conditional here in the first place.

rtibbles: Yeah, that was kind of my thought - this felt like it was leaving some residual, Device specific, CoreBase lying around.

nucleogenesis: In practice the difference seems to be that this way does little more than DRY up the importing and implementation of the SubNav bit across a bunch of components -- which could get particularly hairy when refactoring Coach's umpteen pages. 

Pretty sure I can do this refactor pretty quickly with a little bit of vim macro magic.

radinamatic:
The issue ``0 channels available screen is displayed briefly prior to showing all the available channels`` reported by @pcenov is still extant:

0-resources1

It appears not only on importing a new channel, but also when the user reviews the new resources added prior to importing, so it does have a confusing effect... 😕 

0-resources2

radinamatic:
I don't see the errors in the console while in the **Info** anymore, but the 404 error is still there in the **Permissions** modal:

permissons

As @pcenov observed, no user-facing issues.

radinamatic:
In the task manager I don't see the ``TypeError`` noted by @pcenov any more, it's a different set of errors in the console in all views with no other user/front facing issues.

tasks

rtibbles:
Note the errors you are seeing are this issue: https://github.com/learningequality/kolibri/issues/9531 (which only appear when running from source/dev mode) and are not caused by this PR :)

nucleogenesis:
@rtibbles Once the re-run passes (if they do) I think this will be ready with review of this commit https://github.com/learningequality/kolibri/pull/9475/commits/8477573a6464533d2648c133093a3fc392faac69

and this one https://github.com/learningequality/kolibri/pull/9475/commits/27a40ec7ada16088c7c1d3a22411b76210c5ff22 which removes a now broken reference to ImmersiveToolbar (which is now a private-to-core component exposed via the ImmersivePage component)

pcenov:
I was able to identify the following issues while testing the latest build assets:

| Issue| Screenshot |
| Permissions > View Permissions - Missing checkmarks |  |
| Permissions > Edit Permissions - When I select the 'Make super admin' checkbox, the other checkbox should be selected as well while it is not. If I uncheck and then check again the Make super admin the other checkbox does gets checked. Note that if I save and return to edit the checkboxes the checkmarks are not displayed. |  |
| Missing snackbar messages for updated settings, deleted facility, error log copied to clipboard etc  |  |

@radinamatic

kolibri/plugins/device/assets/src/views/DeviceIndex.vue
```diff
@@ -181,25 +64,19 @@
...
-    },
   };
 
 </script>
 
 
 <style lang="scss" scoped>
 
+  /* Default styles for KPageContainer in Device */
+  /deep/ .device-container {
```
rtibbles: Could we make this more explicit, at the risk of repetition, by adding to the specific uses of KPageContainer? Seems to be relying on the addition of the `device-container` class in each page anyway, so might be better to just be explicit about what that means?

kolibri/plugins/device/assets/src/views/DeviceInfoPage.vue
```diff
@@ -1,72 +1,81 @@
...
-        <th>{{ coreString('deviceNameLabel') }}</th>
-        <td>
-          {{ deviceNameWithId }}
+  <AppBarPage :title="pageTitle">
+
+    <template #subNav>
+      <DeviceTopNav v-if="!hideSubNav" />
+    </template>
+
+    <KPageContainer class="device-container">
```
rtibbles, line 9: For example here, I can imagine this class getting deleted as it seems to be unused in the component?

kolibri/plugins/device/assets/src/views/ManageContentPage/NewChannelVersionPage.vue
```diff
@@ -195,6 +209,7 @@
       },
     },
     mounted() {
+      console.log(this.$router);
```
rtibbles: `log`!

kolibri/plugins/device/assets/src/views/UserPermissionsPage/index.vue
```diff
@@ -1,107 +1,112 @@
 <template>
 
-  <div>
-    <h1 v-if="user === null">
-      {{ $tr('userDoesNotExist') }}
-    </h1>
+  <ImmersivePage
+    v-if="!$store.state.core.loading"
```
rtibbles, line 4: This conditional is the only thing that I can see that is different between this page and the other pages here - can't think what else might be causing the issues in the permissions selection reported by @pcenov.

rtibbles (commented):
All making sense to me - one question about the `device-container` class - not blocking but feels like it's DRY in the same way that the DeviceIndex AppBar imports were.

kolibri/plugins/device/assets/src/views/DeviceIndex.vue
nucleogenesis: What do you think of this compromise https://github.com/learningequality/kolibri/pull/9475/commits/eb2ca57daa43d0c6374d32f1b94405ce33d72e4b

I'm keen to keep the styles living in one place - I figure we're not entirely unlikely to want to adjust all pages at once at some point. This way we get both - explicit "here is why this class is here" and "here is where that lives and is defined".

nucleogenesis:
@pcenov @rtibbles all feedback addressed

big :man_facepalming: on not putting the GlobalSnackbar in the CorePages... :sweat_smile:

pcenov:
@nucleogenesis the Buildkite build is failing for some reason, tried to rebuild but to no avail.

rtibbles:
@pcenov can you use the artifacts from the Github actions instead? https://github.com/learningequality/kolibri/pull/9475#issuecomment-1144249143

kolibri/plugins/device/assets/src/views/DeviceIndex.vue
rtibbles: I like it - keeps a single source of truth, while also being completely explicit within the component.

rtibbles: The other advantage of using a mixin rather than just a defined class that is imported is that it allows for specific deviation within any of the page components if needed.

rtibbles (commented):
Code wise this is good to go - just needs manual QA sign off.

I don't think that the build errors are related to this PR, but I have asked @DXCanas for any leads.

pcenov:
Looks good to me as well, just a couple of instances where the snackbar notification is still not being displayed for some reason:
1. Device > Channels > Channel import/deletion/export - Missing ‘Task started…’ notification
2. Device > Permissions > Edit permissions - Missing ‘Changes saved’ notification

nucleogenesis:
@pcenov just pushed the fix for the snackbars - should be good to go I hope.

pcenov:
@nucleogenesis - gave it another round of testing and all looks good to me know as well. No new issues spotted.

radinamatic (approved):
Manual QA passing, looking good to go! 👏🏽 💯 :shipit:

PR #9477, opened 2022-06-02
Update pytest to v4.1.1

maxbrunet:
## Summary

Hello there :wave:, I figured a good way to get familiar with the code base is to do some housekeeping:

* Upgrade `pytest` to `4.1.1`, it is as far as I got without breakage
* Downgrade `attrs` to `19.1.0` to work around an issue in `pytest`: https://stackoverflow.com/q/58189683
* Ignore `Overriding setting DATABASES` warnings in `test_sqlalchemy_bridge.py`, same as [`test_dbrestore.py`](https://github.com/learningequality/kolibri/blob/e295962435a7ed55f9d120f2184c1596916c0732/kolibri/core/deviceadmin/tests/test_dbrestore.py#L117)
* Remove `python setup.py test` alias, support has been [removed in pytest 3.6.2](https://docs.pytest.org/en/7.1.x/changelog.html#pytest-3-6-2-2018-06-20)

`pytest-django` has been giving me troubles:

* Upgrading further requires to upgrade `pytest-django` to keep the compatibility
* Upgrading `pytest-django` results in the tests being flaky, I suspect it is [the change related to `django.setup()` in `3.4.0`](https://pytest-django.readthedocs.io/en/latest/changelog.html#v3-4-0-2018-08-16)
* Flakiness only happens when running **all** the tests, when selecting a few test files, it does not. I have not been able to make a minimal reproduction to troubleshoot effectively so far. I will try to continue looking into it, but any suggestions on how to approach the issue is welcome

With the upgrade, pytest now surfaces user warnings (e.g. Django deprecations), [the change came in pytest 3.9.0](https://docs.pytest.org/en/7.1.x/changelog.html#pytest-3-9-0-2018-10-15-not-published-due-to-a-release-automation-bug), I may have time to tackle them in a separate PR.

## References

Related to #4618 (I would suggest to keep it open until we get to a version recent enough)

## Reviewer guidance

I have tested it locally against Python 3.9, I would assume the CI will do against all the supported versions.

----

## Testing checklist

- [x] Contributor has fully tested the PR manually
- [ ] If there are any front-end changes, before/after screenshots are included
- [ ] Critical user journeys are covered by Gherkin stories
- [ ] Critical and brittle code paths are covered by unit tests

## PR process

- [ ] PR has the correct target branch and milestone
- [ ] PR has 'needs review' or 'work-in-progress' label
- [ ] If PR is ready for review, a reviewer has been added. (Don't use 'Assignees')
[...]

PR #9483, opened 2022-06-04
Fix deprecation warnings

maxbrunet:
## Summary

Changes:

- Replace `logging.warn()` by `logging.warning()`
- Replace `user.is_anonymous()` fn by `.is_anonymous` property
- Replace `user.is_authenticated()` fn by `.is_authenticated` property
- Replace `assertEquals()` by `assertEqual()`
- Use raw-string instead of regular-string for Regex
- Explicitly set `on_delete=models.CASCADE` for `ForeignKey()`
- Replace `django.core.urlresolvers` imports by `django.urls`
- Fix `django.conf.urls.include()` deprecations (:warning: Please double-check it still works as expected, it may be the least trivial change)
  https://docs.djangoproject.com/en/1.11/releases/1.9/#passing-a-3-tuple-or-an-app-name-to-include
- Replace `base_name` kwarg by `basename` in `router.register()`
- Replace `assertDictContainsSubset()` by `assertEqual(larger_dict, {**larger_dict, **smaller_dict})`
- Replace `@detail_route()`/`@list_route()` by `@action(detail: bool)`
- Replace `assertRaisesRegexp()` by `assertRaisesRegex()`
- Replace `name` kwarg by `field_name` in Filters

Deprecation warnings resolved:

- `DeprecationWarning: The 'warn' method is deprecated, use 'warning' instead`
- `RemovedInDjango20Warning: Using user.is_authenticated() and user.is_anonymous() as a method is deprecated. Remove the parentheses to use it as an attribute.`
- `DeprecationWarning: Please use assertEqual instead.`
- `DeprecationWarning: invalid escape sequence`
- `RemovedInDjango20Warning: on_delete will be a required arg for ForeignKey in Django 2.0. Set it to models.CASCADE on models and in existing migrations if you want to maintain the current default behavior. See https://docs.djangoproject.com/en/1.11/ref/models/fields/#django.db.models.ForeignKey.on_delete`
- `RemovedInDjango20Warning: Importing from django.core.urlresolvers is deprecated in favor of django.urls`
- `RemovedInDjango20Warning: Passing a 3-tuple to django.conf.urls.include() is deprecated. Pass a 2-tuple containing the list of patterns and app_name, and provide the namespace argument to include() instead.`
- `RemovedInDjango20Warning: Specifying a namespace in django.conf.urls.include() without providing an app_name is deprecated. Set the app_name attribute in the included module, or pass a 2-tuple containing the list of patterns and app_name instead.`
- `PendingDeprecationWarning: The `base_name` argument is pending deprecation in favor of `basename`.`
- `DeprecationWarning: assertDictContainsSubset is deprecated`
- <code>DeprecationWarning: `detail_route` is deprecated and will be removed in 3.10 in favor of `action`, which accepts a `detail` bool. Use `@action(detail=True)` instead.</code>
- <code>DeprecationWarning: `list_route` is deprecated and will be removed in 3.10 in favor of `action`, which accepts a `detail` bool. Use `@action(detail=False)` instead.</code>
- `DeprecationWarning: Please use assertRaisesRegex instead.`
- <code>DeprecationWarning: `Filter.name` has been renamed to `Filter.field_name`. See: https://django-filter.readthedocs.io/en/develop/migration.html</code>

Other warnings come from within dependencies which will need to be upgraded (e.g Django, SQLAlchemy)

## References

Fix warnings within the kolibri code base surfaced by #9477

## Reviewer guidance

This can be review commit by commit, I have taken care of making one per deprecation.
This has been tested against Python 3.9 only.
I suspect some of them may fail against Python 2.7 (e.g. deprecations related to `unittest` module), are there still blockers to drop 2.7 support? If there are no concerns, I could make a PR to drop it? (It would also unblock upgrades to pytest 5 and likely many other dependencies)

----

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated


rtibbles:
Yes, we are still supporting Python 2.7 for now. The issue seems to be incompatible code syntax in one of the test files (https://github.com/learningequality/kolibri/runs/6735715542?check_suite_focus=true), this should be easy to resolve though.

rtibbles:
Thank you for addressing all these deprecation warnings, this will massively reduce the workload for upgrading to a newer version of Django specifically. Will review more fully on Monday.

maxbrunet:
> Yes, we are still supporting Python 2.7 for now. The issue seems to be incompatible code syntax in one of the test files ([learningequality/kolibri/runs/6735715542?check_suite_focus=true](https://github.com/learningequality/kolibri/runs/6735715542?check_suite_focus=true)), this should be easy to resolve though.

Yes, this one is easy, but it is only the 1st one it encounters, as I said there are likely more incompatibilities

kolibri/core/auth/test/test_api.py
```diff
@@ -359,9 +359,9 @@ def test_facility_user_can_get_detail(self):
             reverse("kolibri:core:facility-detail", kwargs={"pk": self.facility1.pk}),
             format="json",
         )
-        # .assertDictContainsSubset checks that the first argument is a subset of the second argument
-        self.assertDictContainsSubset(
-            {"name": self.facility1.name}, dict(response.data)
+        self.assertEqual(
+            dict(response.data),
+            {**dict(response.data), **{"name": self.facility1.name}},
```
rtibbles: Note this is the line that is giving Python 2 syntax issues. Can fix by doing:
```suggestion
            dict(name=self.facility1.name, **dict(response.data)),
```

kolibri/core/auth/test/test_api.py
```diff
@@ -1136,7 +1136,9 @@ def post_resetsettings():
         for setting in ["formal", "nonformal", "informal"]:
             set_all_false_and_preset(facility, setting)
             response = post_resetsettings()
-            self.assertDictContainsSubset(mappings[setting], response.data)
+            # Merge smaller dict into larger dict, if the smaller dict is a subset of the larger one, the result should be equal to the larger one
+            # The dict union operator can be used in Python +3.9: assertEqual(larger_dict, larger_dict | smaller_dict)
+            self.assertEqual(response.data, {**response.data, **mappings[setting]})
```
rtibbles: See above for Python 2 syntax compatibility - might need to make an intermediate dict that you update with `mappings[setting]`.

rtibbles (commented):
All the code changes look good to me - as long as we can get tests passing and syntax checks working for Python 2, I think this will be good to merge.

rtibbles:
> Yes, this one is easy, but it is only the 1st one it encounters, as I said there are likely more incompatibilities

Looks like there's just one more, and again to do with Python 2.7 dict shorthand syntax, that makes things a little less verbose, but is still easy to work around.

maxbrunet:
The CI passed for Python 2. It gets to live another day. :tada:
I have not found an issue tracking the removal of Python 2 support, I'll open one.

Edit: #9486

rtibbles (approved):

PR #9485, opened 2022-06-06
Upgrade webpack

rtibbles:
## Summary
Upgrade webpack and vue-loader to version ranges that no longer produce incompatible installation

Fixes this issue: https://github.com/vuejs/vue-loader/issues/1859


nucleogenesis (commented):
What was the problem this fixes? How do you know that this fixes the issue? Should be good to go in any case if the builds all pass, but just curious how you worked this out in case it helps me with my debugging of various build complaints / bugs.

marcellamaki (approved):
Version upgrades all seem to be working harmoniously. Builds and runs without issue - smoke testing all looks good

PR #9503, opened 2022-06-08
Task updates for android

### LINKED ISSUE #9263: Add 'Long running' metadata label to task definition
In the Android context, it is important to know if a task is anticipated to run for a long time (> 10 minutes) or whether its execution will complete more quickly.

This allows us to flag to the user that the process is ongoing, and even more importantly, flag to the Android process manager that this task is taking a long time because it _needs to_ not because it is just hogging resources and should be killed.

We should add this to the the arguments accepted by the RegisteredTask constructor and its decorator - meaning that we can define at the task level whether it is likely to be long running.

I imagine it is plausible that we could add this at the job level also, so that depending on the arguments to the job we conditionalize it, but that seems like additional complexity/premature optimization for now.

### LINKED ISSUE #9264: Add hooks to allow extending the task job storage class with additional callbacks when jobs are enqueued
In order to allow an alternative task runner to fully replace our worker class, we need to allow the Storage class to take additional actions when a task is created in the job store.

This means we will essentially be engaging in some double book keeping - where the Kolibri job store simply becomes a place to update status about a task, and the actual execution is stored in this additional mechanism.

In the Android App context, this means that the information about the scheduling and execution conditions of the task will be stored in the WorkManager API, while the Kolibri jobs database will simply keep a record of progress updates to allow for our usual reporting back of status to the user.

### LINKED ISSUE #9265: Add hooks to allow calling additional callbacks when job progress or metadata are updated
To allow decoupling of job execution from reporting of status to the user, we allow a task to be executed with an additonal callback that will be called with the complete representation of the job object, so that e.g. the existing CLI task progress bars can be displayed, or that a notification in Android can be updated when the progress of a task updates.

rtibbles:
## Summary
* Adds a docstring for the RegisteredTask object
* Adds `long_running` attribute for RegisteredTask and Job objects to flag if a task is anticipated to take a long time
* Consolidates job execution code further to allow for easy reuse in other task runner implementations
* Raises a runtime error to prevent people trying to set empty lists as a default in our options - it produces unexpected results
* Adds an option and ability to handle callbacks for the Storage schedule task, to allow another task runner implementation to hook into our schedule method and receive all necessary data for running a Kolibri task
* Adds optional update hooks so that a Storage object can execute callbacks when the status of a task is updated - primarily intended to be used inside a custom task runner implementation's job execution to message to the task runner
* Moves the ping task into a `tasks.py` module to ensure it is always registered
* Adds automatic retry of failed tasks and updates the ping task to leverage this instead of internally rescheduling itself
* Adds a test for the automatic retry

## References
Fixes #9263
Fixes #9264
Fixes #9265

## Reviewer guidance
Do the tests still pass? Are the interfaces here sufficiently sensible?

----

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/core/tasks/registry.py
+++ b/kolibri/core/tasks/registry.py
@@ -157,7 +157,34 @@ class RegisteredTask(object):
         cancellable=False,
         track_progress=False,
         permission_classes=None,
+        long_running=False,
     ):
+        """
+        :param func: Function to be wrapped as a Registered task
+        :type func: function
+        :param job_id: Fixed job_id to use for any job run from this task, defaults to None
+        :type job_id: str
+        :param queue: The queue to run this task in by default, defaults to DEFAULT_QUEUE
+        :type queue: str
+        :param validator: The job validator, used to validate and serialize JSON into
+        the args, kwargs, and job arguments for running a job of this task,
+        defaults to JobValidator
+        :type validator: JobValidator
+        :param priority: The priority for this task, can be either HIGH or REGULAR,
+        defaults to Priority.REGULAR
+        :type priority: int
+        :param cancellable: Can this task be cancelled while running, defaults to False
+        :type cancellable: bool
+        :param track_progress: Does this task track progress while it's executing, defaults to False
+        :type track_progress: bool
+        :param permission_classes: Classes that determine if a user has read and write access to
+        jobs of this task, defaults to None
+        :type permission_classes: list of BasePermission derived classes or
+        instantiated BasePermission derived classes.
+        :param long_running: In regular operation should this task finish in under ten minutes,
+        defaults to False
+        :type long_running: bool
+        """
         if permission_classes is None:
             permission_classes = []
         if not issubclass(validator, JobValidator):
@@ -172,6 +199,8 @@ class RegisteredTask(object):
             raise TypeError("cancellable must be of bool type.")
         if not isinstance(track_progress, bool):
             raise TypeError("track_progress must be of bool type.")
+        if not isinstance(long_running, bool):
+            raise TypeError("long_running must be of bool type.")
 
         self.func = func
         self.validator = validator
@@ -183,6 +212,7 @@ class RegisteredTask(object):
         self.job_id = job_id
         self.cancellable = cancellable
         self.track_progress = track_progress
+        self.long_running = long_running
 
         # Make this wrapper object look seamlessly like the wrapped function
         update_wrapper(self, func)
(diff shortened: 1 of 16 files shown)


kolibri/core/tasks/storage.py
```diff
@@ -361,9 +377,16 @@ def _update_job(self, job_id, state=None, **kwargs):
         with self.session_scope() as session:
             try:
                 job, orm_job = self._get_job_and_orm_job(job_id, session)
+                for update_hook in self.update_hooks:
+                    update_hook(job, orm_job, state=state, **kwargs)
                 if state is not None:
                     orm_job.state = job.state = state
-                    if state in {State.COMPLETED, State.FAILED, State.CANCELED}:
+                    if state == State.FAILED and orm_job.retry_interval is not None:
```
jredrejo, line 384: great, with this we can refactor our syncing tasks to remove part of the code we have there to retry

kolibri/core/tasks/registry.py
```diff
@@ -157,7 +157,34 @@ def __init__(
...
+        :type priority: int
+        :param cancellable: Can this task be cancelled while running, defaults to False
+        :type cancellable: bool
+        :param track_progress: Does this task track progress while it's executing, defaults to False
+        :type track_progress: bool
+        :param permission_classes: Classes that determine if a user has read and write access to
+        jobs of this task, defaults to None
+        :type permission_classes: list of BasePermission derived classes or
+        instantiated BasePermission derived classes.
+        :param long_running: In regular operation should this task finish in under ten minutes,
```
jredrejo, line 184: where is this limit implemented? Looking at the Jobs class I don't find it

jredrejo (commented):
I have tested the code to ensure there are no regressions. The code also looks good to me, but I haven't been able to find out how the `long_running` attribute works, not been able to test it

kolibri/core/tasks/registry.py
rtibbles: This is purely an advisory flag at the moment - the main reason for having it is that on Android if a task is going to run for longer than ten minutes, we need to tell the OS and foreground it, otherwise it risks being killed.

I had considered allowing this to be a callable rather than a boolean, but it seemed simplest to implement it as a boolean first.

kolibri/core/tasks/storage.py
rtibbles: I think so - yes. @bjester also has an open PR that adds in some extra validation to allow a task to modify its behaviour on retry, so once these are both merged, would be good to consolidate these code paths!

jredrejo (approved):

PR #9507, opened 2022-06-10
Clean up additional deprecation warnings

rtibbles:
## Summary
* Cleans up residual deprecation warnings for Django URLs and `is_anonymous`
* Removes `convert_unicode` parameter that is deprecated in SQLAlchemy
* Replaces `name` with `field_name` for django_filter filters.
* Removes deprecated exception inspection
* Avoids trying to do repair/backup of an in memory SQLite DB.


jredrejo (approved):
It all makes sense, tested anyway to be sure and I didn't find any problem.

PR #9508, opened 2022-06-10
Upgrade requests to latest Py2.7 compatible version

rtibbles:
## Summary
* Upgrades requests to 2.27.1
* Upgrades requests-toolbelt to 0.9.1

### DIFF
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -8,14 +8,14 @@ six==1.14.0
 colorlog==3.2.0 # pyup: <4.0.0
 configobj==5.0.6
 django-mptt==0.9.1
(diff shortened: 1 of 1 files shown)


jredrejo (approved):
LGTM
one weird thing, in python 2.7 everything was fine, but when using 3.8 I saw this message that I hadn't seen before, is pypy.org down?
`WARNING: There was an error checking the latest version of pip`

rtibbles:
Weird - I've not seen that before, but it does sound like a connection issue, certainly. My guess is that it's just a random failure, which thankfully isn't blocking.

PR #9509, opened 2022-06-10
Upgrade cheroot to latest.

rtibbles:
## Summary
* We are a release behind cheroot - the latest version 8.6.0 apparently reduces load when idle, which should slightly improve performance on low end hardware.

### DIFF
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -9,7 +9,7 @@ colorlog==3.2.0 # pyup: <4.0.0
 configobj==5.0.6
 django-mptt==0.9.1
 requests==2.27.1
-cheroot==8.5.2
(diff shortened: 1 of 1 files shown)


jredrejo (approved):
Nothing to object. 
Just one curiosity: looking at this 8.6.0 release, its most important changes is using type hints... even with python 2.7 . I was not aware there were libs that allow using them with 2.7: https://github.com/cherrypy/cheroot/pull/385/files

rtibbles:
It seems like because they put them in `.pyi` files, as a separate spec file, that gets ignored by Python 2.7 so can work? It does mean that you have the TypeScript style of type definitions being separate from the definitions of the objects themselves.

Issue #9469, opened 2022-06-01
Create "Change Facility" Informational screen


marcellamaki:
# Overview 

Presents a user with information about what changing their facility means, the new facility they are joining, and what will happen to their data 

## [Figma](https://www.figma.com/file/MpCG7r5PULOCJsEvnTSti8/Android-Designs-2022?node-id=473%3A6412) 
![Screen Shot 2022-06-01 at 9.03.28 AM.png](https://images.zenhubusercontent.com/6047f6e691103a78d3ad0ace/41761267-e768-4c97-bf11-566368394fc7)

### Acceptance criteria

- [ ] Page is created with correct strings 
- [ ] UI layout to spec with correct tab order and spacing
- [ ] "Your user account type will change from ‘[non-learner-role] to ‘learner’. You will need another admin to change your account type to ‘[non-learner-role]’ again. " only is included for non-learner accounts.

jredrejo:
Closed with #9512

Issue #9325, opened 2022-04-19
Create "Confirm Account" page


marcellamaki:
# Overview 

Adds a confirmation page when a user is changing facility, explaining that they will use their current account to join the new facility.

## Figma 
![Screen Shot 2022-04-20 at 11.17.03 AM.png](https://images.zenhubusercontent.com/6047f6e691103a78d3ad0ace/eb258084-bd5c-46ef-a1ce-791fcb0fe1d9)

### Accessibility notes
Tab order: 
1) "Continue" primary button 
2) "Back" secondary button 
3) "X" close out flow 

### Out of scope 

Adding the route itself is not part of this issue

### Acceptance criteria

- [ ] Page is created with correct strings 
- [ ] UI layout to spec with correct tab order and spacing

jredrejo:
Closed with #9512

PR #9512, opened 2022-06-10
Change facility first screens

### LINKED ISSUE #9324: Create "Select Facility" page
# Overview 

Allows a user to select the facility they want to join.

## Figma 
| If options exist |  No facilities to display |
|---|---|
|  |  | 

### Accessibility notes
Tab order: 
1) Radio buttons (if present)
2) "Add new address"
3) "Continue" primary button 
4) "Back" secondary button 
5) "X" close out flow 

### Out of scope 

### Acceptance criteria

- [ ] Radio button selection autoselects first option in list
- [ ] If a facility cannot be joined, the radio option is disabled
- [ ] Continue button disabled if no facility option selected

jredrejo:
## Summary
Creation of the first screens for the Change facility flow.

In order to do it, some backend additions have been needed:
* `learner_can_sign_up` is added now to the returned fields by ` /api/public/v1/facility/`
* In not _Learner only devices,_ permission to detect other facilities has to be done to any facility user. Currently only superusers and users having access to manage content could discover other devices.
* Created a new public api,` /api/public/facilitysearchuser/` that allows searching for existing (non-superuser) users in another facility, retrieving both the id and usernames for the found users.
* In the `user_profile` plugin, a couple of apis, `remotefacilities` and `remotefacilityuser` are added to proxy the search of information in a different device in the network

In the frontend side:
- Created a main component for the `ChangeFacility/index.vue` , using the new design without `CoreBase` containing the component of the different states, depending on the state machine and the assigned routes
- Some small changes in the `changeMachineFacility.js` to use the new apis
- Changes in the routes to use the new components
 - a provisional component, `ToBeDone.vue` has been created to show information of the machine states that don't have a screen implemented yet.
 - New components `SelectFacility`, `ConfirmAccount` and `ConfirmChangeFacility` to implement some of the first screens in the change facility flow
changefacility1
selectfacility1
selectfacility2
superuser

## References
Closes: #9324 , #9325, #9469 

## Reviewer guidance
- Are the design and behaviour following the guideliines in the proposed design https://www.figma.com/file/MpCG7r5PULOCJsEvnTSti8/Android-Designs-2022?node-id=596%3A10350 ?
- Are all the backend api changes safe?

In order to test it, it would be good to test:
- a normal user, with and without its username existing in the new facility
- a super user, , with and without its username existing in the new facility
Also, for the frontend to work correctly, both the current facility and the new device where the user is moving need to have these api changes. This is not compatible with previous versions of kolibri

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/core/auth/serializers.py
+++ b/kolibri/core/auth/serializers.py
@@ -133,13 +133,23 @@ class FacilitySerializer(serializers.ModelSerializer):
 
 class PublicFacilitySerializer(serializers.ModelSerializer):
     learner_can_login_with_no_password = serializers.SerializerMethodField()
+    learner_can_sign_up = serializers.SerializerMethodField()
 
     def get_learner_can_login_with_no_password(self, instance):
         return instance.dataset.learner_can_login_with_no_password
 
+    def get_learner_can_sign_up(self, instance):
+        return instance.dataset.learner_can_sign_up
+
     class Meta:
         model = Facility
-        fields = ("id", "dataset", "name", "learner_can_login_with_no_password")
+        fields = (
+            "id",
+            "dataset",
+            "name",
+            "learner_can_login_with_no_password",
+            "learner_can_sign_up",
+        )
 
 
 class ClassroomSerializer(serializers.ModelSerializer):
--- a/kolibri/core/device/api.py
+++ b/kolibri/core/device/api.py
@@ -1,7 +1,9 @@
 from datetime import timedelta
 from sys import version_info
 
+import requests
 from django.conf import settings
+from django.core.urlresolvers import reverse
 from django.db.models import Max
 from django.db.models import OuterRef
 from django.db.models.expressions import Subquery
@@ -20,7 +22,9 @@ from rest_framework import mixins
 from rest_framework import status
 from rest_framework import views
 from rest_framework import viewsets
+from rest_framework.exceptions import ValidationError
 from rest_framework.response import Response
+from six.moves.urllib.parse import urljoin
 
 import kolibri
 from .models import DevicePermissions
@@ -38,6 +42,7 @@ from kolibri.core.auth.models import Collection
 from kolibri.core.content.permissions import CanManageContent
 from kolibri.core.content.utils.channels import get_mounted_drive_by_id
 from kolibri.core.content.utils.channels import get_mounted_drives_with_channel_info
+from kolibri.core.device.permissions import IsNotAnonymous
 from kolibri.core.device.permissions import IsSuperuser
 from kolibri.core.device.utils import get_device_setting
 from kolibri.core.discovery.models import DynamicNetworkLocation
@@ -369,3 +374,28 @@ class DriveInfoViewSet(viewsets.ViewSet):
 
     def retrieve(self, request, pk):
         return Response(get_mounted_drive_by_id(pk)._asdict())
+
+
+class RemoteFacilitiesViewset(views.APIView):
+    """
+    Api to retrieve facilities information from a remote device
+    :param str baseurl: url of the server, including port to connect
+    :return : json object containing the list of facilities of the device, with their id, name, learner_can_sign_up and learner_can_login_with_no_password info
+    """
+
+    permission_classes = (IsNotAnonymous,)
+
+    def get(self, request):
+        baseurl = request.query_params.get(
+            "baseurl", request.build_absolute_uri("/")[:-1]
+        )
+        path = reverse("kolibri:core:publicfacility-list")
+        url = urljoin(baseurl, path)
+        try:
+            response = requests.get(url)
+            if response.status_code == 200:
+                return Response(response.json())
+            else:
+                return Response({})
+        except Exception as e:
+            raise ValidationError(detail=str(e))
--- a/kolibri/core/device/api.py
+++ b/kolibri/core/device/api.py
@@ -386,10 +386,8 @@ class RemoteFacilitiesViewset(views.APIView):
     permission_classes = (IsNotAnonymous,)
 
     def get(self, request):
-        baseurl = request.query_params.get(
-            "baseurl", request.build_absolute_uri("/")[:-1]
-        )
-        path = reverse("kolibri:core:publicfacility-list")
+        baseurl = request.query_params.get("baseurl", request.build_absolute_uri("/"))
+        path = reverse("kolibri:core:publicfacility-list").lstrip("/")
         url = urljoin(baseurl, path)
         try:
             response = requests.get(url)
--- a/kolibri/core/public/api.py
+++ b/kolibri/core/public/api.py
@@ -12,8 +12,10 @@ from django.utils import timezone
 from django.utils.decorators import method_decorator
 from django.views.decorators.csrf import csrf_exempt
 from django.views.decorators.gzip import gzip_page
+from django_filters.rest_framework import DjangoFilterBackend
 from morango.constants import transfer_statuses
 from morango.models.core import TransferSession
+from rest_framework import filters
 from rest_framework import status
 from rest_framework import viewsets
 from rest_framework.decorators import api_view
@@ -324,3 +326,16 @@ class SyncQueueViewSet(viewsets.ViewSet):
 
     def update(self, request, pk=None):
         return self.check_queue(request, pk=pk)
+
+
+class FacilitySearchUsernameViewSet(ReadOnlyValuesViewset):
+    filter_backends = (DjangoFilterBackend, filters.SearchFilter)
+    filter_fields = ("facility",)
+    search_fields = ("^username",)
+
+    values = ("id", "username")
+
+    def get_queryset(self):
+        return FacilityUser.objects.filter(roles=None).filter(
+            Q(devicepermissions__is_superuser=False) | Q(devicepermissions__isnull=True)
+        )
--- a/kolibri/core/public/api.py
+++ b/kolibri/core/public/api.py
@@ -25,6 +25,7 @@ from .. import error_constants
 from .constants.user_sync_statuses import QUEUED
 from .constants.user_sync_statuses import SYNC
 from kolibri.core.api import ReadOnlyValuesViewset
+from kolibri.core.auth.models import Facility
 from kolibri.core.auth.models import FacilityUser
 from kolibri.core.content.api import BaseChannelMetadataMixin
 from kolibri.core.content.api import BaseContentNodeMixin
@@ -334,6 +335,25 @@ class FacilitySearchUsernameViewSet(ReadOnlyValuesViewset):
     search_fields = ("^username",)
 
     values = ("id", "username")
+
+    def list(self, request, *args, **kwargs):
+        facility_id = request.query_params.get("facility", None)
+        try:
+            facility = Facility.objects.get(id=facility_id)
+        except AttributeError:
+            # non existing facility
+            return Response({})
+
+        if facility.dataset.learner_can_login_with_no_password:
+            queryset = self.filter_queryset(self.get_queryset())
+            return Response(self.serialize(queryset))
+        else:
+            username = request.query_params.get("search", None)
+            queryset = self.get_queryset().filter(
+                facility=facility_id, username=username
+            )
+            response = {"username": username, "id": None} if queryset else {}
+            return Response(response)
 
     def get_queryset(self):
         return FacilityUser.objects.filter(roles=None).filter(
--- a/kolibri/core/public/api.py
+++ b/kolibri/core/public/api.py
@@ -340,7 +340,7 @@ class FacilitySearchUsernameViewSet(ReadOnlyValuesViewset):
         facility_id = request.query_params.get("facility", None)
         try:
             facility = Facility.objects.get(id=facility_id)
-        except AttributeError:
+        except (AttributeError, Facility.DoesNotExist):
             # non existing facility
             return Response({})
 
--- a/kolibri/core/public/api.py
+++ b/kolibri/core/public/api.py
@@ -24,6 +24,7 @@ from rest_framework.response import Response
 from .. import error_constants
 from .constants.user_sync_statuses import QUEUED
 from .constants.user_sync_statuses import SYNC
+from kolibri.core.api import BaseValuesViewset
 from kolibri.core.api import ReadOnlyValuesViewset
 from kolibri.core.auth.models import Facility
 from kolibri.core.auth.models import FacilityUser
@@ -329,7 +330,7 @@ class SyncQueueViewSet(viewsets.ViewSet):
         return self.check_queue(request, pk=pk)
 
 
-class FacilitySearchUsernameViewSet(ReadOnlyValuesViewset):
+class FacilitySearchUsernameViewSet(BaseValuesViewset):
     filter_backends = (DjangoFilterBackend, filters.SearchFilter)
     filter_fields = ("facility",)
     search_fields = ("^username",)
@@ -338,11 +339,14 @@ class FacilitySearchUsernameViewSet(ReadOnlyValuesViewset):
 
     def list(self, request, *args, **kwargs):
         facility_id = request.query_params.get("facility", None)
+        if facility_id is None:
+            content = "Missing parameter: facility is required"
+            return Response(content, status=status.HTTP_412_PRECONDITION_FAILED)
         try:
             facility = Facility.objects.get(id=facility_id)
         except (AttributeError, Facility.DoesNotExist):
-            # non existing facility
-            return Response({})
+            content = "The facility does not exist in this device"
+            return Response(content, status=status.HTTP_404_NOT_FOUND)
 
         if facility.dataset.learner_can_login_with_no_password:
             queryset = self.filter_queryset(self.get_queryset())
@@ -352,7 +356,13 @@ class FacilitySearchUsernameViewSet(ReadOnlyValuesViewset):
             queryset = self.get_queryset().filter(
                 facility=facility_id, username=username
             )
-            response = {"username": username, "id": None} if queryset else {}
+            response = (
+                [
+                    {"username": username, "id": None},
+                ]
+                if queryset
+                else []
+            )
             return Response(response)
 
     def get_queryset(self):
--- a/kolibri/plugins/user_profile/api_urls.py
+++ b/kolibri/plugins/user_profile/api_urls.py
@@ -1,14 +1,10 @@
 from django.conf.urls import url
 
-from .viewsets import RemoteFacilitiesViewset
 from .viewsets import RemoteFacilityUserViewset
 from .viewsets import UserIndividualViewset
 
 urlpatterns = [
     url(r"userindividual", UserIndividualViewset.as_view(), name="userindividual"),
-    url(
-        r"remotefacilities", RemoteFacilitiesViewset.as_view(), name="remotefacilities"
-    ),
     url(
         r"remotefacilityuser",
         RemoteFacilityUserViewset.as_view(),
--- a/kolibri/plugins/user_profile/api_urls.py
+++ b/kolibri/plugins/user_profile/api_urls.py
@@ -1,7 +1,18 @@
 from django.conf.urls import url
 
+from .viewsets import RemoteFacilitiesViewset
+from .viewsets import RemoteFacilityUserViewset
 from .viewsets import UserIndividualViewset
 
 urlpatterns = [
     url(r"userindividual", UserIndividualViewset.as_view(), name="userindividual"),
+    # url(r"remotefacilities/(?P<baseurl>[^/]+)", RemoteFacilitiesViewset.as_view(), name="remotefacilities"),
+    url(
+        r"remotefacilities", RemoteFacilitiesViewset.as_view(), name="remotefacilities"
+    ),
+    url(
+        r"remotefacilityuser",
+        RemoteFacilityUserViewset.as_view(),
+        name="remotefacilityuser",
+    ),
 ]
--- a/kolibri/plugins/user_profile/api_urls.py
+++ b/kolibri/plugins/user_profile/api_urls.py
@@ -6,7 +6,6 @@ from .viewsets import UserIndividualViewset
 
 urlpatterns = [
     url(r"userindividual", UserIndividualViewset.as_view(), name="userindividual"),
-    # url(r"remotefacilities/(?P<baseurl>[^/]+)", RemoteFacilitiesViewset.as_view(), name="remotefacilities"),
     url(
         r"remotefacilities", RemoteFacilitiesViewset.as_view(), name="remotefacilities"
     ),
--- a/kolibri/plugins/user_profile/assets/src/composables/useIndividualDevice.js
+++ b/kolibri/plugins/user_profile/assets/src/composables/useIndividualDevice.js
@@ -9,6 +9,7 @@ export default function useIndividualDevice() {
     client({ url: urls['kolibri:kolibri.plugins.user_profile:userindividual']() }).then(
       response => {
         isIndividual.value = response.data.individual && !response.data.lod;
+        // isIndividual.value = true;
       }
     );
   });
--- a/kolibri/plugins/user_profile/assets/src/composables/useIndividualDevice.js
+++ b/kolibri/plugins/user_profile/assets/src/composables/useIndividualDevice.js
@@ -9,7 +9,6 @@ export default function useIndividualDevice() {
     client({ url: urls['kolibri:kolibri.plugins.user_profile:userindividual']() }).then(
       response => {
         isIndividual.value = response.data.individual && !response.data.lod;
-        // isIndividual.value = true;
       }
     );
   });
--- a/kolibri/plugins/user_profile/assets/src/machines/changeFacilityMachine.js
+++ b/kolibri/plugins/user_profile/assets/src/machines/changeFacilityMachine.js
@@ -1,3 +1,5 @@
+import client from 'kolibri.client';
+import urls from 'kolibri.urls';
 import { createMachine, assign } from 'xstate';
 // This machine can be visualized and tested at https://stately.ai/viz/c1316a38-033d-4c57-8d9f-e282310e341d
 /* SETCONTEXT event example:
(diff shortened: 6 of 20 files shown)


kolibri/plugins/user_profile/assets/src/composables/useIndividualDevice.js
```diff
@@ -9,6 +9,7 @@ export default function useIndividualDevice() {
     client({ url: urls['kolibri:kolibri.plugins.user_profile:userindividual']() }).then(
       response => {
         isIndividual.value = response.data.individual && !response.data.lod;
+        // isIndividual.value = true;
```
marcellamaki: Is this supposed to be commented out?

kolibri/plugins/user_profile/assets/src/machines/changeFacilityMachine.js
```diff
@@ -108,8 +111,8 @@ export const changeFacilityMachine = createMachine({
         SETCONTEXT: { actions: setInitialContext },
         SELECTFACILITY: {
           actions: assign({
-            targetFacility: () => {
-              return {};
+            targetFacility: (_, event) => {
```
marcellamaki, line 114: This isn't a code review specific question, it's just an "I don't know" question... what does the `_` represent?

kolibri/plugins/user_profile/assets/src/machines/changeFacilityMachine.js
```diff
@@ -157,13 +160,12 @@ export const changeFacilityMachine = createMachine({
...
         NEW: 'createAccount',
         CONTINUE: 'isAdmin',
-        BACK: 'changeFacility',
       },
     },
     isAdmin: {
       always: [
         {
-          cond: context => context.role === 'superadmin',
+          cond: context => context.role === 'superuser',
```
marcellamaki, line 169: Is this a _new_ role type? As in, there are now both superadmins and superusers, depending on other permissions and the type of facility? Or is it renaming 'superadmin' to 'superuser'

marcellamaki (commented):
I have done an initial code read through and though I left a few comments, I don't think anything is a concern so far. I would appreciate if @rtibbles could do a quick read through of the python changes in case there are concerns I might not be aware of, but in terms of readability for a less-expert python coder, I found the changes relatively easy to read through and follow. 

I have to set up some test facilities for manual QA. I will do that tomorrow.

kolibri/plugins/user_profile/assets/src/composables/useIndividualDevice.js
jredrejo: ah, no, I use that code to test in my laptop. I have more than one user in my facility and commenting it out I will see the option to change facility. I left it there on purpose in case someone else developing this part is in the same situation.
`isIndividual` is used to determine if the option to change facility is shown in the profile screen.

For production that can be deleted, that `isIndividual` property must be active in devices with only one user and not being a LOD.

kolibri/plugins/user_profile/assets/src/machines/changeFacilityMachine.js
jredrejo: it comes from my Python practices: if a function receives two parameters and you dont' use the first, you usually replace it by `_` So it represents some data received by this function that's not used inside the function.

kolibri/plugins/user_profile/assets/src/machines/changeFacilityMachine.js
jredrejo: actually, I thought it was something that had decided to be changed in 0.16. I just saw that I was receiving `superuser` instead of  `superadmin` as I expected and I changed it. I didn't investigate why, just supposed it was an intended change. If it is not, we'll have to find out where in the backend this has been changed.

kolibri/plugins/user_profile/assets/src/views/ChangeFacility/ConfirmChangeFacility.vue
```diff
@@ -0,0 +1,113 @@
...
+    },
+
+    $trs: {
+      documentTitle: {
+        message: 'Change Facility',
+        context: 'Title of this step for the change facility page.',
+      },
+      mergeAccounts: {
+        message: 'Merge Accounts',
+        context: 'Button for the merge accounts between facilities button.',
```
MisRob: I think the last word "button" might be a typo?

kolibri/plugins/user_profile/assets/src/views/ChangeFacility/ConfirmChangeFacility.vue
```diff
@@ -0,0 +1,113 @@
...
+      },
+      changeFacilityInfoLine1: {
+        message:
+          'You are about to move your account and progress data to ‘{target_facility}’ learning facility. Your data will still be available to you and will also be accessible to any administrators of this learning facility.',
+        context: 'First line of text explaining what changing to another learning facility means.',
```
MisRob, line 97: @radinamatic Are translators able to recognize components that strings are used for? Wondering if "First line of text"  is helpful since if translators would not be aware of it, they won't know where that first line is. The same question applies to translation strings with "Second line of text" and "Last line of text" in this and other components in this pull request.

kolibri/plugins/user_profile/assets/src/views/ChangeFacility/SelectFacility.vue
```diff
@@ -0,0 +1,279 @@
...
+    components: { AddAddressForm, BottomAppBar },
+
+    mixins: [responsiveWindowMixin, commonCoreStrings, commonSyncElements],
+    setup(props, context) {
+      const {
+        addresses: discoveredAddresses,
+        discoveringPeers,
+        discoveryFailed,
+        discoveredAddressesInitiallyFetched,
+      } = useDynamicAddresses(props);
```
MisRob, line 106: This component has no `props` so passing them feels a bit confusing but maybe it relates rather to the obscure way that `useDynamicAddresses` composable accepts arguments

kolibri/plugins/user_profile/assets/src/views/ChangeFacility/SelectFacility.vue
```diff
@@ -0,0 +1,279 @@
...
+      return {
+        availableAddressIds: [],
+        availableFacilities: [],
+        selectedFacilityId: '',
+        showAddAddressModal: false,
+      };
+    },
+    inject: ['changeFacilityService'],
+    computed: {
+      isAddressAvailable() {
```
MisRob, line 150: Was there any reason for having this defined as a function nested in a computed property when compared to using a function in `methods`?

kolibri/plugins/user_profile/assets/src/views/ChangeFacility/SelectFacility.vue
```diff
@@ -0,0 +1,279 @@
...
+        return function(addressId) {
+          return Boolean(this.availableAddressIds.find(id => id === addressId));
+        };
+      },
+    },
+    watch: {
+      selectedFacilityId(newVal) {
```
MisRob, line 157: I'd suggest comparing the previous value with the new value and only executing the code below if those two values are different to prevent unnecessary actions.

```js
selectedFacilityId(newVal, oldVal) {
  if (newValue === oldVal) { return; }
  ...
}

File: kolibri/plugins/user_profile/assets/src/views/ChangeFacility/SelectFacility.vue [hunk 9]
```diff
@@ -0,0 +1,279 @@
...
+  import commonCoreStrings from 'kolibri.coreVue.mixins.commonCoreStrings';
+  import commonSyncElements from 'kolibri.coreVue.mixins.commonSyncElements';
+  import client from 'kolibri.client';
+  import urls from 'kolibri.urls';
+  import BottomAppBar from 'kolibri.coreVue.components.BottomAppBar';
+  import useSavedAddresses from '../../../../../../core/assets/src/views/sync/SelectAddressModalGroup/useSavedAddresses.js';
+  import useDynamicAddresses from '../../../../../../core/assets/src/views/sync/SelectAddressModalGroup/useDynamicAddresses.js';
+  import AddAddressForm from '../../../../../../core/assets/src/views/sync/SelectAddressModalGroup/AddAddressForm';
+
+  export default {
```
MisRob, line 90: When creating completely new components that use Composable API, I don't see a reason why to mix it with the Options API even though it's technically compatible. I think this strategy is useful when introducing composables to existing components because it doesn't require refactoring a whole component, however for new components I'd say it's a bit confusing since you don't know if you should go to `setup` or to the options. I know about one pitfall with accessing translations in `setup` so I think we may need to use the Options API for properties that need to access `$tr` but other than that, were there any other reasons for mixing both APIs?

kolibri/plugins/user_profile/assets/src/views/ChangeFacility/SelectFacility.vue
```diff
@@ -0,0 +1,279 @@
...
+  <div>
+    <span class="headercontainer">
+      <h1>{{ $tr('documentTitle') }}</h1>
+
+      <transition name="spinner-fade">
+
+        <div v-if="discoveringPeers">
+          <KLabeledIcon>
+            <template #icon>
+              <KCircularLoader :size="16" :stroke="6" class="loader" />
```
MisRob: `"loader"` class doesn't seem to be used

kolibri/plugins/user_profile/assets/src/views/ChangeFacility/SelectFacility.vue
```diff
@@ -0,0 +1,279 @@
...
+    </span>
+    <p v-if="initialFetchingComplete && !availableFacilities.length">
+      {{ $tr('noFacilitiesText') }}
+    </p>
+    <template v-for="f in availableFacilities">
+      <div :key="`div-${f.id}`">
+        <KRadioButton
+          :key="f.id"
+          v-model="selectedFacilityId"
+          class="radio-button"
```
MisRob: This class doesn't seem to be used

kolibri/plugins/user_profile/assets/src/views/ChangeFacility/SelectFacility.vue
```diff
@@ -0,0 +1,279 @@
...
+        marginTop: '34px',
+        paddingTop: '10px',
+        borderTop: `1px solid ${$themePalette.grey.v_300}`
+      }"
+    >
+
+      <KGridItem>{{ $tr('doNotSeeYourFacility') }}</KGridItem>
+      <KGridItem>
+        <KButton
+          class="new-address-button"
```
MisRob: This class doesn't seem to be used

kolibri/plugins/user_profile/assets/src/views/ChangeFacility/ToBeDone.vue
```diff
@@ -0,0 +1,71 @@
+<template>
+
```
MisRob, line 2: Nice, thank you

kolibri/plugins/user_profile/assets/src/views/ChangeFacility/index.vue
```diff
@@ -1,62 +1,30 @@
 <template>
 
-  <CoreBase
```
MisRob, line 3: :tada:

What are your impressions of the new pages architecture in the context of using it together with the state machine?

kolibri/plugins/user_profile/assets/src/routes.js
```diff
@@ -52,68 +56,74 @@ export default [
     children: [
       {
         path: 'change',
-        name: 'CHANGE_FACILITY',
-        component: ChangeFacility,
+        name: 'SELECT_FACILITY',
+        component: SelectFacility,
       },
+      {
+        path: 'change',
```
MisRob: Was using the same `path: 'change'` twice, for both `SELECT_FACILITY` and `CONFIRM_CHANGE_FACILITY`, intentional?

kolibri/plugins/user_profile/assets/src/views/ChangeFacility/index.vue
```diff
@@ -1,62 +1,30 @@
...
-    :immersivePage="false"
-    :appBarTitle="coreString('profileLabel')"
-    :immersivePagePrimary="true"
-  >
-    <KPageContainer>
+  <NotificationsRoot>
+    <div class="main-wrapper" :style="wrapperStyles">
```
MisRob: `"main-wrapper"` class doesn't seem to be used

kolibri/plugins/user_profile/assets/src/views/ChangeFacility/index.vue
```diff
@@ -93,13 +61,27 @@
...
+          paddingTop: this.appBarHeight + 16 + 'px',
+        };
+      },
     },
+
     created() {
       this.service.start();
       this.service.onTransition(state => {
         const stateID = Object.keys(state.meta)[0];
+        console.log(state);
```
MisRob: Not such a big deal to use console logging as the whole feature will be a work in progress for some time but am surprised that lint checks didn't discover it. Maybe better to remove it?

MisRob (commented):
Thank you, @jredrejo.

I've read through code changes and left some comments. None of them are blocking. 

From the high-level perspective, I haven't noticed anything suspicious and the implementation seems to be clear and easy to understand, though I will certainly need to spend some time studying the state machine to get the whole feature better (will send you a message regarding the hangout we chatted about together earlier)

kolibri/plugins/user_profile/assets/src/views/ChangeFacility/ConfirmChangeFacility.vue
jredrejo: mmm, I don't get it, the context says where the first line is, same for the following

kolibri/plugins/user_profile/assets/src/views/ChangeFacility/ConfirmChangeFacility.vue
jredrejo: yeah, too many copy and paste between texts, thanks!

kolibri/plugins/user_profile/assets/src/views/ChangeFacility/SelectFacility.vue
jredrejo: yes, for reactivity. This computed is using the composition api, it's not an "usual" computed section in a vue 2 component.

kolibri/plugins/user_profile/assets/src/views/ChangeFacility/SelectFacility.vue
jredrejo: are you sure this is needed if `newVal` is just a string and not an object? I think the watcher is not triggered unless `selectedFacilityId` changes

kolibri/plugins/user_profile/assets/src/views/ChangeFacility/SelectFacility.vue
jredrejo: sorry, I don't understand this comment. The api I'm using there is not created in this PR, it's an existing api that already exists in Kolibri and it's used in other components. The components are new and they're using these functions in the `setup` section. What are you proposing exactly?

kolibri/plugins/user_profile/assets/src/views/ChangeFacility/SelectFacility.vue
jredrejo: yeah, good catch

kolibri/plugins/user_profile/assets/src/views/ChangeFacility/SelectFacility.vue
jredrejo: yep, too many refactorings, thanks!

kolibri/plugins/user_profile/assets/src/views/ChangeFacility/index.vue
jredrejo: honestly, I don't appreaciate any difference with the way the machine is triggering its changes in this case, using the routes.

kolibri/plugins/user_profile/assets/src/views/ChangeFacility/index.vue
jredrejo: fixed, thanks!

kolibri/plugins/user_profile/assets/src/views/ChangeFacility/index.vue
jredrejo: happy to remove it. I had left a couple of debugging lines in the code to facilitate the work of whoever was going to continue with the following screens, but it's clear it's not needed ;)

kolibri/plugins/user_profile/assets/src/views/ChangeFacility/SelectFacility.vue
jredrejo: actually, the `props` object exists, even if it's not defined, and the `useDynamicAddresses` function uses it, so I think it's safer passing it than a plain `{}`

kolibri/plugins/user_profile/assets/src/routes.js
jredrejo: No, it was an error, thanks!

kolibri/plugins/user_profile/assets/src/views/ChangeFacility/ConfirmChangeFacility.vue
MisRob: I was asking rather if translators know about the context in terms of the UI feature or part of the feature like the modal. Because if not, they might see for example:

`First line of text explaining what changing to another learning facility means.`
`First line of text confirming the username and facility where the user is changing.`
`First line of text ...`
`Second line of text...`

and then I am not sure if it's possible to make sense out of it.

But would need to hear from @radinamatic I think because I'm not sure if translators have some more information available from namespaces, for example. No need to worry about it in this PR way too much, it's rather a general question to learn a bit more about how to write context strings.

kolibri/plugins/user_profile/assets/src/views/ChangeFacility/SelectFacility.vue
MisRob: No, I'm not 100% certain about the object vs string behavior, would need to try. If you're sure it's not triggered, then no problem.

kolibri/plugins/user_profile/assets/src/views/ChangeFacility/SelectFacility.vue
MisRob: I meant Vue APIs
- Composition API - using `setup` and defining methods, computed properties, watchers etc. inside of `setup`
- Options API - declared options (methods, computed, watchers etc.), defined as properties on the object being exported

For example, in this component, `combinedAddresses` is implemented as a computed property by means of the Composition API whereas `isAddressAvailable` is implemented as a computed property by means of the Options API.

I'm proposing using one consistent style when creating new components instead of mixing these two.

MisRob: No need to refactor this one, but would be nice to keep this in mind whenever we know that we need to use the Composition API in a new component. Then we can implement it fully using Composition API so that in the future there's no need to spend time locating methods, watchers, and all the things in two places.

kolibri/plugins/user_profile/assets/src/machines/changeFacilityMachine.js
marcellamaki: @rtibbles - is this something you can answer? 😄

marcellamaki:
@radinamatic, based on the feedback from me and @MisRob, I think there are no more significant code changes to be made here, and this is ready for a quick QA review before ✅ I did a little bit of manual testing but would appreciate an extra set of eyes, especially with some of the different facility scenarios, in case I missed something. thank you!!

kolibri/plugins/user_profile/assets/src/machines/changeFacilityMachine.js
rtibbles: I am fairly sure we've always used `superuser` as the constant value. But we made a design decision to change all references in user facing text at some point to call them a `superadmin` to better relate the role to the `admin` role.

Referencing the core `UserKinds` constants, we have been using `superuser` for some time without any change: https://github.com/learningequality/kolibri/blob/develop/kolibri/core/assets/src/constants.js#L16 (that line has not been updated in 6 years).

kolibri/core/auth/serializers.py
```diff
@@ -133,13 +133,23 @@ class Meta:
 
 class PublicFacilitySerializer(serializers.ModelSerializer):
     learner_can_login_with_no_password = serializers.SerializerMethodField()
+    learner_can_sign_up = serializers.SerializerMethodField()
 
     def get_learner_can_login_with_no_password(self, instance):
         return instance.dataset.learner_can_login_with_no_password
 
+    def get_learner_can_sign_up(self, instance):
+        return instance.dataset.learner_can_sign_up
```
rtibbles, line 142: I know it was like that when you got here - but these kind of n queries chained look ups always bother me.

kolibri/plugins/user_profile/api_urls.py
```diff
@@ -1,7 +1,18 @@
 from django.conf.urls import url
 
+from .viewsets import RemoteFacilitiesViewset
+from .viewsets import RemoteFacilityUserViewset
 from .viewsets import UserIndividualViewset
 
 urlpatterns = [
     url(r"userindividual", UserIndividualViewset.as_view(), name="userindividual"),
+    # url(r"remotefacilities/(?P<baseurl>[^/]+)", RemoteFacilitiesViewset.as_view(), name="remotefacilities"),
```
rtibbles: Can remove the commented line?

kolibri/core/public/api.py
```diff
@@ -324,3 +326,16 @@ def create(self, request):
 
     def update(self, request, pk=None):
         return self.check_queue(request, pk=pk)
+
+
+class FacilitySearchUsernameViewSet(ReadOnlyValuesViewset):
```
rtibbles: This is a spec issue rather than an implementation issue - but I think we might want to revisit how much information about users we let be publicly accessible.

rtibbles (commented):
No immediate concerns with implementation - we might want to revisit the spec from a security perspective though.

kolibri/core/public/api.py
jredrejo: Changed to return just if the username exists when password is needed to login. If not, it does an actual search and returns matching users with their ids

kolibri/core/device/api.py
```diff
@@ -369,3 +374,28 @@ def list(self, request):
...
+    :return : json object containing the list of facilities of the device, with their id, name, learner_can_sign_up and learner_can_login_with_no_password info
+    """
+
+    permission_classes = (IsNotAnonymous,)
+
+    def get(self, request):
+        baseurl = request.query_params.get(
+            "baseurl", request.build_absolute_uri("/")[:-1]
+        )
+        path = reverse("kolibri:core:publicfacility-list")
```
rtibbles: Will need to `lstrip("/")` to ensure it concatenates properly with the baseurl below, in the case where the baseurl has a prefix.

See here where I've just fixed this for the same endpoint in 0.15: https://github.com/learningequality/kolibri/pull/9533/files#diff-d4ed59d24014dff73b18af199c3aa7484e644edae6e7cb162125ddf6d8c8f349R129

kolibri/core/public/api.py
```diff
@@ -324,3 +327,35 @@ def create(self, request):
...
+    filter_backends = (DjangoFilterBackend, filters.SearchFilter)
+    filter_fields = ("facility",)
+    search_fields = ("^username",)
+
+    values = ("id", "username")
+
+    def list(self, request, *args, **kwargs):
+        facility_id = request.query_params.get("facility", None)
```
rtibbles, line 341: Not blocking, but could maybe return a 412 here, if the facility_id is required but not given.

kolibri/core/public/api.py
```diff
@@ -324,3 +327,35 @@ def create(self, request):
...
+        try:
+            facility = Facility.objects.get(id=facility_id)
+        except (AttributeError, Facility.DoesNotExist):
+            # non existing facility
+            return Response({})
```
rtibbles: Maybe should be a 404?

kolibri/core/public/api.py
```diff
@@ -324,3 +327,35 @@ def create(self, request):
...
+
+        if facility.dataset.learner_can_login_with_no_password:
+            queryset = self.filter_queryset(self.get_queryset())
+            return Response(self.serialize(queryset))
+        else:
+            username = request.query_params.get("search", None)
+            queryset = self.get_queryset().filter(
+                facility=facility_id, username=username
+            )
+            response = {"username": username, "id": None} if queryset else {}
```
rtibbles: I think we should return a list here rather than a single object to give a consistent return object regardless of the facility setting.

kolibri/core/public/api.py
```diff
@@ -324,3 +327,35 @@ def create(self, request):
 
     def update(self, request, pk=None):
         return self.check_queue(request, pk=pk)
+
+
+class FacilitySearchUsernameViewSet(ReadOnlyValuesViewset):
```
rtibbles: I think you only want to serve list requests here, so probably better to inherit from `BaseValuesViewset` - otherwise this endpoint is exposing a detail endpoint as well, with no permission checks, and the ability to look up any user by id.

kolibri/plugins/user_profile/viewsets.py
```diff
@@ -22,3 +26,26 @@ def get(self, request, format=None):
...
+class RemoteFacilityUserViewset(APIView):
+    def get(self, request):
+        baseurl = request.query_params.get(
+            "baseurl", request.build_absolute_uri("/")[:-1]
+        )
+        username = request.query_params.get("username", None)
+        facility = request.query_params.get("facility", None)
+        if username is None or facility is None:
+            raise ValidationError(detail="Both username and facility are required")
+        path = reverse("kolibri:core:publicsearchuser-list")
```
rtibbles: `.lstrip("/")`

rtibbles (changes requested):
A few things that can be quickly cleaned up:
* Adding the appropriate `lstrip` methods to URLs generated from Django reverse.
* Inheriting from the `BaseValuesViewset`
* Consistent return object type from the FacilitySearchUsernameViewset

kolibri/core/device/api.py
jredrejo: thanks, I had considered it when using `build_absolute_uri`, but not when a parameter is added. Anyway I've prefered to use `rstrip` to limit `baseurl`. Addinig `lstrip` to the path could cause problems if baseurl does not have the ending `"/"`

kolibri/core/public/api.py
jredrejo: sure, much better. Done!

kolibri/core/public/api.py
jredrejo: yes, better

kolibri/core/public/api.py
jredrejo: fully agree, thanks!

kolibri/core/public/api.py
jredrejo: Of course, I'm afraid I don't remember all the possible viewsets we have in the kolibri code and I picked the first I saw that did the work. Looking at it it's clear using `BaseValuesViewset` limits perfectly the scope.

kolibri/core/device/api.py
rtibbles: `rstrip` on the baseurl gives the wrong result though.

If you do this:
```
urljoin("http://127.0.0.1:8080/kolibri", "/api/public/facility")
```
then you get this output:
```
'http://127.0.0.1:8080/api/public/facility'
```
Which is precisely what we are trying to circumvent.

jredrejo: Ok, I never thought `base_url` would have anything else but the server and the port. I've just changed it according to your suggestion.

rtibbles (approved):
All comments addressed.

PR #9514, opened 2022-06-14
add .trim to v-model for username

### LINKED ISSUE #9471: Trim whitespace during log in authentication/validation
## Observed behavior

During the Uganda training, the team ran into the following problem with logging in to Kolibri: if the participant added a spacebar after their username, it was considered as a wrong username. For example -if the username was ‘learner2’ and if the participant typed in ‘learner2  ’ - with an extra space after 2, Kolibri wasn’t accepting it and said that the username was incorrect
It was difficult to initially figure out what was wrong because the user name was correct. 

## Expected behavior
Authentication and validation should correctly trim whitespace so as to not cause an error

## User-facing consequences
Confusion! Difficulties logging in

## Steps to reproduce
Attempt to into kolibri with a valid account, but add a space at the end of the username

marcellamaki:
## Summary

Adds `.trim` to the v-model for username on login, so that adding a space before or after the username does not cause a validation failure.

## References
Fixes #9471

whitespace-login

Screen Shot 2022-06-14 at 8 53 41 AM

## Reviewer guidance
1. Log in with an existing account, adding a space both before and after the username. The log in should work
2. After entering a username with whitespace and advancing to enter the password, click "change user". This should work and return to a blank form 
3. Add a space or a special character in the middle of the username. This should fail validation

----

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/plugins/user_auth/assets/src/views/SignInPage/index.vue
+++ b/kolibri/plugins/user_auth/assets/src/views/SignInPage/index.vue
@@ -58,7 +58,7 @@
             <KTextbox
               id="username"
               ref="username"
-              v-model="username"
+              v-model.trim="username"
               autocomplete="username"
               :autofocus="true"
               :label="coreString('usernameLabel')"


radinamatic:
Trimming the before and after white space in the username (and reporting an error if it's in the middle) is working a-Ok! 👍🏽 

I'm only wondering if we should do the same for passwords...? 🤔 
In the recording below, the same situation occurs when an accidental white space is introduced before or after the password, resulting in the ``Incorrect password`` message. Thinking really hard if there are any downsides to implementing the same trimming here, and I'm coming up empty...

trim

rtibbles:
> Thinking really hard if there are any downsides to implementing the same trimming here, and I'm coming up empty...

The main downside I can see is that if someone does have a space at the beginning or end of their password, they will never be able to login.

I did a quick scan of the code, and I can't find any place that we do actually remove whitespace from passwords, so it's very possible these may exist. It's also possible I missed where we do this - so if we are already making sure of this, then this change would be ok.

radinamatic:
> if someone does have a space at the beginning or end of their password, they will never be able to login.

But no one is able to save (create account using) the password **with** the whitespace anyways, as there's an error notification in place for that, right?

rtibbles:
> as there's an error notification in place for that, right?

Is there? That's what I was unsure about - whether we do trim or error in this case.

radinamatic:
I was convinced there was the error notification in that case, but after re-checking both 0.15.2 and the asset from this PR, I'm not able to find one (must have been 🤓 dreaming...).  

So yes, given the above it is unwise to implement the trim for passwords, let's merge this then!

radinamatic (approved):
LGTM! 💯 :shipit:

rtibbles:
OK, I think github actions have got confused because this was originally targeted to the develop branch, so the Python 3.4 and 3.5 tests did not run. They should not be affected, so this is safe to merge.

PR #9534, opened 2022-06-23
Onboarding import facility - Mostly designs

### LINKED ISSUE #9310: Design Update: All steps in ImportFacilitySetup component
# Overview

- The whole importFacilitySetup flow has a new design.

# Acceptance Criteria

- [ ] Apply the new designs in Figma
- [ ] Update the ImportFacilitySetup component as needed

## Desktop

Import facility (Figma - [Desktop](https://www.figma.com/file/MpCG7r5PULOCJsEvnTSti8/Android-Designs-2022?node-id=1%3A672) | [Mobile](https://www.figma.com/file/MpCG7r5PULOCJsEvnTSti8/Android-Designs-2022?node-id=1%3A1833))
Loading Facility (Figma - [Desktop](https://www.figma.com/file/MpCG7r5PULOCJsEvnTSti8/Android-Designs-2022?node-id=1%3A793) | [Mobile](https://www.figma.com/file/MpCG7r5PULOCJsEvnTSti8/Android-Designs-2022?node-id=1%3A1842) [Mobile 2](https://www.figma.com/file/MpCG7r5PULOCJsEvnTSti8/Android-Designs-2022?node-id=1%3A1851))
Select super admin (Figma - [Desktop](https://www.figma.com/file/MpCG7r5PULOCJsEvnTSti8/Android-Designs-2022?node-id=1%3A809) | [Mobile](https://www.figma.com/file/MpCG7r5PULOCJsEvnTSti8/Android-Designs-2022?node-id=1%3A1815))
Responsibilities as an administrator (handled in #9309)

nucleogenesis:
## Summary

- Applies new OnboardingStepBase styles to Import Facility flow

https://user-images.githubusercontent.com/6356129/175645233-8e654da9-6a4d-4071-9c1c-9613de103c2a.mp4

- Moves syncTaskUtils into core as it's used across multiple plugins. 
- Avoids errors re: using i18n before it is ready by assigning bound functions to the t8n maps. This could be done instead by making those `thingToTranslationMap` objects into functions that return objects but I did this first and feels like half-dozen in one hand and six in the other in this case.
- Applying style and data management pattern with ImportFacilitySetup as a coordinator component
- Adds event listener and handler to OnboardingStepBase so that you can hit enter to continue while having an INPUT element focused (any form element, radio button).

**_Not included in this PR because the main focus is applying designs_**

- Import actually working
- Robust a11y review re: keyboard navigation, focus trapping modals, etc - needs testing
- Proper handling of when and how users can "Go back" at any point and how to display buttons/options in the case of a failed import and during the importing loading process (ie, what buttons to show) _some tests are skipped right now until we work this out_

## References

Fixes #9310 

## Reviewer guidance

Please walk through the flow OR watch the video above. If you get stuck, wipe your browser's local data and refresh the page. 

- Do designs look to spec?
- a11y notes would be much appreciated, although fixes for them will be put into a separate issue and fixed in follow up

Code review: FWIW I tried to make my commits pretty self-contained overall
- Is it clear why the functions are bound when defined in the now moved to core syncTaskUtils module? Should those maps just be returned by a function instead?
- Does the move, generally, make architectural sense?

----

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated


radinamatic:
Designs look to spec, and I found 2 missed instances of the proper focus trap:

1. Language selector - if the user presses the ``Tab`` key **after the ``CONFIRM`` button** the focus should return to the top, at the first language radio button (which allows to use arrow keys to potentially select another language), but instead it exits the modal and ends up in the browser address bar.
   
    onboarding-lang-modal-focus-not-trapped


2. When the user selects the option to import a facility and presses the ``Enter`` key while the ``CONTINUE`` button is in focus, **Select network address** modal should trap the focus, but instead it is again located behind, in the browser address bar.

    onboarding-focus-trap-select-network

Question for @jtamiace (added a comment in Figma too): Wouldn't it make more sense that in case where there are devices detected at this step, the ``Add new address`` link came **after** the detected devices in the network?

radinamatic:
Radio button text should overflow to the next line, not be cut out with ellipsis.

@marcellamaki Is that something that we should fix in the KDS component?

onboarding-mobile-ellipsis

kolibri/core/assets/src/utils/syncTaskUtils.js
```diff
@@ -1,7 +1,45 @@
...
-import taskStrings from 'kolibri.coreVue.mixins.commonTaskStrings';
-import bytesForHumans from 'kolibri.utils.bytesForHumans';
-import { TaskStatuses, TaskTypes } from '../constants';
+import commonCoreStrings from '../mixins/commonCoreStrings';
+import taskStrings from '../mixins/taskStrings';
+import bytesForHumans from '../utils/bytesForHumans';
+
+export const TaskTypes = {
+  REMOTECHANNELIMPORT: 'kolibri.core.content.tasks.remotechannelimport',
+  REMOTECONTENTIMPORT: 'kolibri.core.content.tasks.remotecontentimport',
```
marcellamaki, line 7: Not blocking: This is a bit nit-picky, and I can see that this was just a "move existing code" but if it's possible, would we have these be more like `REMOTE_CHANNEL_IMPORT` without it being too much of a hassle? I think it would be a lot easier to read. If easier to keep as is, because these are in too many places, no problem though

kolibri/core/assets/src/utils/syncTaskUtils.js
```diff
@@ -1,7 +1,45 @@
...
+  RUNNING: 'RUNNING',
+  QUEUED: 'QUEUED',
+  SCHEDULED: 'SCHEDULED',
+  CANCELED: 'CANCELED',
+  CANCELING: 'CANCELING',
+});
+
+export const TransferTypes = {
+  LOCALEXPORT: 'localexport',
+  LOCALIMPORT: 'localimport',
```
marcellamaki, line 39: Same as above

kolibri/plugins/setup_wizard/assets/src/views/OnboardingStepBase.vue
```diff
@@ -139,6 +147,15 @@
         return availableLanguages[currentLanguage];
       },
     },
+    methods: {
+      /* If the user is focused on a form element and hits enter, continue */
+      handleEnterKey(e) {
```
marcellamaki, line 152: nice! not for now, but I do wonder if we should implement this more widely? 
it is my personal number 1 pet peeve when typing in a text field that enter is not the same as continue (if I'm not focused on the button)

marcellamaki:
> Radio button text should overflow to the next line, not be cut out with ellipsis.

@radinamatic - yes, this is something we should fix in KDS. I am going to work on putting together a KDS dev issues milestone in support of 0.16 for us to review before our retro. I think we have an open issue for this, but will double check!

marcellamaki (commented):
> Is it clear why the functions are bound when defined in the now moved to core syncTaskUtils module? 

It is not clear to me, but overall reading through the code, I did find it easy enough to follow? 

As for whether moving is a good idea, I think on a basic level, having the things in a core util is helpful after a certain point, but I don't know enough context of if there are cons to this. would defer to @rtibbles on that one.

My other comments are small and not blocking. Design/figma look good to me

jtamiace:
> Wouldn't it make more sense that in case where there are devices detected at this step, the Add new address link came after the detected devices in the network?

Yes I agree, and it looks like this is the layout for the other "Select network address" modals already existing in Kolibri. Made the change to all the modals for onboarding in Figma in case that's still helpful

Screen Shot 2022-06-28 at 12 44 30 PM

kolibri/core/assets/src/utils/syncTaskUtils.js
```diff
@@ -1,7 +1,45 @@
-import coreStrings from 'kolibri.coreVue.mixins.commonCoreStrings';
-import taskStrings from 'kolibri.coreVue.mixins.commonTaskStrings';
-import bytesForHumans from 'kolibri.utils.bytesForHumans';
-import { TaskStatuses, TaskTypes } from '../constants';
+import commonCoreStrings from '../mixins/commonCoreStrings';
+import taskStrings from '../mixins/taskStrings';
```
rtibbles: I know it was like this when you got here, but for the taskStrings at least, the mixins module exposes an export for the function: https://github.com/learningequality/kolibri/blob/60dd14c37ecf8d3971b5013290c8a40ca9599668/kolibri/core/assets/src/mixins/taskStrings.js#L119

Which should avoid having to do an awkward destructing off the methods property of the mixin itself.

Would be good to do something similar for commonCoreStrings.

kolibri/core/assets/src/utils/syncTaskUtils.js
```diff
@@ -31,24 +69,28 @@ const syncTaskStatusToStepMap = {
...
+// syncStatusToDescriptionMap[SyncTaskStatuses.SESSION_CREATION]();
+// Note that when getTaskString or coreString is called within a function, this is unnecessary.
 const genericStatusToDescriptionMap = {
-  [TaskStatuses.PENDING]: getTaskString('taskWaitingStatus'),
-  [TaskStatuses.QUEUED]: getTaskString('taskWaitingStatus'),
-  [TaskStatuses.COMPLETED]: getTaskString('taskFinishedStatus'),
-  [TaskStatuses.CANCELED]: getTaskString('taskCanceledStatus'),
-  [TaskStatuses.CANCELING]: getTaskString('taskCancelingStatus'),
-  [TaskStatuses.FAILED]: getTaskString('taskFailedStatus'),
+  [TaskStatuses.PENDING]: getTaskString.bind(null, 'taskWaitingStatus'),
```
rtibbles: This does indeed work, but I wonder if just an inline anonymous arrow function might be more comprehensible?

`() => getTaskString('taskWaitingStatus')`

kolibri/plugins/device/assets/src/views/FacilitiesPage/FacilitiesTasksPage.vue
```diff
@@ -39,13 +39,13 @@
 
 <script>
 
+  import { FacilityTaskPanel } from 'kolibri.coreVue.componentSets.sync';
```
rtibbles, line 41: How do you feel about experimenting with putting some common code (but not really core code) into a module inside `packages/kolibri-common` or some such?

Should then be referenceable as `kolibri-common/path/to/component`.

rtibbles (commented):
`.bind(null...` feels a little opaque.

Wonder if we could start to experiment with another way of sharing dependencies that doesn't overload the core API?

kolibri/core/assets/src/utils/syncTaskUtils.js
nucleogenesis: :facepalm: good call :)

kolibri/plugins/device/assets/src/views/FacilitiesPage/FacilitiesTasksPage.vue
nucleogenesis: If that would help lighten our default front-end bundle load at all with tree shaking, then I think it's a good call for sure. Otherwise I'd be hesitant to separate anything from the apiSpec since, in spite of it being a big monolith, is an easy thing to reason about and change - at least it is once it's clear the object maps to the aliased paths in imports.

rtibbles: Yes - I think a fair amount of the code that is in the default frontend bundle falls into this category, so if we can experiment with this here, it would set us up to migrate more things out of the frontend bundle, and reduce the size considerably, I think.

kolibri/core/assets/src/utils/syncTaskUtils.js
nucleogenesis: The `coreString` signature maps directly to `$tr` a la `(key : string, args: object)` rather than taking some arbitrary number of nameless args - I'll apply the same here to make it more readable / consistent with the other code.

The `commonCoreStrings` pattern though for using it in JS is like:

```
import { coreStrings } from 'kolibri...commonScoreStrings'; // imports the created translator

coreStrings.$tr(...);
```

So maybe it'd be good to export a getter function there as well and unify the pattern altogether. In which case maybe I can move this to a separate issue re: strings things to do before release?

kolibri/core/assets/src/utils/syncTaskUtils.js
nucleogenesis: :100: no prob at all good idea

kolibri/core/assets/src/utils/syncTaskUtils.js
rtibbles: Yeah, I think having the explicit args is better - might be good to open a follow up and hash it out there, yes. I can imagine there will probably be other things to consider before we start moving stuff around more!

nucleogenesis:
@radinamatic I've addressed the comments above.

Note that the solution to wrapping the radio button label requires overriding a default setting in KDS that KRadioButtons be truncated with ellipsis.

@rtibbles all feedback is addressed - I went to address your comment re: the exporting the translator method along with the mixins and create an issue bit figured it'd be just about as quick to just do some of the work up front here https://github.com/learningequality/kolibri/pull/9588

Also, created new issue re: a common module here https://github.com/learningequality/kolibri/issues/9589

radinamatic:
Both missed focus traps are now working well, and the rest of my comments addressed! 👏🏽 

> Note that the solution to wrapping the radio button label requires overriding a default setting in KDS that KRadioButtons be 
truncated with ellipsis.

If I understood @marcellamaki correctly, we are going to address that in KDS before 0.16 is released.

I'll wait for @pcenov to finish his round of testing to approve.

pcenov:
Here is what I was able to identify as possible issues:

| Issue| Screenshot |
| Select facility - 1. Which is the final label? 2. The 'Import learning facility - 1 of X' text is missing. |  |
| Import learning facility - Missing device name, network and facility name values |  |
| Import learning facility - Missing facility name |  |

radinamatic:
Thank you @pcenov, I was mostly concentrated on keyboard navigation, and missed to report these string discrepancies.
 
@nucleogenesis, note that all instances of the string `facility` in the setup wizard are now `learning facility` to give better context for localization. 

I can also replicate the rest of the issues reported above: missing setup step, device name, string change in the third screenshot (`learning facility` is added) plus the facility name is not displayed.

nucleogenesis:
@radinamatic @pcenov 

**re: KRadioButton** - You're right, this will be dealt with in KDS https://github.com/learningequality/kolibri-design-system/issues/302 

**re: "Import learning facility - Missing device name, network and facility name values"** - There are two forms of this message - the one shown on my side is for when there are multiple facilities so the facility name is mentioned in the message there. The one @pcenov shared is for when there is one facility on the device.

**re: Import learning facility - Missing device name, network and facility name values** - fixed in https://github.com/learningequality/kolibri/pull/9534/commits/ad797f8825ec9407236090dab892d7e4b69802e8

**re: string updates**
@radinamatic can you make an issue re: the string updates. There are a lot of uses of the word "facility" - so it may be a bit of work to do that might be a good issue to introduce people to strings later on.

pcenov:
@radinamatic The following issues are still present in: https://buildkite.com/learningequality/kolibri-debian/builds/4738:

| Issue| Screenshot |
| Select facility - 1. Which is the final label? 2. The 'Import learning facility - 1 of X' text is missing. |  |
| Import learning facility - Missing device name, network and facility name values |  |

The facility name is displayed correctly now at the last step of 'Import learning facility, however the 'is' is missing:


@nucleogenesis Will those be fixed here or should I report them as separate issues?

radinamatic:
We agreed with @nucleogenesis to open follow up issue to review the strings, so this should be good to go! 🎉

radinamatic (approved):
Manual QA is looking good! 💯 :shipit:

Issue #9091, opened 2022-02-09
Sync facility data - Constantly seeing 'Most recent sync failed' message


pcenov:
## Observed behavior
Whenever I attempt to sync my facility to KDP I'm seeing a 'Most recent sync failed' message but if I manually refresh the page I can see that the sync has been successful.

## Expected behavior
The user should be shown the correct status of the sync procedure.

## Steps to reproduce the issue
1. Install the following build: https://buildkite.com/learningequality/kolibri-windows/builds/3136#308b9948-f6b3-466a-a661-162ed2f97d63
2. Go to Faclity>Data and scroll down to the 'Sync facility data' section
3. Register your facility to KDP
4. Click the Sync button and wait for the confirmation message.

## Screenshots
![2022-02-09_17-34-42](https://user-images.githubusercontent.com/79847249/153246018-34fb93f0-7083-4031-b8f4-d38cc0aacee5.png)
![2022-02-09_18-31-54](https://user-images.githubusercontent.com/79847249/153246027-2eb392fe-40ad-4407-9de3-55104c22679d.png)

## Usage Details
Windows 10, Ubuntu - Chrome

pcenov:
@radinamatic

radinamatic:
I'm also able to replicate this with 0.15.1b0 on Windows 10, sync is reported as failed in the UI until the page is reloaded.

![kdp3](https://user-images.githubusercontent.com/1457929/153287639-8e206b9a-e376-4f50-9ad6-5c1376b2e02f.gif)

radinamatic:
@marcellamaki Again, impacts the UX, but syncing is still working: not broken, just user facing incorrectly reported until the reload. Priority to be confirmed.

marcellamaki:
@radinamatic and @pcenov  --- thank you! and this maybe be more important than the buttons, since it seems to the user like an _error_

MisRob:
> Again, impacts the UX, but syncing is still working: not broken, just user-facing incorrectly reported until the reload. Priority to be confirmed.

@radinamatic To clarify "syncing is still working", did you mean that
(1) it appeared to work after the page reload because the success message was displayed
(2) data has been synced to KDP

?

pcenov:
@MisRob, yes, both of the above - the syncing is working because after a page reload we see the success message and also the data has been synced to KDP.

radinamatic:
[...]

PR #9541, opened 2022-06-28
Improve facility sync status reporting to users

MisRob:
## Summary

Previously, it wasn't possible to find out whether the last sync was successful or if it failed from the facility object itself. We were only able to derive information about failure from task objects while a task was running and that information wasn't persisted on the page reload, resulting in confusing UX.

This PR improves facility sync status reporting to users by persisting information about the success or failure of a facility sync task on the facility object by replacing the `last_synced` field by `last_successful_sync` and `last_failed_sync` fields (that said, we still continue using the failed task information from the task object to display fast update on the sync process to users without the need to re-fetch a facility when the task fails)

### Preview of possible states

|  Never synced  | There was no successful sync in the past and the most recent sync failed | There was a successful sync in the past and the most recent sync failed | The most recent sync was successful | 
| --- | --- | --- | --- |
| never-synced | sync-failed | success-fail | success |

## References

Resolves https://github.com/learningequality/kolibri/issues/9091

## Reviewer guidance

- From _"Device -> Facilities"_ or from _"Facility -> Data"_, sync a facility to another facility or to KDP to test successful sync information
- We currently have no simple mechanism for end-to-end testing of a sync failure - we're planning to do it at least once together with @bjester though

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/core/auth/api.py
+++ b/kolibri/core/auth/api.py
@@ -34,6 +34,8 @@ from django_filters.rest_framework import DjangoFilterBackend
 from django_filters.rest_framework import FilterSet
 from django_filters.rest_framework import ModelChoiceFilter
 from morango.api.permissions import BasicMultiArgumentAuthentication
+from morango.constants import transfer_stages
+from morango.constants import transfer_statuses
 from morango.models import TransferSession
 from rest_framework import decorators
 from rest_framework import filters
@@ -486,13 +488,28 @@ class FacilityViewSet(ValuesViewset):
     queryset = Facility.objects.all()
     serializer_class = FacilitySerializer
 
-    facility_values = ["id", "name", "num_classrooms", "num_users", "last_synced"]
+    facility_values = [
+        "id",
+        "name",
+        "num_classrooms",
+        "num_users",
+        "last_successful_sync",
+        "last_failed_sync",
+    ]
 
     values = tuple(facility_values + dataset_keys)
 
     field_map = {"dataset": _map_dataset}
 
     def annotate_queryset(self, queryset):
+        transfer_session_dataset_filter = Func(
+            Cast(OuterRef("dataset"), TextField()),
+            Value("-"),
+            Value(""),
+            function="replace",
+            output_field=TextField(),
+        )
+
         return (
             queryset.annotate(
                 num_users=SQCount(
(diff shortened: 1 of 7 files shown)


kolibri/core/auth/test/test_api.py
```diff
@@ -430,6 +436,20 @@ def setUpTestData(cls):
...
+        )
+        cls.transfer_session = TransferSession.objects.create(
+            id=uuid.uuid4().hex,
+            sync_session_id=cls.sync_session.id,
+            push=True,
+            active=False,
+            last_activity_timestamp=timezone.now(),
+            transfer_stage=transfer_stages.CLEANUP,
+            transfer_stage_status=transfer_statuses.COMPLETED,
+        )
```
bjester, line 464: If you recall from our discussion yesterday, the `transfer_session.filter` field is what will match the `dataset`'s ID. The `filter` isn't always that simple, but for a facility sync, it usually is so you can put `filter=facility.dataset_id`

MisRob: Yes, I recall but then struggled a bit to figure out what to use as a value anyways. Thank you for all your help.

kolibri/core/auth/api.py
```diff
@@ -505,15 +522,27 @@ def annotate_queryset(self, queryset):
...
+                last_successful_sync=Subquery(
+                    # the sync command does a pull, then a push, so if the push succeeded,
+                    # the pull likely did too, which means this should represent when the
+                    # facility was last fully and successfully synced
+                    TransferSession.objects.filter(
+                        push=True,
+                        active=False,
+                        transfer_stage=transfer_stages.CLEANUP,
+                        transfer_stage_status=transfer_statuses.COMPLETED,
+                        filter=transfer_session_dataset_filter,
```
jredrejo, line 534: if you're annotating with the facility queryset, why do you need this filter?

kolibri/core/auth/api.py
```diff
@@ -505,15 +522,27 @@ def annotate_queryset(self, queryset):
...
                     TransferSession.objects.filter(
-                        filter=Func(
-                            Cast(OuterRef("dataset"), TextField()),
-                            Value("-"),
-                            Value(""),
-                            function="replace",
-                            output_field=TextField(),
-                        )
+                        transfer_stage_status=transfer_statuses.ERRORED,
+                        filter=transfer_session_dataset_filter,
```
jredrejo, line 545: same as above, I don't understand why this filter is needed when using in the annotation

jredrejo (changes requested):
Look good to me, great work. I only have left a couple of comments with the same doubt about a filter, if we can clarify it, it'd be ready from my pov

kolibri/core/auth/api.py
MisRob: I don't understand this in great detail and worked on this together with @bjester so he can confirm or formulate it better if needed. In my understanding, the `filter` field keeps the information about what facility a transfer session is related to. We use `dataset_id` for that purpose. An example can be seen in tests here https://github.com/learningequality/kolibri/pull/9541/files#diff-bd3440b7bfeda2f7b570e2cbacdfd560a03abff346bd3a788314f4f51e8f792bR450. I think that if `filter` wasn't specified in this way, then the query would return all transfer sessions rather than transfer sessions that have a relationship to this facility and that would be a problem because I need to know the last successful/failed sync status for this particular facility.

MisRob:
Thank you @jredrejo, I've just replied

kolibri/core/auth/api.py
bjester: What @MisRob said!

bjester (approved):
:100:

MisRob:
We've just tested syncing failure and success with @bjester and looks to be good

kolibri/core/auth/api.py
jredrejo: Thanks for the explanation, makes sense to  me now.

jredrejo (approved):

PR #9548, opened 2022-07-04
new script created (review needed)

abdelrahman725:
## Summary

nothing interesting yet, just a basic review for the setup and my approach

## References

…

## Reviewer guidance

…

----

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated


kolibri/core/content/generate_fixtures.py
```diff
@@ -0,0 +1,299 @@
...
+
+def generate_random_id():
+    import uuid
+    return uuid.uuid4().hex 
+
+
+
+
+
+content_kind_to_file_preset = {
```
rtibbles: You can avoid hard coding this by importing:
```
from le_utils.constants import format_presets
```

Each preset should have a `kind` attribute that will allow you to build this mapping programmatically from that.

kolibri/core/content/generate_fixtures.py
```diff
@@ -0,0 +1,299 @@
...
+    "document": ["document","document_thumbnail"],
+    "html5": ["html5_zip","html5_thumbnail"],
+    "slideshow":["slideshow_manifest","slideshow_image"],
+    "h5p" : ["h5p","audio_thumbnail"],
+}
+
+
+
+# code used to generate the following file_preset_to_localfile_extension mapper :
+#     mapper = {}
```
rtibbles: Likewise, the preset data has this information on it as well:
https://github.com/learningequality/le-utils/blob/main/le_utils/resources/presetlookup.json

kolibri/core/content/generate_fixtures.py
```diff
@@ -0,0 +1,299 @@
...
+ 
+
+
+# any generated object will be added here to get deleted later after dumping
+generated_objects = set()
+
+
+
+# keeps track of used (previously generated) languages (the ones that will be dumped) 
+Used_Languages = {}
```
rtibbles: A general code style comment - in the Kolibri code base we tend to use snake_case for most Python variables and function names, so this would be `used_languages`.

For class names we use PascalCase.

kolibri/core/content/generate_fixtures.py
```diff
@@ -0,0 +1,299 @@
...
+tag_names = ["Math.CC.K.CC.C.7","science_related","have_fun","children","experiment","bedo_tag","course","culture","introduction","whatever","another_tag"]
+fixtures_file_path = "fixtures/bedo_fixtures.json"
+license_name = "testing license"
+license_description = "ABC organization authorizes kolibri to use this these resources"
+license_owner="ABC org"
+developer = "bedo khaled"
+
+
+
+def GenerateChannel(root_node):
```
rtibbles: Make sure to install pre-commit:
```
pip install pre-commit
pre-commit install
```

See here for more details: https://kolibri-dev.readthedocs.io/en/develop/getting_started.html#pre-commit-hooks

kolibri/core/content/generate_fixtures.py
```diff
@@ -0,0 +1,299 @@
...
+    "document":"pdf",
+    "epub":"epub",
+    "exercise":"perseus",
+    "h5p":"h5p",
+    "slideshow_manifest":"json",
+}
+ 
+
+
+# any generated object will be added here to get deleted later after dumping
```
rtibbles: If you are properly encapsulating the fixture generation in an in-memory database this is probably superfluous - if it's useful for now can leave it.

kolibri/core/content/generate_fixtures.py
```diff
@@ -0,0 +1,299 @@
...
+            "NAME"   : ":memory:"
+        }
+        try :
+            del connections[db]
+        except AttributeError : pass 
+        call_command("migrate", interactive=False, database=db)
+
+def StartGenerating():
+
+    GetAllLanguages()
```
rtibbles: It seems this is actually _setting_ all the languages on a module global for later lookup? Likewise for the local files below.

kolibri/core/content/generate_fixtures.py
```diff
@@ -0,0 +1,299 @@
...
+# same as above but for local files
+Used_Local_Files = {}
+
+
+
+# extracted languages from the QA channel
+AllLanguages = {}
+#run before switching to memory
+def GetAllLanguages():
+    all_langs= Language.objects.all()
```
rtibbles: Relying on languages that exist in the current database will make this command brittle - you should instead generate the language data from the `le_utils` module: https://github.com/learningequality/le-utils/blob/main/le_utils/constants/languages.py

kolibri/core/content/generate_fixtures.py
```diff
@@ -0,0 +1,299 @@
...
+
+# extracted localfiles from the QA channel
+AllLocalFiles={}
+#run before switching to memory
+def GetAllLocalFiles():
+    # available extensions for now : 
+    exts = ["mp4","png","jpg","zip","mp3","vtt","json","epub","h5p","perseus","pdf"]
+    
+    for ext in exts:
+        current_extension_files = LocalFile.objects.filter(extension=ext)
```
rtibbles: This is also relying on the LocalFiles existing in the database before we move to an in memory DB - this will make this generation brittle. We should programmatically generate the data for the local files instead.

If you are intending to rely specifically on the Kolibri QA channel for either this or the language data, that needs to be specified, and the import needs to happen as part of the overall generation script.

rtibbles (commented):
This seems to be on the right track - please look at how the ChannelBuilder code is organized to get a sense of how you can create more realistic topic tree hierarchies and also organize the code. Having a class that brings this together will definitely be beneficial.

I also think you should turn this into a Django management command - which would naturally give you a `Command` class as a place to unify this state (rather than just living in the module level).

kolibri/core/content/generate_fixtures.py
abdelrahman725: done

kolibri/core/content/generate_fixtures.py
abdelrahman725: done

kolibri/core/content/generate_fixtures.py
abdelrahman725: done

kolibri/core/content/generate_fixtures.py
abdelrahman725: this functions gets all the languages of the QA channel but i can't use them directly, why? because these languages exist in the local db (not the memory db) and i'm only dumping models generated in the memory, so if used them directly in the nodes for example they will not get dumped in the fixtures.So i'm just retrieving them an as you can see in the `Used_Languages` dictionary i keep track of actual generated languages during the script, anyway we have decided that we will discard the usage of QA channel at all, i was just explaining

kolibri/core/content/generate_fixtures.py
abdelrahman725: done !

kolibri/core/content/generate_fixtures.py
abdelrahman725: we have decided to discard the usage of the QA channel at all, i'm not intending to rely on it, because simply i don't know if that's better or not

abdelrahman725: this also done!

Issue #9554, opened 2022-07-07
Reporting an error in bulkimportusers triggers an error


rtibbles:
## Observed behavior
* When attempting to upload an incorrect filetype to bulk import users, if the file is not the correct format, then it will attempt to added an error message to report back on the task.
* When it does this, something in the reporting has a non-JSON serializable object associated with it `proxy` - which causes an error when attempting to save the serialized task to the database.

## Errors and logs
```
Traceback (most recent call last):
File “/usr/lib/python3/dist-packages/kolibri/core/tasks/job.py”, line 174, in to_json
string_result = json.dumps(working_dictionary)
File “/usr/lib/python3.8/json/init.py”, line 231, in dumps
return _default_encoder.encode(obj)
File “/usr/lib/python3.8/json/encoder.py”, line 199, in encode
chunks = self.iterencode(o, _one_shot=True)
File “/usr/lib/python3.8/json/encoder.py”, line 257, in iterencode
return _iterencode(o, 0)
File “/usr/lib/python3.8/json/encoder.py”, line 179, in default
raise TypeError(f’Object of type {o.class.name} ’
TypeError: Object of type proxy is not JSON serializable

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
File “/usr/lib/python3/dist-packages/kolibri/core/tasks/worker.py”, line 61, in handle_finished_future
result = future.result()
File “/usr/lib/python3.8/concurrent/futures/_base.py”, line 437, in result
return self.__get_result()
File “/usr/lib/python3.8/concurrent/futures/_base.py”, line 389, in __get_result
raise self._exception
File “/usr/lib/python3.8/concurrent/futures/thread.py”, line 57, in run
result = self.fn(*self.args, **self.kwargs)
File “/usr/lib/python3/dist-packages/kolibri/core/tasks/job.py”, line 117, in execute_job
result = func(*args, **kwargs)
File “/usr/lib/python3/dist-packages/kolibri/dist/django/core/management/init.py”, line 131, in call_command
return command.execute(*args, **defaults)
File “/usr/lib/python3/dist-packages/kolibri/dist/django/core/management/base.py”, line 330, in execute
output = self.handle(*args, **options)
File “/usr/lib/python3/dist-packages/kolibri/core/tasks/management/commands/base.py”, line 119, in handle
return self.handle_async(*args, **options)
[...]

PR #9557, opened 2022-07-11
Force translations in bulk export/import of user data

jredrejo:
## Summary
After some refactoring in the csv functions, the strings that were translated in the backend were not being translated or serialized in json format when they were still a lazy proxy

This PR:
a) forces all the translation scafolding to be active before calling labeling functions
b) ensures the errors are fully converted to string and not lazy functions before being serialized

## References
Closes: #9554

## Reviewer guidance
When the server is set to other language different than English check that:
a) Bulk export in the Facility-Data tab works fine, check the exported csv file has its headers translated (if using a language Kolibri supports)
b) Modify the exported csv to introduce error, and import it: It should show the error messages correctly translated in the web UI (while triggering an exception in the server console)
c) tests pass

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/core/auth/management/commands/bulkexportusers.py
+++ b/kolibri/core/auth/management/commands/bulkexportusers.py
@@ -108,13 +108,14 @@ output_mappings = {
 }
 
 
-map_output = None  # to be assigned after labels are translated
+def map_output(item):
+    return partial(
+        output_mapper, labels=translate_labels(), output_mappings=output_mappings
+    )(item)
 
 
 def translate_labels():
-    global labels
-    global map_output
-    labels = OrderedDict(
+    return OrderedDict(
         (
             ("id", _("Database ID ({})").format("UUID")),
             ("username", _("Username ({})").format("USERNAME")),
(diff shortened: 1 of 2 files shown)


kolibri/core/auth/management/commands/bulkexportusers.py
```diff
@@ -108,11 +108,12 @@ def kind_of_roles(field, obj):
 }
 
 
-map_output = partial(output_mapper, labels=labels, output_mappings=output_mappings)
+map_output = None  # to be assigned after labels are translated
 
 
 def translate_labels():
     global labels
+    global map_output
```
rtibbles: Happy to merge this as is - but this feels like the map_output should be better encapsulated at time of use rather than assigning to a module global.

rtibbles: Although it seems like this has broken tests - can we just change the reference to map_output to not be a global one?

jredrejo: yep, thanks. I also disliked the smell of the global declarations but I thought It could be fixed with a quick hack. As usual, better doing things correctly and now tests pass perfectly.

rtibbles (approved):
Code looks good, if the actions so yes, then so shall we merge!

PR #9561, opened 2022-07-14
Move CACHES import into function scope to prevent side effects.

rtibbles:
## Summary
* The default cache module involves module level access to the Kolibri OPTIONS object
* This results in evaluation and initialization of the plugin registry
* This causes an issue for anything importing from `main` expecting to be able to enable plugins, as the enabled plugins will not be properly registered

## References
Fixes a reported issue for the Mac app.
I am hopeful that this behaviour would raise a runtime error in develop due to some stricter checking, but it might still need some follow up to prevent this happening in future.

## Reviewer guidance
Test the [mac app](https://github.com/learningequality/kolibri-app/actions/runs/2673793226) that is built from the WHL file from this PR and ensure it starts.

----

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/utils/main.py
+++ b/kolibri/utils/main.py
@@ -25,7 +25,6 @@ from kolibri.core.tasks.main import import_tasks_module_from_django_apps
 from kolibri.core.upgrade import matches_version
 from kolibri.core.upgrade import run_upgrades
 from kolibri.core.utils.cache import process_cache
-from kolibri.deployment.default.cache import CACHES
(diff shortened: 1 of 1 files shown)


jredrejo (approved):
No doubt, thanks. I think I was who introduced the problem :(

radinamatic (approved):
@pcenov also confirmed that it installs and runs on the Mac device, we are good to go! 💯 :shipit:

rtibbles:
> I think I was who introduced the problem :(

Yes - but the fact that nothing shouts at you if an import like this is added is the real problem here!

Issue #9329, opened 2022-04-19
Create "Create Account" / "Edit Account" page


marcellamaki:
# Overview 

Adds a page for a learner account creation or editing an existing account. `CreateLearnerAccountForm.vue` in the Setup Wizard could be used for reference

## Figma 
| No Existing Account |  Edit Existing Account |
|---|---|
| ![Screen Shot 2022-04-20 at 9.43.12 AM.png](https://images.zenhubusercontent.com/6047f6e691103a78d3ad0ace/2940373d-2db3-433f-b4c7-12950945f875) | ![Screen Shot 2022-04-20 at 9.42.40 AM.png](https://images.zenhubusercontent.com/6047f6e691103a78d3ad0ace/67c84f5b-9411-4edb-89bd-bca209c46f76)| 

### Accessibility notes
Tab order: 
1) Text fields (top to bottom)
2) "Continue" (or "Save") primary button 
3) "Back" secondary button 
4) "X" close out flow 

### Out of scope 

Adding the route itself is not part of this issue

### Acceptance criteria

- [ ] Page is created with correct strings 
- [ ] UI layout to spec with correct tab order and spacing
- [ ] If the account exists, the fields are populated with existing details
- [ ] Fields are editable and displayed according to facility settings

MisRob:
Closing as "Create Account" has been already implemented and "Edit account" doesn't need to be implemented because we removed this step from the flow.

PR #9562, opened 2022-07-15
Change Facility - Create Account

MisRob:
## Summary

- Adds _Create Account_ page
- Updates _Create new account_ button to be disabled on _Confirm account username_ page when learners are not allowed to sign up on the target facility
- `FullNameTextbox` now trims the full name before validating it so that names made of empty characters won't be accepted (impacts all forms where we use the component - setup wizard, sign up, user profile settings, and creating/editing users of a facility)
- Sets `max-width` to `700px` on page containers for all change facility steps that have been implemented (the same value that was used in the onboarding setup wizard and is pretty close to `656px` width used on designs) 
- Updates secondary buttons for all change facility steps that have been implemented to have flat appearance

### Screenshots

| | _Confirm account username_ page | _Create account_ page |
| --- | --- | --- |
| The target facility doesn't allow learners to create accounts | cant-sign-up | --- |
| The target facility allows learners to create accounts and it doesn't require a password | can-sign-up | no-pw |
| The target facility allows learners to create accounts and it requires a password | can-sign-up | wants-pw |

#### Max container width and secondary buttons appearance (applies to all change facility steps):

| Before | After |
| ---------| ------- |
| Screenshot from 2022-07-15 15-13-57 | Screenshot from 2022-07-15 15-12-25 |

## References

- Solves the first part of #9329
- [Figma designs](https://www.figma.com/file/MpCG7r5PULOCJsEvnTSti8/Android-Designs-2022?node-id=596%3A10379)

## Notes

- With @jredrejo, @marcellamaki, and @rtibbles we agreed that the API call for creating a new account will happen later during the _Changing learning facility_ step => in this PR, clicking the _Continue_ button on _Create new account_ page only saves the target account data to the state machine's context from where it can be sent to backend later
- It's possible that later I will consider extracting form to its own component since _Edit existing account_ form (second part of #9329) is the same but I decided to open PRs separately
- _Back_ navigation on _Create account_ page will need some more work as soon as we start using this state in other branches of the flow as the button will lead somewhere else. But for that, we will need to implement more states at first.

### Setup

You need to have two facilities to be able to test the change facility workflow:

**Source facility**
1. You can run the development server as usual
2. Your Kolibri instance shouldn't be the learner only device and also you need to make sure there's only one user account, otherwise you won't be able to access the _Change learner facility_ feature in _Profile_

**Target facility**

If you already have a target facility with some user accounts, make sure that in the source facility, you log in as a user with a username that doesn't match any username on the target facility (otherwise you'll be taken to a different part of the flow). Also, it needs to be v0.16.

If you don't have a target facility, you can follow these steps:
1. `cd` to your `kolibri` development directory and make sure to activate your virtual environment like when developing
2. Run ` pip install -e .` to ensure that the target facility has `develop` (`0.16.x`) updates that are required (`kolibri --version` should show something like `0.16.0.dev0+git.20220715072421`)
3. Run Kolibri with the `kolibri` command with `KOLIBRI_HOME` environment variable different from your usual setup so that you don't override your development Kolibri  `KOLIBRI_HOME=~/.tmp kolibri start --foreground`
4. You need to go through the onboarding setup, otherwise the target facility may not be discovered. It seems that we have some onboarding work in progress so it's possible that you won't be able to go through the onboarding forms and complete the setup. I solved this by switching to `0.15.x` version where I did the onboarding and then switched to this branch using the same `KOLIBRI_HOME` and then went through steps 1-3.

#### (1)
In the facility settings of the target facility
- do not _Allow users to create accounts_

On the source facility
1. Navigate to _Profile -> Change facility -> Change_
2. Select the target facility and click _Continue_
3. On _Change Facility_ page click _Continue_
4. On _Confirm account username_ page, check that _Create new account_ button is disabled

#### (2)
In the facility settings of the target facility
- do _Allow users to create accounts_
- do not _Require password for learners_

On the source facility
1. Navigate to _Profile -> Change facility -> Change_
2. Select the target facility and click _Continue_
3. On _Change Facility_ page click _Continue_
4. On _Confirm account username_ page, click _Create new account_ button
6. Check that you can see _Create new account_ page
7. Test _Create new account_ page:
    - full name and username textboxes are visible
    - password textbox is not visible
    - test _Back_ navigation
    - check form validation - you should be able to _Continue_ only when the form is valid
    - test keyboard navigation

#### (3)
In the facility settings of the target facility
- do _Allow users to create accounts_
- do _Require password for learners_

On the source facility
- same as (2) but two password textboxes should show up

Note that there's a focus ring missing around password textboxes when testing keyboard navigation (https://github.com/learningequality/kolibri/issues/9563)

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/core/assets/src/views/userAccounts/FullNameTextbox.vue
+++ b/kolibri/core/assets/src/views/userAccounts/FullNameTextbox.vue
@@ -41,8 +41,11 @@
       };
     },
     computed: {
+      isFullNameValid() {
+        return this.value && this.value.trim() !== '';
+      },
       invalidText() {
-        if (this.value === '') {
+        if (!this.isFullNameValid) {
           return this.coreString('requiredFieldError');
         }
         return '';
@@ -53,12 +56,9 @@
         }
         return '';
       },
-      valid() {
-        return this.invalidText === '';
-      },
     },
     watch: {
-      valid: {
+      isFullNameValid: {
         handler(value) {
           this.$emit('update:isValid', value);
         },
--- a/kolibri/core/assets/src/views/userAccounts/FullNameTextbox.vue
+++ b/kolibri/core/assets/src/views/userAccounts/FullNameTextbox.vue
@@ -25,6 +25,7 @@
     name: 'FullNameTextbox',
     mixins: [commonCoreStrings],
     // NOTE: 'value' and 'isValid' must be .sync'd with parent
+    // You can also pass 'disabled', 'autofocus', and 'autocomplete'
     props: {
       value: {
         type: String,
--- a/kolibri/plugins/user_profile/assets/src/machines/changeFacilityMachine.js
+++ b/kolibri/plugins/user_profile/assets/src/machines/changeFacilityMachine.js
@@ -224,7 +224,15 @@ export const changeFacilityMachine = createMachine({
           target: 'isAdmin',
           actions: setTargetAccount,
         },
-        BACK: 'confirmAccount',
+        BACK: [
+          {
+            cond: context => context.accountExists,
+            target: 'changeFacility',
+          },
+          {
+            target: 'confirmAccount',
+          },
+        ],
       },
     },
     usernameExists: {
--- a/kolibri/plugins/user_profile/assets/src/machines/changeFacilityMachine.js
+++ b/kolibri/plugins/user_profile/assets/src/machines/changeFacilityMachine.js
@@ -83,7 +83,10 @@ export const changeFacilityMachine = createMachine({
     username: '',
     targetAccount: { username: '', id: '' },
     sourceFacility: '',
-    targetFacility: {}, //should be an object with the facility name, id & url
+    // should be an object with the target facility
+    // `name`, `id`, `url`, `learner_can_sign_up` &
+    // `learner_can_login_with_no_password` fields
+    targetFacility: {},
     newSuperAdmin: '',
     taskPolling: false,
     accountExists: false,
--- a/kolibri/plugins/user_profile/assets/src/machines/changeFacilityMachine.js
+++ b/kolibri/plugins/user_profile/assets/src/machines/changeFacilityMachine.js
@@ -221,8 +221,7 @@ export const changeFacilityMachine = createMachine({
           target: 'isAdmin',
           actions: setTargetAccount,
         },
-        BACK: 'changeFacility',
-        BACKMERGING: 'confirmMerge',
+        BACK: 'confirmAccount',
       },
     },
     usernameExists: {
--- a/kolibri/plugins/user_profile/assets/src/machines/changeFacilityMachine.js
+++ b/kolibri/plugins/user_profile/assets/src/machines/changeFacilityMachine.js
@@ -165,7 +165,10 @@ export const changeFacilityMachine = createMachine({
     confirmAccount: {
       meta: { route: 'CONFIRM_ACCOUNT', path: '/change_facility' },
       on: {
-        NEW: 'createAccount',
+        NEW: {
+          cond: context => context.targetFacility && context.targetFacility.learner_can_sign_up,
+          target: 'createAccount',
+        },
         CONTINUE: 'isAdmin',
         BACK: 'changeFacility',
       },
--- a/kolibri/plugins/user_profile/assets/src/machines/changeFacilityMachine.js
+++ b/kolibri/plugins/user_profile/assets/src/machines/changeFacilityMachine.js
@@ -208,8 +208,8 @@ export const changeFacilityMachine = createMachine({
       type: 'final',
     },
     createAccount: {
+      meta: { route: 'CREATE_ACCOUNT', path: '/change_facility' },
       on: {
-        meta: { route: 'CREATE_ACCOUNT', path: '/change_facility' },
         CONTINUE: {
           // event.value must be an object {id:, usename:}
           target: 'isAdmin',
--- a/kolibri/plugins/user_profile/assets/src/machines/changeFacilityMachine.js
+++ b/kolibri/plugins/user_profile/assets/src/machines/changeFacilityMachine.js
@@ -81,7 +81,11 @@ export const changeFacilityMachine = createMachine({
   context: {
     role: 'learner',
     username: '',
-    targetAccount: { username: '', id: '' },
+    targetAccount: {
+      fullName: '',
+      username: '',
+      password: '',
+    },
     sourceFacility: '',
     // should be an object with the target facility
     // `name`, `id`, `url`, `learner_can_sign_up` &
@@ -214,7 +218,6 @@ export const changeFacilityMachine = createMachine({
       meta: { route: 'CREATE_ACCOUNT', path: '/change_facility' },
       on: {
         CONTINUE: {
-          // event.value must be an object {id:, usename:}
           target: 'isAdmin',
           actions: setTargetAccount,
         },
--- a/kolibri/plugins/user_profile/assets/src/views/ChangeFacility/ConfirmAccount/index.vue
+++ b/kolibri/plugins/user_profile/assets/src/views/ChangeFacility/ConfirmAccount/index.vue
@@ -10,6 +10,8 @@
           <KButton
             :primary="false"
             :text="$tr('createAccount')"
+            :disabled="isCreateAccountButtonDisabled"
+            data-test="createNewAccountButton"
             appearance="flat-button"
             @click="to_create"
           />
@@ -55,6 +57,9 @@
       },
       usernameExists() {
         return get(this.state, 'value.accountExists');
+      },
+      isCreateAccountButtonDisabled() {
+        return !get(this.targetFacility, 'learner_can_sign_up');
       },
       firstLine() {
         return this.$tr('confirmAccountLine1', {
--- a/kolibri/plugins/user_profile/assets/src/views/ChangeFacility/ConfirmAccount/index.vue
+++ b/kolibri/plugins/user_profile/assets/src/views/ChangeFacility/ConfirmAccount/index.vue
@@ -30,6 +30,7 @@
 
 <script>
 
+  import get from 'lodash/get';
   import commonCoreStrings from 'kolibri.coreVue.mixins.commonCoreStrings';
   import BottomAppBar from 'kolibri.coreVue.components.BottomAppBar';
 
@@ -47,23 +48,23 @@
     inject: ['changeFacilityService', 'state'],
     computed: {
       targetFacility() {
-        return this.state.value.targetFacility;
+        return get(this.state, 'value.targetFacility');
       },
       username() {
-        return this.state.value.username;
+        return get(this.state, 'value.username');
       },
       usernameExists() {
-        return this.state.value.accountExists;
+        return get(this.state, 'value.accountExists');
       },
       firstLine() {
         return this.$tr('confirmAccountLine1', {
-          target_facility: this.targetFacility.name,
+          target_facility: get(this.targetFacility, 'name', ''),
           username: this.username,
         });
       },
       secondLine() {
         return this.$tr('confirmAccountLine2', {
-          target_facility: this.targetFacility.name,
+          target_facility: get(this.targetFacility, 'name', ''),
         });
       },
     },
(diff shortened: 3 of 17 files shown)


kolibri/core/assets/src/views/userAccounts/FullNameTextbox.vue
```diff
@@ -40,8 +41,11 @@
       };
     },
     computed: {
+      isFullNameValid() {
+        return this.value && this.value.trim() !== '';
```
jredrejo, line 45: nice, this is something we should do in other places in the app for empty text boxes. Maybe it could be moved to be a function in a composable to make it more reusable (or maybe not as it's quite simple, but I'd like to have some kind of reminder for other places)

kolibri/plugins/user_profile/assets/src/machines/changeFacilityMachine.js
```diff
@@ -81,9 +81,16 @@ export const changeFacilityMachine = createMachine({
   context: {
     role: 'learner',
     username: '',
-    targetAccount: { username: '', id: '' },
+    targetAccount: {
+      fullName: '',
```
jredrejo, line 85: the id of the target account will be needed to execute the final merge

kolibri/plugins/user_profile/assets/src/machines/changeFacilityMachine.js
```diff
@@ -158,7 +165,10 @@ export const changeFacilityMachine = createMachine({
     confirmAccount: {
       meta: { route: 'CONFIRM_ACCOUNT', path: '/change_facility' },
       on: {
-        NEW: 'createAccount',
+        NEW: {
+          cond: context => context.targetFacility && context.targetFacility.learner_can_sign_up,
```
jredrejo, line 169: perfect! you got the xstate flow!

kolibri/plugins/user_profile/assets/src/machines/changeFacilityMachine.js
```diff
@@ -208,15 +218,13 @@ export const changeFacilityMachine = createMachine({
       type: 'final',
     },
     createAccount: {
+      meta: { route: 'CREATE_ACCOUNT', path: '/change_facility' },
```
jredrejo, line 221: good catch

kolibri/plugins/user_profile/assets/src/machines/changeFacilityMachine.js
```diff
@@ -208,15 +218,13 @@ export const changeFacilityMachine = createMachine({
...
       on: {
-        meta: { route: 'CREATE_ACCOUNT', path: '/change_facility' },
         CONTINUE: {
-          // event.value must be an object {id:, usename:}
           target: 'isAdmin',
           actions: setTargetAccount,
         },
-        BACK: 'changeFacility',
-        BACKMERGING: 'confirmMerge',
+        BACK: 'confirmAccount',
```
jredrejo: I don't see this clearly looking at the figma design, could we discuss it?

kolibri/plugins/user_profile/assets/src/views/ChangeFacility/ConfirmAccount/index.vue
```diff
@@ -46,23 +50,26 @@
     inject: ['changeFacilityService', 'state'],
     computed: {
       targetFacility() {
-        return this.state.value.targetFacility;
+        return get(this.state, 'value.targetFacility');
```
jredrejo, line 53: I don't understand the need of this change, more taking into account you need to do a new import just for this.

jredrejo (changes requested):
I've left some comments in the code, but hadn't had time to test it yet. I just didn't want to lose these comments before I test it tomorrow

kolibri/core/assets/src/views/userAccounts/FullNameTextbox.vue
MisRob: Maybe having a couple of tests with values that are valid and invalid might help with such issues as well. I find `jest-each` helpful because you can do something like

```
each([' ', '         ', '']).test(`full name textbox doesn't accept "%s" as valid`, value => { 
   test `value` here
})

each(['name', 'full name']).test(`full name textbox accepts "%s" as valid`, value => { 
   test `value` here
})
```

So that you can write tests concisely. This usually makes me to feed tests with values that I may not think of before.

kolibri/plugins/user_profile/assets/src/machines/changeFacilityMachine.js
MisRob: Where should I get it and what do you mean exactly by the "final merge"? `CreateAccount` page only collects information from users who wish to create a completely new account - their username, full name, and password. That account will be created by the backend later on. I assume we don't generate ids for new users on front-end?

 It's different from merging accounts - in that case I assume we will have `id` available.

I guess I could add `id` back with a comment that it can be empty in some cases but maybe it'd make more sense to have two structures in the state machine's context to differentiate these scenarios, `newAccount` and `mergeTargetAccount`?

kolibri/plugins/user_profile/assets/src/machines/changeFacilityMachine.js
MisRob: It's this step: _Confirm account username_ -> _Create new account_ so backwards it's _Create new account_ -> _Confirm account username_

Screenshot from 2022-07-19 11-05-33

Later on, we will need to tweak this logic to be conditional because in the _Username exists_ flow, back action will indeed need to go back to  _Change facility_, but we haven't started implementing this branch of the flow yet.

Screenshot from 2022-07-19 11-13-25

So for now, I just continued from the last state you prepared, _Confirm account username_, which is the one that belongs to _the Default path_, and made sure that this new state goes back to it. Maybe you can try to click through it on the devserver? That's a simple way to see how it works. Let me know if you need to chat about this on a call.

kolibri/plugins/user_profile/assets/src/views/ChangeFacility/ConfirmAccount/index.vue
MisRob: The smoke[ test I added here](https://github.com/learningequality/kolibri/pull/9562/files#diff-fd0b011ddc9a568a5c730be0271e1f026140c6a528245f4e65ee5de91bb7ca4eR28-R31) would fail if I didn't do this when it'd attempt to access `targetFacility` on `undefined`. The purpose of that test is to ensure that the component can run without any runtime errors even before it receives data from other application layers. Making assumptions that some data is available immediately can be tricky in async reactive environments and creates hidden dependencies.

If I wouldn't use lodash `get`, I would typically add a guard `if (this.state && this.state.value) { return this.state.value.targetFacility }` to solve this problem. I learned to use lodash from someone's PR recently for accessing nested properties, it's simpler. No problem to remove it, it's a personal preference. What's the problem with `import`?

kolibri/plugins/user_profile/assets/src/machines/changeFacilityMachine.js
MisRob: Good mentor :)

kolibri/plugins/user_profile/assets/src/machines/changeFacilityMachine.js
jredrejo: It's one of the closest issues in this epic, the PR that implemented it is this one: https://github.com/learningequality/kolibri/pull/9454 , this https://github.com/learningequality/kolibri/blob/develop/kolibri/plugins/user_profile/tasks.py#L62-L78 documents the task that will do the final merge.

kolibri/plugins/user_profile/assets/src/machines/changeFacilityMachine.js
jredrejo: I still don't get it: it was implemented in the machine: depending on the screen you where, the event was different:
        BACK: 'changeFacility',
        BACKMERGING: 'confirmMerge',

so you h ave the two possibilities there, the screen should know which one to call in the back button, so no need to tweak it.

kolibri/plugins/user_profile/assets/src/views/ChangeFacility/ConfirmAccount/index.vue
jredrejo: ok, thanks for the explanation. There's no problem with the import, I just like to avoid adding external functions if it can be done in plain JS but after reading your comment it seems it's needed.

jredrejo:
@MisRob Trying to test it, I have not been able because there's a missing screen in the middle , it seems we're stalled now , how did you test it,  just tweaking the machine to skip one state or I'm missing something?

MisRob:
Could you please post what scenario from those three you were testing and what step exactly didn't work for you in the scenario you tested?

MisRob:
I didn't need to tweak anything

MisRob:
Maybe I had some condition locally present about my facilities that you don't have so more information about what you were exactly testing would be helpful.

kolibri/plugins/user_profile/assets/src/machines/changeFacilityMachine.js
MisRob: I see, from the code it seems you're using "merge" term for both scenarios - creating a new account as well as merging with an existing one. I think I get that now, but still don't understand where you suggest I should get an id when someone's creating a new account. Could you please explain what `id`'s value should be?

jredrejo: nah, forget it. I lost the reference in this task, looking from/to local/remote all the time made me lose the thread. The remote user id will "appear" in the local machine after they sync in the task I reference above. After the remote user is synced to the local machine the task gets the id using the username at https://github.com/learningequality/kolibri/blob/develop/kolibri/plugins/user_profile/tasks.py#L84. So what we really need from the remote facility is what exposed in the validator: its `facility_id`, `username` and `password` (the latest only if the user exists and needs to authenticate).Neither  `fullName` nor `id` are not required.

MisRob:
Just a note if someone's following this, we'll clarify the issues raised above on a call with @jredrejo, and then I will update or post a comment,

MisRob:
With @jredrejo, we found out that there was one more condition that I didn't include in the PR instructions for testing that caused that we were testing different pathways. PR description has been updated.

kolibri/plugins/user_profile/assets/src/machines/changeFacilityMachine.js
MisRob: @jredrejo As we discussed, I replaced `confirmMerge` by `confirmAccount`. But, I thought it'd be better rather to have only one `BACK` event and let the state machine to conditionally decide how to react to it depending on the context rather than having two back events and having `CreateAccount` page decide which one to call. Because in both of those cases, we need to check state machine's `context.accountExists` to find out where to navigate back. I think having this behavior in the state machine itself makes more sense. Please have a look at the latest update and let me know if that's looking good to you now.

MisRob:
There's one more thing around the full name that needs to be clarified with the design team. I'll follow up on Slack.

kolibri/plugins/user_profile/assets/src/views/ChangeFacility/SelectFacility.vue
```diff
@@ -219,11 +220,13 @@
                 name: facility.name,
                 base_url: address.base_url,
                 address_id: address.id,
-                learner_can_sign_up: facility.learner_can_sign_up || true,
+                learner_can_sign_up: facility.learner_can_sign_up,
+                learner_can_login_with_no_password: facility.learner_can_login_with_no_password,
                 kolibri_version: address.kolibri_version,
               };
-              if (!this.availableFacilities.find(f => f.id === facility.id))
+              if (!this.availableFacilities.find(f => f.id === facility.id)) {
```
MisRob, line 227: Just added a little fix. I think that a true Python person wrote this `if` :))))

MisRob:
We're discussing an option not to show the full name textbox, but that work would preferably happen in another PR because we'd need to update gherkins and designs too at first.

All updates are pushed so it's ready for review again. Thanks @jredrejo for all cooperation.

jredrejo (approved):
Thank you Misha, this has been a hard work. I think now it's fine to be merged

PR #9565, opened 2022-07-18
When changing facility , ensure version is at least 0.16.0

jredrejo:
## Summary

- Created a composable to check kolibri versions in JavaScript
- After some comments from @MisRob in a previous PR, passed most of the vue component to use the composition api instead of the options api.

## References
Relates to #9324

## Reviewer guidance
Do changes make sense?
Do tests pass?

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/core/assets/src/composables/__tests__/useMinimumKolibriVersion.spec.js
+++ b/kolibri/core/assets/src/composables/__tests__/useMinimumKolibriVersion.spec.js
@@ -23,5 +23,16 @@ describe(`useMinimumKolibriVersion`, () => {
     it(`higher version is detected, without default`, () => {
       expect(isMinimumKolibriVersion.value('0.16.1', 0, 16, 0)).toBe(true);
     });
+
+    it(`check beta versions work fine with equal values`, () => {
+      expect(isMinimumKolibriVersion.value('0.16.1-b4', 0, 16, 1)).toBe(false);
+    });
+    it(`check beta versions work when betas are included`, () => {
+      expect(isMinimumKolibriVersion.value('0.16.1-b4', 0, 16)).toBe(true);
+    });
+
+    it(`check beta versions work fine with upper values`, () => {
+      expect(isMinimumKolibriVersion.value('0.16.1-b4', 0, 16, 0)).toBe(true);
+    });
   });
 });
--- a/kolibri/core/assets/src/composables/__tests__/useMinimumKolibriVersion.spec.js
+++ b/kolibri/core/assets/src/composables/__tests__/useMinimumKolibriVersion.spec.js
@@ -28,7 +28,7 @@ describe(`useMinimumKolibriVersion`, () => {
       expect(isMinimumKolibriVersion.value('0.16.1-b4', 0, 16, 1)).toBe(false);
     });
     it(`check beta versions work when betas are included`, () => {
-      expect(isMinimumKolibriVersion.value('0.16.1-b4', 0, 16)).toBe(true);
+      expect(isMinimumKolibriVersion.value('0.16.0-b4', 0, 16)).toBe(true);
     });
 
     it(`check beta versions work fine with upper values`, () => {
--- a/kolibri/core/assets/src/composables/__tests__/useMinimumKolibriVersion.spec.js
+++ b/kolibri/core/assets/src/composables/__tests__/useMinimumKolibriVersion.spec.js
@@ -1,0 +1,27 @@
+import useMinimumKolibriVersion from '../useMinimumKolibriVersion';
+
+const { isMinimumKolibriVersion } = useMinimumKolibriVersion();
+
+describe(`useMinimumKolibriVersion`, () => {
+  describe(`isMinimumKolibriVersion computed function`, () => {
+    it(`lower version is detected, with 0.15.0 as default`, () => {
+      expect(isMinimumKolibriVersion.value('0.14.99')).toBe(false);
+    });
+
+    it(`same version is detected, with 0.15.0 as default`, () => {
+      expect(isMinimumKolibriVersion.value('0.15.0')).toBe(true);
+    });
+
+    it(`lower version is detected, without default`, () => {
+      expect(isMinimumKolibriVersion.value('0.15.4', 0, 16, 0)).toBe(false);
+    });
+
+    it(`same version is detected, without default`, () => {
+      expect(isMinimumKolibriVersion.value('0.16.0', 0, 16, 0)).toBe(true);
+    });
+
+    it(`higher version is detected, without default`, () => {
+      expect(isMinimumKolibriVersion.value('0.16.1', 0, 16, 0)).toBe(true);
+    });
+  });
+});
--- a/kolibri/core/assets/src/composables/useMinimumKolibriVersion.js
+++ b/kolibri/core/assets/src/composables/useMinimumKolibriVersion.js
@@ -1,0 +1,27 @@
+import { computed } from 'kolibri.lib.vueCompositionApi';
+
+export default function useMinimumKolibriVersion() {
+  const isMinimumKolibriVersion = computed(() => {
+    return function(version, majorVersion = 0, minorVersion = 15, revisionVersion = 0) {
+      // defaults to kolibri 0.15.0
+      const v = version.split('.');
+      if (v.length < 2) return false;
+      const mayor = parseInt(v[0]);
+      const minor = parseInt(v[1]);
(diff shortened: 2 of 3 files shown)


kolibri/core/assets/src/composables/useMinimumKolibriVersion.js
```diff
@@ -0,0 +1,35 @@
...
+    return function(version, majorVersion = 0, minorVersion = 15, revisionVersion = null) {
+      /*
+      defaults to kolibri 0.15.x
+      If values are not provided for revisionVersion, then any values are allowed.
+      This means that if no value is provided for revisionVersion then
+      alpha and beta versions will be permitted.
+      Just providing 0 as majorVersion and 15 as minorVersion would allow any version
+      greater than or equal to 0.15 (including any alphas or betas)
+      */
+      const v = version.split('.');
```
MisRob, line 14: I know we're not there yet :-) but I'd expect `isMinimumKolibriVersion('1', 0, 15)` and `isMinimumKolibriVersion('1.0', 0, 15)` to return `true`, the same way as `isMinimumKolibriVersion('1.0.0', 0, 15)` does. Would that make sense? Strictly speaking, I'm not actually sure if v1 can be considered the same v1.0 and v1.0.0, it's rather my intuition.

In any case, from a couple of places in this function, it seems that all three parts are needed, so on the line below, maybe we could just throw an error if the version value doesn't satisfy the full format requirement?

```
if (v.length < 3) {
   throw new Error('The full version format (e.g. 0.15.0) is required')
}
```

To ensure that it won't fail silently when returning `false` only because of an unexpected parameter, rather than as a result of comparing versions.

I'd also recommend checking that `version` is defined before trying to split it.

kolibri/plugins/user_profile/assets/src/views/ChangeFacility/SelectFacility.vue
```diff
@@ -120,39 +117,74 @@
...
+
+      function handleAddedAddress() {
+        refreshSavedAddressList();
+        createSnackbar(this.$tr('addAddressSnackbarText'));
+      }
+
+      function resetSelectedAddress() {
+        const enabledFacilities = availableFacilities.value
+          .map(f => f.id)
+          .find(f => isMinimumKolibriVersion.value(f.kolibri_version, 0, 16));
```
MisRob: I think that here you're searching through an array of ids, due to the previous `map`, so `f.kolibri_version` will never be defined

kolibri/plugins/user_profile/assets/src/views/ChangeFacility/SelectFacility.vue
```diff
@@ -120,39 +117,74 @@
...
+        if (enabledFacilities.length !== 0) {
+          const selectedId = storageFacilityId.value || selectedFacilityId.value;
+          selectedFacilityId.value =
+            enabledFacilities.map(f => f.id).find(f => f.id === selectedId) ||
```
MisRob: Seems to be the same problem with `map` as above

MisRob (changes requested):
Thank you @jredrejo, this is a helpful update. Some changes will be needed. I posted some comments and also I can see these errors in the console:
Screenshot from 2022-07-19 15-19-38

kolibri/core/assets/src/composables/useMinimumKolibriVersion.js
jredrejo: I was talking about t his with Richard, if we want to have a complete flexibility on the version naming,  we should use something like https://www.npmjs.com/package/compare-versions 
However, in this case, as we're who are the users of the function and kolibri is going to be a '0.0.0' version always we don't need to add one more dependency. So, yes , you are right for that case... but that case is not going to happen.

kolibri/plugins/user_profile/assets/src/views/ChangeFacility/SelectFacility.vue
jredrejo: similar, but not the same, it's also fixed now. I messed it up with the last refactor I did :blush:

kolibri/core/assets/src/composables/__tests__/useMinimumKolibriVersion.spec.js
```diff
@@ -0,0 +1,38 @@
...
+
+    it(`higher version is detected, without default`, () => {
+      expect(isMinimumKolibriVersion.value('0.16.1', 0, 16, 0)).toBe(true);
+    });
+
+    it(`check beta versions work fine with equal values`, () => {
+      expect(isMinimumKolibriVersion.value('0.16.1-b4', 0, 16, 1)).toBe(false);
+    });
+    it(`check beta versions work when betas are included`, () => {
+      expect(isMinimumKolibriVersion.value('0.16.1-b4', 0, 16)).toBe(true);
```
rtibbles: Maybe just change this to `0.16.0-b4` to better document the intended behaviour here?

kolibri/plugins/user_profile/assets/src/views/ChangeFacility/SelectFacility.vue
```diff
@@ -120,39 +117,73 @@
         return savedAddressesInitiallyFetched.value && discoveredAddressesInitiallyFetched.value;
       });
       const storageFacilityId = useLocalStorage('kolibri-lastSelectedFacilityId', '');
+
+      // data:
+      const availableAddressIds = ref([]);
```
rtibbles, line 122: Are these intended to be persistent across multiple uses of this component?

kolibri/core/assets/src/composables/useMinimumKolibriVersion.js
MisRob: No need for complete flexibility. That's up to you. I imagine that at least communicating that such a decision has been made would help, with a comment or the parameter check. v1 was just one example, this will currently silently fail for non-v1 values too, like  `isMinimumKolibriVersion.value('0.16', 0, 15, 0)` returns `false` whereas `isMinimumKolibriVersion.value('0.16.0', 0, 15, 0)` returns `true`. It's fine to expect all three parts of a version number to be present, my comment is rather about making this clear for anyone who will use the function in the future.

jredrejo: the thing is that this is going to be used to check the version provided by kolibri, and kolibri will always provide three digits. I can add a comment in the function but, while this is been done only to be used inside kolibri, there should not be possible error. I.E. `isMinimumKolibriVersion.value('0.16', 0, 15, 0)` is not a possible case

kolibri/plugins/user_profile/assets/src/views/ChangeFacility/SelectFacility.vue
jredrejo: I am not sure of understanding your question: this is a data variable that's used several times inside the component, not only in the `setup`, but also inside some `watch` methods. It has to be a reactive variable.

kolibri/core/assets/src/composables/__tests__/useMinimumKolibriVersion.spec.js
jredrejo: sure, done!

kolibri/core/assets/src/composables/useMinimumKolibriVersion.js
MisRob: Everything's possible, Jose, Kolibri is written by humans, not machines (at least not yet) ;-)) Leaving philosophy behind, this is not a blocker when compared to console errors, so it's up to you.

kolibri/plugins/user_profile/assets/src/views/ChangeFacility/SelectFacility.vue
rtibbles: Sorry - I think I misread in my initial review, and thought this was defined at the module level, rather than within the setup method (the `setup` method was hidden in the code fold) - disregard!

kolibri/core/assets/src/composables/useMinimumKolibriVersion.js
jredrejo: yes, you're right. I try to keep it simple (or using the external library would be a better option) but throwing an error as you suggest is harmless if we keep the format as it's supposed and does not add complexity. Done!

MisRob (approved):
I've just tested new updates locally and everything's working great

PR #9573, opened 2022-07-21
Set up to track KDS develop

marcellamaki:
## Summary
Sets up the kds dependency to be `kolibri-design-system#develop` so that KDS and Kolibri frontend work can be done in parallel for 0.16. 

## Reviewer guidance
Have been issues with this in the past, possibly due to yarn caching. We are giving it a try, and can make changes as we work with this! But - anything we should do differently right now?

----

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/core/package.json
+++ b/kolibri/core/package.json
@@ -20,7 +20,7 @@
     "intl": "^1.2.4",
     "knuth-shuffle-seeded": "^1.0.6",
     "kolibri-constants": "0.1.38",
-    "kolibri-design-system": "https://github.com/learningequality/kolibri-design-system#v1.3.1-beta0",
+    "kolibri-design-system": "https://github.com/learningequality/kolibri-design-system#develop",
     "lockr": "0.8.4",
     "lodash": "^4.17.21",
     "loglevel": "^1.4.1",
--- a/yarn.lock
+++ b/yarn.lock
@@ -8083,6 +8083,21 @@ kolibri-constants@0.1.38:
   resolved "https://registry.yarnpkg.com/kolibri-constants/-/kolibri-constants-0.1.38.tgz#f6c51c429963210e6e866c72260c9d13d41a1273"
   integrity sha512-IUWZtm+ceA9PF1lCT8+/E6Cjmy+ihiB+xIKfDGI8SrZ6c2VEG7Ku5tyW26Uf0n34olzjeO9zO6nBYjZ+A92Ijw==
 
+"kolibri-design-system@https://github.com/learningequality/kolibri-design-system#develop":
+  version "1.3.0"
+  resolved "https://github.com/learningequality/kolibri-design-system#37dcf4f8db24d0afc5a72586bb35771dd29c8b1d"
+  dependencies:
+    aphrodite "https://github.com/learningequality/aphrodite/"
+    autosize "^3.0.21"
+    css-element-queries "^1.2.0"
+    frame-throttle "^3.0.0"
+    fuzzysearch "^1.0.3"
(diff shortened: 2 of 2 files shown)


MisRob (approved):
Thank you, @marcellamaki. I think it's good to merge now so that we have the latest KDS `develop` version available.

We will then see if we run into caching troubles. Just a note that even if there are no such problems, we will need to open a Kolibri PR anyways any time we need to have some new KDS `develop` updates available, and that PR will need to have `yarn.lock` file updated (after running `yarn upgrade`) since now you can see there that it's frozen to this particular commit`https://github.com/learningequality/kolibri-design-system#37dcf4f8db24d0afc5a72586bb35771dd29c8b1d`.

rtibbles:
Yeah - because yarn does the right thing and pins git dependencies to a specific commit, this will need to be updated in PRs anyway, so I think the bigger fix here is a workflow fix for people to know how to upgrade a yarn installed dependency from github to the latest commit.

Using `yarn upgrade kolibri-design-system` should work to do this, as long as you have a version of yarn later than 0.22 (ref: https://github.com/yarnpkg/yarn/issues/1573#issuecomment-288869754) - if that doesn't work, then installing to the specific commit of HEAD may be the most robust way.

MisRob:
Thanks @rtibbles, handy. I think the issue you referenced may be the description of caching problems that @nucleogenesis mentioned? Let's see if he can confirm.

I think that if upgrading works then it's nice to have `#develop` in package.json - it's a bit clearer on what is meant to point to. And yes, if there's a problem with upgrading, using the latest commit as you suggest would make sense to me. With any of those ways, there will be some work with PRs, and it's not such a huge difference.

Maybe this could be the first nice task for automation one day? I think it should be possible to automatically open a PR in Kolibri when we merge a PR in KDS, but it's not very clear to me if GitHub actions could also generate a new yarn lock during that process...?

rtibbles:
> Maybe this could be the first nice task for automation one day? I think it should be possible to automatically open a PR in Kolibri when we merge a PR in KDS, but it's not very clear to me if GitHub actions could also generate a new yarn lock during that process...?

Definitely possible - the main thing is that we would need to turn this feature off and on as appropriate in Kolibri, to represent when we're 'tracking the latest commit' vs pinned to a specific release.

A middle ground would be to create a manually initiated action that upgrades KDS in the Kolibri repo. That means we have full control, but people don't have to mess about with doing the upgrade procedure in their local dev environment.

MisRob:
Awesome.

> the main thing is that we would need to turn this feature off and on as appropriate in Kolibri, to represent when we're 'tracking the latest commit' vs pinned to a specific release.

That depends on branches so maybe we could use that as criteria? We always want to track the latest KDS `develop` in Kolibri's `develop`. We newly use release branches in KDS similarly to how we do in Kolibri. Upgrading a Kolibri `release` branch to the latest KDS properly tagged release (made on a KDS release branch) won't be as frequent and I imagine rather doing it rather manually anyways since on release branches, we want to be sure that things are stable and properly tested and documented. `develop` is rather to allow for fast iteration in products since doing KDS releases often just for that purpose is burdensome.

MisRob:
@rtibbles ^

rtibbles:
Looking into this further, it starts getting a little more complex - triggering actions on one repo from events on another. It's possible, but we might need to setup explicit actions in the KDS repo that set off triggers for every listening repo, rather than just letting other repos listen to the KDS repo.

I'll dig into it more to validate approaches.

PR #9575, opened 2022-07-21
calculcate max_ length programmatically in multiple models 

abdelrahman725:
## Summary

instead of hard coding the max_length values for `kind`  `preset`  `extension` fields, this argument now can be calculcated automatically in case any of the choices of the **le_utils**  (for the above fields ) change by just running the migrations and the max_length is calculcated for us  instead of changing the values manually.

## References

[le_utils](https://github.com/learningequality/le-utils/tree/main/le_utils/constants) that may change 

## Reviewer guidance

i think by making sure any related db operation is not affected ?

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated


rtibbles:
What issue is this solving? If we create a constant with a larger value, then we can just increase the max_length appropriately. I am not aware of a case where this has been a problem so far.

abdelrahman725:
> What issue is this solving? If we create a constant with a larger value, then we can just increase the max_length appropriately. I am not aware of a case where this has been a problem so far.

yes we can simply change the value, but that's a manual work isn't it? and that's not for one model only . but anyway it's just a good feature to have

rtibbles:
This is introducing an unnecessary model migration that does not solve any current issue, going to close this.

PR #9576, opened 2022-07-22
Change facility - full name updates on the Create Account page

MisRob:
## Summary

Updates the Create Account page to the latest design updates regarding full name
- the full name textbox has been removed
- the current user's full name is displayed in the message below the title to provide context

| Before | After |
| --------- | ------ |
| before | after |

## References

- Related to #9329
- [Figma designs](https://www.figma.com/file/MpCG7r5PULOCJsEvnTSti8/Android-Designs-2022?node-id=596%3A10379)

### Setup

You need to have two facilities to be able to test the change facility workflow:

**Source facility**
- You can run the development server as usual
- Your Kolibri instance shouldn't be the learner only device and also you need to make sure there's only one user account, otherwise you won't be able to access the _Change learner facility_ feature in _Profile_
- If you already have a target facility with some user accounts, make sure that in the source facility, you log in as a user with a username that doesn't match any username on the target facility (otherwise you'll be taken to a different part of the flow)

**Target facility**
- It needs to be v0.16
- In the facility settings, do _Allow users to create accounts_
- Not required to test this PR's changes but note that if you don't _Require password for learners_ in the facility settings, you won't see password textboxes like on the screenshots above

If you don't have a target facility, you can follow these steps:
1. `cd` to your `kolibri` development directory and make sure to activate your virtual environment like when developing
2. Run ` pip install -e .` to ensure that the target facility has `develop` (`0.16.x`) updates that are required (`kolibri --version` should show something like `0.16.0.dev0+git.20220715072421`)
3. Run Kolibri with the `kolibri` command with `KOLIBRI_HOME` environment variable different from your usual setup so that you don't override your development Kolibri  `KOLIBRI_HOME=~/.tmp kolibri start --foreground`
4. You need to go through the onboarding setup, otherwise the target facility may not be discovered. It seems that we have some onboarding work in progress so it's possible that you won't be able to go through the onboarding forms and complete the setup. I solved this by switching to `0.15.x` version where I did the onboarding and then switched to this branch using the same `KOLIBRI_HOME` and then went through steps 1-3.

### Test scenario

On the source facility
1. Navigate to _Profile -> Change facility -> Change_
2. Select the target facility and click _Continue_
3. On _Change Facility_ page click _Continue_
4. On _Confirm account username_ page, click _Create new account_ button
6. Preview _Create new account_ page

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/plugins/user_profile/assets/src/views/ChangeFacility/index.vue
+++ b/kolibri/plugins/user_profile/assets/src/views/ChangeFacility/index.vue
@@ -118,7 +118,6 @@
           value: {
             facility: this.session.facility_id,
             username: this.session.username,
-            fullName: this.session.full_name,
             role: this.getUserKind,
           },
         });
--- a/kolibri/plugins/user_profile/assets/src/views/ChangeFacility/index.vue
+++ b/kolibri/plugins/user_profile/assets/src/views/ChangeFacility/index.vue
@@ -118,6 +118,7 @@
           value: {
             facility: this.session.facility_id,
             username: this.session.username,
+            fullName: this.session.full_name,
             role: this.getUserKind,
           },
         });
--- a/kolibri/plugins/user_profile/assets/src/machines/changeFacilityMachine.js
+++ b/kolibri/plugins/user_profile/assets/src/machines/changeFacilityMachine.js
@@ -25,6 +25,7 @@ const setInitialContext = assign((_, event) => {
   return {
     sourceFacility: event.value.facility,
     username: event.value.username,
+    fullName: event.value.fullName,
     role: event.value.role,
   };
 });
@@ -81,6 +82,7 @@ export const changeFacilityMachine = createMachine({
   context: {
     role: 'learner',
     username: '',
+    fullName: '',
     targetAccount: {
       username: '',
       password: '',
--- a/kolibri/plugins/user_profile/assets/src/machines/changeFacilityMachine.js
+++ b/kolibri/plugins/user_profile/assets/src/machines/changeFacilityMachine.js
@@ -25,7 +25,6 @@ const setInitialContext = assign((_, event) => {
   return {
     sourceFacility: event.value.facility,
     username: event.value.username,
-    fullName: event.value.fullName,
     role: event.value.role,
   };
 });
@@ -82,7 +81,6 @@ export const changeFacilityMachine = createMachine({
   context: {
     role: 'learner',
     username: '',
-    fullName: '',
     targetAccount: {
       username: '',
       password: '',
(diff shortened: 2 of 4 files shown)


kolibri/plugins/user_profile/assets/src/views/ChangeFacility/index.vue
```diff
@@ -118,6 +118,7 @@
           value: {
             facility: this.session.facility_id,
             username: this.session.username,
+            fullName: this.session.full_name,
```
jredrejo: Not sure if we're needing this in the next steps, but it does not harm to keep it anyway

jredrejo (approved):
Everything looks good and works as intended. Thanks!

kolibri/plugins/user_profile/assets/src/views/ChangeFacility/index.vue
MisRob: It's used here https://github.com/learningequality/kolibri/pull/9576/files#diff-3481d531d29d62496d420d7a168388365376baf2c240b7547703c3e4fed49b51R92. Or did you mean some another future step different from this one?

jredrejo: yes, I meant a future step, if it's not needed later, it does not need to be added to the machine context.

MisRob: You're right, it'll make more sense to rather reach to the session directly from the page. Will update.

Just to learn a bit about patterns for the future, I think that how often is data used is not the main criteria, and your comment is rather about that just in the context of the machine itself, it would be an extra field that's not being used in the machine anywhere so that's why it doesn't make much sense, is that right? Therefore it'd be fine to put something in the context even if it's used only once, but that "once" should happen in the machine (e.g. in `cond`)

jredrejo: yes, I would not add it to the session either. In case other screens in the flow need it, I would add it to the context. My question was if other screens are going to need the fullName. If not, it can be just a data variable in the component.
Anything that needs to be used later, in a `cond` as you say, or in an `action`, should be in the context. Actually I see the context as a kind of vuex object, for anything the states can need.

MisRob: By session, I meant `session` getter, the same that's used in ChangeFacility/index.js

MisRob: I am not sure about the other screens, will have a look. Even if they needed it though, I think that they could too simply use `session` getter directly, rather than saving `session.full_name` to the machine's context only so those pages could get it, when there's no need to go this "longer" path to get to the data, especially when it's useless for the machine itself.

PR #9577, opened 2022-07-25
Create new field to store Facility properties

jredrejo:
## Summary
- Created a JSON extra field for the facility dataset. Having it in JSON format will allow us to add new properties for the facilities not having to run migrations that can be expensive in big databases
- Added properties and setters using this field to get the "on my own setup" new facility property
- Use this new `on_my_own_setup` property to determine if the change facility menu will appear to the user

## References
This will be required both in the new wizard setup and the change facility flow for 0.16

## Reviewer guidance
Does it make sense?
Do tests pass?
To test it in the context of the change facility setup this can be used, after running the migrations and calling `kolibri shell` : 
```
f=Facility.get_default_facility()
f.on_my_own_setup = True
```

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated


kolibri/plugins/user_profile/viewsets.py
```diff
@@ -17,14 +16,12 @@ class UserIndividualViewset(APIView):
...
-
         subset_of_users_device = get_device_setting(
             "subset_of_users_device", default=False
         )
-        users_device = FacilityUser.objects.count()
-
+        user_facility = self.request.user.facility
         return Response(
-            {"individual": users_device == 1, "lod": subset_of_users_device}
+            {"individual": user_facility.on_my_own_setup, "lod": subset_of_users_device}
```
MisRob, line 24: Would it make sense to use "on my own setup" naming instead of "individual" in the response too if this is how we're going to call it? So that we don't have more terms in a product where there are already lots of them.

MisRob (commented):
Thank you, @jredrejo, changes make sense to me overall, but I'll let @rtibbles approve because I could miss backend nuances that he'd notice.

kolibri/plugins/user_profile/viewsets.py
jredrejo: yes, as  you prefer. I just didn't want to change too many things in the current code, this was the cleanest approach but you're right, it drags the previous logic in the naming. In case you prefer it, I'll rename the composable and refactor it with en `onMyOwnSetup` variable instead of individual

MisRob: I'm not sure what's my preference since these are my first encounters with various kinds of setups (and probably that's why using more terms makes things a bit confusing for me). I'd go with whatever we officially/most commonly use? Also, in regards to this PR, it's nothing blocking.

rtibbles: Yeah, I think just up date to `on_my_own_setup` for the response in a follow up PR.

rtibbles (approved):

PR #9584, opened 2022-07-29
Screens for create merge accounts

### LINKED ISSUE #9328: Create "Merge Accounts" sub-flow pages: account selection
# Overview 

Creates pages necessary for a mini sub-flow for merging two existing accounts. 

## Figma 
|   |   |
|---|---|
| If the user is manually adding a user name to be merged |  |
| After the user name has been added OR if there are two accounts with the same username  |  |
|  If an admin account is being used to merge the two user accounts |  |
|  Once the two accounts are selected, the user selects the account information they will use in the new facility |  |

### Accessibility notes
Tab order: 
1) Form elements (in reading order)
2) "Continue" primary button 
3) "Back" secondary button 
4) "X" close out flow 

### Out of scope 

Adding the route(s) is not part of this issue

### Acceptance criteria

- [ ] Pages created with correct text
- [ ] Continue button disabled until checkbox is checked

jredrejo:
## Summary
Screen for the user to confirm its credentials, use admin credentials if  they are available and skip it if the target facility allow users without password to login.

While doing this work, some design discussion have taken to changes in the figma design, that have required changes in the xstate machine, the states `showAccounts` and `editAccountDetails` have been removed. Also, a new state `getUserWithoutPasswordInfo` has been added to retrieve the user information if login without  password is allowed in the target facility.

Screencast when the target facility require password for the users, using the user credentials or admin credentials:
normal

Screencast when the target facility does not require password for the users to login:
passwordless

## References
Closes: #9328 

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/core/auth/api.py
+++ b/kolibri/core/auth/api.py
@@ -262,6 +262,9 @@ class PublicFacilityUserViewSet(ReadOnlyValuesViewset):
         "facility",
         "roles__kind",
         "devicepermissions__is_superuser",
+        "id_number",
+        "gender",
+        "birth_year",
     )
     field_map = {
         "is_superuser": lambda x: bool(x.pop("devicepermissions__is_superuser")),
--- a/kolibri/plugins/user_profile/assets/src/machines/changeFacilityMachine.js
+++ b/kolibri/plugins/user_profile/assets/src/machines/changeFacilityMachine.js
@@ -245,7 +245,7 @@ export const changeFacilityMachine = createMachine({
     requireAccountCreds: {
       meta: { route: 'REQUIRE_ACCOUNT_CREDENTIALS', path: '/change_facility' },
       on: {
-        CONTINUE: 'showAccounts',
+        CONTINUE: { actions: setTargetAccount, target: 'showAccounts' },
         USEADMIN: 'useAdminPassword',
         BACK: 'confirmMerge',
       },
@@ -253,7 +253,7 @@ export const changeFacilityMachine = createMachine({
     useAdminPassword: {
       meta: { route: 'ADMIN_PASSWORD', path: '/change_facility' },
       on: {
-        CONTINUE: 'showAccounts',
+        CONTINUE: { actions: setTargetAccount, target: 'showAccounts' },
         BACK: 'requireAccountCreds',
       },
     },
--- a/kolibri/plugins/user_profile/assets/src/machines/changeFacilityMachine.js
+++ b/kolibri/plugins/user_profile/assets/src/machines/changeFacilityMachine.js
@@ -253,13 +253,6 @@ export const changeFacilityMachine = createMachine({
         BACK: 'requireAccountCreds',
       },
     },
-    // showAccounts: {
-    //   meta: { route: 'SHOW_ACCOUNTS', path: '/change_facility' },
-    //   on: {
-    //     CONTINUE: 'confirmAccountDetails',
-    //     BACK: 'requireAccountCreds',
-    //   },
-    // },
     confirmAccountDetails: {
       meta: { route: 'CONFIRM_DETAILS', path: '/change_facility' },
       on: {
@@ -267,13 +260,6 @@ export const changeFacilityMachine = createMachine({
         BACK: 'mergeAccounts',
       },
     },
-    // editAccountDetails: {
-    //   meta: { route: 'EDIT_DETAILS', path: '/change_facility' },
-    //   on: {
-    //     SAVE: { actions: saveAccountDetails },
-    //     BACK: 'confirmAccountDetails',
-    //   },
-    // },
     mergeAccounts: {
       meta: { route: 'MERGE_ACCOUNTS', path: '/change_facility' },
       on: {
--- a/kolibri/plugins/user_profile/assets/src/machines/changeFacilityMachine.js
+++ b/kolibri/plugins/user_profile/assets/src/machines/changeFacilityMachine.js
@@ -1,6 +1,6 @@
 import { createMachine, assign } from 'xstate';
 import {
-  default as remoteFacilityLoginUser,
+  default as remoteFacilityUserData,
   remoteFacilityUsers,
 } from '../composables/useRemoteFacility';
 
@@ -41,7 +41,7 @@ const connectToTargetKolibri = (context, event) => {
 
 const getUserWPasswordInfo = context => {
   const facility = context.targetFacility;
-  return remoteFacilityLoginUser(facility.url, facility.id, context.username, null).then(
+  return remoteFacilityUserData(facility.url, facility.id, context.username, null).then(
     user => user
   );
 };
--- a/kolibri/plugins/user_profile/assets/src/machines/changeFacilityMachine.js
+++ b/kolibri/plugins/user_profile/assets/src/machines/changeFacilityMachine.js
@@ -74,6 +74,7 @@ const setMerging = assign({
 export const changeFacilityMachine = createMachine({
   id: 'machine',
   initial: 'selectFacility',
+  predictableActionArguments: true,
   context: {
     role: 'learner',
     username: '',
--- a/kolibri/plugins/user_profile/assets/src/machines/changeFacilityMachine.js
+++ b/kolibri/plugins/user_profile/assets/src/machines/changeFacilityMachine.js
@@ -1,6 +1,9 @@
-import client from 'kolibri.client';
-import urls from 'kolibri.urls';
 import { createMachine, assign } from 'xstate';
+import {
+  default as remoteFacilityLoginUser,
+  remoteFacilityUsers,
+} from '../composables/useRemoteFacility';
+
 // This machine can be visualized and tested at https://stately.ai/viz/c1316a38-033d-4c57-8d9f-e282310e341d
 /* SETCONTEXT event example:
 {
@@ -31,19 +34,16 @@ const setInitialContext = assign((_, event) => {
 
 const connectToTargetKolibri = (context, event) => {
   const facility = event.value;
-  const params = {
-    baseurl: facility.url,
-    facility: facility.id,
-    username: context.username,
-  };
-  return client({
-    url: urls['kolibri:kolibri.plugins.user_profile:remotefacilityuser'](),
-    params: params,
-  }).then(response => {
-    let users = response.data;
-    if (Object.keys(response.data).length === 0) users = [];
-    return { facility: facility, users: users };
+  return remoteFacilityUsers(facility.url, facility.id, context.username).then(users => {
+    return { facility: facility, ...users };
   });
+};
+
+const getUserWPasswordInfo = context => {
+  const facility = context.targetFacility;
+  return remoteFacilityLoginUser(facility.url, facility.id, context.username, null).then(
+    user => user
+  );
 };
 
 const checkExists = assign((context, event) => {
@@ -70,10 +70,6 @@ const setTargetAccount = assign({
 const setMerging = assign({
   isMerging: () => true,
 });
-
-const saveAccountDetails = () => {
-  // to be done
-};
 
 export const changeFacilityMachine = createMachine({
   id: 'machine',
@@ -245,7 +241,7 @@ export const changeFacilityMachine = createMachine({
     requireAccountCreds: {
       meta: { route: 'REQUIRE_ACCOUNT_CREDENTIALS', path: '/change_facility' },
       on: {
-        CONTINUE: { actions: setTargetAccount, target: 'showAccounts' },
+        CONTINUE: { actions: setTargetAccount, target: 'confirmAccountDetails' },
         USEADMIN: 'useAdminPassword',
         BACK: 'confirmMerge',
       },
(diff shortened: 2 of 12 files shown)


kolibri/core/auth/api.py
```diff
@@ -262,6 +262,9 @@ class PublicFacilityUserViewSet(ReadOnlyValuesViewset):
         "facility",
         "roles__kind",
         "devicepermissions__is_superuser",
+        "id_number",
```
rtibbles, line 265: This feels ok to me as we are still requiring admin authentication to access this data for someone other than the user in question.

kolibri/plugins/user_profile/viewsets.py
```diff
@@ -45,3 +45,28 @@ def get(self, request):
                 return Response({})
         except Exception as e:
             raise ValidationError(detail=str(e))
+
+
+class RemoteFacilityUserLoginViewset(APIView):
```
rtibbles: I found this naming confusing - it made me think that this was an endpoint for logging into a remote device with a specific facility user, rather than getting a facility user while logged in? Could it be `RemoteFacilityUserAuthenticatedViewset`?

Also wonder if this should be a `POST` request to avoid putting the username and password in plain text in the URL?

rtibbles (commented):
Did a quick skim to look at the API endpoints - one possible concern that came to mind as it has been raised previously in security audits.

kolibri/plugins/user_profile/viewsets.py
jredrejo: Happy to refactor the name. I hate naming things :) And for the `POST` recommendation, done too, it's clearly a good idea, thanks!

MisRob:
@jredrejo Could you please add "Reviewer guidance"?

kolibri/plugins/user_profile/assets/src/routes.js
```diff
@@ -100,28 +103,28 @@ export default [
...
-        component: ToBeDone,
-      },
-      {
-        path: 'show_accounts',
-        name: 'SHOW_ACCOUNTS',
-        component: ToBeDone,
+        component: MergeAccountDialog,
       },
+      // {
+      //   path: 'show_accounts',
```
MisRob: Will we need this? Same applies to lines 123-127 below.

kolibri/plugins/user_profile/assets/src/views/ChangeFacility/MergeAccountDialog/__tests__/ConfirmAccountDetails.spec.js
```diff
@@ -0,0 +1,77 @@
...
+
+  it(`clicking the continue button sends the continue event to the state machine`, () => {
+    const wrapper = makeWrapper();
+    clickContinueButton(wrapper);
+    expect(sendMachineEvent).toHaveBeenCalledWith({
+      type: 'CONTINUE',
+    });
+  });
+
+  it(`clicking the back button sends the continue event to the state machine`, () => {
```
MisRob: "...sends the back event..."

kolibri/plugins/user_profile/assets/src/views/ChangeFacility/MergeAccountDialog/__tests__/index.spec.js
```diff
@@ -0,0 +1,147 @@
...
+    },
+    mocks: {
+      $store: {
+        getters: {
+          session: { full_name: fullName, username: username },
+        },
+      },
+    },
+    localVue,
+    store,
```
MisRob: I don't understand why `store` is needed here when above you use the store mock, could you explain?

kolibri/plugins/user_profile/assets/src/views/ChangeFacility/MergeAccountDialog/index.vue
```diff
@@ -0,0 +1,243 @@
+<template>
+
+  <div>
+    <h1>{{ $tr('documentTitle') }}</h1>
+    <h2 data-test="fullName">
+      {{ fullName }}
+    </h2>
+    <h3 data-test="username">
```
MisRob: Headings are used for creating subsections in documents, so I don't think we need `h2` and `h3` here. This would rather create a bit confusing structure as this implementation presents the user name sub-section as a child of the full name sub-section, and all content below `h3` as belonging to the user name sub-section, which doesn't make sense as username and full name merely describe the same user whose username is not subordinate to full name. That's why I suggest removing `h2` and `h3`.

kolibri/plugins/user_profile/assets/src/views/ChangeFacility/MergeAccountDialog/__tests__/index.spec.js
```diff
@@ -0,0 +1,147 @@
...
+    expect(wrapper.exists()).toBeTruthy();
+  });
+
+  it(`Show correct info`, () => {
+    const wrapper = makeWrapper({
+      targetFacility: { name: 'Test Facility' },
+      fullName: 'Test User 1',
+      username: 'test1',
+    });
+    const h2_fullname = wrapper.find('[data-test="fullName"]');
```
MisRob: If you wanted to communicate that `h2` and `h3` are rendered, then I'd suggest rather having a test that will check an element than using a prefix. That said, in another comment, I actually recommended removing `h2` and `h3` from markup.

kolibri/plugins/user_profile/assets/src/views/ChangeFacility/MergeAccountDialog/index.vue
```diff
@@ -0,0 +1,243 @@
...
+    mixins: [commonCoreStrings],
+    setup(_, context) {
+      const changeFacilityService = inject('changeFacilityService');
+      const state = inject('state');
+
+      const isFormSubmitted = ref(false);
+      const isPasswordInvalid = ref(false);
+      const usingAdminPasswordState = ref(false);
+      // hack to get the $store in vue tests with composition api:
+      const store =
```
MisRob, line 99: It's also possible to use `store` together with the composition API like here https://github.com/learningequality/kolibri/blob/develop/kolibri/plugins/learn/assets/src/composables/useUser.js. I didn't try if this will help with your tests, but I imagine you could mock the whole store file, for example.

kolibri/plugins/user_profile/assets/src/views/ChangeFacility/index.vue
```diff
@@ -92,8 +92,9 @@
       this.service.stop();
     },
     mounted() {
-      this.appBarHeight = this.$refs.appBar.$el.clientHeight;
-
+      if (this.$refs.appBar) {
```
MisRob: I think it'd be safer to wrap this into `$nextTick` because [in `mounted` it can happen that not all children are rendered yet](https://v2.vuejs.org/v2/api/#mounted). Therefore, with this condition, if we don't do it within `$nextTick` I think it could happen that we don't set `appBarHeigh` even though the app bar will be shown a while later.

MisRob (commented):
Thank you, @jredrejo, overall it's looking great. I appreciate tests and clarity of code.

I'm posting several first comments. I still want to look into one or two things but for that I need to be able to walkthrough this pathway. Unfortunately, I am getting stuck on _/en/profile/#/change_facility/merge_accounts_. Clicking _Continue_ button does nothing and I can see no errors in the console. I assume I'm in the correct pathway? Any idea what may be happening?
merge-accounts

Requests are too looking fine
Screenshot from 2022-09-01 15-28-21

jredrejo:
> Thank you, @jredrejo, overall it's looking great. I appreciate tests and clarity of code.
> 
> I'm posting several first comments. I still want to look into one or two things but for that I need to be able to walkthrough this pathway. Unfortunately, I am getting stuck on _/en/profile/#/change_facility/merge_accounts_. Clicking _Continue_ button does nothing and I can see no errors in the console. I assume I'm in the correct pathway? Any idea what may be happening? 

Yes, you're in the right path, clicking on continue should take you to the screens this PR implements. In the state you're there are two guards:
```
        CONTINUE: [
          {
            cond: context =>
              context.accountExists && !context.targetFacility.learner_can_login_with_no_password,
            target: 'requireAccountCreds',
          },
          {
            cond: context =>
              context.accountExists && context.targetFacility.learner_can_login_with_no_password,
            target: 'getUserWithoutPasswordInfo',
          },
        ],
```
so, your username should exist as an account in the target facility, and depending on the `learner_can_login_with_no_password` parameter of the target facility will take you to one screen or another. Could it be the username is not in the other, thus it can't be merged?

kolibri/plugins/user_profile/assets/src/routes.js
jredrejo: no, we don't need them. I left those lines to remember we had those routes in the original design and now they don't exist. I thought it could be useful in case of doubts during this development but I am deleting them.

kolibri/plugins/user_profile/assets/src/views/ChangeFacility/MergeAccountDialog/__tests__/ConfirmAccountDetails.spec.js
jredrejo: oh, yes, my typo!, thanks

kolibri/plugins/user_profile/assets/src/views/ChangeFacility/MergeAccountDialog/__tests__/index.spec.js
jredrejo: right, it's not needed. I did so many changes in this file that I left remaining code from some of my previous tests. Deleted!

kolibri/plugins/user_profile/assets/src/views/ChangeFacility/MergeAccountDialog/__tests__/index.spec.js
jredrejo: done, replaced by paragraph tags

kolibri/plugins/user_profile/assets/src/views/ChangeFacility/MergeAccountDialog/index.vue
jredrejo: done, replaced by paragraph tags

kolibri/plugins/user_profile/assets/src/views/ChangeFacility/MergeAccountDialog/index.vue
jredrejo: yes, but didn't work inside the `setup` function of the component because `setup` is executed before anything else, including creating the component and assigning the store. I tested it a lot and was not able to find a better solution to make it work both in the app and when running the tests

kolibri/plugins/user_profile/assets/src/views/ChangeFacility/index.vue
jredrejo: makes sense, I just copied & pasted how this was used in other parts of kolibri where ImmersivePage is used, but changed it here as you suggested and seems to work fine , so I've changed it.

rtibbles (commented):
Backend updates make sense to me, and seem to be reflected in the frontend code as far as I could tell.

kolibri/plugins/user_profile/assets/src/machines/changeFacilityMachine.js
```diff
@@ -74,6 +74,7 @@ const setMerging = assign({
 export const changeFacilityMachine = createMachine({
   id: 'machine',
   initial: 'selectFacility',
+  predictableActionArguments: true,
```
MisRob, line 77: Could you please explain what the reason for this change was?

I am not sure if I can entirely understand the documentation and am a bit concerned about their warning as we use `meta` often
> Please be aware that you might not able to use state from the meta argument when using this flag.

MisRob: https://xstate.js.org/docs/guides/actions.html#actions

MisRob:
> your username should exist as an account in the target facility

Yes, this was the problem. I was able to test it after making sure I have this set.

MisRob (changes requested):
Thank you, @jredrejo, it's looking good to me, except of one thing - I can see there's a deviation from design when it comes to the full name and username:

**Implementation**
Screenshot from 2022-09-05 13-59-05

**Design**
- full name is normal text
- username is small text and I believe it uses `tokens.annotation` color
- spacing between the full name and username is much smaller

Screenshot from 2022-09-05 13-57-47

After this is resolved, I think it will be ready for merge.

kolibri/plugins/user_profile/assets/src/machines/changeFacilityMachine.js
jredrejo: it was recommended by the xstate library in the browser warnings, you can see it in the screenshot you attached at https://github.com/learningequality/kolibri/pull/9584#pullrequestreview-1093031427

jredrejo:
> **Design**
> 
> * full name is normal text
> * username is small text and I believe it uses `tokens.annotation` color
> * spacing between the full name and username is much smaller

Argg, thanks for spotting it, I've never had a good design eye. Changed, now it looks like this: 
image

MisRob (approved):

PR #9585, opened 2022-08-01
[PDF_Viewer] Adding struct tree layer

AlexVelezLl:
## Summary

Adding a new layer that adds a PDF tree structure of tagged PDFs so that screen readers can notice the semantic structure of a tagged PDF.

With this layer screen readers should now be able to recognize among other things:
* Headings of different levels.
* Paragraphs
* Lists/List items
* Tables
* Figures alt text
* Links

## References

This PR solves the lack of semantic structure of the text layer that was missing in #9536.

## Reviewer guidance

1. Go to learn tab
2. pick any reading-type resource that has a tagged PDF (e.g. The pdfs that are in the Kolibri QA channel)
3. Use a screen reader and note if the screen reader is able to identify the semantic structure of the pdf. 

----

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/locale/es_419/LC_MESSAGES/kolibri.core.default_frontend-messages.json
+++ b/kolibri/locale/es_419/LC_MESSAGES/kolibri.core.default_frontend-messages.json
@@ -236,6 +236,7 @@
   "CommonCoreStrings.tasksLabel": "Tareas",
   "CommonCoreStrings.technicalAndVocationalTraining": "Formación técnica y profesional",
   "CommonCoreStrings.tertiary": "Enseñanza superior",
+  "CommonCoreStrings.timeSpentLabel": "Tiempo utilizado",
   "CommonCoreStrings.topicLabel": "Carpeta",
   "CommonCoreStrings.transcript": "Transcripción",
   "CommonCoreStrings.updateAction": "Actualizar",
--- a/kolibri/plugins/pdf_viewer/assets/src/utils/event_utils.js
+++ b/kolibri/plugins/pdf_viewer/assets/src/utils/event_utils.js
@@ -1,0 +1,130 @@
+/* Copyright 2012 Mozilla Foundation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/*
+ * This file has been modified to adapt the component to the needs of Kolibri
+ * The original file is available at:
+ * https://github.com/mozilla/pdf.js/blob/v2.14.305/web/event_utils.js
+ */
+
+// Modified: Just the eventBus class is used
+
+/**
+ * Simple event bus for an application. Listeners are attached using the `on`
+ * and `off` methods. To raise an event, the `dispatch` method shall be used.
+ */
+class EventBus {
+  constructor() {
+    this._listeners = Object.create(null);
+  }
+
+  /**
+   * @param {string} eventName
+   * @param {function} listener
+   * @param {Object} [options]
+   */
+  on(eventName, listener, options = null) {
+    this._on(eventName, listener, {
+      external: true,
+      // Modified: Validate options is defined instead of using optional chain operator
+      // Original line: 90
+      once: options ? options.once : false,
+    });
+  }
+
+  /**
+   * @param {string} eventName
+   * @param {function} listener
+   * @param {Object} [options]
+   */
+  off(eventName, listener, options = null) {
+    this._off(eventName, listener, {
+      external: true,
+      // Modified: Validate options is defined instead of using optional chain operator
+      // Original line: 102
+      once: options ? options.once : false,
+    });
+  }
+
+  /**
+   * @param {string} eventName
+   * @param {Object} data
+   */
+  dispatch(eventName, data) {
+    const eventListeners = this._listeners[eventName];
+    if (!eventListeners || eventListeners.length === 0) {
+      return;
+    }
+    let externalListeners;
+    // Making copy of the listeners array in case if it will be modified
+    // during dispatch.
+    for (const { listener, external, once } of eventListeners.slice(0)) {
+      if (once) {
+        this._off(eventName, listener);
+      }
+      if (external) {
+        // Modified: Validate externalListeners is defined instead of using ||= operator
+        // Original line: 123
+        externalListeners = externalListeners || [];
+        externalListeners.push(listener);
+        continue;
+      }
+      listener(data);
+    }
+    // Dispatch any "external" listeners *after* the internal ones, to give the
+    // viewer components time to handle events and update their state first.
+    if (externalListeners) {
+      for (const listener of externalListeners) {
+        listener(data);
+      }
+      externalListeners = null;
+    }
+  }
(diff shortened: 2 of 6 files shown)


kolibri/locale/es_419/LC_MESSAGES/kolibri.core.default_frontend-messages.json
```diff
@@ -236,6 +236,7 @@
   "CommonCoreStrings.tasksLabel": "Tareas",
   "CommonCoreStrings.technicalAndVocationalTraining": "Formación técnica y profesional",
   "CommonCoreStrings.tertiary": "Enseñanza superior",
+  "CommonCoreStrings.timeSpentLabel": "Tiempo utilizado",
```
AlexVelezLl, line 239: This translation was missing so I put it manually because it gave me many errors in the browser

kolibri/plugins/pdf_viewer/assets/src/utils/event_utils.js
```diff
@@ -0,0 +1,130 @@
...
+ * https://github.com/mozilla/pdf.js/blob/v2.14.305/web/event_utils.js
+ */
+
+// Modified: Just the eventBus class is used
+
+/**
+ * Simple event bus for an application. Listeners are attached using the `on`
+ * and `off` methods. To raise an event, the `dispatch` method shall be used.
+ */
+class EventBus {
```
AlexVelezLl, line 3: I copied this class to handle event based communication similar to how they do in mozilla's pdf reader. I think it could be replaced with native javascript custom events, although I'm not sure how much it would affect cross-browser compatibility.

radinamatic:
Alex, excellent progress, you're spoiling us! 👏🏽 👏🏽 👏🏽 

Headings, lists, links, even tables, all well recognized and announced by the screen reader 🎉 

The only sore point are the images that are not announced, nor their `alt` text is read out, even though it's there as an `aria-label`. I don't think it's because of your implementation, they do not seem to be recognized and announced by NVDA even if I open the PDF directly in the browser. Could it be because the `role=figure` is used? Would it be viable to try `role=img` instead?

Win10 (start)  Running  - Oracle VM VirtualBox_001

radinamatic:
Another point for potential improvement would be to explore the possibility to exclude empty paragraphs from the text layer as those are announced as `blank`, `blank`, `blank` by NVDA, and it's an annoyance.

Win10 (start)  Running  - Oracle VM VirtualBox_002


Sighted PDF users will not miss them when copying text, they can't be highlighted, and are just a bad experience for blind readers.

Empty paragraphs in the Acrobat tag tree view:

kolibri/plugins/pdf_viewer/assets/src/utils/struct_tree_layer_builder.js
```diff
@@ -0,0 +1,139 @@
...
+  TR: 'row',
+  TH: 'columnheader',
+  TD: 'cell',
+  THead: 'columnheader',
+  TBody: null,
+  TFoot: null,
+  // Standard structure type Caption
+  Caption: null,
+  // Standard structure type Figure
+  Figure: 'figure',
```
radinamatic: @AlexVelezLl Yes, ``role="img"``was the key to have the image appropriately announced by NVDA, and read the alternative description!
Please change this remap value so the images render as ``<span role="img" aria-label="...``

kolibri/plugins/pdf_viewer/assets/src/utils/event_utils.js
marcellamaki: As I mentioned during our checkin yesterday, I chatted with @rtibbles about the eventBus, and he suggested following [this `_eventDispatch` pattern that we use in the Kolibri Design System](https://github.com/learningequality/kolibri-design-system/blob/develop/lib/utils/pluginMediator.js#L64). 

As we are using it for the design system, I think it should be fine for all of the browsers we support. 

You can look through this file to see some examples of how it is being used. Take a look and see what you think! Let me or @rtibbles know if you have any questions.

marcellamaki (dismissed):
Hi @AlexVelezLl - this is looking good with an initial read through, and the fact that @radinamatic's manual QA is looking so good is *really* exciting. 🎉 

I added one comment about the eventBus as we discussed yesterday, and linked to the example that Richard shared with me. I think for consistency, it would be good to replace the eventBus with this pattern if possible, since we are using this in one of our other products. Take a look and see if you have any questions. As I mentioned, it's not a piece of code I'm very familiar with, but I would be happy to find some time to go through it together if that would be helpful. 

Great work so far!

kolibri/plugins/pdf_viewer/assets/src/utils/event_utils.js
rtibbles: My main point here was that you can just use an instance of the `Vue` object as an event bus, which is how it is being used in the KDS code that @marcellamaki linked. So this should reduce the amount of vendored code required, and mean that we can rely on Vue's event bus which we are pretty darn sure works!

AlexVelezLl: Nice! I'll use that then

radinamatic:
This is a thing of beauty, now Kolibri PDF viewer offers a slightly better experience for screen reader users compared even to the native browser PDF reader on this aspect... 🤩 😍  Thank you, @AlexVelezLl!!!

Since you've been churning these improvements with such efficiency, here is one last request: let's try to implement a similar rendering of the PDF file pages as ``pdf.js`` does.

You implemented the rendering of the PDF page as a ``section`` HTML element with a ``pdf-page`` class, but class names are not semantic and are not accessible to assistive technology. On another hand  ``pdf.js`` renders a page as ``div`` element with a  `role="region"` and `aria-label="Page N"` which are both revealed to screen readers, and could potentially help navigate the PDF document more easily.

Kolibri |  ``pdf.js`` 
--- | ---
LEDev2202 (start)  Running  - Oracle VM VirtualBox_002 | LEDev2202 (start)  Running  - Oracle VM VirtualBox_003

My thinking goes in direction to try to offer a similar experience and path to knowing the PDF page numbers in Kolibri viewer as in the browser PDF reader, for those users who are already familiar and use the latter.

AlexVelezLl:
@radinamatic It sounds great! I had not noticed that this was missing, I already pushed those changes, although the translations are missing.

marcellamaki (commented):
Code looks great, @AlexVelezLl - waiting for final ✅ from @radinamatic on the last commit. Once she does the final QA on the screen reader, we will merge. 🎉

kolibri/plugins/pdf_viewer/assets/src/views/PdfPage.vue
```diff
@@ -177,8 +191,37 @@
...
+          if (!this.$refs.canvas) {
+            return;
+          }
+          const treeDom = this.structTreeLayer.render(tree);
+          treeDom.classList.add('structTree');
+          this.$refs.canvas.appendChild(treeDom);
+        });
+      },
+    },
+    $trs: {
```
radinamatic, line 223: We discussed including this in the CommonCoreStrings, right @marcellamaki?
https://github.com/learningequality/kolibri/blob/develop/kolibri/core/assets/src/mixins/commonCoreStrings.js#L150

radinamatic:
The new page numbering works well, and I like it 💯 , the only twist is that the lazy loading is messing with the order when one scrolls up and down the document 😳 
I'm guessing there isn't really much we can do about it, right? 

LEDev (start)  Running  - Oracle VM VirtualBox_003

AlexVelezLl:
@radinamatic I think it's a bit complicated because all of that is handled by the RecycleList internally. Although I could search if there is any configuration for that 🤔

radinamatic:
As discussed in our weekly call, this is working great, and ready to merge!

Thank you, I am so looking forward to what's coming next! 😍

radinamatic (approved):
LGTM! 💯 :shipit:

Issue #9326, opened 2022-04-19
Create "Choose Admin" page


marcellamaki:
# Overview 

Adds a page where a user can select a new admin, if the current user leaving the facility is the superadmin.

## Figma 
![Screen Shot 2022-04-20 at 11.08.44 AM.png](https://images.zenhubusercontent.com/6047f6e691103a78d3ad0ace/cca831f3-fc71-4b8f-b145-f6807782f518)

### Accessibility notes
Tab order: 
1) Search for user
2) Radio buttons in read order (top to bottom)
3) "Continue" primary button 
4) "Back" secondary button 
5) "X" close out flow 

### Out of scope 

Adding the route itself is not part of this issue

### Acceptance criteria

- [ ] Page is created with correct strings
- [ ] Results correctly display possible admins to choose from, and using the "search for a user" correctly filters results
- [ ] UI layout to spec with correct tab order and spacing
- [ ] Continue button disabled until a user is selected

MisRob:
Before implementing this issue, we need to decide what happens in the change facility workflow when there's only one user and therefore there's no one to choose the new admin from. This could happen in general and at this point, we don't even allow a user to access the change facility feature if their facility has some other users. Therefore with our current assumptions, implementing this step wouldn't make sense right now since the only user will never have some other users to choose from.

Even if we decided to allow the change facility feature on facilities that have more users, we will still need to clarify what should happen when there's only one since that could happen. I assume that we could simply skip this screen in such a case, but let's confirm that first as it relates to other decisions that need to be made.

Some suggestions and more information can be found [in this Slack thread](https://learningequality.slack.com/archives/CB37UM23A/p1657194662123749), but we didn't make a decision yet.

cc @jtamiace @marcellamaki @jredrejo @rtibbles

jredrejo:
@MisRob is this blocking the work on the following screens?
If it's blocking it I think you could do a blank screen with just a continue button so we continue working in the rest of the epic while a decission is made.

MisRob:
I think it doesn't need to block work on other screens since we use the state machine. We can implement states quite independently on another states, and then at some point, connect the remaining pieces, I'd say?

MisRob:
- we decided to allow access to the change facility workflow on certain kinds of facilities even when they have more users => makes sense to implement it
- when there's only one user we'll skip this screen

PR #9597, opened 2022-08-10
Update UserTable to allow single selection

MisRob:
## Summary

- Adds a basic test suite for `UserTable`, particularly for areas that are later updated by this pull request
- Adds a new property, `enableMultipleSelection` defaulting to `true`. When the table is selectable and `enableMultipleSelection` is `true`, checkboxes will be displayed for user selection which covers all our previous use cases of the selectable table.
- Implements radio buttons for a single user selection when `enableMultipleSelection` is `false`. This is to be used on the choose admin page of the change facility flow (not included in this pull request):

- Moves the table to core components since it's used by more plugins
- Related accessibility improvements of the selectable table (cc @radinamatic)
  - Moved the _Select all_ checkbox outside of the first table column header and used CSS to position it instead. This allowed this `<th>` to be used for describing what that table column contains instead, by means of the _Select user by:_ visually hidden text that precedes the _Full name_ text
   - Used users' full names as labels of checkboxes/radio buttons. Previously, all checkboxes had the same _Select user_ visually hidden label which couldn't communicate well what user is being selected. Based on recommendations [from this WebAIM archive thread](https://webaim.org/discussion/mail_thread?thread=4161).

## References

Related https://github.com/learningequality/kolibri/issues/9326

Opened KDS issues related to some limitations I experienced:
- https://github.com/learningequality/kolibri-design-system/issues/347
- https://github.com/learningequality/kolibri-design-system/issues/348
- https://github.com/learningequality/kolibri-design-system/issues/349

## Reviewer guidance

I wanted to implement radio button selection in an accessible way. Doing it only for radio buttons without making any changes to checkboxes would be complicated since the previous and new solution has different table columns structure. Because there's lots of refactoring in that regard, I'm opening a separate pull request which purpose is to make sure that the new structure makes sense and that there are no regressions at places where we already use `UserTable`. Therefore, during the review, I'd suggest:

- Check that there are no visual regressions in the following features:
  - Coach
    - Enroll users to a group
  - Facility
    - Facility users table 
    - Coaches and learners preview tables on a class page
    - Enroll learners to a class
    - Assign coaches to a class
- Check that there are no functional regressions when selecting multiple users, for example by enrolling learners to a group as a coach
- Check tables raw HTML output, including its visually hidden parts in regards to accessibility, in both selectable and non-selectable mode
- In regards to the new single selection behavior, does code for radio buttons make sense? I will provide a way to preview the table with radio buttons in a subsequent pull request for the choose admin page if that's fine. If you want to try it out sooner, let me know, and I can provide my working branch.

Note: Since I moved the file, if you want to preview actual changes line by line, you will need to go commit by commit.

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated


MisRob:
@radinamatic I'd appreciate if you could have a look at tables in regards to a11y things I mentioned in the description (cc'ed you in the relevant parts)

radinamatic:
@MisRob Thank you for investing effort to find a better output for the screen reader users! 🙏🏽 

I will do some more tests with NVDA to ensure all the visually hidden information makes sense, but one clear issue became evident so far: while for the **Select all** checkbox the label is correctly configured, that is not the case for the rest of the checkboxes related to users. Could you change the `span` element into `label` with a proper `for` attribute? That should probably be sufficient. 

Select all | User
--- | ---
Selection_024 | Selection_023

MisRob:
Thank you @radinamatic.

> while for the Select all checkbox the label is correctly configured, that is not the case for the rest of the checkboxes related to users. Could you change the span element into label with a proper for attribute? That should probably be sufficient.

I should have mentioned this louder, this is one of the KDS issues that I reported here https://github.com/learningequality/kolibri-design-system/issues/347 (linked with some other issues in the description of the PR in the references section). The reason why it works for the "Select all" checkbox is that we use the `KCheckbox` component in a bit different way in that case. But looking at `KCheckbox` again, I think I noticed something that I had missed before that I could use to deal with this better, so let me play around with it before you test it again. I will let you know.

MisRob:
@radinamatic No, so it didn't work. I think it'd be best to prioritize fixing https://github.com/learningequality/kolibri-design-system/issues/347, it should be straightforward and it will also fix other places in Kolibri where we lack checkboxes labels.

MisRob:
@radinamatic Taking into account that it's one of the most important a11y things I'll go ahead and set P0 on that issue and make sure to mention it during the next iteration planning

MisRob:
Now I realize that the previous implementation may be in part governed by the referenced KDS limitation. However, would rather not go back even though this would be temporary regression since it had other problems, it complicated work on radio buttons, and avoiding fixing this in KDS affects negatively more places. I think it'd be fine to leave this PR open though until we address it in KDS to make sure it's okay.

radinamatic:
I'm all for pushing up for a solution in KDS that would have a broader reach, compared with patching things up separately in various places. Keeping an 👁️ on the PR that you mention, hit me when it needs testing! 🙏🏽

MisRob:
Rebased on top of the latest `develop` that contains fixed KCheckbox so this is ready for review and testing again @radinamatic and @marcellamaki

radinamatic:
All the checkboxes are now properly labelled, thank you!!! 👏🏽 💯

radinamatic (approved):
Major accessibility issue fixed and included in this PR, great progress! 👍🏽 :shipit:

MisRob:
Reverting the merge as it was merged by mistake and it also broken tests on `develop`

Issue #9613, opened 2022-08-12
Add sanity check to ensure that Job Storage tables are ready


rtibbles:
## Observed behavior
In 0.15 the deletion and recreation of Job Storage tables at every restart was removed. This means that there are some possible scenarios where Kolibri could attempt to start, but the storage tables do not exist.

## Errors and logs
```
Traceback (most recent call last):
  File "/usr/local/lib/python3.8/site-packages/kolibri/dist/magicbus/base.py", line 215, in _transition
    return self.publish(newstate, *args, **kwargs)
  File "/usr/local/lib/python3.8/site-packages/kolibri/dist/magicbus/base.py", line 291, in publish
    raise exc
magicbus.base.ChannelFailures: ProgrammingError('(psycopg2.errors.UndefinedColumn) column scheduledjobs.saved_job does not exist\nLINE 1: ...peat, scheduledjobs.queue AS scheduledjobs_queue, scheduledj...\n       
```

## Expected behavior
Kolibri should check that these tables are accessible and recreate them if not.

rtibbles:
Fixed in #9722

PR #9615, opened 2022-08-15
Catch exception when job and scheduledjobs tables are not created

jredrejo:
## Summary
  In some cases, as reported in #9613 and https://github.com/learningequality/kolibri-data-portal/issues/407 job tables might not be created before trying to use the scheduler.
This PR captures and resolves the problem as a quick aproximation to fix the issue

## References
Fixes: #9613 

## Reviewer guidance
* Do test pass?

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/core/tasks/scheduler.py
+++ b/kolibri/core/tasks/scheduler.py
@@ -7,6 +7,7 @@ from sqlalchemy import DateTime
 from sqlalchemy import Index
 from sqlalchemy import Integer
 from sqlalchemy import String
+from sqlalchemy.exc import ProgrammingError
 from sqlalchemy.ext.declarative import declarative_base
 
 from kolibri.core.tasks.exceptions import JobNotFound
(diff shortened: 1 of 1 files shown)


kolibri/core/tasks/scheduler.py
```diff
@@ -173,7 +175,12 @@ def schedule(self, dt, func, interval=0, repeat=0, *args, **kwargs):
                 scheduled_time=naive_utc_datetime(dt),
                 saved_job=job.to_json(),
             )
-            session.merge(scheduled_job)
+            try:
+                session.merge(scheduled_job)
+            except ProgrammingError:
+                # needs to create job storage tables if they don't exist
+                job_storage.recreate_tables()
```
rtibbles, line 184: This won't fix the issue, as it's only recreating the job storage, not the scheduled jobs table.

I think rather than trying to catch it at the time of erroring, we should do a startup sanity check like we do for Django migrations: https://github.com/learningequality/kolibri/blob/develop/kolibri/utils/sanity_checks.py#L96

We can then recreate both the `job_storage` and the `scheduled_jobs` tables if needed. Can then run this inside the intiialize method: https://github.com/learningequality/kolibri/blob/develop/kolibri/utils/main.py#L264

rtibbles (changes requested):
Let's do the more systematic sanity check approach, and make sure to check and migrate both job tables (noting that in develop, this is then collapsed back to 1 table).

jredrejo:
I cheated it so it creates the scheduler tables as well, but I do agree it's better to have it in a sanity check in a prior stage of kolibri starting. I just thought this was a quick hack. Closing it by now until I (or someone else) has some bandwidth to do a better implementation.

PR #9616, opened 2022-08-15
Points popover and overflow

### LINKED ISSUE #9276: Remove scrolling nav links on smaller breakpoints and add overflow dropdown menu
This issue is part of the app bar refresh work. The scope of this issue is to remove the scrolling navigation links on smaller screens, which is not user friendly, and place overflow options into a menu.

### Scope
- [ ] Add a dropdown menu to display additional menu options when all will not fit on the screen width

The implementation should be similar to that used in `LearningActivityBar`.

### Acceptance Criteria
- [ ] Layout should be pixel-perfect match to Figma specifications 
- [ ] Main menu should continue to be tab-navigable and tab order should be persisted
- [ ] Dropdown menu should also be in the tab navigation flow. It should open when in focus on `Enter` and close with `Esc`.
- [ ] Update any tests in `app-bar.spec.js` as needed based on your changes

### LINKED ISSUE #9277: Turn points icon into an button on small breakpoints
This change improves the accessibility and functionality of the points display, by using a button rather than a tooltip.

### Scope
- [ ]  On small screens, as defined by `responsiveWindowMixin`, points should not be displayed
- [ ]  The existing tooltip feature for this use case should be removed, replaced with a button and popover. The string can be reused.
- [ ]  Username truncation should remain in its current version

#### Out of scope
- Adding the new icon to KDS to be available for KIconButton. Instead, use any placeholder icon.

### Acceptance Criteria
- [ ]  Button should be keyboard navigable: accessed through tab order. When in focus, it should open the popover on `Enter` and close it on `Esc`
- [ ]  Tests should be added to `app-bar.spec.js` for conditions of when points are displayed, when the icon is a button, and the opening of the popover.

marcellamaki:
## Summary
This PR continues updates to the AppBar, making changes to the display of points and managing the overflow of navigation tabs in a new dropdown menu, rather than a horizontal overflow with scroll. 

## References
Fixes #9276
Fixes #9277 

### Points display 
➡️ Note: out of scope was updating the icon, as I believe the icon design wasn't finalized at the time of dev handoff. 

| Spec | 
|---|
| image | 
| Implementation  |
| points-popover |

@radinamatic I did some preliminary testing using keyboard navigation, and also using Mac voiceover, and based on my very basic understanding, it seems _functional_  😄 but I know things are not always that straight forward. I added an aria label to the icon button, but if there are additional changes that can improve any parts of the accessibility experience, please do let me know. One other consideration I had thought about was - would there be any pros/cons to not having the popover for screenreader users, and rather directly having the aria label just read aloud the points information? I'm not sure if there is an easy way to do this or if there is WCAG guidance on these sorts of things.

### Overflow display

The overflow display creates a new component which is responsible for checking the viewport width, and managing overflow by itself. It receives an array of navigation links. I think there is potential for further refactoring here, potentially to replace the intermediary `[Plugin]TopNav` entirely in a lot of places, but, I wanted to think/talk that through before jumping in. Would be happy to open and take on a follow up issue if that seems worthwhile to folks at this point in time.

| Spec |
|---|
| image | 
| Implementation  |
| en-overflow |
| RTL |
| ar-overflow |

### Points popover 
1. Log in as a learner
2. Ensure you have completed some activities, so you have points
3. Click on the placeholder icon - the popover with the correct number of points should display
4. You should also be able to navigate to this by keyboard 

### Overflow 
1. Logged in as a learner OR as a coach: try out the display on different viewport sizes. While I imagine dynamic resizing the browser may be the less common reason this might pop up, that is a place to start. On hardware devices, or on the chrome dev tools that allow you to test on the viewport widths of mobile devices would be another. 
2. Navigation links should all work, either as a tab, or in the overflow, and direct the user to the correct page on navigation
3. Works in RTL and LTR languages
4. A11y specific: proper keyboard navigation, proper labeling/navigation for screen readers. @radinamatic much like above - I did do testing with voiceover and ensured there were aria labels on the buttons and 🤷‍♀️ I'm not sure if it is a *good* experience, or if it works across devices. Any suggestions for improvements welcome and appreciated 
----

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated


radinamatic:
Tested with the DEB asset on Ubuntu 18.04, browsing as client from Firefox and Chrome on Windows 10 with NVDA.

Keyboard navigation and the screen reader experience is actually not bad: everything is accessible and announced, and all the users have similar experience - points display when the button is pressed, no mater which navigation modality is used. 

points


> pros/cons to not having the popover for screenreader users, and rather directly having the aria label just read aloud the points information

A part from the fact that actively trying to detect screen reader users is discouraged (and I'm not even sure there's a reliable way to do it), I don't think it's warranted here, since all users have the same experience (_press-the-button-to-see-points_).

The only issue I see here is that the popover remains open even when user navigates away: they have to actively press again for button to close, not even mouse click anywhere else does it 😛 

**Bigger concern in this new implementation is the insufficient contrast** for the text in the subpage menu links: too low even for the currently active page, and way, way low for the rest 😞 
cc @jtamiace 

Ubuntu18 (start)  Running  - Oracle VM VirtualBox_010

@pcenov Could you test the non-accessibility pieces (mobile views & RTL) that @marcellamaki listed in the reviewer guidance? Thank you!

pcenov:
I was able to identify just the following couple of issues (valid in all supported browsers):

| Issue| Screenshot |
| Coach Plugin - The top navigation is not working except when a link is clicked through the 'More options'. There's a "TypeError: An id must be specified" error in the console | https://user-images.githubusercontent.com/79847249/185111668-5a928721-7e9c-4961-9192-53fdf527878d.mp4 |
| Mobile view - The tooltip 'More options' is covering the text of the link (valid for RTL view too) |  |

marcellamaki:
Thank you @pcenov and @radinamatic - will follow up on these issues and update! 😊

marcellamaki:
@radinamatic @pcenov - I believe the contrast, coach navigation, popover "navigate away", and tooltip overlap should now all be addressed. 

(Please note that in this PR, for the purposes of testing there is a temporary reference to an in-progress KDS work that supports this. If this is merged before that, the tooltip will still be overlapping until the KDS update is also merged)

pcenov:
@marcellamaki Looking great! Both the tooltip's overlapping of the menu items and coach navigation issues are now fixed. I just have a couple of questions:

| Issue| Screenshot |
| Facility > Classes > Class - Both tables in Enroll learners and Assign coaches are now stretching endlessly - is this a work in progress or an unintentional side effect? |  |
| Mobile view - Is the tooltip actually necessary in mobile as it is shown only after I've already tapped the button and seen what it does. Perhaps it should be removed altogether? |  |

radinamatic:
Re-tested the newest asset, contrast is now within the guidelines! 👏🏽 

I can replicate the finding by @pcenov about the **Enroll learners** table occupying the full width of the page. Is that the new design because the table now has 5 columns instead of 2?

I also second the observation that the tooltip does not seem necessary on touch interfaces. Anyways, it's not something that needs to be fixed in this PR, so going to approve 👍🏽

radinamatic (approved):
Manual and a11y testing passed, good to go! 💯 :shipit:

kolibri/core/assets/src/views/AppBar.vue
```diff
@@ -137,6 +161,25 @@
...
+            this.$refs.pointsButton.$el &&
+            this.$refs.pointsButton.$el.contains(event.target)
+          ) {
+            this.pointsDisplayed = !this.pointsDisplayed;
+          }
+        }
+        return event;
+      },
+      handlePopoverByKeyboard(event) {
+        if ((event.key == 'Tab' || event.keyCode == 27) && this.pointsDisplayed) {
```
MisRob: Was there any reason for using `key` to check for `'Tab'` but `keyCode` for `'Escape'`? [`keyCode` is not recommended ](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/keyCode) so if not, I'd suggest using `key` only.

kolibri/core/assets/src/views/AppBar.vue
```diff
@@ -239,4 +305,15 @@
...
   }
 
+  .points-popover {
+    @extend %dropshadow-4dp;
+
+    position: absolute;
+    right: 50px;
+    z-index: 24;
+    font-size: 12px;
+    background-color: white;
```
MisRob: There's the surface token for such situations https://design-system.learningequality.org/colors/#tokens-surface

kolibri/core/assets/src/views/HorizontalNavBarWithOverflowMenu.vue
```diff
@@ -0,0 +1,151 @@
...
+    },
+    computed: {
+      color() {
+        return this.$themeTokens.textInverted;
+      },
+    },
+    mounted() {
+      this.updateNavigationTabDisplay();
+      window.addEventListener('resize', this.updateNavigationTabDisplay);
+      // document.addEventListener('resize', console.log('hello world'));
```
MisRob: Hello Marcella :)

hello

kolibri/core/assets/src/views/HorizontalNavBarWithOverflowMenu.vue
```diff
@@ -0,0 +1,151 @@
...
+    },
+    beforeDestroy() {
+      document.removeEventListener('resize', this.debouncedUpdateNavigation);
+    },
+    methods: {
+      debouncedUpdateNavigation() {
+        return debounce(this.updateNavigationTabDisplay, 1000);
+      },
+      handleSelect(option) {
+        this.$router.push(this.$router.getRoute(option.value.name));
```
MisRob, line 97: I think it's fine to assume that links passed to this component will have `name`, however it may help to add a validator function to `navigationLinks` prop that would check this because we sometimes use routes that have no `name` associated with it.

kolibri/core/assets/src/views/Navbar/NavbarLink.vue
```diff
@@ -28,14 +22,13 @@
 <script>
 
   import { validateLinkObject } from 'kolibri.utils.validators';
-  import UiIcon from 'kolibri-design-system/lib/keen/UiIcon';
+  // import UiIcon from 'kolibri-design-system/lib/keen/UiIcon';
```
MisRob: Can be removed

kolibri/core/assets/src/views/Navbar/NavbarLink.vue
```diff
@@ -131,10 +135,20 @@
...
     display: inline-block;
-    font-weight: bold;
     text-overflow: ellipsis;
-    text-transform: uppercase;
-    vertical-align: middle;
+  }
+
+  // for small screens
+
+  @media (max-width: 600px) {
```
MisRob: I'd suggest using rather pre-defined breakpoints via `responsiveWindowMixin` so that we have consistency in breakpoints. Also, using media queries, in general, is listed as anti-pattern [in our dev docs](https://kolibri-dev.readthedocs.io/en/develop/frontend_architecture/conventions.html#styling-anti-patterns). I'm not sure if there were some more reasons as I wasn't there when the decision was made, just can remember that we usually try to avoid it.

kolibri/core/assets/src/views/Navbar/index.vue
```diff
@@ -114,33 +30,20 @@
...
   }
 
-  .scroll-button {
-    width: 36px;
-    height: 36px;
-    vertical-align: middle;
-    background-color: inherit;
-    border-style: none;
-    appearance: none;
+  @media (max-width: 840px) {
```
MisRob: This is another place where `responsiveWindowMixin` could be used

kolibri/plugins/device/assets/src/views/DeviceTopNav.vue
```diff
@@ -97,4 +84,25 @@
...
+<style lang="scss" scoped>
+
+  @import '~kolibri-design-system/lib/styles/definitions';
+
+  .menu-icon {
+    position: absolute;
+    right: 4px;
+  }
+
+  .menu-popover {
```
MisRob: Both `menu-popover` and `menu-icon` look to be identical to those defined in _HorizontalNavBarWithOverflowMenu.vue_, I assume there's no need to have them defined twice?

And regarding `menu-popover`, I wonder if it could completely removed from both components since I haven't noticed a place in either Kolibri or KDS where this class would be actually applied, even though I only checked code. Is it really needed?

kolibri/plugins/device/assets/src/views/DeviceTopNav.vue
```diff
@@ -1,83 +1,70 @@
...
+  import responsiveWindowMixin from 'kolibri.coreVue.mixins.responsiveWindowMixin';
 
   export default {
     name: 'DeviceTopNav',
     components: {
-      Navbar,
-      NavbarLink,
+      HorizontalNavBarWithOverflowMenu,
+    },
+    mixins: [commonCoreStrings, responsiveWindowMixin],
```
MisRob: I don't see `responsiveWindowMixin` being used from this component

kolibri/plugins/learn/assets/test/views/learn-top-nav.spec.js
```diff
@@ -1,136 +0,0 @@
...
-  };
-  const setCanAccessUnassignedContent = canAccess => {
-    store.state.canAccessUnassignedContentSetting = canAccess;
-  };
-
-  beforeEach(() => {
-    store = makeStore();
-  });
-
-  it('there are no tabs if showing content unavailable page', () => {
```
MisRob, line 82: What was the reason for removing this whole test suite? I assume that these scenarios are not obsolete, is that right?

kolibri/core/assets/test/horizontal-nav-bar-with-overflow.spec.js
```diff
@@ -0,0 +1,116 @@
+import HorizontalNavBarWithOverflowMenu from 'kolibri.coreVue.components.HorizontalNavBarWithOverflowMenu';
+import { shallowMount, mount } from '@vue/test-utils';
+
+function makeWrapper({ propsData } = {}) {
+  return mount(HorizontalNavBarWithOverflowMenu, { propsData });
+}
+
+const longerNavigationList = [
+  {
+    condition: true,
```
MisRob: Perhaps, `isVisible` instead of `condition` may help with understanding what's this attribute good for?

kolibri/core/assets/src/views/HorizontalNavBarWithOverflowMenu.vue
```diff
@@ -0,0 +1,151 @@
...
+          this.numberOfNavigationTabsToDisplay,
+          this.navigationLinks.length
+        );
+        let options = [];
+        limitedList.forEach(o => options.push({ label: o.title, value: o.link, icon: o.icon }));
+        this.overflowMenuLinks = options;
+      },
+      updateNavigationTabDisplay() {
+        // turn HTML collection into an array
+        const navItems = [].slice.call(document.getElementsByClassName('list-item-navigation'));
```
MisRob: Accessing `NavbarLinks` via classes in this way creates a hidden dependency that makes things quite brittle, e.g. if someone changes the class name in `NavbarLink`, this whole logic would silently break. I think that people usually don't get an idea to search whether a class from a component is used in a parent of its parent (`HorizontalNavBarWithOverflowMenu` -> `Navbar` -> `NavbarLink`). Perhaps `ref`s could be utilized? I'm not sure if that's the best idea, or if it'd be better to consider a different architecture, we can think about it more together if it'd be helpful. Any other ideas?

MisRob (changes requested):
Thank you, @marcellamaki, overall this is great work. I didn't test it as I can see that the QA team has already tested it. I posted some clarifying questions and several cleanup comments. I'd welcome it if we could try to resolve the issue with the hidden dependency before merging. If we don't find a better way, then it'd be good to make sure to at least comment that the problematic class is being used from outside of `NavbarLink`.

kolibri/core/assets/src/views/AppBar.vue
marcellamaki: No, an oversight. Will update!

kolibri/core/assets/src/views/HorizontalNavBarWithOverflowMenu.vue
marcellamaki: Good suggestion, I'll add this.

kolibri/core/assets/src/views/Navbar/NavbarLink.vue
marcellamaki: I have no idea why I did this..... 🤔 I will update to the mixin, unless I come across a reason why I did it this way, in which case I will follow up so we can discuss further

kolibri/core/assets/src/views/HorizontalNavBarWithOverflowMenu.vue
marcellamaki: `refs` is a good idea -- I will try implementing this way, and if I run into a problem today, I'll comment more here or in slack. And thank you for the suggestion and feedback, this is a good point. Even when I was working on it, it seemed a bit... hacky, but I didn't revisit a better approach after "getting it to work". But I should 😄

marcellamaki: This was a very straightforward switch, and I think it's improved. Let me know what you think of the adjustments I made @MisRob and if you have any further ideas to make it better. Thanks again for the great suggestion. Someday, using refs will be something I just think of on my own... but we're not there yet :D

kolibri/plugins/learn/assets/test/views/learn-top-nav.spec.js
marcellamaki: So, my thinking was that the management of these tabs has been fully moved over into the new component. I think there is even potential to remove the "Top Nav" component pattern entirely and just use the new bar which is managing the display ` <HorizontalNavBarWithOverflowMenu />`
It seemed like that would be better addressed as a follow up issue, since it would require some refactoring and making sure there are no regressions across many plugins.
I will double check the deleted file and see if there is any test that I have lost in this process though, that should still be present.

MisRob: Ah right, yeah that makes sense, so I think it's fine, thank you

kolibri/core/assets/src/views/HorizontalNavBarWithOverflowMenu.vue
MisRob: Thank you, it's looking good to me, and I have no further ideas

MisRob (approved):
Thanks, as soon as the remaining media query in _Navbar/index.vue_ is resolved, I think this is ready to be merged. If you found out that the media query needs to stay there, could you just please commit comment on why this was the case?

MisRob:
Oh, I realized we shouldn't actually merge this until we can update package.json, so just posting a note here so that we don't forget

marcellamaki:
> Navbar/index.vue is resolved

Done! It was an oversight to not include with the other commits 🙃

MisRob:
I updated the PR with the following
- Since we merged the corresponding KDS PR https://github.com/learningequality/kolibri-design-system/pull/346/, I removed reference to Marcella's fork repo in favour of referencing the KDS repo in _package.json_
- Resolved conflicts with the latest `develop` as of Sep 8 and did some related fixes of tests, see commit messages for more details
- Resolved `Property or method "disabled" is not defined on the instance but referenced during render.` when the dropdown menu is rendered on smaller screens

MisRob (dismissed):
When working on the above, I noticed that not all places using `KDropdown` are updated to its new API. At some places, we're still using `text`, `disabled`, and `appearance` props that have been removed from the corresponding KDS PR in https://github.com/learningequality/kolibri-design-system/pull/346, for example

https://github.com/learningequality/kolibri/blob/90ec6518156a2648439b9592dd8b0fb3c5d5bff2/kolibri/plugins/facility/assets/src/views/UserPage/index.vue#L57-L63

We need to make sure to update `KDropdown` usage in the whole Kolibri before merging. In the example above, we need to wrap `KDropdown` component in a KDS button component and transfer `text`, `disabled`, and `appearance` properties there. There are more places like this.

 I will see if I can get continue working on this next week, already know that I won't have time to continue this week, so blocking now.

MisRob:
I pushed one more commit that updated our usage of `KDropdownMenu`, which resolved regressions at places we use dropdown menus. @radinamatic, if you could test all dropdown menus you can think of, that'd be helpful. These are places I located based on code:

|  |  |  |  |
| - | - | - | - |
| dropdown-1 | dropdown-2 | dropdown-3 | dropdown-4 |
| dropdown-5 | dropdown-6 | dropdown-7 | dropdown-8 |

Also, there's one more issue resulting in the console warning:



but this needs to be addressed in KDS. I will do it tomorrow.

MisRob:
I dismissed my review - it'd be good if someone else could review the last commit since I am the author. After that change is approved and @radinamatic confirms that this is good to go, should be ready for merge.

MisRob:
A couple of updates in KDS inspired by the latest work I made in this branch are here https://github.com/learningequality/kolibri-design-system/pull/361/. It also fixes 'Property or method "disabled" is not defined on the instance but referenced during render.'

It would be good to update package.json again here to point to the latest KDS develop as soon as https://github.com/learningequality/kolibri-design-system/pull/361/ is reviewed and merged.

MisRob:
@marcellamaki I planned to wrap this up much sooner, but now when you're back, would you like to review the latest changes and finish this? Should be just a matter of updating the package.json again as mentioned one comment above unless review and testing reveal some new issues.

marcellamaki:
Yes, will do @MisRob. Thank you for working on this!

pcenov:
@MisRob I wasn't able to identify any issues caused by your commit for the dropdowns - all is looking just fine!
@marcellamaki after a full regression test of the latest build I found the following issues some of those could be caused by changes made here others are probably caused by something else. Could you please have a look and let me know which ones should be filed as separate bug reports?

| Issue| Screenshot |
| Library > Recent - The ‘View as list’ view is broken in all supported browsers | |
| Bookmarks - Same as above | |
| Learn > Home > Continue learning on your own - Misaligned cards |  |
| Library > Channel - Misaligned channel title and x icon and Uncaught error in the console | |
| Library > Channel with Arabic - Padding issues |   |
| Learn > Resource - Misaligned icon |  |

marcellamaki:
Hi @pcenov - thank you for doing such comprehensive testing! My best guess is the header issues are related/regressions, and the card issues (spacing, view as list) are not, but I will take a look at this before the end of the week to be sure. Anything I can update within the scope of this I will, and then I will clarify which should be follow up issues :)

MisRob:
I merged develop to resolve some conflicts so some new bugs could come from there too. @marcellamaki if you're fine with the latest changes I pushed, and after we install the latest KDS `develop` here again, then I think it'd be fine to merge this before we gather new conflicts with KDS versions (there are some other PRs that need to install it, I believe, but currently can't as it'd break dropdowns) and resolve related regressions/open new issues as part of follow-up work.

pcenov:
@marcellamaki I decided to compare the latest issues reported by me against the latest develop build and can confirm that all of them are present there so I guess I should just file them as separate bug reports?

I also found the following issue - The 'Home' and 'Bookmarks' tabs are being displayed while I'm in the 'Explore without account' flow while they shouldn't. This is something that seems caused by this PR as it's not present on develop:



Hope this helps! :)

marcellamaki:
Thank you @pcenov! I believe most of these, since they are present on `develop` are related to refactoring. I've created [a new issue from your original comment](https://github.com/learningequality/kolibri/issues/9726), and will assign this to myself as follow up. So, no need to make lots of separate issues for this :) 

I will address the tabs though, which you mention, in this PR - I'll try to fix today! Thank you again!

rtibbles (changes requested):
Holding merge until `links` is handled consistently across components, and `KDropDown` bug is re-fixed.

rtibbles (approved):
Final issues resolved.

PR #9642, opened 2022-08-19
update version string to 15.7 beta 1

marcellamaki:


kolibri/__init__.py
```diff
@@ -15,7 +15,7 @@
 
 #: This may not be the exact version as it's subject to modification with
 #: get_version() - use ``kolibri.__version__`` for the exact version string.
-VERSION = (0, 15, 6, "final", 0)
+VERSION = (0, 15, 7, "beta", 1)
```
rtibbles: Only need to update to `"alpha", 0` everything else (until the final release) is inferred from git tags.

marcellamaki: 🤦‍♀️ clearly a sign that I am not doing this frequently enough

rtibbles (approved):

PR #9670, opened 2022-08-29
Added screen to confirm merge

### LINKED ISSUE #9331: Create "Merge Accounts" confirm merge page
# Overview 

Adds a confirmation page to flows when a user is merging existing accounts, explaining the implications of the merge. 

## Figma 
| Not selected  |  Selected |
|---|---|
|  |  | 

### Accessibility notes
Tab order: 
1) Checkbox 
2) "Continue" primary button 
3) "Back" secondary button 
4) "X" close out flow 

### Out of scope 

### Acceptance criteria

- [ ] Continue button disabled until checkbox is checked

jredrejo:
## Summary

confirmMerge

## References
Closes: #9331 

## Reviewer guidance
Following the "merge" path in the change facility setup this screen should appear almost at the end.

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/plugins/user_profile/assets/src/machines/changeFacilityMachine.js
+++ b/kolibri/plugins/user_profile/assets/src/machines/changeFacilityMachine.js
@@ -206,7 +206,15 @@ export const changeFacilityMachine = createMachine({
       meta: { route: 'CONFIRM_MERGE', path: '/change_facility' },
       on: {
         CONTINUE: 'syncChangeFacility',
-        BACK: 'confirmAccountDetails',
+        BACK: [
+          {
+            cond: context => context.role === 'superuser',
+            target: 'chooseAdmin',
+          },
+          {
+            target: 'confirmAccountDetails',
+          },
+        ],
       },
     },
     syncChangeFacility: {
--- a/kolibri/plugins/user_profile/assets/src/machines/changeFacilityMachine.js
+++ b/kolibri/plugins/user_profile/assets/src/machines/changeFacilityMachine.js
@@ -206,7 +206,7 @@ export const changeFacilityMachine = createMachine({
       meta: { route: 'CONFIRM_MERGE', path: '/change_facility' },
       on: {
         CONTINUE: 'syncChangeFacility',
-        BACK: 'changeFacility',
+        BACK: 'confirmAccountDetails',
       },
     },
     syncChangeFacility: {
--- a/kolibri/plugins/user_profile/assets/src/views/ChangeFacility/ConfirmMerge.vue
+++ b/kolibri/plugins/user_profile/assets/src/views/ChangeFacility/ConfirmMerge.vue
@@ -61,10 +61,11 @@
       const isConfirmed = ref(false);
 
       function handleContinue() {
-        if (isConfirmed.value)
+        if (isConfirmed.value) {
           changeFacilityService.send({
             type: 'CONTINUE',
           });
+        }
       }
       function sendBack() {
         changeFacilityService.send({
--- a/kolibri/plugins/user_profile/assets/src/views/ChangeFacility/ConfirmMerge.vue
+++ b/kolibri/plugins/user_profile/assets/src/views/ChangeFacility/ConfirmMerge.vue
@@ -1,0 +1,107 @@
+<template>
+
+  <div>
+    <h1>{{ $tr('documentTitle') }}</h1>
+    <p>{{ $tr('confirmMergeInfoLine') }}</p>
+    <div class="confirm">
+      <KCheckbox
+        ref="kCheckbox"
+        :label="$tr('consequences')"
+        :checked="false"
+        @change="isConfirmed = !isConfirmed"
+        @keydown.enter="handleContinue"
+      />
+    </div>
+
+    <BottomAppBar>
+      <slot name="buttons">
+        <KButtonGroup>
+          <KButton
+            :primary="false"
+            :text="coreString('backAction')"
+            appearance="flat-button"
+            data-test="backButton"
+            @click="sendBack"
+          />
+          <KButton
+            :primary="true"
+            :disabled="!isConfirmed"
+            :text="coreString('continueAction')"
+            data-test="continueButton"
+            @click="handleContinue"
+          />
+        </KButtonGroup>
+      </slot>
+    </BottomAppBar>
+
+  </div>
+
+</template>
+
+
+<script>
+
+  import commonCoreStrings from 'kolibri.coreVue.mixins.commonCoreStrings';
+  import BottomAppBar from 'kolibri.coreVue.components.BottomAppBar';
+  import { inject, ref } from 'kolibri.lib.vueCompositionApi';
+
+  export default {
+    name: 'ConfirmMerge',
+    metaInfo() {
+      return {
+        title: this.$tr('documentTitle'),
+      };
+    },
+    components: { BottomAppBar },
+
+    mixins: [commonCoreStrings],
+
+    setup() {
+      const changeFacilityService = inject('changeFacilityService');
+      const isConfirmed = ref(false);
+
+      function handleContinue() {
+        if (isConfirmed.value)
+          changeFacilityService.send({
+            type: 'CONTINUE',
+          });
+      }
+      function sendBack() {
+        changeFacilityService.send({
+          type: 'BACK',
+        });
+      }
+      return {
+        isConfirmed,
+        sendBack,
+        handleContinue,
+      };
+    },
+
+    $trs: {
+      documentTitle: {
+        message: 'Merge accounts',
+        context: 'Title of this step for the change facility page.',
+      },
+      confirmMergeInfoLine: {
+        message:
+          'You are about to merge all progress data from two different accounts. Progress data includes your interactions with resources, time spent, and points. This cannot be undone.',
+        context: 'Text explaining the consequences merging will have.',
+      },
+      consequences: {
+        message: 'I understand the consequences of merging accounts',
+        context: 'Button to confirm the acceptance of merging.',
+      },
+    },
+  };
+
+</script>
+
+
+<style lang="scss" scoped>
+
+  .confirm {
+    margin-top: 40px;
+  }
+
+</style>
--- a/kolibri/plugins/user_profile/assets/test/views/ChangeFacility/ConfirmMerge.spec.js
+++ b/kolibri/plugins/user_profile/assets/test/views/ChangeFacility/ConfirmMerge.spec.js
@@ -38,6 +38,16 @@ describe(`ChangeFacility/ConfirmMerge`, () => {
     checkbox.trigger('click');
     await wrapper.vm.$nextTick();
     expect(continueButton.vm.disabled).toBeFalsy();
+  });
+
+  it(`clicking continue button sends the continue event to the state machine`, async () => {
+    const wrapper = makeWrapper();
+    clickContinueButton(wrapper);
+    expect(sendMachineEvent).not.toHaveBeenCalled();
+
+    const checkbox = wrapper.find('input[class="k-checkbox-input"]');
+    checkbox.trigger('click');
+    await wrapper.vm.$nextTick();
     clickContinueButton(wrapper);
     expect(sendMachineEvent).toHaveBeenCalledWith({
       type: 'CONTINUE',
--- a/kolibri/plugins/user_profile/assets/test/views/ChangeFacility/ConfirmMerge.spec.js
+++ b/kolibri/plugins/user_profile/assets/test/views/ChangeFacility/ConfirmMerge.spec.js
@@ -1,0 +1,54 @@
+import { mount, createLocalVue } from '@vue/test-utils';
+import ConfirmMerge from '../../../src/views/ChangeFacility/ConfirmMerge';
+
+const localVue = createLocalVue();
+const sendMachineEvent = jest.fn();
+
+function makeWrapper() {
+  return mount(ConfirmMerge, {
+    provide: {
+      changeFacilityService: {
+        send: sendMachineEvent,
+      },
+    },
+    localVue,
+  });
+}
+
+const getBackButton = wrapper => wrapper.find('[data-test="backButton"]');
+const getContinueButton = wrapper => wrapper.find('[data-test="continueButton"]');
+const clickBackButton = wrapper => getBackButton(wrapper).trigger('click');
+const clickContinueButton = wrapper => getContinueButton(wrapper).trigger('click');
+
+describe(`ChangeFacility/ConfirmMerge`, () => {
+  beforeEach(() => {
+    jest.clearAllMocks();
+  });
+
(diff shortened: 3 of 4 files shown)


kolibri/plugins/user_profile/assets/src/views/ChangeFacility/ConfirmMerge.vue
```diff
@@ -0,0 +1,107 @@
...
+
+    mixins: [commonCoreStrings],
+
+    setup() {
+      const changeFacilityService = inject('changeFacilityService');
+      const isConfirmed = ref(false);
+
+      function handleContinue() {
+        if (isConfirmed.value)
+          changeFacilityService.send({
```
MisRob, line 65: Regarding missing curly brackets, I know we discussed this together before and that it's valid JS code though it still seems to be a bit confusing in regards to consistency because we very rarely use this in the whole codebase. It's not blocking but wanted to mention it because we were reviewing with @akolson and it made us stop by here. Perhaps we could discuss code conventions with the team?

kolibri/plugins/user_profile/assets/test/views/ChangeFacility/ConfirmMerge.spec.js
```diff
@@ -0,0 +1,54 @@
...
+  it('continue is disabled until merging is accepted', async () => {
+    const wrapper = makeWrapper();
+    const checkbox = wrapper.find('input[class="k-checkbox-input"]');
+    const continueButton = wrapper.find('[data-test="continueButton"]');
+    expect(continueButton.vm.disabled).toBeTruthy();
+    checkbox.trigger('click');
+    await wrapper.vm.$nextTick();
+    expect(continueButton.vm.disabled).toBeFalsy();
+    clickContinueButton(wrapper);
+    expect(sendMachineEvent).toHaveBeenCalledWith({
```
MisRob, line 52: Nothing blocking, but this part of test actually doesn't test anymore that _'continue is disabled until merging is accepted'_, it rather tests that when the continue button is already enabled, the event will be sent. That's why I think it'd make sense to put this into a separate test case similar to the last test below.

kolibri/plugins/user_profile/assets/src/machines/changeFacilityMachine.js
```diff
@@ -209,7 +209,7 @@ export const changeFacilityMachine = createMachine({
       meta: { route: 'CONFIRM_MERGE', path: '/change_facility' },
       on: {
         CONTINUE: 'syncChangeFacility',
-        BACK: 'changeFacility',
+        BACK: 'confirmAccountDetails',
```
MisRob: We don't always want to navigate to 'confirmAccountDetails', in some cases we need to navigate to the choose a new super admin as illustrated on designs:
Screenshot from 2022-09-06 15-24-04

We already have something similar implemented [here](https://github.com/learningequality/kolibri/blob/develop/kolibri/plugins/user_profile/assets/src/machines/changeFacilityMachine.js#L223-L232)

I think that this should be resolved before merging.

MisRob (changes requested):
Thank you, @jredrejo, code is looking great overall and I appreciate tests that describe the component's behaviour very well. The only blocking thing is the comment about back navigation condition.

Also, I have one question about keyboard vs. mouse navigation:
(1) When using mouse, when I check on the checkbox and then press enter on keyboard, it takes us immediately to another step, whereas (2) when using keyboard, when I check the checkbox with space and then press enter on keyboard, nothing happens and I need to tab to "Continue" button and press enter again to get to the next state. Was this discrepancy intentional? I am actually not sure whether behavior (1) or (2) is preferred.

kolibri/plugins/user_profile/assets/test/views/ChangeFacility/ConfirmMerge.spec.js
jredrejo: yeah, maybe I thought the same but at the end didn't do it because I thought it was going to be too much repeated code. Happy to separate it in a different test.

kolibri/plugins/user_profile/assets/src/views/ChangeFacility/ConfirmMerge.vue
jredrejo: sure, I'm all ears, I will change it if people think it's better (I am so used to it since so many years that I see it as readable or more than using curly brackets) Anwyay, if we decide to use that convention I think we should add a linting rule to detect it. I'm pretty sure you'll find it everywhere in any of the libs we use, as this is a common use in all the languages with syntax derived from C.

> Also, I have one question about keyboard vs. mouse navigation:
> (1) When using mouse, when I check on the checkbox and then press enter on keyboard, it takes us immediately to another step, whereas (2) when using keyboard, when I check the checkbox with space and then press enter on keyboard, nothing happens and I need to tab to "Continue" button and press enter again to get to the next state. Was this discrepancy intentional? I am actually not sure whether behavior (1) or (2) is preferred.

No, it was not intentional, actually I didn't even know it was happening. I prefer (1) because it seems more agile, but I'm not sure that can be achieved just using the keyboard as the events will overlap. Any suggestion?

rtibbles: I must admit, I am also not a huge fan of the single line of code following an `if` statement - and also prefer requiring an expression block.

Seems like we can enforce this via eslint's `curly` rule, with the `all` setting: https://eslint.org/docs/latest/rules/curly

kolibri/plugins/user_profile/assets/src/machines/changeFacilityMachine.js
jredrejo: `confirmAccountDetails` takes you to `isAdmin` that determines if the Choose admin screen must be used:  
```
    isAdmin: {
      always: [
        {
          cond: context => context.role === 'superuser',
          target: 'chooseAdmin',
        },
        {
          target: 'checkIsMerging',
        },
      ],
    },

....

confirmAccountDetails: {
      meta: { route: 'CONFIRM_DETAILS', path: '/change_facility' },
      on: {
        CONTINUE: 'isAdmin',
        BACK: 'mergeAccounts',
      },
    },
```

I see the case where you're already in `chooseAdmin` , then go to `confirmMerge` and then back to a screen previous to `chooseAdmin` but I wonder if it's worth to add more conditions just for this case that will be rare and will take you to the same path anyway. If you think it's worth I'll add it but with the current complexity of the machine I'll rather leave it as it is because I don't see it as a problem.

MisRob:
@MisRob:
> Also, I have one question about keyboard vs. mouse navigation:
(1) When using mouse, when I check on the checkbox and then press enter on keyboard, it takes us immediately to another step, whereas (2) when using keyboard, when I check the checkbox with space and then press enter on keyboard, nothing happens and I need to tab to "Continue" button and press enter again to get to the next state. Was this discrepancy intentional? I am actually not sure whether behavior (1) or (2) is preferred.

@jredrejo:
> No, it was not intentional, actually I didn't even know it was happening. I prefer (1) because it seems more agile, but I'm not sure that can be achieved just using the keyboard as the events will overlap. Any suggestion?

I would guess that users would expect (2) rather than being taken to the next screen automatically after the checkbox is checked without using "Continue" button, but not certain. I think it'd be best to ask @radina what's desired behavior.

kolibri/plugins/user_profile/assets/test/views/ChangeFacility/ConfirmMerge.spec.js
MisRob: One way to deal with repetition is to group related tests under `describe` and use `beforeEach/beforeaAll` in that `describe` block that would set all tests up. But I don't mind if the test setup is repeated two times or so.

kolibri/plugins/user_profile/assets/src/machines/changeFacilityMachine.js
MisRob: > with the current complexity of the machine I'll rather leave it as it is because I don't see it as a problem

Yes, you're right that those conditions make things pretty complex though I don't think that ignoring proper order of back navigation in some cases because of that complexity is a good idea and if so, I'd suggest discussing this with Jessica because on the designs it seems that "Back" button is meant to take you to the correct previous step in all flows? 

I can understand your concerns about adding more conditions, though, as since this problem is not related only to this particular PR, I'd suggest we leave it this way here but make sure to resolve that complexity on higher-level as a follow-up issue before we accumulate tech debt (with that I'm not suggesting you need to be the one to resolve that issue). For similar reasons, I was already pondering a bit about whether it'd make sense to introduce using nested states. Even from designs, it seems to be pretty intuitive to go in that direction. I imagine we could have three root states corresponding to the three main branches of designs (default path, username exists, merge accounts). Those would then contain sub-states representing steps of that particular path. It shouldn't be anything difficult to implement, we could get rid of the majority of conditions that make the machine complicated and I believe it'd make it much better readable. What do you think of this direction?

jredrejo: Honestly, I don't like nested states too much. I only see them useful when there really 2 states to consider at the same time (as one element that's visible or not, and, if, visible, it's in one color or another).
I don't see they add too much value or reduce code or simplicity in other cases.

radinamatic:
Noting that I did not actually test this in practice, so I'm operating under a few assumptions:

> (1) When using mouse, when I check on the checkbox and then press enter on keyboard, it takes us immediately to another step

This _smells_ as a situation where after the checkbox is _checked_, the focus is placed on the (I presume) ``Continue`` button, and pressing the Enter key makes the form proceed to the next page. If that is indeed the case I could say _Ok_ if it weren't the case of mixed navigation modality (mouse and keyboard), and those are not the defaults we usually operate under. If we make sure that the ``Continue`` button, **_if it is indeed in focus_**, also has the proper visual focus indicator, then this could work Ok.

> (2) when using keyboard, when I check the checkbox with space and then press enter on keyboard, nothing happens and I need to tab to "Continue" button and press enter again to get to the next state. 

If the first case above is true, and focus is placed on the ``Continue`` button, this does not seem to be happening when the checkbox is activated by keyboard, instead of the mouse... 🤔 

Actually, now that I think of it, the crux of the question is whether (I have not tested in practice) we are indeed talking about checkboxes, or we are talking **radio buttons**? Since checkboxes allow for multiple selections, it would be premature to send the user to the next state immediately after selecting the first one whichever modality they are using, because they may need/want to check another. On the other hand, if we are talking radio buttons, since they only admit one selected, it would not disrupt the flow if there is a more _agile_ transition to the next state.

kolibri/plugins/user_profile/assets/src/machines/changeFacilityMachine.js
MisRob: I think it may help with some conditions, but it doesn't need to necessarily be nested states, I just don't have a better idea right now. Perhaps a little experiment with machine architecture or researching how others solve conditional branches would help.

For now, I'd suggest at least opening an issue where it will be written down that back navigation order is not working at some places (if you could include which places exactly would be helpful) since the machine's complexity is complicating it. We can discuss the solution and its priority there. Perhaps it will show up to be a low priority, but having an issue will help communicate that it was an intentional choice so other folks won't need to be figuring out whether it's a regression in the future, and also, it will be good to have it logged if we encounter some other issues with growing machine's complexity over time.

jredrejo:
> Noting that I did not actually test this in practice, so I'm operating under a few assumptions:
> 
> > (1) When using mouse, when I check on the checkbox and then press enter on keyboard, it takes us immediately to another step
> 
> This _smells_ as a situation where after the checkbox is _checked_, the focus is placed on the (I presume) `Continue` button, and pressing the Enter key makes the form proceed to the next page. If that is indeed the case I could say _Ok_ if it weren't the case of mixed navigation modality (mouse and keyboard), and those are not the defaults we usually operate under. If we make sure that the `Continue` button, **_if it is indeed in focus_**, also has the proper visual focus indicator, then this could work Ok.

Actually, what's happening is:
if you come from a previous screen where you clicked on continue, the focus is set to continue after you click the checkbox. If you don't come from that previous button because you click anywhere else in the screen and, then click on the checkbox, nothing happens when clicking the Enter key.

Please, let me know exactly if: this is correct or if something needs to be changed, but after this long thread for such a simple screen I still don't know if any of the reviews requires a change or not.

kolibri/plugins/user_profile/assets/src/machines/changeFacilityMachine.js
jredrejo: so, do you want me to change it or do you want to open an issue to change it in the future? I just want to finish this simple task, if you prefer me to add a new state I'll rather do it than continue seeing days pass without a solution.

MisRob: Either of those will do

MisRob:
@jredrejo As posted above, regarding the state machine complexity issue, you can open an issue or resolve it here, it's up to you. Other than that, the code is looking good to me.

Regarding the keyboard and mouse navigation, I assumed that you could resolve it together with @radinamatic as I wasn't certain what's correct behavior and how important it is, I just noticed that discrepancy.

kolibri/plugins/user_profile/assets/src/machines/changeFacilityMachine.js
jredrejo: Thanks, I've implemented it to avoid needing to work on this again in the future.

MisRob:
>  I still don't know if any of the reviews requires a change or not

I usually try to sum this up in the main comment:
> The only blocking thing is the comment about back navigation condition.

If you have an idea about another way to mark blocking issues, let me know.

MisRob (approved):

radinamatic:
> if you come from a previous screen where you clicked on continue, the focus is set to continue after you click the checkbox. If you don't come from that previous button because you click anywhere else in the screen and, then click on the checkbox, nothing happens when clicking the Enter key.

I was searching for a clearer guidance around this issue in a11y specs, but came up empty...

I believe the safest solution which should help users avoid errors and premature progression to the next screen that can happen if the `Continue` button is placed in focus immediately after activating/checking a checkbox, is to **keep the focus on that (checked) checkbox**. This would allow them to `Tab`-navigate through the rest of the checkboxes and eventually get to the `Continue` button to advance to the next screen.

kolibri/plugins/user_profile/assets/src/machines/changeFacilityMachine.js
MisRob: I've just struggled with the same problem on the "Choose a new super admin" screen, but even more complex as besides those three main paths, there are also some sub-paths within them in regards to back navigation. So I've resolved it by adding a simple history implementation to the machine. Won't go into details now (will explain in the PR for the choose new super admin), just heads up if you're working on some new screens, it may be better to not spend time on `BACK` actions since the proposed implementation won't be compatible with the current way.

PR #9682, opened 2022-08-31
Check for an array's length to avoid breaking errors

### LINKED ISSUE #9088: Error when clicking on an exercise in the coach notifications feed when the user didn't generate any attempts
## Observed behavior

In the coach dashboard, I clicked on the purple text in this notification, but nothing happened:
image

## Errors and logs

In the console, I got this traceback:
image
image

It looks as if the user didn't actually have any AttemptLogs for that exercise, and progress was still at 0.

## Expected behavior

I would expect that either:
- The user could still be taken somewhere meaningful when clicking on that link, OR
- The notification text shouldn't be a link, if it's not actionable

## Context

Kolibri 0.15.1b0

nucleogenesis:
## Summary

@akolson and I explored #9088  together and could not replicate it. After some investigation and attempts to find a way to replicate, we determined that the most likely cause of the actual issue reported (that the link didn't work) was just a matter of the unsafe code that is reported in that issue stopping at that exception.

So we just made that code a bit safer by ensuring there is something to get an ID from when pulling the first item from the array there.

## References

Closes #9088 

## Reviewer guidance

Does the code look okay? I think this would be okay with just code review.

----

### DIFF
--- a/kolibri/plugins/coach/assets/src/modules/exerciseDetail/handlers.js
+++ b/kolibri/plugins/coach/assets/src/modules/exerciseDetail/handlers.js
@@ -50,8 +50,10 @@ function showExerciseDetailView({ learnerId, exerciseId, attemptId = null, inter
       return store.dispatch('exerciseDetail/setAttemptLogs').then(attemptLogs => {
         // No attemptId was passed in, so we should trigger a url redirect
         // to the first attempt.
-        if (attemptId === null) {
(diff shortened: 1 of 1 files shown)


MisRob (approved):
Thanks both, looking good to me

akolson (approved):
Looks good, thanks @nucleogenesis

PR #9684, opened 2022-09-01
Add regression test and fix for multiple worker processes duplicating jobs.

### LINKED ISSUE #9641: In Kolibri deployments, multiple frontend instances cause duplicate processing of tasks
## Observed behavior

In Kolibri cloud deployments, such as Kolibri Data Portal (KDP), we start multiple instances of Kolibri, each within an individual Kubernetes pod. Starting several instances of Kolibri means we also start up separate task workers within each pod. Doing so creates the opportunity for the duplication of task processing, since Kolibri's task processing does not protect against this happening.

## Errors and logs
In KDP, we receive errors similar to the following, which are raised in the asynchronous processing of Kolibri syncs:
```
django.db.utils.IntegrityError: duplicate key value violates unique constraint "morango_databasemaxcounter_instance_id_partition_99e4f1fb_uniq"
```
which leads to
```
kolibri_data_portal.management.commands.sync_proceed_to.SyncProceedToError: Failed to finalize <transfer session ID>
```
This would appear to be an issue, but isn't because when you look at the Morango `TransferSession` referenced by the error, you can see that it did complete successfully and didn't actually "fail to finalize". In addition, you can also see that the transfer was completed prior to the `duplicate key value violates unique constraint` error, which indicates that there was another worker trying to process the same task, hence the reason for `duplicate key` error.

## Expected behavior

Kolibri should implement a distributed locking mechanism which prevents Kolibri task workers from picking up tasks which have already been picked up and "locked" by a worker. It may be useful to annotate which worker picked up the task on the job record, and have workers filter for tasks using that; for instance the combination of worker hostname and process ID should work appropriately in deployments like KDP and would be useful for debugging too. 

## User-facing consequences

This can cause significant database bottlenecks on KDP's side since these tasks can retry their operations when they encounter transaction serialization errors due to their more aggressive transaction isolation. Also, the flood of these errors obscure real errors that need to be resolved.

## Context

Kolibri 0.14+0.15
Linux + Google Cloud

rtibbles:
## Summary
* Adds a new job status for when a job has been selected to be run next by the task runner but has not been sent to a worker yet
* Updates to this status upon finding a new job for SQLite
* Adds a row locking query that also skips locked rows for PostgreSQL to ensure that multiple worker processes will not select the same job for running
* Adds a regression test with two workers to ensure this is the case

## References
Fixes #9641

## Reviewer guidance
Are the queries clear enough in terms of what they are doing? Does the test sufficiently cover the concern?

----

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/core/tasks/job.py
+++ b/kolibri/core/tasks/job.py
@@ -47,6 +47,9 @@ class State(object):
     QUEUED means the Job has been sent to the workers for running, but has not
     been run yet (to our knowledge).
 
+    SELECTED means the Job has been picked up by the worker controller as a next
+    job to pass to its workers, but that it has not actually started running yet.
+
     RUNNING means that one of the workers has started running the job, but is not
     complete yet. If the job has been set to track progress, then the job's progress
     and total_progress fields should be continuously updated.
@@ -66,6 +69,7 @@ class State(object):
     PENDING = "PENDING"
     SCHEDULED = "SCHEDULED"
     QUEUED = "QUEUED"
+    SELECTED = "SELECTED"
     RUNNING = "RUNNING"
     FAILED = "FAILED"
     CANCELING = "CANCELING"
--- a/kolibri/core/tasks/storage.py
+++ b/kolibri/core/tasks/storage.py
@@ -6,6 +6,7 @@ from sqlalchemy import DateTime
 from sqlalchemy import func
 from sqlalchemy import or_
 from sqlalchemy import String
+from sqlalchemy import update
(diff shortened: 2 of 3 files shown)


rtibbles:
This is going to give us a bit of a merge conflict going into develop, so that's something to be mindful of.

bjester (approved):
Changes look good, regression test is nice, and tasks are still being processed in my simple testing

PR #9685, opened 2022-09-02
Revert "Update UserTable to allow single selection"

MisRob:
Reverts learningequality/kolibri#9597 which I merged by mistake


MisRob:
Merging promptly as I broke tests on `develop` and want to revert changes before folks start their new branches. Will open a new PR insted.

PR #9686, opened 2022-09-02
Update UserTable to allow single selection - second attempt

MisRob:
Includes changes from reverted https://github.com/learningequality/kolibri/pull/9597 that I merged accidentally to `develop` + new commit that fixes a test.


jredrejo (approved):
Reviewing it following the commits made it really easy. Code looks good to me and I haven't observed any regression in all the places where it can be tested. I would have liked to be able to test it in the setup wizard part but the wizard is still in an unstable situation and that part of the code can't be reached.
Great work!

MisRob:
@jredrejo Yes, if I recall I couldn't reach the setup wizard too, so I rely on that we'll be doing quite thorough QA for things that are currently in progress. Thank you for testing it.

PR #9687, opened 2022-09-02
Create screen to be shown if the username already exists in the facility

### LINKED ISSUE #9671: Create "Change Facility" screen if username exists
# Overview 

Adds a confirmation page to flows when a username in the current facility exists in the targe favility

## Figma 

image

jredrejo:
## Summary
Added screen to be shown when merge facilities is selected and there's already an account with the same username in the target facility.

image

## References
Closes: https://github.com/learningequality/kolibri/issues/9671

## Reviewer guidance
In order to review any of the PR related to the Change facility after device setup project, it's useful to  have in mind the current figma design https://www.figma.com/file/MpCG7r5PULOCJsEvnTSti8/Android-Designs-2022?node-id=611%3A9982

Also to be in context: the common use case for this will be an user who has kolibri installed (usually in a tablet o mobile device) and wants to join to a facility in a school. So two kolibri servers are needed: One kolibri server running the `target` facility where the user will merge his current profile and the kolibri server where these screens will be tested. Both servers must be running the latest code from `develop` (or a kolibri version >= 0.16 after it's released).

Depending on the screen to be tested:

After selecting the target facility (it'll have to be autodetected by the current server) ,  the first screen has two buttons, `Continue` or `Merge Accounts`. In the latter case the current user will be merged with an existing user in the target facility, in the  former case a new user will be created in the target facility. For this PR to be tested, the `Merge Account` button must be clicked.

After clicking the `Merge Account`, depending on the case,  the _target facility_ will need:
* An user created with the same _username_ the current facility has (it applies to this PR)
* An available user to be merged, but not having the same _username_ the current facility has (it applies to different PR)

If all these conditions are fulfilled, the above screen must appear to the user.

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/plugins/user_profile/assets/src/views/ChangeFacility/MergeUsernameExists.vue
+++ b/kolibri/plugins/user_profile/assets/src/views/ChangeFacility/MergeUsernameExists.vue
@@ -1,0 +1,123 @@
+<template>
+
+  <div>
+    <h1>{{ $tr('documentTitle') }}</h1>
+    <p data-test="line1">
+      {{ mergeAccountInfoLine1 }}
+    </p>
+    <p data-test="line2">
+      {{ mergeAccountInfoLine2 }}
+    </p>
+
+    <BottomAppBar>
+      <slot name="buttons">
+        <KButtonGroup>
+          <KButton
+            :primary="false"
+            :text="coreString('backAction')"
+            appearance="flat-button"
+            data-test="backButton"
+            @click="sendBack"
+          />
+          <KButton
+            :primary="true"
+            :text="coreString('continueAction')"
+            data-test="continueButton"
+            @click="handleContinue"
+          />
+        </KButtonGroup>
+      </slot>
+    </BottomAppBar>
+
+  </div>
+
+</template>
+
+
+<script>
+
+  import commonCoreStrings from 'kolibri.coreVue.mixins.commonCoreStrings';
+  import BottomAppBar from 'kolibri.coreVue.components.BottomAppBar';
+  import { computed, inject } from 'kolibri.lib.vueCompositionApi';
+  import get from 'lodash/get';
+
+  export default {
+    name: 'MergeUsernameExists',
+    metaInfo() {
+      return {
+        title: this.$tr('documentTitle'),
+      };
+    },
+    components: { BottomAppBar },
+
+    mixins: [commonCoreStrings],
+
+    setup(_, context) {
+      const changeFacilityService = inject('changeFacilityService');
+      const state = inject('state');
+      const store =
+        context.root.$store !== undefined ? context.root.$store : context.root.$children[0].$store;
+      const session = store.getters['session'];
+      const username = computed(() => session.username);
+      const mergeAccountInfoLine1 = computed({
+        get() {
+          return this.$tr('confirmMergeInfoLine1', {
+            target_facility: get(state, 'value.targetFacility.name', ''),
+            username: username.value,
+          });
+        },
+      });
+      const mergeAccountInfoLine2 = computed({
+        get() {
+          return this.$tr('confirmMergeInfoLine2', {
+            target_facility: get(state, 'value.targetFacility.name', ''),
+          });
+        },
+      });
+
+      function handleContinue() {
+        changeFacilityService.send({
+          type: 'CONTINUE',
+        });
+      }
+      function sendBack() {
+        changeFacilityService.send({
+          type: 'BACK',
+        });
+      }
+      return {
+        mergeAccountInfoLine1,
+        mergeAccountInfoLine2,
+        sendBack,
+        handleContinue,
+      };
+    },
+
+    $trs: {
+      documentTitle: {
+        message: 'Change facility',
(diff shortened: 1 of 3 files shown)


kolibri/plugins/user_profile/assets/src/views/ChangeFacility/MergeUsernameExists.vue
```diff
@@ -0,0 +1,123 @@
+<template>
+
+  <div>
+    <h1>{{ $tr('documentTitle') }}</h1>
```
akolson, line 4: @jredrejo It would be better to add this(`$tr('documentTitle')`) as a computed constant under the `setup()` hook since [ Line:48](https://github.com/learningequality/kolibri/pull/9687/files#diff-b6d35063aff073cfabb167aec1a23a7fc14dc458fd3e26b5204784eb8eb1d9f4R48) has the same implementation. This would improve maintainability of the translated title

kolibri/plugins/user_profile/assets/src/views/ChangeFacility/MergeUsernameExists.vue
```diff
@@ -0,0 +1,123 @@
...
+      const changeFacilityService = inject('changeFacilityService');
+      const state = inject('state');
+      const store =
+        context.root.$store !== undefined ? context.root.$store : context.root.$children[0].$store;
+      const session = store.getters['session'];
+      const username = computed(() => session.username);
+      const mergeAccountInfoLine1 = computed({
+        get() {
+          return this.$tr('confirmMergeInfoLine1', {
+            target_facility: get(state, 'value.targetFacility.name', ''),
```
akolson: It would be good to keep the variable, property, etc names consistent with the known Javascript conventions, unless there is a reason why this should be so for this case. The same would apply to [Line:73](https://github.com/learningequality/kolibri/pull/9687/files#diff-b6d35063aff073cfabb167aec1a23a7fc14dc458fd3e26b5204784eb8eb1d9f4R73)

kolibri/plugins/user_profile/assets/test/views/ChangeFacility/MergeUsernameExists.spec.js
```diff
@@ -0,0 +1,75 @@
...
+    const wrapper = makeWrapper();
+    expect(wrapper.exists()).toBeTruthy();
+  });
+
+  it(`Show correct info`, () => {
+    const wrapper = makeWrapper({
+      targetFacility: { name: 'Test Facility' },
+      username: 'test1',
+    });
+    const line1_paragraph = wrapper.find('[data-test="line1"]');
```
akolson: Would be ideal to keep the naming conventions consistent, please see below as well

kolibri/plugins/user_profile/assets/src/views/ChangeFacility/MergeUsernameExists.vue
```diff
@@ -0,0 +1,123 @@
...
+    },
+    components: { BottomAppBar },
+
+    mixins: [commonCoreStrings],
+
+    setup(_, context) {
+      const changeFacilityService = inject('changeFacilityService');
+      const state = inject('state');
+      const store =
+        context.root.$store !== undefined ? context.root.$store : context.root.$children[0].$store;
```
akolson, line 59: This is non blocking, but I would probably move this logic into a well named function, replace the ternary operator with an if statement and assign it here instead. This would improve readability of this piece of code.

akolson (changes requested):
Hi @jredrejo, code looks good generally. I left a few comments for you to review. Thanks

kolibri/plugins/user_profile/assets/src/views/ChangeFacility/MergeUsernameExists.vue
jredrejo: I don't think that would improve maintainability as the function is already implicit in the `tr` variable. Using a function would mean both strings have to be the same, they're in this design but not in all the screens. I think it's better to keep them separated.
Also adding a computed const won't work to be used in both places.

kolibri/plugins/user_profile/assets/src/views/ChangeFacility/MergeUsernameExists.vue
jredrejo: it was not actually a JS variable but a translation variable, but happy to change it.

kolibri/plugins/user_profile/assets/test/views/ChangeFacility/MergeUsernameExists.spec.js
jredrejo: done

kolibri/plugins/user_profile/assets/src/views/ChangeFacility/MergeUsernameExists.vue
akolson: Thanks @jredrejo for the context. My initial impression was that its the same string required in both places, so there was no need to do the translation twice.

akolson (approved):

PR #9714, opened 2022-09-12
Upgrade morango and lower default sync chunk size through CLI

### LINKED ISSUE #9677: Decrease syncing chunk size to reduce memory load in Morango's buffer API
## Observed behavior

@nucleogenesis's testing on migrating a large database to our cloud Kolibri infrastructure (BCK) encountered increased memory usage when transferring data. This occurred on the server side of a push sync, where Morango POSTs buffers to the API. It resulted in Kolibri getting OOM killed and a 502 error on the client side.

## Expected behavior

Reducing the chunk size a little bit should keep the memory load more stable and should decrease the likelihood our Kolibri application gets OOM killed.

## User-facing consequences

Sync clients may encounter 502 errors or requests that "hang" forever.

## Steps to reproduce

1. Sync a large dataset
2. Monitor memory usage on the server side of the sync

## Context

0.15+

bjester:
## Summary

- Decreases the default chunk size for the sync command to keep memory usage a bit more stable over time; this value matches the value used when syncing through the UI
- Upgrades `morango` to 0.6.15, which includes a fix for the syncing error https://github.com/learningequality/morango/issues/173
- If you were syncing more than 511,500 records (like @nucleogenesis), it was much more likely you'd run into the above overflow error while transferring with a chunk size of 500. An option is to increase the chunk size but that could cause heavy memory usage. So decreasing the chunk size needed the overflow fix
- Updates `make test-with-postgres` to use the port 15432 for postgres to avoid conflicts with host services based off feedback at last demo
  - And some related changes to support this update

## References

Closes https://github.com/learningequality/kolibri/issues/9677

## Reviewer guidance

I ran
```
make test-with-postgres INTEGRATION_TEST=1
```

----

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/utils/tests/test_cli.py
+++ b/kolibri/utils/tests/test_cli.py
@@ -164,7 +164,7 @@ def test_kolibri_listen_port_env(monkeypatch):
 
         test_port = 1234
 
-        test_zip_port = 5432
+        test_zip_port = 1432
 
         os.environ["KOLIBRI_HTTP_PORT"] = str(test_port)
 
--- a/Makefile
+++ b/Makefile
@@ -104,7 +104,7 @@ test-all:
 	export KOLIBRI_DATABASE_USER=postgres; \
 	export KOLIBRI_DATABASE_PASSWORD=postgres; \
 	export KOLIBRI_DATABASE_HOST=127.0.0.1; \
-	export KOLIBRI_DATABASE_PORT=5432; \
+	export KOLIBRI_DATABASE_PORT=15432; \
 	set -ex; \
 	function _on_interrupt() { docker-compose down; }; \
 	trap _on_interrupt SIGINT SIGTERM SIGKILL ERR; \
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -4,7 +4,7 @@ services:
   postgres:
     image: postgres:12
     ports:
-      - "5432:5432"
+      - "15432:5432"
     environment:
       POSTGRES_USER: postgres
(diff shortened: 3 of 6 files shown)


kolibri/utils/tests/test_cli.py
```diff
@@ -164,7 +164,7 @@ def start_mock(port, *args, **kwargs):
 
         test_port = 1234
 
-        test_zip_port = 5432
+        test_zip_port = 1432
```
bjester, line 167: @rtibbles I'm questioning the lack of previous failures on this test if one had postgres running on 5432. Like prior to these changes I would have expected to see a failure using `make test-with-postgres` but I don't recall that :thinking: Is the [assertion](https://github.com/learningequality/kolibri/pull/9714/files#diff-8c2d5a51d09f01b7dfda0dd480f4b502f50a96baa104b92d83f91ab30d6da923R179) on `assert excinfo.code` doing what it needs to?

bjester: To clarify, this test didn't fail, but updating this seemed to resolve issues with another test.

rtibbles: Oh, that is very odd - might be one of the causes of some of the flakiness we see with postgres tests? 

Particlularly odd that it doesn't fail, given that it does explicitly mark the Django DB as needed for the test too.

rtibbles (approved):

PR #9715, opened 2022-09-13
Change Facility: Choose a new super admin screen, error state, back navigation

### LINKED ISSUE #9326: Create "Choose Admin" page

MisRob:
## Summary

**Adds the _Choose a new super_ admin screen to all three main branches of the change facility flow**

That created the need for the following higher-level updates that are also included:

**Adds a new error state to the machine and related logic for redirecting to the unexpected error page as well as resetting the machine when that happens**
  - **Motivation**: To avoid becoming stuck in the flow without any clues when an unexpected error occurs, typically when invoking services in between user-facing transitions
  - **Implementation**: Adds a new `error` state to the state machine and utilizes existing routing logic when transitioning between states to determine whether a user should be redirected to an unexpected error page which will be displayed thanks to our existing Vuex error logic.
 

**Adds history for back navigation to the state machine**
  - **Motivation**: We’ve had issues with complexity of conditions for back navigation previously which has shown to be the case for the _Choose a new super admin_ screen too in an even more complicated manner
  - **Implementation**: Adds `history` to context. States can explicitly push their name to the history. This explicitness is intentional to give us more control; for example, we don’t want to push each state name to history as some of them are not-user facing. `BACK` transition has been removed from all states in favor of one global `BACK` transition that uses `history` context. Inspired by [https://github.com/statelyai/xstate/discussions/1939#discussioncomment-381565](https://github.com/statelyai/xstate/discussions/1939#discussioncomment-381565), when implementing this `BACK` transition logic, `cond` s generated from the object containing all states were utilized.

It also contains a couple of other minor updates, see commit messages.

## References

- [Figma designs](https://www.figma.com/file/MpCG7r5PULOCJsEvnTSti8/Android-Designs-2022?node-id=194%3A2920)
- Closes https://github.com/learningequality/kolibri/issues/9326

## Reviewer guidance

- You need to have another v16 facility set up and running, the so-called target facility, to be able to access the change facility flow. You could use `kolibri start` to run it.
- At your source facility, go to *Profile* → *Change learning facility* to start testing
- The following conditions need to be satisfied for accessing
    - *(1) Default path*
        - on the target facility, there shouldn’t be a user with the same username as your current source facility user
    - *(2) Username exists*
        - on the target facility, there needs to be a user with the same username as your current source facility user
    - *(3) Merge accounts*
        - Currently, I was able to go through this path only when I had a user with the same username on the target facility as my current source facility user
- The whole change facility flow is a work in progress, so not all screens have been implemented yet. It should be possible to access the *Choose a new super admin*. Please let me know if you encounter any issues.

### Choose a new super admin

Please follow [the Figma designs](https://www.figma.com/file/MpCG7r5PULOCJsEvnTSti8/Android-Designs-2022?node-id=194%3A2920) to check the _Choose a new super admin_ page in all _(1) Default path_, _(2) Username exists_, and _(3) Merge accounts_ contexts. While navigating those paths, check:

- [ ] *Chose a new super admin* page is displayed when there are more users on the source device and the current user is the only super user
- [ ] _Choose a new super admin_ page is skipped when
  - the current user is not a super user
  - the device has more super users
  - the device has only one user
 
When _Choose a new super admin_ page is displayed
- [ ] _Continue_ button is disabled, and you can’t move to the next screen until you select a user from the table
- [ ] Keyboard navigation in the table
- [ ] After you click _Continue_ (and confirm merge if needed), the selected user ID is saved to `newSuperAdminId` state machine’s context (the whole context object is printed out at the end of the flow)

### Error state

I haven’t found a better way then to replace the state machine’s `fetchSourceFacilityUsers -> invoke -> src` by

```jsx
src: () => {
  return FacilityUserResource.fetchCollection().then(users => {
    return Promise.reject(;
  });
},
```

Then, try to navigate to the *Chose a new super admin* page.

- [ ] Check that you can see an unexpected error page

### Back navigation
Click through the change facility flow and check

- [ ] navigation back with _Back_ buttons works on pages with _Back_ buttons
- [ ] navigation back via browser's back works

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/core/assets/src/mixins/commonCoreStrings.js
+++ b/kolibri/core/assets/src/mixins/commonCoreStrings.js
@@ -368,7 +368,7 @@ export const coreStrings = createTranslator('CommonCoreStrings', {
   searchForUser: {
     message: 'Search for a user...',
     context:
-      'Text which appears in the search field above the table with users from whose to choose from (e.g. when enrolling learners to a class, selecting users to sync, etc.)',
+      'Text which appears in the search field above the table with users from whom to choose from (e.g. when enrolling learners to a class, selecting users to sync, etc.)',
   },
   findSomethingToLearn: {
     message: 'Find something to learn',
--- a/kolibri/core/assets/src/mixins/commonCoreStrings.js
+++ b/kolibri/core/assets/src/mixins/commonCoreStrings.js
@@ -365,6 +365,11 @@ export const coreStrings = createTranslator('CommonCoreStrings', {
     message: 'Search',
     context: 'Test used to indicate the Kolibri search field.',
   },
+  searchForUser: {
+    message: 'Search for a user...',
+    context:
+      'Text which appears in the search field above the table with users from whose to choose from (e.g. when enrolling learners to a class, selecting users to sync, etc.)',
+  },
   findSomethingToLearn: {
     message: 'Find something to learn',
     context: 'Suggestion located inside the the keyword search field.',
--- a/kolibri/plugins/user_profile/assets/src/machines/changeFacilityMachine.js
+++ b/kolibri/plugins/user_profile/assets/src/machines/changeFacilityMachine.js
@@ -77,11 +77,12 @@ const setSourceFacilityUsers = assign({
   sourceFacilityUsers: (_, event) => event.data,
 });
 
-export const changeFacilityMachine = createMachine({
-  id: 'machine',
-  initial: 'selectFacility',
-  predictableActionArguments: true,
-  context: {
+const resetMachineContext = assign(() => {
+  return generateMachineContext();
+});
+
+const generateMachineContext = () => {
+  return {
     role: 'learner',
     username: '',
     userId: '',
@@ -99,8 +100,23 @@ export const changeFacilityMachine = createMachine({
     taskPolling: false,
     accountExists: false,
     isMerging: false,
-  },
+  };
+};
+
+export const changeFacilityMachine = createMachine({
+  id: 'machine',
+  initial: 'selectFacility',
+  predictableActionArguments: true,
+  context: generateMachineContext(),
   states: {
+    error: {
+      on: {
+        RESET: {
+          target: 'selectFacility',
+          actions: [resetMachineContext],
+        },
+      },
+    },
     profile: {
       meta: { route: 'PROFILE', path: '/' },
       on: {
@@ -198,6 +214,9 @@ export const changeFacilityMachine = createMachine({
         onDone: {
           target: 'checkNeedsNewSuperAdmin',
           actions: [setSourceFacilityUsers],
+        },
+        onError: {
+          target: 'error',
         },
       },
     },
--- a/kolibri/plugins/user_profile/assets/src/machines/changeFacilityMachine.js
+++ b/kolibri/plugins/user_profile/assets/src/machines/changeFacilityMachine.js
@@ -178,12 +178,12 @@ export const changeFacilityMachine = createMachine({
           actions: setMerging,
         },
         {
-          target: 'confirmAccount',
+          target: 'confirmAccountUsername',
         },
       ],
     },
-    confirmAccount: {
-      meta: { route: 'CONFIRM_ACCOUNT', path: '/change_facility' },
+    confirmAccountUsername: {
+      meta: { route: 'CONFIRM_ACCOUNT_USERNAME', path: '/change_facility' },
       on: {
         NEW: {
           cond: context => context.targetFacility && context.targetFacility.learner_can_sign_up,
@@ -244,7 +244,7 @@ export const changeFacilityMachine = createMachine({
           cond: context => !!context.newSuperAdmin,
         },
         SELECTNEWSUPERADMIN: { actions: setNewSuperAdmin },
-        BACK: 'confirmAccount',
+        BACK: 'confirmAccountUsername',
       },
     },
     checkIsMerging: {
@@ -290,7 +290,7 @@ export const changeFacilityMachine = createMachine({
             target: 'changeFacility',
           },
           {
-            target: 'confirmAccount',
+            target: 'confirmAccountUsername',
           },
         ],
       },
--- a/kolibri/plugins/user_profile/assets/src/machines/changeFacilityMachine.js
+++ b/kolibri/plugins/user_profile/assets/src/machines/changeFacilityMachine.js
@@ -175,7 +175,6 @@ export const changeFacilityMachine = createMachine({
         {
           cond: context => !!context.accountExists,
           target: 'usernameExists',
-          actions: setMerging,
         },
         {
           target: 'confirmAccountUsername',
--- a/kolibri/plugins/user_profile/assets/src/machines/changeFacilityMachine.js
+++ b/kolibri/plugins/user_profile/assets/src/machines/changeFacilityMachine.js
@@ -118,10 +118,19 @@ const generateMachineContext = () => {
     // Doesn't necessarily capture all transitions as it is used
     // for user-facing back navigation, therefore we don't want to
     // save  rather internal transitions here.
+    //
     // Each state can save itself explicitly to history by calling
     // `send({ type: 'PUSH_HISTORY', value: <stateName> })` action,
     // typically this would happen when moving forward to another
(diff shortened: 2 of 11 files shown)


MisRob:
@jredrejo Pushed a commit that adds more detailed commentary to the history implementation in regards to points we discussed, as well as mentions some disadvantages of the current solution so that we can refer to it when researching the machine architecture.

kolibri/plugins/user_profile/assets/src/machines/changeFacilityMachine.js
```diff
@@ -71,238 +73,359 @@ const setMerging = assign({
...
-          cond: context => !!context.newSuperAdmin,
-        },
-        SELECTNEWSUPERADMIN: { actions: setNewSuperAdmin },
-        BACK: 'confirmAccount',
+  },
+  isAdmin: {
+    always: [
+      {
+        cond: context => context.role === 'superuser',
+        target: 'fetchSourceFacilityUsers',
```
jredrejo, line 236: I see this is only used in the ChooseAdmin component. I've seen facilities with thousands of users. I don't feel comfortable having such list inside the machine context just to be used in one component. I think it'd be better if this is fetched and, later removed from the browser memory after the component is unmounted, only inside that component.

kolibri/plugins/user_profile/assets/src/machines/changeFacilityMachine.js
```diff
@@ -71,238 +73,359 @@ const setMerging = assign({
...
+    // don't use and it seems that it's made for a bit different use case
+    // (moving to/out nested states and remembering the last children state)
+    // than our history.
+    history: [],
+  };
+};
+
+const states = {
+  error: {
+    on: {
```
jredrejo, line 144: this is a great idea, thanks!

jredrejo (changes requested):
Overall, it looks good to me. As we've been previosly chatting I think we need to find a better architecture for the uses of machine states with Vue in flows that can become as complex as this, and refactor and use it the same way in all the places in the application. I think it'd be good if we can find time to study it in the future, but for now, this is the best solution to continue without blocking any other work.

I've left one comment for something I'd like to change, we can discuss if it's possible, to avoid overloading the browser with much data.

Anyway, I have a final question: this PR does not really do the super admin change, so I'll add an issue to the end of all these screens to do the change if `newSuperAdminId !== '' `

kolibri/core/assets/src/mixins/commonCoreStrings.js
```diff
@@ -365,6 +365,11 @@ export const coreStrings = createTranslator('CommonCoreStrings', {
     message: 'Search',
     context: 'Test used to indicate the Kolibri search field.',
   },
+  searchForUser: {
+    message: 'Search for a user...',
+    context:
+      'Text which appears in the search field above the table with users from whose to choose from (e.g. when enrolling learners to a class, selecting users to sync, etc.)',
```
akolson: @MisRob  please make the following changes to the context so that its more clear.

> Text which appears in the search field above the table with users from ~~whose~~ whom to choose ~~from~~ (e.g. when enrolling learners to a class, selecting users to sync, etc.)

MisRob:
Thank you @jredrejo, that's helpful feedback.

> I have a final question: this PR does not really do the super admin change, so I'll add an issue to the end of all these screens to do the change if newSuperAdminId !== '' 

Yes, that was intentional, as previously we discussed that we would apply all changes during the syncing step (since a user could go back during various steps and change their mind, generally speaking), so I assumed this would happen there too without much thinking. Does that still sound okay? If yes, you're right that it'd be better to have an issue so that we don't forget that this needs to be done. And I am fine with opening it too :)

kolibri/plugins/user_profile/assets/src/machines/changeFacilityMachine.js
MisRob: Thank you, that's an important point.

I considered doing this inside the component, however, source facility users' data is not used only in the `chooseAdmin` state. Even before that, it's used in the machine in the `checkNeedsNewSuperAdmin` that decides whether the machine should go to the `chooseAdmin` or `checkIsMerging` state. In user experience, that translates to the decision whether the _Choose a new super admin_ page should be displayed or skipped, and for that, we need to have that data available before we mount the _Choose a new super admin_ page.

I agree that this should be resolved. I will think about it and let's chat together in a bit as we planned.

kolibri/core/assets/src/mixins/commonCoreStrings.js
MisRob: Thanks, will do

kolibri/plugins/user_profile/assets/src/views/ChangeFacility/index.vue
```diff
@@ -69,22 +70,31 @@
...
-            } else this.$router.push(newRoute);
+        if (state.value === 'error') {
+          this.$store.commit(
+            'CORE_SET_ERROR',
+            `An error occured in the '${this.previousMachineStateName}' state of the change facility machine`
+          );
+          this.service.send('RESET');
+        } else {
+          const stateID = Object.keys(state.meta)[0];
+          if (state.meta[stateID] !== undefined) {
```
akolson: @MisRob I can't seem to understand whats going on in this block, furthermore, the deep nesting makes it harder to read and follow. I think refactoring this into function(s) could make this more clearer

MisRob: I think this connects the machine with the router. You're right that I added one level of nesting with my change, so I'll update this whole part to improve readability and also comment on the part you pointed out on that opportunity in more detail.

akolson: Thanks @MisRob. My apologies, the comment is intended for the whole else block line:[79](https://github.com/learningequality/kolibri/pull/9715/files#diff-57b8e61916c5f184c9b4a498234a098481a08861a0b852b5e898153fe96699c4R79) to [96](https://github.com/learningequality/kolibri/pull/9715/files#diff-57b8e61916c5f184c9b4a498234a098481a08861a0b852b5e898153fe96699c4R97)

MisRob: I think we will need to hear from @jredrejo because I don't recall what was the reason for that block exactly

akolson (changes requested):
Hi @MisRob, code generally looks good, Thanks! Would be good to address the complex sections as pointed out to aide readability and understandability

kolibri/plugins/user_profile/assets/src/views/ChangeFacility/index.vue
MisRob: Regarding additions in this PR, pushed a commit to simplify it

MisRob:
@akolson @jredrejo Thanks both, I've just pushed all updates.

FYI with @jredrejo, we discussed and made some decisions together in regard to his feedback.

Also, here's a follow-up issue for setting the new super admin https://github.com/learningequality/kolibri/issues/9725

kolibri/plugins/user_profile/assets/src/machines/changeFacilityMachine.js
MisRob: With @jredrejo, we decided that for now, I will just make sure to clean that data. We also had some ideas on how to further optimize it, and because there are some product questions, I will raise this for discussion with the team if there are some more suggestions, it will most likely be implemented in the future as an improvement.

Regarding the data clean up, pushed a commit with that updated.

akolson (approved):
Thanks @MisRob code looking good!

jredrejo (approved):
After all our comments and changes I think this is now good to go into develop. Let's continue!

PR #9723, opened 2022-09-15
Added function handler to delay SaveAnswer function

### LINKED ISSUE #9444: Coach>Reports>Quiz score - Incorrect number of attempts
## Observed behavior
I see an incorrect number of attempts at 'Coach>Reports>Quiz score' whenever there's a numeric value input as the answer to a quiz question.
For example if the answer is '12' I will see one failed attempt for the first entered digit '1' and one successful attempt for '12'

1. Install Kolibri 0.15.2
2. Import [Khan Academy's](https://www.khanacademy.org/math/cc-third-grade-math/represent-and-interpret-data/imp-picture-graphs/e/reading_pictographs_2) (English - US curriculum) > Math by grade > 3rd grade > Represent and interpret data > Picture graphs 
3. Create a quiz for 'Read picture graphs (multi-step problems)'
4. Complete the quiz and as a Coach go to Coach>Reports>Quiz score

## Expected result
Correctly displayed number of attempts.

## Usage Details
Ubuntu 20 - Chrome, Firefox

LianaHarris360:
## Summary

This PR adds a debounce function handler for the saveAnswer method, which previously saved incomplete inputs and showed multiple inputs on a completed quiz when the user intended to have one single answer.

## References

Fixes #9444 

## Reviewer guidance

Note: *Multiple answers might still be recorded if a user takes more than a second to input the complete answer.*
1. On Kolibri 0.15.2, import Khan Academy's (English - US curriculum) > Math by grade > 3rd grade > Represent and interpret data > Picture graphs
2. Create a quiz for 'Read picture graphs (multi-step problems)'
3. Complete the quiz 
4. As a Coach go to Coach>Reports>Quiz score

----

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/plugins/learn/assets/src/views/ExamPage/index.vue
+++ b/kolibri/plugins/learn/assets/src/views/ExamPage/index.vue
@@ -38,7 +38,7 @@
               :assessment="true"
               :allowHints="false"
               :answerState="currentAttempt.answer"
-              @interaction="saveAnswer"
+              @interaction="debouncedSaveAnswer"
             />
             <UiAlert v-else :dismissible="false" type="error">
               {{ $tr('noItemId') }}
@@ -297,6 +297,15 @@
         return this.displayNavigationButtonLabel
           ? { position: 'relative', top: '3px', left: '-4px' }
           : {};
+      },
+      debouncedSaveAnswer() {
+        // So as not to share debounced functions between instances of the same component
+        // and also to allow access to the cancel method of the debounced function
+        // best practice seems to be to do it as a computed property and not a method:
+        // https://github.com/vuejs/vue/issues/2870#issuecomment-219096773
+        // 750ms chosen through trial and error designed to allow users to input relatively
+        // slowly while producing one single answer
+        return debounce(this.saveAnswer, 750);
       },
     },
     watch: {


marcellamaki (approved):
Looks good to me! Working as expected. (If we get further feedback on this, we can adjust the debounce time as needed, but the implementation is great!) Nice work @LianaHarris360 🎉

pcenov:
Hi @LianaHarris360, it seems that the issue with the incorrect number of attempts has been fixed however there is something strange going on as a result of which some of my correct answers are not being marked as correct.
Please have a look at the following video where on question 9 the 0 from my answer gets deleted by itself, and also my answer to question 10 is not counted at all:

https://user-images.githubusercontent.com/79847249/193800516-746c88b8-7dcc-4127-b61a-2e6548fc0394.mp4

Screenshot with the results:


I tried completing this quiz with 3 users always using the correct answers and never got a 100% score. Note that the issue is not limited to the numeric fields only as some of the multi-select questions were also marked as incomplete.

LianaHarris360:
Thank you for letting me know @pcenov! I will look into this and see where the problem is occurring.

rtibbles:
Hrm, this might have been why there was no debounce in the first place - to prevent this from happening.

One possibility @LianaHarris360 is that the new `debouncedSaveAnswer` is never getting flushed at any point, so that might be where we are losing responses.

If you can replicate the issue, but can't resolve it, it's OK to revert the change here instead, and we'll come up with another idea for the multiple attempt issue!

LianaHarris360:
I thought about something like that as well. I will investigate a bit and update you if I cannot resolve it @rtibbles.

LianaHarris360:
I was able to replicate the issue @rtibbles. After typing in an answer, the response is lost when a user clicks the next button before the `saveAnswer` function is called. Setting the leading flag on the debounced function causes the first character to be submitted as attempt 1.

rtibbles:
So, with that in mind, I think making sure we flush our new debounced function when we move to the next question may be needed.

So this line: https://github.com/LianaHarris360/kolibri/blob/ee9f2f2f266b0bf5688c44535d061ff96b34bcb2/kolibri/plugins/learn/assets/src/views/ExamPage/index.vue#L393

Would become something like:
```
const saveAnswerPromise = this.debouncedSaveAnswer.flush() || Promise.resolve();
const promise = saveAnswerPromise.then(() => this.debouncedSetAndSaveCurrentExamAttemptLog.flush() || Promise.resolve());
```

So we make sure to flush the `debouncedSaveAnswer` then flush the `debouncedSetAndSaveCurrentExamAttemptLog`.

LianaHarris360:
Yep, that fixed the problem for me. I will update this pr with those changes!

rtibbles:
Great, glad that there was a solution here that kept the original fix too!

PR #9730, opened 2022-09-20
SetupWizard - Partial provisioning, On My Own Flow

nucleogenesis:
## Summary

- Prepares SetupWizard flow so that it can be partially provisioned (ie, a Facility can have been created/imported without the DeviceSettings.is_provisioned flag being set to True) -- this covers the edge case of a user not completing the process of importing users one by one and returning to it without ever saying "Okay I'm done onboarding"
- The FacilityImportResource / FacilityImportViewSet were both set to use the new `is_provisioned` flag when given in a request - it defaults to `False` **honestly not sure what default to use, as is evidenced by it defaulting to True in device settings**
- The same two were also changed so that `createsuperuser` will create a facility with a given `facility_name`.

**Outstanding issues (follow-up things)**

The following is what, at a high level, remains for the Setup Wizard. I will create a new set of issues and add them to the backlog following up on this when approved and before merging.

- Only the "On My Own" flow is setup to work properly here, other flows need work still (high effort)
- The "Step X of Y" logic is not yet implemented (medium effort)
- Wiring up bits around the user's device name and such needs to be implemented when creating facilities and devices (medium effort)
- The Interstitial needs to be designed (low effort)

## Reviewer guidance

- Can you go from start to finish through the On My Own flow?
- Thoughts on what the sane default for "is_provisioned" should be end to end?
- Any concerns with creating the facility inside the "createsuperuser" method?
- Does your language selection during Onboarding get reflected properly? (refresh on the page that says "-- INTERSTITIAL --")

----

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/core/device/serializers.py
+++ b/kolibri/core/device/serializers.py
@@ -48,7 +48,7 @@ class DeviceProvisionSerializer(DeviceSerializerMixin, serializers.Serializer):
     device_name = serializers.CharField(max_length=50, allow_null=True)
     settings = serializers.JSONField()
     allow_guest_access = serializers.BooleanField(allow_null=True)
-    is_provisioned = serializers.BooleanField(allow_null=True)
+    is_provisioned = serializers.BooleanField(default=True)
 
     class Meta:
         fields = (
--- a/kolibri/core/device/serializers.py
+++ b/kolibri/core/device/serializers.py
@@ -48,6 +48,7 @@ class DeviceProvisionSerializer(DeviceSerializerMixin, serializers.Serializer):
     device_name = serializers.CharField(max_length=50, allow_null=True)
     settings = serializers.JSONField()
     allow_guest_access = serializers.BooleanField(allow_null=True)
+    is_provisioned = serializers.BooleanField(allow_null=True)
 
     class Meta:
         fields = (
@@ -57,6 +58,7 @@ class DeviceProvisionSerializer(DeviceSerializerMixin, serializers.Serializer):
             "settings",
             "device_name",
             "allow_guest_access",
+            "is_provisioned",
         )
 
     def validate(self, data):
@@ -109,10 +111,13 @@ class DeviceProvisionSerializer(DeviceSerializerMixin, serializers.Serializer):
             # Create device settings
             language_id = validated_data.pop("language_id")
             allow_guest_access = validated_data.pop("allow_guest_access")
+
             if allow_guest_access is None:
                 allow_guest_access = preset != "formal"
+
             provision_device(
                 device_name=validated_data["device_name"],
+                is_provisioned=validated_data["is_provisioned"],
                 language_id=language_id,
                 default_facility=facility,
                 allow_guest_access=allow_guest_access,
--- a/kolibri/core/device/utils.py
+++ b/kolibri/core/device/utils.py
@@ -90,13 +90,13 @@ def set_device_settings(**kwargs):
         raise DeviceNotProvisioned
 
 
-def provision_device(device_name=None, **kwargs):
+def provision_device(device_name=None, is_provisioned=True, **kwargs):
     from .models import DeviceSettings
 
     device_settings, _ = DeviceSettings.objects.get_or_create(defaults=kwargs)
     if device_name is not None:
         device_settings.name = device_name
-    device_settings.is_provisioned = True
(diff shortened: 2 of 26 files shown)


kolibri/core/device/utils.py
```diff
@@ -90,13 +90,13 @@ def set_device_settings(**kwargs):
         raise DeviceNotProvisioned
 
 
-def provision_device(device_name=None, **kwargs):
+def provision_device(device_name=None, is_provisioned=True, **kwargs):
```
rtibbles, line 93: If `is_provisioned` is not set on the request to the serializer, then `is_provisioned` here will be `None`, not `True` - the default value is only used in the case that the value is not passed to the function at all.

kolibri/core/device/serializers.py
```diff
@@ -48,6 +48,7 @@ class DeviceProvisionSerializer(DeviceSerializerMixin, serializers.Serializer):
     device_name = serializers.CharField(max_length=50, allow_null=True)
     settings = serializers.JSONField()
     allow_guest_access = serializers.BooleanField(allow_null=True)
+    is_provisioned = serializers.BooleanField(allow_null=True)
```
rtibbles: To retain the previous behaviour, this would need to default to `True` rather than allowing `None`.

rtibbles (commented):
Not completely sure, but I think the slight changes to the provision device endpoint might be responsible for the test failures here.

jredrejo (changes requested):
The 'on my own' flow requires to add that info to the extra_fields in the db: https://github.com/learningequality/kolibri/pull/9577
Probably, this is a lack of communication between the devs, but this is being used in the "Merge user into a facility".

OTOH, after refreshing the 'Intersitial' screen, the login screen appears, but the Sign into '' is quite odd. In my opinion, there should be a facility name or we should remove that phrase. 
image

nucleogenesis:
@jredrejo I'll address that in a follow-up issue if that is alright -- if you're okay with that we can merge this and I'll draw up some new issues today for the follow-up on the finishing up the setup wizard altogether

jredrejo:
> @jredrejo I'll address that in a follow-up issue if that is alright -- if you're okay with that we can merge this and I'll draw up some new issues today for the follow-up on the finishing up the setup wizard altogether

The thing is that without that setting, there are features that can't be tested. It's fine if people testing them know that it will require some shell work to do it to test the "Change facility" flow, as it is only meant to be working in devices having that setting.

radinamatic:
From the QA team point of view, we can only answer these 2 questions in the Review guidance:

1. **Can you go from start to finish through the On My Own flow?**

    Not sure if I could call it _from-start-to-finish_ with the `-- INTERSTITIAL --` hickup in between, but if it's going to be addressed soon OK... 🤔

    The bigger concern is that, once we sign in again with the user created in this _On My Own_ flow, what we see is still the full admin user with access to Coach, Facility and the rest of Device page options. We chatted with @marcellamaki to confirm that the way we understood the Figma design for this flow aligns: this onboarding should end up with the user with the same type of permissions as on the current **Learn-only** devices, irregardless they are installing on the app or doing the onboarding in the browser.

    on-my-own2

    Is this also something that will be addressed in the follow-up PRs, or are we misunderstanding something?

2. **Does your language selection during Onboarding get reflected properly?**

    It seems OK, some strings are missing translations at this point, but otherwise the experience is correct.

radinamatic:
Apologies for late reply @nucleogenesis 

There seems to be some progress, but the main concern exposed above still stands: user created with this flow has full superadmin device permissions.

1. No  `-- INTERSTITIAL --` just the spinner which does not get to lead anywhere (I waited for 5 minutes and stopped the recording)

    on-my-own-second-take

2.  After manually going to the sign in page I can still see the normal superadmin UI
   
   https://user-images.githubusercontent.com/1457929/197795862-fd0c96a0-40e7-49db-8251-51ab7a556b02.mp4

I can approve this PR if working of it will make your next task easier, but this is not the state we were expecting the **on-my-own** flow to be, nor we can use it for testing other features like **Merge accounts** and similar.

nucleogenesis:
@radinamatic let's merge this one and I'll follow up on all of the above in a separate PR

PR #9683, opened 2022-08-31
Startup time improvements

rtibbles:
## Summary
* Limit imports of pkg_resources to defer a costly import
* Move code out of module scope and into the core app ready method
* Defer actual registration of zeroconf to the `RUN` event so that it doesn't block `SERVING`
* Default the `no_input` kwarg of initialize to `True` - this will not effect cli invocations that provide their own default value, but will help apps that are calling the `initialize` function and which are running in `no_input` contexts.

~In all, these changes seemed to give approximately a 60-70% reduction in time taken from first start to being ready to serve HTTP requests.~

Unfortunately, this seems overblown, and seems likely a result of being on a dev system, and the code paths that follows.

Also fixes https://github.com/learningequality/kolibri-app/issues/105

----

## Testing checklist

- [x] Contributor has fully tested the PR manually
- [ ] If there are any front-end changes, before/after screenshots are included
- [ ] Critical user journeys are covered by Gherkin stories
- [x] Critical and brittle code paths are covered by unit tests

## PR process

- [x] PR has the correct target branch and milestone
- [x] PR has 'needs review' or 'work-in-progress' label
- [x] If PR is ready for review, a reviewer has been added. (Don't use 'Assignees')
- [ ] If this is an important user-facing change, PR or related issue has a 'changelog' label
- [ ] If this includes an internal dependency change, a link to the diff is provided

## Reviewer checklist

- Automated test coverage is satisfactory
- PR is fully functional
- PR has been tested for [accessibility regressions](http://kolibri-dev.readthedocs.io/en/develop/manual_testing.html#accessibility-a11y-testing)
- External dependency files were updated if necessary (`yarn` and `pip`)
- Documentation is updated
- Contributor is in AUTHORS.md

Issue #9266, opened 2022-04-08
Refactor content import management commands to use utility functions that are shared between the management command and the task implementation to allow for better testing and reusability


rtibbles:
Currently most of our content tasks use the old `AsyncCommand` base command to do double duty as a management command and an asynchronous task. It would be preferable to modularize our task code to allow it to be flexibly reused as needed across the management commands and asynchronous tasks.

The scope of this task is limited to, looking at this file: https://github.com/learningequality/kolibri/blob/develop/kolibri/core/content/tasks.py

rtibbles:
The core of this was resolved by #8269 - the only thing that remains is to remove the dependency on management commands for core functionality for better modularity and easier testing. However, that's a tech debt issue that is beyond the scope of what is needed.

Updating the task accordingly and removing from the epic.

rtibbles:
All required sub issues have been completed, this is complete! Thanks @thesujai !

PR #9732, opened 2022-09-22
Content import composability refactor and single resource import

### LINKED ISSUE #9491: Create a public import metadata endpoint
A new public API endpoint that exposes all the metadata required for importing a specific resource - it should be a detail endpoint where the pk is the node id to be imported.

It should return a JSON object with keys for each of the DB table names for each table used in the metadata import.

Under each of these keys there should be a JSON array of all the rows that need to be imported into the database for the specific resource.

The overall JSON object should also have a `schema_version` key that specifies the schema_version that have been exported as.

The API endpoint can take an optional GET param of `schema_version` and should honour it if capable of returning the metadata in that format - as this endpoint will not be used by older versions of Kolibri, we do not need to honour this GET param for earlier schema_versions than that used in 0.16. If the API endpoint is unable to honour the GET param it should do a permanent redirect to a URL without the GET param.

To honour the request, the appropriate SQLAlchemy schema should be used to read the data from the default database and then returned.

The appropriate data should be:

The resource itself, any ancestors, the channel metadata, any associated files, local files, tags, languages, any through tables. For an exhaustive list, see the channel metadata import machinery: https://github.com/learningequality/kolibri/blob/develop/kolibri/core/content/utils/channel_import.py#L46

### LINKED ISSUE #9492: Enhance channel metadata import to import from a JSON object containing the metadata
At the moment all channel metadata import is imported from a SQLite database source.

In order to allow easier import of smaller chunks of channel metadata, it would be useful to be able to import metadata from a JSON object as well.

This would involve some small updates to the channel metadata machinery.

In the case of SQLite, in order to continue to use the same underlying mechanisms, the simplest way to handle this would be to copy the contents of the JSON object into an in-memory SQLite database. This would allow us to then do ATTACH still for fast import to the default database.

In the case of PostgreSQL, we can instead use the JSON object as the source of the CSV generators for each table that are then used in the COPY FROM operations for import.

These two interventions will then allow us to continue to use the backwards compatibility provided by the the metadata import layer.

### LINKED ISSUE #9552: Create async task for importing a specific resource via partial channel metadata import from JSON
Create an async task that leverages the import metadata endpoint and partial channel metadata import via JSON to allow the import of a single resource node, and its ancestor topics.

rtibbles:
## Summary
* Adds the ability to import content metadata from JSON rather than a SQLite database
* Updates SQLAlchemy version to 1.4 and makes syntax updates
* Allows for import of partial channel data (sub trees rather than the whole tree) and incremental import of the entire tree
* Creates a public API endpoint to return the required metadata JSON for importing a single resource
* Does some refactors to make content import functionality more composable
* Creates an async task for importing a single resource over the network

## References
Fixes #9552
Fixes #9492
Fixes #9491
Completes some of #9266
Last commit also fixes https://github.com/learningequality/kolibri-app/issues/105, which might be cherry-picked to https://github.com/learningequality/kolibri/pull/9683 if that is approved for merge first.

## Reviewer guidance
Do the tests of the new import functionality make sense? Does importing a resource from another instance work?

Does the way I've split up the different functionality of the import manager classes feel readable/practical/sustainable?

----

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated


rtibbles:
DMG build failure appears to be unrelated to these changes. Tracked separately: https://github.com/learningequality/kolibri-app/issues/105

kolibri/core/tasks/utils.py
```diff
@@ -206,3 +212,142 @@ def db_connection():
...
+
+    def update_progress(self, increment=1, message="", extra_data=None):
+        if self.progresstracker:
+            self.progresstracker.update_progress(increment, message)
+        if self.job:
+            self.job.update_progress(
+                self.job.progress + increment, self.job.total_progress
+            )
+            if extra_data and isinstance(extra_data, dict):
+                self.job.update_job_metadata(**extra_data)
```
vkWeb: Umm, I don't see a method `update_job_metadata` defined on `Job` class. Shouldn't we be calling `self.update_job_metadata(**extra_data)` here?

kolibri/core/tasks/utils.py
```diff
@@ -206,3 +212,142 @@ def db_connection():
...
+
+    def is_cancelled(self):
+        try:
+            self.check_for_cancel()
+            return False
+        except (UserCancelledError, KeyError):
+            return True
+
+    def cancel(self):
+        return self.check_for_cancel()
```
vkWeb: Shouldn't we be cancelling the job here? umm, am I missing something sir?

vkWeb (commented):
@rtibbles importing from another kolibri instance is working as expected on my manual test. 

Though the progress update percentage is being reported wrong: 

Screenshot from 2022-10-03 18-36-49

rtibbles:
That's quite a high percentage!

kolibri/core/tasks/utils.py
rtibbles: Ah, good catch - I think I meant to add it to the Job class, because it seemed like a useful thing to be able to set directly on the job, but then didn't...

kolibri/core/tasks/utils.py
rtibbles: This is copied from the original implementation, so I haven't actually changed anything here - but you're right the method name is misleading, what it's doing is reraising a handled cancel, rather than actually initiating a cancel.

kolibri/core/tasks/utils.py
vkWeb: Keeping that function at Job class makes sense to me sir

kolibri/core/tasks/utils.py
vkWeb: Yeap sir then we should cancel the job there.

vkWeb:
> That's quite a high percentage!

Very high indeed sir!

kolibri/core/content/public_api.py
```diff
@@ -0,0 +1,105 @@
...
+        )
+        node_ids = nodes.values_list("id", flat=True)
+        prerequisites = models.ContentNode.has_prerequisite.through.objects.filter(
+            from_contentnode_id__in=node_ids, to_contentnode_id__in=node_ids
+        )
+        related = models.ContentNode.related.through.objects.filter(
+            from_contentnode_id__in=node_ids, to_contentnode_id__in=node_ids
+        )
+        channel_metadata = models.ChannelMetadata.objects.filter(id=node.channel_id)
+
```
bjester, line 74: I doubt this would occur with a list of ancestor nodes, but do any of these `__in` filters in this method above need chunking for SQLite? Or is that handled at a global level-- I can't recall.

kolibri/core/content/public_api.py
```diff
@@ -0,0 +1,105 @@
...
+            # One example is for JSON field data that is stored as a string in the database,
+            # we want to avoid that being coerced to Python objects.
+            cursor.execute(*qs.query.sql_with_params())
+            data[qs.model._meta.db_table] = [
+                dict(zip(raw_fields, row)) for row in cursor
+            ]
+
+        data["schema_version"] = content_schema
+
+        return Response(data)
```
bjester, line 105: So we can mimic this endpoint on Studio to allow remote browsing?

kolibri/core/content/utils/channel_import.py
```diff
@@ -652,73 +702,40 @@ def can_use_sqlite_attach_method(self, model, table_mapper):
...
+                        ).format(
+                            channel_version=existing_channel["version"],
+                            channel_id=self.channel_id,
+                            new_channel_version=self.channel_version,
+                        )
+                    )
+                    return False
+                return True
+            elif existing_channel["version"] < self.channel_version:
                 # We have an older version of this channel, so let's clean out the old stuff first
```
bjester, line 740: Seems this should inherently and correctly handle the case where the previous import was a partial import?

bjester (commented):
Code review looks good! 

In testing, I got an error for a channel in which a new version was available (below), and I also saw the progress % issue @vkWeb noted. 

I clicked `View Changes` and got this
```
ERROR    Job 6c501be7ac2d4a25a2322a2b631858c7 raised an exception: Traceback (most recent call last):
  File "/kolibri/core/tasks/job.py", line 261, in execute
    result = func(*args, **kwargs)
  File "/kolibri/core/tasks/registry.py", line 221, in __call__
    return self.func(*args, **kwargs)
  File "/kolibri/core/content/tasks.py", line 500, in remotechanneldiffstats
    baseurl=baseurl,
  File "/kolibri/core/content/utils/upgrade.py", line 79, in diff_stats
    destination=destination_path,
  File "/kolibri/core/content/utils/channel_import.py", line 1150, in initialize_import_manager
    min_version = channel_metadata.get(
AttributeError: 'unicode' object has no attribute 'get'
```

kolibri/core/content/public_api.py
rtibbles: I think as `node_ids` has not been coerced to a list it should still be being passed as a subquery rather than as an explicit list that would need SQL variables for interpolation, so I think this is safe.

kolibri/core/content/public_api.py
rtibbles: This endpoint is only for import metadata, getting all relevant metadata for importing a single resource (rather than the entire channel database). There are some already implemented public API endpoints that would be needed in addition (a V2 public channel endpoint and several content node endpoints).

They are all exposed under a v2 content metadata router: https://github.com/learningequality/kolibri/blob/develop/kolibri/core/public/api_urls.py#L47

kolibri/core/tasks/utils.py
rtibbles: I think it's actually better to rename the method, as it is never used to initiate a cancel from within the task, only to reraise the cancellation once the task has finished its cleanup tasks.

kolibri/core/content/utils/channel_import.py
rtibbles: As it just deletes everything, it will handle the case when the previous import was partial or when it was complete. We add the check for a current partial import, because we can only do that in the case where the currently installed version of the channel is the same, otherwise we run the risk of mismatched/colliding tree metadata.

rtibbles:
Looks like I've messed up the argument handling for the diff file stats too, will fix!

rtibbles:
Have fixed the import diff stats issue, upgrading now works properly.

I did see a frontend issue during the upgrade process, but the backend experienced no errors. Not sure if the issue is also extant on develop, as I've not touched the frontend here.

Progress updating is now working properly again - I had simply forgotten to set the total progress at the start!
image

bjester:
Upgrading a channel, I got the following frontend error:
```
vue.runtime.esm.js:1897 TypeError: Cannot read properties of undefined (reading 'id')
    at NewChannelVersionPage.vue:237:1
logError @ vue.runtime.esm.js:1897
```

Also, does this mean a resource was a duplicate or that it failed to import?
Screenshot from 2022-12-05 11-52-39

kolibri/core/content/utils/resource_import.py
```diff
@@ -0,0 +1,513 @@
...
+            if isinstance(manifest_file, string_types):
+                manifest_file = open(manifest_file, "r")
+            content_manifest = ContentManifest()
+            try:
+                content_manifest.read_file(manifest_file)
+                node_ids = content_manifest.get_node_ids_for_channel(channel_id)
+                exclude_node_ids = None
+            except ContentManifestParseError:
+                pass
+            manifest_file.close()
```
bjester: I think this manifest parsing would be better placed in a factory function that builds a class instance from it, instead of having it in the class constructor. The reason is that it takes precedence on determining the values of `node_ids` and `exclude_node_ids`, but if it were to fail, it makes that codepath a bit ambiguous on the intended outcome since it just suppresses the parse error.

I see an inherited class is also utilizing this argument, so a static class method might be best for the factory. I would expect that the factory function wouldn't accept `node_ids` and `exclude_node_ids` arguments, thereby making the API clear that it's one or the other.

rtibbles:
> Upgrading a channel, I got the following frontend error:

I got the same issue here, it seemed like a frontend only issue, but can double check this and fix (I think it might be pre-existing in develop but worth fixing here).

> Also, does this mean a resource was a duplicate or that it failed to import?

Yeah, this is a pre-existing issue with resource counting being deduped in one place and not in another. I'll fix it in this PR in case it doesn't get fixed in 0.15.

rtibbles:
Updated - add a `from_manifest` class method to handle the manifest construction and make the keyword arguments explicit.

Cleaned up various refactor induced bugs in the channel update workflow. Should be good to go!

bjester (approved):
LGTM! :100:

PR #9740, opened 2022-09-26
changed _init_.py  from 5 tuple to 3.

### LINKED ISSUE #3251: make __init__.py's version a 3-tuple
### Observed behavior

Currently, `__init__.py` looks like this:

```python
VERSION = (0, 8, 0, 'alpha', 0)
```

The fact that the fourth value (`'alpha'`/`'beta'`/`'final'`) needs to be updated makes creating releases and managing release branches very frustrating:

* we need to do a whole PR+review+merge to set it to `'final'`, create a release tag, then do _another_ PR+review+merge to set it back to either `'alpha'` or `'beta'`
* this creates repeated merge conflicts when syncing old releases back to `develop`, or even worse across multiple release branches
* there doesn't seem to be any difference between `'beta'` and `'alpha'`
* the fifth value (`0`) is completely ignored
* need to create an `alpha0` tag on `develop` after cutting a new beta or we get the confusing error `You have added a final tag without bumping kolibri.VERISON`

### Expected behavior

Since the alpha/beta/final status is reflected in git tags and not in source code, we shouldn't need to update these values in source. Instead, we can simply do:

```python
#: This string does not reflect alpha/beta/final status and is modified by get_version()
#: Use ``kolibri.__version__`` for the exact version string.
VERSION = (0, 8, 0)
```

Then the final/beta/alpha status can be driven by git tags and the process for releasing will be greatly simplified

Final releases should be embedded in the code. Everything else can come from the git tag. Anything interim can be versioned with a git hash, because finding parents is unreliable.

If a tag is made (alpha, beta, rc, or final) that does not match the 3-tuple in `__init__.py`, then the code should throw an error, to prevent tags being made on the wrong branch or code.

### User-facing consequences

No user-facing consequences, but this would improve developer experience for creating new releases and managing release branches.

### Steps to reproduce

Go through the process of creating a patch release and then syncing those changes back to `develop`

AllanOXDi:
This code change  is design to  maka __init__.py's version a 3-tuple
fixes #3251

## Summary

…

## References
fixes #3251

…

## Reviewer guidance

…

----

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/utils/tests/test_version.py
+++ b/kolibri/utils/tests/test_version.py
@@ -28,7 +28,10 @@ class TestKolibriVersion(unittest.TestCase):
 
     @mock.patch("kolibri.utils.version.get_git_describe", return_value=None)
     @mock.patch("kolibri.utils.version.get_version_file", return_value=None)
-    def test_no_tag_no_file_version(self, file_mock, describe_mock):
+    @mock.patch("kolibri.utils.version.get_git_changeset", return_value=None)
+    def test_no_tag_no_file_version_no_git_changeset(
+        self, get_git_changeset_mock, file_mock, describe_mock
+    ):
         """
         Test that when doing something with a 0th alpha doesn't provoke any
         hickups with ``git describe --tag``.
@@ -36,7 +39,24 @@ class TestKolibriVersion(unittest.TestCase):
         then get_prerelease_version should return 0.15.8.dev0
         """
         v = get_version((0, 1, 0))
-        self.assertIn("0.1.0.dev0", v)
+        self.assertEqual("0.1.0.dev0", v)
+
+    @mock.patch("kolibri.utils.version.get_git_describe", return_value=None)
+    @mock.patch("kolibri.utils.version.get_version_file", return_value=None)
+    @mock.patch(
+        "kolibri.utils.version.get_git_changeset", return_value="+git.1234567890"
+    )
+    def test_no_tag_no_file_version_git_changeset(
+        self, get_git_changeset_mock, file_mock, describe_mock
+    ):
+        """
+        Test that when doing something with a 0th alpha doesn't provoke any
+        hickups with ``git describe --tag``.
+        If the version file returns nothing, and get_git_describe returns nothing,
+        then get_prerelease_version should return 0.15.8.dev0
+        """
+        v = get_version((0, 1, 0))
+        self.assertEqual("0.1.0.dev0+git.1234567890", v)
 
     @mock.patch("kolibri.utils.version.get_git_describe", return_value="v0.1.0-alpha1")
     @mock.patch("kolibri.utils.version.get_version_file", return_value=None)
--- a/kolibri/utils/tests/test_version.py
+++ b/kolibri/utils/tests/test_version.py
@@ -18,10 +18,6 @@ from kolibri.utils import version
 get_version = version.get_version.__wrapped__  # @UndefinedVariable
 
 
-def dont_call_me_maybe(msg):
-    raise AssertionError(msg)
-
-
 class TestKolibriVersion(unittest.TestCase):
     def test_version(self):
         """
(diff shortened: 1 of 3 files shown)


kolibri/utils/version.py
```diff
@@ -308,37 +301,30 @@ def get_prerelease_version(version):
...
+                    "inconsistent with kolibri.__version__."
+                    "__version__ is: {}, tag says: {}."
+                ).format(str(version), git_version)
             )
+        if git_version[:3] == version[:3]:
+            if git_version[3] == "final":
+                return major
+        # If the tag was of a final version, we will use it.
 
         return (
```
rtibbles: I think this `return` statement still need to be indented to only run `if git_version[:3] == version[:3]:`

rtibbles:
Looking at the built asset in buildkite (https://buildkite.com/learningequality/kolibri-python-package/builds/6409#0183adc3-aa6b-4b79-a9d7-59b2ffdc1951) it seems that we're not giving a very informative version in this case: `0.15.8.dev0` - it's not wrong, but would be good to retain some of the previous behaviour.

I'm not entirely sure where the change would need to be applied though. I'll try to write a test case to cover what is happening here so we can fix it.

kolibri/utils/tests/test_version.py
```diff
@@ -32,46 +28,56 @@ def test_version(self):
...
-    def test_alpha_0_version(self, file_mock, describe_mock):
+    @mock.patch("kolibri.utils.version.get_git_changeset", return_value=None)
+    def test_no_tag_no_file_version_no_git_changeset(
+        self, get_git_changeset_mock, file_mock, describe_mock
+    ):
         """
         Test that when doing something with a 0th alpha doesn't provoke any
         hickups with ``git describe --tag``.
+        If the version file returns nothing, and get_git_describe returns nothing,
+        then get_prerelease_version should return 0.15.8.dev0
```
marcellamaki: A bit of a nitpick, but I think this should be more general, rather than a specific version. i.e. explaining - what is it exactly that 0.15.8 is an example of? where is this value coming from?

kolibri/utils/tests/test_version.py
```diff
@@ -32,46 +28,56 @@ def test_version(self):
...
+        "kolibri.utils.version.get_git_changeset", return_value="+git.1234567890"
+    )
+    def test_no_tag_no_file_version_git_changeset(
+        self, get_git_changeset_mock, file_mock, describe_mock
+    ):
+        """
+        Test that when doing something with a 0th alpha doesn't provoke any
+        hickups with ``git describe --tag``.
+        If the version file returns nothing, and get_git_describe returns nothing,
+        then get_prerelease_version should return 0.15.8.dev0
```
marcellamaki: and same here

kolibri/utils/tests/test_version.py
```diff
@@ -215,103 +177,104 @@ def test_version_file_final(self, get_git_changeset_mock, describe_mock, file_mo
...
         popen_mock.return_value = process_mock
-        assert get_version((0, 1, 0, "alpha", 1)) == "0.1.0b1.dev0+git.123.abcdfe12"
+        assert get_version((0, 1, 0)) == "0.1.0b1.dev0+git.123.abcdfe12"
 
     @mock.patch("subprocess.Popen")
     @mock.patch("kolibri.utils.version.get_version_file", return_value=None)
     def test_git_random_tag(self, file_mock, popen_mock):
         """
         Test that we don't fail if some random tag appears
+        Always fallback to .dev0 if the tag data is unparseable.
```
marcellamaki: why are we doing this?

kolibri/utils/tests/test_version.py
```diff
@@ -143,69 +129,45 @@ def test_alpha_0_consistent_version_file(
...
     @mock.patch("kolibri.utils.version.get_version_file", return_value="0.1.0b1")
     @mock.patch("kolibri.utils.version.get_git_describe", return_value="v0.0.1")
     @mock.patch("kolibri.utils.version.get_git_changeset", return_value="+git123")
-    def test_version_file_ignored(
+    def test_version_file_overrides(
         self, get_git_changeset_mock, describe_mock, file_mock
     ):
         """
-        Test that the VERSION file is NOT used where git data is available
+        Test that the VERSION file is used when git data is available
```
marcellamaki, line 173: Old:
> Test that the VERSION file is NOT used where git data is available

New: 
> Test that the VERSION file is used when git data is available

So, essentially now it is the opposite? Just making sure that is the correct change

marcellamaki (commented):
This looks good. I added a few very small comments of places where additional detail might be helpful. 

One additional idea is adding a descriptive comment of a few sentences that explains overall what is happening. Maybe at the top of version.py? While yes, this is comprehensible through the tests and the comments there, I find it harder to wrap my mind around everything by going through the conditions one my one in the test file. I have to hold a lot of things in my brain at once over the course of a very long file. A short explanation of how this works at a high level would be helpful for other devs (and future Marcella who understands this now but will undoubtable forget it by January). Even if it doesn't account for every condition an edge case, I think the basics would be sufficient.

kolibri/utils/tests/test_version.py
rtibbles: Yes, previously, we deferred to git data over the VERSION file, now we are doing the opposite (this means we don't ever try to do git things in built environments, but has the effect that in dev environments if you have a VERSION file, then the file will take precedence - but if you're generating a VERSION file you should know what you're doing :) )

marcellamaki (approved):
Looks good to me! Thanks @AllanOXDi and @rtibbles!

PR #9760, opened 2022-10-04
Created screen to do the actual merge

### LINKED ISSUE #9321: "Default Path" for facility change UI is implemented
This is a final issue to ensure that after the state machine, routes, and pages have been created, that the "Default path" flow works as expected. 

Any code adjustments necessary to ensure that the feature is completed is considered in scope for this issue.

### LINKED ISSUE #9725: Change facility: Set a new super admin
In https://github.com/learningequality/kolibri/pull/9715, _Choose a new super admin_ page and related state machine updates have been implemented. When a user chooses a new super admin on that page, we decided that we will only save the new super admin's ID to the machine's context `newSuperAdminId` attribute but won't actually set the new super admin yet on the source facility. That should rather happen together with all other updates related to changing facility during the syncing step (`syncChangeFacilityState`). This issue is here so that we don't forget this needs to be implemented in `syncChangeFacilityState`.

**Background**
This is a general approach we have adopted for all updates to be done when changing facilities for the sake of simplicity since we will know that all updates will happen during a single step, and also to prevent unnecessary actions or inconsistencies when a user changes their mind, go back in the flow, etc.

jredrejo:
## Summary
Implements the last screen to finish the merge process

bueno2

@jtamiace  After implementing it, there are a few glitches that have not been solved and require some design review:
- The task that do the actual merge and syncs is composed of several tasks and its update on the percentage is changing, goes to 100% and then go back to 50% to do the next one and so on. Changing this behaviour is quite challenging and I'd recomment to create a separate issue if it's needed, instead of blocking this merge feature.
- For the same reason explained above, steps go to 7, and then back to 1, and so on. I think in this case removing the numbers for the steps would be a better solution, but I leave it in your hands.

## References
Closes: #9321 , #9334, #9725

## Reviewer guidance
With this final screen, the [whole flow defined in the design](https://www.figma.com/file/MpCG7r5PULOCJsEvnTSti8/Android-Designs-2022?node-id=194%3A2920) is complete.
To test it, two different kolibri installations are needed, both using the current develop version.
Also to be in context: the common use case for this will be an user who has kolibri installed (usually in a tablet o mobile device) and wants to join to a facility in a school. So two kolibri servers are needed: One kolibri server running the `target` facility where the user will merge his current profile and the kolibri server where these screens will be tested. Both servers must be running the latest code from `develop` (or a kolibri version >= 0.16 after it's released).

Also, the kolibri device that's going to transfer its user must have been setup as on_my_own_setup. To do it manually, the database must be modified, using `kolibri shell` (you can write `exit` and enter to finish the shell after introducing the next lines):
```
f=Facility.get_default_facility()
f.on_my_own_setup = True
f.save()
```
The same thing can also being opening the db.sqlite3 file, and adding `{"on_my_own_setup": true}` to the `extra_field` of the existing row in `kolibriauth_facility_dataset` table

After these conditions are fulfilled, transfering the user to the other device facility can be done. There are multiple possibilities and all of them (and its combinations) should be tested:

- The username in the current facility exist in the target facility
- The username wants to be merge with its same username or with a different one
- The username is the only superuser in the current facility (or not)

After the process finishes, the current device must have two facilities: the old one, without the user who ran the process, and the  new one with the user transfered there. The target device must have the user in its facility, created if it didn't exist previously, with the data merged from the other one if it existed previously.

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated


jtamiace:
The progress bar and messages are a little misleading about when the merge process is done, but shouldn't be a blocker to merge this PR. It'd be good to revisit this as a separate issue. I'd also like to see what other options can be before doing something like removing the numbers of the steps.

Redirecting to be signed in was intended to be a shortcut, but isn't essential to this experience. It'd be good to make this a follow up issue.

jtamiace (approved):

jredrejo:
Hi @jtamiace 
> The progress bar and messages are a little misleading about when the merge process is done, but shouldn't be a blocker to merge this PR. It'd be good to revisit this as a separate issue. I'd also like to see what other options can be before doing something like removing the numbers of the steps.

Removing the numbers of the steps can be done quite easily. I've done it and updated the screencast in the description of this PR.

> 
> Redirecting to be signed in was intended to be a shortcut, but isn't essential to this experience. It'd be good to make this a follow up issue.

Finally, after discussing the alternatives and security concerns with @rtibbles I have done an implementation for it that it's working, so that's one less thing to worry about.

pcenov:
Hi @jredrejo,
I'm going to post my observations here as a list of items so that you can quickly scan through those and let me know if any of these don't make sense to you or are out of scope for this PR. I will then create detailed follow-up issues for the relevant to this PR problems.

Issues:

1. Missing ‘Don’t have user’s credentials? Use an admin account’ option at the first step of the Merge accounts wizard
2. The ‘Add new address’ option is not working
3. Merge accounts without knowing the password (by using an Admin account) is not working.
4. Merge accounts when the user is the only super admin is not working - getting an 'Integrity error' in the console
5. If a learner doesn’t exist on the target facility then merging accounts by selecting the ‘Change facility’ option is not working
6. There is a lag of several seconds before the radio button of the target facility gets selected at ‘Select learning facility’
7. If I merge accounts by selecting the ‘Create new account’ option the created user is without a Full name as this field is missing in the ‘Create new account’ form
8. At the ‘Choose a new super admin’ screen you are showing all of the available users including the ones who are already transferred to the target facility

Questions:

1. Why is the account of a merged user still shown in Device > Permissions of the original facility, while it is completely removed from Facility > Users?
2. Syncing between the facilities is not working at all either through Facilities > Sync or by the Learn-only type of sync - is this a work in progress or an omission?
3. The design is not OK when compared to the specs in Figma - is this going to be implemented by someone else?

jredrejo:
> Hi @jredrejo, I'm going to post my observations here as a list of items so that you can quickly scan through those and let me know if any of these don't make sense to you or are out of scope for this PR. I will then create detailed follow-up issues for the relevant to this PR problems.
> 


Hello @pcenov , thanks for the exhaustive review. I will comment in line to your items but I think it would be good if we could coordinate to have a videoconference because there are some points I don't fully understand or I can't reproduce.

> Issues:
> 
> 1. Missing ‘Don’t have user’s credentials? Use an admin account’ option at the first step of the Merge accounts wizard

This has been removed from the design after talking to @jtamiace when that screen was implemented. It's redundant because it appears in the next screen. I see it has not been deleted from the Figma design, but it's fine not having it there.

> 2. The ‘Add new address’ option is not working

Yep , I saw and fixed it

> 3. Merge accounts without knowing the password (by using an Admin account) is not working.

the using an Admin account if what I don't get, let's see it 'in person' if possible

> 4. Merge accounts when the user is the only super admin is not working - getting an 'Integrity error' in the console

There are some things I don't have clear in this point. If the username is also an admin in the target facility, this should not work and provide an error. That was not included in the design, but I have added it after seeing your review. The username in the target facility must be a learner. If not, it will appear a button to retry and begin again the process.

> 5. If a learner doesn’t exist on the target facility then merging accounts by selecting the ‘Change facility’ option is not working

I am also loss in this point, I don't know exactly in what screen you are.

> 6. There is a lag of several seconds before the radio button of the target facility gets selected at ‘Select learning facility’

If you select it manually, it's instant, but if you want to wait for kolibri to select it, yes, it takes some seconds because it has to wait until the full scan of the network looking for other kolibri instances is finished. It does the automatic selection when the list is completed. I don't see any problem (and I think that's how it works in other places of the app using this UI, but we can discuss and change it if needed)

> 7. If I merge accounts by selecting the ‘Create new account’ option the created user is without a Full name as this field is missing in the ‘Create new account’ form

If I am not lost again, that's how the design states it: only username and password are required.

> 8. At the ‘Choose a new super admin’ screen you are showing all of the available users including the ones who are already transferred to the target facility

can you tell me how to reproduce this one? what steps did you take? 


> 
> Questions:
> 
> 1. Why is the account of a merged user still shown in Device > Permissions of the original facility, while it is completely removed from Facility > Users?

I'm not sure if I understand this point either: the user is not removed from the device, is moved to the target facility that now it's in the device, so it'll still appear in the device and in the "target facility"->users

> 2. Syncing between the facilities is not working at all either through Facilities > Sync or by the Learn-only type of sync - is this a work in progress or an omission?

That was not part of the scope of this project. It should be working if it works in kolibri develop, if it does not, it's probably a general issue in kolibri develop. It'd be great if you can confirm if it's not working only after executing this task or it's not working at any case.

> 3. The design is not OK when compared to the specs in Figma - is this going to be implemented by someone else?

it should be, can you specify what parts are not OK to fix them?

pcenov:
Hi @jredrejo as discussed I am listing here the outstanding issues after our discussion:

| Issue| Screenshot |
| ‘Add new address’ - The modal window doesn’t close after successfully adding an address and the facility is not listed | https://user-images.githubusercontent.com/79847249/195650174-89dd4f32-7d6b-4821-a7c0-97781195380f.mp4 |
| The progress bar remains at 50% after successfully merging accounts | https://user-images.githubusercontent.com/79847249/195650323-1a5a4a57-1d30-489a-b554-a99fc76c7b7a.mp4 |
| The Default path is not working (the user is getting into an infinite loop after repeatedly clicking the Continue button) | https://user-images.githubusercontent.com/79847249/195650477-0d28c759-0013-4ec8-993a-9c65e406eaae.mp4 |
| Radio buttons lag - Consider showing the first displayed radio button as selected by default | https://user-images.githubusercontent.com/79847249/195650574-7c195faf-4f13-403a-82b5-b28615c2f064.mp4 |
| Device > Permissions - The merged users are shown in the Device permissions table and it should be further discussed whether that should be the case as it can be a security issue |  |
| Merge accounts without knowing the password and therefore using an Admin or a Super admin account is not working | https://user-images.githubusercontent.com/79847249/195651792-cd91e8ca-07d8-44c2-861d-c4c1cb7a6c8b.mp4 |
| Incorrect message after an attempt to create a user with an existing username | |

kolibri/plugins/user_profile/assets/src/views/ChangeFacility/MergeFacility.vue
```diff
@@ -0,0 +1,353 @@
+<template>
+
+  <div>
+    <h1>{{ $tr('documentTitle') }}</h1>
+    <div class="task-panel" :class="{ 'task-panel-sm': windowIsSmall }">
+      <div class="icon">
+        <transition v-if="!taskError" mode="out-in">
```
marcellamaki, line 7: Two years working on Kolibri and I'd never noticed this `mode="out-in"` code pattern before! Learn something new every day 😁

kolibri/plugins/user_profile/assets/src/views/ChangeFacility/MergeFacility.vue
```diff
@@ -0,0 +1,353 @@
...
+        }).then(() => {
+          redirectBrowser(urls['kolibri:kolibri.plugins.learn:learn']());
+        });
+      }
+
+      function to_retry() {
+        if (taskId.value !== null) {
+          TaskResource.clear(taskId.value);
+        }
+        changeFacilityService.send('TASKERROR');
```
marcellamaki, line 243: This is just because I don't really understand much of our task management, but just as reading through this, I find it a little bit confusing. I read this as "If there is a task Id, clear it. In all cases, send a task error." I don't fully understand this allow the task to restart. Does it somehow put items back into a queue?

marcellamaki: actually, is it that it just updates the state machine, not that it is restarting a specific task?

marcellamaki (commented):
Hi @jredrejo - I did a code read through that focused mainly on the state machine and vue components. I added a few comments but nothing blocking! 

While I found the python readable, and I was able to follow along and understand what was happening, I would appreciate if someone else with more python skill could do a review of that, and give you more helpful feedback. 

I will be happy to do a re-review if you make any front end changes follow QA feedback, but it looks really good. Thank you for your hard work on this :)

kolibri/plugins/user_profile/assets/src/views/ChangeFacility/MergeFacility.vue
jredrejo: Actually, you are right in all your comments: it updates the state machine but, independently, it also clears the existing task if it has failed.  The code uses kind of a Singleton pattern so only one task can be running (for the case two different browsers point to the same url). So, clearing the task is needed to be able to start it again after a failure.

kolibri/plugins/user_profile/assets/src/views/ChangeFacility/MergeFacility.vue
jredrejo: It's documented at https://v2.vuejs.org/v2/guide/transitions.html#Transition-Modes  . I learned it from Sarah Drasner it in 2017 when doing my first vue course https://frontendmasters.com/courses/vue/ that's now a bit oudated but if you like animations, she's the best, you should not miss https://frontendmasters.com/courses/svg-essentials-animation/

kolibri/plugins/user_profile/tasks.py
```diff
@@ -82,5 +95,13 @@ def mergeuser(command, **kwargs):
 
     # Resync with the server to update the merged records
     call_command("sync", **kwargs)
+    new_superuser_id = kwargs.get("new_superuser_id")
+    if new_superuser_id:
+        new_superuser = FacilityUser.objects.get(id=new_superuser_id)
+        # make the user a new super user for this device:
+        new_superuser.facility.add_role(new_superuser, role_kinds.ADMIN)
```
rtibbles, line 99: I wonder if this is necessary? Just the device permissions might be enough for the local device, and we avoid having another mechanism for adding an admin role to a facility (which could be a security risk). I don't think this is a blocker, but just wondering if we can avoid it.

kolibri/plugins/user_profile/viewsets.py
```diff
@@ -70,3 +73,20 @@ def post(self, request, *args, **kwargs):
...
+    """
+    Viewset to login into kolibri using the merged user,
+    after the old user has been deleted
+    """
+
+    def post(self, request):
+        username = request.data.get("username", None)
+        facility = request.data.get("facility", None)
+        token = request.data.get("token", None)
+        if not TokenGenerator().check_token(username, token):
```
rtibbles: Could do the lookup before the token check to still use the user pk like the base implementation does.

The advantage here is that then it's a token for the uuid rather than the username, so it's completely unique to the user (whereas by username is more generic).

kolibri/plugins/user_profile/tasks.py
```diff
@@ -29,6 +35,13 @@ def validate(self, data):
             job_data = super(MergeUserValidator, self).validate(data)
 
         job_data["kwargs"]["local_user_id"] = data["local_user_id"].id
+        if data.get("new_superuser_id"):
+            job_data["kwargs"]["new_superuser_id"] = data["new_superuser_id"].id
+
+        # create token to validate user in the new facility
+        # after it's deleted in the current facility:
+        token = TokenGenerator().make_token(data["username"])
```
rtibbles: Can generate this after the sync has completed and add to the job metadata from inside the task. This will reduce the chance that the token expires after a slow sync, and means that we could use the id of the remote_user, and avoid needing to override to use the username.

rtibbles (commented):
A few questions - I think using the PK to make the token is probably preferable, both for the uniqueness of the token, and to do it after the task has completed so that the token is less likely to time out.

jredrejo:
hello @pcenov , please when you have a chance, review again the issues you found. I have been able to reproduce and fix most of them. But there are some cases I have not been able to do it (main one is the problem adding an external server, I could do it without any problem).
In case any of the scenarios fails, please, take a look at the console, I have added some error logging information that will help to reproduce it.

kolibri/plugins/user_profile/tasks.py
jredrejo: I am not sure this is needed either, but was a core part of the Figma design. Happy to remove the screen to select the new super user and this code if it's not really needed, but better to confirm it. I think it will be needed for the user to be able to add new content if the device is not an "Learner use only" device, aka _SOUD_

kolibri/plugins/user_profile/viewsets.py
jredrejo: good idea, done!

kolibri/plugins/user_profile/tasks.py
jredrejo: yep, thanks for the suggestion. Done too.

pcenov:
Hi @jredrejo after additional testing the following issues are still valid:

| Issue| Screenshot |
| The Default path is not working (the user is getting into an infinite loop after repeatedly clicking the Continue button) | https://user-images.githubusercontent.com/79847249/195650477-0d28c759-0013-4ec8-993a-9c65e406eaae.mp4 |
| Radio buttons lag - Consider showing the first displayed radio button as selected by default | https://user-images.githubusercontent.com/79847249/195650574-7c195faf-4f13-403a-82b5-b28615c2f064.mp4 |
| Device > Permissions - The merged users are shown in the Device permissions table and it should be further discussed whether that should be the case as it can be a security issue |  |
| Merge accounts without knowing the password - Still not working when I am using an Admin account, fixed for super admin |   |
| Incorrect message after an attempt to create a user with an existing username - my user is a learner | |

I was also not able to replicate the issue with adding an external server, but found the following related issues:

| Issue| Screenshot |
| If I attempt to add the address for a 0.15 facility I don't get notified that this is incompatible, I simply see the toast notification that it was added successfully | https://user-images.githubusercontent.com/79847249/197217410-c27a29c7-31da-4b5b-aabe-3f65361e88f3.mp4 |
| If the connection to a facility gets lost the user is not notified in any way. To replicate stop the server of the second facility while being at the  'Select learning facility step' and attempt to proceed with the flow | https://user-images.githubusercontent.com/79847249/197218166-e2820279-67ad-4686-90a7-e71920bcd214.mp4 |

jredrejo:
Hello @pcenov , I'll try to respond in line
> Hi @jredrejo after additional testing the following issues are still valid:
> 
> Issue	Screenshot
> The Default path is not working (the user is getting into an infinite loop after repeatedly clicking the Continue button)	https://user-images.githubusercontent.com/79847249/195650477-0d28c759-0013-4ec8-993a-9c65e406eaae.mp4

I can't reproduce t his one, please let me know the error you see in the console. it should be a text message  like:
`-- error starting task: +` an object error that contains the important information.


> :-------------------------:	:-------------------------:
> Radio buttons lag - Consider showing the first displayed radio button as selected by default	https://user-images.githubusercontent.com/79847249/195650574-7c195faf-4f13-403a-82b5-b28615c2f064.mp4

I have double-checked this and the lag can't be skipped. It's due to the fact that the user server must scan the whole network + contact each of the servers to retrieve their permissions, and that takes time that can not be avoided. We'll have to live with this lag.


> :-------------------------:	:-------------------------:
> Device > Permissions - The merged users are shown in the Device permissions table and it should be further discussed whether that should be the case as it can be a security issue	

Please, fill a new issue for this, I think it's a general problem that is related to any syncing process and I don't think it should be blocking this issue. Let's work on it in a separate discussion.

> :-------------------------:	:-------------------------:
> Merge accounts without knowing the password - Still not working when I am using an Admin account, fixed for super admin	

Good catch! I had been testing it with a superadmin. I have fixed it now.


> :-------------------------:	:-------------------------:
> Incorrect message after an attempt to create a user with an existing username - my user is a learner	
> :-------------------------:	:-------------------------:

Fixed!


> I was also not able to replicate the issue with adding an external server, but found the following related issues:
> 
> Issue	Screenshot
> If I attempt to add the address for a 0.15 facility I don't get notified that this is incompatible, I simply see the toast notification that it was added successfully	https://user-images.githubusercontent.com/79847249/197217410-c27a29c7-31da-4b5b-aabe-3f65361e88f3.mp4

I still think something odd happens in your network setup with the server you're using, probably it's some firewall issue because it should be automatically added to the list. I have reproduced this adding a server that's in Internet, so it's not automatically detected. The server is added correctly to the list and appears correctly greyed because it's a 0.15.x version. Please, test it with kolibri-demo.learningequality.org to check it.


> :-------------------------:	:-------------------------:
> If the connection to a facility gets lost the user is not notified in any way. To replicate stop the server of the second facility while being at the 'Select learning facility step' and attempt to proceed with the flow	https://user-images.githubusercontent.com/79847249/197218166-e2820279-67ad-4686-90a7-e71920bcd214.mp4

I have added a new error message for this case, and the button says "Retry" for the user to retry (sorry, I had my browser in Spanish and some texts appear translated in this screenshot)
image

pcenov:
Hi @jredrejo I have verified the resolved issues and reported separately all the outstanding issues that I was able to identify in the latest build.

About the issue with adding a new address - it seems that this is functioning correctly but as a user I don't have a way of knowing that when my target facility is displayed grayed out it is displayed as such because it is a 0.15 facility. So when I attempt to manually add the address then I get into a case where it says that the facility is added while it isn't and this is confusing. This is an edge case which is not specified anywhere so I'm not sure how we should proceed with it.

About the problem with the transferred users being displayed in the Device > Permissions table - I am actually wondering whether it is a valid scenario as it only happens now because we are simulating the scenario. This should not be possible for a device created through the normal Own my own setup path, right?

radinamatic (approved):
Great job @jredrejo, and thank you @pcenov for the thorough testing! 👏🏽 👏🏽 👏🏽 

We have a few follow-up issues to work on, but this is in a good place to merge 💯 :shipit:

jredrejo:
> Hi @jredrejo I have verified the resolved issues and reported separately all the outstanding issues that I was able to identify in the latest build.
> 

hi @pcenov 

> About the issue with adding a new address - it seems that this is functioning correctly but as a user I don't have a way of knowing that when my target facility is displayed grayed out it is displayed as such because it is a 0.15 facility. So when I attempt to manually add the address then I get into a case where it says that the facility is added while it isn't and this is confusing. This is an edge case which is not specified anywhere so I'm not sure how we should proceed with it.
> 

mmm, when it's added manually, if it's a 0.15 facility, it appears, but greyed, exactly like when it's detected automatically. At least that's what I see when adding the kolibri-demo.learningequality.org site. Did you try with that server? 

> About the problem with the transferred users being displayed in the Device > Permissions table - I am actually wondering whether it is a valid scenario as it only happens now because we are simulating the scenario. This should not be possible for a device created through the normal Own my own setup path, right?

Honestly, I'm not sure, I've lost a bit the global scenario with so many paths and alternatives. I think it is harmless to create a follow up issue, just in case, so we can check it later and close it if it's an impossible situation, or fix it if not.

pcenov:
Hi @jredrejo,  I've tried importing the demo site - yes it's grayed out as you are saying. But what I am flagging as a concern is that since I am assuming that this flow will be used by Learners who are not necessarily technical it might not be crystal clear to them why it is grayed out. Also if I attempt to add it again, it just says that the address was added successfully. And lastly but not least - it's not displayed with the name value which I've provided in the form.

This is very confusing from a user's point of view. @radinamatic what do you think?

https://user-images.githubusercontent.com/79847249/198076838-ae9703a5-35e7-4ba4-926c-d0ea0a14bb5d.mp4

PR #9771, opened 2022-10-13
Bump version string

marcellamaki:


radinamatic (approved):
So much to review, I hardly had the time... 😅

rtibbles (dismissed):
Let's hold off on merging this until https://github.com/learningequality/kolibri/pull/9740 is merged. Then this will only need to update the patch number.

PR #9439, opened 2022-05-16
Push PDF pages rendering below full screen bar

marcellamaki:
## Summary
Updates the PDF renderer index so that the full screen bar does not cover the top of the first page of the PDF.

## References
Fixes #8745

Screen Shot 2022-05-16 at 9 41 01 AM

## Reviewer guidance
Open a PDF in the content renderer. 

Is this spacing okay, or should there be an additional scrolling/height calculation so the page moves up when the bar disappears?

----

## Testing checklist

- [ ] Contributor has fully tested the PR manually
- [ ] If there are any front-end changes, before/after screenshots are included
- [ ] Critical user journeys are covered by Gherkin stories
- [ ] Critical and brittle code paths are covered by unit tests

## PR process

- [ ] PR has the correct target branch and milestone
- [ ] PR has 'needs review' or 'work-in-progress' label
- [ ] If PR is ready for review, a reviewer has been added. (Don't use 'Assignees')
- [ ] If this is an important user-facing change, PR or related issue has a 'changelog' label
- [ ] If this includes an internal dependency change, a link to the diff is provided

## Reviewer checklist

- Automated test coverage is satisfactory
- PR is fully functional
- PR has been tested for [accessibility regressions](http://kolibri-dev.readthedocs.io/en/develop/manual_testing.html#accessibility-a11y-testing)
- External dependency files were updated if necessary (`yarn` and `pip`)
- Documentation is updated
- Contributor is in AUTHORS.md

Issue #9348, opened 2022-04-22
The resource completed dialog appears while I am still watching a video


dylanmccall:
## Observed behavior

Kolibri understands whether a user has watched an entire video based on how much time they have spent watching it, rather than whether they have reached the end. This means if I seek backwards in a video, Kolibri will believe I have finished the video before I have actually reached the end. That is probably an okay compromise, but it leads to the "Resource completed" dialog blocking the video.

## Expected behavior

Instead, Kolibri should wait until the video is _finished_ before showing the "Resource completed" dialog.

## Steps to reproduce

- Choose a video to watch.
- Watch most of the video.
- Rewind to the beginning.
- Continue watching the video.
- The "Resource completed" dialog will appear before you are finished watching.

![Screenshot from 2022-04-22 16-35-04](https://user-images.githubusercontent.com/132063/164817579-e7fc9196-b816-404e-bb75-afac32c3ccd1.png)

## Context

I reproduced this using Kolibri 0.15.1 with the Kolbiri app for GNOME (https://flathub.org/apps/details/org.learningequality.Kolibri).

rtibbles:
This is intended behaviour - I am disinclined to change it, especially as it would require disambiguating between 'finishing' and 'finishing' a resource.

What it does perhaps point to is an issue we have faced before with this kind of modal pop-up, that in some cases not having it is highly detrimental, because learners never move on, and in other cases, it is a pain because it is interfering with a learner's intent to remain within the resource.

At the moment we have managed that with the 'stay here' button which dismisses the modal - there are possibly things that we can do here, but I'd like to hear more direct feedback from users around the interaction.

MisRob:
I saw this too while working on the completion dialog and in some cases, it felt confusing and unpleasant disruption of user experience to me. For example, when you watch half of a video, leave, return to the same resource later and start watching it again. If my memory is correct, in such cases the completion dialog would be displayed when I was in the half of watching. I also reported an issue related to this here https://github.com/learningequality/kolibri/issues/8309#issuecomment-919864207, describing similar problems when rewinding videos.

I think this could be solved by improving the condition for considering a video as completed (or at least the condition for showing the completed dialog) to:

1. Time spent watching is equal to video length
AND
2. Was the end of the video reached?

Something like that could also have another benefit that we'd be more sure that someone has really finished the video (rather than watching the first half of it two times)

@rtibbles I don't think this would cause any problems in the areas you described, or am I missing something here?

[...]

PR #9776, opened 2022-10-14
Fix PDF completion issues

### LINKED ISSUE #8868: PDF resource completion

rtibbles:
## Summary
* Removes debounce of all handling of the `RecycleList` `update` event, and instead just debounces the PDF page rendering which is the expensive call.
* This allows the currently scrolled to position to be more frequently updated, preventing the error that would occur where visiting entire pages would fail to get recorded.
* Generalizes the previous fix for 2 page PDFs to the case where the `scrollTop` is still on the second to last page, but the browser cannot scroll any further. Does this by checking to see if we are on the second to last page, but the browser viewport cannot scroll any further. In this case, we mark both the second to last and last pages as viewed.
* Removes the PDF control panel autohiding, which is now superfluous following #9439
* Fixes an issue that @AllanOXDi and I occasionally saw in testing, whereby the `progress_delta` being sent to the backend could inadvertently be summed to a value greater than 1.
* Adds a regression test for the above.

## References
Fixes #8868

## Reviewer guidance
Continually scroll without stopping from the beginning of a PDF until the end. See that it properly marks as complete.

Zoom out slightly so that more than one page of a PDF is rendered at once. Continually scroll without stopping until the end of the PDF. See that as long as no more than 2 pages are visible on the screen at once, it properly marks as complete.

(This is an edge case where if someone were reading through the PDF in a way that allowed for three or more pages to be visible at once, completion would not be marked - someone would probably need a monitor with an aspect ratio with the height about three times the width at least for this to happen, so I think it's not something to be concerned about).

----

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/plugins/learn/assets/src/composables/__tests__/useProgressTracking.spec.js
+++ b/kolibri/plugins/learn/assets/src/composables/__tests__/useProgressTracking.spec.js
@@ -1,4 +1,4 @@
-import { get } from '@vueuse/core';
+import { get, set } from '@vueuse/core';
 import client from 'kolibri.client';
 import { coreStoreFactory as makeStore } from 'kolibri.coreVue.vuex.store';
 import useProgressTracking from '../useProgressTracking';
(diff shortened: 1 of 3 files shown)


pcenov:
@rtibbles I have started testing this one and so far the only issue I've found is that the completion modal gets displayed immediately on opening a 1 page pdf file, thus completely preventing me to even see what the pdf is about.
That seems to be a slight regression as previously it would allow me to at least see the title of the pdf and the completion modal would show only after scrolling down:

https://user-images.githubusercontent.com/79847249/196216759-b16d08ef-4c1b-475f-9832-16e4008cf474.mp4

rtibbles:
I think properly addressing this would require solving this issue: https://github.com/learningequality/kolibri/issues/9348. I can probably fudge it for now, but just implementing a solution for the above might be easier.

pcenov:
@rtibbles no additional regressions observed during today's round of testing in the rest of the supported browsers and the Android app.

radinamatic (approved):
Manual QA passes, experience will be further improved by working on #9348 afterwards 💯 :shipit:

rtibbles:
OK, have put the above issue into the current patch so that we can try to get this resolved within the same patch. Merging!

Issue #9772, opened 2022-10-14
Create "Sync Schedules" Page


marcellamaki:
### Overview 

Add a new page, "Sync Schedules", that lists all scheduled syncs and their corresponding devices, and provides a way to add, edit or delete these syncs. The table should also display the sync status. It should also add the dropdown menu option on the Facility and Device pages, to add "manage sync schedules" 

_Note_: This page will be shared between the facility and device plugins

### Figma
[Link to Specs](https://www.figma.com/file/XHO5zXznOD3WXcxcA6d6Zv/Scheduled-sync-2022?node-id=0%3A1)
<img width="611" alt="Screen Shot 2022-10-14 at 11 59 57 AM" src="https://user-images.githubusercontent.com/17235236/195890522-d5db5494-5577-42dd-a007-4826add9a228.png">
<img width="826" alt="Screen Shot 2022-10-14 at 11 59 34 AM" src="https://user-images.githubusercontent.com/17235236/195890527-7ab36aa0-b351-4d2a-9072-a3cd57484a30.png">

<img width="830" alt="Screen Shot 2022-10-14 at 11 22 18 AM" src="https://user-images.githubusercontent.com/17235236/195883099-43dc63c1-5139-4b7b-8d83-1a8a6e660903.png">

### Data
The data from the backend that is passed to create this page should be: 
* loading all existing scheduled syncs for this facility from the backend
  * Currently all task fetching on the frontend is concerned with getting currently queued, running, or completed tasks. (One example here for getting the tasks for a specific queue: https://github.com/learningequality/kolibri/blob/87f12da247caf366be7df2b1ee7cfda063926024/kolibri/plugins/device/assets/src/views/FacilitiesPage/facilityTasksQueue.js#L24).
  * We would need to enhance the tasks list endpoint to filter down to:
    1. Only scheduled tasks
    2. Only tasks for a specific facility 
* loading the connection status of KDP and all peer devices  
  * There is already frontend code for handling fetching the peer devices: https://github.com/learningequality/kolibri/blob/87f12da247caf366be7df2b1ee7cfda063926024/kolibri/core/assets/src/views/sync/SelectAddressModalGroup/api.js#L114 - unfortunately, three is not currently any code for checking if we are able to connect to KDP

### Accessibility 
Use existing KComponents (i.e. `KButton`) or Kolibri components as relevant for table, which will ensure page is navigable by keyboard.

### Out of Scope
[...]

PR #9786, opened 2022-10-18
Sync schedule

### LINKED ISSUE #9804: Add edit sync schedule and "add device" model to manage sync schedule page
## Overview

Add edit "Sync Schedules", which allow the users to edit or add the devices that have been scheduled for syncing.  The manage sync schedule table/ page table should be able to reflect these changes.  

Ref #9772

Note: This page will be shared between the facility and device plugins

## Figma

 [here](https://www.figma.com/file/XHO5zXznOD3WXcxcA6d6Zv/Scheduled-sync-2022?node-id=84%3A1792)

image
Screenshot 2022-10-27 at 15 27 10

Admin Account
Sync schedule

## Accessibility 

Use existing kmodels or Kolibri components which will break the user’s attention away from the main UI in order to focus attention on a singular action or task.

## Out of scope

Adding a model that will allow users to remove devices from the scheduled sync is not yet part of this issue; it will be addressed in follow-up issues. Most of the present buttons are still disabled so that this work can be merged in and not be broken.

cc @rtibbles @marcellamaki

AllanOXDi:
## Summary
This pull request implements the new Scheduled Sync feature.

## References
Closes 
#9804
#9772

## Reviewer guidance

See[ figma ](https://www.figma.com/file/XHO5zXznOD3WXcxcA6d6Zv/Scheduled-sync-2022?node-id=83%3A1632&t=DX5jXpDwlmP18KeA-0)specs 

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated


kolibri/plugins/facility/assets/src/routes.js
```diff
@@ -124,4 +127,19 @@ export default [
...
+  {
+    path: '/:facility_id?/editsyncschedule',
+    component: editSyncSchedule,
+    name: 'EditSyncSchedule',
+  },
+  {
+    path: '/:facility_id?/editdevice',
+    component: SyncSchedule,
+    name: 'SyncSchedule',
+  },
```
marcellamaki, line 138: Nice job with the routes!

kolibri/plugins/facility/assets/src/views/DataPage/SyncInterface/ManageSyncSchedule.vue
```diff
@@ -0,0 +1,230 @@
+<template>
+
+  <ImmersivePage
+    :appBarTitle="$tr('toolbarHeader')"
```
marcellamaki: `ImmersivePage` also has a `route` prop. This is what allows you to determine what the page should go back to when the user closes out or uses the back arrow. The [common pattern that you will see is `:route=backRoute`](https://github.com/MisRob/kolibri/blob/9ffeca2a815719bda629affe011fa67e37531ee3/kolibri/plugins/device/assets/src/views/FacilitiesPage/FacilitiesTasksPage.vue#L5) with a computed value for `backRoute` defined below. So for example, you would want this route to bring the user back to the main facility page.  You should include a route each place the `ImmersivePage` is used, so the user can exit the page. If you are ever unsure of where the back route should return the user to, you can ask me, Richard, or ask in the #design channel

kolibri/plugins/facility/assets/src/views/DataPage/SyncInterface/ManageSyncSchedule.vue
```diff
@@ -0,0 +1,230 @@
...
+      </KGrid>
+
+      <CoreTable>
+        <template #tbody>
+          <tbody class="table">
+            <tr>
+              <th>{{ $tr('deviceName') }}</th>
+              <th>{{ $tr('Schedule') }}</th>
+              <th>{{ $tr('Status') }}</th>
+              <th></th>
```
marcellamaki: I think it would be good to still have a header here, even if it is not visible. This will help screen reader users, who hear what is on the page, rather than read it. We can ask @radinamatic what a suitable option might be - perhaps the `visually-hidden` class and a description of what the contents of the column is? 

Here is [some general information about tables and accessibility](https://webaim.org/techniques/tables/) (it is not specific documentation to this case, or a suggestion of a solution - just for your learning!). Once we have some suggestions from Radina, we can work on implementing them correctly.

kolibri/plugins/facility/assets/src/views/DataPage/SyncInterface/ManageSyncSchedule.vue
```diff
@@ -0,0 +1,230 @@
...
+                  @click="editButton"
+                >
+                  edit
+                </KButton>
+              </td>
+            </tr>
+            <tr>
+              <td>
+                <span>MacOS<br>
+                  localhost:8000</span>
```
marcellamaki: We can talk about how to make this not placeholder data, but the steps to actually request the data we need 🙂

kolibri/plugins/facility/assets/src/views/DataPage/SyncInterface/ManageSyncSchedule.vue
```diff
@@ -0,0 +1,230 @@
...
+          </tbody>
+        </template>
+
+      </CoreTable>
+      <KModal
+        v-if="deviceModal"
+        :title="$tr('selectDevices')"
+        size="medium"
+        submitText="CONTINUE"
+        cancelText="CANCEL"
```
marcellamaki: Be sure to use [the translation class](https://kolibri-dev.readthedocs.io/en/develop/i18n.html) and either the common strings or the file-specific strings. for example: [KModal with commonCoreStrings for cancel and submit](https://github.com/MisRob/kolibri/blob/9ffeca2a815719bda629affe011fa67e37531ee3/kolibri/core/assets/src/views/sync/SelectAddressModalGroup/SelectAddressForm.vue#L6)

kolibri/plugins/facility/assets/src/views/DataPage/SyncInterface/ManageSyncSchedule.vue
```diff
@@ -0,0 +1,230 @@
...
+        message:
+          'Set a schedule for Kolibri to automatically try syncing with other Kolibri devices that share this facility. Devices must be connected to the same network at the scheduled sync time.',
+        context: 'Introduction on the manage sync schedule',
+      },
+      homeFacility: {
+        message: 'Home Facility Otodi Allan',
+        context: 'My home facility',
+      },
+      addDevice: {
+        message: 'ADD DEVICE',
```
marcellamaki: You do not need to add the all-capitalized here -- the styling is handled within the button

kolibri/plugins/facility/assets/src/views/DataPage/SyncInterface/ManageSyncSchedule.vue
```diff
@@ -0,0 +1,230 @@
...
+        message: 'Sync Schedules',
+        context: 'Title for the manage sync page',
+      },
+      introduction: {
+        message:
+          'Set a schedule for Kolibri to automatically try syncing with other Kolibri devices that share this facility. Devices must be connected to the same network at the scheduled sync time.',
+        context: 'Introduction on the manage sync schedule',
+      },
+      homeFacility: {
+        message: 'Home Facility Otodi Allan',
```
marcellamaki: We can also talk about how to make this dynamic, based on the data received!

kolibri/plugins/facility/assets/src/views/DataPage/SyncInterface/ManageSyncSchedule.vue
```diff
@@ -0,0 +1,230 @@
...
+
+</script>
+
+
+<style scoped>
+ span{
+   font-size:10px;
+ }
+ KRadioButton{
+   font-size:14px;
```
marcellamaki: no need to add styles here - the default button style options should work, I don't think any of these are custom. The [`KButton` documentation](https://kolibri-design-system.netlify.app/kbutton/) can guide you on the styles as needed. If you still have questions, or the button is not appearing as you expect, let me know.

kolibri/plugins/facility/assets/src/views/DataPage/SyncInterface/SyncSchedule/EditSyncSchedule.vue
```diff
@@ -0,0 +1,138 @@
...
+    },
+    methods: {
+      removeDevice() {
+        this.removeDeviceModal = true;
+      },
+      closeModal() {
+        this.removeDeviceModal = false;
+      },
+      ConfirmRemoveDevice() {
+        this.removeDeviceModal = false;
```
marcellamaki: This is a good start to have a function here for confirm! The next step is to actually save data when submitting. To get you started for context, here is an example you can look at. [On submit, create lesson is called.](https://github.com/MisRob/kolibri/blob/9ffeca2a815719bda629affe011fa67e37531ee3/kolibri/plugins/coach/assets/src/views/plan/LessonCreationPage/index.vue#L27). Then, [this method is responsible for actually saving the data](https://github.com/MisRob/kolibri/blob/9ffeca2a815719bda629affe011fa67e37531ee3/kolibri/plugins/coach/assets/src/views/plan/LessonCreationPage/index.vue#L57). We will talk more about how this works, but reviewing this is a good place to start. See if you can find the different functions that are included here - `createLesson`, `showSnackbarNotification`, and the error handling.

kolibri/plugins/facility/assets/src/views/DataPage/SyncInterface/index.vue
```diff
@@ -266,6 +287,10 @@
         context:
           "Describes the 'sync' button which is used to synchronize data from a facility with the project on the Kolibri Data Portal.",
       },
+      manage_sync: {
+        message: 'Manage Sync Shedule',
```
marcellamaki: small typo: "Manage Sync Schedule"

marcellamaki (commented):
Hi @AllanOXDi 

Great work on this so far! It's so cool to see how much you've been able to learn about Vue and implement in just a few months! 🎉 

Most of my comments are about general patterns and principles that you can use throughout the project, even if I add a specific in-line comment. There are likely multiple places the comment is relevant - such as for internationalization, and the "back" route that I mention for 'ImmersivePage'. 

If it would be helpful for me to add more comments, such as a comment in each place that needs to be updated, let me know and I would be happy to do that. But so far, I have just made a general note of the pattern and included some documentation for your reference.

kolibri/plugins/facility/assets/src/views/DataPage/SyncInterface/ManageSyncSchedule.vue
```diff
@@ -0,0 +1,230 @@
...
+        <KGrid>
+          <KGridItem
+            :layout8="{ span: 4 }"
+            :layout12="{ span: 6 }"
+          >
+            <KButton
+              appearance="basic-link"
+              @click="newAddress"
+            >
+              Add new Address
```
marcellamaki: This needs to be a translatable string too

kolibri/plugins/facility/assets/src/views/DataPage/SyncInterface/ManageSyncSchedule.vue
```diff
@@ -0,0 +1,230 @@
...
+
+          <KGridItem
+            :layout="{ alignment: 'right' }"
+            :layout8="{ span: 4 }"
+            :layout12="{ span: 6 }"
+          >
+            <KButton
+              appearance="basic-link"
+            >
+              Forget
```
marcellamaki: Translatable string

kolibri/plugins/facility/assets/src/views/DataPage/SyncInterface/SyncSchedule/EditSyncSchedule.vue
```diff
@@ -0,0 +1,138 @@
+<template>
+
+  <ImmersivePage
+    :appBarTitle="$tr('toolbarHeader')"
```
marcellamaki: This Immersive Page should also have a "back" link option as I mention above

kolibri/plugins/facility/assets/src/views/DataPage/SyncInterface/SyncSchedule/EditSyncSchedule.vue
```diff
@@ -0,0 +1,138 @@
...
+
+      </KGrid>
+    </KPageContainer>
+    <KPageContainer />
+
+    <KModal
+      v-if="removeDeviceModal"
+      :title="$tr('removeDevice')"
+      size="medium"
+      submitText="REMOVE"
```
marcellamaki: Translatable strings for submit and cancel

kolibri/plugins/facility/assets/src/views/DataPage/SyncInterface/SyncSchedule/EditSyncSchedule.vue
```diff
@@ -0,0 +1,138 @@
...
+      toolbarHeader: {
+        message: 'Edit sync schedule',
+        context: 'Heading for edit schedule page.',
+      },
+      editSyncScheduleSubTitle: {
+        message: 'Edit sync schedule',
+        context: 'Subtitle for the edit sync schedule page',
+      },
+      deviceName: {
+        message: 'LINUX 5 (9C24)',
```
marcellamaki: These messages are okay for now (for example, the one for device name, server time, etc.) but we will update them to be based on the data, rather than set values

kolibri/plugins/facility/assets/src/views/DataPage/SyncInterface/ManageSyncSchedule.vue
```diff
@@ -0,0 +1,230 @@
...
+            :layout8="{ span: 4 }"
+            :layout12="{ span: 6 }"
+          >
+
+            <KRadioButton
+              v-model="deviceModal"
+              label="LINUX"
+              value="linux"
+            >
+              <span>localhost:8040</span>
```
marcellamaki: The KButton label and value, and the <span> will need to be update in the future with user-specific data values, but it is okay for now to have placeholder text !

kolibri/plugins/facility/assets/src/views/DataPage/SyncInterface/SyncSchedule/EditSyncSchedule.vue
```diff
@@ -0,0 +1,138 @@
...
+        <KGridItem>
+          <h1>{{ $tr('editSyncScheduleSubTitle') }}</h1>
+        </KGridItem>
+        <KGridItem>
+          <p>{{ $tr('deviceName') }}</p><br>
+        </KGridItem>
+
+        <KGridItem>
+          <KDropdownMenu
+            text="Every After 5 minutes"
```
marcellamaki: The menu options here should be translatable strings. I just refactored`KDropdownMenu` component, so we can use the new version. Let's plan a time to talk through that -- I'll show you the new documentation and we can talk about how to create a set of options for selection.

marcellamaki (commented):
Hi @AllanOXDi - I've added some additional comments. Hopefully this helps illustrate the "general patterns" that I mention a little bit more. You can reply in the comments I've added if anything is still unclear, or message me on slack, and I will try to explain a different way or add more context 😃

kolibri/plugins/facility/assets/src/views/DataPage/SyncInterface/ManageSyncSchedule.vue
AllanOXDi: Done! Thank you so much for guiding me on this.

kolibri/plugins/facility/assets/src/routes.js
```diff
@@ -1,4 +1,7 @@
 import store from 'kolibri.coreVue.vuex.store';
+import ManageSyncSchedule from '../../../../../packages/kolibri-common/components/SyncSchedule/ManageSyncSchedule';
+import SyncSchedule from '../../../../../packages/kolibri-common/components/SyncSchedule/SyncSchedule';
+import editDeviceSyncSchedule from '../../../../../packages/kolibri-common/components/SyncSchedule/EditDeviceSyncSchedule';
```
marcellamaki: I'm not sure if there was a specific reason why these are living in `kolibri-common` (if so, you can disregard this suggestion), but these could possibly be moved to `kolibri/core/assets` and then accessed in the same way some of the components are reused across plugins. For example: 
[kolibri/core/assets/src/views/language-switcher](https://github.com/learningequality/kolibri/tree/develop/kolibri/core/assets/src/views/language-switcher) is a similarly structured folder with sub-components. You can follow the same pattern that is used in the core api spec, where this component is [imported](https://github.com/learningequality/kolibri/blob/develop/kolibri/core/assets/src/core-app/apiSpec.js#L46), and then [made available](https://github.com/learningequality/kolibri/blob/develop/kolibri/core/assets/src/core-app/apiSpec.js#L166).

When it is [imported in other components (for example, the side nav)](https://github.com/learningequality/kolibri/blob/develop/kolibri/core/assets/src/views/SideNav.vue#L206), the import path is a bit simpler.

packages/kolibri-common/components/SyncSchedule/EditDeviceSyncSchedule.vue
```diff
@@ -0,0 +1,395 @@
...
+      },
+      SyncTime() {
+        return [
+          { label: '1:30 AM', value: 3600 },
+          { label: '2:00 AM', value: 86400 },
+          { label: '2:30 AM', value: 604800 },
+          { label: '3:00 AM', value: 604800 },
+          { label: '3:30 AM', value: 2592000 },
+          { label: '4:00 AM', value: 2592000 },
+          { label: '4:30 AM', value: 2592000 },
```
marcellamaki: Are these the only options we are offering, so that it is during the hours most likely to be offline, or is this just a sample list?

packages/kolibri-common/components/SyncSchedule/EditDeviceSyncSchedule.vue
```diff
@@ -0,0 +1,395 @@
...
+      tue: {
+        message: 'Tuesday',
+        context: 'Day 2',
+      },
+      wed: {
+        message: 'Wednesday',
+        context: 'Day 3',
+      },
+      thur: {
+        message: 'Thrusday',
```
marcellamaki: small typo

packages/kolibri-common/components/SyncSchedule/EditDeviceSyncSchedule.vue
```diff
@@ -0,0 +1,395 @@
...
+        message: 'Friday',
+        context: 'Day 5',
+      },
+      sat: {
+        message: 'Saturday',
+        context: 'Day 6',
+      },
+      sun: {
+        message: 'Sunday',
+        context: 'Day 7',
```
marcellamaki: Might be good to ask @radinamatic if there is standard context we use for days of the week in other forms, or if she has any suggestions

kolibri/plugins/facility/assets/src/routes.js
```diff
@@ -124,4 +127,19 @@ export default [
...
+    component: ManageSyncSchedule,
+    name: 'ManageSyncSchedule',
+  },
+  {
+    path: '/:facility_id?/editsyncschedule/?',
+    component: editDeviceSyncSchedule,
+    name: 'EditSyncSchedule',
+  },
+  {
+    path: '/:facility_id?/editdevice',
```
marcellamaki: I think you might need to pass something like the device id in here, in addition to the facility id, so you can properly load the page

Maybe this route could be something like: 
`'/:facility_id?/:device_id?/edit'`

It doesn't have to be exactly that, but I think you will need both params there

marcellamaki (dismissed):
Hi @AllanOXDi - lots of new work here, which is really exciting to see! 

There is some smoothing out to be done on a few parts of the interface, as well as with some of the CRUD functionality of the device syncing schedule itself.

To start with the syncing, I've added an inline comment about editing, but also wanted to add a little bit more context. When I click on the edit button, I do see the page is created and loads, but there are some console errors and no contents of the page. I'm not sure if this is the step you were at, or if the goal was to have the "edit" feature fully working.

If you want to have the edit feature fully working, it seems like there are two errors happening, one related to your Vuex, and one related to the endpoint returning a 404 error. I'd suggest just debugging one at a time (perhaps with some route adjustments to make sure the device ID is also getting passed, it seems like that might be part of the problem), and then seeing if you can bring both pieces together.

Screenshot 2023-02-17 at 10 43 50 AM


As for the UI, there is some general work that can be done to help it match the figma even more closely, even though it's just small things. 

First up:
- overall, the page text looks a little bit "crowded" here. You can look at the figma and perhaps add a little bit of spacing. In some places, if you are using just `{{ $tr('here's my string') }}` to add text inside of a `div` or a `span`, wrapping it in a `<p>` tag can add some default padding that makes it feel a little bit roomier. You may still need to add a `class` to that `<p>` and adjust the spacing, but that might be a place to start.  
- You can see that the contents shifts when the dropdown is opened, and that it's a bit hard to scroll. I think adding a default, fixed height to the main page container, rather than just having it the size of the content that is there (for example, making it always take up ~80% of the page) might be a good idea. What do you think @jtamiace? 

https://user-images.githubusercontent.com/17235236/219707498-b84f2035-002d-4e1f-87d3-4b5b8b381339.mp4

- Finally, this is a bit of a nitpick, but to save Jessica having to do it, the alignment on the left should all be in one line. One tool I use to help with "pixel perfect" UIs (well, they are pixel perfect in my mind, although Jessica and Tomiwa might have a different opinion) is a Chrome browser extension called "Grid Rule". If you use Chrome you can just search for it and add it on. Turn on the extension, and then you can "pull" lines from the top or the left, to check the alignment of various page elements.

https://user-images.githubusercontent.com/17235236/219709118-11e1f1bc-d3f8-45d8-b63c-ee2fbb480348.mp4

On the main sync schedules page, I have two suggestions. 
- Add some kind of loading state. One option would be to have a loading state for the whole table. The other option would be to have a loading state just for the sync status, if the other pieces of the table (i.e. device name) wouldn't take an extra fetch. This would help reduce the time where it looks like there are no devices
- I'm not sure why there are four statuses here. I tried to figure it out from the code and it seems like `myDevices` is being set to `tasks` - if there is one task per device, that would be right, but here, it seems like there might be more than one task per device, in which case we may need to rethink how this is being displayed to the user. 

https://user-images.githubusercontent.com/17235236/219710099-7028e4ea-5454-4242-bb06-d4133da2714e.mp4

Finally - and this maybe is a question for @rtibbles regarding the expected behavior (maybe with @bjester weighing in if any of the refactoring work you are doing could relate to this?)

I am now running into fun edge cases since I'm on ZeroTier and I can see other's devices available. This morning I saw both Samson's and Allan's computers in the "devices" that were on network. Trying to set a sync schedule on these (logically) resulted in a 403 error. Do we want to filter those devices out entirely? Do we want to filter them out for now, but in the future, add an option to input admin credentials in them, so that they could sync? Just wanted to raise this so we can start thinking about it.

Okay - that's all I have for now, but let me know any questions you have or where you might need more tips or details, and I will do my best to help! Great progress on this!

packages/kolibri-common/components/SyncSchedule/EditDeviceSyncSchedule.vue
AllanOXDi: I am just not sure how to implement this

jtamiace:
> You can see that the contents shifts when the dropdown is opened, and that it's a bit hard to scroll. I think adding a default, fixed height to the main page container, rather than just having it the size of the content that is there (for example, making it always take up ~80% of the page) might be a good idea. What do you think @jtamiace?

Yes, I'm not sure of technical details on how this can be accomplished but the dropdown menu should appear as if it's overlaying above the other components and containers on the page, and not being a part of it
Screenshot 2023-02-21 at 8 58 07 AM


A few other layout details:

Could you reduce the space between the dropdowns to be 16px?
Screenshot 2023-02-21 at 9 00 55 AM

I don't know if this should be addressed through KDS or not, but when a dropdown is inactive, the test should be roughly center-aligned, and the left/right padding should be similar to other text fields with gray backgrounds in Kolibri. The one I screenshotted for reference here is from the sign in page
Screenshot 2023-02-21 at 9 04 39 AM

packages/kolibri-common/components/SyncSchedule/EditDeviceSyncSchedule.vue
rtibbles: For the labels, you can use the `$formatTime` method that is available on `this`: https://github.com/learningequality/vue-intl#formattime

It's a little bit long winded, but you can make a helper function to do this for each time, but to create the Date object, you would need to do something like:
```
const time  = new Date();
now.setHours(8);
now.setMinutes(0);

const label = this.$formatTime(time, {format: 'short'});
```
That would give you the label for `8:00AM` in the appropriate localized format.

packages/kolibri-common/components/SyncSchedule/EditDeviceSyncSchedule.vue
rtibbles: To enumerate the days of the week, you can use a similar trick to the format time above, but using the `formatDate` method: https://github.com/learningequality/vue-intl#formatdate

To get just the day of the week, you should be able to pass a date object to it set to the correct day:
```
this.$formatDate(date, { weekday: 'long' })
```

This _should_ just return the day of the week in long form.

More details here: https://stackoverflow.com/questions/30437134/how-to-get-the-weekday-names-using-intl

N.B. the API they are referencing is wrapped by the `$formatDate` method, so the same options can be passed to either, you just don't need to specify the locale for `$formatDate`.

kolibri/plugins/facility/assets/src/routes.js
rtibbles: I think it would be `'/:facility_id?/:device_id/edit'` without the `?` because `deviceId` is a required parameter. At the moment you are accessing this via a query parameter instead, `$route.query.deviceId` whereas you would access this by `$route.params.deviceId`.

kolibri/plugins/facility/assets/src/routes.js
rtibbles: You can avoid this long relative path and just use `kolibri-common/components/SyncSchedule/ManageSyncSchedule` instead - `kolibri-common` is registered as a package in our local workspace, so you can treat it like any other node_modules package when importing from it.

kolibri/plugins/device/assets/src/views/FacilitiesPage/index.vue
```diff
@@ -270,7 +270,7 @@
       facilityOptions(facility) {
         return [
           {
-            label: this.coreString('registerAction'),
+            label: 'Manage sync Schedule',
```
rtibbles: This string needs to be added as a translatable string!

Also, you have just overwritten the register action option here, you should be adding an additional option alongside `register`.

kolibri/plugins/device/assets/src/views/FacilitiesPage/index.vue
```diff
@@ -284,7 +284,8 @@
         if (option === Options.REMOVE) {
           this.facilityForRemoval = facility;
         } else if (option === Options.REGISTER) {
-          this.facilityForRegister = facility;
+          this.$router.push({ path: '/manage' });
```
rtibbles, line 293: We should add an additional `else if` here for handling the sync schedule option, rather than overwriting the register option.

kolibri/plugins/facility/assets/src/constants.js
```diff
@@ -10,6 +10,7 @@ export const PageNames = {
   DATA_EXPORT_PAGE: 'DATA_EXPORT_PAGE',
   FACILITY_CONFIG_PAGE: 'FACILITY_CONFIG_PAGE',
   ALL_FACILITIES_PAGE: 'ALL_FACILITIES_PAGE',
+  ManageSyncSchedule: 'ManageSyncSchedule',
```
rtibbles: We should follow the same convention for this constant as the one above - use UPPERCASE_SNAKECASE.

It also seems like we are missing one for `EDIT_SYNC_SCHEDULE`?

kolibri/plugins/facility/assets/src/routes.js
```diff
@@ -124,4 +127,19 @@ export default [
       }
     },
   },
+  {
+    path: '/:facility_id?/:device_id?/managesync',
```
rtibbles: If I am understanding correctly, `ManageSyncSchedule` is showing the list of all scheduled syncs for this facility, so we don't need the `device_id` parameter here at all:
`path: '/:facility_id?/managesync',`

kolibri/plugins/facility/assets/src/routes.js
```diff
@@ -124,4 +127,19 @@ export default [
...
+    component: ManageSyncSchedule,
+    name: 'ManageSyncSchedule',
+  },
+  {
+    path: '/:facility_id?/:device_id?/editsyncschedule/?',
```
rtibbles: I think we need the device_id here, so it's not an optional parameter, and I don't think we need the `?` at the end?
`path: '/:facility_id?/:device_id/editsyncschedule/',`

kolibri/plugins/facility/assets/src/routes.js
```diff
@@ -124,4 +127,19 @@ export default [
...
+    component: editDeviceSyncSchedule,
+    name: 'EditSyncSchedule',
+  },
+  {
+    path: '/:facility_id?/:device_id?/editdevice',
```
rtibbles: What is the difference between this and the route above? I think we only need two routes, one for listing all the scheduled syncs, and one for creating/editing the scheduled sync for a specific device?

kolibri/plugins/facility/assets/src/views/DataPage/SyncInterface/index.vue
```diff
@@ -59,16 +59,15 @@
                 >
                   <template #options>
                     <CoreMenuOption
-                      v-if="theFacility.dataset.registered"
```
rtibbles, line 62: We should not be messing with the existence registration code, and should instead be adding an additional `CoreMenuOption` for the manage sync.

kolibri/plugins/facility/assets/src/views/DataPage/SyncInterface/index.vue
```diff
@@ -266,6 +268,10 @@
         context:
           "Describes the 'sync' button which is used to synchronize data from a facility with the project on the Kolibri Data Portal.",
       },
+      manage_sync: {
```
rtibbles: It seems we have lost our linting rule that enforced camelCase message ids, but I would prefer if we maintain this convention: `manageSync`.

Also maybe this should go into coreStrings, as this is needed elsewhere too?

packages/kolibri-common/components/SyncSchedule/SyncSchedule.vue
```diff
@@ -0,0 +1,19 @@
+<template>
+
+  <EditSyncSchedule msg="Remove device from sync schedule" />
```
rtibbles: I don't think this needs to be a separate component - and this message needs to be made translatable. I would recommend moving this message into the `EditSyncSchedule` component, and conditionally displaying it if there is an existing scheduled sync.

packages/kolibri-common/components/SyncSchedule/ManageSyncSchedule.vue
```diff
@@ -0,0 +1,365 @@
...
+
+        </KGridItem>
+
+        <KGridItem>
+          <p>{{ $tr('introduction') }}</p>
+        </KGridItem>
+
+
+        <KGridItem :layout8="{ span: 4 }" :layout12="{ span: 6 }" class="separate">
+          <b>{{ facility.name }}</b>
```
rtibbles, line 21: `facility` starts out `null`, so when we initially try to access `facility.name` here, it gives an error.

We could perhaps conditionalize the `b` on `facility`:
```
<b v-if="facility">{{ facility.name }}</b>
<KCircularLoader v-else />
```
and add an else for a KCircularLoader to indicate to the user that loading is happening in the meantime.

kolibri/plugins/facility/assets/src/constants.js
```diff
@@ -32,6 +33,9 @@ export const pageNameToModuleMap = {
   [PageNames.CLASS_ENROLL_LEARNER]: 'classAssignMembers',
   [PageNames.USER_MGMT_PAGE]: 'userManagement',
   [PageNames.FACILITY_CONFIG_PAGE]: 'facilityConfig',
+  [PageNames.DATA_EXPORT_PAGE]: 'dataExport',
```
rtibbles: You haven't added vuex modules for these three pages, so there's no need to add these at all. These are only needed when we have a corresponding vuex module for the page in question (we are using this less and less as we rely more and more on composables for handling our shared reactive state).

packages/kolibri-common/components/SyncSchedule/EditDeviceSyncSchedule.vue
```diff
@@ -0,0 +1,395 @@
...
+        <KGridItem>
+          <p>{{ device.device_name }} </p>
+        </KGridItem>
+
+        <KGridItem class="select-element">
+          <div
+            class="k-select"
+          >
+            <KSelect
+              v-model="selectedInterval"
```
rtibbles: `selectedInterval` is not defined anywhere, so this is giving an error on render.

Note that the value passed here as the v-model should correspond to one of the `options` (or one of the value properties of one of the options) being passed in from `selectArray`, or `null`, as you have done with `selectedDay` and `selectedTime`.

packages/kolibri-common/components/SyncSchedule/EditDeviceSyncSchedule.vue
```diff
@@ -0,0 +1,395 @@
...
+      },
+
+      cancelBtn() {
+        this.$router.push({ name: PageNames.ManageSyncSchedule });
+      },
+      fetchDevice() {
+        NetworkLocationResource.fetchModel({ id: this.$route.query.id }).then(device => {
+          this.device = device;
+          TaskResource.list({ queue: 'facility_task' }).then(tasks => {
+            this.tasks = tasks.filter(device.extra_metadata.device_id === this.id);
```
rtibbles: Sorry, I totally missed this previously. The `filter` method of an Array takes a callback function that is called with each element of the Array in turn, so this would then become:
```
this.tasks = tasks.filter(task => task.extra_metadata.device_id === device.id);
```
So for each task, you are looking at its extra_metadata and querying to see which one matches the current device_id.
Another gotcha here is that we probably also need to filter by the facility_id in case we have setup syncing to the same device for multiple facilities!

So it would be something like this:
```
this.tasks = tasks.filter(task => task.extra_metadata.device_id === device.id && task.facility_id === this.$store.getters.activeFacilityId);
```

packages/kolibri-common/components/SyncSchedule/EditDeviceSyncSchedule.vue
```diff
@@ -0,0 +1,395 @@
...
+              v-model="selectedInterval"
+              class="selector"
+              :style="selectorStyle"
+              :options="selectArray"
+              label="Repeat"
+            />
+
+          </div>
+          <div
+            v-if="selectedItem.label !== 'Never'
```
rtibbles: You should do this based on `selectedItem.value` here, otherwise this will only work when the interface is in English.

Fortunately that makes the comparison a bit more concise:

```
v-if="selectedItem.value <= 86400"
```

You just need to check if it's equal to the value for every day or less!

packages/kolibri-common/components/SyncSchedule/EditDeviceSyncSchedule.vue
```diff
@@ -0,0 +1,395 @@
...
+        <KGridItem>
+          <h1>{{ $tr('editSyncScheduleSubTitle') }}</h1>
+        </KGridItem>
+
+        <KGridItem>
+          <p>{{ device.device_name }} </p>
+        </KGridItem>
+
+        <KGridItem class="select-element">
+          <div
```
rtibbles: To get the desired responsive behaviour here, you could use a nested `KGrid` within the parent `KGrid` - this should also enforce standard spacing between the `KSelect`s if they are each inside their own `KGridItem`.

packages/kolibri-common/components/SyncSchedule/EditDeviceSyncSchedule.vue
```diff
@@ -0,0 +1,395 @@
...
+        }
+
+        return times;
+      },
+    },
+    beforeMount() {
+      this.fetchDevice();
+    },
+    beforeDestroy() {
+      clearInterval(this.timer);
```
rtibbles, line 255: I think this was copied in to clean up a timer to keep the `now` value refreshed - I think the logic came from the `ElapsedTime` component. As you are displaying the server time to the user here, it would be a good idea to keep it updating!

See the `mounted` lifecycle hook there: https://github.com/learningequality/kolibri/blob/develop/kolibri/core/assets/src/views/ElapsedTime.vue#L34 (you can probably copy it in verbatim).

packages/kolibri-common/components/SyncSchedule/EditDeviceSyncSchedule.vue
```diff
@@ -0,0 +1,395 @@
...
+          <p class="spacing">
+            <KCheckbox>
+              {{ $tr('checboxlabel') }}
+            </KCheckbox>
+          </p>
+          <KButton
+            :vIf="msg"
+            appearance="basic-link"
+            class="spacing"
+            @click="removeDevice"
```
rtibbles: No need to change, but you can also do direct assignment in `@click` handlers, so instead of creating an intermediary method, you could do:
```
@click="removeDeviceModal = true"
```

and it would work just the same!

rtibbles (commented):
Some additional cleanup that would be good here.

Most importantly ensuring that all strings are put into a translator object or `$trs` key on a component.

Let me know if you have any questions!

kolibri/plugins/facility/assets/src/routes.js
AllanOXDi: Removing the` ? `on the device_id prevents the page from loading

packages/kolibri-common/components/SyncSchedule/ManageSyncSchedule.vue
```diff
@@ -0,0 +1,365 @@
...
+      },
+      closeModal() {
+        this.deviceModal = false;
+      },
+      submitModal(id) {
+        this.deviceModal = false;
+        if (id !== ' ') {
+          this.deviceIds.push(id);
+
+          this.$router.push({ path: '/editdevice/?id=' + id });
```
rtibbles: To make this work with the `deviceId` url parameter, you would need to update this (can pass parameters as part of [programmatic navigation](https://v3.router.vuejs.org/guide/essentials/navigation.html)):
```
 this.$router.push({ name: PageNames.EDIT_SYNC_SCHEDULE, params: { deviceId: id } });
```
Once you've defined the `EDIT_SYNC_SCHEDULE` page name and used that for the route.

You would also need to update the EditSyncSchedule component to read the deviceId from params.deviceId rather than from the query.

kolibri/plugins/facility/assets/src/routes.js
rtibbles: I added a comment here about how you'd need to update the navigation to fix this (and updates in reading the deviceId too): https://github.com/learningequality/kolibri/pull/9786/files#r1120769310

kolibri/plugins/device/assets/src/views/FacilitiesPage/index.vue
```diff
@@ -285,6 +289,8 @@
           this.facilityForRemoval = facility;
         } else if (option === Options.REGISTER) {
           this.facilityForRegister = facility;
+        } else if (option === Options.MANAGESYNC) {
```
rtibbles, line 292: Excellent, now you just need to add the `MANAGESYNC` key to the `OPTIONS` constant in line 194 of this file.

https://github.com/learningequality/kolibri/blob/07686f6764fb9b87b1c1c9c2f965c68e45b2631d/kolibri/plugins/device/assets/src/views/FacilitiesPage/index.vue#L194

kolibri/plugins/facility/assets/src/routes.js
```diff
@@ -124,4 +126,14 @@ export default [
       }
     },
   },
+  {
+    path: '/:facility_id?/managesync',
+    component: ManageSyncSchedule,
+    name: PageNames.MANAGE_SYNC_SCHEDULE,
+  },
+  {
+    path: '/:facility_id?/:device_id?/editdevice/:deviceId',
```
rtibbles: Only need `deviceId` once in the URL! So can just have:
`path: '/:facility_id?/editdevice/:deviceId',`

kolibri/plugins/facility/assets/src/views/DataPage/SyncInterface/index.vue
```diff
@@ -59,13 +59,16 @@
                 >
                   <template #options>
                     <CoreMenuOption
-                      v-if="theFacility.dataset.registered"
+                      :style="{ 'cursor': 'pointer', textAlign: 'left' }"
+                      :label="coreString('manageSyncAction')"
+                      @select="managesync()"
+                    />
+                    <CoreMenuOption
```
rtibbles, line 66: Need to add:
`v-if="theFacility.dataset.registered"` back here.

kolibri/plugins/facility/assets/src/views/DataPage/SyncInterface/index.vue
```diff
@@ -59,13 +59,16 @@
...
                       :style="{ 'cursor': 'pointer', textAlign: 'left' }"
                       :label="$tr('register')"
                       @select="displayModal(Modals.REGISTER_FACILITY)"
                     />
                     <CoreMenuOption
-                      v-else
```
rtibbles: Should reinstate the `v-else`

packages/kolibri-common/components/SyncSchedule/EditDeviceSyncSchedule.vue
```diff
@@ -0,0 +1,403 @@
...
+      removeText: {
+        message: 'Remove',
+        context: 'Label for remove button on the remove device modal',
+      },
+      cancelText: {
+        message: 'Cancel',
+        context: 'Label for cancel button on the remove device modal',
+      },
+      removeDeviceLabel: {
+        message: 'Remove device from sync schedule',
```
rtibbles, line 379: :+1:

kolibri/core/assets/src/views/sync/FacilityNameAndSyncStatus/index.vue
```diff
@@ -44,6 +44,11 @@
             {{ $tr('lastSync', { relativeTime: formattedTime(facility.last_successful_sync) }) }}
           </span>
         </template>
+        <KButton
+          :text="$tr('createSync')"
+          appearance="basic-link"
+          @click="manageSync"
```
rtibbles: To make this portable across the facility and device plugins, we will probably want to just emit an event from the component here `$emit('manageSync')` here, and handle that event wherever this component is used. That way we can navigate different in the facility plugin vs the device plugin as needed.

kolibri/plugins/device/assets/src/views/FacilitiesPage/index.vue
```diff
@@ -269,10 +269,14 @@
     methods: {
       facilityOptions(facility) {
         return [
+          {
+            label: this.coreString('manageSyncAction'),
+            value: Options.MANAGESYNC,
+            disabled: facility.dataset.registered,
```
rtibbles: This `disabled` should remain on the `Options.REGISTER` option below, not here.

packages/kolibri-common/components/SyncSchedule/ManageSyncSchedule.vue
```diff
@@ -0,0 +1,355 @@
...
+                  {{ device.extra_metadata.baseurl }}
+                </span>
+              </td>
+              <td>
+                <div>
+                  {{ device.extra_metadata.sync_state }}
+                </div>
+              </td>
+
+              <td v-if="data.length > 0">
```
rtibbles: Here and below, this can cause an issue because data is initialized to `null`, so you would need to do an existence check on data first: `<td v-if="data && data.length > 0">`

packages/kolibri-common/components/SyncSchedule/EditDeviceSyncSchedule.vue
```diff
@@ -0,0 +1,403 @@
...
+
+        <KGridItem>
+          <p>{{ device.device_name }} </p>
+        </KGridItem>
+
+        <KGrid class="align-kselects">
+          <KGrid>
+            <KGridItem>
+              <KSelect
+                v-model="selectedItem"
```
rtibbles, line 23: One of these KSelects is still giving a lot of validator warnings, so we should work out which one(s) and clean up the inputs/options.

rtibbles (approved):
This is 'strings ready' and much more. Still some work to do, but it's all been enumerated either in the existing issues, or this follow up issue: https://github.com/learningequality/kolibri/issues/10173

PR #9803, opened 2022-10-26
Add @rtibbles' script to run kolibri in app mode, yarn scripts

nucleogenesis:
## Summary

Bootstrapping Kolibri to run in app mode w/ the app plugin enabled w/ dev settings.

Provides `yarn app-python-devserver` which runs the script and `yarn app-devserver` which runs the frontend devserver in parallel.

Additionally fixes issue w/ `get_initialize_url` returning an empty string.

## Reviewer guidance

- Should we have a dev server version and a non devserver version? Would QA need a proper dev server?

### DIFF
--- a/docs/manual_testing/app_plugin/index.rst
+++ b/docs/manual_testing/app_plugin/index.rst
@@ -1,0 +1,11 @@
+Testing Kolibri with app plugin enabled
+=============
+
+The Kolibri app plugin
+----------------------------
+
+The Kolibri app plugin is designed to provide features and behavior with the mobile app user in mind. In order to test or develop Kolibri in this mode, there are commands that can be used to initialize Kolibri as needed.
+
+By running the command: `yarn app-python-devserver` you will start Kolibri in development mode. You can also run `yarn app-devserver` to run the frontend devserver in parallel.
+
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -62,5 +64,8 @@
   "browserslist": [
     "extends browserslist-config-kolibri"
   ],
-  "dependencies": {}
+  "dependencies": {},
+  "volta": {
+    "node": "16.18.0"
+  }
```
nucleogenesis, line 70: Been meaning to ask if we can just do this since we suggest volta in our dev docs (and likewise pin node@10 in the release 15 branch) @rtibbles

rtibbles: Yes, this seems eminently sensible to me.

docs/manual_testing/app_plugin/index.rst
```diff
@@ -0,0 +1,11 @@
+Testing Kolibri with app plugin enabled
+=============
```
rtibbles:
```
/home/runner/work/kolibri/kolibri/docs/manual_testing/app_plugin/index.rst:2:Title underline too short.
```

Need to check the docs build! It may also complain about the underline being too long below.

rtibbles (commented):
Looks good, just needs docs tweaks.

rtibbles (approved):

PR #9807, opened 2022-11-02
Tweaks to fix GPG signing in PyPi upload

rtibbles:
## Summary
Transfers lessons learned from prototyping the actions workflow here https://github.com/learningequality/kolibri-oidc-client-plugin/blob/b91521a42bbaeb414b6f82949f57b310e86923c1/.github/workflows/pypi_upload.yml

### DIFF
--- a/.github/workflows/pypi_upload.yml
+++ b/.github/workflows/pypi_upload.yml
@@ -46,11 +46,11 @@ jobs:
         path: dist
     - name: Configure GPG Key
       run: |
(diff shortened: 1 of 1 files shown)


marcellamaki (approved):
This seems fine to me. I looked up the flags that were added (aside from the "") and they make sense. Should be fine to merge to develop and can always adjust as needed if problems pop up

rtibbles:
Having tried every which way on the oidc client plugin, I am pretty confident in this working without any tweaks!

Issue #9589, opened 2022-08-04
Research benefits & LOE of moving common code to package


nucleogenesis:
## Observed behavior

[Follow-up on this comment](https://github.com/learningequality/kolibri/pull/9534#discussion_r908967139) - overall, it may be worthwhile to consider moving some common (although not core) code out of our default front-end bundle.

The idea being that if we expose this code in a package, then when a plugin uses common code, it will be bundled with that plugin's bundle. Conversely, a plugin that does not use some common code will have it shaken out of their builds. 

So - we should see a significant reduction in the size of the default frontend build which is loaded on every page along with a relatively small increase to each plugin's build size.

### Thoughts on differentiating between "core" and "common"

@rtibbles said:

> "What do we think we need to expose, and are prepared to support for people outside Learning Equality building frontend plugins for Kolibri?"

If you take this on and don't feel comfortable deciding on what to move and what not do, prepare and schedule a meeting with the likes of @marcellamaki @rtibbles and @nucleogenesis to walk through the core API spec and to decide on what should go where.

## Acceptance Criteria

- [x] A new package is created in Kolibri called `kolibri-common`
- [ ] The core API spec is reduced in size
- [ ] Provide before & after outcomes for the production and dev build sizes

rtibbles:
Closing in favour of #5488

PR #9410, opened 2022-05-06
Adds aria-live=polite to the global snackbar component

marcellamaki:
## Summary
This adds `aria-live=polite` to the global snackbar component. This setting does not "overly interrupt" on screen readers, but provides update notifications when changes happen.

## References
Fixes #9406 

## Reviewer guidance
The reviewer should test on a screen reader device or on a device with voiceover or similar settings enabled, anywhere where there are snackbar notifications.

I found testing on the "users" table to be easiest -- by creating and deleting a user, I was able to test easily several times.

I have _only_ tested this on MacOS with voiceover... It does seem to be working as expected in this setting, and the reading I have done would indicate that this is fairly universal, but I'm not sure about other devices.

## Testing checklist

- [x] Contributor has fully tested the PR manually
- [ ] If there are any front-end changes, before/after screenshots are included
- [ ] Critical user journeys are covered by Gherkin stories
- [ ] Critical and brittle code paths are covered by unit tests

## PR process

- [x] PR has the correct target branch and milestone
- [x] PR has 'needs review' or 'work-in-progress' label
- [x] If PR is ready for review, a reviewer has been added. (Don't use 'Assignees')
- [x] If this is an important user-facing change, PR or related issue has a 'changelog' label
- [ ] If this includes an internal dependency change, a link to the diff is provided

## Reviewer checklist

- Automated test coverage is satisfactory
- PR is fully functional
- PR has been tested for [accessibility regressions](http://kolibri-dev.readthedocs.io/en/develop/manual_testing.html#accessibility-a11y-testing)
- External dependency files were updated if necessary (`yarn` and `pip`)
- Documentation is updated
- Contributor is in AUTHORS.md

Issue #5488, opened 2019-05-15
Separate frontend core API into 'public' and 'internally shared'


rtibbles:
Currently, all core shared components are loaded onto the core API object, which leads to a large, bloated object that we load on every single page of Kolibri.

In conversation with @indirectlylit, we posited that the purposes of the core API would be better served by a two tier structure:

1) The public, core API. This would be available to external plugins, and accessable on the `kolibriGlobal` object on the `window` object. (this would be a subset of what is currently exposed in this way). These would be available in the `kolibri` core package, under the same file paths as used for import, but for plugin builds, they would map to the kolibriGlobal object at runtime.

2) A shared, internal, library. This would be available at build time for use in Kolibri. The intention of this is to allow us to more easily share components across different internal plugins, without having to overload the core API. These would be put into the `kolibri-common` package, which will not be published, nor supported for external usage.

Lastly - we will continue to use KDS as a place for public, shared components and functionality.

Current single purpose modules that need to be imported independently of the core API (`kolibri_module`, `kolibri_app`, `content_renderer_module`, and `plugin_data`) will be migrated to separate packages that are published to NPM.

The proposed mapping of the current API to this new format would be as follows (null indicates it would be removed):

```
  kolibri_module: 'kolibri-module',
  kolibri_app: 'kolibri-app',
  content_renderer_module: 'kolibri-viewer',
  plugin_data: 'kolibri-plugin-data',
  'kolibri.client': 'kolibri/client',
  'kolibri.heartbeat': 'kolibri/heartbeat',
  'kolibri.lib.logging': 'kolibri/logging',
  'kolibri.lib.vue': 'vue',
  'kolibri.lib.vuex': 'vuex',
  'kolibri.lib.vueCompositionApi': '@vue/composition-api',
  'kolibri.lib.apiResource': 'kolibri/apiResource',
  'kolibri.coreVue.vuex.constants': 'kolibri/constants',
  'kolibri.coreVue.vuex.getters': null,
  'kolibri.coreVue.vuex.actions': null,
  'kolibri.coreVue.vuex.store': null,
  'kolibri.coreVue.vuex.mappers': null, // Remove use of remaining mapper
  'kolibri.coreVue.components.ScrollingHeader': null,
  'kolibri.coreVue.components.Backdrop': 'kolibri-common/components/Backdrop',
  'kolibri.coreVue.components.CoachContentLabel': 'kolibri-common/components/labels/CoachContentLabel',
[...]

PR #9809, opened 2022-11-03
Refactors UserAuth and ProfileEdit pages  not to use Corebase

### LINKED ISSUE #9233: Remove CoreBase from User Profile and User Auth plugins
Remove the CoreBase usage and replace with AppBarCorePage or ImmersivePage, where applicable.

See https://github.com/learningequality/kolibri/pull/9196 or https://github.com/learningequality/kolibri/pull/9222 for examples of how this was done in Facility and Learn.

akolson:
## Summary

This PR refactors UserAuthPage and ProfileEditPage to remove CoreBase usage.
Screenshot 2022-11-03 at 18 17 21

Screenshot 2022-11-03 at 18 14 39

## References

Closes #9233

Please note that the linting build fails due to a build bug fixed by https://github.com/learningequality/kolibri/pull/9810

## Reviewer guidance

View the following pages. Pages should function as before;
- `http://localhost:8000/en/auth/#/signin`
- `http://localhost:8000/en/profile/#/edit`

----

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated


kolibri/core/assets/src/core-app/apiSpec.js
```diff
@@ -173,11 +177,13 @@ export default {
       PointsIcon,
       AuthMessage,
       FilterTextbox,
+      CoreBanner,
       CoreSnackbar,
       CoreMenu,
       CoreMenuDivider,
       CoreMenuOption,
       CoreTable,
+      GlobalSnackbar,
```
nucleogenesis: One thought that occurs here is that perhaps we could generalize the `UserAuthLayout` instead to be a part of the `CorePage` set of components (where the `AppBarPage` and `ImmersivePage` live). This is just because that way the ~`CoreBanner`~, `GlobalSnackbar` and `AppError` can remain private to the `core` app. I'll add more thoughts relevant to this in a separate comment

EDIT: Perhaps we won't be able to avoid it w/ `CoreBanner`

kolibri/plugins/user_auth/assets/src/views/UserAuthLayout.vue
```diff
@@ -0,0 +1,186 @@
...
+        tabindex="-1"
+        class="main"
+      >
+        <slot></slot>
+      </div>
+    </div>
+
+    <div aria-live="polite">
+      <GlobalSnackbar />
+    </div>
```
nucleogenesis, line 45: I'm not familiar with "polite" here and it's new to our code base; could you add a little comment along with this explaining what it does in this context? 

Also, if this is how we should be handling our snackbars, we should add this aria-live to all of them. I wonder if it would be easiest to just add the aria attribute directly in `CoreSnackbar`'s wrapping div?

kolibri/plugins/user_auth/assets/src/views/UserAuthIndex.vue
```diff
@@ -66,8 +41,10 @@
       },
     },
     created() {
-      // Check for redirect param and store it in vuex
-      // otherwise it'll be lost when the route changes.
+      /*
+       * Check for redirect param and store it in vuex
+       * otherwise it'll be lost when the route changes.
+       */
```
nucleogenesis, line 47: Thank you for the clarifying comments!

kolibri/core/assets/src/core-app/apiSpec.js
nucleogenesis: **This thread is not a blocker for merge**, but just have some thoughts to share about some code organization stuff.

I thinking this adds a smidgen of tech debt; however it is not easily avoided because of the need to have the CoreBanner specifically available to the User Auth plugin in any case. I thought about perhaps moving the `CoreBanner` component over to the User Auth plugin altogether, but I think that's more work than its worth.

I'm also hoping to avoid adding to the apiSpec as much as we can as it adds some bloat, especially for something not being used across multiple plugins.

To avoid that bloat perhaps we can make this an exception for our *no importing across packages* "rule". It feels slightly less problematic when we're reaching out for something in core than it would to import across plugins since `core` will always be there. Curious on thoughts from @marcellamaki and @rtibbles as well.

Finally my thought about moving `UserAuthLayout` into `core` is kind of a similar problem where rather than adding one-use components into the ApiSpec, we'd be adding a one-use component into the `core` app

kolibri/plugins/user_profile/assets/src/views/ProfileEditPage.vue
```diff
@@ -62,7 +59,8 @@
         </KButtonGroup>
       </form>
     </KPageContainer>
-  </CoreBase>
+
+  </immersivepage>
```
nucleogenesis: `ImmersivePage` here

nucleogenesis (changes requested):
I've added a non-blocking set of a few thoughts about the overall code organization; I'm maybe just overthinking it :smile: 

This looks great! Thank you for the comments throughout and for the new aria attr for the snackbars! With a few minor tweaks on that and a formatting issue this should be good to go!

kolibri/core/assets/src/core-app/apiSpec.js
rtibbles: > I'm also hoping to avoid adding to the apiSpec as much as we can as it adds some bloat, especially for something not being used across multiple plugins.

This is what motivated me to suggest creating a package in `packages` for this kind of common functionality that is used in multiple plugins, but does not need to be exposed in the apiSpec as something that we are explicitly supporting for any arbitrary third party plugin.

See https://github.com/learningequality/kolibri/issues/5488

 and the additional follow up you created: https://github.com/learningequality/kolibri/issues/9589

kolibri/plugins/user_auth/assets/src/views/UserAuthLayout.vue
akolson: The aria-live was added by @marcellamaki  about 5 months ago to the `GlobalSnackbar` https://github.com/learningequality/kolibri/pull/9410. Based on what it does, i think it would be an important assistive technologies addition to the `CoreSnackbar`. Adding it there automatically adds it to the `GlobalSnackbar`. @marcellamaki any thoughts on this?

marcellamaki: @akolson that might work, I am trying to think through the implications, but I would suggest no matter what that it is worth recruiting @radinamatic to do some testing so we can insure this still works. 😄 

It took some trial and error to come up with placing the `aria-live="polite"` tag in the wrapping div, and the reason it was necessary is because the div itself must _always_ be present in the DOM in order for the `aria-live` interruption to work. Essentially what is happening (cc @nucleogenesis) is that having an empty div with the `aria-live` announces to the screen reader user ("politely" - meaning waiting until they are idle) that the content of that div has changed. In this case, when the snackbar is rendered, that the text associated with that snackbar. 

However, the `aria-live` cannot live _just_ on the component that must be announced itself, because then it is not always "listening" - it is only rendered along with the "changing" content. This implementation does not work across all screen readers (there are some browser-based reader, such as VoiceOver on Mac where this does work). I think it would be possible to add this `aria-live` to the `CoreSnackbar` but only if we are certain that in the course of this refactor, the `CoreSnackbar` component will always be present in DOM, even if it is not always visible. Otherwise, we should follow the same pattern and add an empty div that wraps around the snackbar (and thus will not interfere with the visual UI)

akolson: Thanks @marcellamaki. This is very insightful. At this point I am not too sure that adding it directly to the `CoreSnackbar` would work without making tweaks to the way its rendered to the DOM

kolibri/core/assets/src/core-app/apiSpec.js
akolson: As a way to get us started on #5488  and #9589, a new package `kolibri-common` has been added under packages. This references shared components within the kolibri plugins. References added previously to the `apiSpec.js` have been reverted. This is minimalistic so comments to improve this are welcome. Thanks cc @nucleogenesis  @rtibbles

packages/kolibri-common/index.js
```diff
@@ -0,0 +1,6 @@
+'use strict';
+
+import AppError from '../../kolibri/core/assets/src/views/AppError';
```
rtibbles: My thought was that we could just put the files themselves in the package, wouldn't even need the index.js then, as could just import `kolibri-common/components/GlobalSnackbar` or somesuch.

akolson: Ahh i see. Let me refactor this. Thanks for this clarification

akolson: Ahh i see. Let me refactor this. Thanks for this clarification

nucleogenesis (approved):
Looks good! Thanks Samson! Should be okay to force merge unless the blocking error would be fixed with a re-run of the linting.

PR #9812, opened 2022-11-05
Decorator to use drf_yasg when it's installed

jredrejo:
## Summary
Using _/api_explorer/_ with api endpoints handled by `rest_framework.views.APIView` is not too useful as it usually is not using a db model with a schema to be infered.
`drf_yasg` allows using a manual definition of the schema with the `swagger_auto_schema` but adding it to the code will break the production environment as this library is not installed.

This PR adds a new Python decorator to use `swagger_auto_schema` if it's installed and ignore it otherwise.
The decorator is documented with the params it needs and an example of usage. It supports GET,  PATCH, POST and PUT methods, but it does not allow all the [swagger_auto_schema](https://github.com/axnsan12/drf-yasg/blob/master/src/drf_yasg/utils.py#L33) supports, it's limited to the list of params, options as `required` or `operation_description` are not supported.

## Reviewer guidance
Using the new swagger_auto_schema_available decorator, decorate a get/put function of some api endpoint using  `rest_framework.views.APIView`  and check it appears correctly at the swagger /api_explorer/

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated


rtibbles (commented):
This seems fine, and looks like you've made the API a bit simpler than the default drf_yasg decorator, right?

Do you want to merge this in, or just wait for the feature branch that is using it to merge it in?

kolibri/utils/drf_utils.py
```diff
@@ -0,0 +1,73 @@
+from functools import wraps
```
rtibbles: Oh.. the only thing I can think of is whether we want to put it here or somewhere else? All our current DRF/API related machinery is in `kolibri/core` but I don't have a principled reason for it being in one place vs another.

jredrejo:
> This seems fine, and looks like you've made the API a bit simpler than the default drf_yasg decorator, right?
> 
> Do you want to merge this in, or just wait for the feature branch that is using it to merge it in?

It can wait for the branch to be merged, I cherry-picked it there so it does not matter

kolibri/utils/drf_utils.py
jredrejo: I don't have any preference, if you think moving the fileto `kolibri/core/utils` makes more sense, I'm happy with it

rtibbles: I think the subliminal distinction that I have made is that `kolibri/utils` is for general Python utils and `kolibri/core` functionality is for Django specific stuff. I don't know if the distinction holds true, but that's my interpretation of how we have roughly done things previously.

jredrejo: no problem, moved

rtibbles:
Code already merged in with #9833!

PR #9813, opened 2022-11-07
Disable forward/back buttons until locations are properly loaded.

### LINKED ISSUE #9745: Completed EPUBs  should not be  shown on the most recent
## Observed behavior
Completed content is still being filtered and returned in the recent filter.
The progress bar does not correlate with the consumed content progress. This might be the reason for the bug

## Errors and logs
Video of behaviour:

https://user-images.githubusercontent.com/1680573/192615285-b04348e7-ca04-499c-a114-47de3294114a.mp4

## Expected behavior

When a learner(User) finished interacting with the epub, it should not be returned on the most recent filter. It should be discarded from the filter 

## User-facing consequences

…

## Steps to reproduce

…

## Context

 * Kolibri version 0.15.7
 

…
cc @rtibbles

rtibbles:
## Summary
* If the forward button is pressed before the locations have finished loading, the location for the first page will not be properly set in the `visitedPages` object, and so the progress will not complete.
* This fixes this by disabling the forward and back buttons until the locations are loaded.

## References
Fixes #9745 as far as I am able to replicate it

## Reviewer guidance
Are there any other ways of reproducing #9745 that I am missing?

----

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/plugins/epub_viewer/assets/src/views/EpubRendererIndex.vue
+++ b/kolibri/plugins/epub_viewer/assets/src/views/EpubRendererIndex.vue
@@ -104,6 +104,7 @@
           <NextButton
             v-if="contentIsRtl"
             v-show="!isAtEnd"
+            :disabled="!locationsAreReady"
             :color="navigationButtonColor"
             :isRtl="contentIsRtl"
             :style="{ backgroundColor }"
@@ -112,6 +113,7 @@
           <PreviousButton
             v-else
             v-show="!isAtStart"
+            :disabled="!locationsAreReady"
             :color="navigationButtonColor"
             :isRtl="contentIsRtl"
             :style="{ backgroundColor }"
@@ -130,6 +132,7 @@
           <PreviousButton
             v-if="contentIsRtl"
             v-show="!isAtStart"
+            :disabled="!locationsAreReady"
             :color="navigationButtonColor"
             :isRtl="contentIsRtl"
             :style="{ backgroundColor }"
@@ -138,6 +141,7 @@
           <NextButton
             v-else
             v-show="!isAtEnd"
+            :disabled="!locationsAreReady"
             :color="navigationButtonColor"
             :isRtl="contentIsRtl"
             :style="{ backgroundColor }"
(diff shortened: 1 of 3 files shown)


kolibri/plugins/epub_viewer/assets/src/views/EpubRendererIndex.vue
```diff
@@ -104,6 +104,7 @@
           <NextButton
             v-if="contentIsRtl"
             v-show="!isAtEnd"
+            :disabled="!locationsAreReady"
```
AllanOXDi, line 107: Hey @rtibbles , this appears not to fix properly what we want to fix. It works better when you view a particular Epub content without pausing. But when you pause a bit and come back later to complete it . You will still get the 0.9999 we talked about.

AllanOXDi: We could start thinking about how the current location is being saved and calculated at that moment and how it's being sent to the backend.

AllanOXDi (dismissed):

kolibri/plugins/epub_viewer/assets/src/views/EpubRendererIndex.vue
rtibbles: OK, this was the kind of additional means of replicating the bug that I was interested in. When you say 'pause' do you mean the following?
1. Open an epub
2. Read a few pages
3. Leave the browser open and do something else for a minute
4. Read the remaining pages

AllanOXDi: Yes! You get the completed icon, but it will return the content as not completed.

rtibbles: I'm still unable to replicate this bug - I have added a test and a slight code tweak that I think might help, but I'm still at a loss as to exactly what is happening here.

marcellamaki (approved):
I will move forward with approving this for manual QA and tagging the beta. We can open a follow up issue with more detail if still after this, Allan is able to replicate.

Issue #9781, opened 2022-10-18
Device Settings Updates: Auto Download


marcellamaki:
## Overview 

This issue is a full-stack issue that updates the Device Settings page to include a setting where the user can determine whether or not to automatically download new materials such as lessons and quizzes. **The default behavior is to download automatically**. The setting should be set to true by default, and then the user can change it. There is an additional setting which determines a maximum amount of space that Kolibri can use for auto-downloaded content.

To avoid having to create a new migration for each field, we will be using a feature branch for this project `update-device-settings`. Please make PRs to this branch, which will later be merged into develop when the feature is complete.

### Figma 

Please see [complete figma spec ](https://www.figma.com/file/MpCG7r5PULOCJsEvnTSti8/Android-Designs-2022?node-id=1384%3A16773) for full specs of various states.

➡️ Accessibility notes: 
* The title for this section ('Auto-download') is an `<h2>`
* Please use `KComponents` to ensure that the content is properly passed on to screen readers, and that focus outlining works. 

### API changes 

The changes to the API should be made within the existing API file in the `/device` plugin: `kolibri/plugins/device/assets/src/views/DeviceSettingsPage/api.js` unless it is discovered that there would be some blocker to use this setting. As an alternative, `options.ini` could be considered.

Backend changes should include updating the model `kolibri/core/device/models.py` and related serializer `kolibri/core/device/serializers.py`

## Acceptance Criteria
- [ ]  Frontend is updated with correct strings, using existing components
- [ ]  On saving, the setting is updated on both the frontend and the backend, and a *Settings have been updated* message displays. (This snackbar should already exists and it should appear by default on a successful update to the device settings).
- [ ]  The `DeviceSettingsPage.spec.js` should be updated so that existing tests still pass (as necessary) and there should be at least one test written to cover this scenario

## References 
- [Notion project page](https://www.notion.so/learningequality/Updates-to-Device-Settings-45529cbcc4f74aeb9b19559228c97579)
- [Figma spec](https://www.figma.com/file/MpCG7r5PULOCJsEvnTSti8/Android-Designs-2022?node-id=1384%3A16773)
- [Gherkin scenarios](https://www.notion.so/learningequality/Gherkin-scenarios-02096ac9925844f8902c818cab607d4b#1959b753421648308fb467e21bdaa125)

rtibbles:
Fixed in #9833

Issue #9780, opened 2022-10-18
Device Settings Updates: Primary and Additional Storage Location(s)


marcellamaki:
## Overview 

This issue is a full-stack issue that updates the Device Settings page to include a setting where the user can determine the primary storage path. It also implements the full CRUD functionality, so a storage path can be added, deleted, and read, and the storage location which is the primary can be changed.

To avoid having to create a new migration for each field, we will be using a feature branch for this project `update-device-settings`. Please make PRs to this branch, which will later be merged into develop when the feature is complete.

### Figma 

Please see [complete figma spec](https://www.figma.com/file/MpCG7r5PULOCJsEvnTSti8/Android-Designs-2022?node-id=1384%3A16773).

➡️ Accessibility notes: 
* The title for this section ('Download on metered connection') is an `<h2>`
* Please use `KComponents` for wherever possible to ensure that the content is properly passed on to screen readers, and that focus outlining and focus shift works. (See [KDS docs](https://kolibri-design-system.netlify.app/) for more information)

### API changes 

The changes to the API should be made within the existing API file in the `/device` plugin: `kolibri/plugins/device/assets/src/views/DeviceSettingsPage/api.js` unless it is discovered that there would be some blocker to use this setting. As an alternative, `options.ini` could be considered.

Backend changes should include updating the model `kolibri/core/device/models.py` and related serializer `kolibri/core/device/serializers.py`

## Acceptance Criteria
- [ ]  Frontend is updated with correct strings, using existing components
- [ ]  On saving, the setting is updated on both the frontend and the backend, and a *Settings have been updated* message displays. (This snackbar should already exists and it should appear by default on a successful update to the device settings).
- [ ]  The `DeviceSettingsPage.spec.js` should be updated so that existing tests still pass (as necessary) and there should be at least one test written to cover this scenario

## References 
- [Notion project page](https://www.notion.so/learningequality/Updates-to-Device-Settings-45529cbcc4f74aeb9b19559228c97579)
- [Figma spec](https://www.figma.com/file/MpCG7r5PULOCJsEvnTSti8/Android-Designs-2022?node-id=1384%3A16773)
- [Gherkin scenarios](https://www.notion.so/learningequality/Gherkin-scenarios-02096ac9925844f8902c818cab607d4b#1959b753421648308fb467e21bdaa125)

jredrejo:
[...]

PR #9833, opened 2022-11-15
Device settings updates

### LINKED ISSUE #9779: Device Settings Updates: Download on a metered connection
## Overview 

This issue is a full-stack issue that updates the Device Settings page to include a setting where the user can determine whether or not to download data on metered connections. **The default behavior is to _not_ download on metered connection.** The setting should be set to false by default, and then the user can change it.

To avoid having to create a new migration for each field, we will be using a feature branch for this project `update-device-settings`. Please make PRs to this branch, which will later be merged into develop when the feature is complete.

### Figma 

[Complete figma spec ](https://www.figma.com/file/MpCG7r5PULOCJsEvnTSti8/Android-Designs-2022?node-id=1384%3A16773)

➡️ Accessibility notes: 
* The title for this section ('Download on metered connection') is an `<h2>`
* Please use `<KRadioButtons />` to ensure that the content is properly passed on to screen readers, and that focus outlining works. (You may optionally use Kolibri's `<RadioButtonGroup />` component, which wraps multiple `KRadioButton`s)

Screen Shot 2022-10-18 at 9 26 09 AM

### API changes 

The setting should be a Boolean. 

The changes to the API should be made within the existing API file in the `/device` plugin: `kolibri/plugins/device/assets/src/views/DeviceSettingsPage/api.js`

Backend changes should include updating the model `kolibri/core/device/models.py` and related serializer `kolibri/core/device/serializers.py`

## Acceptance Criteria
- [ ]  Frontend is updated with correct strings, using existing components
- [ ]  On saving, the setting is updated on both the frontend and the backend, and a *Settings have been updated* message displays. (This snackbar should already exists and it should appear by default on a successful update to the device settings).
- [ ]  The `DeviceSettingsPage.spec.js` should be updated so that existing tests still pass (as necessary) and there should be at least one test written to cover this scenario

### Out of Scope
Changing any user behavior or access _based_ on this setting is beyond the scope of this issue. This issue is only to update the interface and allow the user to save the setting, not to make any chances to the download processes based on the setting.

## References 

jredrejo:
## Summary
Implementation of the new device settings UI, according to the [design specs](https://www.figma.com/file/MpCG7r5PULOCJsEvnTSti8/Android-Designs-2022?node-id=1384%3A16773) 

image
image
image
image

## References
Closes: #9779,  #9780, #9781

## Reviewer guidance
This PR implements the needed model design changes in the backend and the frontend to modify them, but it does not act in the backend to change the behaviour of kolibri as stated in the _Out of scope_ section of #9779
However, in the case of altering the storage locations, it does change the behaviour of kolibri as the model is set to use the `options.ini` file as single source of truth. Changes for these settings are not stored in the database (they are done using a normal Django model and drf serialization but the state is saved in this file).

So, to after testing the UI does not  have any glitches. When the new settings are modified and saved, they must appear in:
- the `extra_settings` field of the `device_devicesettings` table 
- For the storage locations, in the `options.ini` file (also, when changing this file, kolibri will restart automatically)

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated


pcenov:
Hi @jredrejo I tested the above described implementation and overall it's looking great. As suggested by you I was able to verify that no changes made through the UI are saved as this requires backend changes. However as pointed out by you, adding other storage locations is possible because these are saved in the options.ini file. I also checked the auto-download and download on metered connections values in the DB and once they are modified there, they are displayed correctly in the UI.

Here are some minor issues I was able to identify (which I can report separately if you think are not in scope):

| Issue| Screenshot |
| When I visit the settings page for the first time and I click the 'Options' drop-down the page moves up by itself  | https://user-images.githubusercontent.com/79847249/203102254-b7a3a32b-614d-4930-99f0-4ea711589ed0.mp4 |
| I can’t change the primary storage location if there is only one location added. Perhaps the Change option should not be displayed in that case. Note that in the current version I can’t change the primary storage location even after I have added additional locations. |  |
| The Remove storage location option should not be available if there is only one storage location. |  |
| The default value of 0 GB is not displayed correctly | |
@radinamatic

LianaHarris360:
Hi @pcenov thank you for pointing these out. I am taking a look at what needs to be changed and then I will make those updates.

LianaHarris360:
@pcenov these are the current updated and fixed issues. The page moving up due to clicking the 'Options' drop-down is still being investigated.

| Issue fixed | Screenshot |
| If there is only one path, this is the view. | onlyOnePath|
| If there is only one writable path, the Change option is disabled. | onlyOneWritablePath |
| Remove storage location option is shown if the 'Other storage locations' section contains more than one path. This is the view if there is are multiple writable paths and no readOnly paths in the 'Other storage locations' section. | multipleWriteablePathsNoReadOnlyPaths|
| The default value of 0 GB is now displayed correctly | floatingLabelCorrectPosition|

pcenov:
Hi @LianaHarris360, I can see now that initially the Change option is disabled + the displaced 0 value is displayed correctly - excellent!
However I'm still facing the following issues:
1. If I have only one storage location and I open the Options drop-down in Other storage locations and I click 'Remove storage location' I still get to see the empty radio button:

https://user-images.githubusercontent.com/79847249/203325837-829afbaa-52b4-476d-a3c6-0fda61eea511.mp4

So it should be further discussed what exactly is the expected behavior in that case.

2. This is valid even if I have added an additional storage location:



3. In this build after refreshing the page all the additional storage locations are removed, the page get's back to its default state. My understanding was these are supposed to be stored in the `options.ini file` so why is this happening:

https://user-images.githubusercontent.com/79847249/203327767-c8b88ebe-d78a-47ea-a4a3-52f4854ea669.mp4

I also noticed the following issue which I have missed in the first round of testing (cc @jredrejo):

1. Why is the (read-only) text being displayed on its own? I would expect it to be displayed next to a read-only storage location:

LianaHarris360:
@pcenov 
1. I am not sure how it happened, but it seems that your 'other storage locations' section contains an empty path. So the empty radio button lists that path which I am assuming is just an empty string. I have tried to add an empty string path to possibly get the same error, but the current code does not allow an empty string to be added as a path.
2. The empty path will show after adding another location because the second location is added to a list of locations, which still includes the first empty path.
3. After making changes to the storage locations, the server should restart and once it reconnects, the changes should be updated and stored in the `options.ini` file. In this case, when you added the location, a 400 (bad request) error was returned and it did a soft save that showed the changes on the page but did not actually store any changes. This issue might be connected to the empty string path. 

A solution could be to validate the storage location paths to catch any that slipped through as empty strings. (@jredrejo)

jredrejo:
> Hi @LianaHarris360, I can see now that initially the Change option is disabled + the displaced 0 value is displayed correctly - excellent! However I'm still facing the following issues:
> 
> 1. If I have only one storage location and I open the Options drop-down in Other storage locations and I click 'Remove storage location' I still get to see the empty radio button:
> 
>  2022-11-22_15-03-10.mp4 
> So it should be further discussed what exactly is the expected behavior in that case.
> 
> 2. This is valid even if I have added an additional storage location:
> 
> 
> 
> 3. In this build after refreshing the page all the additional storage locations are removed, the page get's back to its default state. M y understanding was these are supposed to be stored in the `options.ini file` so why is this happening:
> 
>  2022-11-22_15-11-24.mp4 
> I also noticed the following issue which I have missed in the first round of testing (cc @jredrejo):
> 
> 1. Why is the (read-only) text being displayed on its own? I would expect it to be displayed next to a read-only storage location:
> 
> 

hello @pcenov I have seen this happens if the option has not any value assigned in the options.ini file. For some reason kolibri returns [ "" ] instead of [ ]. It should be fixed now

pcenov:
Thanks @LianaHarris360 and @jredrejo the issue with the empty value is fixed now.
So the only remaining front end issue is the one with the page moving up due to clicking the 'Options' drop-down.

@jredrejo a question for you - currently when I select the option 'Add storage location' and I check the checkbox 'Make this the primary storage location' it doesn't actually make it the primary storage location, rather after the restart of the server the location is displayed under 'Other storage location'. Is this supposed to be functional in the current build or is a pending back end task?

https://user-images.githubusercontent.com/79847249/203550382-0c9a1a09-d674-459c-ac47-f5b9817caae0.mp4

pcenov:
Hi @jredrejo @LianaHarris360 - the latest reported issue here about the storage location is working correctly now.
I have filed a follow up issue specific for the Windows build only: https://github.com/learningequality/kolibri/issues/9880
cc @radinamatic

radinamatic (approved):
Manual QA by @pcenov passed! 👏🏽 💯 
One follow-up issue remains to be solved for Windows :shipit:

marcellamaki:
@jredrejo @LianaHarris360 - thanks for your work on this! The code is really readable and easy to follow, and I don't have any questions in that regard.  🎉 

One thing I ran into when testing was adding a storage location. I tested: 
- adding a new storage location (a valid location) 
- not setting it as the primary location 

When I do this, the kolibri server stops and restarts (and all of the modal explaining this work as expected), but kolibri doesn't successfully reconnect to the server. 

Here is what appears in my terminal - it seems to just stall out after that last log - nothing else prints or reloads, no matter how long I wait.
Screen Shot 2022-12-06 at 1 12 04 PM

Is this because I'm running it on a devserver? Wondering if @pcenov experienced the same thing. (cc @radinamatic)

pcenov:
Hi @marcellamaki - I'm not able to replicate the issue you've experienced in Ubuntu - the server gets restarted successfully. I have previously reported a similar issue for Windows though.

marcellamaki:
Thanks, @pcenov! I'll go ahead with merging this, and can file a separate follow up issue as needed

PR #9808, opened 2022-11-02
Fix: mark as complete modal/logic moved into content page

nucleogenesis:
## Summary

Moves the logic for the Mark As Complete modal so that the modal lives in ContentPage now where the "content session" has been initialized using the `useProgressTracking` module. This seemed like a better alternative to trying to initialize another content session within the context of the modal itself.

Now Learn has a `showCompleteContentModal` state which is set to `false` whenever `ContentPage` is created (just to be sure it doesn't stay open if the user navigates back and to another piece of content). `LearnActivityBar` now just sets that Vuex state to true whenever the "mark as complete" button is clicked.

## References

Fixes #9795 

## Reviewer guidance

You will need two things:

- To run Kolibri with an initialized app plugin; you can use the script in #9803 being sure to set the environment variables as such:
`DJANGO_SETTINGS_MODULE=kolibri.deployment.default.settings.dev KOLIBRI_CENTRAL_CONTENT_BASE_URL=https://hotfixes.studio.learningequality.org python path/to/script.py`
- Then you can use the content @marcellamaki shared with me: `dolok-jajiv` (or one of your own if you have it setup already) 
- Go to a piece of content that has the metadata allowing it to be marked as complete and try to mark it complete. 
- You ought to see a snackbar saying you've completed the resource and the icon should change to the yellow "completed" star.

----

## Testing checklist

- [x] Contributor has fully tested the PR manually
- [ ] If there are any front-end changes, before/after screenshots are included
- [ ] Critical user journeys are covered by Gherkin stories
- [ ] Critical and brittle code paths are covered by unit tests

## PR process

- [ ] PR has the correct target branch and milestone
- [ ] PR has 'needs review' or 'work-in-progress' label
- [ ] If PR is ready for review, a reviewer has been added. (Don't use 'Assignees')
- [ ] If this is an important user-facing change, PR or related issue has a 'changelog' label
- [ ] If this includes an internal dependency change, a link to the diff is provided

## Reviewer checklist

- Automated test coverage is satisfactory
- PR is fully functional
- PR has been tested for [accessibility regressions](http://kolibri-dev.readthedocs.io/en/develop/manual_testing.html#accessibility-a11y-testing)
- External dependency files were updated if necessary (`yarn` and `pip`)
- Documentation is updated
- Contributor is in AUTHORS.md

PR #9830, opened 2022-11-11
Prevent excessive completion modals

rtibbles:
## Summary
* Reverts a previous change that caused completion modals to pop up whenever a user finished a piece of content, and not just on first completion
* Adds a next steps button which will reopen the completion modal
* Add an animation (which gets turned off for users with a reduced motion setting) which emphasizes the next steps button when a finish event is fired again
* Ensures the completion modal is shown when a learner marks a resource as complete

## References
Fixes https://github.com/learningequality/kolibri/issues/9825
Fixes #9824
Fixes https://github.com/learningequality/kolibri/issues/9840

This was discussed with @jtamiace before implementation

## Reviewer guidance
Does this resolve the issue above? Does the new button function as intended? Is the cleanup of the action bar logic in the learning activity bar sensible?

Will this be easy to migrate into develop across the refactor?

[Screencast from 11-10-2022 05:18:18 PM.webm](https://user-images.githubusercontent.com/1680573/201239976-fff85bf4-df32-48a1-86d6-78fe69b21f70.webm)

----

## Testing checklist

- [x] Contributor has fully tested the PR manually
- [x] If there are any front-end changes, before/after screenshots are included
- [ ] Critical user journeys are covered by Gherkin stories
- [ ] Critical and brittle code paths are covered by unit tests

## PR process

- [x] PR has the correct target branch and milestone
- [x] PR has 'needs review' or 'work-in-progress' label
- [x] If PR is ready for review, a reviewer has been added. (Don't use 'Assignees')
- [ ] If this is an important user-facing change, PR or related issue has a 'changelog' label
- [ ] If this includes an internal dependency change, a link to the diff is provided

## Reviewer checklist

- Automated test coverage is satisfactory
- PR is fully functional
- PR has been tested for [accessibility regressions](http://kolibri-dev.readthedocs.io/en/develop/manual_testing.html#accessibility-a11y-testing)
- External dependency files were updated if necessary (`yarn` and `pip`)
- Documentation is updated
- Contributor is in AUTHORS.md

PR #9837, opened 2022-11-16
Mark as complete fix  again

nucleogenesis:
## Summary

Fixes issue that was merged in #9808 where the progress wasn't updating properly. 

I've used the already existing "update progress" method which appears to do some other necessary business logic around properly updating. Also, it seems that the progress value must be a float because passing the int 1 resulted in it trying to update progress to 1.2 and it got mad.

## Reviewer guidance

Code review should be pretty quick. For testing, please look at #9808 for more context.
----

### DIFF
--- a/kolibri/plugins/learn/assets/src/views/ContentPage.vue
+++ b/kolibri/plugins/learn/assets/src/views/ContentPage.vue
@@ -193,7 +193,7 @@
         this.updateContentSession({ progress, interaction }).then(this.cacheProgress);
       },
       updateProgress(progress) {
-        this.updateContentSession({ progress }).then(this.cacheProgress);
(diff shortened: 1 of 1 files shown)


marcellamaki (approved):
This fixes the main problem of the option persisting, and the code looks fine, so this is good to merge. It doesn't address @pcenov's comment that 

> once it does disappear I'm still seeing the ellipsis button instead of directly showing the 'View information' icon to the right of the bookmarks icon

But for now, I think that can be non-blocking and can be addressed in a follow up issue, so that the other PRs that depend on rebasing on top of this can continue moving forward.

rtibbles:
> once it does disappear I'm still seeing the ellipsis button instead of directly showing the 'View information' icon to the right of the bookmarks icon

I think this is probably best handled in #9830

Issue #9820, opened 2022-11-09
Library page - An empty category modal is displayed for a category without child nodes


pcenov:
## Observed behavior
If I have imported a channel from Studio in which for the Category field of a resource I have selected only the main node of a certain category e.g. 'For teachers' then in Kolibri that results in seeing an empty modal window when attempting to filter by the mentioned category.

## Expected behavior
My guess is that in that case it should be possible to filter directly by the selected category by skipping the empty modal.

## Steps to reproduce the issue
1. Install the following build: https://buildkite.com/learningequality/kolibri-python-package/builds/6562
2. Import a channel from https://hotfixes.studio.learningequality.org/ containing a resource with only the main node of a category selected as a Category value.
3. Go to the Library page and attempt to filter by selecting that category value.

## Screenshot

![2022-11-09_18-15-42](https://user-images.githubusercontent.com/79847249/200883033-87664072-500d-4083-86b3-87239396cea2.png)

## Usage Details
Windows 10, Ubuntu - Chrome

pcenov:
@radinamatic @marcellamaki

rtibbles:
This was fixed in #9839

PR #9839, opened 2022-11-17
Category search directly

marcellamaki:
## Summary

If there is a valid tagged category, but no sub-categories, clicking the category button should search directly, rather than opening an empty modal. 

This also cleans up some related display code for non-production environments. 

## References
Fixes https://github.com/learningequality/kolibri/issues/9820

## Reviewer guidance
Did I miss anything? Or, any edge cases I am not considering?

search-category-directly

----

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/plugins/learn/assets/src/constants.js
+++ b/kolibri/plugins/learn/assets/src/constants.js
@@ -44,6 +44,8 @@ export const pageNameToModuleMap = {
   [PageNames.TOPICS_TOPIC_SEARCH]: 'topicsTree',
 };
 
+export const libraryCategories = {};
+
 const availablePaths = {};
 
 plugin_data.categories.map(key => {
@@ -54,9 +56,6 @@ plugin_data.categories.map(key => {
     availablePaths[path] = true;
   }
 });
-
-export const libraryCategories = {};
-
 // Create a nested object representing the hierarchy of categories
 for (let value of Object.values(Categories)
   // Sort by the length of the key path to deal with
--- a/kolibri/plugins/learn/assets/src/constants.js
+++ b/kolibri/plugins/learn/assets/src/constants.js
@@ -48,7 +48,7 @@ export const libraryCategories = {};
 
 const availablePaths = {};
 
-plugin_data.categories.map(key => {
+(plugin_data.categories || []).map(key => {
   const paths = key.split('.');
   let path = '';
   for (let path_segment of paths) {
--- a/kolibri/plugins/learn/assets/src/constants.js
+++ b/kolibri/plugins/learn/assets/src/constants.js
@@ -1,3 +1,6 @@
+import { Categories, CategoriesLookup } from 'kolibri.coreVue.vuex.constants';
+import plugin_data from 'plugin_data';
+
 // a name for every URL pattern
 export const PageNames = {
   ROOT: 'ROOT',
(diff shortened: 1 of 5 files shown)


kolibri/plugins/learn/assets/src/views/EmbeddedSidePanel/index.vue
```diff
@@ -140,20 +140,14 @@
   const resourcesNeeded = {};
   resourcesNeededShown.map(key => {
     const value = ResourcesNeededTypes[key];
-    // TODO rtibbles: remove this condition
     if (plugin_data.learnerNeeds.includes(value) || process.env.NODE_ENV !== 'production') {
```
rtibbles: You removed the TODO but not the condition! 
```suggestion
    if (plugin_data.learnerNeeds.includes(value)) {
```

kolibri/plugins/learn/assets/src/views/LibraryPage.vue
```diff
@@ -256,10 +257,48 @@
...
   import CategorySearchModal from './CategorySearchModal';
   import SearchChips from './SearchChips';
+  import plugin_data from 'plugin_data';
 
   const mobileCarouselLimit = 3;
   const desktopCarouselLimit = 15;
 
+  const availablePaths = {};
+
+  plugin_data.categories.map(key => {
```
rtibbles: Could we put this in a learn specific constants file to share between this and where I assume you copied it from?

kolibri/plugins/learn/assets/src/views/LibraryPage.vue
```diff
@@ -403,9 +442,20 @@
     methods: {
       genContentLink,
       handleShowSearchModal(value) {
-        this.currentCategory = value;
-        this.showSearchModal = true;
-        !(this.windowIsSmall || this.windowIsMedium) ? (this.sidePanelIsOpen = false) : '';
+        // for categories with sub-categories, open the modal
```
rtibbles, line 406: Conditions make sense to me.

rtibbles (changes requested):
Let's finalize the cleanup where I suggested.

The consolidation is a nice to have, but I feel like it's a little more sustainable going forwards.

rtibbles:
Oh, one other thing - should we remove the trailing `>` from the category name in the case where it won't pop up the modal?

kolibri/plugins/learn/assets/src/views/LibraryPage.vue
marcellamaki: yes, that is probably wise!

kolibri/plugins/learn/assets/src/views/EmbeddedSidePanel/index.vue
marcellamaki: updated

marcellamaki:
> Oh, one other thing - should we remove the trailing > from the category name in the case where it won't pop up the modal?

yes, I've updated accordingly

kolibri/plugins/learn/assets/src/constants.js
```diff
@@ -40,3 +43,59 @@ export const pageNameToModuleMap = {
...
+plugin_data.categories.map(key => {
+  const paths = key.split('.');
+  let path = '';
+  for (let path_segment of paths) {
+    path = path === '' ? path_segment : path + '.' + path_segment;
+    availablePaths[path] = true;
+  }
+});
+
+export const libraryCategories = {};
```
rtibbles: I also hadn't realized we now had it in 3 different places, so I was inadvertently invoking the Rule of Three!

kolibri/plugins/learn/assets/src/views/EmbeddedSidePanel/index.vue
```diff
@@ -59,7 +59,7 @@
             :disabled="availableRootCategories &&
               !availableRootCategories[key] &&
               !isKeyActive(key)"
-            iconAfter="chevronRight"
+            :iconAfter="hasNestedCategories(key) ? 'chevronRight' : null"
```
rtibbles: You also have the `category` value available here which is the key for the `libraryCategories` object.

kolibri/plugins/learn/assets/src/views/EmbeddedSidePanel/index.vue
```diff
@@ -305,6 +300,10 @@
       noCategories() {
         this.$emit('input', { ...this.value, categories: { [NoCategories]: true } });
       },
+      hasNestedCategories(key) {
```
rtibbles: If this is `category` instead of `key` as I suggest above, this function can just be:
```
return Object.keys(libraryCategories[category].nested).length > 0;
```

rtibbles (approved):
1 minor tweak possible, but not required.

radinamatic:
Tested in Firefox on Ubuntu 18.04, no empty modals when filtering by main categories, just the direct search results! 💯 👍🏽 

search-category

radinamatic (approved):
Manual QA passes, good to go! 👏🏽 :shipit:

radinamatic:
New top level category is clickable in the modal, filtering working as expected 👏🏽 🥇 

search-category2

rtibbles:
For follow up after this PR, we should look into making the elements in the modal more clickable.

I also have questions about whether the icons should be clickable, and whether the text should be inline with the icons? Would be good to check in with @jtamiace to follow up on this.

PR #9883, opened 2022-11-30
fix too much empty space

### LINKED ISSUE #9500: Mobile view - Too much empty space
## Observed behavior
See the attached screenshot. Tested in real devices as well as using Chrome's dev tools emulator.

1. Install the following build: https://buildkite.com/learningequality/kolibri-python-package/builds/5658 from https://github.com/learningequality/kolibri/pull/9222
2. Go to the Library > Browse channel

## Usage Details
Windows 10 and Ubuntu 20 - Chrome, Firefox

AllanOXDi:
## Summary
This PR fixes too much space on the mobile view when browsing through the channel.

## References
Closes #9500
Before
image
After

Screenshot 2022-11-30 at 20 23 37

## Reviewer guidance

1. Go to the Library > Browse channel.
2. Reduce your screen size to mobile view and observe

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/plugins/learn/assets/src/views/TopicsPage/TopicSubsection.vue
+++ b/kolibri/plugins/learn/assets/src/views/TopicsPage/TopicSubsection.vue
@@ -103,7 +103,7 @@
 <style lang="scss" scoped>
 
   .folder-header-link {
-    margin-top: 15px;
+    margin-top: 25px;
 
     /deep/ .link-text {
       text-decoration: none !important;
--- a/kolibri/plugins/learn/assets/src/views/TopicsPage/TopicSubsection.vue
+++ b/kolibri/plugins/learn/assets/src/views/TopicsPage/TopicSubsection.vue
@@ -103,6 +103,8 @@
 <style lang="scss" scoped>
 
   .folder-header-link {
+    margin-top: 15px;
+
     /deep/ .link-text {
       text-decoration: none !important;
     }
--- a/kolibri/plugins/learn/assets/src/views/TopicsPage/index.vue
+++ b/kolibri/plugins/learn/assets/src/views/TopicsPage/index.vue
@@ -38,26 +38,49 @@
         />
 
         <div class="card-grid">
-          <!-- Filter buttons - shown when not sidebar not visible -->
-          <div v-if="!windowIsLarge" data-test="tab-buttons">
+
+          <!-- catering for a more smaller screen  -->
+          <div v-if="windowIsSmall" data-test="tab-buttons">
             <KButton
               v-if="topics.length"
               icon="topic"
               data-test="folders-button"
-              class="overlay-toggle-button"
+              class="btn-space-mobile-1"
               :text="coreString('folders')"
               :primary="false"
               @click="toggleFolderSearchSidePanel('folder')"
             />
             <KButton
               icon="filter"
-              class="overlay-toggle-button"
+              class="btn-space-mobile-1"
               data-test="filter-button"
               :text="filterTranslator.$tr('filter')"
               :primary="false"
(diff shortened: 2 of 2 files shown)


AlexVelezLl:
Hi! I was checking and I think that with this change the paddings in PC browsers worsens a bit. I think media queries could be used to avoid unexpected changes on other parts.

Before: 
image
After:
image

AllanOXDi:
I think this could be a different issue the we could open. I am not sure if this fix is the one breaking it. cc @marcellamaki

marcellamaki (changes requested):
Hi @AllanOXDi. @AlexVelezLl makes a correct point, and his suggestion about making sure your adjustments don't affect other window sizes is a good idea. 

The `$toolbar-height` variable you change is used in several places and for some other calculations, so there will likely be unintended consequences of changing it - like in the desktop view of the topics page. Rather than using media queries, which is a common strategy that Alex suggests, you can use our `responsiveWindowMixin` which you can see used in this file already. It accomplishes the same thing, by using KDS. The `windowIsLarge`, `windowIsMedium`, and `windowIsSmall` values that you may have seen referenced throughout the Kolibri frontend come from this mixin (which is imported as [part of KDS](https://kolibri-design-system.netlify.app/layout/#responsiveness)). These are designated "breakpoints" which we use across the app for consistency. `windowIsLarge`, `windowIsMedium`, and `windowIsSmall` return a Boolean. This bug is a great way to learn more about how KDS helps with managing responsive design, as we use this extensively in the Kolibri front-end.

One way you could incorporate this mixin into your fix might be to set a condition if `windowIsSmall` is true. You can use a [ternary operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Conditional_Operator) pattern of [setting the class based on if a particular condition is true that follows this example](https://github.com/learningequality/kolibri/blob/develop/kolibri/core/assets/src/views/ExamReport/index.vue#L95), and then in your styles, define two different classes - one to be used for small windows, and one to be used otherwise. That way, you don't need to change the variable at all, and can specify the spacing specifically for small/mobile views. 

Overall, your idea about changing the spacing is right, and the fix looks good on mobile! It's just about making sure the changes are specific enough that they _only_ impact mobile, and updating the code a bit to be sure that is the case. Let me know what questions you might have!

AllanOXDi:
Hi @AlexVelezLl and @marcellamaki, I have added more changes to fix the edge case.

pcenov:
Hi @AllanOXDi I confirm that the mobile view is fixed as specified above. The issue mentioned by @AlexVelezLl though is still present in the channel details modal, especially when viewing the contents of a folder:

AllanOXDi:
Hi @pcenov, please let me know if this is the right spacing.

Screenshot 2022-12-21 at 17 04 38

pcenov:
Hi @AllanOXDi it seems that you should add 5 pixels more but I am just roughly comparing it to the latest develop build so it would be best if you install that yourself and measure the spacing in all locations where your code changes are making an impact:

pcenov:
Hi @marcellamaki adding here a screenshot for comparison between the 0.15 version of the mobile view and the latest build from this PR so that it can be decided how we're going to proceed:

marcellamaki:
Thank you @pcenov! @AllanOXDi - I am sorry to say that the work you have done here, in this case isn't the thing that needs to be fixed, now. I will message you on slack about what we can do instead, and we can open a new PR for it. So, we will be fixing the bug that is causing the "banner" to be hidden, not actually the spacing. So sorry that I missed this earlier with the issue!

PR #9891, opened 2022-12-02
Properly set language direction as string rather than bidi boolean.

### LINKED ISSUE #9887: Ensure that our language globals `lang_direction` property is always a string
The `languageGlobals` object that is being given to `plugin_data` has been returning languages where their `lang_direction` properties are Boolean rather than a string. Using the i18n module, we should be able to rely on that value always being on of 'rtl' or 'ltr'. 

This issue was mitigated by @akolson ( https://github.com/learningequality/kolibri/pull/9863/commits/14e4f8b72e3e1b8544c014114cc7e5efcada313d ) but it always assumes that unless the string is specifically the `languageDirections.LTR` then it should be RTL.

Not sure the implications of this as it may be working reliably, however it would be good to ensure our data structures for the language globals are consistent and predictable.

rtibbles:
## Summary
* Corrects what seems to be a 5 year (maybe more) old bug that improperly sets the language_direction attribute onto the available languages object
* Probable that this does not get noticed previously because the `isRtl` Vue prototype method uses the top level `lang_dir` variable, rather than using the available languages.

## References
Fixes #9887

## Reviewer guidance
This reverts the previous update to the `getLangDir` function but should retain the same overall behaviour.

----

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/core/assets/src/utils/i18n.js
+++ b/kolibri/core/assets/src/utils/i18n.js
@@ -58,16 +58,7 @@ export let currentLanguage = defaultLocale;
 export let languageDirection = languageDirections.LTR;
 
 export function getLangDir(id) {
-  /*
-    The lang_direction property is returned as a boolean and not as a string.
-    The previous logic would fail for non strings thus the change to allow for both
-    string and non string to work as values of lang_direction.
-  */
-  let langDirection = (availableLanguages[id] || {}).lang_direction || languageDirections.LTR;
-  if (typeof langDirection !== 'string' && langDirection) {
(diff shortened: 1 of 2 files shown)


akolson (approved):
LGTM!. Thanks @rtibbles

rtibbles:
Forced pushed to rebase onto latest develop to fix linting failures.

nucleogenesis (approved):
Awesome thanks @rtibbles

PR #9895, opened 2022-12-02
Pins ubuntu versions to ubuntu-20.04 for any task needing Python 2.7.

rtibbles:
## Summary
`ubuntu-latest` is now `ubuntu-22.04`
The `setup/python` action does not have assets for Python 2.7 or 3.6 for Ubuntu 22.04
This pins any actions needing those two Python versions to 20.04, and upgrades Python 3.6 to 3.9 for postgres tests.


bjester (approved):
LGTM

PR #9896, opened 2022-12-02
Setup Wizard Refactor / Updates

nucleogenesis:
# Setup Wizard

**Description revised 31/1/2023**

## Overview

- Introduces updated designs; layouts/architecture supporting new styling
- Refactors the XState machine to manage more context and each stage of the wizard's flow
- Introduces a script and commands to allow starting the devserver in "App Mode" (more on this below) (thanks @rtibbles !)
- Does not include Learn Only Device flow -- but it does lay some foundation for the machine, routes, and templates for the files to be used
- Implements all combinations of following flows: "On my own", "App Mode", "Import Facility", "Create Facility"

### TODO, Follow up work

To let this aging PR find its way into `develop` finally, despite not being finished, I will enumerate the remaining tasks to complete the work.

I am continuing on with the following work. The LOD flow will be done ASAP, some of the other work may become issues to be prioritized in our backlog. _Critical functionality will be done ASAP._

- [ ] Learn-only (Subset of Users) device forms and flows
- [ ] Step X of Y -- I started a path on this but not sure my idea will work out
- [ ] The file layout in the plugin could use some TLC to simplify the architecture
- [ ] The Select Facility page should be skipped whenever there is only 1 Facility available on the network

#### In the browser

The last step does not actually **provision** the device until you click Continue so long as you're in dev mode -- you will see the Setup Wizard Debugger 3000 that will show some data to you and allow you to restart the flow. This helped me a lot during testing of the UI altogether.

You can also always clear your browser's storage -- or specifcally, the `localStorage.savedState` data which is updated with the entire state machine on each transition between states.

#### In the database

`kolibri manage deprovision` will wipe your database for you

### "App Mode"

Be sure  you run `kolibri plugin enable kolibri.plugins.app` to enable the App plugin.

Then you will need to run this from source using the following command:

`yarn app-devserver`

This will start the server and should print out a URL like `http://localhost:8000/initialize/somelonghash123` -- you will need to follow that URL in order to use the server in "App Mode".

**App Mode** here will result in the "On My Own" and "New Facility" superuser being created automatically, skipping the user creation form.

### Flows To Try

**Each should be done with and without the App Mode to be sure all works as expected**

- [ ] On My Own - Should not have to fill out a user form
- [ ] Group Learning -> Full Device -> New Facility
- [ ] Group Learning -> Full Device -> Import Facility -> Create new superuser 
- [ ] Group Learning -> Full Device -> Import Facility -> Authenticate as existing superuser from imported facility


pcenov:
Hi @nucleogenesis, could you let me and @radinamatic whether this is ready for testing? I did try the 'On my own' flow but after I finish the process I am not able to sign in with the credentials I've provided in the wizard:

https://user-images.githubusercontent.com/79847249/205679629-3ac94c8e-8ef8-4429-a670-7b454478bd3f.mp4

nucleogenesis:
Thanks @pcenov I'll test that out today again to see where the issue might be and ping you again when ready.

nucleogenesis:
@pcenov I've resolve the issue where you cannot login

pcenov:
Hi @nucleogenesis here are my observations from today's round of testing. Let me know if you need help in replicating. Posting them here as a long comment, but if you wish I can create follow-up issues as well:

| Issue| Screenshot |
| On My Own - The 'Device' plugin is not available while it should as I cannot import channels |  |
| Group Learning -> Full Device -> Import all data from an existing learning facility - Same as above, I am able to import a facility but the 'Device' plugin is not available  | https://user-images.githubusercontent.com/79847249/206229563-71f398f4-97ce-426e-8dbc-8fb1ceb3f6c1.mp4 |
| Group Learning -> Full Device -> Import all data from an existing learning facility - Cannot login with the new super admin that I've created  | https://user-images.githubusercontent.com/79847249/206228437-d733d14f-5926-4aac-901c-9be510400c6f.mp4 |
| Group Learning -> Full Device -> Create New Learning Facility - Cannot finish the setup process. I tried setting up both a formal and non-formal facility plus changing the default presets.   | https://user-images.githubusercontent.com/79847249/206231152-63130eef-31f9-42a2-920f-cfc5856c142b.mp4  Logs [logs-GrLrn-FullDevice_CreateNew.zip](https://github.com/learningequality/kolibri/files/10177818/logs-GrLrn-FullDevice_CreateNew.zip) |

kolibri/plugins/setup_wizard/api_urls.py
```diff
@@ -1,15 +1,15 @@
-from django.conf.urls import include
-from django.conf.urls import url
 from rest_framework import routers
 
 from .api import FacilityImportViewSet
+from .api import SetupWizardResource
 from .api import SetupWizardRestartZeroconf
 
-router = routers.DefaultRouter()
+router = routers.SimpleRouter()
```
marcellamaki, line 7: is this a Vue upgrade, or a change you decided on? I don't have any context on what the difference, if any, is

kolibri/plugins/setup_wizard/assets/src/machines/wizardMachine.js
```diff
@@ -126,13 +143,16 @@ export const wizardMachine = createMachine({
...
+          // `cond` takes a function that returns a Boolean, continuing to the
+          // `target` when it returns truthy
           cond: isOnMyOwnOrGroup,
           target: 'defaultLanguage',
         },
         {
           cond: isGroupSetup,
           target: 'deviceName',
         },
+        // There is no fallback path here; if neither `cond` above is truthy, this will break
```
marcellamaki: This seems both correct, and also, potentially tricky. Are there other places where we are using a state machine that we have handled a similar scenario?

kolibri/plugins/setup_wizard/assets/src/views/SetupWizardIndex.vue
```diff
@@ -114,6 +114,8 @@
       this.service.start(savedState);
 
       this.service.onTransition(state => {
+        console.log('ontransition', state);
+        console.log(this.service);
```
marcellamaki, line 118: logs

kolibri/plugins/setup_wizard/assets/src/views/importFacility/ImportAuthentication.vue
```diff
@@ -23,21 +23,23 @@
...
+      facility() {
+        return this.wizardService._state.context.selectedFacility;
+      },
+      device() {
+        return this.wizardService._state.context.importDevice;
+      },
+    },
+    mounted() {
+      console.log('ImportAuthentication', this.wizardService.state);
+    },
```
marcellamaki, line 42: log

kolibri/plugins/setup_wizard/assets/src/views/importFacility/SelectFacilityForm.vue
```diff
@@ -75,24 +67,30 @@
...
       },
+      importDeviceId() {
+        return this.wizardService._state.context.importDeviceId;
+      },
       selectedFacility() {
         return this.facilities.find(f => f.id === this.selectedFacilityId);
       },
     },
+    mounted() {
+      console.log('ImportAuthentication', this.wizardService.state);
```
marcellamaki, line 78: and here. now I'm wondering if there is a particular reason these are being logged? if so - sorry :D

kolibri/plugins/setup_wizard/assets/src/views/importFacility/SelectFacilityForm.vue
```diff
@@ -104,6 +102,15 @@
       handleAddressSubmit(address) {
         this.fetchNetworkLocation(address.id).then(() => (this.showSelectAddressModal = false));
       },
+      handleContinue() {
+        //$emit('click_next', { facility: selectedFacility, device: device })
+        console.log(this.wizardService._state);
+        console.log(this.wizardService.state);
```
marcellamaki, line 108: some cleanup

kolibri/plugins/setup_wizard/assets/src/views/onboarding-forms/DefaultLanguageForm.vue
```diff
@@ -24,8 +24,6 @@
     inject: ['wizardService'],
     methods: {
       handleSubmit() {
-        const currentLanguageId = this.$store.state.onboardingData.language_id;
-        this.$store.commit('SET_LANGUAGE', currentLanguageId);
         this.wizardService.send('CONTINUE');
```
marcellamaki, line 27: can you remind me where the language setting is handled now? is it just in the pop up that is outside the scope of the flow?

kolibri/plugins/setup_wizard/assets/src/views/onboarding-forms/PersonalDataConsentForm.vue
```diff
@@ -48,7 +48,24 @@
...
+        console.log('OK NEXT:', nextEvent);
+        if (!nextEvent) {
+          console.error(
+            'Please provide the event you expect where you are using this Component in',
+            "the state machine in the meta field's `nextEvent` property."
+          );
+          return;
+        }
+        // TODO Add an Error State with a "Start over" button? Something better than
+        // "this silently fails" if something goes wrong for the user
```
marcellamaki, line 67: Yes, I was wondering about this too - and I think maybe we chatted about it in a pair programming session. not a blocker, but would possibly be good to discuss before release

kolibri/plugins/setup_wizard/assets/src/views/onboarding-forms/SettingUpKolibri.vue
```diff
@@ -1,50 +1,133 @@
...
+      if (!this.isLearnOnlyDevice) {
+        // 2) Group learning, Full device setup
+        if (this.isNewFacility) {
+          // 2.1) New Facility
+          this.createAndProvisionNewFullFacilityDevice();
+        } else {
+          // 2.2) Import Facility
+          // We already have the facility imported, just provision and redirect
+          this.provisionDevice();
+        }
```
marcellamaki: not at all blocking - I think having some comments here is great, but maybe you don't need all of them? I think the ones that are Booleans of a rather descriptive computed prop are okay to skip, personally, and doing so makes it a bit easier to actually look at the _code_ . But I think it other places like 

> // We already have the facility imported, just provision and redirect

where the logic is something you have to have kept in your brain as you were reading, it's actually really helpful.

kolibri/plugins/setup_wizard/assets/src/views/onboarding-forms/SettingUpKolibri.vue
```diff
@@ -1,50 +1,133 @@
...
+      createAndProvisionNewFullFacilityDevice() {
+        SetupWizardResource.createsuperuser(this.facilityUserData).then(() =>
+          this.provisionDevice()
+        );
+      },
+      provisionDevice() {
+        SetupWizardResource.provisiondevice(this.deviceProvisioningData)
+          .then(() => {
+            // FIXME In dev mode, we'll wait 5 seconds before moving along so we can see the page
+            // ... maybe we keep this?
```
marcellamaki: I think waiting a few seconds might be nice, even if it is mostly like a "foaming toothpaste" effect. Maybe there is some UX research about this? sometimes realllyy short wait times feel glitchy to me.

kolibri/plugins/setup_wizard/assets/src/views/onboarding-forms/UserCredentialsForm.vue
```diff
@@ -13,7 +13,12 @@
          inside a slot default
     -->
     <slot name="form">
+      <!-- Hiding the fullname and username textboxes, but their values are filled in and presumed
+           valid if we're given the user that we're taking credentials for (ie, just entering
```
marcellamaki, line 18: > presumed valid

if there is an error, what happens? do we get the validation errors and the form with fill-in textboxes again?

kolibri/plugins/setup_wizard/assets/src/views/onboarding-forms/UserCredentialsForm.vue
```diff
@@ -91,63 +96,78 @@
...
+      },
+      handleContinue() {
+        // Be sure to sync this as it updates Vuex with the form data so other components
+        // have access to it
+        this.syncOnboardingData();
+        if (!this.formIsValid) {
+          if (this.$refs.password && this.$refs.password.isValid) {
+            // In this case, we have been given a separate password form in our #form slot
+            // and it's ref can tell us if it's valid and whether we should continue
+            console.log('OKGO');
```
marcellamaki: log

marcellamaki (commented):
Hi @nucleogenesis this is great! I did an initial code read through and left some comments (some were just annoying "console log!") and some are more open questions. 

I don't have any major blockers right now (although, I defer proper review of the python code to...anyone not me). I'm going to do a second read and some manual QA on Monday and will write a follow up review then.

kolibri/plugins/setup_wizard/api_urls.py
nucleogenesis: I think @rtibbles might be better able to explain the difference -- but if I didn't do this, something broke and when I changed it things worked :)

kolibri/plugins/setup_wizard/assets/src/machines/wizardMachine.js
nucleogenesis: Not that I know of -- ideally we would have some kind of error state for the Setup Wizard that we could put people into with a "start over" button and some information they can send us to help debug? This would require some design and new strings but could be a worthwhile thing to add? (cc @jtamiace)

kolibri/plugins/setup_wizard/assets/src/views/importFacility/SelectFacilityForm.vue
nucleogenesis: Nope I just didn't do a log audit before making the PR :sweat:

kolibri/plugins/setup_wizard/assets/src/views/onboarding-forms/DefaultLanguageForm.vue
nucleogenesis: The `LanguageSwitcherList` makes the call to the backend to set the default language. This was only stored in Vuex for the frontend's reference previously I believe

kolibri/plugins/setup_wizard/api_urls.py
rtibbles: The DefaultRouter does some introspection to generate an API view at the root of where the router is mounted: https://www.django-rest-framework.org/api-guide/routers/#defaultrouter

My assumption is that this introspection probably breaks because of our own tweaks and adjustments to DRF.

kolibri/plugins/setup_wizard/assets/src/views/onboarding-forms/SettingUpKolibri.vue
nucleogenesis: Thoughts @jtamiace ?

nucleogenesis: And @tomiwaoLE ?

kolibri/plugins/setup_wizard/assets/src/views/onboarding-forms/UserCredentialsForm.vue
nucleogenesis: Basically, IMO, this can use some refactoring because it basically uses one form in two use cases:
- Get a password for an imported user (ie, we imported a facility and they want to use an existing super admin account on that faciliy, so we get the password from them)
- Creating a whole new superuser

The convenient thing is that in the end the `SettingUpKolibri` component gets the same value (namely `password`) from the same place -- but it might just be more clear if we separate those two needs into separate components and data storage (and/or state machine context?)

*But to answer the actual question...*

So the assumption here is that `username` and `fullname` are valid primarily because they are values that we got from another Kolibri in this particular case where we hide the form elements.

kolibri/plugins/setup_wizard/assets/src/views/onboarding-forms/SettingUpKolibri.vue
jtamiace: What part of the onboarding flow is this? Having a hard time telling

kolibri/plugins/setup_wizard/api.py
```diff
@@ -32,6 +32,141 @@ def has_permission(self, request, view):
         return subset_of_users_device
 
 
+class SetupWizardResource(ViewSet):
```
rtibbles, line 35: This appears to be replacing the core provision device endpoint: https://github.com/learningequality/kolibri/blob/develop/kolibri/core/device/api.py#L78

which has decent test coverage of different pathways: https://github.com/learningequality/kolibri/blob/develop/kolibri/core/device/test/test_api.py#L50

I would much rather we add these conditionals to that endpoint and test them, rather than creating a brand new endpoint here.

kolibri/core/auth/models.py
```diff
@@ -1415,6 +1415,23 @@ def add_coaches(self, users):
     def remove_coach(self, user):
         self.remove_role(user, role_kinds.COACH)
 
+    @classmethod
+    def get_or_create(cls, facility_name):
```
rtibbles, line 1419: I am not sure what this class method is giving us over just using `Facility.objects.get_or_create(name=facility_name)`

kolibri/plugins/setup_wizard/assets/src/views/onboarding-forms/SettingUpKolibri.vue
```diff
@@ -1,50 +1,133 @@
...
-        window.location = window.urls['kolibri:kolibri.plugins.user_auth:user_auth']();
+        // 3) Group learning, learn only device
+        this.provisionDevice();
       }
     },
     methods: {
-      isOnMyOwnSetup() {
-        return this.wizardService.state.context.onMyOwnOrGroup == UsePresets.ON_MY_OWN;
+      createAndProvisionOnMyOwnUserDevice() {
+        SetupWizardResource.createonmyownuser(this.facilityUserData).then(() =>
```
rtibbles: I don't understand why the user creation and provisioning steps have been separated in these three instances, when they are then just chained without any intermediary state? This could simply use the existing endpoint without any issues, and one fewer request.

rtibbles (commented):
I am confused by the backend recreation here and the separation of the provision and user creation steps.

The provisioning step needs to be separated from import, but not from the flows where new users are created. I would much rather see any tweaks be made in the existing device provisioning endpoint rather than creating entirely novel, untested code paths here.

If any refactor is warranted, it is probably more that the flows that involve user creation might be better done as asynchronous provisioning tasks (which is when the separation of finalized provisioning from all the main setup would be warranted, so as to be able to monitor tasks as an unlogged in user until provision has been finalized).

kolibri/core/auth/models.py
nucleogenesis: Oh just me not knowing or checking if that's a built-in function

pcenov:
Hi @nucleogenesis, here's a list with issues I was able to identify in the [latest .deb build asset](https://buildkite.com/learningequality/kolibri-debian/builds/5826):

| Issue| Screenshot |
| Group learning > Full device > Create a new learning facility - The user entered 'facility name' value is replaced by a ‘Default Facility Name’ value | https://user-images.githubusercontent.com/79847249/216097560-4397e3eb-aa77-4e4b-b425-f5418cd2a49f.mp4 |
| Group learning > Full device > Create a new learning facility - The already selected/entered data is not displayed when using the ‘Go back’ button | https://user-images.githubusercontent.com/79847249/216099443-9d1e9f8e-9f1b-4436-a273-275baacb6295.mp4 |
| Group learning > Full device > Create a new learning facility - Non-formal - When setting up a ‘Non-formal facility’ the ‘Yes’ option should be selected as default while currently for both ‘Non-formal’ and ‘Formal’ it is set to ‘No’ | https://user-images.githubusercontent.com/79847249/216097800-cf8dfc6b-d448-4059-bb65-9f3daf76139e.mp4 |
| Group learning > Full device > The imported facility is duplicated |  https://user-images.githubusercontent.com/79847249/216098375-7baedc6a-b26f-4dc6-a018-bbbe9c5ae301.mp4 |
| Group learning > Full device > Import all data from an existing learning facility - the ‘Go back’ option is not functional, getting an error in the console |  https://user-images.githubusercontent.com/79847249/216098139-077c3018-b8d7-4306-a8fd-25976ccfaad0.mp4 |
| Group learning > Full device > Import all data from an existing learning facility - ‘Unknown’ displayed while importing  |  |

nucleogenesis:
@pcenov I have addressed all of the issues noted except the loading issues -- I'm seeing "unknown" as well as it goes up to  5000% before completing so I'll be working on that.

Just a note though that unless this is an error specifically in the code here, I may address it in a separate PR -- will update tomorrow.

I will also look for some live code review from @rtibbles 

From that I will make quick fixes and make issues for anything else so that this can be merged and others can take on some work in the setup wizard to become more familiar.

pcenov:
Hi @nucleogenesis, there are still a few issues in the [latest .deb build asset](https://buildkite.com/learningequality/kolibri-debian/builds/5893#0186198f-10b3-417b-9c2f-09ae37a24ca1) the main one being the facility import option which is not functioning at all now:

| Issue| Screenshot |
| Group learning > Full device > Create a new learning facility - Now I am getting a 500 server error if  'Enable passwords on learner accounts' is set to 'Yes' | https://user-images.githubusercontent.com/79847249/216993472-f8b5bd79-26ed-4776-9e9b-18578280235b.mp4 |
| Group learning > Full device > Import all data from an existing learning facility -  Now it's not possible to import a facility at all. I tried several times but I am always getting a 400 bad request error for deviceprovision | https://user-images.githubusercontent.com/79847249/216994946-12554f28-478c-43f8-bf02-d7122e15de8b.mp4 |
| Group learning > Full device > Import all data from an existing learning facility - the ‘Go back’ option is not working | https://user-images.githubusercontent.com/79847249/216996078-04bafdf6-48d9-47e7-b23a-e98d5bb7ebb2.mp4 |

marcellamaki:
Hi @pcenov - thank you so much for the very thorough testing on all of this work. Can you clarify - are these problems _only_ present on the deb asset, or have you just started you testing beginning with that asset first, and you will follow up with additional updates if these problems also exist in the other assets as well? thank you!

pcenov:
Hi @marcellamaki the issues are valid for both the .deb and .exe assets, not OS specific.

nucleogenesis:
@pcenov -- I found a couple other bugs in addressing your comments and have updated the PR -- should be ready for a retest of the workflows On My Own, Full Facility and Facility Import.

Thanks!

pcenov:
Hi @nucleogenesis I've retested the main scenarios for the implemented workflows On My Own, Full Facility and Facility Import and these are the outstanding issues, all of which have been previously reported and can be filed as separate issues:

| Issue| Screenshot |
| Group Learning -> Full Device -> Import all data from an existing learning facility - As previously reported, I am able to import a facility but the 'Device' plugin is not available | https://user-images.githubusercontent.com/79847249/218780914-f993c43e-7239-4adf-9ed4-11512ae6388b.mp4 |
| Group learning > Full device > Import all data from an existing learning facility - the ‘Go back’ option is not working correctly | https://user-images.githubusercontent.com/79847249/218781113-14b3be8b-ae74-486f-a640-abf07e446d3a.mp4 |
| Group learning > Full device > Import all data from an existing learning facility - ‘Unknown’ displayed while importing, incorrect percentage value  |  |

nucleogenesis:
Thank you @pcenov -- I'll merge once this is approved

nucleogenesis:
Noting also that each of those will be either addressed in the Learn-only Device follow-up PR or made into a separate follow-up issue

radinamatic (approved):
It's been a long road but seems like we're almost there, with just a few over-spills 🙂 
Thank you all! 👏🏽 💯 :shipit:

PR #9897, opened 2022-12-03
Minor cosmetic issues with the visible focus outline fixed

### LINKED ISSUE #9484: Keyboard navigation - Minor cosmetic issues with the visible focus outline
## Observed behavior
See the attached screenshots.

1. Install Kolibri v0.15.3-beta2
2. Observe the visible focus outline when importing a channel, changing the permissions or browsing the library

## Screenshots
Device > Permissions:

Device > Facilities > Import facility:

Learn > Library and Learn > Library > Browse channel

## Usage Details
Windows 10 and Ubuntu 20 - Chrome, Firefox

AllanOXDi:
## Summary
This PR fixes minor cosmetic issues with the visible focus outline.
closes #9484
## References
Before:
image## Reviewer guidance

image

After:
Screenshot 2022-12-05 at 16 47 18

Screenshot 2022-12-05 at 16 48 07

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/plugins/coach/assets/src/views/common/HeaderTabs/HeaderTab.vue
+++ b/kolibri/plugins/coach/assets/src/views/common/HeaderTabs/HeaderTab.vue
@@ -79,6 +79,7 @@
     border-top-left-radius: $radius;
     border-top-right-radius: $radius;
     outline: none;
+    outline-offset: -2px !important;
     transition: background-color $core-time ease;
 
     @media print {
--- a/kolibri/plugins/device/assets/src/views/ManagePermissionsPage/UserGrid.vue
+++ b/kolibri/plugins/device/assets/src/views/ManagePermissionsPage/UserGrid.vue
@@ -39,7 +39,7 @@
               <KButton
                 appearance="flat-button"
                 :text="permissionsButtonText(user)"
-                style="margin-top: 6px;"
+                style="margin: 6px;"
                 @click="goToUserPermissionsPage(user.id)"
               />
             </td>
--- a/kolibri/plugins/learn/assets/src/views/TopicsPage.vue
+++ b/kolibri/plugins/learn/assets/src/views/TopicsPage.vue
@@ -76,10 +76,12 @@
             v-if="topics.length"
             :text="coreString('folders')"
             :to="foldersLink"
+            class="tabspace"
           />
           <HeaderTab
             :text="coreString('searchLabel')"
             :to="topics.length ? searchTabLink : {} "
+            class="tabspace"
           />
         </HeaderTabs>
       </div>
@@ -923,6 +925,10 @@
   .chip {
     margin-bottom: 8px;
     margin-left: 8px;
+  }
+  .tabspace {
+    padding: 0;
+    margin-bottom: 14px;
   }
 
 </style>
--- a/kolibri/plugins/learn/assets/src/views/TopicsPage.vue
+++ b/kolibri/plugins/learn/assets/src/views/TopicsPage.vue
@@ -76,12 +76,10 @@
             v-if="topics.length"
             :text="coreString('folders')"
             :to="foldersLink"
-            class="tabspace"
           />
           <HeaderTab
             :text="coreString('searchLabel')"
             :to="topics.length ? searchTabLink : {} "
-            class="tabspace"
           />
         </HeaderTabs>
       </div>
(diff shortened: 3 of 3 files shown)


marcellamaki (changes requested):
Hi @AllanOXDi - this looks good. I see the change you have on the manage permissions page grid. 👍  However, I think maybe you forgot to push some of your changes - the "Add new address" updates aren't here.

AllanOXDi:
I think that's the only change I made. It should be able to work as intended.

pcenov:
Hi @AllanOXDi there are 3 specific issues mentioned in https://github.com/learningequality/kolibri/issues/9484 while the only one that is actually fixed by this PR is the one on the 'Device permissions' page. How do we proceed with the other issues?

AllanOXDi:
@pcenov, This commit should be able to address it. I am worried about the border moving up. Tagging @jtamiace  for her insight on this.

https://user-images.githubusercontent.com/103313919/206770459-3194a5d2-b63e-497a-b699-07da3c1a0512.mp4

cc @radinamatic

jtamiace:
@AllanOXDi I'm not sure what the solution here is via code, but it would be ideal if the border could remain on the gray line.

AllanOXDi:
This would be the best option . But that would require us to make changes from the header tab. My fear is when we make a change from that component, it will break a lot of things.

jtamiace:
I see. I'm not sure if this makes more sense as a separate issue, but if changes are needed to the component itself, I've been looking at an alternative layout for these tabs in designs for a separate project that might also address this outline problem? 

Screen Shot 2022-12-09 at 12 17 53 PM

marcellamaki:
Hi @AllanOXDi and @jtamiace - 

Jessica, these are interesting updates! I think better addressed in a separate issue, and we can make sure that the focus changes are still prioritized there, though. But, thanks for sharing this as an option. When we [create a KTabGroup component](https://github.com/learningequality/kolibri-design-system/issues/385), then at least on the code side, this side panel would be a place we would want to refactor and use it. (Not saying it has to inform your designs.... just that we will be happy to add these designs as we refactor that - we can do it all in one project perhaps!)

Allan - I have a suggestion for a way to update the `HeaderTab.vue` component that doesn't require making too many changes that might break things. My idea is to swap your changes and instead add: 
`outline-offset: -2px !important;`

to the `.header-tab` class to get:
Screen Shot 2022-12-09 at 4 35 48 PM


This is not something you could have reasoned out without knowing more, so here's some context: 
* Generally, we try to avoid using `!important` where we can, as it's an override, and it might indicate that there is another or better way to do something 
* However, this `focusOutline` and it's "[offset](https://developer.mozilla.org/en-US/docs/Web/CSS/outline-offset)" are [set in KDS in the main theme](https://github.com/learningequality/kolibri-design-system/blob/release-v1.4.x/lib/styles/theme.js#L67) to be `4px` (this is kind of KDS trivia, but good to remember!)
* This means that the focus outline being slightly larger than the element is the default behavior
* In almost all instances, this makes sense! It looks cleaner to have the focus outline going around a button, for example, than on top of it, and prevents potentially covering up or cutting off information, especially on smaller sizes. 
* However, I think you may have found an instance where we need an exception, not the rule! There are a handful of other places where we [use the same fix in other places in Kolibri](https://github.com/learningequality/kolibri/blob/develop/kolibri/core/assets/src/views/CoreMenu/CoreMenuOption.vue#L110) (search for "`outline-offset`" and you'll see it) 
* The most common places we have to adjust is situations like this - where we have a focusable element, such a link or button, that is visually distinct from the KDS defaults, or when they are layered or immediately next to other elements. In this example, the focus outline is _there_, it's just behind the rest of the side panel.

Give it a try and see what you think. The outline is not quite as _pretty_ this way, as it is not entirely in the shape of the nice rounded tab, but I think that's okay.

pcenov:
Fantastic @AllanOXDi - the visible focus is now displayed correctly in all 3 cases.

radinamatic (approved):
Great job, thank you @AllanOXDi!

Manual QA passed, good to go! 🥳 💯 :shipit:

marcellamaki (approved):

marcellamaki:
Nice work @AllanOXDi!

PR #9904, opened 2022-12-08
Fix error with null nav items.

rtibbles:
## Summary
* Fixes a previously unreported error that would occur when being on the library page while not logged in.

## Reviewer guidance
This stops the error, but I am not completely sure it's the correct fix. Verify that this fixes both the non-logged in and the logged in case.

----

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/core/assets/src/views/HorizontalNavBarWithOverflowMenu.vue
+++ b/kolibri/core/assets/src/views/HorizontalNavBarWithOverflowMenu.vue
@@ -1,12 +1,11 @@
 <template>
 
   <Navbar>
-    <div v-for="(link, index) in navigationLinks" :key="index" ref="navLinks">
-      <!-- seeting navbarlink to "not hidden" is a bit counter intutive,  -->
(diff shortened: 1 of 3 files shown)


marcellamaki (commented):
I recently saw this and forgot to open an issue - thanks for fixing it. 

I think the fix makes sense, and it fixes the console error. One thing I noticed is that with the fix, the overflow and bookmarks option pops up, on all screen sizes and even when not logged in.


Screen Shot 2022-12-09 at 11 27 53 AM

I think it's a bit out of scope - probably something else brittle or without the proper checks that I did, that is now popping up due to this check. I can open a follow up issue for that (unless you would like to include it here)

rtibbles:
I'll update this here, yup!

rtibbles:
On mobile, logged in:
image

rtibbles:
On mobile logged out:
image

marcellamaki (approved):
LGTM! Thanks for the refactor

PR #9906, opened 2022-12-10
First pass routing and data fetching for remote content browsing and zipcontent remote capability

### LINKED ISSUE #9179: When there is more content in an exercise's folder resources, there is a console error for a missing prop
## Observed behavior
When in an exercise and if I click on the "View folder resources" icon, if there are resources, there is an error for the `nextContent` prop.

## Errors and logs
Screen Shot 2022-03-14 at 9 27 08 AM

## Expected behavior
No errors in console

## User-facing consequences
Doesn't seem to have any user-specific consequences outside of the console.

## Steps to reproduce
1. Login as a learner
2. Click on an exercise/pdf/etc. that has additional resources (e.g. "pH Scale: Basics")
3. Click on the "view resources" icon (icon furthest left on the right-hand side)
4. Note the prop error

## Context
Kolibri 0.15
MacOS
Chrome

### LINKED ISSUE #9387: Allow /content/zipcontent endpoint to accept a baseurl GET parameter for remotely accessing zipcontent
An API endpoint setup under /content/zipcontent that will proxy to a specific remote server using the get param baseurl specified. If no baseurl is specified, it should return a 404.

The API endpoint will then make a further request to <baseurl>/content/zipcontent/.... and then stream the response from that request as the response to the request being made from the client.

Caching these resources is a little trickier, unless we took the decision to stream the entire zipcontent file and attempt to serve the embedded resources from that.

rtibbles:
## Summary
* Adds the capability to leverage a baseurl parameter on zipcontent endpoint requests
* Adds additional filtering for dynamic loading for our subclass of Whitenoise, to allow WSGI apps that we know are serving on a subpath (such as the zipcontent endpoint) to skip attempts at dynamic loading
* Tweaks the frontend to ensure that the baseurl parameter is preserved in Hashi
* Adds frontend routing and data loading support for remote content browsing
* Adds simple list of local devices with links to a device specific library page at the bottom of the main library page
* Updates search, topic tree browsing, and resource viewing to handle remote device browsing

## References
Fixes #9387
Fixes #9179

## Reviewer guidance
Run two Kolibri servers in different home directories - one only needs to run the backend, and should be in your regular home directory that has HTML5 resources loaded, the other should use the devserver and be in an environment without the HTML5 resource you will be trying to load.

Browse the Library Page and see the list of other Kolibri devices at the bottom. Click on one and browse!

----

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated


nucleogenesis (commented):
The code LGTM in general.

@akolson and I tested this manually using local instances on my machine but we got an error. To be sure it wasn't just that I had messed up somewhere we opted to test together over ZeroTier.

Samson downloaded some HTML5 content on his machine and I ran my devserver with the diff you provided applied to my code and we got the same error.

In `learn/../routes/index.js` ~ line 124 in the `/topics/:channel_id`  route definition, the `get(channelsMap)[channel_id].root` line is trying to read a property from undefined. Seems maybe that the list of remote channels isn't being populated?

To confirm ZeroTier was working properly, I downloaded a channel from his instance successfully.

rtibbles:
> In learn/../routes/index.js ~ line 124 in the /topics/:channel_id route definition, the get(channelsMap)[channel_id].root line is trying to read a property from undefined. Seems maybe that the list of remote channels isn't being populated?

Ah yes, sorry my diff will only work if you have imported a different resource from the same channel on the alternate machine - it doesn't cover remote fetching of channel metadata (although the API backend does support that).

kolibri/plugins/learn/assets/src/views/TopicsContentPage.vue
```diff
@@ -385,37 +389,42 @@
...
-        const nextPromise = ContentNodeResource.fetchNextContent(this.content.parent, {
-          topicOnly: true,
-        }).then(nextContent => {
-          // This may return the immediate parent if nothing else is found so let's be sure
-          // not to assign that
-          if (nextContent && this.content.parent !== nextContent.id) {
-            this.nextContent = nextContent;
-          }
-        });
+        const fetchGrandparent = this.content.ancestors.length > 1;
```
akolson, line 392: It would be good to keep the naming consistent with camel case

kolibri/plugins/learn/assets/src/views/CompletionModal/index.vue
```diff
@@ -347,14 +357,37 @@
         this.$el.querySelector('.close-button').focus();
       },
       loadNextContent() {
-        return ContentNodeResource.fetchNextContent(this.contentNodeId).then(data => {
-          this.nextContentNode = data;
+        const fetchGrandparent = this.contentNode.ancestors.length > 1;
```
akolson, line 360: It would be good to keep a consistent camelCase for the variable name

kolibri/plugins/learn/assets/src/views/LibraryPage/index.vue
```diff
@@ -190,14 +207,15 @@
 
       const { currentCardViewStyle } = useCardViewStyle();
 
+      const { baseurl, fetchDevices } = useDevices();
```
akolson, line 210: It would be good to keep a consistent camelCase for the variable name `baseurl`

kolibri/plugins/learn/assets/src/modules/topicsTree/handlers.js
```diff
@@ -4,17 +4,16 @@ import samePageCheckGenerator from 'kolibri.utils.samePageCheckGenerator';
...
+import { _collectionState, contentState } from '../coreLearn/utils';
 
-const { channelsMap } = useChannels();
+const { channelsMap, fetchChannels } = useChannels();
 const { fetchContentNodeTreeProgress } = useContentNodeProgress();
 
-export function showTopicsContent(store, id) {
-  store.commit('CORE_SET_PAGE_LOADING', true);
-  store.commit('SET_PAGE_NAME', PageNames.TOPICS_CONTENT);
+function _loadTopicsContent(store, id, baseurl) {
```
akolson, line 13: It would be good to keep a consistent camelCase for the variable name `baseurl`

akolson (approved):
Thanks @rtibbles. Code generally looks good. Just a general non-blocking observation is that since code was moved around a bit, some functions became quite large. Refactoring these into smaller functions with single responsibilities eventually will bring with it benefits into the code.

kolibri/plugins/learn/assets/src/views/TopicsContentPage.vue
rtibbles: I guess I was thinking of 'grandparent' as one word :)

kolibri/plugins/learn/assets/src/views/LibraryPage/index.vue
rtibbles: My only reason for using `baseurl` like this is that it matches the get parameter used by the API - which means when we pass the value to an API call we can just do `{ baseurl }` rather than remapping.

rtibbles:
I'll take a look to see what I can do to reduce the size of some of the handler functions.

rtibbles:
Have updated the handler functions that were rather large and decomposed them into smaller functions that are called in sequence.

Also cleaned up some unused utility functions around content node state, and made the device loading only happen for logged in users (as guests do not have permissions to access the API endpoint).

PR #9911, opened 2022-12-13
Update Progress Tracking API and frontend composable to send complete context via API

### LINKED ISSUE #9892: Update progress tracking endpoint to remove dependence on backend read content metadata
## Observed behavior
https://github.com/learningequality/kolibri/pull/8525 consolidated all progress tracking into a single endpoint and reduced the amount of data that needed to be sent to the backend to establish context for a session by reading additional required metadata from the backend.

Unfortunately, the need for being able to log progress for non-locally available resources means that this mechanism is no longer suitable.

## Expected behavior
All contextual metadata will now need to be sent when the session starts and not read from the database.

Both the API endpoint and the companion `useProgressTracking` composable will have to be updated with this new interface.

rtibbles:
## Summary
* Updates the progress tracking API to take all context data from the frontend API call rather than relying on reading from the database
* Does this to allow progress tracking for resources that are not locally present (i.e. for remote browsing)
* Updates the frontend composable that communicates with the backend to handle this
* Updates all uses of the frontend composable that pass nodeId to instead pass a whole node object to give all the metadata

## References
Fixes #9892

## Reviewer guidance
Confirm that progress tracking still works as intended.

----

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/core/logger/api.py
+++ b/kolibri/core/logger/api.py
@@ -86,29 +86,31 @@ class StartSessionSerializer(serializers.Serializer):
         if "node_id" not in data and "quiz_id" not in data:
             raise ValidationError("node_id is required if not a coach assigned quiz")
         if "node_id" in data:
-            errors = []
+            errors = {}
             if "kind" not in data:
-                errors.append(ValidationError("kind is required for any node_id"))
+                errors["kind"] = ValidationError("kind is required for any node_id")
             else:
                 if (
                     data["kind"] == content_kinds.EXERCISE
                     and "mastery_model" not in data
                 ):
-                    errors.append(
-                        ValidationError(
-                            "mastery model must be specified for exercise kinds"
-                        )
+                    errors["mastery_model"] = ValidationError(
+                        "mastery model must be specified for exercise kinds"
                     )
                 elif data["kind"] != content_kinds.EXERCISE and "mastery_model" in data:
-                    errors.append(
-                        ValidationError(
-                            "mastery model must not be specified for non-exercise kinds"
-                        )
+                    errors["mastery_model"] = ValidationError(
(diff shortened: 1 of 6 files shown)


kolibri/core/logger/api.py
```diff
@@ -80,6 +85,30 @@ def validate(self, data):
...
+                elif data["kind"] != content_kinds.EXERCISE and "mastery_model" in data:
+                    errors.append(
+                        ValidationError(
+                            "mastery model must not be specified for non-exercise kinds"
+                        )
+                    )
+            if "content_id" not in data:
+                errors.append(ValidationError("content_id is required for any node_id"))
+            if "channel_id" not in data:
+                errors.append(ValidationError("channel_id is required for any node_id"))
```
nucleogenesis: How is this `errors` list being used outside of this code? This looks like a local variable that isn't returned or used anywhere.

nucleogenesis (commented):
Looks great! Simplifies things on the backend without adding much if any complexity to the frontend to compensate. 

I just have one note that is blocking if I'm not missing something there -- if I am then things look good to me on code review.

I will come back to do some manual QA and play with it a bit tomorrow

kolibri/core/logger/api.py
rtibbles: It is not... apparently I forgot to reraise the errors list if non-empty.

Also means that test coverage is not sufficient to catch this, so I'll add some regression tests to ensure this validation happens.

rtibbles:
Have updated the code and added regression tests to make sure we are properly throwing the validation errors!

akolson (approved):
LGTM! Thanks @rtibbles

PR #9910, opened 2022-12-13
Fixes back button behaviour for content pages in Learn

rtibbles:
## Summary
* Properly passes the current route query parameters to link generation to preserve search parameters
* Directly references the id param on the route to infer the currently topicId
* Removes unused `genContentLink` props
* Update code comments
* Delete unused component and references to it
* Fixes unreported bug whereby navigating directly to a search page within a topic would run an unscoped search due to attempting to load the search data before the topic metadata has been loaded
* Refactors `genContentLink` utility into a more general purpose composable that can do one of two things:
  * Return a link that includes query parameters that will create a back link to return to the page the user is currently on
  * Return a link that includes query parameters that will create a back link to return to the page that the current page is pointing at

## References
Fixes #9622
Fixes #9342

## Reviewer guidance
* Use search and navigate to a resource - confirm that the back button returns you to search
* Navigate to a resource via the topic tree - confirm that the back button returns you to the same place in the topic tree
* Navigate to a resource via searching within a channel - confirm that the back button returns you to the same place, and to the same listing of results

[Screencast from 12-12-2022 04:21:26 PM.webm](https://user-images.githubusercontent.com/1680573/207195616-f07aebd3-c5ee-437e-a852-fa1bce7d0b6d.webm)

All back behaviours in various permutations for lesson navigation:
[lesson_back.webm](https://user-images.githubusercontent.com/1680573/207480540-03e9fa9f-b777-4894-b35d-0cb1de3dec2f.webm)

All back behaviours in various permutations for library navigation:
[library_back.webm](https://user-images.githubusercontent.com/1680573/207480567-74007460-3399-4f0e-aa9f-0f8dadb1db39.webm)

----

## Testing checklist

- [x] Contributor has fully tested the PR manually
- [x] If there are any front-end changes, before/after screenshots are included
- [x] Critical user journeys are covered by Gherkin stories
- [ ] Critical and brittle code paths are covered by unit tests

## PR process

- [x] PR has the correct target branch and milestone
- [x] PR has 'needs review' or 'work-in-progress' label
- [x] If PR is ready for review, a reviewer has been added. (Don't use 'Assignees')
- [ ] If this is an important user-facing change, PR or related issue has a 'changelog' label
- [ ] If this includes an internal dependency change, a link to the diff is provided

## Reviewer checklist

- Automated test coverage is satisfactory
- PR is fully functional
[...]

Issue #9043, opened 2022-01-26
Missing visual separation between the last sub-folder of a topic and resources that are direct children of the topic


MisRob:
## Observed behavior

This is the structure of the HTML5 topic in the Kolibri QA Channel. Notice "Slides Slides Slides!" and "Document with links" resources listed below all sub-folders of the topic.

![structure](https://user-images.githubusercontent.com/13509191/151130263-649433e1-4331-4520-83c2-8f669aa01a39.png)

These two resources are direct children of the HTML5 topic. However, in the library, they seem to belong to the SCORM sub-folder due to missing visual separation:

![no-gap](https://user-images.githubusercontent.com/13509191/151130154-5500027f-d938-4999-b80f-932247203d91.png)

## Expected behavior

Resources that live next to sub-folders of a topic are visually separated from sub-folders

## User-facing consequences

Resources seem to belong to a folder that they don't belong to

## Steps to reproduce

1. Import the Kolibri QA Channel by using `nakav-mafak` token
2. Navigate to "Learn" -> "Library" -> "Kolibri QA Channel"
3. Go to "HTML5" sub-folder in the "Folders" tab in the side panel
4. Scroll to the very bottom of the page

## Context

- Kolibri version: Kolibri 0.15.1.dev0+git.20220125194959
- Operating system: Ubuntu 20.04.3 LTS
- Browser: Chrome 97.0.4692.99 (Official Build) (64-bit)

MisRob:
@jtamiace Do you have any suggestions on how to solve this?

jtamiace:
I think another related problem is that the order of folders and resources set on Studio is not preserved when they are both in the same folder level. My suggestion is to use dividers like in 1a here, maybe 24px of space above and below, but can experiment to see what feels best. Or, to simply increase the amount of space used as in 1b, but I don't think that will be clear enough.

If there's any argument to not preserve the order of folders/resources set on Studio, we can have a separate "In this folder" label that isn't clickable, and aggregates any resources at that folder level into one place. This might be fine if the folder is just a collection of things, but could be confusing for a course-like structure if an "intro" video and "conclusion" document/video are both in the same place.

![folders](https://user-images.githubusercontent.com/6668144/151236187-b894ff22-86db-4d94-9e61-e840f8cad7aa.jpg)

MisRob:
Thank you, @jtamiace, I appreciate dividers (1a). Without them, it's not very clear to me even with larger spaces.

[...]

PR #9915, opened 2022-12-15
Implement folder skip logic to avoid navigation down single child folders

### LINKED ISSUE #9860: Remote Content Browsing: Update Channel and Library UI to use new shortened file path criteria
Implementing the skip logic described here:
Frame 873

In addition, if the top level topic has only a single child, then it should instead display the first descendant with more than one folder child or a child that is a resource.

rtibbles:
## Summary
* Updates the backend topic tree endpoint to allow recursive filling out of tree data in the case where there is only a single descendant
* Updates tests
* Updates the frontend to skip topics retrieved from the backend when they only have a single topic child and recurse through the descendants until it finds either multiple topic children or a resource

## References
Fixes #9860

## Reviewer guidance
Import a resource in a channel with some nested paths, with only 1 available resource in a deeply nested folder.
Import multiple resources in the deeply nested path.
Remove one resource, and import another in a parallel nested path.

Breadcrumb link is no longer disabled when showing '3rd grade' but skipping to the child folder, but rather has a `noSkip` query parameter to allow navigation to avoid the skipping:
Screenshot from 2022-12-14 18-17-59

Collapsed child folders within a parent:
Screenshot from 2022-12-14 18-18-12

----

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/core/content/api.py
+++ b/kolibri/core/content/api.py
@@ -940,27 +940,46 @@ class TreeQueryMixin(object):
 
         return depth, next__gt
 
+    def _get_gc_by_parent(self, child_ids):
+        # Use this to keep track of how many grand children we have accumulated per child of the parent node
+        gc_by_parent = {}
+        # Iterate through the grand children of the parent node in lft order so we follow the tree traversal order
+        for gc in (
+            self.filter_queryset(self.get_queryset())
+            .filter(parent_id__in=child_ids)
+            .values("id", "parent_id")
+            .order_by("lft")
+        ):
+            # If we have not already added a list of nodes to the gc_by_parent map, initialize it here
+            if gc["parent_id"] not in gc_by_parent:
+                gc_by_parent[gc["parent_id"]] = []
+            gc_by_parent[gc["parent_id"]].append(gc["id"])
+        return gc_by_parent
+
     def get_grandchild_ids(self, child_ids, depth, page_size):
+        grandchild_ids = []
         if depth == 2:
             # Use this to keep track of how many grand children we have accumulated per child of the parent node
-            gc_by_parent = {}
-            # Iterate through the grand children of the parent node in lft order so we follow the tree traversal order
-            for gc in (
-                self.filter_queryset(self.get_queryset())
-                .filter(parent_id__in=child_ids)
-                .values("id", "parent_id")
-                .order_by("lft")
-            ):
-                # If we have not already added a list of nodes to the gc_by_parent map, initialize it here
-                if gc["parent_id"] not in gc_by_parent:
-                    gc_by_parent[gc["parent_id"]] = []
-                # If the number of grand children for a specific child node is less than the page size
-                # then we keep on adding them to both lists
-                # If not, we just skip this node, as we have already hit the page limit for the node that is
-                # its parent.
-                if len(gc_by_parent[gc["parent_id"]]) < page_size:
-                    gc_by_parent[gc["parent_id"]].append(gc["id"])
-                    yield gc["id"]
+            gc_by_parent = self._get_gc_by_parent(child_ids)
+            singletons = []
+            for child_id in gc_by_parent:
+                gc_ids = gc_by_parent[child_id]
+                if len(gc_ids) == 1:
+                    singletons.append(gc_ids[0])
+                # Only add up to the page size to the list
+                grandchild_ids.extend(gc_ids[:page_size])
+            if singletons:
+                grandchild_ids.extend(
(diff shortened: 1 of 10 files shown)


kolibri/core/content/api.py
```diff
@@ -940,27 +940,46 @@ def validate_and_return_params(self, request):
...
-                # If the number of grand children for a specific child node is less than the page size
-                # then we keep on adding them to both lists
-                # If not, we just skip this node, as we have already hit the page limit for the node that is
-                # its parent.
-                if len(gc_by_parent[gc["parent_id"]]) < page_size:
-                    gc_by_parent[gc["parent_id"]].append(gc["id"])
-                    yield gc["id"]
+            gc_by_parent = self._get_gc_by_parent(child_ids)
+            singletons = []
+            for child_id in gc_by_parent:
```
nucleogenesis, line 969: Should this be `parent_id` in place of `child_id`?

kolibri/plugins/learn/assets/src/views/TopicsPage/index.vue
```diff
@@ -453,16 +459,29 @@
         return this.resourcesDisplayed.length < this.resources.length;
       },
       topics() {
-        return this.contents.filter(content => content.kind === ContentNodeKinds.TOPIC);
+        return this.contents
+          .filter(content => content.kind === ContentNodeKinds.TOPIC)
+          .filter(t => t.children && t.children.results.length)
+          .map(t => {
+            let topicChildren = t.children ? t.children.results : [];
```
nucleogenesis: Looks like the ternary check here will never evaluate to anything except `t.children.results` which should always be something that has a `.length`  because of the filter so maybe could remove the ternary?

nucleogenesis (approved):
Just a couple of notes mostly style related. All-in-all code looks good.

marcellamaki (approved):
Talked through adjustments with Richard, and this looks good. 

It looks like we don't have gherkins for this @rtibbles (cc @jtamiace @radinamatic). I'm not sure the best way to codify the behavior here - it's not clear to me if it should have it's own "scenario" or be used in descriptions of other "scenarios". Regardless, it seems like it would be good to have written down somewhere! Should we open a follow up issue for this?

radinamatic:
I added the label, should be enough!
cc @pcenov

pcenov:
@rtibbles seems that there are some issues with this implementation. I have created the following channel so that it can be easily imported in both Kolibri 0.15 and develop and compared side by side: sukiz-papar

1. I was expecting to see the folders which are shown in the FOLDERS tab in an unchanged fashion exactly as it was before however now it's a bit confusing as the folder skip logic is applied there as well:

https://user-images.githubusercontent.com/79847249/212126386-df9f1796-1db5-4785-9b4b-fe56b00fb4c6.mp4

2. The implementation of Example 2 of the picture provided in https://github.com/learningequality/kolibri/issues/9860 can also be considered confusing as while I am in folder 2 it shows the resources from folders 3 and 4 side by side:



3. Clicking the Back button while viewing a nested resource brings the user back to the Library page instead of the previous screen:

https://user-images.githubusercontent.com/79847249/212124558-ef59058a-bf8a-46d5-aac4-b1e5680ee65c.mp4

rtibbles:
> I was expecting to see the folders which are shown in the FOLDERS tab in an unchanged fashion exactly as it was before however now it's a bit confusing as the folder skip logic is applied there as well:

So we have two choices, either we show the nested folders like we do for the headers, or we show the top level folders (which would give a bit of a mismatch to the headers, and invite navigation into not very useful folders).

> The implementation of Example 2 of the picture provided in https://github.com/learningequality/kolibri/issues/9860 can also be considered confusing as while I am in folder 2 it shows the resources from folders 3 and 4 side by side:

I'll take a closer look at this.

> Clicking the Back button while viewing a nested resource brings the user back to the Library page instead of the previous screen:

I think this is an issue caused by an error in #9910, which I have fixed in #9906

jtamiace:
For the first issue, the experiences in tension are consistency in label names in the side panel/breadcrumb and quicker navigation. Giving this some thought again, seems it'd be best to change the left folder side panel so that the options there align with the breadcrumb - so, changing it to be as it was before. While it will make navigation "slower" if the user chooses to use that panel, it seems it will create more consistency in expected behavior and there is still always the option for faster navigation through the main card section.

On the second issue, there's a solution here to use a divider, extra space, and placing resources in the order as set on Studio, highlighted in "1a" in the mockup https://github.com/learningequality/kolibri/issues/9043#issuecomment-1022561028

PR #9917, opened 2022-12-16
Remove reference to no longer existing and helper.

rtibbles:
## Summary
 *Upgrading @vueuse/core meant that the `and` helper we were referencing was migrated to `@vueuse/math` and renamed as `logicAnd`
* Rather than install a whole other package for this one helper, this replaces the single instance of it


marcellamaki (approved):
lgtm - worked after `rm -rf node_modules`!

PR #9926, opened 2022-12-16
Tweak side panel scroll position setting behaviour.

rtibbles:
## Summary
* Creates separate computed props to generate throttled methods for each throttled method
* Do existence checks and reference $el to ensure we can use DOM element methods.

### DIFF
--- a/kolibri/plugins/learn/assets/src/views/TopicsPage/index.vue
+++ b/kolibri/plugins/learn/assets/src/views/TopicsPage/index.vue
@@ -533,9 +533,11 @@
           return 4;
         } else return null;
       },
+      throttledStickyCalculation() {
+        return throttle(this.stickyCalculation);
+      },
       // calls handleScroll no more than every 17ms
-      throttledHandleScroll() {
-        throttle(this.stickyCalculation);
+      throttledtabPositionCalculation() {
         return throttle(this.tabPositionCalculation);
       },
       activeActivityButtons() {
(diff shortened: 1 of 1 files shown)


marcellamaki (approved):
lgtm

kolibri/plugins/learn/assets/src/views/TopicsPage/index.vue
```diff
@@ -542,9 +541,11 @@
           return 4;
         } else return null;
       },
+      throttledStickyCalculation() {
+        return throttle(this.stickyCalculation);
+      },
       // calls handleScroll no more than every 17ms
-      throttledHandleScroll() {
-        throttle(this.stickyCalculation);
+      throttledtabPositionCalculation() {
```
akolson: Non blocking, but would be good to keep this as fully camel case for better readability. It would turn to `throttledTabPositionCalculation`

akolson (approved):
Changes generally LGTM!

kolibri/plugins/learn/assets/src/views/TopicsPage/index.vue
rtibbles:
```suggestion
      throttledTabPositionCalculation() {
```

kolibri/plugins/learn/assets/src/views/TopicsPage/index.vue
```diff
@@ -615,6 +616,10 @@
     methods: {
       ...mapActions('topicsTree', ['loadMoreContents', 'loadMoreTopics']),
       genContentLink,
+      throttledHandleScroll() {
+        this.throttledStickyCalculation();
+        this.throttledtabPositionCalculation();
```
rtibbles:
```suggestion
        this.throttledTabPositionCalculation();
```

PR #9940, opened 2022-12-22
theme links in navbar fixed

### LINKED ISSUE #9019: Theme links in top navigation bars
Due to technical limitations, we couldn't utilize theme colors fully for links in top navigation bars in the past, e.g. Home, Library, and Bookmarks links in Learn:

links

When a link is active, we use hardcoded `white` colors

https://github.com/learningequality/kolibri/blob/9048d77029e622b57d95009ab8b968f12dd5d99a/kolibri/core/assets/src/views/Navbar/NavbarLink.vue#L111-L112

where we should use [`textInverted` token](https://design-system.learningequality.org/colors/#tokens-textInverted).

It should be now possible to refactor `NavbarLink` to use [router-link scoped slot API ](https://v3.router.vuejs.org/api/#v-slot-api-3-1-0) that's been released meanwhile. That should allow us to use `textInverted` token instead of hardcoded white.

## User-facing consequences

`textInverted` token is used to create sufficient contrast when used on dark backgrounds such as `tokens.primary`. Using hardcoded color instead of the token increases the risk of not having the proper contrast in theme plugins.

amitpanwar789:
## Summary

replaces hardcoded colors with theme tokens 

Noticeable UI changes

before changes:

old

After changes:
new

## References

resolve #9019 

## Reviewer guidance

@marcellamaki 

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/core/assets/src/views/Navbar/NavbarLink.vue
+++ b/kolibri/core/assets/src/views/Navbar/NavbarLink.vue
@@ -71,7 +71,7 @@
       },
       activeClasses() {
         // return both fixed and dynamic classes
-        return `router-link-active ${this.$computedClass({ color: this.$themeTokens.primary })}`;
+        return `router-link-active ${this.$computedClass({ color: this.$themeTokens.textInverted })}`;
       },
     },
   };
--- a/kolibri/core/assets/src/views/Navbar/NavbarLink.vue
+++ b/kolibri/core/assets/src/views/Navbar/NavbarLink.vue
@@ -71,7 +71,9 @@
       },
       activeClasses() {
         // return both fixed and dynamic classes
-        return `router-link-active ${this.$computedClass({ color: this.$themeTokens.textInverted })}`;
+        return `router-link-active ${this.$computedClass({
+          color: this.$themeTokens.textInverted,
+        })}`;
       },
     },
   };
--- a/kolibri/core/assets/src/views/Navbar/NavbarLink.vue
+++ b/kolibri/core/assets/src/views/Navbar/NavbarLink.vue
@@ -2,6 +2,7 @@
 
   <li class="list-item-navigation visible">
     <router-link
+      :active="isActive"
       class="tab"
       :class="$computedClass(tabStyles)"
       :style="windowIsSmall ? smallScreenOverrides : {}"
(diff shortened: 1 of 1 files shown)


kolibri/core/assets/src/views/Navbar/NavbarLink.vue
```diff
@@ -68,6 +70,9 @@
           borderBottomWidth: '2px',
         };
       },
+      isActive() { 
+        return this.$route.path === `/${this.link.name.toLowerCase()}`;
```
MisRob: We can't rely on this comparison to determine whether a link is active.

`link.name` contains our internal route name which in some cases may indeed correspond to the actual URL, but that happens rather by chance.

One place where you can see such solution fails is in _"Coach" --> "Class home"_ tab. Text of the active _"Class home"_ tab is `rgb(250, 250, 250)` (`this.$themePalette.grey.v_50`) instead of expected `rgb(255, 255, 255)` (`this.$themeTokens.textInverted`). Even though the tab is active, `isActive` is evaluated to `false` due to `link.name` being `HomePage`whereas the actual URL contains only `home`. In some cases, it is even slightly more complicated as `$route.path` may contain some more information, e.g. a class ID.

Looking into `$route.name` won't help us either, as again in some cases, like when sub-routing, it is different from `link.name`. You can, for example, log both values in _"Coach" -> "Plan" -> "Quizzes"_ to see it.

MisRob (changes requested):
@amitpanwar789 Thank you for your time and work, we appreciate it. Unfortunately, this solution is based on a problematic comparison, please see the explanation. If you'd like to still work on this issue, I think we'd need a different approach. There are some suggestions in the issue description that may be worth exploring or you could also try different ideas if you have some. Let me know if you have any questions.

MisRob:
@amitpanwar789 I've just discovered a place in our codebase that utilizes Vue Router's active class with our `$computedClass` ([docs](https://kolibri-dev.readthedocs.io/en/develop/frontend_architecture/core.html?highlight=computed%20class#dynamic-core-theme)) for a similar purpose. Perhaps this could help us here?

https://github.com/learningequality/kolibri/blob/d7eaed56c7480a2adb1833664667b610b6765be3/kolibri/plugins/coach/assets/src/views/common/HeaderTabs/HeaderTab.vue#L7

https://github.com/learningequality/kolibri/blob/d7eaed56c7480a2adb1833664667b610b6765be3/kolibri/plugins/coach/assets/src/views/common/HeaderTabs/HeaderTab.vue#L34-L37

amitpanwar789:
@MisRob done, thank you for the help

MisRob (approved):
Thank you, @amitpanwar789, code is looking good to me, and can see the expected colors on links now. If you'd like, you're welcome to add yourself to the list of contributors https://github.com/learningequality/kolibri/blob/develop/AUTHORS.md before we merge this PR

PR #9961, opened 2023-01-03
Facility export usage data section design updated

### LINKED ISSUE #9704: Implement new design for "Export usage data"
## Observed behavior

The existing design:
Screenshot from 2022-09-09 10-04-03

## Expected behavior

In the new design:
- The notes should be relocated to appear in modals on click of a "Learn more" link
- The new design should be responsive to varying viewport sizes down to mobile
  - the <kbd>Download</kbd> and <kbd>Generate log</kbd> buttons should wrap below when the viewport size is too small for them to appear on the same row

Screenshot from 2022-09-09 10-04-34

## User-facing consequences

- The new design is a more fluid layout that is easier to make responsive and will better support all screen sizes overall

LianaHarris360:
## Summary

This PR updates the 'Export usage data' section under the Facility > Data tab. The changes include the notes being relocated to appear in modals when the "Learn More" link is clicked and a responsive design for varying viewport sizes down to mobile. The `Download` and `Generate` log buttons now wrap below when the viewport size is too small for them to appear on the same row.

Before
image

After
ExportUsageDataAfter

ExportUsageDataAfterResponsive

ExportUsageDataAfterResponsive2

## References

[Figma mockup](https://www.figma.com/file/SgTZugaKkES68aipU5dCk8SR/Github-issues-%26-random-fixes?node-id=1055%3A2073&t=LWDWhNNeD7A6srJo-0)
Closes #9704

## Reviewer guidance

1. Navigate to Facility > Data tab.
2. Confirm `Generate` and `Download` buttons display the expected behavior.
3. Confirm `Learn more` links open modals that contain the correct information.
4. Adjust the screen viewport to confirm that the responsive design responds as expected.

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/plugins/facility/assets/src/views/DataPage/LearnMoreModal.vue
+++ b/kolibri/plugins/facility/assets/src/views/DataPage/LearnMoreModal.vue
@@ -1,0 +1,80 @@
+<template>
+
+  <KModal
+    :title="logType === 'summary' ? $tr('summaryLogs') : $tr('sessionLogs')"
+    :submitText="$tr('close')"
+    size="medium"
+    @cancel="closeModal"
+    @submit="closeModal"
+  >
+    <p class="description">
+      {{ getMessage() }}
+    </p>
+  </KModal>
+
+</template>
+
+
+<script>
+
+  export default {
+    name: 'LearnMoreModal',
+    props: {
+      logType: {
+        type: String,
+        required: true,
+      },
+    },
+    methods: {
+      closeModal() {
+        this.$emit('cancel');
+      },
+      getMessage() {
+        let message = '';
+        if (this.logType === 'summary') {
+          message = this.$tr('summaryLogText');
+        } else if (this.logType === 'session') {
+          message = this.$tr('sessionLogText');
+        }
+        return message;
+      },
+    },
+    $trs: {
+      summaryLogs: {
+        message: 'Summary Logs',
+        context: 'Title for Summary logs description',
+      },
+      sessionLogs: {
+        message: 'Session Logs',
+        context: 'Title for Session logs description',
+      },
+      summaryLogText: {
+        message:
+          'A user may visit the same resource multiple times. This file records the total time and progress each user has achieved for each resource, summarized across possibly more than one visit. Anonymous usage is not included.',
+        context:
+          "Text description on the 'Learn More' pop-up window in the Facility > Data > Export usage data section.\n",
+      },
+      sessionLogText: {
+        message:
+          'When a user views a resource, we record how long they spend and the progress they make. Each row in this file records a single visit a user made to a specific resource. This includes anonymous usage, when no user is signed in.',
+        context:
+          "Text description on the 'Learn More' pop-up window in the Facility > Data > Export usage data section.\n",
+      },
+      close: {
+        message: 'Close',
+        context:
+          "Refers to the 'Close' button on the 'Learn More' pop-up window in the Facility > Data > Export usage data section.",
+      },
+    },
+  };
+
+</script>
+
+
+<style lang="scss" scoped>
+
+  .description {
+    margin-top: 0;
+  }
+
+</style>
--- a/kolibri/plugins/facility/assets/src/views/DataPage/LearnMoreModal.vue
+++ b/kolibri/plugins/facility/assets/src/views/DataPage/LearnMoreModal.vue
@@ -1,14 +1,14 @@
 <template>
 
   <KModal
-    :title="logType === 'summary' ? $tr('summaryLogs') : $tr('sessionLogs')"
+    :title="modalTitle"
     :submitText="$tr('close')"
     size="medium"
     @cancel="closeModal"
     @submit="closeModal"
   >
     <p class="description">
-      {{ getMessage() }}
+      {{ modalMessage }}
     </p>
   </KModal>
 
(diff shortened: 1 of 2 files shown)


pcenov:
Hi @LianaHarris360, the first issue which I'm observing is that when I am viewing the page in full screen with plenty of space the 'Download' button always remains above the 'Generate log file' button while I am expecting to see it to the left in that case as specified in Figma:

https://user-images.githubusercontent.com/79847249/210816333-c681629c-95c0-47ed-a28c-39b262fdf7d5.mp4

Also in Figma the button's label is 'Generate log' while in the current build it is 'Generate log file' - which one is correct? :)

Lastly when you open the modal for both Session and Summary logs the 'Logs' part of the title is capitalized, while it shouldn't.

LianaHarris360:
Hi, @pcenov Thank you for catching this, I will update this PR to fix these issues!

kolibri/plugins/facility/assets/src/views/DataPage/LearnMoreModal.vue
```diff
@@ -0,0 +1,80 @@
...
+      },
+      getMessage() {
+        let message = '';
+        if (this.logType === 'summary') {
+          message = this.$tr('summaryLogText');
+        } else if (this.logType === 'session') {
+          message = this.$tr('sessionLogText');
+        }
+        return message;
+      },
```
bjester: Very minor, since this will only be invoked once, but this looks like it could be a computed property.

kolibri/plugins/facility/assets/src/views/DataPage/LearnMoreModal.vue
```diff
@@ -0,0 +1,80 @@
...
+
+<script>
+
+  export default {
+    name: 'LearnMoreModal',
+    props: {
+      logType: {
+        type: String,
+        required: true,
+      },
```
bjester, line 37: Since your message parsing wouldn't fail but silently handle log types that aren't `summary` nor `session`, you could add a validator to ensure that this `logType` prop is one of those.

kolibri/plugins/facility/assets/src/views/DataPage/index.vue
```diff
@@ -284,12 +290,33 @@
...
+  }
+
+  .subheading-buttons {
+    display: inline-block;
+    margin-right: 0.313rem;
+  }
+
+  .generated-time {
+    margin: 0.313rem 0;
+    color: #616161;
```
bjester: Kolibri can be themed with different colors, so using tokenized colors is recommended for this. This color seems to match the `$themeTokens.annotation` color. See https://design-system.learningequality.org/colors/#tokens-annotation

Note that you'll need to use `$themeTokens.annotation` in the component template or JS, not within the CSS

bjester (changes requested):
Code review looks good, but left some comments for things that I think should change

LianaHarris360:
Hi @pcenov `Generate Log` is the correct string, it should be corrected now.

This is the smallest screen size where the `Download` button remains beside the `Generate Log` button

Screenshot 2023-01-06 at 2 10 05 PM

Now, on smaller screen sizes, the `Generate Log` button should be in front of the `Download` button

Screenshot 2023-01-06 at 1 32 58 PM


The modal title for both Session and Summary logs should be displayed correctly now as well.

Screenshot 2023-01-09 at 9 02 11 AM

Screenshot 2023-01-09 at 9 02 03 AM

kolibri/plugins/facility/assets/src/views/DataPage/index.vue
```diff
@@ -168,6 +176,14 @@
       pollForTasks() {
         return this.$route.name === PageNames.DATA_EXPORT_PAGE;
       },
+      windowSizeStyle() {
+        if (this.windowIsMedium || this.windowIsSmall) {
+          return {
+            float: 'left',
```
bjester: Floats can be quite buggy across browsers. Based off what I see on [Can I Use](https://caniuse.com/?search=order), you could use `order` which I think is a bit more accurate to the intent here.

With either `float` or `order`, I think there's an important note regarding accessibility in the [MDN `order` docs](https://developer.mozilla.org/en-US/docs/Web/CSS/order):
> Since order is only meant to affect the visual order of elements and not their logical or tab order...

> Using the order property will create a disconnect between the visual presentation of content and DOM order. This will adversely affect users experiencing low vision navigating with the aid of assistive technology such as a screen reader. If the visual (css) order is important, then screen reader users will not have access to the correct reading order.

I don't think the DOM order is all that important, but specifically the keyboard navigable ordering seems important. The [linked article](https://tink.uk/flexbox-the-keyboard-navigation-disconnect/) has some recommendations.

LianaHarris360: This looks interesting and you make a good point about the keyboard navigation, I will try implementing `order` instead and look into the recommendations listed in the article.

pcenov:
@LianaHarris360 I confirm the latest changes are implemented correctly as specified above. The only additional issue which I noticed is that the buttons are not aligned to the text as they are in Figma:



Figma:

kolibri/plugins/facility/assets/src/views/DataPage/index.vue
```diff
@@ -284,12 +316,45 @@
...
+    padding-top: 20px;
+    padding-bottom: 15px;
+  }
+
+  .summary-section-buttons {
+    padding-top: 20px;
+  }
+
+  /deep/ .section-buttons-flex div {
+    display: flex;
```
bjester, line 358: Is this rule required?

bjester (approved):
Looks good to me. One comment on some CSS I'm not sure that's needed

kolibri/plugins/facility/assets/src/views/DataPage/index.vue
LianaHarris360: If the `display: flex` is taken away, the `Generate` button is not placed in front of the `Download` button on smaller screen sizes. The `section-buttons-flex` class is only applied when `windowIsMedium` or `windowIsSmall` is true.

https://user-images.githubusercontent.com/46411498/211926665-a451430b-1fcb-4275-b09d-0f1019c61d97.mov

bjester: Ah gotcha! Sorry my usage search of it didn't bring up the definition of it in the computed prop. :+1:

LianaHarris360: Do you think I should add a comment stating that this class is conditional based on the computed prop `windowSizeStyle` and that it supports the `order` property?

bjester: Yeah a comment is never terrible :smile:

pcenov:
Looks perfect now - manual QA passed.

PR #9767, opened 2022-10-11
Anroid bottom nav bar

marcellamaki:
## Summary

@radinamatic and @pcenov - thank you for taking a first look through these changes! I have a few last hiccups with coach routing that I still have not sorted out, but I think a first pass would be very helpful. I promise I will have proper gherkins written up when the PR is ready for final review, and I've done my best to outline testing scenarios clearly below for what the _currently expected_ functionality should be based on what I've done so far. There will also be some slight visual/UI updates to make this to-spec to the figma that I'll brush up early next week.

#### Regression testing
- These changes should ONLY affect the "app context", probably best tested on a physical device, as I have been mocking the changes for development 
- Therefore, testing of the "usual" browser-based workflow should be done to ensure that there have been no regressions
- Key areas of focus would be all things navigation - the side navigation panel, and the routing for the tabs within each plugin. All should function normally and have no visual changes

#### New feature testing
While the new feature is not 100% complete yet, there are some key pieces of functionality that are ready for testing

**Bottom bar for the learner:**
- In the App Context, the bottom bar for learner should be functional
- bottom bar should only appear in the app context 
- learner/user should be able to navigate with any of the bottom icon buttons
-  learner/user should be able to navigate with any of the bottom icon buttons using TAB key, and focus outline should be present
- the MENU icon, should open an overlay that contains learner information (very similar to the current browser sidebar) 
- within this overlay menu, the "LEARN" button should open and close, to reveal that same navigation tab options (home, library, bookmarks)
- these options should work for navigation
- all other menu options should also work
- the menu should be keyboard navigable, with the focus outline visible, and the focus trapped (contained within the menu while open, not disappearing across the page)

| Descption  | Screenshot |
|---|---|
| Bottom bar is present  | Screenshot 2023-01-06 at 6 59 38 PM |
|  Overlay menu is visible | Screenshot 2023-01-06 at 6 44 08 PM |
| Overlay menu has an option open | Screenshot 2023-01-06 at 6 49 35 PM |

**Bottom bar for the non-learner**:
- In the App Context, the bottom bar for learner should be functional
- bottom bar should only appear in the app context 
- For all functioning navigation routes, no matter what plugin they are currently in (For example, if currently on the device page, "Home" on the bottom bar should bring to the LEARN home page, "Library" to the Library page, etc. If in the learn plugin an ")
-  user should be able to navigate with any of the bottom icon buttons using TAB key, and focus outline should be present
[...]

PR #9984, opened 2023-01-10
App bottom bar

marcellamaki:
## Summary

@radinamatic and @pcenov - thank you for taking a first look through these changes! I have a few last hiccups with coach routing that I still have not sorted out, but I think a first pass would be very helpful. I promise I will have proper gherkins written up when the PR is ready for final review, and I've done my best to outline testing scenarios clearly below for what the _currently expected_ functionality should be based on what I've done so far. There will also be some slight visual/UI updates to make this to-spec to the figma that I'll brush up early next week.

#### Regression testing
- These changes should ONLY affect the "app context", probably best tested on a physical device, as I have been mocking the changes for development 
- Therefore, testing of the "usual" browser-based workflow should be done to ensure that there have been no regressions
- Key areas of focus would be all things navigation - the side navigation panel, and the routing for the tabs within each plugin. All should function normally and have no visual changes
- Please note that there is a KNOWN REGRESSION with the search function on library and topics page in learn. See image below. Richard is making changes in parallel that are changing this, and I will integrate those changes once he has done so. 

Screenshot 2023-01-24 at 6 56 36 PM

#### New feature testing
While the new feature is not 100% complete yet, there are some key pieces of functionality that are ready for testing

**Bottom bar for the learner:**
- In the App Context, the bottom bar for learner should be functional
- bottom bar should only appear in the app context 
- learner/user should be able to navigate with any of the bottom icon buttons
-  learner/user should be able to navigate with any of the bottom icon buttons using TAB key, and focus outline should be present
- the MENU icon, should open an overlay that contains learner information (very similar to the current browser sidebar) 
- within this overlay menu, the "LEARN" button should open and close, to reveal that same navigation tab options (home, library, bookmarks)
- these options should work for navigation
- all other menu options should also work (The home and bookmarks icons should only appear while a user is logged in. For just the guest user, it should only be the LIBRARY icon and main menu icon.)
- the menu should be keyboard navigable, with the focus outline visible, and the focus trapped (contained within the menu while open, not disappearing across the page)
- ESC key should close the menu

| Descption  | Screenshot |
|---|---|
| Bottom bar is present  | Screenshot 2023-01-06 at 6 59 38 PM |
|  Overlay menu is visible | Screenshot 2023-01-06 at 6 44 08 PM |
| Overlay menu has an option open | Screenshot 2023-01-06 at 6 49 35 PM |

**Bottom bar for the non-learner**:
- In the App Context, the bottom bar for learner should be functional
- bottom bar should only appear in the app context 
- For all functioning navigation routes, no matter what plugin they are currently in (For example, if currently on the device page, "Home" on the bottom bar should bring to the LEARN home page, "Library" to the Library page, etc.)
-  user should be able to navigate with any of the bottom icon buttons using TAB key, and focus outline should be present
- the MENU icon, should open an overlay that plugin information (very similar to the current browser sidebar) 
- within this overlay menu, the "LEARN" button should open and close, to reveal that same navigation tab options (home, library, bookmarks). These options should work for navigation
- within this overlay menu, the "COACH" button should open and close, to reveal that same navigation tab options. If the user is within the coach plugin, these options should work and direct to the main page. If the user is not currently in the Coach plugin, they will be directed to the main class page. (i.e. currently in Device > open menu > click Coach>Reports > see class main page). _This will be updated and is just an MVP_
- within this overlay menu, the "DEVICE" button should open and close, to reveal that same navigation tab options (home, library, bookmarks). These options should work for navigation
- within this overlay menu, the "FACILITY" button should open and close, to reveal the navigation tab options. If the user only has one facility, these options should work. If they have multiple facilities, it will NOT work.  _This will be updated and is just an MVP_
- The menu should be keyboard navigable, with the focus outline visible, and the focus trapped (contained within the menu while open, not disappearing across the page)

| Descption  | Screenshot |
|---|---|
| Bottom bar is present  |  Screenshot 2023-01-06 at 6 59 23 PM |
| Overlay menu has an option open | Screenshot 2023-01-06 at 6 59 17 PM |

## References

…

## Reviewer guidance

…

----

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated


marcellamaki:
Build issues present in https://github.com/learningequality/kolibri/pull/9767 have been resolved and this should now be ready for initial testing @pcenov @radinamatic. Thank you!

pcenov:
Hi @marcellamaki I have started testing this PR and so far I've been able to identify the following issues:

Regression testing on the .deb asset:

| Issue| Screenshot |
| The 'Sign-in' form is not displayed at all. I tried installing the build both by upgrading a 0.15 one or directly installing the 0.16 to the same result. |  |
| The bottom bar is being displayed while it shouldn't as I am not in app context |  https://user-images.githubusercontent.com/79847249/214612403-572dc05f-c684-4e4f-b90f-a8b5b38ef8d6.mp4 |

New feature testing in the Android app:

| Issue| Screenshot |
| There's no indication which is the currently selected tab and there are no tab labels |  |
| The menu doesn't open to the plugin in which the user is currently in, all menu items are collapsed | |
| The navigation is not working correctly if I am switching between the Home and the Menu tab or while I am at a specific page and I attempt to select the same page from the menu |  https://user-images.githubusercontent.com/79847249/214615325-1de262a0-b763-4373-937a-1a70a0f2ad82.mp4 |
| Missing “Coach > Class home” link in the side-bar |  |

I'll continue testing tomorrow as well. Let me know if you need any of these reported as separate issues.

marcellamaki:
Hi @pcenov - thanks for beginning your testing on this! I've pushed one change which was an oversight on my part which I think is causing the debian issue. It should be resolved now 🤞 (and this should prevent you from running into a similar issue on other operating systems as well). I will beginning addressing remaining feedback tomorrow and Friday! No need for separate follow up issues for now :)

pcenov:
Hi @marcellamaki here's a list with a few more:

| Issue| Screenshot |
| Windows/Debian - the bottom bar is still being displayed in the latest build |  |
| The user points are not displayed in the sidebar/Learner should be capitalized | |
| If I am signed in as a super admin, admin or coach then the name, username and type of user labels are not displayed in the sidebar |  |
| Figma: Both the top bar and the navigation bar should hide as the user scrolls down, and show again as they scroll up | |
| The plugin name should be positioned to the left of the screen |  |
| For Learn-only device the Permissions, Facilities, Info and Settings links should be hidden |  |
| For Learn-only device: Missing indication that this is a Learn-only device  |  |
| Some of the strings that I know we have translations for are not translated when I switch to a language different than English  | |

marcellamaki:
Hi @pcenov - I have made a first batch of changes but still need to add those for the learn only devices. I should have these finished tomorrow and will update you when it is ready for retesting. Thank you!

marcellamaki:
Hi @pcenov. The following issues should now be resolved: 

- There's no indication which is the currently selected tab and there are no tab labels
- The menu doesn't open to the plugin in which the user is currently in, all menu items are collapsed
- The navigation is not working correctly if I am switching between the Home and the Menu tab or while I am at a specific page and I attempt to select the same page from the menu
- Missing “Coach > Class home” link in the side-bar
- The user points are not displayed in the sidebar/Learner should be capitalized
- If I am signed in as a super admin, admin or coach then the name, username and type of user labels are not displayed in the sidebar
- The plugin name should be positioned to the left of the screen
- For Learn-only device the Permissions, Facilities, Info and Settings links should be hidden

The following issues either I could not replicate or can be managed in follow up: 
- Figma: Both the top bar and the navigation bar should hide as the user scrolls down, and show again as they scroll up: There is a new bug in the main header component (it should also work here - it's a regression). I will create and assign myself that as an issue, and ensure as part of it, we confirm it works here as well 
- Some of the strings that I know we have translations for are not translated when I switch to a language different than English: Due to some refactoring, the IDs used to reference the strings have changed in some places, or are unavailable. 
- For Learn-only device: Missing indication that this is a Learn-only device: I'm not able to replicate this... I'll do some more testing and can open a follow up issue if it persists or isn't resolved. I'm hoping I resolved it unknowingly with some of my other fixes, and that's why I can't see the problem now on my setup. 
- I'm still not sure what is happening with the Windows/Debian bottom bar. Do you see the bottom bar if you are running the .pex file in the browser? 

Thank you for your attentive testing, as always!!

pcenov:
Hi @marcellamaki here's a list with new issues in the latest build:

| Issue| Screenshot |
| Sidebar - Not sure if you are using the correct icon and also the points get displayed only after tapping on the icon while in Figma the points are displayed to the right of the icon |  |
| Sidebar - The opened tab and current page should be highlighted | |
| Sidebar - All links in Coach are broken | https://user-images.githubusercontent.com/79847249/217255432-b37bbb9b-3252-44b0-9e34-10260caaeedb.mp4 |
| App bar - When I launch the app for the first time all the labels for Home, Library and Bookmarks are displayed in blue. It should also be specified whether the labels should be displayed for all icons or only for the currently selected one as is displayed in Figma  |  |
| Library page - Some strange links are displayed, not sure if this is caused by changes made here | https://user-images.githubusercontent.com/79847249/217258306-72d14cf0-2b02-4ba2-afdb-59652e0fdb90.mp4 |
| App bar - All tabs appear as selected while I am at the 'Profile' page |  |
| App bar - If the 'Home' label is a longer string it is not displayed correctly positioned under the icon. Try with Fulfulde Mbororoore or Ukrainian for example |  |
| App bar - When I tap the 'Menu' icon it remains highlighted with a transparent circle |  |
| App bar - When I tap on an icon and then tap again I see the sidebar and the icon gets highlighted | https://user-images.githubusercontent.com/79847249/217263909-fe08f3dc-df51-4863-b0cb-b6560ed33c8d.mp4 |

Regarding the issues which you are not able to replicate:
- For the first 2 points - ok, these can be tracked in follow-up issues
- For the learn-only device indication - I'm not able to setup a learn-only device as the wizard crashes and also for some reason the app won't let me install it on top of a 0.15 version which I previously used as a workaround... so I am not able to retest the issue with the missing text for Learn-only device.
- The app bar is no longer being displayed when using the .deb asset.

marcellamaki:
Hi @pcenov - I believe 🤞 I have addressed all of your new feedback. Thank you!

pcenov:
Hi @marcellamaki, I am testing with the following build assets: .[deb](https://buildkite.com/learningequality/kolibri-debian/builds/6029#01867ad0-8c51-457c-b427-0d958ba33401) and [Android](https://buildkite.com/learningequality/kolibri-android-installer/builds/6225#01867ad0-881c-4dde-9a33-7d261afd38e5) and the main issue is that accessing any of the pages in the Coach plugin is still not working in both the desktop version and the mobile app: https://user-images.githubusercontent.com/79847249/217255432-b37bbb9b-3252-44b0-9e34-10260caaeedb.mp4 Perhaps you were testing with leftover data from a previous installation so try with a clean installation if you are not able to replicate it.

The following minor issues from the list above are also still valid:

| Issue| Screenshot |
| App bar - When I tap the 'Menu' icon it remains highlighted with a transparent circle |  |
| App bar - When I tap on an icon and then tap on it again the icon gets highlighted | https://user-images.githubusercontent.com/79847249/220905815-21d644fd-cf37-4d08-ac22-079a0e340eb8.mp4 |

The rest is looking good!

marcellamaki:
@pcenov @radinamatic this should be ready for re-review (with "is it string freeze mergeable?" in mind). I have not been able to test the coach problem on a physical android device, but did resolve the browser issue and started testing from a clean state, as Peter recommended, so I am hoping this is resolved 🤞 @rtibbles adding you as a code reviewer. I will be opening a separate issue tomorrow to deal with the remaining coach routing redirect issues which I have not completed yet, and will add any remaining post-string-freeze follow up tasks there

kolibri/core/assets/src/constants.js
```diff
@@ -102,6 +102,7 @@ export const UPDATE_MODAL_DISMISSED = 'UPDATE_MODAL_DISMISSED';
 
 export const NavComponentSections = {
   ACCOUNT: 'account',
+  LEARN: 'LearnSideNavEntry',
```
rtibbles: We happen to be using this for learn now, but really this is just registration for the bottom navigation right? So we could just call it 'BottomBar' navigation, or something similar. It's more descriptive of the role it ends up playing, and means it could be flexibly used for other things going forwards.

kolibri/plugins/facility/assets/src/views/FacilityManagementSideNavEntry.vue
```diff
@@ -16,24 +32,132 @@
...
+      },
+      handleNav(route) {
+        this.isActiveLink(route) ? this.toggleAndroidMenu() : null;
+      },
     },
     role: UserKinds.ADMIN,
     priority: 10,
+    $trs: {
+      data: {
+        message: 'Data',
```
rtibbles: Presumably these are already defined somewhere else? Maybe just move them into a common translator object. https://github.com/learningequality/kolibri/blob/develop/kolibri/plugins/facility/assets/src/views/FacilityTopNav.vue#L54

kolibri/plugins/device/assets/src/views/DeviceManagementSideNavEntry.vue
```diff
@@ -1,36 +1,166 @@
...
       device: {
         message: 'Device',
         context:
           'The device is the physical or virtual machine that has the Kolibri server installed on it.',
       },
-    },
-    computed: {
-      url() {
-        return urls['kolibri:kolibri.plugins.device:device_management']();
+      permissionsLabel: {
```
rtibbles: Can move these into CommonDeviceStrings to avoid duplication with https://github.com/learningequality/kolibri/blob/develop/kolibri/plugins/device/assets/src/views/DeviceTopNav.vue#L68

kolibri/core/assets/src/views/AndroidNavigationNestedMenu.vue
```diff
@@ -0,0 +1,550 @@
...
+      /**
+       * @public
+       * Focuses on correct first element for FocusTrap.
+       */
+      focusLastEl() {
+        this.$refs.closeButton.$el.focus();
+      },
+    },
+    $trs: {
+      navigationLabel: {
```
rtibbles: These strings are just duplicates of strings in SideNav.vue.

kolibri/core/assets/src/views/AndroidNavigationNestedMenu.vue
```diff
@@ -0,0 +1,550 @@
+<template>
```
rtibbles: I know there are some differences here, but the wholesale duplication of the SideNav component with some mild tweaks feels fairly problematic here. I would much rather we either conditionalize the existing SideNav component with props, or break it up slightly to allow the display as a slide out with modal backdrop. Either way, as all the strings in here already exist in SideNav, we can reuse those and ensure strings stay in the SideNav component.

rtibbles (changes requested):
I think the implementation still needs some work to avoid having to come back and do a fairly heavy refactor after release. Most of the strings here are not new, but rather have been copied from another place in the codebase.

In a separate PR for string freeze, let's _move_ those strings into appropriate common translator objects rather than duplicating them, and update the references accordingly. Any actually new strings in here appear to be in common translator objects already so should be easy to add.

kolibri/core/assets/src/views/AndroidNavigationNestedMenu.vue
marcellamaki: yes, that is a really good idea.

pcenov:
Hi @marcellamaki - no new issues observed while testing this one and I can confirm that the Coach plugin issues are fixed now (tested with a physical Android device).

kolibri/core/assets/src/views/CoreMenu/CoreMenuOption.vue
```diff
@@ -83,13 +128,56 @@
...
         }
       },
+      isActiveLink(route) {
+        return route.includes(this.$router.currentRoute.path);
+      },
+      submenuShouldBeOpen() {
+        // which plugin are we currently in?
+        if (this.subRoutes) {
+          const key = Object.keys(this.subRoutes)[0];
+          this.subRoutes[key].route.includes(window.location.pathname)
```
rtibbles: This might lead to false positives in the case of say, Learn, where MyDownloads is nested under the learn plugin, but not at the top level.

We need to ensure an exact match of the pathname rather than `includes`, as your use of the pathname has already excluded any query parameters and the hash.

kolibri/core/assets/src/views/MenuNav.vue
```diff
@@ -228,7 +272,7 @@
   ];
 
   export default {
-    name: 'SideNav',
+    name: 'MenuNav',
```
rtibbles: You have renamed the component, but this component contains strings, this breaks string freeze. Will need to revert the rename here.

kolibri/plugins/coach/assets/src/routes/index.js
```diff
@@ -47,6 +47,7 @@ export default [
     },
   },
   {
+    name: ClassesPageNames.HOME_PAGE,
```
rtibbles: This creates an implicit coupling between the HomePage component name and this constant - it might be safer to use this constant to name the HomePage, to ensure that this is not accidentally broken by someone later deciding to rename the HomePage component to something else!

kolibri/plugins/coach/assets/src/views/CoachSideNavEntry.vue
```diff
@@ -2,8 +2,8 @@
 
   <CoreMenuOption
     :label="coreString('coachLabel')"
-    :link="url"
```
rtibbles, line 5: Just to make sure I am understanding correctly - the top level option now no longer links, and instead reveals the top level nav items to click on?

kolibri/plugins/coach/assets/src/views/CoachSideNavEntry.vue
```diff
@@ -16,17 +16,57 @@
   import commonCoreStrings from 'kolibri.coreVue.mixins.commonCoreStrings';
   import navComponents from 'kolibri.utils.navComponents';
   import urls from 'kolibri.urls';
+  import { PageNames } from '../constants';
+  import { generateNavRoute } from '../../../../../core/assets/src/utils/generateNavRoutes';
```
rtibbles: This should reference the coreAPI spec, if it is here at all.

kolibri/plugins/coach/assets/src/views/CoachSideNavEntry.vue
```diff
@@ -16,17 +16,57 @@
...
+    mixins: [commonCoach, commonCoreStrings, coachStringsMixin],
     computed: {
       url() {
         return urls['kolibri:kolibri.plugins.coach:coach']();
       },
+      coachRoutes() {
+        return {
+          home: {
+            text: this.coreString('classHome'),
+            route: this.generateNavRoute(PageNames.HOME_PAGE),
```
rtibbles: This still feels very unclean as an interface, and I think we might regret it.

kolibri/plugins/device/assets/src/views/DeviceManagementSideNavEntry.vue
rtibbles: These strings appear to be new here still.

kolibri/plugins/facility/assets/src/views/FacilityManagementSideNavEntry.vue
```diff
@@ -27,13 +30,72 @@
...
+            route: this.generateNavRoute(FacilityPageNames.DATA_EXPORT_PAGE),
+          },
+        };
+      },
+    },
+    methods: {
+      generateNavRoute(route) {
+        // if class id
+        let params;
+        if (this.$store.getters.currentFacilityId) {
```
rtibbles: This is incomplete and will break navigation in the case that a user is navigating a facility that is not their own.

This appears to be compounding the regression noted here: https://github.com/learningequality/kolibri/issues/10294

kolibri/plugins/facility/assets/src/views/FacilityManagementSideNavEntry.vue
rtibbles: More new strings that have not been cleaned up.

kolibri/plugins/learn/assets/src/views/LearnSideNavEntry.vue
```diff
@@ -14,20 +15,64 @@
...
       CoreMenuOption,
     },
     computed: {
-      url() {
-        return urls['kolibri:kolibri.plugins.learn:learn']();
+      learnRoutes() {
+        return {
+          home: {
+            text: this.coreString('homeLabel'),
+            icon: 'dashboard',
```
rtibbles: My suggestion was to just add a boolean to these routes:
```
bottomBar: true,
```

Rather than define them all over again.

package.json
```diff
@@ -65,7 +65,9 @@
   "browserslist": [
     "extends browserslist-config-kolibri"
   ],
-  "dependencies": {},
+  "dependencies": {
+    "path-to-regexp": "1.8.0"
```
rtibbles: This needs to be added to the `kolibri/core/package.json` not at the top level package.

kolibri/core/assets/src/constants.js
rtibbles: Guessing that maybe this just isn't used at all now?

rtibbles (commented):
Getting closer, but still quite a bit of cleanup needed, and string freeze is being broken in a few places here.

I think we can clean up the route generation part of the interface even more to allow that to just pass a definition rather than a generated route, and leave those implementation details up to the core.

kolibri/core/assets/src/views/MenuNav.vue
marcellamaki: 🤦‍♀️ right

kolibri/plugins/coach/assets/src/views/CoachSideNavEntry.vue
marcellamaki: Correct, unless there are not other nav items (i.e. profile page)

kolibri/plugins/coach/assets/src/routes/index.js
marcellamaki: I see what you're saying. And, is this different than some of the other ways that the page name constants are used throughout coach? i.e. [other class routes](https://github.com/marcellamaki/kolibri/blob/ebc5bea002ed58900d58f5e842f121394a10c7f9/kolibri/plugins/learn/assets/src/routes/classesRoutes.js#L29). Maybe it's not, and we are just trying to move away from that. I just used it for reference and am wondering if there is something specific here that is unique that we are trying to avoid.

rtibbles: One thing we did differently in some coach routes was rely on the vue router implicit behaviour of using the component name as the route name. We are then making that implicit behaviour explicit, but wanted to make sure we don't make a footgun for ourselves.

marcellamaki: Ah, that makes sense. Thanks

marcellamaki: So, I tried  

> use this constant to name the HomePage

And I got an error that the component didn't have a name... maybe I was not doing what you were actually suggesting? 
`name: PageNames.HOME_PAGE`
(I realized it doesn't have to be `ClassesPageNames` at all)

rtibbles: If it's a pain, don't worry about it. Just maybe leave a code comment where the constant is defined that it needs to be the same as the HomePage component name.

kolibri/core/assets/src/views/CoreMenu/CoreMenuOption.vue
```diff
@@ -178,6 +186,10 @@
       handleNav(route) {
         this.isActiveLink(route) ? this.toggleAndroidMenu() : null;
       },
+      generateNavRoute(route) {
+        console.log('url', this.link, 'route', route, 'subroutes', this.subRoutes);
+        return generateNavRoute(this.link, route, this.subRoutes);
```
rtibbles: Not sure why we need to pass in all the subroutes here - we have the route object for this specific subroute, it doesn't need to know about the others. I think cleaning that up will simplify the logic in the generateNavRoute function too.

The other thing is - do we want to conditionalize this - if we are in `this.link` we can use vue router to generate the link, if not we need to use the route generation.

This is also the place where we could reinject params from the current route if we are on the SPA in question.

kolibri/core/assets/src/views/CoreMenu/CoreMenuOption.vue
```diff
@@ -141,6 +145,10 @@
           this.$emit('select');
         }
       },
+      createLabel(text) {
+        const string = `${text.toLowerCase()}Label`;
+        return this.coreString(string);
```
rtibbles: This creates an implicit coupling between the plugin and the core strings, so this makes this not very pluggable.

kolibri/plugins/learn/assets/src/views/LearnSideNavEntry.vue
```diff
@@ -16,63 +17,25 @@
...
 
-  const url = urls['kolibri:kolibri.plugins.learn:learn']();
   const component = {
     name: 'LearnSideNavEntry',
     mixins: [commonLearnStrings, commonCoreStrings],
     components: {
       CoreMenuOption,
     },
     computed: {
-      learnRoutes() {
```
rtibbles: I think you've gone too far in just using the static baseRoutes. You need to be able to reference strings here in order to get the labels in a sufficiently flexible way, and there's also no specification of 'show on bottom bar' in the baseRoutes you have defined.

marcellamaki:
@rtibbles this is ready for (hopefully) final review. 

@radinamatic and @pcenov - if Richard ✅  the coding changes (a significant refactor), after that it will be ready for QA re-review. 

One **meaningful change** is [the decision to include the routes within the "regular" side menu](https://learningequality.slack.com/archives/C0LKG14NL/p1679076874303889), in addition to the within the "app" menu. 

Additional notes for some extra edge-case QA: 
- Moving back and forth between facilities and classes pages, specifically. (And, with no classes, with one class, and with multiple classes. Combined with - devices having one facility, devices having more than one facility). I have done testing here and tried to handle all combinations, but a fresh round of testing would be very helpful 
- Log in and log out working as expected, and the proper option (i.e. if you are signed in "Sign out" is displayed, etc.) is visible
- Correct display of options based on user permissions
- I suspect that despite my best efforts, we will need a follow up issue to align this with the PIN settings changes. I would prefer to merge this and handle that in follow up if it _is needed_, as long as @rtibbles is on board with that approach

marcellamaki:
@radinamatic @pcenov confirming I have updated after merging Alex's "My Downloads" PR, so that option in the menu should be working according to the refactored code for managing menu items

kolibri/core/assets/src/utils/generateNavRoutes.js
```diff
@@ -0,0 +1,34 @@
...
+
+export function generateNavRoute(rootUrl, pathReference, params = {}) {
+  let compiledRoute;
+
+  // when there is a direct path
+  compiledRoute = `${rootUrl}#${pathReference}`;
+
+  // if the path requires params but none exist, send to root
+  const pathParams = [];
+  pathToRegexp(compiledRoute, pathParams);
```
rtibbles, line 11: This is not a knock to your code, but the API of `path-to-regexp` - mutating the input parameters, really?!

kolibri/core/assets/src/views/CoreMenu/CoreMenuOption.vue
```diff
@@ -53,18 +91,27 @@
...
+        visibleSubMenu: false,
+      };
     },
-    inject: ['showActive'],
     computed: {
-      active() {
-        const showActive = typeof this.showActive !== 'undefined' ? this.showActive : true;
-        return showActive && window.location.pathname.startsWith(this.link);
+      isActive() {
+        return window.location.pathname.includes(this.label.toLowerCase());
```
rtibbles: This seems incorrect - this should be `this.link` I think?

Also, my previous comment here about `includes` still stands - this feels like it will mark the Learn page as active when we are on the MyDownloads page.

kolibri/core/assets/src/views/CoreMenu/CoreMenuOption.vue
```diff
@@ -83,12 +130,56 @@
...
+    mounted() {
+      this.submenuShouldBeOpen();
     },
     methods: {
-      conditionalEmit() {
-        if (!this.link) {
-          this.$emit('select');
+      isActiveLink(route) {
+        // temporary second condition to account for downloads route structure
+        if (!this.$route.name || this.$route.name != 'MY_DOWNLOADS') {
```
rtibbles: Doing proper matching above should fix this.

kolibri/core/assets/src/views/CoreMenu/CoreMenuOption.vue
```diff
@@ -83,12 +130,56 @@
...
+        if (this.subRoutes && this.subRoutes.length > 0) {
+          // temporary second condition to account for downloads route structure
+          window.location.pathname.includes(this.link) && this.$route.name != 'MY_DOWNLOADS'
+            ? (this.visibleSubMenu = true)
+            : (this.visibleSubMenu = false);
         }
+        return false;
+      },
+      subpathStyles(route) {
+        if (this.isActiveLink(route) && this.$router.currentRoute.name != 'PROFILE') {
```
rtibbles: Why is Profile special?

If it is special, then this should be specified as a property in the Profile nav item, and then conditionalized on that.

kolibri/plugins/coach/assets/src/routes/planRoutes.js
```diff
@@ -4,12 +4,18 @@ import GroupsPage from '../views/plan/GroupsPage';
...
 import { showGroupsPage } from '../modules/groups/handlers';
+import HomePage from '../views/home/HomePage';
 import planLessonsRoutes from './planLessonsRoutes';
 import planExamRoutes from './planExamRoutes';
 
 export default [
   ...planLessonsRoutes,
   ...planExamRoutes,
+  {
+    name: 'HomePage',
```
rtibbles: This route is now defined twice: https://github.com/learningequality/kolibri/blob/develop/kolibri/plugins/coach/assets/src/routes/index.js#L50 I am not sure what purpose defining it here has, but it should only be defined once, especially as the definition here  lacks a handler.

kolibri/plugins/coach/assets/src/views/CoachAppBarPage.vue
```diff
@@ -6,7 +6,7 @@
   >
     <AppBarPage :title="appBarTitle || defaultAppBarTitle">
       <template
-        v-if="showSubNav"
+        v-if="showSubNav && !isAppContext"
```
rtibbles, line 9: It seems odd that the `TopNavbar` is conditionally shown based on `!isAppContext` here, but no replacement is specified within the context of this page. Either the conditionality should all be within this page, or not.

(Also makes me suspect that the contents of `TopNavbar` might now be redundant with the contents of our routes specification)

kolibri/plugins/coach/assets/src/views/CoachSideNavEntry.js
```diff
@@ -0,0 +1,39 @@
+import navComponents from 'kolibri.utils.navComponents';
+import urls from 'kolibri.urls';
+import coreStrings from 'kolibri.utils.coreStrings';
+import { UserKinds } from 'kolibri.coreVue.vuex.constants';
+import baseRoutes from '../routes/baseRoutes';
+import { coachStrings } from './common/commonCoachStrings';
+
+const component = {
```
rtibbles: A slight nit - maybe we call it something else now that this definitely isn't a component?

`const sideNavConfig = {`?

kolibri/plugins/coach/assets/src/views/common/commonCoachStrings.js
```diff
@@ -31,6 +31,11 @@ const coachStrings = createTranslator('CommonCoachStrings', {
     context:
       "In the Plan > Quizzes section, coaches can create new quizzes using the 'New quiz' button.",
   },
+  plan: {
```
rtibbles: This looks like a new string to me? Breaking string freeze.

kolibri/plugins/user_auth/assets/src/views/LogoutSideNavEntry.js
```diff
@@ -0,0 +1,30 @@
+import { UserKinds, NavComponentSections } from 'kolibri.coreVue.vuex.constants';
```
rtibbles: The logout side nav entry is defined in core: https://github.com/learningequality/kolibri/blob/develop/kolibri/core/assets/src/views/LogoutSideNavEntry.vue

Probably  best that it stay here (as it's not tied to any plugin).

Worth noting that when you convert it to a config object rather than a component, you would need to make sure the translator namespace is identical to the component `name`: https://github.com/learningequality/kolibri/blob/develop/kolibri/core/assets/src/views/LogoutSideNavEntry.vue

kolibri/plugins/user_auth/kolibri_plugin.py
```diff
@@ -55,3 +55,8 @@ def url(self):
 @register_hook
 class LogInNavAction(NavigationHook):
     bundle_id = "login_side_nav"
+
+
+@register_hook
+class LogOutNavAction(NavigationHook):
```
rtibbles: See `kolibri/core/kolibri_plugin.py` for where this should be (and currently is) defined.

rtibbles (changes requested):
Some special cases that should be eliminated. Need to get rid of the salt on the string freeze here too.

kolibri/plugins/coach/assets/src/views/CoachAppBarPage.vue
rtibbles: Note: this does not have to be fixed before merge, but I'd like us to file a follow up issue for post-0.16 to address this.

kolibri/plugins/coach/assets/src/views/common/commonCoachStrings.js
marcellamaki: Ah, good catch. This is already in commonCoach from before string freeze and this is a duplicate that I didn't intent to re-add with this commit.

kolibri/core/assets/src/views/BottomNavigationBar.vue
```diff
@@ -0,0 +1,119 @@
...
+    },
+    methods: {
+      generateNavRoute(route, url) {
+        const params = this.$route.params;
+        return generateNavRoute(url, route, params);
+      },
+      isActiveLink(route) {
+        return (
+          route.includes(this.$router.currentRoute.path) &&
+          this.$router.currentRoute.name != 'PROFILE'
```
rtibbles: What's the reason for this special case for the PROFILE page?

The corollary here being, we probably shouldn't be special casing any pages, it smells.

kolibri/core/assets/src/views/BottomNavigationBar.vue
```diff
@@ -0,0 +1,119 @@
+<template>
+
+  <div class="bottom-bar" :style="{ backgroundColor: $themeTokens.textInverted }">
+    <span v-for="(routeDefinition, key) in bottomMenuOptions[0].routes" :key="key">
+      <a
```
rtibbles, line 5: Maybe I am misunderstanding - but isn't the BottomNavigationBar not always showing links for the currently active SPA? Could we not use VueRouter then, rather than the generateNavRoute?

kolibri/core/assets/src/views/CoreMenu/CoreMenuOption.vue
```diff
@@ -1,39 +1,77 @@
...
-        <div
-          v-if="secondaryText"
-        >{{ secondaryText }}</div>
-      </slot>
-    </a>
-  </li>
+  <div>
+    <!-- routes that have subpaths -->
+    <li v-if="subRoutes && subRoutes.length > 0">
+      <a
```
rtibbles, line 6: Is this an `a` tag just to make it appear clickable?

kolibri/core/assets/src/views/CoreMenu/CoreMenuOption.vue
```diff
@@ -83,12 +130,52 @@
...
+      },
+      toggleAndroidMenu() {
+        this.$emit('toggleAndroidMenu');
+      },
+      handleNav(route) {
+        this.isActiveLink(route) ? this.toggleAndroidMenu() : null;
+      },
+      generateNavRoute(route) {
+        const params = this.$route.params;
+        return generateNavRoute(this.link, route, params);
```
rtibbles, line 178: Feels like this could still be conditionalized to use vue router in the case we are on the SPA?

rtibbles (changes requested):
Unless there's a very good reason, I'd like the PROFILE page special casing removed.

Aside from that I think my queries about using vue router rather than the path to regexp utility functions can be migrated to a follow up.

kolibri/core/assets/src/views/BottomNavigationBar.vue
marcellamaki: it is not. in the current design - it is always showing the learn routes. I was supposed to follow up with Jessica on this and forgot. will ask in slack now

kolibri/core/assets/src/views/CoreMenu/CoreMenuOption.vue
marcellamaki: I thought it should actually be an <a> tag because it has to do with page navigation, like from an accessibility perspective, but maybe it doesn't have to be...

rtibbles: Yeah my only thought is that semantically it's not a link, so having it as an `a` feels a bit odd. But it should definitely be _something_ that has a pointer property that makes it look clickable, and will be in the tab sequence for keyboard navigation.

marcellamaki:
Follow up issue started here: https://github.com/learningequality/kolibri/issues/10395

kolibri/core/assets/src/views/BottomNavigationBar.vue
```diff
@@ -0,0 +1,124 @@
...
+    methods: {
+      generateNavRoute(route, url) {
+        const params = this.$route.params;
+        return generateNavRoute(url, route, params);
+      },
+      isActiveLink(route, url) {
+        return window.location.pathname === url && route == this.$router.currentRoute.path;
+      },
+      handleNav(route) {
+        if (this.isActiveLink(route)) {
```
rtibbles: This looks like it will always be false, because `url` is undefined here.

kolibri/core/assets/src/views/BottomNavigationBar.vue
```diff
@@ -0,0 +1,124 @@
...
+        };
+      },
+    },
+    methods: {
+      generateNavRoute(route, url) {
+        const params = this.$route.params;
+        return generateNavRoute(url, route, params);
+      },
+      isActiveLink(route, url) {
+        return window.location.pathname === url && route == this.$router.currentRoute.path;
```
rtibbles, line 79: I think route might be a vue-router route specification, in which case it will include parameter names that will never be in the path.

rtibbles: As we are in the SPA in question here, we can just check the route.name against `this.$router.name`, right?

rtibbles (commented):
A couple of questions about active links - also I don't see where we conditionally style the top level side nav option when we are on its associated SPA? Did I miss that?

rtibbles (approved):
This provides a stable API for the side nav and nested navigation routes - I think we can merge this and move forward with incremental internal changes to the implementation without breaking that API.

Issue #9884, opened 2022-12-01
A .pdf file gets marked as completed before showing the completion modal to the learner


pcenov:
## Observed behavior
This one was hard to spot in testing, but when one is scrolling down slowly while reading a .pdf file, the completion modal get's displayed after the start icon for the completion is displayed and only once the user has scrolled down completely.

## Expected behavior
The completion modal should be displayed prior or simultaneously with the completion icon.

## Steps to reproduce the issue
1. Install the the following [build ](https://github.com/learningequality/kolibri/releases/tag/v0.15.9-beta4) 
2. While viewing a small .pdf file (for example one from the QA Channel) start scrolling down slowly but don't scroll down to the end of the document while being at the last page.
3. Observe the header section where the resource will get marked as completed before showing the completion modal.

## Video

https://user-images.githubusercontent.com/79847249/205079305-53bc48ba-c34a-46f1-bc8c-cf7111d23c96.mp4

## Usage Details
Windows 10, Ubuntu - Chrome, Firefox

pcenov:
@radinamatic

AllanOXDi:
Let me take a look.

rtibbles:
This is intended behaviour, and was the purpose of the change made in addressing https://github.com/learningequality/kolibri/issues/9348 - that we now explicitly wait for the user to be at the end of a resource before displaying the completion modal, even if their progress has been set to complete.

pcenov:
In that case we have an interesting philosophical question - is there a difference between the intended behaviour and the expected behaviour? :) I think the intended behaviour here can be confusing to a varying degree to a learner but it's also reasonable to expect that a learner would be able to scroll all the way down so definitely not a major issue...

rtibbles:
In discussing this with @AllanOXDi - the only thing I could think of is tightening the 'page complete' requirement for the PDF renderer to require a learner to scroll all the way to the bottom of a page for it to count as complete - but that also feels kind of finicky, because then a user might not get it marked as complete when they expect it would. I think I'd prefer it this way round than the other.

nucleogenesis:
Not sure if I should make a new issue for this, but I just found this while looking through some PDFs and seems relevant to the issue discussed above.

If you have a PDF with one page and the whole page is visible when the page loads, then you get the completion modal before you even have a chance to read anything.

[...]

PR #9985, opened 2023-01-10
Only display the completion modal on the finish event when the resource is also complete

### LINKED ISSUE #9983: The completion modal for a .pdf resource get's displayed prematurely when using the bookmarks
## Observed behavior
The completion modal for a .pdf resource get's displayed prematurely when using the bookmarks and this happens every time I open the .pdf file until it actually gets completed.

## Expected behavior
The .pdf resource should be marked as completed only after it actually is completed. 

1. Install the the following [build ](https://buildkite.com/learningequality/kolibri-python-package/builds/6937) 
2. While viewing a .pdf file (for example one from the QA Channel) first click a bookmark bringing you to the end of the document and then scroll down to complete the pdf.

## Video

https://user-images.githubusercontent.com/79847249/211343071-2c9ad481-2708-44f4-b572-1cfbf447c5c9.mp4

## Usage Details
Windows 10, Ubuntu - Chrome, Firefox

rtibbles:
## Summary
* Fixes an issue where the completion modal would display even if the resource isn't complete, simply because the `finish` event was emitted.
* Does this by either checking whether complete is true, or adding a short lived watcher for complete, in case the progress update happens in a small interval after the finish event.

## References
Fixes #9983

## Reviewer guidance
Confirm the fix of #9983 PDFs
Also confirm that the completion modal still properly displays when videos complete

----

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/plugins/learn/assets/src/views/ContentPage.vue
+++ b/kolibri/plugins/learn/assets/src/views/ContentPage.vue
@@ -209,7 +209,7 @@
         this.hideMarkAsCompleteModal();
         // Do this immediately to remove any delay
         // before the completion modal displays if appropriate.
-        this.onFinished();
+        this.displayCompletionModal();
         return this.updateProgress(1.0)
           .then(() => {
             this.$store.dispatch('createSnackbar', this.learnString('resourceCompletedLabel'));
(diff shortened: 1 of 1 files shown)


pcenov:
@rtibbles I confirm that the issue described in https://github.com/learningequality/kolibri/issues/9983 is fixed now. No regression issues observed with video completions and the completion modal is correctly displayed there.

I'm noting that the issue which is already reported in https://github.com/learningequality/kolibri/issues/9884 can also be replicated by using the bookmarked links in the pdf in which case the experience of the learner is that they do not see the completion modal unless they scroll all the way down to the end of the .pdf:

https://user-images.githubusercontent.com/79847249/211820777-0dec1809-7ce8-41ad-8bb8-2c113fb4e1c5.mp4

In order to see the completion modal in the Android app I have to keep zooming in and out the pages as just scrolling down to the last page doesn't do the trick.

marcellamaki (approved):
looks good. As discussed with @rtibbles, the additional note that @pcenov flagged will be difficult to handle in the scope of this, and is part of an ongoing discussion about the appropriate times for the completion modal to appear (or not appear). As it is not a regression, and there are several different possible paths forward, we will not address in the scope of this pr/patch release.

PR #9993, opened 2023-01-12
added modal for setting PIN

### LINKED ISSUE #9992: Allow PIN creation from the facility settings page
## Observed behaviour
As a device admin, I should be able to create a PIN from the facility page so that I can manage my facility settings.

## Expected behaviour

The user should be prompted to enter a four-digit PIN 
After clicking the create a PIN button

## Steps to reproduce
Navigate to the facility settings page

AllanOXDi:
## Summary
This PR introduces a modal to handle  PIN creating on the device facility setting page
## References
Closes #9992
## Reviewer guidance
Navigate to the facility  setting page, locate the button to create PIN, click and observe the behaviour

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/plugins/facility/assets/src/views/FacilityConfigPage/ChangePinModal.vue
+++ b/kolibri/plugins/facility/assets/src/views/FacilityConfigPage/ChangePinModal.vue
@@ -13,7 +13,7 @@
 
       <KTextbox
         v-model="name"
-        type="number"
+        input="number"
         :label="$tr('enterPinLabel')"
         :maxlength="4"
         @blur="nameBlurred = true"
--- a/kolibri/plugins/facility/assets/src/views/FacilityConfigPage/ChangePinModal.vue
+++ b/kolibri/plugins/facility/assets/src/views/FacilityConfigPage/ChangePinModal.vue
@@ -2,8 +2,8 @@
 
   <KModal
     :title="$tr('title')"
-    :submitText="$tr('save')"
-    :cancelText="$tr('cancel')"
+    :submitText="coreString('saveAction')"
+    :cancelText="coreString('cancelAction')"
     @submit="submit"
     @cancel="$emit('cancel')"
   >
@@ -12,10 +12,10 @@
       <p>{{ $tr('setPin') }}</p>
 
       <KTextbox
-        ref="myfocus"
+        ref="pinFocus"
         v-model="pin"
         input="number"
-        :label="$tr('enterPinLabel')"
+        :label="coreString('enterPinPlaceholder')"
         :maxlength="4"
         :invalid="true"
         :invalidText="pinError"
@@ -61,7 +61,7 @@
         this.showSnackbarNotification('pinUpadeted');
       },
       focus: function() {
-        this.$refs.myfocus.focus();
+        this.$refs.pinFocus.focus();
       },
     },
     $trs: {
@@ -77,18 +77,6 @@
       setPin: {
         message: 'Choose four numbers to set as your new PIN.',
         context: 'Label to allow user to choose numbers to set PIN',
-      },
-      cancel: {
-        message: 'cancel',
-        context: 'cancel btn text',
-      },
-      save: {
-        message: 'save',
-        context: 'Text for save button',
-      },
-      enterPinLabel: {
-        message: 'Enter PIN',
-        context: 'Enter PIN label',
       },
     },
   };
--- a/kolibri/plugins/facility/assets/src/views/FacilityConfigPage/ChangePinModal.vue
+++ b/kolibri/plugins/facility/assets/src/views/FacilityConfigPage/ChangePinModal.vue
@@ -1,0 +1,65 @@
+<template>
+
+  <KModal
+    :title="$tr('title')"
+    :submitText="$tr('save')"
+    :cancelText="$tr('cancel')"
+    @submit="$emit('submit')"
+    @cancel="$emit('cancel')"
+  >
+    <div>
+      <p>{{ $tr('needToSync') }}</p>
+      <p>{{ $tr('setPin') }}</p>
+
+      <KTextbox
+        v-model="name"
+        type="number"
+        :label="$tr('enterPinLabel')"
+        :maxlength="4"
+        @blur="nameBlurred = true"
+      />
+    </div>
+  </KModal>
+
+</template>
+
+
+<script>
+
+  import commonCoreStrings from 'kolibri.coreVue.mixins.commonCoreStrings';
+
+  export default {
+    name: 'ChangePinModal',
+    mixins: [commonCoreStrings],
+    methods: {},
+
+    $trs: {
+      title: {
+        message: 'Change device management PIN',
+        context: 'Title for the change pin modal.',
+      },
+      needToSync: {
+        message:
+          'You will need to sync this device with other devices that share this facility in order to use this PIN.',
+        context: 'Reminder to sync devices',
+      },
+      setPin: {
+        message: 'Choose four numbers to set as your new PIN.',
+        context: 'Label to allow user to choose numbers to set PIN',
+      },
+      cancel: {
+        message: 'cancel',
+        context: 'cancel btn text',
+      },
+      save: {
+        message: 'save',
+        context: 'Text for save button',
+      },
+      enterPinLabel: {
+        message: 'Enter PIN',
+        context: 'Enter PIN label',
+      },
+    },
+  };
+
+</script>
--- a/kolibri/plugins/facility/assets/src/views/FacilityConfigPage/ChangePinModal.vue
+++ b/kolibri/plugins/facility/assets/src/views/FacilityConfigPage/ChangePinModal.vue
@@ -4,7 +4,7 @@
     :title="$tr('title')"
     :submitText="$tr('save')"
     :cancelText="$tr('cancel')"
-    @submit="$emit('submit')"
+    @submit="submit"
     @cancel="$emit('cancel')"
   >
     <div>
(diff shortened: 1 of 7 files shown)


kolibri/plugins/facility/assets/src/views/FacilityConfigPage/CreateManagementPinModal.vue
```diff
@@ -0,0 +1,65 @@
...
+    @submit="$emit('submit')"
+    @cancel="$emit('cancel')"
+  >
+    <div>
+      <p>{{ $tr('newToSync') }}</p>
+      <p>{{ $tr('enterPin') }}</p>
+
+      <KTextbox
+        v-model="name"
+        type="number"
```
marcellamaki: A **non-blocking** comment: 

I can totally understand why you decided to add type 'number' here, since, it's a number. However, number text fields do have the increment/decrement arrows.

Screenshot 2023-01-20 at 11 11 10 AM

I wonder if having this alongside the need for a 4 digit input might be confusing? Part of me thinks yes, and part of me thinks "people know what a PIN is, nobody is going to click the arrow 1000 times to generate a 4 digit PIN".... 

Curious if you have any thoughts @AllanOXDi. Or if @akolson or @rtibbles have ideas.

kolibri/plugins/facility/assets/src/views/FacilityConfigPage/index.vue
```diff
@@ -146,13 +166,16 @@
...
+      BottomAppBar,
+      CreateManagementPinModal,
     },
     mixins: [commonCoreStrings, responsiveWindowMixin],
     data() {
       return {
         showModal: false,
         showEditFacilityModal: false,
         settingsCopy: {},
+        CreatePinShow: false,
```
marcellamaki: Very small nit-pick, I would change this to `createPinShow`, just for consistent cases

kolibri/plugins/facility/assets/src/views/FacilityConfigPage/index.vue
```diff
@@ -247,6 +270,9 @@
       copySettings() {
         this.settingsCopy = Object.assign({}, this.settings);
       },
+      handlecreatePin() {
+        this.CreatePinShow = true;
```
marcellamaki: same as above

kolibri/plugins/facility/assets/src/views/FacilityConfigPage/index.vue
```diff
@@ -299,6 +325,19 @@
...
+        message: 'Device management PIN',
+        context: 'The title for the device management pin',
+      },
+      deviceManagementDescription: {
+        message:
+          'This 4-digit PIN allows users to manage content and other settings on learn-only devices',
+        context: 'Description for the device management',
+      },
+      createPinBtn: {
+        message: 'CREATE PIN',
```
marcellamaki: Non-blocking, but just an "FYI" - the button should handle the capitalization of this for you, automatically 

In other instances when you might need to change the case of the string, you can use various string methods like [toUpperCase()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/toUpperCase). This can be helpful also if you might need to use the same string with different capitalization in different places.

marcellamaki (changes requested):
Hi @AllanOXDi - this looks great! I added one _very_ small change request, and two non-blocking comments. Please let me know if you have any questions! 

Awesome work. I'm really impressed with how your frontend skills are coming along! You're learning all of this so quickly! Great job 😺

kolibri/plugins/facility/assets/src/views/FacilityConfigPage/CreateManagementPinModal.vue
rtibbles: On the flip side, specifying `number` means that on devices with on screen keyboards, it will display an on screen keyboard with just numeric input, which is a distinct usability win.

That said, it looks like we can get the same result by using an input type text or password and adding the `inputmode` attribute set to `numeric` https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/inputmode

So, I'd suggest using a text input, with the `inputmode` set to `numeric` and then setting the `pattern` attribute to do the number validation: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#pattern

Can use a simple regex like `[0-9]{4}` to ensure the 4 digit PIN?

AllanOXDi: Hi @marcellamaki @rtibbles, Thanks for pointing this out and thanks for the suggestions.

kolibri/plugins/facility/assets/src/views/FacilityConfigPage/ChangePinModal.vue
```diff
@@ -0,0 +1,65 @@
+<template>
+
+  <KModal
+    :title="$tr('title')"
+    :submitText="$tr('save')"
+    :cancelText="$tr('cancel')"
```
marcellamaki: For "Save" and "Cancel" actions in all of these modals, you can use the `commonCoreString`s for these, as the strings already exist. 

For example, [this modal uses them](https://github.com/learningequality/kolibri/blob/develop/kolibri/plugins/facility/assets/src/views/FacilityConfigPage/EditFacilityNameModal.vue#L5). Don't forget that you will need to [import commonCoreStrings](https://github.com/learningequality/kolibri/blob/develop/kolibri/plugins/facility/assets/src/views/FacilityConfigPage/EditFacilityNameModal.vue#L32) in all of your files.

kolibri/plugins/facility/assets/src/views/FacilityConfigPage/CreateManagementPinModal.vue
```diff
@@ -0,0 +1,95 @@
+<template>
+
+  <KModal
+    :title="$tr('title')"
+    :submitText="$tr('save')"
+    :cancelText="$tr('cancel')"
```
marcellamaki: Same as above

kolibri/plugins/facility/assets/src/views/FacilityConfigPage/CreateManagementPinModal.vue
```diff
@@ -0,0 +1,95 @@
...
+    },
+    computed: {},
+    methods: {
+      submit() {
+        if (!this.pin) {
+          this.showErrorText = true;
+          this.pinError = 'This field cannot be empty';
+          this.focus();
+        } else {
+          if (this.pinPattern.test(this.pin)) {
```
marcellamaki, line 53: Nice work with the validation!

kolibri/plugins/facility/assets/src/views/FacilityConfigPage/CreateManagementPinModal.vue
```diff
@@ -0,0 +1,95 @@
...
+    :cancelText="$tr('cancel')"
+    @submit="submit"
+    @cancel="$emit('cancel')"
+  >
+    <div>
+      <p>{{ $tr('newToSync') }}</p>
+      <p>{{ $tr('enterPin') }}</p>
+
+      <KTextbox
+        ref="myfocus"
```
marcellamaki: While you can label the ref with anything, using a label that describes the component (for example, "pinInput") rather than what you are using the ref _for_  (for example, what you have here, referring to the focus) can be helpful to keep track of what a ref is pointing to, especially for future coding. This ref is only used once, and there is only one in this file, but in some files, you might have multiple refs. And sometimes, you might be using them across files. 

You can see [an example where there are two refs within the same file, and their descriptive names](https://github.com/learningequality/kolibri/blob/develop/kolibri/core/assets/src/views/userAccounts/PasswordTextbox.vue#L5).

kolibri/plugins/facility/assets/src/views/FacilityConfigPage/RemovePinModal.vue
```diff
@@ -0,0 +1,48 @@
+<template>
+
+  <KModal
+    :title="$tr('title')"
+    :submitText="$tr('removePin')"
+    :cancelText="$tr('cancel')"
```
marcellamaki: Same as above with common strings

kolibri/plugins/facility/assets/src/views/FacilityConfigPage/ViewPinModal.vue
```diff
@@ -0,0 +1,42 @@
+<template>
+
+  <KModal
+    :title="$tr('title')"
+    :cancelText="$tr('close')"
+    @cancel="$emit('cancel')"
+  >
+    <div>
+      <p>{{ $tr('pin') }}</p>
```
marcellamaki: Here, we will want to actually pass in the value of each user's pin, dynamically, from the database. 

Right now, of course, that isn't set up yet, so it is okay to use a placeholder! But, for the placeholder, it might be better to use a `prop` or `computed` value, since that is most likely what you will end up displaying once you continue connecting this to the backend. 

So this could look something like 

```
... 

<p>{{ userPin }}</p>

...

computed: {
      userPin() {
        // TO DO, add computed PIN value after backend is set up
        return 1234;
      },
 
```

Depending on what you and @akolson are thinking for the data here, you might have a different idea of how to do this, which is fine. This is just one suggestion :)

kolibri/plugins/facility/assets/src/views/FacilityConfigPage/index.vue
```diff
@@ -67,28 +67,34 @@
...
+            {{ $tr('createPinBtn') }}
+          </KButton>
+
+          <KButton
+            hasDropdown
+            :text="$tr('optionBtn')"
+            style="marginLeft:15px"
+          >
+            <template #menu>
+              <KDropdownMenu
```
marcellamaki, line 88: nice work with the dropdowns!

marcellamaki (changes requested):
Hi @AllanOXDi - great improvements and new work on this! So exciting to see this coming together 😄 

I've made a few small suggestions - but nothing major. Overall it's looking very good. The manual QA looks good too! ✅

marcellamaki (approved):
Nice work @AllanOXDi! I think this is good to merge. If any other adjustments need to be made when connecting this with the work Samson has done, I think it's fine to do that in follow up.

PR #10008, opened 2023-01-17
[EPUB] Prevent multiple results match highlight

### LINKED ISSUE #4239: EPUB search results highlight multiple instances of the search token within a single result
### Observed behavior

During search, it's possible that multiple results appear within the same context:

However then a separate link and context is shown for each result. This makes it look like there are 4 results instead of 2.

image

### Expected behavior

Only have a single link and quoted context section on the right-hand bar. This will make clear that the two results are the only ones visible 

### User-facing consequences

Initially I thought there were 4 results instead of 2

### Steps to reproduce

* Go to http://kolibri-beta.learningequality.org/learn/#/topics/c/3a721243e176458aae5563fd030c9858
* search for 'chord'

### Context

develop

----

cc @christianmemije

AlexVelezLl:
## Summary

During search, it's possible that multiple matches appear within the same list item because the text preview of the reference can contain a sentence with multiple occurrences of the same word. This PR tries to disambiguate which specific match each item in the result list refers to, so that for each item there is only one highlighted match

| Before | After |
| :-----: | :----: |
| image <br> search query "a" | image <br> search query "a" |
| image <br> search query "The" | image <br> search query "The" |

## References

closes #4239

## Reviewer guidance

Go to any epub resource (e.g. http://localhost:8000/en/learn/#/topics/c/1c5a63eb626f4aa883eaef0cfe37634f) and search for common queries that may appear multiple times in the same sentence (e.g. "The"), and verify that just one match is highlighted per each result item.

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/plugins/epub_viewer/assets/src/views/SearchSideBar.vue
+++ b/kolibri/plugins/epub_viewer/assets/src/views/SearchSideBar.vue
@@ -91,7 +91,9 @@
 <script>
 
   import Mark from 'mark.js';
-  import CFI from 'epub-cfi-resolver';
+  import { EpubCFI } from 'epubjs';
+  import isEqual from 'lodash/isEqual';
+
   import SideBar from './SideBar';
 
   /**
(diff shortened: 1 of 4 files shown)


kolibri/plugins/epub_viewer/assets/src/views/SearchSideBar.vue
```diff
@@ -88,6 +91,7 @@
 <script>
 
   import Mark from 'mark.js';
+  import CFI from 'epub-cfi-resolver';
```
rtibbles: Unfortunately, `epub-cfi-resolver` is licensed under an AGPL license, which is not compatible with the MIT license that we use for Kolibri (MIT is more permissive, and so including AGPL software in our codebase would require us to adopt a similarly restrictive license to AGPL for compatibility).

If what you are doing works with the EPUBJS CFI parser: https://github.com/futurepress/epub.js/blob/master/src/epubcfi.js I would suggest just using that instead. I note that the author of this library says that the epubcfi parser is incomplete, but I assume it's also using it for other things, so at least the breakage will be consistent.

rtibbles (dismissed):
Need to switch CFI parser implementation for licensing reasons.

pcenov:
@radinamatic  I've tested this in both Windows and Ubuntu and it is functioning as described.

rtibbles:
This looks good - just needs a rebase on develop to resolve the merge conflict with #10048 - and can add the follow up from there to remove the `markjs` dependency there too.

AlexVelezLl:
Hi @rtibbles ! Sorry for the delay, the PR has already been updated, removing the markjs dependency in the SearchSideBar, the results are still the same as the pr summary.

rtibbles (changes requested):
Code looks correct to me - @pcenov can you reconfirm the fix now that the implementation has been updated?

One thing still needs to be done which is removing the markjs dependency:

```
yarn workspace kolibri-document-epub-render-plugin remove mark.js
```

The name of the workspace is given by the name given in this plugin's package.json:
https://github.com/learningequality/kolibri/blob/develop/kolibri/plugins/epub_viewer/package.json#L2

AlexVelezLl:
Oh @rtibbles we can not remove the markjs dependency yet, because it is still used to mark the epub content. Which I think would take a bit longer to migrate

rtibbles:
Ah, ok - that makes sense - the EPUB content is significantly less under our control!

rtibbles (approved):

pcenov:
@rtibbles I've reconfirmed the fix - no issues observed. Thank you @AlexVelezLl!

Issue #9783, opened 2022-10-18
Device Settings Updates: Feature Disabled


marcellamaki:
## Overview 

This issue is a full-stack, but read-only issue that updates the Device Settings page to have disabled sections based on particular Kolibri configurations. 

This should be completed _after_ the other issues for adding the UI for these settings is complete. It should add a `disabled` attribute to each of these settings, and set conditions to check it. It should also add the banner to the front end, if one or more settings attributes is disabled. You can refer to the [KDS Docs](https://kolibri-design-system.netlify.app/) for more details on this setting with `KComponents`.

### Figma 

[Complete figma spec](https://www.figma.com/file/MpCG7r5PULOCJsEvnTSti8/Android-Designs-2022?node-id=1384%3A16773).
<img width="903" alt="Screen Shot 2022-10-21 at 1 44 08 PM" src="https://user-images.githubusercontent.com/17235236/197257151-5566f53a-5f81-46fe-8d33-1e1fc38faf3a.png">

### ➡️ Accessibility Notes: 
The banner at the top of the page should be closable through tab navigation. It does not need `aria-polite` as it will not be _added_ to the DOM once the user is already there (i.e. it is not a pop-up banner notification), but rather it is information about the state of Kolibri configuration.

## Acceptance Criteria
- [ ]  Frontend is updated so that device settings have a "disabled" option.
- [ ]  If one or more settings is disabled, the informational banner should appear, and be dismissible.
- [ ]  The `DeviceSettingsPage.spec.js` should be updated so that existing tests still pass (as necessary) and there should be tests written to cover this scenario as necessary.

## References 
- [Notion project page](https://www.notion.so/learningequality/Updates-to-Device-Settings-45529cbcc4f74aeb9b19559228c97579)
- [Figma spec](https://www.figma.com/file/MpCG7r5PULOCJsEvnTSti8/Android-Designs-2022?node-id=1384%3A16773)
- [Gherkin scenarios](https://www.notion.so/learningequality/Gherkin-scenarios-02096ac9925844f8902c818cab607d4b#1959b753421648308fb467e21bdaa125)

Issue #10019, opened 2023-01-18
Automatic resource syncing


rtibbles:
## Expected behavior

In order to ensure a seamless user experience when taking Kolibri learner only devices away from the main server, it is expected that not only will the metadata for quizzes and lessons be synced to the learner only device, but also that any relevant resources to support that lesson and quiz will also be imported onto the device, so that they can be effectively perused away from the server.

On the main server, coaches will expect to be able to monitor not just the data syncing (which they can currently do) but also the status of any resource syncing - particularly the case when the learner only device has run out of storage space and is unable to import the requisite resources. This will allow the coach to take remediating action, such as disabling outdated lessons and quizzes, or freeing up space on the learner only device.

In order to achieve this status reporting, ~importing into a learner-only device will create an implicit relationship between the originating device and the learner-only device which will be instantiated by a shared partition in a new morango profile `manageddevicedata`.~ we will create a new facility model to store the sync status of a device.

~The exact structure of the partitions might need some additional speccing to ensure that we are not preventing syncing from happening potentially to multiple parent servers.~

This new base LearnerDeviceStatus includes a CharField that is not validated, but contains a status constant string. It is intentionally not validated, so as to allow newer status strings from newer versions of Kolibri to still successfully sync. In addition, this model will contain a referential store of the Learner's unique id, and the instance id - the partition of the model will be based on the learner id, so that single user syncing devices only sync the statuses for relevant learners.

We should then enhance the learner only sync command to also run automated import of necessary resources for lessons and quizzes, which purges any unneeded files for lessons and quizzes which are no longer active. To do this using minimal disk space for channel metadata, it will use the single resource import functionality for each of the imported files.

This import functionality should report back if there is insufficient space, or if the import was successful, and store the appropriate status in the LearnerDeviceStatus object. Once this has been done, another sync of the facility data will happen to sync this status back to the server.

rtibbles:
The first part of this has been resolved by #10075 - there is now a model to track the status of the resource syncing.

rtibbles:
This functionality is now fully implemented. Any further issues can be treated as bugs, but this is complete!

Issue #8087, opened 2021-05-14
Device Settings Updates: Changes to "Allow guest access settings" are not immediately enforced on Sign-in page


jonboiser:
## Observed behavior
This comes from https://kolibri-beta.learningequality.org/en/user/#/signin, but I am also able to reproduce this locally

I turned off guest access in Device Settings:

![image](https://user-images.githubusercontent.com/10248067/118328945-438b0380-b4bb-11eb-83ae-d4bdb718dedb.png)

But I still see the "Explore..." button

![image](https://user-images.githubusercontent.com/10248067/118329029-62899580-b4bb-11eb-9f8d-5a0796221c9f.png)

If you view the source for the UserPage, the `plugin_data` script is out of date

Line 364:

```
window['kolibriPluginDataGlobal'] = window['kolibriPluginDataGlobal'] \|\| {};
  | window['kolibriPluginDataGlobal']['kolibri.plugins.user.app'] = JSON.parse("
{\"oidcProviderEnabled\":false,\"allowGuestAccess\":true}");
 
```

This looks like a similar behavior that was fixed in https://github.com/learningequality/kolibri/pull/5601 (which removed bootstrapping code from the Django template that was also causing stale information to be sent to the browser).

## Expected behavior

The guest access setting is reflected immediately after an admin changes it.

## Steps to reproduce

1. Change the guest access setting in device settings
2. Go back to the sign-in page
3. You will see the "explore without an account" button

marcellamaki:
Moving this to the Device Settings milestone, as it can be updated as part of that project.

jredrejo:
Finished in #10025

PR #10025, opened 2023-01-19
Complete device settings

### LINKED ISSUE #9960: Device Setting Updates: Ensure new Device Settings update the user's experience
## Overview 

This issue is a full-stack issue that makes sure all of the new device settings that have been implemented actually are used within the UI and are correctly updating the user's experience. It should be completed after the earlier device settings issues are finished, or at the same time as the remaining issues. 

## Acceptance Criteria
- [ ]  On the new device settings page, each of the new settings, when selected, not only updates that UI, but the saved values actually change the user's experience in accordance with that change 

## References 

jredrejo:
## Summary
Complete the missing parts for the new device settings updates.

After checking all the new features, these parts were  not working yet and have been implemented:
- Deselecting “Enable auto-download” should also deselect all descendants 
- Selecting a descendant should enable “Enable auto-download”
- Enable pages (learn, coach, etc.) was neither in the UX not working
- Download on metered connection should appear only in Android devices

A [follow up issue](https://github.com/learningequality/kolibri/issues/10019) has been created to ensure the metered connection on Android devices is implemented.

## References
Closes: #9960  #8087

## Reviewer guidance
Check all the missed points in the description and working properly. Also, review the [defined Gherkin scenarios ](https://www.notion.so/learningequality/Gherkin-scenarios-02096ac9925844f8902c818cab607d4b#1959b753421648308fb467e21bdaa125) are correctly described and implemented 

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/plugins/device/assets/src/modules/deviceInfo/index.js
+++ b/kolibri/plugins/device/assets/src/modules/deviceInfo/index.js
@@ -20,6 +20,12 @@ export default {
       state.deviceName = null;
     },
   },
+  getters: {
+    getDeviceOS(state) {
+      // Channel data just needs to exist, but doesn't need to be available
+      return state.deviceInfo.os;
+    },
+  },
   actions: {
     updateDeviceName(store, name) {
       return client({
--- a/kolibri/plugins/device/assets/src/modules/deviceInfo/index.js
+++ b/kolibri/plugins/device/assets/src/modules/deviceInfo/index.js
@@ -22,7 +22,6 @@ export default {
   },
   getters: {
     getDeviceOS(state) {
-      // Channel data just needs to exist, but doesn't need to be available
       return state.deviceInfo.os;
     },
   },
--- a/kolibri/plugins/device/assets/src/views/DeviceSettingsPage/ServerRestartModal.vue
+++ b/kolibri/plugins/device/assets/src/views/DeviceSettingsPage/ServerRestartModal.vue
@@ -100,7 +100,8 @@
         context: 'Description for restarting the server.',
       },
       newPluginsState: {
-        message: 'Changing the state of the enabled plugins will restart this server.',
+        message:
+          'Changing the state of the enabled plugins will restart this server. This page has to be refrehed to see the changes.',
         context: 'Description for restarting the server.',
       },
       selectedPath: {
--- a/kolibri/plugins/device/assets/src/views/DeviceSettingsPage/ServerRestartModal.vue
+++ b/kolibri/plugins/device/assets/src/views/DeviceSettingsPage/ServerRestartModal.vue
@@ -35,7 +35,7 @@
     mixins: [commonCoreStrings],
     props: {
       changedSetting: {
-        type: String, // primary, remove, add
+        type: String, // primary, remove, add, plugins
         required: true,
       },
       path: {
@@ -52,15 +52,22 @@
     methods: {
       getMessage() {
         let message = '';
-        if (this.changedSetting === 'primary') {
-          message = this.$tr('newPrimaryLocationRestartDescription');
-        } else if (this.changedSetting === 'remove') {
-          message = this.$tr('removeLocationRestartDescription');
-        } else if (this.changedSetting === 'add') {
-          message = this.$tr('newLocationRestartDescription');
+        switch (this.changedSetting) {
+          case 'primary':
+            message = this.$tr('newPrimaryLocationRestartDescription');
+            break;
+          case 'remove':
+            message = this.$tr('removeLocationRestartDescription');
+            break;
+          case 'add':
+            message = this.$tr('newLocationRestartDescription');
+            break;
+          case 'plugin':
+            message = this.$tr('newPluginsState');
+            break;
         }
 
-        return message + this.$tr('serverRestartDecription');
+        return message + this.$tr('serverRestartDescription');
       },
       handleSubmit() {
         if (this.changedSetting === 'add') {
@@ -87,9 +94,13 @@
         message: 'Removing a storage location will restart this server.',
         context: 'Reason to restart the server.',
       },
-      serverRestartDecription: {
+      serverRestartDescription: {
         message:
           ' Anyone using Kolibri on this server right now will temporarily be unable to use it.',
+        context: 'Description for restarting the server.',
+      },
+      newPluginsState: {
+        message: 'Changing the state of the enabled plugins will restart this server.',
         context: 'Description for restarting the server.',
       },
       selectedPath: {
--- a/kolibri/plugins/device/assets/src/views/DeviceSettingsPage/ServerRestartModal.vue
+++ b/kolibri/plugins/device/assets/src/views/DeviceSettingsPage/ServerRestartModal.vue
@@ -66,7 +66,7 @@
             message = this.$tr('newPluginsState');
             break;
         }
-
+        // message is a separate sentence, concatenating them is not problematic
         return message + this.$tr('serverRestartDescription');
       },
       handleSubmit() {
--- a/kolibri/plugins/device/assets/src/views/DeviceSettingsPage/index.vue
+++ b/kolibri/plugins/device/assets/src/views/DeviceSettingsPage/index.vue
@@ -244,24 +244,11 @@
           </p>
 
           <KCheckbox
-            :label="$tr('enabledLearn')"
-            :checked="enabledLearn"
-            @change="enabledLearn = $event"
-          />
-          <KCheckbox
-            :label="$tr('enabledCoach')"
-            :checked="enabledCoach"
-            @change="enabledCoach = $event"
-          />
-          <KCheckbox
-            :label="$tr('enabledFacility')"
-            :checked="enabledFacility"
-            @change="enabledFacility = $event"
-          />
-          <KCheckbox
-            :label="$tr('enabledProfile')"
-            :checked="enabledProfile"
-            @change="enabledProfile = $event"
+            v-for="plugin in dataPlugins"
+            :key="plugin.id"
+            :label="plugin.name"
+            :checked="plugin.enabled"
+            @change="plugin.enabled = $event"
           />
         </div>
       </section>
@@ -333,6 +320,7 @@
   import { mapGetters } from 'vuex';
   import find from 'lodash/find';
   import urls from 'kolibri.urls';
+  import { ref } from 'kolibri.lib.vueCompositionApi';
   import commonCoreStrings from 'kolibri.coreVue.mixins.commonCoreStrings';
   import { availableLanguages, currentLanguage } from 'kolibri.utils.i18n';
   import sortLanguages from 'kolibri.utils.sortLanguages';
@@ -374,51 +362,34 @@
     mixins: [commonCoreStrings],
     setup() {
       const { restart } = useDeviceRestart();
-      const { plugins, togglePlugin } = usePlugins();
+      const { plugins, fetchPlugins, togglePlugin } = usePlugins();
+      const dataPlugins = ref(null);
 
-      function getPluginState(pluginName) {
-        if (plugins.value === null) {
-          return false;
-        } // plugins not loaded yet
-        const plugin = plugins.value.find(p => p.id.includes(pluginName));
-        if (plugin === undefined) {
-          return false;
-        } // plugin not found
-        return plugin.enabled;
-      }
+      fetchPlugins.then(() => {
+        dataPlugins.value = plugins.value.map(plugin => ({ ...plugin }));
+      });
 
-      function getPluginId(pluginName) {
-        if (plugins.value === null) {
-          return null;
-        } // plugins not loaded yet
-        const plugin = plugins.value.find(p => p.id.includes(pluginName));
-        if (plugin === undefined) {
-          return null;
-        } // plugin not found
-        return plugin.id;
-      }
-
-      function checkAndTogglePlugin(pluginState, pluginName) {
-        let restart = null;
-        if (pluginState !== getPluginState(pluginName)) {
-          const plugin_id = getPluginId(pluginName);
-          if (plugin_id == null) {
-            return null;
+      function checkAndTogglePlugins() {
+        dataPlugins.value.forEach((plugin, index) => {
+          if (plugin.enabled !== plugins.value[index].enabled) {
+            togglePlugin(plugin.id, plugin.enabled);
           }
-          togglePlugin(plugin_id, pluginState);
-          restart = 'plugins';
-        }
-        return restart;
+        });
       }
 
-      function checkPluginChanges(pluginStates, pluginNames) {
-        const previousPluginStates = pluginNames.map(pluginName => getPluginState(pluginName));
-        return pluginStates.some(
-          (pluginState, index) => pluginState !== previousPluginStates[index]
-        );
+      function checkPluginChanges() {
+        // returns true if any of the plugins have changed its
+        // enabled state
+        const unchanged = dataPlugins.value.every((plugin, index) => {
+          if (plugin.enabled !== plugins.value[index].enabled) {
+            return false;
+          }
+          return true;
+        });
+        return !unchanged;
       }
 
-      return { restart, plugins, getPluginState, checkPluginChanges, checkAndTogglePlugin };
+      return { restart, dataPlugins, checkPluginChanges, checkAndTogglePlugins };
     },
     data() {
       return {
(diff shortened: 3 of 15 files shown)


marcellamaki (commented):
Hi @jredrejo - thank you for working on this. I've taken a look and the first settings seem to be working as intended. 

One question I have is regarding the expected behavior for disabling or enabling various plugins. I'm not sure if this is new, or if it's an existing bug is that although I do see the plugin removed from the side navigation, when I log in I can still access the page. 

For example, here as an admin, I've disabled the coach plugin. But if I log in as coach, I still see the coach home page (which... I suppose makes sense, because what else would they see), but the plugin is not visible in the navigation. It just seems a bit confusing to me. At the same time, if access were completely reduced, I'm not sure what the coach would see... Maybe just redirected to Learn?

Screenshot 2023-02-01 at 9 50 35 AM

I've also added @radinamatic so she or @pcenov could confirm the android-specific notification on an actual device. 
> Download on metered connection should appear only in Android devices

jredrejo:
> Hi @jredrejo - thank you for working on this. I've taken a look and the first settings seem to be working as intended.
> 
> One question I have is regarding the expected behavior for disabling or enabling various plugins. I'm not sure if this is new, or if it's an existing bug is that although I do see the plugin removed from the side navigation, when I log in I can still access the page.
> 
> For example, here as an admin, I've disabled the coach plugin. But if I log in as coach, I still see the coach home page (which... I suppose makes sense, because what else would they see), but the plugin is not visible in the navigation. It just seems a bit confusing to me. At the same time, if access were completely reduced, I'm not sure what the coach would see... Maybe just redirected to Learn?
> 
> Screenshot 2023-02-01 at 9 50 35 AM
> 
> I've also added @radinamatic so she or @pcenov could confirm the android-specific notification on an actual device.
> 
> > Download on metered connection should appear only in Android devices

Hi @marcellamaki good catch. I think this is an UX thing where @jtamiace might help. In the code I've done, the navigation bar is the affected part, I didn't check that use case. Knowing the exact context for the purpose of this feature would help, to know if it makes sense that coaches can access to the coach plugin after login or not.

rtibbles:
> For example, here as an admin, I've disabled the coach plugin. But if I log in as coach, I still see the coach home page (which... I suppose makes sense, because what else would they see), but the plugin is not visible in the navigation. It just seems a bit confusing to me. At the same time, if access were completely reduced, I'm not sure what the coach would see... Maybe just redirected to Learn?

This seems potentially like a caching issue - if the coach plugin is disabled, the coach plugin should not be viewable, it should simply 404 (and more importantly, with the coach plugin disabled, the coach would never be redirected there after login in the first place).

jtamiace:
Yeah I think redirecting to another page on sign in makes sense

For facility admin users:
* If the Facility plugin is disabled, redirect to Coach
* If the Facility and Coach plugins are disabled: redirect to Learn

For coach users:
* If Coach is disabled, redirect to Learn

If an admin or coach somehow manually inputs the URL to the Coach or Facility plugin while it's disabled, we should probably block it? With a message like "Coach/Facility features have been disabled on this device." 

Could use this page layout:
Screenshot 2023-02-02 at 11 19 46 AM

jredrejo:
> > For example, here as an admin, I've disabled the coach plugin. But if I log in as coach, I still see the coach home page (which... I suppose makes sense, because what else would they see), but the plugin is not visible in the navigation. It just seems a bit confusing to me. At the same time, if access were completely reduced, I'm not sure what the coach would see... Maybe just redirected to Learn?
> 
> This seems potentially like a caching issue - if the coach plugin is disabled, the coach plugin should not be viewable, it should simply 404 (and more importantly, with the coach plugin disabled, the coach would never be redirected there after login in the first place).

Actually, the plugin is not disabled, it's just the menu is not visible, that's why this is happening. I didn't understand in the mockups the plugin should be disabled, but it seems that's the case. I'll change it so it will behave as @jtamiace  has commented below.

rtibbles:
Ah, yes, I see this isn't actually using the composable for managing plugins at all... this should be doing that: https://github.com/learningequality/kolibri/blob/develop/kolibri/plugins/device/assets/src/composables/usePlugins.js

rtibbles:
The viewset this is connecting to is defined here: https://github.com/learningequality/kolibri/blob/develop/kolibri/core/device/api.py#L306

rtibbles:
The other general thing here is that we should not be encoding details about plugins in either a) other plugins, or b) in core.

The names of the plugins are exposed translated via the API endpoint defined here, so that we don't include their names in Device, e.g.: https://github.com/learningequality/kolibri/blob/develop/kolibri/plugins/learn/kolibri_plugin.py#L30 I would suggest either making a general purpose string for the tooltip that substitutes in the plugin name, or to add them to the plugin data so they can be translated independently.

jredrejo:
> Ah, yes, I see this isn't actually using the composable for managing plugins at all... this should be doing that: https://github.com/learningequality/kolibri/blob/develop/kolibri/plugins/device/assets/src/composables/usePlugins.js

@marcellamaki @rtibbles  changes done to enable/disable plugins instead of hiding the menu options. Thanks for the suggestions!

kolibri/plugins/device/assets/src/modules/deviceInfo/index.js
```diff
@@ -20,6 +20,12 @@ export default {
       state.deviceName = null;
     },
   },
+  getters: {
+    getDeviceOS(state) {
+      // Channel data just needs to exist, but doesn't need to be available
```
rtibbles: I don't understand this code comment.

kolibri/plugins/device/assets/src/views/DeviceSettingsPage/ServerRestartModal.vue
```diff
@@ -52,15 +52,22 @@
...
+            break;
+          case 'add':
+            message = this.$tr('newLocationRestartDescription');
+            break;
+          case 'plugin':
+            message = this.$tr('newPluginsState');
+            break;
         }
 
-        return message + this.$tr('serverRestartDecription');
```
rtibbles, line 63: Could you add a caveat comment here to explain that because each string is a separate sentence, then concatenating them here is not a concern.

kolibri/plugins/device/assets/src/views/DeviceSettingsPage/ServerRestartModal.vue
```diff
@@ -52,15 +52,22 @@
     methods: {
       getMessage() {
         let message = '';
-        if (this.changedSetting === 'primary') {
-          message = this.$tr('newPrimaryLocationRestartDescription');
-        } else if (this.changedSetting === 'remove') {
-          message = this.$tr('removeLocationRestartDescription');
-        } else if (this.changedSetting === 'add') {
-          message = this.$tr('newLocationRestartDescription');
+        switch (this.changedSetting) {
```
rtibbles, line 55: Not blocking - but the overall stylistic tendency in the Kolibri codebase has been for `if`/`else if` code blocks rather than switch statements, mostly because of the ease with which one can forget to `break` in a switch statement, and so get the incorrect result.

kolibri/plugins/device/assets/src/views/DeviceSettingsPage/index.vue
```diff
@@ -900,6 +1044,46 @@
...
       },
+      enabledPages: {
+        message: 'Enabled pages',
+        context: 'Label for enabled pages section',
+      },
+      enabledPagesDescription: {
+        message: 'Unselect a page to hide it even if the user has permission to access it.',
+        context: "Description for the 'Enabled pages' section.",
+      },
+      enabledLearn: {
```
rtibbles: This string is available, translated, from the API call. As are all the other names of the screens here.

kolibri/plugins/device/assets/src/views/DeviceSettingsPage/index.vue
```diff
@@ -900,6 +1044,46 @@
...
+        message: 'Learn',
+        context: 'Label for the Learn page',
+      },
+      enabledLearnDescription: {
+        message: 'Unselect to hide the Learn page even if the user has permission to access it.',
```
rtibbles: As these descriptions are completely generic, it seems like we could just use a single string here:
```
message: 'Unselect to hide the { page } page even if the user has permission to access it.',
```

kolibri/plugins/device/assets/src/views/DeviceSettingsPage/index.vue
```diff
@@ -236,6 +234,40 @@
...
+
+        <div class="fieldset">
+          <h2>
+            {{ $tr('enabledPages') }}
+          </h2>
+          <p class="info-description">
+            {{ $tr('enabledPagesDescription') }}
+          </p>
+
+          <KCheckbox
```
rtibbles, line 246:
```suggestion
          <KCheckbox
            v-for="plugin in plugins"
            :label="plugin.name"
            :checked="plugin.enabled"
            :description="$tr('pluginDescription', { page: plugin.name })"
            @change="togglePlugin(plugin.id)"
          />
```
This would seem to suffice (except for not tracking that a restart could happen - that would have to dig into some other logic).

rtibbles (dismissed):
Should remove knowledge of the plugins from the device plugin here.

I'd also prefer to minimize the diff and reduce risk of future edits by not converting the `if/else` blocks to switch statements, but if you feel strongly about it, it's not a blocker.

kolibri/plugins/device/assets/src/views/DeviceSettingsPage/index.vue
jtamiace: While reviewing the strings here, we realized the description text under each plugin checkbox label wasn't necessary, and have deleted them. 

Screenshot 2023-02-06 at 4 06 03 PM

rtibbles: Oh, in that case, no string at all, and we can remove any reference to other plugins from within this plugin!

kolibri/plugins/device/assets/src/modules/deviceInfo/index.js
jredrejo: Me neither, I must have been left it after changing this code several times. Deleted

kolibri/plugins/device/assets/src/views/DeviceSettingsPage/ServerRestartModal.vue
jredrejo: Done

kolibri/plugins/device/assets/src/views/DeviceSettingsPage/ServerRestartModal.vue
jredrejo: I'd agree for just two or even three `if/else if `, but having 4 different options I think readability is much better using a `switch`. Forgetting a `break` is one code error as any other that should be fixed in a review. I was not aware of that stylistic decision. I don't have any problem in changing it if you prefer it though.

kolibri/plugins/device/assets/src/views/DeviceSettingsPage/index.vue
jredrejo: I'll explain this in a latter comment

kolibri/plugins/device/assets/src/views/DeviceSettingsPage/index.vue
jredrejo: I see two problems with this implementation (which is obviously nicer and I had tested it before submitting the deifinitive code):

- The way  `usePlugins` composable is built (which I don't like too much,  depends on a Promise to have the plugins, but the promise is not public for the composable user, so I had to add a watch to know when plugins had been read) would make the page blink when the plugins info are received, as there would not be any plugin when the component is mounted. This is the same reason not to use the translation provided by the api.
- It would automatically enable/disable the plugin, not giving the user the oportunity to cancel it if the user does not want to restart kolibri. Also, as you mentioned, the logic to avoid restarting kolibri three times if three plugins are changed would not work.

For the first point I think the change might be acceptable if you think the UX is not bad. For the `@change` event, I think it should be kept as it is now.

rtibbles: > which I don't like too much, depends on a Promise to have the plugins, but the promise is not public for the composable user, so I had to add a watch to know when plugins had been read

Seems like the best thing to do here is to move the automatic fetching out of the composable, or to add a loading state ref in there, and use that to display a loading state for the plugins.

> For the @change event, I think it should be kept as it is now.

Yeah, I think doing additional tracking for the change event is fine, that was the part that I was most iffy about.

> It would automatically enable/disable the plugin

This could be reverted if they didn't want to, but yes, it would need additional tracking of state. Please do edit the composable for the needs here, this is the only place it is used, and I just put it in place to have a starting point for interfacing with the API.

kolibri/plugins/device/assets/src/views/DeviceSettingsPage/ServerRestartModal.vue
rtibbles: I think our default eslint configuration enables this rule https://eslint.org/docs/latest/rules/no-fallthrough - so I think the danger here is mitigated, and it just becomes a stylistic choice, so we can leave as is.

kolibri/plugins/device/assets/src/views/DeviceSettingsPage/index.vue
jredrejo: Changes done. Thanks for all the feedback and good ideas.

pcenov:
Hi @jredrejo I've manually tested the following points and they are implemented correctly as described:

- Deselecting “Enable auto-download” should also deselect all descendants
- Selecting a descendant should enable “Enable auto-download”
- Enable pages (learn, coach, etc.) was neither in the UX not working
- Download on metered connection should appear only in Android devices

Here are some minor issues which should probably be addressed here as well but let me know if they are not in scope for this PR:

| Issue| Screenshot |
| Enabled pages - After disabling a page and after the server has restarted the page is still displayed in the sidebar. If possible it should no longer be displayed in that case. The reverse is also valid - once I enable a page it should be available in the sidebar. |  https://user-images.githubusercontent.com/79847249/217857911-b73f0a7f-192b-4b67-a5b8-a360f3a3a656.mp4 |
| Change of primary storage - The server is no longer auto-restarting. Tested in both Windows and Ubuntu. | https://user-images.githubusercontent.com/79847249/217859270-33c37d24-16b6-4782-80fc-8dbb3040b0a3.mp4 |
| Enabled pages - The list with pages is not being displayed in the Android app |  |

rtibbles:
> Enabled pages - After disabling a page and after the server has restarted the page is still displayed in the sidebar. If possible it should no longer be displayed in that case. The reverse is also valid - once I enable a page it should be available in the sidebar.

@jredrejo the easiest way to achieve this would probably be by refreshing the page.

jredrejo:
thanks @pcenov  for the tests. For the three issues you've found, I've solved one, won't solve another and can't solve the third. In more detail:

> Enabled pages - After disabling a page and after the server has restarted the page is still displayed in the sidebar.

This one won't be solved because:

- refreshing the page from JavaScript does not change the situation as the menu is cached
- even if it would solve it, the page won't change for the rest of the users connected to the server until they refresh their browsers

So I have left it as it was, but adding more info to the modal message, telling to the user  he needs to refresh the browser to see the pages, @radinamatic @jtamiace this is a new string that was not present in the Figma designs because they didn't expect restarting the server in this case, but it's needed
refresh

> Change of primary storage - The server is no longer auto-restarting. 

I had moved the restart until the save button was clicked, but I've restored the previous behaviour as it was probably better for the user, so this should be solved

> Enabled pages - The list with pages is not being displayed in the Android app

This is a new issue on Android you've discovered and can't be solved in this context. It has to be done in the Android development. #[10102](https://github.com/learningequality/kolibri/issues/10102) has been created to solve it

Thanks again for the tests, let me know if you see any other glitch!

pcenov:
Thank you @jredrejo I have verified the fix in both Windows and Ubuntu. No other issues observed - LGTM!

marcellamaki (approved):
this should be good to go! Thanks for the team effort everyone

radinamatic:
@jredrejo This is the new version of the string in the modal that came up in the conversation between @jtamiace, @carinediaz and me. Does it still describe well the circumstance to the user?

> When you enable or disable a page, Kolibri will restart, and you must refresh the browser to see the changes. Anyone using Kolibri on this server at this time will be temporarily disconnected. 

Note that we are favoring the term `page` instead of `plugin` and we are not using the latter in the user-facing UI.

jredrejo:
> Note that we are favoring the term `page` instead of `plugin` and we are not using the latter in the user-facing UI.
@radinamatic 
I'm happy with whatever is more correct for the users, if you provide me the exact wording I'll add it in the latest PR for DeviceSettings I'm currently working on (to close #9783)

Issue #9796, opened 2022-10-24
Facility Settings Updates: PIN usage on Learn-only Devices


marcellamaki:
## Overview 

This issue is a full-stack issue that updates allows the Device page to be accessed on learn-only devices by entering a PIN.

We will be using a feature branch for this project `update-facility-settings`. Please make PRs to this branch, which will later be merged into develop when the feature is complete.

### Figma 

When the user picks "Device" on the menu, the modal opens and prompts for the PIN to be entered. If the correct PIN is entered, the user should be authorized to access the device page. 

<img width="830" alt="Screen Shot 2022-10-24 at 11 53 02 AM" src="https://user-images.githubusercontent.com/17235236/197570382-10f9761e-e257-440e-bce6-500c49321842.png">

**Please see [complete figma spec](https://www.figma.com/file/MpCG7r5PULOCJsEvnTSti8/Android-Designs-2022?node-id=1473%3A16838) for details on states for PIN modal**

➡️ Accessibility notes: 
* Please use `KComponents` for wherever possible to ensure that the content is properly passed on to screen readers, and that focus outlining and focus shift works. (See [KDS docs](https://kolibri-design-system.netlify.app/) for more information)

### API changes 

Authentication will need to be updated to support a PIN granting access as an admin. A user should not be able to navigate directly via a URL without being prompted to authenticate, but should not have to enter the PIN twice, once when pressing "Device" in the navigation, and then again when redirected. One approach would be to save something in the session to say that it is authenticated, i.e. a 15 second cookie that can be used if the PIN has already been entered, and can be checked during navigation to the Device plugin.

Access should expire on return to Learn plugin and/or with a timeout.

## Acceptance Criteria
- [ ]  Frontend is updated with correct strings, using `KComponents`
- [ ] PIN authentication is implemented as described above
- [ ]  The `FacilitySettingsPage.spec.js` should be updated so that existing tests still pass (as necessary) and there should be at least one test written to cover this scenario

## References 
- [Notion project page](https://www.notion.so/learningequality/Updates-to-Facility-Settings-f35f3be6e5fb45fb8de0abc0d898c017)
- [Figma spec](https://www.figma.com/file/MpCG7r5PULOCJsEvnTSti8/Android-Designs-2022?node-id=969%3A16669)
- [Gherkin scenarios](https://www.notion.so/learningequality/Gherkin-scenarios-02096ac9925844f8902c818cab607d4b#1959b753421648308fb467e21bdaa125)

[...]

PR #10031, opened 2023-01-20
Implements facility settings pin setup

### LINKED ISSUE #9789: Facility Settings Updates: PIN setup
## Overview 

This issue is a full-stack issue that updates the Facility Settings page to include a setting where the user can set a PIN for accessing admin-level settings on a learner-only device. It also implements the full CRUD functionality, so a PIN can be added, deleted, viewed, or edited. 

To avoid having to create a new migration for each field, we will be using a feature branch for this project `update-facility-settings`. Please make PRs to this branch, which will later be merged into develop when the feature is complete.

### Figma 

**Please see [complete figma spec](https://www.figma.com/file/MpCG7r5PULOCJsEvnTSti8/Android-Designs-2022?node-id=969%3A16669) for details on states for PIN modal**

Tab order of the Facility settings page, and placement of the new "Create PIN" button: 

Screen Shot 2022-10-21 at 1 55 53 PM

➡️ Accessibility notes: 
* Please use `KComponents` for wherever possible to ensure that the content is properly passed on to screen readers, and that focus outlining and focus shift works. (See [KDS docs](https://kolibri-design-system.netlify.app/) for more information)

### API changes 

The `FacilityDatasetResource` should be usable as it has beens et up, and related actions (i.e. `kolibri/plugins/facility/assets/src/modules/facilityConfig/actions.js`) should be easily updatable to include the extra data.

Backend changes should include updating the `FacilityDataset` model `kolibri/core/auth/models.py` and any relevant serializers or viewsets.

## Acceptance Criteria
- [ ]  Frontend is updated with correct strings, using existing components
- [ ]  On saving, the setting is updated on both the frontend and the backend, and a *Settings have been updated* message displays. (This snackbar should already exists and it should appear by default on a successful update to the device settings).
- [ ]  The `DeviceSettingsPage.spec.js` should be updated so that existing tests still pass (as necessary) and there should be at least one test written to cover this scenario

## References 

akolson:
## Summary

This pr implements api endpoints for the facility pin set up. Two resources have been created for this purpose. 

- `facilitydataset/<dataset-id>/update-pin`.  This resource updates the pin in the facility settings.  
- `facilitydataset/<dataset-id>/is-pin-valid`. This resource checks if the input pin matches what is stored within the settings. This was created in preparation for Issue #9796.

## References

Contributes to #9789

## Reviewer guidance

- Use the functions(in `/modules/facilityConfig/actions.js`) below to consume the resources on the frontend as desired
  - `setPin(store, payload)`
  - `unset(store)`
  - `isPinValid(store, payload)`
- where `payload` is `{pin_code: "1234"}`
- import the functions from `kolibri/plugins/facility/assets/src/modules/facilityConfig/index.js` into the file where you want to consume the resources
- Please note that you will need to run data migrations using `kolibri manage migrate` or a combination of `kolibri manage makemigrations` and `kolibri manage migrate` for the resources to work correctly. The migrations add validation to the FacilityDataset `extra_fields`
----

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated


kolibri/plugins/facility/assets/src/modules/facilityConfig/index.js
```diff
@@ -54,10 +62,19 @@ export default {
         return f.name === payload.oldName;
       }).name = payload.newName;
     },
+    IS_FACILITY_PIN_VALID(state, valid) {
+      state.isFacilityPinValid = valid;
+    },
```
nucleogenesis: A minor note but calling this mutation `IS_` something makes it read more like a getter so maybe putting `SET_` on front might give the name of the mutation a bit more clarity.

kolibri/core/auth/api.py
```diff
@@ -195,6 +197,50 @@ def resetsettings(self, request, pk):
...
+                "A Pin is a required and it must be a number and 4 characters long"
+            )
+
+        try:
+            dataset = FacilityDataset.objects.get(pk=pk)
+            if dataset.extra_fields is None:
+                dataset.extra_fields = {}
+            dataset.extra_fields["pin_code"] = (
+                pin_code if pin_code is None else str_pin_code
+            )
```
nucleogenesis: A non-blocking thought but this logic around what is assigned to the `extra_fields["pin_code"]` on 217 I think would live better up where those variables are defined. The `pin_code if pin_code is None` bit of logic in particular would flow better around there.

I think maybe just assigning `str_pin_code` directly with either `None` or the str() version of the data received would simplify a bit as then you can just assign the `str_pin_code` to the extra_fields.

Or is the `not str_pin_code.isdigit()` call expecting that `str_pin_code` is assigning the string `"None"` when `pin_code` is `None`?

nucleogenesis (commented):
Came back through everything and looks good, particularly on the front-end side of things. I left a couple non-blocking comments.

I think that maybe clarifying that bit I mentioned would be helpful but await additional review for more thoughts there.

kolibri/core/auth/api.py
akolson: This piece of code has been refactored to put into considerations the suggestions above. Thanks @nucleogenesis

kolibri/core/auth/models.py
```diff
@@ -160,7 +177,12 @@ class FacilityDataset(FacilityDataSyncableModel):
     learner_can_delete_account = models.BooleanField(default=True)
     learner_can_login_with_no_password = models.BooleanField(default=False)
     show_download_button_in_learn = models.BooleanField(default=True)
-    extra_fields = JSONField(null=True, blank=True)
+    extra_fields = JSONField(
+        null=True,
+        blank=True,
+        validators=[JSON_Schema_Validator(extra_fields_schema)],
```
bjester, line 183: Nice! I like the use of a schema validator here :100:

bjester: If the model is saved with `extra_fields` missing the new `pin_code` field, does it pass validation?

akolson: Validation will fail with a `400`. Generally, one of fields should be present for validation to pass, preventing unknown fields from being saved. There should be a test covering this scenario

bjester: So if an existing database is upgraded to 0.16, and they haven't set a pin so the field doesn't exist, will other saves to the model fail because of that?

akolson: I don't think so as the `extra_fields` fields is an optional field and is `null` by default. But also, default values have been defined for the fields just in case.

kolibri/core/auth/models.py
```diff
@@ -75,11 +75,28 @@
...
 
 logger = logging.getLogger(__name__)
 
+extra_fields_schema = {
+    "type": "object",
+    "additionalProperties": False,
+    "properties": {
+        "facility": {"type": "object"},
+        "on_my_own_setup": {"type": "boolean"},
+        "pin_code": {"type": "string"},
```
bjester: If the `pin_code` is supposed to be a 4-digit number/integer, why is the type `string`?

akolson: The discussion here is what prompted my decision to use a string instead https://github.com/learningequality/kolibri/pull/9993#discussion_r1082758692. But I am also open to here any particular wins we could benefit from using `int` that may not be available with string.

bjester: Thanks for sharing that context, that reasoning makes sense to me for the input field. 

Thinking about the API separately though, it seems it likely accepts a JSON post body? If so, one benefit of the JSON body is that you can disambiguate between strings and numbers, since it's JSON, unlike with normal form-encoded post bodies. Of course, your API validation enforces that the string is essentially a number so aligning the parameter type and schema could make the API a bit clearer without needing the entire context of how the pin code is input or validated.

akolson: Thanks for this insight. Makes sense!

akolson: I have updated the code to take into consideration these recommendations. Thanks again @bjester

bjester (approved):
Looks good to me

PR #10049, opened 2023-01-29
[EPub_Viewer] Fix createMarks

### LINKED ISSUE #10048: EPub viewer rerendering old search query results
## Observed behavior

When I make some consecutive searches the results list eventually get bugged due to mark.js adding to the text some previous results.

For example with http://localhost:8000/en/learn/#/topics/c/1c5a63eb626f4aa883eaef0cfe37634f?prevName=HOME. If first I search by "a", this is the result:

image

After that, if I search by "The", this is the result:

image

but if I search for "a" again, now the results are bugged.

image

 Is like the results of the first search are merged with the results of the 2nd search. All these bugs are introduced due to the `createMarks` method, because if I deactivate the marks the results are correct again no matter how many times I search for the same queries.

## Expected behavior

Search results lists should always be with the correct matches no matter how many times I search for the same query.

## User-facing consequences

The results are very confusing for users since we are giving them portions of text that do not really exist as results.

## Steps to reproduce

Described in Observed behavior.

## Context

* Kolibri version: 0.16.0.dev0+git.20230117123659
* Operating system: macOS Monterrey 12.2.1
* Browser: Google Chrome

AlexVelezLl:
## Summary

There are wrong behaviors in the EPub search in specific situations like the one described in #10048. These behaviors are probably due to some incompatibility between Mark.js's way of marking DOM content and Vue's way of tracking DOM updates, which causes these DOM updates to sometimes fail with inconsistent results.

This PR changes the way of displaying the list of results from v-show to v-if so that vue does not set the `display: none` property but instead recognizes the rendering changes of the new results as a completely new DOM component and Mark .js always mark the results over a "clean" component.

## References

closes #10048

## Reviewer guidance

Perform the steps described in #10048 and check that those inconsistent results doesn't appear anymore.

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/plugins/epub_viewer/assets/src/views/SearchSideBar.vue
+++ b/kolibri/plugins/epub_viewer/assets/src/views/SearchSideBar.vue
@@ -62,7 +62,7 @@
     </transition>
 
     <ol
-      v-show="searchHasBeenMade && !searchIsLoading && searchResults.length > 0"
+      v-if="searchHasBeenMade && !searchIsLoading && searchResults.length > 0"
       ref="searchResultsList"
       class="search-results-list"
     >
@@ -153,7 +153,6 @@
       searchIsLoading: false,
       maxSearchResultsExceeded: false,
       searchHasBeenMade: false,
-      markInstance: null,
     }),
     computed: {
       numberOfSearchResults() {
@@ -187,29 +186,20 @@
                 this.maxSearchResultsExceeded = true;
               }
               this.searchResults = searchResults.slice(0, MAX_SEARCH_RESULTS);
-
+              this.$emit('newSearchQuery', searchQuery);
+              this.searchIsLoading = false;
               // Wait for list to be updated
               this.$nextTick().then(() => {
-                if (this.markInstance) {
-                  this.markInstance.unmark({
-                    done: () => this.createMarks(searchQuery),
-                  });
-                } else {
-                  this.createMarks(searchQuery);
-                }
+                this.createMarks(searchQuery);
               });
             }
           );
         }
       },
       createMarks(searchQuery) {
-        this.markInstance = new Mark(this.$refs.searchResultsList);
-        this.markInstance.mark(searchQuery, {
+        const markInstance = new Mark(this.$refs.searchResultsList);
+        markInstance.mark(searchQuery, {
           separateWordSearch: false,
-          done: () => {
-            this.$emit('newSearchQuery', searchQuery);
-            this.searchIsLoading = false;
-          },
         });
       },
     },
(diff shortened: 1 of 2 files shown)


kolibri/plugins/epub_viewer/assets/src/views/SearchSideBar.vue
```diff
@@ -187,29 +186,20 @@
                 this.maxSearchResultsExceeded = true;
               }
               this.searchResults = searchResults.slice(0, MAX_SEARCH_RESULTS);
-
+              this.$emit('newSearchQuery', searchQuery);
+              this.searchIsLoading = false;
               // Wait for list to be updated
               this.$nextTick().then(() => {
-                if (this.markInstance) {
-                  this.markInstance.unmark({
```
AlexVelezLl, line 194: Since the node component now is always new, there is. no need to unmark anything, because nothing is marked yet.

kolibri/plugins/epub_viewer/assets/src/views/SearchSideBar.vue
```diff
@@ -62,7 +62,7 @@
     </transition>
 
     <ol
-      v-show="searchHasBeenMade && !searchIsLoading && searchResults.length > 0"
+      v-if="searchHasBeenMade && !searchIsLoading && searchResults.length > 0"
       ref="searchResultsList"
       class="search-results-list"
     >
```
rtibbles, line 68: One thought that occurs is that we could simply do the highlight styling ourselves, and pass in the styled text into the `KButton` slot, rather than passing it as text and then doing the post hoc highlighting with Mark.

Something like:
```
        <KButton
          :text="item.excerpt"
          appearance="basic-link"
          class="search-results-list-item-button"
          @click="$emit('navigateToSearchResult', item)"
        >
        <span v-for="i, text in item.excerpt.split(searchQuery)" :key="i" :class="text.toLowerCase() === searchQuery.toLowerCase() ? 'highlight' : ''>
           {{ text }}
       </span>
    </KButton>
```

Obviously you would need to make the 'split' a bit more nuanced, as we are doing case insensitive searches, so you'd probably need a method to handle this better. And would need to define a `highlight` class to handle the highlighting.

rtibbles (commented):
Code changes make sense, but I have not yet tested.

Made one suggestion that we can file as a follow up issue.

kolibri/plugins/epub_viewer/assets/src/views/SearchSideBar.vue
AlexVelezLl: Yes, I think it's a super good suggestion! it could be added to #10008 since precisely in that PR I compute the `before the match`, `match` and `after the match` text portions in separate spans, and that preprocessing is calculated in another method that has in account the case insensitive.

pcenov:
@AlexVelezLl by manually testing this I confirm that the issue described in https://github.com/learningequality/kolibri/pull/10048 is fixed now as specified above.

kolibri/plugins/epub_viewer/assets/src/views/SearchSideBar.vue
rtibbles: Great, I'll leave that to you to add in #10008!

rtibbles (approved):
Manual QA passes, code looks good. Followup will happen in #10008

PR #10056, opened 2023-01-30
Learner insufficient storage notifications

### LINKED ISSUE #9844: Automatic Syncing Updates: Create Learner-facing notifications for insufficient storage
## Overview 

This issue is a full-stack issue that creates a banner-style in-Kolibri notification if a learner has run out of storage but their library is trying to sync, or when they are choosing to download more content. **Notifications are present across user permission levels (as some learners will be by-default super admins of their own device)**. 

The notifications change slightly based on how Kolibri was configured.

### Figma 

Please see [complete figma spec ](https://www.figma.com/file/MpCG7r5PULOCJsEvnTSti8/Android-Designs-2022?node-id=1040%3A20712&t=bVNzyZsQkYnGNOPe-0) for full specs of various states.

Screen Shot 2022-11-18 at 4 51 25 PM
Screen Shot 2022-11-18 at 4 51 31 PM

### Accessibility requirements
Each banner should include a `visually-hidden` `h1` which will allow the screen reader user to more readily discover the notification and it's context. Please check in with @radinamatic and @marcellamaki with any questions and for final wording. For now, you can use placeholder text of "Storage notification".

The `aria-live` with `polite` should also be used, to interrupt the screen-reader user with the notification. For more details, please see the [MDN Docs on aria-live regions](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Live_Regions).

## Acceptance Criteria
Learn interface has notifications for: 
Setup: “On my own” and “Full device”
- [ ]   Does not have device permissions, Nothing in My downloads
- [ ]   Does not have device permissions, Has things in My downloads
- [ ]   Super admin, insufficient storage

Setup: Learn-only device
- [ ] Not enough room for assigned things
- [ ] Removed things in My downloads to make room for assigned things

### Out of scope
Implementing banner or other push-style notifications for browser learn-only devices is not part of this issue.

## References 
- [Notion project page](https://www.notion.so/learningequality/Remote-Content-Syncing-Automatic-syncing-d0a28f150b324d4da019790228117068)
- [Figma spec](https://www.figma.com/file/MpCG7r5PULOCJsEvnTSti8/Android-Designs-2022?node-id=194%3A2922&t=bVNzyZsQkYnGNOPe-0)
- [Gherkin scenarios](https://www.notion.so/learningequality/Gherkin-scenarios-02096ac9925844f8902c818cab607d4b#8d768ab75449478783adb84732968e18)

AllanOXDi:
## Summary
This pull request adds a feature that notifies learners who have run out of storage. The notifications will be present across all user permission levels tailored to the configuration.  See figma spec[ here](https://www.figma.com/file/MpCG7r5PULOCJsEvnTSti8/Android-Designs-2022?node-id=194%3A2922&t=npGaBhD8GfIR9Mbr-0) 

### Screenshots
| Side nav: synced | Side nav: storage | Header |
| --- | --- | --- |
| Screenshot from 2023-03-01 12-29-24 | Screenshot from 2023-03-01 12-46-44 | Screenshot from 2023-03-01 12-52-33 |

## References
Closes #9844 

## Reviewer guidance

Please see[ figma ](https://www.figma.com/file/MpCG7r5PULOCJsEvnTSti8/Android-Designs-2022?node-id=194%3A2922&t=npGaBhD8GfIR9Mbr-0)specs for review guidance 
## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated


kolibri/core/assets/src/views/CorePage/AppBarPage.vue
```diff
@@ -20,6 +20,9 @@
         type="indeterminate"
         :delay="false"
       />
+      <div aria-live="polite">
+        <StorageNotification />
```
marcellamaki: Nice! This looks good from the code side. Let's remember to have @radinamatic do some testing on this once we get the backend hooked up to verify!

kolibri/core/assets/src/views/SideNav.vue
```diff
@@ -51,6 +52,8 @@
               <p :style="{ color: $themeTokens.annotation, fontSize: '12px', marginTop: 0 }">
                 {{ getUserKind }}
               </p>
+
+              <SideNavStorageNotification />
               <!-- display sync status, when relevant -->
               <div v-if="isSubsetOfUsersDevice" data-test="syncStatusInDropdown">
                 <div class="sync-status">
```
marcellamaki, line 59: What you've done here is add a "storage notification" in addition to the existing "sync status" notification. That makes sense from a code perspective, because as we discussed with @bjester and @rtibbles, they are different types of syncs. 

However, in the current designs, we already have a "Device status" header and in the figma, these two types of statuses are displayed under "Device status". They're... conflated, for lack of a better term. 

From the user perspective, having the two different statuses might be confusing - "why are there two different types of syncs?" However, I'm not sure if there might be any edge cases that would warrant it. Perhaps @bjester could weigh in, based on his work on the backend that supports these notifications for now. 

My initial thought is that _in the sidenav_, we add this additional string for the insufficient storage to the existing sync status display. If there is insufficient storage, display that string (no matter the other successes or failures of the facility data syncing. Otherwise, display the existing sync status, whatever it may be. 

What do you think? Blaine, Richard, @jtamiace - any thoughts?

kolibri/core/assets/src/views/StorageNotification.vue
```diff
@@ -0,0 +1,177 @@
...
+        return message;
+      },
+      closeBanner() {
+        this.bannerOpened = false;
+        if (this.previouslyFocusedElement) {
+          this.previouslyFocusedElement.focus();
+        }
+      },
+      manageChannel() {
+        this.$router.push('/');
```
marcellamaki, line 119: Let's remember to follow up on the cross-plugin routing after my android PR is merge! I think some of that can be repurposed here if needed

kolibri/plugins/coach/assets/src/views/plan/CoachExamsPage/index.vue
```diff
@@ -83,7 +83,7 @@
                   v-if="!exam.active && !exam.archive"
                   :text="coachString('openQuizLabel')"
                   appearance="flat-button"
-                  @click="showOpenConfirmationModal = true; modalQuizId = exam.id"
+                  @click="showOpeåppanConfirmationModal = true; modalQuizId = exam.id"
```
marcellamaki: seems like a typo!

marcellamaki (commented):
Great start @AllanOXDi! I added one main question regarding the sync statuses that I'd like a few others to weigh in on. Other than that it's looking good! You can continue to tweak this as needed as the backend pieces develop.

kolibri/core/assets/src/views/SideNav.vue
bjester: Thanks for looping me in, @marcellamaki. The API endpoint that returns the value passed to prop `userSyncStatus` on this component, which is then passed to `SyncStatusDisplay` below, will have a new status string `INSUFFICIENT_STORAGE`. We can only know the device status if it synced successfully, so the error states are mutually exclusive.

kolibri/core/assets/src/views/AppBar.vue
```diff
@@ -88,16 +88,16 @@
...
   import UiToolbar from 'kolibri.coreVue.components.UiToolbar';
   import KIconButton from 'kolibri-design-system/lib/buttons-and-links/KIconButton';
-  import { SyncStatus } from 'kolibri.coreVue.vuex.constants';
+  // import { SyncStatus } from 'kolibri.coreVue.vuex.constants';
   import themeConfig from 'kolibri.themeConfig';
   import responsiveWindowMixin from 'kolibri.coreVue.mixins.responsiveWindowMixin';
   import navComponentsMixin from '../mixins/nav-components';
   import SkipNavigationLink from './SkipNavigationLink';
-  import plugin_data from 'plugin_data';
+  // import plugin_data from 'plugin_data';
```
marcellamaki: You can remove these commented out imports if they are not required anymore

kolibri/core/assets/src/views/AppBar.vue
```diff
@@ -121,19 +121,16 @@
     data() {
       return {
         pointsDisplayed: false,
-        userSyncStatus: null,
         isPolling: false,
         // poll every 10 seconds
-        pollingInterval: 10000,
-        isSubsetOfUsersDevice: plugin_data['isSubsetOfUsersDevice'],
+        // pollingInterval: 10000,
```
marcellamaki: and this :)

kolibri/core/assets/src/views/SideNavStorageNotification.vue
```diff
@@ -0,0 +1,81 @@
...
+      } = useUserSyncStatus();
+      return { queued, lastSynced, status, deviceStatus, deviceStatusSentiment };
+    },
+    props: {},
+    beforeMount() {
+      console.log(this.deviceStatus);
+      console.log(this.lastSynced);
+      console.log(this.status);
+      console.log(this.deviceStatusSentiment);
+      console.log('this is a sidenav notification');
```
marcellamaki: Not sure if these logs are intentional. I know we do have some console logging in places, so maybe some would be useful here, but some of this might be leftover from debugging

kolibri/core/assets/src/views/StorageNotification.vue
```diff
@@ -0,0 +1,195 @@
+<template>
+
+  <div v-if="bannerOpened" class="banner" :style="{ background: $themeTokens.surface }">
+    <div class="banner-inner">
+      <h1 style="display: none">
+        Storage notification
```
marcellamaki: This will need to be a string - even though it is not visible and used for accessibility purposes, we will still want it to be translated. Then, the screen reader can present the header information to the user in the language they are using Kolibri in

kolibri/core/assets/src/views/StorageNotification.vue
```diff
@@ -0,0 +1,195 @@
...
+      },
+    },
+    data() {
+      return {
+        isSubsetOfUsersDevice: plugin_data.isSubsetOfUsersDevice,
+        bannerOpened: false,
+        hasDevicePermissions: false,
+        // isSuperAdmin: false,
+        availableDownloads: false,
+        // learnOnlyDevice: true,
```
marcellamaki: commented code cleanup

kolibri/core/assets/src/views/CorePage/AppBarPage.vue
```diff
@@ -67,6 +80,10 @@
         type: Boolean,
         default: null,
       },
+      // userId: {
+      //   type: String,
+      //   required: false,
+      // },
```
marcellamaki: stray commented code

kolibri/core/assets/src/views/StorageNotification.vue
```diff
@@ -0,0 +1,195 @@
...
+          this.bannerOpened = true;
+        }
+      },
+      focusChange(e) {
+        // We need the element prior to the close button and more info
+        if (
+          (this.$refs.close_button && e.target != this.$refs.close_button.$el) ||
+          (this.$refs.open_button && e.target != this.$refs.open_button.$el)
+        ) {
+          this.previouslyFocusedElement = e.target;
```
marcellamaki, line 132: nice thinking through of the focus here!

marcellamaki (dismissed):
Great work @AllanOXDi! This is looking really good. I've added a few small comments that are just some small polishing up, but very close to being ready I think! 

I imagine there might be a few edge cases that pop up once we do some robust manual QA and have all of the different pieces of syncing and downloading in place -- and that's to be expected. Overall I think this is a very good initial implementation. Nice job working on setting up the composable with @bjester too :)

kolibri/core/assets/src/composables/useUserSyncStatus.js
```diff
@@ -0,0 +1,74 @@
...
+
+export function useUser() {
+  const isUserLoggedIn = computed(() => store.getters.isUserLoggedIn);
+
+  return {
+    isUserLoggedIn,
+  };
+}
+
+export function fetchUserSyncStatus(store, param) {
```
bjester: It does not appear that store should be an argument for this function. What line 38 sends looks like the value for `param`

kolibri/core/assets/src/composables/useUserSyncStatus.js
```diff
@@ -0,0 +1,74 @@
...
+    if (syncData && syncData[0]) {
+      queued.value = syncData[0].queued;
+      lastSynced.value = syncData[0].last_synced;
+      status.value = syncData[0].status;
+      deviceStatus.value = syncData[0].device_status;
+      deviceStatusSentiment.value = syncData[0].device_status_sentiment;
+    }
+  });
+
+  return status;
```
bjester: I don't think this should return `status`. `fetchUserSyncStatus` returns a promise, so the code inside the callback passed to `then` could occur asynchronously and outside the top-down flow of the function, meaning `status.value = syncData[0].status;` doesn't necessarily occur before `return status`.

See my other comment about `useTimeoutPoll`, which would require this to return the promise:
```js
return fetchUserSyncStatus({ user: store.state.core.session.user_id }).then(syncData => {
    if (syncData && syncData[0]) {
      queued.value = syncData[0].queued;
      lastSynced.value = syncData[0].last_synced;
      status.value = syncData[0].status;
      deviceStatus.value = syncData[0].device_status;
      deviceStatusSentiment.value = syncData[0].device_status_sentiment;
    }
  });
```

kolibri/core/assets/src/composables/useUserSyncStatus.js
```diff
@@ -0,0 +1,74 @@
...
+}
+
+export function useUserSyncStatus() {
+  onMounted(() => {
+    usageCount.value++;
+
+    if (usageCount.value === 1) {
+      setTimeout(pollUserSyncStatusTask, 1000);
```
bjester: I recently discovered [`useTimeoutPoll`](https://vueuse.org/core/useTimeoutPoll/) which is perfect for this. It will require that `pollUserSyncStatusTask` returns the Promise from `fetchUserSyncStatus`. You could create a global instance of the `useTimeoutPoll` composable at the same scope as the `ref`s, and then use the returned `resume()` right here inside `onMounted` and `pause()` in place of `clearTimeout`

kolibri/core/assets/src/views/StorageNotification.vue
```diff
@@ -0,0 +1,197 @@
...
+    props: {
+      showBanner: {
+        type: String,
+        required: false,
+        default: null,
+      },
+    },
+    data() {
+      return {
+        isSubsetOfUsersDevice: plugin_data.isSubsetOfUsersDevice,
```
bjester: Is there an existing example you took this from? I haven't see something like this

bjester: Oh nevermind, in the PR deletions there's an example :+1:

kolibri/core/assets/src/views/AppBar.vue
```diff
@@ -121,19 +119,14 @@
     data() {
       return {
         pointsDisplayed: false,
-        userSyncStatus: null,
         isPolling: false,
```
bjester: Looks like you can delete this data var and the sole usage of it in this file.

kolibri/core/assets/src/composables/useUserSyncStatus.js
```diff
@@ -44,23 +44,21 @@ export function pollUserSyncStatusTask() {
       deviceStatusSentiment.value = syncData[0].device_status_sentiment;
     }
   });
-
-  return status;
 }
 
 export function useUserSyncStatus() {
+  const { pause, resume } = useTimeoutPoll(pollUserSyncStatusTask, 1000);
```
bjester: Since the call to `useTimeoutPoll` is inside this function, every time `useUserSyncStatus` is invoked it's going to create a new timer, which will not play nice with the usage tracking.

Here's a situation where this could go wrong:
1) Component A calls `useUserSyncStatus`, creates timer A
2) Since Component A is the first usage of `useUserSyncStatus`, it will call `resume()` for timer A
3) Component B calls `useUserSyncStatus`, creates timer B
4) Since Component B is the second usage, it does not call `resume()` for timer B
5) Component A is destroyed
6) Since there is still one usage by Component B (still mounted), `pause()` for either timer won't be called after Component A is destroyed 
7) Component B is destroyed
8) Since Component B was the last usage of `useUserSyncStatus`, it will call `pause()`... BUT that `pause` function refers to the timer B not timer A. Timer A is the active timer (the one we called `resume()`), and timer B never had `resume()` called, meaning timer A will continue forever.

Since the goal is to de-duplicate the polling if more than one component uses `useUserSyncStatus` and store that polling result for any usage of `useUserSyncStatus`, we only need one timer. That means calling `useTimeoutPoll(pollUserSyncStatusTask, 1000);` should exist outside of the composable function `useUserSyncStatus`. This is what I meant in my previous comment about it being at the same scope as the refs.

kolibri/core/assets/src/composables/useUserSyncStatus.js
```diff
@@ -44,23 +44,21 @@ export function pollUserSyncStatusTask() {
       deviceStatusSentiment.value = syncData[0].device_status_sentiment;
     }
   });
-
-  return status;
```
bjester: Thanks for clearing this up. Although, there was another item I mentioned about making this work well with `useTimeoutPoll`.

Currently, you have set this up to poll every second (`1000` passed to `useTimeoutPoll`). What if the server doesn't respond within 1 second? If the server doesn't respond, then the next request will be started before the previous is completed. It could continue opening requests to the server without the prior being finished. This isn't beneficial because the concurrent requests don't update the frontend anymore than a single request, and they only add load to the server side. Obviously, JavaScript has 2 timer functions `setTimeout` and `setInterval`. The reason the composable is called `useTimeoutPoll` and not `useIntervalPoll` is that it relies on the return value of the function passed to `useTimeoutPoll` in order to determine when it should call the function again. So `useTimeoutPoll` will wait 1000ms after the promise returned by the function is resolved. In this case, the promise is resolved when the request completes. BUT! This function isn't returning that promise, so it wouldn't wait for it to be completed. You need to return that promise here!

bjester:
@AllanOXDi I think the only priority for fixing here before merge for string freeze is [this comment](https://github.com/learningequality/kolibri/pull/10056#discussion_r1116296578) as that looks like it will break things. The other comments, [here](https://github.com/learningequality/kolibri/pull/10056#discussion_r1119072819) and [here](https://github.com/learningequality/kolibri/pull/10056#discussion_r1119088350), are straightforward to also fix but can be followed up in a separate PR.

kolibri/core/assets/src/api-resource.js
```diff
@@ -1057,7 +1057,7 @@ export class Resource {
       'color: red'
     );
     console.log(`Error occured for ${this.name} resource on page ${window.location.href}`);
-    if (store.state.route) {
+    if (store?.state?.route) {
```
rtibbles, line 1060: Nice - I hadn't even seen this had come in in ES2020!

rtibbles (approved):
Still some follow up to do here, so we should create an issue to capture it if existing issues don't cover already, but this is looking in reasonable shape to merge for string freeze with more to come.

kolibri/core/assets/src/api-resource.js
bjester: I didn't know either, but I saw this in @akolson's cookie code! Definitely will be using it

Issue #9850, opened 2022-11-21
Automatic Syncing Updates: Add modal into Coach workflow when toggling resource visibility to on-network learn-only devices


marcellamaki:
## Overview 

This issue is a front-end issue that updates the Coach notifications when assigning and un-assigning resources to learners who are using learn-only devices. The additional modal that is added when changing the state of "visible to learners," learners who are learner-only devices will have their libraries synced automatically, to add or remove the visible resource(s).

### Figma 

 [Complete figma spec ](https://www.figma.com/file/MpCG7r5PULOCJsEvnTSti8/Android-Designs-2022?node-id=194%3A2922&t=yKUs0H8sJ8FrRN6r-0)
<img width="809" alt="Screen Shot 2022-11-21 at 4 27 38 PM" src="https://user-images.githubusercontent.com/17235236/203161561-ddd13c93-dd9b-4a87-8161-f55f562061c7.png">
<img width="550" alt="Screen Shot 2022-11-21 at 4 27 48 PM" src="https://user-images.githubusercontent.com/17235236/203161569-b1612370-d8e0-4b66-9f9b-885a2b9c4daf.png">

### Accessibility requirements
Please us `KModal` for accessibility criteria included out-of-the-box.

## Acceptance Criteria
Coach interface has correct intermediary modal that displays for each scenario in the figma

## References 
- [Notion project page](https://www.notion.so/learningequality/Remote-Content-Syncing-Automatic-syncing-d0a28f150b324d4da019790228117068)
- [Figma spec](https://www.figma.com/file/MpCG7r5PULOCJsEvnTSti8/Android-Designs-2022?node-id=194%3A2922&t=yKUs0H8sJ8FrRN6r-0)
- [Gherkin scenarios](https://www.notion.so/learningequality/Gherkin-scenarios-02096ac9925844f8902c818cab607d4b#45dcdae16db8450b98023b559724277f)

marcellamaki:
Closed with https://github.com/learningequality/kolibri/pull/10452

Issue #9848, opened 2022-11-21
Automatic Syncing Updates: Resource size is displayed for content in Reports and Plan tabs


marcellamaki:
## Overview 

This issue is a full-stack issue that updates the Coach Plugin's Reports and Plan tabs to include information about resource size, both within the resource and in the main table. 

### Figma 

Please see [complete figma spec](https://www.figma.com/file/MpCG7r5PULOCJsEvnTSti8/Android-Designs-2022?node-id=960%3A16822&t=yKUs0H8sJ8FrRN6r-0) for all relevant screens and details. Below is just one example
<img width="735" alt="Screen Shot 2022-11-21 at 10 50 33 AM" src="https://user-images.githubusercontent.com/17235236/203098767-e0620d5b-c765-4ff1-8e0a-d087ac09052e.png">

## Acceptance Criteria
- [ ]  Frontend is updated with resources sizes on both Lesson and Quiz tables, with a total, and by-resource in each row
- [ ] Each resource has a total size available when you navigate into the resource

### Out of Scope
➡️ Accessibility notes: 
* Accessibility remediations for the reports page will be handled in . However, please do not make anything worse 😄 

## References 
- [Notion project page](https://www.notion.so/learningequality/Remote-Content-Syncing-Automatic-syncing-d0a28f150b324d4da019790228117068)
- [Figma spec](https://www.figma.com/file/MpCG7r5PULOCJsEvnTSti8/Android-Designs-2022?node-id=194%3A2922&t=yKUs0H8sJ8FrRN6r-0)
- [Gherkin scenarios](https://www.notion.so/learningequality/Gherkin-scenarios-02096ac9925844f8902c818cab607d4b#45dcdae16db8450b98023b559724277f)

PR #10058, opened 2023-01-31
File sizes

marcellamaki:
## Summary
"Lesson" resources size updates for Automatic syncing work. 

The size of each individual lesson is now available: 
* On the table row in the report page 
* On the table row in the plan page
* On the lesson page (as accessed from the report page, by clicking on the lesson) 
* On the lesson page (as accessed from the plan page, by clicking on the lesson) 

I do still have to add an additional condition so that [the modal shown is only displayed to the coach if there are learn only devices on the network](https://www.figma.com/file/MpCG7r5PULOCJsEvnTSti8/Android-Designs-2022?node-id=194%3A2922&t=hNQaKzOeRq5SF6t9-0).

| Description | Reference |
|---|---|
| Report: Lesson Table | Screenshot 2023-01-31 at 4 40 01 PM |
| Report: Lesson Detail View | Screenshot 2023-01-31 at 4 40 07 PM |
| Plan: Lesson Table | Screenshot 2023-01-31 at 4 40 16 PM |
| Plan: Lesson Detail View | Screenshot 2023-01-31 at 4 40 22 PM |
| Make visible modal | Screenshot 2023-01-31 at 4 40 31 PM |
| Make not visible modal | Screenshot 2023-01-31 at 4 40 38 PM |

## References

Supports the first half of https://github.com/learningequality/kolibri/issues/9850 and https://github.com/learningequality/kolibri/issues/9848

[Figma](https://www.figma.com/file/MpCG7r5PULOCJsEvnTSti8/Android-Designs-2022?node-id=194%3A2922&t=hNQaKzOeRq5SF6t9-0)

## Reviewer guidance

There must be a different fetch method I can use, but I tried `fetchDetailModel` and it didn't work. Or, improvements to the `@action` in general to make it easier to read. Not sure about any _optimization_ concerns either...

----

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated


marcellamaki:
@radinamatic @pcenov - this is ready for initial review. As I mention above, I do still need to ensure the modal _only_ appears when there are learn-only devices on the network (currently, it always appears, unless the user dismisses it). Other than that, this should be ready for first pass of manual QA! 

@rtibbles - I will follow up with you re: all things python and fixing my tests... but would appreciate recommendations for improvements to the `@action` which I suspect could be written more... simply.

kolibri/core/lessons/viewsets.py
```diff
@@ -97,3 +101,26 @@ def annotate_queryset(self, queryset):
         return annotate_array_aggregate(
             queryset, lesson_assignment_collections="lesson_assignments__collection"
         )
+
+    @action(detail=False)
+    def size(self, request, **kwargs):
+        lessons = self.get_queryset()
+        if lessons:
+            lessons_nodes_list = [{lesson.id: lesson.resources} for lesson in lessons]
+            lessons_sizes_dict = {}
```
rtibbles: Should move this outside of the `if` statement - this is not defined unless `lessons` is truthy.

kolibri/core/lessons/viewsets.py
```diff
@@ -97,3 +101,26 @@ def annotate_queryset(self, queryset):
         return annotate_array_aggregate(
             queryset, lesson_assignment_collections="lesson_assignments__collection"
         )
+
+    @action(detail=False)
+    def size(self, request, **kwargs):
+        lessons = self.get_queryset()
+        if lessons:
+            lessons_nodes_list = [{lesson.id: lesson.resources} for lesson in lessons]
```
rtibbles: This initial comprehension is probably unnecessary, as you can just do this in your for loop below.

kolibri/core/lessons/viewsets.py
```diff
@@ -97,3 +101,26 @@ def annotate_queryset(self, queryset):
...
+            lessons_sizes_dict = {}
+
+            for lesson_node_item in lessons_nodes_list:
```
rtibbles: If you don't do the above list comprehension, you can just do:
```
for lesson in lessons:
```

Also if you do this ^ then you don't even need the `if` statement, as iterating over an empty iterator will do no harm!

kolibri/core/lessons/viewsets.py
```diff
@@ -97,3 +101,26 @@ def annotate_queryset(self, queryset):
...
+                current_lesson_id = list(lesson_node_item.keys())[0]
+                current_lesson_resources_list = list(lesson_node_item.values())[0]
+                resource_nodes = []
```
rtibbles: `total_file_size` takes a queryset of ContentNodes, so the simplest thing to do here is to create one from the lesson resources:
```
resources_nodes = ContentNode.objects.filter(id__in=[r["contentnode_id"] for r in lesson.resources])
```

kolibri/core/lessons/viewsets.py
```diff
@@ -97,3 +101,26 @@ def annotate_queryset(self, queryset):
...
+
+                for resource in current_lesson_resources_list:
+                    resource_nodes.append(
+                        ContentNode.objects.filter(id=resource["contentnode_id"])
+                    )
+
+                lessons_sizes_dict[current_lesson_id] = total_file_size(
+                    resource_nodes[0]
```
rtibbles: I think you probably want all the `resource_nodes` here, otherwise you'll just be getting the size of the first one!
```
resource_nodes
```

rtibbles (commented):
Some quick suggestions on the backend implementation.

For tests, I would recommend just copy pasting some of the existing lesson viewset tests and generating the URL for the size endpoint instead!

pcenov:
Hi @marcellamaki, here's a list with some minor issues I was able to identify. Looks absolutely awesome so far:

| Issue| Screenshot |
| Initial state - Some errors in the console and not sure whether the text 'Total size...' should be displayed when there are no resources |  |
| Initial state - All page content disappears if I refresh the page | https://user-images.githubusercontent.com/79847249/217574159-f9eb9de4-7416-4b24-815c-d3dc7f380882.mp4|
| The text 'Total size...' should either have a value of 0 MB when there are no resources visible to learners or hidden altogether |  |
| Reports tab - NaN value is displayed when there are no resources, this is also valid for the modal window | |
| Lesson details - Should the modal be displayed there as well for consistency? |  https://user-images.githubusercontent.com/79847249/217577978-6a85414d-5bb6-4c9a-a906-1840183dcc80.mp4 |
| Reports - The option 'Don't show this message again' is not working |  https://user-images.githubusercontent.com/79847249/217578650-db491ffb-b4e2-4155-a911-3acf91dd1099.mp4 |
| Total size rounding - It could be a bit confusing if the total is different from the total sum of the resources |  |

marcellamaki:
Thank you @pcenov will work on getting these updated, as well as adding the quiz functionality today!

@rtibbles for @pcenov's comment above 
> Total size rounding - It could be a bit confusing if the total is different from the total sum of the resources

Should I do something similar here as we discussed in the "my downloads" page planning and do a sum of the displayed values, to prevent this rounding mismatch?

rtibbles:
Our discussion on the downloads page was more focused on the mismatch between duplicated and deduplicated total sizes. Here this is where the sum of rounded values has produced less than the rounded sum - values like 1.4MB and 27.4MB could produce this, as when summed would produce 28.8MB -> Rounding to 29MB.

One way to mitigate this would be to use 1dp precision for the displayed individual sizes, then keep 0dp precision for the total.

marcellamaki:
@radinamatic @pcenov this is ready for manual QA re-review and the functionality has been added to the quiz tabs as well - in both report, and learn, as well as the quiz report visibility when looking at an individual quiz. Thank you!

pcenov:
Hi @marcellamaki here are the results after retesting on Windows and Ubuntu:

| Issue| Screenshot |
| Lesson/Quiz details - 'NaN B' value is displayed immediately after having created a new empty lesson/quiz. Refreshing the page fixes the issue. |  |
| Initial state - All page content disappears if I refresh the page (this is still valid in the current build and it's not happening on latest develop).  Error in the console: data must be instantiated with some data |  https://user-images.githubusercontent.com/79847249/218494000-b5773920-c068-44d4-98ce-f0c596341ffb.mp4 |
| Coach > Plan > Lesson - The 'New lesson' button is not positioned correctly initially |   |
| Coach > Plan > Lesson - The size value remains not updated immediately after having added resources. Again, gets displayed correctly after reloading the page. This is valid for the quizzes as well and for the modal. |  https://user-images.githubusercontent.com/79847249/218495264-a126f541-993d-4678-939d-84c48cf13519.mp4 |
| Coach > Plan > Quizzes - The text 'Total size of quizzes that are visible to learners: XX MB' is missing/not implemented yet? |  |
| Lesson/Quiz details - There's too much space between the Size label and the XX MB value under it. Is this intentional? |  |

marcellamaki:
Thank you @pcenov!

marcellamaki:
@pcenov this is ready for re-review. thank you!

pcenov:
Hi @marcellamaki there's a blocking issue: When I go to the Coach plugin I am not redirected to the Class Home page instead I see the Classes table and if I click on the class name I am again not being redirected to the Class Home page. Here are the errors in the console:

marcellamaki:
Hmmm, I have not experienced this but will try to replicate!! Thank you @pcenov, and sorry about that!

marcellamaki:
@pcenov - can I ask what asset you are testing with? I have not been able to replicate this locally with either a dev or production build, and now I am trying with the pex and still no luck. I will keep investigating, but just wanted to see if there are any other conditions I should keep in mind. Thank you!

pcenov:
Hmmm, I was testing with the following build asset: https://buildkite.com/learningequality/kolibri-debian/builds/6006#01866b56-7ca3-4014-9d57-7da494a5498a having removed all previous Kolibri data, starting with a clean installation of Kolibri.

marcellamaki:
@rtibbles I have gone through this again for some manual QA and I think the "fixes" I have implemented have resolved the problems that would be blocking for string freeze. However, it does mean that the file size itself is now not actually displayed everywhere 😄 so consistent and properly reactive implementation of that would have to be covered in follow up. I also have to deal with the linting and python test situation which I will do now

kolibri/core/exams/test/test_exam_api.py
```diff
@@ -332,3 +332,13 @@ def test_exam_with_valid_question_sources_succeeds(self):
...
+
+    def test_can_get_quiz_size(self):
+        self.login_as_admin()
+        response = self.client.get(
+            reverse("kolibri:core:exam-size"),
+            {
+                "class_id": self.classroom.id,
+            },
+        )
+        self.assertEqual(response.status_code, 200)
```
rtibbles, line 344: For this and the lesson size tests, we should file a follow up issue to test that the response actually contains the correct file sizes.

kolibri/core/exams/api.py
```diff
@@ -130,3 +134,15 @@ def perform_update(self, serializer):
...
+    def size(self, request, **kwargs):
+        exams = self.get_queryset()
+        exams_sizes_dict = {}
+        for exam in exams:
+            quiz_nodes = ContentNode.objects.filter(
+                id__in={source["exercise_id"] for source in exam.question_sources}
+            )
+            exams_sizes_dict[exam.id] = total_file_size(quiz_nodes)
+
+        return Response([exams_sizes_dict])
```
rtibbles, line 148: Returning the map of ids to file size in a list feels a bit odd. We should either return a list of `{"id": <id>, "file_size": <file_size>}` or make it a detail endpoint that just returns the map.

kolibri/core/lessons/viewsets.py
```diff
@@ -97,3 +101,15 @@ def annotate_queryset(self, queryset):
...
+    def size(self, request, **kwargs):
+        lessons = self.get_queryset()
+        lessons_sizes_dict = {}
+        for lesson in lessons:
+            resource_nodes = ContentNode.objects.filter(
+                id__in=[r["contentnode_id"] for r in lesson.resources]
+            )
+            lessons_sizes_dict[lesson.id] = total_file_size(resource_nodes)
+
+        return Response([lessons_sizes_dict])
```
rtibbles, line 115: See comment above about the expected return of this endpoint.

kolibri/plugins/coach/assets/src/modules/classSummary/index.js
```diff
@@ -4,6 +4,7 @@ import flatten from 'lodash/flatten';
 import find from 'lodash/find';
 import Vue from 'kolibri.lib.vue';
 import ClassSummaryResource from '../../apiResources/classSummary';
+import { ExamResource, LessonResource } from '../../../../../../core/assets/src/api-resources';
```
rtibbles: These should reference `'kolibri.resources'`.

kolibri/plugins/coach/assets/src/modules/classSummary/index.js
```diff
@@ -225,6 +226,14 @@ export default {
         Object.values(state.examLearnerStatusMap).map(learnerMap => Object.values(learnerMap))
       );
     },
+    /*
+     * quizzesSizes := [
+     *   { quiz_id: size in bytes for humans }, ...
+     * ]
+     */
+    quizzesSizes(state) {
```
rtibbles, line 234: I am guessing this is the incomplete vuex state? As it appears to be identical to the getter for exams?

kolibri/plugins/coach/assets/src/modules/classSummary/index.js
```diff
@@ -261,6 +270,14 @@ export default {
         Object.values(getters.lessonLearnerStatusMap).map(learnerMap => Object.values(learnerMap))
       );
     },
+    /*
+     * lessonsSizes := [
+     *   { lesson_id: size in bytes for humans }, ...
+     * ]
+     */
+    lessonsSizes(state) {
```
rtibbles, line 278: Same here, except for lessons.

kolibri/plugins/coach/assets/src/modules/classSummary/index.js
```diff
@@ -418,13 +435,40 @@ export default {
       state.examLearnerStatusMap = { ...state.examLearnerStatusMap };
       state.contentLearnerStatusMap = { ...state.contentLearnerStatusMap };
     },
+    SET_CLASS_LESSONS_SIZES(state, sizes) {
+      state.lessonsSizes = sizes;
```
rtibbles, line 439: You need to add an empty `{}` to the default state for each of these maps, otherwise setting the value here will not produce a reactive update. If you don't want to declare this state in advance in the initial state, you'd have to use `Vue.set` here instead of direct property assignment because Vue2.

rtibbles (approved):
Let's get this in for string freeze, but there's some specific follow up to do here:

Firstly, let's evaluate exactly where the state needs to live, and ensure it's set as part of the initial state in vuex.

We can then also decide whether we:
a) Return a bare map of ids to sizes as the response
b) Return an array of objects with id and file_size
c) Decide we can optimize the file size calculation sufficiently that we just bung it onto the base lesson and quiz viewsets and store it there (or possibly even on each individual resource).

We can chat through possible approaches after string freeze is done!

PR #10073, opened 2023-02-02
Reverts pin to string to accomodate pins starting with 0

akolson:
## Summary

This pull request reverts pin to be stored as string due to the failure to store an integer pin `0000` as such values would be truncated of all zeros and the most significant value retained. This fails the length validation.

## References

Contributes to #9789 

## Reviewer guidance

Please see https://github.com/learningequality/kolibri/pull/10031

----

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated


kolibri/core/auth/test/test_api.py
```diff
@@ -1501,24 +1507,24 @@ def test_facility_admin_can_check_is_pin_valid_unset_pin(self):
...
 
-        response = self.is_pin_valid({"pin_code": 1234})
+        response = self.is_pin_valid({"pin_code": "1234"})
         self.assertEqual(response.data["is_pin_valid"], False)
 
     def test_facility_admin_can_check_is_pin_valid_empty_pin_specified(self):
         self.client.login(username=self.superuser.username, password=DUMMY_PASSWORD)
-        response = self.update_pin({"pin_code": 1234})
-        response = self.is_pin_valid({"pin_code": None})
+        response = self.update_pin({"pin_code": "1234"})
```
nucleogenesis: There are several spots in these tests where you assign `response = self.update_pin(...)` but then never end up using _that_ `response` variable but instead overwrite it shortly after. Just calling `self.update_pin` directly should be fine in any case where you're not asserting against that particular response.

Super minor and not a blocker.

nucleogenesis (approved):
Looks good! :+1:

kolibri/core/auth/test/test_api.py
akolson: Thanks @nucleogenesis  for pointing this out. I have done cleanup on these as they are minor changes

bjester (approved):
Sorry about the regression, looks good

PR #10093, opened 2023-02-06
Update changelog to prepare for 0.15.11 release

marcellamaki:


rtibbles (approved):

radinamatic (approved):

rtibbles:
Something outside of our control seems to have prevented the Python 3.4 and 3.5 test runs from even being created in the matrix. Merging as the change log will not have affected this.

PR #10110, opened 2023-02-13
Implements pin check on learn-only devices

### LINKED ISSUE #9796: Facility Settings Updates: PIN usage on Learn-only Devices

akolson:
## Summary

This PR implements a pin check on learn-only devices to restrict access to the device plugin in kolibri

## References

Closes #9796 

## Reviewer guidance

See https://github.com/learningequality/kolibri/issues/9796 for reviewer guidance
To set up pin, see https://github.com/learningequality/kolibri/issues/9789

----

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/plugins/device/assets/src/views/DeviceIndex.vue
+++ b/kolibri/plugins/device/assets/src/views/DeviceIndex.vue
@@ -41,6 +41,7 @@
   import plugin_data from 'plugin_data';
 
   const welcomeDimissalKey = 'DEVICE_WELCOME_MODAL_DISMISSED';
+  const isPinAuthenticatedKey = 'IS_PIN_AUTHENTICATED';
 
   export default {
     name: 'DeviceIndex',
@@ -53,6 +54,8 @@
     data() {
       return {
         showModal: true,
+        currentFacility: {},
+        isPinAuthenticated: window.sessionStorage.getItem(isPinAuthenticatedKey) === 'true',
       };
     },
     computed: {
@@ -64,16 +67,10 @@
       facilities() {
         return this.$store.state.core.facilities;
       },
-      currentFacility() {
-        return find(this.facilities, { id: this.currentFacilityId }) || {};
-      },
       isPinSet() {
         const dataset = this.currentFacility['dataset'] || {};
         const extraFields = dataset['extra_fields'] || {};
         return extraFields['pin_code'];
-      },
-      isPinAuthenticated() {
-        return this.$store.state.core.session['is_pin_authenticated'];
       },
       userIsAuthorized() {
         if (this.pageName === PageNames.BOOKMARKS) {
@@ -93,12 +90,13 @@
         return this.$route.name;
       },
       requirePinAuthentication() {
-        return this.authenticateWithPin && !this.isPinSet;
+        return this.authenticateWithPin && this.isPinSet;
       },
     },
     watch: {
-      currentFacility(newValue) {
-        const { dataset } = newValue;
+      facilities(newValue) {
+        this.currentFacility = find(newValue, { id: this.currentFacilityId }) || {};
+        const { dataset } = this.currentFacility;
         this.$store.commit('facilityConfig/SET_STATE', {
           facilityDatasetId: dataset.id, //Required for pin authentication
         });
@@ -117,7 +115,8 @@
         return (this.showModal = false);
       },
       submit() {
-        this.$store.commit('CORE_SET_SESSION', { is_pin_authenticated: true });
+        window.sessionStorage.setItem(isPinAuthenticatedKey, true);
+        this.isPinAuthenticated = window.sessionStorage.getItem(isPinAuthenticatedKey) === 'true';
       },
     },
   };
--- a/kolibri/plugins/device/assets/src/views/DeviceIndex.vue
+++ b/kolibri/plugins/device/assets/src/views/DeviceIndex.vue
@@ -9,6 +9,14 @@
     </template>
 
     <transition name="delay-entry">
+      <PinAuthenticationModal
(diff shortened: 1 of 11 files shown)


kolibri/plugins/device/assets/src/views/DeviceIndex.vue
```diff
@@ -8,6 +8,14 @@
       <DeviceTopNav />
     </template>
 
+    <transition name="delay-entry">
```
rtibbles, line 11: I think the designs call for this to be shown when clicking on the Device nav item, prior to navigation to the Device page, so we may need to include this in the DeviceSideNav also, and then persist the authentication state across the page reload with a cookie.

(Although, this still _also_ needs to be here to allow for authentication with the PIN if someone navigates directly to the page).

akolson: Thanks @rtibbles! Yes, you are right about this. I did notice this during visual testing. Thanks for the hint!

marcellamaki:
Hi @akolson - I've read through the code and started doing QA on this. it looks good so far and close to done, although I am running into one error with the pin input, despite the correct values being stored. I will write up my comments first thing tomorrow when I have a more clear mind :)

marcellamaki:
Hi @akolson - my follow up comments with a morning brain :) I'll explain my manual QA testing scenario here: 

1. I've added the pin for the facility, and I have confirmed it is saved. 
2. I updated the plugins files so that subset of users device would return true for the `plugin_data`
3. I sign in as a learner, and update the URL manually to `en/device` (separately, I think ensuring that the `Device` option appears for LOD navigation is part of @nucleogenesis's work on the setup wizard, but just making sure I flag here as well so it's not a case of each thinking the other was resolving it in their work)
4. I have console logged the existing pin, which is available in the frontend, so that's good ✅ 
5. I enter the pin as the learner, and get a 403 error. 

My initial guess is that maybe this is because a learner is accessing the `/device` endpoint plugins? I'm not sure how much of this has to do with other setup configurations (see above re: setup wizard) and how much of this requires updates to this PR... not sure if anyone else has an idea here (@rtibbles?) 

https://user-images.githubusercontent.com/17235236/219419199-f537dec6-76e7-4bf8-b84f-7af4f856d7d3.mp4

akolson:
@marcellamaki  Thanks for this early morning brain comments :). I think this is a permission related issue regarding access to the is_pin_valid() resource. We will need to make changes to the backend to allow the learner to access this resource. I think the issue is related to https://github.com/learningequality/kolibri/pull/10073. I however think, it should be okay to fix this here as it directly ties into the functionality of this PR. @rtibbles ?

rtibbles:
I think the correct resolution for this is slightly different. If I am understanding correctly, @marcellamaki is attempting to access the Device page using a learner account that does not have superadmin permissions. This user should just be told that they do not have permission to access this page, as was previously the case. This is because the PIN feature does not grant any additional permissions to users, but rather restricts access to the Device page for users who have superadmin permissions, but are not facility admins.

So, the resolution here is to only do any of the PIN modal/PIN access checks in the case that the user has SUPERADMIN permissions but is not also a facility admin.

akolson:
Ahh I see, let me make this change. Thanks @rtibbles

akolson:
@rtibbles @marcellamaki  also just to clarify, Is this a check that should only happen on learner only devices or any device?

rtibbles:
Oh right - yes, this is targeted at LODs specifically!

akolson:
@marcellamaki  a validation check has been added to ensure that the pin will only be required for superusers who are not facility admins.

marcellamaki (approved):
Looks good @akolson. I think it's okay to make any further adjustments as we discover them through some robust QA scenarios and bug bashing :) 

Also okay to merge despite mac build being broken!

Issue #10066, opened 2023-02-01
Update UI for more visibility/clarity to support manually adding addresses for syncing


jredrejo:
## Observed behavior

Server discovery to sync does not work in wide networks due to zeroconf (i.e mDNS) limitations

## Expected behavior
In  schools with networks having some switches/routers that split different LAN segments,  users of tablets can not discover the server because mDNS does not cross the segments.

We should add a new Device setting to assign the manual address of the server in the tablets, so they can connect directly to its address to sync when they can not rely on Zeroconf to find the server.

**Updated 4 April 2024:**
Per discussion on Slack, which also referenced [this comment of @pcenov 's in 0.16.0 testing](https://github.com/learningequality/kolibri/issues/11556#issuecomment-1825768457), it would appear that single user syncing on manually-set addresses does in fact "just work" as an extension of other 0.16 updates. If the address is manually entered _somewhere_, then it should get registered for syncing.

However, the expected behavior - easily manually setting the address in the UI - still needs design/UX updates to make this an easy process. UI text to explain this capability, and maybe some explicit functionality that allows you to add additional servers in the future, would be good to include. Conversation between @rtibbles @jtamiace and perhaps @jredrejo to fully understand the desired use case from the user perspective would be helpful to make sure Jessica knows what's needed and has a path forward.

Issue #9598, opened 2022-08-10
Select network address: distinguish between discovered instances and contactable instances


bjester:
## Observed behavior

Kolibri instances on the same network discover each other using a Zeroconf mDNS broadcast. We use this functionality to surface the network addresses of these instances when selecting an address for syncing a facility (one example). In the event the network instance has a typical firewall setup, where incoming packets are dropped on most if not all ports, we may have discovered the instance but would be unable to connect to the port on which Kolibri is running.

If the firewall drops incoming packets, requests will appear to "hang", including [the request](https://github.com/learningequality/kolibri/blob/develop/kolibri/core/discovery/api.py#L49) in our network locations API.

## Errors and logs

Forever loading
![183738670-4a78f005-4f3b-4441-99e6-634e76ea9eb1](https://user-images.githubusercontent.com/819838/183964266-8b7fc159-e775-4b6e-b2ec-29c5e0dc2a07.png)

## Expected behavior

Ideally, we should set reasonable timeouts on requests to network instances to avoid forever loading states in the _Select network address_ modal. Additionally, discovering an instance through Zeroconf, but being unable to complete requests to it, is useful information to surface to the user. This sort of situation would likely be indicative of a firewall active on the network instance.

[Figma for new designs](https://www.figma.com/file/SgTZugaKkES68aipU5dCk8SR/Github-issues-%26-other-fixes?node-id=1721%3A4307)

## User-facing consequences

> Would Kolibri consider improving its user interface [...] — so that in future people don't again waste hours/days every time — fighting this same battle?
https://github.com/learningequality/kolibri/issues/9594#issuecomment-1209768398

## Steps to reproduce

1) Set up 2 instances on a network, both with the same facility
2) On one instance, add a simple firewall that will block access to the port Kolibri is running on from the local network
3) From the instance without a firewall, attempt to sync to the firewalled instance
4) Observe issues with selecting a network address

## Context

Kolibri 0.15
OS: any
Browser: N/A

holta:
Thank you @bjester for writing this up, to help out regular everyday citizens trying to establish Sneakernet-of-Alexandria civic/learning networks — w/o drowning in complexity !!

💯

nucleogenesis:
I think that this might require some new strings to indicate the error state or issue to the user (ie, is the source bad, should they try again, etc) so I'll put this in the pipeline for 0.16

bjester:
[...]

PR #10133, opened 2023-02-21
Connection status for network locations with refactoring of network discovery hooks

### LINKED ISSUE #7374: Emphasize "Network address" to "Device" in copy for facility sync workflows
### Observed behavior

Some background from @jamalex : "Devices" are Kolibri peer servers that have a permanent ID (`instance_id` in Morango and other places) and might have a more transient URL/protocol.

The current UIs emphasize the _address_ as the main entity, instead of the _device_

### Expected behavior

* UI copy for p2p-facility-sync flows emphasizes "Device" over "Address"
* (Bonus) the code implementing this stuff possibly be renamed for DX purposes

### Needs Clarification

* What headers and non-header copy needs to be changed?
* Do we still retain the original "Network address" copy for import/export flows?
* Does "Device" work as a unifying term for both Kolibri Peers + Studio/Content sources so it can be used for both facility-sync + content import/export flows?

### Errors and logs

…

### Steps to reproduce

…

### Context

…

bjester:
## Summary

Overall this supports many new 0.16 features beyond the [original scoped issue](https://github.com/learningequality/kolibri/issues/9598) focused on the 'Select network device' modal, through providing access to connection metadata for network locations 
- Refactors zeroconf listeners to operate through tasks/jobs and to allow 'discovery' of static locations when there's a network change
- Integrates new connection status metadata into network location models with appropriate coordination for keeping it up-to-date through new tasks/jobs
- Adds feature for establishing network discovery hooks within individual Kolibri plugins
- Refactors 'Select network device' modal to consolidate static/dynamic address handling in the frontend
- ~Adds new strings for surfacing connection status to the user in the 'Select network device' modal~
- Fixes undocumented bug where device list wouldn't be filtered when intended to be filtered to those having a specific facility or channel
- Re-enables frontend tests for 'Select network device' modal that were marked as skipped

## References

Supports https://github.com/learningequality/kolibri/issues/10066
Addresses https://github.com/learningequality/kolibri/issues/9598 (potential followups below)
Resolves https://github.com/learningequality/kolibri/issues/7374

### Followup tasks
Followup to integrate 'Connection testing' modal https://github.com/learningequality/kolibri/issues/10153
Other misc followup tasks https://github.com/learningequality/kolibri/issues/10186

## Reviewer guidance

- Open 'Select network device' modal for syncing multiple facilities, with network instances that have and haven't the same facility
- Open 'Select network device' modal for importing channel content, with network instances that have and haven't the same channel
- In any situation, connect and disconnect from networks while 'Select network device' modal is open
- Use new 'Change facility' workflow when provisioned as `on_my_own_setup`
- Add and remove manually added devices
- Verify devices discovered through Zeroconf

### How it works
When you open the 'Select network device' modal:
- it should fetch all dynamically discovered devices on the network and any manually added (static) devices
- it should filter and hide devices that do not have the same facility when the intent is to sync a particular facility
- it should filter and hide devices that do not have the same channel when the intent is to import more of a particular channel
- it should show all devices when the intent is to import a facility or channel not already on the device

While the 'Select network device' modal is open:
- it should disable selection of any device if the device becomes unavailable (disconnects from the network, firewall issue, doesn't respond to requests, etc.)
- it should periodically fetch any new dynamically discovered devices on the network, showing the loading spinner when doing so
- it should periodically test the connection for each, showing the loading spinner when doing so
- it should show the message `Some devices are not responding. ...` if any device becomes unavailable (disabled)

~When the 'Select network device' modal is submitted:~ needs https://github.com/learningequality/kolibri/issues/10153

## Screenshots
| Modal: at open | Modal: generic issue<br/> Modal: at submit | Modal: after submit failure |
| --- | --- | --- |
| Screenshot from 2023-02-23 12-14-24 | Screenshot from 2023-02-23 12-15-01 | Screenshot from 2023-02-23 12-15-20 |

----

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated


kolibri/plugins/user_profile/assets/src/views/ChangeFacility/SelectFacility.vue
```diff
@@ -95,58 +97,90 @@
...
-        addresses: savedAddresses,
-        refreshSavedAddressList,
-        savedAddressesInitiallyFetched,
-      } = useSavedAddresses(props, context);
-      const combinedAddresses = computed(() => {
-        return [...savedAddresses.value, ...discoveredAddresses.value];
-      });
-      const initialFetchingComplete = computed(() => {
-        return savedAddressesInitiallyFetched.value && discoveredAddressesInitiallyFetched.value;
+      const { devices: _devices, isFetching, hasFetched, fetchFailed, forceFetch } = useDevices({
```
jredrejo: `useDevices` returns `devices` empty before doFetch fulfills its Promise, and it's not updated anymore (`device` is not reactive). So the screen does not show any device even if the composable gets its value.

kolibri/plugins/user_profile/assets/src/views/ChangeFacility/SelectFacility.vue
```diff
@@ -95,58 +97,90 @@
...
+        {
+          getKey: device => device.id,
+        }
+      );
+
       // computed properties (functions):
-      const { isMinimumKolibriVersion } = useMinimumKolibriVersion();
+      const availableAddressIds = computed(() => devices.value.filter(d => d.available));
+      const availableFacilities = computedAsync(async () => {
+        const results = await Promise.all(devices.value.map(d => fetchDeviceFacilities(d)));
```
jredrejo: fails when `devices` is `undefined`

kolibri/core/assets/src/views/sync/SelectAddressModalGroup/SelectAddressForm.vue
```diff
@@ -295,63 +323,120 @@
...
+
+        this.uiAlertText = null;
+        this.isSubmitChecking = true;
+        this.doCheck(match.id).then(device => {
+          if (device.available) {
+            this.$emit('submit', device);
+          } else {
+            this.isSubmitChecking = false;
+            switch (device.connection_status) {
+              case ConnectionStatus.Unknown:
```
jredrejo: As `break` does not exist, this will go to the `default` leg of the switch, not sure if this is done on purpose

jredrejo (changes requested):
Thanks for this im pressive piece of code. 
I'll be using some of these composables in my upcoming code .
I can follow most of the code and looks good to me (for some parts of the code I'm not familiar I'm not sure there will be other devs who can say more). 

However, it is not perfect :) For the kolibri/plugins/user_profile/assets/src/views/ChangeFacility/SelectFacility.vue component, I've found some problems that I've commented inline.

kolibri/core/assets/src/views/sync/SelectAddressModalGroup/SelectAddressForm.vue
bjester: It doesn't actually go to the `default` leg, it gets combined with the next switch for `ConnectionStatus.ConnectionFailure`

jredrejo: can `device.connection_status` be `ConnectionStatus.Unknown` & `ConnectionStatus.ConnectionFailure ` at the same time? If not I only see the `default` leg as possible without the `break`. If it can, I don't see the need to query for `ConnectionStatus.Unknown`
In summary, I don't feel comfortable removing the `break` statement inside a `switch`, it removes the readability that switch adds over a series of chained `if-else` and there are usually cleaner options.

kolibri/core/assets/src/api-resources/index.js
```diff
@@ -41,3 +29,8 @@ export { default as AttemptLogResource } from './attemptLog';
 export { default as PingbackNotificationResource } from './pingbackNotification';
 export { default as PingbackNotificationDismissedResource } from './pingbackNotificationDismissed';
 export { default as PortalResource } from './portal';
+export {
+  NetworkLocationResource,
+  StaticNetworkLocationResource,
+  DynamicNetworkLocationResource,
+} from './networkLocation';
```
bjester, line 37: Relocated so I could add methods specifically for these resources.

kolibri/core/assets/src/composables/useMinimumKolibriVersion.js
```diff
@@ -1,34 +1,41 @@
...
-          if (revisionVersion === null || revision >= revisionVersion) return true;
-        }
-      }
-      return false;
-    };
+    }
+    return false;
   });
 
   return {
```
bjester, line 41: Minor changes here to allow setting the version once at invocation of the composable.

kolibri/core/assets/src/mixins/commonSyncElements.js
```diff
@@ -60,12 +60,6 @@ export default {
       // TODO switch to using the last 4 characters
       return this.getCommonSyncString('nameWithIdFragment', { name, id: id.slice(0, 4) });
     },
-    createStaticNetworkLocation({ base_url, device_name }) {
-      return StaticNetworkLocationResource.createModel({
-        base_url,
-        nickname: device_name,
-      }).save();
-    },
```
bjester, line 96: Unused

kolibri/core/assets/src/views/sync/SelectAddressModalGroup/SelectAddressForm.vue
bjester: In this situation, it becomes an OR statement on either `ConnectionStatus.Unknown` or `ConnectionStatus.ConnectionFailure`:
```
> let x = 'a';
undefined
> switch (x) {
... case 'a':
... case 'b':
...   console.log('case a or b')
...   break;
... default:
...   console.log('no case')
... }
case a or b
```
This type of situation is exactly why would tend to use a `switch`, because you can map multiple values to a single value with more clarity than a typical block of multiple `if` statements

kolibri/core/assets/src/views/sync/SelectAddressModalGroup/api.js
```diff
@@ -1,146 +1,60 @@
...
-      return false;
-    });
+export function createDevice(address) {
+  return StaticNetworkLocationResource.createModel({
+    base_url: address.base_url,
+    nickname: address.nickname,
+  }).save();
 }
 
-function versionCompare(version) {
```
bjester: Replaced with @jredrejo's `useMinimumKolibriVersion`

kolibri/core/assets/src/views/sync/SelectAddressModalGroup/api.js
```diff
@@ -1,146 +1,60 @@
...
-            resolve({ ...location, hasContent: isAvailable });
-          });
-        });
-      });
-      return Promise.all(locationsWithAvailabilityPromises);
-    }
-
-    // If facilityId is not provided, then we are at the initial Facility Import workflow
-    // disable any locations unless it is unavailable/offline
-    return locations.map(location => ({ ...location, hasContent: location.available }));
```
bjester: Consolidated this logic out of the API utilities, similar for both channels and facilities, into filtering composable structure

kolibri/core/assets/src/views/sync/SelectAddressModalGroup/useConnectionChecker.js
```diff
@@ -0,0 +1,96 @@
...
+ * @param {Number} interval - time between running individual checks
+ * @param {Number} concurrency - the number of simultaneous connection checks
+ * @return {{
+ *  checkFailed: Ref<bool>, isChecking: Ref<bool>, hasChecked: Ref<bool>,
+ *  doCheck: (function(id: string): Promise<NetworkLocation>),
+ * }}
+ */
+export default function useConnectionChecker(
+  devices,
+  { threshold = 5, interval = 2, concurrency = 3 } = {}
```
bjester: This updates the connection status when keeping it refreshed is important, like the user sitting with the network device modal open.

With these parameters, I tried to balance the situation where there might be many devices on the network. Open to tweaking these

kolibri/core/assets/src/views/sync/SelectAddressModalGroup/useDevices.js
```diff
@@ -0,0 +1,179 @@
...
+  });
+
+  return {
+    // use computed array that depends on availableIds/unavailableIds
+    devices: computed(() => {
+      return get(devices)
+        .filter(d => get(unavailableIds).indexOf(d.id) < 0)
+        .map(d => {
+          // set unavailable if we haven't determined if it should be filtered out yet
+          d.available = get(availableIds).indexOf(d.id) >= 0;
```
bjester: This was a fly-by decision to maintain visibility into the underlying processing, by showing locations as unavailable until we've determined the filtering result from `getIsAvailable`. If it takes a long time to call `getIsAvailable`, without this the user would see an empty list. The downside is that if the device gets filtered out, the device would vanish from the list.

kolibri/core/auth/management/utils.py
```diff
@@ -159,27 +159,27 @@ def is_portal_sync(baseurl):
...
     except NetworkLocationNotFound:
-        raise CommandError("Unable to connect to: {}".format(baseurl))
+        raise CommandError("Unable to connect to: {}".format(address))
 
 
-def get_network_connection(baseurl):
+def get_network_connection(address):
     controller = MorangoProfileController(PROFILE_FACILITY_DATA)
-    network_connection = controller.create_network_connection(get_baseurl(baseurl))
+    network_connection = controller.create_network_connection(get_baseurl(address))
```
bjester, line 182: Overall, using `address` here until it's been resolved to a `base_url`

kolibri/core/content/tasks.py
```diff
@@ -253,14 +253,16 @@ def validate(self, data):
...
+            client = NetworkClient.build_for_address(peer["base_url"])
+            if client.device_info[
+                "application"
+            ] == "kolibri" and not version_matches_range(
+                client.device_info["kolibri_version"], MIN_RESOURCE_IMPORT_VERSION
             ):
                 raise IncompatibleVersionError(
                     "Remote Kolibri instance must be 0.16.0 or higher"
                 )
-            peer["base_url"] = client.baseurl
```
bjester, line 284: This appeared to be a bug. `baseurl` doesn't exist on `NetworkClient`

kolibri/core/device/api.py
```diff
@@ -405,28 +401,6 @@ def retrieve(self, request, pk):
...
-        try:
-            response = requests.get(url)
-            if response.status_code == 200:
-                return Response(response.json())
-            else:
-                return Response({})
-        except Exception as e:
-            raise ValidationError(detail=str(e))
-
-
```
bjester, line 438: Consolidated with existing endpoint in `discovery/api.py`

kolibri/core/discovery/test/test_network_utils.py
```diff
@@ -23,6 +29,21 @@ def test_valid_ipv4_address(self):
...
+                "http://192.168.0.1:8080/",
+                "http://192.168.0.1/",
+                "http://192.168.0.1:8008/",
+                "http://192.168.0.1:8000/",
+                "http://192.168.0.1:5000/",
+                "https://192.168.0.1:8080/",
+                "https://192.168.0.1/",
+            ],
+        )
+
```
bjester, line 46: If the passed in URL already had a port, it would produce two URLs with the same port if the port was one of the standard ports. This test ensures we dedupe the url variations.

kolibri/core/discovery/utils/network/client.py
```diff
@@ -13,84 +18,213 @@
...
+        super(NetworkClient, self).__init__()
+
+        self.base_url = base_url
+        self.timeout = timeout or (DEFAULT_CONNECT_TIMEOUT, DEFAULT_READ_TIMEOUT)
+        self.session = None
+        self.device_info = None
+        self.remote_ip = None
+
+    @classmethod
+    def build_for_address(cls, address, timeout=None):
```
bjester, line 51: Originally, this logic always occurred in the constructor of the `NetworkClient`. By moving it to a factory method, it makes the client a little more useful as a utility

kolibri/core/discovery/utils/network/client.py
```diff
@@ -13,84 +18,213 @@
...
+        logger.info(
+            "Attempting connections to variations of the URL: {}".format(address)
+        )
+        if timeout is None:
+            if get_current_job() is not None:
+                # when we're within a job, then we can use longer timeouts
+                timeout = (DEFAULT_CONNECT_TIMEOUT, DEFAULT_ASYNC_READ_TIMEOUT)
+            else:
+                # if we're within a request thread, then we limit it for an overall time
+                timeout = (DEFAULT_CONNECT_TIMEOUT, DEFAULT_SYNC_READ_TIMEOUT)
```
bjester, line 70: I tried to make the client a bit smart on handling timeouts when it's going to try many URL variations

packages/kolibri-tools/lib/apiSpecExportTools.js
```diff
@@ -103,6 +103,7 @@ const distSpecFilePath = path.resolve(__dirname, '../dist/apiSpec.json');
 try {
   apiSpec = specModule(specFilePath);
 } catch (e) {
+  console.error(e);
```
bjester, line 106: Logging this error helped clarify what was going wrong when confronted with an error about missing `apiSpec.json`

kolibri/core/assets/src/composables/useMinimumKolibriVersion.js
```diff
@@ -1,34 +1,41 @@
...
-      if (major === majorVersion) {
-        if (minor > minorVersion) return true;
-        if (minor === minorVersion) {
-          if (revisionVersion === null || revision >= revisionVersion) return true;
-        }
-      }
-      return false;
-    };
+    }
+    return false;
```
bjester, line 38: Refactored this for re-use with new device composables. Its change in interface didn't affect existing usage.

jredrejo (changes requested):
I have tried to test this after the latest changes and there's no way, either with a previous db or starting it from scratch. I always get this error:
```
ERROR    2023-03-13 20:22:49,780 Error in 'RUN' listener <bound method ZeroConfPlugin.RUN of <kolibri.utils.server.ZeroConfPlugin object at 0x7f719deedc00>>
  File "/datos/le/mio/kolibri/venv/lib/python3.10/site-packages/magicbus/base.py", line 273, in publish
    result = listener(*args, **kwargs)
  File "/datos/le/mio/kolibri/kolibri/utils/server.py", line 302, in RUN
    from kolibri.core.discovery.utils.network.search import NetworkLocationListener
  File "/datos/le/mio/kolibri/kolibri/core/discovery/utils/network/search.py", line 3, in <module>
    from kolibri.core.discovery.tasks import add_dynamic_network_location
  File "/datos/le/mio/kolibri/kolibri/core/discovery/tasks.py", line 222, in <module>
    def add_dynamic_network_location(broadcast_id, instance):
  File "/datos/le/mio/kolibri/kolibri/core/tasks/decorators.py", line 41, in register_task
    return RegisteredTask(
  File "/datos/le/mio/kolibri/kolibri/core/tasks/registry.py", line 218, in __init__
    raise ValueError(
ValueError: High priority tasks must specify a status_fn to inform the user of why it is important
```
Note: I've made a make clean and pip/yarn install and rebuilt everything to avoid conflict with previous code without success

bjester:
@jredrejo yeah, I need to push the fix for that. I had talked with @rtibbles about this new status function constraint for high priority tasks-- they need wired up to the status functions to pass the validation.

jredrejo:
> @jredrejo yeah, I need to push the fix for that. I had talked with @rtibbles about this new status function constraint for high priority tasks-- they need wired up to the status functions to pass the validation.

ok, please, let me know when I can re-test again. I'm very interested in using the composables you wrote here

jredrejo (approved):
Tested and works great in the frontend. I'm looking forward to using it soon.
Thanks!

rtibbles (approved):
Code makes sense on a read through, and all previous comments have been addressed.

If there are any issues here, we're going to find them by using this!

PR #10127, opened 2023-02-17
Move facility settings buttons to the page container for Android

MisRob:
## Summary

In the Android native app, _"Save changes"_ and _"Reset to defaults"_ buttons in the facility settings are now moved to the bottom of the main page container. There are no changes in the browser app where the buttons are in the bottom bar.

| Android - before | Android - after (1/3) | Android - after (2/3) | Android - after (3/3) | Browser (no changes, for reference) |
| --------------------- | -------------------------- | -------------------------- | -------------------------- | ------------------------------------------------ |
| android-before | android-after-large | android-after-medium | android-after-small | browser |

## References

Closes #9794
[Figma designs](https://www.figma.com/file/MpCG7r5PULOCJsEvnTSti8/Android-Designs-2022?node-id=969%3A16669&t=Zpbks4IizxDtKuoJ-0)

## Reviewer guidance

1. Change `isAppContext` getter return value to `true` to simulate the Android environment:

https://github.com/learningequality/kolibri/blob/2b52b5b2af694afd5f8be78c22fc7673d30514f4/kolibri/core/assets/src/state/modules/session.js#L77-L79

2. Go to _"Facility"_ -> _"Settings"_
3. See the buttons

- [ ] Buttons should follow the new designs in the Android native app
- [ ] There should be no changes in the browser app

## Testing checklist

- [x] Contributor has fully tested the PR manually
- [x] If there are any front-end changes, before/after screenshots are included
- [ ] Critical user journeys are covered by Gherkin stories
- [x] Critical and brittle code paths are covered by unit tests

## PR process

- [x] PR has the correct target branch and milestone
- [x] PR has 'needs review' or 'work-in-progress' label
- [x] If PR is ready for review, a reviewer has been added. (Don't use 'Assignees')
- [ ] If this is an important user-facing change, PR or related issue has a 'changelog' label
- [ ] If this includes an internal dependency change, a link to the diff is provided

## Reviewer checklist

- Automated test coverage is satisfactory
- PR is fully functional
- PR has been tested for [accessibility regressions](http://kolibri-dev.readthedocs.io/en/develop/manual_testing.html#accessibility-a11y-testing)
- External dependency files were updated if necessary (`yarn` and `pip`)
- Documentation is updated
- Contributor is in AUTHORS.md

PR #10074, opened 2023-02-02
Pin setup integration

AllanOXDi:
## Summary
This pull request integrates the front-end with the back-end to allow the creation, update and deflation of pins on the facility setting page.

## Reviewer guidance
 Navigate to facility> settings >create pin
## Testing checklist

- [ ] Contributor has fully tested the PR manually
- [ ] If there are any front-end changes, before/after screenshots are included
- [ ] Critical user journeys are covered by Gherkin stories
- [ ] Critical and brittle code paths are covered by unit tests

## PR process

- [ ] PR has the correct target branch and milestone
- [ ] PR has 'needs review' or 'work-in-progress' label
- [ ] If PR is ready for review, a reviewer has been added. (Don't use 'Assignees')
- [ ] If this is an important user-facing change, PR or related issue has a 'changelog' label
- [ ] If this includes an internal dependency change, a link to the diff is provided

## Reviewer checklist

- Automated test coverage is satisfactory
- PR is fully functional
- PR has been tested for [accessibility regressions](http://kolibri-dev.readthedocs.io/en/develop/manual_testing.html#accessibility-a11y-testing)
- External dependency files were updated if necessary (`yarn` and `pip`)
- Documentation is updated
- Contributor is in AUTHORS.md

PR #10135, opened 2023-02-22
Update facility settings

akolson:
## Summary

This PR updates develop with the latest updates from facility settings updates.

## References

https://github.com/learningequality/kolibri/pull/9993
https://github.com/learningequality/kolibri/pull/10031
https://github.com/learningequality/kolibri/pull/10073
https://github.com/learningequality/kolibri/pull/10074
https://github.com/learningequality/kolibri/pull/10125
https://github.com/learningequality/kolibri/pull/10110
https://github.com/learningequality/kolibri/pull/10127 (added on Feb 22 by @MisRob) 

## Reviewer guidance

Please check the [spec](https://www.figma.com/file/MpCG7r5PULOCJsEvnTSti8/Android-Designs-2022?node-id=1473%3A16838) and the PRs above for reviewer guidance.

----

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated


marcellamaki:
@radinamatic and @pcenov - this has all had code review for each PR on the previous "working branch" `update facility settings` and now the full feature should be in place and ready for testing. We will do a follow up code review now with everything in one place, but this would definitely benefit from some robust manual QA. Please do let me, @akolson or @AllanOXDi know if you have any specific questions about testing.

pcenov:
Hi @marcellamaki, to fully test the device management PIN functionality I have to install the build asset from this PR in one VM and then install Kolibri 0.15.11 on another so that I can first setup a Learn-only device and then install the build from this PR on top of it (because setting up a Learn-only device is not working via the 0.16 setup wizard).
However while setting up the Learn-only device on Kolibri 0.15.11 by importing a learner from the full 0.16 device I am getting a "7 of 7: Failed" error:

https://user-images.githubusercontent.com/79847249/220661211-be30d69a-3192-437c-8c11-8ac56fa9316b.mp4

Here are the logs:
[logs.zip](https://github.com/learningequality/kolibri/files/10805049/logs.zip)

Note that I am not able to add a Facility name in Facility settings of the 0.16 device as well and it is empty because of a known issue with the setup wizard which is fixed in latest develop:

https://user-images.githubusercontent.com/79847249/220662319-7c5427e1-cbf1-46fe-a90a-efd0daa13ac9.mp4

I am able to replicate these issues consistently.

kolibri/plugins/facility/assets/src/views/FacilityConfigPage/index.vue
```diff
@@ -247,6 +360,18 @@
       copySettings() {
         this.settingsCopy = Object.assign({}, this.settings);
       },
+      handleCreatePin() {
```
nucleogenesis, line 365: I believe the issue @pcenov noted about not being able to save the facility name is due to the call to `this.saveFacilityName` in `FacilityConfigPage.methods.sendFacilityName` where it calls that method that is not in scope. As that `saveFacilityName`  is set as an action in the `facilityConfig` Vuex module, perhaps we just need to `...mapActions({ saveFacilityName: 'facilityConfig/saveFacilityName'})` in the methods there to provide `this.saveFacilityName` as a function.

akolson: @nucleogenesis the failure to save has been fixed with https://github.com/learningequality/kolibri/pull/10144. Once review is complete and the pr merged, the issue will be fixed on feature branch too! cc @pcenov

marcellamaki: @akolson @pcenov - https://github.com/learningequality/kolibri/pull/10144 reviewed and merged ✅

kolibri/core/assets/src/utils/cookieUtils.js
```diff
@@ -0,0 +1,13 @@
...
+  const expires = `expires=${currentDate.toUTCString()}`;
+  window.document.cookie = `${key}=${value}; ${expires}; path=/`;
+}
+
+export function getCookie(key) {
+  return window.document.cookie
+    .split('; ')
+    .find(cookie => cookie.startsWith(`${key}=`))
+    ?.split('=')[1];
+}
```
bjester: I realize these utilities were added in a separate PR, but since they're becoming global utilities, we should think about using a library for this to cover edge cases. For instance, the cookie value needs to be URI encoded in case it has commas, semicolons or whitespace. There may be other edge cases I'm not recalling

bjester: Sounds like we've used `cookiejs` ([link](https://www.npmjs.com/package/cookiejs)) before, but removed it when we no longer needed it. Looking at its source, it seems to have a few tricks that likely cover edge cases like the one I mentioned.

akolson: Thanks @bjester  for raising this. I will proceed to use the `cookiejs`. I was initially hesitant to introduce a new library into the project as i felt it would be an overkill to a really basic functionality. But since it covers for edge cases better, then we can introduce it back

akolson: @bjester [`js-cookie`](https://www.npmjs.com/package/js-cookie) seems more robust and documentation more clearer. Any thoughts?

akolson: https://github.com/learningequality/kolibri/pull/10164 is tracking the library addition changes

kolibri/core/assets/src/mixins/notificationStrings.js
```diff
@@ -110,6 +110,24 @@ export default createTranslator('NotificationStrings', {
...
+  pinUpdated: {
+    message: 'PIN updated',
+    context: 'A confimation message for updating a pin',
+  },
+  pinRemove: {
+    message: 'PIN removed',
+    context: 'A confirmation message for removing a pin',
+  },
+  pinAuthenticate: {
+    message: 'PIN authenticated succesfully',
```
marcellamaki: small typo, should be "successfully" (not trying to randomly nit pick a few lines of code, I'm just prepping for the final string review 😄)

kolibri/core/assets/src/mixins/notificationStrings.js
```diff
@@ -110,6 +110,24 @@ export default createTranslator('NotificationStrings', {
...
+    context:
+      'A confirmation message indicating that the pin provided for authentication is correct',
```
marcellamaki: I think this should be "the PIN provided"

kolibri/plugins/device/assets/src/views/PinAuthenticationModal.vue
```diff
@@ -0,0 +1,93 @@
...
+        this.$refs.pinFocus.focus();
+      },
+    },
+    $trs: {
+      title: {
+        message: 'Enter PIN',
+        context: 'Title for the pin modal.',
+      },
+      incorrectPin: {
+        message: 'Incorrect pin, please try again',
```
marcellamaki: "PIN" here too

kolibri/plugins/facility/assets/src/views/FacilityConfigPage/ViewPinModal.vue
```diff
@@ -0,0 +1,44 @@
...
+      },
+    },
+    methods: {},
+
+    $trs: {
+      title: {
+        message: 'Device management PIN',
+        context: 'Title for the view pin modal.',
+      },
+      close: {
```
marcellamaki: we can reuse a core string here. I think it's `closeAction`

marcellamaki (commented):
Hi! Adding a few small string comments as I prep for string reviews.

kolibri/core/assets/src/utils/cookieUtils.js
bjester: Nice! Sounds good to me :100:

rtibbles (approved):
Each individual PR in this feature branch has been reviewed, and all comments on this PR have been addressed as a whole.

This will still need follow up QA to give total assurance of the feature before release, but this is good enough to go for now.

pcenov:
Hi @marcellamaki after having created a PIN and having successfully synced my Learn-only device with the main facility I am still not being prompted to enter the PIN on the learn-only device. Tested with [the .deb Buildkite asset](https://buildkite.com/learningequality/kolibri-debian/builds/6154#0186a01e-0e29-4303-b2f5-411cd205682d).

akolson:
Hi @pcenov, Thanks for the feedback. Please confirm that the user logged into the learn only device is s superuser and not a facility admin. This is the only condition under which the pin prompt will be displayed

pcenov:
Hi @akolson the user is a Learner with access to the Device plugin.

akolson:
I am thinking thats why it is not showing. Based on comments in [this pr](https://github.com/learningequality/kolibri/pull/10110), the only condition under which the prompt is displayed is when the user is a superuser and not a facility admin. @rtibbles and @marcellamaki are we missing something?

pcenov:
My guess is that in that case it will be possible to test this on an actual Learn-only device only after the new Setup Wizard is finalized and only when there will be an actual superuser created for each Learn-only device. Currently I am testing by creating a LOD on a 0.15 version of  Kolibri and upgrading with the build asset from this PR in which case the only users available on the LOD are either an Admin or a Learner. :)

nucleogenesis:
As of now in #10139 the first LOD imported / created user will not be a facility admin OR a superuser (technically) -- but they will have `can_manage_content` -- unless I missed something noting the technical change on that in the Slack thread yesterday (cc @rtibbles )

rtibbles:
I think I finally resolved in that Slack thread, that the first LOD imported user should be a superuser (but not a facility admin).

That said, it's possible that we did not consider `can_manage_content` permissions for the PIN, which it should _also_ apply to.

akolson:
Yes @rtibbles this was missed. I will open a pr to add this as well.

akolson:
So the two cases for which the pin prompt will be displayed on a learn-only device are;
1. When a user is a superuser and not a facility admin
2. When a user can manage content

cc @rtibbles @marcellamaki

rtibbles:
Roughly, yes.

If the PIN prompt is enabled for a user's facility, then on a Learn-only device, if that user is not a facility admin, but they do have access to the Device tab (superuser or can manage content permissions), then the prompt will be displayed.

As the Device tab doesn't display for users without superuser or manage content permissions, hopefully that's a bit moot :) so I guess all we actually really need to check is whether they are _not_ a facility admin.

akolson:
Hi @pcenov. A Tweaks were made to also show the PIN prompt for learners who are not admin but have permission to manage the content(`can_manage_content`). Testing artifacts can be found [here in this PR](https://github.com/learningequality/kolibri/pull/10178)

pcenov:
Thanks @akolson - I will be able to fully test the scenario with the PIN prompt only after https://github.com/learningequality/kolibri/pull/10139 is merged in develop.

Issue #9861, opened 2022-11-21
Remote Content Browsing: Create 'My Downloads' Page


marcellamaki:
## Overview 

This issue is a front-end issue to add an "My Downloads" page that will display all content that has been downloaded to the device's Kolibri. 

### Figma 

Please see [complete figma spec ](https://www.figma.com/file/MpCG7r5PULOCJsEvnTSti8/Android-Designs-2022?node-id=194%3A2922&t=yAtcmSpF4FxfLCXa-0) for full specs.

<img width="462" alt="Screenshot 2023-01-31 at 9 10 32 AM" src="https://user-images.githubusercontent.com/17235236/215784144-a73e4122-98e2-4eac-967f-5dec1c07b25c.png">
<img width="508" alt="Screenshot 2023-01-31 at 9 10 48 AM" src="https://user-images.githubusercontent.com/17235236/215784158-45478573-4c7a-4f47-b514-f7b40ce90532.png">
<img width="526" alt="Screenshot 2023-01-31 at 9 11 46 AM" src="https://user-images.githubusercontent.com/17235236/215784189-01a393ae-9544-445a-9dd8-c59e2ad855e2.png">
<img width="575" alt="Screenshot 2023-01-31 at 9 15 19 AM" src="https://user-images.githubusercontent.com/17235236/215784397-95ee245f-844b-4a99-89cc-aca014d82f29.png">
<img width="771" alt="Screenshot 2023-01-31 at 9 17 30 AM" src="https://user-images.githubusercontent.com/17235236/215785383-f8fc0367-4dff-491d-a0d4-c821c4251297.png">
<img width="688" alt="Screenshot 2023-01-31 at 9 17 41 AM" src="https://user-images.githubusercontent.com/17235236/215785398-a5db9cc5-ad1e-42f7-984d-43e4de572c93.png">

## Acceptance Criteria
- [ ]  The "My downloads" page has been added to Kolibri, and is visible in the main navigation menu
- [ ]  A user can filter and sort the results 
- [ ] The downloads page is paginated in increments of 25 results
- [ ] A user can select one, multiple, or all results on the page to remove them from their downloads
- [ ] On "remove", the user is sees a confirmation modal
- [ ] Selection of results is persisted across pages
- [ ] The "Total size of downloads" can just be a sum of the file size data (i.e. we do not have to build around edge cases such as duplicate resources, rounding, etc.) 
[...]

PR #10137, opened 2023-02-23
My downloads page

AlexVelezLl:
## Summary

New my downloads page.

Preview:

https://user-images.githubusercontent.com/51239030/225904378-8642ca49-4658-465e-b089-22919de4d50c.mov

## References

Specs in #9861

## Reviewer guidance

1. Go to menu
2. Click on "My Downloads" menu item

----

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated


kolibri/plugins/my_downloads/assets/src/views/DownloadsList/PaginatedListContainerWithBackend.vue
```diff
@@ -0,0 +1,132 @@
+<template>
```
rtibbles: This file has been copied from https://github.com/learningequality/kolibri/blob/develop/kolibri/plugins/facility/assets/src/views/PaginatedListContainerWithBackend.vue - I think we'd be better off moving that component into `kolibri-common` and then just reusing it here rather than duplicating.

rtibbles (commented):
I think for now, this can live as an additional single page app within the learn plugin (which will mostly just involve moving stuff from where you have it into Learn, and adding a sub-route for the my-downloads SPA to learn).

My reasoning for this is that while this is a separate single page app from Learn, with a separate side nav item, it is entirely dependent on learn, as that's the only place you can initiate downloads from.

kolibri/plugins/my_downloads/assets/src/views/DownloadsList/PaginatedListContainerWithBackend.vue
AlexVelezLl: Perfect, I'll do it!

kolibri/plugins/learn/assets/src/composables/useContentLink.js
```diff
@@ -47,6 +47,29 @@ export default function useContentLink(store) {
     return _makeLink(id, isResource, query);
   }
 
+  function genExternalContentURLBackLinkCurrentPage(id) {
```
AlexVelezLl, line 50: Didnt find a better solution for getting the url of the content renderer to move from my_downloads to learn :/. Due to the fact that they are differents SPA they dont have the same router object

rtibbles: This seems tractable, if a little complex. Another option here would be to create a content preview page within the downloads SPA, but this can work for now, especially as they both live in the same plugin, so cannot be deactivated separately.

It's possible that some of the work that @marcellamaki is doing in her app bar PR could be reused here to simplify this slightly.

marcellamaki: Yes, we may be able to repurpose that here, and once this is merged, I can also do a follow up to change the side navigation item to match the new `CoreMenuOption` API @rtibbles

marcellamaki (approved):
Looks good @AlexVelezLl ! Nice work. 

I think this piece is ready to merge. Can you create a follow up issue that tracks the additional pieces that are remaining to get this working, once the API is ready, and not mocked? (code cleanup of dummy data, API, filtering using the API endpoint for the paginated list containers, etc.) Let me know if you have questions and we can go over it together.

Issue #10126, opened 2023-02-16
SetupWizard Bugs & Follow-up tasks


nucleogenesis:
## Overview

This is going to maintain a task list, notes, and discussion around several bugs. I am putting a lot of this in the same place as the solutions to many of the reported bugs during our last bug bash will likely be fixed together and are likely related to the same root issues.

```[tasklist]
### Tasks
- [x] Handle `apiError` such that it provides a "Start Over" button which will delete the `savedState` value where the interpreted machine is serialized between transitions (there is a path in the state machine -- send it the `START_OVER` event)
[Notion Report: Getting stuck forever, no "start over"](https://www.notion.so/learningequality/Import-facility-flow-completely-fails-after-selection-of-device-to-import-from-if-the-selected-devic-248c03dce87b49c2b09fd2207ca25261?pvs=4) -
[Notion Report: Cannot go back from "Select Facility" page](https://www.notion.so/learningequality/Possible-to-get-into-a-state-in-the-import-facility-work-flow-where-you-cannot-go-back-5ee3bf6124ca4f2382bf55a5ccaf4803?pvs=4) -
- [x] Add a guard against transitioning beyond the credentials form -- likely this error is related to the already-entered username / password not being clear during navigation.
- [x] Import facility "Retry" button should provide better error handling, escape hatch to start over fresh.
- [x] In final steps of Facility Import -- be sure that the user selected at the end is set as the "Superuser" for the device
- [x] Update the SelectFacility page such that it automatically selects the first option and continues without loading the page at all if there is only one possible facility to choose from. Must ensure that "Go Back" then from the next page returns to the page *previous* to the  "Select Facility" page
- [x] The list of users on the Multiple Select is dim (as if disabled?) -- dig into the cause of this and make sure everything is legible and using proper contrasting colors per our design system
- [ x] Move the timeout in the polling function in `ImportIndividualUsers` so that it won't run until the previous poll call is resolved. Investigate that this is the case also in the `LoadingTaskPage`, `MultipleUsers`.
[...]

PR #10139, opened 2023-02-24
Setup Wizard + Learn Only Device Import User

nucleogenesis:
## Summary

Implements new version of Learn Only Device - Import user. I hacked away at the existing machine and flow, taking what fit the new architecture.

There were some issues w/ the API where errors were coming back malformed which I updated so they'd jive with existing code.

The "Multiple Users" import has some additional features not mentioned in the designs:
- It will track what users were imported using credentials such that if you import a single user with their credentials, then go in as an admin to import more, you'll see the previously imported user as "Imported"
- The "Import" button turns to a KCircularLoader while the user is being imported by keeping a list of "being imported users" which are then put onto the "imported users" box when completed (or removed altogether in the case of an error which I'm not exactly sure how to handle / message the user about)
- The "go back" arrow will always take the user to the "LoadingTaskPage" as they may have unfinished tasks that are running. They will get the "Finish" option and the "Import another user" link which will allow them to go back to the auth form.

## Reviewer guidance

**As of PR creation**
`Learn only device -> Import from facility` - This flow should work. You should be able to import users with their username and password - you should be able to import multiple users at once easily by signing in as a super admin. You should be able to go back to import more users and at any point, when hitting continue, should be taken to the "loading task" page - or to the final step.

**Please test the following issues as well:**
`Facility Import` - The error where the user was not being created as a "superuser", resulting in Device not being available, should be resolved.

`Facility -> New Facility` - Please test this flow as there were issues with this during the bug bash that should be resolved

---

_Currently, creating a new user on a remote facility is not implemented, but should be ready for testing first thing next week._

**Code review** will happen in person, but any feedback is greatly appreciated.


kolibri/core/auth/tasks.py
```diff
@@ -792,13 +792,23 @@ def validate(self, data):
...
         validate_and_create_sync_credentials(
             baseurl, facility_id, username, password, user_id=user_id
         )
+        job_data["extra_metadata"]["user_id"] = user_id
+        job_data["extra_metadata"]["username"] = user_info["username"]
+
         job_data["kwargs"]["user"] = user_id
 
+        job_data["kwargs"].update(
+            dict(
```
bjester, line 827: I believe you can drop the `dict` wrapping here. `update` should support either a dictionary or kwargs directly

kolibri/core/device/serializers.py
```diff
@@ -75,7 +77,13 @@ def validate(self, data):
         ):
             data["os_user"] = True
         elif "superuser" not in data:
-            raise serializers.ValidationError("Superuser is required for provisioning")
+            # We don't need a superuser if we're on an SoUD
+            print(data)
+            print(data.get("is_soud"))
```
bjester: :printer: :interrobang:

kolibri/plugins/setup_wizard/assets/src/views/ImportMultipleUsers.vue
```diff
@@ -0,0 +1,208 @@
...
+                }
+              }
+            });
+          }
+        });
+        if (this.isPolling) {
+          setTimeout(() => {
+            this.pollImportTask();
+          }, 2000);
+        }
```
bjester, line 186: You might consider setting the next timeout after previous call has resolved. We have other places like this where we poll but don't wait for the previous call to finish, creating the chance for multiple concurrent requests if the server gets bogged down

kolibri/plugins/setup_wizard/assets/src/views/ImportIndividualUserForm.vue
```diff
@@ -151,7 +153,32 @@
...
+              name: data.device_name,
+              id: data.device_id,
+              baseurl: data.device_address,
+            };
+            this.selectedFacilityId = this.facilities[0].id;
+            this.loadingNewAddress = false;
+          })
+          .catch(error => {
+            // TODO handle disconnected peers error more gracefully
+            this.$store.dispatch('showError', error);
```
bjester, line 203: Looks like `fetchNetworkLocationFacilities` already catches, and incorrectly because it returns an array, but `{facilities: []}` would be expected.

kolibri/core/auth/tasks.py
bjester: Looks like other places in the same file use `dict` so just food for thought, no need to change

pcenov:
Hi @nucleogenesis I have started testing this today and here is some initial feedback.

What's working:

The only flow that is fully functioning is `Group learning > Full device > Create a new learning facility`

What's still not there:

1. `On my own` - One is able to go through the flow but the `Device plugin` is missing

2. `Group learning > Full device > Import all data from an existing learning facility` -  Here I am getting a 400 error at the end:



3. `Group learning > Learn-only device > Import one or more existing user accounts from an existing facility` - I can go through the flows but the Device plugin is also not being displayed after having gone through the Wizard with any of the Learn-only flows. Tried on both Windows and Ubuntu and the Android app:

https://user-images.githubusercontent.com/79847249/221227273-809868e2-2c39-4a7a-a79c-16808ec63bc4.mp4


4. `Group learning > Learn-only device > Import one or more existing user accounts from an existing facility` - Perhaps here the scenario where and Admin is importing several learners should be discussed with @jtamiace as I have no way of verifying the implementation without designs. Here's how this page looks now, with the names of the users being grayed out for some reason:




5. `Group learning > Learn-only device > Create a new user account for an existing facility` - this is a WIP as mentioned by you in the description of the PR and is currently not working

nucleogenesis:
@pcenov I'm very sorry - as I was working on this a bit late I failed to push my fixes for the device page issue you noted above :disappointed: after I tested it successfully -- I've pushed the changes now. I know this takes a long time to test the scenarios, I apologize for the oversight here

kolibri/core/auth/tasks.py
nucleogenesis: This is where, with Python, I don't know *for sure* what type(s) dict.update accepts so I used your logic and went with what I saw that ostensibly worked in others' code :laughing:

kolibri/core/auth/tasks.py
```diff
@@ -241,7 +241,7 @@ def validate(self, data):
             facility_name = facility.name
         else:
             facility_id = facility
-            facility_name = data["facility_name"]
```
rtibbles: This was put here to require the frontend to send the facility name to the backend, so that we can display it properly to the user during progress tracking - I think this must be available on the frontend to send.

kolibri/core/auth/tasks.py
```diff
@@ -809,8 +819,8 @@ def validate(self, data):
     queue=soud_sync_queue,
     permission_classes=[IsSuperAdmin() | NotProvisioned()],
 )
-def peeruserimport(**kwargs):
-    call_command("sync", **kwargs)
+def peeruserimport(command, **kwargs):
```
rtibbles, line 843: Did this raise an error with no arg here?

kolibri/core/device/serializers.py
```diff
@@ -75,7 +78,11 @@ def validate(self, data):
         ):
             data["os_user"] = True
         elif "superuser" not in data:
-            raise serializers.ValidationError("Superuser is required for provisioning")
+            # We don't need a superuser if we're on an SoUD
```
rtibbles: Why not? We still need a superuser, otherwise, there is no one who can administer the device. By default the first imported user should be the superuser.

nucleogenesis:
@pcenov note that the Full Facility Import I found that the device isn't showing (despite it working last night :cry: ) - will address this and let you know when it is ready to test

rtibbles (commented):
A couple of questions and comments.

I'd also like this cleaned up from the previous PR: https://github.com/learningequality/kolibri/pull/9896#discussion_r1049878834 and any uses of it.

kolibri/core/auth/tasks.py
nucleogenesis: Yep

pcenov:
Hi @nucleogenesis,
I had a look at the latest [build assets for Windows and Ubuntu](https://buildkite.com/learningequality/kolibri-python-package/builds/7428) and now the only user flows that are not working are `Learn-only device > Create a new user account for an existing facility` and `Learn-only device > Import one or more existing user accounts from an existing facility`.

Here's a video for `Learn-only device > Import one or more existing user accounts from an existing facility` so that you can see the errors:

https://user-images.githubusercontent.com/79847249/221617215-328d0b63-9905-47cc-8f5a-d659f0bb9c46.mp4



And here's for the case where I am importing just a single learner:

https://user-images.githubusercontent.com/79847249/221617619-83e64042-89e0-4304-b4a2-c906603723a7.mp4

And here's a screenshot for the 'Select a user' screen where the Full name and Username values are in light gray color and hard to read:



Also while importing a Full facility I saw the following error in the logs when I go to the 'Create super admin' screen with seemingly no user facing consequences but I am going to mention it here just in case you know what's going on there:

```
ERROR    2023-02-27 10:16:06,716 Internal Server Error: /api/tasks/tasks/
  File "/usr/lib/python3/dist-packages/kolibri/dist/django/core/handlers/exception.py", line 41, in inner
  File "/usr/lib/python3/dist-packages/kolibri/dist/django/core/handlers/base.py", line 187, in _get_response
  File "/usr/lib/python3/dist-packages/kolibri/dist/django/core/handlers/base.py", line 185, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/usr/lib/python3/dist-packages/kolibri/dist/django/views/decorators/csrf.py", line 58, in wrapped_view
  File "/usr/lib/python3/dist-packages/kolibri/dist/rest_framework/viewsets.py", line 116, in view
    return self.dispatch(request, *args, **kwargs)
  File "/usr/lib/python3/dist-packages/kolibri/dist/rest_framework/views.py", line 495, in dispatch
    response = self.handle_exception(exc)
  File "/usr/lib/python3/dist-packages/kolibri/dist/rest_framework/views.py", line 455, in handle_exception
    self.raise_uncaught_exception(exc)
  File "/usr/lib/python3/dist-packages/kolibri/dist/rest_framework/views.py", line 492, in dispatch
    response = handler(request, *args, **kwargs)
  File "/usr/lib/python3/dist-packages/kolibri/core/tasks/api.py", line 122, in list
    jobs_response.append(self._job_to_response(job))
  File "/usr/lib/python3/dist-packages/kolibri/core/tasks/api.py", line 76, in _job_to_response
    orm_job = job_storage.get_orm_job(job_id=job.job_id)
  File "/usr/lib/python3/dist-packages/kolibri/core/tasks/storage.py", line 316, in get_orm_job
    raise JobNotFound()
kolibri.core.tasks.exceptions.JobNotFound
```

nucleogenesis:
@pcenov thank you - I'm going to be doing everything I can to wrap this up today; will note here again later on (tomorrow morning you're time) when the mentioned issues are ready for testing

nucleogenesis:
@pcenov I was not able to get the "add a new user" issue finished tonight. I'll continue on with the dev on this feature in the morning.

Both "Lod -> Import One" and "Lod -> Import Multiple" should work now and log in the first imported user when completed.

pcenov:
Hi @nucleogenesis,  the learn-only device setup is partially working in the latest build and there are significant differences with the 0.15 version so I am going to post here video comparisons of the main user flows for either importing a single learner or multiple learners.

Import of multiple learners:

Here I see the following issues: after importing 2 learners, the second one is imported as a `Super admin` having access to all plugins while the first one is imported as a learner having access only to the Learn plugin. If I attempt to sign in with the actual Super admin that’s not possible:

https://user-images.githubusercontent.com/79847249/221868126-b2e69b51-9057-47a0-bf9b-834d02c70f75.mp4


Here’s the same user flow on a 0.15 device where you can see that after the import of the learners I am signed in as an `Admin` user with access to all plugins (only the the Channels tab is available in Device) and I can sign in with each of the learners and each of them has access to the Device and Learn plugins.


https://user-images.githubusercontent.com/79847249/221868240-4c86ed8d-e2f8-4ebf-aa3d-55048755f074.mp4



Import of a single user:

Here after importing the learner I don’t see the Full name of the user and the name of the facility, only a green checkmark. After clicking Continue I am signed in as a `Super admin` user with full access to the tabs in the Device plugin and all the other plugins.



https://user-images.githubusercontent.com/79847249/221868323-a2edd004-a85e-4379-a870-3f0ac2cd01d6.mp4


In comparison in 0.15 the user is signed in as a `Learner` with access to the Device plugin and again only the Channels tab is available.



https://user-images.githubusercontent.com/79847249/221868409-3ea46bfe-d3ad-4582-b1d1-0dd46f037761.mp4

nucleogenesis:
@pcenov  @radinamatic 

**FIXES**

1) I have added strings including the "Step X of Y" strings - all messages should be in.
2) All flows should be working properly
3) I have fixed the first user not being made to the superuser when importing multiple for an LOD -- previously I was using the first one who successfully imported. The first user added (whether by the admin multiple-select UI _or_ by entering multiple users' credentials one by one).
4) Learn Only users should only be presented with Learn & Device when they're the superuser; only Learn otherwise
5) Text color on the table of users has been fixed
6) The page to select the facility is skipped when there is only one facility to select.
7) When the Select Facility page is skipped, the "Step X of Y" message counts as if it never existed (ie, Select Facility is always Step 1 -- unless we skip it, then the next page for the given flow is Step 1 and the total number of steps is one fewer as it doesn't count the skipped Select Facility step.
8) If you hit an error and see the "start over" button, it clears the state machine.
9) On the "Import LOD User" flow, you can "Go Back" from the page where you sign in **ONLY** if you haven't yet imported any users; once you import a user you will no longer be able to go back.
10) The "Loading Task Page" will track multiple tasks and only show the "Finish" button when all tasks are complete _(this can only happen if you, as an admin, select to import many users and quickly click the back arrow, which will take you to the loading page)._
11) Loading tasks should show proper task messaging (NOTE: This is different than the Figma but that seems to have been a decision made outside of the scope of _this_ project as that flow/UI is used throughout Kolibri).
12) The Loading Task page will say "Continue" for any flow that goes to the "Personal Data Consent" form; otherwise it will show "Finish" (which then goes to the last step.

**KNOWN ISSUES**
_I am not able to check off items in the checklist on #10126  - so the remaining incomplete items there that aren't mentioned as being fixed above are still outstanding_

- User selection of Full Device vs LOD is not saved when going back to that page

---

I believe I've addressed all previous comments -- my apologies if I've missed anything you've noted. My main focus has been trying to get functionality in place and as the pieces finally came together I was able to put time into many of the cosmetic issues and work on the strings.

Thank you for all of your work testing this, I really hope that this next testing experience is smooth :crossed_fingers:

pcenov:
Hi @nucleogenesis here's a list with the issues I was able to identify today after going through your list above:

| Issue| Screenshot |
| Learn-only device > Create a new user account - Nothing happens when I attempt to create an already existing user, there should be displayed the validation message 'Username already exists' under the Username field  | https://user-images.githubusercontent.com/79847249/222199524-4e6a996b-3d7f-4c40-b3f0-ac2e204f4443.mp4 |
| Learn-only device > Create a new user account - I can create a new user account but the user is created without the Full name value and for some reason the user doesn’t have the option to edit their profile | https://user-images.githubusercontent.com/79847249/222199710-85a35fbd-143d-4cd5-b3bf-9ae5c89ba5a5.mp4 |
| Learn-only device > Create a new user account - The 'Cancel' option is not working and I can create multiple users by clicking the browser's Back button and filling in the form again. This is also valid for the learner import flow |  https://user-images.githubusercontent.com/79847249/222202019-435c6203-fc68-4a1c-baee-9c1c70a6b977.mp4 |
| Learn-only device > Import one or more.. -  I'm getting a 500 error if I attempt to use invalid admin credentials | https://user-images.githubusercontent.com/79847249/222200461-a87009fb-9fb1-4991-8b8e-5ff27630a3ef.mp4 |
| On my own  -  The full name field is supposed to accept up to 120 characters but entering more than 50 is causing an error | https://user-images.githubusercontent.com/79847249/222204048-a95d3be0-c156-4b9f-a36c-ffdc482316b9.mp4 |

I'm not sure what is the final decision on the `Learn-only device > Import one or more..` user flow and perhaps that should be further discussed with the team: Should the Super admin user be imported along with the learners or only the learners should be imported? Why is the `password` value displayed as 'NOT_SPECIFIED' in the debugger? Also why is only one of the imported learners being able to see the Device plugin? Here's a video for that scenario:

https://user-images.githubusercontent.com/79847249/222209801-d0e95cb7-9aa9-45fa-9902-f274f6695be4.mp4

nucleogenesis:
> I'm not sure what is the final decision on the Learn-only device > Import one or more.. user flow and perhaps that should be further discussed with the team: Should the Super admin user be imported along with the learners or only the learners should be imported? Why is the password value displayed as 'NOT_SPECIFIED' in the debugger? Also why is only one of the imported learners being able to see the Device plugin? Here's a video for that scenario:

@rtibbles noted that the first imported learner should be set as the Device superuser for a LoD - perhaps he and @jtamiace can discuss whether or not there may be a better path forward. My suggestion **which would need to be decided very quickly** due to string freeze - is that if we keep it with the first user being set as the superuser, then we should mention that in the copy on the page where the user enters credentials _and_ where a super admin selects multiple users. **Alternatively** we could make all imported learners have the same permissions where they have access to Device content management.

If you use super admin creds to do the multiple import, we don't actually import that admin user at all - we just authenticate with the server that we are an admin so we can get the list of all learners in the facility to present for selection.

When we import users as an admin, we don't get their passwords for each imported user. The user's hashed password is synced, but the browser has no way of knowing each user's password is so I used a code constant `NOT_SPECIFIED` because the call to the API requires that _something_ be in the password field for validation even though, in this particular case, we don't need or use the password at all.

nucleogenesis:
@pcenov @radinamatic 

The issue w/ the import user form being missing is resolved now.

Note -- all non-blocking bugs noted here will become separate issues and adding to our backlog.

Any blocking bugs I will resolve in this PR _or_ ASAP in a separate PR depending on whether this gets merged for strings purposes or not.

There is a mention in #10135 of this PR -- with the import flow now working, should this give you what you need to test those settings as needed?

nucleogenesis:
I've added extant issues/bugs here: https://github.com/learningequality/kolibri/issues/10126 -- the issues are in a checklist.

kolibri/core/auth/tasks.py
rtibbles: > Looks like other places in the same file use dict so just food for thought, no need to change

I think this was my fault, I had also somehow failed to realize that you could pass **kwargs directly to `update` rather than passing in a dict.

kolibri/core/auth/utils/users.py
```diff
@@ -33,13 +34,9 @@ def get_remote_users_info(baseurl, facility_id, username, password):
              'users' containing the list of users of the facility if the user had rights.
     """
     user_info_url = reverse_remote(baseurl, "kolibri:core:publicuser-list")
-    params = {
-        "facility_id": facility_id,
-    }
     try:
         response = requests.get(
             user_info_url,
-            data=params,
```
rtibbles, line 42: Was there a reason to remove this? @jredrejo had been reporting issues with this function being able to properly authenticate against a remote server as well, for the user merging flow, so wondering if there's something similar happening here?

kolibri/core/device/serializers.py
```diff
@@ -146,7 +148,23 @@ def create(self, validated_data):
                         full_name=validated_data["superuser"].get("full_name"),
                     )
             else:
-                superuser = None
+                superuser = FacilityUser.objects.get_or_create_os_user(
```
rtibbles, line 151: Nice - consolidates the provisioning logic in one place.

kolibri/plugins/setup_wizard/api.py
```diff
@@ -40,131 +41,24 @@ class SetupWizardResource(ViewSet):
...
-        the_facility = Facility.get_or_create(facility_name=facility_name)
-        try:
-            superuser = FacilityUser.objects.create_superuser(
-                request.data.get("username"),
-                request.data.get("password"),
-                facility=the_facility,
-                full_name=request.data.get("full_name"),
-            )
-            return Response({"username": superuser.username})
+    def createuseronremote(self, request):
```
rtibbles, line 44: @jredrejo did an interesting thing for this in merge user flow, where this was done during the validation of the task for user merging - so that if this failed for some reason there was instant feedback. Wondering if this might be an approach we could take here too.

kolibri/plugins/setup_wizard/api.py
```diff
@@ -297,30 +151,15 @@ def listfacilitylearners(self, request):
...
         user_info = facility_info["user"]
         roles = user_info["roles"]
         admin_roles = (user_kinds.ADMIN, user_kinds.SUPERUSER)
         if not any(role in roles for role in admin_roles):
             raise PermissionDenied()
         students = [u for u in facility_info["users"] if not u["roles"]]
         return Response({"students": students, "admin": facility_info["user"]})
-
-
-class SetupWizardRestartZeroconf(ViewSet):
```
rtibbles: Are we just using the regular device restart now, or are we triggering this update in another way?

kolibri/core/auth/utils/users.py
nucleogenesis: IIRC it was something around this `PublicFacilityUserViewset.get_queryset()` method that gets the `facility_id` from the query params since it's a `GET` -- in retrospect I think the actual issue is that `requests.get` doesn't take a `data` parameter it takes a `params` parameter (`requests.post` expects `data=` though) so maybe this should just be `params=params`?

rtibbles: Yep, that seems right! https://requests.readthedocs.io/en/latest/user/quickstart/#passing-parameters-in-urls

bjester: seems like this still needs a resolution

kolibri/plugins/setup_wizard/assets/src/views/ImportIndividualUserForm.vue
bjester: This will be resolved with 'Handle `apiError`...'? It's dead code because of https://github.com/learningequality/kolibri/blob/develop/kolibri/core/assets/src/mixins/commonSyncElements.js#L77

bjester (approved):
Code-only review looks fine! The [feedback](https://github.com/learningequality/kolibri/pull/10139/files#r1125157030) about the removal of the `facility_id` param is rather important though and needs resolution, either through fix here or followup

PR #10145, opened 2023-02-27
Remove cross component translator references

### LINKED ISSUE #9037: Remove all instances of the `crossComponentTranslator` before the next string freeze
## Observed behavior

After the string freeze for the `0.15` release, we couldn't modify strings and their namespaces anymore. That also means that we couldn't move strings between components. We have the `crossComponentTranslator` utility which allowed us to access translation strings in a component from another component so that we could temporarily work around this. Before the next string freeze, we should clean up all places where the `crossComponentTranslator` is used, move strings to components from where they are really used or to common string files, and remove all components that are not used but needed to be left in the codebase because they contain some translations.

### Example

Translations strings from the `OverallStatus` component

https://github.com/learningequality/kolibri/blob/2f93cb01fe9de910843bbaf462ae82c5b7a088ea/kolibri/plugins/learn/assets/src/views/AssessmentWrapper/OverallStatus.vue#L16-L23

should be moved to the `LessonMasteryBar` component

https://github.com/learningequality/kolibri/blob/2f93cb01fe9de910843bbaf462ae82c5b7a088ea/kolibri/plugins/learn/assets/src/views/AssessmentWrapper/LessonMasteryBar.vue#L30

and used directly instead of using the `crossComponentTranslator`

https://github.com/learningequality/kolibri/blob/2f93cb01fe9de910843bbaf462ae82c5b7a088ea/kolibri/plugins/learn/assets/src/views/AssessmentWrapper/LessonMasteryBar.vue#L28

https://github.com/learningequality/kolibri/blob/2f93cb01fe9de910843bbaf462ae82c5b7a088ea/kolibri/plugins/learn/assets/src/views/AssessmentWrapper/LessonMasteryBar.vue#L8

`OverallStatus` component should then be removed.

Note that removing a component might not be desired at all places that need to be cleaned up since sometimes we share strings between components that are all used. In such cases, the `crossComponentTranslator` should still be removed and strings can be duplicated in both components, alternatively, they could be moved to shared string files like for example `commonCoreStrings`, `commonLearnStrings`, etc. depending on what location is appropriate.

Search codebase for the `crossComponentTranslator` term to find more.

marcellamaki:
## Summary

- Removes use of cross-component translator
- Deletes unused components and moves the strings when appropriate (several refactored components existed only for the strings)
- Move some strings to common string files, to reduce the number of references and unnecessary multiple translations

## References
Fixes #9037 

## Reviewer guidance
There were a lot of moving parts here. Fresh eyes able to see anything I missed or dropped? 

----

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated


rtibbles (commented):
Code changes seem to make sense to me. There are some failing frontend tests, that suggests maybe something isn't quite working as intended though?

marcellamaki:
Thanks for flagging - I did not check tests for dependent strings... that seems to be the issue. Will update. Hopefully a small fix.

kolibri/plugins/learn/assets/src/views/CurrentlyViewedResourceMetadata.vue
```diff
@@ -256,6 +253,16 @@
         }
       },
     },
+    $tr: {
```
rtibbles: Seems like these strings are redundant, or if they aren't, we should be changing references to them to read from `learnString`?

kolibri/plugins/learn/assets/src/views/CurrentlyViewedResourceMetadata.vue
```diff
@@ -12,7 +12,7 @@
         }"
         data-test="beginners-chip"
       >
-        {{ coreString("ForBeginners") }}
+        {{ coreString(forBeginners) }}
```
rtibbles: This appears to be missing `''`.

kolibri/plugins/learn/assets/src/views/LibraryPage/index.vue
```diff
@@ -318,10 +315,6 @@
         message: 'View as grid',
         context: 'Label for a button used to view resources as a grid.',
       },
-      clearAll: {
```
rtibbles, line 321: I think that we can remove the `results` and `moreThanXResults` strings above as well?

kolibri/plugins/learn/assets/src/views/SidePanelModal/SidePanelResourcesList.vue
```diff
@@ -16,21 +16,6 @@
     name: 'SidePanelResourcesList',
     $trs: {
       /* eslint-disable kolibri/vue-no-unused-translations */
-      noOtherLessonResources: {
```
rtibbles, line 19: According to the comment in the template, we can delete this file entirely, so we should do so!

rtibbles (changes requested):
Still a little more cleanup needed!

rtibbles (approved):
Code changes look correct, and I couldn't find any related errors from a manual check through.

PR #10149, opened 2023-02-27
Return Android specific system value from public device info endpoint.

rtibbles:
## Summary
* Creates a constant to return in place of `platform.system` when we are on an android device
* Returns it from the public device info endpoint


bjester (approved):
looks good

PR #10182, opened 2023-03-03
Task notifications

rtibbles:
## Summary
* Adds capability for tasks to create their own translated notification strings
* Adds requirement for long running and high priority tasks to do so
* Uses kolibri's version of Django translation utils to allow these to be accessed even before Django has been initialized
* Adds long_running flag to all potentially long running tasks (I have deliberately omitted single user syncs - is this correct or a mistake?)
* Adds status functions for all long running and high priority tasks
* Adds them for optional display on some other user relevant tasks
* Adds a dummy update hook function, and uses it in the integration testing app script
* Tweaks how we pick up the update and schedule hooks to make sure they are always set regardless of how the Storage object is initialized

## Reviewer guidance
Does the API for registering the status functions make sense?
Is the expected input and return value of the status functions clear?

When you run `python integration_testing/scripts/run_kolibri_app_mode.py` (having built frontend assets with yarn run build) and do content import actions, do the notifications get logged to the terminal?

----

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/integration_testing/scripts/run_kolibri_app_mode.py
+++ b/integration_testing/scripts/run_kolibri_app_mode.py
@@ -1,8 +1,11 @@
 #!/usr/bin/env python3
+import atexit
 import logging
+import os
 
 from magicbus.plugins import SimplePlugin
 
+from kolibri.main import disable_plugin
 from kolibri.main import enable_plugin
 from kolibri.plugins.app.utils import interface
 from kolibri.utils.cli import initialize
@@ -29,6 +32,10 @@ logging.info("Initializing Kolibri and running any upgrade routines")
 
 # activate app mode
 enable_plugin("kolibri.plugins.app")
+atexit.register(disable_plugin, "kolibri.plugins.app")
+
+# Add a task update hook
+os.environ["KOLIBRI_UPDATE_HOOKS"] = "kolibri.core.tasks.job.log_status"
 
 # we need to initialize Kolibri to allow us to access the app key
 initialize()
--- a/kolibri/core/device/task_notifications.py
+++ b/kolibri/core/device/task_notifications.py
@@ -1,0 +1,14 @@
+"""
+Tiny module to store strings for task notifications for device tasks.
+TODO: This can be migrated into kolibri/core/device/tasks.py in 0.17
+"""
+from kolibri.core.tasks.job import JobStatus
+from kolibri.utils.translation import ugettext as _
+
+
+def status_fn(job):
+    # Translators: A notification title shown to users when Kolibri is looking for other Kolibri devices on the network.
+    searching = _("Searching")
+    # Translators: Notification text shown to users when Kolibri is looking for other Kolibri devices on the network.
+    notification_text = _("Looking for other Kolibri devices")
+    return JobStatus(searching, notification_text)
--- a/kolibri/core/auth/tasks.py
+++ b/kolibri/core/auth/tasks.py
@@ -239,7 +239,6 @@ def exportuserstocsv(facility=None, locale=None):
 
 class SyncJobValidator(JobValidator):
     facility = serializers.PrimaryKeyRelatedField(queryset=Facility.objects.all())
-    facility_name = serializers.CharField(default="")
     command = serializers.ChoiceField(choices=["sync", "resumesync"], default="sync")
     sync_session_id = HexOnlyUUIDField(format="hex", required=False, allow_null=True)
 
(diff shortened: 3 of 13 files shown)


bjester:
>I have deliberately omitted single user syncs - is this correct or a mistake?

I don't think so. Since we have the queuing architecture for performance and the data is a smaller subset anyway, I don't think these need to be classified as long running

kolibri/core/device/task_notifications.py
```diff
@@ -0,0 +1,14 @@
+"""
+Tiny module to store strings for task notifications for device tasks.
+TODO: This can be migrated into kolibri/core/device/tasks.py in 0.17
```
bjester, line 3: So I guess the underlying meaning of this is that we can't move this until 0.17 because we'll lose translations?

integration_testing/scripts/run_kolibri_app_mode.py
```diff
@@ -29,6 +32,10 @@ def RUN(self):
 
 # activate app mode
 enable_plugin("kolibri.plugins.app")
+atexit.register(disable_plugin, "kolibri.plugins.app")
```
bjester, line 35: > and do content import actions

Is this just disabling the web interface? I couldn't login to do the above. Maybe I should try a fresh install that isn't provisioned as SoUD?

Screenshot from 2023-03-03 10-29-04

bjester (approved):
No issues that I can see to get this merged in for string freeze, but I have questions and I think the second item is probably important:
- How do the notifications get removed?
- I see for some tasks, you made it aware of the job state, but for others you didn't. Like for jobs that are enqueued with a delay (single-user syncing) the notification gets posted immediately even though the job hasn't even started:
```
INFO     2023-03-03 10:20:22,833 Requeueing to sync with server http://192.168.7.15:8080/ for user b34733a9617fd05032cfe445e7d6f060 in 60 seconds
INFO     2023-03-03 10:20:22,836 Data syncing in progress
INFO     2023-03-03 10:20:22,837 Do not disconnect your device from the network.
```
- Is it an issue if the status gets fired many many times? (see output below)
```
INFO     2023-03-03 10:20:19,547 Locally preparing data to send
INFO     2023-03-03 10:20:19,548 Data syncing in progress
INFO     2023-03-03 10:20:19,548 Do not disconnect your device from the network.
INFO     2023-03-03 10:20:20,869 Data syncing in progress
INFO     2023-03-03 10:20:20,869 Do not disconnect your device from the network.
INFO     2023-03-03 10:20:20,873 Data syncing in progress
INFO     2023-03-03 10:20:20,873 Do not disconnect your device from the network.
INFO     2023-03-03 10:20:20,875 Data syncing in progress
INFO     2023-03-03 10:20:20,875 Do not disconnect your device from the network.
INFO     2023-03-03 10:20:20,876 Data syncing in progress
INFO     2023-03-03 10:20:20,876 Do not disconnect your device from the network.
INFO     2023-03-03 10:20:21,180 Data syncing in progress
INFO     2023-03-03 10:20:21,180 Do not disconnect your device from the network.
INFO     2023-03-03 10:20:21,181 Data syncing in progress
INFO     2023-03-03 10:20:21,181 Do not disconnect your device from the network.
INFO     2023-03-03 10:20:21,182 Data syncing in progress
INFO     2023-03-03 10:20:21,182 Do not disconnect your device from the network.
INFO     2023-03-03 10:20:21,182 Data syncing in progress
INFO     2023-03-03 10:20:21,183 Do not disconnect your device from the network.
INFO     2023-03-03 10:20:21,183 No records transferred
INFO     2023-03-03 10:20:21,183 Data syncing in progress
INFO     2023-03-03 10:20:21,183 Do not disconnect your device from the network.
INFO     2023-03-03 10:20:22,829 Data syncing in progress
INFO     2023-03-03 10:20:22,829 Do not disconnect your device from the network.
```

rtibbles:
> How do the notifications get removed?

I think for now we'll either autodismiss them when complete or make them dismissable. Probably the former as that is less intrusive. We can then potentially nuance that behaviour in future by adding a flag to the JobStatus class. 

> I see for some tasks, you made it aware of the job state, but for others you didn't. Like for jobs that are enqueued with a delay (single-user syncing) the notification gets posted immediately even though the job hasn't even started:

For the Android context I think some of this will be mitigated as the notifications won't get used until the task has actually started, so it's probably more of a question of how much we want to encode that here vs in the Android app.

> Is it an issue if the status gets fired many many times?

This should be fine because it should just update the existing notification for the task in place.

PR #10187, opened 2023-03-03
Update strings across develop after final string review session

marcellamaki:
This should be cross-referenced with the Notion page and/or Ditto to ensure alignment after final string review took place.

### DIFF
--- a/kolibri/plugins/device/assets/src/views/DeviceSettingsPage/index.vue
+++ b/kolibri/plugins/device/assets/src/views/DeviceSettingsPage/index.vue
@@ -1043,7 +1043,8 @@
         context: 'Label for enabled pages section',
       },
       enabledPagesDescription: {
-        message: 'Unselect a page to hide it even if the user has permission to access it.',
+        message:
+          'When you uncheck a page, it will make it invisible to users even if they have permission to access it.',
         context: "Description for the 'Enabled pages' section.",
       },
       alertDisabledOptions: {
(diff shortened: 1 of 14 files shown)


kolibri/plugins/facility/assets/src/views/FacilityConfigPage/CreateManagementPinModal.vue
```diff
@@ -78,7 +78,7 @@
         context: 'Reminder to sync devices',
       },
       enterPin: {
-        message: 'Enter four numbers to set as your new PIN',
+        message: 'Choose a 4-digit number to set as your new PIN',
```
radinamatic: This is identical as the string in the Change PIN modal. Shouldn't we consolidate and put in the `commonCore`?

marcellamaki: good catch! I'll update

kolibri/plugins/device/assets/src/views/DeviceSettingsPage/index.vue
```diff
@@ -1043,7 +1043,8 @@
         context: 'Label for enabled pages section',
       },
       enabledPagesDescription: {
-        message: 'Unselect a page to hide it even if the user has permission to access it.',
+        message:
+          'When you uncheck a page, it will make it invisible to users even if they have permission to access it.',
```
radinamatic: Change to:
`When you uncheck a page, it will become invisible to users even if they have permission to access it.`

radinamatic (approved):
One last string touch-up and we are good to go! 👏🏽 💯 :shipit:

PR #10192, opened 2023-03-06
"Change Facility" flow string finalization

marcellamaki:
This PR adds or corrects missing strings from the "Change Facility" flow based on the finalized Ditto


radinamatic (approved):
All good, merge away! 👏🏽 💯 :shipit:

PR #10193, opened 2023-03-06
"Content Syncing" Flows string finalization

marcellamaki:
Most of these strings were up to date, since Samson used the update figma. Only one small correction here, which is from me pulling strings from Alex's "My Downloads" PR.


radinamatic (approved):
One more for the road! 🎉 :shipit:

PR #10194, opened 2023-03-06
Add string audit tooling.

rtibbles:
## Summary
* Checks for duplicate message ids whenever we are about to merge two message objects into each other
* Adds CLI command to take a Ditto string export and check that any strings marked as FINAL are represented in the codebase
* Also checks for strings that exist in the codebase that are exact matches for each other and marks them as potential duplicates in the output CSV
* Adds section to our i18n docs describing the new command

### DIFF
--- a/docs/i18n.rst
+++ b/docs/i18n.rst
@@ -411,3 +411,15 @@ It takes a comma separated list of ``intl_code`` language codes. It can also tak
 
 * ``kolibri-supported`` will include all languages listed in ``KOLIBRI_SUPPORTED_LANGUAGES``
 * ``kolibri-all`` will include all languages defined in *language_info.json*
+
+
+Auditing strings
+----------------
+
+Much of our string workflow before developer implementation happens using `Ditto <https://dittowords.com>`__. In order to do a full audit of newly added strings from Ditto, a CSV of the newly added strings can be exported from Ditto, and then our internal audit tool can be run to generate a CSV report of any strings that are potentially missing:
+
+.. code-block:: bash
+
+  yarn run yarn run auditdittostrings --ditto-file <path to Ditto CSV>
+
+This will produce an output CSV file in kolibri/locale/en/LC_MESSAGES/profiles/ditto.csv that contains an audit report on which strings from the Ditto file that are marked as FINAL were not found in the codebase (using an exact match method, so this may produce false positives if strings are not in ICU syntax on Ditto), and also any strings that have been discovered in the codebase to be an exact match - i.e. when we appear to have duplicate strings in our codebase (again, these may be false positives, as some strings may be repeated for different senses).
--- a/docs/i18n.rst
+++ b/docs/i18n.rst
@@ -420,6 +420,6 @@ Much of our string workflow before developer implementation happens using `Ditto
 
 .. code-block:: bash
 
-  yarn run yarn run auditdittostrings --ditto-file <path to Ditto CSV>
+  yarn run auditdittostrings --ditto-file <path to Ditto CSV>
 
 This will produce an output CSV file in kolibri/locale/en/LC_MESSAGES/profiles/ditto.csv that contains an audit report on which strings from the Ditto file that are marked as FINAL were not found in the codebase (using an exact match method, so this may produce false positives if strings are not in ICU syntax on Ditto), and also any strings that have been discovered in the codebase to be an exact match - i.e. when we appear to have duplicate strings in our codebase (again, these may be false positives, as some strings may be repeated for different senses).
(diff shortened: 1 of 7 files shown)


packages/kolibri-tools/lib/cli.js
```diff
@@ -641,6 +641,33 @@ _addPathOptions(program.command('i18n-profile'))
     profileStrings(pathInfo, options.ignore, options.outputFile, options.verbose);
   });
 
+// I18N Ditto Audit
+_addPathOptions(program.command('i18n-audit'))
+  .option(
+    '--output-file <outputFile>',
+    'File path and name to which to write out the audit to',
+    filePath
```
nucleogenesis, line 649: I was having a hard time figuring out why this `filePath` token was here or what it did. I found that it's a function -- is this basically a callback function that takes in the user input as its param?

packages/kolibri-tools/lib/i18n/auditMessages.js
```diff
@@ -0,0 +1,105 @@
...
+  forEachPathInfo(pathInfo, pathData => {
+    const namespace = pathData.namespace;
+    if (!extractedMessages[namespace]) {
+      const filePathMessages = getAllMessagesFromFilePath(pathData.moduleFilePath, ignore, verbose);
+      for (const otherNamespace in extractedMessages) {
+        const nameSpaceMessages = extractedMessages[otherNamespace];
+        if (checkForDuplicateIds(nameSpaceMessages, filePathMessages)) {
+          logging.error(
+            `Duplicate message ids across namespaces ${namespace} and ${otherNamespace}`
+          );
```
nucleogenesis, line 48: > Checks for duplicate message ids whenever we are about to merge two message objects into each other

Is this primarily for us to find where we can consider making `common` a particular message because they share message ids?

packages/kolibri-tools/lib/i18n/utils.js
```diff
@@ -1,9 +1,31 @@
...
+  const potentialDuplicates = intersection(Object.keys(obj1), Object.keys(obj2));
+  const actualDuplicates = [];
+  for (const potentialDuplicate of potentialDuplicates) {
+    const message1 = obj1[potentialDuplicate].message;
+    const message2 = obj2[potentialDuplicate].message;
+    if (message1 !== message1) {
+      logging.error(
+        `${potentialDuplicate} messageId is repeated with different strings '${message1}' and '${message2}'`
+      );
+      actualDuplicates.push(potentialDuplicate);
```
nucleogenesis, line 24: Did you end up seeing errors about generic message ids like `headingLabel` which are going to naturally have different message? Just wondering what the false positive rate here is in practice

docs/i18n.rst
```diff
@@ -411,3 +411,15 @@ It takes a comma separated list of ``intl_code`` language codes. It can also tak
...
+
+
+Auditing strings
+----------------
+
+Much of our string workflow before developer implementation happens using `Ditto <https://dittowords.com>`__. In order to do a full audit of newly added strings from Ditto, a CSV of the newly added strings can be exported from Ditto, and then our internal audit tool can be run to generate a CSV report of any strings that are potentially missing:
+
+.. code-block:: bash
+
+  yarn run yarn run auditdittostrings --ditto-file <path to Ditto CSV>
```
nucleogenesis: `yarn run yarn run` duplicate here

nucleogenesis (changes requested):
Minor typo needs changes.

Just a few questions - I ran it locally and didn't see any error messages about duplicate message IDs so I think maybe I'm not clear what is being checked. Is `message ID` the fully qualified `Component.key` or is it just the `key` part? I was expecting something like 

> `titleLabel` messageID is repeated with different strings

for example when it goes through coach as that is a message ID (`key` without `Component`) we use across multiple components with different messages.

packages/kolibri-tools/lib/i18n/auditMessages.js
rtibbles: No, this is indicative of an error - because we ultimately merge all the messageIds into a single lookup when we pull in from crowdin, so they should be unique across all namespaces.

packages/kolibri-tools/lib/i18n/utils.js
rtibbles: No - it's keyed by the full messageId, so is not flagging those as errors.

docs/i18n.rst
rtibbles: D'oh!

packages/kolibri-tools/lib/cli.js
rtibbles: Yes, it's a validation function - unfortunately commander allows some not very intuitive argument overloading, and if you pass a callback as the third argument, it uses it to validate the argument and returns the value from the function instead.

rtibbles:
>  Is message ID the fully qualified Component.key

It is just this. It's only to ensure we have globally unique full messageIds, to prevent collisions when pulling in the translated messages from crowdin.

The good news was, when I ran this, this was not the case.

nucleogenesis (approved):
Thanks for the clarifications! All looks good and ran successfully :+1:

PR #10196, opened 2023-03-06
Add or update strings for settings flows

marcellamaki:
This PR covers the "Settings" flows page in Ditto. 

It also adds a few missing strings that had inadvertently been merged in during the 'PIN' workflow that had not been added as internationalized strings, but just as text. So, those are added to commonCore and the text has been replaced in the files.


radinamatic (approved):
Good to go! 👍🏽 💯 :shipit:

PR #10205, opened 2023-03-07
Restart zeroconf before provisioning

nucleogenesis:
## Summary
The code that restarted ZeroConf was left unused in the new provisioning call. This ensures that we call to restart ZeroConf just before we provision.

I moved the logic from the JS API Resource method directly into the `SettingUpKolibri` device provisioning.

## References

Tracked in #10126 

## Reviewer guidance
Provision Kolibri a couple times using a couple different flows - keep an eye on the Devtools -> Network tab to see that the `/restart` call comes back `200`.

Is there anything I'm missing here?

### DIFF
--- a/kolibri/core/device/api.py
+++ b/kolibri/core/device/api.py
@@ -1,3 +1,4 @@
+import logging
 from datetime import timedelta
 from sys import version_info
 
@@ -74,6 +75,9 @@ from kolibri.utils.server import STATUS_RUNNING
 from kolibri.utils.system import get_free_space
 from kolibri.utils.time_utils import local_now
 
+logger = logging.getLogger(__name__)
+logger.info("Updating our Kolibri instance on the Zeroconf network now")
+
 
 class DevicePermissionsViewSet(viewsets.ModelViewSet):
     queryset = DevicePermissions.objects.all()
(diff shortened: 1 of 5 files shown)


bjester:
I think we want the restart to be called _after_ provisioning, right? The point is that provisioning updates the device settings (like `subset_of_users_device`) and that needs to update the Zeroconf broadcast, which is done by restarting.

kolibri/plugins/setup_wizard/assets/src/views/onboarding-forms/SettingUpKolibri.vue
```diff
@@ -194,6 +195,9 @@
         return this.wizardService.state.context[key];
       },
       provisionDevice() {
+        // Restart ZeroConf (needs to be done before provisioning)
+        FinishSoUDSyncingResource.postListEndpoint('restart');
```
rtibbles: As this is a non-blocking async call, this is going to cause a weird race condition, and as @bjester pointed out, this should be happening _after_ provisioning has occurred, so that we restart zeroconf when the `subset_of_users_device` flag has been updated.

rtibbles (changes requested):

kolibri/plugins/setup_wizard/assets/src/views/onboarding-forms/SettingUpKolibri.vue
nucleogenesis: Bah I didn't remember off hand when I read your comment, but I did that initially but if I try that API endpoint after provisioning I'd get 403 because those SetupWizard-specific permissions classes rely on the device not yet being provisioned.

So I just did the thing in the view after the serializer validation has passed and we've logged the user in before returning from the device provisioning view.

rtibbles: That also seems fine to me!

kolibri/core/device/api.py
```diff
@@ -97,6 +97,14 @@ def create(self, request, *args, **kwargs):
         if APP_KEY_COOKIE_NAME in request.COOKIES:
             app_key = request.COOKIES[APP_KEY_COOKIE_NAME]
             response_data["app_key"] = app_key
+
+        # Restart zeroconf before moving along
+        import logging
```
rtibbles: Can we move the logging import and logger initialization into the module scope? And maybe we conditionalize this on whether we are provisioning the device as a subset of users device?

rtibbles (approved):

PR #10214, opened 2023-03-08
String updates

marcellamaki:


kolibri/plugins/device/assets/src/views/DeviceSettingsPage/index.vue
```diff
@@ -1002,7 +1002,7 @@
       },
       allowLearnersDownloadDescription: {
         message:
-          'Allow users to mark resources for Kolibri to download when available in their network.',
+          'Allow users to explore resources they don’t have and mark it for Kolibri to automatically download when it’s available in their network.',
```
radinamatic: Wrong apostrophe character here? Did we agree on `'`?

rtibbles: I believe we did!

radinamatic (approved):
Just two apostrophes to turn into `'`, otherwise good to go! 👍🏽

kolibri/plugins/device/assets/src/views/DeviceSettingsPage/index.vue
marcellamaki: I will be doing a find and replace for all. These are coming from any strings that are copy/pasted from ditto and figma. So, we have some in these new strings (I should have been more attentive today as I was doing this) and some existing strings as well.

rtibbles (approved):

PR #10215, opened 2023-03-08
Fix license check

rtibbles:
## Summary
* The license check was running in the bare container, which meant that any system python packages were getting picked up
* Apparently we also weren't installing Python dependencies, so this was failing to actually check Python libraries
* Fixes both these things by adding a virtualenv for the check and then installing the dependencies into them


jredrejo (approved):
LGTM

PR #10216, opened 2023-03-09
Remote Content Browsing:  Clean the library page when displaying other available libraries

AllanOXDi:
## Summary
This PR clean the library page when displaying other available libraries.

## Reviewer guidance

Please see the figma [design](https://www.figma.com/file/MpCG7r5PULOCJsEvnTSti8/Android-Designs-2022?node-id=194%3A2922&t=CrKIgrjUb27SiVJA-0)
## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated


kolibri/plugins/learn/assets/src/views/LibraryPage/UnPinnedDevices.vue
```diff
@@ -53,6 +57,11 @@
         required: false,
         default: 0,
       },
+      allDevices: {
+        type: Object,
+        required: false,
+        default: null,
+      },
```
nucleogenesis: With `default: null` here you can run into an issue where if the prop is not provided at all it can result in a `Error: Cannot read properties of null/undefined` in any code that tries to read into `allDevices`, which they assume is an object.

It doesn't look like you're checking for if it's nullish before you try to read from the object's value (ie, your `v-if` around line 7) -- so if `allDevices` really _isn't_ required, then making the default value `{}` _or_ checking if `allDevices` is null before reading from it (or maybe using the `?.`) would be safer.

kolibri/plugins/learn/assets/src/views/LibraryPage/PinnedNetworkResources.vue
```diff
@@ -30,66 +25,56 @@
...
+            this.pinnedDevices.push(element);
+          });
+        });
+        console.log(this.pinnedDevices);
+      });
+    },
+    methods: {
+      getDeviceIcon(device) {
+        if (device['operating_system'] === 'Darwin') {
+          return 'laptop';
```
nucleogenesis: Since we default at the end to showing a laptop this seems unnecessary -- is there some particular reason we should check first for "Darwin" before "Android" or could we remove the first `if` here to the same effect?

nucleogenesis (changes requested):
Overall code looks good.

I think that making the `allDevices` prop a bit safer or required would help a bit. Note that making it required will just mean that Vue will complain to you in the devtools whenever you don't pass a value to the prop - so it may just be worth doing null checks before reading into the object in any case.

kolibri/plugins/learn/assets/src/views/LibraryPage/PinnedNetworkResources.vue
akolson: +1

kolibri/plugins/learn/assets/src/views/LibraryPage/UnPinnedDevices.vue
```diff
@@ -62,12 +71,26 @@
...
           marginBottom: `${this.windowGutter}px`,
           minHeight: `${this.overallHeight}px`,
+          textAlign: 'center',
         };
       },
     },
+    methods: {
+      getDeviceIcon(device) {
+        if (device['operating_system'] === 'Darwin') {
+          return 'laptop';
```
akolson: Please see comment above from @nucleogenesis  regarding this.

kolibri/plugins/learn/assets/src/views/LibraryPage/index.vue
akolson, line 1: Just a general comment on `PinnedNetworkResources` and `PinnedNetworkResources` components. It is probably better to load the devices alongside there channels in this file and just pass the results as props to the components since they are in the same file. This comes with benefits like better state management, better reloading capabilities and most importantly implementing the DRY(Don't repeat yourself) principal.

kolibri/plugins/learn/assets/src/views/LibraryPage/PinnedNetworkResources.vue
```diff
@@ -30,66 +25,56 @@
...
 
   export default {
     name: 'PinnedNetworkResources',
     components: {
-      CardContent,
+      ChannelCardGroupGrid,
     },
     setup() {
       const { windowBreakpoint } = useKResponsiveWindow();
+      const { baseurl, fetchDevices } = useDevices();
```
akolson, line 40: Please see my general comment on the `LibraryPage/index.,vue` file

kolibri/plugins/learn/assets/src/views/LibraryPage/MoreNetworkDevices.vue
akolson, line 1: Please see my general comment on the `LibraryPage/index.,vue` file

akolson (changes requested):
Hi @AllanOXDi, Great work. Code generally LGTM. I have left some comments that will help improve the code even further.

kolibri/plugins/learn/assets/src/views/LibraryPage/index.vue
```diff
@@ -49,63 +47,59 @@
...
+          :pinnedDevices="pinnedDevices"
+        />
         <!-- More  -->
-        <div v-if="!deviceId">
-          <h2>
-            {{ $tr('moreLibraries') }}
-          </h2>
+        <div>
+          <h2>{{ $tr('moreLibraries') }}</h2>
           <MoreNetworkDevices />
```
akolson: I think applying the same concept as seen with `PinnedNetworkResources` would work great. Creating an `unpinnedDevices`  `data` property and passing it into the `PinnedNetworkResources` as a prop would help with refreshing state of these devices(I think there is a refresh button somewhere). That would mean also getting rid of the `fetchDevices` and `fetchChannels` in the `MoreNetworkDevices` component as it would not be necessary since you are passing the content(devices and their channels) as `props`. For now its okay for them to display the same devices. Later we can apply filtering based on results returned from the pinning API that we co-hacked with Jacob yesterday

akolson (changes requested):
Overall code looks good. Thanks @AllanOXDi. Just a final change and we should be good for merging

akolson (approved):

kolibri/plugins/learn/assets/src/views/LibraryPage/UnPinnedDevices.vue
```diff
@@ -62,12 +77,26 @@
...
+      getDeviceIcon() {
+        if (this.operatingSystem === 'Android') {
+          return 'device';
+        } else if (!this.allDevices.subset_of_users_device) {
+          return 'cloud';
+        } else {
+          return 'laptop';
+        }
+      },
+    },
```
marcellamaki, line 96: I had to add another `}` here when running locally.

marcellamaki:
Hi @AllanOXDi - I was doing a little bit of manual QA on this and I noticed that the channel cards that are displayed here are my channels, not the channels of the remote devices. 

Maybe this PR is just for the card implementation, but I just wanted to mention this because I am not sure if it was intentional, and in case we need a follow up issue for it. It looks like the Explore Libraries page that @akolson was working on has this in place, so maybe it's just a matter of a follow up issue using the pattern there. 
https://user-images.githubusercontent.com/17235236/228255591-794f7b4a-0471-46d3-b614-14a3c436e7f2.mp4

Nice work overall! Great to see this Library Page coming along so well :)

akolson:
@marcellamaki thanks for raising this. Based on the design spec, i think we shouldn't be having this
image @AllanOXDi ?

kolibri/plugins/learn/assets/src/views/LibraryPage/index.vue
```diff
@@ -300,28 +295,12 @@
...
-          if (this.deviceId === KolibriStudioId) {
-            return this.learnString('kolibriLibrary');
-          } else {
-            return this.$tr('libraryOf', { device: this.deviceName });
-          }
-        } else {
-          return this.coreString('channelsLabel');
-        }
-      },
+      // ...mapGetters(['isUserLoggedIn']),
```
akolson, line 303: This should be removed completely

kolibri/plugins/learn/assets/src/views/LibraryPage/index.vue
```diff
@@ -351,11 +330,24 @@
...
     created() {
       this.search();
-      if (this.isUserLoggedIn) {
-        this.fetchDevices().then(devices => {
-          this.devices = devices.filter(d => d.available);
+      // if (this.isUserLoggedIn) {
+      //   this.fetchDevices().then(devices => {
+      //     this.devices = devices.filter(d => d.available);
+      //   });
+      // }
```
akolson, line 337: All commented code should be removed completely

kolibri/plugins/learn/assets/src/views/LibraryPage/index.vue
```diff
@@ -386,6 +374,10 @@
         message: 'Showing all available libraries around you.',
         context: 'Connection state when the device is connected and shows other libraries',
       },
+      refresh: {
+        message: 'Refresh',
+        context: 'Link for refreshing library',
+      },
```
akolson, line 380: I think adding/moving around translations strings has been suspended for now due to the string freeze. cc @marcellamaki  @radinamatic

AllanOXDi: I think these strings were already in. Might have come due to a rebase yesterday

kolibri/plugins/learn/assets/src/views/ChannelCard.vue
```diff
@@ -113,6 +117,11 @@
         required: false,
         default: null,
       },
+      isPinned: {
```
akolson, line 120: I think this prop is misleading. I think the wifi icon is meant to specify whether the resource is available or not. @jtamiace , @marcellamaki ?

kolibri/plugins/learn/assets/src/views/ChannelCard.vue
```diff
@@ -48,14 +48,18 @@
...
       v-if="version"
       class="version-wrapper"
       :style="versionStyle"
     >
       <p>{{ $tr('version', { version: version }) }}</p>
     </div>
+    <div
+      style="text-align:right;margin-right:10px;margin-top:10px"
+    >
+      <KIcon v-if="isPinned" icon="wifi" />
```
akolson, line 61: I think the `v-if="isPinned"` directive should be moved to the parent(`div`) as hiding just the `KIcon` will leave additional margins that may not be desirable. Also, see my comments below regarding the `isPinned` prop

kolibri/plugins/learn/assets/src/views/LibraryPage/CardContent.vue
akolson, line 1: This file should be deleted as its no longer used in light of the [discussion here](https://learningequality.slack.com/archives/CB37UM23A/p1680033871832199)

kolibri/plugins/learn/assets/src/views/LibraryPage/index.vue
akolson: I just checked and the refresh string already exists in the `commonCoreStrings.js`. So using this instead and deleting the new one would be an acceptable solution to this.

https://github.com/learningequality/kolibri/blob/develop/kolibri/core/assets/src/mixins/commonCoreStrings.js#L1524

akolson (changes requested):
Thanks @AllanOXDi! Code looks good. Just a few minor changes and we should be good to go!

AllanOXDi:
Closing this PR again. It's giving me a nightmare updating it.Probably because I deleted a base branch. Please see #10354. Sorry!

Issue #9058, opened 2022-01-28
Some resources cards are not optimized for very small screens


MisRob:
## Observed behavior

For very small screens, there are several problems on resource cards on the Library and Bookmarks pages:

| | | |
| --- | --- | --- |
| Library | ![library-card](https://user-images.githubusercontent.com/13509191/151541151-2565efca-b3ab-4918-813c-b9976eb355bc.png) | - unnecessary white space on the left side of the thumbnail<br>- the thumbnail overlaps with the learning activity label  |
| Bookmarks | ![bookmark-card](https://user-images.githubusercontent.com/13509191/151541323-598cfc7d-920f-4ff3-9a7e-9651de345c68.png) |-unnecessary vertical white space below the thumbnail<br>-the progress bar at the bottom overlaps with the info and close controls |

## Expected behavior

- no unnecessary white space
- no overlapping elements
- learning activity labels are readable

## User-facing consequences

- bad readability
- it's not possible to recognize learning activity in some cases
- more scrolling on mobile devices due to unnecessary vertical white space

## Steps to reproduce

Preview cards on the Bookmarks and Library pages in Learn for a small screen width (tested around 320px)

## Context

- Kolibri version: Kolibri 0.15.1.dev0+git.20220127215827
- Operating system: Ubuntu 20.04.3 LTS
- Browser: Chrome Version 97.0.4692.99 (Official Build) (64-bit)

MisRob:
Many of these issues could be solved by more robust CSS allowing for auto-wrapping on very small screens. This would automatically prevent issues similar to
![Screenshot from 2022-01-28 13-03-10](https://user-images.githubusercontent.com/13509191/151543896-d64a3c45-2dc2-4fc0-b470-6516140214ac.png)
without much extra effort needed.

It may be helpful to merge and clean up our many card components introduced during hybrid learning work first since it's possible that some of them will become obsolete (related #8533)

iprasandeep:
Hey, @MisRob I would like to work on this issue. thanks :)

MisRob:
Hi @prsndp, thank you, you're welcome to take this on

MisRob:
@prsndp Please assign yourself as soon as you start working on it (even though I'm not going to work on it, I'm assigning myself now too only so we know that it's been taken already so that we don't assign it to someone else accidentally during our work planning)

iprasandeep:
[...]

PR #10217, opened 2023-03-09
Removing void spaces and adding media query to the bookmark cards

shrinishant:
## Summary

I removed the min-height css property of mobile-card and card in CardList.vue so that the extra void space could be removed between resource description and bookmarked status.

Also added media query to reduce further void spaces on screens with width less than 720px.

I manually verified the results in the inspection mode for upto 320 px width screens and also for iPhone SE device. I also ran yarn run test to test for the frontend before creating this PR.

Before :-

Screenshot from 2023-03-04 11-37-55

After :-
Screenshot from 2023-03-09 18-04-34

## References

#9058 

## Reviewer guidance

…

----

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/plugins/learn/assets/src/views/CardList.vue
+++ b/kolibri/plugins/learn/assets/src/views/CardList.vue
@@ -246,7 +246,6 @@
 
   .card {
     position: relative;
-    min-height: 246px;
     vertical-align: top;
     border-radius: 8px;
     transition: box-shadow $core-time ease;
@@ -261,10 +260,6 @@
     display: block;
     width: 100%;
     text-decoration: none;
-  }
-
-  .mobile-card {
-    min-height: 400px;
   }
 
   .metadata-info {
@@ -320,5 +315,11 @@
   .text-area {
     margin-bottom: $footer-height;
   }
+
+  @media only screen and (max-width: 720px) {
+    .text-area {
+      margin-bottom: 0;
+    }
+}
 
 </style>
--- a/kolibri/plugins/learn/assets/src/views/CardList.vue
+++ b/kolibri/plugins/learn/assets/src/views/CardList.vue
@@ -2,7 +2,6 @@
 
   <div
     class="card container drop-shadow"
-    :class="{ 'mobile-card': isMobile }"
     :style="{ backgroundColor: $themeTokens.surface }"
   >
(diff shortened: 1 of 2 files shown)


kolibri/plugins/learn/assets/src/views/CardList.vue
```diff
@@ -321,4 +316,10 @@
     margin-bottom: $footer-height;
   }
 
+  @media only screen and (max-width: 720px) {
```
MisRob: [Blocking] This indeed reduces some space. However, looking at our designs, it seems that we want to keep some space between the text area and the progress bar. We need to only address unwanted or unexpected spaces. Therefore this needs to be reverted. Thank you for your effort.

kolibri/plugins/learn/assets/src/views/CardList.vue
```diff
@@ -263,10 +262,6 @@
     text-decoration: none;
   }
 
-  .mobile-card {
```
MisRob, line 266: [Nice to have] Since you removed `.mobile-card` class styles, it'd be good if you could clean up its assignments from the component markup, for example

```
:class="{ 'mobile-card': isMobile }"
```
as this code is now obsolete.

MisRob (changes requested):
Thank you, @shrinishant. Removing `min-height` seems to be fine to me in this context since `CardList` is only used on the Bookmarks page where we never show cards in a grid.

There is one issue that needs to be resolved before merging, please see below.

MisRob:
@shrinishant One more thing - the linter check will need to be fixed too

shrinishant:
> @shrinishant One more thing - the linter check will need to be fixed too

@MisRob sir, I've made the required changes and commited it but I couldn't check or verify for the linter check although I did run the "yarn run lint-frontend".

MisRob:
Thank you, @shrinishant. Everything's looking good to me. I started new checks run.

Before we merge this, you're welcome to add yourself to the list of contributors in the _AUTHORS.md_ file. I will let this PR open for a while, but if it gets merged before you manage to do so, do not hesitate to open a follow-up PR.

MisRob (approved):

PR #10218, opened 2023-03-09
 Manage and edit sync schedule clean up

AllanOXDi:
## Summary

…

## References

…

## Reviewer guidance

…

----

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated


packages/kolibri-common/components/SyncSchedule/EditDeviceSyncSchedule.vue
```diff
@@ -20,7 +20,7 @@
           <KGrid>
             <KGridItem>
               <KSelect
-                v-model="selectedItem"
```
rtibbles: Hrm, you probably want to keep these as `v-model` not change them to `:value` otherwise, when the value changes in the KSelect, this will not be updated in data.

packages/kolibri-common/components/SyncSchedule/EditDeviceSyncSchedule.vue
```diff
@@ -173,11 +173,13 @@
...
         now: now(),
-        selectedItem: '',
+        selectedItem: {},
         tasks: [],
-        selectedDay: null,
-        selectedTime: null,
+        selectedDay: {},
+        selectedTime: {},
         removeBtn: false,
+        serverTime: now(),
```
rtibbles: Probably good to add a `setTimeout` to update this periodically, so that it's not fixed at page load.

An example of a component doing something similar is here: https://github.com/learningequality/kolibri/blob/develop/kolibri/core/assets/src/views/ElapsedTime.vue#L34

rtibbles (commented):
Clicking on create scheduled sync in the device page is not doing anything just yet, as you need to add an event handler for the `manageSync` event here: https://github.com/learningequality/kolibri/blob/develop/kolibri/plugins/device/assets/src/views/FacilitiesPage/index.vue#L46

You will also need to add additional routes for managing sync schedules and editing sync schedules here: https://github.com/learningequality/kolibri/blob/develop/kolibri/plugins/device/assets/src/routes/index.js#L65 (you've so far added them to the facility plugin, but the device plugin defines its routes separately).

rtibbles:
I am replicating the 403 when trying to save, but have not worked out the cause just yet.

PR #10206, opened 2023-03-07
Fix user in change facility

jredrejo:
## Summary
This PR solves several problems that have raised when testing the merge facility flow:
- The one described at #9791 was caused when the target facility requires a password for the user. Previous design didn't have this scenario in mind.  This has required changes in the state machine to add a new state and a new form for the user to enter it.
- Some changes in the facility model were breaking Morango Sync
- Some changes in the auth tasks required one more parameter to validate
- New wording for one of the strings

New window for the user to set a password if needed:
image

## References
Closes: #9791 

## Reviewer guidance
The steps @pcenov  described in #9791 to reproduce the issue can be used here to test it

## Testing checklist

- [x] Contributor has fully tested the PR manually
- [x] If there are any front-end changes, before/after screenshots are included
- [x] Critical user journeys are covered by Gherkin stories
- [x] Critical and brittle code paths are covered by unit tests

## PR process

- [ ] PR has the correct target branch and milestone
- [ ] PR has 'needs review' or 'work-in-progress' label
- [ ] If PR is ready for review, a reviewer has been added. (Don't use 'Assignees')
- [ ] If this is an important user-facing change, PR or related issue has a 'changelog' label
- [ ] If this includes an internal dependency change, a link to the diff is provided

## Reviewer checklist

- Automated test coverage is satisfactory
- PR is fully functional
- PR has been tested for [accessibility regressions](http://kolibri-dev.readthedocs.io/en/develop/manual_testing.html#accessibility-a11y-testing)
- External dependency files were updated if necessary (`yarn` and `pip`)
- Documentation is updated
- Contributor is in AUTHORS.md

Issue #9791, opened 2022-10-24
Change facility - Misleading message after an attempt to create a user with an existing username


pcenov:
## Observed behavior
The message which is displayed is 'User ‘user’ already exists in ‘TARGET’ is not a learner. Please choose a different username.'

## Expected behavior
The text should be further specified and that should be reflected in Figma as I wasn't able to find it there.

## Steps to reproduce the issue
1. Install the following build: https://buildkite.com/organizations/learningequality/pipelines/kolibri-debian/builds/5203/jobs/0183fbda-239d-4c8c-8201-420d48d92d2c/artifacts/0184090a-8e30-43f9-af55-d55929fe454a
2. Modify the DB as specified in https://github.com/learningequality/kolibri/pull/9760
3. Create a learner account and sign in as the learner
4. Go to Profile and click the Change button
5. Select the target learning facility and press Continue
6. At the next screen select the 'Create new account' option
7. Fill in the username of an existing user.

## Screenshot
![2022-10-24_18-28-39](https://user-images.githubusercontent.com/79847249/197567253-7b97d4e5-6dd6-4362-a4c1-dfd844cc2466.png)

## Usage Details
Windows 10, Ubuntu - Chrome

pcenov:
@jredrejo @radinamatic

PR #10222, opened 2023-03-09
Fix error for user merge data

jredrejo:
## Summary
When merging the user to a different facility, if the remote facility didn't have any gender, birthdate or full name assigned, the result of the merge set them in blank for the final user. 

## References
Related: #10206

## Reviewer guidance
When merging an user as described in #9791 if the local user has set Gender, Birth year, identifier or fullname, they should stay after the merge when seing the profile in the local device.

NOTE: Checked that the remote facility does not get this info because users are usually in a morango partition that's read only. It's a known issue that has no solution by now according to @bjester 

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/plugins/user_profile/tasks.py
+++ b/kolibri/plugins/user_profile/tasks.py
@@ -111,7 +111,7 @@ def mergeuser(command, **kwargs):
     merge_users(local_user, remote_user)
 
     # Resync with the server to update the merged records
-    # kwargs["no_pull"] = True
+    kwargs["no_pull"] = True
     del kwargs["no_push"]
     call_command("sync", **kwargs)
     new_superuser_id = kwargs.get("new_superuser_id")
--- a/kolibri/plugins/user_profile/tasks.py
+++ b/kolibri/plugins/user_profile/tasks.py
@@ -111,7 +111,6 @@ def mergeuser(command, **kwargs):
     merge_users(local_user, remote_user)
 
     # Resync with the server to update the merged records
-    kwargs["no_pull"] = True
     del kwargs["no_push"]
     call_command("sync", **kwargs)
     new_superuser_id = kwargs.get("new_superuser_id")
--- a/kolibri/core/auth/utils/migrate.py
+++ b/kolibri/core/auth/utils/migrate.py
@@ -26,10 +26,10 @@ logger = logging.getLogger(__name__)
 
 
 def _merge_user_models(source_user, target_user):
-    for f in ["gender", "birth_year", "id_number"]:
+    for f in ["gender", "birth_year", "id_number", "full_name"]:
         source_value = getattr(source_user, f, None)
         target_value = getattr(target_user, f, None)
-        if not target_value and source_value:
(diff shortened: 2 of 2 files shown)


rtibbles:
Wait.. if this is running as a single user sync - does the merging of the learner data on the local device ever actually work? It will get set locally, but then never sent back, right?

bjester:
> Wait.. if this is running as a single user sync - does the merging of the learner data on the local device ever actually work? It will get set locally, but then never sent back, right?

The user account is read-only, but if you're referring to learner activity, it should get synced because that shouldn't have a read-only partition.

kolibri/plugins/user_profile/tasks.py
```diff
@@ -111,7 +111,7 @@ def mergeuser(command, **kwargs):
     merge_users(local_user, remote_user)
 
     # Resync with the server to update the merged records
-    # kwargs["no_pull"] = True
+    kwargs["no_pull"] = True
```
bjester: While it is unlikely that an issue would arise from this, having this set to `no_pull` does leave a possibility.

rtibbles:
> The user account is read-only

Right - that's what I was thinking, we're merging the data from the local user onto the synced in remote user, so this will never get synced back.

bjester:
I believe that's why @jredrejo opened this one afterwards https://github.com/learningequality/kolibri/pull/10225

bjester:
Maybe there's still a piece missing, sending that additional info when creating the new user on the remote

rtibbles:
Yeah - with the follow up PR, if the user doesn't exist in the remote, we will at least preserve the locally set demographic data and full name. But if the user already exists in the remote, once we've synced it in, there's no way to propagate that data back, so the merging (pre-existing to @jredrejo's PR) and the fixes made here, are not doing anything :/

rtibbles:
Almost feels like we made a mistake with the user demographic data, and it should have been in a separate model FKed to the user, with a user specific partition (but rectifying that seems like a whole other mess).

bjester:
Oh I see, the existing user scenario makes sense

kolibri/plugins/user_profile/tasks.py
jredrejo: ok, logic tells me this was a good idea and would accelerate the process. It worked in all the tests I did, but the you're the expert in Morango syncs, so I'll remove it if you think there might be some scenario where it could cause problems.

jredrejo:
> Yeah - with the follow up PR, if the user doesn't exist in the remote, we will at least preserve the locally set demographic data and full name. But if the user already exists in the remote, once we've synced it in, there's no way to propagate that data back, so the merging (pre-existing to @jredrejo's PR) and the fixes made here, are not doing anything :/

Yes, that's what I've seen in my tests but I saw it as a natural consequence of having two sources of truth and I always thought that was expected when this process was designed.

rtibbles (approved):

PR #10225, opened 2023-03-09
When creating a new remote user send all the local user information

### LINKED ISSUE #10223: Modify current api to create users to accept more parameters
## Observed behavior
As stated in #10222 when syncing users from one facility to another, the user information is not saved because the Morango partition is in read-only mode. This is also related to #8849

However, when merging an user to a remote facility, if the user needs to be created in the remote facility we only can provide username/facility/password using the current API, so the created user does not include full name, gender, id or birth  year if the previous user have them.
It'd be great to be able to pass this info to the API when creating the user in the remote facility so this information is saved.

## Expected behavior
Be able to create a new user in a remote facility providing all the user information, not just its username and password

## Context
kolibri 0.16

jredrejo:
## Summary
In the merge facility process, when a new user needs to be created in the remote facility, send all the user information, not just the username and password.
Pushing the sync info won't work because the Morango partition is read only, so we need to ensure the creation includes all the existing information

## References
Closes: #10223
Depends:  #10222

## Reviewer guidance
Same process as in #10222 
This PR must not be merged until #10222 has been merged

## Testing checklist

## PR process

- [s] PR has 'needs review' or 'work-in-progress' label

## Reviewer checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/plugins/user_profile/tasks.py
+++ b/kolibri/plugins/user_profile/tasks.py
@@ -53,6 +53,9 @@ class MergeUserValidator(PeerImportSingleSyncJobValidator):
             "password": data["password"],
             "facility": facility,
         }
+        for f in ["gender", "birth_year", "id_number", "full_name"]:
+            if getattr(data["local_user_id"], f, None):
+                user_data[f] = getattr(data["local_user_id"], f, None)
         public_signup_url = reverse_remote(baseurl, "kolibri:core:publicsignup-list")
         response = requests.post(public_signup_url, data=user_data)
         if response.status_code != HTTP_201_CREATED:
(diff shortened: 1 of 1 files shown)


kolibri/plugins/user_profile/tasks.py
```diff
@@ -53,6 +53,9 @@ def create_remote_user(self, data):
             "password": data["password"],
             "facility": facility,
         }
+        for f in ["gender", "birth_year", "id_number", "full_name"]:
+            if getattr(data["local_user_id"], f, None):
+                user_data[f] = getattr(data["local_user_id"], f, None)
```
rtibbles, line 58: Should we also exclude the NOT_SPECIFIED value here too? https://github.com/learningequality/kolibri/blob/develop/kolibri/core/auth/constants/demographics.py#L5

jredrejo: yep, good catch.
Changed!

rtibbles (approved):

Issue #9784, opened 2022-10-18
When the root folders of channels appear in search results, they appear as unadorned folder cards rather than channels


rtibbles:
## Observed behavior
Search can result in folders being displayed in search results. When a folder is the root folder of a channel, it does correctly link to the channel root in Kolibri, but does not display the additional channel metadata that is already available in the client side.

See the first folder here, when clicked, links to the root of the channel

![image](https://user-images.githubusercontent.com/1680573/196498119-c22a40bb-2cb7-46d1-ae00-17e9c5132c8b.png)

https://kolibri-catalog-en.learningequality.org/en/learn/#/library?channels=1e378725d3924b47aa5e1260628820b5

## Expected behavior
Root folders should display channel thumbnails etc.

photon0205:
Hi! I would like to take up this issue

rtibbles:
Sure! You should look at the  current implementation of Channel display in order to see how to display the channels here:
https://github.com/learningequality/kolibri/blob/develop/kolibri/plugins/learn/assets/src/views/ChannelCard.vue

(I would recommend just using this component directly as a first pass and seeing how it looks).

photon0205:
@rtibbles if i use the ChannelCard component, then this is how it looks:
<img width="1433" alt="Screenshot 2023-03-11 at 12 33 59 PM" src="https://user-images.githubusercontent.com/99784265/224470589-90993083-1b17-4add-a3e8-0f6776866846.png">
Ans if i just use channel_thumbnails for the content not having anything in their thumbnails, then it looks like this:
<img width="1433" alt="Screenshot 2023-03-11 at 12 35 04 PM" src="https://user-images.githubusercontent.com/99784265/224470632-638f8d59-8b74-4675-8b4b-54ace25f7147.png">
Now you suggest which one should i go for

rtibbles:
Fixed in #10231

PR #10231, opened 2023-03-12
ptch: display channel thumbnails in root folder

photon0205:
Solved #9784 by displaying the channel thumbnail of the root folders
ss


kolibri/core/assets/src/utils/contentNodeUtils.js
```diff
@@ -9,6 +9,9 @@ import { MasteryModelTypes } from 'kolibri.coreVue.vuex.constants';
  * from `files` is returned.
  */
 export function getContentNodeThumbnail(contentnode) {
+  if (!contentnode.thumbnail) {
```
rtibbles: Thanks for the pull request! Unfortunately, this fix is too broad, and would cause the channel thumbnail to be used for every ContentNode that does not have a thumbnail. It will also be used everywhere this function is used, so it might be nicer to limit the scope of this to the Learn plugin.

I would suggest instead to make the fix in `ContentNodeThumbnail.vue`, check the return value of this function, and if it is falsy, then you can check whether the `contentNode` has a truthy `parent` attribute. If the `parent` attribute is falsy, that means it is a root node, and you can safely return the channel thumbnail.

Also, the `channel_thumbnail` property is not reliably defined, so you are better off using this function from the `useChannels` composable: https://github.com/learningequality/kolibri/blob/develop/kolibri/plugins/learn/assets/src/composables/useChannels.js#L51

See here for an example of its use: https://github.com/learningequality/kolibri/blob/develop/kolibri/plugins/learn/assets/src/views/HybridLearningContentCard/index.vue#L113

rtibbles (dismissed):
This fix would be a bit too broad, let's do something more targeted, and use a more reliable method to get the channel thumbnail, scoped to the learn plugin only.

photon0205:
with this change as well i am getting the same result

rtibbles (commented):
Excellent - thanks for making these updates! I'll let our QA team confirm the fix, and then this can be merged.

In the meantime, if you would like, please push an additional commit adding yourself to the AUTHORS.md file: https://github.com/learningequality/kolibri/blob/develop/AUTHORS.md

pcenov:
No issues observed while manually testing this - LGTM! @radinamatic

radinamatic (approved):
Manual QA passes, good to go! 🎉 💯 :shipit:

rtibbles:
@photon0205 thank you for your first contribution to Kolibri - it is very much appreciated, great work!

PR #10252, opened 2023-03-15
Final string prep after crowdin review for translators

marcellamaki:
Final catch all for tidying up: 
- Add missing AND
- Bare strings
- Misc. capitalization and cleanup


MisRob (approved):
Code is looking good to me, thank you

radinamatic (approved):
Good to gooooooooooo!!! 💥 :shipit:

PR #10247, opened 2023-03-14
Handle network problems when merging users

jredrejo:
## Summary

When trying to move one user to another facility device network errors could happen:
1. before the merge has actually happened
2. after the merge has happened

For case 1, this PR just stop the process and show an error message to the user
For case 2, this PR schedules a new sync and inform the user the merge has happened

error_merge

## References
Closes: #9792 @pcenov 

## Reviewer guidance
Prepare an scenario with two kolibri devices running and start the merge project.
Stop it after the process has begun and check things happen as this PR summary states

## Testing checklist

- [x] Contributor has fully tested the PR manually
- [x] If there are any front-end changes, before/after screenshots are included
- [ ] Critical user journeys are covered by Gherkin stories
- [ ] Critical and brittle code paths are covered by unit tests

## PR process

- [x] PR has the correct target branch and milestone
- [x] PR has 'needs review' or 'work-in-progress' label
- [x] If PR is ready for review, a reviewer has been added. (Don't use 'Assignees')
- [ ] If this is an important user-facing change, PR or related issue has a 'changelog' label
- [ ] If this includes an internal dependency change, a link to the diff is provided

## Reviewer checklist

- Automated test coverage is satisfactory
- PR is fully functional
- PR has been tested for [accessibility regressions](http://kolibri-dev.readthedocs.io/en/develop/manual_testing.html#accessibility-a11y-testing)
- External dependency files were updated if necessary (`yarn` and `pip`)
- Documentation is updated
- Contributor is in AUTHORS.md

PR #10258, opened 2023-03-15
Fix string references.

rtibbles:
## Summary
* Fixes some string references that were still to coreStrings but should have been to commonSyncStrings
* Fixes typo in coreStrings
* Adds profileStrings mixin to SelectFacility.vue that was missing it

## References
Fixes missing strings noted in https://github.com/learningequality/kolibri/pull/10247#issuecomment-1470099760

### DIFF
--- a/kolibri/plugins/user_profile/assets/src/views/ChangeFacility/SelectFacility.vue
+++ b/kolibri/plugins/user_profile/assets/src/views/ChangeFacility/SelectFacility.vue
@@ -40,7 +40,7 @@
       <KGridItem>{{ $tr('doNotSeeYourFacility') }}</KGridItem>
       <KGridItem>
         <KButton
-          :text="coreString('addNewAddressAction')"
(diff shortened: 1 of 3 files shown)


jredrejo (approved):
Thanks!

kolibri/plugins/user_profile/assets/src/views/ChangeFacility/SelectFacility.vue
```diff
@@ -40,7 +40,7 @@
       <KGridItem>{{ $tr('doNotSeeYourFacility') }}</KGridItem>
       <KGridItem>
         <KButton
-          :text="coreString('addNewAddressAction')"
+          :text="getCommonSyncString('addNewAddressAction')"
```
bjester, line 43: Shoot, I should have checked elsewhere-- I noticed this in the `SelectDeviceForm`

PR #10259, opened 2023-03-15
Sync schedule cleaning

### LINKED ISSUE #10173: Finalize sync schedule work
This is a follow up issue to #9786 which has been slightly prematurely merged in order to get strings in before string freeze.

The following issues need to still be addressed:

```[tasklist]
### Tasks
- [x] Handle click events in the common status component by emitting an event that is handled in the parent component in order to avoid encoding facility specific navigation events in the core component (see https://github.com/learningequality/kolibri/pull/9786/files#r1122278683)
- [x] Reinstate disabling of register option: https://github.com/learningequality/kolibri/pull/9786/files#r1122279739
- [x] Add MANAGE_SYNC constant to `OPTIONS` so that it has a value rather than being `undefined`: https://github.com/learningequality/kolibri/pull/9786/files#r1122207691
- [x] Reinstate conditional display of the Register Menu option: https://github.com/learningequality/kolibri/pull/9786/files#r1122209233
- [x] Handle loading state for facility data: https://github.com/learningequality/kolibri/pull/9786/files#r1119538350
- [x] Handle unloaded network locations: https://github.com/learningequality/kolibri/pull/9786/files#r1122283756
- [x] Clean up validation warnings from KSelects: https://github.com/learningequality/kolibri/pull/9786/files#r1122285628
- [x] Check and resolve styling issues for KSelects: https://github.com/learningequality/kolibri/pull/9786#issuecomment-1438852763
- [x] Finalize editing/creation functionality to make sure fully functional
- [x] Add implementation and routing to device plugin as well as facility
```

AllanOXDi:
## Summary

- Added a click events in the common status component  to avoid encoding facility specific navigation events in the core component .
- Reinstated register option.
- Added MANAGE_SYNC constant to `OPTIONS` so that it has a value .
- Reinstated conditional display of the Register Menu option.
- fixes loading state for facility data.
- fixes unloaded network locations.
-  Resolve styling issues for KSelects.
- fixes editing/creation functionality to make sure fully functional.
- fixes  routing to device plugin as well as facility

closes #10173

# Reviewer guidance
Please see[ figma](https://www.figma.com/file/XHO5zXznOD3WXcxcA6d6Zv/Scheduled-sync-2022?node-id=83-1632) design 
## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated


kolibri/core/assets/src/views/sync/FacilityNameAndSyncStatus/index.vue
```diff
@@ -61,6 +61,7 @@
   import { now } from 'kolibri.utils.serverClock';
   import taskStrings from 'kolibri.coreVue.mixins.commonTaskStrings';
   import commonCoreStrings from 'kolibri.coreVue.mixins.commonCoreStrings';
+  import { PageNames } from '../../../../../../plugins/facility/assets/src/constants';
```
nucleogenesis: We want to avoid importing from a plugin like this. The `core` front-end module cannot have a dependency because if Kolibri was run without the Facility plugin enabled, then the whole app wouldn't build (not that this is a particularly likely situation, but best to avoid the possibility).

It looks like the value you use from this is also being added to the Device plugin's constants too.

If there is some sort of cross-plugin use case for this set of constants then I'd suggest moving them into `core/assets/src/constants.js`, which will then allow them to be imported into any plugin in Kolibri a la `import { SyncPageNames } from 'kolibri.coreVue.vuex.constants';`

If you decide to go that route, then you'll just need to remove them from the Facility and Device plugins and update references to them accordingly.

AllanOXDi:
It was closed because I deleted my forked repo. Reopening it

kolibri/core/assets/src/views/sync/FacilityNameAndSyncStatus/index.vue
```diff
@@ -111,6 +112,12 @@
         }
         return this.$formatRelative(datetime, { now: this.now });
       },
+      manageSync() {
```
rtibbles, line 115: You mention this in your PR description:
> Added a click events in the common status component to avoid encoding facility specific navigation events in the core component .

But this is doing the exact opposite - it's removing the `manageSync` event emitting and instead encoding the route navigation here.

kolibri/plugins/device/assets/src/modules/pluginModule.js
```diff
@@ -9,6 +9,7 @@ export default {
   state() {
     return {
       authenticateWithPin: false,
+      isDevicePlugin: false,
```
rtibbles, line 12: This vuex module is only used in the device plugin, so this will always be true when this is being used. Better to find another way to handle this conditionality, wherever you might be using it.

kolibri/plugins/device/assets/src/views/DeviceSettingsPage/index.vue
```diff
@@ -707,6 +707,7 @@
       },
       getFacilitySettingsPath(facilityId = '') {
         const getUrl = urls['kolibri:kolibri.plugins.facility:facility_management'];
+        window.ur;
```
rtibbles, line 710: Typo?

kolibri/plugins/device/assets/src/views/FacilitiesPage/index.vue
```diff
@@ -290,7 +292,10 @@
         } else if (option === Options.REGISTER) {
           this.facilityForRegister = facility;
         } else if (option === Options.MANAGESYNC) {
-          this.$router.push({ path: '/manage' });
+          this.$router.push({
```
rtibbles, line 295: This is good - you can also attach an event handler to the `FacilityNameAndSyncStatus` to handle the `manageSync` event to navigate to this page as well, so that both means for getting to this page are handled.

kolibri/plugins/facility/assets/src/constants.js
```diff
@@ -10,8 +10,6 @@ export const PageNames = {
   DATA_EXPORT_PAGE: 'DATA_EXPORT_PAGE',
   FACILITY_CONFIG_PAGE: 'FACILITY_CONFIG_PAGE',
   ALL_FACILITIES_PAGE: 'ALL_FACILITIES_PAGE',
-  MANAGE_SYNC_SCHEDULE: 'MANAGE_SYNC_SCHEDULE',
```
rtibbles, line 13: Why are we removing these pages from the facility? We want to be able to manage sync schedules in both the facility _and_ the device plugins.

kolibri/plugins/facility/assets/src/modules/pluginModule.js
```diff
@@ -48,7 +48,7 @@ export default {
       // parameter if we're in a multi-facility situation
       const params = {};
       if (getters.userIsMultiFacilityAdmin) {
-        params.facility_id = getters.activeFacilityId;
+        params.facilityId = getters.activeFacilityId;
```
rtibbles, line 51: The route uses the parameter `facility_id` below which is what this is encoding, so you should revert this change.

kolibri/plugins/facility/assets/src/routes.js
```diff
@@ -128,12 +132,20 @@ export default [
...
+    name: SyncPageNames.MANAGE_SYNC_SCHEDULE,
   },
   {
-    path: '/:facility_id?/editdevice/:deviceId',
+    path: '/:facility_id?/editdevice/:deviceId/',
     component: editDeviceSyncSchedule,
-    name: PageNames.EDIT_SYNC_SCHEDULE,
+    name: SyncPageNames.EDIT_SYNC_SCHEDULE,
+    props: route => {
+      return { facilityId: route.params.facilityId || store.getters.currentFacilityId };
```
rtibbles, line 146: This should be `route.params.facility_id` - the name of the parameter matches what is specified in the route path:
`path: '/:facility_id?/editdevice/:deviceId/',`

kolibri/plugins/facility/assets/src/views/DataPage/SyncInterface/index.vue
```diff
@@ -130,7 +130,7 @@
   import CoreMenu from 'kolibri.coreVue.components.CoreMenu';
   import CoreMenuOption from 'kolibri.coreVue.components.CoreMenuOption';
   import { TaskStatuses } from 'kolibri.utils.syncTaskUtils';
-  import { PageNames } from './../../../constants';
+  import { PageNames } from '../../../../../../device/assets/src/constants';
```
rtibbles, line 133: We should revert this - the page names for the facility plugin are distinct from the page names for the device plugin.

kolibri/plugins/device/assets/src/constants.js
```diff
@@ -14,6 +14,8 @@ export const PageNames = {
   AVAILABLE_CHANNELS_PAGE: 'AVAILABLE_CHANNELS',
   FACILITIES_PAGE: 'FACILITIES_PAGE',
   FACILITIES_TASKS_PAGE: 'FACILITIES_TASKS_PAGE',
+  MANAGE_SYNC_SCHEDULE: 'MANAGE_SYNC_SCHEDULE',
```
rtibbles, line 17: I see you have defined these in the `kolibri-common` package - I like that idea! That does mean you don't have to repeat them here. For explicitness sake, you could import those SyncPageNames within this file and add them to the `PageNames` constant here using the object spread operator:

```
export const PageNames = {
  MANAGE_CONTENT_PAGE: 'MANAGE_CONTENT_PAGE',
  MANAGE_PERMISSIONS_PAGE: 'MANAGE_PERMISSIONS_PAGE',
  USER_PERMISSIONS_PAGE: 'USER_PERMISSIONS_PAGE',
  DEVICE_INFO_PAGE: 'DEVICE_INFO_PAGE',
  DEVICE_SETTINGS_PAGE: 'DEVICE_SETTINGS_PAGE',
  REARRANGE_CHANNELS: 'REARRANGE_CHANNELS',
  DELETE_CHANNELS: 'DELETE_CHANNELS',
  EXPORT_CHANNELS: 'EXPORT_CHANNELS',
  MANAGE_TASKS: 'MANAGE_TASKS',
  MANAGE_CHANNEL: 'MANAGE_CHANNEL',
  SELECT_CONTENT: 'SELECT_CONTENT',
  NEW_CHANNEL_VERSION_PAGE: 'NEW_CHANNEL_VERSION_PAGE',
  AVAILABLE_CHANNELS_PAGE: 'AVAILABLE_CHANNELS',
  FACILITIES_PAGE: 'FACILITIES_PAGE',
  FACILITIES_TASKS_PAGE: 'FACILITIES_TASKS_PAGE',
  ...SyncPageNames,
};
```

This will add all the keys and values from the `SyncPageNames` object to the `PageNames` object in this file.

packages/kolibri-common/components/SyncSchedule/EditDeviceSyncSchedule.vue
```diff
@@ -270,15 +270,21 @@
           });
       },
       handleSaveSchedule() {
-        FacilityResource.fetchModel({ id: this.$store.getters.activeFacilityId, force: true }).then(
+        FacilityResource.fetchModel({ id: this.$route.params.facilityId, force: true }).then(
```
rtibbles, line 273: Relying on the route parameters here might be a bit brittle - may be better to pass in both the facilityId and the deviceId as props to this component, so that we're not reliant on a certain route being in place when this is used.

packages/kolibri-common/components/SyncSchedule/EditDeviceSyncSchedule.vue
```diff
@@ -292,18 +298,19 @@
       },
 
       cancelBtn() {
-        this.$router.push({ name: PageNames.ManageSyncSchedule });
+        this.$router.push({ name: SyncPageNames.MANAGE_SYNC_SCHEDULE });
```
rtibbles, line 301: This will lose the context of the facility in case the route needs one - we may be better off passing in this 'back' and 'cancel' routes as parameters that we can configure in the different plugins.

packages/kolibri-common/components/SyncSchedule/EditDeviceSyncSchedule.vue
```diff
@@ -292,18 +298,19 @@
...
       },
       fetchDevice() {
         NetworkLocationResource.fetchModel({ id: this.$route.params.deviceId }).then(device => {
           this.device = device;
+          this.baseurl = device.base_url;
```
rtibbles, line 306: Where do we use this assigned value later?

packages/kolibri-common/components/SyncSchedule/EditDeviceSyncSchedule.vue
```diff
@@ -292,18 +298,19 @@
...
           TaskResource.list({ queue: 'facility_task' }).then(tasks => {
             this.tasks = tasks.filter(
               task =>
                 task.extra_metadata.device_id === device.id &&
                 task.facility_id === this.$store.getters.activeFacilityId
             );
-            if (this.tasks.length != 0) {
+            if (this.tasks) {
```
rtibbles, line 313: You do need to do at least `if (this.tasks && this.tasks.length)` as otherwise if tasks is an empty Array, this will still be `true`.

packages/kolibri-common/components/SyncSchedule/ManageSyncSchedule.vue
```diff
@@ -261,11 +276,49 @@
...
-          this.$router.push({ name: PageNames.EDIT_SYNC_SCHEDULE, params: { deviceId: value } });
+          this.$router.push({
+            name: SyncPageNames.EDIT_SYNC_SCHEDULE,
+            params: { deviceId: value, facilityId: this.facilityId },
+          });
         } else {
           return window.location.href;
         }
       },
+      ScheduleTime(time, timestamp) {
```
rtibbles, line 287: Maybe change this to camelCase (scheduleTime), as we usually use that for Javascript/Vue methods.

packages/kolibri-common/components/SyncSchedule/ManageSyncSchedule.vue
```diff
@@ -261,11 +276,49 @@
       },
       editButton(value) {
         if (value !== ' ') {
-          this.$router.push({ name: PageNames.EDIT_SYNC_SCHEDULE, params: { deviceId: value } });
+          this.$router.push({
```
rtibbles, line 279: The fact that some of our routes are using `facility_id` as the facility parameter, means that this isn't going to work universally.

rtibbles (commented):
A bit to clean up here, good progress on displaying the current sync schedules, and it's starting to come together overall. I haven't provided a direct path forward here for dealing with the routing, but there is definitely a bit of a blocker with the seeming inconsistency between `facility_id` and `facilityId` in the route parameter name which we'll need to resolve.

kolibri/plugins/device/assets/src/views/DeviceSettingsPage/index.vue
AllanOXDi: Yes, fixing it.

rtibbles:
All comments in this PR have been addressed in #10463, so closing this as superseded by #10463

PR #10264, opened 2023-03-16
Allows superusers and those with can_manage_content permision to access '/ispinvalid' API resource

### LINKED ISSUE #10238: The device management PIN functionality is not functioning correctly on a Learn-only device
## Observed behavior
After creating a Learn-only device I do see the PIN modal however it's not accepting the the PIN. Also after removing the PIN from the main facility and restarting both servers, the PIN modal is still being displayed.

## Expected behavior
Fully functional PIN functionality.

1. Install the the [following build ](https://buildkite.com/learningequality/kolibri-debian/builds/6281) 
2.  Setup a full facility and add a Device management PIN at Facility > Settings
3. Setup a Learn-only facility and attempt to import a channel.

## Usage Details
Windows 10, Ubuntu 22  - Chrome

akolson:
## Summary

This pr fixes issues raised [here](https://github.com/learningequality/kolibri/issues/10238) related to the device management pin feature picked up during manual QA. The `/ispinvalid`(previously `/is-pin-valid` before this fix) resource has been updated to allow access to both superusers and users with the `can_manage_content` permission.

## References

Fixes https://github.com/learningequality/kolibri/issues/10238 https://github.com/learningequality/kolibri/issues/10275

## Reviewer guidance

Please see https://github.com/learningequality/kolibri/pull/10135

----

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/core/auth/api.py
+++ b/kolibri/core/auth/api.py
@@ -71,6 +71,7 @@ from kolibri.core.api import ReadOnlyValuesViewset
 from kolibri.core.api import ValuesViewset
 from kolibri.core.auth.constants.demographics import NOT_SPECIFIED
 from kolibri.core.auth.permissions.general import _user_is_admin_for_own_facility
+from kolibri.core.auth.permissions.general import DenyAll
 from kolibri.core.device.utils import allow_guest_access
 from kolibri.core.device.utils import allow_other_browsers_to_connect
 from kolibri.core.device.utils import valid_app_key_on_request
@@ -154,6 +155,14 @@ class KolibriAuthPermissions(permissions.BasePermission):
         return False
 
 
+class IsPINValidPermissions(DenyAll):
+    def has_permission(self, request, view):
+        return request.user.is_superuser or request.user.can_manage_content
+
+    def has_object_permission(self, request, view, obj):
+        return self.has_permission(request, view)
+
+
 class FacilityDatasetViewSet(ValuesViewset):
     permission_classes = (KolibriAuthPermissions,)
     filter_backends = (KolibriAuthPermissionsFilter,)
@@ -219,8 +228,11 @@ class FacilityDatasetViewSet(ValuesViewset):
         except FacilityDataset.DoesNotExist:
             raise Http404("Facility not found")
 
-    @decorators.action(methods=["post"], detail=True, url_path="is-pin-valid")
-    def is_pin_valid(self, request, pk):
+
+class IsPINValidView(views.APIView):
+    permission_classes = (IsPINValidPermissions,)
+
+    def post(self, request, pk):
         serializer = ExtraFieldsSerializer(data=request.data)
         if not serializer.is_valid() or serializer.data.get("pin_code") is None:
             return HttpResponseBadRequest("Invalid pin input")
--- a/kolibri/core/auth/api_urls.py
+++ b/kolibri/core/auth/api_urls.py
@@ -6,6 +6,7 @@ from .api import FacilityDatasetViewSet
 from .api import FacilityUsernameViewSet
 from .api import FacilityUserViewSet
 from .api import FacilityViewSet
+from .api import IsPINValidView
 from .api import LearnerGroupViewSet
 from .api import MembershipViewSet
 from .api import RoleViewSet
@@ -48,6 +49,11 @@ urlpatterns = (
             r"^usernameavailable$",
             UsernameAvailableView.as_view(),
             name="usernameavailable",
+        ),
+        url(
+            r"^ispinvalid/(?P<pk>[a-f0-9]{32})$",
+            IsPINValidView.as_view(),
+            name="ispinvalid",
         ),
     ]
 )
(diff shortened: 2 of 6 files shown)


rtibbles (commented):
Code read through makes sense to me. QA sign off and we can merge!

pcenov:
Hi @akolson I confirm that the PIN modal is displayed correctly now on a Learn-only device and the PIN is accepted. I have reported separately the following issues: https://github.com/learningequality/kolibri/issues/10273, https://github.com/learningequality/kolibri/issues/10274, https://github.com/learningequality/kolibri/issues/10275

It's also worth mentioning that I am being prompted to enter the PIN several times while in the Device plugin which seems a bit excessive and should be further discussed:

https://user-images.githubusercontent.com/79847249/225896993-fc6c1bea-3f58-4fb0-baf3-2305a997041b.mp4

akolson:
Hi @pcenov I appreciate the feedback. Concerning the pin prompt being requested several times, the authentication expires in 15 seconds according to the spec. @marcellamaki, @rtibbles  and probably @jtamiace,@tomiwaoLE can weigh in on this on how we could better align this

marcellamaki:
Hi @akolson and @pcenov - Thanks for flagging. You definitely implemented according to spec, but I did not write a very well-thought-out spec. Sorry about this! 

So, for this part:
> A user should not be able to navigate directly via a URL without being prompted to authenticate, but should not have to enter the PIN twice, once when pressing "Device" in the navigation, and then again when redirected

The 15-second here was not very well thought out, and I was thinking only of the experience of the user being redirected to the route. I do not have enough context of your implementation at the top of my mind, Samson, but depending on how you did this, there is probably a check on the route that prompts the user to enter a PIN on route enter. 

When the user presses "Device", they are prompted to enter the PIN on that event, even though they have not actually been redirected, yet. Then, we want to make sure that they don't get that second prompt, based on the check associated with the "on route enter" check.

But, now that I think about it more, I do think there would be value in having this cookie persist longer, although I am not sure what interval might be appropriate. Perhaps 15 minutes, while the admin is working on the tablet? And/or, perhaps the cookie could be cleared on navigation back to Learn. So, if you are doing a task, and you have completed and you pass the device back to the learner, you can just click back into learn, and know that they will not have the ability to continue the admin session in device (intentionally or inadvertently)

@rtibbles any other thoughts, ideas, or context? Sorry @akolson that we are figuring some of this out very much on the fly. Thank you for your patience.

akolson:
Thanks @marcellamaki for this, and no worries. Some features only get better after such refinements :). I still appreciate the detail put into these issues regardless. 

Now concerning the implementation, on a very high level, the PIN check(if set up in facility settings) happens just before a user is redirected to any of the device plugin's routes. The stored cookie is checked in the process to avoid re-authentication if it is still valid. This all happens in the `beforeResolve` navigation guard of the device plugin.

My thinking during implementation of this issue was to remove cookie expiry when browsing within the device plugin and set it with expiry after navigating away to other plugins but my approach highly depended on using the `beforeResolve` navigation guard within the context of cross plugin navigation which currently has it's limitations as each plugin's navigation is treated in complete isolation. I am very happy to hear thoughts on other ways this could be achieved, while maintaining the same approach or otherwise @rtibbles?

rtibbles:
A simple fix for not forcing reauthentication with the PIN while navigating within the Device plugin might be to first try to read the value from a stored value in Javascript (so some global state in Javascript) and then to read it from the cookie. If the cookie is set, also set the global JS state.

If the cookie is not set, do the PIN auth, and then set both the cookie and the JS state. The JS state will persist within navigation within the device tab unless someone does a hard refresh on the page, but will not persist once someone has navigated away from the Device plugin. I feel like this might prevent continual reauthentication while still making it not persist too long?

akolson:
Thanks Richard for the suggestion. Let me try this out and revert.

akolson:
Multiple PIN prompting while navigating within the Device plugin has been resolved, thanks to @rtibbles @marcellamaki. Below is a description of the various scenarios under which it will be shown provided PIN authentication has been setup;

- Initially prior to inputing of the PIN
- Navigating to Device plugin, authenticating with PIN, navigating to other plugin, and returning back to the device plugin after 15 second(The time taken for the cookie to expire)
- If you refresh your current page within the Device plugin when the cookie session has expired(Expires within 15 second)

pcenov:
Hi @akolson now no matter what I do the PIN modal never shows up. I'm testing using the following build: https://buildkite.com/organizations/learningequality/pipelines/kolibri-debian/builds/6373/jobs/018705db-d71e-4a07-85d3-8c024059e354/artifacts/01870924-419a-497b-a711-c1b7f1b36b8e
Could you please check as well?

akolson:
Hi @pcenov , My apologies, I committed a piece of test code that disables the PIN prompt on LODs. This has been reverted and all should be fine now.

pcenov:
@akolson I confirm - now the PIN modal is displayed as described in your comment and it's much more user friendly. Thanks!

radinamatic (approved):
We pinned the PIN, go us!!! 🎉 💯 :shipit:

PR #10265, opened 2023-03-16
Temporarily disable raspberry pi image generation for releases

rtibbles:
## Summary
* Currently the debian build on GHA does not produce deb files with back compatible compression, which breaks the Pi image build


jredrejo:
is there a follow up issue to restore the build after the packaging is updated?

marcellamaki (approved):

marcellamaki:
> is there a follow up issue to restore the build after the packaging is updated?

https://github.com/learningequality/kolibri/issues/10266 😃

PR #10277, opened 2023-03-17
Normalize use of BACK in setup wizard

### LINKED ISSUE #10240: SetupWizard: Verify proper functionality when going backward in the forms
## Observed behavior

@nucleogenesis has done a bit of this, but this would be worth reviewing.

During the code review, @jredrejo noted that you can use [XState's History](https://xstate.js.org/docs/guides/history.html) to go directly back to a previous state -- however it should be noted that a user's selection from forms are saved to the state machine's context _on transition_ -- so if you go back directly to the previous state the user's selection will not be reflected.

In this case, we should be sure that the user's selection is reflected in the various forms by not using the "history go back" functionality in XState, but rather to ensure that each form populates it's initial value if there is already a selection for its relevant value within the machine's context.

**[Notion Report](https://www.notion.so/learningequality/Bypass-of-credential-screen-in-Import-facility-flow-a8770add596e4c418c9dc29498c88db4?pvs=4) -**
[Another Notion Report](https://www.notion.so/learningequality/A-user-should-not-be-able-to-proceed-past-the-Select-facility-page-if-an-import-address-has-not-be-93ffa32011374777b5e49171c5791b35?pvs=4) -

#### NOTE This is a specifically noted issue to review.
 **LOD -> Create new user:** Once a user has been created, there should not be any more "Back" buttons available to the user - the`OnboardingStepBase -> noBackAction` prop is used to hide the back button.

jredrejo:
## Summary
As done in the change facility state machine , `kolibri/plugins/user_profile/assets/src/machines/changeFacilityMachine.js` back button is managed through a history list used as a stack.
This method makes easier for the wizard to handle to which previous state the user wanted to go back when several states can drive to one same state.
It also helps to keep the consistency in our code using the same method in both machines.

There are a special cases where going back to a previous state requires the execution of some cleaning actions that have been kept as they were in the code, renaming the event to avoid conflict with the general `BACK` event.

## References
Closes: #10240 

## Reviewer guidance
Same as in #10139 paying special attention to the behaviour of the back button.

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/plugins/setup_wizard/assets/src/machines/wizardMachine.js
+++ b/kolibri/plugins/setup_wizard/assets/src/machines/wizardMachine.js
@@ -43,7 +43,7 @@ import { DeviceTypePresets, FacilityTypePresets, LodTypePresets, UsePresets } fr
  */
 
 /* eslint-disable-next-line */
-import { assign, createMachine } from 'xstate';
+import { assign, createMachine, send } from 'xstate';
 
 // NOTE: Uncomment the following function if you're using the visualizer
 // const checkCapability = capabilityToCheck => ["get_os_user"].includes(capabilityToCheck);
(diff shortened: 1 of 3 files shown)


kolibri/plugins/setup_wizard/assets/src/machines/wizardMachine.js
```diff
@@ -68,407 +68,522 @@ const initialContext = {
...
+        actions: [pushHistoryItem],
       },
+      BACK: Object.keys(states).map(state => {
+        return {
+          target: state,
+          cond: context =>
+            context.history.length && context.history[context.history.length - 1] === state,
+          actions: [removeLastHistoryItem],
+        };
+      }),
```
nucleogenesis, line 579: This is nice - love that it's just an array of these `target/cond/actions` objects set here.

kolibri/plugins/setup_wizard/assets/src/machines/wizardMachine.js
```diff
@@ -68,407 +68,522 @@ const initialContext = {
...
+  importLodUsers: {
+    initial: 'selectLodSetupType',
+    states: {
+      selectLodSetupType: {
+        meta: { route: { name: 'LOD_SETUP_TYPE' } },
         on: {
-          CONTINUE: 'createSuperuserAndFacility',
-          BACK: 'requirePassword',
+          // #<name> points to a state w/ an `id` property; wizard is the root
+          BACK_LOD: { target: '#wizard.fullOrLearnOnlyDevice', actions: 'clearFullOrLOD' },
```
nucleogenesis: I think that this one would probably work just fine by going back on the history stack as the action here only sets that value to `null` -- this may not actually be so important as ultimately we probably want the user's selection to be saved when we land on the "Full or LOD" page.

One thing to note is that in that "Full or LOD" is already reading from the context's value OR returning the `FULL` constant -- so it's probably actually better to just use the `BACK` action here.

kolibri/plugins/setup_wizard/assets/src/machines/wizardMachine.js
```diff
@@ -68,407 +68,522 @@ const initialContext = {
...
+  },
 
-      // Facility Creation Path
-      setFacilityPermissions: {
-        meta: { route: { name: 'FACILITY_PERMISSIONS' } },
+  // It's own little baby state machine
+  importFacility: {
+    initial: 'selectFacilityForm',
+    states: {
+      selectFacilityForm: {
```
nucleogenesis, line 398: When I click "Go Back" from here - taking me to the `fullDeviceNewOrImportFacility` state - I can no longer go back from there.

To be clear:
Group Learning -> Full -> Import (and select device + facility) -> Go back -> (can no longer go back from here)

nucleogenesis (changes requested):
One requirement noted in #10240 is that we expect users' selections to be retained as they go back through the forms.

With this implementation, we may need to also maintain the values of particular pieces of context, conditionally.

For example on the first page - the default selection is "On my own". If you click "Group learning" -> Continue -> Go back then the default selection is shown, rather than what the user selected originally.

This isn't a huge UX issue (IMO) for that particular example -- but once the user gets into the "Full Device Creation" flow, they'll be going through the handful of questions and if they go back to re-read something then they shouldn't have their selections reset.

The complication comes in because, particularly in the "BACK_SKIP_FACILITY_FORM" event where we run the `revertFullDeviceImport`, we are resetting the selected device and facility. These context properties are used across multiple flows -- so if a user selects a device and facility to import from for a _full device_ - but realizes they really meant to do _learn only_ - then their previous selection would be automatically kept.

Now that I write that out, it actually seems like it might be best not to clear it out because why would they select the wrong device+facility in one flow, go to the other and do a different flow?

---

In any case, it seems that whether we use history and just pop back to the last one, we then will have to manually maintain the context -- except in the specific cases where we deem it necessary to just go back to where we were context and all (ie, if we want to clear out the user's selected devices when they go back from an import without having already imported anything).

I wonder if we could just have `BACK` and something like `BACK_WITH_CONTEXT` where the first does what it does here, then the latter could apply the context of the _current_ state to the target state's context. 

The pattern I used whereby I only set the context whenever the user clicks "continue" means that the previous state will always nullify the context that the transition to the current state set a value to. So one way or the other, we need to be able to maintain previous context when going back.

jredrejo:
> One requirement noted in #10240 is that we expect users' selections to be retained as they go back through the forms.
> 
> With this implementation, we may need to also maintain the values of particular pieces of context, conditionally.
> 
> For example on the first page - the default selection is "On my own". If you click "Group learning" -> Continue -> Go back then the default selection is shown, rather than what the user selected originally.
> 
> This isn't a huge UX issue (IMO) for that particular example -- but once the user gets into the "Full Device Creation" flow, they'll be going through the handful of questions and if they go back to re-read something then they shouldn't have their selections reset.
> 
> The complication comes in because, particularly in the "BACK_SKIP_FACILITY_FORM" event where we run the `revertFullDeviceImport`, we are resetting the selected device and facility. These context properties are used across multiple flows -- so if a user selects a device and facility to import from for a _full device_ - but realizes they really meant to do _learn only_ - then their previous selection would be automatically kept.
> 
> Now that I write that out, it actually seems like it might be best not to clear it out because why would they select the wrong device+facility in one flow, go to the other and do a different flow?
> 
> In any case, it seems that whether we use history and just pop back to the last one, we then will have to manually maintain the context -- except in the specific cases where we deem it necessary to just go back to where we were context and all (ie, if we want to clear out the user's selected devices when they go back from an import without having already imported anything).
> 
> I wonder if we could just have `BACK` and something like `BACK_WITH_CONTEXT` where the first does what it does here, then the latter could apply the context of the _current_ state to the target state's context.
> 
> The pattern I used whereby I only set the context whenever the user clicks "continue" means that the previous state will always nullify the context that the transition to the current state set a value to. So one way or the other, we need to be able to maintain previous context when going back.

@nucleogenesis Honestly, if this `BACK_WITH_CONTEXT` is a requirement, I'd close this PR discarding it. Adding it would require quite a bunch of code in every page to use the defaults or the context instead (not to mention the needed tests). I don't think it's worth if the current code is already doing it.

kolibri/plugins/setup_wizard/assets/src/machines/wizardMachine.js
jredrejo: cool, that means deleting code, and I always love doing that. Done!

kolibri/plugins/setup_wizard/assets/src/machines/wizardMachine.js
jredrejo: Fixed, the way `BACK` is handled can't work inside a child machine

MisRob:
@nucleogenesis @jredrejo @rtibbles 

Seeing this PR and discussion
and recalling that we stumbled upon the history issue in the change facility flow in the past and spent a good chunk of time discussing and dealing with it (that's what gave rise to the first attempt for some custom history handling that @jredrejo re-introduced here)
and imagining this will be again needed in the future
and the fact that our workflows tend to be pretty complex

it seems to me that it may be helpful to invest some time into implementing history logic that we could use easily in all our machines (with context included). That would also help us to use the same technique everywhere instead of dealing with it for each machine separately. Wondering if we could inspire ourselves here https://github.com/marcelkalveram/xstate-undoable ([source script](https://github.com/marcelkalveram/xstate-undoable/blob/main/src/index.js))

What do you think about it?

MisRob:
> @nucleogenesis Honestly, if this BACK_WITH_CONTEXT is a requirement, I'd close this PR discarding it. Adding it would require quite a bunch of code in every page to use the defaults or the context instead (not to mention the needed tests). I don't think it's worth if the current code is already doing it.

@jredrejo thank you for all the effort here. Seems that this has shown to be more complex than the change facility where I didn't account for context if I recall. Hopefully, we can figure out a way forward.

nucleogenesis:
@MisRob xstate-undoable sounds like a great place to start for sure! I think that it would be a great project and could help us smooth out some rough edges on using XState

nucleogenesis:
Closing as #10660 will serve as a follow-up on making the decisions necessary around all of the above.

PR #10129, opened 2023-02-17
Download Request functionality for resource cards in remote libraries

rtibbles:
## Summary
* Updates the backend to add the admin_imported flag for ContentNodes (ref: #10079) and adds an upgrade task for that
* Updates the backend to return this data from the backend, and annotate this data onto remotely returned resources, so that we have a track of local state for the resources being browsed. Does this without adding additional bloat to the public ContentNode endpoint by creating an inherited mixin for internal usage.
* Creates a skeleton composable for exposing a map of current download requests, and fetching download requests. This will probably need to be embellished to allow for more selective fetches, in the case that the list needs to be paginated.
* Creates a separate HybridLearningFooter component to consolidate our footer information/actions. Includes the ProgressBar, coach content icons, and 'clear bookmark', 'show info' 'download resource', 'remove resource' actions.
* Does a small refactor of HybridLearningContentCard and CardList components to give them the same props API as ResourceCard. Adds a standard 'footer' slot to each card to allow injection of the new footer component.
* Adds a deviceId prop to LibraryPage and TopicsPage to flag whether we are on a remote device or not, uses this to propagate props down to enable or disable download request functionality.
* Also uses this deviceId prop to hide the Resume section on the remote device library page.

## References
Fixes https://github.com/learningequality/kolibri/issues/9857

Note: we will need a follow up issue for properly connecting the composable to the backend once it has been created.

## Reviewer guidance
Any regressions? Anything deleted that should not have been? Are all the footer action buttons working as intended?

Screencast of the new functionality:
[Screencast from 02-17-2023 10:03:58 AM.webm](https://user-images.githubusercontent.com/1680573/219741884-52de6484-4852-4690-bb8a-3f0585da9f0a.webm)

Screenshot of non-remote device:
image

Screencast of bookmarking:
[Screencast from 02-17-2023 10:12:22 AM.webm](https://user-images.githubusercontent.com/1680573/219744838-1ac76d35-e472-4184-ba73-4d7617741a0a.webm)

----

## Testing checklist

- [ ] Contributor has fully tested the PR manually
- [ ] If there are any front-end changes, before/after screenshots are included
- [ ] Critical user journeys are covered by Gherkin stories
- [ ] Critical and brittle code paths are covered by unit tests

## PR process

- [ ] PR has the correct target branch and milestone
- [ ] PR has 'needs review' or 'work-in-progress' label
- [ ] If PR is ready for review, a reviewer has been added. (Don't use 'Assignees')
[...]

PR #10285, opened 2023-03-21
[DO NOT MERGE] Remote download on a content page

MisRob:
WIP

Adds the ability to download remote content from the learning bar on a remote content page

Peek 2023-03-21 18-09

**TODO**

- [ ] Rebase on top of the latest `develop` branch after https://github.com/learningequality/kolibri/pull/10137 is merged (some smaller tweaks may be needed as `useDownloadRequests` interface has been updated

### DIFF
--- a/kolibri/plugins/learn/assets/src/views/TopicsContentPage.vue
+++ b/kolibri/plugins/learn/assets/src/views/TopicsContentPage.vue
@@ -27,6 +27,7 @@
       data-test="learningActivityBar"
       @navigateBack="navigateBack"
       @toggleBookmark="toggleBookmark"
+      @download="handleRemoteDownloadRequest"
       @viewResourceList="toggleResourceList"
       @viewInfo="openSidePanel"
       @completionModal="openCompletionModal"
@@ -208,7 +209,7 @@
       const { fetchLesson } = useLearnerResources();
       const { back } = useContentLink();
       const { baseurl } = useDevices();
-      const { downloadRequestMap } = useDownloadRequests();
+      const { addDownloadRequest, downloadRequestMap } = useDownloadRequests();
       return {
         baseurl,
         canDownload,
@@ -217,6 +218,7 @@
         fetchContentNodeTreeProgress,
         fetchLesson,
         back,
+        addDownloadRequest,
         downloadRequestMap,
       };
     },
@@ -508,6 +510,9 @@
         if (this.$refs.contentPage) {
           this.$refs.contentPage.displayCompletionModal();
         }
+      },
+      handleRemoteDownloadRequest() {
+        this.addDownloadRequest(this.content);
       },
     },
     $trs: {
--- a/kolibri/plugins/learn/assets/src/views/TopicsContentPage.vue
+++ b/kolibri/plugins/learn/assets/src/views/TopicsContentPage.vue
@@ -22,7 +22,8 @@
       :contentProgress="contentProgress"
       :allowMarkComplete="allowMarkComplete"
       :contentKind="content.kind"
-      :showBookmark="isUserLoggedIn"
+      :showBookmark="allowBookmark"
+      :showDownload="allowRemoteDownload"
       data-test="learningActivityBar"
       @navigateBack="navigateBack"
       @toggleBookmark="toggleBookmark"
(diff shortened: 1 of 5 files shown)


MisRob:
@jtamiace @rtibbles @marcellamaki When you look at the recording above, should the bookmark button appear immediately after clicking the download button which only initiates the download request that may take some time to complete? I'd appreciate design feedback as well as technical feedback because I'm not entirely sure that it's fine to bookmark something that hasn't been downloaded yet because the download could potentially fail.

kolibri/plugins/learn/assets/src/views/TopicsContentPage.vue
```diff
@@ -302,6 +312,18 @@
...
         return this.lesson && this.lesson.resources.some(c => !c.contentnode);
       },
+      isRemoteContent() {
+        return Boolean(this.deviceId);
+      },
+      isDownloadedRemoteContent() {
+        return this.isRemoteContent && Boolean(this.downloadRequestMap[this.content.id]);
+      },
+      allowRemoteDownload() {
+        return this.isUserLoggedIn && this.isRemoteContent && !this.isDownloadedRemoteContent;
```
MisRob: @rtibbles I saw that in your [HybridLearningFooter](https://github.com/learningequality/kolibri/pull/10129/files#diff-e985e6e0a76dfd6c79d3937a14c90d2e97d2fa6ff5125cb1d5e860db918cd6b3R154-R158) updates, you also used `admin_imported` flag as one of the conditions to determine whether the download should be allowed. Is it something to be used here too? What would it mean in this context?

rtibbles: Yes - `admin_imported` means that it has been imported to the device by the admin, and in the designs it prevents users from marking resources for download in this case, so you can add an extra conditional for this case.

rtibbles:
Currently bookmarking of resources that are not locally imported will fail too, so if we don't change this behaviour, we will have to fix that.

At the very least, we should provide some sort of loading state to prevent an accidental fat finger click that both requests the download and bookmarks the resource - but if we're doing the interstitial loading state, maybe we should just have the 'ready for download' state until it has actually been downloaded. The main question is whether we have interface strings for explanatory tooltips/aria-labels for that.

marcellamaki:
> The main question is whether we have interface strings for explanatory tooltips/aria-labels for that.

We have 'Waiting to download' in commonCoreStrings which might be useful? 

I think having some sort of intermediary state before the bookmark icon appears might be helpful. And, it might even be worth a short "wait" if the download happens quickly. i.e. at least a couple of seconds before the icon changes? Somehow that feels less jarring to me.

jtamiace:
In terms of how to visually represent this intermediary state, I see three options: a determinate circular loader, indeterminate circular loader, or disabling the download icon. 

Disabling the download icon while it's downloading could be all right if the wait time is only a few moments, but not sufficient feedback for larger files and longer wait times. An indeterminate loader is better at communicating that downloading is happening but leaves questions on how much longer it's going to take.

I think a determinate loader would be the most informative here - is this something that's possible to calculate & display?

kolibri/plugins/learn/assets/src/views/TopicsContentPage.vue
MisRob: Alright, will do

MisRob:
@jtamiace @marcellamaki @rtibbles Thank you all  for your feedback here and on Slack.

@jtamiace 
> I think a determinate loader would be the most informative here - is this something that's possible to calculate & display?

We're not there yet, but it will be possible.

---

To summarize the next steps in regards to the loader:
- In this PR, I will implement an indeterminate loader. There will be a minimal amount of time for the loader to be displayed to prevent too fast UX update from the loader to the bookmark icon when a download is fast.
- I will open a follow-up issue to replace the indeterminate loader with the determinate loader as soon as backend is ready to provide download progress.
- For the loader tooltip, I will use the existing 'Waiting to download' string and will open a follow-up issue to replace it with a new 'Downloading' string

MisRob:
Closing in favor of https://github.com/learningequality/kolibri/pull/10558

PR #10293, opened 2023-03-22
Misc SetupWizard fixes

### LINKED ISSUE #10126: SetupWizard Bugs & Follow-up tasks

nucleogenesis:
## Summary

Misc fixes and clean-up:
- Remove console logs
- Fixed error on super admin form w/ string missing a parameter
- Removed "go back" from personal data consent
- When we have a core error in vuex, onboarding step base will present the error along with a "start over" button

## References

Fixes #10126 

This will close that issue and anything not fixed here in that issue will be made into individual issues (namely the "fix the number of characters allowed on name/device name").

## Reviewer guidance

I need help here -- I thought I could use the `AppError` component here but that's pretty specifically setup for a non setup-wizard context. It _does_ have some messaging that I'd like to hijack so I was considering one of the two following:
- Update the `AppError` component to take some props and configure it to show only some of the more relevant-to-this-context messages (ie, "Something went wrong") -- and then could even update the buttons to say "Start Over" instead
- `crossComponentTranslator` :vomiting_face: 

Any additional thoughts would be appreciated.

NOTE also that there is not a robust use of `handleApiError` action throughout Setup Wizard.

### DIFF
--- a/kolibri/plugins/setup_wizard/assets/src/views/ImportAuthentication.vue
+++ b/kolibri/plugins/setup_wizard/assets/src/views/ImportAuthentication.vue
@@ -60,9 +60,6 @@
         return this.wizardService.state.context.importDevice;
       },
     },
-    mounted() {
-      console.log('ImportAuthentication', this.wizardService.state);
-    },
     methods: {
       callSubmitCredentials() {
         const $credentialsForm = this.$refs.credentialsForm;
--- a/kolibri/plugins/setup_wizard/assets/src/views/OnboardingStepBase.vue
+++ b/kolibri/plugins/setup_wizard/assets/src/views/OnboardingStepBase.vue
@@ -42,7 +42,15 @@
       :noPadding="true"
     >
 
-      <div class="content">
+      <div v-if="coreError" style="padding: 2em;">
+        <h2>{{ coreError }}</h2>
+        <KButton
+          :text="coreString('startOverAction')"
+          :primary="true"
+          @click="startOver"
+        />
+      </div>
+      <div v-else class="content">
         <!-- Optional back arrow to show at the top for longer content views -->
         <KIconButton
           v-if="showBackArrow"
@@ -208,8 +216,15 @@
       selectedLanguage() {
         return availableLanguages[currentLanguage];
       },
+      coreError() {
+        return this.$store.state.core.error;
+      },
     },
     methods: {
+      startOver() {
+        this.wizardService.send('START_OVER');
+        this.$store.dispatch('clearError');
+      },
       /* If the user is focused on a form element and hits enter, continue */
       handleEnterKey(e) {
         e.preventDefault();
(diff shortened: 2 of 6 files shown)


jredrejo (approved):
This looks good to me, but there's a frontend test that's not passing.

Issue #10221, opened 2023-03-09
Library - The breadcrumb navigation for a folder is not being displayed correctly


pcenov:
## Observed behavior
The breadcrumb navigation for a folder is not being displayed correctly

## Expected behavior
The breadcrumb navigation for a folder should be displayed correctly

## Steps to reproduce the issue
1. Install the the latest [0. 16 build ](https://buildkite.com/learningequality/kolibri-python-package/builds/7597) 
2. Import the QA channel and go to Library > QA Channel > African Story Books

## Video

https://user-images.githubusercontent.com/79847249/224100562-8d0f99f8-b2b0-4389-b9ba-41647fe5e013.mp4

## Usage Details
Ubuntu 22  - Chrome, Firefox

pcenov:
@radinamatic

rtibbles:
The TypeError appears to be coming from `KResponsiveWindow` - it's possible that this breakage is then causing the failure to resize other elements, and could be the root of the bug.

shrinishant:
Hey @pcenov, I couldn't find QA channel while searching for import. Could you please let me know how to find that channel?

radinamatic:
@shrinishant  Kolibri QA channel is unlisted, but you can use the token `nakav-mafak` to import it.

shrinishant:
> The TypeError appears to be coming from `KResponsiveWindow` - it's possible that this breakage is then causing the failure to resize other elements, and could be the root of the bug.

@rtibbles @radinamatic @pcenov  i'm not getting any TypeError at my end.

@rtibbles sir, after exploring a bit i found that for this specific QA channel - a css property (top: 64px) is being imposed on the side panel unlike for other channels. Further, that 64px is being dynamically set under stickyCalculation() at line 661 of kolibri/plugins/learn/assets/src/views/TopicsPage/index.vue file.

I tried removing it and then this issue is gone and that sidePanel is no longer overlapping over the breadcrumbs. But, i'm not sure whether this would impact something else or not.

rtibbles:
Fixed in #10298

PR #10298, opened 2023-03-22
breadcrumb navigation bug

thanksameeelian:
## Summary
- addresses issue #10221 

- removes existence check for header that was causing function to return early. months ago, the `stickyCalculation()` function experienced errors because the header was not being targeted correctly. when refactored, the header was correctly targeted but this additional check/catch was also built-in. this early return meant that in certain cases, there was an absence of the sidebar styling that is applied later on in the function, and the `headerBottom` const was sometimes null/undefined rather than being set to `0` in cases where that wasn't the intended behavior.

- there is still an existence check for the header, as part of the variable assignment for `headerBottom`, so that underlying functionality is not lost 

**_BEFORE - when first navigating to the page (before scrolling down)_**
image

_**AFTER - when first navigating to the page (before scrolling down)**_
image

- does not address page skipping described in previous drafts of this PR. after much time spent thinking that the breadcrumb bug and its potential fixes were related to the page jumping up and down bug (screen recordings of the latter bug exist in previous versions of this write-up that you should be able to see by looking at its edits), misha helped me see the two were entirely separate. 
                                  ~~>  _**DEETS FOR THOSE INTERESTED**_: misha and i will be continuing to explore the page jumping bug during a cohacking session on friday, and if our work isn't fruitful we'll be sure to produce a detailed issue about it that potentially anyone could take on. our current notes are on a branch of misha's, but her latest thinking is that the jumping could be solved by effectively not calling `stickyCalculation()` once the header is out of sight & if we're scrolling down (as my thinking is that the jumpy loop may be caused by the calling of that function, it causing changes in how the sidebar is rendered, the function being re-called since changes occurred, then it overwriting those sidebar changes [e.g. because it now recognizes the existence of the header since a DOM change has occurred] with different styling, the function being called since things have changed, and the cycle repeats -- just a theory maybe something like that is happening). we'll be exploring possible solutions surrounding that on friday!

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/plugins/learn/assets/src/views/TopicsPage/index.vue
+++ b/kolibri/plugins/learn/assets/src/views/TopicsPage/index.vue
@@ -595,10 +595,13 @@
       document.documentElement.style.position = '';
     },
     created() {
-      window.addEventListener('scroll', this.throttledHandleScroll);
+      // window.addEventListener('scroll', this.throttledHandleScroll);
       if (this.subTopicId) {
         this.handleLoadMoreInSubtopic(this.subTopicId);
       }
+    },
+    mounted() {
+      window.addEventListener('scroll', this.throttledHandleScroll);
     },
     methods: {
       ...mapActions('topicsTree', ['loadMoreContents', 'loadMoreTopics']),
@@ -711,6 +714,8 @@
       // stickyCalculation() {
       //   const header = this.$refs.header;
       //   const topbar = document.querySelector('.scrolling-header');
+
+      // // make sure you're referencing the element there -- checking here may not be the place it needs to happen
       //   const headerBottom = header ? header.getBoundingClientRect().bottom : 0;
       //   const topbarBottom = topbar ? topbar.getBoundingClientRect().bottom : 0;
       //   if (headerBottom < Math.max(topbarBottom, 0)) {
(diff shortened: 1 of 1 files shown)


kolibri/plugins/learn/assets/src/views/TopicsPage/index.vue
```diff
@@ -647,12 +651,46 @@
...
       stickyCalculation() {
+
+        // // IDEA: first determine we are in the right location/page (since created/mounted is getting called earlier
+        // if we navigate to this page rather than starting on it) tried using presence of a class belonging to another
+        // DOM element on the page or another signifier to ensure page is rendered before header existence is assessed
+
+
+        // OLD CODE replaced 2 months ago -- because header const did not include this.$refs.header.$el,
+        // getBoundingClientRect() couldn't be validly called on it and the function quietly failed and errored out in
+        // the console. until it was addressed, this error coincidentally stopped the "crazy jumping page" bug from
```
marcellamaki: I think you're really onto something here with the assessment of what when wrong in the old code and why. 

One thing I tried as I was reviewing your notes here was just to 

- revert the sticky calculation function to [what it was in 0.15 ](https://github.com/learningequality/kolibri/blob/release-v0.15.x/kolibri/plugins/learn/assets/src/views/TopicsPage.vue#L733)
- test it out
- replicate the problem you describe (yep) 
- and then re-think - okay, so what **fundamentally** is the problem? I'll give you a hint - you already describe it here :) 

Sometimes we fix a bug, but coming back to it from scratch with a different approach or fresh eyes might be more useful than trying to work with what's here right now. You don't necessarily have to work within the "fix" that's in `develop` - you can go back and bring that new perspective and see if something else might work better or more robustly!

If you haven't done so, I'd encourage you to try this approach! I did a little bit of testing, and following your assessment of the situation **_very literally_**, I got a preliminary solution. It would warrant a bit more testing in some of the use cases you describe here to make sure it works in all of the edge cases, but... see if you can find a fix using this approach. Let me know how it goes and if you have more questions along the way!

thanksameeelian: thanks for taking a look! i also pinged you on slack, but i'd love to huddle and chat for a couple minutes about this on monday (or beyond) if you're available!

MisRob (approved):
Thank you, @thanksameeelian, this is looking good to me

PR #10315, opened 2023-03-24
Fix resource import counts

### LINKED ISSUE #10287: Task manager: Import containing duplicate resources results in a mismatch
## Observed behavior
 Successful import of a channel with duplicate resources results in a mismatch in counts of total imports

## Errors and logs
image (17)

## Expected behavior
We should fix this by setting a non-deduped count for the total number of resources downloaded.

## User-facing consequences

Confusion about the number of resources imported, or if the import was successful.
…

## Steps to reproduce
Import a channel with duplicate resources

## Context
[Alpha 7 bug bash](https://www.notion.so/learningequality/Radina-has-previously-reported-this-but-not-sure-we-have-an-issue-anywhere-Successful-import-of-a--4125b0fb17284fed8772bf17e590763f?pvs=4)

rtibbles:
## Summary
* Fixes the import counts for resources

## References
Fixes https://github.com/learningequality/kolibri/issues/10287

## Reviewer guidance
Import the entire Kolibri QA channel. Ensure that it shows that all resources are imported if it is successful.

Screenshot from 2023-03-24 08-07-12

----

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/core/content/utils/import_export_content.py
+++ b/kolibri/core/content/utils/import_export_content.py
@@ -195,7 +195,7 @@ def get_import_export_nodes(  # noqa: C901
         # Only bother with this query if there were any resources returned above.
 
         if nodes_query.count() > 0:
-            nodes_queries_list.append(nodes_query)
+            nodes_queries_list.append(nodes_query.distinct())
 
     return nodes_queries_list
 
--- a/kolibri/core/content/utils/resource_import.py
+++ b/kolibri/core/content/utils/resource_import.py
@@ -232,7 +232,6 @@ class ResourceImportManagerBase(with_metaclass(ABCMeta, JobProgressMixin)):
         self.resources_before_transfer = (
             ContentNode.objects.filter(channel_id=self.channel_id, available=True)
             .exclude(kind=content_kinds.TOPIC)
-            .values("content_id")
             .count()
         )
 
@@ -318,7 +317,6 @@ class ResourceImportManagerBase(with_metaclass(ABCMeta, JobProgressMixin)):
         self.resources_after_transfer = (
             ContentNode.objects.filter(channel_id=self.channel_id, available=True)
             .exclude(kind=content_kinds.TOPIC)
-            .values("content_id")
             .count()
         )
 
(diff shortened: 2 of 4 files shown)


kolibri/core/content/utils/resource_import.py
```diff
@@ -232,7 +232,6 @@ def prepare_for_import(self):
         self.resources_before_transfer = (
             ContentNode.objects.filter(channel_id=self.channel_id, available=True)
             .exclude(kind=content_kinds.TOPIC)
-            .values("content_id")
```
jredrejo, line 235: good cleaning. Happily, Django is smart enough to create the same sql query with or without this line, so same logic with less code :)
I checked it's always this one, with or without `content_id`
`SELECT COUNT(*) AS "__count" FROM "content_contentnode" WHERE ("content_contentnode"."channel_id" = %s AND "content_contentnode"."available" = %s AND NOT ("content_contentnode"."kind" = %s))`

kolibri/core/content/utils/import_export_content.py
```diff
@@ -195,7 +195,7 @@ def get_import_export_nodes(  # noqa: C901
         # Only bother with this query if there were any resources returned above.
 
         if nodes_query.count() > 0:
-            nodes_queries_list.append(nodes_query)
+            nodes_queries_list.append(nodes_query.distinct())
```
jredrejo, line 198: I think this change is totally logical, however, this is breaking the tests.
As `nodes_queries` has an outer join with `File`, it caused repeated appearances of the same node when the node had more than one file in the relation.
Tests need to be changed to reflect the right amount of nodes.

jredrejo (approved):
Code changes are perfect, but tests need to be changed as well to use the correct quantities.

kolibri/core/content/utils/import_export_content.py
rtibbles: Ah thanks - I checked the import export tests, but not the device API tests.

kolibri/core/content/utils/resource_import.py
rtibbles: Yeah, which means that ironically, what this `values` call was meant to be doing (uniqueifying by `content_id`) was actually doing nothing anyway!

pcenov:
@radinamatic no issues observed here - the resource count is correct and the the duplicated resources are being counted as separate items when importing them.

PR #10338, opened 2023-03-28
App mode to getting_started docs

nucleogenesis:
## Summary

Documents the "app mode" startup script in "Getting Started".

## Reviewer guidance

Anything I should add or adjust?

### DIFF
--- a/docs/getting_started.rst
+++ b/docs/getting_started.rst
@@ -297,6 +297,23 @@ and in the second terminal, start the webpack build process for frontend assets:
   yarn run frontend-devserver
 
 
+Running in App Mode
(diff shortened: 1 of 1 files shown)


rtibbles:
Important to highlight that you will need to access Kolibri via the initialize_url that gets logged in the terminal.

I think the main other thing to flag is that the appContext shibboleth remains set in cookies after you have finished running this, so even when you access in the browser from the dev server, it will still register as the appContext until cookies are cleared.

nucleogenesis:
Thanks @rtibbles I had a nagging feeling I'd forgotten something :sweat_smile:

rtibbles (approved):

PR #10347, opened 2023-03-29
Remote connection status

jredrejo:
## Summary
Adds a connection status icon when browsing a remote library content
Peek 2023-03-29 20-13

## References
Closes: #9859 

## Reviewer guidance
As in the attached video, explore a remote library and connect and disconnect the remote device to test if it works fine.

## Testing checklist

- [x] Contributor has fully tested the PR manually
- [x] If there are any front-end changes, before/after screenshots are included
- [ ] Critical user journeys are covered by Gherkin stories
- [ ] Critical and brittle code paths are covered by unit tests

## PR process

- [x] PR has the correct target branch and milestone
- [x] PR has 'needs review' or 'work-in-progress' label
- [x] If PR is ready for review, a reviewer has been added. (Don't use 'Assignees')
- [ ] If this is an important user-facing change, PR or related issue has a 'changelog' label
- [ ] If this includes an internal dependency change, a link to the diff is provided

## Reviewer checklist

- Automated test coverage is satisfactory
- PR is fully functional
- PR has been tested for [accessibility regressions](http://kolibri-dev.readthedocs.io/en/develop/manual_testing.html#accessibility-a11y-testing)
- External dependency files were updated if necessary (`yarn` and `pip`)
- Documentation is updated
- Contributor is in AUTHORS.md

PR #10227, opened 2023-03-10
adds logic to LibraryPage to optionally display a particular device's…

akolson:
## Summary

This PR implements the ability to explore a device's library. Please see https://github.com/learningequality/kolibri/issues/9854 for spec details and figma designs

## References

Closes https://github.com/learningequality/kolibri/issues/9854

## Reviewer guidance

To review;

- Login into Kolibri
- Go to `/learn/#/explore_libraries` via the browser address bar
- On load, click on the `Explore` button of any of the libraries displayed

## Testing checklist

- [x] Contributor has fully tested the PR manually
- [ ] If there are any front-end changes, before/after screenshots are included
- [ ] Critical user journeys are covered by Gherkin stories
- [ ] Critical and brittle code paths are covered by unit tests

## PR process

- [x] PR has the correct target branch and milestone
- [x] PR has 'needs review' or 'work-in-progress' label
- [x] If PR is ready for review, a reviewer has been added. (Don't use 'Assignees')
- [ ] If this is an important user-facing change, PR or related issue has a 'changelog' label
- [ ] If this includes an internal dependency change, a link to the diff is provided

## Reviewer checklist

- Automated test coverage is satisfactory
- PR is fully functional
- PR has been tested for [accessibility regressions](http://kolibri-dev.readthedocs.io/en/develop/manual_testing.html#accessibility-a11y-testing)
- External dependency files were updated if necessary (`yarn` and `pip`)
- Documentation is updated
- Contributor is in AUTHORS.md

PR #10354, opened 2023-03-30
Remote Content Browsing:  Clean the library page when displaying other available libraries

AllanOXDi:
## Summary
This PR clean the library page when displaying other available libraries.

## Reviewer guidance

Please see the figma [design](https://www.figma.com/file/MpCG7r5PULOCJsEvnTSti8/Android-Designs-2022?node-id=194%3A2922&t=CrKIgrjUb27SiVJA-0)
## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated


holta:
Thank you @AllanOXDi

(This appears very promising!)

kolibri/plugins/learn/assets/src/views/LibraryPage/PinnedNetworkResources.vue
```diff
@@ -1,26 +1,34 @@
...
+          <ChannelCardGroupGrid
+            data-test="channel-cards"
+            class="grid"
+            :contents="device.channels"
+            :isPinnedDevice="true"
+          >
+            <ExploreCard
+              :style="cardStyle"
+              class="card-main-wrapper"
+            />
```
marcellamaki, line 27: This looks great! I would suggest updating this to include the <KGridItem> in the slot, wrapped around the explore card, like this 

```
            <KGridItem :layout="{ span: cardColumnSpan, alignment: 'auto' }">
              <ExploreCard
                :style="cardStyle"
              />
            </KGridItem>
``` 

This helps the card go smoothly into the existing grid, and sets the width of the card using the grid, so it changes based on the screen size, just like the other channel cards. 


https://user-images.githubusercontent.com/17235236/229165501-e991493d-bc1a-4237-8679-437493fca10f.mp4



It is not the most straightforward, and it does require adding this computed prop `cardColumnSpan` from [where it is defined in ChannelCardGroupGrid](https://github.com/learningequality/kolibri/blob/a324be1ab3035772d9b5dc7d9e43164e4a6440b1/kolibri/plugins/learn/assets/src/views/ChannelCardGroupGrid.vue#L47) for consistency. (You had been using it before and then cleaned it up. Sorry to ask you to put it back 😄) 

If anyone has a suggestion for a slightly less messy approach (cc @akolson @MisRob ?), I would be happy for you to recommend something here to Allan that might be better :) This was just the best way I could figure out adding the card into the grid without changing that component too much

kolibri/plugins/learn/assets/src/views/LibraryPage/index.vue
```diff
@@ -1,11 +1,9 @@
 <template>
 
   <LearnAppBarPage
-    :appBarTitle="appBarTitle"
+    :appBarTitle="learnString('learnLabel')"
```
marcellamaki, line 4: I'm not sure why this is changed?

kolibri/plugins/learn/assets/src/views/LibraryPage/index.vue
```diff
@@ -49,63 +47,82 @@
...
-                </KButton>
-                <KIcon icon="disconnected" />
-              </p>
-              <p v-else>
+            <KGridItem
+              :layout12="{ span: 4 }"
+              :layout8="{ span: 4 }"
+              :layout4="{ span: 4 }"
+            >
+              <div v-if="searching" style="padding-top:20px">
```
marcellamaki, line 66: You do have this condition here, which is great, but right now it is just set to `True` always. Maybe @jredrejo's recent PR with the device connection status can be helpful to reference once it is merged, and the same logic can be used here. 

The string that is displayed should change based on whether Kolibri is looking for devices on the network, or displaying devices.

The styles for this text should be 
```
fontSize: '12px',
color: this.$themeTokens.annotation,
```

kolibri/plugins/learn/assets/src/views/LibraryPage/index.vue
```diff
@@ -49,63 +47,82 @@
...
-            <KGridItem :layout="{ span: 6 }">
-              <span>
-                <h1>
-                  {{ $tr('otherLibraries') }}
-                </h1>
-              </span>
+        <div>
+          <KGrid>
+            <KGridItem
+              :layout12="{ span: 8 }"
```
marcellamaki, line 53: I think keeping this with `{ span: 6 }` as you had makes the spacing closer to the spec.

kolibri/plugins/learn/assets/src/views/LibraryPage/index.vue
```diff
@@ -49,63 +47,82 @@
...
+                      {{ coreString('refresh') }}
+                    </KButton>
+                  </div>
+                  <div>
+                    <KIcon
+                      icon="wifi"
+                    />
+                  </div>
+                </div>
+              </div>
```
marcellamaki, line 82: This could be simplified to have just slightly fewer `<div>`s 

```
<div v-if="searching" class="sync-status">
     <span>
         {{ $tr('searchingOtherLibrary') }}
      </span>
      <div class="network-device-refresh">
      <KButton appearance="basic-link" >
             {{ coreString('refresh') }}
        </KButton>
         <KIcon  icon="wifi" />
     </div>
</div>
```

I also started playing around with classes regarding the spacing. You don't have to add exactly this, but here are some ideas to get you going: 

```
  .sync-status {
    position: absolute;
    right: 16px;
    top : 40%;
  }

  .network-device-refresh {
    display: inline-block;
    margin: 0 4px;
  }
```

To have the `position: absolute` work in the right place (within the parent), the parent container - in this case the `KGridItem` also needs a position. I tried adding 
```
.sync-display {
    position: relative;
  }
```

There are three things I am trying to achieve with this code: 
- When there is space enough on the screen, keep the text all in one line with all items in a row. Try to not wrap the text, even if the items are adjacent: 
Screenshot 2023-03-31 at 12 24 24 PM

- As the window gets smaller, have the "Refresh" with the icon "break" together, meaning that the icon and the "Refresh" text always move together
- Keep this information aligned to the right of the grid

I think what I have is a start, but I encourage you to play around with it a little bit yourself!

marcellamaki: My second thought here is actually a question for @radinamatic - I am now thinking for both this display, and for @jredrejo's [connection status display PR](https://github.com/learningequality/kolibri/pull/10347), we might want to have an 
`aria-live="polite"` and an extra label that conveys the "meaning" of the wifi icon. (It is clearer in José's PR, because the text says "Disconnected" when it appears and he has added the `ariaLabel` to the icon button, but we may still need the "interruption added"). 

For this scenario, there are three possible strings to display based on connection status: 
- Searching for libraries around you.
- No other libraries around you right now
- Showing all available libraries around you.

Unfortunately I did not think about specific wording here, but we do have various connection status indicators, including `'Successfully reconnected!'` elsewhere in the codebase that could perhaps be repurposed. 

I am imagining something like the following possible combinations as the "announcement" (these are all existing strings) - maybe both included in the "interruption" but the first part associated with the icon and the second part would be the relevant text?
1. "Your device seems to be disconnected." "Searching for libraries around you." 
2. "Successfully reconnected!" "No other libraries around you right now"
3. "Successfully reconnected!" "Showing all available libraries around you."

kolibri/plugins/learn/assets/src/views/LibraryPage/ExploreCard.vue
```diff
@@ -0,0 +1,81 @@
...
+
+  import commonCoreStrings from 'kolibri.coreVue.mixins.commonCoreStrings';
+  import { PageNames } from './../../constants';
+
+  export default {
+    name: 'ExploreCard',
+    mixins: [commonCoreStrings],
+    computed: {
+      exploreLibraries() {
+        return { name: PageNames.EXPLORE_LIBRARIES };
```
marcellamaki, line 31: I think you may want to use a link that goes directly to the library of this device, similar to what @akolson uses within the `LibraryItem`

```
      libraryPageRoute() {
        return {
          name: PageNames.LIBRARY,
          params: {
            deviceId: this.deviceId,
          },
        };
      },
```

To me, if the user is clicking on the explore within a particular library, it seems helpful to bring them to that specific library, rather than the entire "explore libraries" page.

kolibri/plugins/learn/assets/src/views/LibraryPage/index.vue
```diff
@@ -424,4 +480,18 @@
...
+  .view-all-text {
+    padding: 60px;
+    margin-left: 60px;
+    font-size: 20px;
+    font-weight: bold;
+    text-align: center;
+  }
+
+  .view-all-card {
+    width: 370px;
```
marcellamaki, line 492: This width is overriding the use of the grid. You can remove it. I do think it would be good to add a _height_ however - the max height of the cards in figma is 80px. This should be updated here and in the `UnpinnedDevice` cards.

kolibri/plugins/learn/assets/src/views/LibraryPage/UnPinnedDevices.vue
```diff
@@ -1,24 +1,34 @@
...
+              :text="deviceName"
+              :maxHeight="52"
+            />
+          </span>
+        </h2>
+        <div class="channels">
+          <p v-if="channels">
+            {{ channels }} {{ $tr('channels', { count: channels }) }}
+          </p>
+        </div>
```
marcellamaki: This is another place where the number of `div`s can be reduced. 

The spacing and alignment of cards is tricky to get right. Here is how I thought about this card (please excuse my terrible penmanship): 

Screenshot 2023-03-31 at 1 32 46 PM
Screenshot 2023-03-31 at 1 05 40 PM

- First, you can set a specific height on your card, based on the spec (80px)
- The next thing I looked for was how the card was divided up. I see a few different sections: 
1. The left part of the card has the icon in it, only. 
2. The other piece of the card has text, and the text is left-aligned 
3. Using the text truncation is a great idea 👍 
4. I think using an absolute position for the "Channels" text would make the cards look cleaner. By this I mean, the channels text should always appear in the same bottom part of the card, whether the title portion is one or two lines. Using absolute positioning can help with this.

marcellamaki (changes requested):
Hi @AllanOXDi - this PR has some _amazing_ improvement. The accurate channels are displaying, you've done some nice code cleanup, and it's definitely getting close to being done. 

From my perspective, there are three remaining things to address. 
1. Polishing up the UI. You're close, but I know we can make the designers 😍 by getting it pixel perfect. I've added some suggestions - including some things that are more general suggestions and some specific code recommendations to try. Cards in particular are _hard_. I would suggest you take a first attempt at implementing some of this feedback, but we can have a cohack early next week for anything you get stuck on and work through it together. (It may just be me googling "How to do ____ with CSS" 😄 
2. A few accessibility questions. These could be done in follow up, but I don't want to neglect them. I've tagged @radinamatic for her input here. 
3. There are some routing things that I am a bit confused by. I think we might be missing a page - the page that is neither the "Explore Libraries" page, nor actually a "Topic" or content page _within_ a remote library, but the page for that device's library itself. For example "Richard's Classroom Library". This is what should open when you click the "Explore" button, either here, or from the "Explore Libraries" Page. I'm not clear on whether this still has to be implemented, or if I don't know what the correct route is that I should recommend to you. Do you or @akolson have any thoughts here? It's totally fine if it is still on the to-do list and I am happy to help with the implementation as well. 

Screenshot 2023-03-31 at 1 47 25 PM

kolibri/plugins/learn/assets/src/views/LibraryPage/PinnedNetworkResources.vue
```diff
@@ -107,9 +109,10 @@
 
     position: relative;
     display: inline-flex;
-    width: 100%;
-    max-height: 258px;
+    width: 350px;
```
marcellamaki, line 112: I think you should be able to remove all of the styles on this page, as they seem to be only used for the `Explore` card? If so, anything for that care can be defined in that file. It is redundant and makes it a little bit confusing to debug, and things might be overwriting each other in confusing ways.

marcellamaki:
> There are some routing things that I am a bit confused by. I think we might be missing a page - the page that is neither the "Explore Libraries" page, nor actually a "Topic" or content page within a remote library, but the page for that device's library itself. For example "Richard's Classroom Library". This is what should open when you click the "Explore" button, either here, or from the "Explore Libraries" Page. I'm not clear on whether this still has to be implemented, or if I don't know what the correct route is that I should recommend to you. Do you or @akolson have any thoughts here? It's totally fine if it is still on the to-do list and I am happy to help with the implementation as well.

I just checked in with @rtibbles about this and he will add this page in follow-up, based on his previous work, which is just a conditionalized display of the existing library page. Sorry for my confusion here!

akolson:
Hi @marcellamaki,  I think this page already exists. Actually we reused the Library page to create this intermediary page in https://github.com/learningequality/kolibri/pull/10227/ and that should cover it. cc @rtibbles. Below is an example
image

kolibri/plugins/learn/assets/src/views/LibraryPage/PinnedNetworkResources.vue
MisRob: Yes, I think that generally speaking, using grid items within a grid ensures consistency, that's how grid components are intended to be used.

MisRob: It may be simpler to wrap the newly added slot in the `ChannelCardGroupGrid` in `KGridItem`:

```
<KGridItem>
  <slot></slot>
</KGridItem>
```

(however it depends on how that new slot was intended, I'm not familiar with the whole pull request)

kolibri/plugins/learn/assets/src/views/LibraryPage/index.vue
AllanOXDi: I am note sure too! I am cross checking.

kolibri/plugins/learn/assets/src/views/LibraryPage/index.vue
AllanOXDi: The connection states should  be dressed  by @jredrejo PR, If am not mistaken. cc @akolson  for more guidance on this.

kolibri/plugins/learn/assets/src/views/LibraryPage/ExploreCard.vue
AllanOXDi: That was my initial thought. But when I was chatting with @akolson about this. He suggested that we should route to explore Library page. If I got him right. we could make a follow up issue to address this in a different PR. As this PR is just to display the remote content on the library page?

akolson: Hi @AllanOXDi, I think I didn't understand the context of your question, my apologies, but Marcella is right about how this should be implemented and swapping out the current `exploreLibraries` property for the `libraryPageRoute` property as suggested Marcella should do the trick without having to open a different PR as i think we already have a `devivcId` prop within `LibraryPage` component. Please confirm this if you get to proceed with this :)

AllanOXDi: Great! Thanks, I have updated it.

kolibri/plugins/learn/assets/src/views/LibraryPage/index.vue
akolson: Hmm, not entirely.  Jose's PR checks for a particular device's availability. I think we would need to do more to achieve the other states that Marcella mentions here. Also, 2. and 3. are a consequence of successfully reconnecting back to a particular device and a successful reload of libraries within that device

kolibri/plugins/learn/assets/src/views/LibraryPage/PinnedNetworkResources.vue
marcellamaki: > It may be simpler to wrap the newly added slot in the ChannelCardGroupGrid in KGridItem

Yes, that's a good idea @MisRob!

AllanOXDi:
Review feedbacks resolved in #10420

Issue #10363, opened 2023-03-31
Remove vendored ipaddress module when we drop Python 2.7 support


akolson:
## Summary

In order to better handle functionalities associated with ip addresses, there is need to implement a utility library that can be used across the platform.  [`ipaddress`](https://github.com/python/cpython/blob/main/Lib/ipaddress.py) has been vendored however, it only supports python 3+. 

As such,  this task is to remove the vendored `ipaddress`, which we added to provide support for `python 2.7`.

rtibbles:
I think we can probably let this lie, and just focus on unvendoring it when we drop Python 2.7 support.

rtibbles:
Fixed in https://github.com/learningequality/kolibri/pull/11654

PR #10364, opened 2023-03-31
Add is local attribute

### LINKED ISSUE #10220: Add `is_local` attribute to `NetworkLocation`
Depends on: https://github.com/learningequality/kolibri/pull/10133

## Summary
In order to support 0.16 feature work, we need a way to determine what devices (aka `NetworkLocation` model records) are on the local network versus on the internet. With the above PR, we'll start tracking the remote IP address of the device, which we can use to determine if it's local or on the internet.

## Desired behavior

- The `NetworkLocation` model should have a new boolean field `is_local`
- For static `NetworkLocation`s, the function `capture_network_state` should set `is_local` using the information `client.remote_ip`-- this should support IPv4 and IPv6 addresses
- For dynamic `NetworkLocation`s `is_local` should always be `True`
- For py2.7 support, the py3+ standard lib `ipaddress` could be vendored ([discussion](https://learningequality.slack.com/archives/CB37UM23A/p1676048338939419))

akolson:
## Summary

This pr adds the `is_local` attribute to the `NetworkLocation` model. To provide functionality associated with the `is_local` attribute, the `ipaddress` library has been vendored. 

`python2.7+` support has been added for the vendored `ipaddress` and also unnecessary functionality has been removed so that only functionality that suits our uses cases remains.

 Due to the spaghetti like nature of the code, only a few parts of the code have been purged. There is need to invest more time in creating a version that better suits our use cases. A follow up issue has been created https://github.com/learningequality/kolibri/issues/10363

## References

Closes https://github.com/learningequality/kolibri/issues/10220

## Reviewer guidance

The `is_local` attribute should be returned when accessing `networklocation` data. 

----

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated


nucleogenesis (approved):
Kolibri code LGTM. 

I think that it might be worthwhile to put a link to that open issue in the ipaddress.py file with a bit of context around why we needed to vendor it in the first place (ie, is this for 2.7 compatibility, what's lacking with [ipaddress](https://docs.python.org/3/library/ipaddress.html) that this vendored copy makes up for)

kolibri/core/discovery/utils/network/client.py
```diff
@@ -77,6 +76,8 @@ def build_for_address(cls, address, timeout=None):
 
     @classmethod
     def build_from_network_location(cls, network_location, timeout=None):
+        from kolibri.core.discovery.models import ConnectionStatus
```
bjester: What prompted this change?

kolibri/core/discovery/models.py
```diff
@@ -118,8 +124,17 @@ class Meta:
...
+        self.is_local = self.capture_network_state()
         return super(StaticNetworkLocation, self).save(*args, **kwargs)
 
+    @classmethod
+    def capture_network_state(self):
+        try:
+            client = NetworkClient.build_from_network_location(self)
+            return ip_address(client.remote_ip).is_private
+        except (AddressValueError, ValueError):
+            return False
```
bjester: Would you please add this functionality to [this function `capture_network_state`](https://github.com/learningequality/kolibri/blob/develop/kolibri/core/discovery/utils/network/connections.py#L33) instead?

bjester (changes requested):
One requested change, but otherwise looks good :+1:

kolibri/core/discovery/utils/network/client.py
akolson: I ran into circular dependency issues so was just a way to circumvent the issue. But I think this should be resolved with the changes you requested above, so i will revert.

kolibri/core/discovery/models.py
```diff
@@ -117,7 +120,16 @@ class Meta:
         proxy = True
 
     def save(self, *args, **kwargs):
+        from kolibri.core.discovery.utils.network.connections import (
+            capture_network_state,
+        )
+        from kolibri.core.discovery.utils.network.client import NetworkClient
```
akolson: The two imports are causing circular dependency bugs when imported from the top of the file, so they have been added as part of the function to solve the issue.

kolibri/core/discovery/models.py
```diff
@@ -117,7 +120,16 @@ class Meta:
...
+
         self.dynamic = False
+
+        client = NetworkClient.build_from_network_location(self)
+        self.is_local = capture_network_state(self, client)
```
bjester: @akolson This wasn't what I meant by my previous requested change, but I can guess why you went about this and it's good motive. 

Of course, we need to make a request in order to determine the IP address, and therefore the value for this new boolean field. Although, doing this within the model's save method isn't acceptable because it could generate HTTP requests and block the save. It could also throw errors if the location isn't contactable. The circular dependency issues are red flags that should alert you that something isn't meshing right. The relationship between the two operate in reverse-- the function `capture_network_state` is called with the `NetworkLocation` object and it does so elsewhere in the code where we have dedicated processing for updating network location data such as this.

So your changes to `capture_network_state` should set the value on the `NetworkLocation` object inside of it, and the function shouldn't return anything. Doing so will eliminate these circular dependencies. You'll also likely need to update the tests for `capture_network_state`

bjester (changes requested):

kolibri/core/discovery/models.py
akolson: Thanks Blaine for the clarification. Changes have been implemented as requested

kolibri/core/discovery/test/test_connections.py
```diff
@@ -141,6 +141,12 @@ def test_okay__dynamic(self):
         update_network_location(self.mock_location)
         self.assertNotEqual(self.mock_location.last_known_ip, "192.168.101.101")
 
+    def test_okay__static(self):
+        self.mock_location.dynamic = False
+        update_network_location(self.mock_location)
+        self.assertEqual(self.mock_location.last_known_ip, "192.168.101.101")
+        self.assertIn(self.mock_location.is_local, [True, False])
```
bjester: Seems like "192.168.101.101" should always be a local IP address?

akolson: It doesn't seem to be when run on `python 2.7`. Maybe  underlying issue with some parts of the porting? Let me try to fiddle with this and revert.

bjester: I saw [this issue](https://github.com/learningequality/kolibri/issues/10363) you added. The main reason to vendor the library is to support 2.7, which I'm sure required you to make compatibility tweaks to the vendored code. But if the functionality that we need isn't working in 2.7, then that seems problematic!

akolson: Issue has been resolved and test made more specific (`self.assertTrue(self.mock_location.is_local)`)

bjester (approved):
Nice work figuring out the 2.7 compatibility issues, @akolson!

PR #10365, opened 2023-03-31
Automatic content import / removal initiated by facility syncing

### LINKED ISSUE #10079: Remote Content Browsing: Create support for user download requests, download management, and garbage collection
## Overview 

This issue is an overview issue that connects support for user download requests, as well as the management of downloaded resources (whether user initiated or auto-synced) as well as the necessary garbage collection to support automatic syncing. 

In the MVP, we will **not** do complex resource seeking, to handle "I was downloading this from Person A. Person A went offline. Person B has the same resource. Download this resource from Person B". However, we may want to implement a setting where, if there is bandwidth and the user is allowing us to use it, we could just find the resource on Studio and download it directly from there. 

### Figma 

Please see [complete figma spec ](https://www.figma.com/file/MpCG7r5PULOCJsEvnTSti8/Android-Designs-2022?node-id=194%3A2922&t=yAtcmSpF4FxfLCXa-0) for front end references that may be a helpful visual reference to support the requirements and acceptance criteria.

## Acceptance Criteria

- [x] Create `DownloadRequests` model, which contains all user downloads. It should have a foreign key to the user who initiated the download, the reason (i.e. user initiated, sync, etc.), the facility ID (to support cleanup), the status of the request (i.e. started, failed, etc.),  date added, and all resource details (title, file size, learning activities) so that the metadata can still be displayed to a user after the resource itself is no longer locally present
- [x] This should live in the Device app (not plugin) or its own core app (@bjester: note I added it to the content app)
- [x] A flag should exist that designates admin imports vs. other imports (user initiated, synced). Admin imports are NOT garbage collected. 
- [x] When upgrading, any existing resources on the device before 0.16 should be flagged to an admin import
- [x] Implement a garbage collection mechanism, which is triggered periodically, or when necessary to clean up learner devices to make space for new content that needs to be synced.
- [x] When doing garbage collection, learner downloads should leave first. If there is still insufficient room, the coach should receive the insufficient storage notification, prompting them to remove stale resources. (Please note, we may need to solidify how much of this happens automatically, and how much is coach-determined at this phase).
- [ ] See [prioritization in Notion](https://www.notion.so/learningequality/Data-and-content-architecture-proposal-8e03a6da67394c578d40336b7a7953bf#62e26c23d9f946a3a5e2edd1f0347122) for more details on how resource cleanup should be prioritized
- [ ] Use the Tasks API for automated retries
- [ ] Create an API endpoint to query "Is there more than just my delete request for this resource" to manage a "remove download" scenario for 2+ user accounts on the same device, where they have downloaded the same resource, and one person wants to remove it.

### Out of scope
Implementation of the "My Downloads" page, which is covered by https://github.com/learningequality/kolibri/issues/9861

## References 
- [Figma spec](https://www.figma.com/file/MpCG7r5PULOCJsEvnTSti8/Android-Designs-2022?node-id=194%3A2922&t=zDk6ybjN91p5hH6m-0)
- [Gherkin scenarios](https://www.notion.so/learningequality/Gherkin-scenarios-02096ac9925844f8902c818cab607d4b#0a205e75b2ea49a9be07acf15bf13b7a)

bjester:
## Summary

- Adds ability to assign priority to Kolibri sync hook operations (this update is worthwhile even though I implemented this with intermediary changes, but with refactoring it became less vital to the new functionality in this PR)
- Sets priority for existing lessons and exams sync hooks to be prioritized first in the cleanup sync stage
- Refactors the existing `pre_transfer` and `post_transfer` sync hooks to operation outside of sync signals so that we can hook into syncs outside of transactions wrapping portions of the sync in the `sync` command
- Updates `NetworkClient` exceptions for `HTTPError` to include response object for determining the status code, either 4xx or 5xx in this scenario
- Adds `ContentAssignmentManager` model manager that can be attached to models for determining models affected by a sync and parsing associated content dependencies from them, initially attached to `Lesson`, `Exam`, `IndividualSyncableLesson` and `IndividualSyncableExam`
- Adds new `ContentRequest` model and proxy models `ContentDownloadRequest` and `ContentRemovalRequest` for tracking both sync and user initiated content requests
- Adds content plugin sync hook that creates `ContentRequest` models from content dependencies parsed out of `ContentAssignmentManager` usages during the cleanup stage of a sync
- Adds task and processing for `ContentRequest`s to automatically import content and remove content as necessary

### Functionality notes
- Either the device is unprovisioned or the device setting to allow 'Auto-downloads' is true in order to enable automatic content import
- The content sync hook cleanup operation generates corresponding content download and removal requests using the records modified by a sync
- The content sync hook `post_transfer` imports content metadata in order to accurately determine the storage usage of necessary imports
- After completing the import of metadata, the hook enqueues a task to perform the actual import of content files
- The task will re-attempt to import metadata if necessary (there exists requests without metadata)
- The task processes content download requests one by one, attempting to download as many resources as possible within remaining free space
- When there are pending download requests but no requests would fit in remaining free space, the task then processes removals in this order, repeating download attempts in between until all download requests have been completed:
  1) Sync initiated removal requests
  2) User initiated removal requests
  3) User initiated downloads without explicit removal requests
- If all downloads cannot be accommodated after processing removals and user initiated downloads, an insufficient storage status is saved to `LearnerDeviceStatus`

## References

https://github.com/learningequality/kolibri/issues/10019
Resolves https://github.com/learningequality/kolibri/issues/10079

### Followup
https://github.com/learningequality/kolibri/issues/10378
https://github.com/learningequality/kolibri/issues/10379
https://github.com/learningequality/kolibri/issues/10381
https://github.com/learningequality/kolibri/issues/10382
https://github.com/learningequality/kolibri/issues/10385
https://github.com/learningequality/kolibri/issues/10386

## Reviewer guidance

…

----

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated


kolibri/core/content/tasks.py
```diff
@@ -275,7 +276,7 @@ class RemoteResourceImportValidator(ResourceNodeValidator):
     )
 
     def validate(self, data):
-        job_data = super(RemoteImportMixin, self).validate(data)
+        job_data = super(RemoteResourceImportValidator, self).validate(data)
```
rtibbles, line 279: Ah yes, that is rather an important tweak!

kolibri/core/content/tasks.py
```diff
@@ -335,6 +336,18 @@ def remoteresourceimport(
     import_manager.run()
 
 
+@register_task(
+    queue=QUEUE,
+    long_running=False,
```
rtibbles: It might be safer to set this to `True` - I am still a little iffy about how exactly we should be determining this, but I tend to assume that importing resources might always take > 10 minutes on some occasions.

kolibri/core/content/utils/channel_import.py
```diff
@@ -50,7 +50,10 @@
 models_not_to_overwrite = [LocalFile]
 
 models_to_exclude = [
-    apps.get_model(CONTENT_APP_NAME, "ChannelMetadata_included_languages")
+    apps.get_model(CONTENT_APP_NAME, "ChannelMetadata_included_languages"),
```
rtibbles, line 53: Doesn't need to be fixed here, but this is making me realize we should probably be accounting for these `models_to_exclude` when we do the SQLAlchemy schema generation. No need to generate a schema for models we are not importing from or to.

kolibri/core/content/utils/content_request.py
```diff
@@ -0,0 +1,525 @@
...
+        logger.warning(str(e))
+
+
+def _get_import_metadata(client, contentnode_id):
+    """
+    :type client: NetworkClient
+    :type contentnode_id: str
+    :rtype: None|dict
+    """
+    url_path = reverse_path(
```
rtibbles, line 247: Another thought about possible future refactors:

Wondering if we should defer all remote requests like this to a client object. (rather than several places where we use the reverse_remote utility, and then just use requests).

kolibri/core/content/utils/content_request.py
```diff
@@ -0,0 +1,525 @@
...
+        else len(contentnode_ids)
+    )
+    processed_count = 0
+    logger.info("Importing content metadata for {} nodes".format(total_count))
+    for contentnode_id in contentnode_ids:
+        import_metadata = _get_import_metadata(client, contentnode_id)
+        # if the request 404'd, then we wouldn't have this data
+        if import_metadata:
+            processed_count += 1
+            with db_lock_sqlite_only():
```
rtibbles: If we remove this context manager, does the error still happen?

Wondering if we should have some thread local property for `db_lock_sqlite_only` that prevents it trying to make a lock if it's already inside an existent lock?

kolibri/core/content/utils/content_request.py
```diff
@@ -0,0 +1,525 @@
...
+            )
+
+        # delete any related download requests
+        related_downloads.delete()
+
+    # avoid adding extra logs
+    if not new_contentnode_ids:
+        return
+
+    # TODO: we can't yet do this during a sync because the sync command creates a lock / transaction
```
rtibbles: Hrm... odd, the db_lock doesn't seem to do anything special to make sure the transaction is shared between Django and SQLAlchemy: https://github.com/learningequality/kolibri/blob/develop/kolibri/core/utils/lock.py#L37

Not sure what is different about the transaction created during the sync - it should be fine to created a nested transaction within that parent transaction.

kolibri/core/content/utils/content_request.py
```diff
@@ -0,0 +1,525 @@
...
+        reason=ContentRequestReason.UserInitiated
+    )
+    complete_user_downloads = ContentDownloadRequest.objects.filter(
+        status=ContentRequestStatus.Completed, reason=ContentRequestReason.UserInitiated
+    )
+
+    # loop while we have pending downloads
+    while incomplete_downloads_with_metadata.exists():
+        # grab the next request that will fit within current free space
+        free_space = get_free_space()
```
rtibbles: `get_free_space` defaults to the KOLIBRI_HOME dir which by default will be on the same drive as the content directory, but is configurable to be elsewhere.

Should be:
```
free_space = get_free_space(OPTIONS["Paths"]["CONTENT_DIR"])
```

kolibri/core/content/utils/content_request.py
```diff
@@ -0,0 +1,525 @@
...
+def process_content_removal_requests(queryset):
+    """
+    Garbage collects content requests marked for removal (removed_at is not null)
+
+    :param queryset: a ContentRemovalRequest queryset
+    :type queryset: django.db.models.QuerySet
+    """
+    # exclude admin imported nodes
+    removable_nodes = ContentNode.objects.filter(
+        Q(admin_imported__isnull=True)
```
rtibbles: This should only be happening in develop environments, as the upgrade handler should prevent this ever actually being null - just in case this has performance implications.

rtibbles (commented):
A couple of thoughts, a few tweaks, but overall from a code read through, I understand what is happening, and I like it!

kolibri/core/content/utils/channel_import.py
bjester: I'll look into this

kolibri/core/content/utils/content_request.py
bjester: Does the SQLAlchemy side use a separate DB connection? That was my assumption for why it was blocking, for the reason that sqlite uses serializable transactions, so only one connection's transaction can execute at a time. If it does use the same connection, or if we can make it do that, my guess is that is shouldn't block like you said.

kolibri/core/content/utils/content_request.py
bjester: I'll need to test this again. With fresh eyes, I'm guessing this was still blocking the post-sync import of metadata for the same reason the sync caused it.

kolibri/core/content/utils/content_request.py
bjester: Yeah since the client tries to set appropriate request timeouts based off context, that would be ideal. I had looked into making those timeouts global somehow, but it seemed that required runtime patching of code in `requests`.

kolibri/core/content/utils/content_request.py
rtibbles: Yes, it should be using the same connection, precisely to allow the db lock to be used - we use a special connection generator for SQLAlchemy that references the Django connection.

bjester: okay, maybe I'll try the "thread local property" approach you mentioned

bjester: So this change seems to have fixed it. It appears it wasn't sharing the connection after all, but looking at tests, this seemed like the appropriate change
https://github.com/learningequality/kolibri/pull/10365/commits/1baed22525776e5353ab5fde7ad78187e84f47fa#diff-76f6082e35fb73d50fa01980aa366b8455b9c2488734ab0d985582ab11a584bcR110

kolibri/core/device/models.py
```diff
@@ -179,6 +180,35 @@ def secondary_storage_locations(self):
...
+        Allows use with `get_device_setting` utility
+        :param name: A str name of the `extra_settings` field
+        :return: mixed
+        """
+        return self.extra_settings.get(name, extra_settings_default_values[name])
+
+    @property
+    def allow_download_on_metered_connection(self):
+        # TODO: fix typo
+        return self._get_extra("allow_download_on_mettered_connection")
```
rtibbles: This has been fixed on develop, so we can update this before merge.

```suggestion
        return self._get_extra("allow_download_on_metered_connection")
```

rtibbles (approved):

PR #10380, opened 2023-04-04
Policies plugin initialization fixes

nucleogenesis:
## Summary

The Policies plugin was not loading as it was missing the urls configuration in the plugin setup.

Additionally the CookiePolicy component was not loading due to a string missing its parameters (which took far too long to identify thanks to Chrome devtools' useless traces...) and both components needed to import ImmersivePage.

This plugin sets up for follow-up work on the front-end, namely:

- Figuring out how to get the CookiesBanner hooked into wherever we want it
- Handling the `route` prop of `ImmersivePage` so that we can send the user back to where they were reliably.
- Some styling issues

## References

#10148 - won't put a closing keyword as there is follow-up to do

## Reviewer guidance

Anything that should be included here rather than in a follow-up?

### DIFF
--- a/kolibri/plugins/policies/assets/src/app.js
+++ b/kolibri/plugins/policies/assets/src/app.js
@@ -1,4 +1,3 @@
-import router from 'kolibri.coreVue.router';
 import PageRoot from 'kolibri.coreVue.components.PageRoot';
 import routes from './routes';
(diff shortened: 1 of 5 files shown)


rtibbles (approved):
This all looks sensible to me - we can merge and keep iterating in develop.

Issue #10326, opened 2023-03-27
SetupWizard > On My Own -- Handle username and device name length validation errors


nucleogenesis:
## Observed behavior

During the "On my own" flow, the user's device name is automatically generated to something like "Personal Device for <the users' full name>". 

An issue arises wherein the `device name` can only be up to 50 characters long -- but the user's `full_name` can be up to 120 which can easily result in the device name being too long during provisioning.

## Expected behavior

_Idea 1: Interpolation_

@rtibbles suggestion (if I understand correctly) - is that we would actually store the user's FacilityUser ID in the string and then use JS on the front-end to interpolate the user's full name in the ID's place when rendering the string. This is a step in a complex direction and, considering the UUID is 32 characters long, that only leaves 18 characters for the rest of the translated strings which could still lead to an issue.

One concern with the idea of _increasing the validation's limit of 50 characters_ on the device name is that we may run into issues with ZeroConf / discovery stuff -- but that concern was discussed in the context of making it 170 characters (or whatever -- to accommodate the longest possible) -- _however_ - we may find that increasing it by 20-30 characters could accommodate the solution above.

_Idea 2: Truncation_

Rather than handling this on the backend at all, we may instead opt to just save a truncated version of the user's full_name, if needed, to set the device name to the "Personal device for ..." such that it will only ever be 50 characters long. Something like:

```
let baseTranslatedLength = this.$tr('defaultDeviceName', { name: '' }).length
let truncatedFullName = this.user.full_name.slice(0, baseTranslatedLength-1);
let actualStringWeUse = this.$tr('defaultDeviceName', { name: truncatedFullName });
```

should do it and take into account translated message links.

---

_Idea 3: ???_

This one can get a bit hairy, so any additional thoughts from @bjester @rtibbles would  be much appreciated here.

## User-facing consequences

Cannot provision a device if you have a long full name or speak German and have a medium full name.

holta:
Coincidentally I had the exact same question yesterday: what validation of KOLIBRI_USER is wise and/or necessary?!

But not knowing the exact answer (specifically in the case of Linux here) I just validated the input file to be non-empty for now:

```
if [ -s /etc/kolibri/username ]    # Test that file exists with non-zero size
```

@jredrejo might have better ideas!  See line 349 here, in what is basically a 2-line fix to Kolibri's Debian/Ubuntu installer:

- PR learningequality/kolibri-installer-debian#117

rtibbles:
[...]

PR #10392, opened 2023-04-06
setup wizard - conflict in allowed char length between device_name and user's full_name

thanksameeelian:
## Summary
Addresses #10326, a bug reported where validation would fail at the end of "On My Own" setup if a user's `full_name` was too long to fit completely within the character limits of device and facility names once their respective descriptive strings were also added ("Personal device for `{full_name}`" & "Home facility for `{full_name}`"). 

Resolves the bug by slicing the device & facility names to the required amount of characters once the translated string is combined with a user's `full_name`.

## References
Addresses #10326

## Reviewer guidance
- working off of `develop` with a fresh kolibri instance, follow the "on my own" setup and give your user a `full_name` of at least 31 characters. as you try to finalize setup, you will be unable to progress forward and you'll find a validation error reported in the terminal, as `device_name` exceeds 50 chars (if `full_name` is at least 83 chars, both `device_name` and facility name will fail)
- working off of this branch, following the same path should result in no errors and a finalized, usable setup.

----

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/plugins/setup_wizard/assets/src/views/onboarding-forms/SettingUpKolibri.vue
+++ b/kolibri/plugins/setup_wizard/assets/src/views/onboarding-forms/SettingUpKolibri.vue
@@ -169,8 +169,8 @@
 
           // // suggested fix(-ish)
           // let baseTranslatedLength = this.$tr('defaultDeviceName', { name: '' }).length
-let       // truncatedFullName = this.user.full_name.slice(0, baseTranslatedLength-1);
-let       // actualStringWeUse = this.$tr('defaultDeviceName', { name: truncatedFullName });
+          // let truncatedFullName = this.user.full_name.slice(0, baseTranslatedLength-1);
+          // let actualStringWeUse = this.$tr('defaultDeviceName', { name: truncatedFullName });
 
           // // if there is a device name available, use it, otherwise get superuser's full name
           // // and insert into (translation of) 'personal device of {}'
--- a/kolibri/plugins/setup_wizard/assets/src/views/onboarding-forms/SettingUpKolibri.vue
+++ b/kolibri/plugins/setup_wizard/assets/src/views/onboarding-forms/SettingUpKolibri.vue
@@ -81,7 +81,8 @@
         const usersName = get(this.wizardContext('superuser'), 'full_name', '');
         const facilityName =
           this.wizardContext('facilityName') ||
-          this.$tr('onMyOwnFacilityName', { name: usersName });
+          // full names may be up to 120 chars, but facility names can only be 100 chars
+          this.$tr('onMyOwnFacilityName', { name: usersName }).slice(0, 100);
         const selectedFacility = this.wizardContext('selectedFacility');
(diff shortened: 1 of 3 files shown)


kolibri/plugins/setup_wizard/assets/src/views/onboarding-forms/SettingUpKolibri.vue
```diff
@@ -162,6 +162,18 @@
...
+
+          // // is there anywhere else device name can get assigned to full name (saw nickname elsewhere)
+
+          // // suggested fix(-ish)
+          // let baseTranslatedLength = this.$tr('defaultDeviceName', { name: '' }).length
+          // let truncatedFullName = this.user.full_name.slice(0, baseTranslatedLength-1);
+          // let actualStringWeUse = this.$tr('defaultDeviceName', { name: truncatedFullName });
+
+          // // if there is a device name available, use it, otherwise get superuser's full name
+          // // and insert into (translation of) 'personal device of {}'
```
nucleogenesis: Looking back at this now I guess we could just truncate what we send in any case below when we're assigning the `device_name` field.

So you could take line 179 here and just cut it down to 50 characters to appease all of the validators. This should resolve the originally reported error where long names break the final step of the setup and hopefully won't be an issue for users as I believe it will only be used when showing the device on a network (ie, during sync / import / remote browsing).

We'll need to apply this to the Facility Name too I believe and should be good.

nucleogenesis (approved):
Code looks good to me and should cover everything. I think that with sign-off from @pcenov / @radinamatic this should be good to go

radinamatic:
Manual QA confirms the issue as fixed, according to the reviewer guidance.

https://user-images.githubusercontent.com/1457929/231311561-6db51f85-6759-4f5e-98aa-3b5d75d10b9c.mp4

---
One thing I noticed is that even though the `Full name` field allows for 120 characters, the resulting facility name get shortened to 100, and the resulting device name gets shortened to 50. Maybe we should rethink and shorten the allowed character limit in the first place... 🤔 

1 | 2 | 3 | 4
-- | -- | -- | --
Ubuntu20 04  Running  - Oracle VM VirtualBox_001 | Ubuntu20 04  Running  - Oracle VM VirtualBox_002 | Ubuntu20 04  Running  - Oracle VM VirtualBox_003 | Ubuntu20 04  Running  - Oracle VM VirtualBox_004

rtibbles:
> One thing I noticed is that even though the Full name field allows for 120 characters

But that's the full name for the user, which is allowed to be that long. It's just that we then reuse it to generate a facility name and a device name, so we wouldn't want to shorten the full name just for these cases.

radinamatic:
> full name for the user, which is allowed to be that long

I got that far 😛 

What seemed awkward is that we allow so long full names (where did the number 120 come from, really?), but on personal devices & facilities when the user goes to rename their names, they'll have _only_ 100 and 50, respectively. Not really a problem, just looked a bit weird.

PR #10414, opened 2023-04-07
Fix scroll bar jumping on topics pages

MisRob:
## Summary

Fixes jumping scrollbar when reaching the bottom of the page for certain page heights. The issue is caused by the document scroll height being changed by the sidebar's switching position from absolute to fixed in the sticky calculation, resulting in an endless cycle of the sticky calculation being called and the sidepanel alternating between fixed and absolute position over and over. Setting min width prevents this by making sure that the document scroll height won't change on the sidebar positioning updates.

### Before
https://drive.google.com/file/d/1RB_-w0gmQmb2UBoX20us6aNU9ch_NKpN/view?usp=share_link

### After
https://drive.google.com/file/d/1D-gE0hzv-jeJbT8__DINuFobWUOvWOCr/view?usp=share_link

## Reviewer guidance

If you can, find a place where you can reproduce the scroll bar jumping and check that this PR fixes it. Note that this issue occurs under pretty specific conditions in regard to page height. @thanksameeelian and me could commonly reproduce it when there were two rows of cards on a page, for example on _Learn --> Library --> Kolibri QA Channel --> African Story Books_ page.

## Comments

Note that in some cases, this change may add a bit extra space under the card grid.

I was experimenting with more robust solutions incorporating changes to the page layout and the sticky calculation, however, it was complicated by the complexity of the component and would result in a larger refactor that I don't have the capacity for right now. Reported in #10415.

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/plugins/learn/assets/src/views/TopicsPage/index.vue
+++ b/kolibri/plugins/learn/assets/src/views/TopicsPage/index.vue
@@ -35,7 +35,7 @@
 
       <main
         class="main-content-grid"
-        :style="gridOffset"
+        :style="gridStyle"
       >
         <KBreadcrumbs
           v-if="breadcrumbs.length && windowIsSmall"
@@ -544,10 +544,30 @@
           return 346;
         }
       },
-      gridOffset() {
-        return this.isRtl
-          ? { marginRight: `${this.sidePanelWidth + 24}px` }
-          : { marginLeft: `${this.sidePanelWidth + 24}px` };
+      gridStyle() {
+        let style = {};
+        /*
+          Fixes jumping scrollbar when reaching the bottom of the page
+          for certain page heights and when side bar is present.
+          The issue is caused by the document scroll height being changed
+          by the sidebar's switching position from absolute to fixed in
+          the sticky calculation, resulting in an endless cycle
+          of the calculation being called and the sidepanel alternating between
+          fixed and absolute position over and over. Setting min height prevents
+          this by making sure that the document scroll height won't change
+          on the sidebar positioning updates.
+        */
+        if (this.windowIsLarge) {
+          style = {
+            minHeight: '900px',
+          };
+        }
+        if (this.isRtl) {
+          style.marginRight = `${this.sidePanelWidth + 24}px`;
+        } else {
+          style.marginLeft = `${this.sidePanelWidth + 24}px`;
+        }
+        return style;
       },
       numCols() {
         if (this.windowBreakpoint > 1 && this.windowBreakpoint < 2) {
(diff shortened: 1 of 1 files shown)


kolibri/plugins/learn/assets/src/views/TopicsPage/index.vue
```diff
@@ -740,6 +740,19 @@
...
     top: $toolbar-height;
+
+    /*
+      Fixes jumping scrollbar when reaching the bottom of the page
+      for certain page heights. The issue is caused by the document
+      scroll height being changed by the sidebar's switching position
+      from absolute to fixed in the sticky calculation, resulting in
+      an endless cycle of the sticky calculation being called
+      and the sidepanel alternating between fixed and absolute position
+      over and over. Setting min width prevents this by making sure that
```
rtibbles: Should this be `min height` not `min width`?

rtibbles: Also - does this happen across all breakpoints? Just thinking about the blanket application of this on mobile.

MisRob: Thank you, yes it should say "min height", will fix it.

> does this happen across all breakpoints? Just thinking about the blanket application of this on mobile.

That's a good point, I don't know. I assumed that in the blanket state, there would be a sidebar anyways, but I really didn't test it and it may not be the case in all cases because `TopicsPage` has many modes. Adding this style just for some breakpoints may be safer. I will look at it.

akolson (commented):
Code changes look okay. However, as pointed out by Richard, this could pose significant challenges on the mobile side of things.

MisRob:
I tried a couple of pages for smaller resolutions and I didn't see the scrollbar jumping there. I think it's because we actually don't display the sidebar in those situations. So I limited the fix only to large screens where the sidebar is present. I also opened https://github.com/learningequality/kolibri/issues/10663.

kolibri/plugins/learn/assets/src/views/TopicsPage/index.vue
```diff
@@ -544,10 +544,30 @@
           return 346;
         }
       },
-      gridOffset() {
-        return this.isRtl
-          ? { marginRight: `${this.sidePanelWidth + 24}px` }
-          : { marginLeft: `${this.sidePanelWidth + 24}px` };
+      gridStyle() {
+        let style = {};
```
rtibbles, line 548: I think once this is pushed into develop, this will cause a linting error because of the addition of the prefer const linting rule.
```suggestion
        const style = {};
```

rtibbles (approved):

akolson (approved):
Code looks good to me. Not sure why github is complaining linting and some tests, but this should be good for merging once the failures have been resolved. Thanks @MisRob

rtibbles:
Oh, because I misread the code, so my 'change' was actually completely unhelpful.

kolibri/plugins/learn/assets/src/views/TopicsPage/index.vue
rtibbles: This was an error, I have removed it.

rtibbles:
Test failures are due to github actions failing, not this PR.

kolibri/plugins/learn/assets/src/views/TopicsPage/index.vue
MisRob: Ah yeah, no worries, using two syntaxes (first time rewriting the object, and the second time only adding a new attribute) for the very same thing didn't help much with this

PR #10424, opened 2023-04-10
Updated docs from issue #10376

jose-hy:
## Summary

Updated documentation (specifically the Getting Started Section) to remove any 0.15 version recommendations

## References

from Issue #10376

## Reviewer guidance

View changes in server following the editing docs section in the documentation

----

## Testing checklist

- [ ] Contributor has fully tested the PR manually
- [ ] If there are any front-end changes, before/after screenshots are included
- [ ] Critical user journeys are covered by Gherkin stories
- [ ] Critical and brittle code paths are covered by unit tests

## PR process

- [ ] PR has the correct target branch and milestone
- [ ] PR has 'needs review' or 'work-in-progress' label
- [ ] If PR is ready for review, a reviewer has been added. (Don't use 'Assignees')
- [ ] If this is an important user-facing change, PR or related issue has a 'changelog' label
- [ ] If this includes an internal dependency change, a link to the diff is provided

## Reviewer checklist

- Automated test coverage is satisfactory
- PR is fully functional
- PR has been tested for [accessibility regressions](http://kolibri-dev.readthedocs.io/en/develop/manual_testing.html#accessibility-a11y-testing)
- External dependency files were updated if necessary (`yarn` and `pip`)
- Documentation is updated
- Contributor is in AUTHORS.md

PR #10434, opened 2023-04-10
Remove 0.15 setup workaround from getting started docs

marcellamaki:
Deletes the 0.15 setup workaround from when setup wizard was in progress


rtibbles:
Closing in favour of #10424

PR #10441, opened 2023-04-11
Removes double focus outline on sign-in page

### LINKED ISSUE #10281: Double focus online on the password for the sign-in page
## Observed behavior
Sometimes, there is a double focus online on the password for the sign-in page

## Errors and logs
Screenshot 2023-03-20 at 3 30 36 PM

## Expected behavior
Should only be one focus outline

## User-facing consequences
It doesn't look very nice, and might be confusing to those navigating by focus outline (and possible screen reader, if two different elements are indeed focused).

## Steps to reproduce
Sign in, and then use the TAB key to navigate. Does not reproduce _all_ of the time so it might take some wrangling. 

Does seem to also happen on hover

## Context
`develop`, Mac 13, Chrome 

This might need to be fixed in KDS

LianaHarris360:
## Summary

This pr updates the sign-in page to remove the double focus outline that occurs when running the dev server with hot reload enabled.

Before:

usernameBefore

passwordbefore

After:
usernameafter

passwordAfter

## References

Fixes #10281 

## Reviewer guidance

1. Run the Kolibri dev server with hot reload enabled.
2. On the sign-in page, use TAB key to navigate to input field.
3. Confirm that there is no double focus outline for username and password input.

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/plugins/user_auth/assets/src/views/AuthBase.vue
+++ b/kolibri/plugins/user_auth/assets/src/views/AuthBase.vue
@@ -430,4 +430,10 @@
     vertical-align: middle;
   }
 
+  /deep/ .ui-textbox-input {
+    &:hover {
+      outline: none;
+    }
+  }
+
 </style>


rtibbles:
Does this only happen when hot-reload is enabled?

marcellamaki:
Someone definitely mentioned this was only reproducible with hot reload - although I can't remember who. Maybe somebody mentioned in standup?

rtibbles:
The main reason for my question was if this only happens during hot-reload, what impact does this have on a production environment - as it seems to be doing a deep selector on the UI text box to remove the outline state, which might have unintended consequences, if this is a development only bug?

LianaHarris360:
This is a good point. I could only replicate this when hot reload is enabled, so it seems to be a development only bug for me. Besides being a slight annoyance when running the dev server with hot reload and since it has no impact on a production environment, should this issue just be closed?

radinamatic:
Probably mentioned this previously, but I don't recall ever seeing this when testing with production assets.
I can give a whirl to assets from this PR to see that the change has no effect on focus outline in production Kolibri, if we want to get rid of that slight annoyance in the dev environment, lmk!

radinamatic:
After cursory check on Ubuntu 20.04 with Kolibri DEB asset from this PR, visual focus indicator in both Firefox and Chrome  looks correct for all interactive elements.

Testing on Windows 10 next to confirm!

radinamatic:
Same uneventful experience on Windows with the EXE asset from this PR installed and tested on Firefox, Chrome & Edge.

No regressions regarding the visual focus outline observed.

rtibbles (approved):

rtibbles:
Not sure what is going on here - but this seems to solve it. I would think that it might be a deeper issue with KDS, but it doesn't seem to occur anywhere else.

PR #10455, opened 2023-04-11
Fixes loading states in general

### LINKED ISSUE #10336: Ensure Vuex loading state is properly managed in all pages post CoreBase Refactor
## Observed behavior
Since the `CoreBase` refactor, loading state has been inconsistently managed across all plugins, resulting in "jumpy" or flashing pages and an overall non-smooth user experience.

## Expected behavior
As the "loading" state in Vuex was not changed by the refactor, this should still be available for tracking the loading state, and should be used to properly display the page linear loader.

Tasks: 
- [ ] #10455 
- [x] #10729 

## User-facing consequences
Glitchy or flashing UI. Loading status is not clearly communicated and can be confusing.

## Context
`develop`

nucleogenesis:
## Summary

A handful of fixes to various loading states in Learn, smoothing things out a bit.

Overall, the `LearnAppBarPage` was taking a `loading` prop but it only got that prop from two components that use it. Ultimately, I think that should have never been a prop and should just be set at the highest level to reflect the value of the Core `loading` state. 

The result now is that when we load content of a page initially, we show the KLinearLoader in the AppBar -- per the [KDS loaders guidelines](https://design-system.learningequality.org/loaders/#linearloaders) -- but then use the [circular loaders](https://design-system.learningequality.org/loaders/#circularloaders) when loading new information.

I also removed and updated logic around the loaders on the BookmarkPage where now the circular loader only shows when "loading more" to the same effect.

## References

Fixes #10336 

## Reviewer guidance

Note that some of these things may be easier to observe with network throttling.

1) User table shows loader while fetching users

https://github.com/learningequality/kolibri/assets/6356129/ced1f702-1716-42f7-be15-8dd6c53f842a

2) Lesson playlist page does not show "no resources" message while fetching, app bar linear loader is active

https://github.com/learningequality/kolibri/assets/6356129/156fc540-d771-4e0d-8ef5-98114e8c8ac2

3) Flash in the Topics Page navigation

https://github.com/learningequality/kolibri/assets/6356129/0d592eb0-e0fb-464b-b4b5-963bb2c3cc90

---

### DIFF
--- a/kolibri/core/assets/src/views/CoreTable.vue
+++ b/kolibri/core/assets/src/views/CoreTable.vue
@@ -1,9 +1,11 @@
 <script>
 
   import get from 'lodash/get';
+  import KCircularLoader from 'kolibri-design-system/lib/loaders/KCircularLoader';
 
   export default {
     name: 'CoreTable',
+    components: { KCircularLoader },
     props: {
       selectable: {
         type: Boolean,
@@ -13,6 +15,11 @@
       emptyMessage: {
         type: String,
         default: null,
+      },
+      dataLoading: {
+        type: Boolean,
+        default: false,
+        required: false,
       },
     },
     computed: {
@@ -73,9 +80,19 @@
         }
       });
 
-      // Insert an empty message as a <p> at the end if it is provided and the
-      // table has no rows.
-      const showEmptyMessage = this.emptyMessage && !tableHasRows;
+      // If we have loaded the data, but have no empty message and no rows, we log an error.
+      if (!this.dataLoading && this.emptyMessage && !tableHasRows) {
+        console.error('CoreTable: No rows in table, but no empty message provided.');
+      }
+
+      /*
+       * If data is still loading, then we show a loader. Otherwise, we show the
+       * empty message if there are no rows in the table. If we have loaded data, have
+       * an emptyMessage and have no rows.
+       */
+      var noDataEl = this.dataLoading
+        ? createElement('p', [createElement(KCircularLoader)])
+        : createElement('p', this.emptyMessage);
 
       return createElement('div', { class: 'core-table-container' }, [
         createElement('table', { class: 'core-table' }, [
@@ -83,7 +100,7 @@
           theadEl,
           tbodyCopy,
         ]),
-        showEmptyMessage && createElement('p', this.emptyMessage),
+        noDataEl,
       ]);
     },
   };
--- a/kolibri/core/assets/src/views/CoreTable.vue
+++ b/kolibri/core/assets/src/views/CoreTable.vue
@@ -1,7 +1,10 @@
 <script>
 
   import get from 'lodash/get';
+  import logger from 'kolibri.lib.logging';
   import KCircularLoader from 'kolibri-design-system/lib/loaders/KCircularLoader';
+
+  const logging = logger.getLogger(__filename);
 
   export default {
     name: 'CoreTable',
@@ -82,7 +85,7 @@
 
       // If we have loaded the data, but have no empty message and no rows, we log an error.
       if (!this.dataLoading && this.emptyMessage && !tableHasRows) {
-        console.error('CoreTable: No rows in table, but no empty message provided.');
+        logging.error('CoreTable: No rows in table, but no empty message provided.');
       }
 
       /*
--- a/kolibri/core/assets/src/views/CoreTable.vue
+++ b/kolibri/core/assets/src/views/CoreTable.vue
@@ -88,11 +88,11 @@
       /*
        * If data is still loading, then we show a loader. Otherwise, we show the
        * empty message if there are no rows in the table. If we have loaded data, have
-       * an emptyMessage and have no rows.
+       * an emptyMessage and have no rows. If we have rows, then we show the table alone
        */
-      var noDataEl = this.dataLoading
+      var dataStatusEl = this.dataLoading
         ? createElement('p', [createElement(KCircularLoader)])
-        : createElement('p', this.emptyMessage);
+        : !tableHasRows && createElement('p', this.emptyMessage); // Only show message if no rows
 
       return createElement('div', { class: 'core-table-container' }, [
         createElement('table', { class: 'core-table' }, [
@@ -100,7 +100,7 @@
           theadEl,
           tbodyCopy,
         ]),
-        noDataEl,
+        dataStatusEl,
       ]);
     },
   };
--- a/kolibri/core/assets/src/views/UserTable/index.vue
+++ b/kolibri/core/assets/src/views/UserTable/index.vue
@@ -17,7 +17,7 @@
       data-test="selectAllCheckbox"
       @change="selectAll($event)"
     />
-    <CoreTable>
+    <CoreTable :emptyMessage="emptyMessage" :dataLoading="dataLoading">
       <template #headers>
         <th data-test="fullNameHeader" :style="{ minWidth: '32px' }">
           <span v-if="selectable" class="visuallyhidden">
@@ -195,13 +195,6 @@
       </template>
     </CoreTable>
 
-    <p
-      v-if="(!users || !users.length) && !$store.state.core.loading"
-      class="empty-message"
-    >
-      {{ emptyMessage }}
-    </p>
-
   </div>
 
 </template>
@@ -276,6 +269,10 @@
       selectedStyle: {
         type: String,
         default: '',
+      },
+      dataLoading: {
+        type: Boolean,
+        default: false,
       },
     },
     computed: {
--- a/kolibri/core/assets/src/views/UserTable/index.vue
+++ b/kolibri/core/assets/src/views/UserTable/index.vue
@@ -62,7 +62,14 @@
       </template>
 
       <template #tbody>
-        <tbody>
+        <tbody v-if="$store.state.core.loading">
+          <tr style="overflow:hidden;">
+            <td>
+              <KCircularLoader class="center-loader" />
+            </td>
+          </tr>
+        </tbody>
+        <tbody v-else>
           <tr
             v-for="user in users"
             :key="user.id"
@@ -189,7 +196,7 @@
     </CoreTable>
 
     <p
-      v-if="!users || !users.length"
+      v-if="(!users || !users.length) && !$store.state.core.loading"
       class="empty-message"
     >
       {{ emptyMessage }}
@@ -387,6 +394,21 @@
 
   td.id-col {
     max-width: 120px;
+  }
+
+  .center-loader {
+    &.ui-progress-circular {
+      position: static !important;
+      width: 100% !important;
+      overflow: hidden;
+    }
+
+    /deep/ svg {
+      top: 48px;
+      width: 32px;
+      height: 32px;
+      overflow: hidden;
+    }
   }
 
 </style>
--- a/kolibri/core/assets/src/views/UserTable/index.vue
+++ b/kolibri/core/assets/src/views/UserTable/index.vue
@@ -62,14 +62,7 @@
       </template>
 
       <template #tbody>
-        <tbody v-if="$store.state.core.loading">
-          <tr style="overflow:hidden;">
-            <td>
-              <KCircularLoader class="center-loader" />
-            </td>
-          </tr>
-        </tbody>
-        <tbody v-else>
+        <tbody>
           <tr
             v-for="user in users"
             :key="user.id"
--- a/kolibri/core/assets/src/views/UserTable/index.vue
+++ b/kolibri/core/assets/src/views/UserTable/index.vue
@@ -386,19 +386,4 @@
     max-width: 120px;
   }
 
-  .center-loader {
-    &.ui-progress-circular {
-      position: static !important;
-      width: 100% !important;
-      overflow: hidden;
-    }
-
-    /deep/ svg {
-      top: 48px;
-      width: 32px;
-      height: 32px;
-      overflow: hidden;
-    }
-  }
-
 </style>
--- a/kolibri/plugins/coach/assets/src/modules/pluginModule.js
+++ b/kolibri/plugins/coach/assets/src/modules/pluginModule.js
@@ -21,7 +21,7 @@ const logging = logger.getLogger(__filename);
 export default {
   state() {
     return {
-      busy: false,
+      dataLoading: false,
       classList: [],
       pageName: '',
       toolbarRoute: {},
@@ -37,6 +37,9 @@ export default {
     },
     SET_TOOLBAR_TITLE(state, title) {
       state.toolbarTitle = title;
+    },
+    SET_DATA_LOADING(state, dataLoading) {
+      state.dataLoading = dataLoading;
     },
     SET_TOOLBAR_ROUTE(state, toolbarRoute) {
       state.toolbarRoute = toolbarRoute;
@@ -60,13 +63,15 @@ export default {
   },
   actions: {
     setClassList(store, facilityId) {
+      store.commit('SET_DATA_LOADING', true);
       return ClassroomResource.fetchCollection({
         getParams: { parent: facilityId || store.getters.currentFacilityId, role: 'coach' },
       })
         .then(classrooms => {
           store.commit('SET_CLASS_LIST', classrooms);
         })
-        .catch(error => store.dispatch('handleApiError', error));
+        .catch(error => store.dispatch('handleApiError', error))
+        .finally(() => store.commit('SET_DATA_LOADING', false));
     },
     /**
       * Handle coach page errors.
--- a/kolibri/plugins/coach/assets/src/modules/pluginModule.js
+++ b/kolibri/plugins/coach/assets/src/modules/pluginModule.js
@@ -64,6 +64,7 @@ export default {
   actions: {
     setClassList(store, facilityId) {
       store.commit('SET_DATA_LOADING', true);
+      store.commit('SET_CLASS_LIST', []); // Reset the list if we're loading a new one
       return ClassroomResource.fetchCollection({
         getParams: { parent: facilityId || store.getters.currentFacilityId, role: 'coach' },
       })
--- a/kolibri/plugins/coach/assets/src/modules/pluginModule.js
+++ b/kolibri/plugins/coach/assets/src/modules/pluginModule.js
@@ -70,9 +70,12 @@ export default {
       })
         .then(classrooms => {
           store.commit('SET_CLASS_LIST', classrooms);
+          store.commit('SET_DATA_LOADING', false);
         })
-        .catch(error => store.dispatch('handleApiError', error))
-        .finally(() => store.commit('SET_DATA_LOADING', false));
+        .catch(error => {
+          store.dispatch('handleApiError', error);
+          store.commit('SET_DATA_LOADING', false);
+        });
     },
     /**
       * Handle coach page errors.
--- a/kolibri/plugins/device/assets/src/modules/managePermissions/handlers.js
+++ b/kolibri/plugins/device/assets/src/modules/managePermissions/handlers.js
@@ -19,6 +19,7 @@ function fetchFacilityUsers() {
 
 export function showManagePermissionsPage(store) {
   const shouldResolve = samePageCheckGenerator(store);
+  store.commit('managePermissions/SET_LOADING_FACILITY_USERS', true);
   const promises = Promise.all([fetchFacilityUsers(store), fetchDevicePermissions()]);
   return promises
     .then(function onSuccess([users, permissions]) {
@@ -31,5 +32,8 @@ export function showManagePermissionsPage(store) {
     })
     .catch(function onFailure(error) {
       shouldResolve() ? store.dispatch('handleError', error) : null;
+    })
+    .finally(() => {
+      store.commit('managePermissions/SET_LOADING_FACILITY_USERS', false);
     });
 }
--- a/kolibri/plugins/device/assets/src/modules/managePermissions/handlers.js
+++ b/kolibri/plugins/device/assets/src/modules/managePermissions/handlers.js
@@ -22,18 +22,17 @@ export function showManagePermissionsPage(store) {
   store.commit('managePermissions/SET_LOADING_FACILITY_USERS', true);
   const promises = Promise.all([fetchFacilityUsers(store), fetchDevicePermissions()]);
   return promises
-    .then(function onSuccess([users, permissions]) {
+    .then(([users, permissions]) => {
       if (shouldResolve()) {
         store.commit('managePermissions/SET_STATE', {
           facilityUsers: users,
           permissions,
         });
       }
+      return store.commit('managePermissions/SET_LOADING_FACILITY_USERS', false);
     })
-    .catch(function onFailure(error) {
-      shouldResolve() ? store.dispatch('handleError', error) : null;
-    })
-    .finally(() => {
+    .catch(error => {
       store.commit('managePermissions/SET_LOADING_FACILITY_USERS', false);
+      return shouldResolve() ? store.dispatch('handleError', error) : null;
     });
 }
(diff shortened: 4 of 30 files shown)


marcellamaki:
@nucleogenesis - I've started a code read through and this looks good. But I do just want to note that the original scope of the issue was across the whole frontend, not just in learn. If it would be helpful, I can create follow up issues to address this in the rest of the plugins, so they could be split up? Let me know what you prefer. 

@radinamatic - some robust manual QA would be helpful here, for sure, although I'll be doing some as well.

nucleogenesis:
@marcellamaki I did dig through the other plugins but didn't note any other spots where I saw no loading states. 

I didn't do anything on change facility or setup wizard though so can double check there. 

If you have any specific spots where this has been reported I'm happy to add those onto this PR.

marcellamaki:
Ah, yes, sure! [I've been noticing it in coach](https://learningequality.slack.com/archives/CB37UM23A/p1675981260280819), mostly because I've been working in there a bit recently, but an initial response from @pcenov doing some replication this indicate it is fairly widespread at least in Learn and Coach, although I suppose it maybe have been resolved incidentally in the interim?  I'm not sure about the device and facility plugins - it may not be as noticeable there also because of the types of pages and interactions.

He shared this with @radinamatic 
image

My takeaway is more that this is not a matter of QA'ing to find a handful of places where it is missing, since it is inconsistently implemented. I can help with some replication later this week of that would be helpful!

marcellamaki (changes requested):
Thanks for continuing the work on this @nucleogenesis! I can see that you've dug into a lot of things here, and that you're really working across a lot of the plugins to get things resolved. 

My main issue with this is although it may work more smoothly (I haven't done robust manual QA yet, but I can already tell from the code that there will be improvements), the loading state still seems inconsistent and a bit confusing while reading through. 

For example - in the AppBarPage, you've added `v-if="!$store.state.core.loading"` on the main-wrapper div, which makes a lot of sense to me. But there's still a `loading` prop, and the `KLinearLoader` is visible based on that prop. 

Within `FacilityAppBarPage` you are passing: 
```
<AppBarPage :title="title" :loading="$store.state.core.loading">
``` 
My suspicion is that this combination of "what is loading and where is it coming from" is just an oversight from refactoring, but I think it's worth cleaning up now, across these files and making sure the same type of thing isn't popping up in other places. I can see myself coming to this code later and wondering "why both? maybe the prop isn't even passed in? has it been deleted in the parent and now it's just undefined and not doing anything?"

I can see how tangled up a lot of this is (and you've alluded it it a few times during standups 😄) and I don't want to send you too far down a rabbit hole. I also know your feelings about Vuex! But, I think having a secondary goal of a consistent pattern for using the loading state(s) -- and documenting it, even if only in a few code comments in the most high level shared page components -- would be really helpful to our future selves. I suspect you are already working towards identifying and implementing that consistent pattern, but writing it out might be just the reference point you need to go back through and ensure consistency within and across plugins and pages.  

I'm happy to talk more about this and brainstorm some ideas with you and the pros and cons of approaches. I think it's fine to use both Vuex and props, but in my opinion, it's confusing to use both in the same component. Something like the vuex in the highest parent component, and then passing down the props to other child components as needed could work, and might eliminate the need to be using vuex everywhere. I could imagine there might be cases where that won't work though, or where it might be a little more complex.

Let me know your thoughts on this and if you'd want to talk though anything more. Another option might be to split this into several issues and once we establish a consistent pattern, divide up the work per-plugin so that it doesn't all have to be on you. 

Thanks again for the ongoing efforts here - I know this is not an easy problem to sort out.

nucleogenesis:
@marcellamaki 
> I can see myself coming to this code later and wondering "why both? maybe the prop isn't even passed in? has it been deleted in the parent and now it's just undefined and not doing anything?"

Ahh yeah - I meant to use the `$store.state.core.loading` value as the primary indicator of "something is loading" throughout.

So in other changes I've made, I used that value to toggle circular loaders.

It honestly feels weird that the whole app has one "something is loading" state value like we could run into a race condition if we are loading multiple things simultaneously if they all are setting/unsetting the loading state. 

In the cases where I use it it seems to work alright but it still feels odd.

---

In any case - I think that using something like `mapState({ loading: state => state.core.loading})` wherever we need the value will work. It doesn't look like there are many places even passing the `loading` prop up to the various AppBar pages and so it shouldn't be too much to clean all that up and include it here. I'll re-request code review when done.

nucleogenesis:
@pcenov hey! I've added before/after videos (both are in the same video) for each of the main fixes here. I'll add some more changes, but none that will require re-review

pcenov:
Hi @nucleogenesis,
I'm not able to verify your first point 'User table shows loader while fetching users' as currently there's a problem with the Users page as none of the newly created users get displayed in the table - so something is broken there:

https://github.com/learningequality/kolibri/assets/79847249/5cab4fe0-d83c-4b2f-8453-1eec72543a38

The other two points are implemented as specified in the reviewer guidance and demonstrated by your videos.

I have also recorded two videos, the first one as a Learner and the second one as a Super admin interacting with Kolibri while the Slow 3g throttling is on. This is just for illustration that we have 3 different loading indicators - the Kolibri logo, a spinner and a green line under the top bar and in some cases no indication at all. So to me it seems that areas that can benefit from additional improvements would be the sign-in page where there is no loading indicator, everything in the Device and Coach plugins where there's also no loading indicator, viewing a Quiz as a learner for the first time and anything else that doesn't clearly indicate that things are loading.

Learner:

https://github.com/learningequality/kolibri/assets/79847249/f7d66ced-54e0-4f82-ab9c-eb88243fc1e7

Super admin:

https://github.com/learningequality/kolibri/assets/79847249/470a536a-4caa-4723-842d-9b2193553747

nucleogenesis:
Just noting that I'm working on updating to resolve the conflicts. I'll be moving on to complete @pcenov 's listed issues above after I determine whether or not I'm causing the Users table to not populate with newly created users.

Will aim to have this ready for re-review after tomorrow

marcellamaki:
@nucleogenesis - just confirming I see the update loader state in the user table. I did a little bit of manual QA - not as robust as @pcenov of course :D - but initially it does seem to be good, although Peter's point about the multiple loading states is well taken. This should certainly be something we discuss for an upcoming release. 

Jacob, let me know when this is ready for final code review and I'll take a look!

nucleogenesis:
@marcellamaki I went to revise a few things to clean things up on the "one or the other" front re: Vuex vs Component state and such -- I opted to put the loading states for discrete areas where things might be loading into more local Vuex module. This way you can introspect not just if _something_ is loading, but whether a specific piece of data is loading.

This brought with it some non-comprehensive refactors (as in, there are more places where this can be applied). I figure that getting review on it where it is now would be good and I can follow-up with a list of other places where this pattern can be applied.

For code review, I suggest going commit-by-commit as in most cases they should represent a self-contained change.

nucleogenesis:
Also, this fixes user tables broadly where it will no longer show "No users to show" while loading. So it would be helpful if @pcenov you could check the user tables in:

- facility -> users
- facility -> classes -> class management
- facility -> classes -> class editing
- facility -> permissions (see video - this video's fixes should be similar in all instances of the "user table")


https://github.com/learningequality/kolibri/assets/6356129/2c138e02-0a20-4bf5-b262-cc7c1b385ad2

- coach -> group edit/create
- coach -> plan -> lesson assignment table
- device -> channels list (see video)

https://github.com/learningequality/kolibri/assets/6356129/bfe265b4-1311-48d1-a7bf-831c34f1f1fa

nucleogenesis:
@radinamatic -- I pinged Peter on this yesterday (forgot he was out), would you be able to run through the [scenarios here](https://github.com/learningequality/kolibri/pull/10455#issuecomment-1558163087)?

kolibri/core/assets/src/views/CoreTable.vue
```diff
@@ -73,17 +80,27 @@
...
+      if (!this.dataLoading && this.emptyMessage && !tableHasRows) {
+        console.error('CoreTable: No rows in table, but no empty message provided.');
+      }
+
+      /*
+       * If data is still loading, then we show a loader. Otherwise, we show the
+       * empty message if there are no rows in the table. If we have loaded data, have
+       * an emptyMessage and have no rows. If we have rows, then we show the table alone
+       */
+      var dataStatusEl = this.dataLoading
```
thanksameeelian, line 96: **non-blocking**, just curious!

is there a reason this is a `var` vs. a `const`?  it appears like its value doesn't get reassigned (just conditionally assigned) and it doesn't need to be accessed beyond this scope, but i think i'm missing something & i'd love to learn what it is! 🙂

kolibri/plugins/facility/assets/src/modules/classManagement/handlers.js
```diff
@@ -3,23 +3,28 @@ import samePageCheckGenerator from 'kolibri.utils.samePageCheckGenerator';
...
+      classrooms => {
+        if (shouldResolve()) {
+          store.commit('classManagement/SET_STATE', {
+            modalShown: false,
+            classes: [...classrooms],
+          });
+          store.commit('CORE_SET_PAGE_LOADING', false);
+        }
+      },
+      error => {
```
thanksameeelian: **non-blocking:** 

is this `error` handling in `.then` effectively the same as a `.catch`, or doing something slightly different/additional? if it's the same, it might be nice to standardize this and isolate the error with a `.catch` to make the pattern more recognizable

kolibri/plugins/facility/assets/src/modules/userManagement/handlers.js
```diff
@@ -18,19 +19,22 @@ export function showUserPage(store, toRoute, fromRoute) {
...
+      users => {
+        if (shouldResolve()) {
+          store.commit('userManagement/SET_STATE', {
+            facilityUsers: users.results.map(_userState),
+            totalPages: users.total_pages,
+            usersCount: users.count,
+          });
+        }
+      },
+      error => {
```
thanksameeelian: **non-blocking:** 

i have the same question here as my comment in `...classManagement/handlers.js`: is handling the error here in `.then` doing something different from a standard `.catch`? if so, i'm really interested to learn the nuance, as i still have a lot to learn about promises and the ins & outs of `.then`, `.catch`, `.finally` in general. 

if this _is_ effectively the same as a `.catch`, i think placing this explicitly within a `.catch` might be slightly more clear/readily understandable.

kolibri/plugins/learn/assets/src/views/LibraryPage/index.vue
```diff
@@ -3,7 +3,6 @@
   <LearnAppBarPage
     :appBarTitle="appBarTitle"
     :appearanceOverrides="{}"
-    :loading="loading"
```
thanksameeelian, line 6: **non-blocking:**

along the same lines at @pcenov's [earlier comment](https://github.com/learningequality/kolibri/pull/10455#issuecomment-1544148276) (and hopefully not contrary to his & others' suggestions), i was struck by the variety of different loading indicators, which felt especially evident here in `Learn` - while `Bookmarks` provides the helpful spinny loader, when you navigate to `Library` or `Home`, the loader is the subtler green line loader, and since this was the last part of kolibri i visited during review, the difference felt pretty incongruous, the line loader feels comparatively hard to notice, especially when you're not looking for it, and the reasoning behind having a spinning loader in one subtopic vs the line loader in another subtopic within the same plugin feels a bit hard to understand as a user. even though one may be linked to particular content while the other may be linked to a whole page loading, i think that differentiation isn't obvious (and my understanding may be wrong!) and to a user the effect is the same - they're waiting, but they may not notice the loading indication in all cases.

i don't think this should necessarily be addressed in this PR (or at all if it's a minority opinion), but i did want to make a note of it here. 

https://github.com/learningequality/kolibri/assets/43046460/8f44204b-59b3-4cdb-9b59-70179942db45

kolibri/plugins/coach/assets/src/modules/pluginModule.js
```diff
@@ -21,7 +21,7 @@ const logging = logger.getLogger(__filename);
 export default {
```
thanksameeelian, line 21: **non-blocking but potentially important to address in a further loading state or coach-specific PR if not here:**

i just wanted to re-raise the issue here as it is not caused or worsened by your work but is slightly affected by it: the oddness in loading in `Coach` as a multi-facility user (outlined in #10714 & detailed further [in this comment](https://github.com/learningequality/kolibri/issues/10714#issuecomment-1558262576)) was also noticeable during review of your PR, because now when a multi-facility user goes to look at the classes in one of their imported facilities and the default facility's class list is briefly displayed before the imported facility's class list renders and replaces it, the default facility's class list is also decorated with a loading spinner. 

https://github.com/learningequality/kolibri/assets/43046460/0efde4e2-d1fe-42ce-b21c-04656334f3eb

the scenario definitely was not caused by your PR, but is affected by the addition of the loading spinner. just FYI the same doesn't occur in the somewhat parallel scenario in `Facility` - it's all looking good there:

https://github.com/learningequality/kolibri/assets/43046460/5a33235e-5a24-439c-8e54-6132c9b0d8d5

kolibri/plugins/device/assets/src/views/ManageContentPage/index.vue
```diff
@@ -45,20 +45,23 @@
...
-          <ChannelPanel
-            v-for="channel in sortedChannels"
-            :key="channel.id"
-            :channel="channel"
-            :disabled="channelIsBeingDeleted(channel.id)"
-            :showNewLabel="showNewLabel(channel.id)"
-            @select_delete="deleteChannelId = channel.id"
-            @select_manage="handleSelectManage(channel.id)"
-          />
+          <KCircularLoader v-if="channelListLoading" />
```
thanksameeelian, line 53: **non-blocking, hoping for additions in a follow-up PR:**

having the loading spinner here & in `Permissions` is **really nice**, and its helpfulness is highlighted in comparison to the behavior in `Info` & `Facilities` within this same plugin, which don't appear to provide any loading indicators. it would be nice to add it in there as well in the future, if possible!

https://github.com/learningequality/kolibri/assets/43046460/5c945457-d582-4c60-925a-6fc7fffba47b

thanksameeelian (commented):
this looks good to me!

i reviewed the code, ran through the scenarios described, and tried to visit every page affected by these changes, and i believe it's all looking good! i have some non-blocking questions and observations, but many are related to separate extant & tracked issues as well as potential future follow-ups, and none require action. 

i've spent a long time digging into this, as it's a previously-unexplored area for me, and the code & variable naming seemed clear and relatively easy to understand from my level of limited insight. 

i'd love to know if @marcellamaki feels like her concerns have been addressed and hear what @radinamatic or @pcenov think of the latest updates! this is looking good and functional to me and it seems like patterns established here open the door for future tweaks and expansions 🙂

kolibri/core/assets/src/views/CoreTable.vue
```diff
@@ -73,17 +80,27 @@
         }
       });
 
-      // Insert an empty message as a <p> at the end if it is provided and the
-      // table has no rows.
-      const showEmptyMessage = this.emptyMessage && !tableHasRows;
+      // If we have loaded the data, but have no empty message and no rows, we log an error.
+      if (!this.dataLoading && this.emptyMessage && !tableHasRows) {
+        console.error('CoreTable: No rows in table, but no empty message provided.');
```
rtibbles: In general, it's preferable to use the Kolibri logger here rather than `console` directly.

https://github.com/learningequality/kolibri/blob/6055ec5b8b02e368150ae6dc0e871c2f424b7a90/kolibri/core/assets/src/utils/navComponents.js#L2 <-- example.

One distinct advantage of this is that we just merged a PR that silences the logger during tests, allowing the test output to be much much quieter!

kolibri/core/assets/src/views/UserTable/index.vue
```diff
@@ -389,4 +386,19 @@
     max-width: 120px;
   }
 
+  .center-loader {
```
rtibbles: This class isn't used in this component - is this intended to be here, reaching into the CoreTable component? Should it be in CoreTable? Or should it not be here at all?

kolibri/plugins/device/assets/src/modules/managePermissions/handlers.js
```diff
@@ -31,5 +32,8 @@ export function showManagePermissionsPage(store) {
     })
     .catch(function onFailure(error) {
       shouldResolve() ? store.dispatch('handleError', error) : null;
+    })
+    .finally(() => {
```
rtibbles: I think core-js should polyfill this properly, but worth testing this on IE11 just to be absolutely sure.

kolibri/plugins/device/assets/src/modules/managePermissions/index.js
```diff
@@ -15,5 +16,8 @@ export default {
     RESET_STATE(state) {
       Object.assign(state, defaultState());
     },
+    SET_LOADING_FACILITY_USERS(state, loadingFacilityUsers) {
+      Object.assign(state, { loadingFacilityUsers });
```
rtibbles: Any reason to use Object.assign here rather than `state.loadingFacilityUsers = loadingFacilityUsers`?

kolibri/plugins/facility/assets/src/modules/classEditManagement/handlers.js
```diff
@@ -18,8 +18,11 @@ export function showClassEditPage(store, classId) {
     ClassroomResource.fetchCollection({ getParams: { parent: facilityId }, force: true }),
   ];
   const shouldResolve = samePageCheckGenerator(store);
```
rtibbles: I suspect that the samePageCheckGenerator can be removed here now too - as it was to guard the global shared state that is no longer being used due to the removal of references to CORE_SET_PAGE_LOADING

kolibri/plugins/facility/assets/src/modules/classManagement/handlers.js
rtibbles: It's the optional `onRejected` [parameter](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/then#parameters) to `then` - I tend to agree that use distinct `then` and `catch` is more explicit, but it was also like that when @nucleogenesis got here!

kolibri/plugins/learn/assets/src/views/LearnAppBarPage.vue
```diff
@@ -4,7 +4,7 @@
     :is="page"
     :appBarTitle="appBarTitle"
     :appearanceOverrides="appearanceOverrides"
-    :loading="loading"
+    :loading="$store.state.core.loading"
```
rtibbles: I thought we had deliberately done these via props to make it easier to do away with the core loading state in future. I feel like it would be better to continue on that trajectory, and make the substitutions in the consumers of the `LearnAppBarPage` now, so that they can be phased out on a case by case basis.

Also, if a page is using `LearnAppBarPage` but is not deferring to the global loading state, this will cause unpredictable behaviour.

kolibri/plugins/learn/assets/src/views/TopicsPage/index.vue
```diff
@@ -7,210 +7,216 @@
...
       :appBarTitle="barTitle"
       :appearanceOverrides="{}"
       :primary="!allowDownloads"
       class="page"
     >
-      <template #actions>
+      <template v-if="deviceId" #actions>
         <DeviceConnectionStatus :deviceId="deviceId" color="white" />
       </template>
-      <!-- Header with thumbail and tagline -->
```
rtibbles, line 25: Just to verify here that the only real change here is that all the code has been indented under a `v-else`? The diff is not very informative!

rtibbles (commented):
This seems a practical way forward to me, although I'd prefer we not roll back the path to cleaning out the core loading state entirely in the future.

rtibbles:
The only other thought that occurred is that you changed a lot of `SET_BUSY` to `SET_DATA_LOADING` and equivalent renaming of states. I think it is more semantic, but please do double check that you've caught all the relevant names!

pcenov:
Hi @nucleogenesis, it great to see so much progress made on this front so far! As suggested by you I have confirmed that the issue with "No users to show" while loading is fixed now for the following tables:

Facility > Users
Facility > Classes > class management
Facility > Classes -> class editing
Device > Permissions 
Coach > Plan -> group edit/create
Coach > Plan -> lesson assignment table
Device > Channels list

While further testing specifically for this type of a misleading loading state I was also able to identify the following locations:

Facility > Classes > Class > Enroll learners
Facility > Classes > Class > Assign coaches
Coach > Plan > Lessons
Coach > Plan > Groups

Video:

https://github.com/learningequality/kolibri/assets/79847249/5119133f-e896-4bd6-95c7-1670f80d6d38

kolibri/plugins/learn/assets/src/views/LearnAppBarPage.vue
nucleogenesis: Aye - this is maybe a bit nebulous where this is supposed to be the "Page is loading" state where we have the indeterminate linear loader under the AppBar which is, more or less in my mind, what I figure the `core` loading state ought to impact.

As an alternative, perhaps this should be a prop that gets drilled but defaults to the core loading state in the `CoreAppBarPage` - makes more sense for the core component to default to the core loading state, but then still provides control to whoever uses `LearnAppBarPage`?

kolibri/core/assets/src/views/CoreTable.vue
nucleogenesis: I needed it to be a `var` in an earlier failed attempt to write this bit and I forgot to change it to const after getting it settled

thanksameeelian: i just assumed it was for some technical reason that went over my head! 😅

kolibri/plugins/learn/assets/src/views/LibraryPage/index.vue
nucleogenesis: I agree it's odd -- and looking back at this I think that the [per our KDS docs](https://design-system.learningequality.org/loaders/#linearloaders) I've not applied the correct pattern as this is the "initial page load" - so it should have been the linear loader only (which jives a bit w/ a comment Richard noted about the props in `LearnAppBarPage` relying on the core loading state rather than taking in a more localized prop)

I think that with passing the props up to that page in this case will be a relatively simple update to make, which I can make note of in a follow-up.

kolibri/plugins/coach/assets/src/modules/pluginModule.js
nucleogenesis: Hmm... I'll see if maybe we can clear the contents of the page in coach whenever we make a new requests to a different facility or something.

kolibri/plugins/learn/assets/src/views/TopicsPage/index.vue
nucleogenesis: I'll double check to be sure 👍🏻

marcellamaki (approved):
Thanks @nucleogenesis - it seems like critical feedback has been addressed and that there is a general pattern that we can build and extend on this pattern, so I feel okay about merging this and building on it in follow up. 

One **_essential_** point of immediate follow up however, is that the loading state is not well managed in remote content browsing, whereas the improvements in the rest of learn are pretty clear. I'm not sure why it's different than the current library page, but I imagine it's related to the remote-browsing-ness. I would really like for us to work to get this resolved, and it isn't clearly covered by existing issues. 

Here's a short video example of what I'm talking about, but overall, I think if you explore the remote browsing feature with a throttled connection (or even a not throttled connection - this was just for it to be well captured in screengrab) you will see that there is a lot of sort of.... cached + overriding situations happening that would be good to smooth out. 

If you have a chance to open an issue, that would be great. Otherwise, I will open one tomorrow. 


https://github.com/learningequality/kolibri/assets/17235236/44406c12-6c5f-45b2-b0d7-7d0a4ca13aa6

PR #8156, opened 2021-06-15
Remove double negation in query

jredrejo:
## Summary
Executing the get_unused_files function is problematic in databases with many records, making it unusable when the database times out as it's happening in UNW servers.
This change remove a double negation in a query improving the performance .
Measuring it in the Mexico UNW server it passes from timeout to be executed in Time: 2068.917 ms (00:02.069)

## References
Closes: #8155

```
Explain SELECT "content_localfile"."id",
       "content_localfile"."extension",
       "content_localfile"."available",
       "content_localfile"."file_size"
FROM "content_localfile"
LEFT OUTER JOIN "content_file" ON ("content_localfile"."id" = "content_file"."local_file_id")
WHERE ((NOT ("content_localfile"."id" IN
               (SELECT U1."local_file_id" AS Col1
                FROM "content_file" U1
                INNER JOIN "content_contentnode" U2 ON (U1."contentnode_id" = U2."id")
                WHERE U2."available" =TRUE))
        OR "content_file"."id" IS NULL)
       AND "content_localfile"."available" = TRUE)


 Merge Left Join  (cost=29.99..477901437.97 rows=42855 width=42)
   Merge Cond: ((content_localfile.id)::text = (content_file.local_file_id)::text)
   Filter: ((NOT (SubPlan 1)) OR (content_file.id IS NULL))
   ->  Index Scan using content_localfile_pkey on content_localfile  (cost=0.41..1056.17 rows=31650 width=42)
         Filter: available
   ->  Index Scan using content_file_local_file_id_9780c2ab on content_file  (cost=0.42..3464.65 rows=85751 width=49)
   SubPlan 1
     ->  Materialize  (cost=29.16..10970.53 rows=84033 width=33)
           ->  Merge Join  (cost=29.16..9893.37 rows=84033 width=33)
                 Merge Cond: (u2.id = u1.contentnode_id)
                 ->  Index Scan using content_contentnode_pkey on content_contentnode u2  (cost=0.29..5561.79 rows=51595 width=16)
                       Filter: available
                 ->  Index Scan using content_file_contentnode_id_d4089e6e on content_file u1  (cost=0.42..3147.88 rows=85751 width=49)
```
versus
```
        explain        SELECT "content_localfile"."id",
       "content_localfile"."extension",
       "content_localfile"."available",
       "content_localfile"."file_size"
FROM "content_localfile"
WHERE ("content_localfile"."id" IN
         (SELECT U0."id" AS Col1
          FROM "content_localfile" U0
[...]

PR #10460, opened 2023-04-12
Attempt to reduce query times and memory usage during channel deletion operations

rtibbles:
## Summary
* Ensures final deletion progress always reaches 100%
* Simplifies unused file query considerably, hopefully making it considerably more performant
* Does a values modifier on the unused files query to avoid the cost of inflating large numbers of models, in local testing on my machine, this caused a roughly 8x speed up in iterating over the queryset for ~100 000 LocalFile objects
* For large ContentNode trees do batched deletion of ContentNodes to reduce memory usage caused by Django recording foreign key relationships to traverse and cascade delete

## References
Recent issues on BCK, and previous experience as shown in https://github.com/learningequality/kolibri/pull/8156 means that when there are a lot of channels in a database, our current deletion process can cause very long running queries, thanks to the outer joins on the file and content node tables.

## Reviewer guidance
I tweaked the tests to no longer check the deleted file bytes, as the values call removes the file size, and it is not used in production code.

----

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/core/content/management/commands/deletecontent.py
+++ b/kolibri/core/content/management/commands/deletecontent.py
@@ -42,22 +42,20 @@ def delete_metadata(channel, node_ids, exclude_node_ids, force_delete):
         .count()
     )
 
-    unused_files = (
-        LocalFile.objects.filter(
-            available=True,
-            files__contentnode__channel_id=channel.id,
-            files__contentnode__available=False,
-        )
-        .distinct()
-        .values("id", "file_size", "extension")
-    )
-
-    deleted_bytes = unused_files.aggregate(size=Sum("file_size"))["size"] or 0
-
     if force_delete:
         # Do this before we delete all the metadata, as otherwise we lose
         # track of which local files were associated with the channel we
         # just deleted.
+
+        unused_files = (
+            LocalFile.objects.filter(
+                available=True,
+                files__contentnode__channel_id=channel.id,
+                files__contentnode__available=False,
+            )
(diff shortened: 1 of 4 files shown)


jredrejo (commented):
For the PR, nice, everything makes sense, including tests.



@rtibbles about your comment to optimize `get_unused_files` I've done this patch :

```
--- a/kolibri/core/content/models.py
+++ b/kolibri/core/content/models.py
@@ -298,13 +298,11 @@ class LocalFileQueryset(models.QuerySet, FilterByUUIDQuerysetMixin):
         return self.filter(files__isnull=True).delete()
 
     def get_unused_files(self):
-        ids = LocalFile.objects.filter(
-            Q(files__contentnode__available=False) | Q(files__isnull=True)
-        )
         return (
-            self.filter(id__in=ids)
-            .exclude(files__contentnode__available=True)
-            .filter(available=True)
+            self.filter(available=True)
+            .exclude(Q(files__contentnode__available=True))
+            .filter(Q(files__contentnode__available=False) | Q(files__contentnode=None))
         )

```
I've tested it in my sqlite installation and times passes from 5430 ms to 28 ms. In postgresql (in the huge KDP database) it passes from 228 seconds to 16 seconds

jredrejo (approved):
This is a great improvement on performance, specially in the `unused_files` in the command and `get_unused_files` in the model

PR #10461, opened 2023-04-12
Removes facility after last user changes facility and sets migrated u…

### LINKED ISSUE #10333: Removing last user from "on my own facility" does not remove the facility from the device
## Observed behavior
Reported by @rtibbles in the alpha9 bug bash 

When the last user is migrated out of an on my own facility, that facility is not removed from the device

## Expected behavior
The facility should be removed from the device

LianaHarris360:
…ser as superadmin

## Summary

This pr updates the change facility workflow for the last user in a facility with the removal of the user's source facility from the device and sets the migrated user as the `superadmin` on the device.
- The `ChangeFacilityMachine` has been updated to include a `setAsSuperAdmin` property that is false unless the current user is the last user in the source facility.
- The `mergeuser` task receives a `set_as_super_user` argument that is true if the user is the last user in their source facility. If this argument exists, the user is set as a superuser on the device and the source facility is removed.
- The user no longer sees this string if they will be set as the superadmin after the facility change:
     - `Your user account type will change from '{role}' to 'learner' and you will no longer be able to manage resources on this device. You will need someone with admin permissions in '{facility}' to change your account type back to '{role}'.`

If the user is the last user in the facility they are migrating from, they are met with this screen after selecting a facility to migrate to:
Screenshot 2023-04-11 at 8 17 19 AM

https://user-images.githubusercontent.com/46411498/231536484-a04b15d6-0eb9-4d5a-8158-900e79d024e5.mp4

Note: When the source facility is being removed, the terminal displays the warning: `Deleted {count} database records but expected to delete {total_to_delete} records for facility {facility.name}`. Despite this warning, the workflow still works as intended. The cause for this warning is being investigated.

## References

Closes #10333 

## Reviewer guidance

1. On a learning facility, change the facility of the last user.
2. After the facility change, the user should be a superuser on the device and only the target facility should be on the device.
3. Test that the `chooseAdmin` page still displays if there are other users in the source facility.
4. Test the workflow when creating a new account or merging accounts with an existing user in the remote facility.

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated

### DIFF
--- a/kolibri/plugins/user_profile/tasks.py
+++ b/kolibri/plugins/user_profile/tasks.py
@@ -25,6 +25,15 @@ from kolibri.core.utils.urls import reverse_remote
 from kolibri.utils.translation import ugettext as _
 
 
+def set_superuser(user_id):
+    new_superuser = FacilityUser.objects.get(id=user_id)
+    # make the user a new super user for this device:
+    new_superuser.facility.add_role(new_superuser, role_kinds.ADMIN)
+    DevicePermissions.objects.create(
+        user=new_superuser, is_superuser=True, can_manage_content=True
+    )
+
+
 class MergeUserValidator(PeerImportSingleSyncJobValidator):
     local_user_id = serializers.PrimaryKeyRelatedField(
         queryset=FacilityUser.objects.all()
@@ -140,12 +149,7 @@ def mergeuser(command, **kwargs):
 
     new_superuser_id = kwargs.get("new_superuser_id")
     if new_superuser_id:
-        new_superuser = FacilityUser.objects.get(id=new_superuser_id)
-        # make the user a new super user for this device:
-        new_superuser.facility.add_role(new_superuser, role_kinds.ADMIN)
-        DevicePermissions.objects.create(
-            user=new_superuser, is_superuser=True, can_manage_content=True
-        )
+        set_superuser(new_superuser_id)
 
     # create token to validate user in the new facility
     # after it's deleted in the current facility:
@@ -157,14 +161,10 @@ def mergeuser(command, **kwargs):
     job.save_meta()
     job.update_progress(1.0, 1.0)
 
-    # check if current user should be promoted to superuser:
+    # check if current user should be set as superuser:
     set_as_super_user = kwargs.get("set_as_super_user")
     if set_as_super_user:
-        # set current user as the new superuser of this device:
-        remote_user.facility.add_role(remote_user, role_kinds.ADMIN)
-        DevicePermissions.objects.create(
-            user=remote_user, is_superuser=True, can_manage_content=True
-        )
+        set_superuser(remote_user.id)
         delete_facility(local_user.facility)
         set_device_settings(default_facility=remote_user.facility)
     else:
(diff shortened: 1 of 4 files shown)


kolibri/plugins/user_profile/tasks.py
```diff
@@ -152,4 +156,16 @@ def mergeuser(command, **kwargs):
     job.extra_metadata["remote_user_pk"] = remote_user_pk
     job.save_meta()
     job.update_progress(1.0, 1.0)
-    local_user.delete()
+
+    # check if current user should be promoted to superuser:
+    set_as_super_user = kwargs.get("set_as_super_user")
+    if set_as_super_user:
+        # set current user as the new superuser of this device:
+        remote_user.facility.add_role(remote_user, role_kinds.ADMIN)
```
jredrejo: I'd add an inner function for these two lines as they are used also if `new_superuser_id` exists. Not for the sake of DRY as they're just two lines, but to be sure mainteinance is simple in the future if changes need to be added when an user is promoted to superuser

jredrejo (approved):
Tested and it works fine. Code looks good for me. I've added a comment for a slight improvement but not too important.

kolibri/plugins/user_profile/tasks.py
```diff
@@ -24,6 +25,15 @@
 from kolibri.utils.translation import ugettext as _
 
 
+def set_superuser(user_id):
+    new_superuser = FacilityUser.objects.get(id=user_id)
+    # make the user a new super user for this device:
+    new_superuser.facility.add_role(new_superuser, role_kinds.ADMIN)
```
rtibbles: This user is brought in using a learner only sync - this means two things, they shouldn't be made a facility admin, and even when we are doing so here, it will not sync anywhere.

We should be setting them as a superuser as below, but not a facility admin.

rtibbles (changes requested):
Shouldn't be promoting users to facility admin, only device admin permissions.

kolibri/plugins/user_profile/tasks.py
jredrejo: ah, none of the facility menus appeared after the user logins so I didn't noticed that, but it makes sense. In that case, @LianaHarris360 you can discard my previous comment, it would not make sense create a function for just one line of code.

LianaHarris360: I updated the code so that the migrated user is a superuser and not specifically a facility admin. I am able to log in/out and see the side navigation normally. But once the server is restarted and I log back in, I have the same problem as @jredrejo, the menu items disappear. However, this issue seems to be unrelated to the code changes I've made because this also happens on the `develop` branch. I can go ahead and open a separate issue for this.

Screenshot 2023-04-13 at 9 26 37 AM

rtibbles: Yeah, this is not caused by your changes, definitely, @LianaHarris360 - it's either a caching issue where the side nav items are not being properly deregistered, or it's an issue with 'contamination' where plugins are being unsafely referenced from outside the plugins themselves.

rtibbles (approved):

PR #10463, opened 2023-04-12
Sync cleaning all

AllanOXDi:
## Summary

This PR is following a review comments on #10259

## References

#10259

…

## Testing checklist

## PR process

## Reviewer checklist

- PR is fully functional
- Documentation is updated


kolibri/plugins/facility/assets/src/views/DataPage/SyncInterface/index.vue
```diff
@@ -130,7 +130,7 @@
   import CoreMenu from 'kolibri.coreVue.components.CoreMenu';
   import CoreMenuOption from 'kolibri.coreVue.components.CoreMenuOption';
   import { TaskStatuses } from 'kolibri.utils.syncTaskUtils';
-  import { PageNames } from './../../../constants';
+  import { PageNames } from '../../../../../../device/assets/src/constants';
```
rtibbles: Better to import `SyncPageNames` here like you do above, I think?

```
import { SyncPageNames } from 'kolibri-common/components/SyncSchedule/constants';
```

kolibri/plugins/facility/assets/src/views/DataPage/SyncInterface/index.vue
```diff
@@ -173,7 +173,10 @@
     },
     methods: {
       managesync() {
-        this.$router.push({ name: PageNames.MANAGE_SYNC_SCHEDULE });
+        this.$router.push({
+          name: PageNames.MANAGE_SYNC_SCHEDULE,
```
rtibbles: Can change this to `SyncPageNames.MANAGE_SYNC_SCHEDULE`.

kolibri/plugins/facility/assets/src/routes.js
```diff
@@ -95,6 +96,9 @@ export default [
     name: PageNames.DATA_EXPORT_PAGE,
     component: DataPage,
     path: '/:facility_id?/data',
+    props: route => {
+      return { isFromFacility: route.params.isFromFacility };
```
rtibbles: `isFromFacility` is not a parameter on the route, so this will always evaluate to `undefined`.

kolibri/plugins/device/assets/src/modules/pluginModule.js
```diff
@@ -9,6 +9,7 @@ export default {
   state() {
     return {
       authenticateWithPin: false,
+      isDevicePlugin: false,
```
rtibbles: Should clean up the changes in here that are no longer required.

kolibri/plugins/facility/assets/src/modules/pluginModule.js
```diff
@@ -48,7 +48,7 @@ export default {
       // parameter if we're in a multi-facility situation
       const params = {};
       if (getters.userIsMultiFacilityAdmin) {
-        params.facility_id = getters.activeFacilityId;
+        params.facilityId = getters.activeFacilityId;
```
rtibbles: The param name is `facility_id` so this change shouldn't be made still.

packages/kolibri-common/components/SyncSchedule/ManageSyncSchedule.vue
```diff
@@ -211,22 +224,22 @@
     },
     computed: {
       goBack() {
```
rtibbles: Can also just directly reference `this.goBackRoute` where this is used in the code.

rtibbles (commented):
Still a little bit of cleanup needed, but this is close to merge.

kolibri/plugins/facility/assets/src/routes.js
```diff
@@ -128,12 +129,23 @@ export default [
   },
   {
     path: '/:facility_id?/managesync',
+    props: route => {
+      return {
+        facilityId: route.params.facilityId || store.getters.currentFacilityId,
```
rtibbles:
```suggestion
        facilityId: route.params.facility_id || store.getters.currentFacilityId,
```

kolibri/plugins/facility/assets/src/routes.js
```diff
@@ -128,12 +129,23 @@ export default [
...
   {
-    path: '/:facility_id?/editdevice/:deviceId',
+    path: '/:facility_id?/editdevice/:deviceId/',
     component: editDeviceSyncSchedule,
-    name: PageNames.EDIT_SYNC_SCHEDULE,
+    name: SyncPageNames.EDIT_SYNC_SCHEDULE,
+    props: route => {
+      return { facilityId: route.params.facilityId || store.getters.currentFacilityId };
+      // return { device_id: route.params.device_id || store.getters.currentDeviceId,
+      //     facility_id:store.getters.currentFacilityId };
```
rtibbles:
```suggestion
      return { facilityId: route.params.facility_id || store.getters.currentFacilityId };
```

kolibri/plugins/facility/assets/src/views/DataPage/SyncInterface/index.vue
```diff
@@ -173,7 +173,10 @@
     },
     methods: {
       managesync() {
-        this.$router.push({ name: PageNames.MANAGE_SYNC_SCHEDULE });
+        this.$router.push({
+          name: SyncPageNames.MANAGE_SYNC_SCHEDULE,
+          params: { deviceId: this.$store.getters.activeFacilityId },
```
rtibbles:
```suggestion
          params: { facility_id: this.$store.getters.activeFacilityId },
```

packages/kolibri-common/components/SyncSchedule/EditDeviceSyncSchedule.vue
```diff
@@ -152,8 +151,8 @@
   import { now } from 'kolibri.utils.serverClock';
   import commonCoreStrings from 'kolibri.coreVue.mixins.commonCoreStrings';
   import { TaskTypes } from 'kolibri.utils.syncTaskUtils';
-  import { PageNames } from '../../../../kolibri/plugins/facility/assets/src/constants';
-
+  import { SyncPageNames } from './constants';
+  // ... from 'kolibri-common/Components/SyncSchedule/constants'
```
rtibbles:
```suggestion
```

rtibbles (commented):
I think that there is still a bit of an issue with separation of concerns here.

Currently, because the device plugin routes use `facilityId` as the parameter name for the facility in the URL, this is working - but the facility plugin routes are using `facility_id` instead.

This is why I recommended passing in `facilityId` as a prop to the components, so that these differences didn't matter, but this code is currently in a half way house between the two.

kolibri/plugins/device/assets/src/views/FacilitiesPage/index.vue
```diff
@@ -290,7 +292,10 @@
         } else if (option === Options.REGISTER) {
           this.facilityForRegister = facility;
         } else if (option === Options.MANAGESYNC) {
-          this.$router.push({ path: '/manage' });
+          this.$router.push({
+            name: SyncPageNames.MANAGE_SYNC_SCHEDULE,
+            params: { facility_id: facility.id },
```
akolson: It looks like this should be `facilityId` and not `facility_id` as device plugin uses the former. Please confirm before making the change.

kolibri/plugins/facility/assets/src/views/DataPage/SyncInterface/index.vue
akolson: As indicated by Richard, this should be `facility_id` and not `deviceId`

packages/kolibri-common/components/SyncSchedule/ManageSyncSchedule.vue
```diff
@@ -210,23 +223,20 @@
       };
     },
     computed: {
-      goBack() {
-        return { name: PageNames.DATA_EXPORT_PAGE };
+      theFacilityId() {
```
akolson: The `theFacilityId()` computed property is probably not necessary at this point as you already have a prop `facilityId` that seems to have fail-safe values whenever passed into this component(See route files for device and facility plugins)

packages/kolibri-common/components/SyncSchedule/constants.js
```diff
@@ -0,0 +1,8 @@
+export const SyncPageNames = {
+  MANAGE_SYNC_SCHEDULE: 'MANAGE_SYNC_SCHEDULE',
+  EDIT_SYNC_SCHEDULE: 'EDIT_SYNC_SCHEDULE',
+};
+
+// These are copied from (facility | device)/../constants.js' PageNames
+export const FacilityReturnPage = 'DATA_EXPORT_PAGE';
+export const DeviceReturnPage = 'FACILITIES_PAGE';
```
akolson: It doesn't look like you are using Line 7 and 8 anywhere in your code. It's probably better to get rid of it.

akolson (changes requested):
Thanks @AllanOXDi! Code is looking good overall. There is just a few more issues to sort and we should be good. Also, cross-checking with @rtibbles whether the issues he raised have been handled as well.

kolibri/core/assets/src/views/sync/FacilityNameAndSyncStatus/index.vue
```diff
@@ -111,6 +112,12 @@
         }
         return this.$formatRelative(datetime, { now: this.now });
       },
+      manageSync() {
+        this.$router.push({
+          name: SyncPageNames.MANAGE_SYNC_SCHEDULE,
+          params: { facilityId: this.facility.id },
```
rtibbles: This is still going to cause issues, because this component is being used in the device plugin and the facility plugin. The facility plugin uses `facility_id` as a route parameter, and the device plugin is currently using `facilityId` as a route parameter. As long as they use a different parameter, this is going to cause issues.

For the sake of moving forward, I think it might be simplest, just to standardize the new device plugin routes you have created to use `facility_id` as the route parameter for facility, so that we can set this consistently.

Ideally, this is why this shouldn't be handled internally to the component, so that this standalone component doesn't need to 'know' about the details of the route implementations in the different plugins, but we can open follow up issues to handle this as technical debt.

kolibri/plugins/device/assets/src/routes/index.js
```diff
@@ -75,6 +78,31 @@ const routes = [
...
     },
   },
+  {
+    name: SyncPageNames.MANAGE_SYNC_SCHEDULE,
+    component: withAuthMessage(ManageSyncSchedule, 'superuser'),
+    path: '/facilities/:facilityId/managesync',
+    props: route => {
+      return {
+        goBackRoute: { name: PageNames.FACILITIES_PAGE },
+        facilityId: route.params.facilityId || store.getters.currentFacilityId,
```
rtibbles: See my note above about standardizing `facility_id` as the route parameter.

kolibri/plugins/device/assets/src/views/FacilitiesPage/index.vue
rtibbles: @akolson see my note above, as we still have code that is referencing the route parameters within both plugins, we might be better off just using `facility_id` across both plugins so that it works, even if it's not the cleanest separation of concerns.

kolibri/plugins/facility/assets/src/views/DataPage/SyncInterface/index.vue
```diff
@@ -173,7 +173,10 @@
     },
     methods: {
       managesync() {
-        this.$router.push({ name: PageNames.MANAGE_SYNC_SCHEDULE });
+        this.$router.push({
+          name: SyncPageNames.MANAGE_SYNC_SCHEDULE,
+          params: { facilityId: this.facility.id },
```
rtibbles: Within the facility plugin, the route parameter is `facility_id`. I think we should just standardize on that for the route parameters to avoid this ongoing confusion.

packages/kolibri-common/components/SyncSchedule/EditDeviceSyncSchedule.vue
```diff
@@ -270,40 +275,44 @@
...
-        );
+            .catch(() => {
+              this.createTaskFailedSnackbar();
+            });
+        });
       },
 
       cancelBtn() {
-        this.$router.push({ name: PageNames.ManageSyncSchedule });
+        this.$router.push({ name: SyncPageNames.MANAGE_SYNC_SCHEDULE });
```
rtibbles: We should mix in the `facility_id` parameter from the current route here, or this will break on the facility plugin when we are editing sync schedules for the non-default facility, and all the time for the device plugin where the facility_id is required.

rtibbles (changes requested):
The route parameters are still causing confusion here, so I think it is best to change all the device plugin routes to use `facility_id` for the facility id parameter in the route.

akolson (commented):
I think we should be good with this, @rtibbles any additional thoughts?

kolibri/plugins/device/assets/src/routes/index.js
```diff
@@ -75,6 +77,42 @@ const routes = [
...
+    },
+  },
+  {
+    name: PageNames.EDIT_SYNC_SCHEDULE,
+    component: withAuthMessage(EditDeviceSyncSchedule, 'superuser'),
+    path: '/facilities/:device_id/:facilityId/editdevice',
+    props: route => {
+      return {
+        goBackRoute: { name: PageNames.MANAGE_SYNC_SCHEDULE },
+        facilityId: route.params.facilityId || store.getters.currentFacilityId,
```
rtibbles, line 108: This does not need to be changed, but just a comment that as the facilityId is a required parameter, I am not sure the `||` is needed here.

rtibbles (approved):
Excellent, I see no more confusions between `facility_id` and `facilityId` and it seems like it does not rely on shared URL parameters either between the plugins.

kolibri/plugins/device/assets/src/routes/index.js
akolson: I agree. cc @AllanOXDi

akolson (approved):
LGTM. Thanks @AllanOXDi

